PR #16928, opened 2025-06-19
chore: add node synced helper

### LINKED ISSUE #16911: Add helper Future type for getting notified when the node synced
### Describe the feature

ref https://github.com/paradigmxyz/reth/discussions/16906

we can convert this logic:

https://github.com/paradigmxyz/reth/blob/a86e18fa1bdfaa94982cd82ec75fcce6d3596834/crates/rpc/rpc/src/eth/pubsub.rs#L178-L198

into a helper future/stream type that resolves once the node no longer syncs, this could technically also use a tick rate of 12s or so instead of using the subscription stream, so this type could be something like

```rust
struct SyncListener<N> {
  tick: Pin<Box<dyn Stream<()>,
  network_info: N
}
```

### Additional context

_No response_

shane-moore:
Closes https://github.com/paradigmxyz/reth/issues/16911

### DIFF
--- a/crates/rpc/rpc/src/eth/helpers/sync_listener.rs
+++ b/crates/rpc/rpc/src/eth/helpers/sync_listener.rs
@@ -6,10 +6,10 @@ use reth_network_api::NetworkInfo;
 use std::{
     future::Future,
     pin::Pin,
-    task::{Context, Poll},
+    task::{ready, Context, Poll},
 };
 
-/// Future that resolves once the node is no longer syncing.
+/// This future resolves once the node is no longer syncing: [`NetworkInfo::is_syncing`].
 #[must_use = "futures do nothing unless polled"]
 #[pin_project]
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


crates/rpc/rpc/src/eth/helpers/sync_listener.rs
```diff
@@ -0,0 +1,130 @@
...
+            match tick_event {
+                Some(_) => {
+                    if !this.network_info.is_syncing() {
+                        return Poll::Ready(());
+                    }
+                }
+                None => return Poll::Ready(()),
+            }
+        }
+        Poll::Pending
```
mattsse: I think you can make this slightly more ergonomic with

```
loop {
  match ready!(this.tick.as_mut().poll_next(cx))
}

File: crates/rpc/rpc/src/eth/helpers/sync_listener.rs [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+use futures::Stream;
+use pin_project::pin_project;
+use reth_network_api::NetworkInfo;
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// Future that resolves once the node is no longer syncing.
```
mattsse: let's add some more context here, 

This future resolves once the node is no longer syncing: [`NetworkInfo::is_syncing`].

crates/rpc/rpc/src/eth/helpers/sync_listener.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Create a new [`SyncListener`] using the given tick stream.
+    pub const fn new(network_info: N, tick: St) -> Self {
+        Self { tick, network_info }
+    }
+}
+
+impl<N, St> Future for SyncListener<N, St>
+where
+    N: NetworkInfo,
+    St: Stream<Item = ()> + Unpin,
```
mattsse: I think we could even relax the Item type here with an additional generic impl<N,St,Out> ?

crates/rpc/rpc/src/eth/pubsub.rs
mattsse, line 1: can we keep this unchanged for this pr?

mattsse (changes requested):
great, some nits+suggestions

mattsse (approved):
sweet

Issue #16876, opened 2025-06-18
Move `bytecode_by_hash` from `StateProvider` to a dedicated `BytecodeReader`


hai-rise:
### Describe the feature

This would allow us to relax the state provider for mempool transaction validators from a full `StateProvider` to just `AccountReader + BytecodeReader`.

https://github.com/paradigmxyz/reth/blob/55134742d693cebbee1d1cbe19366458a36b3c14/crates/transaction-pool/src/validate/eth.rs#L484-L502

Then, a custom mempool transaction validator (implementing `TransactionValidator`) with a custom cache layer (from shreds for us) wouldn't need to implement verbose, error-prone unreachable traits like `StateRootProvider`, `StorageRootProvider`, etc.

We can work on this if it's indeed desirable 🙏.

nadtech-hub:
I may take it up as well

emhane:
wdym `StateRootProvider` and `StorageRootProvider` are unreachable? :eyes:

mattsse:
imo this makes sense, with 7097 we likely need some additional bytecode related things

hai-rise:
> wdym `StateRootProvider` and `StorageRootProvider` are unreachable? 👀

I meant for a custom storage provider for a custom transaction validator. It should only provide `basic_account` and `bytecode_by_hash` to validate transactions, but the current requirement is a full `StateProvider`.

Currently:

```rust
impl StateProvider for CustomTxValidatorProvider { ... }

impl AccountReader for CustomTxValidatorProvider { ... }

// All methods below are not used, but we still need to 
// implement to satisfy `P: StateProvider` in `validate_one_against_state`
impl BlockHashReader for CustomTxValidatorProvider { ... }

impl StateRootProvider for CustomTxValidatorProvider { ... }

impl HashedPostStateProvider for CustomTxValidatorProvider { ... }

impl StorageRootProvider for CustomTxValidatorProvider { ... }

impl StateProofProvider for CustomTxValidatorProvider { ... }
```

Ideally:
```rust
impl AccountReader for CustomTxValidatorProvider { ... }

impl BytecodeReader for CustomTxValidatorProvider { ... }
```

In our case, we listen to preconfs and increase the balance of newly funded accounts, so their transactions can be accepted into the mempool before the block with the funding transaction is canonicalized.

mattsse:
yep makes perfect sense, maybe @Soubhik-10 wants to take this

emhane:
[...]

PR #16929, opened 2025-06-19
chore(tx-validation): remove redundant validate methods

hai-rise:
Random refactoring towards #16876, but this removes redundant `validate_*` methods by inlining used-only-once(s) and thin forwarding(s).


mattsse (approved):
these are indeed not needed anymore because they just shadow trait functions

PR #16930, opened 2025-06-19
feat: require only account & bytecode reader for tx validation

hai-rise:
Last piece for #16876.

### DIFF
--- a/crates/storage/storage-api/src/state.rs
+++ b/crates/storage/storage-api/src/state.rs
@@ -92,6 +92,10 @@ pub trait StateProvider:
     }
 }
 
+/// Minimal requirements to read a full account, for example, to validate its new transactions
+pub trait AccountInfoReader: AccountReader + BytecodeReader {}
+impl<T: AccountReader + BytecodeReader> AccountInfoReader for T {}
+
 /// Trait implemented for database providers that can provide the [`reth_trie_db::StateCommitment`]
 /// type.
 #[cfg(feature = "db-api")]
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -28,7 +28,7 @@ use reth_primitives_traits::{
     constants::MAX_TX_GAS_LIMIT_OSAKA, transaction::error::InvalidTransactionError, Block,
     GotExpected, SealedBlock,
 };
-use reth_storage_api::{StateProvider, StateProviderFactory};
+use reth_storage_api::{AccountInfoReader, StateProviderFactory};
 use reth_tasks::TaskSpawner;
 use std::{
     marker::PhantomData,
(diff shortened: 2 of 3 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -241,7 +241,7 @@ where
                 if maybe_state.is_none() {
                     match self.client.latest() {
                         Ok(new_state) => {
-                            *maybe_state = Some(new_state);
+                            *maybe_state = Some(Box::new(new_state));
```
hai-rise, line 244: We cannot cast `Box<dyn StateProvider>` directly to `Box<dyn TxValidatorProvider>`, so we wrap it in another `Box` to treat `Box<dyn StateProvider>` as a `dyn TxValidatorProvider` instead.

crates/storage/storage-api/src/state.rs
```diff
@@ -92,6 +92,10 @@ pub trait StateProvider:
     }
 }
 
+/// Minimal provider requirements to validate mempool transactions
+pub trait TxValidatorProvider: AccountReader + BytecodeReader {}
```
mattsse: hmm, this now uses the current usecase of the trait as the name, 
I think this is like `AccountInfoReader` because all the functions add up to accountinfo type which is nonce,balance,code

crates/transaction-pool/src/validate/eth.rs
mattsse: this seems okay, the only other way we could make this work is `fn as_account -> dyn` or use the upcast_dyn crate but this seems okay, this will then be boxbox

crates/transaction-pool/src/validate/eth.rs
```diff
@@ -241,7 +241,7 @@ where
                 if maybe_state.is_none() {
                     match self.client.latest() {
```
mattsse, line 242: the txvalidator will still use StateProvider internal, so I'm not entirely sure I fully get your use case.

are you always using `Some` with a special provider?

mattsse (changes requested):
naming nit

crates/transaction-pool/src/validate/eth.rs
hai-rise: Yeah, we always have a `Some` from our custom cache. We actually use `.pending()` as the fallback provider, but overall, this `self.client.latest()` is still a good default for the `None` case (if we fail to create the custom provider, there's no extra cache in the first place, etc.).

Basically calling `OpTransactionValidator::validate_one_with_state(origin, tx, &mut my_custom_provider)` in our custom `TransactionValidator` implementation.

crates/storage/storage-api/src/state.rs
hai-rise: Fixed, sir :pray:.

crates/transaction-pool/src/validate/eth.rs
mattsse: if this makes it easier, then this seems fine

mattsse (approved):

crates/transaction-pool/src/validate/eth.rs
hai-rise: Yes, it does allow us to pass in `Some(Box::new(impl AccountInfoReader))` instead of a full `impl StateProvider`!

PR #16931, opened 2025-06-19
chore: Expose payload_id

SozinM:
Expose payload_id so it's possible to use this function in id calculation outside of client.
In our use case we handle FCU on proxy level and need to determine future payload_id before client answered to FCU


mattsse (approved):
makes sense

PR #16934, opened 2025-06-19
feat(alloy-provider): implement fetch block

cakevm:
Add the `block(...)`, `block_by_id(...)` implementation for the `alloy-provider`. 

Thanks to 🤖 [Claude Code](https://claude.ai/code) for support

### DIFF
--- a/crates/alloy-provider/src/lib.rs
+++ b/crates/alloy-provider/src/lib.rs
@@ -436,8 +436,11 @@ where
     Node: NodeTypes,
     BlockTy<Node>: TryFromBlockResponse<N>,
 {
-    fn block_by_id(&self, _id: BlockId) -> ProviderResult<Option<Self::Block>> {
-        Err(ProviderError::UnsupportedProvider)
+    fn block_by_id(&self, id: BlockId) -> ProviderResult<Option<Self::Block>> {
+        match id {
+            BlockId::Number(number_or_tag) => self.block_by_number_or_tag(number_or_tag),
+            BlockId::Hash(hash) => self.block_by_hash(hash.block_hash),
+        }
     }
 
     fn sealed_header_by_id(
--- a/crates/alloy-provider/src/lib.rs
+++ b/crates/alloy-provider/src/lib.rs
@@ -359,18 +359,9 @@ where
     }
 
     fn block(&self, id: BlockHashOrNumber) -> ProviderResult<Option<Self::Block>> {
-        let block_response = match id {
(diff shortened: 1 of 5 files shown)


crates/primitives-traits/src/rpc.rs
```diff
@@ -0,0 +1,12 @@
+//! RPC conversion traits for node types.
+
+use crate::NodePrimitives;
+
+/// Trait for converting RPC blocks to primitive blocks for a specific node type.
+pub trait RpcBlockConversion: NodePrimitives {
+    /// The RPC block type.
+    type RpcBlock: serde::Serialize + serde::de::DeserializeOwned + 'static;
```
mattsse: instead of extending NodePrimitives

I'd like to try something else, which is kinda the inverse of

https://github.com/paradigmxyz/reth/blob/54cd8b34a40d1c22c5c00bb49ec4eb855ae77690/crates/rpc/rpc-types-compat/src/transaction.rs#L95-L95

something like

```
pub trait TryFromBlockResponse<N: alloy_network::Network> {
    type Error: Error + Send + Sync + Unpin;
    
    fn from_block_response(N::BlockResponse) -> Result<Self, Self::Error>;
}
```

mattsse (changes requested):

crates/primitives-traits/src/rpc.rs
cakevm: Thanks @mattsse, I took your implementation 1:1 https://github.com/paradigmxyz/reth/compare/matt/prototype-conversion. It works like a charm. I currently have no app for optimism, and did not moved forward to test it.

crates/alloy-provider/src/lib.rs
```diff
@@ -356,8 +358,30 @@ where
...
+                self.provider.get_block_by_hash(hash).full().await.map_err(ProviderError::other)
+            }),
+            BlockHashOrNumber::Number(num) => self.block_on_async(async {
+                self.provider
+                    .get_block_by_number(num.into())
+                    .full()
+                    .await
+                    .map_err(ProviderError::other)
+            }),
+        }?;
```
mattsse: this should be equivalent to `get_block(id.into())`

mattsse (changes requested):

crates/alloy-provider/src/lib.rs
cakevm: Cool!, much simpler.

mattsse (approved):

PR #16938, opened 2025-06-19
perf(trie): box subtries in parallel sparse trie

shekhirin:
Storing 256 `SparseSubtrie` items on the stack isn't a good idea


mediocregopher (approved):
SGTM, I'll need to update my PR then to set the `branch_node_tree/hash_masks`, `reveal_node` is supposed to be doing that I believe

PR #16943, opened 2025-06-19
perf(trie): `SparseSubtrie::update_hashes`

shekhirin:
Towards https://github.com/paradigmxyz/reth/issues/16866

Implements the `SparseSubtrie::update_hashes` method for calculating the hashes of a subtrie. Will be called in parallel from `ParallelSparseTrie::update_subtrie_hashes`.


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -565,10 +569,329 @@ impl SparseSubtrie {
...
+                                // SAFETY: it's a child, so it's never empty
+                                let last_child_nibble = child_path.last().unwrap();
+
+                                // Determine whether we need to set trie mask bit.
+                                let should_set_tree_mask_bit = if let Some(store_in_db_trie) =
+                                    child_node_type.store_in_db_trie()
+                                {
+                                    // A branch or an extension node explicitly set the
+                                    // `store_in_db_trie` flag
+                                    store_in_db_trie
```
mediocregopher, line 753: Is this right? Is this not the same bug as in HashBuilder, where we're setting tree_mask bit on extension node children?

shekhirin: yep, the behaviour in the sparse trie was just duplicated from the hash builder, but I didn't dare to change it because wasn't sure it won't break anything 😅 I think we should change it everywhere at the same time when we do it in the hash builder.

mediocregopher (approved):
Not much changed in the main logic since last time I looked, LGTM :+1:

Rjected (approved):
lgtm, makes sense to have this on sparsesubtrie + bring over run_hash_builder

PR #16944, opened 2025-06-19
perf: Reuse CachedPrecompileMetrics across block executions

mediocregopher:
This saves some time allocating the metrics on every instantiation of a CachedPrecompile.

Fixes #16860


mattsse (approved):
cool,
this looks complete already

@Rjected @shekhirin not recording Hit/miss during prewarming is better because this way we get accurate metrics for regular block exec

Rjected (approved):
yep makes sense to me

PR #16945, opened 2025-06-19
bench(trie): prepare trie outside of routine, use large input size

shekhirin:
`SparseTrie` is not so large that it requires `PerIteration`, so let's use `LargeInput`. Also, prepare the trie outside of routine, and only prepare a list of leaves to update/remove inside the setup closure.


Rjected (approved):
makes sense

PR #16956, opened 2025-06-19
feat(trie): add ParallelSparseTrie::update_leaf

### LINKED ISSUE #16868: Implement `update_leaf` for `ParallelSparseTrie`
### Describe the feature

We need an `update_leaf` method similar to the one on the regular `SparseTrie`, this is mainly used for tests and benchmarks which we would want to port over to the parallel sparse trie:
https://github.com/paradigmxyz/reth/blob/051cef53bc7372c22d545a583427b0e598529e87/crates/trie/sparse/src/trie.rs#L1586-L1717

### Additional context

_No response_

Rjected:
Implements `update_leaf` for `ParallelSparseTrie`. This is done by implementing a method `next_node` which traverses and updates the trie one step at a time towards the path being updated. This is done for the first two levels of the trie. When this causes a node to be created that should be in a lower subtrie, it's moved to the lower subtrie. This also updates and creates lower subtries as necessary.

closes https://github.com/paradigmxyz/reth/issues/16868


Rjected:
Update: this is currently failing one of the tests I've added, still figuring out why

Rjected:
This now passes all of the tests added and is ready for review

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -181,16 +225,110 @@ impl ParallelSparseTrie {
     /// provider returns an error.
     pub fn update_leaf(
         &mut self,
-        key_path: Nibbles,
+        path: Nibbles,
```
mediocregopher: For `remove_leaf` we differentiated this path as being a "full path", so it's clear that it's not a node path. Could we do that here too?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -841,6 +980,198 @@ impl SparseSubtrie {
...
+                if &current == path {
+                    unreachable!("we already checked leaf presence in the beginning");
+                }
+
+                // find the common prefix
+                let common = current.common_prefix_length(path);
+
+                // update existing node
+                let new_ext_key = current.slice(current.len() - current_key.len()..common);
+                *node = SparseNode::new_ext(new_ext_key);
```
mediocregopher, line 1066: If there was an existing leaf at the root (`0x`) with a key of `0x123` and you were trying to update the value `0x456`, I think you'd end up with an extension with an empty key here; at this point `current_key` would be `0x123`, `current` would also be `0x123` because it was empty but then extended with `current_key`, `common` would be zero, and so `new_ext_key` would be `current[0..0]`.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -841,6 +980,198 @@ impl SparseSubtrie {
         current_level == child_level
     }
 
+    /// Updates or inserts a leaf node at the specified key path with the provided RLP-encoded
+    /// value.
+    ///
+    /// This method updates the internal prefix set and, if the leaf did not previously exist,
```
mediocregopher: > updates the internal prefix set

Where is this happening? The only prefix set updates I see are in the `ParallelSparseTrie::update_leaf` call.

Actually I think maybe this comment might just be out of date, it also mentions returning the split branch

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -92,6 +92,50 @@ impl ParallelSparseTrie {
...
+    /// at 0x1234, 0x1235, and 0x1236:
+    /// - Incorrect: Create subtrie with path 0x12. subtrie.update_leaf would not be able to
+    /// traverse here because it starts at the subtrie path.
+    /// - Correct: Create subtrie with path 0x123. Here subtrie.update_leaf would start at 0x123,
+    /// correctly find the branch node, and proceed normally.
+    ///
+    /// The method also handles path updates: if a subtrie exists with path 0x12345 and we
+    /// insert a node at 0x1234, for example as a result of shortening the extension node and
+    /// creating a branch node, the subtrie path is updated to 0x1234.
+    fn get_or_create_lower_subtrie(
```
mediocregopher: So this isn't something I accounted for in either `reveal_node` or `remove_leaf`. I will need to update `remove_leaf` accordingly, but I think if you incorporated this logic into the existing `lower_subtrie_for_path` and `subtrie_for_path` methods it would fix `reveal_node` without any other changes being necessary, and there wouldn't need to be this extra method.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -841,6 +980,198 @@ impl SparseSubtrie {
...
+    ///
+    /// This will add or update any nodes in the trie as necessary.
+    ///
+    /// Returns a `LeafUpdateStep` containing the next node to process (if any) and
+    /// the paths of nodes that were inserted during this step.
+    fn next_node(
+        &mut self,
+        mut current: Nibbles,
+        path: &Nibbles,
+        provider: impl BlindedProvider,
```
mattsse, line 1033: this will be used for doing IO?

separately, I still find BlindedProvider slightly confusing because this or more like BlindedNodeProvider because BlindedProvider makes me think that the provider itself is blinded

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -181,16 +225,110 @@ impl ParallelSparseTrie {
...
+        // Traverse the upper subtrie to find the node to update or the subtrie to update.
+        //
+        // First we check the upper subtrie, and if we would update solely the upper subtrie, we
+        // call update_leaf on the upper subtrie.
+        //
+        // We stop when the next node to traverse would be in a lower subtrie, or if there are no
+        // more nodes to traverse.
+        while next.is_some_and(|next| next.len() < UPPER_TRIE_MAX_DEPTH) {
+            // checked this above, so unwrap is safe
+            let current = next.unwrap();
```
```suggestion
        while let Some(current) = next.filter(|next| next.len() < UPPER_TRIE_MAX_DEPTH) {
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -181,16 +225,110 @@ impl ParallelSparseTrie {
...
+            let subtrie_idx = path_subtrie_index_unchecked(node_path);
+            let node = self.upper_subtrie.nodes.remove(node_path).unwrap();
+
+            // If it's a leaf node, extract its value before getting mutable reference to subtrie
+            let leaf_value = if let SparseNode::Leaf { key, .. } = &node {
+                let mut leaf_full_path = *node_path;
+                leaf_full_path.extend(key);
+                Some((
+                    leaf_full_path,
+                    self.upper_subtrie.inner.values.remove(&leaf_full_path).unwrap(),
```
shekhirin: how often do we expect to have this branch hit? If often, I would try to avoid inserting into `self.upper_subtrie.inner.values` in the beginning of the function only to remove it here.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -841,6 +980,198 @@ impl SparseSubtrie {
...
+        };
+        match node {
+            SparseNode::Empty => {
+                // We need to insert the node with a different path and key depending on the path of
+                // the subtrie.
+                let path = path.slice(self.path.len()..);
+                *node = SparseNode::new_leaf(path);
+                Ok(LeafUpdateStep::complete_with_insertions(vec![current]))
+            }
+            &mut SparseNode::Hash(hash) => {
```
shekhirin: do we need `&mut` here?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -841,6 +980,198 @@ impl SparseSubtrie {
...
+                Err(SparseTrieErrorKind::BlindedNode { path: current, hash }.into())
+            }
+            SparseNode::Leaf { key: current_key, .. } => {
+                current.extend(current_key);
+
+                // this leaf is being updated
+                if &current == path {
+                    unreachable!("we already checked leaf presence in the beginning");
+                }
```
shekhirin: should this be a debug assertion?

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: This is basically what I'm thinking, for reference: https://github.com/paradigmxyz/reth/commit/fb9f1570495d237d0f2d8543a029d6d619ee2694

I'm going to work on fixing remove_leaf in that branch, if @Rjected if you wanted to use those changes I could rebase off you later or vice-versa

crates/trie/sparse-parallel/src/trie.rs
shekhirin: this will do I/O, yes

agreed on naming

crates/trie/sparse-parallel/src/trie.rs
Rjected: yeah I just copy pasted the comment from update_leaf, will update

crates/trie/sparse-parallel/src/trie.rs
Rjected: I think we can merge that change probably sooner / independently, but makes sense, having one method would be nice

crates/trie/sparse-parallel/src/trie.rs
Rjected: we don't, removing

crates/trie/sparse-parallel/src/trie.rs
Rjected: Yes probably, tbh I am not 100% sure when we would hit this

crates/trie/sparse-parallel/src/trie.rs
Rjected: that's true but it would be overwritten by this
```rust
                self.nodes.insert(
                    branch_path,
                    SparseNode::new_split_branch(
                        current.get_unchecked(common),
                        path.get_unchecked(common),
                    ),
                );
```
`current` is 0x123, `path` is 0x456, common is zero, we would get:
* `branch_path = current.slice(..0)` which would be the empty path
* `new_leaf_path = path.slice(..=0)` which would be `0x4`
* `existing_leaf_path = current.slice(..=0)` which would be `0x1`

so I think extension node creation would get overwritten by insertion.

I think we test this case as well

crates/trie/sparse-parallel/src/trie.rs
Rjected: not often, but if we for example create two leaves at level 2 (say w a split branch at level 1) after updating the upper trie and need to move those leaves down to subtries, we'll need both values for these leaves, not just the one passed as an argument. so imo it's actually more convenient to add in values so that we have both if this happens

crates/trie/sparse-parallel/src/trie.rs
Rjected: changed to `debug_assert!`

Rjected: test

crates/trie/sparse-parallel/src/trie.rs
Rjected: How about `TrieNodeProvider`? or `RevealedNodeProvider`?

shekhirin: `TrieNodeProvider` makes sense to me, as it's in theory for any node, not just blinded ones. And then the method would be `TrieNodeProvider::node` or smth like that

shekhirin: https://github.com/paradigmxyz/reth/issues/17155

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I see, yeah it looks like `test_progressive_branch_creation` does indeed test it so it must work :+1:

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: https://github.com/paradigmxyz/reth/pull/17143 is merged so the updated `lower_subtrie_for_path` should be available for you to use here

Rjected: Using the new subtrie for path method now

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+        //
+        // We use `next` to keep track of the next node that we need to traverse to, and
+        // `inserted_nodes` to keep track of any nodes that were created during the traversal.
+        let mut new_nodes = Vec::new();
+        let mut next = Some(Nibbles::default());
+
+        // Traverse the upper subtrie to find the node to update or the subtrie to update.
+        //
+        // First we check the upper subtrie, and if we would update solely the upper subtrie, we
+        // call update_leaf on the upper subtrie.
```
shekhirin: is this comment outdated? I see that we're calling `update_leaf` only for lower subtries on line 281, and the upper subtrie is updated in `next_node`.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1467,6 +1748,41 @@ impl SparseSubtrieInner {
...
+        /// The next node path to process
+        next_node: Nibbles,
+    },
+    /// Update is complete with nodes inserted
+    Complete {
+        /// The node paths that were inserted during this step
+        inserted_nodes: Vec<Nibbles>,
+    },
+    /// No operation needed (node doesn't exist)
+    NoOp,
```
shekhirin: can we rename it to `NodeNotFound` or smth like that? `NoOp` makes me think that we don't need to do anything after calling `next_node`, but still can keep calling `next_node` until `Complete`.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+        }
+
+        // Start at the root, traversing until we find either the node to update or a subtrie to
+        // update.
+        //
+        // We first traverse the upper subtrie for two levels, and moving any created nodes to a
+        // lower subtrie if necessary.
+        //
+        // We use `next` to keep track of the next node that we need to traverse to, and
+        // `inserted_nodes` to keep track of any nodes that were created during the traversal.
```
mediocregopher:
```suggestion
        // `new_nodes` to keep track of any nodes that were created during the traversal.
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+        }
+
+        // Move nodes from upper subtrie to lower subtries
+        for node_path in &new_nodes {
+            // Skip nodes that belong in the upper subtrie
+            if SparseSubtrieType::path_len_is_upper(node_path.len()) {
+                continue
+            }
+
+            let node = self.upper_subtrie.nodes.remove(node_path).unwrap();
```
```suggestion
            let node = self.upper_subtrie.nodes.remove(node_path).expect("node belongs to upper subtrie");
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+
+            // If it's a leaf node, extract its value before getting mutable reference to subtrie
+            let leaf_value = if let SparseNode::Leaf { key, .. } = &node {
+                let mut leaf_full_path = *node_path;
+                leaf_full_path.extend(key);
+                Some((
+                    leaf_full_path,
+                    self.upper_subtrie.inner.values.remove(&leaf_full_path).unwrap(),
```
```suggestion
                    self.upper_subtrie.inner.values.remove(&leaf_full_path).expect("leaf nodes have associated values entries"),
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+                subtrie.inner.values.insert(leaf_full_path, value);
+            }
+
+            // Insert the node into the lower subtrie
+            subtrie.nodes.insert(*node_path, node);
+        }
+
+        // If we reached the max depth of the upper trie, we may have had more nodes to insert.
+        if let Some(next_path) = next.filter(|n| n.len() >= UPPER_TRIE_MAX_DEPTH) {
+            // Use get_or_create_lower_subtrie to ensure the subtrie has the correct path.
```
mediocregopher:
```suggestion
            // Use subtrie_for_path to ensure the subtrie has the correct path.
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -885,6 +974,198 @@ impl SparseSubtrie {
...
+        Ok(())
+    }
+
+    /// Processes the current node, returning what to do next in the leaf update process.
+    ///
+    /// This will add or update any nodes in the trie as necessary.
+    ///
+    /// Returns a `LeafUpdateStep` containing the next node to process (if any) and
+    /// the paths of nodes that were inserted during this step.
+    fn next_node(
```
shekhirin: just to be clear that it belongs to the `update` part of the sparse trie, and also that it modifies the structure, not just iterates over it
```suggestion
    fn update_next_node(
```

mediocregopher (approved):
Two minor comment fixes but otherwise LGTM

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1467,6 +1748,41 @@ impl SparseSubtrieInner {
...
+    pub const fn complete_with_insertions(inserted_nodes: Vec<Nibbles>) -> Self {
+        Self::Complete { inserted_nodes }
+    }
+}
+
+impl Default for LeafUpdateStep {
+    fn default() -> Self {
+        Self::NoOp
+    }
+}
```
shekhirin: think can be done with `#[default]` attribute on the enum variant

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
     /// provider returns an error.
     pub fn update_leaf(
         &mut self,
-        key_path: Nibbles,
+        full_path: Nibbles,
         value: Vec<u8>,
-        masks: TrieMasks,
-        provider: impl BlindedProvider,
+        provider: impl BlindedProvider + Clone,
```
```suggestion
        provider: impl BlindedProvider,
```

shekhirin (commented):
nits

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+        let mut next = Some(Nibbles::default());
+
+        // Traverse the upper subtrie to find the node to update or the subtrie to update.
+        //
+        // First we check the upper subtrie, and if we would update solely the upper subtrie, we
+        // call update_leaf on the upper subtrie.
+        //
+        // We stop when the next node to traverse would be in a lower subtrie, or if there are no
+        // more nodes to traverse.
+        while let Some(current) = next.filter(|next| next.len() < UPPER_TRIE_MAX_DEPTH) {
```
mediocregopher:
```suggestion
        while let Some(current) = next.filter(|next| SparseSubtrieType::path_len_is_upper(next.len())) {
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -184,16 +184,104 @@ impl ParallelSparseTrie {
...
+            if let Some((leaf_full_path, value)) = leaf_value {
+                subtrie.inner.values.insert(leaf_full_path, value);
+            }
+
+            // Insert the node into the lower subtrie
+            subtrie.nodes.insert(*node_path, node);
+        }
+
+        // If we reached the max depth of the upper trie, we may have had more nodes to insert.
+        if let Some(next_path) = next.filter(|n| n.len() >= UPPER_TRIE_MAX_DEPTH) {
```
mediocregopher:
```suggestion
        if let Some(next_path) = next.filter(|n| !SparseSubtrieType::path_len_is_upper(n.len())) {
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -885,6 +974,198 @@ impl SparseSubtrie {
...
+                    *key = current.slice(current.len() - key.len()..common);
+
+                    // If branch node updates retention is enabled, we need to query the
+                    // extension node child to later set the hash mask for a parent branch node
+                    // correctly.
+                    if self.inner.updates.is_some() {
+                        // Check if the extension node child is a hash that needs to be revealed
+                        if self.nodes.get(&current).unwrap().is_hash() {
+                            if let Some(RevealedNode { node, tree_mask, hash_mask }) =
+                                provider.blinded_node(&current)?
```
mattsse, line 1113: this will eventually do IO?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -885,6 +974,198 @@ impl SparseSubtrie {
...
+                    // find the common prefix
+                    let common = current.common_prefix_length(path);
+                    *key = current.slice(current.len() - key.len()..common);
+
+                    // If branch node updates retention is enabled, we need to query the
+                    // extension node child to later set the hash mask for a parent branch node
+                    // correctly.
+                    if self.inner.updates.is_some() {
+                        // Check if the extension node child is a hash that needs to be revealed
+                        if self.nodes.get(&current).unwrap().is_hash() {
```
mattsse: why does this hold?

not obvious from the code above this line why this is guaranteed to exist

mattsse (commented):
I could mostly follow along, the state enums helped.

only have one question

crates/trie/sparse-parallel/src/trie.rs
shekhirin: yes

crates/trie/sparse-parallel/src/trie.rs
shekhirin: this should have an `expect` call with a msg

it holds because when you iterate over the existing sparse trie, all children of extension and branch nodes should exist in the trie. So when you check a child of an extension node, it is guaranteed to exist, because the only places where we insert extension nodes are when we update existing nodes.

crates/trie/sparse-parallel/src/trie.rs
Rjected: yep makes sense

crates/trie/sparse-parallel/src/trie.rs
Rjected: yeah outdated comment

crates/trie/sparse-parallel/src/trie.rs
Rjected: yep makes sense, renamed to NodeNotFound

crates/trie/sparse-parallel/src/trie.rs
Rjected: Added an `expect`, with a message, lmk if it makes sense

mattsse (approved):

shekhirin (approved):

PR #16665, opened 2025-06-05
perf: remove some clones around `eth_call`

hai-rise:
Remove a few clones around `eth_call` and clean code accordingly.

PR #16958, opened 2025-06-20
perf(rpc): optimize EVM reuse in `eth_estimateGas`

thaodt:
This PR is inspired by [my mentor's comment](https://github.com/paradigmxyz/reth/pull/16665#discussion_r2127878002).

- Within a single `eth_estimateGas` call, the EVM instance is created once and reused throughout the binary search process
- Multiple `evm.transact()` calls reuse the same EVM instance.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/estimate.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/estimate.rs
@@ -7,7 +7,7 @@ use alloy_rpc_types_eth::{state::StateOverride, transaction::TransactionRequest,
 use futures::Future;
 use reth_chainspec::MIN_TRANSACTION_GAS;
 use reth_errors::ProviderError;
-use reth_evm::{ConfigureEvm, Database, Evm, EvmEnvFor, TransactionEnv, TxEnvFor};
+use reth_evm::{ConfigureEvm, Database, Evm, EvmEnvFor, EvmFor, TransactionEnv, TxEnvFor};
 use reth_revm::{database::StateProviderDatabase, db::CacheDB};
 use reth_rpc_eth_types::{
     error::{api::FromEvmHalt, FromEvmError},
@@ -82,19 +82,14 @@ pub trait EstimateCall: Call {
         }
 
         // Check if this is a basic transfer (no input data to account with no code)
-        let is_basic_transfer = if tx_env.input().is_empty() {
+        let mut is_basic_transfer = false;
+        if tx_env.input().is_empty() {
             if let TxKind::Call(to) = tx_env.kind() {
                 if let Ok(code) = db.db.account_code(&to) {
-                    code.map(|code| code.is_empty()).unwrap_or(true)
-                } else {
-                    false
+                    is_basic_transfer = code.map(|code| code.is_empty()).unwrap_or(true);
                 }
-            } else {
-                false
             }
-        } else {
-            false
-        };
+        }
 
         // Check funds of the sender (only useful to check if transaction gas price is more than 0).
         //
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -81,28 +81,20 @@ pub trait EstimateCall: Call {
...
+                    code.map(|code| code.is_empty()).unwrap_or(true)
+                } else {
+                    false
                 }
+            } else {
+                false
             }
-        }
+        } else {
+            false
```
mattsse: we can make this a bit nicer and init the var with false and only set it to true when this is a transfer

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -116,10 +108,31 @@ pub trait EstimateCall: Call {
         // If the provided gas limit is less than computed cap, use that
         tx_env.set_gas_limit(tx_env.gas_limit().min(highest_gas_limit));
 
-        trace!(target: "rpc::eth::estimate", ?evm_env, ?tx_env, "Starting gas estimation");
+        // Create EVM instance once and reuse it throughout the entire estimation process
+        let mut evm = self.evm_config().evm_with_env(&mut db, evm_env);
```
mattsse, line 107: this makes sense here

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -128,7 +141,27 @@ pub trait EstimateCall: Call {
...
+                    }
+                    ExecutionResult::Revert { output, .. } => {
+                        // reverted again after bumping the limit
+                        Err(RpcInvalidTransactionError::Revert(RevertError::new(output))
+                            .into_eth_err())
+                    }
+                    ExecutionResult::Halt { reason, .. } => {
+                        Err(Self::Error::from_evm_halt(reason, req_gas_limit))
+                    }
+                };
```
mattsse: why do we need to do this here now?

we didn't transact in this case previously

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -155,7 +188,29 @@ pub trait EstimateCall: Call {
...
+                        }
+                        ExecutionResult::Revert { output, .. } => {
+                            // reverted again after bumping the limit
+                            Err(RpcInvalidTransactionError::Revert(RevertError::new(output))
+                                .into_eth_err())
+                        }
+                        ExecutionResult::Halt { reason, .. } => {
+                            Err(Self::Error::from_evm_halt(reason, req_gas_limit))
+                        }
+                    }
```
mattsse: same question

mattsse (changes requested):
this makes a lot of sense and was in our backlog for a while,

only have a question about the changed error handling

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
thaodt: Good catch! You're right to question this.

  **Previously**: The call `self.map_out_of_gas_err()` which internally used `self.transact()`. This created a new EVM instance for each retry. 
I met some difficulties to resolve the trait bounds issues therefore I basically inlined the logic inside `fn map_out_of_gas_err()`, but now I tried a different approach and it works. Please see new commit :pray: 

  **Now**: we reuse the same EVM instance throughout the entire process, I've updated it to accept an EVM instance parameter instead of creating a new one.
The logic itself remains identical - we're just reusing the existing EVM instance.

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
thaodt: Same as above - both locations now call the updated `map_out_of_gas_err()` which reuses the EVM instance.

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
thaodt: fixed sir! thank you :fire:

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
mattsse: ah I overlooked that map_out_of_gas_err needs to re-execute

mattsse (commented):
this looks pretty good, since estimate requires a bunch of transact calls this should result in a nice speed up.

pending @klkvr

klkvr (approved):
nice

mattsse (approved):

PR #16959, opened 2025-06-20
feat: added DbTx and DbTxMut traits for Either

Rimeeeeee:


crates/storage/db-api/src/db_either.rs
```diff
@@ -0,0 +1,353 @@
...
+    cursor::{
+        DbCursorRO, DbCursorRW, DbDupCursorRO, DbDupCursorRW, DupWalker, RangeWalker,
+        ReverseWalker, Walker,
+    },
+    table::{DupSort, Encode, Table},
+    transaction::{DbTx, DbTxMut},
+};
+use reth_storage_errors::db::DatabaseError;
+use std::ops::{Bound, RangeBounds};
+use tokio_util::either::Either;
```
mattsse: let's use the `either` crate for this instead

crates/storage/db-api/src/db_either.rs
mattsse, line 1: we can name this either.rs

mattsse (changes requested):

mattsse (approved):
this looks great!

but I'm holding off on merging this until we know how we need to use this

mattsse (commented):
this came up internally, and with the approach we want to take next we likely dont need this anymore

sorry about this @Rimeeeeee

PR #16960, opened 2025-06-20
chore(trie): Replace magic numbers in ParallelSparseTrie code

mediocregopher:


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -18,6 +18,13 @@ use reth_trie_sparse::{
 use smallvec::SmallVec;
 use tracing::trace;
 
+/// The maximum length of a path, in nibbles, which belongs to the upper subtrie of a
+/// [`ParallelSparseTrie`]. All longer paths belong to a lower subtrie.
+const UPPER_TRIE_MAX_DEPTH: usize = 2;
+
+/// Number of lower subtries which are managed by the [`ParallelSparseTrie`].
+const NUM_LOWER_SUBTRIES: usize = 1 << (UPPER_TRIE_MAX_DEPTH * 4);
```
shekhirin: can we do `pow` here for readability? 😅

mediocregopher: Done, didn't realize it was a `const fn` :laughing:

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -636,11 +645,12 @@ struct ChangedSubtrie {
...
 
-/// Convert first two nibbles of the path into a lower subtrie index in the range [0, 255].
+/// Convert first [`UPPER_TRIE_MAX_DEPTH`] nibbles of the path into a lower subtrie index in the
+/// range [0, [`NUM_LOWER_SUBTRIES`]).
 ///
 /// # Panics
 ///
-/// If the path is shorter than two nibbles.
+/// If the path is shorter than [`UPPER_TRIE_MAX_DEPTH`] nibbles.
 fn path_subtrie_index_unchecked(path: &Nibbles) -> usize {
```
shekhirin, line 654: `path_subtrie_index_unchecked` doesn't work with variable size max depth, but I think it doesn't worth it to make it dynamic. Let's just add an assertion.
```suggestion
fn path_subtrie_index_unchecked(path: &Nibbles) -> usize {
    debug_assert_eq!(UPPER_TRIE_MAX_DEPTH, 2);
```

shekhirin (approved):

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: Good catch :+1:

PR #16964, opened 2025-06-20
chore: feature gate all op rpc types compat impl

mattsse:
these should be feature gated because we want to avoid pulling in these impls for regular eth.

Introduced an additional op feature in reth-rpc-eth-api for convenience


RomanHodulak (approved):
Makes sense.

You mentioned earlier that we were moving away from `op` feature flag, but I suppose that's only for stuff like extending enum variants?

PR #16965, opened 2025-06-20
feat: Add reproducible debian packaging and a corresponding CI workflow job

MoeMahhouk:
This pull request introduces support for building, packaging, and testing Debian `.deb` packages for the project. It includes updates to the GitHub Actions workflow, the `Makefile`, and a new script for verifying reproducibility of Debian packages. Below are the most important changes grouped by theme:

### GitHub Actions Workflow Enhancements
* Added new Debian build targets (`x86_64-unknown-linux-gnu-deb` and `aarch64-unknown-linux-gnu-deb`) to the matrix in `.github/workflows/release.yml`. These targets include settings for architecture-specific GCC packages and Rust targets.
* Introduced steps for installing Debian build dependencies, adding Rust targets, and installing `cargo-deb` for `.deb` package creation.
* Added separate steps for building and signing Debian packages, generating checksums, and uploading `.deb` artifacts as part of the release process. [[1]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R223-R281) [[2]](diffhunk://#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R223-R281)

### Makefile Updates for Debian Packaging
* Added new `Makefile` targets for building `.deb` packages (`deb-cargo`, `deb-cargo-x86_64`, `deb-cargo-aarch64`, and `deb-cargo-all`), testing reproducibility, and installing or removing packages locally.
* Introduced a `clean-deb` target to clean up Debian packaging artifacts and a `help-deb` target to display Debian-specific build instructions.

### Reproducibility Testing
* Added a new script `scripts/test-deb-reproducible.sh` to test the reproducibility of `.deb` packages by comparing two builds and running `diffoscope` for detailed analysis if differences are found.

### Documentation and Release Notes
* Updated the release notes template in `.github/workflows/release.yml` to include information about Debian packages, their installation instructions, and links to PGP signatures and checksums.

These changes streamline the process of building and distributing Debian packages, ensuring reproducibility and providing clear documentation for end-users.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,13 +97,13 @@ jobs:
           - target: x86_64-unknown-linux-gnu-deb
             os: ubuntu-24.04
             profile: maxperf
-            allow_fail: false
+            allow_fail: true
             rust_target: x86_64-unknown-linux-gnu
             gcc_package: gcc
           - target: aarch64-unknown-linux-gnu-deb
             os: ubuntu-24.04
             profile: maxperf
-            allow_fail: false
+            allow_fail: true
             rust_target: aarch64-unknown-linux-gnu
             gcc_package: gcc-aarch64-linux-gnu
         build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,7 +116,7 @@ jobs:
       - uses: rui314/setup-mold@v1
       - uses: dtolnay/rust-toolchain@stable
         with:
-          target: ${{ matrix.configs.target }}
+          target: ${{ endsWith(matrix.configs.target, '-deb') && matrix.configs.rust_target || matrix.configs.target }}
       - name: Install cross main
         id: cross_main
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,6 +93,19 @@ jobs:
             os: ubuntu-24.04
             profile: maxperf
             allow_fail: true
+          # Debian package builds
+          - target: x86_64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
+            rust_target: x86_64-unknown-linux-gnu
+            gcc_package: gcc
+          - target: aarch64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
+            rust_target: aarch64-unknown-linux-gnu
+            gcc_package: gcc-aarch64-linux-gnu
         build:
           - command: build
             binary: reth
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -93,6 +93,19 @@ jobs:
...
+          - target: x86_64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
+            rust_target: x86_64-unknown-linux-gnu
+            gcc_package: gcc
+          - target: aarch64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
```
mattsse: let's make this true

mattsse (commented):
cant really review this properly, this should also not prevent the existing release setup otherwise we're stuck with this

Rjected:
How critical is this to be in the main reth repo rather than a fork, or on some other infrastructure with automation? i.e., if this were not in the main repo, what would be lost?

We cannot have the following situation happen:
* We try to release
* Something in the reproducible build / deb workflow fails
* We can't debug, or takes too long
* No release that day

MoeMahhouk:
> How critical is this to be in the main reth repo rather than a fork, or on some other infrastructure with automation? i.e., if this were not in the main repo, what would be lost?
> 
> We cannot have the following situation happen:
> 
> * We try to release
> * Something in the reproducible build / deb part of the workflow fails
> * We can't debug, or takes too long
> * No release that day
> 
> So this should not be in any workflow that is a prereq for release, or would block release if failing

Thanks @Rjected for your thorough response!
I do understand your concerns and they make total sense to not block release cycles for too long on this.
Why in main repo:

User trust: Official .deb packages from paradigmxyz/reth carry much more weight than third-party packages
Security: Same build infrastructure = clear provenance chain
Reach: Opens reth to users who prefer debian packages over manual install

Intermediate solution:
Let's make debian builds allow_fail: true so they never block releases. We can help on fixing issues on the reproducible debian packaging without impacting your release velocity. Binary releases continue as normal, debian packages become "nice to have" additions.
Once we've proven stability over a few releases, we can revisit making them required. The implementation reuses your existing build-reproducible target, so the risk is minimal.
Would this non-blocking approach work?

.github/workflows/release.yml
```diff
@@ -93,6 +93,19 @@ jobs:
...
+          - target: x86_64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
+            rust_target: x86_64-unknown-linux-gnu
+            gcc_package: gcc
+          - target: aarch64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: true
```
Rjected: why is one of these `allow_fail: true` and the other `false`?

.github/workflows/release.yml
```diff
@@ -93,6 +93,19 @@ jobs:
...
             profile: maxperf
             allow_fail: true
+          # Debian package builds
+          - target: x86_64-unknown-linux-gnu-deb
+            os: ubuntu-24.04
+            profile: maxperf
+            allow_fail: false
+            rust_target: x86_64-unknown-linux-gnu
+            gcc_package: gcc
+          - target: aarch64-unknown-linux-gnu-deb
```
Rjected: if we use `target` here to distinguish between deb and non-deb builds, the `rust-toolchain` step will fail because the `-deb` names are not valid targets.

Makefile
```diff
@@ -504,3 +504,93 @@ check-features:
...
+	
+	@echo "✅ Package built successfully!"
+	@find target/aarch64-unknown-linux-gnu/debian -name "*.deb" -exec ls -la {} \;
+
+.PHONY: deb-cargo-all
+deb-cargo-all: deb-cargo-x86_64 deb-cargo-aarch64 ## Build .deb for all architectures
+
+.PHONY: test-deb-reproducible
+test-deb-reproducible: ## Test reproducibility of cargo-deb packages
+	@./scripts/test-deb-reproducible.sh
```
Rjected, line 537: I think we should add this to https://github.com/paradigmxyz/reth/blob/main/.github/workflows/reproducible-build.yml

bin/reth/debian/reth.service
```diff
@@ -0,0 +1,37 @@
...
+    --authrpc.port 8551 \
+    --http \
+    --http.addr 127.0.0.1 \
+    --http.port 8545 \
+    --http.api "eth,net,web3,trace,rpc,debug,txpool" \
+    --ws \
+    --ws.addr 127.0.0.1 \
+    --ws.port 8546 \
+    --ws.api "eth,net,trace,web3,rpc,debug,txpool" \
+    --log.stdout.format json \
```
Rjected: why json?

bin/reth/debian/reth.service
```diff
@@ -0,0 +1,37 @@
...
+    --log.file.max-files 0 \
```
Rjected: if we do this / are relying on journald for logs, can set the `RUST_LOG` to `debug`? If people run this and something happens to the node, not having debug logs would be rough

bin/reth/debian/postrm
```diff
@@ -0,0 +1,17 @@
...
+
+if [ "$1" = "purge" ]; then
+    # Remove user (optional - some prefer to keep it)
+    if getent passwd reth >/dev/null 2>&1; then
+        deluser reth || true
+    fi
+    
+    # Remove data directory (be careful!)
+    # rm -rf /var/lib/reth || true
+    # rm -rf /var/log/reth || true
```
Rjected, line 12: this should be run right?

bin/reth/Cargo.toml
```diff
@@ -130,3 +131,26 @@ min-trace-logs = [
 [[bin]]
 name = "reth"
 path = "src/main.rs"
+
+[package.metadata.deb]
+maintainer = "Paradigm <team@paradigm.xyz>"
+extended-description = """\
+Reth is a new Ethereum full node implementation that is focused on being \
+user-friendly, highly modular, as well as being fast and efficient."""
```
Rjected:
```suggestion
Modular, contributor-friendly and blazing-fast implementation of the Ethereum protocol, in Rust."""
```

bin/reth/Cargo.toml
```diff
@@ -130,3 +131,26 @@ min-trace-logs = [
 [[bin]]
 name = "reth"
 path = "src/main.rs"
+
+[package.metadata.deb]
+maintainer = "Paradigm <team@paradigm.xyz>"
```
Rjected, line 136: not sure this is the right email, checking

Rjected (changes requested):
I have some initial comments / questions, I think building deb packages as part of releases makes sense. I want to make sure that any failures in build-reproducible are caught before we do a release. I am currently leaning towards `allow_fail: true` on the debian package builds, because we previously had a delayed release due to a failure in `make build-reproducible`.

.github/workflows/release.yml
MoeMahhouk: it was left by mistake, I will change both to true

.github/workflows/release.yml
MoeMahhouk: This is handled later [here ](https://github.com/paradigmxyz/reth/pull/16965/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R171-R180)for example

Makefile
MoeMahhouk: we could have it there as well, but in general, you wouldnt want to publish debian packages that are not reproducible, or do you want to separate that as well and let it publish them despite of that?

bin/reth/debian/postrm
MoeMahhouk: I left this commented for feedback if you want to include it as postrm or not. It is more of a product question.
I can include it if you find it useful

bin/reth/debian/reth.service
MoeMahhouk: no particular reason, it can be anything.
Such systemd file would in most cases be overwritten by the user for their specific use-cases

bin/reth/debian/reth.service
MoeMahhouk: What is the minimalistic set of flags to set for this that you suggest to have here and I will change it.
The main idea is to provide it in the packaging and potentially leave it to the user to modify it with any potential extra flags. 
The ones I included here are from our buildernet instance but probably we should do a simple set of flags that you (reth team) recommend.

.github/workflows/release.yml
Rjected: right, but we run this first
```yml
      - uses: dtolnay/rust-toolchain@stable
        with:
          target: ${{ matrix.configs.target }}
```
which would mean the `-deb` builds would always fail I think?

Makefile
Rjected: > let it publish them despite of that?

I think this, just add `make test-deb-reproducible` to the `reproducible-build` workflow. The `reproducible-build` workflow should test that the build is actually reproducible, and give a warning if something is causing us to not be reproducible. It's currently run via cron

bin/reth/debian/reth.service
Rjected: IMO let's remove this then and keep it default

Makefile
```diff
@@ -504,3 +504,93 @@ check-features:
...
+	@find target/x86_64-unknown-linux-gnu/debian -name "*.deb" -exec ls -la {} \;
+
+.PHONY: deb-cargo-x86_64
+deb-cargo-x86_64: ## Build .deb for x86_64 architecture
+	$(MAKE) deb-cargo
+
+.PHONY: deb-cargo-aarch64  
+deb-cargo-aarch64: ## Build .deb for aarch64 architecture
+	@echo "Building .deb package for aarch64..."
+	$(MAKE) build-reproducible RUST_TARGET=aarch64-unknown-linux-gnu PROFILE=$(PROFILE)
```
Rjected: ah, I think the `build-reproducible` part of the makefile needs to change as well because it currently hardcodes `--target x86_64-unknown-linux-gnu` and does not respect `RUST_TARGET`

MoeMahhouk: good catch. I will address that

Makefile
MoeMahhouk: I added it there now and found out that the reproducible-build workflow is failing because the reproducible-build was not generating reproducible binaries.
So I fixed that now and now testing the workflow if it will pass with both the binaries and debian packaging reproducibility.

bin/reth/Cargo.toml
MoeMahhouk: Sure, I will change it to the correct mail once you share that with me

Rjected:
Need to review this again, would like to have this

MoeMahhouk:
closing this PR as it was later on done within this [PR](https://github.com/paradigmxyz/reth/pull/18615)

PR #16967, opened 2025-06-20
feat(transaction-pool): enforce EIP-2681

### LINKED ISSUE #16966: Missing EIP-2681 validation
### Describe the feature

The current code is missing [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681) nonce validation, this can easily be added just before these lines

https://github.com/paradigmxyz/reth/blob/24f0365340889630358f1577b1fe9d7b296530b0/crates/transaction-pool/src/validate/eth.rs#L296-L303

### Additional context

_No response_

0xMushow:
closes #16966

### DIFF
--- a/crates/rpc/rpc-eth-types/src/error/mod.rs
+++ b/crates/rpc/rpc-eth-types/src/error/mod.rs
@@ -851,7 +851,9 @@ impl From<InvalidPoolTransactionError> for RpcPoolError {
             }
             InvalidPoolTransactionError::OversizedData(_, _) => Self::OversizedData,
             InvalidPoolTransactionError::Underpriced => Self::Underpriced,
-            InvalidPoolTransactionError::Eip2681 => Self::Invalid(RpcInvalidTransactionError::NonceMaxValue),
+            InvalidPoolTransactionError::Eip2681 => {
+                Self::Invalid(RpcInvalidTransactionError::NonceMaxValue)
+            },
             InvalidPoolTransactionError::Other(err) => Self::PoolTransactionError(err),
             InvalidPoolTransactionError::Eip4844(err) => Self::Eip4844(err),
             InvalidPoolTransactionError::Eip7702(err) => Self::Eip7702(err),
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -293,6 +293,15 @@ where
             }
         };
 
+        // Reject transactions with a nonce higher or equal to U64::max according to EIP-2681
+        let tx_nonce = transaction.nonce();
+        if tx_nonce >= u64::MAX {
```
mattsse: >= doesn't make sense we can just check for ==

crates/transaction-pool/src/error.rs
```diff
@@ -247,6 +247,9 @@ pub enum InvalidPoolTransactionError {
         /// Balance of account.
         balance: U256,
     },
+    /// EIP-2681 error thrown if the nonce is higher or equal than U64::max
```
mattsse: can we link the eip here

mattsse (changes requested):
smol nit

crates/transaction-pool/src/validate/eth.rs
0xMushow: yep realised afterwards, thanks!

crates/transaction-pool/src/error.rs
0xMushow: done :)

crates/rpc/rpc-eth-types/src/error/mod.rs
```diff
@@ -775,6 +775,9 @@ pub enum RpcPoolError {
     /// Custom pool error
     #[error(transparent)]
     PoolTransactionError(Box<dyn PoolTransactionError>),
+    /// EIP-2681 related error
+    #[error("nonce is not valid")]
+    Eip2681,
```
mattsse: we can get rid of this and instead simply map this to RpcInvalidTransactionError::NonceMaxValue

which is the variant that we already have for this

mattsse (changes requested):
smol nit

mattsse (approved):

PR #16969, opened 2025-06-20
feat: bump revm v26

klkvr:


mattsse (approved):

crates/storage/provider/src/writer/mod.rs
```diff
@@ -792,24 +807,22 @@ mod tests {
...
             },
         )]));
-        state.merge_transitions(BundleRetention::Reverts);
-
-        // Block #7: Change storage.
         state.commit(HashMap::from_iter([(
             address1,
             RevmAccount {
-                status: AccountStatus::Touched,
+                status: AccountStatus::Touched | AccountStatus::SelfDestructed,
```
rakita: Should this be reverted?

mattsse: I think so?

maybe this does actually explain the failing selfdestruct error?

rakita: @klkvr we should add back 

before this commit:
```
        state.merge_transitions(BundleRetention::Reverts);

        // Block #7: Change storage.
 ```
 and after this commit:  
 ```
         state.merge_transitions(BundleRetention::Reverts);
```

mattsse: added this back and the test works again

PR #16976, opened 2025-06-20
fix: add missing historical RPC endpoints for Optimism pre-bedrock

mattsse:
Added missing endpoints that check `IsOptimismPreBedrock` in op-geth:
- `eth_estimateGas`
- `eth_createAccessList`  
- `eth_getProof`

Also fixed `eth_getStorageAt` to parse block ID from correct parameter position (2 instead of 1), as parameters are `[address, key, block]`.

Found by comparing op-geth's usage of `HistoricalRPCService()` with our implementation.


Rjected (approved):
nice, lgtm

PR #16980, opened 2025-06-21
fix: use empty withdrawals if parent has withdrawals root

mattsse:
for pending block we were always providing None, but we can easily check the parent

### DIFF
--- a/crates/rpc/rpc/src/eth/helpers/pending_block.rs
+++ b/crates/rpc/rpc/src/eth/helpers/pending_block.rs
@@ -72,7 +72,7 @@ where
             prev_randao: B256::random(),
             gas_limit: parent.gas_limit(),
             parent_beacon_block_root: parent.parent_beacon_block_root().map(|_| B256::ZERO),
-            withdrawals: None,
+            withdrawals: parent.withdrawals_root().map(|_| Default::default()),
         })
     }
 }


crates/rpc/rpc/src/eth/helpers/pending_block.rs
```diff
@@ -72,7 +72,7 @@ where
             prev_randao: B256::random(),
             gas_limit: parent.gas_limit(),
             parent_beacon_block_root: parent.parent_beacon_block_root().map(|_| B256::ZERO),
-            withdrawals: None,
+            withdrawals: parent.withdrawals_root().map(|_| Default::default()),
```
RomanHodulak, line 70: Can you help me understand what this means?

So the EVM ENV for the next block has withdrawals in it and the withdrawals are set to empty vector if the parent block has withdrawal root?

mattsse: yeah exactly, withdrawals_root is Some if the parent block already included withdrawals.

if that's the case we also should configure empty withdrawals for the next block, same thing with the `parent_beacon_block_root`.

by checking the parent, we effectively determine if the corresponding hardfork was already active when the parent block was created

RomanHodulak: Gotcha, so this should be only `None` for pre-merge blocks?

RomanHodulak (approved):

crates/rpc/rpc/src/eth/helpers/pending_block.rs
mattsse: yep, the distinction really is `Some()` vs `None`.

for example, in opstack there are no withdrawals, but op mirrors L1, so there is only `None` vs `Some(vec![])`

PR #15909, opened 2025-04-24
feat: introduce reth era export

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/15495.

Only start introducing the functional logic (not the interface with the cli)  : 

For a certain number of block set up in the config, it fetches the data (header, block body, receipts) to reconstruct the block history and write them directly into era files. 

Still to do : 
- [x] Add writing logic,
- [x]  Add tests with import/export rountrip --> started but not finished

PR #16984, opened 2025-06-21
chore(era): complete doc for `ClientWithFakeIndex`

lean-apple:
Some preparation to help to finalize https://github.com/paradigmxyz/reth/pull/15909.

- Move `ClientWithFakeIndex` to `/test/it/main.rs` instead of `/test/it/history.rs`, it can also be moved to a `src/test_utils` file (probably better),
- Make doc for `ClientWithFakeIndex` more explicit,
- make impl for ClientWithFakeIndex,  `get` method used async directly,
- use some const for the index url/response.

cc @RomanHodulak, @mattsse.


mattsse (approved):
cool

PR #16985, opened 2025-06-21
chore: simplify test HttpClient implementations

mattsse:
Removes unnecessary boxing and type casts in test HTTP clients.

## Changes
- Use `async fn` directly instead of `future::ready` wrapper
- Use `futures::stream::iter` for already Unpin streams instead of `stream::once` with boxing
- Use `Bytes::from_static` for zero-copy static data instead of `Bytes::from`
- Use `tokio_util::either::Either` in ClientWithFakeIndex to avoid boxing different stream types

## Benefits
- Eliminates heap allocations in test code
- Removes dynamic dispatch overhead
- Cleaner, more readable test implementations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8058,6 +8058,7 @@ dependencies = [
  "reth-storage-api",
  "tempfile",
  "tokio",
+ "tokio-util",
  "tracing",
 ]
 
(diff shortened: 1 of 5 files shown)


RomanHodulak (approved):
Nice

PR #17002, opened 2025-06-23
chore: simplify `RpcConverter`

klkvr:
Given that it gets `ConfigureEvm` which already has primitives we don't need a separate `N: NodePrimitives` generic here


mattsse (approved):
ah nice, that makes sense

@RomanHodulak for vis

RomanHodulak (approved):

PR #17004, opened 2025-06-23
chore: fix typo execution.rs

kks-code:


crates/stages/stages/src/stages/execution.rs
```diff
@@ -71,7 +71,7 @@ where
     evm_config: E,
     /// The consensus instance for validating blocks.
     consensus: Arc<dyn FullConsensus<E::Primitives, Error = ConsensusError>>,
-    /// The consensu
+    /// The consensus
     /// The commit thresholds of the execution stage.
```
mattsse: I think this comment is totally wrong here and should be removed entirely, likely this was intended for the field above

mattsse (changes requested):

crates/stages/stages/src/stages/execution.rs
kks-code: done!

mattsse (approved):

PR #17007, opened 2025-06-23
chore: rm standalone fn

mattsse:
this is just bloat and we should always go through the builder api


RomanHodulak (approved):
:+1: Decreases maintenance burden.

PR #17008, opened 2025-06-23
chore: add size field in the new_header_stream method

### LINKED ISSUE #16997: Add optional `size` field to `eth_subscribe` `newHeads` response
### Describe the feature

Currently, `eth_subscribe` `newHeads` response does not include the `size` field, which represents the RLP-encoded size of the block in bytes. This aligns with geth's implementation (confirmed in https://github.com/ethereum/go-ethereum/issues/28700), but several RPC providers have extended the response to include this field, causing compatibility issues with tools that expect it.

## Situation
  - Some RPC providers (e.g., alchemy, chainstack) include the `size` field in their `newHeads` responses despite `Geth` not including it
  - Libraries like viem include `size` in their type definitions for `newHeads`
  - Applications like Ponder expect this field to be present
  - Users need to make an additional `eth_getBlockByNumber` call to get the block size, adding unnecessary RPC overhead.

## Proposed Solution

Add the "size" field to the `newHeads` subscription response. 
The implementation would be minimal, modifying the `newHeads` stream to have access to full block data (not just headers) to calculate the size, similar to how it's done for uncle blocks where `block.length()` is used to compute the size:

https://github.com/paradigmxyz/reth/blob/4f5ad18682ab332beb9c969fe29da6f3dee11da8/crates/rpc/rpc/src/eth/pubsub.rs#L354-L361

How do you think?

## Pros:
  - Eliminates the need for additional RPC calls to get block size
  - Improves compatibility with popular RPC providers and tools
  - Minimal performance impact - block data is already available in the stream
  - Additive change that won't break existing clients

## Compatibility Considerations

Geth Compatibility: Geth explicitly does not include this field. However:
  - The Ethereum JSON-RPC spec doesn't strictly define the newHeads response format
  - Multiple major providers already include additional fields beyond Geth's implementation
  - The field would be additive and shouldn't break existing clients
  - This creates a practical compatibility issue where tools built against provider APIs don't work with Geth-compatible clients

kamuik16:
Closes #16997 

The approach seemed simple enough not to use helper functions. Let me know if still, those helper fns are still needed.

### DIFF
--- a/crates/rpc/rpc/src/eth/pubsub.rs
+++ b/crates/rpc/rpc/src/eth/pubsub.rs
@@ -2,7 +2,7 @@
 
 use std::sync::Arc;
 
-use alloy_primitives::TxHash;
+use alloy_primitives::{TxHash, U256};
 use alloy_rpc_types_eth::{
     pubsub::{Params, PubSubSyncStatus, SubscriptionKind, SyncStatusMetadata},
     Filter, Header, Log,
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -353,10 +353,18 @@ where
...
-                headers.into_iter().map(|h| Header::from_consensus(h.into(), None, None)),
-            )
+            let headers = new_chain
+                .committed()
+                .blocks_iter()
+                .map(|block| {
+                    Header::from_consensus(
+                        block.clone_sealed_header().into(),
+                        None,
+                        Some(U256::from(block.rlp_length())),
```
mattsse, line 363: ah forgot that we already had this

mattsse (approved):
lgtm!

PR #17009, opened 2025-06-23
refactor(rpc): Rename `TransactionCompat` => `RpcConvert`

RomanHodulak:
The new name fits better with the trait as it's responsible for conversions from and into JSON-RPC io objects.


crates/rpc/rpc-types-compat/src/transaction.rs
```diff
@@ -21,7 +21,7 @@ use std::{convert::Infallible, error::Error, fmt::Debug, marker::PhantomData};
 use thiserror::Error;
 
 /// Builds RPC transaction w.r.t. network.
```
mattsse: should we update docs here?

mattsse (changes requested):
smol doc nit

pending @klkvr

mattsse (approved):
lgtm

PR #17010, opened 2025-06-23
docs(rpc): Add documentation for `RpcConverter`

RomanHodulak:
The documentation for this struct is almost non-existent. It would be useful to have some, especially when the `RpcConvert` docs are referring to it.


mattsse (approved):
cool

PR #16727, opened 2025-06-09
perf: `U256` nybbles

shekhirin:
Bump alloy-trie to 0.9.0 and nybbles to 0.4.0.

This patches all Alloy deps to a commit. The patch will be removed once new Alloy version is released.

Brings performance improvements from migrating nybbles to `U256` representation.

latency_histogram

PR #17011, opened 2025-06-23
perf(trie): Place the root nodes of the lower SparseSubtries in those tries

mediocregopher:
Previously the root nodes of each lower SparseSubtrie (0x00, 0x01, 0x02, etc..) would get placed in the upper SparseSubtrie. This is inconvenient for `update_hashes` and other methods which we are implementing, so this PR moves those nodes into their respective lower trie instance.

In implementing this I found a couple of couplings between the logic of the `SparseSubtrieType` and `reveal_node`; a new method is added to `SparseSubtrieType` to remove this coupling.


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -119,8 +119,9 @@ impl ParallelSparseTrie {
         match node {
             TrieNode::Branch(branch) => {
                 // If a branch is at the cutoff level of the trie then it will be in the upper trie,
-                // but all of its children will be in a lower trie.
-                if path.len() == UPPER_TRIE_MAX_DEPTH {
+                // but all of its children will be in a lower trie. Check if a child node would be
+                // in the lower subtrie, and reveal accordingly.
+                if !SparseSubtrieType::path_len_is_upper(path.len() + 1) {
```
shekhirin, line 124: can we instead do `SparseSubtrieType::from_path(&path).is_upper()` here? Feels easier to read IMO

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -137,10 +138,8 @@ impl ParallelSparseTrie {
             TrieNode::Extension(ext) => {
                 let mut child_path = path.clone();
                 child_path.extend_from_slice_unchecked(&ext.key);
-                if child_path.len() > UPPER_TRIE_MAX_DEPTH {
-                    self.lower_subtrie_for_path(&child_path)
-                        .expect("child_path must have a lower subtrie")
-                        .reveal_node_or_hash(child_path, &ext.child)?;
+                if let Some(subtrie) = self.lower_subtrie_for_path(&child_path) {
+                    subtrie.reveal_node_or_hash(child_path, &ext.child)?;
```
shekhirin, line 142: makes sense 👍

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -583,24 +582,29 @@ impl SparseSubtrie {
 /// Sparse Subtrie Type.
 ///
 /// Used to determine the type of subtrie a certain path belongs to:
-/// - Paths in the range `0x..=0xff` belong to the upper subtrie.
-/// - Paths in the range `0x000..` belong to one of the lower subtries. The index of the lower
-///   subtrie is determined by the path first nibbles of the path.
+/// - Paths in the range `0x..=0xf` belong to the upper subtrie.
+/// - Paths in the range `0x00..` belong to one of the lower subtries. The index of the lower
+///   subtrie is determined by the path first [`UPPER_TRIE_MAX_DEPTH`] nibbles of the path.
```
```suggestion
///   subtrie is determined by the first [`UPPER_TRIE_MAX_DEPTH`] nibbles of the path.
```

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: To do this we would need to `push` a nibble onto `path`, which I wanted to avoid cause it might incur an allocation... but maybe with your U256 changes in #16727 that's no longer the case? Not 100% on the context there

shekhirin: I see, I'm okay with having this method then 👍

shekhirin (approved):
LGTM, also needs Dan's review

Rjected (approved):
this makes sense

PR #17013, opened 2025-06-23
refactor(rpc): Rename crate `reth_rpc_types_compat` => `reth_rpc_convert`

RomanHodulak:
Changes the name of the mentioned crate to fit better with it's purpose which is RPC conversions.


mattsse (approved):
makes sense, 

unsure how many currently import this as dep directly but I think we should do this

RomanHodulak:
Little worried about that too, I suppose we communicate this change via release notes? Maybe we can add `reth_rpc_types_compat` crate as a re-export of `reth_rpc_convert`, but we've already made breaking change with the `RpcConvert` trait rename.

PR #17022, opened 2025-06-23
feat(tx-pool): add getter methods for `EthTransactionValidator` internal fields

### LINKED ISSUE #17014: Exposer more getters for EthTransactionValidator
### Describe the feature

https://github.com/paradigmxyz/reth/blob/dc67f0237ff01048e79aec2f6f346ad378dc3afc/crates/transaction-pool/src/validate/eth.rs#L46-L46

could be reused for more custom validation:

https://github.com/paradigmxyz/reth/blob/dc67f0237ff01048e79aec2f6f346ad378dc3afc/crates/transaction-pool/src/validate/eth.rs#L238-L245

but we're currently missing some getters to obtain settings:

https://github.com/paradigmxyz/reth/blob/dc67f0237ff01048e79aec2f6f346ad378dc3afc/crates/transaction-pool/src/validate/eth.rs#L156-L177

## TODO
* add missing getters to EthTransactionValidator to access internals

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17014.

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -61,6 +61,62 @@ impl<Client, Tx> EthTransactionValidator<Client, Tx> {
     pub fn client(&self) -> &Client {
         &self.inner.client
     }
+
+    /// Returns the tracks activated forks relevant for transaction validation
+    pub fn fork_tracker(&self) -> &ForkTracker {
+        &self.inner.fork_tracker
+    }
+
+    /// Returns if there are EIP-2718 type transactions
+    pub fn eip2718(&self) -> bool {
+        self.inner.eip2718
+    }
+
+    /// Returns if there are EIP-1559 type transactions
+    pub fn eip1559(&self) -> bool {
+        self.inner.eip1559
+    }
+
+    /// Returns if there are EIP-4844 blob transactions
+    pub fn eip4844(&self) -> bool {
+        self.inner.eip4844
+    }
+
+    /// Returns if there are EIP-7702 type transactions
+    pub fn eip7702(&self) -> bool {
+        self.inner.eip7702
+    }
+
+    /// Returns the current max gas limit
+    pub fn block_gas_limit(&self) -> &AtomicU64 {
+        &self.inner.block_gas_limit
+    }
+
+    /// Returns the current tx fee cap limit in wei locally submitted into the pool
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -61,6 +61,62 @@ impl<Client, Tx> EthTransactionValidator<Client, Tx> {
...
+    /// Returns the config to handle [`TransactionOrigin::Local`](TransactionOrigin) transactions..
+    pub fn local_transactions_config(&self) -> &LocalTransactionConfig {
+        &self.inner.local_transactions_config
+    }
+
+    /// Returns the maximum size in bytes a single transaction can have in order to be accepted into
+    /// the pool.
+    pub fn max_tx_input_bytes(&self) -> &usize {
+        &self.inner.max_tx_input_bytes
+    }
```
mattsse: no need to return a ref here
```suggestion
    /// Returns the maximum size in bytes a single transaction can have in order to be accepted into
    /// the pool.
    pub fn max_tx_input_bytes(&self) -> usize {
        self.inner.max_tx_input_bytes
    }
```

crates/transaction-pool/src/validate/eth.rs
```diff
@@ -61,6 +61,62 @@ impl<Client, Tx> EthTransactionValidator<Client, Tx> {
...
+        self.inner.eip4844
+    }
+
+    /// Returns if there are EIP-7702 type transactions
+    pub fn eip7702(&self) -> bool {
+        self.inner.eip7702
+    }
+
+    /// Returns the current max gas limit
+    pub fn block_gas_limit(&self) -> &AtomicU64 {
```
mattsse: we can call this function here

https://github.com/paradigmxyz/reth/blob/cf8ff9829cca8a16bbed99a2de9e3d85b8dfe92b/crates/transaction-pool/src/validate/eth.rs#L674-L676

and return u64

mattsse (changes requested):
smol nits

crates/transaction-pool/src/validate/eth.rs
lean-apple: yup but I moved to the other impl to have the right bounds or we can reuse the whole logic `self.block_gas_limit.load(std::sync::atomic::Ordering::Relaxed) ` if we want to keep there

mattsse (approved):

PR #17024, opened 2025-06-23
feat: add configurable RPC middleware to RpcAddOns

### LINKED ISSUE #16942: Make RpcMiddleWare customizable when launching RPC server
### Describe the feature

ref #16940

The rpcserverconfig supports:
https://github.com/paradigmxyz/reth/blob/aa725dd0cf84462cd3bf13257f0aed8e1a9eebe4/crates/rpc/rpc-builder/src/lib.rs#L1078-L1079

we currently don't have a way to customize this in RpcAddOns
we can introduce a field for this in RpcAddons, mirroring RpcServerConfig

```
    /// Configurable RPC middleware
    rpc_middleware: RpcServiceBuilder<RpcMiddleware>,
```
or add an additional argument for Middleware when launching the server

https://github.com/paradigmxyz/reth/blob/aa725dd0cf84462cd3bf13257f0aed8e1a9eebe4/crates/node/builder/src/rpc.rs#L528-L528

The former likely allows more usecases, we'd then need to configure the middleware here before launching the node

https://github.com/paradigmxyz/reth/blob/aa725dd0cf84462cd3bf13257f0aed8e1a9eebe4/crates/node/builder/src/rpc.rs#L549-L550

### Additional context

_No response_

mattsse:
## Summary

Adds configurable RPC middleware support to `RpcAddOns`, allowing users to customize middleware layers for RPC servers.

## Changes

- Added a new generic parameter `RpcMiddleware` to `RpcAddOns` with a default of `Identity`
- Added `rpc_middleware: RpcServiceBuilder<RpcMiddleware>` field to the struct
- Updated the constructor to accept the middleware parameter
- Added `set_rpc_middleware()` method for runtime middleware configuration
- Integrated middleware with RPC server launching (HTTP, WS, IPC)
- Updated Ethereum and Optimism node implementations to use default middleware

## Implementation Details

The middleware is applied when launching RPC servers through the `launch_rpc_server` and `launch_add_ons_with` methods. The implementation mirrors the existing `RpcServerConfig` middleware pattern, providing consistent API design.

Users can now:
- Use default Identity middleware (no-op) without specifying type parameters
- Configure custom middleware layers for request logging, metrics, rate limiting, etc.
- Change middleware at runtime using `set_rpc_middleware()`

Closes #16942

### DIFF
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -431,7 +431,10 @@ pub struct RpcAddOns<
     engine_validator_builder: EV,
     /// Builder for `EngineApi`
     engine_api_builder: EB,
-    /// Configurable RPC middleware
+    /// Configurable RPC middleware stack.
+    ///
+    /// This middleware is applied to all RPC requests across all transports (HTTP, WS, IPC).
+    /// See [`RpcAddOns::set_rpc_middleware`] for more details.
     rpc_middleware: RpcServiceBuilder<RpcMiddleware>,
 }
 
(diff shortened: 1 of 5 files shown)


crates/node/builder/src/rpc.rs
```diff
@@ -455,28 +466,98 @@ where
...
+    ///     .layer(metrics_layer);
+    /// let with_full_stack = rpc_addons.set_rpc_middleware(middleware_stack);
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// - Middleware is applied to the RPC service layer, not the HTTP transport layer
+    /// - The default middleware is `Identity` (no-op), which passes through requests unchanged
+    /// - Middleware layers are applied in the order they are added via `.layer()`
+    pub fn set_rpc_middleware<T>(
```
klkvr:
```suggestion
    pub fn with_rpc_middleware<T>(
```

klkvr (approved):
nice

Issue #15117, opened 2025-03-18
Feature: Allow `RpcAddOns` to accept custom transaction requests


fedemagnani:
### Describe the feature

**Description**:
I am working on attaching custom bytes to a transaction and encountered limitations in the current trait bounds within `RpcAddOns`. While defining new `NodePrimitives` and implementing `NodeComponents` was straightforward due to good trait abstractions, integrating a `CustomTransactionRequest` within the RPC framework presents challenges.

**Problem**:

- `RpcAddOns` requres that `eth_api_builder` implements `EthApiBuilder`, whose associated type `EthApi` must implement `FullEthApiServer` and all its related trait bounds.

https://github.com/paradigmxyz/reth/blob/03c5abba887452a00680ab5e8a1aee34811158a1/crates/node/builder/src/rpc.rs#L349-L364

- Some of these trait bounds (such as`EthCall`, `Call`) enforce the use of `alloy_rpc_types_eth::TransactionRequest`, preventing abstraction for custom transaction requests.

https://github.com/paradigmxyz/reth/blob/03c5abba887452a00680ab5e8a1aee34811158a1/crates/rpc/rpc-eth-api/src/helpers/call.rs#L51-L62

- `EthCall` depends on `FullEthApiTypes`, which provides access to `Self::NetworkTypes: RpcTypes`, that is a 
subset of `alloy_network::Network` limited to response-associated types.

https://github.com/paradigmxyz/reth/blob/03c5abba887452a00680ab5e8a1aee34811158a1/crates/rpc/rpc-eth-api/src/types.rs#L15-L34

### Additional context

**Proposed Solution**:
Expanding `RpcTypes` to include request-associated types (or directly using `alloy_network::Network`) would allow greater flexibility in handling transaction requests.

**Expected Benefits**:

- Enables custom transaction structures within the existing RPC framework.

- Improves extensibility for projects requiring non-standard transaction formats.

- Reduces the need for workarounds when integrating custom transaction logic.

**Questions**:
Would it be feasible to extend `RpcTypes` for request-associated types, or is there an alternative approach that aligns better with the current design? Open to discussion on possible solutions.

mattsse:
> Would it be feasible to extend RpcTypes for request-associated types, or is there an alternative approach that aligns better with the current design?

I think we can definitely try this, it is already part of the `NetworkTypes`

[...]

PR #17025, opened 2025-06-23
feat(rpc): Propagate the RPC transaction request from `Network` and `RpcTypes`

RomanHodulak:
Part of #15117 

Propagates the associated transaction request type throughout the RPC API.

For example, `Optimism` has `OpTransactionRequest` associated with it and that's the target of `TryIntoSimTx` and `TryIntoTxEnv` impls now enforced by the `OpEthApi`.

There are some places that still work directly with the `TransactionRequest` struct, accessing its fields. Those should be probably abstracted out.

The struct `TransactionRequest` is still the input of some of the `EthApiServer` methods, despite now having a generic for it.

In this PR we simply use `From<TransactionRequest>` bound to make it work.

### DIFF
--- a/crates/rpc/rpc-convert/src/transaction.rs
+++ b/crates/rpc/rpc-convert/src/transaction.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     fees::{CallFees, CallFeesError},
-    RpcTransaction, RpcTypes,
+    RpcTransaction, RpcTxReq, RpcTypes,
 };
 use alloy_consensus::{error::ValueError, transaction::Recovered, EthereumTxEnvelope, TxEip4844};
 use alloy_primitives::{Address, TxKind, U256};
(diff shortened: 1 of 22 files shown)


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -499,7 +495,7 @@ pub mod op {
             block_env: &BlockEnv,
         ) -> Result<OpTransaction<TxEnv>, Self::Err> {
             Ok(OpTransaction {
-                base: self.try_into_tx_env(cfg_env, block_env)?,
+                base: self.as_ref().clone().try_into_tx_env(cfg_env, block_env)?,
```
mattsse, line 498: this will go away when we finalize 

https://github.com/alloy-rs/op-alloy/pull/525/

mattsse (approved):
all of this lgtm, pending @klkvr 
imo it's fine if we cant fully get rid of the txrequest constraint here but using Into is still an improvement

need to finalize https://github.com/alloy-rs/op-alloy/pull/525 next

and maybe extend the txrequest with some helpers so that we can make this fully configurable

PR #17028, opened 2025-06-23
feat: introduce max_tx_gas_limit feature to enforce per-transaction gas limits

niran:
**Description**

This PR adds a configurable maximum gas limit for individual transactions in the reth transaction pool. The feature allows node operators to reject transactions that exceed a specified gas limit, providing additional control over resource usage and transaction size.

Key changes:
- Adds `--txpool.max-tx-gas` CLI argument to configure the maximum gas limit per transaction
- Implements validation logic to reject transactions exceeding the configured limit
- Uses `Option<u64>` where `None` disables the limit (default behavior)

**Tests**

Added comprehensive unit tests covering:
- Transaction validation when max gas limit is exceeded (should reject)
- Transaction validation when max gas limit is disabled (should accept)
- Transaction validation when transaction is within the configured limit (should accept)
- Proper error propagation through the transaction pool
- Integration with the existing transaction validation pipeline

**Additional context**

By default, transactions can use as much gas as is available to an entire block. This is reasonable on Ethereum mainnet where the block gas limit can be executed in a small fraction of the block time, but on high throughput rollups, it allows worst case transactions to push execution time beyond the block time—sometimes far beyond the block time.  Flashblocks introduce an even smaller constraint of one-tenth of a block.

Both state growth and gas target increases change the equilibrium for worst case transactions that the latest opcode pricing was intended to achieve. In between opcode repricings, we need a way to mitigate this problem. A per-transaction gas maximum prevents a single large worst case transaction (e.g. XEN) from stalling the chain, and limits the excess execution time that a block full of worst case transactions can cause.

The limit can be set based on how underpriced we expect opcodes to become before we fork to reprice them. For instance, a 10x underpriced opcode can 10x execution time without a transaction gas maximum. With a transaction gas maximum of 1/10th of the block gas limit, a 10x underpriced opcode can only double the execution time. A 5x underpriced opcode would be able to 1.5x the execution time with the same limit, and a 20x underpriced opcode would 3x the execution time.

[EIP 7825](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7825.md) aims to introduce a similar per-transaction gas limit at 30 million gas that would apply to both the txpool and block validation. High throughput rollups will typically want a lower limit chosen in relation to the block gas limit, and single-sequencer rollups can avoid block validation rules that require a hard fork to adjust as gas dynamics evolve.


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -1259,7 +1288,7 @@ mod tests {
         let blob_store = InMemoryBlobStore::default();
         let validator = EthTransactionValidatorBuilder::new(provider)
             .set_tx_fee_cap(0) // no cap
-            .build(blob_store);
+            .build(blob_store.clone());
```
mattsse: please undo unrelated changes

crates/transaction-pool/src/validate/eth.rs
```diff
@@ -1096,7 +1123,9 @@ pub fn ensure_intrinsic_gas<T: EthPoolTransaction>(
         spec_id,
         transaction.input(),
         transaction.is_create(),
-        transaction.access_list().map(|l| l.len()).unwrap_or_default() as u64,
+        transaction
+            .access_list()
+            .map(|l| l.len()).unwrap_or_default() as u64,
```
mattsse: does this just need fmt?
unsure what changed here

crates/transaction-pool/src/error.rs
```diff
@@ -316,6 +319,7 @@ impl InvalidPoolTransactionError {
                 }
             }
             Self::ExceedsGasLimit(_, _) => true,
+            Self::MaxTxGasLimitExceeded(_, _) => true,
```
mattsse: this is a bit ambiguous because these are still technically valid, if this setting is used on L1 it could lead to issues where broadcasting these txs would get peers penalized, so I think we either need to make this one false for now or also configure a setting for how this error should be treated

crates/node/core/src/args/txpool.rs
```diff
@@ -69,6 +69,10 @@ pub struct TxPoolArgs {
     #[arg(long = "txpool.gas-limit", default_value_t = ETHEREUM_BLOCK_GAS_LIMIT_30M)]
     pub enforced_gas_limit: u64,
```
mattsse, line 70: this setting already exists here, which is currently used by the pool directly,

I think we should instead rethink this and unify these two settings.

we can make this an option instead for example

mattsse (changes requested):
this makes sense,

Some nits and I believe we need to unify some settings, because the current enforced_gas_limit doesn't behave as expected and this setting should really be enforced in the validator instead

crates/node/core/src/args/txpool.rs
niran: `enforced_gas_limit` is described as a per-transaction maximum when set as `PoolConfig.gas_limit`, but `AllTransactions` uses it as the _initial_ value of `block_gas_limit`, which is then updated from incoming block headers, so the setting does not constrain the mempool at all. The stateless validation in `EthTransactionValidatorInner` never seems to use `enforced_gas_limit`—it starts from the constant `ETHEREUM_BLOCK_GAS_LIMIT_30M`. Unless I'm missing something, `enforced_gas_limit` is really more like `initial_block_gas_limit`.

crates/transaction-pool/src/error.rs
niran: This should definitely be false, I just didn't understand what that value did before. Updated in e3a8475a29410035bc8dc4804f37bdc14c83387a

crates/transaction-pool/src/validate/eth.rs
niran: Removed in fe8cff398544fd72682d68b2461b97ba04881129.

crates/transaction-pool/src/validate/eth.rs
niran: Done in fe8cff398544fd72682d68b2461b97ba04881129.

crates/node/core/src/args/txpool.rs
mattsse: yeah, this is just initial block gas limit and we should probably remove this setting

mattsse (approved):

PR #17030, opened 2025-06-24
perf(trie): implement `ParallelSparseTrie::root` method

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/16866

This PR implements the method `ParallelSparseTrie::root` that does the following:
1. Call `ParallelSparseTrie::update_subtrie_hashes` to update the hashes of lower subtries
2. Call `ParallelSparseTrie::update_upper_subtrie_hashes` to update the hashes of upper subtrie.

`ParallelSparseTrie::update_upper_subtrie_hashes` has almost the same functionality as `SparseSubtrie::update_hashes`, but instead of just looking into the upper subtrie, it also checks the lower subtries if the path is long enough and we expect the node to be there. To abstract this common behaviour of calculating the RLP and hashes of nodes, `SparseSubtrieInner::rlp_node` is used.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -242,12 +242,13 @@ impl ParallelSparseTrie {
         self.update_subtrie_hashes();
         
         // Step 2: Prepare the stack with lower subtrie root nodes
-        // Only insert nodes that are actually referenced by the upper subtrie
+        // Only insert nodes whose paths are in the prefix set (need to be updated)
         let mut lower_roots = Vec::new();
+        let mut prefix_set_frozen = core::mem::take(&mut self.prefix_set).freeze();
         
         for subtrie in self.lower_subtries.iter_mut().flatten() {
-            // Only include this subtrie's root if it's referenced in the upper subtrie
-            if self.upper_subtrie.nodes.contains_key(&subtrie.path) {
+            // Only include this subtrie's root if its path is in the prefix set
+            if prefix_set_frozen.contains(&subtrie.path) {
                 // Get the node at the subtrie's root path
                 if let Some(node) = subtrie.nodes.get(&subtrie.path) {
                     // Get the hash of this node (should have been computed by update_hashes)
@@ -288,7 +289,7 @@ impl ParallelSparseTrie {
         // After update_subtrie_hashes, self.prefix_set contains the unchanged keys
         // that don't belong to any lower subtrie - these are exactly the keys
         // that belong to the upper subtrie and need to be updated.
-        let mut prefix_set = core::mem::take(&mut self.prefix_set).freeze();
+        let mut prefix_set = prefix_set_frozen;
         let root_rlp = self.upper_subtrie.update_hashes(&mut prefix_set);
         
         // Step 4: Return the root hash
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -57,19 +57,6 @@ impl Default for ParallelSparseTrie {
 }
 
 impl ParallelSparseTrie {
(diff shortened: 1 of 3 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -226,6 +227,43 @@ impl ParallelSparseTrie {
...
+
+        debug_assert!(self.upper_subtrie.inner.buffers.path_stack.is_empty());
+        self.upper_subtrie.inner.buffers.path_stack.push(RlpNodePathStackItem {
+            path: Nibbles::default(), // Start from root
+            is_in_prefix_set: None,
+        });
+
+        while let Some(stack_item) = self.upper_subtrie.inner.buffers.path_stack.pop() {
+            let path = stack_item.path;
+            let node = if path.len() < UPPER_TRIE_MAX_DEPTH {
```
mediocregopher, line 243: I introduced [subtrie_for_path](https://github.com/paradigmxyz/reth/pull/17035/files#diff-20fbf719d9bb610898f9f3bd9c621f1a581ba2c371da857b3dbcae9ea1a73e26R81) in my `remove_leaf` MR, you could use that here to reduce some logic duplication.

mediocregopher:
nitpick: Maybe rename `update_subtrie_hashes` to `update_lower_subtrie_hashes`, and make it private if it's not needed externally

crates/trie/sparse-parallel/src/trie.rs
shekhirin: unfortunately not, because `self` will then be borrowed mutably fully, and i won't be able to use `self.upper_subtrie.inner.rlp_node` mutably as well 🥲

shekhirin:
@mediocregopher renamed, but still need it public, because we'll be calling this method like do now with `calculate_below_level ` https://github.com/paradigmxyz/reth/blob/3a13f161f8da42bff2095268530396bdadfc690c/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs#L255

mediocregopher (approved):

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -644,12 +677,64 @@ struct SparseSubtrieInner {
...
+    /// # Node Processing Details
+    ///
+    /// - **Empty/Hash nodes**: Return pre-computed or cached RLP immediately
+    /// - **Pre-computed nodes**: If the node hash is already computed, and the node path is not in
+    ///   the prefix set, return the pre-computed RLP
+    /// - **Leaf nodes**: Encode the key-value pair and update the node's hash
+    /// - **Extension nodes**: May need to defer processing until child is computed, on the next
+    ///   invocation update the node's hash.
+    /// - **Branch nodes**: May need to defer processing until all children are computed, on the
+    ///   next invocation update the node's hash.
```
Rjected: I think some abbreviated versions of these would fit well as comments in the branches themselves

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -644,12 +677,64 @@ struct SparseSubtrieInner {
...
+        trace!(
+            target: "trie::parallel_sparse",
+            ?path,
+            ?node,
+            "Calculating node RLP"
+        );
+
+        // Check if the path is in the prefix set.
+        // First, check the cached value. If it's `None`, then check the prefix set, and update
+        // the cached value.        let prefix_set_contains =
```
```suggestion
        // the cached value.
```

Rjected (approved):
lgtm, comments are mostly about docs / comments

crates/trie/sparse-parallel/src/trie.rs
shekhirin: lol

mattsse (commented):
all of this makes sense to me so far

PR #17035, opened 2025-06-24
perf(trie): implement remove_leaf for ParallelSparseTrie

### LINKED ISSUE #16869: Add `remove_leaf` for `ParallelSparseTrie`
### Describe the feature

`SparseTrie` has a method called `remove_leaf` that is used in tests, we should have a similar method in `ParallelSparseTrie` so that we can test it similarly:
https://github.com/paradigmxyz/reth/blob/051cef53bc7372c22d545a583427b0e598529e87/crates/trie/sparse/src/trie.rs#L1719-L1922

### Additional context

_No response_

mediocregopher:
Implements the remove_leaf method for ParallelSparseTrie to match the functionality available in SparseTrie. The method handles leaf removal with proper trie node collapsing and merging, including:

- Branch nodes with single remaining children collapse to extension/leaf nodes
- Extension nodes merge with their children when appropriate
- Proper handling of blinded nodes requiring revelation during collapse
- Value migration between upper and lower subtries when nodes merge
- Comprehensive test coverage for all removal scenarios

Closes #16869


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,426 @@ impl ParallelSparseTrie {
...
+    /// Returns the next node in the traversal path from the given path towards the leaf for the
+    /// given full key, or an error if any node along the traversal path is not revealed.
+    ///
+    /// ## Invariants
+    ///
+    /// - `from_path` must be a prefix of `key`.
+    fn find_next_to_leaf(
+        from_path: &Nibbles,
+        from_node: &SparseNode,
+        key: &Nibbles,
```
shekhirin: let's use `path` when talking about paths to nodes, and `key` when talking about leaf/extension node key that extends the path to the node to get the full path

mediocregopher: This is a reference to what I guess would be called the full path of a leaf, which is what gets used as a key in the `values` HashMap. I wanted to differentiate it from just "path", because it doesn't appear in the `nodes` HashMap so it was getting confusing.

Maybe `leaf_full_path` makes sense?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,426 @@ impl ParallelSparseTrie {
...
+        // and grandparent.
+
+        let leaf_path;
+        let leaf_subtrie;
+
+        let mut branch_parent_path: Option<Nibbles> = None;
+        let mut branch_parent_node: Option<SparseNode> = None;
+
+        let mut ext_grandparent_path: Option<Nibbles> = None;
+        let mut ext_grandparent_node: Option<SparseNode> = None;
```
shekhirin, line 416: grandparent can also be a branch node

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,426 @@ impl ParallelSparseTrie {
...
+                    return Ok(FindNextToLeafOutcome::NotFound)
+                }
+                Ok(FindNextToLeafOutcome::ContinueFrom(child_path))
+            }
+            SparseNode::Branch { state_mask, .. } => {
+                if key.len() == from_path.len() {
+                    return Ok(FindNextToLeafOutcome::NotFound)
+                }
+
+                let nibble = key.get(from_path.len()).unwrap();
```
```suggestion
                let nibble = key.get_unchecked(from_path.len());
```

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: The `ext_grandparent_*` variables will only get set to an extension node grandparent. Additionally there's [some logic](https://github.com/paradigmxyz/reth/pull/17035/files#diff-20fbf719d9bb610898f9f3bd9c621f1a581ba2c371da857b3dbcae9ea1a73e26R439) to ensure that they get unset if a second branch comes in, so if there's a situation like extension->branch->branch the extension variables don't stay set.

crates/trie/sparse-parallel/src/trie.rs
Rjected: `full_path` or something similar makes sense to me

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,433 @@ impl ParallelSparseTrie {
...
+            return;
+        }
+
+        if let SparseNode::Leaf { key, .. } = new_parent_node {
+            let prev_child_subtrie =
+                if let Some(subtrie) = self.lower_subtrie_for_path(prev_child_path) {
+                    subtrie
+                } else {
+                    return;
+                };
```
```suggestion
                let Some(subtrie) = self.lower_subtrie_for_path(prev_child_path) else {
                    return
                };
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,433 @@ impl ParallelSparseTrie {
...
+        debug_assert!(remaining_child_path.len() > parent_path.len());
+        debug_assert!(remaining_child_path.starts_with(parent_path));
+
+        let remaining_child_nibble = remaining_child_path.get_unchecked(parent_path.len());
+
+        // If we swap the branch node out either an extension or leaf, depending on
+        // what its remaining child is.
+        match remaining_child_node {
+            SparseNode::Empty | SparseNode::Hash(_) => {
+                unreachable!("remaining child must have been revealed already")
```
Rjected: is this actually unreachable? Or should it just panic?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,433 @@ impl ParallelSparseTrie {
...
+    }
+
+    /// Given the path to a parent branch node and a child node which is the sole remaining child on
+    /// that branch after removing a leaf, returns a node to replace the parent branch node and a
+    /// boolean indicating if the child should be deleted.
+    ///
+    /// ## Invariants
+    ///
+    /// - Both parent and child nodes must be already revealed.
+    /// - The parent's path must be a prefix of the child's path.
```
Rjected: should note panic conditions

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,433 @@ impl ParallelSparseTrie {
...
+    }
+
+    /// Given the path to a parent extension and its key, and a child node (not necessarily on this
+    /// subtrie), returns an optional replacement parent node. If a replacement is returned then the
+    /// child node should be deleted.
+    ///
+    /// ## Invariants
+    ///
+    /// - Both parent and child nodes must be already revealed.
+    /// - The parent's path must be a prefix of the child's path.
```
Rjected: should also note panic conditions here

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -175,23 +193,433 @@ impl ParallelSparseTrie {
...
+                            )?;
+                            remaining_child_subtrie.nodes.get(&remaining_child_path).unwrap()
+                        } else {
+                            return Err(SparseTrieErrorKind::Other(Box::<
+                                dyn core::error::Error + Send + Sync,
+                            >::from(
+                                format!("remaining child node {remaining_child_path:?} not found in provider")
+                            ))
+                            .into())
+                        }
```
Rjected: in the regular sparse trie we don't return an error here, not sure that we need to here. But if we do, there should be an easier way to convert into the other error variant

Rjected (commented):
Mostly style nits, I also think the name `find_next_to_leaf` could be better, but I can't think of anything better

crates/trie/sparse-parallel/src/trie.rs
shekhirin: agreed on `full_path`

mediocregopher: Renamed it to `leaf_full_path` everywhere

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: It's unreachable based on the invariants documented on the method. Basically it would be improper to pass Empty/Hash into the method, and since it's a private method we can say that it if it does occur it's a bug within the crate. It seemed to me that was what `unreachable!` is for.

Maybe that's not the right way to think about it though, I don't mind changing it if so.

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: In the regular sparse trie we would end up returning a `SparseTrieErrorKind::BlindedNode` if `blinded_node` returned `None`, which seemed confusing to me, since we'd expect there to always be a corresponding entry for the hashed node (I think?)

wdyt about adding a new variant to `SparseTrieErrorKind`? I added one in my recent commit to show the idea.

Rjected: the new error variant makes sense

crates/trie/sparse-parallel/src/trie.rs
Rjected: ~~I would prefer panic here, I like `unreachable` more when it's explicitly checked in the fn body, or `assert`ed~~

Actually, in this case, I can see it, `unreachable` is fine here

Rjected (approved):
lgtm

PR #17044, opened 2025-06-24
chore: use earliest block number

mattsse:
now that we can expiry history, we should use this value for the `earliest` block


joshieDo (approved):
lgtm, i wonder though if this can become some kind of footgun, since headers are never expired

PR #17046, opened 2025-06-24
chore: replace revm_utils with alloy_evm helpers

### LINKED ISSUE #17000: Reuse new alloy-evm utils
### Describe the feature

alloy-evm now has additional utils for everything in

https://github.com/paradigmxyz/reth/blob/a0c3bbf92083ca54bdab4841b754c7b5370b112e/crates/rpc/rpc-eth-types/src/revm_utils.rs#L1-L1

that we can reuse

## TODO
* remove revm_utils and replace with alloy-evm helpers, will require some from conversions for the new error types

### Additional context

_No response_

iTranscend:
resolves https://github.com/paradigmxyz/reth/issues/17000


crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -759,7 +763,9 @@ pub trait Call:
             apply_block_overrides(*block_overrides, db, &mut evm_env.block_env);
         }
         if let Some(state_overrides) = overrides.state {
-            apply_state_overrides(state_overrides, db)?;
+            apply_state_overrides(state_overrides, db).map_err(|e| {
+                <EthApiError as From<alloy_evm::overrides::StateOverrideError<_>>>::from(e)
```
mattsse: can we also use Self::Error::from_eth_err here?

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -770,7 +776,8 @@ pub trait Call:
             if tx_env.gas_price() > 0 {
                 // If gas price is specified, cap transaction gas limit with caller allowance
                 trace!(target: "rpc::eth::call", ?tx_env, "Applying gas limit cap with caller allowance");
-                let cap = caller_gas_allowance(db, &tx_env)?;
+                let cap = caller_gas_allowance(db, &tx_env)
+                    .map_err(<EthApiError as From<alloy_evm::call::CallError<_>>>::from)?;
```
mattsse: same q

mattsse (changes requested):
great, only 1 q re error mapping

crates/rpc/rpc-eth-api/src/helpers/call.rs
iTranscend: with the current bounds, rustc can't seem to figure out what `From` impl to use when using `Self::Error::from_eth_err`

It's prioritizing the `From` at https://github.com/paradigmxyz/reth/blob/b1e91ea2698f39de18151f04f9a3ee3860809e92/crates/rpc/rpc-eth-api/src/helpers/call.rs#L736-L739

wdyt?

mattsse (approved):
very nice,

I converted these verbose err maps into helper fns to make this a bit simpler

PR #17050, opened 2025-06-25
feat: make jwt-secret argument consistent across reth-bench commands

shekhirin:
Previously, reth-bench commands used inconsistent argument names:
- `new-payload-fcu` and `new-payload-only` used `--jwtsecret`
- `send-payload` used `--jwt-secret`

This change standardizes all commands to use `--jwt-secret` with hyphens for consistency with other CLI tools.

### DIFF
--- a/crates/node/core/src/args/benchmark_args.rs
+++ b/crates/node/core/src/args/benchmark_args.rs
@@ -21,7 +21,7 @@ pub struct BenchmarkArgs {
     ///
     /// If no path is provided, a secret will be generated and stored in the datadir under
     /// `<DIR>/<CHAIN_ID>/jwt.hex`. For mainnet this would be `~/.reth/mainnet/jwt.hex` by default.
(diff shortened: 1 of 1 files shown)


klkvr (approved):

crates/node/core/src/args/benchmark_args.rs
```diff
@@ -21,7 +21,7 @@ pub struct BenchmarkArgs {
     ///
     /// If no path is provided, a secret will be generated and stored in the datadir under
     /// `<DIR>/<CHAIN_ID>/jwt.hex`. For mainnet this would be `~/.reth/mainnet/jwt.hex` by default.
-    #[arg(long = "jwtsecret", value_name = "PATH", global = true, required = false)]
+    #[arg(long = "jwt-secret", value_name = "PATH", global = true, required = false)]
```
mattsse: could we keep the jwtsecret as alias so that this doesnt break existing setups`

PR #17052, opened 2025-06-25
feat(trie): add assert_eq_parallel_sparse_trie_proof_nodes

### LINKED ISSUE #17041: Add `assert_eq_parallel_sparse_trie_proof_nodes`
### Describe the feature

In the regular sparse trie, we have the following method that checks for equivalence between the sparse trie structure and the proof nodes returned from the hash builder:
https://github.com/paradigmxyz/reth/blob/f7a7112330d3e36de99fe7b46e8850198682290c/crates/trie/sparse/src/trie.rs#L2612-L2656

We should create a new method, `assert_eq_parallel_sparse_trie_proof_nodes`, that also checks that the parallel sparse trie structure contains the structure represented by the proof nodes.

This requires doing a check across the upper and lower subtries

### Additional context

_No response_

i-m-aditya:
resolves #17041

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -1241,17 +1241,15 @@ mod tests {
             .lower_subtries
             .iter()
             .filter_map(Option::as_ref)
-            .flat_map(|subtrie| subtrie.nodes.iter())
-            .sorted_by_key(|(path, _)| *path);
-        let upper_sparse_nodes = sparse_trie.upper_subtrie.nodes.iter().collect::<Vec<_>>();
+            .flat_map(|subtrie| subtrie.nodes.iter());
 
-        let sparse_nodes = lower_sparse_nodes
-            .into_iter()
-            .chain(upper_sparse_nodes)
-            .sorted_by_key(|(path, _)| *path);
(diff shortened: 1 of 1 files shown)


shekhirin (commented):
LGTM, just needs some clippy fixes

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1241,14 +1241,13 @@ mod tests {
             .lower_subtries
             .iter()
             .filter_map(Option::as_ref)
-            .map(|subtrie| subtrie.nodes.iter().collect::<Vec<_>>())
-            .flatten()
-            .collect::<Vec<_>>();
+            .flat_map(|subtrie| subtrie.nodes.iter())
+            .sorted_by_key(|(path, _)| *path);
```
shekhirin: no need to sort here, as it'll be sorted below

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1241,14 +1241,13 @@ mod tests {
...
         let upper_sparse_nodes = sparse_trie.upper_subtrie.nodes.iter().collect::<Vec<_>>();
```
```suggestion
        let upper_sparse_nodes = sparse_trie.upper_subtrie.nodes.iter();
```

shekhirin (approved):
LGTM

PR #17058, opened 2025-06-25
feat: add per-address metrics for precompile cache

shekhirin:
Record cache hit/miss metrics separately for each precompile address.

Changes:
- Create metrics instances with address labels using new_with_labels
- Remove shared metrics field, create per-address instances
- Update both prewarm and main execution integration points

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -10,7 +10,7 @@ use crate::{
 use alloy_consensus::BlockHeader;
 use alloy_eips::{merge::EPOCH_SLOTS, BlockNumHash, NumHash};
 use alloy_evm::block::BlockExecutor;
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2443,11 +2443,16 @@ where
 
         if !self.config.precompile_cache_disabled() {
             executor.evm_mut().precompiles_mut().map_precompiles(|address, precompile| {
+                let metrics = self
+                    .precompile_cache_metrics
+                    .entry(*address)
+                    .or_insert_with(|| CachedPrecompileMetrics::new_with_address(*address))
+                    .clone();
```
mattsse, line 2450: I see, this will lazily init them and then reuse

mattsse (approved):

PR #17062, opened 2025-06-25
docs: error fixes for clarity

sky-coderay:
Description:

crates/engine/invalid-block-hooks/src/witness.rs
`deterministric` - `deterministic` -- fix typo


mattsse (approved):
you again

PR #17089, opened 2025-06-26
chore: make SparseStateTrie generic with respect to SparseTrie implementation

### LINKED ISSUE #17088: Make SparseStateTrie type generic with respect to trie implementation
### Describe the feature

We want to have the ability to use the serial sparse trie in some circumstances and the parallel in others, for example based on feature/command-line flag.

In order to accomplish this we need to pull the methods belonging to RevealedSparseTrie which are used by SparseStateTrie into a trait, and have SparseStateTrie accept that trait as a generic parameter. For now we will continue using the existing serial sparse trie implementation; this is only a refactoring in preparation for future changes.

### Additional context

_No response_

mediocregopher:
This refactoring introduces generics that allow using different trie implementations while maintaining the same interface.

**Changes made:**

- Renamed `RevealedSparseTrie` to `SerialSparseTrie` to better contrast it against the `ParallelSparseTrie`, and to make room for a trait encompassing both.

- Extracted revealed trie functionality into a new `RevealedSparseTrie` trait that abstracts over different sparse trie implementations.

- Made `SparseStateTrie` generic over both account and storage trie types (`R` and `S`), defaulting to `SerialSparseTrie`.

Fixes #17088

### DIFF
--- a/crates/trie/sparse/src/trie.rs
+++ b/crates/trie/sparse/src/trie.rs
@@ -52,7 +52,7 @@ impl TrieMasks {
     }
 }
 
-/// A struct for keeping the hashmaps from `RevealedSparseTrie`.
+/// A struct for keeping the hashmaps from `SerialSparseTrie`.
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct SparseTrieState {
     /// Map from a path (nibbles) to its corresponding sparse trie node.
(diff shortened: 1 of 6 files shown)


crates/trie/sparse/src/trie.rs
mediocregopher, line 1: It's hard to verify by looking at it, but there's no logic changes in here, only moving methods around and adding generic parameters to impls. Moving methods was required primarily so that `SerialSparseTrie` could implement the `RevealedSparseTrie` trait, and so all methods involved had to get moved into that impl block.

mediocregopher:
This work has been superceded by  #17166 and is going to be done in a more piecemeal way (though the idea remains roughly the same). Closing for now

PR #17092, opened 2025-06-26
feat(rpc): Use generic transaction request as input

### LINKED ISSUE #15117: Feature: Allow `RpcAddOns` to accept custom transaction requests

RomanHodulak:
Closes #17051

Depends on `alloy` features alloy-rs/alloy#2622 alloy-rs/alloy#2623 alloy-rs/alloy#2624 alloy-rs/alloy#2631 alloy-rs/alloy#2662 alloy-rs/alloy#2674 alloy-rs/alloy#2677

The work so far associated with #15117 paved a road to implement the generic input transaction request type that is introduced here.

This allows for custom RPC request bodies in custom nodes while reusing all the default components. To make this possible a `TransactionBuilder` trait is used as a bound for the generic transaction request type.

### DIFF
--- a/crates/net/network-api/src/noop.rs
+++ b/crates/net/network-api/src/noop.rs
@@ -21,7 +21,6 @@ use reth_eth_wire_types::{
 use reth_network_p2p::{sync::NetworkSyncUpdater, NoopFullBlockClient};
 use reth_network_peers::NodeRecord;
 use reth_network_types::{PeerKind, Reputation, ReputationChangeKind};
-use reth_rpc_convert::RpcTypes;
 use reth_tokio_util::{EventSender, EventStream};
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::UnboundedReceiverStream;
@@ -218,11 +217,4 @@ where
     fn peers_handle(&self) -> &PeersHandle {
         &self.peers_handle
     }
-}
-
-impl RpcTypes for NoopNetwork<EthNetworkPrimitives> {
-    type Header = alloy_consensus::Header;
-    type TransactionRequest = alloy_rpc_types_eth::transaction::TransactionRequest;
-    type TransactionResponse = alloy_rpc_types_eth::Transaction;
-    type Receipt = alloy_rpc_types_eth::TransactionReceipt;
 }
--- a/crates/net/network-api/src/noop.rs
+++ b/crates/net/network-api/src/noop.rs
@@ -6,6 +6,13 @@
 use core::{fmt, marker::PhantomData};
 use std::net::{IpAddr, SocketAddr};
 
+use crate::{
+    events::{NetworkPeersEvents, PeerEventStream},
+    test_utils::{PeersHandle, PeersHandleProvider},
+    BlockDownloaderProvider, DiscoveryEvent, NetworkError, NetworkEvent,
+    NetworkEventListenerProvider, NetworkInfo, NetworkStatus, PeerId, PeerInfo, PeerRequest, Peers,
+    PeersInfo,
+};
 use alloy_rpc_types_admin::EthProtocolInfo;
 use enr::{secp256k1::SecretKey, Enr};
 use reth_eth_wire_types::{
@@ -17,14 +24,6 @@ use reth_network_types::{PeerKind, Reputation, ReputationChangeKind};
 use reth_tokio_util::{EventSender, EventStream};
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::UnboundedReceiverStream;
-
-use crate::{
-    events::{NetworkPeersEvents, PeerEventStream},
-    test_utils::{PeersHandle, PeersHandleProvider},
-    BlockDownloaderProvider, DiscoveryEvent, NetworkError, NetworkEvent,
-    NetworkEventListenerProvider, NetworkInfo, NetworkStatus, PeerId, PeerInfo, PeerRequest, Peers,
-    PeersInfo,
-};
 
 /// A type that implements all network trait that does nothing.
 ///
--- a/crates/net/network-api/src/noop.rs
+++ b/crates/net/network-api/src/noop.rs
@@ -21,6 +21,7 @@ use reth_eth_wire_types::{
 use reth_network_p2p::{sync::NetworkSyncUpdater, NoopFullBlockClient};
 use reth_network_peers::NodeRecord;
 use reth_network_types::{PeerKind, Reputation, ReputationChangeKind};
+use reth_rpc_convert::RpcTypes;
 use reth_tokio_util::{EventSender, EventStream};
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::UnboundedReceiverStream;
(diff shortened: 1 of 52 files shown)


crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -521,7 +524,7 @@ pub trait Call:
     /// Executes the call request at the given [`BlockId`].
     fn transact_call_at(
         &self,
-        request: TransactionRequest,
+        request: RpcTxReq<<Self::RpcConvert as RpcConvert>::Network>,
```
mattsse, line 527: we could maybe introduce an alias for this?

crates/rpc/rpc-eth-types/src/simulate.rs
```diff
@@ -126,65 +127,62 @@ where
...
     }
 
     // if we can't build the _entire_ transaction yet, we need to check the fee values
-    if tx.buildable_type().is_none() {
-        match tx_type {
-            TxType::Legacy | TxType::Eip2930 => {
-                if tx.gas_price.is_none() {
-                    tx.gas_price = Some(block_base_fee_per_gas as u128);
-                }
+    if tx.output_tx_type_checked().is_none() {
```
mattsse: need to double check if these two fn behave the same

mattsse (commented):
pretty cool
crucial feature for enabling custom tx types

we do need a new alloy patch for the simulate endpoint, but can do that as a followup

crates/rpc/rpc-eth-types/src/simulate.rs
```diff
@@ -126,65 +131,62 @@ where
...
         Primitives: NodePrimitives<SignedTx = Tx>,
-        Network: RpcTypes<TransactionRequest: From<TransactionRequest>>,
+        Network: alloy_network::Network
+                     + RpcTypes<TransactionRequest: TransactionBuilder<<T as RpcConvert>::Network>>,
     >,
 {
     // If we're missing any fields we try to fill nonce, gas and
     // gas price.
-    let tx_type = tx.preferred_type();
+    let tx_type = tx.output_tx_type();
```
mattsse: this is equivalent, so this is fine

crates/rpc/rpc-eth-types/src/simulate.rs
mattsse: it is

mattsse (commented):
I think all of this looks good,

there are likely a few things we can cleanup after, like the signer type

crates/net/network-api/src/noop.rs
```diff
@@ -219,3 +219,10 @@ where
         &self.peers_handle
     }
 }
+
+impl RpcTypes for NoopNetwork<EthNetworkPrimitives> {
+    type Header = alloy_consensus::Header;
+    type TransactionRequest = alloy_rpc_types_eth::transaction::TransactionRequest;
+    type TransactionResponse = alloy_rpc_types_eth::Transaction;
+    type Receipt = alloy_rpc_types_eth::TransactionReceipt;
+}
```
klkvr: why do we need this?

crates/optimism/rpc/src/eth/call.rs
```diff
@@ -39,7 +39,10 @@ where
                 >,
             >,
             RpcConvert: RpcConvert<TxEnv = TxEnvFor<Self::Evm>, Network = Self::NetworkTypes>,
-            NetworkTypes: RpcTypes<TransactionRequest: From<TransactionRequest>>,
+            NetworkTypes: op_alloy_network::Network
+                              + RpcTypes<
+                TransactionRequest: TransactionBuilder<<Self::RpcConvert as RpcConvert>::Network>,
+            >,
```
klkvr: could we avoid requiring `Network` here?

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -62,17 +63,17 @@ impl<T> OpNodeCore for T where T: RpcNodeCore<Provider: BlockReader> {}
 /// This type implements the [`FullEthApi`](reth_rpc_eth_api::helpers::FullEthApi) by implemented
 /// all the `Eth` helper traits and prerequisite traits.
 #[derive(Clone)]
-pub struct OpEthApi<N: OpNodeCore, NetworkT = Optimism> {
+pub struct OpEthApi<N: OpNodeCore, NetworkT: op_alloy_network::Network = Optimism> {
```
klkvr: `Network` bound is not very convinient here because it forces all custom nodes to define not just RPC types but consensus type as well. What do we need to not require it?

crates/rpc/rpc-eth-types/src/simulate.rs
RomanHodulak: Thanks for checking

crates/optimism/rpc/src/eth/mod.rs
RomanHodulak: I see, should be droppable. It stems from here https://github.com/paradigmxyz/reth/blob/ba7c15d0b77a863c8367f6d01a5d57484f9fb04f/crates/rpc/rpc/src/eth/core.rs#L273

https://github.com/paradigmxyz/reth/blob/ba7c15d0b77a863c8367f6d01a5d57484f9fb04f/crates/rpc/rpc-eth-api/src/helpers/spec.rs#L112

https://github.com/paradigmxyz/reth/blob/b011ad0d8d63baad0ae211c822a61ab17006014f/crates/rpc/rpc-convert/src/rpc.rs#L32

crates/optimism/rpc/src/eth/call.rs
RomanHodulak: I think that's required by `TransactionBuilder<N: Network>`

klkvr (approved):

PR #17097, opened 2025-06-27
feat: add graph selection option to newpayload latency comparison script

mediocregopher:
Allow users to choose which graphs to generate when comparing newpayload latency between different reth versions. The `--graphs` option accepts:
- `histogram` - shows distribution of percent differences
- `line` - shows latency over time/block number
- `all` - generates both graphs (default behavior)

**Additional improvements:**
- Dynamic color coding for comparison line (green if faster, red if slower)
- Changed baseline color from blue to orange for better contrast


mattsse (approved):
lgtm

PR #17098, opened 2025-06-27
feat: Add StatelessTrie trait for `reth-stateless`

kevaundray:
The current implementation of SparseStateTrie has been shown to be in-efficient for zkVMs -- while this is the case, this PR creates a trait such that downstream users can switch out the trie while keeping all of the other logic

### DIFF
--- a/crates/stateless/src/trie.rs
+++ b/crates/stateless/src/trie.rs
@@ -14,7 +14,7 @@ use reth_trie_sparse::{
 };
 
 /// Trait for stateless trie implementations that can be used for stateless validation.
-pub trait StatelessTrieTrait: core::fmt::Debug {
+pub trait StatelessTrie: core::fmt::Debug {
     /// Initialize the stateless trie using the `ExecutionWitness`
     fn new(
         witness: &ExecutionWitness,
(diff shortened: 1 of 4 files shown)


crates/stateless/src/validation.rs
```diff
@@ -133,6 +137,31 @@ pub fn stateless_validation<ChainSpec, E>(
...
     ChainSpec: Send + Sync + EthChainSpec + EthereumHardforks + Debug,
     E: ConfigureEvm<Primitives = EthPrimitives> + Clone + 'static,
+{
+    stateless_validation_with_trie::<StatelessTrie, ChainSpec, E>(
+        current_block,
+        witness,
+        chain_spec,
+        evm_config,
+    )
+}
```
kevaundray: This wrapper method means that this does not break downstream users who were using the previous API.

crates/stateless/src/trie.rs
```diff
@@ -99,6 +128,30 @@ impl StatelessTrie {
     }
 }
 
+impl StatelessTrieTrait for StatelessTrie {
```
rkrasiuk:
```suggestion
impl StatelessTrie for StatelessSparseTrie {
```

rkrasiuk (approved):

PR #17099, opened 2025-06-27
chore: pass provider into SparseTrie and SparseStateTrie via impl argument in update/remove_leaf

mediocregopher:
Working towards #17088

Only the `remove_leaf` and `update_leaf` methods of the RevealedSparseTrie actually use the Provider. This PR takes the Provider off the trie struct and passes it into those methods as an argument.

By not keeping the Provider on the trie itself it makes the trie into a pure datastructure, opening up our options in terms of re-using it across payloads.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -38,7 +38,7 @@ where
     /// Sparse Trie initialized with the blinded provider factory.
     ///
     /// It's kept as a field on the struct to prevent blocking on de-allocation in [`Self::run`].
-    pub(super) trie: SparseStateTrie<BPF>,
+    pub(super) trie: SparseStateTrie,
     pub(super) metrics: MultiProofTaskMetrics,
     /// Blinded node provider factory.
     blinded_provider_factory: BPF,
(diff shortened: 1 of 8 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -38,13 +38,15 @@ where
     /// Sparse Trie initialized with the blinded provider factory.
     ///
     /// It's kept as a field on the struct to prevent blocking on de-allocation in [`Self::run`].
-    pub(super) trie: SparseStateTrie<BPF>,
```
mattsse, line 41: nice!
I like this a lot

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -129,19 +132,24 @@ where
                 "Updating sparse trie"
             );
 
-            let elapsed = update_sparse_trie(&mut self.trie, update).map_err(|e| {
-                ParallelStateRootError::Other(format!("could not calculate state root: {e:?}"))
-            })?;
+            let elapsed =
+                update_sparse_trie(&mut self.trie, update, &self.blinded_provider_factory)
+                    .map_err(|e| {
+                        ParallelStateRootError::Other(format!(
```
mattsse, line 138: idea for follow up:

make other a `Box<dyn Error>` and add `ParallelStateRootError::msg(err: Display)`

mattsse (approved):
lgtm, pending @shekhirin @Rjected

shekhirin (approved):
very nice

Rjected (approved):
lgtm

PR #17100, opened 2025-06-27
docs: debug clarify healtyh node rpc url setting

mattsse:


crates/node/core/src/args/debug.rs
```diff
@@ -80,6 +80,11 @@ pub struct DebugArgs {
     pub invalid_block_hook: Option<InvalidBlockSelection>,
 
     /// The RPC URL of a healthy node to use for comparing invalid block hook results against.
+    ///
+    ///Debug setting that enables execution witness comparison for troubleshooting bad blocks.
```
mediocregopher, line 84:
```suggestion
    /// Debug setting that enables execution witness comparison for troubleshooting bad blocks.
```

mediocregopher (approved):
LGTM sans one super duper nitpick

PR #16816, opened 2025-06-14
feat: relax EthereumAddons trait bounds to support generic validators

mattsse:
Similar to #16615 for OpAddOns, this relaxes the trait bounds on EthereumAddons to support custom engine validators and API builders. Also updates TestNode in exex test-utils to match the new generic structure.

## Summary

This PR makes `EthereumAddons` more flexible by:
- Adding generic type parameters for the ETH API builder, engine validator, and engine API builder
- Relaxing trait bounds to use `EthChainSpec + EthereumHardforks` instead of hardcoded `ChainSpec`
- Making `EthereumEngineValidatorBuilder` generic over `ChainSpec`
- Adding methods `with_engine_api()` and `with_engine_validator()` to allow customization

## Changes

- Modified `EthereumAddons` to have generic parameters `<N, EthB, EV, EB>`
- Updated `TestNode` in exex test-utils to match the new structure
- Added necessary imports and trait bounds

This allows downstream users to customize validators while reusing EthereumAddons, avoiding code duplication.

fyi @rezbera

PR #17108, opened 2025-06-27
feat: remove redundant generic from `EthereumEngineValidatorBuilder`

klkvr:


mattsse (commented):
I believe we added this for bera

https://github.com/paradigmxyz/reth/pull/16816

so that they can reuse it

https://github.com/berachain/bera-reth/blob/413c4be403dfadc3c127131888412fc2e86602d2/src/node/mod.rs#L115

mattsse (approved):

PR #17109, opened 2025-06-27
feat: Add `From<ChainConfig> for ChainSpec`

kevaundray:
Adds a From impl for ChainSpec from ChainConfig.

This allows us to convert a `ChainConfig` into a `ChainSpec`. Once the `ChainConfig` struct is serializable, we will be able to use that when sending data from the host to the guest in a zkVM context. The PR for that is here: https://github.com/alloy-rs/alloy/pull/2630

### DIFF
--- a/crates/chainspec/src/spec.rs
+++ b/crates/chainspec/src/spec.rs
@@ -17,7 +17,7 @@ use alloy_consensus::{
 use alloy_eips::{
     eip1559::INITIAL_BASE_FEE, eip7685::EMPTY_REQUESTS_HASH, eip7892::BlobScheduleBlobParams,
 };
-use alloy_genesis::Genesis;
+use alloy_genesis::{ChainConfig, Genesis};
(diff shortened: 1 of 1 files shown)


crates/chainspec/src/spec.rs
```diff
@@ -637,22 +637,33 @@ impl ChainSpec {
 
 impl From<Genesis> for ChainSpec {
     fn from(genesis: Genesis) -> Self {
+        let mut chain_spec = ChainSpec::from(genesis.config.clone());
+        chain_spec.genesis_header =
+            SealedHeader::new_unhashed(make_genesis_header(&genesis, &chain_spec.hardforks));
+        chain_spec.genesis = genesis;
+
+        chain_spec
+    }
```
kevaundray: Main change was moving the config related components from here into the new From<ChainConfig> impl

kevaundray:
spoke with @mattsse and we can instead serialize the ChainConfig and then construct a Genesis structure from the config

PR #17122, opened 2025-06-28
polishing docs and comments

zeroprooff:
two fixes:

in _crates/net/eth-wire/src/eth_snap_stream.rs_
`snao` - `snap` - fix typo

in _crates/net/network/src/session/conn.rs_
`Returns__access` - `Returns _access` deleted double space


mattsse (commented):
superseded

PR #17123, opened 2025-06-28
feat: made bench command operate on PayloadTypes

### LINKED ISSUE #17087: Make bench command operate on PayloadTypes
### Describe the feature

we currently do this:

https://github.com/paradigmxyz/reth/blob/d635035be72d5fc63329b66e9fa40365a2c8b05a/bin/reth-bench/src/bench/new_payload_fcu.rs#L48-L72

this is problematic if we want to use this for both op and eth

instead this should be converted into a command that accepts a closure for the conversion, or some other helper trait, simular to

https://github.com/paradigmxyz/reth/blob/d635035be72d5fc63329b66e9fa40365a2c8b05a/crates/node/builder/src/launch/debug.rs#L44-L44

because we need special block -> payload conversion

## TODO
* introduce rpcblock -> payload conversion helper for this entrypoint https://github.com/paradigmxyz/reth/blob/d635035be72d5fc63329b66e9fa40365a2c8b05a/bin/reth-bench/src/bench/mod.rs#L48-L48
this can be a closure that takes the rpcblock and converts to `Payload`, hence the command also needs a NodeTypes generic hint

### Additional context

_No response_

Soubhik-10:
closes #17087


mattsse (commented):
superseded by #17351

which turned out to be easier, sorry about this @Soubhik-10

PR #17125, opened 2025-06-28
Replace Book with Docs references

PixelPil0t1:
Old word: "Book" -> New word: "Docs"
Old link: https://github.com/paradigmxyz/reth/tree/main/book -> New link: https://github.com/paradigmxyz/reth/tree/main/docs
Why these changes are needed:
The documentation has been moved from the book directory to the docs directory, so we need to update the references in CONTRIBUTING.md to point to the correct location. This ensures contributors can find the latest documentation.
I believe this is a good replacement as it aligns with the recent documentation restructuring. If you have a better suggestion for the documentation reference, please let me know and I'll update accordingly.


mattsse (approved):
ty

PR #17132, opened 2025-06-29
feat(era1): add subcommand `export-era`

### LINKED ISSUE #15495: Add support for exporting history as era store
### Describe the feature

once we have #15371 we need a way to generate an era store with the current database.

for this we simply need to read all the tables and pipe these into an era writer.

## TODO
* add new command and impl for `era export <>`

### Additional context

This approach can be similar to https://github.com/paradigmxyz/reth/pull/15737
but we only need a regular provider that reads block data and writes them to an era file

ref: https://github.com/ethereum/go-ethereum/blob/9b4eab6a29704f55fa7b4b92e296094f0dbcee22/cmd/utils/cmd.go#L403-L405

this can accept a config struct for first,last,path etc.
in a separate step we can add a cli command that then invokes this

lean-apple:
(Really) Closes https://github.com/paradigmxyz/reth/issues/15495. 

Still to do: 
- [x] add command doc
- [x] test more with import
- [x] fix export file name

### How to test
```shell
cargo run --bin reth -- export-era
cargo run --bin reth -- export-era --path ./my-era1-exports

RUST_LOG=debug  cargo run --bin reth -- import-era --chain sepolia
# then
cargo run --bin reth -- export-era --chain sepolia --first-block-number 201 --last-block-number 300

#or with --max-blocks-per-file
cargo run --bin reth -- export-era \
  --chain sepolia \
  --first-block-number 0 \
  --last-block-number 100 \
  --max-blocks-per-file 10
```

@mattsse, @RomanHodulak, in this pr I followed the same subcommand model than `import-era`, and propose `export-era`,
 
Yet in the initial [issue](https://github.com/paradigmxyz/reth/issues/15495), it was mentioned to have a subcommand `era`, and so it would do  `reth era import`, `reth era export`,  and I think it would be more intuitive to have subcommand `reth era1`,  (and later) `reth era` for the post-merge feature, wdyt ? 

I can submit another pr to have this before this one.


crates/cli/commands/src/export_era.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> ExportEraCommand<C> {
+    /// Execute `export-era` command
+    pub async fn execute<N>(self) -> eyre::Result<()>
+    where
+        N: CliNodeTypes<ChainSpec = C::ChainSpec>,
+        <<N as NodeTypes>::Primitives as NodePrimitives>::BlockHeader: Into<Header>,
+        <<N as NodeTypes>::Primitives as NodePrimitives>::Block:
+            Into<BlockBody<<<N as NodeTypes>::Primitives as NodePrimitives>::SignedTx, Header>>,
```
mattsse: hmm, why do we need these conversions?

crates/cli/commands/src/export_era.rs
```diff
@@ -0,0 +1,114 @@
...
+                .join(ERA1_EXPORT_FOLDER_NAME),
+        };
+
+        let export_config = era1::ExportConfig {
+            network: self.env.chain.chain().to_string(),
+            first_block_number: self.export.first_block_number.unwrap_or(0),
+            last_block_number: self
+                .export
+                .last_block_number
+                .unwrap_or(MAX_BLOCKS_PER_ERA1 as u64 - 1),
```
mattsse, line 70: this feels a bit weird, does this default to a single era file?

crates/era-utils/src/export.rs
```diff
@@ -135,7 +142,20 @@ where
...
         let headers = provider.headers_range(start_block..=end_block)?;
 
-        let era1_id = Era1Id::new(&config.network, start_block, block_count as u32);
+        // Extract first 4 bytes of last block's state root as historical identifier
+        let historical_root = headers
+            .last()
+            .map(|header| {
+                let header_converted: Header = header.clone().into();
+                let state_root = header_converted.state_root;
+                [state_root[0], state_root[1], state_root[2], state_root[3]]
```
mattsse: I think we can use the BlockHeader trait here for state_root()

crates/era-utils/src/export.rs
```diff
@@ -69,6 +69,13 @@ impl ExportConfig {
             return Err(eyre!("Max blocks per file cannot be zero"));
         }
 
+        if self.network != "mainnet" && self.network != "sepolia" {
+            return Err(eyre!(
+                "Network '{}' is not supported. Only 'mainnet' and 'sepolia' are allowed",
+                self.network
+            ));
+        }
```
mattsse: do we need this check because these are the only ones that should support era1?

crates/cli/commands/src/export_era.rs
```diff
@@ -0,0 +1,114 @@
...
+use reth_era_utils as era1;
+use reth_node_api::NodeTypes;
+use reth_primitives_traits::NodePrimitives;
+use reth_provider::DatabaseProviderFactory;
+use std::{path::PathBuf, sync::Arc};
+use tracing::info;
+
+const ERA1_EXPORT_FOLDER_NAME: &str = "era1-export";
+
+#[derive(Debug, Parser)]
```
mattsse, line 16: needs one line doc

mattsse (changes requested):
cool, some questions

crates/cli/commands/src/export_era.rs
lean-apple: For the [`export`](https://github.com/paradigmxyz/reth/blob/fcf58cb5acc2825e7c046f6741e90a8c5dab7847/crates/era-utils/src/export.rs#L79) fn, I didn't find a better way to make it simpler or more readable

crates/era-utils/src/export.rs
lean-apple: I dont think goerli is supported for `era1` files (https://eth-clients.github.io/history-endpoints/) - only sepolia as "pre-merge" testnet.

crates/cli/commands/src/export_era.rs
lean-apple: Kind of, it's more a default `first/last` block number duo with `0/8191` to have the 8192 max blocks per file

crates/era-utils/src/export.rs
lean-apple: thanks

crates/cli/commands/src/export_era.rs
lean-apple: Solved with https://github.com/paradigmxyz/reth/pull/17312.

mattsse (approved):

PR #17138, opened 2025-06-29
feat(chainspec): move Holocene next-block base-fee logic into OpChainSpec

### LINKED ISSUE #17090: Move holocene next block base fee handling into opChainspec
### Describe the feature

after #16927 we can move 

https://github.com/paradigmxyz/reth/blob/ad8fec466156245afb1fc6cbd1afed6dcbfc8b7b/crates/optimism/consensus/src/validation/mod.rs#L189-L206

into `next_block_base_fee` of OpChainSpec so that post holocene decoding is respected

## TODO
* Call this fn from opchainspec's fn

### Additional context

_No response_

rose2221:
fixes #17090

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -283,9 +283,6 @@ split-debuginfo = "unpacked"
 # Speed up tests.
 [profile.dev.package]
 proptest.opt-level = 3
-rand_chacha.opt-level = 3
-rand_xorshift.opt-level = 3
-unarray.opt-level = 3
 
 # Meant for testing - all optimizations, but with debug assertions and overflow checks.
 [profile.hivetests]
@@ -318,12 +315,7 @@ lto = "fat"
 codegen-units = 1
 
 [workspace.dependencies]
-# reth
-op-reth = { path = "crates/optimism/bin" }
-reth = { path = "bin/reth" }
-reth-alloy-provider = { path = "crates/alloy-provider" }
 reth-basic-payload-builder = { path = "crates/payload/basic" }
-reth-bench = { path = "bin/reth-bench" }
 reth-chain-state = { path = "crates/chain-state" }
 reth-chainspec = { path = "crates/chainspec", default-features = false }
 reth-cli = { path = "crates/cli/cli" }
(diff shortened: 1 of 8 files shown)


Cargo.toml
mattsse, line 1: do we need all of those changes here?

I think we can keep this unchanged?

crates/optimism/consensus/src/lib.rs
```diff
@@ -96,12 +94,12 @@ where
...
                 }
                 .into(),
-            ))
+            ));
         }
 
         // Check transaction root
         if let Err(error) = block.ensure_transaction_root_valid() {
-            return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
+            return Err(ConsensusError::BodyTransactionRootDiff(error.into()));
```
mattsse, line 102: let's avoid the additional `;`

crates/optimism/chainspec/src/lib.rs
```diff
@@ -227,6 +228,29 @@ impl OpChainSpec {
     }
 }
 
+/// Extracts the Holocene 1599 parameters from the encoded extra data from the parent header.
+///
+/// Caution: Caller must ensure that holocene is active in the parent header.
+///
+/// See also [Base fee computation](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/holocene/exec-engine.md#base-fee-computation)
+pub fn decode_holocene_base_fee<H>(
```
mattsse, line 236: can we move this to basefee.rs file so that this is a bit more organised?

crates/optimism/evm/src/lib.rs
```diff
@@ -187,7 +186,10 @@ where
             prevrandao: Some(attributes.prev_randao),
             gas_limit: attributes.gas_limit,
             // calculate basefee based on parent block's gas usage
-            basefee: next_block_base_fee(self.chain_spec(), parent, attributes.timestamp)?,
+            basefee: self
+                .chain_spec()
+                .next_block_base_fee(&parent, attributes.timestamp)
+                .unwrap_or_default(),
```
mattsse, line 192: I think this would be a bit problematic because now an error is treated as 0 value, imo we can keep using the regular next_block_base_fee function here

crates/optimism/consensus/src/validation/mod.rs
```diff
@@ -154,62 +152,25 @@ fn compare_receipts_root_and_logs_bloom(
...
 
     Ok(())
 }
 
-/// Extracts the Holocene 1599 parameters from the encoded extra data from the parent header.
-///
-/// Caution: Caller must ensure that holocene is active in the parent header.
-///
-/// See also [Base fee computation](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/holocene/exec-engine.md#base-fee-computation)
-pub fn decode_holocene_base_fee(
```
mattsse, line 174: since we removed this, we can re-export this function so that we dont need to change any other imports

mattsse (changes requested):

PR #17143, opened 2025-06-30
fix(trie): correct ParallelSparseTrie lower subtrie path management

mediocregopher:
Fixes incorrect behavior in both the `remove_leaf` and `reveal_node` methods of ParallelSparseTrie related to how the `path` field of the lower sparse tries is treated.

**Background**

The `path` field on lower sparse tries should always point to the root node of that lower trie, i.e. the node with the shortest path which still has `path.len() >= UPPER_TRIE_MAX_DEPTH`.

**reveal_nodes**

The `reveal_nodes` method will now set the `path` field of lower sparse tries to the path of nodes being revealed if the node is the first node in that lower sparse trie. When the lower sparse trie already exists, but its path is longer than the node being revealed, then the path is updated to the new shorter path.

**remove_leaf**

The `remove_leaf` method's changes are more complex. When removing a leaf which is the last node of its lower subtrie the lower subtrie is now completely deleted. This is done because there is no way to update the `path` to a value which makes sense in this case.

When removing a leaf results in a branch node being collapsed into its parent extension node as an extension, and that branch was previously the root node of a lower subtrie, then the lower subtrie's `path` is updated to be the path of the node the new extension node is pointing to.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -285,6 +285,47 @@ impl ParallelSparseTrie {
         }
     }
 
+    /// Used by `remove_leaf` to ensure that when a node is removed from a lower subtrie that any
+    /// externalities are handled. These can include:
+    /// - Removing the lower subtrie completely, if it is now empty.
+    /// - Updating the `path` field of the lower subtrie to indicate that its root node has changed.
+    ///
+    /// This method assumes that the caller will deal with putting all other nodes in the trie into
+    /// a consistent state after the removal of this one.
+    ///
+    /// ## Panics
+    ///
+    /// - If the removed node was not a leaf or extension.
+    fn remove_node(&mut self, path: &Nibbles) {
+        let subtrie = self.subtrie_for_path(path);
+        let node = subtrie.nodes.remove(path);
+
+        let Some(idx) = SparseSubtrieType::from_path(path).lower_index() else {
+            // When removing a node from the upper trie there's nothing special we need to do to fix
+            // its `path` field; the upper trie's path is always empty.
+            return;
+        };
+
+        match node {
+            Some(SparseNode::Leaf { .. }) => {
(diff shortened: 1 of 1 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -282,6 +285,47 @@ impl ParallelSparseTrie {
...
+    /// - Removing the lower subtrie completely, if it is now empty.
+    /// - Updating the `path` field of the lower subtrie to indicate that its root node has changed.
+    ///
+    /// This method assumes that the caller will deal with putting all other nodes in the trie into
+    /// a consistent state after the removal of this one.
+    ///
+    /// ## Panics
+    ///
+    /// - If the removed node was not a leaf or extension.
+    fn remove_node(&mut self, path: &Nibbles) {
```
shekhirin, line 299: can we make it accept a subtrie as an argument, so that we don't call `subtrie_for_path` twice in cases when we already have the subtrie available? 

https://github.com/paradigmxyz/reth/blob/2819ab2c0e32d546996d2c7534901cfe68ad6c16/crates/trie/sparse-parallel/src/trie.rs#L486 https://github.com/paradigmxyz/reth/blob/2819ab2c0e32d546996d2c7534901cfe68ad6c16/crates/trie/sparse-parallel/src/trie.rs#L569

shekhirin (commented):
one perf nit

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I just gave it a shot and I'm not sure it's workable... the subtries are all mutable references with the same lifetime as `self`, so when you call `fn remove_node(&mut self, ...)` any refs to the subtries are discarded. At least that's how I'm understanding it...

In any case I don't think this is going to be a bottleneck, ultimately it's a constant-time lookup in an array, plus some simple checks.

shekhirin (approved):

PR #17150, opened 2025-06-30
fix: use safe math for withdrawals check

mattsse:
this is user input and could lead to a overflow panic


RomanHodulak (approved):
It's good to use safe addition, but if the overflow actually occurred, if that's realistic, wouldn't it still be a problem?

mattsse:
not realistic and would then fail validation or even earlier, but this should get rid of a potential panic

PR #16886, opened 2025-06-18
fix: move `bytecode_by_hash` from `StateProvider` to a dedicated `BytecodeReader`

0xNarumi:
toward [16876](https://github.com/paradigmxyz/reth/issues/16876)

PR #17156, opened 2025-07-01
fix: move remaining trait methods of `StateProvider` to `AccountReader`

meohyun2:
📄 Summary
	•	As suggested in the review ([link](https://github.com/paradigmxyz/reth/pull/16886#pullrequestreview-2938915991)), this PR moves the relevant trait methods into AccountReader.
	•	Refactored the dependent components accordingly to resolve the resulting dependency issues.

⸻

🔧 Changes
	•	Moved the following trait methods into AccountReader: `storage`, `account_code`, `account_balance`, `account_nonce`
	•	Updated implementations and usages across the codebase.
	•	Resolved dependency cycles and refactored impacted modules.

### DIFF
--- a/crates/storage/storage-api/src/state.rs
+++ b/crates/storage/storage-api/src/state.rs
@@ -29,7 +29,6 @@ pub trait StateReader: Send + Sync {
 pub type StateProviderBox = Box<dyn StateProvider>;
 
 /// An abstraction for a type that provides state data.
-#[auto_impl(&, Arc, Box)]
 pub trait StateProvider:
     BlockHashReader
     + AccountReader
@@ -40,6 +39,19 @@ pub trait StateProvider:
     + HashedPostStateProvider
     + Send
     + Sync
+{
+}
+
+impl<T> StateProvider for T where
+    T: BlockHashReader
+        + AccountReader
+        + BytecodeReader
+        + StateRootProvider
+        + StorageRootProvider
+        + StateProofProvider
+        + HashedPostStateProvider
+        + Send
+        + Sync
 {
 }
 
--- a/crates/storage/storage-api/src/state.rs
+++ b/crates/storage/storage-api/src/state.rs
@@ -3,9 +3,8 @@ use super::{
     StorageRootProvider,
 };
 use alloc::boxed::Box;
-use alloy_consensus::constants::KECCAK_EMPTY;
 use alloy_eips::{BlockId, BlockNumberOrTag};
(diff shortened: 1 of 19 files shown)


crates/storage/storage-api/src/state.rs
```diff
@@ -42,54 +41,6 @@ pub trait StateProvider:
     + Send
     + Sync
 {
-    /// Get storage of given account.
```
mattsse: now that we've removed all functions from this type, we can add a blanket impl for T that satisfies all traits, then we dont need to impl StateProvider manually anymore

mattsse (changes requested):
I think this makes sense, encapsulates the logic really nicely.

I only have one request

crates/storage/storage-api/src/state.rs
meohyun2: Thanks for the great suggestion!

crates/storage/storage-api/src/state.rs
```diff
@@ -30,7 +29,6 @@ pub trait StateReader: Send + Sync {
 pub type StateProviderBox = Box<dyn StateProvider>;
 
 /// An abstraction for a type that provides state data.
-#[auto_impl(&, Arc, Box)]
```
mattsse: hmm, does this mean these are no longer StateProvider?

meohyun2: It was my mistake. After your comment, I looked into whether it’s possible to implement both blanket impl and auto_impl for StateProvider at the same time. As it turns out, it seems that this would only be possible once Specialization (which is not yet stable) is introduced. So, I think we should roll back to the previous commit (6b7c1d5) and exclude the blanket impl for now.

That said, I want to make sure I understood your review comment correctly.

Could you explain in more detail how blanket impl would be applied in this case?

meohyun2: @mattsse  I did some testing, and it seems that when using blanket_impl, the functionality provided by auto_impl is already being handled.

I tested the following code:
```rust
let s: Arc<dyn StateProvider> = Arc::new(state_provider.clone().state_provider);
let s2: Box<dyn StateProvider> = Box::new(state_provider.state_provider);
let s3: &dyn StateProvider = &state_provider.state_provider;
```
The code snippets were declared inside the test code of CachedStateProvider (in `cached_state.rs`, line 683), and each variable was written to verify that the CachedStateProvider trait compiles correctly when wrapped or referenced using Arc, Box, or &. and it compiles without any issues.

Therefore, I think `#[auto_impl(&, Arc, Box)]` can be safely removed.

PR #17166, opened 2025-07-01
chore(trie): factor out SparseTrieState

mediocregopher:
This change removes the need for the SparseTrieState type, and the associated `SparseTrie::AllocatedEmpty` variant, by instead passing around the reset RevealedSparseTrie itself in order to re-use allocations across payloads.

The `SparseTrie::Blind` variant is modified to now optionally carry a `RevealedSparseTrie`, and the SparseTrieTask will take that out of the SparseStateTrie to re-use.

The `RevealedSparseTrie::from_root` method is transformed into `RevealedSparseTrie::with_root`, which performs essentially the same function but allows us to re-use the RevealedSparseTrie while maintaining correct semantics around the empty root node.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -11,7 +11,7 @@ use reth_trie_parallel::root::ParallelStateRootError;
 use reth_trie_sparse::{
     blinded::{BlindedProvider, BlindedProviderFactory},
     errors::{SparseStateTrieResult, SparseTrieErrorKind},
-    SparseStateTrie, SparseTrieState,
+    SparseStateTrie, SparseTrie,
 };
 use std::{
     sync::mpsc,
@@ -66,40 +66,39 @@ where
         }
     }
 
-    /// Creates a new sparse trie, populating the accounts trie with the given cleared
-    /// `SparseTrieState` if it exists.
+    /// Creates a new sparse trie, populating the accounts trie with the given `SparseTrie`, if it
+    /// exists.
     pub(super) fn new_with_stored_trie(
         executor: WorkloadExecutor,
         updates: mpsc::Receiver<SparseTrieUpdate>,
         blinded_provider_factory: BPF,
         trie_metrics: MultiProofTaskMetrics,
-        sparse_trie_state: Option<SparseTrieState>,
+        sparse_trie: Option<SparseTrie>,
     ) -> Self {
-        if let Some(sparse_trie_state) = sparse_trie_state {
+        if let Some(sparse_trie) = sparse_trie {
             Self::with_accounts_trie(
                 executor,
                 updates,
                 blinded_provider_factory,
                 trie_metrics,
-                sparse_trie_state,
+                sparse_trie,
             )
         } else {
             Self::new(executor, updates, blinded_provider_factory, trie_metrics)
         }
     }
 
-    /// Creates a new sparse trie task, using the given cleared `SparseTrieState` for the accounts
+    /// Creates a new sparse trie task, using the given [`SparseTrie::Blind`] for the accounts
     /// trie.
     pub(super) fn with_accounts_trie(
         executor: WorkloadExecutor,
         updates: mpsc::Receiver<SparseTrieUpdate>,
         blinded_provider_factory: BPF,
         metrics: MultiProofTaskMetrics,
-        sparse_trie_state: SparseTrieState,
+        sparse_trie: SparseTrie,
     ) -> Self {
-        let mut trie = SparseStateTrie::new().with_updates(true);
-        trie.populate_from(sparse_trie_state);
-
+        debug_assert!(sparse_trie.is_blind());
+        let trie = SparseStateTrie::new().with_updates(true).with_accounts_trie(sparse_trie);
         Self { executor, updates, metrics, trie, blinded_provider_factory }
     }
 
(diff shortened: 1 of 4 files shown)


crates/trie/sparse/src/trie.rs
```diff
@@ -2110,8 +2058,6 @@ impl SparseTrieUpdates {
     }
 
     /// Clears the updates, but keeps the backing data structures allocated.
-    ///
-    /// Sets `wiped` to `false`.
```
Rjected: this seems to still be true in the new fn?

crates/trie/sparse/src/trie.rs
```diff
@@ -451,59 +416,48 @@ impl RevealedSparseTrie {
...
-            self.updates = Some(SparseTrieUpdates::default());
-        }
+        self.updates = retain_updates.then(|| {
+            self.updates
+                .map(|mut u| {
+                    u.clear();
+                    u
+                })
+                .unwrap_or_default()
+        });
```
Rjected: is this to reuse the updates vec?

crates/trie/sparse/src/state.rs
```diff
@@ -75,21 +75,22 @@ impl SparseStateTrie {
         self
     }
 
+    /// Set the state root to the given `SparseTrie`.
```
Rjected: I wouldn't use the word root here, since it sets the whole trie, maybe accounts trie is more accurate
```suggestion
    /// Set the accounts trie to the given `SparseTrie`.
```

Rjected (commented):
this makes sense, I have some doc comments and a question, it seems like this is slightly easier now that there are no generics in the sparse trie

crates/trie/sparse/src/trie.rs
mediocregopher: Yup, since this gets called as part of `with_root`, and the goal of all this is to reuse buffers, I thought it would be better to not throw away the updates buffer here.

crates/trie/sparse/src/trie.rs
mediocregopher: You're right, I had changed this previously but changed it back, forgot to revert the comment.

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -154,8 +153,8 @@ where
         self.metrics.sparse_trie_final_update_duration_histogram.record(start.elapsed());
         self.metrics.sparse_trie_total_duration_histogram.record(now.elapsed());
 
-        // take the account trie
-        let trie = self.trie.take_cleared_account_trie_state();
+        // take the account trie so that we can re-use its already allocated datastructures.
```
shekhirin:
```suggestion
        // take the account trie so that we can re-use its already allocated data structures.
```

crates/trie/sparse/src/state.rs
```diff
@@ -574,6 +575,7 @@ impl SparseStateTrie {
         Ok(Some(root_node))
     }
 
+    #[cfg(test)]
```
shekhirin: it's a public API that can be used by someone else, so I'd remove this cfg attr

crates/trie/sparse/src/trie.rs
```diff
@@ -451,59 +416,48 @@ impl RevealedSparseTrie {
...
 
     /// Configures the trie to retain information about updates.
     ///
     /// If `retain_updates` is true, the trie will record branch node updates and deletions.
     /// This information can then be used to efficiently update an external database.
     pub fn with_updates(mut self, retain_updates: bool) -> Self {
-        if retain_updates {
-            self.updates = Some(SparseTrieUpdates::default());
-        }
+        self.updates = retain_updates.then(|| {
```
shekhirin: hmm, I think this would change the existing behaviour. If you re-use a sparse trie that had `with_updates(true)`, but now you don't call `with_updates(true)`, it will still collect updates, as the pre-allocated updates vec is `Some`.

crates/trie/sparse/src/trie.rs
```diff
@@ -267,6 +219,19 @@ impl SparseTrie {
         let revealed = self.as_revealed_mut()?;
         Some((revealed.root(), revealed.take_updates()))
     }
+
+    /// Returns a [`SparseTrie::Blind`] based on this one. If this instance was revealed, or was
+    /// itself a `Blind` with a pre-allocated [`RevealedSparseTrie`], this will return
+    /// a `Blind` carrying a reset pre-allocated [`RevealedSparseTrie`].
+    pub fn blinded(self) -> Self {
```
shekhirin: the method name is a bit confusing IMO, as it doesn't communicate that we reset the trie to re-use allocations. Maybe just `clear`?

crates/trie/sparse/src/trie.rs
```diff
@@ -848,11 +805,13 @@ impl RevealedSparseTrie {
     }
 
     /// This clears all data structures in the sparse trie, keeping the backing data structures
-    /// allocated.
+    /// allocated. A [`SparseNode::Empty`] is inserted at the root.
     ///
     /// This is useful for reusing the trie without needing to reallocate memory.
-    pub fn clear(&mut self) {
+    pub fn reset(&mut self) {
```
shekhirin: Think it would be easier if the method name was `clear` similar to how std names such methods, like `HashMap::clear` or `Vec::clear`

crates/trie/sparse/src/trie.rs
mediocregopher: I think you're correct... it wasn't a problem previously because we didn't directly re-use the RevealedSparseTrie, so updates would always start as None. I think the solution will be to have reset/clear always set updates to None for now, and revert this change. If we want to re-use the updates buffer we can figure out something for that later, we aren't doing it up till now anyway.

shekhirin: sgtm

crates/trie/sparse/src/trie.rs
mediocregopher: This change has been reverted after [this discussion](https://github.com/paradigmxyz/reth/pull/17166#discussion_r2179660611)

shekhirin (approved):
LGTM

mattsse (approved):
cool, lgtm!

PR #17167, opened 2025-07-01
feat: include chain-id query param for etherscan v2 API

iTranscend:
resolves https://github.com/paradigmxyz/reth/issues/17142

We'll need to update the default api urls to V2 in the [alloy-chains](https://github.com/alloy-rs/chains/blob/09cafe326bd9b9a43124de6087a93803158088c5/src/named.rs#L1290-L1673) crate. I've opened a [PR](https://github.com/alloy-rs/chains/pull/174) to alloy-chains that does this.

Is this an ideal approach? @mattsse

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7548,6 +7548,7 @@ dependencies = [
  "reth-tracing",
  "ringbuffer",
  "serde",
+ "serde_json",
  "tokio",
 ]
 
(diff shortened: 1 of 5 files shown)


mattsse (changes requested):
this looks great,

I've also released a new patch release of alloy-chains, which we should bump here as well

mattsse (approved):
ty!

PR #17171, opened 2025-07-02
chore: update hashes in parallel

### LINKED ISSUE #17161: Update `ParallelSparseTrie` hashes in parallel
https://github.com/paradigmxyz/reth/blob/4199dd46767af78ac9c4b360c3c218e80d54b2e1/crates/trie/sparse-parallel/src/trie.rs#L680-L687

varun-doshi:
Closes #17161

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -678,13 +678,23 @@ impl ParallelSparseTrie {
         self.prefix_set = unchanged_prefix_set;
 
         // Update subtrie hashes in parallel
-        // TODO: call `update_hashes` on each subtrie in parallel
         let (tx, rx) = mpsc::channel();
-        for ChangedSubtrie { index, mut subtrie, mut prefix_set } in subtries {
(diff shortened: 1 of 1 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -678,13 +678,23 @@ impl ParallelSparseTrie {
...
+                    subtrie.update_hashes(&mut prefix_set);
+                    tx.send((index, subtrie)).unwrap();
+                })
+            })
+            .collect();
+
         drop(tx);
+        for handle in handles {
+            handle.join().unwrap();
+        }
```
mattsse, line 697: I think we need to think about how to spawn these, either rayon or reuse a blocking pool, so this likely requires more consideration

mediocregopher:
FYI, I implemented this at #17173 before I saw this being open, sorry :grimacing: Using rayon there, feature gated behind `std`, so it seems like the better way forward

mediocregopher:
reth team will be implementing this ourselves, thanks giving it a shot though!

PR #17174, opened 2025-07-02
chore: relax rpc middleware generic

### LINKED ISSUE #17076: Add RPC middleware configuration support to node builder
## Description
Currently, while `reth` has excellent RPC middleware support internally (`RpcAddOns::with_rpc_middleware`), there's no clean way for external projects (like L2 implementations or custom nodes) to inject middleware during the node building process. 

Operators need to:
- Monitor slow RPC requests (especially `eth_getLogs` that can crash nodes)
- Log problematic requests with full payload details for debugging
- Implement rate limiting for specific RPC methods
- Add custom metrics and monitoring

For example, we're experiencing crashes from massive `eth_getLogs` requests. We need to log requests taking >200ms with full payload details to identify problematic queries, but can't easily inject this middleware.

## Proposed Solution
A solution I have for now is adding a `with_rpc_middleware()` method to the node builder that allows configuring middleware before RPC server launch.

This way is no changes to core RPC infrastructure, can be flexible support any tower-compatible middleware.

mattsse:
closes #17076

this can be modified via with_add_ons or map_add_ons

fyi @thaodt

### DIFF
--- a/crates/ethereum/node/src/node.rs
+++ b/crates/ethereum/node/src/node.rs
@@ -35,7 +35,7 @@ use reth_node_builder::{
 use reth_provider::{providers::ProviderFactoryBuilder, EthStorage};
 use reth_rpc::{eth::core::EthApiFor, ValidationApi};
 use reth_rpc_api::{eth::FullEthApiServer, servers::BlockSubmissionValidationApiServer};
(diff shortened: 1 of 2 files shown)


fgimenez (approved):
lgtm

PR #17179, opened 2025-07-02
feat: add v5 flashbots relay block validation api for Fusaka

varun-doshi:
Closes #17147

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -450,24 +450,24 @@ reth-ress-protocol = { path = "crates/ress/protocol" }
 reth-ress-provider = { path = "crates/ress/provider" }
 
 # revm
-revm = { version = "27.0.1", default-features = false }
-revm-bytecode = { version = "6.0.0", default-features = false }
-revm-database = { version = "7.0.0", default-features = false }
-revm-state = { version = "7.0.0", default-features = false }
+revm = { version = "27.0.2", default-features = false }
+revm-bytecode = { version = "6.0.1", default-features = false }
+revm-database = { version = "7.0.1", default-features = false }
+revm-state = { version = "7.0.1", default-features = false }
 revm-primitives = { version = "20.0.0", default-features = false }
-revm-interpreter = { version = "23.0.0", default-features = false }
-revm-inspector = { version = "8.0.1", default-features = false }
-revm-context = { version = "8.0.1", default-features = false }
-revm-context-interface = { version = "8.0.0", default-features = false }
-revm-database-interface = { version = "7.0.0", default-features = false }
-op-revm = { version = "8.0.1", default-features = false }
-revm-inspectors = "0.26.0"
+revm-interpreter = { version = "23.0.1", default-features = false }
+revm-inspector = { version = "8.0.2", default-features = false }
+revm-context = { version = "8.0.2", default-features = false }
+revm-context-interface = { version = "8.0.1", default-features = false }
+revm-database-interface = { version = "7.0.1", default-features = false }
+op-revm = { version = "8.0.2", default-features = false }
+revm-inspectors = "0.26.5"
 
 # eth
 alloy-chains = { version = "0.2.0", default-features = false }
 alloy-dyn-abi = "1.2.0"
 alloy-eip2124 = { version = "0.2.0", default-features = false }
-alloy-evm = { version = "0.13", default-features = false }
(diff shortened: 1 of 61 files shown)


varun-doshi:
Missing an impl
Working on it

Cargo.toml
```diff
@@ -489,9 +489,9 @@ alloy-rpc-client = { version = "1.0.16", default-features = false }
 alloy-rpc-types = { version = "1.0.16", features = ["eth"], default-features = false }
 alloy-rpc-types-admin = { version = "1.0.16", default-features = false }
 alloy-rpc-types-anvil = { version = "1.0.16", default-features = false }
-alloy-rpc-types-beacon = { version = "1.0.16", default-features = false }
+alloy-rpc-types-beacon = { version = "1.0.17", default-features = false }
 alloy-rpc-types-debug = { version = "1.0.16", default-features = false }
-alloy-rpc-types-engine = { version = "1.0.16", default-features = false }
+alloy-rpc-types-engine = { version = "1.0.17", default-features = false }
```
mattsse: we always need to increase these in lockstep, so we should bump all alloy deps here from 1.0.16 to 1.0.17

crates/rpc/rpc/src/validation.rs
```diff
@@ -365,6 +366,29 @@ where
 
         Ok(versioned_hashes)
     }
+    /// Validates the given [`BlobsBundleV1`] and returns versioned hashes for blobs.
+    pub fn validate_blobs_bundl_v2(
```
mattsse: smol typo

crates/rpc/rpc/src/validation.rs
```diff
@@ -365,6 +366,29 @@ where
...
+            return Err(ValidationApiError::InvalidBlobsBundle)
+        }
+
+        let versioned_hashes = blobs_bundle
+            .commitments
+            .iter()
+            .map(|c| kzg_to_versioned_hash(c.as_slice()))
+            .collect::<Vec<_>>();
+
+        let sidecar = blobs_bundle.pop_sidecar(blobs_bundle.blobs.len());
```
mattsse: we can now use the new `.try_into_sidecar` function on blobs_bundle here

so that this becomes

`blobs_bundle.try_into_sidecar().map_err(InvalidBlobsBundle)?.validate()`

then we don't really need the function because this is then just a one liner

mattsse (changes requested):
great, smol nits

crates/rpc/rpc/src/validation.rs
```diff
@@ -365,6 +366,29 @@ where
...
+    /// Validates the given [`BlobsBundleV1`] and returns versioned hashes for blobs.
+    pub fn validate_blobs_bundl_v2(
+        &self,
+        mut blobs_bundle: BlobsBundleV2,
+    ) -> Result<Vec<B256>, ValidationApiError> {
+        if blobs_bundle.commitments.len() != blobs_bundle.proofs.len() ||
+            blobs_bundle.commitments.len() != blobs_bundle.blobs.len()
+        {
+            return Err(ValidationApiError::InvalidBlobsBundle)
+        }
```
bharath-123: We need to check `len(commitments) == len(blobs)` and `len(proofs) = len(blobs) * CELLS_PER_EXT_BLOB`

crates/rpc/rpc/src/validation.rs
bharath-123:
```suggestion
    pub fn validate_blobs_bundle_v2(
```

bharath-123:
Also would recommend changing the PR title. `add flashbots relay support for Fusaka` is misleading as it implies you are adding a relay in reth. It should be on the lines of `add v5 flashbots relay block validation api for Fusaka` or something like that.

crates/rpc/rpc/src/validation.rs
varun-doshi: correct me if Im wrong here but signature of validate still looks like this:
```
validate(
        &self,
        blob_versioned_hashes: &[B256],
        proof_settings: &c_kzg::KzgSettings,
    )
```
    
This returns a `Result<(),_>`
    
Can we modify this to return `Result<Vec<B256>,_>` instead?
    
If you think its a good idea, I can raise a PR for that

mattsse (approved):

PR #15347, opened 2025-03-28
feat(cli): add bootnode command

frankudoags:
closes #15324

PR #17180, opened 2025-07-02
Add bootnode cmd to cli runner

ferranbt:
This is a follow up for https://github.com/paradigmxyz/reth/pull/15347 which adds the new Bootnode command to the cli.


mattsse (approved):
this does make sense,

however the p2p command struct is now a bit of mess, which could be problematic since it now has a bunch of other args

but imo we can clean this up separately

PR #17181, opened 2025-07-02
perf(trie): sparse trie trait

### LINKED ISSUE #17160: Trait for sparse trie that will be used in `SparseTrie` enum
### Describe the feature

-

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/17160

### DIFF
--- a/crates/trie/sparse/src/traits.rs
+++ b/crates/trie/sparse/src/traits.rs
@@ -1,21 +1,56 @@
 //! Traits for sparse trie implementations.
 
+use core::fmt::{Debug, Display};
+
 use alloc::vec::Vec;
 use alloy_primitives::B256;
 use reth_execution_errors::SparseTrieResult;
 use reth_trie_common::{Nibbles, TrieNode};
 
-use crate::{
-    blinded::BlindedProvider,
-    SparseTrieUpdates, TrieMasks,
-};
+use crate::{blinded::BlindedProvider, SparseTrieUpdates, TrieMasks};
 
 /// Trait defining common operations for revealed sparse trie implementations.
 ///
 /// This trait abstracts over different sparse trie implementations (serial vs parallel)
 /// while providing a unified interface for the core trie operations needed by the
-/// SparseTrie enum.
-pub trait SparseTrieInterface: Default + Clone + PartialEq + Eq + core::fmt::Debug {
+/// [`crate::SparseTrie`] enum.
+pub trait SparseTrieInterface: Default + Clone + Debug + Display {
+    /// Creates a new revealed sparse trie from the given root node.
+    ///
+    /// This function initializes the internal structures and then reveals the root.
+    /// It is a convenient method to create a trie when you already have the root node available.
+    ///
+    /// # Returns
+    ///
+    /// Self if successful, or an error if revealing fails.
+    fn from_root(root: TrieNode, masks: TrieMasks, retain_updates: bool) -> SparseTrieResult<Self>;
+
+    /// Configures the trie to retain information about updates.
+    ///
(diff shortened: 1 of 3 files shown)


crates/trie/sparse/src/traits.rs
```diff
@@ -0,0 +1,196 @@
...
+use reth_trie_common::{Nibbles, TrieNode};
+
+use crate::{blinded::BlindedProvider, SparseTrieUpdates, TrieMasks};
+
+/// Trait defining common operations for revealed sparse trie implementations.
+///
+/// This trait abstracts over different sparse trie implementations (serial vs parallel)
+/// while providing a unified interface for the core trie operations needed by the
+/// [`crate::SparseTrie`] enum.
+pub trait SparseTrieInterface: Default + Clone + Debug {
```
mediocregopher: Is the trie ever actually cloned? That seems like a mistake if it does happen

crates/trie/sparse/src/traits.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// # Arguments
+    ///
+    /// * `root` - The root node to reveal
+    /// * `masks` - Trie masks for root branch node
+    /// * `retain_updates` - Whether to track updates
+    ///
+    /// # Returns
+    ///
+    /// Self if successful, or an error if revealing fails.
+    fn with_root(
```
mediocregopher, line 53: Comment should indicate that this will panic if called on a trie which is not new/cleared

crates/trie/sparse/src/traits.rs
```diff
@@ -0,0 +1,196 @@
...
+    ///
+    /// Self for method chaining.
+    fn with_updates(self, retain_updates: bool) -> Self;
+
+    /// Reserves capacity for additional trie nodes.
+    ///
+    /// # Arguments
+    ///
+    /// * `additional` - The number of additional trie nodes to reserve capacity for.
+    fn reserve_nodes(&mut self, additional: usize);
```
mediocregopher, line 80: I wasn't sure how this would end up getting implemented for ParallelSparseTrie, would we just make it a no-op?

mediocregopher (commented):
1 comment and 2 questions

crates/trie/sparse/src/traits.rs
shekhirin: yeah, thinking of just a noop https://github.com/paradigmxyz/reth/commit/64d76c61d9c6e2e1f504cdf653abde0e228ea69c

crates/trie/sparse/src/traits.rs
shekhirin: yeah, think it's not necessary. If we need it, we can add it later

mediocregopher (approved):

mediocregopher (approved):

crates/trie/sparse/src/traits.rs
```diff
@@ -0,0 +1,307 @@
...
+use reth_trie_common::{Nibbles, TrieNode};
+
+use crate::blinded::BlindedProvider;
+
+/// Trait defining common operations for revealed sparse trie implementations.
+///
+/// This trait abstracts over different sparse trie implementations (serial vs parallel)
+/// while providing a unified interface for the core trie operations needed by the
+/// [`crate::SparseTrie`] enum.
+pub trait SparseTrieInterface: Default + Debug {
```
mattsse, line 21: can we find something other than Interface?

mattsse (approved):
lgtm, but maybe we can come up with something better than Interface, although not a blocker here

PR #17183, opened 2025-07-02
feat(txpool): add minimal priority fee configuration for transaction pool

0x00101010:
# Add minimum priority fee parameter for transaction pool

  ## Summary

  This PR adds a `--txpool.minimum-priority-fee` CLI parameter that enforces a minimum priority
  fee for transactions entering the pool. Node operators can use this to reject transactions with
   priority fees below a specified threshold, providing better control over transaction pool
  quality and preventing spam.

  ## Changes

  ### CLI Parameter
  - Added `--txpool.minimum-priority-fee` parameter (optional, no default)
  - When not specified, no minimum priority fee validation is performed
  - Integrated with existing transaction pool configuration system

  ### Transaction Pool Validation
  - Enhanced `EthTransactionValidator` to validate priority fees in `validate_one_no_state()`
  - Validation applies to External and Private transactions
  - Local transactions are exempt from minimum priority fee validation
  - Only applies to dynamic fee transactions (EIP-1559) that have `max_priority_fee_per_gas`

  ### Error Handling
  - Added `PriorityFeeBelowMinimum` error variant to `InvalidPoolTransactionError`
  - Proper RPC error mapping with descriptive error messages including the minimum required fee
  - Error message: "transaction priority fee below minimum required priority fee
  {minimum_priority_fee}"

  ### Architecture Integration
  - Updated both Ethereum and Optimism node builders to pass the configuration
  - Updated `with_minimum_priority_fee()` method to `EthTransactionValidatorBuilder`
  - Configuration flows from CLI args → pool config → validator builder → validator

  ## Compatibility

  This feature is compatible with both Ethereum and Optimism nodes due to shared transaction pool
   infrastructure. The validation occurs in the `EthTransactionValidator` which is used by both
  node types.

  Affected transaction sources:
  - RPC-submitted transactions (`eth_sendRawTransaction`, `eth_sendTransaction`)
  - P2P propagated transactions
  - Private mempool transactions

  Local transactions are exempt from this validation to maintain local transaction privileges.

  ## Breaking Changes

  None. The feature is disabled by default (parameter not specified) to maintain backward
  compatibility.

  ## Usage

  ```bash
  # Set minimum priority fee to 1 gwei
  reth node --txpool.minimum-priority-fee 1000000000

  # Set minimum priority fee to 10 gwei
  reth node --txpool.minimum-priority-fee 10000000000

  # No minimum priority fee (default behavior)
  reth node

  Implementation Details

  The validation is implemented in EthTransactionValidator::validate_one_no_state() and only
  applies to dynamic fee transactions (EIP-1559). The check occurs after basic transaction
  validation but before state-dependent validation.

  The feature uses the existing transaction origin system to exempt local transactions while
  applying validation to external and private transactions. This maintains the existing local
  transaction privileges while providing spam protection for network-sourced transactions.


crates/transaction-pool/src/error.rs
```diff
@@ -381,6 +389,7 @@ impl InvalidPoolTransactionError {
                 Eip7702PoolTransactionError::InflightTxLimitReached => false,
                 Eip7702PoolTransactionError::AuthorityReserved => false,
             },
+            Self::PriorityFeeBelowMinimum { .. } => false,
```
mattsse, line 390: yep, this def needs to be false

crates/node/core/src/args/txpool.rs
```diff
@@ -65,6 +66,11 @@ pub struct TxPoolArgs {
     #[arg(long = "txpool.minimal-protocol-fee", default_value_t = MIN_PROTOCOL_BASE_FEE)]
     pub minimal_protocol_basefee: u64,
 
+    /// Minimum priority fee required for transaction acceptance into the pool.
+    /// Transactions with priority fee below this value will be rejected. Set to 0 to disable.
+    #[arg(long = "txpool.minimal-priority-fee", default_value_t = DEFAULT_MINIMAL_PRIORITY_FEE)]
+    pub minimal_priority_fee: u64,
```
mattsse: this was indeed missing from the cli args

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1678,6 +1695,14 @@ impl<T: PoolTransaction> AllTransactions<T> {
             })
         }
 
+        if transaction.priority_fee_or_price() < self.minimal_priority_fee as u128 {
+            return Err(InsertErr::PriorityFeeBelowMinimum {
+                priority_fee: transaction.priority_fee_or_price(),
+                minimal_priority_fee: self.minimal_priority_fee as u128,
+                transaction: Arc::new(transaction),
+            })
```
mattsse: we currently enforce this check here during the validation part:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/transaction-pool/src/validate/eth.rs#L439-L449

which seems more appropriate

although this only applies this check to dynamic fee txs because for legacy txs this setting would be dynamic

mattsse (changes requested):
this setting was indeed missing from the CLI args

but we have it as part of the validator, so I suggest that we instead keep the validation in the validator and configure it properly via:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/transaction-pool/src/validate/eth.rs#L930-L930

here:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/ethereum/node/src/node.rs#L425-L425

and here:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/optimism/node/src/node.rs#L806-L806

I'm open to changing how this check is enforced with non 1559 txs

crates/node/core/src/args/txpool.rs
mattsse: we can keep make this an option here then

0x00101010: Done

crates/transaction-pool/src/pool/txpool.rs
0x00101010: Great, updated code to reflect that. Also don't think we need to handle legacy txs as they don't have a priority fee concept.

mattsse (approved):

PR #17184, opened 2025-07-02
feat(p2p): separate args for (header, body)

### LINKED ISSUE #17182: Cleanup p2p subcommand
### Describe the feature

these additional args are currently on the p2p subcommand level

https://github.com/paradigmxyz/reth/blob/b286a61db8a46e51649be721d4db7d3f3b3fbddb/crates/cli/commands/src/p2p/mod.rs#L23-L53

but only need for 

https://github.com/paradigmxyz/reth/blob/b286a61db8a46e51649be721d4db7d3f3b3fbddb/crates/cli/commands/src/p2p/mod.rs#L62-L73

so we should group these differently instead

# TODO
* introduce Args container for download subcommands and use for those (header,body) execlusively,
* then move all of the setup code in there

https://github.com/paradigmxyz/reth/blob/b286a61db8a46e51649be721d4db7d3f3b3fbddb/crates/cli/commands/src/p2p/mod.rs#L81-L120

ref #17180

### Additional context

_No response_

Dyslex7c:
Closes #17182


crates/cli/commands/src/p2p/mod.rs
```diff
@@ -99,35 +116,54 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
             .manager()
             .await?;
         let network = net.handle().clone();
+        let handle = Arc::new(network);
         tokio::task::spawn(net);
 
-        let fetch_client = network.fetch_client().await?;
-        let retries = self.retries.max(1);
-        let backoff = ConstantBuilder::default().with_max_times(retries);
+        let fetch_client = handle.fetch_client().await?;
```
mattsse: this we can leave to the caller even because we return the handle

crates/cli/commands/src/p2p/mod.rs
```diff
@@ -99,35 +116,54 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
             .boot_nodes(boot_nodes.clone())
             .apply(|builder| {
                 self.network.discovery.apply_to_builder(builder, rlpx_socket, boot_nodes)
             })
-            .build_with_noop_provider(self.chain)
+            .build_with_noop_provider(chain)
             .manager()
             .await?;
         let network = net.handle().clone();
+        let handle = Arc::new(network);
```
mattsse: this we dont need to arc because the handle is clone

crates/cli/commands/src/p2p/mod.rs
```diff
@@ -75,11 +82,21 @@ pub enum Subcommands {
     Rlpx(rlpx::Command),
 }
 
-impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>> Command<C> {
-    /// Execute `p2p` command
-    pub async fn execute<N: CliNodeTypes<ChainSpec = C::ChainSpec>>(self) -> eyre::Result<()> {
-        let data_dir = self.datadir.clone().resolve_datadir(self.chain.chain());
-        let config_path = self.config.clone().unwrap_or_else(|| data_dir.config());
+impl DownloadArgs {
```
mattsse: this is a bit out of order, we should define the DownloadArgs after the Commands impl and then move this block there so that we have

struct Command

impl Command

struct DownloadArgs

impl DownloadArgs

crates/cli/commands/src/p2p/mod.rs
```diff
@@ -75,11 +82,21 @@ pub enum Subcommands {
...
+    pub async fn setup<C, N>(
+        &self,
+        chain: Arc<C>,
+        config: Option<PathBuf>,
```
mattsse: I think we can also move these two fields into the download args?

mattsse (changes requested):
cool, good start

Dyslex7c:
@mattsse I've added all the changes you mentioned in the latest commit, please do check

mattsse (approved):
very nice

PR #17185, opened 2025-07-02
replace reference monitoring.mdx

strmfos:
Updated the [metrics] link in `monitoring.mdx` to point to a new documentation location:
- Removed: `https://github.com/paradigmxyz/reth/blob/main/docs/design/metrics#current-metrics`
- Added: `https://github.com/paradigmxyz/reth/blob/main/docs/design/metrics.md`


mattsse (commented):
we replaced this with a link to the docs instead

PR #17205, opened 2025-07-03
chore(trie): make SparseStateTrie generic with respect to trie implementation

### LINKED ISSUE #17088: Make SparseStateTrie type generic with respect to trie implementation

mediocregopher:
The SparseStateTrie now allows for different implementations of the sparse trie to be used, with the ability to have account trie use a different implementation than storage tries as well.

Closes #17088

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10611,6 +10611,7 @@ dependencies = [
  "proptest-arbitrary-interop",
  "rand 0.8.5",
  "rand 0.9.1",
+ "rayon",
  "reth-execution-errors",
  "reth-primitives-traits",
  "reth-trie",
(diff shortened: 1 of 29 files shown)


Rjected (approved):
lgtm, nice

shekhirin (approved):

mattsse (approved):

PR #17213, opened 2025-07-03
fix: Returns Arc<RecoveredBlock> in BlockAndReceiptsResult

### LINKED ISSUE #17210: Returns Arc<RecoveredBlock> in BlockAndReceiptsResult
### Describe the feature

this is currently not super efficient

https://github.com/paradigmxyz/reth/blob/037be8d7ac27d53ed160489366593a87bfe3bd42/crates/rpc/rpc-eth-api/src/helpers/block.rs#L23-L26

because most of the time

we end up cloning a bunch of things:

https://github.com/paradigmxyz/reth/blob/037be8d7ac27d53ed160489366593a87bfe3bd42/crates/rpc/rpc-eth-api/src/helpers/block.rs#L142-L150

this is only a SealedBlock because of:

https://github.com/paradigmxyz/reth/blob/037be8d7ac27d53ed160489366593a87bfe3bd42/crates/rpc/rpc-eth-api/src/helpers/block.rs#L130-L130

but we already have:

https://github.com/paradigmxyz/reth/blob/037be8d7ac27d53ed160489366593a87bfe3bd42/crates/storage/storage-api/src/block.rs#L85-L85

so we should change:

https://github.com/paradigmxyz/reth/blob/037be8d7ac27d53ed160489366593a87bfe3bd42/crates/storage/storage-api/src/block.rs#L89-L91

to just return recoveredblock instead.

## TODO
* transition these functions and types to operate on recoveredBlock
* starting from the type alias, then functions and then callsite

### Additional context

_No response_

frankudoags:
closes https://github.com/paradigmxyz/reth/issues/17210

### DIFF
--- a/crates/alloy-provider/src/lib.rs
+++ b/crates/alloy-provider/src/lib.rs
@@ -34,9 +34,7 @@ use reth_db_api::{
 };
 use reth_errors::{ProviderError, ProviderResult};
 use reth_node_types::{BlockTy, HeaderTy, NodeTypes, PrimitivesTy, ReceiptTy, TxTy};
-use reth_primitives::{
-    Account, Bytecode, RecoveredBlock, SealedBlock, SealedHeader, TransactionMeta,
-};
+use reth_primitives::{Account, Bytecode, RecoveredBlock, SealedHeader, TransactionMeta};
 use reth_provider::{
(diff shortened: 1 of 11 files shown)


frankudoags:
wip

mattsse (approved):
great!

PR #17221, opened 2025-07-04
chore: use encoded txns while conversions

### LINKED ISSUE #17186: reuse encoded txs when doing the payload to RecoveredBlock conversion
### Describe the feature

Blocked by https://github.com/paradigmxyz/reth/issues/17413

currently we do:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/ethereum/node/src/engine.rs#L46-L51

after doing:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/ethereum/payload/src/validator.rs#L81-L81

however with 

https://github.com/alloy-rs/alloy/blob/c808022d571aaffb97a3429a0fdba6fc06e8f7aa/crates/rpc-types-engine/src/payload.rs#L1180-L1193

we could reuse the input bytes directly for recovery and avoiding another rlp(tx) operation when doing recovery.

e.g. we decode txs from the bytes and also use the bytes to recover the sender and store in a vec,

then we can make use of:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/primitives-traits/src/block/recovered.rs#L95-L99

directly.

to do this we need to change the return value of

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/ethereum/payload/src/validator.rs#L72-L75

to what we have here:

https://github.com/paradigmxyz/reth/blob/60940dd243918a1ba55f5cce5046a443bc75611d/crates/ethereum/node/src/engine.rs#L46-L49

### Additional context

_No response_

dawksh:
`try_into_block_with_sidecar`: does a direct RLP decode of transaction bytes into transaction struct
`try_recover_sealed_unchecked`: uses already decoded transaction

Resolves #17186

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12220,9 +12220,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.46.0"
+version = "1.46.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1140bb80481756a8cbe10541f37433b459c5aa1e727b4c020fbfebdc25bf3ec4"
+checksum = "0cc3a2344dafbe23a245241fe8b09735b521110d30fcefbbd5feb1797ca35d17"
 dependencies = [
  "backtrace",
  "bytes",


crates/ethereum/payload/src/validator.rs
```diff
@@ -72,14 +72,17 @@ impl<ChainSpec: EthereumHardforks> EthereumExecutionPayloadValidator<ChainSpec>
...
+    ) -> Result<RecoveredBlock<Block<T>>, PayloadError> {
         let ExecutionData { payload, sidecar } = payload;
 
         let expected_hash = payload.block_hash();
 
         // First parse the block
         let sealed_block = payload.try_into_block_with_sidecar(&sidecar)?.seal_slow();
 
+        // Get Senders
+        let senders = sealed_block.senders().unwrap();
```
mattsse: this is effectively the same thing as before

mattsse (changes requested):
this doesn't address the intended change yet,

we need to actually reuse the input bytes

crates/ethereum/payload/src/validator.rs
dawksh: apologies, pushed the actual fix, my first reth contribution, was a little confused

dawksh: have posted my reasoning in the PR description, hoping this is correct

dawksh: figuring this out, wip

dawksh: please take a look now, explained my reasoning in PR Description, my first time working with reth so got a bit confused

PR #17223, opened 2025-07-04
fix(trie): track branch node updates only in ParallelSparseTrie, not subtries

### LINKED ISSUE #17172: Don't lost branch updates on `remove_leaf` calls
### Describe the feature

It's possible that during a `remove_leaf` call an entire lower subtrie will end up being deleted. If this subtrie had previously collected any branch node updates these would then be lost.

Fixing this is likely blocked on #16956, which also uses the updates set

### Additional context

* Create a ParallelSparseTrie with roughly this structure:
```
0x: Branch
0x1: Leaf
0x2: Extension
0x234: Branch
0x2345: Leaf
0x2346: Leaf
```

Deleting `0x2345` will result in the branch at `0x234` being temporarily replaced with an extension, and the lower subtrie for `0x23` will collect a corresponding branch node update in its `updates` set. Then the leaf at `0x2346` gets collapsed into the temporary extension, which gets collapsed into the extension at `0x2`. The final structure is:

```
0x: Branch
0x1: Leaf
0x2: Leaf (key: 346)
```

The `0x23` lower subtrie will have been deleted completely, losing the branch node updates.

mediocregopher:
This commit contains 3 bug fixes related to branch node update retention.

**Fix 1**

The ParallelSparseTrie previously kept the `Option<SparseTrieUpdates>` field, used to both enable and collect branch node updates, in multiple places:

* 1 on the ParallelSparseTrie
* 1 on the upper SparseSubtrie
* 1 on each of the 256 lower SparseSubtries.

This caused issues in `remove_leaf`, which needs to fully remove a lower subtrie if it no longer has any nodes, and would therefore lose any accumulated branch node updates.

The fix is to only track branch node updates on the ParallelSparseTrie. This introduces a new wrinkle, which is that `update_hashes` runs in parallel on each lower subtrie and needs to make changes to the global update set. To get around this we track these changes using a `Vec<SparseTrieUpdatesAction>` on each lower subtrie, and then apply those actions in `update_subtrie_hashes`.

**Fix 2**

`with_updates` wasn't properly propagating the setting to the lower subtries. Existing lower subtries wouldn't receive the setting, and any created afterwards wouldn't either. This has been fixed, in that the lower subtries don't track the setting at all anymore, and have it passed to their methods when needed.

**Fix 3**

Fixing the above revealed that `update_leaf` had a bug which went unnoticed before because update retention wasn't actually active. When turning an extension into a branch, it's possible `update_leaf` would need to reveal the previous child of the extension. This was failing in cases where the extension was on the upper trie and the child on a lower one. This has been fixed and a test case added.

Fixes #17172

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -76,14 +76,7 @@ impl SparseTrieInterface for ParallelSparseTrie {
     }
 
     fn with_updates(mut self, retain_updates: bool) -> Self {
-        self.updates = Some(Default::default());
-
-        for subtrie in core::iter::once(&mut self.upper_subtrie)
-            .chain(self.lower_subtries.iter_mut().flatten())
-        {
-            subtrie.set_updates(retain_updates);
-        }
-
+        self.updates = retain_updates.then(|| Default::default());
         self
     }
 
@@ -150,6 +143,8 @@ impl SparseTrieInterface for ParallelSparseTrie {
             return Ok(())
         }
 
+        let retain_updates = self.updates_enabled();
+
         // Start at the root, traversing until we find either the node to update or a subtrie to
         // update.
         //
@@ -170,7 +165,7 @@ impl SparseTrieInterface for ParallelSparseTrie {
         {
             // Traverse the next node, keeping track of any changed nodes and the next step in the
             // trie
-            match self.upper_subtrie.update_next_node(current, &full_path)? {
+            match self.upper_subtrie.update_next_node(current, &full_path, retain_updates)? {
                 LeafUpdateStep::Continue { next_node } => {
                     next = Some(next_node);
                 }
@@ -267,7 +262,7 @@ impl SparseTrieInterface for ParallelSparseTrie {
 
             // If we didn't update the target leaf, we need to call update_leaf on the subtrie
             // to ensure that the leaf is updated correctly.
-            subtrie.update_leaf(full_path, value, provider)?;
+            subtrie.update_leaf(full_path, value, provider, retain_updates)?;
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -514,8 +551,10 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
         drop(tx);
 
-        // Return updated subtries back to the trie
-        for (index, subtrie) in rx {
+        // Return updated subtries back to the trie after executing any actions required on the
+        // top-level `SparseTrieUpdates`.
+        for (index, mut subtrie) in rx {
+            Self::apply_subtrie_update_actions(self.updates.as_mut(), &mut subtrie);
             self.lower_subtries[index] = Some(subtrie);
         }
```
mattsse: the order in which we process them doesn't matter right?

mediocregopher: The order with respect to lower subtrie doesn't matter, because each lower subtrie is only going to produce updates for its own branch nodes (ie those branch nodes with the subtrie's two-nibble prefix)

shekhirin (approved):
LGTM

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1403,17 +1459,24 @@ impl SparseSubtrie {
...
     /// # Panics
     ///
     /// If the node at the root path does not exist.
-    #[instrument(level = "trace", target = "engine::tree", skip_all, fields(root = ?self.path), ret)]
-    pub fn update_hashes(&mut self, prefix_set: &mut PrefixSet) -> RlpNode {
+    #[instrument(level = "trace", target = "trie::parallel_sparse", skip_all, fields(root = ?self.path), ret)]
+    fn update_hashes(
+        &mut self,
+        prefix_set: &mut PrefixSet,
+        update_actions: &mut Option<Vec<SparseTrieUpdatesAction>>,
```
Rjected, line 1478: is the idea here that we will be allocation a few of these action lists and reusing then?

Rjected (approved):
lgtm, makes sense, I do have one question about allocation

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I wanted to leave open room to do that, yeah, also it just looked weird to me to be accepting prefix set as a `&mut` but then return the Vec. At this point it doesn't matter much either way though.

mediocregopher: Oh! and it also moves the logic for deciding if update retention is enabled outside of `update_hashes`, because SparseSubtrie doesn't have a way of knowing that anymore

PR #16889, opened 2025-06-18
feat: load KZG settings on EthTransactionValidator startup

iTranscend:
resolves https://github.com/paradigmxyz/reth/issues/16861

PR #17224, opened 2025-07-04
chore: load kzg settings in background

mattsse:
followup to https://github.com/paradigmxyz/reth/pull/16889

this turned out to be quite expensive so we should load this in the background


shekhirin (approved):

klkvr (approved):
lgtm, triggering hive on top of this branch

PR #17225, opened 2025-07-04
chore: make receipt root mismatch log more useful

klkvr:
this currently logs just a debug impl of a vec which is pretty much unusable if you want to programatically compare it with actual receipts

this PR changes this to vec of rlp encodings which can then be directly compared against valid rlp or decoded into e.g alloy type


mattsse (approved):

mattsse:
makes sense

PR #17234, opened 2025-07-04
chore(doc): update exclude list for doc/cli

fintechup:
This updates the `exclude` list in `Cargo.toml` to include proper path for `doc/cli`. Also removes old paths.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,7 +169,7 @@ members = [
     "crates/tracing-otlp",
 ]
 default-members = ["bin/reth"]
-exclude = ["doc/cli"]
+exclude = ["docs/cli"]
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -169,7 +169,7 @@ members = [
     "crates/tracing-otlp",
 ]
 default-members = ["bin/reth"]
-exclude = ["book/sources", "book/cli"]
+exclude = ["doc/cli"]
```
mattsse:
```suggestion
exclude = ["docs/cli"]
```
should be docs, right?

mattsse (approved):

Cargo.toml
marianor7: yes, my b

PR #17235, opened 2025-07-04
fix: logical error in pruning test for storage_history PruneMode::Full

Galoretka:
Corrected a logical error in the pruning test: when checking for PruneMode::Full on storage_history, the test now correctly asserts that storage_indexing_stage.execute returns an error, instead of incorrectly calling acc_indexing_stage.execute. This ensures the test accurately verifies unsupported full pruning for storage history. No other logic or functionality was changed.


Galoretka:
Also corrected a typo

mattsse (approved):
yeah this appears to actually test the wrong thing

PR #17237, opened 2025-07-04
feat: make `Receipt` generic over `TxType`

### LINKED ISSUE #17081: Make receipt generic over TxType
### Describe the feature

we can make this type reusable if we make the txtype generic and keep eth txtype as the default

https://github.com/paradigmxyz/reth/blob/471f6a375eefc116480bf0ec828ff1ac3286c940/crates/ethereum/primitives/src/receipt.rs#L25-L27

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17081.

### DIFF
--- a/crates/ethereum/primitives/src/receipt.rs
+++ b/crates/ethereum/primitives/src/receipt.rs
@@ -7,7 +7,7 @@ use alloy_eips::{
     eip2718::{Eip2718Result, Encodable2718, IsTyped2718},
     Decodable2718,
 };
-use alloy_primitives::{Bloom, Log, B256};
+use alloy_primitives::{bytes, Bloom, Log, B256};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use reth_primitives_traits::{proofs::ordered_trie_root_with_encoder, InMemorySize};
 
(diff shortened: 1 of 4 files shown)


crates/ethereum/primitives/src/receipt.rs
```diff
@@ -232,13 +265,13 @@ impl Encodable for Receipt {
...
 
-impl Decodable for Receipt {
+impl Decodable for Receipt<TxType> {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         Ok(Self::network_decode(buf)?)
     }
 }
 
-impl TxReceipt for Receipt {
+impl TxReceipt for Receipt<TxType> {
```
rezzmah: Consider making these implementations generic. Otherwise it's not really possible to implement for different `TxTypes` due to error below. Saves an unnecessary wrap

```rs
error[E0117]: only traits defined in the current crate can be implemented for types defined outside of the crate
  --> src/node/evm/receipt.rs:37:1
   |
37 | impl TxReceipt for Receipt<TxTypeCustom> {
   | ^^^^^^^^^^^^^^^^^^^---------------------
   |                    |
   |                    `reth_ethereum_primitives::Receipt` is not defined in the current crate
```

mattsse (approved):
cool,

double checked the expanded code which is equivalent, because we dont expect any changes to the format, we can simply use the expanded code here

PR #17238, opened 2025-07-04
transaction.rs: enhance error handling in RPC transaction helpers

crazykissshout:
As the title described


mattsse (commented):
these are all errors that we dont expect or are too verbose

PR #17239, opened 2025-07-04
refactor(txpool): Remove txhash from PoolUpdate

frankudoags:
- introduce `prune_transaction_by_id` and  `remove_transaction_by_id` functions that operate on `tx_id` but equivalent to `prune_transaction_by_hash` and remove_transaction_by_hash`

- remove txhash field from poolupdate to keep mem print low

closes #17226

PR #17242, opened 2025-07-05
chore: update size metrics once

mattsse:
after https://github.com/paradigmxyz/reth/pull/17239

we have a dedicated fn for removing txs based on updates.

we can update the metrics once after we've processed all updates


Rjected (approved):
makes sense, remove / prune are only used here

PR #17245, opened 2025-07-05
fix: call state_root() method instead of field in state root mismatch error

Galoretka:
Corrects a typo in the error message for state root mismatch by calling the state_root() method instead of referencing it as a field. This ensures the actual state root value is displayed in the error output, improving debugging accuracy.


mattsse (commented):
using the field is fine here

Issue #17197, opened 2025-07-03
Add E2E Test Infrastructure for RPC Methods


fgimenez:
### Describe the feature

Extend the E2E testing framework for RPC methods to cover existing functionality and catch regressions in CI. This includes building reusable test actions and scenarios that can validate the correctness of RPC endpoints under various conditions.

### Additional context

# TODO
Add reusable actions and test scenarios for RPC methods in `crates/e2e-test-utils/sc/testsuite`, starting with eth_getLogs as the first implementation.

PR #17253, opened 2025-07-07
refactor: extract import functionality to separate module

fgimenez:
towards #17197

extracts the import functionality into a separate module without clap dependencies, making it easily reusable.


fgimenez:
successful hive execution here https://github.com/paradigmxyz/reth/actions/runs/16111948843 this uses `reth import` for the `rpc-compat` suite

mattsse (approved):
makes sense

PR #17256, opened 2025-07-07
fix(trie): `ParallelSparseTrie::default` should have an empty root node

shekhirin:
This needs to match the `RevealedSparseTrie::default` implementation https://github.com/paradigmxyz/reth/blob/44b361a4e290267083d85ab00e9334fc9a44379f/crates/trie/sparse/src/trie.rs#L365-L377


mediocregopher (approved):
LGTM :+1:

PR #17259, opened 2025-07-07
fix(trie): correctly handle path field on cleared ParallelSparseTrie lower subtries

mediocregopher:
**Background**

The lower subtries of ParallelSparseTries have a `path` field which indicates the "root node" of the lower subtrie, that is the node with the shortest path held by the lower sparse trie. This field is used in a few methods to aid in walking a lower subtrie without first having to walk the upper.

The `path` was not getting properly reset on `clear()`, resulting in state leaking between re-use of the ParallelSparseTrie.

**Fix**

For maximum correctness a new enum is introduced, LowerSparseSubtrie, which is used to track whether or not a lower subtrie is revealed, or just blinded while re-using a previously allocated `SparseSubtrie`. We never reveal a LowerSparseSubtrie without also revealing its root path at the same time.

Another small fix here is to insert an empty node into the root path of the _upper_ subtrie on `clear()`. This makes the ParallelSparseTrie semantically consistent with its serial implementation, though should have very little effect otherwise.


crates/trie/sparse-parallel/src/lower.rs
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+        };
+    }
+
+    /// Clears the subtrie and transitions it to the blinded state, preserving a cleared
+    /// [`SparseSubtrie`] if possible.
+    pub(crate) fn clear(&mut self) {
+        *self = match core::mem::take(self) {
+            Self::Blind(allocated) => Self::Blind(allocated),
```
shekhirin: let's add a debug assertion here that checks `allocated.is_empty()`

crates/trie/sparse-parallel/src/lower.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub(crate) fn reveal(&mut self, path: &Nibbles) {
+        match self {
+            Self::Blind(None) => {
+                *self = Self::Revealed(Box::new(SparseSubtrie::new(*path)));
+            }
+            entry @ Self::Blind(Some(_)) => {
+                let Self::Blind(Some(mut subtrie)) = core::mem::take(entry) else { unreachable!() };
+                subtrie.path = *path;
+                *entry = Self::Revealed(subtrie);
+            }
```
shekhirin: I think we can just collapse these two arms into one to avoid doing the `unreachable!()`

```suggestion
            Self::Blind(entry) => {
                if let Some(mut subtrie) = entry.take() {
                    subtrie.path = *path;
                    *self = Self::Revealed(subtrie);
                } else {
                    *self = Self::Revealed(Box::new(SparseSubtrie::new(*path)));
                }
            }
```

crates/trie/sparse-parallel/src/lower.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    /// Takes ownership of the underlying [`SparseSubtrie`] if revealed and the predicate returns
+    /// true.
+    ///
+    /// If the subtrie is revealed, and the predicate function returns `true` when called with it,
+    /// then this method will take ownership of the subtrie and transition this `LowerSparseSubtrie`
+    /// to the blinded state. Otherwise, returns `None`.
+    pub(crate) fn take_if<P>(&mut self, predicate: P) -> Option<Box<SparseSubtrie>>
```
shekhirin: I don't see it being used, do we need it?

shekhirin (commented):
the approach makes sense to me

crates/trie/sparse-parallel/src/lower.rs
mediocregopher: It's used in `take_changed_lower_subtries`: https://github.com/paradigmxyz/reth/blob/242c64d22e82272e494f829b3deecdd73d422afa/crates/trie/sparse-parallel/src/trie.rs#L974-L978

Definitely open to better ideas of how to handle this case though, I wasn't super happy with this

crates/trie/sparse-parallel/src/lower.rs
mediocregopher: Nice, done!

shekhirin (approved):

crates/trie/sparse-parallel/src/lower.rs
```diff
@@ -0,0 +1,100 @@
...
```
mattsse: we could name this `take_revealed_if` ?

mattsse (approved):
lgtm
optional naming suggestion

PR #17260, opened 2025-07-07
feat(test): allow to create testing nodes with specific datadir

fgimenez:
towards #17197

this will allow us to spin up testing nodes with imported datadirs


mattsse (approved):
makes sense

PR #17261, opened 2025-07-07
feat(test): spin up e2e test nodes with imported data

fgimenez:
towards #17197

just spins up the nodes with imported rlp data, will be integrated with the e2e framework in a followup


mattsse (approved):
love how easy it is to follow what the tests are doing

PR #17262, opened 2025-07-07
fix(trie): add lower subtrie root paths to upper subtrie prefix set

shekhirin:
https://github.com/paradigmxyz/reth/blob/2614c5a36bdd3183e3aa0ac276a96fc2952342af/crates/trie/sparse-parallel/src/trie.rs#L1009-L1013

If we don't do this, when calculating the upper subtrie hashes, we would skip lower subtrie root extension and branch nodes here https://github.com/paradigmxyz/reth/blob/2614c5a36bdd3183e3aa0ac276a96fc2952342af/crates/trie/sparse-parallel/src/trie.rs#L1631-L1635 https://github.com/paradigmxyz/reth/blob/2614c5a36bdd3183e3aa0ac276a96fc2952342af/crates/trie/sparse-parallel/src/trie.rs#L1684-L1688


mediocregopher (approved):

Rjected (approved):
i see, this makes sense

PR #17263, opened 2025-07-07
feat(test): add apply_with_import method to e2e Setup

fgimenez:
towards #17197
requires #17261

* `Setup::apply_with_import` similar to regular `apply` but using nodes created with `setup_engine_with_chain_import`
* extracts common import testing utils to separate module
* adds e2e action to assert chain tip by number
* adds e2e test that exercises import


crates/e2e-test-utils/src/testsuite/examples.rs
```diff
@@ -15,6 +16,107 @@ use reth_chainspec::{ChainSpecBuilder, MAINNET};
...
+    let rlp_path = temp_dir.path().join("test_chain.rlp");
+    write_blocks_to_rlp(&test_blocks, &rlp_path)?;
+
+    // Create setup with imported chain
+    let mut setup =
+        Setup::default().with_chain_spec(chain_spec).with_network(NetworkSetup::single_node());
+
+    // Create environment and apply setup with import
+    let mut env = crate::testsuite::Environment::<EthEngineTypes>::default();
+    setup.apply_with_import::<EthereumNode>(&mut env, &rlp_path).await?;
```
mattsse, line 53: nice

crates/e2e-test-utils/src/testsuite/examples.rs
```diff
@@ -15,6 +16,107 @@ use reth_chainspec::{ChainSpecBuilder, MAINNET};
...
+            alloy_rpc_types_eth::Transaction,
+            alloy_rpc_types_eth::Block,
+            alloy_rpc_types_eth::Receipt,
+            alloy_rpc_types_eth::Header,
+        >::block_by_number(
+            &client.rpc,
+            alloy_eips::BlockNumberOrTag::Number(10),
+            true, // Include full transaction details
+        )
+        .await;
```
mattsse, line 84: maybe we could even replace this with alloy provider which would likely be less verbose, perhaps we add some utils to the env/node_clients for that, similar to what we have for ServerHandles

mattsse (approved):
very nice

crates/e2e-test-utils/src/testsuite/examples.rs
fgimenez: makes total sense, on it

fgimenez: done here https://github.com/paradigmxyz/reth/pull/17263/commits/95d3916d957c65822e952156ec4fe4b759d94991

PR #17267, opened 2025-07-07
chore(trie): add Either type for SparseTrieInterface

Rjected:
Adds an `Either<A, B>` type which we could use for the sparse trie in the payload processor. The idea is to switch on the cli flag in the payload processor, to build either a `ParallelSparseTrie` or a `RevealedSparseTrie` as the accounts trie. The idea is that we would use this as the account trie type.

ref https://github.com/paradigmxyz/reth/pull/17266 for `from_root`
based on https://github.com/paradigmxyz/reth/pull/17268

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10625,6 +10625,7 @@ dependencies = [
  "assert_matches",
  "auto_impl",
  "codspeed-criterion-compat",
+ "either",
  "itertools 0.14.0",
  "metrics",
  "pretty_assertions",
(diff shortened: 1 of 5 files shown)


crates/trie/sparse/src/either.rs
```diff
@@ -0,0 +1,141 @@
...
+use reth_trie_common::{Nibbles, TrieNode};
+
+/// An `Either` type for types that implement the sparse trie interface.
+#[derive(Debug)]
+pub enum Either<A, B> {
+    /// The left sparse trie
+    Left(A),
+    /// The right sparse trie
+    Right(B),
+}
```
mattsse: should we impl this directly for either::Either ?

Rjected: hmm yeah let's just do it for either::Either

Rjected: now doing this for `either::Either`, also re-exporting

shekhirin (approved):

PR #17269, opened 2025-07-07
feat(trie): add generics to SparseTrieTask

Rjected:
We will need these to configure whether we will use the parallel sparse trie, or the regular sparse trie.


shekhirin (approved):
@mediocregopher PTAL

mattsse (approved):
seems fine, we could maybe make SparseTrieInterface object-safe by accepted dyn blindedprovider, but this seems fine

PR #17274, opened 2025-07-08
docs(guides): Add history expiry

### LINKED ISSUE #16547: Document all things history expiry
### Describe the change

The goal is to ensure that there is sufficient documentation about how history expiry works with Reth and fill in any and all missing spots.

### Additional context

_No response_

RomanHodulak:
Closes #16547

Adds a documentation chapter on history expiry serving as a guide.

Notes:
* `export-era` is not yet implemented #17132 , maybe that section should be added later?
* Pruning: maybe add a section on pruning expired historical data? Matt could help here.
* I couldn't find a proper section for this so I added directory `guides/`. Should it be put somewhere else?
* I came up with the format, should it be any different?

### DIFF
--- a/docs/vocs/docs/pages/guides/history-expiry.mdx
+++ b/docs/vocs/docs/pages/guides/history-expiry.mdx
@@ -1,0 +1,49 @@
+---
+description: Usage of tools for importing, exporting and pruning historical blocks
+---
+
+# History Expiry
+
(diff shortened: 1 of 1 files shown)


mattsse (approved):
great

mattsse:
will add the section about pruning

PR #17286, opened 2025-07-08
perf: remove block cloning from `is_descendant` check

rkrasiuk:
## Description

Remove unnecessary block cloning from `is_descendant` check.

Screenshot 2025-07-08 at 18 43 46


shekhirin (approved):
LGTM

PR #17287, opened 2025-07-08
feat: make build_receipt infallable

### LINKED ISSUE #17231: Make build_receipt infallible
### Describe the feature

This currently returns an error: https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/rpc/rpc-eth-types/src/receipt.rs#L12-L13

if:

https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/rpc/rpc-eth-types/src/receipt.rs#L24-L26

fails,

we can change the tx argument:

https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/rpc/rpc-eth-types/src/receipt.rs#L13-L13

to `Recovered<&T>` instead and require that the caller already provides the recovered address

now that we operate on the recovered block when we invoke this:

https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/primitives-traits/src/block/recovered.rs#L327-L333

we can use the helper function:

https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/primitives-traits/src/block/recovered.rs#L301-L305

and remove recovery entirely

but then need to do recovery here:

https://github.com/paradigmxyz/reth/blob/d101fb7b900ca998cb117a1bd52c92c1e134c211/crates/rpc/rpc/src/eth/helpers/receipt.rs#L36-L36
 

### Additional context

_No response_

catconcat:
Closes #17231

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -416,7 +416,7 @@ clippy-op-dev:
 	--all-features
 
 lint-codespell: ensure-codespell
-	codespell --skip "*.json" --skip "./testing/ef-tests/ethereum-tests"
+	# codespell --skip "*.json" --skip "./testing/ef-tests/ethereum-tests"
 
 ensure-codespell:
 	@if ! command -v codespell &> /dev/null; then \
--- a/Makefile
+++ b/Makefile
@@ -416,7 +416,7 @@ clippy-op-dev:
 	--all-features
 
 lint-codespell: ensure-codespell
-	# codespell --skip "*.json" --skip "./testing/ef-tests/ethereum-tests"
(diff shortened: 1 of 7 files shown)


mattsse (approved):
ty!

I applied some additional optimizations that this now unlocked 

sorry @iTranscend I know you were assigned to this -.-

@catconcat let's respect the process next time and ping the assignee

cc @hai-rise this should give a nice speedup for all full block receipt responses

PR #16926, opened 2025-06-18
feat(node): add `LocalPayloadAttributesAddOn` trait

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/16873.

Currently working on the last point : 
- [x] add `UnsupportedLocalPayloadAttributesAddOn` and `UnsupportedPayloadAttributesBuilder` - still needed to change trait signature from `impl PayloadAttributesBuilder` to associated type `Builder`

PR #17297, opened 2025-07-08
feat(sdk): add `local_payload_attributes_builder` to `DebugNodeLauncher`

### LINKED ISSUE #16873: Promote LocalPayloadAttributesBuilder to NodeAddOn instead
### Describe the feature

we're currently requiring:

https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/node/builder/src/launch/engine.rs#L82-L82

which is problematic because this leads to workarounds like this:

https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/engine/local/src/payload.rs#L69-L71

Instead we can Introduce a `trait LocalPayloadAttributesAddon` that returns `impl PayloadAttributesBuilder` for `fn local_payload_attributes_builder()`, similar to:

https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/node/api/src/node.rs#L128-L132

Also fallibe, so that we also have a way to not support it

for this

https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/payload/primitives/src/traits.rs#L125-L125

then we can replace this

https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/node/builder/src/launch/engine.rs#L259-L259

## TODO
* introduce trait
* Integrate in EthereumAddons and OpAddOns, similar to https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/ethereum/node/src/node.rs#L276-L276
* replace https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/node/builder/src/launch/engine.rs#L82-L84 with an additional `+ LocalPayloadAttributesAddon`
* Remove workaround for unsupported https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/crates/engine/local/src/payload.rs#L71-L71 and instead add a `struct UnsupportedLocalPayloadAttributesAddOn` that returns an error instead, so that we can get rid of https://github.com/paradigmxyz/reth/blob/cb11ab04751cebea8c39cb8b44ee0559695c9e44/examples/custom-engine-types/src/main.rs#L79-L80

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/16873 (attempt 2)

Follow-up for https://github.com/paradigmxyz/reth/pull/16926#pullrequestreview-2971264558 (and https://github.com/paradigmxyz/reth/pull/16926).


crates/node/builder/src/launch/debug.rs
```diff
@@ -56,6 +57,18 @@ pub trait DebugNode<N: FullNodeComponents>: Node<N> {
...
+    ///
+    ///  It will be used by the `LocalMiner` when dev mode is enabled.
+    ///
+    /// The builder is responsible for creating the payload attributes that define how blocks should
+    /// be constructed during local mining.
+    fn local_payload_attributes_builder(
+        chain_spec: &Self::ChainSpec,
+    ) -> impl PayloadAttributesBuilder<
+        <<Self as reth_node_api::NodeTypes>::Payload as PayloadTypes>::PayloadAttributes,
+    >;
```
lean-apple, line 71: @klkvr I don't think it's possible to have a default impl because of the `PayloadAttributes`, or is it?

crates/engine/local/src/payload.rs
```diff
@@ -64,19 +64,3 @@ where
         }
     }
 }
-
-/// A temporary workaround to support local payload engine launcher for arbitrary payload
-/// attributes.
-// TODO(mattsse): This should be reworked so that LocalPayloadAttributesBuilder can be implemented
-// for any
-pub trait UnsupportedLocalAttributes: Send + Sync + 'static {}
```
mattsse, line 72: great

crates/node/builder/src/launch/debug.rs
mattsse: possibly, but this seems fine

mattsse (commented):
this does make a lot of sense

wdyt @klkvr

crates/ethereum/node/src/node.rs
```diff
@@ -353,6 +355,14 @@ impl<N: FullNodeComponents<Types = Self>> DebugNode<N> for EthereumNode {
     fn rpc_to_primitive_block(rpc_block: Self::RpcBlock) -> reth_ethereum_primitives::Block {
         rpc_block.into_consensus().convert_transactions()
     }
+
+    fn local_payload_attributes_builder(
+        chain_spec: &Self::ChainSpec,
+    ) -> impl PayloadAttributesBuilder<
+        <<Self as reth_node_api::NodeTypes>::Payload as PayloadTypes>::PayloadAttributes,
```
klkvr:
```suggestion
        <Self::Payload as PayloadTypes>::PayloadAttributes,
```

crates/optimism/node/src/node.rs
```diff
@@ -246,6 +247,14 @@ where
     fn rpc_to_primitive_block(rpc_block: Self::RpcBlock) -> reth_node_api::BlockTy<Self> {
         rpc_block.into_consensus()
     }
+
+    fn local_payload_attributes_builder(
+        chain_spec: &Self::ChainSpec,
+    ) -> impl PayloadAttributesBuilder<
+        <<Self as reth_node_api::NodeTypes>::Payload as PayloadTypes>::PayloadAttributes,
```
klkvr:
```suggestion
        <Self::Payload as PayloadTypes>::PayloadAttributes,
```

klkvr (approved):
nice, lgtm

PR #17299, opened 2025-07-08
chore: remove Beacon from type names

### LINKED ISSUE #17176: Rename `Beacon` prefixed types
### Describe the feature

`beacon` refers to the regular engine api and we're no longer particularly tied to this

hence we should drop the prefix

## TODO
* rename `BeaconConsensusEngineHandle` and `BeaconConsensusEngineEvent` 
* keep an alias so that this change becomes non breaking (must also be added where these types are re-exported)

### Additional context

_No response_

petarjuki7:
Fixes #17176 

Renamed `BeaconConsensusEngineHandle` and `BeaconConsensusEngineEvent` to `ConsensusEngineHandle` and `ConsensusEngineEvent`

Added type aliases where needed


mattsse:
very sorry, this drifted too much unfortunately

PR #17301, opened 2025-07-09
fix(docs): correct duplicated function reference in documentation

Bilogweb3:
Fix a duplicated function reference in the documentation where `try_recover_unchecked` was mentioned twice. Replace the first occurrence with `try_recover` to correctly show the two different recovery functions available


mattsse (approved):
ah yes, this should be try_recover

PR #17313, opened 2025-07-09
perf(txpool): reduce one BTree lookup operation in `add_transaction`

nekomoto911:


mattsse (changes requested):
is this essentialy just replacing the assert!

I think we can achieve the same thing by downgrading the assert to a debug_assert!

can we do that instead?

nekomoto911:
Yes, I've considered changing it to `debug_assert!`, but I'm unsure if this `assert!` could truly be triggered in a production environment here. If a duplicated transaction were to occur due to some vulnerability and didn't panic immediately, could it lead to serious problems?

mattsse:
I think we can downgrade this to debug_assert, initially this was only used as sanity check but I think we covered all checks 

we can add cfg(debug_assertions) to the test so that it keeps catching the panic.

imo not panicking in the case this is reachable would be better

mattsse (approved):

PR #17317, opened 2025-07-09
fix(trie): ParallelSparseTrie::update_leaf: add moved leaves to the prefix set

mediocregopher:
In the case where `leafA` starts off in the upper trie, and then `update_leaf(leafB)` results in `leafA` being moved to a lower trie, we need to add `leafA` to the prefix set even if it wasn't previously there. If we don't then its lower subtrie won't necessarily be included in the call to `update_subtrie_hashes`, resulting in a missing hash.

This also includes a fuzz test which is similar to the existing one, exccept that it tests against the existing serial sparse trie implementation, rather than the hash builder.


shekhirin (approved):
nice find

PR #17322, opened 2025-07-09
fix(trie): move masks to `ParallelSparseTrie` level

shekhirin:
If branch masks are kept on the subtrie level, we may query upper subtrie masks when calling `rlp_node` for a lower subtrie https://github.com/paradigmxyz/reth/blob/162568b297e4e20cd95530d78d2c0ed9ef26c124/crates/trie/sparse-parallel/src/trie.rs#L958-L969 https://github.com/paradigmxyz/reth/blob/162568b297e4e20cd95530d78d2c0ed9ef26c124/crates/trie/sparse-parallel/src/trie.rs#L1858-L1864

This PR moves the branch masks to `ParallelSparseTrie` level and passes them as read-only references to all necessary functions.

### DIFF
--- a/crates/era-utils/src/export.rs
+++ b/crates/era-utils/src/export.rs
@@ -1,7 +1,7 @@
 //! Logic to export from database era1 block history
 //! and injecting them into era1 files with `Era1Writer`.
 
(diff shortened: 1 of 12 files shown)


mediocregopher (approved):
:rocket:

PR #17326, opened 2025-07-09
chore(trie): Implement ParallelSparseTrie::find_leaf

### LINKED ISSUE #17196: Implement `find_leaf` for `ParallelSparseTrie`
### Describe the feature

The regular sparse trie has a method for traversing the trie and finding specific leaves, generating either a proof of existence or exclusion proof.

https://github.com/paradigmxyz/reth/blob/051cef53bc7372c22d545a583427b0e598529e87/crates/trie/sparse/src/trie.rs#L1439-L1456

We will need to add a method like this for the parallel sparse trie.

Any tests should be ported over.

### Additional context

_No response_

mediocregopher:
This implements the final method of the SparseTrieInterface trait. We don't need this method in the immediate term, but it'd be better to round out the trait rather than have someone get rugged by it not being there later.

The implementation is able to take advantage of `find_next_to_leaf`, and so is fairly simple. The tests are all taken from the RevealedSparseTrie more or less verbatim.

Fixes #17196


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -628,10 +633,84 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
+        full_path: &Nibbles,
+        expected_value: Option<&Vec<u8>>,
+    ) -> Result<LeafLookup, LeafLookupError> {
+        // Inclusion proof
+        //
+        // First, do a quick check if the value exists in either the upper or lower subtrie's values
+        // map. We assume that if there exists a leaf node, then its value will be in the `values`
+        // map.
+        if let Some(actual_value) =
+            [Some(self.upper_subtrie.as_ref()), self.lower_subtrie_for_path(full_path)]
```
shekhirin: can just use `self.subtrie_for_path` I think?

shekhirin (approved):
Very nice that we can reuse `find_next_to_leaf`. Only one nit, otherwise LGTM.

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I don't think so unfortunately, because we only know the full_path (node path + key), but `subtrie_for_path` can only take the node path. The node might be in either the upper or lower subtrie depending on the overall structure.

mediocregopher:
@shekhirin I confirmed with @kevaundray that we can remove `diverged_at` entirely, so I went ahead and did that, simplifies things nicely.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -657,10 +662,67 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
```
mattsse: I think this could also be `std::iter::once().chain(`?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -657,10 +662,67 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
+                .iter()
+                .flatten()
+                .filter_map(|subtrie| subtrie.inner.values.get(full_path))
+                .next()
+        {
+            // We found the leaf, check if the value matches (if expected value was provided)
+            return expected_value
+                .is_none_or(|v| v == actual_value)
+                .then_some(LeafLookup::Exists)
```
mattsse, line 681: just to double check, `None == LeafLookup::Exists`?

mattsse (commented):
made sense, I have one suggestion and 1 q

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: Based on [the original implementation](https://github.com/paradigmxyz/reth/blob/d7aa75137944c1c8f82f2fe7170cb9fcbdeccc3b/crates/trie/sparse/src/trie.rs#L934-L962) yes. Basically None == "we don't know/care about the value, just that the leaf exists"

PR #17339, opened 2025-07-09
chore: add helpers for disabling read-tx timeout

### LINKED ISSUE #17328: Example how to remove timeout on StateProviderBox
### Describe the change

Hi,

As per our discussion with @mattsse in TG, creating an issue on how to disable the timeout on `StateProviderBox` for the external use.

### Additional context

_No response_

mattsse:
Improves the documentation for disabling read transaction timeouts to make it clearer when and why users might want to disable this safety mechanism.

The updated documentation:
- Clarifies that timeouts are automatic by default
- Explains that free list growth occurs when the database is actively being modified
- Makes it clear when it's safe to disable timeouts (static databases with no writes)
- Uses more concise and direct language

Also adds a mutable setter method `max_read_transaction_duration()` to match the existing builder pattern in `DatabaseArguments`.

Closes #17328

### DIFF
--- a/crates/storage/provider/src/providers/database/builder.rs
+++ b/crates/storage/provider/src/providers/database/builder.rs
@@ -160,7 +160,7 @@ impl ReadOnlyConfig {
     ///
     /// Caution: Keeping database transaction open indefinitely can cause the free list to grow if
     /// changes to the database are made.
-    pub fn disable_long_read_transaction_safety(mut self) -> Self {
+    pub const fn disable_long_read_transaction_safety(mut self) -> Self {
         self.db_args.max_read_transaction_duration(Some(MaxReadTransactionDuration::Unbounded));
         self
     }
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/providers/database/builder.rs
```diff
@@ -71,6 +74,29 @@ impl<N> ProviderFactoryBuilder<N> {
     ///         .unwrap();
     /// }
     /// ```
+    ///
+    /// # Open an instance with disabled read-transaction timeout
+    ///
+    /// By default, read transactions are automatically terminated after a timeout to prevent
+    /// database free list growth, which can degrade performance when the database is actively
```
Rjected: I think the main reason is to prevent db growth now, we don't have degradation due to large freelist anymore

crates/storage/provider/src/providers/database/builder.rs
```diff
@@ -129,6 +155,16 @@ impl ReadOnlyConfig {
         Self::from_dirs(datadir.join("db"), datadir.join("static_files"))
     }
 
+    /// Disables long-lived read transaction safety guarantees, such as backtrace recording and
```
Rjected: Would remove backtrace recording notes because it's only relevant if the feature is enabled

Rjected (commented):
some comments on the docs

Rjected (approved):
LGTM

PR #17343, opened 2025-07-10
fix(docs): fix wrong path in pruning.mdx

nekomoto911:


mattsse (commented):
@yash-atreya is this right?

PR #17347, opened 2025-07-10
chore: make `OpAddonsBuilder` generic over middleware

### LINKED ISSUE #17273: make `OpAddOnsBuilder` generic over middleware
### Describe the feature

It looks like I still got the type constraint when injecting middleware, for e.g.:
```bash
110 |             addons.with_rpc_middleware(
    |                    ------------------- arguments to this method are incorrect
111 |                 SlowRequestLogMiddleware::new(self.slow_request_log_config.clone())
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Identity`, found `SlowRequestLogMiddleware`
    |
help: the return type of this call is `SlowRequestLogMiddleware` due to the type of the argument passed
```

https://github.com/paradigmxyz/reth/blob/36d568a404df3bca08b217ebade69cf4f1b38d0e/crates/optimism/node/src/node.rs#L285-L291

I think we need to add `Identity` here and update the `builder()` method to be generic over `RpcMiddleware`
https://github.com/paradigmxyz/reth/blob/36d568a404df3bca08b217ebade69cf4f1b38d0e/crates/optimism/node/src/node.rs#L301-L308

Also, update the `OpAddOnsBuilder` to support custom middleware.

I have a working solution already, I'd like to take this if you're open to it!

### Additional context

relating to #17076

thaodt:
resolve #17273

### DIFF
--- a/crates/optimism/node/src/node.rs
+++ b/crates/optimism/node/src/node.rs
@@ -295,6 +295,7 @@ impl<N, NetworkT> Default
         OpEthApiBuilder<NetworkT>,
         OpEngineValidatorBuilder,
         OpEngineApiBuilder<OpEngineValidatorBuilder>,
+        Identity,
     >
 where
     N: FullNodeComponents<Types: NodeTypes>,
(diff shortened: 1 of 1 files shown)


mattsse (approved):
clean

PR #17351, opened 2025-07-10
feat: support isthmus in reth-bench

### LINKED ISSUE #17087: Make bench command operate on PayloadTypes

klkvr:
Adds support for isthmus `withdrawalsRoot` field to reth-bench by determining whether we're on OP chain and using `OpExecutionPayloadV4` type if relevant

closes https://github.com/paradigmxyz/reth/issues/17087

### DIFF
--- a/bin/reth-bench/src/valid_payload.rs
+++ b/bin/reth-bench/src/valid_payload.rs
@@ -2,53 +2,19 @@
 //! response. This is useful for benchmarking, as it allows us to wait for a payload to be valid
 //! before sending additional calls.
 
-use alloy_eips::eip7685::RequestsOrHash;
-use alloy_primitives::B256;
-use alloy_provider::{ext::EngineApi, Network, Provider};
+use alloy_provider::{ext::EngineApi, network::AnyRpcBlock, Network, Provider};
 use alloy_rpc_types_engine::{
-    ExecutionPayload, ExecutionPayloadInputV2, ExecutionPayloadSidecar, ExecutionPayloadV1,
-    ExecutionPayloadV3, ForkchoiceState, ForkchoiceUpdated, PayloadAttributes, PayloadStatus,
+    ExecutionPayload, ExecutionPayloadInputV2, ForkchoiceState, ForkchoiceUpdated,
+    PayloadAttributes, PayloadStatus,
 };
 use alloy_transport::TransportResult;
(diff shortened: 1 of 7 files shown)


bin/reth-bench/src/valid_payload.rs
```diff
@@ -323,16 +194,49 @@ pub(crate) async fn call_new_payload<N, P: EngineApiValidWaitExt<N>>(
...
+    version: EngineApiMessageVersion,
+    params: serde_json::Value,
+) -> TransportResult<()> {
+    let method = match version {
+        EngineApiMessageVersion::V1 => "engine_newPayloadV1",
+        EngineApiMessageVersion::V2 => "engine_newPayloadV2",
+        EngineApiMessageVersion::V3 => "engine_newPayloadV3",
+        EngineApiMessageVersion::V4 => "engine_newPayloadV4",
+        EngineApiMessageVersion::V5 => "engine_newPayloadV5",
+    };
```
mattsse: should we add this as a helper fn to EngineApiMessageVersion?

mattsse (approved):
this should do it and make it work again with op chains

@shekhirin do you want to try this first or can we merge?

bin/reth-bench/src/valid_payload.rs
klkvr: yep makes sense 75ab97476

shekhirin (approved):
We're back

Image

PR #17352, opened 2025-07-10
perf(trie): reuse update action buffers in parallel sparse trie processing

### LINKED ISSUE #17348: Preserve SparseTrieUpdatesAction vecs across root computations
### Describe the feature

We are seeing that 5% of the entire SparseTrieTask is being taken by deallocations within `apply_subtrie_update_actions`. By adding some buffer reuse we could easily factor that out.

### Additional context

_No response_

mediocregopher:
This change optimizes memory allocation during sparse trie hash updates by reusing `SparseTrieUpdatesAction` buffers across operations instead of allocating new vectors each time.

Fixes #17348

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -49,6 +49,9 @@ pub struct ParallelSparseTrie {
     branch_node_tree_masks: HashMap<Nibbles, TrieMask>,
     /// When a bit is set, the corresponding child is stored as a hash in the database.
     branch_node_hash_masks: HashMap<Nibbles, TrieMask>,
+    /// Reusable buffer pool used for collecting [`SparseTrieUpdatesAction`]s during hash
+    /// computations.
+    update_actions_buffers: Vec<Vec<SparseTrieUpdatesAction>>,
 }
 
 impl Default for ParallelSparseTrie {
@@ -63,6 +66,7 @@ impl Default for ParallelSparseTrie {
             updates: None,
             branch_node_tree_masks: HashMap::default(),
             branch_node_hash_masks: HashMap::default(),
+            update_actions_buffers: Vec::default(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -604,25 +609,38 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
 
         // Return updated subtries back to the trie after executing any actions required on the
         // top-level `SparseTrieUpdates`.
-        for (index, subtrie, update_actions) in rx {
-            self.apply_subtrie_update_actions(update_actions);
+        for (index, subtrie, update_actions_buf) in rx {
+            if let Some(mut update_actions_buf) = update_actions_buf {
+                self.apply_subtrie_update_actions(
+                    #[allow(clippy::iter_with_drain)]
+                    update_actions_buf.drain(..),
```
shekhirin, line 639: nice, this leaves the allocation in place

shekhirin (approved):
LGTM, clean change

crates/trie/sparse-parallel/src/trie.rs
mattsse, line 639: this is the same as .clear() minus the clippy

> // A full range clears the vector, like `clear()` does
v.drain(..);

mattsse (commented):
smol nit

crates/trie/sparse-parallel/src/trie.rs
shekhirin: `clear()` doesn't return the values, we need to pass them into `apply_subtrie_update_actions`

mediocregopher: ^ Right, I think what clippy was yelling at me to do here was pass in `update_actions_buf.iter()` then `update_actions_buf.clear()` afterwards, which would have been logically equivalent, but it seemed more efficient to do both as a single step using drain.

PR #17354, opened 2025-07-10
chore: add helper convert into error object

mattsse:
this is more helpeful because from impls dont give proper intellisense


Rjected (approved):
yep seems reasonable

PR #17355, opened 2025-07-10
feat(trie): wire parallel trie config to PayloadProcessor

### LINKED ISSUE #17163: Have a way to enable parallel sparse trie in engine tests
### Describe the feature

Same as we have with forcing the old non- sparse trie way of calculating the state root

### Additional context

_No response_

Rjected:
We have to use static dispatch because the types that implement `SparseTrieInterface` require constructors that prevent object safety. This PR is a set of helper methods to prevent the complexity of dispatching to multiple sparse trie implementations from being in the `spawn` method.

There are four methods introduced:
* `spawn_sparse_trie_task` which takes in the config option for the parallel sparse trie
* `dispatch_trie_spawn`, which extracts the actual trie type (serial or parallel), and dispatches it to a spawn method depending on the type.
* `spawn_trie_task`, which spawns the task and convert the trie in the result back into the proper type

This previously used `Either`, but because a generic `Either<A,B>` can't impl `From<A>` and `From<B>`, we had to have two `spawn_` methods, one for the serial sparse trie and one for the parallel sparse trie. So I removed it and replaced it with `ConfiguredSparseTrie`.

Fixes https://github.com/paradigmxyz/reth/issues/17163 , we would just enable the value in the `TreeConfig` on the e2e tests


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -196,24 +206,20 @@ where
...
-
         // wire the sparse trie to the state root response receiver
         let (state_root_tx, state_root_rx) = channel();
-        self.executor.spawn_blocking(move || {
-            let res = sparse_trie_task.run();
-            let _ = state_root_tx.send(res);
-        });
+
+        // Take the stored sparse trie
+        let stored_trie = self.sparse_trie.take();
```
mediocregopher, line 213: We could probably enshrine it better in the types, but the SparseTrie at this point is always Blind, so a lot of the checks downstream of this could probably be simplified.

mattsse (commented):
I see, the changes in the payload processor setup are basically just so that we handle matches

shekhirin (approved):
LGTM, unfortunate that we can't do `impl From for Either` :( https://github.com/rayon-rs/either/issues/33

One nit: can we refactor generics out from `SparseTrieTask` in the future, so that we have just `trie: SparseStateTrie<ConfiguredSparseTrie, SerialSparseTrie>`?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -318,6 +324,144 @@ where
...
+            }
+            Some(SparseTrie::Blind(Some(boxed))) => {
+                self.dispatch_trie_spawn(
+                    *boxed,
+                    sparse_trie_rx,
+                    proof_task_handle,
+                    state_root_tx,
+                    // this is blind
+                    false,
+                );
```
shekhirin: can probably be merged into into match arm with `stored_trie.is_revealed()` as last argument

Rjected: yep, collapsed

PR #17357, opened 2025-07-10
feat(cli): add enable-parallel-sparse-trie flag

Rjected:
Adds an `--engine.enable-parallel-sparse-trie` flag.

### DIFF
--- a/crates/node/core/src/args/engine.rs
+++ b/crates/node/core/src/args/engine.rs
@@ -34,6 +34,10 @@ pub struct EngineArgs {
     #[arg(long = "engine.disable-caching-and-prewarming")]
     pub caching_and_prewarming_disabled: bool,
 
+    /// Enable the parallel sparse trie in the engine.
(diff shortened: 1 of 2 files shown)


crates/node/core/src/args/engine.rs
```diff
@@ -34,6 +34,10 @@ pub struct EngineArgs {
     #[arg(long = "engine.disable-caching-and-prewarming")]
     pub caching_and_prewarming_disabled: bool,
 
+    /// Enable the parallel sparse trie in the engine.
+    #[arg(long = "engine.enable-parallel-sparse-trie", default_value = "false")]
+    pub enable_parallel_sparse_trie: bool,
+
```
shekhirin: for consistency with other perf flags

```suggestion
    /// Enable the parallel sparse trie in the engine.
    #[arg(long = "engine.parallel-sparse-trie", default_value = "false")]
    pub parallel_sparse_trie_enabled: bool,

```

shekhirin (approved):

PR #17361, opened 2025-07-10
chore(trie): impl TrieUpdates::drain_into_sorted

Rjected:
Realized that `clear` for these methods is not sufficient if we want to keep the space around, because we sort the `TrieUpdates` and `Arc` the sorted version.


mediocregopher (approved):
Method LGTM, is it intentional that nothing is using this yet?

mattsse (approved):

shekhirin (approved):

Rjected:
> Method LGTM, is it intentional that nothing is using this yet?

Yeah, it should be a small change to use this and reuse trie input allocs once the other drain method is in though

PR #17340, opened 2025-07-10
feat: make ethereum Cli generic over node

klkvr:
Based on https://github.com/paradigmxyz/reth/pull/17330

Removes `execution` and `build_block` debug commands and adds helper methods to `Cli` in a non-breaking way allowing to reuse the logic for custom nodes.

PR #17330, opened 2025-07-09
feat: `re-execute` command

klkvr:
Closes #17285 

Introduces `reth re-execute` commands which should be run in scope of an archive node and will re-execute entire chain (or selected block range) on top of historical state read from database.

This command allows to test historical sync correctness substantially faster than via full resync.

PR #17363, opened 2025-07-11
feat: make ethereum Cli generic over node and remove debug commands

klkvr:
applies changes from https://github.com/paradigmxyz/reth/pull/17340 which weren't merged in scope of https://github.com/paradigmxyz/reth/pull/17330 because of my mistake

Removes all debug commands this time


mattsse (approved):
good stuff

PR #17367, opened 2025-07-11
fix(`docs`): rustdocs search functionality

yash-atreya:
- Fix search index in the vocs output before injecting
- Correct the paths

### DIFF
--- a/docs/vocs/scripts/inject-cargo-docs.ts
+++ b/docs/vocs/scripts/inject-cargo-docs.ts
@@ -70,13 +70,18 @@ async function injectCargoDocs() {
       .replace(/data-static-root-path="\.\.\/static\.files\/"/g, `data-static-root-path="${BASE_PATH}/static.files/"`)
       
       // Fix search index paths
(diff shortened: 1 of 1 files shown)


docs/vocs/scripts/inject-cargo-docs.ts
```diff
@@ -70,13 +70,18 @@ async function injectCargoDocs() {
       .replace(/data-static-root-path="\.\.\/static\.files\/"/g, `data-static-root-path="${BASE_PATH}/static.files/"`)
       
       // Fix search index paths
-      .replace(/data-search-index-js="([^"]+)"/g, `data-search-index-js="${BASE_PATH}/static.files/$1"`)
+      .replace(/data-search-index-js="[^"]+"/g, `data-search-index-js="${BASE_PATH}/search-index.js"`)
```
mattsse, line 73: ah yes, this matches with what I saw in the browser

mattsse (approved):

PR #17369, opened 2025-07-11
perf(engine): clear accounts trie in background to not block state root

### LINKED ISSUE #17349: Perform `take_cleared_accounts_trie` in the background
### Describe the feature

7% of the SparseSubtrieTask is taken on `take_cleared_accounts_trie`, which is a cleanup step after computation is done. Theoretically this could be done asynchronously, if not the taking then at least the clearing which is what takes the most time.

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/17349

This PR moves `SparseTrie::clear()` call for accounts trie reuse out of the hot path, meaning that the state root will be returned ASAP, and the clearing will happen in background.


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -197,7 +200,16 @@ where
...
-        let sparse_trie = self.sparse_trie.take();
+        let sparse_trie = match self.accounts_trie_rx.try_recv() {
+            Ok(trie) => Some(trie),
+            // If the accounts trie is not available, create a new one.
+            //
+            // This may happen in two cases:
+            // 1. It's the first payload that we're processing.
+            // 2. The accounts trie hasn't been returned yet.
+            Err(TryRecvError::Empty) => None,
+            Err(TryRecvError::Disconnected) => panic!("sparse trie channel disconnected"),
```
Rjected: doing `try_recv` makes sense here

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -108,7 +108,26 @@ where
...
+    ) -> (Result<StateRootComputeOutcome, ParallelStateRootError>, SparseTrie<A>) {
+        // run the main loop to completion
+        let result = self.run_to_completion();
+        // take the account trie so that we can reuse its already allocated data structures.
+        let trie = self.trie.take_accounts_trie();
+
+        (result, trie)
+    }
+
+    /// Run the sparse trie task to completion.
```
Rjected: these both now have docs mentioning running the sparse trie to completion

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -108,7 +108,26 @@ where
...
+    ///
+    /// See [`Self::run`] for more information.
+    fn run_to_completion(&mut self) -> Result<StateRootComputeOutcome, ParallelStateRootError> {
```
Rjected: tbh i think only one of these should have `run` in it? `run` and `run_to_completion` seem kinda similar

Rjected (commented):
mainly comments about the method naming and docs, otherwise lgtm

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -210,9 +222,12 @@ where
 
         // wire the sparse trie to the state root response receiver
         let (state_root_tx, state_root_rx) = channel();
+        let sparse_trie_tx = self.accounts_trie_tx.clone();
         self.executor.spawn_blocking(move || {
-            let res = sparse_trie_task.run();
+            let (res, trie) = sparse_trie_task.run();
             let _ = state_root_tx.send(res);
+            // Clear accounts trie and return it back to the channel
+            let _ = sparse_trie_tx.send(trie.clear());
```
mattsse: I see, this now first sends the stateroot then the cleared trie

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -68,9 +68,10 @@ where
     precompile_cache_disabled: bool,
     /// Precompile cache map.
     precompile_cache_map: PrecompileCacheMap<SpecFor<Evm>>,
-    /// A cleared sparse trie, kept around to be reused for the state root computation so that
-    /// allocations can be minimized.
-    sparse_trie: Option<SparseTrie>,
+    /// A cleared accounts sparse trie, kept around to be reused for the state root computation so
+    /// that allocations can be minimized.
+    accounts_trie_tx: mpsc::SyncSender<SparseTrie>,
+    accounts_trie_rx: mpsc::Receiver<SparseTrie>,
```
mattsse: from what I can tell, this is using a channel like a mutex?

this should not introduce any race condition issues because the trie must be empty here so it's either available or not

but should this be an `Arc<Mutex` instead?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -197,7 +200,16 @@ where
         });
 
         // take the sparse trie if it was set
-        let sparse_trie = self.sparse_trie.take();
+        let sparse_trie = match self.accounts_trie_rx.try_recv() {
```
mediocregopher: It would be extremely unlikely admittedly, but if it took more time for the ParallelSparseTrie to clear itself than it did for the next payload to come in for execution then the channel would not be written to and this would hit the Empty case. If that happens then a new trie is created, but the old one will eventually write itself to this channel. At that point there will always be one extra trie floating around, effectively a memory leak.

We could get around this by pre-sending an empty trie onto the channel at initialization, and doing `recv()` here.

Again, not likely to happen

shekhirin: that's a valid point that's possible in `reth-bench` runs, will just switch to an `Arc<Mutex<_>>` as per Matt's suggestion

mattsse: yeah I believe this is ruled out because we still need to handle the stateroot result on the receiver and the sparse trie is immediately set after sending the stateroot

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -210,9 +210,12 @@ where
 
         // wire the sparse trie to the state root response receiver
         let (state_root_tx, state_root_rx) = channel();
+        let accounts_trie = self.accounts_trie.clone();
         self.executor.spawn_blocking(move || {
-            let res = sparse_trie_task.run();
+            let (res, trie) = sparse_trie_task.run();
             let _ = state_root_tx.send(res);
+            // Clear accounts trie and return it back to the payload processor
+            accounts_trie.lock().replace(trie);
```
mattsse: we still need to clear this here?

mattsse (changes requested):

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2314,9 +2314,6 @@ where
                                 "State root task returned incorrect state root"
                             );
                         }
-
-                        // hold on to the sparse trie for the next payload
-                        self.payload_processor.set_sparse_trie(trie);
```
mattsse, line 2319: 👍

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -210,9 +210,12 @@ where
 
         // wire the sparse trie to the state root response receiver
         let (state_root_tx, state_root_rx) = channel();
+        let accounts_trie = self.accounts_trie.clone();
         self.executor.spawn_blocking(move || {
-            let res = sparse_trie_task.run();
+            let (res, trie) = sparse_trie_task.run();
             let _ = state_root_tx.send(res);
+            // Clear accounts trie and return it back to the payload processor
```
mattsse: replace return with store it

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -210,9 +210,12 @@ where
 
         // wire the sparse trie to the state root response receiver
         let (state_root_tx, state_root_rx) = channel();
+        let accounts_trie = self.accounts_trie.clone();
```
mattsse: pedantic style nit, use `Arc::clone`

mattsse (approved):
lgtm, pending @Rjected

crates/engine/tree/src/tree/payload_processor/mod.rs
shekhirin: we take it from the payload processor above, and now we return it. why not?

Rjected (approved):
lgtm

PR #17371, opened 2025-07-11
perf(trie): reuse allocated trie input in payload processor

### LINKED ISSUE #17292: Reuse `TrieInput` to reduce allocations
### Describe the feature

Profile: https://share.firefox.dev/3IjDzET

Image

When constructing a `TrieInput`, we do multiple re-allocations when doing `HashedPostState::extend_ref` and `TrieUpdates::extend_ref`. What we should do instead, is to clear the `TrieInput` struct, save it in the `PayloadProcessor`, and reuse on the next block to minimize allocations.

### Additional context

_No response_

Rjected:
Fixes https://github.com/paradigmxyz/reth/issues/17292

Puts a `TrieInput` in the payload processor, letting us reuse it. Changes `new_from_input` to use the `drain_into_sorted` methods.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1400,6 +1400,7 @@ where
                 self.persisting_kind_for(block.recovered_block().header()),
                 self.provider.database_provider_ro()?,
                 block.recovered_block().parent_hash(),
+                None,
             )?;
             // Extend with block we are generating trie updates for.
             trie_input.append_ref(block.hashed_state());
@@ -2174,8 +2175,10 @@ where
         persisting_kind: PersistingKind,
         provider: TP,
         parent_hash: B256,
+        allocated_trie_input: Option<TrieInput>,
     ) -> ProviderResult<TrieInput> {
-        let mut input = TrieInput::default();
+        // get allocated trie input or use a default trie input
+        let mut input = allocated_trie_input.unwrap_or_default();
 
         let best_block_number = provider.best_block_number()?;
 
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -304,8 +304,8 @@ where
     }
 
     /// Takes the trie input from the inner payload processor, if it exists.
-    pub const fn take_trie_input(&mut self) -> Option<TrieInput> {
-        self.trie_input.take()
+    pub fn take_trie_input(&mut self) -> Option<TrieInput> {
+        self.trie_input.take().map(|input| input.cleared())
     }
 
     /// Returns the cache for the given parent hash.
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -304,8 +304,8 @@ where
     }
 
     /// Takes the trie input from the inner payload processor, if it exists.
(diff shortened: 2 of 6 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -1452,6 +1452,7 @@ where
                 self.persisting_kind_for(block.recovered_block().header()),
                 self.provider.database_provider_ro()?,
                 block.recovered_block().parent_hash(),
+                None,
```
shekhirin, line 1403: why not reuse the allocated trie input here as well?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -309,6 +314,11 @@ where
         CacheTaskHandle { cache, to_prewarm_task: Some(to_prewarm_task), cache_metrics }
     }
 
+    /// Takes the trie input from the inner payload processor, if it exists.
+    pub const fn take_trie_input(&mut self) -> Option<TrieInput> {
+        self.trie_input.take()
```
shekhirin, line 308: this should also clear the trie input, right? Otherwise its contents reused

Rjected: added this

crates/engine/tree/src/tree/mod.rs
Rjected: we don't have the allocated trie input here, the payload processor is in the
engine validator

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -298,6 +303,11 @@ where
         CacheTaskHandle { cache, to_prewarm_task: Some(to_prewarm_task), cache_metrics }
     }
 
+    /// Takes the trie input from the inner payload processor, if it exists.
+    pub fn take_trie_input(&mut self) -> Option<TrieInput> {
+        self.trie_input.take().map(|input| input.cleared())
+    }
```
mattsse: I see, this is the only way how we can obtain the trieinput again and we ensure here that it's cleared

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -82,16 +82,20 @@ pub(super) struct MultiProofConfig<Factory> {
...
+    ) -> (TrieInput, Self) {
+        let config = Self {
             consistent_view,
-            nodes_sorted: Arc::new(input.nodes.into_sorted()),
-            state_sorted: Arc::new(input.state.into_sorted()),
-            prefix_sets: Arc::new(input.prefix_sets),
-        }
+            nodes_sorted: Arc::new(input.nodes.drain_into_sorted()),
+            state_sorted: Arc::new(input.state.drain_into_sorted()),
+            prefix_sets: Arc::new(input.prefix_sets.clone()),
```
mattsse, line 96: we don't clear the prefix set here?

this deviates from `TrieInput::cleared`
should we call cleared here?

mattsse (commented):
one q

crates/engine/tree/src/tree/payload_processor/multiproof.rs
Rjected: hmm, yeah maybe we should call cleared here instead of in `take_trie_input`

Rjected: put the `cleared` call here

mattsse (approved):

PR #17372, opened 2025-07-11
perf(trie): parallelize ParallelSparseTrie::reveal_nodes

### LINKED ISSUE #17350: Perform reveal_nodes in parallel
### Describe the feature

34% of the SparseTrieTask is taken on node revealing now. We can reveal nodes on the ParallelSparseTrie in parallel. This will require grouping the nodes to be revealed and then splitting the work via rayon, similar to how `update_subtrie_hashes` works.

About a third of the time spent in node revealing is actually spent reserving space in the `nodes` HashMaps within the subtries. When we group the nodes-to-be-revealed we should also call `reserve` on each HashMap ahead of time in order to pre-reserve this space in one go, which should save some back-and-forth there.

### Additional context

_No response_

mediocregopher:
Replace `reveal_node` with `reveal_nodes` on the SparseTrieInterface, allowing us to reveal nodes in parallel on the ParallelSparseTrie.

This change required some refactoring in the SparseStateTrie as well. We were previously sorting proof nodes in SparseStateTrie prior to passing them individually to `reveal_node`. We now sort within `reveal_nodes`, as that allows us to use different sorting algorithms in each trie implementation.

Additionally in SparseStateTrie, the `filter_revealed_nodes` function has been refactored to both filter proof nodes and map them into the new RevealedSparseNode struct. This saves a couple Vec allocations, and cleans up logic related to pulling out the root node to treat it specially.

Fixes #17350


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+            let mut num_upper_nodes = 0;
+            let upper_nodes =
+                nodes.iter().take_while(|n| SparseSubtrieType::path_len_is_upper(n.path.len()));
+
+            for node in upper_nodes {
+                // TODO once reveal_node is not on the trait we can probably make it accept
+                // references, and skip these clones.
+                self.reveal_node(node.path, node.node.clone(), node.masks.clone())?;
+                num_upper_nodes += 1;
+            }
```
Rjected: I think we can do something like this
```suggestion
    // find where upper nodes end
    let split_pos = nodes
        .iter()
        .position(|n| !SparseSubtrieType::path_len_is_upper(n.path.len()))
        .unwrap_or(nodes.len());
    
    // get owned nodes by draining
    for node in nodes.drain(..split_pos) {
        self.reveal_node(node.path, node.node, node.masks)?;
    }
```
and we can create the slice after

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+            nodes_slice = &nodes_slice[num_upper_nodes..];
+        }
+
+        #[cfg(not(feature = "std"))]
+        // Reveal lower subtrie nodes serially if nostd
+        for node in nodes_slice {
+            for node in nodes {
+                self.reveal_node(node.path, node.node.clone(), node.masks.clone())?;
+            }
+        }
```
Rjected: why the double loop here? nodes should also still be owned here so I think we can just do `for node in nodes`?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+                })
+                .collect();
+
+            // Take the lower subtries in the same order that the nodes were grouped into, so that
+            // the two can be zipped together. This also must be collected into a Vec for rayon's
+            // `zip` to be happy.
+            let lower_subtries: Vec<_> = node_groups
+                .iter()
+                .map(|nodes| {
+                    let node = nodes.first().expect("no empty groups");
```
Rjected: could probably switch to `nodes[0]` given chunk_by only returns non empty slices

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+                    let Some(idx) = SparseSubtrieType::from_path(&node.path).lower_index() else {
+                        panic!("upper subtrie node {node:?} found amongst lower nodes");
```
Rjected: i think can be replaced with expect?
```suggestion
                    let idx = SparseSubtrieType::from_path(&node.path).lower_index().expect("upper subtrie node {node:?} found with lower nodes");
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+                self.lower_subtries[subtrie_idx] = LowerSparseSubtrie::Revealed(subtrie);
+                if res.is_err() {
+                    any_err = res;
+                }
+            }
+
+            // TODO we're just returning the last seen error here, not sure if there's a more
+            // idiomatic thing to do...
+            any_err
+        }
```
Rjected: i would probably just loop through the `rx` results here and if we get an err, return early (maybe just by doing `res?`). and the fn would have Ok(()) at the end

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+                    subtrie.nodes.reserve(nodes.len());
+
+                    for node in nodes {
+                        // Reveal each node in the subtrie, returning early on any errors
+                        let res = subtrie.reveal_node(node.path, &node.node, node.masks.clone());
+                        if res.is_err() {
+                            return (subtrie_idx, subtrie, res)
+                        }
+                    }
+                    (subtrie_idx, subtrie, Ok(()))
```
Rjected: can we skip the `return` here and just return `(subtrie_idx, subtrie, res)`?

Rjected (changes requested):
I have mostly style comments, the idea makes sense to me so far, I think the `reveal_nodes` method makes sense direction wise.

crates/trie/sparse-parallel/src/trie.rs
shekhirin: we can probably even do something like
```rust
let nodes_iter = nodes.into_iter();
let upper_nodes = nodes_iter.by_ref().take_while(|n| SparseSubtrieType::path_len_is_upper(n.path.len()));
for lower_node in nodes { ... }
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+            }
+        }
+
+        let mut nodes_slice = nodes.as_slice();
+
+        // Due to the sorting, if there are any nodes to be revealed in the upper trie then the
+        // first node will be an upper trie node. Handle all upper trie reveals using
+        // `ParallelSparseTrie::reveal_node`, which deals with revealing recursively into
+        // lower-subtries as necessary.
+        if SparseSubtrieType::from_path(&nodes[0].path).lower_index().is_none() {
```
shekhirin: can introduce a new method

```suggestion
        if SparseSubtrieType::from_path(&nodes[0].path).is_upper() {
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1234,135 @@ impl ParallelSparseTrie {
...
+
+        // Sort nodes first by their subtrie, and secondarily by their path. This allows for
+        // grouping nodes by their subtrie using `chunk_by`.
+        nodes.sort_unstable_by(
+            |RevealedSparseNode { path: path_a, .. }, RevealedSparseNode { path: path_b, .. }| {
+                let subtrie_type_a = SparseSubtrieType::from_path(path_a);
+                let subtrie_type_b = SparseSubtrieType::from_path(path_b);
+                subtrie_type_a.cmp(&subtrie_type_b).then(path_a.cmp(path_b))
+            },
+        );
```
shekhirin: we need to not forget to remove this https://github.com/paradigmxyz/reth/blob/332c65661778b71d8fdc655c5178cb9d4718feb7/crates/trie/sparse/src/state.rs#L901 to avoid sorting twice

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I think draining off the front of the Vec will cause the rest of the elements to get shifted left, so that's incurring a potentially large copy if there's a lot of proofs. The original way also only iterates over the upper nodes once, not twice, so I think overall will be more performant (though admittedly much uglier)

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: Good catch, mistake on my part from copy-pasting I think

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: I think `expect` doesn't allow a formatting parameter, and I didn't want to incur the formatting unless it was necessary, so I went with the `else`/`panic!`

crates/trie/sparse-parallel/src/trie.rs
shekhirin: I think my approach with iterator also iterates only once and doesn't copy? Or does it not work for some reason?

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: We need to read all results off of `rx` in order to get the taken lower sparse trie's back and put them back into the ParallelSparseTrie, otherwise the whole struct will be in an inconsistent state.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1304,68 @@ impl ParallelSparseTrie {
...
+    /// # Arguments
+    ///
+    /// * `path` - The path where the node should be revealed
+    /// * `node` - The trie node to reveal
+    /// * `masks` - Trie masks for branch nodes
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if successful, or an error if the node was not revealed.
+    fn reveal_upper_node(
```
mediocregopher, line 1327: This is the same as the old `reveal_node` implementation, minus these lines at the top:
```rust
        // Store masks
        if let Some(tree_mask) = masks.tree_mask {
            self.branch_node_tree_masks.insert(path, tree_mask);
        }
        if let Some(hash_mask) = masks.hash_mask {
            self.branch_node_hash_masks.insert(path, hash_mask);
        }

        if let Some(subtrie) = self.lower_subtrie_for_path_mut(&path) {
            return subtrie.reveal_node(path, &node, masks);
        }
```

crates/trie/sparse/src/state.rs
```diff
@@ -506,7 +502,11 @@ where
                         storage_trie_entry
                             .as_revealed_mut()
                             .ok_or(SparseTrieErrorKind::Blind)?
-                            .reveal_node(path, trie_node, TrieMasks::none())?;
+                            .reveal_nodes(vec![RevealedSparseNode {
```
mediocregopher: Not ideal, but only used in stateless atm

crates/trie/sparse-parallel/src/trie.rs
Rjected: ah I see

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: This ended up all being changed and this new method wouldn't get used here anyway

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: Sorry @shekhirin , GH hadn't updated with your comment and I didn't see it 🤦

I ended up finding yet a third approach, which is kind of a mix of me and Dan's. It allows for reserving capacity on the upper trie's nodes map without introducing any new allocs. We're back to two iterations over upper nodes though... can't have everything 🤷

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: What would `res` be outside of the loop?

mediocregopher:
@shekhirin @Rjected RFR, I left some TODO questions around the code in places where I think possibly more could/should be done

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -95,62 +98,135 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
+        // them off from the rest to be handled specially by
+        // `ParallelSparseTrie::reveal_upper_node`.
+        let num_upper_nodes = nodes
+            .iter()
+            .position(|n| !SparseSubtrieType::path_len_is_upper(n.path.len()))
+            .unwrap_or(nodes.len());
+
+        let all_nodes = nodes.as_slice();
+        let upper_nodes = &all_nodes[..num_upper_nodes];
+        let lower_nodes = &all_nodes[num_upper_nodes..];
```
shekhirin: this is basically `partition`?

shekhirin: actually no, `partition` will iterate until the end, but we can just find the position of first matching element and split on it

it should be `split_once`, but it's [unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.split_once), so nvm 😄

crates/trie/sparse-parallel/src/trie.rs
Rjected: oh, I think i misread this, this makes sense

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -95,62 +98,135 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
+                })
+                .collect();
+
+            // Take the lower subtries in the same order that the nodes were grouped into, so that
+            // the two can be zipped together. This also must be collected into a Vec for rayon's
+            // `zip` to be happy.
+            let lower_subtries: Vec<_> = node_groups
+                .iter()
+                .map(|nodes| {
+                    let node = &nodes[0]; // SAFETY: chunk_by won't produce empty groups
```
```suggestion
                    // NOTE: chunk_by won't produce empty groups
                    let node = &nodes[0];
```
I prefer `SAFETY` to be reserved for `unsafe` usage

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2160,6 +2298,24 @@ impl SparseSubtrieType {
...
 
+impl Ord for SparseSubtrieType {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match (self, other) {
+            (Self::Upper, Self::Upper) => Ordering::Equal,
+            (Self::Upper, Self::Lower(_)) => Ordering::Less,
+            (Self::Lower(_), Self::Upper) => Ordering::Greater,
+            (Self::Lower(idx_a), Self::Lower(idx_b)) if idx_a == idx_b => Ordering::Equal,
+            (Self::Lower(idx_a), Self::Lower(idx_b)) => idx_a.cmp(idx_b),
+        }
```
Rjected, line 2312: would be nice to have a comment here explaining the order

crates/trie/sparse-parallel/src/trie.rs
Rjected: I see, that seems fine then

crates/trie/sparse-parallel/src/trie.rs
Rjected: Does this work?
```suggestion
        let upper_nodes = &nodes[..num_upper_nodes];
        let lower_nodes = &nodes[num_upper_nodes..];
```

crates/trie/sparse/src/state.rs
```diff
@@ -178,31 +177,35 @@ where
         proof: impl IntoIterator<Item = (Nibbles, Bytes)>,
     ) -> SparseStateTrieResult<()> {
         assert!(!self.retain_updates);
+        // TODO this method doesn't seem to be used, can we remove it?
```
Rjected: I see, yeah it does look unused, I think we can file an issue to remove, wdyt @shekhirin ?

crates/trie/sparse/src/state.rs
```diff
@@ -217,14 +220,15 @@ where
...
```
Rjected: same for this

crates/trie/sparse/src/state.rs
Rjected: maybe we should keep the old `reveal_node` for this?

crates/trie/sparse/src/state.rs
```diff
@@ -377,56 +384,50 @@ where
...
         #[cfg(feature = "metrics")]
         {
             self.metrics.increment_total_storage_nodes(_total_nodes as u64);
             self.metrics.increment_skipped_storage_nodes(_skipped_nodes as u64);
         }
-        let mut nodes = nodes.into_iter().peekable();
 
-        if let Some(root_node) = Self::validate_root_node_decoded(&mut nodes)? {
+        if let Some(root_node) = root_node {
+            // Perform sanity check.
```
Rjected: this is because we separate the root node from `nodes`?

crates/trie/sparse/src/state.rs
```diff
@@ -753,7 +736,7 @@ where
             trace!(target: "trie::sparse", ?address, "Retrieving storage root from account leaf to update account");
             // The account was revealed, either...
             if let Some(value) = self.get_account_value(&address) {
-                // ..it exists and we should take its current storage root or...
+                // ..it exists and we should take it's current storage root or...
```
Rjected:
```suggestion
                // ..it exists and we should take its current storage root or...
```

crates/trie/sparse/src/state.rs
```diff
@@ -312,47 +319,47 @@ where
...
-                        branch_node_tree_masks.get(&path).copied(),
-                    )
-                } else {
-                    (None, None)
-                };
+            // Reveal root node if it wasn't already.
+            let trie =
+                self.state.reveal_root(root_node.node, root_node.masks, self.retain_updates)?;
 
-                trace!(target: "trie::sparse", ?path, ?node, ?hash_mask, ?tree_mask, "Revealing account node");
```
Rjected, line 266: let's add a trace level log that shows number of nodes being revealed?

crates/trie/sparse/src/state.rs
```diff
@@ -377,56 +384,50 @@ where
...
-                let (hash_mask, tree_mask) = if let TrieNode::Branch(_) = node {
-                    (
-                        storage_subtree.branch_node_hash_masks.get(&path).copied(),
-                        storage_subtree.branch_node_tree_masks.get(&path).copied(),
-                    )
-                } else {
-                    (None, None)
-                };
-
-                trace!(target: "trie::sparse", ?account, ?path, ?node, ?hash_mask, ?tree_mask, "Revealing storage node");
```
Rjected, line 340: re-adding a trace log showing num storage nodes being revealed would also be nice here

crates/trie/sparse/src/state.rs
```diff
@@ -889,16 +878,31 @@ fn filter_revealed_nodes(
...
 
         result.new_nodes += 1;
+
         // If it's a branch node, increase the number of new nodes by the number of children
         // according to the state mask.
-        if let TrieNode::Branch(branch) = &node {
+        if let TrieNode::Branch(branch) = &proof_node {
             result.new_nodes += branch.state_mask.count_ones() as usize;
         }
+        // TODO should we increase by 1 if it's an extension?
```
Rjected: wdym by this?

crates/trie/sparse/src/trie.rs
```diff
@@ -413,146 +414,12 @@ impl SparseTrieInterface for SerialSparseTrie {
...
-        // If the node is already revealed and it's not a hash node, do nothing.
-        if self.nodes.get(&path).is_some_and(|node| !node.is_hash()) {
-            return Ok(())
-        }
-
-        if let Some(tree_mask) = masks.tree_mask {
-            self.branch_node_tree_masks.insert(path, tree_mask);
+    fn reveal_nodes(&mut self, mut nodes: Vec<RevealedSparseNode>) -> SparseTrieResult<()> {
+        nodes.sort_unstable_by(|a, b| a.path.cmp(&b.path));
+        self.nodes.reserve(nodes.len());
```
Rjected: During multiproof revealing we reserve nodes then call reveal_nodes, so not sure if we should include this

Rjected (changes requested):
I have a few more style comments, and some comments about the TODOs

crates/trie/sparse/src/state.rs
```diff
@@ -178,31 +177,35 @@ where
...
             let node = TrieNode::decode(&mut &bytes[..])?;
-            trie.reveal_node(path, node, TrieMasks::none())?;
+            nodes_to_reveal.push(RevealedSparseNode { path, node, masks: TrieMasks::none() });
 
             // Track the revealed path.
             self.revealed_account_paths.insert(path);
         }
 
-        Ok(())
+        Ok(trie.reveal_nodes(nodes_to_reveal)?)
```
shekhirin: can we make `reveal_nodes` accept an iterator instead, so that we can filter and decode the nodes from `proof` without collecting it into a vec?

crates/trie/sparse/src/state.rs
shekhirin: If it's an extension node, it will have one child, so we need to do `result.new_nodes += 1`. Similar to branch nodes with N children.

crates/trie/sparse/src/state.rs
mediocregopher: This was a check that `validate_root_node_decoded` used to do, I'm not sure what the context is which makes it necessary, i just moved it here for backwards compatibility.

crates/trie/sparse/src/state.rs
mediocregopher: I will try it... we will have to iterate over `proof` once to pull out the root node anyway, but that can be done as a one-off operation separate from filtering/mapping

crates/trie/sparse/src/trie.rs
mediocregopher: True, it's not necessary here, will remove it :+1:

crates/trie/sparse/src/state.rs
shekhirin: yeah, we can do it the same way as it was before using `peekable`, that will also make the diff smaller

mediocregopher: Hmmm actually, the first thing that both reveal_nodes implementations need to do is sort the proofs, so they'll need to collect into a Vec anyway. Seems more flexible to do it outside the method.

(also fyi the method this comment is on is one of the ones which isn't used, we should focus on `reveal_decoded_account_multiproof` and `reveal_decoded_storage_multiproof` I think)

crates/trie/sparse/src/state.rs
mediocregopher: I re-introduced the `reveal_node` method to the trait, with a default implementation which is essentially this. For SerialSparseTrie we can easily implement the method to call the same inner method that `reveal_nodes` calls, for ParallelSparseTrie it would require two different implementations, probably not worth it until we can show that it is, and maybe better to focus on refactoring this code to use `reveal_nodes` anyway.

crates/trie/sparse/src/state.rs
Rjected: I see, I think this check makes sense anyways

crates/trie/sparse/src/state.rs
```diff
@@ -312,47 +312,36 @@ where
...
-                trace!(target: "trie::sparse", ?path, ?node, ?hash_mask, ?tree_mask, "Revealing account node");
-                trie.reveal_node(path, node, TrieMasks { hash_mask, tree_mask })?;
+            // Reserve the capacity for new nodes ahead of time, if the trie implementation
+            // supports doing so.
+            trie.reserve_nodes(new_nodes);
 
-                // Track the revealed path.
-                self.revealed_account_paths.insert(path);
-            }
+            trace!(target: "trie::sparse", "Revealing {} account nodes", nodes.len());
```
Rjected:
```suggestion
            trace!(target: "trie::sparse", total_nodes = ?nodes.len(), "Revealing account nodes");
```

crates/trie/sparse/src/state.rs
```diff
@@ -377,56 +366,39 @@ where
...
-                    (None, None)
-                };
-
-                trace!(target: "trie::sparse", ?account, ?path, ?node, ?hash_mask, ?tree_mask, "Revealing storage node");
-                trie.reveal_node(path, node, TrieMasks { hash_mask, tree_mask })?;
-
-                // Track the revealed path.
-                revealed_nodes.insert(path);
-            }
+            trace!(target: "trie::sparse", ?account, "Revealing {} storage nodes", nodes.len());
```
Rjected:
```suggestion
            trace!(target: "trie::sparse", ?account, total_nodes = ?nodes.len(), "Revealing storage nodes");
```

crates/trie/sparse/src/trie.rs
```diff
@@ -412,147 +413,20 @@ impl SparseTrieInterface for SerialSparseTrie {
...
 
-        if let Some(tree_mask) = masks.tree_mask {
-            self.branch_node_tree_masks.insert(path, tree_mask);
-        }
-        if let Some(hash_mask) = masks.hash_mask {
-            self.branch_node_hash_masks.insert(path, hash_mask);
+    fn reveal_nodes(&mut self, mut nodes: Vec<RevealedSparseNode>) -> SparseTrieResult<()> {
+        nodes.sort_unstable_by(|a, b| a.path.cmp(&b.path));
+        for node in &nodes {
+            self.reveal_node_ref(node.path, &node.node, &node.masks)?;
```
Rjected: if we have an owned nodes here, can we not just have the regular `reveal_node` fn, which doesn't clone?

Rjected (commented):
Just a few final comments, otherwise this looks good

crates/trie/sparse/src/trie.rs
mediocregopher: Yup good point :+1:

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -95,62 +98,135 @@ impl SparseTrieInterface for ParallelSparseTrie {
...
+            // the two can be zipped together. This also must be collected into a Vec for rayon's
+            // `zip` to be happy.
+            let lower_subtries: Vec<_> = node_groups
+                .iter()
+                .map(|nodes| {
+                    // NOTE: chunk_by won't produce empty groups
+                    let node = &nodes[0];
+                    let Some(idx) = SparseSubtrieType::from_path(&node.path).lower_index() else {
+                        panic!("upper subtrie node {node:?} found amongst lower nodes");
+                    };
```
shekhirin: can do `unwrap_or_else` with panic in the closure

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1230,6 +1306,68 @@ impl ParallelSparseTrie {
...
+    /// * `masks` - Trie masks for branch nodes
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if successful, or an error if the node was not revealed.
+    fn reveal_upper_node(
+        &mut self,
+        path: Nibbles,
+        node: &TrieNode,
+        masks: &TrieMasks,
```
shekhirin: we can derive `Copy` for `TrieMasks`, because it's just two `Option<TrieMask>` fields

crates/trie/sparse/src/traits.rs
```diff
@@ -83,7 +73,25 @@ pub trait SparseTrieInterface: Sized + Debug + Send + Sync {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `nodes` - The nodes to be revealed, each having a path and optional set of branch node
+    ///   masks. The nodes will be unsorted.
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if successful, or an error if any of the nodes was not revealed.
+    fn reveal_nodes(&mut self, nodes: Vec<RevealedSparseNode>) -> SparseTrieResult<()>;
```
shekhirin, line 94: we can benefit from accepting an iterator here, so that we don't do a vec heap allocation in `reveal_node` and instead just use `core::iter::once`. Can be done later in a separate PR.

shekhirin: oh, that's probably not that straightforward, because we benefit from `nodes` being a vector in `ParallelSparseTrie::reveal_nodes` implementation

crates/trie/sparse/src/trie.rs
```diff
@@ -548,14 +548,22 @@ impl SparseTrieInterface for SerialSparseTrie {
...
+                    self.values.insert(full, leaf.value.clone());
                 }
             },
         }
 
         Ok(())
     }
 
+    fn reveal_nodes(&mut self, mut nodes: Vec<RevealedSparseNode>) -> SparseTrieResult<()> {
+        nodes.sort_unstable_by(|a, b| a.path.cmp(&b.path));
```
shekhirin:
```suggestion
        nodes. sort_unstable_by_key(|node| node.path);
```

shekhirin (approved):

Rjected (approved):
LGTM

Issue #17228, opened 2025-07-04
Add Middleware generic to AuthServerConfig


mattsse:
### Describe the feature

as prep for #17195

we should introduce a Middleware = Identity generic to:

https://github.com/paradigmxyz/reth/blob/342bab5e82bfb4d3fcf185fded4f8944f93e90eb/crates/rpc/rpc-builder/src/auth.rs#L24-L24

that we then use to bootstrap this:

https://github.com/paradigmxyz/reth/blob/342bab5e82bfb4d3fcf185fded4f8944f93e90eb/crates/rpc/rpc-builder/src/auth.rs#L54-L56

see also:

https://github.com/paradigmxyz/reth/blob/342bab5e82bfb4d3fcf185fded4f8944f93e90eb/crates/rpc/rpc-builder/src/lib.rs#L1241-L1242

https://github.com/paradigmxyz/reth/blob/342bab5e82bfb4d3fcf185fded4f8944f93e90eb/crates/rpc/rpc-builder/src/lib.rs#L1312-L1312

iTranscend:
can i?

emmajam:
Hey! We're doing some spring cleaning on our issues backlog 🧹 Closing old issues to keep things tidy. If this is still relevant, please feel free to re-open — we appreciate your contribution!

PR #17373, opened 2025-07-11
feat: add Middleware generic to AuthServerConfig

iTranscend:
towards https://github.com/paradigmxyz/reth/issues/17228


crates/rpc/rpc-builder/src/auth.rs
```diff
@@ -85,12 +104,13 @@ impl AuthServerConfig {
...
 /// Builder type for configuring an `AuthServerConfig`.
 #[derive(Debug)]
-pub struct AuthServerConfigBuilder {
+pub struct AuthServerConfigBuilder<Middleware = Identity> {
     socket_addr: Option<SocketAddr>,
     secret: JwtSecret,
     server_config: Option<ServerConfigBuilder>,
     ipc_server_config: Option<IpcServerBuilder<Identity, Identity>>,
     ipc_endpoint: Option<String>,
+    rpc_middleware: Option<Middleware>,
```
mattsse: we can remove the option here because Identity is more or less the same

crates/rpc/rpc-builder/src/auth.rs
```diff
@@ -48,17 +54,30 @@ impl AuthServerConfig {
     }
 
     /// Convenience function to start a server in one step.
-    pub async fn start(self, module: AuthRpcModule) -> Result<AuthServerHandle, RpcError> {
-        let Self { socket_addr, secret, server_config, ipc_server_config, ipc_endpoint } = self;
+    pub async fn start(self, module: AuthRpcModule) -> Result<AuthServerHandle, RpcError>
+    where
+        Middleware: RethRpcMiddleware,
```
mattsse: for this usecase we need to install this as the `tower::ServiceBuilder::new()` middleware instead

crates/rpc/rpc-builder/src/auth.rs
```diff
@@ -48,17 +54,30 @@ impl AuthServerConfig {
...
+            ipc_server_config,
+            ipc_endpoint,
+            rpc_middleware,
+        } = self;
 
         // Create auth middleware.
         let middleware =
             tower::ServiceBuilder::new().layer(AuthLayer::new(JwtAuthValidator::new(secret)));
 
+        let rpc_middleware = RpcServiceBuilder::default().layer(rpc_middleware);
```
mattsse, line 88: this we can also do but not in this pr which should focus on the tower middleware

crates/rpc/rpc-builder/src/config.rs
```diff
@@ -212,7 +212,7 @@ impl RethRpcServerConfig for RpcServerArgs {
     fn auth_server_config(&self, jwt_secret: JwtSecret) -> Result<AuthServerConfig, RpcError> {
         let address = SocketAddr::new(self.auth_addr, self.auth_port);
 
-        let mut builder = AuthServerConfig::builder(jwt_secret).socket_addr(address);
+        let mut builder = AuthServerConfig::<Identity>::builder(jwt_secret).socket_addr(address);
```
mattsse: this should not be necessary if we move the builder fn into `impl AuthServerConfig` scope

crates/rpc/rpc-builder/tests/it/utils.rs
```diff
@@ -33,7 +34,7 @@ pub const fn test_address() -> SocketAddr {
 
 /// Launches a new server for the auth module
 pub async fn launch_auth(secret: JwtSecret) -> AuthServerHandle {
-    let config = AuthServerConfig::builder(secret).socket_addr(test_address()).build();
+    let config = AuthServerConfig::<Identity>::builder(secret).socket_addr(test_address()).build();
```
mattsse: same

mattsse (changes requested):
good start but the middleware we want to target in this pr is for the tower middleware so we need a different trait bound and perhaps a helper trait for that

mattsse (approved):
okay, we actually also want the rpc middleware here as well, so this doesn't hurt

PR #16646, opened 2025-06-04
feat: make RpcAddOns::launch_add_ons_with composable

fgimenez:
Closes #15985

PR #17376, opened 2025-07-11
feat(sdk): make engine API (auth server) optional for custom consensus integrations

fgimenez:
followup #16646 

When integrating reth with non-Ethereum consensus mechanisms, the engine API server is unnecessary overhead. These changes allow SDK users to opt out of running the auth server when their consensus implementation doesn't require it.

* added `--disable-auth-server` CLI argument to optionally disable the auth server
* modified `RpcAddOns` to support conditional auth server launching
* auth server remains enabled by default


mattsse (approved):
this is an easy workaround
cc @loocapro

PR #17378, opened 2025-07-12
refactor(docs): improve regex pattern robustness in CLI help parser

reallesee:
Replace hardcoded spaces with whitespace metacharacter in regex pattern to enhance parsing reliability and follow regex best practices. This change ensures the parser can handle various CLI output formats while maintaining backward compatibility.


mattsse (commented):
we use this for args formatting and this is working fine

PR #17379, opened 2025-07-12
fix: correct command syntax in benchmark documentation

Fallengirl:
Add missing space between 'latest' and '--full' in the cast block command example


mattsse (commented):
thanks, I combined these

PR #17381, opened 2025-07-12
docs: Fix broken internal links in history-expiry guide

VolodymyrBg:
Replaced the internal link to node in history-expiry.mdx with a site-relative path (/cli/reth/node). This ensures the link works correctly on the synchronized documentation website and directs users to the appropriate node synchronization pipeline section.


mattsse (commented):
thanks, I combined these

PR #17382, opened 2025-07-12
refactor(rpc): change receipt to `Cow<R>` for `build_receipt`

### LINKED ISSUE #17334: Use `Cow` for build_receipt  receipt: &R,
### Describe the feature

https://github.com/paradigmxyz/reth/blob/959323fa6fa31401ab9def6bfc41f67878062088/crates/rpc/rpc-eth-types/src/receipt.rs#L16-L16

We have instances where we already own the receipt 

https://github.com/paradigmxyz/reth/blob/959323fa6fa31401ab9def6bfc41f67878062088/crates/rpc/rpc/src/eth/helpers/receipt.rs#L39-L39

and where we borrow:

https://github.com/paradigmxyz/reth/blob/959323fa6fa31401ab9def6bfc41f67878062088/crates/rpc/rpc/src/eth/helpers/block.rs#L62-L62

we can avoid the clone call:

https://github.com/paradigmxyz/reth/blob/959323fa6fa31401ab9def6bfc41f67878062088/crates/rpc/rpc-eth-types/src/receipt.rs#L51-L56

for `Cow::Owned`

## TODO
* change argument to `Cow<R>`, update all calls and get rid of the logs clone for Cow::Owned

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17334.


crates/rpc/rpc-eth-types/src/receipt.rs
```diff
@@ -48,28 +51,31 @@ where
...
-            log_index: Some((num_logs + tx_log_idx) as u64),
-            removed: false,
-        })
-        .collect();
-
-    let rpc_receipt = alloy_rpc_types_eth::Receipt {
-        status: receipt.status_or_post_state(),
-        cumulative_gas_used: receipt.cumulative_gas_used(),
-        logs,
+    macro_rules! build_rpc_logs {
```
mattsse, line 54: another way to write this would be using a closure for the map, but we can remove this once

https://github.com/alloy-rs/alloy/pull/2712 is released so this is fine for now

mattsse (approved):

crates/rpc/rpc-eth-types/src/receipt.rs
lean-apple: Will keep in mind to do the change once there is `alloy`'s release 🫡

lean-apple: https://github.com/paradigmxyz/reth/pull/17569.

PR #17383, opened 2025-07-12
feat: add `into_logs()` to `TxReceipt` for `Receipt`/`OpReceipt`

lean-apple:
Override default [`into_logs()`](https://github.com/alloy-rs/alloy/blob/1f157f03af5d9472044d6972ad84056f89db7d4c/crates/consensus/src/receipt/mod.rs#L94) to avoid cloning, necessary for `to_vec()`


mattsse (approved):
nice catch

PR #17388, opened 2025-07-13
perf(blob): optimize blob store gets

nekomoto911:
This PR avoids redundant blob deep copy operations in `get_one` and `get_all`, and reduces the critical section of the `blob_cache` lock in `get_all`.

### DIFF
--- a/crates/transaction-pool/src/blobstore/disk.rs
+++ b/crates/transaction-pool/src/blobstore/disk.rs
@@ -424,10 +424,9 @@ impl DiskFileBlobStoreInner {
         if let Some(blob) = self.blob_cache.lock().get(&tx) {
             return Ok(Some(blob.clone()))
         }
-        let blob = self.read_one(tx)?;
 
-        if let Some(blob) = &blob {
-            let blob_arc = Arc::new(blob.clone());
+        if let Some(blob) = self.read_one(tx)? {
+            let blob_arc = Arc::new(blob);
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/blobstore/disk.rs
```diff
@@ -542,11 +541,18 @@ impl DiskFileBlobStoreInner {
...
             return Ok(res)
         }
+        let from_disk = from_disk
+            .into_iter()
+            .map(|(tx, data)| {
+                let data = Arc::new(data);
+                res.push((tx, data.clone()));
+                (tx, data)
+            })
+            .collect::<Vec<_>>();
```
mattsse, line 551: I think this makes sense, we're okay with doing this extra vec upfront if it reduces time(lock)

crates/transaction-pool/src/blobstore/disk.rs
```diff
@@ -424,10 +424,9 @@ impl DiskFileBlobStoreInner {
         if let Some(blob) = self.blob_cache.lock().get(&tx) {
             return Ok(Some(blob.clone()))
         }
-        let blob = self.read_one(tx)?;
 
-        if let Some(blob) = &blob {
-            let blob_arc = Arc::new(blob.clone());
+        if let Some(blob) = self.read_one(tx)? {
```
mattsse, line 428: nice

mattsse (approved):
ty!

PR #17392, opened 2025-07-14
perf: optimize txpool_status RPC by avoiding full transaction collection

keanji-x:
## Summary
Optimizes the `txpool_status` RPC endpoint by introducing a new method to get transaction counts directly without collecting all transactions into memory.

## Changes
- Added `queue_and_pendng_txn_count()` method to `TransactionPool` trait that returns `(pending_count, queued_count)` tuple
- Updated `txpool_status` RPC handler to use the new count method instead of `all_transactions()`
- Added count methods to internal pool structures:
  - `pending_transactions_count()` in `TxPool`
  - `queued_transactions_count()` in `TxPool`
- Updated iterator return types to `ExactSizeIterator` for efficient counting in `PendingPool` and `ParkedPool`
- Implemented the new method in `NoopTransactionPool` (returns `(0, 0)`)

## Performance Impact
This change significantly improves the performance of the `txpool_status` RPC call by:
- Avoiding allocation of potentially large transaction collections
- Reducing memory pressure when the transaction pool contains many transactions
- Providing O(1) count operations instead of O(n) collection + counting


crates/rpc/rpc/src/txpool.rs
```diff
@@ -84,8 +84,8 @@ where
     /// Handler for `txpool_status`
     async fn txpool_status(&self) -> RpcResult<TxpoolStatus> {
         trace!(target: "rpc::eth", "Serving txpool_status");
-        let all = self.pool.all_transactions();
-        Ok(TxpoolStatus { pending: all.pending.len() as u64, queued: all.queued.len() as u64 })
+        let (pending, queued) = self.pool.pending_and_queue_txn_count();
+        Ok(TxpoolStatus { pending: pending as u64, queued: queued as u64 })
```
mattsse: great

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -434,6 +434,11 @@ impl<T: TransactionOrdering> TxPool<T> {
         self.pending_pool.all()
     }
 
+    /// Returns the number of transactions from the pending sub-pool
+    pub(crate) fn pending_transactions_count(&self) -> usize {
+        self.pending_pool.all().count()
```
mattsse: I've optimized this a bit more so that this returns just .len() because now this does a bunch of Arc clones and this is equivalent to .len()

mattsse (approved):
this makes a lot of sense, I made this a bit more efficient

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -434,6 +434,11 @@ impl<T: TransactionOrdering> TxPool<T> {
         self.pending_pool.all()
     }
 
+    /// Returns the number of transactions from the pending sub-pool
+    pub(crate) fn pending_transactions_count(&self) -> usize {
+        self.pending_pool.len()
+    }
+
```
kien-rise, line 441: @hai-rise We'll need to make this function `pub` so it can be accessed downstream.

Basically we need to add (copy) this code to our `RisePool` implementation.

```rust
    fn pending_and_queued_txn_count(&self) -> (usize, usize) {
        let data = self.pool.get_pool_data();
        let pending = data.pending_transactions_count();
        let queued = data.queued_transactions_count();
        (pending, queued)
    }
```

However, the `data.pending_transactions_count()` and `data.queued_transactions_count()` are private function calls.

----

@keanji-x FYI, we are from RISE and we are having `impl<V, T> TransactionPool for RisePool<V, T>` in our code.

cc @mattsse

PR #17393, opened 2025-07-14
fix(docs): prevent infinite refresh loop in rustdoc search and handle…

yash-atreya:
… missing scripts gracefully

Previously, when the rustdoc search scripts failed to load, the sendSearchForm() function would try to submit a non-existent form, causing continuous page refreshes.

This fix:
- Adds a failsafe counter to prevent more than 2 reload attempts
- Shows "No Results" instead of trying to submit a form when scripts fail
- Dynamically detects the search script filename from HTML files
- Fixes path resolution issues that were causing script loading failures

The search now works correctly both via query parameters (/docs?search=term) and through the search interface, with graceful degradation when scripts cannot be loaded.

### DIFF
--- a/docs/vocs/scripts/inject-cargo-docs.ts
+++ b/docs/vocs/scripts/inject-cargo-docs.ts
@@ -86,6 +86,22 @@ async function injectCargoDocs() {
     await fs.writeFile(file, content, 'utf-8');
   }
 
+  // Find the actual search JS filename from the HTML files
+  let actualSearchJsFile = '';
+  for (const htmlFile of htmlFiles) {
+    const htmlContent = await fs.readFile(htmlFile, 'utf-8');
+    const searchMatch = htmlContent.match(/data-search-js="[^"]*\/([^"]+)"/);
+    if (searchMatch && searchMatch[1]) {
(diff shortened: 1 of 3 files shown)


docs/vocs/scripts/inject-cargo-docs.ts
```diff
@@ -102,9 +141,28 @@ async function injectCargoDocs() {
...
+      content = content.replace(
+        /window\.searchState=\{/g,
+        'window.searchFailureCount=(window.searchFailureCount||0);window.searchState={'
+      );
+      
       content = content.replace(
         /function sendSearchForm\(\)\{document\.getElementsByClassName\("search-form"\)\[0\]\.submit\(\)\}/g,
-        'function sendSearchForm(){/* Fixed: No form submission needed - search loads via script */}'
+        `function sendSearchForm(){
+          /* Fixed: Prevent infinite reload loop */
```
mattsse, line 153: why do we need to do this?

yash-atreya: We don't - closing. 

Testing here first https://github.com/yash-atreya/reth/pull/5

PR #17396, opened 2025-07-14
fix(txpool): Propagate promoted transactions on account updates

keanji-x:
## What's Changed?
This PR fixes an issue where transactions promoted to the pending pool via the update_accounts method were not being broadcast to the network.
The update_accounts function is called to synchronize account states, for instance, when the pool's state has drifted. This can cause parked or queued transactions (e.g., those with a future nonce) to be promoted to the pending pool. However, unlike transactions promoted during a canonical state change (on_canonical_state_change), these promoted transactions were not being announced to the pending_transaction_listener.
This change adds logic to update_accounts to iterate over the promoted transactions and notify the pending_transaction_listener, which is responsible for network propagation. This ensures that these newly pending transactions are correctly broadcast to peers, preventing them from getting "stuck" in the local mempool.
The new notification logic is consistent with existing propagation mechanisms, correctly filtering for transactions that are allowed to be propagated (respecting the propagate flag) to prevent the broadcast of private transactions.

## Why is this change necessary?
Without this fix, transactions promoted due to asynchronous account state updates (not tied to a new block) would become pending locally but would fail to be shared with the rest of the network. This could significantly delay their eventual inclusion in a block. This patch ensures reliable transaction propagation in all scenarios where a transaction enters the pending state.


mattsse (approved):
very nice find, ty!

PR #17403, opened 2025-07-14
fix: optimize iterator chain by removing unnecessary Vec creation

aso20455:
Replace inefficient pattern of collecting an iterator into a Vec just to convert it back to an iterator for reverse traversal. This change directly calls .rev() on the iterator, avoiding temporary memory allocation and improving performance.


Rjected:
I don't think we can optimize like this, because the `BlockState` can only create an iterator by iterating through its parents, which are stored as an `Arc`ed field. So this iterator can only be turned into a double-ended iterator by collecting first.

PR #17405, opened 2025-07-14
feat(trie): add ParallelSparseTrieMetrics

Rjected:
This adds metrics to the parallel sparse trie, right now the metrics struct just contains a histogram noting the number of subtries updated when we call `update_subtries`


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -587,6 +592,9 @@ impl SparseTrieInterface for ParallelSparseTrie {
         let mut prefix_set = core::mem::take(&mut self.prefix_set).freeze();
         let (subtries, unchanged_prefix_set) = self.take_changed_lower_subtries(&mut prefix_set);
 
+        // update metrics
+        self.update_hash_metrics(subtries.len());
```
shekhirin: can we instead feature flag the call and the method, so that we don't have arguments prefixed with `_`?

Rjected: sure, that makes sense

mediocregopher:
Another that just came to mind: would be nice to know how often we're hitting the cases in `update_leaf` and `remove_leaf` where we have to reveal a blinded node... I have a theory the answer for `remove_leaf` is "never", but it'd be nice to have some certainty on that.

Rjected:
> Another that just came to mind: would be nice to know how often we're hitting the cases in update_leaf and remove_leaf where we have to reveal a blinded node... I have a theory the answer for remove_leaf is "never", but it'd be nice to have some certainty on that.

I think I'll add some metrics on how many nodes we're fetching to the proof task manager for now, it's not splitting by update_leaf / remove_leaf, but it's easier to implement properly

crates/trie/sparse-parallel/src/metrics.rs
```diff
@@ -0,0 +1,23 @@
...
+pub(crate) struct ParallelSparseTrieMetrics {
+    /// A histogram for the number of subtries updated when calculating hashes.
+    pub(crate) subtries_updated: Histogram,
+    /// A histogram for the time it took to update lower subtrie hashes.
+    pub(crate) subtrie_hash_update_latency: Histogram,
+    /// A histogram for the time it took to update the upper subtrie hashes.
+    pub(crate) subtrie_upper_hash_latency: Histogram,
+}
+
+impl PartialEq for ParallelSparseTrieMetrics {
```
mattsse, line 16: ah this is because we derive this for ParallelSparseTrie, do we even this this derive there?
since this is pub crate this is fine tho

mattsse (approved):
lgtm!

crates/trie/sparse-parallel/src/metrics.rs
Rjected: > do we even this this derive there?

its mainly for tests
https://github.com/paradigmxyz/reth/blob/3772535220784420370fff9c13f85d17d15eb786/crates/trie/sparse-parallel/src/trie.rs#L4623

Issue #17293, opened 2025-07-08
Do not calculate the transactions root in `insert_block_inner`


shekhirin:
### Describe the feature

Profile: https://share.firefox.dev/4lLWTJq

<img width="1271" height="773" alt="Image" src="https://github.com/user-attachments/assets/a91b91cb-d914-4f80-a1ac-cbd4f9c2db42" />

We don't need to calculate it again in `insert_block_inner`, because it's already calculated in `try_into_block_with_sidecar`.

mattsse:
completed with payload refactor

PR #17406, opened 2025-07-14
chore(consensus): refactor fork and ommers check into standalone fn

Rjected:
ref https://github.com/paradigmxyz/reth/issues/17293

If want to avoid calculating the transactions root twice, we at least need a fn that contains all checks in `validate_block_pre_execution`, without the transactions root check.

Right now this is called `validate_block_fork_fields`, but would like opinions on a rename because that name is vague

### DIFF
--- a/crates/consensus/common/src/validation.rs
+++ b/crates/consensus/common/src/validation.rs
@@ -140,6 +140,31 @@ where
     B: Block,
     ChainSpec: EthereumHardforks,
 {
+    post_merge_hardfork_fields(block, chain_spec)?;
+
+    // Check transaction root
+    if let Err(error) = block.ensure_transaction_root_valid() {
+        return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
+    }
+
+    Ok(())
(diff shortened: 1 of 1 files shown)


crates/consensus/common/src/validation.rs
```diff
@@ -136,6 +136,32 @@ pub fn validate_block_pre_execution<B, ChainSpec>(
...
+
+    // Check transaction root
+    if let Err(error) = block.ensure_transaction_root_valid() {
+        return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
+    }
+
+    Ok(())
+}
+
+// TODO: rename this fn
```
mattsse: maybe post_merge_hardfork_fields?

mattsse (approved):

shekhirin (approved):

crates/consensus/common/src/validation.rs
Rjected: ah yeah that name is better

PR #17407, opened 2025-07-15
refactor(consensus): add Consensus methods for fork fields and tx root

Rjected:
ref https://github.com/paradigmxyz/reth/pull/17406 and https://github.com/paradigmxyz/reth/issues/17293

This would allow us to omit the transactions root check in the engine, since we do that on payload conversion.

Need naming advice for the pre-execution validation method that does not have tx root checks.

### DIFF
--- a/crates/ethereum/consensus/src/lib.rs
+++ b/crates/ethereum/consensus/src/lib.rs
@@ -21,6 +21,7 @@ use reth_consensus_common::validation::{
     validate_against_parent_eip1559_base_fee, validate_against_parent_hash_number,
     validate_against_parent_timestamp, validate_block_pre_execution, validate_body_against_header,
     validate_header_base_fee, validate_header_extra_data, validate_header_gas,
+    validate_post_merge_fork_fields,
 };
 use reth_execution_types::BlockExecutionResult;
 use reth_primitives_traits::{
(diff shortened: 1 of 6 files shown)


crates/ethereum/consensus/src/lib.rs
```diff
@@ -128,6 +129,19 @@ where
     fn validate_block_pre_execution(&self, block: &SealedBlock<B>) -> Result<(), Self::Error> {
         validate_block_pre_execution(block, &self.chain_spec)
     }
+
+    fn validate_transactions(&self, block: &SealedBlock<B>) -> Result<(), Self::Error> {
+        // Check transaction root
+        if let Err(error) = block.ensure_transaction_root_valid() {
+            return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
+        }
```
mattsse, line 137: should we make this the default impl?

mattsse (approved):
this makes sense, one suggestion re txroot default

pending @klkvr

klkvr (commented):
isn't the current plan for perf improvements to refactor engine so that we start executing transactions right away without blocking on any validations? meaning that it's fine to keep the logic as is as long as it's done in parallel

`Consensus` traits are already a bit confusing so if we don't have a strong need in such changes I'd prefer holding off

Rjected:
closing this, assuming we will do this in the payload validator

PR #17409, opened 2025-07-15
feat: add generic database support for Receipt<T>

rezzmah:
Enable Receipt<T> to work with custom transaction types by updating the database compression macro to handle the generic type parameter.

This fixes the extensibility gap from PR #17237 where downstream projects couldn't use Receipt<CustomTxType> due to the macro only supporting concrete Receipt types.

**Change**: Updated impl_compression_for_compact\! macro to use `Receipt<T>` instead of Receipt, enabling automatic database support for any transaction type that implements the required traits.


mattsse (approved):
that should do it

PR #17410, opened 2025-07-15
fix(`docs`): rustdoc search functionality 

yash-atreya:
- Locate the relevant search-*.js file which powers the search functionality


docs/vocs/scripts/inject-cargo-docs.ts
```diff
@@ -120,9 +135,10 @@ async function injectCargoDocs() {
       );
       
       // Fix the search-js variable to return just the filename
+      // Use the detected search filename
       content = content.replace(
         /getVar\("search-js"\)/g,
-        `"search-f7877310.js"`
+        `"${actualSearchJsFile}"`
```
yash-atreya, line 142: The `search-js` file was hardcoded previously which would not work across new builds. 

We're now trying to fetch the `search-*` file on `bun run build`. The build / ci job will fail if we're not able to locate it.

mattsse (approved):
ty

Issue #17413, opened 2025-07-15
Make payload validation in engine customizable


mattsse:
### Describe the feature

for perf reasons #17291 and flexibility reasons we want to make payload/block validation(execution) entirely configurable in

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L228-L228

this affects the

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L522-L525

function and

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L2033-L2034

because both currently call:

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L2052-L2052

ideally we make it configurable how the payload/block is validated, for payload this currently happens here where we first convert the payload to a block:

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L555-L555

but we repurpose this trait

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/primitives/src/lib.rs#L109-L111

so that it can validate blocks and payloads, so instead of

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/primitives/src/lib.rs#L124-L127

we can replace this with  `validate_payload(Payload, ct) -> Result<Outcome>`, then we also need `validate_block(block, ctx)` because when we download the block from the network we already have the block itself, we also need to provide some context from the tree itself.
This could be `struct TreeCtx<'a> {}` which the `EngineApiTreeHandler` would initialize before calling the function, the pupose of the TreeCtx would be to provide access to obtaining the state like:

https://github.com/paradigmxyz/reth/blob/253721d22604850180a1af07f7f3e7657dea25c8/crates/engine/tree/src/tree/mod.rs#L2128-L2128

then the flow could look like from the `EngineApiTreeHandler`

```rust
fn on_newpayload(payload) {
  // sanity checks for whether we're currently syncing
[...]

PR #17429, opened 2025-07-15
feat(engine): add validate_payload and validate_block methods to EngineValidator trait

fgimenez:
towards #17413 

Adds new validation methods to the `EngineValidator` trait as a first step towards repurposing it for more flexible payload and block validation.

- add `validate_payload` and `validate_block` methods with default implementations to maintain backward compatibility

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -16,8 +16,9 @@ use reth_consensus::{ConsensusError, FullConsensus};
 use reth_engine_primitives::{InvalidBlockHook, PayloadValidator};
 use reth_evm::{ConfigureEvm, SpecFor};
 use reth_payload_primitives::{
-    EngineApiMessageVersion, EngineObjectValidationError, InvalidPayloadAttributesError,
-    NewPayloadError, PayloadAttributes, PayloadOrAttributes, PayloadTypes,
+    BuiltPayload, EngineApiMessageVersion, EngineObjectValidationError,
+    InvalidPayloadAttributesError, NewPayloadError, PayloadAttributes, PayloadOrAttributes,
+    PayloadTypes,
 };
 use reth_primitives_traits::{
     AlloyBlockHeader, Block, BlockBody, GotExpected, NodePrimitives, RecoveredBlock, SealedHeader,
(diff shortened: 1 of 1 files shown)


crates/engine/primitives/src/lib.rs
```diff
@@ -42,6 +42,27 @@ pub use invalid_block_hook::InvalidBlockHook;
...
+
+    /// Returns the current canonical block number  
+    fn canonical_block_number(&self) -> u64;
+}
+
+/// Context providing access to tree state during validation
+pub struct TreeCtx<'a> {
+    /// The engine API tree state
+    pub state: &'a dyn TreeStateProvider,
+}
```
mattsse: we likely need a generic for N as well because I think we need to fetch parent block for example

crates/engine/primitives/src/lib.rs
```diff
@@ -126,6 +147,24 @@ pub trait PayloadValidator: Send + Sync + Unpin + 'static {
         payload: Self::ExecutionData,
     ) -> Result<RecoveredBlock<Self::Block>, NewPayloadError>;
 
+    /// Validates a payload received from engine API.
+    fn validate_payload(
+        &self,
+        _payload: Self::ExecutionData,
+        _ctx: TreeCtx<'_>,
+    ) -> Result<(), NewPayloadError> {
```
mattsse: for the result we likely need a new outcome struct that includes the block and the validation status, e.g. we still want to return an invalid block in full

fgimenez: makes sense, done here https://github.com/paradigmxyz/reth/pull/17429/commits/48b5476bde5d9c5c64c7b074b69f96fc0eed3ae6

crates/engine/primitives/src/lib.rs
fgimenez: makes sense, added a generic `State` which later we can specify in impls as `EngineApiTreeState<N: NodePrimitives>`

crates/engine/primitives/src/lib.rs
```diff
@@ -106,7 +130,7 @@ pub trait EngineTypes:
 
 /// Type that validates an [`ExecutionPayload`].
 #[auto_impl::auto_impl(&, Arc)]
-pub trait PayloadValidator: Send + Sync + Unpin + 'static {
+pub trait PayloadValidator<State = ()>: Send + Sync + Unpin + 'static {
```
mattsse: we likely don't need this generic if we make `TreeCtx<'_, Self::Block>`

fgimenez: i believe it would be easier if `TreeCtx` can hold a full state instance, so that we can call the new methods like `validate_payload(payload, TreeCtx { state: &self.state})` if we make `TreeCtx` generic over block this would be more difficult, or am i missing something?

mattsse: let#s figure this out later

fgimenez: the trait is now generic over `N: NodePrimitives`

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -973,8 +973,10 @@ where
 }
 
 /// Type that validates the payloads processed by the engine.
-pub trait EngineValidator<Types: PayloadTypes>:
-    PayloadValidator<ExecutionData = Types::ExecutionData>
+pub trait EngineValidator<
+    Types: PayloadTypes,
+    N: NodePrimitives = reth_ethereum_primitives::EthPrimitives,
+>: PayloadValidator<ExecutionData = Types::ExecutionData>
```
klkvr: `NodePrimitives` are already available as `Types::BuiltPayload::Primitives`

fgimenez: awesome thx! done ptal

mattsse (approved):

klkvr (approved):

PR #17431, opened 2025-07-15
feat(tx-pool): add submit methods to `TransactionPool`

### LINKED ISSUE #17421: Add helper submit function to txpool
### Describe the feature

we currently need to do this

```
    // Recover the transaction
    let transaction = transaction.try_into_recovered()?;

    // Convert to pool transaction type
    let pool_transaction =
        <FC::Pool as TransactionPool>::Transaction::try_from_consensus(transaction)
            .map_err(|e| eyre::eyre!("Failed to convert to pool transaction: {e}"))?;

    // Submit the transaction to the pool and get event stream
    let tx = node
        .pool()
        .add_transaction(TransactionOrigin::Local, pool_transaction)
        .await?;
```

but we can easily do this ins something like

```rust
fn `add_consensus_transaction(tx: Recovered<Self::Transaction::Consensus>, origin)`
```

this would then do the conversion `<FC::Pool as TransactionPool>::Transaction::try_from_consensus(transaction)` internally 

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17421.

### DIFF
--- a/crates/transaction-pool/src/traits.rs
+++ b/crates/transaction-pool/src/traits.rs
@@ -1614,7 +1614,7 @@ impl<Tx: PoolTransaction> NewSubpoolTransactionStream<Tx> {
             match self.st.try_recv() {
                 Ok(event) => {
                     if event.subpool == self.subpool {
-                        return Ok(event);
+                        return Ok(event)
                     }
                 }
                 Err(e) => return Err(e),
@@ -1631,7 +1631,7 @@ impl<Tx: PoolTransaction> Stream for NewSubpoolTransactionStream<Tx> {
             match ready!(self.st.poll_recv(cx)) {
                 Some(event) => {
                     if event.subpool == self.subpool {
-                        return Poll::Ready(Some(event));
+                        return Poll::Ready(Some(event))
                     }
                 }
                 None => return Poll::Ready(None),
--- a/crates/transaction-pool/src/traits.rs
+++ b/crates/transaction-pool/src/traits.rs
@@ -52,7 +52,7 @@
 
 use crate::{
     blobstore::BlobStoreError,
-    error::{InvalidPoolTransactionError, PoolResult},
+    error::{InvalidPoolTransactionError, PoolError, PoolResult},
     pool::{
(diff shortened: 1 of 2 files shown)


examples/txpool-tracing/src/submit.rs
```diff
@@ -115,19 +108,35 @@ where
...
     let transaction: TransactionSigned =
         NetworkWallet::<Ethereum>::sign_request(wallet, request).await?.into();
-    // Recover the transaction
-    let transaction = transaction.try_into_recovered()?;
 
+    // Recover the transaction and prepare the pool
+    let (tx_hash, pool_transaction) = prepare_pool_transaction::<FC>(transaction)?;
+
+    // Submit the transaction to the pool
+    node.pool().add_transaction(TransactionOrigin::Local, pool_transaction).await?;
```
lean-apple: is another helper needed to include also this submission?

crates/transaction-pool/src/traits.rs
```diff
@@ -176,6 +176,42 @@ pub trait TransactionPool: Clone + Debug + Send + Sync {
...
 
+    /// Submit a consensus transaction directly to the pool
+    fn add_consensus_transaction(
+        &self,
+        tx: Recovered<<Self::Transaction as PoolTransaction>::Consensus>,
+        origin: TransactionOrigin,
+    ) -> impl Future<Output = PoolResult<TxHash>> + Send {
+        async move {
+            let tx_hash = *tx.tx_hash();
+
```
mattsse, line 187: this is a bit unfortunate,

what we could do is change the error type and require that this at least the returns the tx hash, e.g.:

https://github.com/paradigmxyz/reth/blob/1e208710432fcb0d3a4b367a3c3977661305a1d1/crates/transaction-pool/src/traits.rs#L1248-L1249

but we can do this in a separate step

mattsse (approved):
this is great UX improvement, we can make this a bit nicer even by changing the error, but the hash copy isnt a big deal

PR #17434, opened 2025-07-16
refactor(txpool): restructure transaction storage to group by sender

keanji-x:
## What does this PR do?

This pull request refactors the core data structure for storing transactions in the `AllTransactions` pool.

It changes `txs` from a single, large `BTreeMap<TransactionId, PoolInternalTransaction>` to a nested `FxHashMap<SenderId, BTreeMap<TransactionId, PoolInternalTransaction>>`.

This groups all transactions by their sender, which aligns the data structure more closely with the pool's primary access patterns (e.g., nonce checking, per-sender transaction limits).

## Why is this change necessary?

The previous structure, a single B-Tree map for all transactions, resulted in performance degradation as the number of transactions in the pool grew. Key operations like lookups, insertions, and per-sender iterations had a time complexity of $O(\log N)$, where $N$ is the total number of transactions in the pool.

By grouping transactions by sender, the complexity of these common operations is reduced to approximately $O(\log M)$, where $M$ is the number of transactions for a single sender ($M \ll N$). This provides a significant performance improvement, especially in a busy pool.

### DIFF
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -173,11 +173,11 @@ impl<T: TransactionOrdering> TxPool<T> {
         }
 
         let mut next_expected_nonce = on_chain.nonce;
-        for (id, tx) in self.all().descendant_txs_inclusive(&on_chain) {
-            if next_expected_nonce != id.nonce {
+        for (nonce, tx) in self.all().descendant_txs_inclusive(&on_chain) {
+            if next_expected_nonce != *nonce {
                 break
             }
-            next_expected_nonce = id.next_nonce();
+            next_expected_nonce = nonce + 1;
             last_consecutive_tx = Some(tx);
         }
 
@@ -1023,10 +1023,11 @@ impl<T: TransactionOrdering> TxPool<T> {
             let descendant =
                 self.all_transactions.descendant_txs_exclusive(&id).map(|(id, _)| *id).next();
             if let Some(descendant) = descendant {
-                if let Some(tx) = self.remove_transaction(&descendant) {
+                let descendant_id = TransactionId::new(id.sender, descendant);
+                if let Some(tx) = self.remove_transaction(&descendant_id) {
                     removed.push(tx)
                 }
-                id = descendant;
+                id = descendant_id;
             } else {
                 return
             }
@@ -1228,7 +1229,7 @@ pub(crate) struct AllTransactions<T: PoolTransaction> {
     /// _All_ transactions identified by their hash.
     by_hash: HashMap<TxHash, Arc<ValidPoolTransaction<T>>>,
     /// _All_ transaction in the pool sorted by their sender and nonce pair.
-    txs: FxHashMap<SenderId, BTreeMap<TransactionId, PoolInternalTransaction<T>>>,
+    txs: FxHashMap<SenderId, BTreeMap<u64, PoolInternalTransaction<T>>>,
     /// Tracks the number of transactions by sender that are currently in the pool.
     tx_counter: FxHashMap<SenderId, usize>,
     /// The current block number the pool keeps track of.
@@ -1280,14 +1281,14 @@ impl<T: PoolTransaction> AllTransactions<T> {
 
     /// Returns the internal transaction with additional metadata
     pub(crate) fn get(&self, id: &TransactionId) -> Option<&PoolInternalTransaction<T>> {
-        self.txs.get(&id.sender).and_then(|sender_txs| sender_txs.get(id))
+        self.txs.get(&id.sender).and_then(|sender_txs| sender_txs.get(&id.nonce))
     }
 
     pub(crate) fn get_mut(
         &mut self,
         id: &TransactionId,
     ) -> Option<&mut PoolInternalTransaction<T>> {
-        self.txs.get_mut(&id.sender).and_then(|sender_txs| sender_txs.get_mut(id))
+        self.txs.get_mut(&id.sender).and_then(|sender_txs| sender_txs.get_mut(&id.nonce))
     }
 
     /// Increments the transaction counter for the sender
@@ -1363,8 +1364,6 @@ impl<T: PoolTransaction> AllTransactions<T> {
         // pre-allocate a few updates
         let mut updates = Vec::with_capacity(64);
 
-        let mut iter = self.txs.iter_mut().flat_map(|(_, txs)| txs.iter_mut()).peekable();
-
         // Loop over all individual senders and update all affected transactions.
         // One sender may have up to `max_account_slots` transactions here, which means, worst case
         // `max_accounts_slots` need to be updated, for example if the first transaction is blocked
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1578,7 +1608,16 @@ impl<T: PoolTransaction> AllTransactions<T> {
...
+        let internal =
+            if let hash_map::Entry::Occupied(mut sender_entry) = self.txs.entry(tx.sender_id()) {
+                let internal_opt = sender_entry.get_mut().remove(&tx.transaction_id);
+                if sender_entry.get().is_empty() {
+                    sender_entry.remove();
+                }
+                internal_opt
+            } else {
+                None
+            }?;
```
mattsse: we now have this a few times, could we perhaps extract this to a single fn?

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1228,7 +1249,7 @@ pub(crate) struct AllTransactions<T: PoolTransaction> {
     /// _All_ transactions identified by their hash.
     by_hash: HashMap<TxHash, Arc<ValidPoolTransaction<T>>>,
     /// _All_ transaction in the pool sorted by their sender and nonce pair.
-    txs: BTreeMap<TransactionId, PoolInternalTransaction<T>>,
+    txs: FxHashMap<SenderId, BTreeMap<TransactionId, PoolInternalTransaction<T>>>,
```
mattsse: I think this makes sense

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -234,8 +239,12 @@ impl<T: TransactionOrdering> TxPool<T> {
...
                 for tx in removed {
                     let to = {
-                        let tx =
-                            self.all_transactions.txs.get_mut(tx.id()).expect("tx exists in set");
+                        let tx = self
+                            .all_transactions
+                            .txs
+                            .get_mut(&tx.id().sender)
+                            .and_then(|sender_txs| sender_txs.get_mut(tx.id()))
+                            .expect("tx exists in set");
```
mattsse: this we also have a few times now, could we make this a new helper of alltransactions?

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -191,7 +191,12 @@ impl<T: TransactionOrdering> TxPool<T> {
 
     /// Returns all senders in the pool
     pub(crate) fn unique_senders(&self) -> HashSet<Address> {
-        self.all_transactions.txs.values().map(|tx| tx.transaction.sender()).collect()
+        self.all_transactions
+            .txs
+            .values()
+            .flat_map(|tx| tx.values())
+            .map(|tx| tx.transaction.sender())
```
mattsse: I think this now collects all tx senders but we only need to do this per value, so this would collect a bunch of redundant addresses, because we only need `tx.values().first()`

mattsse (changes requested):
I think this makes sense, I'd like to see some numbers for this

perhaps @hai-rise could help here, also curios what @hai-rise thinks

crates/transaction-pool/src/pool/txpool.rs
keanji-x: Good point! Done! 👍

crates/transaction-pool/src/pool/txpool.rs
hai-rise: Oh, this is interesting and makes sense. Maybe we can help benchmark with our heavy pool scenarios.

Also, at a glance, we can greedily demote the inner `TransactionId` to a single `u64` for nonce now?

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1523,7 +1530,10 @@ impl<T: PoolTransaction> AllTransactions<T> {
         sender: SenderId,
     ) -> impl Iterator<Item = (&TransactionId, &PoolInternalTransaction<T>)> + '_ {
         self.txs
-            .range((sender.start_bound(), Unbounded))
+            .get(&sender)
+            .map(|sender_txs| sender_txs.iter())
+            .into_iter()
+            .flatten()
             .take_while(move |(other, _)| sender == other.sender)
```
hai-rise: We don't need these anymore? Demoting the inner `TransactionId` to a `u64` nonce may help yield compile errors for all places we should double-check :).

crates/transaction-pool/src/pool/txpool.rs
mattsse: > TransactionId to a single u64 for nonce now?

this makes sense, this would make ordering even more performant I believe @keanji-x

keanji-x: Makes sense, will update this shortly

keanji-x: done

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1356,131 +1364,120 @@ impl<T: PoolTransaction> AllTransactions<T> {
...
-                next_sender!(iter);
-                continue 'transactions
             }
+            // Get the first transaction for this sender
+            if let Some((nonce, tx)) = tx_iter.next() {
+                let id = TransactionId::new(*sender_id, *nonce);
+
+                // track the balance if the sender was changed in the block
+                // check if this is a changed account
+                let changed_balance = if let Some(info) = changed_accounts.get(sender_id) {
```
mattsse: this lookup seems redundant if we're now guaranteed to be in the same sender context in this loop, I guess this was even redundant before.

we can either change it in this pr or in a followup

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1356,131 +1364,120 @@ impl<T: PoolTransaction> AllTransactions<T> {
...
-                    }
-                };
-            }
-
-            // track the balance if the sender was changed in the block
-            // check if this is a changed account
-            let changed_balance = if let Some(info) = changed_accounts.get(&id.sender) {
+        for (sender_id, sender_txs) in &mut self.txs {
+            let mut tx_iter = sender_txs.iter_mut().peekable();
+            if let Some(info) = changed_accounts.get(sender_id) {
```
mattsse: these are significant changes, and I'm not fully convinced yet that this is entirely correct

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1356,131 +1364,120 @@ impl<T: PoolTransaction> AllTransactions<T> {
...
-                    next_sender!(iter);
-                    continue 'transactions
-                }
+                // the next expected nonce after this transaction: nonce + 1
+                let mut next_nonce_in_line = tx.transaction.nonce().saturating_add(1);
 
-                // update for next iteration of this sender's loop
-                next_nonce_in_line = next_nonce_in_line.saturating_add(1);
+                // Update all consecutive transaction of this sender
+                for (peek_nonce, tx) in tx_iter {
```
mattsse: this peek_ is now confusing because this doesnt peek

hmm, this entire section feels off, because this now consume the iter but we need to peek into consecutive nonce txs to update the balances accordingly (because we need the context of previous txs

mattsse (changes requested):
still need to review the loop logic changes a bit more but I think the lookahead logic isnt entirely accurate

crates/transaction-pool/src/pool/txpool.rs
mattsse: I'm realizing that we could also improve this if we create our own peek_back iterator but we could tackle this separately.

keanji-x: Yeah,  Agreed! We can definitely create our own peek_back iterator in this PR. However, the original approach isn't efficient since it needs to iterate through all transactions. Let's keep this PR focused and simple for now, and we can address the efficiency optimization in a follow-up PR.

I‘ll fixed

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1356,7 +1364,13 @@ impl<T: PoolTransaction> AllTransactions<T> {
...
         let mut updates = Vec::with_capacity(64);
 
-        let mut iter = self.txs.iter_mut().peekable();
+        let mut iter = self
+            .txs
+            .iter_mut()
+            .flat_map(|(sender, txs)| {
+                txs.iter_mut().map(move |(nonce, tx)| (TransactionId::new(*sender, *nonce), tx))
+            })
+            .peekable();
```
keanji-x, line 1373: The current PR involves modifications to many lines and changes a fundamental data structure, making it highly prone to conflicts. To ensure better reviewability, we should minimize logical changes in this PR. Therefore, I've created an iterator that's compatible with the existing code, which can be optimized away in the next PR. 

For better performance, the follow-up PR can focus on the update function and add more tests to ensure accuracy.

hai-rise:
> cool, @hai-rise any chance you could run this and check some numbers? 🙏

cc @kien-rise who are interested in running the numbers!

kien-rise:
@mattsse Hi, I just started my benchmark script. It will finish after about 2 hours.
Then, I will need another 2 hours to benchmark the "before" state.
The data I collect is the run duration of `on_canonical_state_change`.
Then, I will make a chart to compare the "before" and the "after".

------

Overall, this refactoring PR looks good to me. Using a hash map is definitely a more natural way than `BTreeMap` to represent this particular data.

keanji-x:
### Test Environment

- **Benchmark:** #17688  
- **CPU:** AMD EPYC 7K62 48-Core Processor  
- **Memory:** 94.2 GB  

---

### Benchmark Results

| Senders / Txs | BTreeMap (µs) | HashMap (µs) | Change (%) | Conclusion              |
|---------------|---------------|--------------|------------|-------------------------|
| 500 / 1       | 26.213        | 27.448       | +4.06%     | No significant change   |
| 500 / 5       | 127.49        | 115.91       | -9.10%     | Improved ✅              |
| 500 / 10      | 213.22        | 186.71       | -12.40%    | Improved ✅              |
| 1000 / 1      | 34.390        | 36.153       | +4.64%     | Regressed 🔻            |
| 1000 / 5      | 161.27        | 149.98       | -7.05%     | Improved ✅              |
| 1000 / 10     | 268.20        | 238.25       | -11.73%    | Improved ✅              |
| 2000 / 1      | 46.464        | 49.909       | +7.38%     | Regressed 🔻            |
| 2000 / 5      | 219.29        | 215.82       | -0.83%     | No significant change   |
| 2000 / 10     | 508.57        | 489.14       | -3.77%     | No significant change   |

---

### Analysis

The results indicate that for single-transaction cases (`txs = 1`), **BTreeMap consistently outperforms HashMap**, with performance regressions of **4–7%**. However, as the number of transactions per sender increases, **HashMap demonstrates clear advantages**, especially with 5–10 transactions per sender, yielding **7–12% performance gains**.

When the transaction-to-account ratio is **1:1**, the nested HashMap structure performs worse than the flattened BTreeMap. This suggests:

1. Iterating over a BTreeMap<(SenderId, nonce), Txn> is faster than traversing a HashMap<SenderId, BTreeMap<nonce, Txn>> in the `update` function, especially when the inner BTreeMaps are very small.
2. When each account has only one transaction, removing a transaction also deletes the corresponding BTreeMap, resulting in additional overhead.

PR #17436, opened 2025-07-16
feat: make revm_spec generic over header type

rezzmah:
## Summary

Makes `revm_spec` function generic over any type implementing `BlockHeader` instead of being hardcoded to `alloy_consensus::Header`. This enables custom chain implementations using the Ethereum EVM (like bera-reth) to use their own header types while maintaining full compatibility with the EVM configuration.

## Changes

- Updated `revm_spec` function to accept generic `H: BlockHeader` parameter
- Changed header field access from direct field access to trait methods (`header.timestamp()`, `header.number()`)
- Updated test cases to explicitly use `Header::default()` instead of `Default::default()`

## Use Case

This change primarily benefits chains using the Ethereum EVM that need custom header types with additional consensus fields while still being able to determine the correct EVM spec for execution.

## Backward Compatibility

The change maintains full backward compatibility as `alloy_consensus::Header` implements `BlockHeader`, so existing code continues to work unchanged.

### DIFF
--- a/crates/ethereum/evm/src/config.rs
+++ b/crates/ethereum/evm/src/config.rs
@@ -100,6 +100,7 @@ where
 mod tests {
     use super::*;
     use crate::U256;
+    use alloy_consensus::Header;
     use reth_chainspec::{ChainSpecBuilder, MAINNET};
 
     #[test]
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this seems fine

PR #17439, opened 2025-07-16
chore: simplify blob count extraction using new blob_count() method

cakevm:
Make use of the new `blob_count()`, improve readability and remove unwrap.


mattsse (approved):
this is identical yes

PR #17440, opened 2025-07-16
perf(trie): Process multiproof reveals for storage tries in parallel

### LINKED ISSUE #17430: Reveal decoded storage multiproofs in parallel
### Describe the feature

We currently reveal storage multiproofs serially, we should be able to do this in parallel for each storage trie
https://github.com/paradigmxyz/reth/blob/73f2edb90c979afa0ea6e1fe93577b8388d1eced/crates/trie/sparse/src/state.rs#L285-L288

### Additional context

_No response_

mediocregopher:
Make `SparseStateTrie::reveal_decoded_multiproof` parallel with respect to storage root revealing. The account multiproof revealing is already parallelized by ParallelSparseTrie.

Implementing this change required splitting the `storages` and `revealed_storage_paths` fields of `SparseStateTrie` into their own inner struct, so that they could be borrowed at the same time via a `get_or_create` style method. As part of this I also introduced re-use of the `revealed_storage_paths` HashSets across payloads.

## Benchmarks (mainnet, 1000 blocks)

```
Timestamp: 2025-07-21 17:15:05 UTC
Baseline: main
Feature:  origin/mediocregopher/17430-parallel-storage-reveal

Performance Changes:
  NewPayload Latency: -5.98%
  FCU Latency:        +0.67%
  Total Latency:      -5.76%
  Gas/Second:         +6.11%
  Blocks/Second:      +6.11%

Baseline Summary:
  Blocks: 1000, Gas: 17906209584, Duration: 37.58s
  Avg NewPayload: 36.32ms, Avg FCU: 1.24ms, Avg Total: 37.56ms
  Started: 2025-07-21 17:26:14 UTC, Ended: 2025-07-21 17:30:48 UTC

Feature Summary:
  Blocks: 1000, Gas: 17906209584, Duration: 35.42s
  Avg NewPayload: 34.14ms, Avg FCU: 1.25ms, Avg Total: 35.40ms
  Started: 2025-07-21 17:32:08 UTC, Ended: 2025-07-21 17:36:44 UTC
```

latency_comparison

Fixes #17430

### DIFF
--- a/crates/trie/sparse/src/state.rs
+++ b/crates/trie/sparse/src/state.rs
@@ -60,7 +60,7 @@ pub struct SparseStateTrie<
     /// Collection of revealed account trie paths.
     revealed_account_paths: HashSet<Nibbles>,
     /// State related to storage tries.
-    storage: SparseStoragesStateTrie<S>,
+    storage: StorageTries<S>,
     /// Flag indicating whether trie updates should be retained.
     retain_updates: bool,
     /// Reusable buffer for RLP encoding of trie accounts.
@@ -235,20 +235,6 @@ where
         {
             use rayon::iter::{ParallelBridge, ParallelIterator};
 
-            struct ParallelRevealInput<S> {
-                account: B256,
-                storage_subtree: DecodedStorageMultiProof,
-                revealed_nodes: HashSet<Nibbles>,
-                trie: SparseTrie<S>,
-            }
-
-            struct ParallelRevealOutput<S> {
-                account: B256,
-                revealed_nodes: HashSet<Nibbles>,
-                trie: SparseTrie<S>,
-                result: SparseStateTrieResult<ProofNodesMetricValues>,
-            }
-
             let (tx, rx) = std::sync::mpsc::channel();
             let retain_updates = self.retain_updates;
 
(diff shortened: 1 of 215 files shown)


crates/trie/sparse/src/state.rs
```diff
@@ -58,14 +57,10 @@ pub struct SparseStateTrie<
...
-    /// Sparse storage tries.
-    storages: B256Map<SparseTrie<S>>,
-    /// Cleared storage tries, kept for re-use
-    cleared_storages: Vec<SparseTrie<S>>,
     /// Collection of revealed account trie paths.
     revealed_account_paths: HashSet<Nibbles>,
-    /// Collection of revealed storage trie paths, per account.
-    revealed_storage_paths: B256Map<HashSet<Nibbles>>,
+    /// State related to storage tries.
+    storage: SparseStoragesStateTrie<S>,
```
Rjected: I would maybe call this `StorageTries`, wdyt?

crates/trie/sparse/src/state.rs
```diff
@@ -236,12 +220,92 @@ where
...
+                revealed_nodes: HashSet<Nibbles>,
+                trie: SparseTrie<S>,
+            }
+
+            struct ParallelRevealOutput<S> {
+                account: B256,
+                revealed_nodes: HashSet<Nibbles>,
+                trie: SparseTrie<S>,
+                result: SparseStateTrieResult<ProofNodesMetricValues>,
+            }
```
Rjected: why not just use tuples inside the `map`?

Rjected (approved):
just have nits, this makes sense to me. I think `SparseStoragesStateTrie` is a bit verbose, but could only come up with `StorageTries`, which might be confusing as well

crates/trie/sparse/src/state.rs
mediocregopher: Fine by me :+1:

crates/trie/sparse/src/state.rs
mediocregopher: I thought it would make it easier to follow if we could see explicitly what fields were what type, once you get into 4-tuples it's kinda messy imo :P I don't feel strongly about it though, changed it to tuples

PR #17441, opened 2025-07-16
feat: enable CLI support for custom block headers

rezzmah:
## Summary

Enable CLI commands to work with custom block header types by requiring `From<alloy_consensus::Header>` conversion. This allows custom chain implementations to use extended headers while maintaining full CLI compatibility.

## Changes

- Make `init_state` command generic over `BlockHeader` types with `From<Header>` constraint
- Update header factory to convert from standard headers to custom types
- Fix header field access to use trait methods instead of direct field access
- Add proper type bounds to CLI interface for custom header support

**Note**: This is not an optimal solution. Ideally, the CLI should work with custom headers without requiring conversion from standard Ethereum headers. This constraint is a pragmatic workaround to enable custom header support while maintaining CLI compatibility.


crates/cli/commands/src/init_state/mod.rs
```diff
@@ -103,7 +105,10 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> InitStateC
                     &provider_rw,
                     SealedHeader::new(header, header_hash),
                     total_difficulty,
-                    |number| Header { number, ..Default::default() },
+                    |number| {
+                        let header = Header { number, ..Default::default() };
+                        <<N::Primitives as NodePrimitives>::BlockHeader>::from(header)
```
rezzmah, line 110: We could add `set_number` or smthing in Blockheader trait but thats kinda jank.

crates/cli/commands/src/init_state/without_evm.rs
```diff
@@ -12,14 +12,16 @@ use reth_stages::{StageCheckpoint, StageId};
 use reth_static_file_types::StaticFileSegment;
 use std::{fs::File, io::Read, path::PathBuf};
 use tracing::info;
-
 /// Reads the header RLP from a file and returns the Header.
-pub(crate) fn read_header_from_file(path: PathBuf) -> Result<Header, eyre::Error> {
+pub(crate) fn read_header_from_file<H>(path: PathBuf) -> Result<H, eyre::Error>
+where
+    H: BlockHeader + Decodable,
```
mattsse: I think we dont even need the header bound here
```suggestion
    H: Decodable,
```

crates/cli/commands/src/init_state/mod.rs
```diff
@@ -72,7 +72,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> InitStateC
     where
         N: CliNodeTypes<
             ChainSpec = C::ChainSpec,
-            Primitives: NodePrimitives<BlockHeader = alloy_consensus::Header>,
+            Primitives: NodePrimitives<BlockHeader: BlockHeader + From<Header>>,
```
mattsse, line 75: since this command is mainly used to fill the db with fake headers, I think this seems fine

mattsse (commented):
not opposed to this, wdyt @klkvr

mattsse (approved):
this seems fine to relax

PR #17442, opened 2025-07-16
feat(pool): return state of an added tx

hai-rise:
This allows our `eth_sendRawTransactionSync` implementation (hopefully also more use cases) to know if the transaction was added as pending to keep waiting, or to return an early error that the user needs to fill a missing nonce or to top up their balance to "unlock" the transaction.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -61,8 +61,8 @@ use reth_primitives_traits::SignedTransaction;
 use reth_tokio_util::EventStream;
 use reth_transaction_pool::{
     error::{PoolError, PoolResult},
-    GetPooledTransactionLimit, PoolTransaction, PropagateKind, PropagatedTransactions,
-    TransactionPool, ValidPoolTransaction,
+    AddedTransactionOutcome, GetPooledTransactionLimit, PoolTransaction, PropagateKind,
+    PropagatedTransactions, TransactionPool, ValidPoolTransaction,
 };
 use std::{
     collections::{hash_map::Entry, HashMap, HashSet},
@@ -81,7 +81,8 @@ use tracing::{debug, trace};
 /// The future for importing transactions into the pool.
 ///
 /// Resolves with the result of each transaction import.
-pub type PoolImportFuture = Pin<Box<dyn Future<Output = Vec<PoolResult<TxHash>>> + Send + 'static>>;
+pub type PoolImportFuture =
+    Pin<Box<dyn Future<Output = Vec<PoolResult<AddedTransactionOutcome>>> + Send + 'static>>;
 
 /// Api to interact with [`TransactionsManager`] task.
 ///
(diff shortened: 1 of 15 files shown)


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -494,6 +494,7 @@ where
 
                 let added = pool.add_transaction(tx, balance, state_nonce, bytecode_hash)?;
                 let hash = *added.hash();
+                let subpool = added.subpool();
```
hai-rise: Main sauce.

crates/transaction-pool/tests/it/listeners.rs
```diff
@@ -113,7 +113,7 @@ async fn txpool_listener_all() {
 
     let added_result =
         txpool.add_transaction(TransactionOrigin::External, transaction.transaction.clone()).await;
-    assert_matches!(added_result, Ok(hash) if hash == *transaction.transaction.get_hash());
+    assert_matches!(added_result, Ok((hash, subpool)) if hash == *transaction.transaction.get_hash() && subpool == SubPool::Pending);
```
hai-rise: I suppose we can add more subpool tests, but trying to keep the changes minimal in this PR.

crates/transaction-pool/src/traits.rs
```diff
@@ -130,7 +130,7 @@ pub trait TransactionPool: Clone + Debug + Send + Sync {
     fn add_external_transaction(
         &self,
         transaction: Self::Transaction,
-    ) -> impl Future<Output = PoolResult<TxHash>> + Send {
+    ) -> impl Future<Output = PoolResult<(TxHash, SubPool)>> + Send {
```
hai-rise: Wonder if we want a "fuller" struct, but for now I feel `(TxHash, SubPool)` is clear and clean enough.

crates/net/network/src/transactions/mod.rs
```diff
@@ -81,7 +81,8 @@ use tracing::{debug, trace};
 /// The future for importing transactions into the pool.
 ///
 /// Resolves with the result of each transaction import.
-pub type PoolImportFuture = Pin<Box<dyn Future<Output = Vec<PoolResult<TxHash>>> + Send + 'static>>;
+pub type PoolImportFuture =
+    Pin<Box<dyn Future<Output = Vec<PoolResult<(TxHash, SubPool)>>> + Send + 'static>>;
```
mattsse: generally supportive, but I'm not a fan of this tuple because I very much disklike dealing with tuples for this in general, also this does leak the subpool internals.

my suggestion would be something like `Outcome { hash, state: Pending,Queued}`

mattsse (changes requested):
not opposed to this because this seems useful,
but I'd like to approach this differently and introduce a dedicated type for this, then we'll have an easier type in case we need to change anything else

crates/net/network/src/transactions/mod.rs
hai-rise: Ok, I've added more proper types. Would love to break `AddedTransactionState::Queued` into more concrete states but ran out of time, so maybe later..

hai-rise: @mattsse I cannot break down `AddedTransactionState::Queued` any time soon, wonder if we can merge this and open an issue in case the community can help? The `Pending`/`Queued` distinction is currently good enough for us, too.

mattsse:
@hai-rise I can get this over the line shortly

hai-rise:
@mattsse Sir :smiling_face_with_tear:..

mattsse (approved):
lgtm,

merging this after #17556 is out

PR #17444, opened 2025-07-16
feat: Forward transactions to a specified endpoint

### LINKED ISSUE #15990: Native eth forwarding support
### Describe the feature

**Have the ability to forward transactions received via the JSON RPC layer onto a specified endpoint and not forward them onto other peers / into the mempool.**

Have a use case for natively supporting forwarding transactions received on the JSON RPC layer rather than sending them to the mempool / other peers

One use case is you run a RETH node us its JSON RPC layer for eth_sendRawTransaction and want private transactions. Let's say you want to forward onto a builder endpoint rather than into the mempool.

You can achieve this outside of RETH with a lightweight proxy - Unfortunately we have entities who are asking for it to be specifically done via RETH - they don't want to maintain the proxy / do the code lift (even if small) / have another failure point

Would be great to be able to do this via a config / cli option

--------------------

Spoke to @mattsse  on the RETH telegram group he mentioned 

> If you open an issue for native eth forwarding support, same as opstack does, maybe we can also make this an option for vanilla reth

## TODO
* we'd need to install some hook next to 
https://github.com/paradigmxyz/reth/blob/b3d722f1fdc56de2354248dfd9a0e82f03941586/crates/rpc/rpc/src/eth/core.rs#L287-L288
that we invoke when we receive a new raw tx: https://github.com/paradigmxyz/reth/blob/b3d722f1fdc56de2354248dfd9a0e82f03941586/crates/rpc/rpc/src/eth/helpers/transaction.rs#L27-L27

similar to https://github.com/paradigmxyz/reth/blob/b3d722f1fdc56de2354248dfd9a0e82f03941586/crates/optimism/rpc/src/eth/transaction.rs#L46-L61

louisbrown0212:
Resolves #15990


crates/rpc/rpc-eth-types/src/builder/config.rs
```diff
@@ -45,6 +45,8 @@ pub struct EthConfig {
     pub fee_history_cache: FeeHistoryCacheConfig,
     /// The maximum number of getproof calls that can be executed concurrently.
     pub proof_permits: usize,
+    /// The raw transaction forwarder.
+    pub raw_tx_forwarder: Option<String>,
```
mattsse: I'd like to convert this to a ForwardConfig type, then it will be easier to add additional settings to this later

mattsse (changes requested):
only have a settings nit

I guess we could eventually unify this behaviour with the opethapi but having this as additional setting natively also seems okay

mattsse (approved):

PR #17449, opened 2025-07-16
fix: Rename `--authrpc.jwtsecret` to `--authrpc.jwt-secret`

maradini77:
**Description:**  
- Renamed the CLI flag and documentation references from `--authrpc.jwtsecret` to `--authrpc.jwt-secret` for consistency and clarity.


mattsse (commented):
this is wrong, because this arg is named

https://github.com/paradigmxyz/reth/blob/1e208710432fcb0d3a4b367a3c3977661305a1d1/crates/node/core/src/args/rpc_server.rs#L111-L111

PR #17450, opened 2025-07-16
refactor: move receipt conversions to `RpcConverter`

klkvr:
Added `RpcConverter::convert_receipts` and `RpcConverter` trait that allow to move the receipt convertions out of the EthApi/OpEthApi

Additionally changed `EthApiInner` to store and be generic over `RpcConvert` implementation


mattsse (approved):
makes sense, lgtm

PR #17451, opened 2025-07-16
feat: add TreePayloadValidator

fgimenez:
towards #17413

implements in `validate_payload` the same validation, execution and state root calculation logic as in `insert_block_inner`

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -1,0 +1,967 @@
+//! Concrete implementation of the `PayloadValidator` trait.
+
+use crate::tree::{
+    cached_state::CachedStateProvider,
+    instrumented_state::InstrumentedStateProvider,
+    payload_processor::PayloadProcessor,
+    precompile_cache::{CachedPrecompile, CachedPrecompileMetrics, PrecompileCacheMap},
+    ConsistentDbView, EngineApiMetrics, EngineApiTreeState, InvalidHeaderCache, PersistingKind,
+    StateProviderDatabase, TreeConfig,
+};
+use alloy_eips::BlockNumHash;
+use alloy_evm::{block::BlockExecutor, Evm};
+use alloy_primitives::B256;
+use reth_chain_state::CanonicalInMemoryState;
+use reth_consensus::{ConsensusError, FullConsensus};
+use reth_engine_primitives::InvalidBlockHook;
+use reth_evm::{ConfigureEvm, SpecFor};
+use reth_payload_primitives::NewPayloadError;
+use reth_primitives_traits::{
+    AlloyBlockHeader, Block, BlockBody, GotExpected, NodePrimitives, RecoveredBlock, SealedHeader,
+};
+use reth_provider::{
+    BlockExecutionOutput, BlockNumReader, BlockReader, DBProvider, DatabaseProviderFactory,
+    HashedPostStateProvider, HeaderProvider, ProviderError, StateCommitmentProvider, StateProvider,
+    StateProviderFactory, StateReader,
+};
+use reth_revm::db::State;
+use reth_trie::{updates::TrieUpdates, HashedPostState, TrieInput};
+use reth_trie_db::{DatabaseHashedPostState, StateCommitment};
+use reth_trie_parallel::root::{ParallelStateRoot, ParallelStateRootError};
+use std::{collections::HashMap, sync::Arc, time::Instant};
+use tracing::{debug, trace};
+
+/// Outcome of validating a payload
+#[derive(Debug)]
+pub enum PayloadValidationOutcome<Block: reth_primitives_traits::Block> {
+    /// Payload is valid and produced a block
+    Valid {
+        /// The block created from the payload
+        block: RecoveredBlock<Block>,
+        /// The trie updates from state root computation
+        trie_updates: reth_trie::updates::TrieUpdates,
+    },
+    /// Payload is invalid but block construction succeeded
+    Invalid {
+        /// The block created from the payload
+        block: RecoveredBlock<Block>,
+        /// The validation error
+        error: NewPayloadError,
+    },
+}
+
+/// Information about the current persistence state for validation context
+#[derive(Debug, Clone, Copy)]
+pub struct PersistenceInfo {
+    /// The last persisted block
+    pub last_persisted_block: BlockNumHash,
+    /// The current persistence action, if any
+    pub current_action: Option<PersistenceAction>,
+}
+
+impl PersistenceInfo {
+    /// Creates a new persistence info with no current action
+    pub const fn new(last_persisted_block: BlockNumHash) -> Self {
+        Self { last_persisted_block, current_action: None }
+    }
+
+    /// Creates persistence info with a saving blocks action
+    pub const fn with_saving_blocks(
+        last_persisted_block: BlockNumHash,
+        highest: BlockNumHash,
+    ) -> Self {
+        Self {
+            last_persisted_block,
+            current_action: Some(PersistenceAction::SavingBlocks { highest }),
+        }
+    }
+
+    /// Creates persistence info with a removing blocks action
+    pub const fn with_removing_blocks(
+        last_persisted_block: BlockNumHash,
+        new_tip_num: u64,
+    ) -> Self {
+        Self {
+            last_persisted_block,
+            current_action: Some(PersistenceAction::RemovingBlocks { new_tip_num }),
+        }
+    }
+}
+
+/// The type of persistence action currently in progress
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PersistenceAction {
+    /// Saving blocks to disk
+    SavingBlocks {
+        /// The highest block being saved
+        highest: BlockNumHash,
+    },
+    /// Removing blocks from disk
+    RemovingBlocks {
+        /// The new tip after removal
+        new_tip_num: u64,
+    },
+}
+
+/// Context providing access to tree state during validation
+pub struct TreeCtx<'a, N: NodePrimitives> {
+    /// The engine API tree state
+    state: &'a EngineApiTreeState<N>,
+    /// Information about the current persistence state
+    persistence_info: PersistenceInfo,
+    /// Reference to the canonical in-memory state
+    canonical_in_memory_state: &'a CanonicalInMemoryState<N>,
+}
+
+impl<'a, N: NodePrimitives> std::fmt::Debug for TreeCtx<'a, N> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TreeCtx")
+            .field("state", &"EngineApiTreeState")
+            .field("persistence_info", &self.persistence_info)
+            .field("canonical_in_memory_state", &self.canonical_in_memory_state)
+            .finish()
+    }
+}
+
+impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
+    /// Creates a new tree context
+    pub const fn new(
+        state: &'a EngineApiTreeState<N>,
+        persistence_info: PersistenceInfo,
+        canonical_in_memory_state: &'a CanonicalInMemoryState<N>,
+    ) -> Self {
+        Self { state, persistence_info, canonical_in_memory_state }
+    }
+
+    /// Returns a reference to the engine API tree state
+    pub const fn state(&self) -> &'a EngineApiTreeState<N> {
+        self.state
+    }
+
+    /// Returns a reference to the persistence info
+    pub const fn persistence_info(&self) -> &PersistenceInfo {
+        &self.persistence_info
+    }
+
+    /// Returns a reference to the canonical in-memory state
+    pub const fn canonical_in_memory_state(&self) -> &'a CanonicalInMemoryState<N> {
+        self.canonical_in_memory_state
+    }
+}
+
+/// A helper type that provides reusable payload validation logic for network-specific validators.
+///
+/// This type contains common validation, execution, and state root computation logic that can be
+/// used by network-specific payload validators (e.g., Ethereum, Optimism). It is not meant to be
+/// used as a standalone component, but rather as a building block for concrete implementations.
+pub struct TreePayloadValidator<N, P, C>
+where
+    N: NodePrimitives,
+    P: DatabaseProviderFactory<Provider: BlockReader + BlockNumReader + HeaderProvider>
+        + BlockReader
+        + BlockNumReader
+        + StateProviderFactory
+        + StateReader
+        + StateCommitmentProvider
+        + HashedPostStateProvider
+        + HeaderProvider<Header = N::BlockHeader>
+        + Clone
+        + 'static,
+    C: ConfigureEvm<Primitives = N> + 'static,
+{
+    /// Provider for database access.
+    provider: P,
+    /// Consensus implementation for validation.
+    consensus: Arc<dyn FullConsensus<N, Error = ConsensusError>>,
+    /// EVM configuration.
+    evm_config: C,
+    /// Configuration for the tree.
+    config: TreeConfig,
+    /// Payload processor for state root computation.
+    payload_processor: PayloadProcessor<N, C>,
+    /// Precompile cache map.
+    precompile_cache_map: PrecompileCacheMap<SpecFor<C>>,
+    /// Precompile cache metrics.
+    precompile_cache_metrics: HashMap<alloy_primitives::Address, CachedPrecompileMetrics>,
+    /// Tracks invalid headers to prevent duplicate hook calls.
+    invalid_headers: InvalidHeaderCache,
+    /// Hook to call when invalid blocks are encountered.
+    invalid_block_hook: Box<dyn InvalidBlockHook<N>>,
+    /// Metrics for the engine api.
+    metrics: EngineApiMetrics,
+}
+
+impl<N, P, C> std::fmt::Debug for TreePayloadValidator<N, P, C>
+where
+    N: NodePrimitives,
+    P: DatabaseProviderFactory<Provider: BlockReader + BlockNumReader + HeaderProvider>
+        + BlockReader
+        + BlockNumReader
+        + StateProviderFactory
+        + StateReader
+        + StateCommitmentProvider
+        + HashedPostStateProvider
+        + HeaderProvider<Header = N::BlockHeader>
+        + Clone
+        + std::fmt::Debug
+        + 'static,
+    C: ConfigureEvm<Primitives = N> + 'static,
+{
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,127 @@
...
+    evm_config: C,
+    /// Configuration for the tree.
+    config: TreeConfig,
+    /// Payload processor for state root computation.
+    payload_processor: PayloadProcessor<N, C>,
+    /// Precompile cache map.
+    precompile_cache_map: PrecompileCacheMap<SpecFor<C>>,
+    /// Precompile cache metrics.
+    precompile_cache_metrics: HashMap<alloy_primitives::Address, CachedPrecompileMetrics>,
+}
```
mattsse: this is so much nicer than having all of this part of the tree impl,

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,127 @@
...
+impl<N, P, C> TreePayloadValidator<N, P, C>
+where
+    N: NodePrimitives,
+    P: BlockReader
+        + StateProviderFactory
+        + StateReader
+        + StateCommitmentProvider
+        + Clone
+        + Unpin
+        + 'static,
```
mattsse: likely that we dont need to enforce all of these on struct def + in this scope

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,127 @@
...
+        + StateReader
+        + StateCommitmentProvider
+        + ChainSpecProvider
+        + Clone
+        + Unpin
+        + 'static,
+    C: ConfigureEvm<Primitives = N> + 'static,
+{
+    type Block = N::Block;
+    type ExecutionData = ExecutionData;
```
mattsse: now what is missing here is some additional trait component from which we get this, because this is is different eth vs op,

but perhaps we dont impl the trait for this type and instead reuse this container type for the op+eth impls

I think let's first try making this work for eth and then we can revisit what we need for op

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    fn validate_payload(
+        &self,
+        payload: Self::ExecutionData,
+        _ctx: TreeCtx<'_, EngineApiTreeState<N>>,
+    ) -> Result<PayloadValidationOutcome<Self::Block>, NewPayloadError> {
+        match self.ensure_well_formed_payload(payload) {
+            Ok(block) => Ok(PayloadValidationOutcome::Valid { block }),
+            Err(error) => Err(error),
+        }
```
mattsse: here we now need the logic this is currently in insert_block_inner

mattsse (commented):
nice progress

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,1008 @@
...
+    /// EVM configuration.
+    evm_config: C,
+    /// Configuration for the tree.
+    config: TreeConfig,
+    /// Payload processor for state root computation.
+    /// Uses Mutex for interior mutability since the `spawn` method requires &mut self
+    payload_processor: Mutex<PayloadProcessor<N, C>>,
+    /// Precompile cache map.
+    /// Uses Mutex for interior mutability since `cache_for_address` requires &mut self
+    precompile_cache_map: Mutex<PrecompileCacheMap<SpecFor<C>>>,
```
mattsse: I think we can avoid the mutexes here because from tree pov we can always have mutable access so we maybe just need to change the functions that access this

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,1008 @@
...
+    /// Precompile cache metrics.
+    ///
+    /// We use `Mutex` here for interior mutability because:
+    /// 1. The `PayloadValidator` trait requires `&self` (not `&mut self`) to support `Arc<dyn
+    ///    PayloadValidator>`
```
mattsse: hmm, not sure if we need this, perhaps this some flashbots validation limitation, but I feel we can solve this differently

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,1008 @@
...
+            invalid_block_hook,
+            metrics,
+        }
+    }
+
+    /// Converts an execution payload to a recovered block.
+    fn payload_to_block(
+        &self,
+        payload: ExecutionData,
+    ) -> Result<RecoveredBlock<N::Block>, NewPayloadError>
```
mattsse: this we can have here, this is ethereum specific logic this belongs to the ethereum specific wrapper

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,1008 @@
...
+        // Configure executor for the block
+        let mut executor = self.evm_config.executor_for_block(&mut db, block);
+
+        // Configure precompile caching if enabled
+        if !self.config.precompile_cache_disabled() {
+            // Get the spec id before the closure
+            let spec_id = *self.evm_config.evm_env(block.header()).spec_id();
+
+            // Lock both the cache map and metrics map for the duration of precompile setup
+            let mut cache_map = self.precompile_cache_map.lock().unwrap();
```
mattsse: if we end up using mutexes here then this should always be parkinglot

crates/engine/primitives/src/lib.rs
```diff
@@ -59,11 +66,79 @@ pub enum PayloadValidationOutcome<Block: reth_primitives_traits::Block> {
...
 /// Context providing access to tree state during validation
 #[derive(Debug)]
-pub struct TreeCtx<'a, State> {
+pub struct TreeCtx<'a, State, N: NodePrimitives = reth_ethereum_primitives::EthPrimitives> {
     /// The engine API tree state
     pub state: &'a State,
+    /// Information about the current persistence state
+    pub persistence_info: PersistenceInfo,
+    /// Reference to the canonical in-memory state
+    pub canonical_in_memory_state: &'a CanonicalInMemoryState<N>,
```
mattsse: these should all be private

crates/engine/primitives/src/lib.rs
```diff
@@ -59,11 +66,79 @@ pub enum PayloadValidationOutcome<Block: reth_primitives_traits::Block> {
...
+}
+
 /// Context providing access to tree state during validation
 #[derive(Debug)]
-pub struct TreeCtx<'a, State> {
+pub struct TreeCtx<'a, State, N: NodePrimitives = reth_ethereum_primitives::EthPrimitives> {
     /// The engine API tree state
     pub state: &'a State,
+    /// Information about the current persistence state
+    pub persistence_info: PersistenceInfo,
```
mattsse: do we need all of this info for state root validation?

crates/engine/primitives/src/lib.rs
```diff
@@ -130,7 +205,9 @@ pub trait EngineTypes:
 
 /// Type that validates an [`ExecutionPayload`].
 #[auto_impl::auto_impl(&, Arc)]
-pub trait PayloadValidator<State = ()>: Send + Sync + Unpin + 'static {
+pub trait PayloadValidator<State = (), N: NodePrimitives = reth_ethereum_primitives::EthPrimitives>:
+    Send + Sync + Unpin + 'static
```
mattsse: we could even leave this unchanged for now, then it would be easier to integrate all of this in a next step

crates/engine/primitives/src/lib.rs
fgimenez: sure done, added getters

crates/engine/primitives/src/lib.rs
fgimenez: good point, no changes in `PayloadValidator` here now the branch targets main

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: yep, removed for now and methods get `&mut self`

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: yep, removed for now

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: we use this method in `validate_payload`, should we do the conversion payload -> block in other way?

crates/engine/primitives/src/lib.rs
fgimenez: yep, it is required for determine if we can run parallel state root computation (either no blocks being persisted or the blocks being persisted are ancestors of the current block) and to efficiently calculate trie input (if we're in PersistingDescendant mode, the system filters out blocks that have already been persisted)

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: removed `payload_to_block` and left only reusable logic here https://github.com/paradigmxyz/reth/pull/17451/commits/eec53b6623a0c8dd836f06feb4feb1d438dd19e3 ptal

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+            // Fallback to database if not in tree state
+            let header: N::BlockHeader = self
+                .provider
+                .header(&parent_hash)
+                .map_err(|e| ConsensusError::Other(e.to_string()))?
+                .ok_or_else(|| {
+                    ConsensusError::Other(format!("Parent header not found: {parent_hash}"))
+                })?;
+            SealedHeader::seal_slow(header)
+        };
```
klkvr: can this use `self.get_parent_header`? also wondering if this should be a helper on `ctx`

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+                        parent_hash.into(),
+                    )));
+                    return Ok(PayloadValidationOutcome::Invalid { block, error });
+                }
+                Err(e) => {
+                    let error = NewPayloadError::Other(Box::new(e));
+                    return Ok(PayloadValidationOutcome::Invalid { block, error });
+                }
+            }
+        };
```
klkvr: ideally we should also hide this in context type, feel free to skip for this PR though

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+
+    /// Creates a payload handle for the given block.
+    ///
+    /// This method decides whether to use full spawn (with background state root tasks)
+    /// or cache-only spawn based on the current conditions.
+    ///
+    /// Returns a tuple of (`PayloadHandle`, `use_state_root_task`) where `use_state_root_task`
+    /// indicates whether the state root task was actually enabled (it may be disabled
+    /// if prefix sets are non-empty).
+    fn create_payload_handle(
```
klkvr: should we call this `spawn_payload_tasks` or similar given that is spawns things?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+        &mut self,
+        block: &RecoveredBlock<N::Block>,
+        provider_builder: crate::tree::StateProviderBuilder<N, P>,
+        use_state_root_task: bool,
+        parent_hash: B256,
```
klkvr: nit: we can omit the `parent_hash` arg as it's available in the `block`

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+            }
+        }
+    }
+
+    /// Computes state root with appropriate strategy based on configuration.
+    #[allow(clippy::too_many_arguments)]
+    fn compute_state_root_with_strategy(
+        &self,
+        block: &RecoveredBlock<N::Block>,
+        parent_hash: B256,
```
klkvr: same here

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+    /// Executes the given block using the provided state provider.
+    fn execute_block<S>(
+        &mut self,
+        state_provider: &S,
+        block: &RecoveredBlock<N::Block>,
+        handle: &crate::tree::PayloadHandle,
+    ) -> Result<(BlockExecutionOutput<N::Receipt>, Instant), NewPayloadError>
+    where
+        S: StateProvider,
+        N::Block: Block<Body: BlockBody<Transaction = N::SignedTx>>,
```
klkvr:
```suggestion
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+    /// 2. `InstrumentedStateProvider` for metrics (if enabled)
+    fn execute_block_with_state_provider<S>(
+        &mut self,
+        state_provider: S,
+        block: &RecoveredBlock<N::Block>,
+        handle: &crate::tree::PayloadHandle,
+    ) -> Result<(BlockExecutionOutput<N::Receipt>, Instant), NewPayloadError>
+    where
+        S: StateProvider,
+        N::Block: Block<Body: BlockBody<Transaction = N::SignedTx>>,
```
klkvr:
```suggestion
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+    /// Attempts to get the state root from the background task.
+    fn try_state_root_from_task(
+        &self,
+        handle: &mut crate::tree::PayloadHandle,
+        block: &RecoveredBlock<N::Block>,
+        execution_time: Instant,
+    ) -> Option<(B256, TrieUpdates)>
+    where
+        N::Block: Block,
+    {
```
klkvr:
```suggestion
    ) -> Option<(B256, TrieUpdates)> {
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+        tree_state: &EngineApiTreeState<N>,
+        persisting_kind: PersistingKind,
+        run_parallel_state_root: bool,
+        use_state_root_task: bool,
+        handle: &mut crate::tree::PayloadHandle,
+        execution_time: Instant,
+    ) -> Result<(B256, TrieUpdates), NewPayloadError>
+    where
+        N::Block: Block,
+    {
```
klkvr:
```suggestion
    ) -> Result<(B256, TrieUpdates), NewPayloadError> {
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,938 @@
...
+        block: &RecoveredBlock<N::Block>,
+        provider_builder: crate::tree::StateProviderBuilder<N, P>,
+        use_state_root_task: bool,
+        parent_hash: B256,
+        tree_state: &EngineApiTreeState<N>,
+        persistence_info: &reth_engine_primitives::PersistenceInfo,
+    ) -> (crate::tree::PayloadHandle, bool)
+    where
+        N::Block: Block<Body: BlockBody<Transaction = N::SignedTx>>,
+    {
```
klkvr:
```suggestion
    ) -> (crate::tree::PayloadHandle, bool) {
```

klkvr (changes requested):
looks good to me, only have nits

`TreePayloadValidator` is quite huge which imo flags that it still not fully feasible for custom nodes to define entire engine themselves

but we can likely decompose this more and make more parts reusable further

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: good point, with `TreeCtx` defined in `reth-engine-primitives` creating there a helper method to generate `StateProviderBuilder` would create a circular dep, for now i've created the helper here in `TreePayloadValidator`

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: of course good catch. for now left the helper in `TreePayloadValidator`, we don't have provider access in ctx at this point

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: done

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: much better, done

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: nice thx! done

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -0,0 +1,922 @@
...
+            );
+            let error = NewPayloadError::Other(Box::new(ConsensusError::BodyStateRootDiff(
+                GotExpected { got: state_root, expected: block.header().state_root() }.into(),
+            )));
+            return Ok(PayloadValidationOutcome::Invalid { block, error });
+        }
+
+        // Check if the block is a fork
+        let is_fork =
+            ensure_ok!(self.is_fork(block.sealed_header(), tree_state, canonical_in_memory_state));
```
mattsse: should we determine this here if the tree itself can determine this outside of this impl?

fgimenez: great point, removed

crates/engine/primitives/src/lib.rs
```diff
@@ -42,6 +44,115 @@ pub use invalid_block_hook::InvalidBlockHook;
...
+        /// The new tip after removal
+        new_tip_num: u64,
+    },
+}
+
+/// Context providing access to tree state during validation
+#[derive(Debug)]
+pub struct TreeCtx<'a, State, N: NodePrimitives = reth_ethereum_primitives::EthPrimitives> {
+    /// The engine API tree state
+    state: &'a State,
```
mattsse: perhaps we can remove this generic later and we can avoid this by using just the engineState

mattsse (approved):
this now more or less mirrors what we currently have in the tree itself.

I expect that we'd need to fine tune a few things when we embed this in the eth+op variant.

but since this mirrors the current impl, I think we should merge this as is since this isnt not integrated yet, this doesnt affect existing code

klkvr (approved):

crates/engine/primitives/src/lib.rs
fgimenez: sure this was a leftover, using now the proper `state: &'a EngineApiTreeState<N: NodePrimitives>`

PR #17456, opened 2025-07-17
feat(alloy-provider): implement `receipt_by_hash` method

cakevm:
Implementing the `receipt_by_hash` using the same pattern as for the rpc block.

I implemented some conversion in the primitives for eth and op. I was undecided if it is worth to feature gate them. The op conversions follow a similar pattern as the eth. Have a conversion for the consensus type and the rpc type.


crates/ethereum/primitives/Cargo.toml
```diff
@@ -22,6 +22,7 @@ alloy-eips = { workspace = true, features = ["k256"] }
 alloy-primitives.workspace = true
 alloy-consensus = { workspace = true, features = ["serde"] }
 alloy-rlp.workspace = true
+alloy-rpc-types-eth.workspace = true
```
mattsse: this is quite a heavy import just from conversion

mattsse (changes requested):
doesnt feel right to add rpc deps just for this,

we could either feature gate this or perhaps better, try implementing the trait manually

crates/ethereum/primitives/Cargo.toml
cakevm: Mhh, I have the feeling with a feature flag it is easier to support for other networks since we not need to extend the rpc-convert. Maybe I add it quick, and can look how it feels.

cakevm:
How do you feel about the flag? As alternative I can try push the conversion to `rpc-convert` for eth / op. But it also feels like it makes sense to have a rpc -> reth primitives conversion close to the type.

cakevm:
I forgot about the AnyNetwork. It is currently not supported, but would add more complexity to the primitives. So I pushed the conversion now to `rpc-convert`.

mattsse (approved):

PR #17464, opened 2025-07-17
feat: add helper for building pending block env

klkvr:
Introduces `PendingEnvBuilder` trait that encapsulates building of the EVM environment for the pending block. It defaults to `BasicPendingEnvBuilder` which relies on the `NextBlockEnvAttributes` providing implementation directly via `BuildPendingEnv`: 
https://github.com/paradigmxyz/reth/blob/69d0bd49773c6c50f3df79772b741d5c9ddeedc5/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L357-L384

More complex implementations can be provided via `EthApiBuilder::with_pending_env_builder`:
https://github.com/paradigmxyz/reth/blob/69d0bd49773c6c50f3df79772b741d5c9ddeedc5/crates/rpc/rpc/src/eth/builder.rs#L146-L150

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9244,6 +9244,7 @@ dependencies = [
  "reth-optimism-primitives",
  "reth-primitives-traits",
  "reth-revm",
+ "reth-rpc-eth-api",
  "revm",
  "thiserror 2.0.12",
 ]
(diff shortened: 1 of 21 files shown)


mattsse (approved):
makes sense

mattsse (approved):
lgtm! 🔥

PR #17473, opened 2025-07-17
fix: Update Base Sequencer URL to mainnet.base.org

maradini77:
**Description:**  
Replaces the outdated Base Sequencer endpoint (mainnet-sequencer.base.org) with the correct and working mainnet.base.org URL.


mattsse (commented):
this should be the sequencer ulr not the public rpc url

PR #17474, opened 2025-07-17
docs: correct typo in comment

rnkrtt:
Fixed simple typo
`specfified` - `specified`


mattsse (commented):
there's currently a lot of movement in this crate so I want to avoid uneccessary conflicts

PR #17476, opened 2025-07-18
feat: make basic block builder pub

0xOsiris:
I would like to create a type wrapper around this struct for a block builder that can have some pre-state loaded into it. It would be useful to have this public.


mattsse (approved):
any concerns here @klkvr ?

Issue #16895, opened 2025-06-18
Support for ERA file format


RomanHodulak:
### Describe the feature

Currently we have the encoding for ERA1 implemented in the `reth_era` crate.

https://github.com/paradigmxyz/reth/blob/4a6b2837e6c1662d2b21a06638b5e1228d986c6e/crates/era/Cargo.toml#L2

The goal is to add support for the ERA format, as described in the parent issue, or ensure that the support already exists.

PR #17477, opened 2025-07-18
feat(era): add era types

lean-apple:
Part 1 for https://github.com/paradigmxyz/reth/issues/16895.
(the full file format would be done in a second pr).


mattsse (approved):
very nice

the indexentry trait makes a lot of sense so we dont need to duplicate this logic

PR #17488, opened 2025-07-18
perf(trie): Re-use storage tries across payloads

mediocregopher:
This is implemented by re-using the entire SparseStateTrie across payloads, so that we keep reusing the accounts trie as well.

There's a small change to how storage tries are fetched/created in SparseStateTrie, the rest of this change has to do with passing SparseStateTrie around. A new type is introduced for that, ClearedSparseStateTrie, which makes it easy to enforce at compile-time that only a cleared state is being used in a payload task.

## Benchmarks (mainnet, 1000 blocks)

```
Baseline: main
Feature:  origin/mediocregopher/17457-reuse-storage-tries

Performance Changes:
  NewPayload Latency: -4.11%
  FCU Latency:        -6.72%
  Total Latency:      -4.18%
  Gas/Second:         +4.36%
  Blocks/Second:      +4.36%

Baseline Summary:
  Blocks: 1000, Gas: 17906209584, Duration: 43.10s
  Avg NewPayload: 41.96ms, Avg FCU: 1.12ms, Avg Total: 43.08ms
  Started: 2025-07-18 13:31:54 UTC, Ended: 2025-07-18 13:36:20 UTC

Feature Summary:
  Blocks: 1000, Gas: 17906209584, Duration: 41.30s
  Avg NewPayload: 40.23ms, Avg FCU: 1.04ms, Avg Total: 41.28ms
  Started: 2025-07-18 13:37:35 UTC, Ended: 2025-07-18 13:41:59 UTC
```

latency_comparison

Fixes #17457

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -77,7 +77,7 @@ where
     /// A cleared `SparseStateTrie`, kept around to be reused for the state root computation so
     /// that allocations can be minimized.
     sparse_state_trie: Arc<
-        tokio::sync::Mutex<Option<ClearedSparseStateTrie<ConfiguredSparseTrie, SerialSparseTrie>>>,
+        parking_lot::Mutex<Option<ClearedSparseStateTrie<ConfiguredSparseTrie, SerialSparseTrie>>>,
     >,
     /// Whether to use the parallel sparse trie.
     use_parallel_sparse_trie: bool,
@@ -325,8 +325,8 @@ where
     {
         // Reuse a stored SparseStateTrie, or create a new one using the desired configuration if
         // there's none to reuse.
-        let mut sparse_state_trie_lock_guard = self.sparse_state_trie.clone().blocking_lock_owned();
-        let sparse_state_trie = sparse_state_trie_lock_guard.take().unwrap_or_else(|| {
+        let cleared_sparse_trie = Arc::clone(&self.sparse_state_trie);
+        let sparse_state_trie = cleared_sparse_trie.lock().take().unwrap_or_else(|| {
             let accounts_trie = if self.use_parallel_sparse_trie {
                 ConfiguredSparseTrie::Parallel(Default::default())
             } else {
@@ -352,9 +352,10 @@ where
             let (result, trie) = task.run();
             // Send state root computation result
             let _ = state_root_tx.send(result);
-            // Clear the SparseStateTrie and replace it back into the mutex _after_ returning, so
-            // that time spent clearing doesn't block the step after this one.
-            sparse_state_trie_lock_guard.replace(ClearedSparseStateTrie::cleared(trie));
+
+            // Clear the SparseStateTrie and replace it back into the mutex _after_ acquiring the
+            // lock, so that time spent clearing doesn't block the step after this one.
+            cleared_sparse_trie.lock().replace(ClearedSparseStateTrie::cleared(trie));
         });
     }
 }
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -332,10 +332,11 @@ where
             } else {
                 ConfiguredSparseTrie::Serial(Default::default())
             };
-            SparseStateTrie::new()
-                .with_accounts_trie(SparseTrie::Blind(Some(Box::new(accounts_trie))))
-                .with_updates(true)
-                .into()
+            ClearedSparseStateTrie::cleared(
+                SparseStateTrie::new()
+                    .with_accounts_trie(SparseTrie::Blind(Some(Box::new(accounts_trie))))
+                    .with_updates(true),
+            )
         });
 
         let task =
@@ -353,7 +354,7 @@ where
             let _ = state_root_tx.send(result);
             // Clear the SparseStateTrie and replace it back into the mutex _after_ returning, so
             // that time spent clearing doesn't block the step after this one.
-            sparse_state_trie_lock_guard.replace(trie.into());
+            sparse_state_trie_lock_guard.replace(ClearedSparseStateTrie::cleared(trie));
         });
     }
 }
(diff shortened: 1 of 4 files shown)


crates/trie/sparse/src/state.rs
```diff
@@ -166,6 +200,16 @@ where
...
 
+    /// Retrieves the storage trie for the given address, creating a new one if it doesn't exist.
+    ///
+    /// This method should always be used to create a storage trie, as it will re-use previously
+    /// allocated and cleared storage tries when possible.
+    fn get_or_create_storage_trie(&mut self, address: B256) -> &mut SparseTrie<S> {
+        self.storages
+            .entry(address)
+            .or_insert_with(|| self.cleared_storages.pop().unwrap_or_default())
+    }
```
mattsse, line 207: I see, this will either retrieve an existing trie or reuse an existing, cleared trie

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -75,9 +74,11 @@ where
     precompile_cache_disabled: bool,
     /// Precompile cache map.
     precompile_cache_map: PrecompileCacheMap<SpecFor<Evm>>,
-    /// A cleared accounts sparse trie, kept around to be reused for the state root computation so
+    /// A cleared `SparseStateTrie`, kept around to be reused for the state root computation so
     /// that allocations can be minimized.
-    accounts_trie: Arc<Mutex<Option<SparseTrie<ConfiguredSparseTrie>>>>,
+    sparse_state_trie: Arc<
+        tokio::sync::Mutex<Option<ClearedSparseStateTrie<ConfiguredSparseTrie, SerialSparseTrie>>>,
```
mattsse: Is there a specific reason for using tokio mutex here, afaict we only hold the lock guard across memory operations

what this does is basically `future::block_on(mutex.lock())`, 
a parkinglot mutex would be more appropriate here, since we never use this in async context

crates/trie/sparse/src/state.rs
```diff
@@ -20,6 +20,40 @@ use reth_trie_common::{
...
+    A = SerialSparseTrie, // Account trie implementation
+    S = SerialSparseTrie, // Storage trie implementation
+>(SparseStateTrie<A, S>);
+
+impl<A, S> From<SparseStateTrie<A, S>> for ClearedSparseStateTrie<A, S>
+where
+    A: SparseTrieInterface,
+    S: SparseTrieInterface,
+{
+    fn from(mut trie: SparseStateTrie<A, S>) -> Self {
```
mattsse: I'd prefer to have this as `fn cleared(SparseStateTrie<A, S>)` instead, which should make it clear how the sparsestatetrie is converted

crates/trie/sparse/src/state.rs
```diff
@@ -20,6 +20,40 @@ use reth_trie_common::{
...
+    }
+}
+
+impl<A, S> ClearedSparseStateTrie<A, S>
+where
+    A: SparseTrieInterface + Default,
+    S: SparseTrieInterface + Default,
+{
+    /// Returns the cleared `SparseStateTrie`, consuming this instance.
+    pub fn unwrap(self) -> SparseStateTrie<A, S> {
```
mattsse: I'd like to name this into_inner, so that this cant be mistaken for a fallible Option::unwrap

mattsse (changes requested):
cool, this makes a lot of sense

mostly style/naming nits

crates/engine/tree/src/tree/payload_processor/mod.rs
mediocregopher: In the previous iteration of the code it was still possible that the next payload would be started before the previous one had finished clearing its SparseStateTrie and replaced it into the `Arc<Mutex<Option<...>>>`. In this case a new SparseStateTrie would be created, with completely un-allocated account/storage tries and other buffers. It seemed to me better to block briefly to wait 

This issue occurred because we released the mutex outside the `self.executor.spawn_blocking`, and then re-grabbed it inside. I fixed this by not releasing the lock initially, and instead moving the `sparse_state_trie_lock_guard` into the `self.executor.spawn_blocking` closure, and then call `replace` on that guard to release it. But to do this we need a mutex/guard which is `Send`, which afaict the parking_lot mutex is not.

mattsse (approved):
lgtm

pending @shekhirin or @Rjected

crates/trie/sparse/src/state.rs
```diff
@@ -20,6 +20,36 @@ use reth_trie_common::{
...
+>(SparseStateTrie<A, S>);
+
+impl<A, S> ClearedSparseStateTrie<A, S>
+where
+    A: SparseTrieInterface + Default,
+    S: SparseTrieInterface + Default,
+{
+    /// Creates a [`ClearedSparseStateTrie`] by clearing all the existing internal state of a
+    /// [`SparseStateTrie`] and then storing that instance for later re-use.
+    pub fn cleared(mut trie: SparseStateTrie<A, S>) -> Self {
```
shekhirin: I think something like `from_state_trie` should fit better

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -320,128 +312,52 @@ where
...
-                SparseTrie::Blind(opt) => {
-                    SparseTrie::Blind(opt.map(|t| Box::new(ConfiguredSparseTrie::from(*t))))
-                }
-                SparseTrie::Revealed(t) => {
-                    SparseTrie::Revealed(Box::new(ConfiguredSparseTrie::from(*t)))
-                }
-            };
-            accounts_trie.lock().replace(trie.clear());
+            // Clear the SparseStateTrie and replace it back into the mutex _after_ acquiring the
+            // lock, so that time spent clearing doesn't block the step after this one.
```
shekhirin: wdym by this comment?

shekhirin (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
mediocregopher: Out of date comment, fixed it

PR #17490, opened 2025-07-18
feat: `HeaderConverter`

klkvr:
Introduces `HeaderConverter` which allows to customize the primitive -> RPC header convertions. That way we are able to remove last bound on RPC types embedded into RPC traits: 
https://github.com/paradigmxyz/reth/blob/65a63e129e90b792c51591f8b1ae5126dce93d22/crates/optimism/rpc/src/eth/pending_block.rs#L27-L29


crates/primitives-traits/src/block/recovered.rs
```diff
@@ -608,44 +608,23 @@ mod rpc_compat {
         ///
         /// The `tx_resp_builder` closure transforms each transaction into the desired response
         /// type.
-        pub fn into_rpc_block<T, F, E>(
+        pub fn into_rpc_block<T, RpcH, F, E>(
```
mattsse, line 612: can we clarify in the docs that the size of the clisure here is the rlp_length of the block?

crates/primitives-traits/src/block/recovered.rs
```diff
@@ -671,27 +650,31 @@ mod rpc_compat {
         ///
         /// Consumes self and returns [`BlockTransactions::Hashes`] containing only transaction
         /// hashes.
-        pub fn into_rpc_block_with_tx_hashes<T>(self) -> Block<T, Header<B::Header>> {
+        pub fn into_rpc_block_with_tx_hashes<T, E, RpcHeader>(
+            self,
+            f: impl Fn(SealedHeader<B::Header>, usize) -> Result<RpcHeader, E>,
```
mattsse: since we can convert `Header<B::Header>` into `SealedHeader<B::Header>`, I wonder if we should change the closure to that?

mattsse: can this be FnOnce?

crates/rpc/rpc-eth-api/src/helpers/block.rs
```diff
@@ -64,9 +63,11 @@ pub trait EthBlocks:
         async move {
             let Some(block) = self.recovered_block(block_id).await? else { return Ok(None) };
 
-            let block = block.clone_into_rpc_block(full.into(), |tx, tx_info| {
-                self.tx_resp_builder().fill(tx, tx_info)
```
mattsse, line 68: this is slightly less efficient, the reason why `clone_into_rpc_block` exists is that this is a lot cheaper for the `Kind::Hash` case

and since this block is coming from the cache, this is almost never a single arc instance

crates/primitives-traits/src/block/recovered.rs
```diff
@@ -608,44 +608,23 @@ mod rpc_compat {
...
-        /// For full transactions, the entire block is cloned.
-        ///
-        /// The `tx_resp_builder` closure transforms each transaction into the desired response
-        /// type.
-        pub fn clone_into_rpc_block<T, F, E>(
-            &self,
-            kind: BlockTransactionsKind,
-            tx_resp_builder: F,
-        ) -> Result<Block<T, Header<B::Header>>, E>
-        where
```
mattsse: could we keep that?

we could also maybe change this `self: Arc<Self>`

see explainer why this function exists

mattsse (changes requested):
makes sense

left 1 idea re closure, and smol request to keep the clone_ function

mattsse (approved):
lgtm

PR #17497, opened 2025-07-18
feat: Add IPC socket permission configuration

### LINKED ISSUE #15806: Add config parameter for UNIX socket permissions
### Describe the feature

At present the UNIX socket file permissions cannot be explicitly specified. Would it be possible to add a config parameter to allow this?

### Additional context

_No response_

danielrachi1:
## Issue Addressed

Fixes #15806 

## Proposed Changes

- Add `ipc.permissions` setting to `node`.
- Handle the transmission of this setting down to the IPC server builder.
- Use `os::unix::fs::PermissionsExt` to write the provided permissions to the file/socket.

### DIFF
--- a/crates/rpc/ipc/src/server/mod.rs
+++ b/crates/rpc/ipc/src/server/mod.rs
@@ -19,7 +19,6 @@ use jsonrpsee::{
 use std::{
     future::Future,
     io,
-    os::unix::fs::PermissionsExt,
     pin::{pin, Pin},
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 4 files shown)


crates/rpc/ipc/src/server/mod.rs
```diff
@@ -139,7 +140,15 @@ where
             .to_fs_name::<GenericFilePath>()
             .and_then(|name| ListenerOptions::new().name(name).create_tokio())
         {
-            Ok(listener) => listener,
+            Ok(listener) => {
+                if let Some(perms_str) = &self.cfg.ipc_socket_permissions {
+                    if let Ok(mode) = u32::from_str_radix(&perms_str.replace("0o", ""), 8) {
```
mattsse: this should do it, can't think of a better way to do this, and I also don't think we should parse this on cli because configuring string seems easier for this

mattsse (approved):
cool, tysm!

PR #17498, opened 2025-07-18
feat(storage): add max read transaction duration support

### LINKED ISSUE #17418: feat(provider): expose MaxReadTransactionDuration
### Describe the feature

As per our convo with Matthias, creating an issue on exposing `MaxReadTransactionDuration` to the public API of `ProviderFactory`. 

Reasons:

If an application is using `ProviderFactory` as an argument, it would be very useful to be able to check what `MaxReadTransactionDuration` was set to.

### Additional context

I am building an AMMS-rs alternative that uses `ProviderFactory` as a sole-provider. The `StateProvider` sometimes times out due to a lengthy initial sync, therefore, I need a way to check if the safety timeout was set. This will allow me to either continues the normal logic or spawn a separate thread to keep alive the `StateProvider` by closing and reopening the tx right before the timelimit.

Lil-Duckling-22:
## Description:
Added support for configuring and exposing max_read_transaction_duration in ProviderFactory

### Сhanges:

- Added max_read_transaction_duration field to ProviderFactory

- New constructor: new_with_max_read_transaction_duration

- Integrated with ProviderFactoryBuilder

- Added tests for duration handling

### DIFF
--- a/crates/storage/provider/src/providers/database/mod.rs
+++ b/crates/storage/provider/src/providers/database/mod.rs
@@ -15,6 +15,7 @@ use reth_chainspec::ChainInfo;
 use reth_db::{init_db, mdbx::DatabaseArguments, DatabaseEnv};
 use reth_db_api::{database::Database, models::StoredBlockBodyIndices};
 use reth_errors::{RethError, RethResult};
+use reth_libmdbx::MaxReadTransactionDuration;
 use reth_node_types::{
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/database/mod.rs
mattsse, line 1: hmm, this doesn't do anything really, because this setting is a setting for the database itself

mattsse (commented):
we've now duplicated this setting on the providerfactory without it being used there.

and we already have a way to configure this setting on the DatabaseArgs.
we need to rethink this a bit but atm I dont have an idea

PR #17503, opened 2025-07-19
chore: fix reth-engine-tree dev-dependencies import

viktorking7:
Updates `reth-evm-ethereum` dependency in `crates/engine/service/Cargo.toml` to include the `test-utils` feature for development dependencies.


mattsse (commented):
these arent necessary

PR #17504, opened 2025-07-19
fix: remove duplicate comment in capabilities.rs

rnkrtt:
Remove duplicate comment "The list of all supported Engine capabilities available over the engine endpoint" that was repeated on line 20. The comment was already present on line 3 and was redundant between the struct definition and its implementation

image


mattsse (commented):
this is a comment for the type, I think we can keep this one

PR #17505, opened 2025-07-19
fix: temporary file leak in atomic_write_file

anim001k:
## Description

Fixed a logical error in the `atomic_write_file` function that caused temporary files to remain on disk when `write_fn` failed. The function now properly cleans up temporary files before returning errors, preventing file system pollution.

### Changes

- **Replaced** `map_err` chain with `match` statement for proper error handling
- **Added** temporary file cleanup on `write_fn` failure
- **Improved** code comments for better maintainability

### Problem

The original code would leave `.tmp` files on disk when `write_fn` returned an error, leading to:
- Resource leaks
- File system clutter  
- Potential confusion for users

### Solution

Now when `write_fn` fails, the function:
1. Removes the temporary file
2. Returns the original error wrapped in `FsPathError::Write`

### DIFF
--- a/crates/fs-util/src/lib.rs
+++ b/crates/fs-util/src/lib.rs
@@ -323,10 +323,21 @@ where
     let mut file =
         File::create(&tmp_path).map_err(|err| FsPathError::create_file(err, &tmp_path))?;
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I see, that makes sense

PR #17509, opened 2025-07-20
docs: normalize dynamic CLI defaults in help generation

AJStonewee:
Fixes inconsistent CLI documentation caused by dynamic default values.

Two CLI args have defaults calculated from CPU cores:
- `--engine.max-proof-task-concurrency` (CPU cores * 8)  
- `--engine.reserved-cpu-cores` (min(2, CPU cores))

These showed different values on different machines, making docs inconsistent.

Now replaces actual numbers with descriptive text like existing `--rpc.max-tracing-requests` handling.

**Before:** `[default: 16]`  
**After:** `[default: <DYNAMIC: CPU cores * 8>]`


mattsse (approved):
makes sense

PR #17512, opened 2025-07-20
ethereum: Fix hardcoded error type in Ethereum payload builder

LandLord529:
When marking invalid transactions, the code was always returning `TxTypeNotSupported` instead of the actual error from validation.

This meant users would see "transaction type not supported" for any invalid tx - insufficient funds, bad nonce, etc. Makes debugging really confusing.

Now it properly forwards the real error type.


mattsse (commented):
this doesn't compile,

this error is likely fine but we'd need to find a different way to convert the error type here

PR #17514, opened 2025-07-21
Rename 'transaction' field to 'header' in SealedHeader bincode test for clarity

davidjsonn:
Description:  
Renamed the field to match its actual type and usage


mattsse (commented):
this doesnt compile, since this is just a test, this is fine

PR #17515, opened 2025-07-21
docs: typos

crStiv:
just fixed typos in documentation as usual


mattsse (commented):
these are auto-generated from the cli docs, so we'd need to fix them where the structs are defined

since this is a different change, I'm closing this one

PR #17517, opened 2025-07-21
docs(trace): document trace format and response structure

avorylli:
## Motivation
The trace API lacked formal documentation of its return formats, making it hard for users to interpret trace responses.

## Solution
Added detailed documentation for:

- TraceResults

- LocalizedTransactionTrace

- All action types (call, create, suicide, reward)

- result, stateDiff, and vmTrace formats

Helps developers better understand and integrate with trace RPC methods.


mattsse (approved):
this looks correct,

it's nice to have an actual reference for this

PR #17518, opened 2025-07-21
fix: error handling by properly unwrapping Result with the `?`

mdqst:


mattsse (commented):
this is fine because we always want to record the elapsed time

PR #17520, opened 2025-07-21
feat: make payload validation functions generic over block header type

rezzmah:
## Summary

Updates payload validation functions to be generic over block header types instead of being hardcoded to specific block implementations. This change enables chains with custom block headers to utilize the existing validation logic.

## Changes

- Added generic header type parameter `H` to payload validation functions in `cancun.rs` and `prague.rs`
- Updated `BlockBody<T>` usage to `BlockBody<T, H>` for type consistency
- Maintains identical validation behavior while improving flexibility for downstream chains

This change is particularly beneficial for chains like Optimism, Berachain, and other custom implementations that extend Ethereum's block structure with additional header fields.


mattsse (approved):
makes sense

PR #17524, opened 2025-07-21
refactor: rename `AlloyRethProvider` to `RpcBlockchainProvider` and move to storage

cakevm:
To make the purpose of the implementation clearer, it has been renamed to `RpcBlockchainProvider`, as it functions like `BlockchainProvider` but is based on RPC. Since it is closely related to `reth-provider`, it has also been moved to the storage folder and renamed to `reth-storage-rpc-provider`. Remove the `AsyncDbWrapper` since it is not required for this scenario.


mattsse (approved):
no strong opinion here, but this makes sense

PR #17526, opened 2025-07-21
fix: ensure required revm features are activated

mattsse:
regression from #17450

we need to force activate these features for the node

also added asm-keccak feature to ethereum-node for consistency


mediocregopher (approved):

klkvr (approved):
oops sorry about this

PR #17529, opened 2025-07-21
fix: remove Temporary ETL Files on Write Error

anim001k:
## Description

This PR improves the reliability of the ETL collector by ensuring that temporary files are properly removed if an error occurs during file creation or writing.  
Now, if writing to a temp ETL file fails, the code attempts to delete the file to avoid leaving garbage on disk.

- Prevents orphaned temp files on disk
- Adds clear inline comments for maintainability
- No changes to public API

### DIFF
--- a/crates/etl/src/lib.rs
+++ b/crates/etl/src/lib.rs
@@ -131,30 +131,29 @@ where
     }
 
     fn flush(&mut self) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this isn't needed because the tempfile is only created in new and on drop this wipes

PR #17533, opened 2025-07-21
feat: `ComponentsFor` type alias

klkvr:
Adds a type alias and a default generic for `NodeAdapter` making it easier to use for custom node impls


mattsse (approved):
cool

PR #17534, opened 2025-07-21
fix: Use Vec for Block Hash Ancestors in WitnessDatabase

prestoalvarez:
## Description

This PR replaces the `BTreeMap<u64, B256>` with a `Vec<B256>` for storing block hash ancestors in `WitnessDatabase`.  
This change improves access speed and simplifies the code, as block numbers are expected to be contiguous.

### DIFF
--- a/crates/stateless/src/witness_db.rs
+++ b/crates/stateless/src/witness_db.rs
@@ -23,10 +23,9 @@ where
 {
     /// Map of block numbers to block hashes.
     /// This is used to service the `BLOCKHASH` opcode.
-    // TODO: use Vec instead -- ancestors should be contiguous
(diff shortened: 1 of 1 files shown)


crates/stateless/src/witness_db.rs
```diff
@@ -107,7 +106,7 @@ where
     /// Returns an error if the hash for the given block number is not found in the map.
     fn block_hash(&mut self, block_number: u64) -> Result<B256, Self::Error> {
         self.block_hashes_by_block_number
-            .get(&block_number)
+            .get(block_number as usize)
```
mattsse, line 109: this does not work

mattsse (commented):
this needs architectural changes instead, we cant easily do this and imo the map is fine

PR #17536, opened 2025-07-21
fix: Add Safe ok_or to WithPeerId<Option<T>> and Deprecate unwrap Methods

prestoalvarez:


mattsse (commented):
these were added for convenience, these are fine

PR #17538, opened 2025-07-21
docs(trie): update ParallelSparseTrie documentation

### LINKED ISSUE #17042: Add docs outlining the upper / lower parallel sparse trie structure
### Describe the feature

Currently we split the sparse trie in the parallel sparse trie across one upper trie, and 256 lower subtries. The roots of the lower subtries are contained in the lower subtries, and are not contained in the upper subtries. This means the upper subtrie can have at max 17 nodes, one for the root and up to 16 for its children. This should be documented on the parallel sparse trie struct, potentially with some diagrams.

It may also be beneficial to include some detailed comments on how we manage this when:
* adding leaves
* removing nodes
* updating hashes

### Additional context

_No response_

mediocregopher:
Now that PST is more stable we can fill out its documentation a bit.

Fixes #17042

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -64,15 +64,17 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
 ///
 /// ## Leaf Operations
 ///
-/// **Update**: When updating a leaf, the `prefix_set` is marked to track the change. The value is
-/// stored in the appropriate subtrie's values map. If the leaf is new, the trie structure is
-/// updated by walking to the leaf from the root, creating necessary intermediate branch nodes.
+/// **Update**: When updating a leaf, the new value is stored in the appropriate subtrie's values
+/// map. If the leaf is new, the trie structure is updated by walking to the leaf from the root,
+/// creating necessary intermediate branch nodes.
 ///
 /// **Removal**: Leaf removal may require parent node modifications. The algorithm walks up the
 /// trie, removing nodes that become empty and converting single-child branches to extensions.
 ///
 /// During leaf operations the overall structure of the trie may change, causing nodes to be moved
 /// from the upper to lower trie or vice-versa.
+///
+/// The `prefix_set` is modified during both leaf updates and removals to track changed leaf paths.
 ///
 /// ## Root Hash Calculation
 ///
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -32,6 +32,47 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
 
 /// A revealed sparse trie with subtries that can be updated in parallel.
 ///
+/// ## Structure
+///
+/// The trie is divided into two tiers for efficient parallel processing:
+/// - **Upper subtrie**: Contains nodes with paths shorter than `UPPER_TRIE_MAX_DEPTH` (2 nibbles)
+/// - **Lower subtries**: An array of 256 subtries, each handling nodes with paths of 2+ nibbles
+///
+/// Node placement is determined by path depth:
+/// - Paths with < 2 nibbles go to the upper subtrie
+/// - Paths with >= 2 nibbles go to lower subtries, indexed by their first 2 nibbles
+///
+/// ## Node Revealing
+///
+/// The trie uses lazy loading to efficiently handle large state tries. Nodes can be:
+/// - **Blind nodes**: Stored as hashes (`SparseNode::Hash`), representing unloaded trie parts
+/// - **Revealed nodes**: Fully loaded nodes (Branch, Extension, Leaf) with complete structure
+///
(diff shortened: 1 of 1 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,47 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
 
 /// A revealed sparse trie with subtries that can be updated in parallel.
 ///
+/// ## Structure
+///
+/// The trie is divided into two tiers for efficient parallel processing:
+/// - **Upper subtrie**: Contains nodes with paths shorter than `UPPER_TRIE_MAX_DEPTH` (2 nibbles)
```
shekhirin: can use rustdoc links to items, in other places too

```suggestion
/// - **Upper subtrie**: Contains nodes with paths shorter than [`UPPER_TRIE_MAX_DEPTH`] (2 nibbles)
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,47 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
...
+/// - **Lower subtries**: An array of 256 subtries, each handling nodes with paths of 2+ nibbles
+///
+/// Node placement is determined by path depth:
+/// - Paths with < 2 nibbles go to the upper subtrie
```
shekhirin: I wonder if we should use the `UPPER_TRIE_MAX_DEPTH` const here just so that we're consistent about either using it, or just `2`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,47 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
...
+/// - Paths with >= 2 nibbles go to lower subtries, indexed by their first 2 nibbles
```
shekhirin: we should also clarify what does the path of a lower subtrie mean, because it can be longer than just 2 nibbles

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,47 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
...
+///
+/// Note: An empty trie contains an `EmptyRoot` node at the root path, rather than no nodes at all.
+/// A trie with no nodes is blinded, its root may be `EmptyRoot` or some other node type.
+///
+/// Revealing is generally done using pre-loaded node data provided to via `reveal_nodes`. In
+/// certain cases, such as edge-cases when updating/removing leaves, nodes are revealed on-demand.
+///
+/// ## Leaf Operations
+///
+/// **Update**: When updating a leaf, the `prefix_set` is marked to track the change. The value
```
shekhirin: should mention that nodes can move from upper to lower subtries

crates/trie/sparse-parallel/src/trie.rs
mediocregopher: Modified to use UPPER_TRIE_MAX_DEPTH everywhere for consistency

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,58 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
 
 /// A revealed sparse trie with subtries that can be updated in parallel.
 ///
+/// ## Structure
+///
+/// The trie is divided into two tiers for efficient parallel processing:
+/// - **Upper subtrie**: Contains nodes with paths shorter than [`UPPER_TRIE_MAX_DEPTH`]
+/// - **Lower subtries**: An array of [`NUM_LOWER_SUBTRIES`] subtries, each handling nodes with
+///   paths of [`UPPER_TRIE_MAX_DEPTH`]+ nibbles
```
```suggestion
///   paths of at least [`UPPER_TRIE_MAX_DEPTH`] nibbles
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -32,6 +32,58 @@ pub const NUM_LOWER_SUBTRIES: usize = 16usize.pow(UPPER_TRIE_MAX_DEPTH as u32);
...
+///
+/// **Update**: When updating a leaf, the `prefix_set` is marked to track the change. The value is
+/// stored in the appropriate subtrie's values map. If the leaf is new, the trie structure is
+/// updated by walking to the leaf from the root, creating necessary intermediate branch nodes.
+///
+/// **Removal**: Leaf removal may require parent node modifications. The algorithm walks up the
+/// trie, removing nodes that become empty and converting single-child branches to extensions.
+///
+/// During leaf operations the overall structure of the trie may change, causing nodes to be moved
+/// from the upper to lower trie or vice-versa.
```
Rjected: let's move the comment about modifying the prefix set under `Leaf Operations`, and explain that the prefix set is meant to store changed paths, and will be modified for both updates and removals

Rjected (approved):
just nits, otherwise looks good

PR #17541, opened 2025-07-21
feat(reth-bench): auto-create output directory

shekhirin:
Automatically create the output directory if it doesn't exist when the --output argument is provided to reth-bench, improving user experience by eliminating the need to manually create directories before running benchmarks.


Rjected (approved):
lgtm, thanks claude

PR #17542, opened 2025-07-21
ci: mark system eest tests as passing

mattsse:
I believe these edge cases are now covered by revm directly

https://github.com/paradigmxyz/reth/actions/runs/16424596986/job/46413198786


klkvr (approved):

fgimenez (approved):
nice

PR #17548, opened 2025-07-22
feat(merkle): add IntermediateRootState for storage root progress

### LINKED ISSUE #17547: Implement incremental storage root calculation
### Describe the feature

The issue https://github.com/paradigmxyz/reth/issues/12527 turns out to be because the XEN contract causes a from-scratch storage root calculation to allocate ~100G.

This assumption from the original root code is now false due to how large XEN is:
https://github.com/paradigmxyz/reth/blob/39f1ee879517eae7d377ef5905c8bf517aab5110/crates/trie/trie/src/trie.rs#L195-L201

And the task described in the TODO is now relevant. This involves:
1. Adding helpers for restoring a storage root computation progress
2. Methods for stopping the storage root calculation after a threshold is met
3. Data structures for saving the storage root progress
4. Logic for stopping / resuming an incremental storage root calculation in StateRoot

### Additional context

_No response_

Rjected:
fixes https://github.com/paradigmxyz/reth/issues/17547

This introduces intermediate storage root state, to keep track of storage root progress. This is included in the `MerkleCheckpoint` so that we can save / reload this progress when we have many buffered storage trie updates. This allows us to limit memory usage in the merkle stage for arbitrarily large contracts while continuing to make progress calculating the storage root.

### DIFF
--- a/crates/stages/types/src/checkpoints.rs
+++ b/crates/stages/types/src/checkpoints.rs
@@ -53,7 +53,7 @@ impl reth_codecs::Compact for MerkleCheckpoint {
 
         len += self.state.to_compact(buf);
 
-        // Serialize the optional storage root checkpoint
+        // Encode the optional storage root checkpoint
         match &self.storage_root_checkpoint {
             Some(checkpoint) => {
                 // one means Some
@@ -88,9 +88,8 @@ impl reth_codecs::Compact for MerkleCheckpoint {
 
         let (state, mut buf) = HashBuilderState::from_compact(buf, 0);
 
-        // Deserialize the optional storage root checkpoint
+        // Decode the storage root checkpoint if it exists
         let (storage_root_checkpoint, buf) = if buf.is_empty() {
-            // For backward compatibility, if no more data, assume None
             (None, buf)
         } else {
             match buf.get_u8() {
@@ -168,7 +167,12 @@ impl reth_codecs::Compact for StorageRootMerkleCheckpoint {
         // Encode account fields
         buf.put_u64(self.account_nonce);
         len += 8;
+
+        let balance_len = self.account_balance.byte_len() as u8;
+        buf.put_u8(balance_len);
+        len += 1;
         len += self.account_balance.to_compact(buf);
+
         buf.put_slice(self.account_bytecode_hash.as_slice());
         len += 32;
 
@@ -193,7 +197,8 @@ impl reth_codecs::Compact for StorageRootMerkleCheckpoint {
 
         // Decode account fields
         let account_nonce = buf.get_u64();
-        let (account_balance, mut buf) = U256::from_compact(buf, 0);
+        let balance_len = buf.get_u8() as usize;
+        let (account_balance, mut buf) = U256::from_compact(buf, balance_len);
         let account_bytecode_hash = B256::from_slice(&buf[..32]);
         buf.advance(32);
 
@@ -545,6 +550,7 @@ stage_unit_checkpoints!(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloy_primitives::b256;
     use rand::Rng;
     use reth_codecs::Compact;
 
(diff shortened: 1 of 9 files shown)


Rjected:
This currently outputs correct storage roots, but we still get an incorrect state root for some reason and the code needs cleanup

Rjected:
The merkle stage on hoodi now passes with constant, low memory usage, with progress increasing linearly.

Here is the before, with stalls in progress being correlated with storage root calculation for large contracts and high memory usage:
Screenshot 2025-07-24 at 10 51 13 PM
Screenshot 2025-07-24 at 10 52 10 PM


Here is after this change:
Screenshot 2025-07-24 at 10 53 39 PM
Screenshot 2025-07-24 at 10 53 49 PM

Rjected:
This completed on base without issue, here is the merkle stage before the change, note over 100G of memory usage:
Screenshot 2025-07-24 at 11 37 55 PM
Screenshot 2025-07-24 at 11 38 06 PM

And here is the merkle stage after the change, memory is now consistently less than 150MB:
Screenshot 2025-07-25 at 1 07 23 AM
Screenshot 2025-07-25 at 1 07 34 AM

crates/stages/types/src/checkpoints.rs
```diff
@@ -68,8 +86,133 @@ impl reth_codecs::Compact for MerkleCheckpoint {
             buf = rest;
         }
 
-        let (state, buf) = HashBuilderState::from_compact(buf, 0);
-        (Self { target_block, last_account_key, walker_stack, state }, buf)
+        let (state, mut buf) = HashBuilderState::from_compact(buf, 0);
+
+        // Decode the storage root checkpoint if it exists
+        let (storage_root_checkpoint, buf) = if buf.is_empty() {
```
mattsse, line 92: I see, after decoding the HashBuilderState we still might have the checkpoint if the buf is non empty (backwards compat)

mattsse (commented):
all of this makes sense to me, but would appreciate a thorough review by @mediocregopher on the trie changes

lovely xen mitigation -.-

crates/trie/trie/src/progress.rs
```diff
@@ -1,34 +1,90 @@
...
 }
+
+/// The intermediate state of a state root computation, whether account or storage root.
+#[derive(Debug)]
+pub struct IntermediateRootState {
+    /// Previously constructed hash builder.
+    pub hash_builder: HashBuilder,
+    /// Previously recorded walker stack.
+    pub walker_stack: Vec<CursorSubNode>,
+    /// The last hashed key processed.
```
mediocregopher, line 78: nit: Could we include in the comment that for StorageRootProgress this corresponds to the `hashed_account` of the storage root? That wasn't immediately intuitive when I first saw it.

crates/trie/trie/src/trie.rs
```diff
@@ -154,34 +161,130 @@ where
...
+                        .get(&hashed_address)
+                        .cloned()
+                        .unwrap_or_default(),
+                    #[cfg(feature = "metrics")]
+                    self.metrics.storage_trie.clone(),
+                )
+                .with_intermediate_state(Some(storage_state.state))
+                .with_threshold(remaining_threshold);
+
+                match storage_root_calculator.calculate(retain_updates)? {
```
mediocregopher: Would it be possible to break this logic out into a closure like:
```
|storage_root_calulator| -> Option<StorageRootProgress> {
    match storage_root_calculator.calculate(retain_updates)? {
        ...
    }
}
```

It looks like it's essentially the same exact expression twice in two places, it's big enough to be worth deduplication imo. Maybe there's some difference between them I'm missing though...

mediocregopher (changes requested):
Mostly looks good to me, changes make sense and the code is clear. Just one note about deduplicating a big block of code if it's possible.

crates/trie/trie/src/progress.rs
Rjected: yeah, this would be the hashed storage key for storage tries, and the hashed account for the account trie, will add a comment

crates/trie/trie/src/trie.rs
Rjected: i ended up just putting the state required in a new struct called `StateRootContext`, and the result handling

mediocregopher: Love it :100:

mediocregopher (approved):

mattsse (approved):
lgtm

PR #17551, opened 2025-07-22
fix(exex): update batch threadshold calculate processed blocks

nkysg:
https://github.com/paradigmxyz/reth/blob/7b76a1e00fe2da5eb8cabd7d4e00bd9b9dfe6774/crates/exex/exex/src/backfill/job.rs#L125-L134

within BackfillJob::execute_range, the call to thresholds.is_end_of_batch uses block_number - *self.range.start() as the num_blocks argument. This calculates the number of processed blocks incorrectly (off by 1, as it's checked after processing the current block but subtracts without adding 1).

For max_blocks = Some(1) and range 1..=2:
Process block 1 (block_number = 1, check 1 - 1 = 0 >= 1? No.
Process block 2 (block_number = 2, check 2 - 1 = 1 >= 1? Yes, break.
Result: Processes 2 blocks despite max_blocks=1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,9 +97,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.6"
+version = "0.2.5"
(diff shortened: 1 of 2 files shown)


mattsse (approved):
ah, I see, this should indeed use +1

ty

PR #17557, opened 2025-07-22
ethereum: Fix hardcoded error type in Ethereum payload builder(NEW)

LandLord529:
This fixes the issue where all invalid transactions during payload building were incorrectly reported as "transaction type not supported" instead of showing the actual error.

Followup to https://github.com/paradigmxyz/reth/pull/17512 which got closed due to compilation issues. @mattsse mentioned the error was likely fine but we'd need to find a different way to convert the error type - so that's what this does.

The problem was trying to directly convert `Box<dyn InvalidTxError>` to `InvalidTransactionError`, which doesn't work. Instead I created a simple wrapper that implements `PoolTransactionError` and uses `InvalidPoolTransactionError::Other`.

Now when a transaction fails because of insufficient funds, nonce issues, etc., users will actually see the real error instead of getting confused by "tx type not supported".

@mattsse  - added some basic tests to make sure the wrapper preserves the original error info. Not sure if they belong here or if the approach is right, feel free to remove them if they don't fit.

**Before:**
```rust
InvalidPoolTransactionError::Consensus(InvalidTransactionError::TxTypeNotSupported)
```
**After:**  
```rust
InvalidPoolTransactionError::Other(Box::new(ExecutionError(error)))
```

Should make debugging a lot less frustrating.

### DIFF
--- a/crates/ethereum/payload/src/lib.rs
+++ b/crates/ethereum/payload/src/lib.rs
@@ -26,17 +26,43 @@ use reth_evm_ethereum::EthEvmConfig;
 use reth_payload_builder::{BlobSidecars, EthBuiltPayload, EthPayloadBuilderAttributes};
 use reth_payload_builder_primitives::PayloadBuilderError;
 use reth_payload_primitives::PayloadBuilderAttributes;
-use reth_primitives_traits::transaction::error::InvalidTransactionError;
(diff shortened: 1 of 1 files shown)


LandLord529:
hi @Rjected  sorry for teh tagging, This PR seems to have been shelved for a long time. would you mind taking a check? :)

PR #17559, opened 2025-07-22
refactor: move EngineValidator trait to reth-engine-tree

fgimenez:
towards #17413 

having `EngineValidator` in `reth-engine-tree` will allow to add methods that use `TreeCtx` (which in turn uses `EngineApiTreeState`) without cyclic dependencies


crates/engine/primitives/src/tree_state.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.trie_updates.as_deref()
+    }
+}
+
+/// Provides access to the tree state for payload validation.
+///
+/// This trait abstracts over the tree state operations needed by the payload validator,
+/// allowing lower-level crates to depend on this interface without creating circular
+/// dependencies with higher-level crates that contain the actual tree state implementation.
+pub trait TreeStateProvider<N: NodePrimitives>: Send + Sync {
```
klkvr: what would this be implemented on? unsure I understand what's the circular dep, can't validator impls depend on engine-tree directly?

fgimenez: the cyclic dep was happening when trying to use `EngineApiTreeState<N>` as the type of `state` in `TreeCtx`, `EngineApiTreeState` is defined in `reth-engine-tree` (very high in the dep tree), and this way we would be making `reth-engine-primitives` (low in the dep tree) depend on it. lmk if there's a different approach i'm missing

klkvr: engine-tree is the only consumer of the trait we're defining so I think we can just move the validator trait consumed by the tree to `reth-engine-tree`?

fgimenez: you mean moving `PayloadValidator` trait? in the current approach we are repurposing the existing `PayloadValidator`, adding the `validate_payload` and `validate_block` there in this PR https://github.com/paradigmxyz/reth/pull/17559, and this trait is not only used in engine-tree, we use it also in rpc https://github.com/paradigmxyz/reth/blob/0c7bd1e5dd59852e119b0ec6033d9f44888ae1c3/crates/rpc/rpc/src/validation.rs#L479

mattsse: I think the rpc usecase we can solve separately, independently of this

mattsse: I think we can actually get rid of this entirely because all we need is just:

https://github.com/paradigmxyz/reth/blob/53df3b803ad97fd692e2e0b79540556bffee8248/crates/rpc/rpc/src/validation.rs#L393-L393

klkvr: we already have `EngineValidator`/`PayloadValidator` separation so we could move `EngineValidator` to engine-tree and keep `PayloadValidator` for everything rpc-related

fgimenez: @mattsse @klkvr ok `EngineValidator` moved to `reth-engine-tree`, the rest of types are already in place after #17451 ptal

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -965,3 +971,47 @@ where
         Ok(input)
     }
 }
+
+/// Type that validates the payloads processed by the engine.
+pub trait EngineValidator<Types: PayloadTypes>:
+    PayloadValidator<ExecutionData = Types::ExecutionData>
+{
```
mattsse, line 978: I see, #17429 requires that we move this here to avoid the generic.

these layered traits are a bit weird, and I think these validation fns are only used in the engineapi server, but I guess it's easier to move them here for now? @klkvr otherwise we'd need to redesign the trait and all the builder types first

mattsse, line 978: I see, #17429 requires that we move this here to avoid the generic.

these layered traits are a bit weird, and I think these validation fns are only used in the engineapi server, but I guess it's easier to move them here for now? @klkvr otherwise we'd need to redesign the trait and all the builder types first

mattsse, line 978: I see, #17429 requires that we move this here to avoid the generic.

these layered traits are a bit weird, and I think these validation fns are only used in the engineapi server, but I guess it's easier to move them here for now? @klkvr otherwise we'd need to redesign the trait and all the builder types first

mattsse (approved):
this seems okay, at least for the time being, but imo we should revisit this entire trait once we're done with the integration

pending @klkvr

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: > I see, #17429 requires that we move this here to avoid the generic.

exactly, allows us to use `EngineApiTreeState<N: NodePrimitives>` (without cyclic deps) instead of the generic `State`

klkvr (approved):
makes sense for now, but we should clean this up at some point, and remove unnecesarry tree dep from rpc crates

PR #17560, opened 2025-07-22
test(sdk): Add test for using node builder with noop components

emhane:
Fixes `NoopNetwork` impl to make it possible to disable any component in node builder by using a noop component. Adds test to verify the node builds with all noop components.


mattsse (approved):
makes sense, ty

PR #17564, opened 2025-07-22
feat(storage): add trait fn and impl for block_body

Dyslex7c:
Closes #17522

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -1165,6 +1165,7 @@ impl<TX: DbTx + 'static, N: NodeTypes> BlockNumReader for DatabaseProvider<TX, N
 
 impl<TX: DbTx + 'static, N: NodeTypesForProvider> BlockReader for DatabaseProvider<TX, N> {
     type Block = BlockTy<N>;
+    type BlockBody = BodyTy<N>;
 
     fn find_block_by_hash(
         &self,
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1165,6 +1165,7 @@ impl<TX: DbTx + 'static, N: NodeTypes> BlockNumReader for DatabaseProvider<TX, N
 
 impl<TX: DbTx + 'static, N: NodeTypesForProvider> BlockReader for DatabaseProvider<TX, N> {
     type Block = BlockTy<N>;
+    type BlockBody = BodyTy<N>;
```
mattsse, line 1168: this type can be obtained from the Block type directly because this satisfies trait Block

https://github.com/paradigmxyz/reth/blob/1effb1952c58acb7435fb4ea2257e47ed868ab83/crates/storage/storage-api/src/block.rs#L59-L63

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1207,6 +1208,35 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> BlockReader for DatabaseProvid
...
+                let Some(transactions) = self.transactions_by_block(number.into())? else {
+                    return Ok(None)
+                };
+
+                // We need a header to pass to read_block_bodies, but we only care about the body
+                if let Some(header) = self.header_by_number(number)? {
+                    let body = self
+                        .storage
+                        .reader()
+                        .read_block_bodies(self, vec![(&header, transactions)])?
```
mattsse, line 1228: oh I totally forgot that we actually need the header as input here... sigh

then this kinda defeats the purpose, I cant think of a workaround here.

hmm, then this feature no longer makes sense unfortunately...

very sorry about that

mattsse (commented):
I was wrong in my initial assumption that this would lead to more performant IO, if we always need to fetch the header, which we need, then this doesnt make sense because then this is always as expensive as reading the entire block...

sorry about that

PR #17567, opened 2025-07-22
fix: Added 0x prefix

anim001k:


mattsse (commented):
this is just test code and the hash has no significance so we can keep it asa is

PR #17568, opened 2025-07-22
fix: Update alloy-evm to 0.15

bharath-123:
The EIP-2395 blockhashes contract system calls fail in the `fusaka-devnet-3` branch as the calls have a gas limit of 30M. 

https://github.com/paradigmxyz/reth/commit/5b45d7bfe6c0947af1c2ce1147645e4ee1ef7789 updated the fusaka-devnet-3 to use the latest revm fusaka-devnet-3 branch. But it didn't update `alloy-evm` to 0.15 which includes the updates done in the latest fusaka-devnet-3 revm branch.

Some clones are introduced since `TracingCtx` struct lost the Copy trait https://github.com/alloy-rs/evm/blob/main/crates/evm/src/tracing.rs#L20

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6056,7 +6056,7 @@ dependencies = [
 [[package]]
 name = "op-revm"
 version = "8.0.3"
-source = "git+https://github.com/bluealloy/revm?branch=fusaka-devnet-3#7e98625d1391b11b0f5fb2fa88dae2735959994b"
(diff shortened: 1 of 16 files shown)


mattsse (commented):
@rkrasiuk should we instead rebase this entire branch?

rkrasiuk:
the devnet 3 branch has been rebased on top of main

Issue #17565, opened 2025-07-22
E2E support for custom headers


klkvr:
### Describe the feature

Right now custom headers are properly propagated all the way down to the EVM, allowing to have header fields affecting execution via `BlockExecutionCtx`.

However, existing payload builder implementations are not able to build blocks with custom headers as this would require supporting custom payload attributes and custom `Evm::NextBlockEnvCtx` which are usually locked to hardcoded types:
https://github.com/paradigmxyz/reth/blob/39f1ee879517eae7d377ef5905c8bf517aab5110/crates/optimism/payload/src/builder.rs#L209

What we could do is either keep this as is given that payload builder implementations are usually custom, or introduce a trait that would specify payload attributes -> NextBlockEnvCtx conversion, allowing to propagate any custom fields to EVM. This is a bit tricky because this conversion might rely on some additional context like chainspec

nadtech-hub:
I'd like to make it

nadtech-hub:
Why was I unassigned?

PR #17570, opened 2025-07-22
feat: abstraction for attributes -> NextBlockEnv conversion

klkvr:
towards #17565 

Introduces `BuildNextEnv` trait that should be implemented on `NextBlockEnvCtx` and defines how to construct it from payload attributes, parent header and arbitrary ctx.

Integrates it into `OpPayloadBuilder` in a non-breaking way

### DIFF
--- a/crates/optimism/payload/src/builder.rs
+++ b/crates/optimism/payload/src/builder.rs
@@ -3,14 +3,13 @@
 use crate::{
     config::{OpBuilderConfig, OpDAConfig},
     error::OpPayloadBuilderError,
-    payload::{OpBuiltPayload, OpPayloadBuilderAttributes},
-    OpPayloadPrimitives,
+    payload::OpBuiltPayload,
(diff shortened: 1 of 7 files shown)


crates/optimism/payload/src/builder.rs
```diff
@@ -42,12 +40,18 @@ use reth_revm::{
...
 /// Optimism's payload builder
-#[derive(Debug, Clone)]
-pub struct OpPayloadBuilder<Pool, Client, Evm, Txs = ()> {
+#[derive(Debug)]
+pub struct OpPayloadBuilder<
+    Pool,
+    Client,
+    Evm,
+    Txs = (),
+    Attrs = OpPayloadBuilderAttributes<TxTy<<Evm as ConfigureEvm>::Primitives>>,
```
mattsse, line 53: these are pretty self explanatorily

mattsse (approved):
makes sense, but as discussioned, ideally we find a way to simplify

PR #17573, opened 2025-07-22
feat: convert BlockExecutionErrors

### LINKED ISSUE #17356: BlockExecutionError returns as internal server error in rpc
### Describe the feature

due to: https://github.com/paradigmxyz/reth/blob/b3d722f1fdc56de2354248dfd9a0e82f03941586/crates/rpc/rpc-eth-types/src/error/mod.rs#L367-L371

we currently emit this as internal error as well:

https://github.com/alloy-rs/evm/blob/6ba140eb91bae718b50fd6b8944ffeef5ac28c5b/crates/evm/src/block/error.rs#L10-L10

> "ErrorObject { code: InternalError, message: \"EVM reported invalid transaction (0x8aee37f94206fba821c7b6e44add8ba20982f5538f3cb6f9f2ec93bdfeafe4df): lack of funds (0) for max fee (255)\", data: None }"

this is a bit tricky because of the dyn InvalidTxerr we can't perform this conversion:

https://github.com/paradigmxyz/reth/blob/b3d722f1fdc56de2354248dfd9a0e82f03941586/crates/rpc/rpc-eth-types/src/error/mod.rs#L657-L657

### Additional context

_No response_

stevencartavia:
closes #17356


crates/rpc/rpc-eth-types/src/error/mod.rs
```diff
@@ -371,7 +371,22 @@ impl From<RethError> for EthApiError {
...
     fn from(error: BlockExecutionError) -> Self {
-        Self::Internal(error.into())
+        match error {
+            BlockExecutionError::Validation(validation_error) => match validation_error {
+                BlockValidationError::InvalidTx { error, .. } => Self::InvalidTransaction(
+                    RpcInvalidTransactionError::other(rpc_error_with_code(
+                        EthRpcErrorCode::TransactionRejected.code(),
+                        error.to_string(),
+                    )),
+                ),
```
mattsse: this might not be sufficient because we have different error codes for this error

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/rpc/rpc-eth-types/src/error/mod.rs#L603-L617

and ideally we want to do through this conversion:

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/rpc/rpc-eth-types/src/error/mod.rs#L665-L665

so what we can here is try to cast the dyn error to InvalidTransaction

we could even add this as default to the trait 

```
fn as_invalid_tx_err(&self) -> Option<&InvalidTransaction> {
  // try cast
}
```

then this would also work for optimism

mattsse (changes requested):

crates/rpc/rpc-eth-types/src/error/mod.rs
```diff
@@ -23,9 +23,22 @@ use revm::context_interface::result::{
...
+    /// Try to cast the dyn `InvalidTxError` to `InvalidTransaction`
+    fn as_invalid_tx_err(&self) -> Option<&InvalidTransaction>;
+}
+
+impl InvalidTxErrorExt for dyn InvalidTxError {
+    fn as_invalid_tx_err(&self) -> Option<&InvalidTransaction> {
+        let any_ref: &dyn Any = self;
+        any_ref.downcast_ref::<InvalidTransaction>()
+    }
+}
```
mattsse: this doesnt really belong here and we should rather add this to the existing InvalidTxError trait in alloy-evm

because the way this behaves rn is a bit pointless since its impled on the dyn

mattsse (changes requested):

mattsse (approved):

PR #17575, opened 2025-07-23
fix: correct prune mode assignments in HistoryIndexingStages

zeroprooff:
Fix incorrect prune mode assignments in the HistoryIndexingStages implementation. The IndexStorageHistoryStage was incorrectly receiving account_history prune mode, and IndexAccountHistoryStage was incorrectly receiving storage_history prune mode. This change ensures each stage receives its corresponding prune mode configuration


mattsse (approved):
oh nice, yeah the defaults should be the same for those but this is a right fix

PR #17576, opened 2025-07-23
feat: make CompactEnvelope trait public for external crate usage

rezzmah:
Makes CompactEnvelope and related traits public to enable external crates to implement efficient transaction serialization following Reth patterns. This change allows chain-specific implementations like Berachain (https://github.com/berachain/bera-reth/pull/50) to leverage the standardized CompactEnvelope system for database storage compatibility while maintaining backwards compatability


klkvr (approved):
makes sense, right now we have a blanket impl for this trait but no way to actually use it

PR #17578, opened 2025-07-23
refactor: reuse evm tracing utils for trace call many

ongyimeng:
Optimize trace_call_many by reusing EVM tracer instance

### Description

This PR optimizes the `trace_call_many` RPC method by reusing a single EVM tracer instance across multiple calls, rather than creating a new instance for each call. This addresses the performance issue identified in #17484.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -466,15 +466,15 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> {
     where
         Self: LoadPendingBlock + Call,
         F: Fn(
-            usize, // call index
-            TracingCtx<
-                '_,
-                TxEnvFor<Self::Evm>,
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -446,4 +446,70 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> {
...
+                    .create_tracer(
+                        StateCacheDbRefMutWrapper(&mut db),
+                        evm_env,
+                        inspector_setup()
+                    );
+
+                let mut call_idx = 0;
+                let results = tracer
+                    .try_trace_many(
+                        tx_envs.iter().cloned(),
```
mattsse: I think we can use into_iter here?

mattsse (changes requested):
great start,

what's missing to get ci green is a few clippy things

mattsse (commented):
okay, I looked at this again,

this is pretty much complete but there are a few more blockers we need to resolve first before we can merge this

PR #17582, opened 2025-07-23
chore: move validation to standalone fns

mattsse:
this makes the validation fns reusable without the container helper type


fgimenez (approved):
awesome thanks!

PR #17583, opened 2025-07-23
feat: add Eth EngineValidator impl

fgimenez:
towards #17413


crates/ethereum/payload/src/tree_validator.rs
```diff
@@ -0,0 +1,205 @@
...
+    + StateProviderFactory
+    + StateReader
+    + StateCommitmentProvider
+    + HashedPostStateProvider
+    + HeaderProvider<Header = <EthPrimitives as reth_primitives_traits::NodePrimitives>::BlockHeader>
+    + Clone
+    + Unpin
+    + 'static
+{
+}
```
mattsse, line 41: this is just for convenience?

crates/ethereum/payload/src/tree_validator.rs
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Debug)]
+pub struct EthPayloadValidator<P, C, Spec>
+where
+    P: EthProvider,
+    C: ConfigureEvm<Primitives = EthPrimitives> + 'static,
+{
+    /// Reusable validation logic provider
+    tree_validator: TreePayloadValidator<EthPrimitives, P, C>,
+    /// Chain spec for hardfork checks
+    chain_spec: Arc<Spec>,
```
mattsse, line 71: I think this is a good base, we could even get rid of the chainspec field if we require that P is also chainspec provider, at the cost of an arc clone which should be negligible

PR #17585, opened 2025-07-23
feat: add Op EngineValidator impl

fgimenez:
towards #17413


crates/optimism/payload/src/tree_validator.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    fn validate_block_post_execution_with_hashed_state(
+        &self,
+        _state_updates: &HashedPostState,
+        _block: &RecoveredBlock<Self::Block>,
+    ) -> Result<(), ConsensusError> {
+        // Default implementation - no additional validation needed for Optimism
+        Ok(())
+    }
```
mattsse, line 141: I think this is missing

https://github.com/paradigmxyz/reth/blob/2c5a967898aeb6018d3f3a9a67a206b192289b7a/crates/optimism/node/src/engine.rs#L133-L137

but I guess now that we have control over execution+validation we could also retire this function

crates/optimism/payload/src/tree_validator.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+impl<P, C, Spec> EngineValidator<OpPayloadTypes> for OpPayloadValidator<P, C, Spec>
```
mattsse, line 144: it might make sense to split off this trait entirely but we can think about this next

crates/optimism/payload/src/tree_validator.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl<P, C, Spec> PayloadValidator for OpPayloadValidator<P, C, Spec>
+where
+    P: OpProvider,
+    C: ConfigureEvm<Primitives = OpPrimitives> + 'static,
+    Spec: OpHardforks + EthereumHardforks + Send + Sync + 'static,
+{
+    type Block = OpBlock;
```
mattsse, line 115: this is more restrictive than

https://github.com/paradigmxyz/reth/blob/2c5a967898aeb6018d3f3a9a67a206b192289b7a/crates/optimism/node/src/engine.rs#L121-L121

but this should also ago away if we restructure this

PR #17588, opened 2025-07-23
fix: typo in Cargo.toml

sashaodessa:
**Description:**  
Corrected a minor typo in the comment of `crates/ethereum/node/Cargo.toml`, changing `poperly` to `properly`


mattsse (changes requested):
there's one more case of this

https://github.com/paradigmxyz/reth/blob/bf36f9521125d455d2c6d7bbe512f8d12c93c80c/crates/optimism/node/Cargo.toml#L49-L49

lets fix that as well?

mattsse (approved):

PR #17592, opened 2025-07-23
test(era1): add more `Receipt` tests to verify decoding

lean-apple:
- Add more era1 tests for for receipts (reth-ethereum-primitives and alloy's one), with the decode part in `Vec<RecordWithBloom>`
- Add examples in `execution_types` code docs

### DIFF
--- a/crates/era/src/era1_types.rs
+++ b/crates/era/src/era1_types.rs
@@ -163,11 +163,11 @@ mod tests {
         test_utils::{create_sample_block, create_test_block_with_compressed_data},
         DecodeCompressed,
     };
+    use alloy_consensus::ReceiptWithBloom;
     use alloy_primitives::{B256, U256};
-    use reth_ethereum_primitives::Receipt;
 
     #[test]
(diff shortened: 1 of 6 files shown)


mattsse (approved):
I see these are all tests, anything we actually need to fix wrt import/export?

lean-apple:
> I see these are all tests, anything we actually need to fix wrt import/export?

When we import from the files, we only insert headers and bodies into the db, https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/era-utils/src/history.rs#L317 and https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/era-utils/src/history.rs#L320C38-L321C1.

and so we only need to decode these : 
https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/era-utils/src/history.rs#L244-L255.

Regarding the export to era1 files, we encode and compress so it was ok.

PR #17595, opened 2025-07-24
feat: add bsc Parlia support

clydemeng:
# BSC Parlia-support : snapshots, tolerant RLP, test helpers, **Maxwell timestamp rule**

This PR brings in the first generic‐Reth changes required by the BNB Smart Chain (Parlia) integration that lives in `loocapro_reth_bsc`.  
All changes are **opt-in**; Ethereum behaviour remains untouched.

---

## Motivation

Parlia deviates from Ethereum in four visible areas:

1. **Maxwell hard-fork:** block interval drops to 0.75 s, so consecutive
   blocks can share the *same* `timestamp`.  
2. Block propagation (`NewBlockHashes`, `NewBlock`) carries extra RLP
   fields (timestamp, proposer, side-cars).  
3. The consensus layer stores a compressed “checkpoint snapshot” every
   1 024 blocks.  
4. E2E tests need new helpers to spin up a payload builder, inject
   Parlia-style `extraData`, and broadcast the extended messages.

---

### 1  Consensus – **Maxwell timestamp relaxation**
* `consensus/common/src/validation.rs`
  ```rust
  // reject only if child.timestamp < parent.timestamp
  // (== is now allowed due to 0.75 s Maxwell interval)
  ```
  The classic “strictly greater” check became “not less than” so Parlia
  blocks that arrive within the same second no longer fail validation.
  This affects **only** chains that feed such headers; Ethereum blocks
  still satisfy `timestamp > parent.timestamp` naturally.

### 2  Networking (`eth-wire-types`)
* Manual, **tolerant RLP decoders** for
  `NewBlockHashes` & `NewBlock` that ignore trailing BSC-specific
  fields.

### 3  Execution (`evm/execution-types`)
* `snapshot_pool.rs` – global accumulator for Parlia checkpoint blobs.
* `ExecutionOutcome` gains `snapshots: Vec<(BlockNumber, Vec<u8>)>`.

### 4  Storage (`db-api`)
* New `ParliaSnapshots` table to persist the checkpoint snapshots.

### 5  Node builder
* `ParliaBuilder` trait to inject Parlia consensus into the generic node
  builder.

### 6  E2E test utils
* Retry logic, “wait-for-payload”, and lightweight network/RPC
  shims to exercise the new flows.

### 7  Cargo changes
```toml
once_cell = { version = "1.19", default-features = false, features = ["std"] }
```
added to `reth-execution-types` for the snapshot pool.

---

## Backwards compatibility

* Ethereum peers see **identical** RLP output; only the *decoder* got
  looser.
* `ExecutionOutcome` introduces a new field with a default value—no
  caller breakage.
* The new DB table is inert unless the BSC executor pushes snapshots.
* Timestamp rule change is purely *relaxed*; it cannot invalidate
  Ethereum blocks.

---

## Tests & validation

* Unit-tests for tolerant decoders and snapshot SER/DE.
* Regression test covering `timestamp == parent.timestamp` passes.
* `cargo test --all --all-features` green on Linux/macOS.

---

## Follow-ups / out of scope
* Snapshot production wired only from the BSC executor (downstream).  
* Seal / attestation verification to follow.  
* CLI flags & metrics for snapshots still TODO.

---

Thanks for reviewing!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7508,6 +7508,7 @@ dependencies = [
  "reth-consensus",
  "reth-ethereum-primitives",
  "reth-primitives-traits",
+ "tracing",
 ]
 
 [[package]]
@@ -8405,6 +8406,7 @@ dependencies = [
  "arbitrary",
  "bincode 1.3.3",
  "derive_more",
+ "once_cell",
  "rand 0.9.1",
  "reth-ethereum-primitives",
  "reth-primitives-traits",
--- a/crates/consensus/common/Cargo.toml
+++ b/crates/consensus/common/Cargo.toml
@@ -11,6 +11,8 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
+tracing.workspace = true
+
 # reth
 reth-chainspec.workspace = true
 reth-consensus.workspace = true
--- a/crates/consensus/common/src/validation.rs
+++ b/crates/consensus/common/src/validation.rs
@@ -304,7 +304,7 @@ pub fn validate_against_parent_timestamp<H: BlockHeader>(
     parent: &H,
 ) -> Result<(), ConsensusError> {
     // As BSC Maxwell hardfork reduces block interval to 0.75s,
-    // assuming header.timestamp() <= parent.timestamp() is no longer valid.
+    // assuming header.timestamp() strictly larger than parent.timestamp() is no longer valid.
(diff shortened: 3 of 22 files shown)


clydemeng:
This PR is stale as most of changes can be  moved to  loocapro/reth-bsc

Issue #17291, opened 2025-07-08
Start executing transactions ASAP on `engine_newPayload`


shekhirin:
### Describe the feature

Blocked by https://github.com/paradigmxyz/reth/issues/17413

Profile: https://share.firefox.dev/4lLWTJq

In this Mainnet block, we spend:
- 17ms on payload to block conversion (validations, transaction decoding, sender recovery), trie input creation, multiproof config creation <img width="2552" height="772" alt="Image" src="https://github.com/user-attachments/assets/fbd9a5ae-8bfc-46bc-ba2e-24ea4ad95bd6" />
- 27ms on the actual EVM execution <img width="2555" height="775" alt="Image" src="https://github.com/user-attachments/assets/18ff6525-b0e2-4618-be63-fe3effc2e8ec" />
- 27ms on waiting for state root  <img width="2554" height="776" alt="Image" src="https://github.com/user-attachments/assets/26c7a8f2-334c-4549-b669-395cd646883a" />

What we should do instead, is to start executing (and prewarming) the transactions at T0, meaning that:
- For every transaction decoded and its sender recovered, we should send it to execution (prewarming & actual execution) 
- Block validation should happen at the same time as we execute
- Trie input and multiproof config creation should happen in trie threads

For a payload that takes 70ms, doing this can save us at least 25% of time, because we start the execution ASAP. Potentially, it can be more, because we also start prewarming sooner.

PR #17598, opened 2025-07-24
refactor: introduce Enginvalidator in tree

klkvr:
Towards #17291 

Refactors `EngineApiTreeHandler` so that for happy path `ensure_well_formed_payload` call is happening right before the block execution, allowing to abstract it away along with execution logic as a single call


mattsse (approved):
checks out, this makes sense and is a prerequisite for logic changes that we'd need to do anyways

crates/node/builder/src/launch/engine.rs
```diff
@@ -207,6 +207,15 @@ where
...
             .maybe_store_messages(node_config.debug.engine_api_store.clone());
 
+        let engine_validator = BasicEngineValidator::new(
+            ctx.blockchain_db().clone(),
+            consensus.clone(),
+            ctx.components().evm_config().clone(),
+            engine_payload_validator,
+            engine_tree_config.clone(),
+            ctx.invalid_block_hook().await?,
+        );
```
mattsse, line 217: while this locks in the basicengineValidator as the only supported implementation, I think this is still fine for now

examples/custom-engine-types/src/main.rs
```diff
@@ -239,15 +238,6 @@ where
...
     }
-
-    fn validate_payload_attributes_against_header(
-        &self,
-        _attr: &<T as PayloadTypes>::PayloadAttributes,
-        _header: &<Self::Block as reth_ethereum::primitives::Block>::Header,
-    ) -> Result<(), InvalidPayloadAttributesError> {
-        // skip default timestamp validation
-        Ok(())
-    }
```
mattsse, line 250: this is now a bit problematic, because we can no longer easily configure this because this is baked into the `BasicEngineValidator` and there's no way to override this network specific behaviour anymore, which would break a few use cases where this check must be disabled, e.g. non increasing timestamps.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -160,7 +104,7 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
 /// used by network-specific payload validators (e.g., Ethereum, Optimism). It is not meant to be
 /// used as a standalone component, but rather as a building block for concrete implementations.
 #[derive(derive_more::Debug)]
-pub struct TreePayloadValidator<P, Evm>
+pub struct BasicEngineValidator<P, Evm, V>
```
mattsse, line 139: one limitation of this approach long term is that we're baking in all of these internals as the defacto standard, and any network specific optimizations we could do (op vs eth) now must be expressed as part of a trait that we then need to enforce as part of the launcher.

In order to do the optimizations we have in mind we'd either need to extend the PayloadValidator and/or the payload type.

mattsse (changes requested):
this is definitely an improvement,
but this currently removes some features that users rely on which is configuring the payloadattributes vs parent.timestamp rule

which we could perhaps solve differently entirely.

My main worry here is that enshrining the Basic type isnt flexible enough moving forward.

mediocregopher:
Question, mostly because this area of the code still goes over my head quite a bit: At what point exactly is the payload validation concurrency going to happen in the future? My guess is within `insert_block_or_payload`, where `execute` and `convert_to_block` can be made to happen simultaneously, but not 100% I'm understanding correctly.

mattsse (approved):
lgtm,

this should unblock us from doing more optimizations and cleanup

PR #17600, opened 2025-07-24
feat: tracked  State for local pending block

### LINKED ISSUE #17584: Track State for local pending block
### Describe the feature

we currently only track block + receipts

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/rpc/rpc-eth-types/src/pending_block.rs#L76-L85

but we should also track the state at this block

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/chain-state/src/in_memory.rs#L754-L763

if we want to support flashblocks then we also need to store
* parent canonical hash
* a Vec of the `ExecutedBlockWithTrieUpdates` leading up to that block, so we can obtain the state via

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/chain-state/src/memory_overlay.rs#L36-L49

## TODO
* add additional context to struct Pendingblock

for this we also need to change the return type here:

https://github.com/paradigmxyz/reth/blob/f2f3af43d4f4a3ad8ab03ff6a3ab13cc3a33d906/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L185-L196

and if this needs to support multiple sub blocks, we need to provide the flashblock state as input or the parent txs

### Additional context

_No response_

Rimeeeeee:
fixes #17584

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
@@ -341,7 +341,7 @@ pub trait LoadPendingBlock:
         id: &BlockId,
     ) -> impl std::future::Future<Output = Result<StateProviderBox, Self::Error>> + Send {
         async move {
-            if let BlockId::Number(BlockNumberOrTag::Pending) = id {
+            if matches!(id, BlockId::Number(BlockNumberOrTag::Pending)) {
                 let pending_block = self.pending_block().lock().await;
                 let block = pending_block.as_ref().ok_or_else(|| {
                     Self::Error::from_eth_err(EthApiError::HeaderNotFound(
--- a/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
@@ -4,7 +4,7 @@
 use super::SpawnBlocking;
 use crate::{EthApiTypes, FromEthApiError, FromEvmError, RpcNodeCore};
 use alloy_consensus::{BlockHeader, Transaction};
-use alloy_eips::eip7840::BlobParams;
+use alloy_eips::{eip7840::BlobParams, BlockId};
 use alloy_primitives::{B256, U256};
 use alloy_rpc_types_eth::BlockNumberOrTag;
 use futures::Future;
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -322,10 +322,47 @@ pub trait LoadPendingBlock:
...
+                ..Default::default()
+            }),
+            hashed_state: Arc::new(hashed_state),
+        })
+    }
+
+    /// Returns the pending state for the given block id.
+    fn get_pending_state(
+        &self,
+        id: &BlockId,
```
mattsse: here we don't need the block id

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -322,10 +322,47 @@ pub trait LoadPendingBlock:
...
+    ) -> impl std::future::Future<Output = Result<StateProviderBox, Self::Error>> + Send {
+        async move {
```
mattsse: what this should do is, is first obtain the latest state

and then use 

https://github.com/paradigmxyz/reth/blob/876e964cbcbd85ade6fdec40bb35f08690332854/crates/chain-state/src/memory_overlay.rs#L36-L46

if we have a pending block, so that the pending state extends the latest state

mattsse (changes requested):

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -322,10 +324,53 @@ pub trait LoadPendingBlock:
...
+                .map_err(Self::Error::from_eth_err)?;
+
+            let pending_block = self.pending_block().lock().await;
+            if let Some(_pending_block) = pending_block.as_ref() {
+                let ex_block = self.build_block(&latest_header);
+                let overlay = MemoryOverlayStateProviderRef::new(
+                    historical_state,
+                    vec![ExecutedBlockWithTrieUpdates {
+                        block: ex_block.unwrap(),
+                        trie: ExecutedTrieUpdates::Missing,
```
Rimeeeeee: To be replaced with actual ExecutedTrieUpdates(TODO)

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -322,10 +324,53 @@ pub trait LoadPendingBlock:
...
+            execution_output: Arc::new(ExecutionOutcome {
+                receipts: vec![execution_result.receipts],
+                ..Default::default()
+            }),
+            hashed_state: Arc::new(hashed_state),
+        })
+    }
+
+    /// Returns the pending state by combining with latest state.
+    fn get_pending_state(
```
mattsse: I excluded this fn from this pr because I realized we need to rethink this a bit

mattsse (approved):

PR #17601, opened 2025-07-24
docs(example): extract full contract state from db

sergey-melnychuk:


Cargo.toml
```diff
@@ -169,6 +169,7 @@ members = [
     "testing/ef-tests/",
     "testing/testing-utils",
     "crates/tracing-otlp",
+    "examples/full-contract-state",
```
mattsse: this is out of order here

examples/full-contract-state/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "full-contract-state"
```
mattsse:
```suggestion
name = "example-full-contract-state"
```

examples/full-contract-state/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+# Ethereum node and provider types
+reth-ethereum = { workspace = true, features = ["node"] }
+
+# Storage and database APIs
+reth-storage-api.workspace = true
+reth-db-api.workspace = true
+reth-storage-errors.workspace = true
+
+# Primitive traits
+reth-primitives-traits.workspace = true
```
mattsse: I think all of those should be available via reth-ethereum, if not we need to re-export

mattsse (changes requested):
cool, this seems useful, 

left some pedantic nits

mattsse (approved):

PR #17603, opened 2025-07-24
feat: integrate `EngineValidator` into `EngineApiTreeHandler`

klkvr:
Based on https://github.com/paradigmxyz/reth/pull/17598

Changes in the PR:
1. `TreePayloadValidator` is renamed to `BasicEngineValidator` and made generic over `PayloadValidator`
2. Logic in `TreePayloadValidator` is changed to be almost 1:1 identical to current logic in `insert_block_inner` 
3. `insert_block_inner` is changed to call `EngineValidator` implementation


crates/engine/service/src/service.rs
```diff
@@ -87,7 +87,7 @@ where
         evm_config: C,
     ) -> Self
     where
-        V: EngineValidator<N::Payload, Block = BlockTy<N>>,
+        V: EngineValidator<N::Payload>,
```
mattsse, line 90: do we even need EngineValidator here or is PayloadValidator sufficient?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -160,7 +104,7 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
 /// used by network-specific payload validators (e.g., Ethereum, Optimism). It is not meant to be
 /// used as a standalone component, but rather as a building block for concrete implementations.
 #[derive(derive_more::Debug)]
-pub struct TreePayloadValidator<P, Evm>
+pub struct BasicEngineValidator<P, Evm, V>
```
mattsse, line 107: ah Inverting this makes a lot of sense actually

mattsse (commented):
need to take another closer pass at this but this is looking great

crates/engine/service/src/service.rs
klkvr: we now have 2 EngineValidators, that one doesn't have any supertraits and only used by the tree

another one we should rename to EngineApiValidator eventually

mattsse (approved):
great,

PR #17607, opened 2025-07-25
feat: added db max_readers arg

### LINKED ISSUE #17597: Add max_readers option to DatabaseArguments
### Describe the feature

we currently do: https://github.com/paradigmxyz/reth/blob/876e964cbcbd85ade6fdec40bb35f08690332854/crates/storage/db/src/implementation/mdbx/mod.rs#L377-L378

but ideally this setting should come from

https://github.com/paradigmxyz/reth/blob/876e964cbcbd85ade6fdec40bb35f08690332854/crates/storage/db/src/implementation/mdbx/mod.rs#L72-L72

## TODO
* add max_readers: Option to `DatabaseArgs`
* use this to configure the max readers in the setup instead

### Additional context

_No response_

Soubhik-10:
closes #17597


Soubhik-10:
Will be opening one with fixed book

Issue #17605, opened 2025-07-25
Allow a custom Tokio runtime for the RPC server


hai-rise:
`jsonrpsee` allows us to spawn RPC tasks onto a custom runtime:
https://docs.rs/jsonrpsee-server/latest/jsonrpsee_server/struct.ServerConfigBuilder.html#method.custom_tokio_runtime

However, Reth nodes don't configure this, so always set `tokio_runtime` to `None` (default).
https://github.com/paradigmxyz/reth/blob/876e964cbcbd85ade6fdec40bb35f08690332854/crates/rpc/rpc-builder/src/config.rs#L163-L194

It would be nice to be able to configure this. Worst case, it's an extra feature for Reth node builders; best case, it unlocks more improvements and experiments for #17604.

mattsse:
not the RpcServerArgs because these are CLI focused

we can do this on this type first

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc-builder/src/lib.rs#L1010-L1010

and then we figure out how to propagate this setting up the stack

mattsse:
@mablr do you still plan on doing the followup for #17611?

mattsse:
the use would need to configure this here so for rpcaddons we can also add a handle field like we did for #17611 and then we use that when we configure the actual server, basically bubbling up the setting

PR #17611, opened 2025-07-25
feat(rpc): add method to configure custom tokio runtime for RPC server

mablr:
towards #17605 by adding `with_tokio_runtime()` method to `RpcServerConfig` allowing to configure a custom Tokio runtime for RPC servers (any kind of transport) instead of using the default runtime.

Follow-up: propagate this setting up the stack

### DIFF
--- a/crates/rpc/rpc-builder/src/lib.rs
+++ b/crates/rpc/rpc-builder/src/lib.rs
@@ -1191,6 +1191,22 @@ impl<RpcMiddleware> RpcServerConfig<RpcMiddleware> {
         self
     }
 
+    /// Configures a custom tokio runtime for the rpc server.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
great, next step we'd need to bubble this up to

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/node/builder/src/rpc.rs#L422-L428

so that we configure this

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/node/builder/src/rpc.rs#L663-L663


after looking at this, perhaps we can rethink how we actually launch all of this

PR #17613, opened 2025-07-25
refactor: improve storage error messages

CreeptoGengar:
Improve error message specificity in storage layer components:

- Replace generic "{_0}" with "failed to decode RLP data: {_0}" in provider errors
- Add context "storage lock error: {_0}" for lockfile operations
- Specify "database error: {_0}" for database-related failures


mattsse (commented):
these act as transparent

PR #17614, opened 2025-07-25
test: modify discv5 startup test to use a random port

ssolit:
This PR addresses a reliability issue in the test_discv5_and_rlpx_same_socket_ok_without_discv4 integration test. Previously, the test used the DEFAULT_DISCOVERY_PORT, which could already be occupied by another process on the developer’s machine—causing intermittent test failures.

To improve developer experience during local testing and eliminates flakiness due to port conflicts, the test is now modified to dynamically select a random available UDP port at runtime. This ensures the test runs reliably regardless of the local environment setup.

### DIFF
--- a/crates/net/network/tests/it/startup.rs
+++ b/crates/net/network/tests/it/startup.rs
@@ -97,19 +97,25 @@ async fn test_discv5_and_discv4_same_socket_fails() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_discv5_and_rlpx_same_socket_ok_without_discv4() {
+    let test_port: u16 = TcpListener::bind("127.0.0.1:0") // 0 means OS assigns a free port
+        .await
+        .expect("Failed to bind to a port")
+        .local_addr()
+        .unwrap()
+        .port();
+
     let secret_key = SecretKey::new(&mut rand_08::thread_rng());
     let config = NetworkConfigBuilder::eth(secret_key)
-        .listener_port(DEFAULT_DISCOVERY_PORT)
+        .listener_port(test_port)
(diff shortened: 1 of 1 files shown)


crates/net/network/tests/it/startup.rs
```diff
@@ -97,19 +97,25 @@ async fn test_discv5_and_discv4_same_socket_fails() {
...
+        .await
+        .expect("Failed to bind to a port")
+        .local_addr()
+        .unwrap()
+        .port();
+
     let secret_key = SecretKey::new(&mut rand_08::thread_rng());
     let config = NetworkConfigBuilder::eth(secret_key)
-        .listener_port(DEFAULT_DISCOVERY_PORT)
+        .listener_port(test_port)
```
mattsse, line 109: this should just use port 0

mattsse (changes requested):

crates/net/network/tests/it/startup.rs
ssolit: I created a test_port variable because it is used in three places, and we are trying to test having the same value in all three places. If the value is 0, it's unclear if the code will grab (different) free ports multiple times, or use the same port three times

ssolit: Unless you are trying to say that the listener port should be zero, but the other two should be the same non-zero value. If so, I can make the change

ssolit: Hey @mattsse, wanted to follow up on this

mattsse (approved):
yeah, I guess this is a simple workaround for finding an unused port for this particular test setup

PR #17615, opened 2025-07-25
feat: added debug_storageRangeAt support

### LINKED ISSUE #17609: Add debug_storageRangeAt support
### Describe the feature

we're missing 

https://www.quicknode.com/docs/ethereum/debug_storageRangeAt

## TODO
* add fn to DebugApi
* implement endpoint similar to https://github.com/paradigmxyz/reth/pull/17601 for which we likely need another provider trait function that works like this pr

cc @Wollac

maybe @Rimeeeeee ?

### Additional context

_No response_

Rimeeeeee:
fixes #17609

### DIFF
--- a/crates/rpc/rpc-eth-api/src/node.rs
+++ b/crates/rpc/rpc-eth-api/src/node.rs
@@ -8,7 +8,7 @@ use reth_node_api::{FullNodeComponents, NodePrimitives, PrimitivesTy};
 use reth_primitives_traits::{BlockTy, HeaderTy, ReceiptTy, TxTy};
 use reth_rpc_eth_types::EthStateCache;
 use reth_storage_api::{
-    BlockReader, BlockReaderIdExt, StageCheckpointReader, StateProviderFactory,
+    BlockReader, BlockReaderIdExt, StageCheckpointReader, StateProviderFactory, StorageReader,
 };
 use reth_transaction_pool::{PoolTransaction, TransactionPool};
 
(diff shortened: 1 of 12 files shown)


crates/rpc/rpc-eth-api/src/node.rs
```diff
@@ -62,7 +63,9 @@ pub trait RpcNodeCore: Clone + Send + Sync + Unpin + 'static {
 
 impl<T> RpcNodeCore for T
 where
-    T: FullNodeComponents<Provider: ChainSpecProvider<ChainSpec: EthereumHardforks>>,
+    T: FullNodeComponents<
+        Provider: StorageReader + ChainSpecProvider<ChainSpec: EthereumHardforks>,
```
mattsse, line 67: these we can avoid if we add the `StorageReader` bound to

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/storage/provider/src/traits/full.rs#L13-L13

crates/rpc/rpc/src/debug.rs
```diff
@@ -1265,13 +1265,24 @@ where
...
         _tx_idx: usize,
-        _contract_address: Address,
-        _key_start: B256,
-        _max_result: u64,
-    ) -> RpcResult<()> {
-        Ok(())
+        contract_address: Address,
+        key_start: B256,
+        max_result: u64,
+    ) -> RpcResult<(Vec<(B256, StorageKey, StorageValue)>, Option<B256>)> {
```
mattsse, line 1273: for this, let's actually add this as a new type to alloy rpc-types-debug

https://github.com/ethereum/go-ethereum/blob/16117eb7cddc4584865af106d2332aa89f387d3d/eth/api_debug.go#L199-L203

mattsse (changes requested):
great start

some suggestions, and prep work that will make this nicer

Pana:
I will also impl this method

PR #17618, opened 2025-07-26
chore: remove duplicate deps

lean-apple:
Remove duplicate deps between `dep` and `dev-dep` (I checked if it was possible to keep only in dev-dep) after running subcommand `zepter lint duplicate-deps`. 

Didn't add the `zepter` check to ci yet as it may be premature.


mattsse (approved):
nice,

merging after #17620

PR #17622, opened 2025-07-26
chore: era-downloader: Remove debug print statements

LandLord529:
**Remove inappropriate debug output** in `reth-era-downloader`
   - Remove debug print statements that shouldn't be in production code
   - Replace with appropriate code comment


mattsse (approved):
makes sense

mattsse:
sorry, this has been fixed separately

PR #17623, opened 2025-07-26
Split blob helpers into BlobPoolExt (fixes #294)

chainguru1014:
### Motivation

L2s like RISE and OP do not need EIP-4844 blob transaction helpers. Currently, the core `TransactionPool` trait requires every consumer to implement these blob-related methods, even when they are unused. That forces downstream crates to depend on `alloy-eips` and `blobstore`, increasing compile time and dependency size.

This change splits those methods into a new trait `BlobPoolExt` behind the optional `"blob"` feature. This allows downstream projects to disable the feature cleanly via `default-features = false`.

---

### What changed

- Introduced new trait: `BlobPoolExt`, gated behind `#[cfg(feature = "blob")]`
- Moved 9 blob-specific helpers (e.g. `get_blob`, `delete_blobs`, etc.) out of `TransactionPool`
- Implemented `BlobPoolExt` for `Pool<V, T, S>` and `NoopTransactionPool`
- Updated tests, examples, and integration points
- CI passes with and without `--no-default-features`

### DIFF
--- a/crates/ethereum/payload/Cargo.toml
+++ b/crates/ethereum/payload/Cargo.toml
@@ -39,3 +39,7 @@ alloy-primitives.workspace = true
 
 # misc
 tracing.workspace = true
+
+[features]
+default = ["blob"]
+blob = ["reth-transaction-pool/blob"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3249,6 +3249,7 @@ dependencies = [
  "futures-util",
  "reqwest",
  "reth-ethereum",
+ "reth-transaction-pool",
  "serde",
  "serde_json",
  "thiserror 2.0.12",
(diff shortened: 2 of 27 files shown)


crates/ethereum/payload/Cargo.toml
```diff
@@ -39,3 +39,7 @@ alloy-primitives.workspace = true
 
 # misc
 tracing.workspace = true
+
+[features]
+default = ["blob"]
+blob = ["reth-transaction-pool/blob"]
```
kien-rise, line 45: This crate cannot live without `reth-transaction-pool/blob`.

Apply these changes:

```diff
- reth-transaction-pool.workspace = true
+ reth-transaction-pool = { workspace = true, features = ["blob"] }
```

and remove the features section:

```diff
- [features]
- default = ["blob"]
- blob = ["reth-transaction-pool/blob"]
```

PR #17631, opened 2025-07-27
feat: suggest gas price based on latest block's transactions

yiweichi:
This PR focus on implementing a simplified version of the GPO without needing to look at the tx pool, however look at the status of latest block. Suggest gas price base on if the latest block is at capacity. this will help to:
- reduce node configuration complexity by removing congestion threshold parameter
- reduce RPC node overhead by removing expensive query of tx pool size
- enable private mempool while any RPC node is still able to provide reasonable gas price estimate

Two JSON_RPC methods are effected.
1. `eth_gasPrice`, if the latest block is not at capacity, this will return a minimal suggest gas price.
2. `eth_feeHistory`,  if the latest block is not at capacity, the reward field of return will be the minimal suggest gas price.


mattsse (commented):
unsure if this pr was intended for reth

PR #17633, opened 2025-07-27
perf: box larger futures

mattsse:
these are quite large and polled very often

reducing size of the txpool handlers isnt very feasible


Rjected (approved):
makes sense to box large futures

PR #17634, opened 2025-07-27
chore: standardize console output in beacon-api-sidecar-fetcher example

NAIHANU:
Use consistent `println!` for all output messages in the beacon-api-sidecar-fetcher example instead of mixing `println!` and `eprintln!`. This maintains output consistency and follows best practices for example code.

Tested with `cargo check --package example-beacon-api-sidecar-fetcher`.


mattsse (commented):
this should be eprintln because an error

PR #17635, opened 2025-07-28
feat: add pending block mode configuration

adust09:
## Summary
- Add `PendingBlockMode` configuration to RPC server arguments
- Propagate pending block mode through the RPC builder chain
- Fix missing field in `EthApiBuilder` struct initialization

This PR adds the ability to configure how pending blocks are built for RPC requests with three modes:
- `full`: Build pending blocks with all transactions from the mempool (default)
- `empty`: Build pending blocks without any transactions  
- `none`: Don't build pending blocks, return None for pending requests

## Test plan
- [ ] Verify RPC server starts with default pending block mode
- [ ] Test setting different pending block modes via CLI args
- [ ] Ensure pending block requests behave according to the configured mode

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10126,6 +10126,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-eth",
  "alloy-sol-types",
+ "clap",
  "derive_more",
  "futures",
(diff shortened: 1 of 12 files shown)


mattsse (commented):
we have merged this

PR #17638, opened 2025-07-28
perf: use FuturesOrdered instead of join_all to yield results

### LINKED ISSUE #17610: Use Stream for fetching receipts chunks
### Describe the feature

This currently awaits _all_ spawned sub tasks

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc/src/eth/filter.rs#L1139-L1139

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc/src/eth/filter.rs#L1110-L1113

this means we're only yielding the next chunk after we've processed all:

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc/src/eth/filter.rs#L1154-L1154

we can replace this with a `FuturesOrdered` instead and store this as a field in

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc/src/eth/filter.rs#L1010-L1012

for this we need to rewrite the logic of the function slightly:

so that we first poll the futuresordered here:

https://github.com/paradigmxyz/reth/blob/a7cbf81b65acacbca5e089593d0516ade936550f/crates/rpc/rpc/src/eth/filter.rs#L1023-L1026

which we can do with `poll_fn` for example

### Additional context

assigning @ongyimeng

ongyimeng:
Resolves #17610

### DIFF
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -1048,6 +1048,7 @@ impl<
                     continue;
                 }
                 // Task completed with empty results, continue to check for more work
+                trace!(target: "rpc::eth::filter", "Parallel task completed with no receipt results");
                 continue;
             }
 
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -1099,13 +1099,14 @@ impl<
             }
         }
 
-        // If we reach here, something unexpected happened
+        // If we reach here, a task completed with empty results but other tasks are still pending.
+        // Return None and let the next call handle remaining tasks.
         debug!(
             target: "rpc::eth::filter",
             pending_tasks_len = self.pending_tasks.len(),
             queued_results_len = self.next.len(),
             headers_remaining = self.iter.len(),
-            "RangeBlockMode reached unexpected state - returning None"
+            "Task returned empty results with pending tasks remaining - returning None"
         );
         Ok(None)
     }
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -1085,7 +1085,7 @@ impl<
             if remaining_headers >= PARALLEL_PROCESSING_THRESHOLD {
                 self.spawn_parallel_tasks(range_headers);
             } else {
-                self.spawn_sequential_task(range_headers);
+                return self.process_small_range(range_headers).await;
             }
 
             // Now wait for the newly spawned task to complete
@@ -1111,46 +1111,43 @@ impl<
         Ok(None)
     }
 
-    /// Spawn a sequential task for processing a small range of headers
+    /// Process a small range of headers sequentially
     ///
     /// This is used when the remaining headers count is below [`PARALLEL_PROCESSING_THRESHOLD`].
-    fn spawn_sequential_task(
+    async fn process_small_range(
         &mut self,
         range_headers: Vec<SealedHeader<<Eth::Provider as HeaderProvider>::Header>>,
-    ) {
-        let filter_inner = self.filter_inner.clone();
-        let task = Box::pin(async move {
-            let mut results = Vec::new();
+    ) -> Result<Option<ReceiptBlockResult<Eth::Provider>>, EthFilterError> {
+        // Process each header individually to avoid queuing for all receipts
+        for header in range_headers {
+            // First check if already cached to avoid unnecessary provider calls
+            let (maybe_block, maybe_receipts) = self
+                .filter_inner
+                .eth_cache()
+                .maybe_cached_block_and_receipts(header.hash())
+                .await?;
 
-            for header in range_headers {
-                // First check if already cached to avoid unnecessary provider calls
-                let (maybe_block, maybe_receipts) =
-                    filter_inner.eth_cache().maybe_cached_block_and_receipts(header.hash()).await?;
-
-                let receipts = match maybe_receipts {
-                    Some(receipts) => receipts,
-                    None => {
-                        // Not cached - fetch directly from provider
-                        match filter_inner.provider().receipts_by_block(header.hash().into())? {
-                            Some(receipts) => Arc::new(receipts),
-                            None => continue, // No receipts found
-                        }
+            let receipts = match maybe_receipts {
+                Some(receipts) => receipts,
+                None => {
+                    // Not cached - fetch directly from provider
+                    match self.filter_inner.provider().receipts_by_block(header.hash().into())? {
+                        Some(receipts) => Arc::new(receipts),
+                        None => continue, // No receipts found
                     }
-                };
-
-                if !receipts.is_empty() {
-                    results.push(ReceiptBlockResult {
-                        receipts,
-                        recovered_block: maybe_block,
-                        header,
-                    });
                 }
+            };
+
+            if !receipts.is_empty() {
+                self.next.push_back(ReceiptBlockResult {
+                    receipts,
+                    recovered_block: maybe_block,
+                    header,
+                });
             }
+        }
 
-            Ok(results)
-        });
-
-        self.pending_tasks.push_back(task);
+        Ok(self.next.pop_front())
     }
 
     /// Spawn parallel tasks for processing a large range of headers
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -1011,6 +1011,10 @@ impl<
     }
 }
 
+/// Type alias for the complex future type used in `RangeBlockMode`
+type ReceiptFetchFuture<P> =
+    Pin<Box<dyn Future<Output = Result<Vec<ReceiptBlockResult<P>>, EthFilterError>> + Send>>;
+
 /// Mode for processing blocks using range queries for older blocks
 struct RangeBlockMode<
     Eth: RpcNodeCoreExt<Provider: BlockIdReader, Pool: TransactionPool> + EthApiTypes + 'static,
@@ -1020,9 +1024,7 @@ struct RangeBlockMode<
     next: VecDeque<ReceiptBlockResult<Eth::Provider>>,
     max_range: usize,
     // Stream of ongoing receipt fetching tasks
-    pending_tasks: FuturesOrdered<
-        Pin<Box<dyn Future<Output = Result<Vec<ReceiptBlockResult<Eth::Provider>>, EthFilterError>> + Send>>
-    >,
+    pending_tasks: FuturesOrdered<ReceiptFetchFuture<Eth::Provider>>,
 }
 
 impl<
@@ -1108,13 +1110,11 @@ impl<
         let filter_inner = self.filter_inner.clone();
         let task = Box::pin(async move {
             let mut results = Vec::new();
-            
+
             for header in range_headers {
                 // First check if already cached to avoid unnecessary provider calls
-                let (maybe_block, maybe_receipts) = filter_inner
-                    .eth_cache()
-                    .maybe_cached_block_and_receipts(header.hash())
-                    .await?;
+                let (maybe_block, maybe_receipts) =
+                    filter_inner.eth_cache().maybe_cached_block_and_receipts(header.hash()).await?;
 
                 let receipts = match maybe_receipts {
                     Some(receipts) => receipts,
@@ -1138,7 +1138,7 @@ impl<
 
             Ok(results)
         });
-        
+
         self.pending_tasks.push_back(task);
     }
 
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+    // Stream of ongoing receipt fetching tasks
+    pending_tasks: FuturesOrdered<ReceiptFetchFuture<Eth::Provider>>,
 }
 
 impl<
         Eth: RpcNodeCoreExt<Provider: BlockIdReader, Pool: TransactionPool> + EthApiTypes + 'static,
     > RangeBlockMode<Eth>
 {
     async fn next(&mut self) -> Result<Option<ReceiptBlockResult<Eth::Provider>>, EthFilterError> {
+        // First, try to return any already processed results
```
emhane:
```suggestion
        // First, try to return any already processed result
```

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+            if let Some(task_result) = self.pending_tasks.next().await {
+                match task_result {
+                    Ok(chunk_results) => {
+                        for result in chunk_results {
+                            self.next.push_back(result);
+                        }
+                        return Ok(self.next.pop_front());
+                    }
+                    Err(e) => return Err(e),
+                }
```
emhane:
```suggestion
                for result in task_results? {
                    self.next.push_back(result);
                }
                return Ok(self.next.pop_front());
```
less loc, more readable 

this can also be optimised to only add elements `task_results[1..]` to the `VecDeque`, as `self.next.pop_front()` here will always return `task_results[0]` since the code at start of this method ensures we don't add more items to `next` unless `next` is empty.

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+            }
+        }
+
+        // If no pending tasks and no queued results, try to start new tasks
+        if self.pending_tasks.is_empty() {
+            // Try to get the next header(s) to process
```
emhane: here is an example of why this would be better as a proper `Stream` impl instead of custom method, because the appropriate thing to do here is schedule for wake up and return `Poll::Pending`, out of scope of this pr though. cc @mattsse @fgimenez

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+            while range_headers.len() < self.max_range {
+                let Some(peeked) = self.iter.peek() else { break };
+                let Some(last_header) = range_headers.last() else { break };
 
-        let mut range_headers = Vec::with_capacity(self.max_range);
-        range_headers.push(next_header);
+                let expected_next = last_header.number() + 1;
+                if peeked.number() != expected_next {
+                    break; // Non-consecutive block, stop here
+                }
```
emhane: probably helpful to add a debug log here

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+                let Some(next_header) = self.iter.next() else { break };
+                range_headers.push(next_header);
+            }
 
-            let expected_next = last_header.header().number() + 1;
-            if peeked.header().number() != expected_next {
-                break; // Non-consecutive block, stop here
+            // Check if we should use parallel processing for large ranges
+            let remaining_headers = self.iter.len() + range_headers.len();
+            if remaining_headers >= PARALLEL_PROCESSING_THRESHOLD as usize {
```
emhane: wondering if `PARALLEL_PROCESSING_THRESHOLD` can be a `usize` const so this cast isn't needed

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+            if let Some(task_result) = self.pending_tasks.next().await {
+                match task_result {
+                    Ok(chunk_results) => {
+                        for result in chunk_results {
+                            self.next.push_back(result);
+                        }
+                        return Ok(self.next.pop_front());
+                    }
+                    Err(e) => return Err(e),
+                }
```
emhane: same here, see suggestion above on reducing loc for better readability

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
         }
 
-        // Check if we should use parallel processing for large ranges
-        let remaining_headers = self.iter.len() + range_headers.len();
-        if remaining_headers >= PARALLEL_PROCESSING_THRESHOLD as usize {
-            self.process_large_range(range_headers).await
-        } else {
-            self.process_small_range(range_headers).await
-        }
+        // If we reach here, something unexpected happened
```
emhane: that's an indicator we want a debug log here, maybe even `debug_assert` to panic in debug mode

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+        Ok(None)
     }
 
-    /// Process small range headers
-    async fn process_small_range(
+    /// Spawn a sequential task for processing a small range of headers
```
emhane: pls link to the parallel threshold constant in these docs

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,130 @@ struct RangeBlockMode<
...
+                    filter_inner.eth_cache().maybe_cached_block_and_receipts(header.hash()).await?;
+
+                let receipts = match maybe_receipts {
+                    Some(receipts) => receipts,
+                    None => {
+                        // Not cached - fetch directly from provider
+                        match filter_inner.provider().receipts_by_block(header.hash().into())? {
+                            Some(receipts) => Arc::new(receipts),
+                            None => continue, // No receipts found
+                        }
```
emhane: a cache usually works so that there is one method, e.g. `load_receipt`, which handles reading from disk and caching values on so called _cache miss_. i.e. this logic should be abstracted away into `EthStateCache` as follow up issue. cc @mattsse @fgimenez

emhane (changes requested):
nice!

ongyimeng:
hi emilia, thanks so much for your feedback! I have made the changes, let me know if this looks good to you 🦀 @emhane

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,143 @@ struct RangeBlockMode<
...
         }
 
-        // Check if we should use parallel processing for large ranges
-        let remaining_headers = self.iter.len() + range_headers.len();
-        if remaining_headers >= PARALLEL_PROCESSING_THRESHOLD as usize {
-            self.process_large_range(range_headers).await
-        } else {
-            self.process_small_range(range_headers).await
-        }
+        // If we reach here, something unexpected happened
```
emhane: we reach here if the next pending task returns empty chunk results, and there are still pending tasks in the queue. can be more clear about this in the comment and log message.

emhane (approved):
nice! smol comment, asides that lgtm

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,144 @@ struct RangeBlockMode<
...
-            return Ok(None);
-        };
+        // Next, try to get a completed task result if there are pending tasks
+        if !self.pending_tasks.is_empty() {
+            if let Some(task_result) = self.pending_tasks.next().await {
+                let mut chunk_results = task_result?;
+                if !chunk_results.is_empty() {
+                    let first_result = chunk_results.remove(0);
+                    self.next.extend(chunk_results);
+                    return Ok(Some(first_result));
```
mattsse: this is okay, alternatively we could also wrap this section in a loop but this also works

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,144 @@ struct RangeBlockMode<
...
-            range_headers.push(next_header);
+            // Now wait for the newly spawned task to complete
+            if let Some(task_result) = self.pending_tasks.next().await {
+                let mut chunk_results = task_result?;
+                if !chunk_results.is_empty() {
+                    let first_result = chunk_results.remove(0);
+                    self.next.extend(chunk_results);
+                    return Ok(Some(first_result));
+                }
+            }
```
mattsse: repeating this feels okay here as well

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,144 @@ struct RangeBlockMode<
...
+        );
+        Ok(None)
     }
 
-    /// Process small range headers
-    async fn process_small_range(
+    /// Spawn a sequential task for processing a small range of headers
+    ///
+    /// This is used when the remaining headers count is below [`PARALLEL_PROCESSING_THRESHOLD`].
+    fn spawn_sequential_task(
```
mattsse: this doesn't actually spawn a task

I'd like to keep this function at this was previously because this behaviour didn't change, and we can instead await this call directly

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,89 +1023,144 @@ struct RangeBlockMode<
...
 
-            let expected_next = last_header.header().number() + 1;
-            if peeked.header().number() != expected_next {
-                break; // Non-consecutive block, stop here
+            // Check if we should use parallel processing for large ranges
+            let remaining_headers = self.iter.len() + range_headers.len();
+            if remaining_headers >= PARALLEL_PROCESSING_THRESHOLD {
+                self.spawn_parallel_tasks(range_headers);
+            } else {
+                self.spawn_sequential_task(range_headers);
```
mattsse: here we'd need to await the sequential task directly and if it's parallel we we'd effectively loop or self.pending_tasks.next().await

mattsse (changes requested):
a few more suggestions

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,48 +1023,82 @@ struct RangeBlockMode<
...
-            let Some(next_header) = self.iter.next() else { break };
-            range_headers.push(next_header);
-        }
+                let expected_next = last_header.number() + 1;
+                if peeked.number() != expected_next {
+                    debug!(
+                        target: "rpc::eth::filter",
+                        last_block = last_header.number(),
+                        next_block = peeked.number(),
+                        expected = expected_next,
```
fgimenez, line 1067: would be good to include the range size here
```suggestion
                        expected = expected_next,
                        range_size = range_headers.len(),
```

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1006,6 +1011,10 @@ impl<
     }
 }
 
+/// Type alias for the complex future type used in `RangeBlockMode`
```
fgimenez: we could mention here that this is used to manage parallel receipt fetching tasks

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1014,48 +1023,82 @@ struct RangeBlockMode<
...
-        };
+            // Try to get a completed task result if there are pending tasks
+            if let Some(task_result) = self.pending_tasks.next().await {
+                let chunk_results = task_result?;
+                if !chunk_results.is_empty() {
+                    // Add all results to buffer, first one will be returned in next loop iteration
+                    self.next.extend(chunk_results);
+                    // Continue loop to return the first result from buffer
+                    continue;
+                }
```
fgimenez: we could log here when `chunk_results.is_empty()`, something like `trace!(target: "rpc::eth::filter", "Task completed with empty results");` wdyt?

fgimenez (approved):
looks good, minor nits, pending @mattsse

crates/rpc/rpc/src/eth/filter.rs
ongyimeng: sounds good!

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -1106,52 +1154,46 @@ impl<
...
-                    }
-                }
-                Ok(Err(e)) => return Err(e),
-                Err(_join_err) => {
-                    return Err(EthFilterError::InternalError);
+                // Await the blocking task and handle the result
+                match chunk_task.await {
+                    Ok(Ok(chunk_results)) => Ok(chunk_results),
+                    Ok(Err(e)) => Err(e),
+                    Err(_join_err) => Err(EthFilterError::InternalError),
```
fgimenez: we could log details about what happened here before losing context:
```suggestion
                    Err(join_err) => {
                        trace!(target: "rpc::eth::filter", error = ?join_err, "Task join error");
                        Err(EthFilterError::InternalError)
                    }
```

fgimenez (approved):
looks good, smol log suggestion

crates/rpc/rpc/src/eth/filter.rs
ongyimeng: slfr I have made the edit, thanks for your suggestion!! @fgimenez

mattsse (approved):

PR #17641, opened 2025-07-28
feat: separate EngineValidator from PayloadValidator

### LINKED ISSUE #17639: Separate `EngineValidator` from `PayloadValidator`
### Describe the feature

This trait is now only used by the engineAPI itself:

https://github.com/paradigmxyz/reth/blob/9d1af5a09cc7794a767858eb3219a24b7e52fc16/crates/engine/primitives/src/lib.rs#L107-L123

https://github.com/paradigmxyz/reth/blob/9d1af5a09cc7794a767858eb3219a24b7e52fc16/crates/rpc/rpc-engine-api/src/engine_api.rs#L74-L79

and we can relax this trait so that it is a standalone implementation
we only need to use this when we create the engineapi for op and eth

https://github.com/paradigmxyz/reth/blob/9d1af5a09cc7794a767858eb3219a24b7e52fc16/crates/node/builder/src/rpc.rs#L1101-L1101

## TODO
* remove supertrait PayloadValidator from EngineValidator

### Additional context

_No response_

fgimenez:
closes #17639

after the separation we need a new `Block` associated type in `EngineValidator`. also, `BasicEngineValidator` requires `PayloadValidator`m the bound is added to `EngineValidatorAddOn` and `EngineValidatorBuilder`

### DIFF
--- a/crates/engine/primitives/src/lib.rs
+++ b/crates/engine/primitives/src/lib.rs
@@ -106,9 +106,6 @@ pub trait EngineTypes:
 
 /// Type that validates the payloads processed by the engine.
 pub trait EngineValidator<Types: PayloadTypes>: Send + Sync + Unpin + 'static {
-    /// The block type used by the engine.
-    type Block: Block;
-
     /// Validates the presence or exclusion of fork-specific fields based on the payload attributes
     /// and the message version.
     fn validate_version_specific_fields(
--- a/crates/engine/primitives/src/lib.rs
+++ b/crates/engine/primitives/src/lib.rs
@@ -105,7 +105,10 @@ pub trait EngineTypes:
 }
 
 /// Type that validates the payloads processed by the engine.
-pub trait EngineValidator<Types: PayloadTypes>: PayloadValidator<Types> {
+pub trait EngineValidator<Types: PayloadTypes>: Send + Sync + Unpin + 'static {
+    /// The block type used by the engine.
+    type Block: Block;
+
     /// Validates the presence or exclusion of fork-specific fields based on the payload attributes
     /// and the message version.
     fn validate_version_specific_fields(
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -11,7 +11,7 @@ use reth_chain_state::CanonStateSubscriptions;
 use reth_chainspec::{ChainSpecProvider, EthereumHardforks};
 use reth_node_api::{
     AddOnsContext, BlockTy, EngineTypes, EngineValidator, FullNodeComponents, FullNodeTypes,
-    NodeAddOns, NodeTypes, PayloadTypes, PrimitivesTy,
+    NodeAddOns, NodeTypes, PayloadTypes, PayloadValidator, PrimitivesTy,
 };
 use reth_node_core::{
     node_config::NodeConfig,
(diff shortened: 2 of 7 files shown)


crates/engine/primitives/src/lib.rs
```diff
@@ -105,7 +105,10 @@ pub trait EngineTypes:
 }
 
 /// Type that validates the payloads processed by the engine.
-pub trait EngineValidator<Types: PayloadTypes>: PayloadValidator<Types> {
+pub trait EngineValidator<Types: PayloadTypes>: Send + Sync + Unpin + 'static {
+    /// The block type used by the engine.
+    type Block: Block;
```
mattsse: where do we need this block type?

crates/node/builder/src/rpc.rs
```diff
@@ -991,6 +991,7 @@ pub trait EthApiBuilder<N: FullNodeComponents>: Default + Send + 'static {
 pub trait EngineValidatorAddOn<Node: FullNodeComponents>: Send {
     /// The Validator type to use for the engine API.
     type Validator: EngineValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
+        + PayloadValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
         + Clone;
```
mattsse: this makes sense because:

https://github.com/paradigmxyz/reth/blob/b1f1e9d71174a8e6456825a59f40c3103ed2d7e8/crates/node/builder/src/launch/engine.rs#L193-L193

but this also means that we could get rid of ` EV: EngineValidatorBuilder<N>,` for the engineapi entirely

https://github.com/paradigmxyz/reth/blob/b1f1e9d71174a8e6456825a59f40c3103ed2d7e8/crates/node/builder/src/rpc.rs#L1076-L1088

if we repurpose the trait for this there, then the engineapi validation would be its standalone thing

and next we could start changing

https://github.com/paradigmxyz/reth/blob/b1f1e9d71174a8e6456825a59f40c3103ed2d7e8/crates/node/builder/src/rpc.rs#L1017-L1028

to something that creates:

https://github.com/paradigmxyz/reth/blob/b1f1e9d71174a8e6456825a59f40c3103ed2d7e8/crates/node/builder/src/launch/engine.rs#L210-L217

mattsse (commented):
this makes sense as incremental next step, only have one q re block type

and some ideas for next changes ptal @klkvr

crates/engine/primitives/src/lib.rs
fgimenez: the constraints come from `EngineValidatorBuilder` and `EngineValidatorAddOn`, maybe better remove them?
```
error[E0220]: associated type `Block` not found for `reth_node_api::EngineValidator`
   --> crates/node/builder/src/rpc.rs:993:74
    |
993 |     type Validator: EngineValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
    |                                                                          ^^^^^ associated type `Block` not found

error[E0220]: associated type `Block` not found for `reth_node_api::EngineValidator`
    --> crates/node/builder/src/rpc.rs:1035:69
     |
1035 |     Validator: EngineValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
     |                                                                     ^^^^^ associated type `Block` not found

error[E0220]: associated type `Block` not found for `reth_node_api::EngineValidator`
    --> crates/node/builder/src/rpc.rs:1021:74
     |
1021 |     type Validator: EngineValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
     |                                                                          ^^^^^ associated type `Block` not found
```

klkvr: yeah ideally we should be able to get rid of it but I guess we rely on it somewhere in RPC

klkvr (approved):
lgtm

crates/engine/primitives/src/lib.rs
fgimenez: makes sense, done

fgimenez: no rpc requirements at the end

mattsse (approved):

mattsse:
@fgimenez actually we can either merge this as is and followup with the changes or iterate on this here, both make sense to me

PR #17643, opened 2025-07-28
chore(trie): fully reveal sparse tries prior to leaf updates/removals

### LINKED ISSUE #17571: Updates and removals of leaves cause sparse trie thread to block
### Describe the feature

Because of blinded node fetching. We should do something about it.

### Additional context

_No response_

mediocregopher:
Fixes #17571

**Background**

The Serial/ParallelSparseTrie's have their nodes revealed using the DecodedMultiProofs generated based on the previous database state and the changed leafset. Once revealed, the leaf updates/removals are applied to the sparse tries and root hashes are calculated.

There are specific edge-cases during leaf adding/removal where a node which is outside the changeset is required to complete the operation. In these cases we were falling back to a singular database lookup to fill in the gap.

By modifying the generation of the DecodedMultiProofs to include those extra required nodes we can simplify the sparse trie implementations significantly, as well as the surrounding engine code which supports these one-off lookups.

**Changes**

This primarily relies on a change to the ProofRetainer in alloy-trie, made here:

https://github.com/alloy-rs/trie/pull/109

These changes are enabled by a flag which the payload validator enables; other places where proofs are generated remain unnaffected.

There is a further change here to the trie walker to not skip over branch nodes when they might be involved in a leaf removal. A change to support leaf addition is not required, because that case involves children of extension nodes, and we don't ever skip over extension nodes anyway.

The sparse trie code to fallback to the db for missing nodes remains in place for now, but we emit a warning when it happens. This will let us track if there's any edge-cases we're missing.

**Benchmarks - 2k blocks on main**

No significant change

```
Timestamp: 2025-07-28 10:37:12 UTC
Baseline: main
Feature:  origin/mediocregopher/17571-leaf-updates-removals

Performance Changes:
  NewPayload Latency: -1.24%
  FCU Latency:        -1.74%
  Total Latency:      -1.26%
  Gas/Second:         +1.27%
  Blocks/Second:      +1.27%

Baseline Summary:
  Blocks: 2000, Gas: 36118922440, Duration: 74.02s
  Avg NewPayload: 35.69ms, Avg FCU: 1.29ms, Avg Total: 36.99ms
  Started: 2025-07-28 10:48:20 UTC, Ended: 2025-07-28 10:56:39 UTC

Feature Summary:
  Blocks: 2000, Gas: 36118922440, Duration: 73.08s
  Avg NewPayload: 35.25ms, Avg FCU: 1.27ms, Avg Total: 36.52ms
  Started: 2025-07-28 10:58:35 UTC, Ended: 2025-07-28 11:06:55 UTC
```


Rjected:
> The sparse trie code to fallback to the db for missing nodes remains in place for now, but we emit a warning when it happens. This will let us track if there's any edge-cases we're missing.

Assuming there are no warnings from the benches? Are there any changes in the metrics (nodes fetched for ex)

mediocregopher:
> > The sparse trie code to fallback to the db for missing nodes remains in place for now, but we emit a warning when it happens. This will let us track if there's any edge-cases we're missing.
> 
> Assuming there are no warnings from the benches? Are there any changes in the metrics (nodes fetched for ex)

I swear there were no warnings when I ran this against mainnet on monday, but now I'm seeing two related to leaf removal, so there's a bit more investigation to do here :facepalm: 

Regarding metrics, I just reran the bench (2k blocks on mainnet) and this time it came out 3% faster :shrug: there were indeed more multiproofs fetched and more db txs generally though.

shot-1753882921
shot-1753883001

Summary:
Total multiproof account nodes: +20.4118%
Total multiproof storage nodes: +19.1292%
Opened Read-Write Transactions: +13.9773%
Opened Read-Only Transactions: +12.8246%

crates/trie/trie/src/walker.rs
```diff
@@ -150,10 +171,28 @@ impl<C> TrieWalker<C> {
...
+            let key_is_only_nonremoved_child =
+                self.added_removed_keys.as_ref().is_some_and(|added_removed_keys| {
+                    node.full_key_is_only_nonremoved_child(added_removed_keys.as_ref())
+                });
+
+            let _full_key = node.full_key();
+            trace!(
+                target: "trie::walker",
+                ?key_is_only_nonremoved_child,
+                ?_full_key,
```
Rjected: can we do this?
```suggestion
                ?key_is_only_nonremoved_child,
                full_key=?node.full_key(),
```

crates/trie/common/src/added_removed_keys.rs
```diff
@@ -0,0 +1,211 @@
...
+                if *val == U256::ZERO {
+                    storage_removed_keys.insert_removed(*key);
+                } else {
+                    storage_removed_keys.remove_removed(key);
+                }
+            }
+
+            if !account.is_empty() {
+                self.account.remove_removed(hashed_address);
+            }
```
Rjected: hmm, is it possible for these account empty cases to be hit? not suggesting they be removed, just wondering, we might want to just add docs if they can't be

crates/trie/parallel/src/proof_task.rs
```diff
@@ -265,19 +268,37 @@ where
...
+            .unwrap_or_else(|| Arc::new(MultiAddedRemovedKeys::new()));
+        let added_removed_keys = multi_added_removed_keys.get_storage(&input.hashed_address);
+
+        let span = tracing::trace_span!(
+            target: "trie::proof_task",
+            "Storage proof calculation",
+            hashed_address=?input.hashed_address,
+            // Add a random id because we often have parallel storage proof calculations for the
+            // same hashed address, and we want to differentiate them during trace analysis.
+            span_id=rand::random::<u64>(),
```
Rjected: I would prefer not to be doing `rand` operations here, maybe we could just assign each task a number for this?

Rjected (commented):
the comments are very helpful, I mainly have one question about empty account edge cases, and a suggestion wrt the span ID

crates/trie/parallel/src/proof_task.rs
mediocregopher: Done, added an `id` field to the ProofTaskTx

crates/trie/common/src/added_removed_keys.rs
mediocregopher: I originally thought it wasn't possible, but they do actually get hit. For example if I comment out these `self.account` lines I get this error on block 22729256 of mainnet:

```
WARN Branch node child not revealed in remove_leaf, falling back to db child_path=Nibbles(0xba982e83) leaf_full_path=Nibbles(0xba982e827371ca20eb274c77c02d21b0acb7e04fed4f6b28a73948060688dc6e)
```

This corresponds to https://etherscan.io/address/0xecb7ca9ec4cb52536b61227176993216cf4f4154, which gets self-destructed in that tx without having ever called a tx itself (and therefore having a zero nonce).

crates/trie/common/src/added_removed_keys.rs
```diff
@@ -0,0 +1,217 @@
...
+                    self.account.insert_removed(*hashed_address);
+                }
+                continue
+            }
+
+            let storage_removed_keys =
+                self.storages.entry(*hashed_address).or_insert_with(default_added_removed_keys);
+
+            for (key, val) in &storage.storage {
+                if *val == U256::ZERO {
```
mattsse: there's is_zero()
```suggestion
                if val.is_zero() {
```

crates/trie/trie/src/proof/mod.rs
```diff
@@ -183,7 +184,7 @@ where
 
 /// Generates storage merkle proofs.
 #[derive(Debug)]
-pub struct StorageProof<T, H> {
+pub struct StorageProof<T, H, K = AddedRemovedKeys> {
```
mattsse, line 187: ah I see, this way this works for both, proofretainer and addedremovedkeys

makes sense

mattsse (commented):
I believe I was able to follow along here

all of this made sense, lgtm but also want @Rjected to take a final look at this

Rjected (approved):
LGTM, nice

PR #17645, opened 2025-07-28
perf: add benchmark for on_canonical_state_change

mattsse:
## Summary

Adds a comprehensive benchmark for the transaction pool's `on_canonical_state_change` method to help evaluate performance optimizations.

## Description

This PR adds a new benchmark that measures the performance of `Pool::on_canonical_state_change`, which is called when the canonical chain state changes and the pool needs to update its internal state accordingly.

The benchmark:
- Tests with varying pool sizes: 100 to 50,000 senders with 1-10 transactions each
- Simulates realistic state changes where all accounts have their nonce incremented and balance decreased
- Uses the same benchmark infrastructure as existing pool benchmarks for consistency

## Context

This benchmark is being added in preparation for evaluating the optimizations proposed in #17434, which aims to improve the performance of canonical state updates in the transaction pool.

## Testing

Run the benchmark with:
```bash
cargo bench -p reth-transaction-pool --bench canonical_state_change
```

For specific test cases:
```bash
cargo bench -p reth-transaction-pool --bench canonical_state_change -- "senders: 1000 | txs_per_sender: 5"
```

### DIFF
--- a/crates/transaction-pool/benches/canonical_state_change.rs
+++ b/crates/transaction-pool/benches/canonical_state_change.rs
@@ -65,38 +65,6 @@ async fn fill_pool(pool: &TestPoolBuilder, txs: Vec<MockTransaction>) -> HashMap
     sender_nonces
 }
 
-/// Create a canonical state update with changed accounts for all senders
-fn create_canonical_update(
-    sender_nonces: HashMap<Address, u64>,
-) -> (Box<SealedBlock<Block>>, CanonicalStateUpdate<'static, Block>) {
-    // Create a mock block - using default Ethereum block
-    let header = Header::default();
-    let body = BlockBody::default();
-    let block = Block { header, body };
-    let sealed_block = Box::new(SealedBlock::seal_slow(block));
-    let sealed_block_ref = Box::leak(sealed_block.clone());
-
-    let changed_accounts: Vec<ChangedAccount> = sender_nonces
-        .into_iter()
-        .map(|(address, nonce)| ChangedAccount {
-            address,
-            nonce: nonce + 1, // Increment nonce as if transactions were mined
-            balance: U256::from(9_000_000_000_000_000u64), // Decrease balance
-        })
-        .collect();
-
-    let update = CanonicalStateUpdate {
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/benches/canonical_state_change.rs
```diff
@@ -0,0 +1,156 @@
...
+                tx.set_sender(sender);
+                tx.set_nonce(nonce as u64);
+            }
+
+            txs.push(tx);
+        }
+    }
+
+    txs
+}
```
Rjected, line 47: we have `create_transactions_for_sender` in the truncate tests, maybe we could combine these two fns? Or include them somewhere where they can be reused? I can see us writing more txpool benches

crates/transaction-pool/benches/canonical_state_change.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Create a canonical state update with changed accounts for all senders
+fn create_canonical_update(
+    sender_nonces: HashMap<Address, u64>,
+) -> CanonicalStateUpdate<'static, Block> {
+    // Create a mock block - using default Ethereum block
+    let header = Header::default();
+    let body = BlockBody::default();
+    let block = Block { header, body };
+    let sealed_block = Box::leak(Box::new(SealedBlock::seal_slow(block)));
```
Rjected: I see, this is so we force `'static`? why do we need `'static` here?

Rjected (commented):
benchmark looks good, we should make more, just have 2 questions

crates/transaction-pool/benches/canonical_state_change.rs
mattsse: yeah, I could just inline this fn then this wouldnt be necessary but shouldnt matter for this

crates/transaction-pool/benches/canonical_state_change.rs
mattsse: that makes sense, but would like to get this in first

Rjected (approved):
lgtm

PR #17647, opened 2025-07-28
perf(txpool): avoid redundant notifications

mattsse:
these listeners are mainly intended to be used programatically and not usually used, but currently we always process tx events.

this adds checks for is_empty to return early if there's no listener installed


Rjected (approved):
nice, this makes sense

PR #17655, opened 2025-07-29
refactor: move invalid block hook creation from LaunchContext to AddOnsContext

fgimenez:
* extracted `invalid_block_hook` logic from `LaunchContextWith`
* made it accessible from `AddOnsContext` by creating an extension trait `InvalidBlockHookExt` that adds the `create_invalid_block_hook` method to `AddOnsContext`
* moved the `NoopInvalidBlockHook` and `InvalidBlockHooks` types to `reth-engine-primitives` to avoid cyclic dependencies


mattsse (approved):
lgtm!

PR #17656, opened 2025-07-29
refactor(cli): replace From<Header> with CliHeader trait

### LINKED ISSUE #17642: Remove `From<Header>` bound from CLI
### Describe the feature

Right now for reusing `Cli` with custom header you need to define a `From<Header>` implementation: 
https://github.com/paradigmxyz/reth/blob/9d1af5a09cc7794a767858eb3219a24b7e52fc16/crates/ethereum/cli/src/interface.rs#L129

This is not great because we only need this for a single cli command but this impl might be accidentally misused.

What we could do is instead introduce
```rust
trait CliHeader {
    fn set_number(&mut self, number: u64);
}
```

and require it to be implemented instead, That way we will be able to use this trait here https://github.com/paradigmxyz/reth/blob/9d1af5a09cc7794a767858eb3219a24b7e52fc16/crates/cli/commands/src/init_state/mod.rs#L108-L111

and get rid of `From<Header>` requirement

### Additional context

_No response_

shiyasmohd:
Closes #17642

### DIFF
--- a/crates/cli/commands/src/common.rs
+++ b/crates/cli/commands/src/common.rs
@@ -143,7 +143,7 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
         {
             if factory.db_ref().is_read_only()? {
                 warn!(target: "reth::cli", ?unwind_target, "Inconsistent storage. Restart node to heal.");
-                return Ok(factory)
+                return Ok(factory);
             }
 
             // Highly unlikely to happen, and given its destructive nature, it's better to panic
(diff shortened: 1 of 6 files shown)


crates/optimism/cli/src/commands/init_state.rs
```diff
@@ -71,7 +75,7 @@ impl<C: ChainSpecParser<ChainSpec = OpChainSpec>> InitStateCommandOp<C> {
             } else if last_block_number > 0 && last_block_number < BEDROCK_HEADER.number {
                 return Err(eyre::eyre!(
                     "Data directory should be empty when calling init-state with --without-ovm."
-                ))
+                ));
```
klkvr:
```suggestion
                ))
```

crates/cli/commands/src/common.rs
```diff
@@ -143,7 +143,7 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
         {
             if factory.db_ref().is_read_only()? {
                 warn!(target: "reth::cli", ?unwind_target, "Inconsistent storage. Restart node to heal.");
-                return Ok(factory)
+                return Ok(factory);
```
klkvr:
```suggestion
                return Ok(factory)
```

klkvr (approved):

PR #17658, opened 2025-07-29
fix: formatting and syntax in ECDSA recovery benchmark

VolodymyrBg:
- Fix missing space in hex literal assignment
- Correct indentation and add missing semicolon


mattsse (commented):
this fmt is fine

PR #17661, opened 2025-07-29
refactor: stream transactions while executing payload

klkvr:
Preparation for https://github.com/paradigmxyz/reth/issues/17291

Changes main execution code-paths to not depend on the block and instead take `EvmEnv` directly and take transactions as an iterator of `OwnedExecutableTx`.

`OwnedExecutableTx` is a new trait intoroduced as a helper to bound `'static` types that can be passed to block executors.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -9,15 +9,14 @@ use crate::tree::{
     sparse_trie::SparseTrieTask,
     StateProviderBuilder, TreeConfig,
 };
-use alloy_consensus::{transaction::Recovered, BlockHeader};
 use alloy_evm::block::StateChangeSource;
 use alloy_primitives::B256;
 use executor::WorkloadExecutor;
 use multiproof::{SparseTrieUpdate, *};
 use parking_lot::RwLock;
 use prewarm::PrewarmMetrics;
-use reth_evm::{ConfigureEvm, OnStateHook, SpecFor};
-use reth_primitives_traits::{NodePrimitives, SealedHeaderFor};
+use reth_evm::{execute::OwnedExecutableTxFor, ConfigureEvm, EvmEnvFor, OnStateHook, SpecFor};
+use reth_primitives_traits::NodePrimitives;
 use reth_provider::{
     providers::ConsistentDbView, BlockReader, DatabaseProviderFactory, StateCommitmentProvider,
     StateProviderFactory, StateReader,
@@ -32,13 +31,10 @@ use reth_trie_sparse::{
     provider::{TrieNodeProvider, TrieNodeProviderFactory},
     ClearedSparseStateTrie, SerialSparseTrie, SparseStateTrie, SparseTrie,
 };
-use std::{
-    collections::VecDeque,
-    sync::{
-        atomic::AtomicBool,
-        mpsc::{self, channel, Sender},
-        Arc,
-    },
+use std::sync::{
+    atomic::AtomicBool,
+    mpsc::{self, channel, Sender},
+    Arc,
 };
 
 use super::precompile_cache::PrecompileCacheMap;
@@ -146,15 +142,15 @@ where
     ///
     /// This returns a handle to await the final state root and to interact with the tasks (e.g.
     /// canceling)
-    pub fn spawn<P>(
+    pub fn spawn<P, T: OwnedExecutableTxFor<Evm>>(
         &mut self,
-        header: SealedHeaderFor<N>,
-        transactions: VecDeque<Recovered<N::SignedTx>>,
+        env: ExecutionEnv<Evm>,
+        transactions: impl ExactSizeIterator<Item = T> + Send + 'static,
         provider_builder: StateProviderBuilder<N, P>,
         consistent_view: ConsistentDbView<P>,
         trie_input: TrieInput,
         config: &TreeConfig,
-    ) -> PayloadHandle
+    ) -> PayloadHandle<T>
     where
         P: DatabaseProviderFactory<Provider: BlockReader>
             + BlockReader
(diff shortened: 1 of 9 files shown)


crates/evm/evm/src/execute.rs
```diff
@@ -552,6 +552,35 @@ where
...
+}
+
+impl<T, Evm: ConfigureEvm> OwnedExecutableTxFor<Evm> for T where
+    T: OwnedExecutableTx<TxEnvFor<Evm>, TxTy<Evm::Primitives>>
+{
+}
+
+/// A helper trait marking a 'static type that can be converted into an [`ExecutableTx`] for block
+/// executor.
+pub trait OwnedExecutableTx<TxEnv, Tx>: Clone + Send + 'static {
```
mattsse, line 569: these make sense

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -418,6 +450,13 @@ impl PayloadHandle {
     pub(super) fn terminate_caching(&mut self, block_output: Option<BundleState>) {
         self.prewarm_handle.terminate_caching(block_output)
     }
+
+    /// Returns iterator yielding transactions from the stream.
+    pub fn iter_transactions(&mut self) -> impl Iterator<Item = Tx> + '_ {
+        core::iter::repeat_with(|| self.transactions.recv())
+            .take_while(|res| res.is_ok())
+            .map(|res| res.unwrap())
```
mattsse, line 460: this makes sense

crates/evm/evm/src/metrics.rs
```diff
@@ -75,20 +75,19 @@ pub struct ExecutorMetrics {
...
+        let (gas_used, output) = f();
         let execution_duration = execute_start.elapsed().as_secs_f64();
 
         // Update gas metrics.
-        self.gas_processed_total.increment(block.header().gas_used());
-        self.gas_per_second.set(block.header().gas_used() as f64 / execution_duration);
-        self.gas_used_histogram.record(block.header().gas_used() as f64);
+        self.gas_processed_total.increment(gas_used);
+        self.gas_per_second.set(gas_used as f64 / execution_duration);
+        self.gas_used_histogram.record(gas_used as f64);
```
mattsse, line 90: perhaps we get rid of this function entirely and inline this instead, but can do that later

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -492,6 +531,60 @@ impl ExecutionCache {
     }
 }
 
+/// Container for receiving end of executable transactions stream.
+#[derive(Debug)]
+pub struct ExecutableTxReceiver<T> {
+    rx: mpsc::Receiver<T>,
+    len: usize,
+}
```
mattsse: I think if we want to do smth like

payload -> (env, iter, remainder)

then this iterator must yield results, because the iter must handle the Bytes -> Tx transformation

mattsse (changes requested):
all of this makes sense, but one limitation that we'll Hit next is that we need to yield Result of tx instead because the inteded operations that we want to yield is Bytes -> Recovered<Tx>

another thing we need to make easier next is the regular execution process that does

payload -> (env, iter) and ideally we can then reuse this create the final block after execution while we're waiting for the state root outcome

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -492,6 +533,50 @@ impl ExecutionCache {
...
+pub struct ExecutionEnv<Evm: ConfigureEvm> {
+    /// Evm environment.
+    pub evm_env: EvmEnvFor<Evm>,
+    /// Hash of the block being executed.
+    pub hash: B256,
+    /// Hash of the parent block.
+    pub parent_hash: B256,
+}
+
+impl<Evm: ConfigureEvm> Default for ExecutionEnv<Evm>
```
mediocregopher, line 547: nit: could we add derive Default?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -290,12 +286,14 @@ where
     ///
     /// Note: Since here are no ordering guarantees this won't the state the txs produce when
     /// executed sequentially.
-    fn transact_batch(self, txs: &[Recovered<N::SignedTx>], sender: Sender<PrewarmTaskEvent>) {
-        let Some((mut evm, evm_config, metrics, terminate_execution)) = self.evm_for_ctx() else {
-            return
-        };
+    fn transact_batch(
```
mediocregopher, line 297: nit: doc for this method is out of date now

crates/engine/tree/src/tree/payload_processor/mod.rs
klkvr: we couldn't because derive will add bound `Evm: Default` which doesn't guarantee `EvmEnvFor<Evm>: Default`

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -71,41 +62,57 @@ where
...
+                    });
+
+                    handles.push(tx);
+                }
 
-        for chunk in &self.pending.drain(..).chunks(chunk_size) {
-            let sender = self.actions_tx.clone();
-            let ctx = self.ctx.clone();
-            let pending_chunk = chunk.collect::<Vec<_>>();
+                let _ = handles[task_idx].send(executable);
```
mattsse: ah I see, this basically does batching via task_idx -> spawned sub task

mattsse (approved):
very nice

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -71,41 +62,65 @@ where
...
-            self.prewarm_outcomes_left += pending_chunk.len();
-            self.executor.spawn_blocking(move || {
-                ctx.transact_batch(&pending_chunk, sender);
-            });
-        }
+                executing += 1;
+            }
+
+            // drop handle and wait for all tasks to finish and drop theirs
+            // todo: do we even need this? tasks are spawned anyway and will see all transactions
```
mattsse: this is fine, we can remove the todo here

we can perhaps get rid of this outer spawn entirely if we track the exit condition differently, and keep track of the finished execution handles at top level

but we can do this in a potential followup

mattsse (approved):
lgtm!

let's do it

PR #17664, opened 2025-07-29
refactor: move BasicEngineValidator creation into EngineApiValidatorBuilder trait

fgimenez:
closes #17659

* removes `EngineValidator` bound from `EngineValidatorAddOn` trait
* introduces dedicated `EngineApiValidatorBuilder` trait
* renames `EngineValidator` to `EngineApiValidator`
* moves `BasicEngineValidator` creation from launch/engine.rs into builder

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -73,7 +73,7 @@ use crate::tree::error::AdvancePersistenceError;
 pub use block_buffer::BlockBuffer;
 pub use invalid_headers::InvalidHeaderCache;
 pub use payload_processor::*;
-pub use payload_validator::{BasicEngineValidator, EngineApiValidator};
+pub use payload_validator::{BasicEngineValidator, EngineValidator};
 pub use persistence_state::PersistenceState;
 pub use reth_engine_primitives::TreeConfig;
 
@@ -300,7 +300,7 @@ where
         BlockReader<Block = N::Block, Header = N::BlockHeader>,
     C: ConfigureEvm<Primitives = N> + 'static,
     T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>,
-    V: EngineApiValidator<T>,
+    V: EngineValidator<T>,
 {
     /// Creates a new [`EngineApiTreeHandler`].
     #[expect(clippy::too_many_arguments)]
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -73,7 +73,7 @@ use crate::tree::error::AdvancePersistenceError;
 pub use block_buffer::BlockBuffer;
 pub use invalid_headers::InvalidHeaderCache;
 pub use payload_processor::*;
-pub use payload_validator::{BasicEngineValidator, EngineValidator};
+pub use payload_validator::{BasicEngineValidator, EngineApiValidator};
 pub use persistence_state::PersistenceState;
 pub use reth_engine_primitives::TreeConfig;
 
@@ -300,7 +300,7 @@ where
         BlockReader<Block = N::Block, Header = N::BlockHeader>,
     C: ConfigureEvm<Primitives = N> + 'static,
     T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>,
-    V: EngineValidator<T>,
+    V: EngineApiValidator<T>,
 {
     /// Creates a new [`EngineApiTreeHandler`].
     #[expect(clippy::too_many_arguments)]
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -41,7 +41,7 @@ use tracing::{debug, error, info, trace, warn};
 
 /// Context providing access to tree state during validation.
 ///
-/// This context is provided to the [`EngineValidator`] and includes the state of the tree's
+/// This context is provided to the [`EngineApiValidator`] and includes the state of the tree's
 /// internals
 pub struct TreeCtx<'a, N: NodePrimitives> {
     /// The engine API tree state
@@ -130,7 +130,7 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
 
 /// A helper type that provides reusable payload validation logic for network-specific validators.
 ///
-/// This type satisfies [`EngineValidator`] and is responsible for executing blocks/payloads.
+/// This type satisfies [`EngineApiValidator`] and is responsible for executing blocks/payloads.
 ///
 /// This type contains common validation, execution, and state root computation logic that can be
 /// used by network-specific payload validators (e.g., Ethereum, Optimism). It is not meant to be
@@ -793,7 +793,7 @@ pub type ValidationOutcome<N, E = InsertBlockError<BlockTy<N>>> =
 /// Type that validates the payloads processed by the engine.
 ///
 /// This provides the necessary functions for validating/executing payloads/blocks.
-pub trait EngineValidator<
+pub trait EngineApiValidator<
     Types: PayloadTypes,
     N: NodePrimitives = <<Types as PayloadTypes>::BuiltPayload as BuiltPayload>::Primitives,
 >: Send + Sync + 'static
@@ -841,7 +841,7 @@ pub trait EngineValidator<
     ) -> ValidationOutcome<N>;
 }
 
-impl<N, Types, P, Evm, V> EngineValidator<Types> for BasicEngineValidator<P, Evm, V>
+impl<N, Types, P, Evm, V> EngineApiValidator<Types> for BasicEngineValidator<P, Evm, V>
 where
     P: DatabaseProviderFactory<Provider: BlockReader>
         + BlockReader<Header = N::BlockHeader>
@@ -878,7 +878,7 @@ where
         ctx: TreeCtx<'_, N>,
     ) -> ValidationOutcome<N, InsertPayloadError<N::Block>> {
         let block = self.validator.ensure_well_formed_payload(payload)?;
-        Ok(EngineValidator::<Types>::validate_block(self, block, ctx)?)
+        Ok(EngineApiValidator::<Types>::validate_block(self, block, ctx)?)
     }
 
     fn validate_block(
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -41,7 +41,7 @@ use tracing::{debug, error, info, trace, warn};
 
 /// Context providing access to tree state during validation.
 ///
-/// This context is provided to the [`EngineApiValidator`] and includes the state of the tree's
+/// This context is provided to the [`EngineValidator`] and includes the state of the tree's
 /// internals
 pub struct TreeCtx<'a, N: NodePrimitives> {
     /// The engine API tree state
@@ -130,7 +130,7 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
 
 /// A helper type that provides reusable payload validation logic for network-specific validators.
 ///
-/// This type satisfies [`EngineApiValidator`] and is responsible for executing blocks/payloads.
+/// This type satisfies [`EngineValidator`] and is responsible for executing blocks/payloads.
 ///
 /// This type contains common validation, execution, and state root computation logic that can be
 /// used by network-specific payload validators (e.g., Ethereum, Optimism). It is not meant to be
@@ -793,7 +793,7 @@ pub type ValidationOutcome<N, E = InsertBlockError<BlockTy<N>>> =
 /// Type that validates the payloads processed by the engine.
 ///
 /// This provides the necessary functions for validating/executing payloads/blocks.
-pub trait EngineApiValidator<
+pub trait EngineValidator<
     Types: PayloadTypes,
     N: NodePrimitives = <<Types as PayloadTypes>::BuiltPayload as BuiltPayload>::Primitives,
 >: Send + Sync + 'static
@@ -841,7 +841,7 @@ pub trait EngineApiValidator<
     ) -> ValidationOutcome<N>;
 }
 
-impl<N, Types, P, Evm, V> EngineApiValidator<Types> for BasicEngineValidator<P, Evm, V>
+impl<N, Types, P, Evm, V> EngineValidator<Types> for BasicEngineValidator<P, Evm, V>
 where
     P: DatabaseProviderFactory<Provider: BlockReader>
         + BlockReader<Header = N::BlockHeader>
@@ -878,7 +878,7 @@ where
         ctx: TreeCtx<'_, N>,
     ) -> ValidationOutcome<N, InsertPayloadError<N::Block>> {
         let block = self.validator.ensure_well_formed_payload(payload)?;
-        Ok(EngineApiValidator::<Types>::validate_block(self, block, ctx)?)
+        Ok(EngineValidator::<Types>::validate_block(self, block, ctx)?)
     }
 
     fn validate_block(
--- a/crates/ethereum/node/src/node.rs
+++ b/crates/ethereum/node/src/node.rs
@@ -248,7 +248,7 @@ where
         Evm: ConfigureEvm<NextBlockEnvCtx = NextBlockEnvAttributes>,
     >,
     EthB: EthApiBuilder<N>,
-    EV: EngineApiValidatorBuilder<N> + Unpin,
+    EV: EngineApiValidatorBuilder<N>,
     EB: EngineApiBuilder<N>,
     EthApiError: FromEvmError<N::Evm>,
     EvmFactoryFor<N::Evm>: EvmFactory<Tx = TxEnv>,
@@ -293,7 +293,7 @@ where
         Evm: ConfigureEvm<NextBlockEnvCtx = NextBlockEnvAttributes>,
     >,
     EthB: EthApiBuilder<N>,
-    EV: EngineApiValidatorBuilder<N> + Unpin,
+    EV: EngineApiValidatorBuilder<N>,
     EB: EngineApiBuilder<N>,
     EthApiError: FromEvmError<N::Evm>,
     EvmFactoryFor<N::Evm>: EvmFactory<Tx = TxEnv>,
@@ -305,8 +305,7 @@ where
     }
 }
 
-impl<N, EthB, EV, EB, RpcMiddleware> EngineValidatorAddOn<N>
-    for EthereumAddOns<N, EthB, EV, EB, RpcMiddleware>
+impl<N, EthB, EV, EB> EngineValidatorAddOn<N> for EthereumAddOns<N, EthB, EV, EB>
 where
     N: FullNodeComponents<
         Types: NodeTypes<
@@ -321,7 +320,6 @@ where
     EB: EngineApiBuilder<N>,
     EthApiError: FromEvmError<N::Evm>,
     EvmFactoryFor<N::Evm>: EvmFactory<Tx = TxEnv>,
-    RpcMiddleware: RethRpcMiddleware,
 {
     type ValidatorBuilder = EV;
 
--- a/crates/ethereum/node/src/node.rs
+++ b/crates/ethereum/node/src/node.rs
@@ -175,7 +175,7 @@ where
 pub struct EthereumAddOns<
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
-    PVB: PayloadValidatorBuilder<N>,
+    PVB,
     EB = BasicEngineApiBuilder<PVB>,
     EVB = BasicEngineValidatorBuilder<PVB>,
     RpcMiddleware = Identity,
@@ -187,7 +187,6 @@ impl<N, EthB, PVB, EB, EVB, RpcMiddleware> EthereumAddOns<N, EthB, PVB, EB, EVB,
 where
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
-    PVB: PayloadValidatorBuilder<N>,
 {
     /// Creates a new instance from the inner `RpcAddOns`.
     pub const fn new(inner: RpcAddOns<N, EthB, PVB, EB, EVB, RpcMiddleware>) -> Self {
@@ -222,7 +221,6 @@ impl<N, EthB, PVB, EB, EVB, RpcMiddleware> EthereumAddOns<N, EthB, PVB, EB, EVB,
 where
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
-    PVB: PayloadValidatorBuilder<N>,
 {
     /// Replace the engine API builder.
     pub fn with_engine_api<T>(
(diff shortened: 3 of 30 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -300,7 +300,7 @@ where
         BlockReader<Block = N::Block, Header = N::BlockHeader>,
     C: ConfigureEvm<Primitives = N> + 'static,
     T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>,
-    V: EngineValidator<T>,
+    V: EngineApiValidator<T>,
```
mattsse: can we now shift this to PayloadValidator?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -793,7 +793,7 @@ pub type ValidationOutcome<N, E = InsertBlockError<BlockTy<N>>> =
 /// Type that validates the payloads processed by the engine.
 ///
 /// This provides the necessary functions for validating/executing payloads/blocks.
-pub trait EngineValidator<
+pub trait EngineApiValidator<
```
mattsse: we should keep `EngineValidator` name here because this is no longer tied to the engine api and then this is also distinct from the `EngineApiValidator`

crates/node/builder/src/rpc.rs
```diff
@@ -1070,6 +1072,67 @@ pub trait EngineApiBuilder<Node: FullNodeComponents>: Send + Sync {
...
+    /// and adds block execution, state validation, and fork handling.
+    fn build_tree_validator<P, Evm>(
+        self,
+        ctx: &AddOnsContext<'_, Node>,
+        provider: P,
+        consensus: Arc<dyn FullConsensus<PrimitivesTy<Node::Types>, Error = ConsensusError>>,
+        evm_config: Evm,
+        tree_config: reth_engine_primitives::TreeConfig,
+        invalid_block_hook: Box<dyn InvalidBlockHook<PrimitivesTy<Node::Types>>>,
+    ) -> impl Future<
```
mattsse: hmm, can we not obtain all of this directly from in build fn?

crates/node/builder/src/rpc.rs
```diff
@@ -1070,6 +1072,67 @@ pub trait EngineApiBuilder<Node: FullNodeComponents>: Send + Sync {
...
+/// Builder trait for creating validators specifically for the Engine API.
+///
+/// This trait is responsible for building validators that the Engine API will use
+/// to validate payloads. Unlike the general `EngineValidatorBuilder`, this is
+/// specifically designed for Engine API validation requirements.
+pub trait EngineApiValidatorBuilder<Node: FullNodeComponents>: Send + Sync + Clone {
+    /// The validator type that will be used by the Engine API.
+    type Validator: EngineApiValidator<<Node::Types as NodeTypes>::Payload>
+        + PayloadValidator<<Node::Types as NodeTypes>::Payload, Block = BlockTy<Node::Types>>
+        + Clone;
```
mattsse: here we only need the `PayloadValidator` trait bound to entirely separate this from engineapi

mattsse (changes requested):
some suggestions, what we need to do here is entirely separating the `PayloadValidator` from the engineapi.

naming is a bit rough here, but the `EngineApiValidator` trait is now only required for setting up the engine api, and we should move that there:

https://github.com/paradigmxyz/reth/blob/e38e247b408bc6ad1e55f4d4f7af5b1fdced8948/crates/node/builder/src/rpc.rs#L1080-L1080

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: makes sense done, i misunderstood a comment in the issue

crates/node/builder/src/rpc.rs
fgimenez: good point, done and cleaned up the bounds

crates/engine/tree/src/tree/mod.rs
fgimenez: the `validate_block` and `validate_payload` methods we need are in `EngineValidator`, am i missing something?

crates/node/builder/src/rpc.rs
fgimenez: ok, moved the bound to the implementations, ptal

crates/node/builder/src/launch/engine.rs
```diff
@@ -191,15 +190,23 @@ where
...
+            .await?;
 
         let consensus_engine_stream = UnboundedReceiverStream::from(consensus_engine_rx)
             .maybe_skip_fcu(node_config.debug.skip_fcu)
             .maybe_skip_new_payload(node_config.debug.skip_new_payload)
             .maybe_reorg(
                 ctx.blockchain_db().clone(),
                 ctx.components().evm_config().clone(),
-                engine_payload_validator.clone(),
+                payload_validator,
```
mattsse: oh I see we still need this here...

let's see if we can get rid of this

crates/node/builder/src/rpc.rs
```diff
@@ -987,56 +996,106 @@ pub trait EthApiBuilder<N: FullNodeComponents>: Default + Send + 'static {
...
+    ) -> impl Future<Output = eyre::Result<Self::EngineApi>> + Send;
+}
+
+/// Builder trait for creating validators specifically for the Engine API.
+///
+/// This trait is responsible for building validators that the Engine API will use
+/// to validate payloads.
+pub trait EngineApiValidatorBuilder<Node: FullNodeComponents>: Send + Sync + Clone {
+    /// The validator type that will be used by the Engine API.
+    type Validator: PayloadValidator<
```
klkvr: should this build the `EngineApiValidator`?

crates/node/builder/src/rpc.rs
```diff
@@ -987,56 +996,106 @@ pub trait EthApiBuilder<N: FullNodeComponents>: Default + Send + 'static {
...
+            Ok(reth_engine_tree::tree::BasicEngineValidator::new(
+                ctx.node.provider().clone(),
+                Arc::new(ctx.node.consensus().clone()),
+                ctx.node.evm_config().clone(),
+                inner_validator,
+                tree_config,
+                invalid_block_hook,
+            ))
+        }
+    }
```
klkvr: I think what we want here is to erase the `BasicEngineValidator` type and turn it into AT instead

crates/node/builder/src/rpc.rs
```diff
@@ -987,56 +996,106 @@ pub trait EthApiBuilder<N: FullNodeComponents>: Default + Send + 'static {
...
+        ctx: &AddOnsContext<'_, Node>,
+        tree_config: TreeConfig,
+    ) -> impl Future<
+        Output = eyre::Result<BasicEngineValidator<Node::Provider, Node::Evm, Self::Validator>>,
+    > + Send
+    where
+        Node::Provider: DatabaseProviderFactory<Provider: BlockReader>
+            + StateReader
+            + StateCommitmentProvider
+            + HashedPostStateProvider,
```
klkvr: this bound should not be needed

crates/node/builder/src/launch/engine.rs
klkvr: we should be able to, because `BasicEngineValidator` already exposes the `ensure_well_formed_payload` which I think is the only method we use there

klkvr (changes requested):

crates/node/builder/src/rpc.rs
fgimenez: good point, we needed here a `PayloadValidator` for `maybe_reorg`, i have removed that constraint and changed this to `EngineApiValidator`

crates/node/builder/src/rpc.rs
fgimenez: ok, removed from here but added to the concrete impls, ptal

crates/node/builder/src/rpc.rs
fgimenez: indeed, done, we now have a `TreeValidator` AT and the default impl is removed, each implementor creates its own

crates/node/builder/src/launch/engine.rs
fgimenez: ok, i've adapted the reorg functionality to use `EngineValidator` instead of `PayloadValidator`, ptal

crates/exex/test-utils/src/lib.rs
```diff
@@ -124,6 +124,11 @@ impl NodeTypes for TestNode {
 impl<N> Node<N> for TestNode
 where
     N: FullNodeTypes<Types = Self>,
+    N::Provider: reth_provider::BlockReader
+        + reth_provider::StateReader
+        + reth_provider::StateCommitmentProvider
+        + reth_provider::HashedPostStateProvider,
+    <N::Provider as reth_provider::DatabaseProviderFactory>::Provider: reth_provider::BlockReader,
```
klkvr: this bound should not be needed, if something is missing we should add it to `FullProvider` instead

examples/custom-node/src/lib.rs
```diff
@@ -54,6 +60,8 @@ impl NodeTypes for CustomNode {
 impl<N> Node<N> for CustomNode
 where
     N: FullNodeTypes<Types = Self>,
+    N::Provider: BlockReader + StateReader + StateCommitmentProvider + HashedPostStateProvider,
+    <N::Provider as DatabaseProviderFactory>::Provider: BlockReader,
```
klkvr: same here

examples/custom-node/src/engine.rs
```diff
@@ -296,16 +305,38 @@ pub enum CustomError {
 #[non_exhaustive]
 pub struct CustomEngineValidatorBuilder;
 
-impl<N> EngineValidatorBuilder<N> for CustomEngineValidatorBuilder
+impl<N> EngineApiValidatorBuilder<N> for CustomEngineValidatorBuilder
 where
     N: FullNodeComponents<Types = CustomNode>,
+    N::Provider: StateReader + StateCommitmentProvider + HashedPostStateProvider,
+    <N::Provider as DatabaseProviderFactory>::Provider: BlockReader,
```
klkvr: and here

crates/node/builder/src/rpc.rs
```diff
@@ -1070,6 +1031,39 @@ pub trait EngineApiBuilder<Node: FullNodeComponents>: Send + Sync {
...
+/// to validate payloads.
+pub trait EngineApiValidatorBuilder<Node: FullNodeComponents>: Send + Sync + Clone {
+    /// The validator type that will be used by the Engine API.
+    type Validator: EngineApiValidator<<Node::Types as NodeTypes>::Payload>;
+
+    /// The tree validator type that will be used by the consensus engine.
+    type TreeValidator: EngineValidator<
+        <Node::Types as NodeTypes>::Payload,
+        <Node::Types as NodeTypes>::Primitives,
+    >;
```
klkvr: it feels wrong that `EngineApi` validator (which is optional) coupled with the core `TreeValidator` component, shouldn't those be separate?

crates/ethereum/node/src/node.rs
```diff
@@ -173,71 +179,105 @@ where
 pub struct EthereumAddOns<
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
-    EV,
-    EB = BasicEngineApiBuilder<EV>,
+    EV: Send,
```
klkvr: why do we need this generic now?

klkvr (changes requested):

crates/exex/test-utils/src/lib.rs
fgimenez: sure done

crates/node/builder/src/rpc.rs
fgimenez: `TreeValidator` is the AT just introduced to avoid building the concrete `BasicEngineValidator`, any suggestions to improve here are very welcome!

crates/ethereum/node/src/node.rs
fgimenez: yep, this was introduced to decouple `EV` from `EB` lmk wdyt

crates/node/builder/src/rpc.rs
klkvr: the `EngineApiValidatorBuilder` should only be responsible for building the `Validator`, `TreeValidator` should be built by a separate `EngineValidatorBuilder` with default impl being smth like `BasicEngineValidatorBuilder<EV>` where `EV` is `PayloadValidatorBuilder`

so what we can do is rename `EngineApiValidatorBuilder` to `PayloadValidatorBuilder`. And have `RpcAddOns` look like this:
```rust
RpcAddOns<Node, EthB, EV, EngineApi = BasicEngineApiBuilder<EV>, Engine = BasicEngineValidatorBuilder<EV>,  RpcMiddleware = Identity>
```

then `EV` will be required to implement `PayloadValidatorBuilder`, and `BasicEngineApiBuilder` will expect it to be smth like `PayloadValidatorBuilder<Validator: EngineApiValidator>`

fgimenez: makes total sense, done ptal

crates/ethereum/node/src/node.rs
klkvr: what will break if we will remove it? addons should only be generic over builders, not the types being built

fgimenez: good point, now it can be completely removed, done

crates/ethereum/node/src/node.rs
```diff
@@ -173,71 +175,94 @@ where
 pub struct EthereumAddOns<
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
-    EV,
-    EB = BasicEngineApiBuilder<EV>,
+    PVB: PayloadValidatorBuilder<N>,
```
klkvr:
```suggestion
    PVB,
```

let's not require this bound here

crates/optimism/node/src/node.rs
```diff
@@ -270,11 +272,17 @@ impl NodeTypes for OpNode {
 /// This type provides optimism-specific addons to the node and exposes the RPC server and engine
 /// API.
 #[derive(Debug)]
-pub struct OpAddOns<N: FullNodeComponents, EthB: EthApiBuilder<N>, EV, EB, RpcMiddleware = Identity>
-{
+pub struct OpAddOns<
+    N: FullNodeComponents,
+    EthB: EthApiBuilder<N>,
+    PVB: PayloadValidatorBuilder<N>,
```
klkvr:
```suggestion
    PVB,
```

crates/optimism/node/src/node.rs
```diff
@@ -326,16 +362,17 @@ where
     }
 }
 
-impl<N, EthB, EV, EB, RpcMiddleware> OpAddOns<N, EthB, EV, EB, RpcMiddleware>
+impl<N, EthB, PVB, EB, EVB, RpcMiddleware> OpAddOns<N, EthB, PVB, EB, EVB, RpcMiddleware>
 where
     N: FullNodeComponents,
     EthB: EthApiBuilder<N>,
+    PVB: PayloadValidatorBuilder<N>,
```
klkvr:
```suggestion
```

crates/optimism/node/src/node.rs
```diff
@@ -344,23 +381,27 @@ where
...
-    pub fn with_engine_validator<T>(
+    /// Maps the [`PayloadValidatorBuilder`] builder type.
+    pub fn with_payload_validator<T>(
         self,
-        engine_validator_builder: T,
-    ) -> OpAddOns<N, EthB, T, EB, RpcMiddleware> {
+        payload_validator_builder: T,
+    ) -> OpAddOns<N, EthB, T, EB, EVB, RpcMiddleware>
+    where
+        T: PayloadValidatorBuilder<N>,
```
klkvr:
```suggestion
```
let's remove all similar bounds as well

examples/custom-node/src/lib.rs
```diff
@@ -82,6 +83,13 @@ where
...
     fn add_ons(&self) -> Self::AddOns {
-        self.inner.add_ons_builder().build()
+        self.inner
+            .add_ons_builder::<CustomRpcTypes>()
+            .build::<
+                NodeAdapter<N>,
+                CustomEngineValidatorBuilder,
+                CustomEngineApiBuilder,
+                BasicEngineValidatorBuilder<CustomEngineValidatorBuilder>,
+            >()
```
klkvr: do we need this annotation now?

crates/node/builder/src/rpc.rs
```diff
@@ -1070,38 +1104,134 @@ pub trait EngineApiBuilder<Node: FullNodeComponents>: Send + Sync {
     ) -> impl Future<Output = eyre::Result<Self::EngineApi>> + Send;
 }
 
+/// Builder trait for creating payload validators specifically for the Engine API.
+///
+/// This trait is responsible for building validators that the Engine API will use
+/// to validate payloads.
+pub trait PayloadValidatorBuilder<Node: FullNodeComponents>: Send + Sync + Clone {
+    /// The validator type that will be used by the Engine API.
+    type Validator: EngineApiValidator<<Node::Types as NodeTypes>::Payload>;
```
klkvr: this should be `Validator: PayloadValidator`, `EngineApiValidator` we can enforce on `BasicEngineApiBuilder`

crates/node/builder/src/launch/engine.rs
```diff
@@ -191,15 +190,34 @@ where
...
+        #[expect(clippy::if_then_some_else_none)]
+        let reorg_validator = if node_config.debug.reorg_frequency.is_some() {
+            Some(
+                validator_builder
+                    .build_tree_validator(&add_ons_ctx, engine_tree_config.clone())
+                    .await?,
+            )
+        } else {
+            None
+        };
```
klkvr: this is a bit confusing but I see why we need this, what we could do is instead have a closure `impl FnOnce() -> Validator` and pass it into `maybe_reorg` to avoid the `Option` stuff

crates/node/builder/src/rpc.rs
```diff
@@ -1070,38 +1104,134 @@ pub trait EngineApiBuilder<Node: FullNodeComponents>: Send + Sync {
...
+    ) -> impl Future<Output = eyre::Result<Self::Validator>> + Send;
+}
+
+/// Builder trait for creating engine validators for the consensus engine.
+///
+/// This trait is responsible for building validators that the consensus engine will use
+/// for block execution, state validation, and fork handling.
+pub trait EngineValidatorBuilder<Node: FullNodeComponents>: Send + Sync + Clone {
+    /// The tree validator type that will be used by the consensus engine.
+    type TreeValidator: EngineValidator<
```
klkvr: should this be called `EngineValidator`/`Validator`?

crates/ethereum/node/src/node.rs
fgimenez: done, i've removed all the occurrences where possible and relaxed to `Send` where not

crates/node/builder/src/launch/engine.rs
fgimenez: ok added closure ptal

crates/node/builder/src/rpc.rs
fgimenez: good point, changed

crates/node/builder/src/rpc.rs
fgimenez: indeed, `EngineValidator` is more appropriate imo, changed

examples/custom-node/src/lib.rs
fgimenez: nope! removed :smile:

crates/node/builder/src/launch/engine.rs
```diff
@@ -191,15 +190,29 @@ where
...
                 ctx.components().evm_config().clone(),
-                engine_payload_validator.clone(),
+                || {
+                    // This closure is only called when reorg frequency is Some
+                    futures::executor::block_on(
+                        validator_builder
+                            .build_tree_validator(&add_ons_ctx, engine_tree_config.clone()),
+                    )
+                    .expect("Failed to build tree validator for reorg")
+                },
```
klkvr: we could make `maybe_reorg` async and just pass the future instead of doing `block_on`

mattsse: yeah this seems better

mattsse: alternatively we could just use tokio::block_in_place perhaps

examples/engine-api-access/src/main.rs
```diff
@@ -35,7 +33,7 @@ async fn main() {
         .with_database(db)
         .with_types::<OpNode>()
         .with_components(op_node.components())
-        .with_add_ons(OpAddOns::default().with_engine_api(engine_api))
+        .with_add_ons(op_node.add_ons().with_engine_api(engine_api))
```
klkvr: does `Default` no longer work here?

klkvr (approved):
lgtm, pending reorg future nit

examples/engine-api-access/src/main.rs
fgimenez: no, i get these errors:
```
error[E0277]: the trait bound `Identity: EngineValidatorBuilder<NodeAdapter<..., ...>>` is not satisfied
   --> examples/engine-api-access/src/main.rs:36:23
    |
 36 |         .with_add_ons(OpAddOns::default().with_engine_api(engine_api))
    |          ------------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound
    |          |
    |          required by a bound introduced by this call
    |
```
and 
```
error[E0599]: the method `on_component_initialized` exists for struct `NodeBuilderWithComponents<FullNodeTypesAdapter<OpNode, Arc<...>, ...>, ..., ...>`, but its trait bounds were not satisfied
   --> examples/engine-api-access/src/main.rs:37:10
    |
 32 |       let _builder = NodeBuilder::new(config)
    |  ____________________-
 33 | |         .with_database(db)
 34 | |         .with_types::<OpNode>()
 35 | |         .with_components(op_node.components())
 36 | |         .with_add_ons(OpAddOns::default().with_engine_api(engine_api))
 37 | |         .on_component_initialized(move |ctx| {
    | |         -^^^^^^^^^^^^^^^^^^^^^^^^ method cannot be called due to unsatisfied trait bounds
    | |_________|
```

crates/node/builder/src/launch/engine.rs
fgimenez: ok made `maybe_reorg` async ptal

crates/optimism/node/src/node.rs
```diff
@@ -270,11 +272,17 @@ impl NodeTypes for OpNode {
...
 /// API.
 #[derive(Debug)]
-pub struct OpAddOns<N: FullNodeComponents, EthB: EthApiBuilder<N>, EV, EB, RpcMiddleware = Identity>
-{
+pub struct OpAddOns<
+    N: FullNodeComponents,
+    EthB: EthApiBuilder<N>,
+    PVB,
+    EB = OpEngineApiBuilder<PVB>,
+    EVB = BasicEngineValidatorBuilder<PVB>,
```
mattsse, line 280: hmm, why do we need three now?

isnt payload and engine decoupled now, so we should only need two?

klkvr: yep, we need 2 builders for engine api and engine validator, and one more that builds inner validators for them

examples/engine-api-access/src/main.rs
klkvr: fixed here 362b4aa55

crates/optimism/node/src/node.rs
mattsse: I see

examples/engine-api-access/src/main.rs
mattsse: I think this was just 1 missing default generic

mattsse (approved):
lgtm

I think we can expect a few more changes around this but this is a change that we should make because unlocks a few more things

PR #17666, opened 2025-07-29
Add feature‑gated `blob` flag and empty `BlobPoolExt` trait (step 1/3 for #294)

chainguru1014:
### Motivation
Downstream L2s (RISE, OP, …) disable blob txs; keeping nine blob‑specific
helpers in `TransactionPool` forces them to compile EIP‑4844 types.

### What changed
* new Cargo feature **`blob`** in `reth‑transaction‑pool` (enabled by default)
* introduced empty marker trait **`BlobPoolExt`**
* moved nothing yet – purely preparatory.

No functional change; all tests/benches unchanged.

### Migration
_No migration needed (no public item removal)._

### DIFF
--- a/crates/transaction-pool/Cargo.toml
+++ b/crates/transaction-pool/Cargo.toml
@@ -73,6 +73,8 @@ serde_json.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread"] }
 
 [features]
+blob = []
+default = ["serde", "blob"]
 serde = [
     "dep:serde",
     "reth-execution-types/serde",
--- a/crates/transaction-pool/benches/canonical_state_change.rs
+++ b/crates/transaction-pool/benches/canonical_state_change.rs
@@ -75,8 +75,7 @@ fn canonical_state_change_bench(c: &mut Criterion) {
             let total_txs = num_senders * txs_per_sender;
 
             let group_id = format!(
-                "txpool | canonical_state_change | senders: {} | txs_per_sender: {} | total: {}",
-                num_senders, txs_per_sender, total_txs
+                "txpool | canonical_state_change | senders: {num_senders} | txs_per_sender: {txs_per_sender} | total: {total_txs}"
             );
 
             // Create the update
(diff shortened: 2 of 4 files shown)


mattsse (commented):
I dont think this is worth doing, because you could just impl these as noops

what's the context for this change?

chainguru1014:
Thanks for the quick look, @mattsse – happy to clarify the motivation.
(Let me know if a different direction would be preferable.)

Context / pain‑point
Down‑stream use‑case: RISE, OP Stack and several other L2s compile reth with --no-default-features because they currently do not accept EIP‑4844 blob transactions on their canonical chain.

In our Docker images we disable everything blob‑related to shave off artefacts we don’t need.

The nine blob helpers in TransactionPool reference alloy_eips::{eip4844, eip7594, …}.
Even if we implement them as no‑ops, those types remain in the public surface, so crates that only depend on reth‑transaction‑pool still have to pull in 30‑ish additional crates (kzg, ark‑*, etc.).

Compile‑time overhead: On CI the extra deps add ≈ 40 s build time and ~120 MiB incremental target dir size for us.

image


So the split‑trait approach is the only one that removes the heavy deps yet leaves the original TransactionPool untouched for code that doesn’t care.

Why submit in three small PRs
PR 1 (this) – introduce the gated feature and empty trait (no functional change).

PR 2 – move the nine helpers behind #[cfg(feature="blob")] + blanket‑impl impl<T: TransactionPool> BlobPoolExt for T.

PR 3 – migrate internal call‑sites (engine‑api, payload‑builder, tests) to add the bound only where blobs are actually used.

Keeping them incremental should make review & bisect easier; happy to squash if you prefer a single patch.

Open to alternatives
If you feel the compile‑time wins don’t justify the additional trait, we can:

prototype an alternative with cfg(blob) on the existing methods (breaking but gated), or

close this and keep patching downstream.

Let me know which direction you’d prefer – I’m happy to adjust.

Thanks again for the review!

PR #17669, opened 2025-07-29
fix: RPC: feeHistoryEntry should return 0.0 when blob_params.max_blob_count is zero

acerone85:
The documentation of `FeeHistoryEntry` specifies that the `blob_gas_used_ratio` should be `0.0` when there are no blobs in a block. 

https://github.com/paradigmxyz/reth/blob/c45b1cd314f69281115fab3ac736d1584f65e089/crates/rpc/rpc-eth-types/src/fee_history.rs#L352-L354

However, some custom EVM chain disable blobs by specifying a value of `0` for `max_blob_count` in the genesis file's `blobSchedules`. In this case, calling the `eth_feeHistory` method will return a list of `feeHistoryEntry` whose `blob_gas_used_ratio` is `NaN`, serialized to `null`. 

For those chains, this has the impact of having foundry script deployments failing with the following error: 

```
Error: Failed to get EIP-1559 fees
``` 

This PR fixes the impementation of `FeeHistoryEntry` so that when `blob_params.max_blob_count` is `0`, then `blob_gas_used_ratio` is set to `0.0` rather than `NaN`, as per the `FeeHistoryEntry` documentation.

A unit test `fee_history::tests::test_fee_history_entries_when_max_blob_count_is_zero` is provided to validate the fix.

### DIFF
--- a/crates/rpc/rpc-eth-types/src/fee_history.rs
+++ b/crates/rpc/rpc-eth-types/src/fee_history.rs
@@ -412,35 +412,3 @@ where
         })
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use alloy_eips::eip7840::BlobParams;
-    use alloy_primitives::B256;
-    use reth_chain_state::{test_utils::TestBlockBuilder, EthPrimitives};
-
-    #[test]
-    fn test_fee_history_entries_when_max_blob_count_is_zero() {
-        // Arrange
-
-        let blob_params = BlobParams {
-            target_blob_count: 0,
-            max_blob_count: 0,
-            update_fraction: 1,
-            min_blob_fee: 0,
-            max_blobs_per_tx: 0,
-        };
-
-        let mut test_block_builder: TestBlockBuilder<EthPrimitives> = TestBlockBuilder::default();
-
-        // The block will be generated with no blob gas used
-        let block = test_block_builder.generate_random_block(1u64, B256::ZERO);
-
-        // Act
-        let fee_history_enty =
-            super::FeeHistoryEntry::new(&block.sealed_block(), Some(blob_params.clone()));
-
-        // Assert
-        assert_eq!(fee_history_enty.blob_gas_used_ratio, 0.0);
-    }
-}
--- a/crates/rpc/rpc-eth-types/src/fee_history.rs
+++ b/crates/rpc/rpc-eth-types/src/fee_history.rs
@@ -409,3 +409,35 @@ where
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use alloy_eips::eip7840::BlobParams;
+    use alloy_primitives::B256;
+    use reth_chain_state::{test_utils::TestBlockBuilder, EthPrimitives};
+
+    #[test]
+    fn test_fee_history_entries_when_max_blob_count_is_zero() {
+        // Arrange
+
+        let blob_params = BlobParams {
+            target_blob_count: 0,
+            max_blob_count: 0,
+            update_fraction: 1,
+            min_blob_fee: 0,
+            max_blobs_per_tx: 0,
+        };
+
+        let mut test_block_builder: TestBlockBuilder<EthPrimitives> = TestBlockBuilder::default();
+
+        // The block will be generated with no blob gas used
+        let block = test_block_builder.generate_random_block(1u64, B256::ZERO);
+
+        // Act
+        let fee_history_enty =
+            super::FeeHistoryEntry::new(&block.sealed_block(), Some(blob_params.clone()));
+
+        // Assert
+        assert_eq!(fee_history_enty.blob_gas_used_ratio, 0.0);
+    }
+}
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -409,3 +415,35 @@ where
...
+            max_blob_count: 0,
+            update_fraction: 1,
+            min_blob_fee: 0,
+            max_blobs_per_tx: 0,
+        };
+
+        let mut test_block_builder: TestBlockBuilder<EthPrimitives> = TestBlockBuilder::default();
+
+        // The block will be generated with no blob gas used
+        let block = test_block_builder.generate_random_block(1u64, B256::ZERO);
```
acerone85: Asking for an opinion to the maintainers: I could not find a more explicit way to generate a sealed block with `0` blob count. Is there a better way to do it?

crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -376,12 +376,18 @@ where
             base_fee_per_blob_gas: header
                 .excess_blob_gas()
                 .and_then(|excess_blob_gas| Some(blob_params?.calc_blob_fee(excess_blob_gas))),
-            blob_gas_used_ratio: block.body().blob_gas_used() as f64 /
-                blob_params
-                    .as_ref()
-                    .map(|params| params.max_blob_gas_per_block())
```
mattsse, line 382: right this can end up being 0.0

crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -376,12 +376,18 @@ where
...
+                // equal to zero.
+                0.0
+            } else {
+                block.body().blob_gas_used() as f64 /
+                    blob_params
+                        .as_ref()
+                        .map(|params| params.max_blob_gas_per_block())
+                        .unwrap_or(alloy_eips::eip4844::MAX_DATA_GAS_PER_BLOCK_DENCUN)
+                        as f64
+            },
```
mattsse: we have the same thing here once more:

https://github.com/paradigmxyz/reth/blob/e38e247b408bc6ad1e55f4d4f7af5b1fdced8948/crates/rpc/rpc-eth-api/src/helpers/fee.rs#L189-L190

what we can do is add a smol helper function for this that checks the max_blob_gas_per_block for 0.0

like fn checked_gas_used_ratio

crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -409,3 +415,35 @@ where
...
+    fn test_fee_history_entries_when_max_blob_count_is_zero() {
+        // Arrange
+
+        let blob_params = BlobParams {
+            target_blob_count: 0,
+            max_blob_count: 0,
+            update_fraction: 1,
+            min_blob_fee: 0,
+            max_blobs_per_tx: 0,
+        };
```
mattsse: I think this test setup is a bit uneccessary 

if we introduce a helper fn for this in here:

https://github.com/paradigmxyz/reth/blob/e38e247b408bc6ad1e55f4d4f7af5b1fdced8948/crates/rpc/rpc-eth-types/src/utils.rs#L1-L1

then we only need to test this function

mattsse (changes requested):
nice find, tiny suggestion

crates/rpc/rpc-eth-types/src/fee_history.rs
acerone85: Done in 325dd2769. I did not return `Option<f64>` even though the function starts with `checked`. Happy to change either the signature or name if needed. 

Also in the case  of a division `n/0` where `n != 0`, a value of infinity will be returned and `null` will be returned in the rpc fee_history entries. This makes sense to me as it should never be the case that `blob_gas_used` is greater than `max_blob_gas_per_block`, but I am happy to change it if needed.

crates/rpc/rpc-eth-types/src/fee_history.rs
acerone85: Done in ce2a5d135

acerone85:
Thanks for the review @mattsse. I have addressed the comments

crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -159,7 +161,7 @@ where
                 .collect::<Vec<_>>();
 
             if result.is_empty() {
-                return None
+                return None;
```
mattsse:
```suggestion
                return None
```

crates/rpc/rpc-eth-types/src/fee_history.rs
```diff
@@ -323,7 +325,7 @@ where
         // Empty blocks should return in a zero row
         if transactions.is_empty() {
             rewards_in_block.push(0);
-            continue
+            continue;
```
mattsse:
```suggestion
            continue
```

mattsse (approved):
that should do it, ty!

mattsse:
@acerone85 last nits,

unfortunately can't apply them myself because opened from org fork

acerone85:
Tried to make the changes needed to make the CI pass. Unfortunately I can't see the output from CI runs (nor I would expect to), and I tried to run the CI locally with `act` but my laptop was struggling. 

Let me know if there are any changes needed and I'll apply them (or I can cherry-pick the fix commit on top of my branch if you have one)

Issue #12811, opened 2024-11-23
RPC: Batch tx submissions to insert into the mempool together


hai-rise:
### Describe the feature

Currently, the RPC server submits each transaction individually which competes for a write lock of the pool. This can be very congested for performance chains that expect 100k+ new transactions every second. Even more challenging if the block time is small -- canonical state changes also compete for the write lock several times a second.

Ideally, the RPC server would buffer transactions to insert in a batch when the pool write lock is unavailable, instead of competing for the write lock (and block anyway) for every insertion request.

### Additional context

Related: #12806.

hai-rise:
> Can you assign me to this :)

cc @mattsse 

Also, a kind reminder that #12806 is ready and is blocking this 🙏.

mattsse:
kinda related #13165

imo this is a non-goal for the regular ethapi impl, but this behaviour can already be modified by overriding the ethapi traits (see OpEthApi), although this requires some additional impls.

but we could come up with a way to make this easier to modify

kdonthi:
Just wondering, @hai-rise @mattsse is there anything I need to consider extra before working on this?

gakonst:
I think adding a queued between the EthApi impl and the txpool makes sense personally -- open minded to whether this is a parking spot before the txpool or after the ethapi -- could see it both ways.

hai-rise:
> Just wondering, [@hai-rise](https://github.com/hai-rise) [@mattsse](https://github.com/mattsse) is there anything I need to consider extra before working on this?

Oops sorry I missed this. The decision isn't final in Reth and we have fixed it downstream so overall there should be more high-impact and [newcomer-friendly issues](https://github.com/paradigmxyz/reth/issues?q=is%3Aissue%20state%3Aopen%20label%3AD-good-first-issue) for you 🙏. 

> open minded to whether this is a parking spot before the txpool or after the ethapi -- could see it both ways.

We originally proposed the RPC server but eventually buffered transactions at `impl<V, T> TransactionPool for RisePool<V, T>` as we've customized the pool anyway and the code change is pretty small. 

> allows controlling the flow so that we don't overwhelm any downstream component

[...]

PR #17670, opened 2025-07-29
feat(txpool): Batch insertions into the Tx Pool

0xKitsune:
Addresses #12811

This PR adds support for batch transaction insertions into the TxPool to reduce lock contention and improve throughput under high-load scenarios.

When enabled, transactions are funneled through a `TxBatcher`, which accumulates txs in an internal buffer. Insertions are triggered either on a fixed interval or when a threshold is reached, whichever comes first. This avoids excessive locking and per-tx overhead during highly concurrent `send_raw_transaction` calls.

The Batcher configuration is exposed via CLI flags:
- `--txpool.max-batch-size`: Enables batch insertion and sets the maximum number of transactions that will be processed in a batch.

When batching is enabled, the `EthApi` routes transactions through the batcher. Otherwise, it will falls back to the default logic executing each tx indivudally.

Internally, the batcher uses a channel and a background task to process batches. Batched transactions are marked as `TransactionOrigin::Local` and validated before being added.

Opening a draft PR for initial feedback. I am still writing benchmarks as well as adding logging/metrics. Also happy to adjust if there is a preferred approach different than the current implementation.


crates/rpc/rpc/src/eth/core.rs
```diff
@@ -290,6 +302,9 @@ pub struct EthApiInner<N: RpcNodeCore, Rpc: RpcConvert> {
 
     /// Builder for pending block environment.
     next_env_builder: Box<dyn PendingEnvBuilder<N::Evm>>,
+
+    /// Optional transaction batcher for batching tx insertions
+    tx_batcher: Option<TxBatcher<N::Pool>>,
```
mattsse: maybe we just make this behaviour default, wdyt @klkvr

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    /// Process batch transaction insertions
+    pub async fn process_batches(
+        pool: Pool,
+        batch_interval: Duration,
+        batch_threshold: usize,
+        pending_count: Arc<AtomicU64>,
+        mut request_rx: mpsc::Receiver<BatchTxRequest<Pool::Transaction>>,
+    ) {
+        let mut interval = tokio::time::interval(batch_interval);
```
mattsse: I'd like to write this as a native future, then we can get rid of the batch_interval more easily, because we can simply poll the receivers until either drained or we reached the threshold, this would be easier by handrolling the future than in async code

crates/transaction-pool/src/traits.rs
```diff
@@ -1631,7 +1631,7 @@ impl<Tx: PoolTransaction> Stream for NewSubpoolTransactionStream<Tx> {
             match ready!(self.st.poll_recv(cx)) {
                 Some(event) => {
                     if event.subpool == self.subpool {
-                        return Poll::Ready(Some(event))
+                        return Poll::Ready(Some(event));
```
mattsse:
```suggestion
                        return Poll::Ready(Some(event))
```

crates/transaction-pool/src/traits.rs
```diff
@@ -1614,7 +1614,7 @@ impl<Tx: PoolTransaction> NewSubpoolTransactionStream<Tx> {
             match self.st.try_recv() {
                 Ok(event) => {
                     if event.subpool == self.subpool {
-                        return Ok(event)
+                        return Ok(event);
```
mattsse:
```suggestion
                        return Ok(event)
```

crates/transaction-pool/src/lib.rs
```diff
@@ -499,7 +501,7 @@ where
         transactions: Vec<Self::Transaction>,
     ) -> Vec<PoolResult<AddedTransactionOutcome>> {
         if transactions.is_empty() {
-            return Vec::new()
+            return Vec::new();
```
mattsse:
```suggestion
            return Vec::new()
```

mattsse (changes requested):

crates/transaction-pool/src/batcher.rs
0xKitsune: Ive updated the PR to process batches as a native future as well as simplify some of the logic. Let me know if this is the direction you had in mind or if anything else should change.

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,273 @@
...
+    Pool::Transaction: Send + Sync,
+{
+    type Output = ();
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
+
+        loop {
+            // Drain all available requests from the receiver
+            let mut batch = Vec::with_capacity(*this.channel_buffer_size);
```
mattsse: I think we can get away with allocating only a few slots here maybe just two, 1 should also be fine

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,273 @@
...
+
+/// A single batch transaction request
+/// All transactions processed through the batcher are considered local
+/// transactions (`TransactionOrigin::Local`) when inserted into the pool.
+#[derive(Debug)]
+pub struct BatchTxRequest<T: PoolTransaction> {
+    /// Tx to be inserted in to the pool
+    pool_tx: T,
+    /// Channel to send result back to caller
+    response_tx: oneshot::Sender<Result<B256, BatchTxError>>,
```
mattsse: we kinda need two layers here one for the actual Resul<AddedTransactionOutcome, PoolError> and then another outer layer for the case the channel closed, which we could avoid if we copy the hash so we can manually create a poolerror, but this feels a bit uneccesary, so I think we should use `Result<Result` or introduce another error type that unifies BatchTxError and PoolErr

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,273 @@
...
+};
+use tokio::sync::{mpsc, oneshot};
+
+/// Configuration for tx pool batch insertion
+#[derive(Debug, Clone)]
+pub struct BatchTxConfig {
+    /// Channel buffer size for incoming batch tx requests
+    pub channel_buffer_size: usize,
+    /// Maximum number of transactions to batch insert at a time
+    pub max_batch_size: usize,
```
mattsse: I think we only really need the batch size setting here?

the buffer size seems not redundant, because I think we can use an unbounded channel here, because the tx is already available on the request level and restricting the channel size doesnt really restrict anything if we drain + spawn, so we can use an unbounded channel instead

mattsse (changes requested):
a few more suggestions

crates/transaction-pool/src/batcher.rs
0xKitsune: Sounds good, I have updated to sue an unbounded channel.

crates/transaction-pool/src/batcher.rs
0xKitsune: I updated the PR to remove `BatchTxError` and send `Result<AddedTransactionOutcome, PoolError>` through the `response_tx` directly. I also updated  `EthApiError` to include `BatchTxSendError` and `BatchTxRecvError` variants which eliminates the need for `Result<Result<...>>`.
 
 Let me know if this approach works well or if you have further changes in mind!

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,243 @@
...
+        for (response_tx, pool_result) in response_tx.into_iter().zip(pool_results) {
+            let _ = response_tx.send(pool_result);
+        }
+    }
+}
+
+impl<Pool> Future for BatchTxProcessor<Pool>
+where
+    Pool: TransactionPool + Clone + Send + Sync + 'static,
+    Pool::Transaction: Send + Sync,
```
klkvr:
```suggestion
    Pool: TransactionPool + 'static,
```

crates/rpc/rpc/src/eth/helpers/transaction.rs
```diff
@@ -34,11 +34,21 @@ where
...
+            batch_sender.send(request).map_err(|_| EthApiError::BatchTxSendError)?;
+            let AddedTransactionOutcome { hash, .. } = response_rx.await??;
+
+            Ok(hash)
+        } else {
+            // submit the transaction to the pool with a `Local` origin
+            let AddedTransactionOutcome { hash, .. } =
+                self.pool().add_transaction(TransactionOrigin::Local, pool_transaction).await?;
+            Ok(hash)
+        }
```
klkvr: there are a few more places where we send the tx to the pool — e.g in op implementation:
https://github.com/paradigmxyz/reth/blob/38edbde4542d47034148fa3f3f42786866de4a01/crates/optimism/rpc/src/eth/transaction.rs#L48-L54

also there's `sendTransaction` and `sendRawTransactionConditional`. Ideally we expose this logic as a helper on `EthApiInner` I think

crates/transaction-pool/src/batcher.rs
```diff
@@ -0,0 +1,243 @@
...
+    pool: Pool,
+    max_batch_size: usize,
+    #[pin]
+    request_rx: mpsc::UnboundedReceiver<BatchTxRequest<Pool::Transaction>>,
+}
+
+impl<Pool> BatchTxProcessor<Pool>
+where
+    Pool: TransactionPool + Clone + Send + Sync + 'static,
+    Pool::Transaction: Send + Sync,
```
klkvr:
```suggestion
    Pool: TransactionPool + 'static,
```

klkvr (changes requested):

crates/rpc/rpc/src/eth/core.rs
klkvr: I'd be fine with that, can make the default batch size 1

crates/rpc/rpc/src/eth/helpers/transaction.rs
0xKitsune: Sounds good, currently all transactions inserted via the `tx_batch_sender` are marked as `TransactionOrigin::Local` however there are instances where [other transaction origins are used](https://github.com/paradigmxyz/reth/blob/59e4a5556fa54f1c210e45412b6a91f2351bea19/crates/optimism/rpc/src/eth/ext.rs#L163). I can update the batch insertion logic to sort transactions by origin before insertion unless you have a more optimal approach in mind.

crates/rpc/rpc/src/eth/core.rs
0xKitsune: I updated the PR to use tx batching by default for transactions submitted via `send_raw_transaction`. The default batch size has been updated to 1.

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -45,12 +45,8 @@ where
                     tracing::debug!(target: "rpc::eth", %err, hash=% *pool_transaction.hash(), "failed to forward raw transaction");
                 })?;
 
-            // Retain tx in local tx pool after forwarding, for local RPC usage.
```
klkvr, line 48: let's keep the comment

crates/rpc/rpc/src/eth/builder.rs
```diff
@@ -330,6 +346,11 @@ where
             }),
         );
 
+        // Create tx pool insertion batcher
+        let (processor, tx_batch_sender) =
+            BatchTxProcessor::new(components.pool().clone(), max_batch_size);
+        task_spawner.spawn_critical("tx-batcher", Box::pin(processor));
```
klkvr: let's move this setup into `EthApiInner::new` and make it take `max_batch_size` instead

klkvr (dismissed):

mattsse (approved):
lgtm,

we probably want to increase the max batch size but let's start with 1 for now, the overhead for a single 1 should be negligible here

PR #17674, opened 2025-07-29
perf(trie): use ConfiguredSparseTrie for storage roots

Rjected:
This would cause the parallel sparse trie to be used for storage tries as well, because some contracts are very large and their transactions cause lots of reveals + root calculations that are entirely serial

TODO:
 * oops need to actually enable parallel sparse trie here


Rjected:
this needs to be done differently

PR #17675, opened 2025-07-30
chore: move historical status part into single doc

Pana:
Since reth 1.0 has been released for a while, and now is mature enough for everyone to use, so the historical status part is no need put into reademe.


mattsse (commented):
I think we can keep this in the main readme instead

PR #17679, opened 2025-07-30
fix(GPO): calculate `max_tx_gas_used` corner case

yiweichi:
This PR fix a corner case in function `op_suggest_tip_cap` of gas price oracle. I think there are basically three issues:
1. When the receipts length is 1, we considered `max_tx_gas_used` as `None`, but it should be `receipts[0].cumulative_gas_used()`.
2. When the receipts length is larger than 1, this always fail to return the gas used for the first receipt. if we have [receipt_one, receipt_two], this would return `receipt_two.cumulative_gas_used - receipt_one.cumulative_gas_used`. so it miss `receipt_one.cumulative_gas_used`.
3. When the receipts length is 0, we better still update the `inner.last_price` cache.

In this case if the `gas_used` of this single transaction block is large enough, we won't suggest an underpriced gas tip.

### DIFF
--- a/crates/rpc/rpc-eth-types/src/gas_oracle.rs
+++ b/crates/rpc/rpc-eth-types/src/gas_oracle.rs
@@ -306,27 +306,18 @@ where
             .await?
             .ok_or(EthApiError::ReceiptsNotFound(BlockId::latest()))?;
 
-        let max_tx_gas_used = match receipts.len() {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah nice, yeah this is totally correct, tysm!

mattsse (approved):
ah nice, yeah this is totally correct, tysm!

PR #17681, opened 2025-07-30
feat: delay block -> payload conversions

klkvr:
Closes #17640 
Depends on https://github.com/alloy-rs/evm/pull/147


mattsse (approved):
all of this makes sense, let's merge into perf-engine branch

kaxhnet (approved):

kaxhnet (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -265,16 +269,20 @@ where
...
         let (prewarm_tx, prewarm_rx) = mpsc::channel();
         let (execute_tx, execute_rx) = mpsc::channel();
         self.executor.spawn_blocking(move || {
-            for transaction in transactions {
+            for tx in transactions {
+                let tx = tx.map(|tx| WithTxEnv { tx_env: tx.to_tx_env(), tx });
                 // only send Ok(_) variants to prewarming task
-                if let Ok(tx) = &transaction {
+                if let Ok(tx) = &tx {
                     let _ = prewarm_tx.send(tx.clone());
```
mattsse, line 283: do we technically nicht to clone the entire `WithTxEnv` here for prewarming?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -384,20 +433,44 @@ where
...
             (output, execution_finish)
         } else {
-            ensure_ok!(self.execute_block(&state_provider, env, &block, &mut handle))
+            ensure_ok!(self.execute_block(&state_provider, env, &input, &mut handle))
         };
 
         // after executing the block we can stop executing transactions
         handle.stop_prewarming_execution();
 
+        let block = self.convert_to_block(input)?;
```
mattsse, line 446: we could potentially reuse the collected exectuabaleiterator here potentially and get rid of the additional  decode + recovery step here but we can look at this in a followup

crates/ethereum/evm/src/lib.rs
```diff
@@ -275,6 +279,89 @@ where
...
+            cfg_env.set_max_blobs_per_tx(blob_params.max_blobs_per_tx);
+        }
+
+        // derive the EIP-4844 blob fees from the header's `excess_blob_gas` and the current
+        // blobparams
+        let blob_excess_gas_and_price =
+            payload.payload.as_v3().map(|v3| v3.excess_blob_gas).zip(blob_params).map(
+                |(excess_blob_gas, params)| {
+                    let blob_gasprice = params.calc_blob_fee(excess_blob_gas);
+                    BlobExcessGasAndPrice { excess_blob_gas, blob_gasprice }
```
mattsse, line 319: maybe we should introduce a helper for this next to calc_blob_fee so that we can use do `calc_blob_prices().into()`

crates/evm/evm/src/execute.rs
```diff
@@ -554,30 +554,38 @@ where
...
 }
 
-/// A helper trait marking a 'static type that can be converted into an [`ExecutableTx`] for block
-/// executor.
-pub trait OwnedExecutableTx<TxEnv, Tx>: Clone + Send + 'static {
-    /// Converts the type into an [`ExecutableTx`] for block executor.
-    fn as_executable(&self) -> impl IntoTxEnv<TxEnv> + RecoveredTx<Tx> + Copy;
+/// A container for a transaction and a transaction environment.
+#[derive(Debug, Clone)]
+pub struct WithTxEnv<TxEnv, T> {
```
mattsse, line 569: perhaps smth we should consider upstreaming?

crates/optimism/evm/src/lib.rs
```diff
@@ -217,6 +224,76 @@ where
...
+        SignedTx = R::Transaction,
+        BlockHeader = Header,
+        BlockBody = alloy_consensus::BlockBody<R::Transaction>,
+        Block = alloy_consensus::Block<R::Transaction>,
+    >,
+    OpTransaction<TxEnv>: FromRecoveredTx<N::SignedTx> + FromTxWithEncoded<N::SignedTx>,
+    R: OpReceiptBuilder<Receipt: DepositReceipt, Transaction: SignedTransaction>,
+    Self: Send + Sync + Unpin + Clone + 'static,
+{
+    fn evm_env_for_payload(&self, payload: &OpExecutionData) -> EvmEnvFor<Self> {
```
mattsse, line 242: likely that we want this as a standalone function, so perhaps we can consider moving this conversion to alloy-evm somehow, like `ToEvmEnv` or somth

examples/custom-node/src/evm/config.rs
```diff
@@ -81,3 +90,29 @@ impl ConfigureEvm for CustomEvmConfig {
...
+    fn tx_iterator_for_payload(
+        &self,
+        payload: &CustomExecutionData,
+    ) -> impl ExecutableTxIterator<Self> {
+        payload.inner.payload.transactions().clone().into_iter().map(|encoded| {
+            let tx = CustomTransaction::decode_2718_exact(encoded.as_ref())
+                .map_err(Into::into)
+                .map_err(PayloadError::Decode)?;
+            let signer = tx.try_recover().map_err(NewPayloadError::other)?;
+            Ok::<_, NewPayloadError>(WithEncoded::new(encoded, tx.with_signer(signer)))
```
mattsse, line 115: nice, we could leave a note here about why we're doing this

mattsse (approved):
cool, all of this makes sense

crates/engine/tree/src/tree/payload_processor/mod.rs
klkvr: not really, we only need non-TxEnv context for this log https://github.com/paradigmxyz/reth/blob/04fb7c4efd4a9523bcc3075e0a990e5c2f873749/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L321-L322

crates/optimism/evm/src/lib.rs
klkvr: we need a chainspec for this so this might be a bit tricky

crates/ethereum/evm/Cargo.toml
```diff
@@ -12,10 +12,12 @@ workspace = true
 
 [dependencies]
 # Reth
+reth-engine-primitives.workspace = true
 reth-execution-types.workspace = true
 reth-chainspec.workspace = true
 reth-ethereum-forks.workspace = true
 reth-primitives-traits.workspace = true
+reth-payload-primitives.workspace = true
```
mattsse: I think these are causing the riscv target failures because these include some non compatible deps

mattsse (changes requested):
we likely need to keep the keep them evm crates riscv compatible

if we now pull in all the engine types in here this is causing issues, maybe we could flip this around here?

crates/ethereum/evm/Cargo.toml
klkvr: 3e52d48 added `engine` feature to evm crates to work around this

mattsse (approved):
gg

Rjected (approved):
lgtm

PR #17682, opened 2025-07-30
chore(sdk): Add example for building offline `TraceApi` with node builder

### LINKED ISSUE #15681: Add example for using rpc apis like TraceApi against a readonly db on eth and op chains
### Describe the feature

I've asked how to do this on telegram and was directed to open an issue here.
Basically I would like to create an external process like in the [rpc-db example](https://github.com/paradigmxyz/reth/tree/main/examples/rpc-db) to execute transaction traces of most transactions of a chain for doing some data extraction.

To avoid the overhead of running an rpc server and calling it via http plus the json encoding/decondig cost I've tried to use the TraceApi directly in the code, I was able to compile it for EthereumNode but not for OpNode (because of the tx type I think).

It would be nice to have an example on both how to use those rpc modules directly as they provide a nice way of interacting with the db (without having to use the evm directly for example) and how to do the same on the OP chains.

### Additional context

This is an example code that compiles (but I couldn't test as the db I have is the OP chain)

```rust
async fn main() -> eyre::Result<()> {
    let db_path = std::env::var("RETH_DB_PATH")?;
    let db_path = Path::new(&db_path);
    let db = Arc::new(open_db_read_only(
        db_path.join("db").as_path(),
        DatabaseArguments::new(ClientVersion::default()),
    )?);
    let spec = Arc::new(ChainSpecBuilder::mainnet().build());
    let factory = ProviderFactory::<NodeTypesWithDBAdapter<EthereumNode, Arc<DatabaseEnv>>>::new(
        db.clone(),
        spec.clone(),
        StaticFileProvider::read_only(db_path.join("static_files"), true)?,
    );

    let provider: BlockchainProvider<NodeTypesWithDBAdapter<EthereumNode, Arc<DatabaseEnv>>> =
        BlockchainProvider::new(factory)?;

    let eth_api = EthApiBuilder::new(
        provider.clone(),
        NoopTransactionPool::default(),
        NoopNetwork::default(),
        EthEvmConfig::mainnet(),
    )
    .build();

    let trace = TraceApi::new(
        eth_api.clone(),
        BlockingTaskGuard::new(10),
        EthConfig::default(),
    );

    let traces = trace.trace_block(BlockId::latest()).await?;

    println!("{:?}", traces);

    Ok(())
}
```

and this is the same example for OpNode that fails:

```rust
async fn main() -> eyre::Result<()> {
    let db_path = std::env::var("RETH_DB_PATH")?;
    let db_path = Path::new(&db_path);
    let db = Arc::new(open_db_read_only(
        db_path.join("db").as_path(),
        DatabaseArguments::new(ClientVersion::default()),
    )?);
    let spec = Arc::new(OpChainSpecBuilder::optimism_mainnet().build());
    let factory = ProviderFactory::<NodeTypesWithDBAdapter<OpNode, Arc<DatabaseEnv>>>::new(
        db.clone(),
        spec.clone(),
        StaticFileProvider::read_only(db_path.join("static_files"), true)?,
    );

    let provider: BlockchainProvider<NodeTypesWithDBAdapter<OpNode, Arc<DatabaseEnv>>> =
        BlockchainProvider::new(factory)?;

    let eth_api = EthApiBuilder::new(
        provider.clone(),
        NoopTransactionPool::default(),
        NoopNetwork::default(),
        OpEvmConfig::mainnet(),
    )
    .build();

    let trace = TraceApi::new(
        eth_api.clone(),
        BlockingTaskGuard::new(10),
        EthConfig::default(),
    );

    let traces = trace.trace_block(BlockId::latest()).await?;

    println!("{:?}", traces);

    Ok(())
}
```

emhane:
Closes https://github.com/paradigmxyz/reth/issues/15681

Note: It isn't possible to build this offline `TraceApi` using the sdk by going via `NodeBuilder`, because that path only returns a `FullNode` and that doesn't work in the `EthApi` atm, because `FullNode` drops the consensus component, so it can't implement `FullNodeComponents`. `EthApi` requires a node type that implements `FullNodeComponents` currently. There is a lot of redundancy here, it's messy. We need to clean it up. The effort of introducing `RpcNodeCore` as trait bound instead of `FullNodeComponents` throughout RPC was initiated to mitigate stuff like this.


mattsse (approved):
cool

PR #17685, opened 2025-07-31
feat(trie): add blinded node metrics in ProofTaskManager

Rjected:
This tracks the number of blinded account / storage nodes that we fetch during the lifetime of a proof task manager. There is usually one per block, this counts all requests and records them into histograms when the task is terminated.


mediocregopher (approved):
Thanks for this, LGTM :+1:

PR #17690, opened 2025-07-31
fix(config): default back gas limit to 45M on mainnet

### LINKED ISSUE #17689: fix: default gas limit to 45M on mainnet
### Describe the feature

After Berlin Interop, we all agreed to bump the gas limit to 45M, and testnets to 60M. However currently the mainnet is also set to 60M.

### Additional context

_No response_

0xMushow:
closes #17689


mattsse (approved):
oopsie

PR #17691, opened 2025-07-31
fix: use primitive header type for fetching header

mattsse:
we need to use the primitive type here


klkvr (approved):
ah right, hopefully this is the last one of these

PR #17693, opened 2025-07-31
feat(rpc): Add support for custom Tokio runtime configuration in EthereumAddOns

mablr:
Towards #17605 (follow-up for #17611)

This change bubbles up the `with_tokio_runtime()` method from the `RpcServerConfig` to `EthereumAddOns`, allowing to configure nodes with custom Tokio runtimes.

The optional runtime handle is now passed through the `RpcAddOns` structure and applied to RPC server configuration when provided.


crates/node/builder/src/rpc.rs
```diff
@@ -725,7 +779,15 @@ where
             engine_handle,
         } = setup_ctx;
 
-        let server_config = config.rpc.rpc_server_config().set_rpc_middleware(rpc_middleware);
+        let server_config = if let Some(tokio_runtime) = tokio_runtime {
+            config
+                .rpc
+                .rpc_server_config()
+                .set_rpc_middleware(rpc_middleware)
+                .with_tokio_runtime(tokio_runtime)
```
mattsse: for this we can add another fn that takes an Option<Handle> then is is a bit cleaner

mattsse (changes requested):
great, we can also do this for opaddons

crates/ethereum/node/tests/it/builder.rs
```diff
@@ -72,6 +72,44 @@ async fn test_eth_launcher() {
             });
 }
 
+#[test]
+fn test_eth_launcher_with_tokio_runtime() {
+    // #[tokio::test] can not be used here because we need to create a custom tokio runtime
+    // and it would be dropped before the test is finished, resulting in a panic.
```
mattsse, line 78: thanks for flagging this, I also added this to the addons fns

mattsse (approved):
ty!

I made the functions accept an option this makes this slightly more ergonomic when dealing with this

PR #17695, opened 2025-07-31
fix(txpool): also emit promoted pending tx on pool drift

mattsse:
this function is invoked when the pool drifted, e.g. synced initially, then all accounts are updated which can promote transactions to pending.

we should also emit the pending txs here because it is expected that the pending subscription receives all new pending txs (e.g. eth_subscribe)


Rjected (approved):
makes sense

PR #17698, opened 2025-07-31
test(op-chainspec): add isthmus checks

lean-apple:
Add isthmus time check with the mainnet timestamp referred in https://docs.optimism.io/notices/upgrade-15.

cc @emhane


mattsse (approved):
lgtm, pending @emhane

emhane (approved):

PR #17699, opened 2025-07-31
fix: skip pending tx updates with higher prio value

### LINKED ISSUE #17696: Incorrect transaction ordering by gas price within flashblocks on Base mainnet
### Describe the bug

**Description:**
After the recent introduction of flashblocks on the Base main, I observed that transactions within the same flashblock are not consistently ordered by gas price. Specifically, I've found multiple cases where a transaction with a much higher gas price appears after one with a lower gas price within the same flashblock, which contradicts expected behavior.

To confirm this issue, I subscribed to the websocket stream of flashblocks and identified several instances demonstrating this unexpected ordering.

I have collected examples demonstrating these discrepancies and attached them below for reference.

I have talked about the issue with @haardikk21 in Base Discord, the consensus is that the issue is on the Reth side (not rollup-boost or op-rbuilder)

**Expected behavior:**
Transactions within the same flashblock should be consistently ordered by descending gas price.

**Actual behavior:**
Transactions within the same flashblock are occasionally ordered in a seemingly arbitrary or incorrect sequence, with higher gas price transactions sometimes appearing after lower gas price ones.

**Attachments:**
Attached is a file containing specific examples where transaction ordering by gas price appears incorrect, structure is: block number, flashblock index + info for two incorrectly ordered transactions (gasPrice, sender, tx index, hash)

[incorrect_order.log](https://github.com/user-attachments/files/21534477/incorrect_order.log)

### Steps to reproduce

**Steps to reproduce:**

Subscribe to the websocket stream of flashblocks on the Base network.

Analyze the transactions ordering within each received flashblock.

Observe that transaction ordering does not consistently align with descending gas price.

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Other

none

none

Base Mainnet

Archive (default)

_No response_

_No response_

### Code of Conduct

mattsse:
while this is a neat feature, this breaks expected behaviour of this iterator

there's an argument for still yielding the best available tx, so perhaps we should also make this configurable somehow with settings for updates, but enforcing this always for new live updates seems appropriate

closes #17696
closes #17325

### DIFF
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -983,7 +983,7 @@ mod tests {
             transaction: Arc::new(valid_new_higher_fee_tx.clone()),
             priority: Priority::Value(U256::from(u64::MAX)),
         };
-        tx_sender.send(pending_tx.clone()).unwrap();
+        tx_sender.send(pending_tx).unwrap();
 
         // ensure that the higher prio tx is skipped since we yielded a lower one
         for tx in best {
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -2,7 +2,7 @@ use crate::{
     error::{Eip4844PoolTransactionError, InvalidPoolTransactionError},
     identifier::{SenderId, TransactionId},
     pool::pending::PendingTransaction,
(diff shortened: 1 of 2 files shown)


Rjected (approved):
i see, so we can still include new txs from the channel but they won't break the ordering

PR #17702, opened 2025-08-01
fix: ensure transactions are added to subpool in the order of nonce

### LINKED ISSUE #17701: BestTransactions may iterate transactions in nonce disorder
### Describe the bug

When a newly added transaction is added to the pending subpool, some other transactions may also need to be moved to the pending subpool. 
https://github.com/paradigmxyz/reth/blob/a5f2d58650664abea51f36a12dd3ae1c156d7579/crates/transaction-pool/src/pool/txpool.rs#L676-L702
The execution order is:
1. The new transaction is first added to the pending subpool.
2. Other transactions are moved to pending. 
The `PendingPool` will send the newly added transaction in insertion order to existing `BestTransactions` instances. If the transactions are not added in nonce order, the `BestTransactions` will receive and iterate transactions in a nonce disorder. 
The scenario that transactions are not added in nonce order is: when "other `updates` transactions` contain a transaction whose sender is the same of the newly added transaction but with la ower nonce. I will give an example below in the reproduction steps. 

The problematic symptom is: the transaction list output by `BestTransactions` does not adhere to the partial order by transaction nonce. 

### Steps to reproduce

1. Add this test in https://github.com/paradigmxyz/reth/blob/a5f2d58650664abea51f36a12dd3ae1c156d7579/crates/transaction-pool/src/pool/txpool.rs 
```rust

    #[test]
    fn test_insertion_disorder() {
        let mut f = MockTransactionFactory::default();
        let mut pool = TxPool::new(MockOrdering::default(), Default::default());

        let sender = address!("1234567890123456789012345678901234567890");
        let tx0 = f.validated_arc(
            MockTransaction::legacy().with_sender(sender).with_nonce(0).with_gas_price(10),
        );
        let tx1 = f.validated_arc(
            MockTransaction::eip1559()
                .with_sender(sender)
                .with_nonce(1)
                .with_gas_limit(1000)
                .with_gas_price(10),
        );
        let tx2 = f.validated_arc(
            MockTransaction::legacy().with_sender(sender).with_nonce(2).with_gas_price(10),
        );
        let tx3 = f.validated_arc(
            MockTransaction::legacy().with_sender(sender).with_nonce(3).with_gas_price(10),
        );

        let mut best = pool.best_transactions();

        // tx0 should be put in the pending subpool
        pool.add_transaction((*tx0).clone(), U256::from(1000), 0, None).unwrap();
        let t0 = best.next().expect("tx0 should be put in the pending subpool");
        assert_eq!(t0.id(), tx0.id());
        // tx1 should be put in the queued subpool due to insufficient sender balance
        pool.add_transaction((*tx1).clone(), U256::from(1000), 0, None).unwrap();
        assert!(best.next().is_none());
        // tx2 should be put in the pending subpool, and tx1 should be promoted to pending
        pool.add_transaction((*tx2).clone(), U256::MAX, 0, None).unwrap();
        let t1 = best.next().expect("tx2 should be put in the pending subpool");
        let t2 = best.next().expect("tx2 should be put in the pending subpool");
        assert_eq!(t1.id(), tx1.id());
        assert_eq!(t2.id(), tx2.id());
        // tx3 should be put in the pending subpool,
        pool.add_transaction((*tx3).clone(), U256::MAX, 0, None).unwrap();
        let t3 = best.next().expect("tx3 should be put in the pending subpool");
        assert_eq!(t3.id(), tx3.id());
    }
```
The test does the following things:
1. It first adds a transaction `tx0` (nonce=0), directly put in the pending subpool. 
2. Then transaction `tx1` (nonce=1) is added, whose cost is higher than the sender's balance. It is put in the queued subpool. 
3. Transaction `tx2` (nonce=2) is added. In the meantime, the sender's balance is updated to be sufficiently large. `tx2` should be put in the pending subpool and `tx1` should be moved to the queued subpool. 
4. As the logic in `TxPool::add_transaction`, `tx2` will be first added to the subpool and then `tx1`. The existing `BestTransaction` instances will first receive `tx2` and then `tx1`. 
5. As a result, the `BestTransaction` iterator will first output `tx2` and then `tx1`, causing a nonce gap in the output transaction list. 

### Node logs

```text
N/A
```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

git commit `54a4a23f64a2522903f9aafb02d2b65f5a45ad7d`

N/A

N/A

Archive (default)

_No response_

_No response_

### Code of Conduct

Troublor:
As mentioned in #17701, the `updates` in `crates/transaction-pool/src/pool/txpool.rs::TxPool::add_transaction()` may contain both transactions with lower nonce and higher nonce than the newly added transaction. The original implementation performs the update on existing transactions after adding the new transaction, causing insertion of transactions into subpools not ordered by nonce. 

This PR fixes the issue by partitioning transactions in `updates` and ensure that transactions with lower nonce are moved first before adding the new transaction to subpool, while the transactions with higher nonce are move afterwards.

### DIFF
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -675,12 +675,30 @@ impl<T: TransactionOrdering> TxPool<T> {
 
         match self.all_transactions.insert_tx(tx, on_chain_balance, on_chain_nonce) {
             Ok(InsertOk { transaction, move_to, replaced_tx, updates, .. }) => {
+                // Bipartition the updates into two lists:
+                // 1. those needed to move before adding the new transaction. Typically,
+                //    transactions with same sender as the newly added one but lower nonce should be
+                //    moved first.
+                // 2. those needed to move after adding the new transaction. Typically, transactions
+                //    with same sender as the newly added one but higher nonce should be moved
+                //    after.
+                // We ensure that transactions of the same sender are added to subpool in the order
+                // of nonce.
+                let (pre_updates, post_updates) =
+                    self.bipartition_updates(transaction.id(), updates);
+                // process updates before adding new transaction
+                let UpdateOutcome { promoted: pre_promoted, discarded: pre_discarded } =
+                    self.process_updates(pre_updates);
                 // replace the new tx and remove the replaced in the subpool(s)
                 self.add_new_transaction(transaction.clone(), replaced_tx.clone(), move_to);
                 // Update inserted transactions metric
                 self.metrics.inserted_transactions.increment(1);
-                let UpdateOutcome { promoted, discarded } = self.process_updates(updates);
-
+                // process updates after adding new transaction
+                let UpdateOutcome { promoted: post_promoted, discarded: post_discarded } =
+                    self.process_updates(post_updates);
+                // merge update outcome
+                let promoted = [pre_promoted, post_promoted].concat();
+                let discarded = [pre_discarded, post_discarded].concat();
                 let replaced = replaced_tx.map(|(tx, _)| tx);
 
                 // This transaction was moved to the pending pool.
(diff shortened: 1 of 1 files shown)


mattsse (commented):
thanks for this, 

the explanation makes sense, I'm exploring if we can make this a bit more efficient

this primarily affects the pending pool, right?

mattsse:
@Troublor I think this makes more sense, but since this only includes tx from the same sender we can also do this on the callsite since we know the tx's nonce

mattsse:
hmm @Troublor is this an edge case when the the new balance would move txs < newtx.nonce to pending?

because cant think of another case where a new tx would move tx with a lower nonce to pending

perhaps we pending insert should also be adjusted to check for lower nonce?

Troublor:
@mattsse Yes, I think this is an edge case. What do you mean by "pending insert should also be adjusted to check for lower nonce?"?

mattsse:
@Troublor I believe this is an issue with live feeding we can probably fix this edge case and not emit if the tx has an ancestor

mattsse:
Superseded by #22308, which includes a more complete fix that handles the gap-fill and mixed-promotion scenarios as well. Thanks @Troublor for the original report and fix attempt!

PR #17708, opened 2025-08-01
feat: Execute execution-apis-tests in e2e tests

shemnon:
Add a test case that implements the instructions in the rpc-e2e-tests README.md file. If the EXECUTION_APIS_TEST_PATH is not set the test will pass, otherwise all configured tests must pass.


mattsse (approved):
makes sense

PR #17709, opened 2025-08-01
fix: forward unknown hashes pre bedrock

mattsse:
we dont have the hashes pre-bedrock so if we cant find the blocknumber to a blockid hash request then we assume it's pre-bedrock and forward


Rjected (approved):
i see lol

PR #17711, opened 2025-08-02
feat: introduced configurable version globals

### LINKED ISSUE #17705: Introduce configurable version globals
### Describe the feature

we pull all of these in during cargo build
https://github.com/paradigmxyz/reth/blob/2170f1b97eede4e41de40fbade18a9d3dd681e8b/crates/node/core/src/version.rs#L10-L74

this has a side-effect that if the CLi is reused it still is reporting those constants

https://github.com/paradigmxyz/reth/blob/2170f1b97eede4e41de40fbade18a9d3dd681e8b/crates/ethereum/cli/src/interface.rs#L32-L32

we can fix this if we

introduce a container struct and a lazy static, that allows setting this and convert constant access to functions, that then check the global lazy static

if nothing has been installed this should install a container with the current constants

### Additional context

cc @rezbera

Rimeeeeee:
fixes #17705

### DIFF
--- a/crates/node/core/src/version.rs
+++ b/crates/node/core/src/version.rs
@@ -1,4 +1,6 @@
 //! Version information for reth.
+use std::sync::OnceLock;
+
 use alloy_primitives::Bytes;
 use alloy_rpc_types_engine::ClientCode;
 use reth_db::ClientVersion;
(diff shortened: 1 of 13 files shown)


crates/node/core/src/version.rs
```diff
@@ -1,63 +1,61 @@
...
+    /// # Example
+    ///
+    /// ```text
+    /// reth/v{major}.{minor}.{patch}-{sha1}/{target}
+    /// ```
+    /// e.g.: `reth/v0.1.0-alpha.1-428a6dc2f/aarch64-apple-darwin`
+    pub p2p_client_version: &'a str,
+
+    /// extra data used for payload building
+    pub extra_data: String,
```
mattsse: I'd like to make all of these fields a `Cow<'static` then this can also be configured with `String`

crates/node/core/src/version.rs
```diff
@@ -82,9 +80,53 @@ pub fn default_extra_data_bytes() -> Bytes {
...
-        git_sha: VERGEN_GIT_SHA.to_string(),
-        build_timestamp: VERGEN_BUILD_TIMESTAMP.to_string(),
+        version: env!("CARGO_PKG_VERSION").to_string(),
+        git_sha: env!("VERGEN_GIT_SHA_SHORT").to_string(),
+        build_timestamp: env!("VERGEN_BUILD_TIMESTAMP").to_string(),
+    }
+}
+
+/// Global static version metadata
+static VERSION_METADATA: OnceLock<RethCliVersionConsts<'static>> = OnceLock::new();
```
mattsse: let's move this top of file

crates/node/core/src/version.rs
```diff
@@ -82,9 +80,53 @@ pub fn default_extra_data_bytes() -> Bytes {
...
+
+/// Initialize the global version metadata.
+pub fn init_version_metadata(metadata: RethCliVersionConsts<'static>) {
+    let _ = VERSION_METADATA.set(metadata);
+}
```
mattsse: we can name this `try_int_version_metadata` and return the error as well, also move top of file

crates/node/core/src/version.rs
```diff
@@ -82,9 +80,53 @@ pub fn default_extra_data_bytes() -> Bytes {
...
+
+/// Get a reference to the initialized version metadata.
+///
+/// # Panics
+/// If init_version_metadata() hasn't been called.
+pub fn version_metadata() -> &'static RethCliVersionConsts<'static> {
+    VERSION_METADATA.get().expect("Version metadata not initialized")
```
mattsse: for this we want a `get_version_metadata` that does an or_init with the `default_version_metadata` so that this function does the init

mattsse (changes requested):
great, some suggestions

crates/node/core/src/version.rs
```diff
@@ -82,9 +90,48 @@ pub fn default_extra_data_bytes() -> Bytes {
...
+/// # Panics
+/// If `init_version_metadata()` hasn't been called.
+pub fn version_metadata() -> &'static RethCliVersionConsts {
+    VERSION_METADATA.get().expect("Version metadata not initialized")
+}
+
+/// Get a reference to the global version metadata
+pub fn get_version_metadata() -> &'static RethCliVersionConsts {
+    VERSION_METADATA.get_or_init(default_version_metadata)
+}
```
mattsse: these can be easily misused, I'd like to only keep `get_version_metadata` which does lazy init but rename it to `version_metadata`

crates/node/core/src/version.rs
```diff
@@ -82,9 +90,48 @@ pub fn default_extra_data_bytes() -> Bytes {
...
+
+/// default version metadata using compile-time env! macros.
+pub fn default_version_metadata() -> RethCliVersionConsts {
```
mattsse: let's name this `default_reth_version_metadata`

mattsse (changes requested):
last pedantic nit

mattsse (approved):
tysm

fyi @rezbera

PR #17712, opened 2025-08-02
perf: use debug_assert for parked pool lookup

malik672:
would call it optimize but doubt this check is hardly ever hit tbh


mattsse (changes requested):
actually, let's just convert this to a debug_assert instead

we did the same for the pending pool, and these checks are mostly safeguards but since this is in the hotpath we should just make this a debug_assert

malik672:
> actually, let's just convert this to a debug_assert instead
> 
> we did the same for the pending pool, and these checks are mostly safeguards but since this is in the hotpath we should just make this a debug_assert

can the `contains()` function be removed

mattsse (commented):
this would only avoid the second lookup in the unhappy path(panic case), which doesnt matter so we can keep it as this

mattsse:
actually, I've converted this to debug_assert which removes the lookup entirely

mattsse (approved):

PR #17718, opened 2025-08-03
test(generators): add topics_count parameter to random_receipt

SkandaBhat:
Adds a `topics_count` parameter to the `random_receipt` test utility function to allow control over the number of topics generated in receipt logs during testing.

This is passed through to the existing `random_log` function which already supports configurable topic counts.


mattsse (approved):
makes sense

PR #17720, opened 2025-08-04
fix: enforce propagate on getpooledtx

mattsse:
the assumption here was that because this is pull based the remote must already know the hash.

but we've documented that this fn respects propagate setting so we should filter them here

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -341,7 +341,7 @@ where
         }
     }
 
-    /// Returns pooled transactions for the given transaction hashe sthat are allowed to be propagated.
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -341,7 +341,7 @@ where
         }
     }
 
-    /// Returns pooled transactions for the given transaction hashes.
+    /// Returns pooled transactions for the given transaction hashe sthat are allowed to be propagated.
```
bharath-123:
```suggestion
    /// Returns pooled transactions for the given transaction hashes that are allowed to be propagated.
```

Rjected (approved):
makes sense, yeah we do document it like that

PR #17723, opened 2025-08-04
feat: enable `crypto-backend` for signer recovery

kevaundray:


crates/primitives-traits/Cargo.toml
```diff
@@ -16,7 +16,7 @@ workspace = true
 reth-codecs = { workspace = true, optional = true }
 
 # ethereum
-alloy-consensus = { workspace = true, features = ["k256"] }
+alloy-consensus = { workspace = true, features = ["k256", "crypto-backend"] }
```
mattsse, line 19: since we don't use this directly and users can always force this, perhaps we dont need this?

kevaundray: yeah thats true, this only saves users from adding an extra feature flag

PR #17729, opened 2025-08-05
fix: typo in file deletion error message

radik878:
Replace "kokot" with "file" in the error message when deleting
outdated ERA files outside the working range.

This fixes a typo that was present in the debug output message
in the delete_outside_range method of EraClient.


mattsse (approved):
okay this one i let slide

PR #17730, opened 2025-08-05
fix: execution-apis eth_syncing should return false

shemnon:
e2e test setup does not update the syncing state of the test node. This causes eth_syncing to falsely report the node is still syncing (the default state of all nodes).

Added canonical eth_syncing.io test and updated name of test_eth_get_logs_compat to test_local_rpc_tests_compat so it reflects reality.


fgimenez (approved):
good point, for e2e tests in which the CL role is played by the test itself it makes more sense to start in idle state, thanks!

PR #17737, opened 2025-08-06
feat(chain-state): add `IndexedTx` helper

### LINKED ISSUE #17735: Introduce helper for IndexedTx
### Describe the feature

we can make this a lot more convenient if we introduce a helper type and a function that returns `fn find_indexed(&TxHash)-> Indexed<B>`
where Indexed whould borrow the entire block and have the index and transaction matching the hash, then we can move this TransactionMeta as a helper function so that this code can be replaced by `find_index(txhash).meta()`

https://github.com/paradigmxyz/reth/blob/4d96ea0343b1f5f716449da5ac2c08335e31edd1/crates/chain-state/src/in_memory.rs#L556-L574

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17735.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -709,28 +709,7 @@ impl<N: NodePrimitives> BlockState<N> {
         tx_hash: TxHash,
     ) -> Option<(N::SignedTx, TransactionMeta)> {
         self.chain().find_map(|block_state| {
-            block_state
-                .block_ref()
-                .recovered_block()
-                .body()
-                .transactions_iter()
-                .enumerate()
-                .find(|(_, tx)| tx.trie_hash() == tx_hash)
-                .map(|(index, tx)| {
-                    let meta = TransactionMeta {
-                        tx_hash,
-                        index: index as u64,
-                        block_hash: block_state.hash(),
-                        block_number: block_state.block_ref().recovered_block().number(),
-                        base_fee: block_state.block_ref().recovered_block().base_fee_per_gas(),
-                        timestamp: block_state.block_ref().recovered_block().timestamp(),
-                        excess_blob_gas: block_state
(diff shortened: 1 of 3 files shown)


crates/chain-state/src/in_memory.rs
```diff
@@ -725,30 +709,65 @@ impl<N: NodePrimitives> BlockState<N> {
...
+/// Transaction with its index and block reference for efficient metadata access.
+#[derive(Debug)]
+pub struct IndexedTx<'a, N: NodePrimitives> {
+    /// Recovered block containing the transaction
+    block: &'a RecoveredBlock<N::Block>,
+    /// Transaction matching the hash
+    tx: &'a N::SignedTx,
+    /// Index of the transaction in the block
+    index: usize,
+}
```
mattsse: great!

I'd like to move even move this into the reth-primitives-trat crate

crates/chain-state/src/in_memory.rs
```diff
@@ -725,30 +709,65 @@ impl<N: NodePrimitives> BlockState<N> {
...
+
+    /// Finds a transaction by hash and returns it with its index and block context.
+    pub fn find_indexed(&self, tx_hash: TxHash) -> Option<IndexedTx<'_, N>> {
+        self.block_ref()
+            .recovered_block()
+            .body()
+            .transactions_iter()
+            .enumerate()
+            .find(|(_, tx)| tx.trie_hash() == tx_hash)
+            .map(|(index, tx)| IndexedTx { block: self.block_ref().recovered_block(), tx, index })
```
mattsse: then we can make this a function of `RecoveredBlock`

mattsse (changes requested):

mattsse (approved):
lgtm! ty

PR #17738, opened 2025-08-06
refactor: Minor cleanups and improvements in ef-tests suite

torrpriius:
1.  testing/ef-tests/tests/tests.rs:
    -   Replaced an unnecessary `format!("{}", ...)` call with a more direct and efficient `.to_string()` in the `blockchain_test!` macro.

2.  testing/ef-tests/src/case.rs:
    -   Improved the documentation for `Cases::run` to explicitly state what the function returns, enhancing clarity for future developers.

3.  testing/ef-tests/src/models.rs:
    -   Clarified the documentation for the `requests_hash` field in the `Header` struct, specifying that it represents the root of the requests trie.

4.  testing/ef-tests/src/cases/blockchain_test.rs:
    -   Removed an unprofessional smiley face `:)` from a code comment in the `should_skip` function.

### DIFF
--- a/testing/ef-tests/src/case.rs
+++ b/testing/ef-tests/src/case.rs
@@ -33,7 +33,7 @@ pub struct Cases<T> {
 }
 
 impl<T: Case> Cases<T> {
-    /// Run the contained test cases.
+    /// Runs all test cases and returns a vector of `CaseResult`.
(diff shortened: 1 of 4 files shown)


mattsse (commented):
these are fine

PR #17739, opened 2025-08-06
fix: add missing semicolon in wallet generation loop

Fibonacci747:
Fixed missing semicolon in wallets.push(wallet) statement inside Wallet::derive_wallets implementation.


mattsse (approved):
i like this more

Issue #17683, opened 2025-07-30
Use parallel sparse trie for storage tries


Rjected:
### Describe the feature

Right now we use `SerialSparseTrie` for storage tries even if we have enabled parallel sparse trie. Using parallel sparse tries for storage tries can be good for larger storage tries, for example new root calculation for some tries on mainnet can take up to 5ms.

To do this we need a way to initialize the `SparseStateTrie` with parallel storage tries. This is not trivial because the sparse state trie will initialize storage tries itself based on the configured storage trie type using `Default`:
https://github.com/paradigmxyz/reth/blob/3772535220784420370fff9c13f85d17d15eb786/crates/trie/sparse/src/state.rs#L825-L842

We could maybe rewrite `ConfiguredSparseTrie` to contain an enum like it has now, and a flag that causes `Default` to use the `Parallel` variant

PoulavBhowmick03:
Can I take this up?

Rjected:
> Can I take this up?

Hey, I think this needs some design before attempting implementation, if you let me know how you would solve this I can give some feedback!

PR #17741, opened 2025-08-06
perf(trie): use ParallelSparseTrie for storage roots

Rjected:
ref https://github.com/paradigmxyz/reth/issues/17683

this uses parallel sparse trie for storage tries, this does not involve any configuration, just replaces the storage tries.


fynnss:
Is this ready for test?

PR #17743, opened 2025-08-06
feat(evm): add metrics for individual tx execution latency

Rjected:
records individual tx execution latency as a metric


mattsse (commented):
this is one extra clock syscall per tx

should be fine I think?

PR #17746, opened 2025-08-06
refactor: migrate RPC hooks from build-time storage to launch-time parameters

fgimenez:
Refactors RPC addon architecture to migrate RPC hooks from build-time storage to launch-time configuration, while maintaining the Node generic parameter for type safety.

  - moved `on_rpc_started` and `extend_rpc_modules` from being stored in RPC addons to being passed at launch
  - introduced `NodeBuilderWithRpcHooks` wrapper for configuring RPC hooks
  - kept the N parameter in `RpcAddOns` with `PhantomData` for proper hook typing

### DIFF
--- a/crates/node/builder/src/builder/states.rs
+++ b/crates/node/builder/src/builder/states.rs
@@ -7,7 +7,7 @@
 
 use crate::{
     components::{NodeComponents, NodeComponentsBuilder},
-    hooks::NodeHooks,
+    hooks::{NodeHooks, OnNodeStartedHook, RpcHooks},
     launch::LaunchNode,
     rpc::{RethRpcAddOns, RethRpcServerHandles, RpcContext},
     AddOns, ComponentsFor, FullNode,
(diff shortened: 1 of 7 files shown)


crates/optimism/node/tests/it/priority.rs
```diff
@@ -136,7 +136,7 @@ async fn test_custom_block_priority_config() {
         .with_database(db)
         .with_types_and_provider::<OpNode, BlockchainProvider<_>>()
         .with_components(build_components(config.chain.chain_id()))
-        .with_add_ons(OpNode::new(Default::default()).add_ons())
+        .with_add_ons(OpNode::new(Default::default()).default_add_ons())
```
klkvr: does `.add_ons()` not work anymore?

crates/optimism/node/src/node.rs
klkvr, line 1: this file is pretty tricky to review, could we maybe preserve the order of impls here?

crates/node/builder/src/builder/states.rs
```diff
@@ -268,23 +278,19 @@ where
             + Send
             + 'static,
     {
-        self.map_add_ons(|mut add_ons| {
-            add_ons.hooks_mut().set_on_rpc_started(hook);
-            add_ons
-        })
+        self
```
klkvr, line 319: I see, those are just noops now, should we remove them or figure out a way to pass them to rpc addons?

crates/optimism/node/tests/it/priority.rs
fgimenez: nope, having removed the `N` generic forces us to indicate somehow the types used here, old `add_ons` gives now:
```
error[E0283]: type annotations needed
   --> crates/optimism/node/tests/it/priority.rs:139:55
    |
139 |         .with_add_ons(OpNode::new(Default::default()).add_ons())
    |                                                       ^^^^^^^
    |
    = note: cannot satisfy `_: reth_node_api::FullNodeTypes`
    = help: the following types implement trait `reth_node_api::FullNodeTypes`:
              reth_node_api::FullNodeTypesAdapter<Types, DB, Provider>
              reth_node_builder::NodeAdapter<T, C>
    = note: required for `reth_optimism_node::OpNode` to implement `reth_node_builder::Node<_>`
help: try using a fully qualified path to specify the expected types
    |
139 -         .with_add_ons(OpNode::new(Default::default()).add_ons())
139 +         .with_add_ons(<reth_optimism_node::OpNode as reth_node_builder::Node<N>>::add_ons(&OpNode::new(Default::default())))
```
and we need to define `N` too

crates/optimism/node/src/node.rs
fgimenez: indeed, the different steps in the migration have meessed things up, pushed changes to preserve the original order ptal

crates/node/builder/src/builder/states.rs
fgimenez: makes sense, working on passing them to rpc addons

fgimenez: ok done, now we store the hooks in a specialized `NodeBuilderWithRpcHooks` and, at launch, addons are wrapped with their hooks in `AddOnsWithRpcHooks` and passed through the existing launch flow. these methods are no longer no-ops and the api is kept ptal

crates/optimism/node/tests/it/priority.rs
fgimenez: `.add_ons()` works again with the latest changes

mattsse (commented):
All of this makes sense and I like that this allows us to get rid of the hooks_mut fn entirely.

My only concern is that we're now starting to layer addons logic and maybe this isn flexible enough 
I need to think about this a bit more

unsure if layering the addons with AddOnsWithRpcHooks will be flexible enough moving forward because now we need to pass this down via launch_add_ons_with_hooks for example

maybe, instead of layering the addons in side the nodebuilder we should perhaps revisit the way the rpc server is launched, but I don't have anything specific just yet

PR #17751, opened 2025-08-07
fix(stages): use correct block number in error message

reallesee:
Fix error message in next_block_range_with_transaction_threshold to use end_block_number instead of target_block when block body indices are not found. This makes the error message more accurate and helps with debugging


mattsse (approved):
makes sense

PR #17754, opened 2025-08-07
fix: add validation against parent header in `reth-stateless`

Wollac:
This PR adds the previously missing `validate_header_against_parent` check to the stateless validation process, which is a critical validation step.

The rest of the refactoring, such as using `SealedHeader` for ancestor blocks, is a result of enabling this check and helps avoid unnecessary rehashing.

### DIFF
--- a/crates/stateless/src/validation.rs
+++ b/crates/stateless/src/validation.rs
@@ -12,15 +12,16 @@ use alloc::{
     vec::Vec,
 };
 use alloy_consensus::{BlockHeader, Header};
-use alloy_primitives::B256;
-use alloy_rlp::Decodable;
+use alloy_primitives::{keccak256, B256};
 use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_consensus::{Consensus, HeaderValidator};
 use reth_errors::ConsensusError;
 use reth_ethereum_consensus::{validate_block_post_execution, EthBeaconConsensus};
 use reth_ethereum_primitives::{Block, EthPrimitives};
 use reth_evm::{execute::Executor, ConfigureEvm};
-use reth_primitives_traits::{block::error::BlockRecoveryError, Block as _, RecoveredBlock};
+use reth_primitives_traits::{
+    block::error::BlockRecoveryError, Block as _, RecoveredBlock, SealedHeader,
+};
 use reth_trie_common::{HashedPostState, KeccakKeyHasher};
 
 /// Errors that can occur during stateless validation.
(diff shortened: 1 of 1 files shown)


crates/stateless/src/validation.rs
```diff
@@ -167,38 +168,36 @@ where
...
         .iter()
-        .map(|serialized_header| {
-            let bytes = serialized_header.as_ref();
-            Header::decode(&mut &bytes[..])
+        .map(|bytes| {
+            let hash = keccak256(bytes);
+            alloy_rlp::decode_exact::<Header>(bytes)
+                .map(|h| SealedHeader::new(h, hash))
                 .map_err(|_| StatelessValidationError::HeaderDeserializationFailed)
         })
```
kevaundray, line 179: what was the rationale for this change?

crates/stateless/src/validation.rs
```diff
@@ -231,31 +230,30 @@ where
...
     block: &RecoveredBlock<Block>,
+    parent: &SealedHeader<Header>,
 ) -> Result<(), StatelessValidationError>
 where
     ChainSpec: Send + Sync + EthChainSpec<Header = Header> + EthereumHardforks + Debug,
 {
     let consensus = EthBeaconConsensus::new(chain_spec);
 
     consensus.validate_header(block.sealed_header())?;
+    consensus.validate_header_against_parent(block.sealed_header(), parent)?;
```
kevaundray, line 256: Hmm I wonder why the execution spec tests didn't pick this up

crates/stateless/src/validation.rs
```diff
@@ -167,38 +168,36 @@ where
...
+    let mut ancestor_headers: Vec<_> = witness
         .headers
         .iter()
-        .map(|serialized_header| {
-            let bytes = serialized_header.as_ref();
-            Header::decode(&mut &bytes[..])
+        .map(|bytes| {
+            let hash = keccak256(bytes);
+            alloy_rlp::decode_exact::<Header>(bytes)
+                .map(|h| SealedHeader::new(h, hash))
```
mattsse, line 177: ah, right. 

this makes sense

mattsse (approved):
lgtm, pending @kevaundray

crates/stateless/src/validation.rs
mattsse: decode only pops from the buffer, there can be trailing bytes, shouldn't have been an issue before because we sealed these again, but this is better

Wollac: Using `decode` ignores trailing bytes leading, which can lead to malleability issues. In the worst case, there have even been consensus failures due to clients handling trailing RLP bytes differently. Since this function is intended for creating execution proofs, I think it is always good practice to use `decode_exact`, even if there might be no attack vectors in this particular case.

Wollac: The second change is a performance optimization: Since we already have the exact RLP bytes and RLP is deterministic by definition, we can use the RLP bytes directly for the hash. This saves us one memory allocation and one RLP encoding step per header. For a maximum of 256 headers, this gives us a significant performance improvement. However, since there are no more headers in the witness, this speedup is negligible.
Thus, I think it's also absolutely okay to decode this into a plain `Header` and perform an additional `hash_slow()`. Ultimately, it comes down to personal preference.

kevaundray (approved):
LGTM too

PR #17755, opened 2025-08-07
feat: allow external block recovery in `reth-stateless`

Wollac:
This PR introduces the `RecoverableBlock` trait to decouple the expensive block recovery step from the core validation logic.

Previously, `stateless_validation` accepted a Block and immediately performed sender recovery, which can be a significant bottleneck (especially in zkVMs).

By making the validation function generic over a new `RecoverableBlock` trait, consumers of this crate can now provide their own optimized block recovery implementations. This allows for specialized, high-performance recovery logic without altering the validation process itself. The default implementation still converts a `Block` to a `RecoveredBlock` as before.


crates/stateless/src/validation.rs
```diff
@@ -128,17 +127,18 @@ pub enum StatelessValidationError {
 ///
 /// If all steps succeed the function returns `Some` containing the hash of the validated
 /// `current_block`.
-pub fn stateless_validation<ChainSpec, E>(
-    current_block: Block,
+pub fn stateless_validation<B, ChainSpec, E>(
+    current_block: B,
     witness: ExecutionWitness,
     chain_spec: Arc<ChainSpec>,
```
kevaundray: For historical reference; can we say roughly what the custom recovery algorithm that you are thinking or/using is?

Wollac: It makes a lot of sense for a zkVM to recover the public keys from the signatures on the host, which is much more performant, and pass them as part of the witness. This way, the zkVM only needs to verify the keys, which is much more efficient than full recovery.
While there are other ways to achieve this, these changes enable passing either a regular or a recovered block.

crates/stateless/src/validation.rs
```diff
@@ -152,20 +152,20 @@ where
...
     T: StatelessTrie,
+    B: RecoverableBlock<Block>,
     ChainSpec: Send + Sync + EthChainSpec<Header = Header> + EthereumHardforks + Debug,
     E: ConfigureEvm<Primitives = EthPrimitives> + Clone + 'static,
 {
-    let current_block = current_block
-        .try_into_recovered()
-        .map_err(|err| StatelessValidationError::SignerRecovery(Box::new(err)))?;
+    let current_block =
+        current_block.try_into_recovered().map_err(|_| StatelessValidationError::SignerRecovery)?;
```
mattsse: how would an implementation for this look like, is it assumed that the addresses are already available here or what is being called here exactly?

because to me this now looks euqivalent to just accepting `RecoveredBlock` here

mattsse (commented):
one q

crates/stateless/src/validation.rs
Wollac: The goal was to accept a block that allows users of the API to implement their own methods for handling recovery.
The input used to be a `RecoveredBlock` and was changed to a `Block`, so I assumed there were reasons why a `RecoveredBlock` was not desirable. That's why I implemented something that accepts both options. However, you are correct that, for my particular use case, a `RecoveredBlock` would work as well an be simpler. I am more than happy to change it when that would also be acceptable.

Wollac: In general, I think the most logical thing conceptually would be to accept an `ExecutionPayload` instead of a block, as it contains the exact information required to create and execute a new block by definition. However, this would be a significant change to how to use reth-stateless, so it should probably be left for another time.

kevaundray: I had reverted RecoveredBlock because I thought that when we attempt to standardize the interface, it will be made more difficult due to RecoveredBlock not being a standard object across ELs.

I don't have a strong opinion here, If @mattsse thinks RecoveredBlock or ExecutionPayload is better, we can go with that, or we can delay the decision and avoid the trait by having a function that takes in RecoveredBlock. So maybe something like:

```rust
func stateless_validate_with_block(block : Block, witness : Witness) {
    recovered_block := block.recover()
    stateless_validate(recovered_block, witness)
}

func stateless_validate(recovered_block, witness) {
 /// All of the logic in stateless_validation except signer recovery
}
```

kevaundray: (I do think avoiding the new trait would be a good direction to go in here)

kevaundray (approved):

crates/stateless/src/validation.rs
Wollac: So, then which route do we wanna take?
- Go back to `RecoveredBlock` for the input
- Add separate methods for `Block` input. I guess then we'd need another version for `stateless_validation_with_trie` as well
- (I don't don't think `TryInto<RecoveredBlock>` would be an option as due to the Rust generic system it cannot be cleanly implemented.)

kevaundray: I think `RecoveredBlock` would be easiest path forward here -- if someone wants the current interface, then they could do the following in the their guest program:

```rust
fn main() {
   let block : Block = env::read();
   let witness : ExecutionWitness = env::read();
   let recovered_block = block.recover_signers();
   stateless_validation(recovered_block, witness)
}
```

ie they do the recovery in their main functions guest program rather than in here

crates/stateless/src/validation.rs
```diff
@@ -153,20 +151,20 @@ where
...
```
mattsse: I don't think this makes a lot of sense because this is more or less same as accepting

```
current_block: RecoveredBlock<B>,
```

just with extra steps

can we accept recoveredblock here instead?

mattsse (changes requested):

crates/stateless/src/validation.rs
mattsse: exactly, a recoveredblock can be created like this

https://github.com/paradigmxyz/reth/blob/69a1951f54e6f0c92e1968c6a6103362622e681a/crates/primitives-traits/src/block/recovered.rs#L76-L76


https://github.com/paradigmxyz/reth/blob/69a1951f54e6f0c92e1968c6a6103362622e681a/crates/primitives-traits/src/block/recovered.rs#L102-L102

kevaundray: yeah this was my fault for reverting :D

crates/stateless/src/validation.rs
Wollac: OK, I've changed everything to a `RecoveredBlock` as discussed in ba8f616bfac87713e15bec22db07d456cb1159f4.
I've also changed the `SignerRecovery` error, to a `Custom` error. It seemed strange to keep the SignerRecovery if there is no recovery anymore, but the calling code might want to return a `StatelessValidationError` when they do custom block recovery.

mattsse (approved):

PR #17756, opened 2025-08-07
feat: persist origin on pooled tx backup for propagation setting

### LINKED ISSUE #17721: Persists propagate setting when backing up pooled tx on shotdown
### Describe the feature

we currently lose the origin context when we re-insert txs from disk:

https://github.com/paradigmxyz/reth/blob/d5f59070bb0be1cf9bba94e01b895fede1ae7078/crates/transaction-pool/src/maintain.rs#L637-L637

because we simply store the tx as rlp:

https://github.com/paradigmxyz/reth/blob/d5f59070bb0be1cf9bba94e01b895fede1ae7078/crates/transaction-pool/src/maintain.rs#L661-L661

what we should do instead is store these as objects so that we can store rlp + propagated

the reload part should however be backwards compat for a bit, so we should also attempt to rlp decode the contents in case that failes.

todo:

introduce helper type that is `struct TxBackup{rlp, propagate}` and the we store it as json vec instead
the additional overhead should be negligible here 

### Additional context

_No response_

iTranscend:
resolves https://github.com/paradigmxyz/reth/issues/17721

This PR introduces a helper type `TxBackup` that retains the origin for propagation setting when transactions in the mempool are backed up on shutdown.

`TxBackup` is stored as json and reloading is backwards compatible with the previous rlp-only encoded list.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -24,8 +24,7 @@ use reth_primitives_traits::{
 };
 use reth_storage_api::{errors::provider::ProviderError, BlockReaderIdExt, StateProviderFactory};
 use reth_tasks::TaskSpawner;
-// #[cfg(feature = "serde_json")]
-// use serde_json;
+use serde::{Deserialize, Serialize};
 use std::{
     borrow::Borrow,
     collections::HashSet,
@@ -38,9 +37,6 @@ use tokio::{
     time::{self, Duration},
 };
 use tracing::{debug, error, info, trace, warn};
-
-#[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
 
 /// Maximum amount of time non-executable transaction are queued.
 pub const MAX_QUEUED_TRANSACTION_LIFETIME: Duration = Duration::from_secs(3 * 60 * 60);
@@ -640,7 +636,13 @@ where
                     let pool_tx =
                         <P::Transaction as PoolTransaction>::try_from_consensus(recovered).ok()?;
 
-                    Some((backup.origin, pool_tx))
+                    let origin = if backup.propagate {
+                        TransactionOrigin::Local
+                    } else {
+                        TransactionOrigin::Private
+                    };
+
+                    Some((origin, pool_tx))
                 })
                 .collect()
         } else {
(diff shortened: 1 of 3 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -622,25 +623,63 @@ where
...
+    info!(target: "txpool", txs_file =?file_path, num_txs=%num_txs, "Successfully reinserted local transactions from file");
     reth_fs_util::remove_file(file_path)?;
     Ok(())
 }
 
+/// A transaction backup that is saved as json to a file for
+/// reinsertion into the pool
+#[derive(Deserialize, Serialize)]
+struct TxBackup {
+    rlp: Vec<u8>,
```
mattsse: this serializes as [], we should use Bytes here

crates/transaction-pool/src/maintain.rs
```diff
@@ -622,25 +623,63 @@ where
...
-    info!(target: "txpool", txs_file =?file_path, num_txs=%outcome.len(), "Successfully reinserted local transactions from file");
+    info!(target: "txpool", txs_file =?file_path, num_txs=%num_txs, "Successfully reinserted local transactions from file");
     reth_fs_util::remove_file(file_path)?;
     Ok(())
 }
 
+/// A transaction backup that is saved as json to a file for
+/// reinsertion into the pool
+#[derive(Deserialize, Serialize)]
+struct TxBackup {
```
mattsse: we can make this one pub because we write this type to file

crates/transaction-pool/src/maintain.rs
```diff
@@ -622,25 +623,63 @@ where
...
+    rlp: Vec<u8>,
+    propagate: bool,
```
mattsse: we can actually store the origin here then we dont lose any context

mattsse (changes requested):
cool, some nits

crates/transaction-pool/src/maintain.rs
iTranscend: true; was leaning towards that initially. will do

mattsse (approved):
ty!

PR #17758, opened 2025-08-08
feat(rpc): add Other(String) variant to RethRpcModule for custom RPC modules

fgimenez:
Adds support for custom RPC modules by introducing an `Other(String)` variant to `RethRpcModule`. This allows users to configure and use RPC modules beyond the predefined set, enabling greater extensibility for custom implementations.

The PR also includes validation to warn users when they configure RPC modules that haven't been registered with the RPC builder, helping catch configuration errors early.


docs/vocs/docs/pages/cli/reth/node.mdx
```diff
@@ -258,7 +258,7 @@ RPC:
       --http.api <HTTP_API>
           Rpc Modules to be configured for the HTTP server
 
-          [possible values: admin, debug, eth, net, trace, txpool, web3, rpc, reth, ots, flashbots, miner, mev]
+          [possible values: admin, debug, eth, net, trace, txpool, web3, rpc, reth, ots, flashbots, miner, mev, other]
```
mattsse: this seems weird, this shouldn't get displayed, this is coming from

let values = RethRpcModule::all_variant_names().

I believe

crates/rpc/rpc-builder/src/lib.rs
```diff
@@ -1956,6 +1960,73 @@ impl TransportRpcModules {
...
         Ok(())
     }
+
+    /// Validates that all configured modules have been registered with actual RPC methods.
+    ///
+    /// This method checks for configured but unregistered modules (especially custom/Other modules)
+    /// and logs warnings for any modules that were configured but never had methods registered.
+    ///
+    /// Returns a list of unregistered module names for each transport (http, ws, ipc).
+    pub fn validate_configured_modules(&self) -> (Vec<String>, Vec<String>, Vec<String>) {
```
mattsse: hmm, not obvious to me what this does, or why we need this

I think checking by method_names.count == 0 is a bit weird

crates/rpc/rpc-builder/src/lib.rs
```diff
@@ -981,7 +980,9 @@ where
                         // only relevant for Ethereum and configured in `EthereumAddOns`
                         // implementation
                         // TODO: can we get rid of this here?
-                        RethRpcModule::Flashbots => Default::default(),
+                        // Custom modules are not handled here - they should be registered via
+                        // extend_rpc_modules
+                        RethRpcModule::Flashbots | RethRpcModule::Other(_) => Default::default(),
```
mattsse: I think we should actually skip this here

mattsse (changes requested):
introducing a new string based variant makes sense,

however we need a few additional cli level parse tests here that ensure that this behaves as expected,

and the validation part is a bit to restrict, if we skip ::Other then we likely dont need this?

however validation here is missing, the problematic part is something like `--http.api eht` which is clearly a typo that we currently dont catch,

since this is CLI specific, we need to make this behaviour somehow configurable.
my suggestion here would be introducing something liek

```
RethCliParsers {
  type ChainSpec:
  type RpcModules
}
```
and replacing C: ChainSpecParser with that

https://github.com/paradigmxyz/reth/blob/61662098aa7acb110a78d8fc8dfc0f6e6ed7de3b/crates/ethereum/cli/src/interface.rs#L30-L30

then we can configure how rpc modules are parsed

docs/vocs/docs/pages/cli/reth/node.mdx
fgimenez: exactly, fixed here https://github.com/paradigmxyz/reth/pull/17758/commits/3462b40033bedc39ba695d0b62083dab056df7c2

crates/rpc/rpc-builder/src/lib.rs
fgimenez: yep, removed

crates/rpc/rpc-builder/src/lib.rs
fgimenez: indeed, skipped now ptal

crates/cli/util/src/parsers.rs
```diff
@@ -1,13 +1,31 @@
...
+/// Trait for configuring CLI parsers across reth.
+///
+/// This trait provides configurable parsers for various CLI components,
+/// allowing customization of validation behavior.
+pub trait RethCliParsers: Clone + Send + Sync + 'static {
+    /// The chain specification parser type.
+    type ChainSpecParser: reth_cli::chainspec::ChainSpecParser + std::fmt::Debug;
+
+    /// The RPC module validator type.
+    type RpcModuleValidator: reth_rpc_server_types::RpcModuleValidator;
```
mattsse, line 20: should we follow the same convention as ChainSpecParser here and move the parse_rpc_modules in the trait?

crates/rpc/rpc-server-types/src/module.rs
```diff
@@ -347,10 +396,109 @@ impl Serialize for RethRpcModule {
...
+    // - Distance is 2 and strings are at least 3 chars (for transpositions like "eth" -> "eht")
+    match distance {
+        1 => true,
+        2 => input.len() >= 3 && target.len() >= 3,
+        _ => false,
+    }
+}
+
+/// Calculate edit distance between two strings.
+fn edit_distance(s1: &str, s2: &str) -> usize {
```
mattsse: dont think we need this

either the user input is valid, or it's not, no need to check for typos

crates/rpc/rpc-builder/src/lib.rs
```diff
@@ -915,81 +915,95 @@ where
...
-                                    EthBundle::new(
-                                        eth_api.clone(),
-                                        self.blocking_pool_guard.clone(),
-                                    )
-                                    .into_rpc(),
-                                )
-                                .expect("No conflicts");
+            .filter_map(|namespace| {
+                // Skip Other variants - they should be registered via extend_rpc_modules
+                if matches!(namespace, RethRpcModule::Other(_)) {
```
mattsse: I prefer introducing is_other functions than adding matches! for checks like

mattsse (changes requested):
after reviewing I understood why we don't integrate an rpcmoduleparser, we can't really propagate the module parser down to the rpcserverargs because then we leak an unconstrained generic all over the place, right?

so perhaps we enforce this either when we do `cli::parse` or we enforce this in the Nodecommand::run somewhere early

so we do need to instead validate the parsed modules after somewhere, unsure where the proper location for that is.

maybe this should actually be enforced by the parser impl for the Cli itself?

crates/cli/util/src/parsers.rs
fgimenez: makes sense, moved

crates/rpc/rpc-builder/src/lib.rs
fgimenez: yep, much beter, done

crates/rpc/rpc-server-types/src/module.rs
fgimenez: ok, removed

fgimenez (commented):
> so perhaps we enforce this either when we do cli::parse or we enforce this in the Nodecommand::run somewhere early

thx, great catch! now this is done after cli parsing, in `with_runner_and_components` during `NodeCommand::run`, ptal

mattsse (commented):
great, I like this

I'd like for @klkvr to take a look as well before we merge this

since this is breaking we should time merging this appropriately

crates/node/core/src/version.rs
```diff
@@ -113,7 +113,7 @@ pub fn default_reth_version_metadata() -> RethCliVersionConsts {
         vergen_git_sha: Cow::Owned(env!("VERGEN_GIT_SHA_SHORT").to_string()),
         vergen_build_timestamp: Cow::Owned(env!("VERGEN_BUILD_TIMESTAMP").to_string()),
         vergen_cargo_target_triple: Cow::Owned(env!("VERGEN_CARGO_TARGET_TRIPLE").to_string()),
-        vergen_cargo_features: Cow::Owned(env!("VERGEN_CARGO_FEATURES").to_string()),
+        vergen_cargo_features: Cow::Owned(String::new()),
```
klkvr, line 116: hmm not sure about this

crates/ethereum/cli/src/interface.rs
```diff
@@ -27,11 +27,10 @@ use tracing::info;
 /// This is the entrypoint to the executable.
 #[derive(Debug, Parser)]
 #[command(author, version =version_metadata().short_version.as_ref(), long_version = version_metadata().long_version.as_ref(), about = "Reth", long_about = None)]
-pub struct Cli<C: ChainSpecParser = EthereumChainSpecParser, Ext: clap::Args + fmt::Debug = NoArgs>
-{
+pub struct Cli<P: RethCliParsers = EthereumCliParsers, Ext: clap::Args + fmt::Debug = NoArgs> {
```
klkvr, line 30: I'm wondering if a less invasive change would be to do `Cli<C: ChainSpecParser = EthereumChainSpecParser, Ext: clap::Args + fmt::Debug = NoArgs, R: RpcModuleValidator = DefaultRpcModuleValidator>`

That way we won't need to change most of the cli code

fgimenez:
closing in favor of https://github.com/paradigmxyz/reth/pull/18160

PR #17760, opened 2025-08-08
docs(stages): fix MerkleStage doc link to StorageHashingStage

dizer-ti:
Correct the intra-doc link in merkle.rs from AccountHashingStage to StorageHashingStage so it points to the correct module


mattsse (commented):
not worth it

PR #17764, opened 2025-08-08
chore: Expose payload builder handle and metrics

SozinM:
1. I've exposed payload builder handle. In our core we use it to add intermidiate executed blocks in the cache (we cannot guarantee which executed block will be included to the chain, so we need to preserve all of them, so we would use cache for executed block when it arrives trough p2p or newPayload).
2. Added metrics that are useful for builders, FCU with attributes and number of cache misses for executed block.


mattsse (approved):
makes sense

PR #17766, opened 2025-08-08
fix(network): push header before next header check in `get_headers_response`

colinlyguo:
This PR is a robustness fix that pushes the header to result before checking the next header in `get_headers_response`. Previously, when hitting a limit, one valid header was omitted from the response.

The spec: https://github.com/ethereum/devp2p/blob/40ab248bf7e017e83cc9812a4e048446709623e8/caps/eth.md#blockheaders-0x04


colinlyguo:
@mattsse @Rjected could you help take a look?

mattsse (approved):
I see, makes sense, ty!

PR #17767, opened 2025-08-08
chore: rename gas to gas_used in the node logs

schoentgenaa:
From the reth logs, I found the `gas` field was used to represent the gas used in the block, but it was not clear enough, so try to rename it to `gas_used` for clarity.

```
INFO Block added to canonical chain number=23092863 hash=0xa977637bc74218fc35ff21893bcd32b63244f7348aacd4efa4f7abcebd816471 peers=4 txs=201 gas=22.28 Mgas gas_throughput=223.76 Mgas/second gas_limit=44.96 Mgas full=49.6% base_fee=0.41gwei blobs=0 excess_blobs=0 elapsed=99.581054ms
```

BTW, should we remove the whitespace between the number and unit, like `gas_throughput=223.76 Mgas/second` to `gas_throughput=223.76Mgas/second`?


mattsse (commented):
makes sense, also on removing whitespace would be nic

wdyt @Rjected @shekhirin

Rjected (approved):
lgtm, also supportive of removing the whitespace

PR #17768, opened 2025-08-08
feat: arb-reth

tiljrd:
This PR is a draft for adding arb-reth, a version that can be an alternative execution client in addition to nitro geth.


tiljrd:
> @mattsse do we want this?

Happy to close this PR for now since we'll anyway use reth as an sdk instead

PR #17769, opened 2025-08-08
docs: fix grammatical error in pending.rs comment

sashaodessa:
**What was changed:**

- Changed: `/// could potentially be become invalid at point of execution`
- To: `/// could potentially become invalid at point of execution`

**Files affected:**
- `crates/transaction-pool/src/pool/pending.rs`


mattsse (commented):
I draw the line at single typos

PR #17771, opened 2025-08-08
chore(cli-docs): remove extra prompt char from update.sh log output

crazykissshout:
Removes an unnecessary “$” from the log line in docs/cli/update.sh, prevents confusion when copying commands from logs; aligns with typical logging style.

before 
`  Running: $ <path>/help.rs --root-dir /docs --out-dir /pages/cli reth
`

after

`  Running: <path>/help.rs --root-dir /docs --out-dir /pages/cli reth
`


mattsse (commented):
this is fine because this indicates a command

PR #17774, opened 2025-08-09
Use .as_slice() for parent_beacon_block_root in payload_id hashing

sashass1315:
- Align hashing with other B256 fields by calling .as_slice()
- Edited payload_id to hash parent_beacon_block_root via .as_slice() for consistency with other fields.
- Verified by running crate tests; all green.


mattsse (commented):
this is the same

PR #17776, opened 2025-08-09
chore: replace ~/.cargo with $CARGO_HOME

schoentgenaa:
In the make help output, it outputs `~/.cargo` as the default cargo home, but someone maybe set it to other directories, so try to replace it with `$CARGO_HOME`.


mattsse (approved):
this is better, but unused in the commands, but still

PR #17783, opened 2025-08-10
fix: block_meta was not commited

schoentgenaa:
Seems StaticFileWriters will commit the `headers, transactions, receipts`, but the `block_meta` was not committed, I'm not sure if this is a bug or not?

### DIFF
--- a/crates/storage/provider/src/providers/static_file/writer.rs
+++ b/crates/storage/provider/src/providers/static_file/writer.rs
@@ -66,7 +66,7 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
     }
 
     pub(crate) fn commit(&self) -> ProviderResult<()> {
-        for writer_lock in [&self.headers, &self.transactions, &self.receipts] {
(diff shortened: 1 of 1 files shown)


DaniPopes (dismissed):

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -66,7 +66,7 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
     }
 
     pub(crate) fn commit(&self) -> ProviderResult<()> {
-        for writer_lock in [&self.headers, &self.transactions, &self.receipts] {
+        for writer_lock in [&self.headers, &self.transactions, &self.receipts, &self.block_meta] {
```
mattsse, line 69: this is actually missing

https://github.com/paradigmxyz/reth/blob/6a79d80ec5cfc4fea22bf889e39a388a880571b6/crates/storage/provider/src/providers/static_file/writer.rs#L234-L234

unsure why @joshieDo

mattsse (commented):
okay, turns out this is obsolete with

https://github.com/paradigmxyz/reth/issues/17828

PR #17784, opened 2025-08-10
fix: reth dev node implement the `--block-max-transactions` arg

schoentgenaa:
This argument seems was not implemented in the reth dev node, so try to implement it.


mattsse (approved):
makes sense, although this doesnt guarantee that the block will have exactly max_transactions, but it's better than nothing

PR #17789, opened 2025-08-11
docs(e2s_file): clarify automatic version insertion and entries behavior

Bashmunta:
Update write_version docs to state the version is auto-written before the first non-version entry if not explicit.
Change entries docs to “Read all entries” (it returns Vec<Entry>, not an iterator).


mattsse (approved):
I let this one slide because this is indeed more accurate

mattsse (approved):

PR #17790, opened 2025-08-11
fix: use StateCommitment

robinsdan:
The current implementation ignored the `StateCommitment` constraint

### DIFF
--- a/crates/storage/provider/src/providers/state/historical.rs
+++ b/crates/storage/provider/src/providers/state/historical.rs
@@ -19,11 +19,8 @@ use reth_storage_api::{
 };
 use reth_storage_errors::provider::ProviderResult;
 use reth_trie::{
-    proof::{Proof, StorageProof},
-    updates::TrieUpdates,
-    witness::TrieWitness,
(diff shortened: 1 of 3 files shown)


mattsse (commented):
i dont understand what this fixes, can you elaborate

does this make sense @Rjected

robinsdan:
@mattsse Wasn't the introduction of the `StateCommitment` trait intended to abstract the implementation of the Database? However, the current implementation uses a fixed type, `StateRoot`, `StorageRoot`, etc, which assumes that `StateCommitment` is equivalent to `MerklePatriciaTrie`. https://github.com/paradigmxyz/reth/blob/fe53ec5be98e4d7c49c4dce645b11dfc81e94587/crates/trie/db/src/commitment.rs#L27-L39

Rjected:
The intention of this abstraction was to allow customization of the state root algorithm yes, but I think it has failed at doing that because there are too many parts of the code which assume MPT and the abstraction only handles certain db related traits. I am going to close this PR as I think it makes the code more verbose, but I would instead support removing the `StateCommitment` trait

Issue #17545, opened 2025-07-21
Refactor `RpcConverter`


klkvr:
### Describe the feature

We have recently abstracted all of the RPC conversions and intorudced `RpcConverter` type:
https://github.com/paradigmxyz/reth/blob/39f1ee879517eae7d377ef5905c8bf517aab5110/crates/rpc/rpc-convert/src/transaction.rs#L390-L410

It is still somewhat messy and is not super easy to build. What we want is to have 2 ways to define each of the conversions (header, receipt, tx, sim tx, txenv):
a. by directly implementing some trait on a primitive type (e.g impl `TryIntoTxEnv` on `TransactionRequest`)
b. by having a closure or fully custom type that does the conversion and potentially contains some useful context

What we can do is have something like `RpcConverter<Tx = (), TxEnv = (), SimTx = (), Receipt = (), Header = ()>` where each of the generics is a certain converter, for example:

`Tx` generic should implement `trait RpcTxConverter<PrimitiveTx, RpcTx> { ... }` which has 2 blanket impls
1. for `()` where `RpcTx: FromConsensusTx<PrimitiveTx>`
2. for `impl Fn(PrimitiveTx) -> RpcTx` 

then we should also have a helper method `with_tx_converter<NewTx>(self, new: NewTx) -> RpcConverter<NewTx, ...>` on `RpcConverter` to set custom values.

That way, `RpcConverter::default()` would be a fully valid type as long as you implement all of the needed traits on your rpc/primitive types, and if you don't do that because of foreign type rules or needing more context, you can still override all of the `()`'s in a more verbose way via a nice builder pattern.

Also, worth exploring whether the converter traits should get `N: RpcNodeCore` or provider as input to extend their basic capabilities

lean-apple:
Hello, may I try?

mattsse:
marking this as completed

PR #17792, opened 2025-08-11
refactor(rpc): add `TxEnv` converter to `RpcCoverter`

lean-apple:
Part 1 to go towards https://github.com/paradigmxyz/reth/issues/17545.

Still to do:
- [x] Add closure support,
- [x]  Update doc

Add convertion support only for `TxEnv`, other conversions will be done in next prs.


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -382,6 +382,66 @@ impl TryIntoTxEnv<TxEnv> for TransactionRequest {
...
+
+    fn convert_tx_env<Spec>(
+        &self,
+        request: RpcReq,
+        cfg_env: &CfgEnv<Spec>,
+        block_env: &BlockEnv,
+    ) -> Result<T, Self::Error> {
+        //TODO: check if we should use only the chain id - and some other specific fields
+        // or the whole `CfgEnv`
+        let cfg_env_base = unsafe { std::mem::transmute::<&CfgEnv<Spec>, &CfgEnv<()>>(cfg_env) };
```
mattsse: don't think we can do this because there's no guarantee that spec is sized

do we need this closure impl?, or can we add another generic instead of () for the spec

mattsse (changes requested):

crates/rpc/rpc-convert/src/transaction.rs
lean-apple: Added as a generic to match the initial issue requirements, but ok to remove it also, 
what do you think @RomanHodulak ?

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -137,10 +137,10 @@ pub trait RpcConvert: Send + Sync + Unpin + Clone + Debug + 'static {
 
     /// Creates a transaction environment for execution based on `request` with corresponding
     /// `cfg_env` and `block_env`.
-    fn tx_env<Spec>(
+    fn tx_env(
         &self,
         request: RpcTxReq<Self::Network>,
-        cfg_env: &CfgEnv<Spec>,
+        cfg_env_chain_id: u64,
```
mattsse: we cant really make that change because this can cause side-effects because then we create the cfgenv twice, so we need to keep this as is

mattsse (changes requested):
can we undo the function signature change

crates/rpc/rpc-convert/src/transaction.rs
RomanHodulak: I think we should keep it as it is part of the input for the conversion.

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -369,6 +369,78 @@ impl TryIntoSimTx<EthereumTxEnvelope<TxEip4844>> for TransactionRequest {
...
+        block_env: &BlockEnv,
+    ) -> Result<TxEnv, Self::Error> {
+        tx_req.try_into_tx_env(cfg_env, block_env)
+    }
+}
+
+/// Converts rpc transaction requests into transaction environment using a closure.
+impl<F, TxReq, TxEnv, E> TxEnvConverter<TxReq, TxEnv> for F
+where
+    F: Fn(TxReq, &CfgEnv<()>, &BlockEnv) -> Result<TxEnv, E>
```
mattsse: where would this be applicable?

cant think of a usecase where we have () for spec

mattsse (commented):
question about the closure impl

this doesnt seem useful because cfg<()> isnt a thing I believe?

crates/rpc/rpc-convert/src/transaction.rs
lean-apple: Indeed, it does not make sense also for me, let's remove it

klkvr (approved):

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -369,6 +369,45 @@ impl TryIntoSimTx<EthereumTxEnvelope<TxEip4844>> for TransactionRequest {
...
 
+/// Converts `TxReq` into `TxEnv`.
+///
+/// Where:
+/// * `TxReq` is a transaction request received from an RPC API
+/// * `TxEnv` is the corresponding transaction environment for execution
+///
+/// The `TxEnvConverter` has a blanket implementation:
+/// * `()` assuming `TxReq` implements [`TryIntoTxEnv`] and is used as default for [`RpcConverter`].
+pub trait TxEnvConverter<TxReq, TxEnv>: Debug + Send + Sync + Unpin + Clone + 'static {
```
klkvr: can we add a blanket implementation for closures in scope of this PR?

mattsse: I don't think this works because this requires `Spec` generic,

should we make this an AT as well?

klkvr: ah right

yeah we can make `Spec` an AT on `RpcConvert`, and do `TxEnvConverter<TxReq, TxEnv, Spec>`. we should know the spec type from the `Evm`

mattsse (approved):
lgtm!

PR #17795, opened 2025-08-11
feat(node): add --override.<hardfork> cli args

### LINKED ISSUE #16523: Add --override.<hardfork> cli args
### Describe the feature

geth and op-geth have additional args to overrid hardfork activations

https://github.com/ethereum-optimism/op-geth/blob/d200eea2a5adce482c4daf36210fea512c52eb38/cmd/utils/flags.go#L287-L291

https://github.com/ethereum/go-ethereum/blob/8b9f2d4e36dfcdb2af94dc0d3b63f30c8ad159a9/cmd/utils/flags.go#L241-L245

which are quite useful for testing

we should add these as well

## TODO
* add define cli in `RollupArgs` for opstack and for eth in a new Args type https://github.com/paradigmxyz/reth/blob/3336e56b096c8a5a49620f019ea665f3071aeffe/crates/node/core/src/args/mod.rs#L1-L1

cc @VeerChaurasia 

### Additional context

_No response_

mablr:
## Motivation
Close #16523

## Solution
- Added a new trait `HardforksOverrides` to apply hardfork timestamp overrides to chainspec.
- Updated relevant structures and CLI arguments to support these overrides for both Ethereum and OP chainspecs.
- Added unit tests on CLI args

## Remarks
- It may be necessary to restrict the `--override.<hardfork>` cli args to dev mode
- Eth Hardfork overrides are applied in `NodeCommand` impl as they are defined in `EthereumHardforkOverrideArgs`
- OP ones are applied in the op-reth main bin as as they are defined in `RollupArgs`
- A follow-up task may be to merge `HardforksOverrides` into alloy's existing `EthereumHardforks` trait
- It would be nice to add integration test, but I am not sure at which level (Cli, NodeCommand...)

### DIFF
--- a/crates/optimism/bin/src/main.rs
+++ b/crates/optimism/bin/src/main.rs
@@ -1,6 +1,7 @@
 #![allow(missing_docs, rustdoc::missing_crate_level_docs)]
 
 use clap::Parser;
+use reth_optimism_chainspec::HardforksOverrides;
 use reth_optimism_cli::{chainspec::OpChainSpecParser, Cli};
 use reth_optimism_node::{args::RollupArgs, OpNode};
 use tracing::info;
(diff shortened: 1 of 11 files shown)


crates/optimism/bin/src/main.rs
```diff
@@ -17,8 +18,16 @@ fn main() {
...
     if let Err(err) =
-        Cli::<OpChainSpecParser, RollupArgs>::parse().run(async move |builder, rollup_args| {
+        Cli::<OpChainSpecParser, RollupArgs>::parse().run(async move |mut builder, rollup_args| {
             info!(target: "reth::cli", "Launching node");
+
+            // Apply OP hardfork overrides to the chainspec before launching
+            let op_hardfork_overrides = rollup_args.op_hardforks_as_vec();
+            if !op_hardfork_overrides.is_empty() {
+                builder.config_mut().chain =
+                    builder.config().chain.apply_hardforks_overrides(op_hardfork_overrides).into();
```
mablr, line 28: There is maybe a cleaner way to update chainspec

mattsse (commented):
okay, this impl looks pretty good, but I'm actually no longer sure if we actually still need this

mattsse (commented):
fusaka devnets are coming to an and there was no use case for this after all

sorry about this @mablr

PR #17799, opened 2025-08-11
chore: use proper error handling instead of unwrap to prevent receipt handling panic

schoentgenaa:
There could panic if ther iterator of `receipts_iter.next()` was exhausted or if there was a mismatch between expected transaction numbers and actual receipts, so try to use error matching instead of `unwrap` to prevent the panic.


schoentgenaa:
@Rjected @mattsse It will be appreciated if you could look at this.

PR #17800, opened 2025-08-11
chore(test-vectors): remove TxDeposit compact TODO

schoentgenaa:
Add the op TxDeposit compact test

### DIFF
--- a/crates/cli/commands/Cargo.toml
+++ b/crates/cli/commands/Cargo.toml
@@ -15,7 +15,7 @@ reth-chainspec.workspace = true
 reth-cli.workspace = true
 reth-cli-runner.workspace = true
 reth-cli-util.workspace = true
-reth-codecs.workspace = true
+reth-codecs = { workspace = true, features = ["op"] }
 reth-config = { workspace = true, features = ["serde"] }
 reth-consensus.workspace = true
 reth-db = { workspace = true, features = ["mdbx"] }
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/test_vectors/compact.rs
```diff
@@ -94,7 +95,7 @@ compact_types!(
         Log,
         // BranchNodeCompact, // todo requires arbitrary
         TrieMask,
-        // TxDeposit, TODO(joshie): optimism
+        TxDeposit,
```
Rjected:
```suggestion
```

I would rather remove this TODO, we should be special casing this in the op-reth test-vectors command, we also shouldn't enable the `op` reth-codec feature by default
https://github.com/paradigmxyz/reth/blob/81d76aa565df6a20dbc72d32f7ebf7c898f49332/crates/optimism/cli/src/commands/test_vectors.rs#L57-L61

Rjected (changes requested):

crates/cli/commands/Cargo.toml
```diff
@@ -15,7 +15,7 @@ reth-chainspec.workspace = true
 reth-cli.workspace = true
 reth-cli-runner.workspace = true
 reth-cli-util.workspace = true
-reth-codecs.workspace = true
+reth-codecs = { workspace = true }
```
Rjected:
```suggestion
reth-codecs.workspace = true
```

Rjected (changes requested):
just a nit otherwise this looks good

schoentgenaa:
> just a nit otherwise this looks good

Thanks!

Rjected (approved):
lgtm

PR #17801, opened 2025-08-11
feat(metrics): add sstore/sload opcode metrics

schoentgenaa:
Add metrics for the `sstore` and `sload` opcodes.


schoentgenaa:
@Rjected @mattsse Would you plz take a look at this? :)

PR #17802, opened 2025-08-11
feat(tx-pool): add `add_transactions_with_origins` helper

### LINKED ISSUE #17794: Add helper for inserting `Vec<(origin, tx)>` to poll
### Describe the feature

we have 

https://github.com/paradigmxyz/reth/blob/69a1951f54e6f0c92e1968c6a6103362622e681a/crates/transaction-pool/src/traits.rs#L168-L177

but no way to individually specify the origin

but we have 

https://github.com/paradigmxyz/reth/blob/69a1951f54e6f0c92e1968c6a6103362622e681a/crates/transaction-pool/src/validate/mod.rs#L200-L207

on the validation fn

## TODO
* add missing trait function for `Vec<(TransactionOrigin, Self::Transaction)>, `

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17794.

### DIFF
--- a/crates/transaction-pool/src/lib.rs
+++ b/crates/transaction-pool/src/lib.rs
@@ -298,6 +298,7 @@ use alloy_eips::{
 };
 use alloy_primitives::{Address, TxHash, B256, U256};
 use aquamarine as _;
+use futures_util::future::join_all;
 use reth_chainspec::{ChainSpecProvider, EthereumHardforks};
 use reth_eth_wire_types::HandleMempoolData;
 use reth_execution_types::ChangedAccount;
@@ -504,6 +505,17 @@ where
         let validated = self.validate_all(origin, transactions).await;
 
         self.pool.add_transactions(origin, validated.into_iter().map(|(_, tx)| tx))
+    }
+
+    async fn add_transactions_with_individual_origins(
+        &self,
+        transactions: Vec<(TransactionOrigin, Self::Transaction)>,
+    ) -> Vec<PoolResult<AddedTransactionOutcome>> {
+        if transactions.is_empty() {
+            return Vec::new()
+        }
+        join_all(transactions.into_iter().map(|(origin, tx)| self.add_transaction(origin, tx)))
+            .await
     }
 
     fn transaction_event_listener(&self, tx_hash: TxHash) -> Option<TransactionEvents> {
--- a/crates/transaction-pool/src/lib.rs
+++ b/crates/transaction-pool/src/lib.rs
@@ -298,7 +298,6 @@ use alloy_eips::{
 };
 use alloy_primitives::{Address, TxHash, B256, U256};
 use aquamarine as _;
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/lib.rs
```diff
@@ -506,6 +507,17 @@ where
...
+        &self,
+        transactions: Vec<(TransactionOrigin, Self::Transaction)>,
+    ) -> Vec<PoolResult<AddedTransactionOutcome>> {
+        if transactions.is_empty() {
+            return Vec::new()
+        }
+        join_all(transactions.into_iter().map(|(origin, tx)| self.add_transaction(origin, tx)))
+            .await
+    }
+
```
lean-apple: not sure using the https://github.com/paradigmxyz/reth/blob/69a1951f54e6f0c92e1968c6a6103362622e681a/crates/transaction-pool/src/validate/mod.rs#L200-L207 will be so efficient as we have to retrieve the origin after and select  the one to push to `add_transactions` (or group per origin), or do we need to modify `PoolInner::add_transactions` to handle multiple origins per tx?

mattsse: yeah here we'd need to use

```
 self.pool
            .validator()
            .validate_transactions
```

and also add another helper next to

https://github.com/paradigmxyz/reth/blob/main/crates/transaction-pool/src/pool/mod.rs#L582-L586

which accepts this iter type, then we can keep the logic entirely and make ` pub fn add_transactions(` just forward as `transactions.into_iter.map()`

mattsse (changes requested):
smol suggestion

crates/transaction-pool/src/lib.rs
lean-apple: https://github.com/paradigmxyz/reth/pull/17802/commits/2c35e6e8fbe91d71e42cbe64645afce84c1f539e

crates/transaction-pool/src/lib.rs
```diff
@@ -379,17 +391,16 @@ where
...
-        transaction: V::Transaction,
-    ) -> (TxHash, TransactionValidationOutcome<V::Transaction>) {
-        let hash = *transaction.hash();
-
-        let outcome = self.pool.validator().validate_transaction(origin, transaction).await;
-
-        (hash, outcome)
+        transactions: Vec<(TransactionOrigin, V::Transaction)>,
+    ) -> Vec<(TransactionOrigin, TransactionValidationOutcome<V::Transaction>)> {
+        let origins: Vec<_> = transactions.iter().map(|(origin, _)| *origin).collect();
```
mattsse, line 401: I see,

hmm, we could perhaps consider making this part of the response, but this isn't too bad because this usecase isn't that common

mattsse (approved):
ty, lgtm!

PR #17804, opened 2025-08-11
Prevent starvation of RlpxSatelliteStream

jnoorchashm37:
Spamming messages to the underlying subprotocols in `RlpxSatelliteStream` caused starvation - fix by moving the inner `P2PStream` sends into the `Sink` impl on the `RlpxSatelliteStream`


crates/net/network/src/session/active.rs
mattsse, line 1: please undo

mattsse (changes requested):
only style nits,

can't apply changes because opened from main,

mind opening this from a different branch instead

crates/net/network/src/session/active.rs
jnoorchashm37: Sorry - my formatter was being wierd: https://github.com/paradigmxyz/reth/pull/17830

jnoorchashm37:
PR from different branch - https://github.com/paradigmxyz/reth/pull/17830

PR #17808, opened 2025-08-12
chore: remove whitespace of the gas unit and threshold

schoentgenaa:
This is a follow up PR of https://github.com/paradigmxyz/reth/pull/17767


Rjected (approved):
lgtm, thanks!

PR #17811, opened 2025-08-12
fix: remove unused import from execute.rs

phrwlk:
Removed unused import core::fmt::Debug from crates/evm/evm/src/execute.rs.


mattsse (approved):
sure -1 not bad

PR #17816, opened 2025-08-12
perf: optimize condition ordering in ParkedPool for better short-circuiting

malik672:
Reorder conditions in ParkedPool::truncate_pool to leverage logical short-circuiting.

Using LLVMCA the diff is:

1000 fewer instructions when pool is empty
1384 fewer CPU cycles per check
700 fewer micro-operations

https://godbolt.org/z/oTYY5h3v5


DaniPopes (approved):
Sure, `limit.is_exceeded` is 2 integer comparisons but whatever

PR #17819, opened 2025-08-12
fix: import should count on the delta

tmelhao:
We're using the count in table to count the imported blocks, this is incorrect, we should use the delta count to check.


DaniPopes:
Taking a look, I can reproduce with the exact command that CI runs and stable rust toolchain, but nothing else

tmelhao:
@mattsse ping

DaniPopes (approved):
Makes sense, leaving to @mattsse

mattsse (approved):
sure

tmelhao:
ping @mattsse

PR #17820, opened 2025-08-12
perf: iterate through nibbles

### LINKED ISSUE #17810: perf: avoid allocation when storing nibbles
### Describe the feature

In this file `.0.to_vec` is used to write both `StoredNibbles` and `StoredNibblesSubKey`. This creates an unnecessary heap allocation when we know that nibbles has a maximum size of 64 (unpacked) bytes.

cc @shekhirin 

### Additional context

_No response_

dawksh:
fix #17810 by iterating through the nibbles and putting to buffer directly without creating a new Vec in the heap

### DIFF
--- a/crates/trie/common/src/nibbles.rs
+++ b/crates/trie/common/src/nibbles.rs
@@ -28,7 +28,9 @@ impl reth_codecs::Compact for StoredNibbles {
     where
         B: bytes::BufMut + AsMut<[u8]>,
     {
-        buf.put_slice(&self.0.to_vec());
+        for i in 0..self.0.len() {
+            buf.put_u8(self.0.get_byte(i).unwrap());
+        }
         self.0.len()
     }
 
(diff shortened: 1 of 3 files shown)


crates/trie/common/src/nibbles.rs
```diff
@@ -77,7 +79,9 @@ impl reth_codecs::Compact for StoredNibblesSubKey {
         assert!(self.0.len() <= 64);
 
         // right-pad with zeros
-        buf.put_slice(&self.0.to_vec());
+        for i in 0..self.0.len() {
+            buf.put_u8(self.0.get_byte(i).unwrap());
```
mattsse: there's get_unchecked

mattsse (changes requested):
this makes sense

@Rjected should we add a helper iterator type for this to nybbles?

EDIT: quickly added this

crates/trie/common/src/nibbles.rs
dawksh: oops, my bad, pushed the changes

mattsse (approved):
lgtm

@Rjected we can either merge this now or wait until we have the iter pr released

mattsse (approved):
ah sigh, there's no extend apparently

so this is okay

Rjected (approved):

PR #17821, opened 2025-08-12
feat(rpc): Add `SimTxConverter` to allow for providing custom converters with extra context

RomanHodulak:
Part of #17545

### Motivation
If the conversion algorithm of a custom node for Rpc Transaction Request -> Simulate Transaction requires more information than just the input, then it's really difficult to accomplish.

The reason is that `RpcConverter` implements `RpcConvert` only if your Transaction Request type implements `TryIntoSimTx`. The `TryIntoSimTx` cannot carry more information than the input. Therefore, if you need to implement the conversion differently, you need to override the entire `RpcConvert` implementation, which is way more than you really need.

### Solution
Adds a new trait called `SimTxConverter` with two blanket implementations:
* `()` assuming `TxReq` implements `TryIntoSimTx` and is used as default for `RpcConverter`
* `Fn(TxReq) -> Result<SimTx, ValueError<TxReq>>` and can be applied using `RpcConverter::with_sim_tx_converter`

### DIFF
--- a/crates/rpc/rpc-convert/src/transaction.rs
+++ b/crates/rpc/rpc-convert/src/transaction.rs
@@ -247,7 +247,7 @@ pub trait SimTxConverter<TxReq, SimTx>: Clone + Debug + Unpin + Send + Sync + 's
 
 impl<TxReq, SimTx> SimTxConverter<TxReq, SimTx> for ()
 where
-    TxReq: TryIntoSimTx<SimTx> + Sized,
+    TxReq: TryIntoSimTx<SimTx>,
 {
     fn convert_sim_tx(&self, tx_req: TxReq) -> Result<SimTx, ValueError<TxReq>> {
         tx_req.try_into_sim_tx()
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -223,6 +223,52 @@ where
...
+///
+/// One should prefer to implement [`TryIntoSimTx`] for `TxReq` to get the `SimTxConverter`
+/// implementation for free, thanks to the blanket implementation, unless the conversion requires
+/// more context. For example, some configuration parameters or access handles to database, network,
+/// etc.
+pub trait SimTxConverter<TxReq, SimTx>: Clone + Debug + Unpin + Send + Sync + 'static {
+    /// Performs the conversion from `tx_req` into `SimTx`.
+    ///
+    /// See [`SimTxConverter`] for more information.
+    fn convert_sim_tx(&self, tx_req: TxReq) -> Result<SimTx, ValueError<TxReq>>;
```
fgimenez: will `ValueError` be enough to cover all the potential error cases now available with `SimTxConverter`? aiui `TryIntoSimTx` models stateless transformations, and `ValueError` is ok to represent problems in the tx fields. But  `SimTxConverter` allows for contextual/stateful conversions, `ValueError` might be very limiting to express what went wrong, wdyt?

RomanHodulak: Good point

mattsse (approved):
lgtm!

PR #17822, opened 2025-08-12
perf: remove some clones around `eth` tracing

hai-rise:
Basically #16665 but for tracing, which is a huge bottleneck for indexers of performance chains. We should also do the equivalence of #16958, and hopefully much more...


DaniPopes (approved):
LGTM

mattsse (approved):
ah yes, this api was necessary earlier

this clone is now super redundant

PR #17825, opened 2025-08-12
feat(ecies): implement constant-time comparison for MAC verification

moricho:
## Summary
This PR changes the MAC verification in the ECIES implementation to use constant-time comparison to prevent timing attacks.

## Background
The previous implementation used regular comparison operators (`!=`) for MAC tag verification. This could potentially cause the following issues:
- Timing attack vulnerability: Regular comparisons exit early when the first mismatch is detected, which could allow attackers to guess MAC tags by exploiting response time differences.
- Security risk: Timing attacks on MAC verification pose a serious security risk, especially in encrypted communications.

## Implementation Details
- HMAC-SHA256 Verification Improvement
  - Before: Calculate tag using `hmac_sha256()` and verify with regular comparison
  - After: Use newly added `verify_hmac_sha256()` function with HMAC's built-in `verify_slice()` method for constant-time comparison
- Ethereum MAC Verification Improvement
  - Before: Calculate MAC using `digest()` and verify with regular comparison
  - After: Use newly added `verify()` method with `subtle::ConstantTimeEq::ct_eq()` for constant-time comparison 
    - `hmac` crate uses [`subtle`](https://github.com/dalek-cryptography/subtle) in a verification step as well

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,7 +133,7 @@ dependencies = [
  "secp256k1 0.30.0",
  "serde",
  "serde_with",
- "thiserror 2.0.12",
+ "thiserror 2.0.14",
 ]
 
 [[package]]
@@ -170,7 +170,7 @@ dependencies = [
  "futures",
  "futures-util",
  "serde_json",
- "thiserror 2.0.12",
+ "thiserror 2.0.14",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


mattsse:
>Timing attack vulnerability: Regular comparisons exit early when the first mismatch is detected, which could allow attackers to guess MAC tags by exploiting response time differences.
Security risk: Timing attacks on MAC verification pose a serious security risk, especially in encrypted communications.

hmm, I don't think these really apply here because this all of this is behind IO anyway


wdyt @Rjected

PR #17826, opened 2025-08-12
chore(cli): rename file import_op to import_core for clarity

tmelhao:
The file import_op is a little confused to the optimism, so rename it to import_core to clarify its purpose.


mattsse (approved):
okay, this kinda makes sense

PR #17827, opened 2025-08-12
feat(rpc): Add `RpcTxConverter` to allow for providing custom converters with extra context

RomanHodulak:
Part of #17545

### Motivation
Custom nodes with custom transactions need their own Consensus Transaction -> Rpc Transaction Response algorithm.

The input for this conversion is the consensus transaction, it's signer address and an associated TxInfo type.

### Solution
Adds a new trait called `RpcTxConverter` with two blanket implementations:
* `()` assuming `Tx` implements `IntoRpcTx` and is used as default for `RpcConverter`
* `Fn(Tx, Address, TxInfo) -> RpcTx` and can be applied using `RpcConverter::with_rpc_tx_converter`

### Discussion
Compared to TxEnv and SimTx, custom conversions are easy to achieve thanks to `TxInfoMapper` being able to generate a custom extra context object that is then used as part of the input for `IntoRpcTx`. This is being actively used for Optimism deposit transactions.

Therefore, it is somewhat questionable how useful this change is. At the very least it is being consistent with `SimTxConverter` and `TxEnvConverter`. To make it more useful, we may merge `TxInfoMapper` with `RpcTxConverter`, but then we're broadening the scope of the custom overrides (just tx info vs rpc tx conversion).

### DIFF
--- a/crates/optimism/rpc/src/error.rs
+++ b/crates/optimism/rpc/src/error.rs
@@ -11,7 +11,7 @@ use reth_rpc_eth_api::{AsEthApiError, EthTxEnvError, TransactionConversionError}
 use reth_rpc_eth_types::{error::api::FromEvmHalt, EthApiError};
 use reth_rpc_server_types::result::{internal_rpc_err, rpc_err};
 use revm::context_interface::result::{EVMError, InvalidTransaction};
-use std::fmt::Display;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm!

PR #17830, opened 2025-08-12
fix: rlpx satellite stream starvation

jnoorchashm37:
Spamming messages to the underlying subprotocols in RlpxSatelliteStream caused starvation - fix by moving the inner P2PStream sends into the Sink impl on the RlpxSatelliteStream


mattsse (approved):
makes sense, now the remaining egress should be handled on poll_ready

there's still an issue that the conn is polled when we already have too many eth requests in flight but we could consider also polling the satellite egress

jnoorchashm37:
> this breaks existing tests, likely because they relay on the the stream being entirely driven by poll_next

Should I change the test here, or is this desired behavior? After looking at the code - the underlying impl should still work in the prod version.

PR #17835, opened 2025-08-12
chore: Remove `BlockMeta` variants and unused code

### LINKED ISSUE #17828: Remove BlockMeta variants and unused code
### Describe the feature

ref #15335

## TODO
* remove all BlockMeta handling such as:

these functions:

https://github.com/paradigmxyz/reth/blob/6a79d80ec5cfc4fea22bf889e39a388a880571b6/crates/storage/provider/src/providers/static_file/writer.rs#L557-L574

and 

https://github.com/paradigmxyz/reth/blob/6a79d80ec5cfc4fea22bf889e39a388a880571b6/crates/static-file/types/src/segment.rs#L41-L43

because this is unused and obsolete since this context is expected to come from era files

### Additional context

_No response_

mablr:
Closes #17828

- Removed `BlockMeta` from `StaticFileSegment` and all related code paths (writer APIs, masks, readers, CLI, producer).
- `block_body_indices` now served only from the database; static-file masks and reads for it were deleted.
- Updated `HighestStaticFiles` and `StaticFileTargets` to drop `block_meta`; adjusted tests accordingly.
- Minor cleanup in S3 static file list.

### Remark
To preserve existing trait bounds (since `BlockReader` requires `BlockBodyIndicesProvider`), I added a minimal impl for `StaticFileProvider` that returns `UnsupportedProvider`. This avoids rippling trait changes while removing `BlockMeta` support, but maybe some further modifications are needed.

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -22,8 +22,8 @@ use reth_chainspec::{ChainInfo, ChainSpecProvider, EthChainSpec};
 use reth_db::{
     lockfile::StorageLock,
     static_file::{
-        iter_static_files, BlockHashMask, BodyIndicesMask, HeaderMask, HeaderWithHashMask,
-        ReceiptMask, StaticFileCursor, TDWithHashMask, TransactionMask,
+        iter_static_files, BlockHashMask, HeaderMask, HeaderWithHashMask, ReceiptMask,
+        StaticFileCursor, TDWithHashMask, TransactionMask,
     },
 };
 use reth_db_api::{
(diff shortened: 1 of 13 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1804,28 +1791,15 @@ impl<N: FullNodePrimitives<SignedTx: Value, Receipt: Value, BlockHeader: Value>>
 }
 
 impl<N: NodePrimitives> BlockBodyIndicesProvider for StaticFileProvider<N> {
```
mattsse, line 1793: do we even still need this trait impl then?

perhaps we can remove entirely?

mattsse (changes requested):
smol q re provider trait impl

also needs make update-book-cli

crates/storage/provider/src/providers/static_file/manager.rs
mablr: I initially wanted to remove it but this breaks the trait bound:

https://github.com/mablr/reth/blob/82f1cc09ff8b9c8c884c168cef931051cba55376/crates/storage/storage-api/src/block.rs#L53-L53

mattsse (approved):
I see

lgtm

PR #17836, opened 2025-08-12
perf: use unwrap and save 198 cycles 

malik672:
Since we've checked that there's at least one sender in the pool, we can use unwrap here and save 198 cycles

https://godbolt.org/z/YK5TxoPs8

### DIFF
--- a/crates/transaction-pool/src/pool/parked.rs
+++ b/crates/transaction-pool/src/pool/parked.rs
@@ -189,7 +189,8 @@ impl<T: ParkedOrd> ParkedPool<T> {
         while limit.is_exceeded(self.len(), self.size()) && !self.last_sender_submission.is_empty()
         {
             // NOTE: This will not panic due to `!last_sender_transaction.is_empty()`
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/parked.rs
```diff
@@ -189,7 +189,8 @@ impl<T: ParkedOrd> ParkedPool<T> {
         while limit.is_exceeded(self.len(), self.size()) && !self.last_sender_submission.is_empty()
         {
             // NOTE: This will not panic due to `!last_sender_transaction.is_empty()`
-            let sender_id = self.last_sender_submission.last().expect("not empty").sender_id;
+            // Unwrap here is safe because we checked that there is at least one sender
```
Rjected: imo the `NOTE` comment is sufficient
```suggestion
```

Rjected (commented):
one very minor nit otherwise lgtm

malik672:
> one very minor nit otherwise lgtm

Done

Rjected (approved):

PR #17838, opened 2025-08-12
refactor: BasicEngineValidator -> EngineApiValidator

anim001k:
renames EngineValidator to EngineApiValidator


mattsse (commented):
not something we want to do

PR #17841, opened 2025-08-12
feat: Stateless validation function receives public keys corresponding to each transaction

kevaundray:
Taken and modified from: https://github.com/risc0/zeth/compare/wollac/wip/pre-recover

This PR changes the interface to use block and public keys, allowing the host to recover the public keys outside of the guest and do signature verification inside of the guest


crates/stateless/src/validation.rs
```diff
@@ -86,11 +87,77 @@ pub enum StatelessValidationError {
...
 }
 
+/// Verifies a transaction using its signature and the given public key.
+///
+/// Note: If the signature or the public key is incorrect, then this method
+/// will return an error.
+///
+/// Returns the address derived from the public key.
+fn recover_sender(
+    vk: &VerifyingKey,
```
kevaundray: This PR requires VerifyingKey which is a secp256k1/k256 specific type that is not exposed via alloy_consensus/crypto AFAICT, so we directly import it here

Wollac: Exactly for that reason I don't believe that is the best place to implement signature verification. It even locks you into using k256 which is undesirable.

Ideally the verification as well as the public key wrapper type should be added to `alloy_consensus/crypto`:
That module already contains `recover_signer_unchecked` (pre Homestead) and `recover_signer` (post Homestead).

Wollac: In my opinion, it should either be implemented in Alloy or left to the user of reth-stateless as it is currently.

crates/stateless/src/validation.rs
```diff
@@ -86,11 +87,77 @@ pub enum StatelessValidationError {
...
+) -> Result<Address, StatelessValidationError> {
+    let sig = tx.signature();
+
+    // non-normalized signatures are only valid pre-homestead
+    let sig_is_normalized = sig.normalize_s().is_none();
+    if is_homestead && !sig_is_normalized {
+        return Err(StatelessValidationError::HomesteadSignatureNotNormalized);
+    }
+
+    sig.to_k256()
```
Wollac: I am not entirely sure, but I believe that k256 verification fails when `s` is not normalized, while before EIP-2 it was explicitly allowed to have non-normalized signatures.

crates/stateless/src/validation.rs
kevaundray: Yep I agree re alloy

crates/stateless/src/validation.rs
kevaundray: I think the function said it accepts values between 1 and N, whereas the homestead check was for values between 1 and N/2 -- Same check can be seen here in geth: https://github.com/ethereum/go-ethereum/blob/3ff99ae52c420477020ae957a61c5c216ac7e7f5/crypto/crypto.go#L298

kevaundray: Inlining code here:

```rust
    /// Returns the inner ECDSA signature.
    #[cfg(feature = "k256")]
    #[inline]
    pub fn to_k256(self) -> Result<k256::ecdsa::Signature, k256::ecdsa::Error> {
        k256::ecdsa::Signature::from_scalars(self.r.to_be_bytes(), self.s.to_be_bytes())
    }

    /// Create a [`Signature`] from the serialized `r` and `s` scalar values
    /// which comprise the signature.
    ///
    /// # Returns
    /// - `Ok(signature)` if the `r` and `s` components are both in the valid
    ///   range `1..n` when serialized as concatenated big endian integers.
    /// - `Err(err)` if the `r` and/or `s` component of the signature is
    ///   out-of-range when interpreted as a big endian integer.
    pub fn from_scalars(r: impl Into<FieldBytes<C>>, s: impl Into<FieldBytes<C>>) -> Result<Self> {
        let r = ScalarPrimitive::from_slice(&r.into()).map_err(|_| Error::new())?;
        let s = ScalarPrimitive::from_slice(&s.into()).map_err(|_| Error::new())?;

        if r.is_zero().into() || s.is_zero().into() {
            return Err(Error::new());
        }

        Ok(Self { r, s })
    }
```
    
The homestead check is also only for `s` value, so might be a bit weird to have the same function be called for `r` too

crates/stateless/src/validation.rs
kevaundray: @mattsse any thoughts on the general PR? opened this up to be more concrete about the changes

crates/stateless/src/validation.rs
kevaundray: We can add a test for this though -- I didn't see an explicit one in the alloy_consesus/crypto.rs module

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -376,6 +386,21 @@ fn pre_execution_checks(
...
+{
+    txs.into_iter()
+        .enumerate()
+        .map(|(i, tx)| {
+            tx.signature()
+                .recover_from_prehash(&tx.signature_hash())
+                .map_err(|e| format!("failed to recover signature for tx #{i}: {e}").into())
+        })
+        .collect::<Result<Vec<_>, _>>()
+}
```
mattsse, line 440: oh I see, this returns the verifying keys

crates/stateless/src/validation.rs
mattsse: how about you we solve this with a trait that also has k256 and secp features that we expect to be in sync with alloy's crypto.

this can even be more advanced like

```rust
trait StatelessSignerRecoverable {

  fn recover_signer<Crypto>(&self, is_homestead: bool)
}
```

where crypto is like

```rust
struct K256;
struct Secp;
```

and then knows how to do it?

crates/stateless/src/validation.rs
mattsse: the way we solved this on reth is that we have unchecked and checked, fns for this

mattsse (commented):
some thoughts on trait support for recovery, but I'm also okay with proceeding with this

crates/stateless/src/validation.rs
kevaundray: I think we would still need secp and k256 to expose VerifyingKey -- I can prototype this to be more concrete

kevaundray: Went with a similar approach but instead changed the public key type to be bytes and then added feature flags on the function

crates/stateless/src/validation.rs
kevaundray: added unchecked and checked

crates/stateless/src/recover_block.rs
```diff
@@ -0,0 +1,130 @@
...
+    let block_hash = block.hash_slow();
+    Ok(RecoveredBlock::new(block, senders, block_hash))
+}
+
+/// Verifies a transaction using its signature and the given public key.
+///
+/// Note: If the signature or the public key is incorrect, then this method
+/// will return an error.
+///
+/// Returns the address derived from the public key.
```
kevaundray, line 52: This function and the ones below it could be placed into alloy next to recover_signer

crates/stateless/Cargo.toml
```diff
@@ -36,3 +36,11 @@ thiserror.workspace = true
...
 serde.workspace = true
 serde_with.workspace = true
+
+k256 = { workspace = true, optional = true }
+secp256k1 = { workspace = true, optional = true }
+
+[features]
+default = ["k256"]
+k256 = ["dep:k256"]
+secp256k1 = ["dep:secp256k1"]
```
mattsse, line 46: k256 is already in the dep graph here so we could just make this the defacto default without feature gating and then use secp if that is enabled

EDIT: nvm I see that k256 should be prioritized over secp?
why?

crates/stateless/src/recover_block.rs
```diff
@@ -0,0 +1,130 @@
...
+        verify_and_compute_sender_unchecked_k256(vk, sig, sig_hash)
+    }
+    #[cfg(not(any(feature = "secp256k1", feature = "k256")))]
+    {
+        let _ = vk;
+        let _ = tx;
+        let _: B256 = sig_hash;
+        let _: &Signature = sig;
+
+        unimplemented!("Must choose either k256 or secp256k1 feature")
```
mattsse, line 85: should this even result in a compile_error! ?

crates/stateless/src/recover_block.rs
```diff
@@ -0,0 +1,130 @@
...
+) -> Result<Address, StatelessValidationError> {
+    let sig = tx.signature();
+
+    // non-normalized signatures are only valid pre-homestead
+    let sig_is_normalized = sig.normalize_s().is_none();
+    if is_homestead && !sig_is_normalized {
+        return Err(StatelessValidationError::HomesteadSignatureNotNormalized);
+    }
+    let sig_hash = tx.signature_hash();
+    #[cfg(all(feature = "k256", feature = "secp256k1"))]
```
mattsse, line 66: so if both are activated we choose k256?

mattsse (commented):
fine with me,

left one more q

crates/stateless/Cargo.toml
kevaundray: No strong opinion here, I think its fine if we can freely switch between the two

crates/stateless/src/recover_block.rs
kevaundray: yeah, this was somewhat arbitrary -- we can switch if you prefer

mattsse: no strong opinion here either, fine with both

crates/stateless/src/recover_block.rs
kevaundray: Tried to go with the suggested cleaner route, though it means that all methods called by this method would have a `cfg(any(secp256k1 || k256))` too -- I may be missing something re making this cleaner without leaking it to all of the methods that call it

mattsse (approved):

PR #17844, opened 2025-08-13
perf(engine): cache prewarming execution results

### LINKED ISSUE #17833: perf: Reuse transaction prewarming results
### Describe the feature

Right now we throw away the results in transaction prewarming, these could actually be reused if we track the state read during the tx, and compare against the state before sequential execution. If the state is the same (except for the beneficiary balance) then we can skip sequential execution for that transaction.

- for prewarming, during sequential execution, we store all the states in mem so that dont have to go to the disk to get the data

- for reusuing the result would be saving the state diff for the transaction and we are going to keep track of what state the trasnaction is read. so when we get to the account during sequential execution, we see if the state is the same as prewarming. if so, we apply the state diff. 

- when it would it not be the same state as prewarming → 2 transactions in a row that update an account, this is because txn executes in parallel and so we cant really cache the state diff because they start with the same state.
- most transactions are independent, therefore theres a larger probability that we can make use of the first order state diff that we cached in parallel - see : https://dependency.pics/index.html

### Additional context
- Initial experiments [here](https://github.com/paradigmxyz/reth/pull/17844). They were not faster, hit rate for the cache was pretty low, I would try to understand why hit rate is low at 60% for `execution results caching`

- I can't think of anything else wrt metrics, but maybe some trace / debug logs that would be useful for figuring things out in more detail. Like, when do we cache certain transactions vs when do we execute them sequentially. But not sure if we have the metrics required to understand the problem fully.

- I was also thinking, maybe this is related to how we chunk transactions in prewarming

### Testing
- For this, it would be time spent in execution actually. So if I were confident in the implementation I'd compare the two branches execution metrics for the same block range against each other. That and trying to increase hit rate.

-  would also be interesting to see if we spend less time in execution as a result. 

- Tracking gas skip and transaction skipped.

Rjected:
WIP, ref https://github.com/paradigmxyz/reth/issues/17833

needs rebase / conflicts fixes


yongkangc:
Initial tests shows cache hit ratio of 57% 
image

PR #17845, opened 2025-08-13
fix(e2c-tests): use chunk check to avoid stack outflow 

tmelhao:
Try to fix the job fail found in this job: https://github.com/paradigmxyz/reth/actions/runs/16923802640/job/47955444747?pr=17819

### DIFF
--- a/crates/rpc/rpc-e2e-tests/src/rpc_compat.rs
+++ b/crates/rpc/rpc-e2e-tests/src/rpc_compat.rs
@@ -523,4 +523,42 @@ mod tests {
         assert!(result.is_err());
         assert!(result.unwrap_err().to_string().contains("Number mismatch"));
     }
+
+    #[test]
+    fn test_compare_json_values_large_array() {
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-e2e-tests/src/rpc_compat.rs
```diff
@@ -118,10 +118,22 @@ impl RunRpcCompatTests {
                             b.len()
                         ));
                     }
-                    // Add array elements to work stack in reverse order
-                    // so they are processed in correct order
-                    for (i, (av, bv)) in a.iter().zip(b.iter()).enumerate().rev() {
-                        work_stack.push((av, bv, format!("{current_path}[{i}]")));
```
mattsse, line 124: hmm, this is a bit odd, unsure what causes larger stack here

mattsse (commented):
wdyt @DaniPopes 

does anything stand out here to you?

PR #17846, opened 2025-08-13
chore(backfill): remove unused BlockHeader import in job.rs

yyhrnk:
Remove use alloy_consensus::BlockHeader; from crates/exex/exex/src/backfill/job.rs
Import is unused, no functional changes
Reduces noise/warnings and keeps the module clean


mattsse (approved):
let's see

PR #17849, opened 2025-08-13
chore(noop): use futures_core::ready for consistency

kurahin:
Replace use of use futures_util::{ready, StreamExt}; with use futures_core::ready; and keep use futures_util::StreamExt;.
Reason: Using use futures_util::{ready, StreamExt}; works, but other parts of the project use futures_core::ready. Unifying imports for consistency. No behavior changes.


mattsse (commented):
this is the same type

PR #17850, opened 2025-08-13
Update generate-jwt.sh

vtjl10:
### Description/Motivation

Improve reliability and security of the JWT token generation script.

- • Reliability: enable `set -euo pipefail` to fail fast on errors and unset variables.
- • Security:
  - create the `jwttoken` directory with strict permissions using `install -d -m 700`;
  - write the token to a single `JWT_FILE` path with `umask 077` and enforce `chmod 600`;
  - avoid `tee` to prevent leaking the secret to stdout.
- • Maintainability: introduce `JWT_FILE` variable, use `[[ ! -f "$JWT_FILE" ]]` for accurate existence checks, and improve the user-facing message.

### DIFF
--- a/etc/generate-jwt.sh
+++ b/etc/generate-jwt.sh
@@ -1,12 +1,16 @@
 #!/bin/bash
+set -euo pipefail
 # Borrowed from EthStaker's prepare for the merge guide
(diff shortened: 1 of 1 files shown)


mattsse (commented):
idk why we need this

PR #17854, opened 2025-08-13
feat: added flashblocks-url to op RollupArgs(wip)

Rimeeeeee:


crates/optimism/node/src/args.rs
```diff
@@ -66,6 +66,10 @@ pub struct RollupArgs {
     /// Minimum suggested priority fee (tip) in wei, default `1_000_000`
     #[arg(long, default_value_t = 1_000_000)]
     pub min_suggested_priority_fee: u64,
+
+    /// Optional websocket endpoint for Flashblocks connection.
+    #[arg(long = "flashblocks-url", value_name = "FLASHBLOCKS_WS_URL")]
+    pub flashblocks_url: Option<String>,
```
mattsse, line 72: this def should be websocket yes

crates/optimism/rpc/src/eth/pending_block.rs
```diff
@@ -3,16 +3,40 @@
...
 };
+use reth_storage_api::{ProviderBlock, ProviderReceipt, ReceiptProvider, StateProviderFactory};
+use reth_transaction_pool::{
+    error::InvalidPoolTransactionError, BestTransactionsAttributes, TransactionPool,
+};
+use revm::context::Block;
+use serde_json::Value;
+use tokio::time::Duration;
+
+pub(super) type SharedFlashblockState = Arc<tokio::sync::RwLock<FlashBlock>>;
```
emhane, line 39: prefer if we don't have this alias, so that it is easier to spot where locks are being used inline to make it easier to debug (locks are statistically bug prone sync primitive compared to others)

crates/optimism/rpc/src/eth/pending_block.rs
```diff
@@ -41,21 +65,192 @@ where
...
+                let existing_hashes: std::collections::HashSet<_> =
+                    s.diff.transactions.iter().cloned().collect();
+
+                s.diff.transactions.extend(new_fb.diff.transactions.into_iter().filter(|tx| {
+                    let recovered: Recovered<
+    < <N as RpcNodeCore>::Primitives as reth_primitives_traits::NodePrimitives >::SignedTx
+> = recover_raw_transaction(tx).unwrap();
+
+                    !existing_hashes.contains(recovered.tx_hash().as_slice())
+                }));
```
emhane, line 250:
```suggestion
            let is_new_block = new_fb.metadata.block_number > s.metadata.block_number;
            if is_new_block {
                *state.write().await = new_fb;
            } else {
                // atomic update of transactions
                let mut s = state.write().await;
                
                let existing_raw_txns HashSet<_> = s.diff.transactions.iter().collect();
                
                let new_txns = new_fb.diff.transactions.retain(|tx| !existing_raw_txns.contains(&tx));
                
                s.diff.transactions.extend(new_txns);
```
1. always hold a write lock for the absolute minimal time possible. a good way to avoid holding on to a write lock for unnecessary time, is to avoid binding it, unless explicitly necessary to make some instructions an atomic operation.
2. it's inefficient to decode the transactions just for finding new txns. simply compare the bytes.

emhane (changes requested):
thanks @Rimeeeeee ! first pass

crates/optimism/rpc/src/eth/pending_block.rs
emhane: @mattsse do the transactions in the alloy `Flashblock` type have to be in a particular order hence the `Vec<Bytes>`? otherwise wdyt ab making transactions generic on flashblocks? would be great for us to use a `HashSet<Bytes>` here directly, then we just have to do
```rust
for raw_tx in new_fb.diff.transactions {
    state.write().await.diff.transaction.insert(raw_tx)
}
```
and can skip cloning state's existing txns

mattsse (commented):
this is superseded by https://github.com/base/node-reth/tree/main that we'll upstream

sorry about this

PR #17863, opened 2025-08-13
feat: trie state reuse the available in-memory trie node

wetkeyboard:
The TODO comments indicated that the `StorageRootProvider` were not reusing available in-memory trie nodes.
The current implementation was try to extract storage data from the trie input.

And in the meanwhile, we cached the keccak address in each function, to reduce the keccak calculation.


mattsse (commented):
is this right @shekhirin @mediocregopher ?

mattsse (commented):
this is now outdated

PR #17868, opened 2025-08-14
chore: remove Beacon from type names

### LINKED ISSUE #17176: Rename `Beacon` prefixed types

mattsse:
supersedes #17299

closes https://github.com/paradigmxyz/reth/issues/17176

this removes the Beacon prefix from some types while keeping a deprecated alias, because beacon is kinda confusing


Rjected (approved):
lgtm

PR #17869, opened 2025-08-14
perf: avoid unnecessary iterator clone in CORS validation

JackDrogon:
perf: avoid unnecessary iterator clone in CORS validation

Replace iter.clone().any() with direct string contains check for wildcard detection. This eliminates an unnecessary iterator clone and improves performance in the CORS domain validation path.


mattsse (commented):
this isn't equivalent because we need to check the split items and this only clones the iter

JackDrogon:
> this isn't equivalent because we need to check the split items and this only clones the iter
@mattsse yep, u are right. original code just check domain like ",*" or ",*,", not like "aa*bb". my code is wrong.

PR #17873, opened 2025-08-14
refactor(era): add era types and file traits for shared behavior

lean-apple:
Still towards https://github.com/paradigmxyz/reth/issues/16895. (should be the last before the finale pr to close the issue).

Add `EraFileFormat`, `StreamReader`,  `FileReader`, and `StreamWriter`, `FileWrite` traits for shared behavior between era and era 1 files :
"only" the data content change - compressed encoded block data -, but the file behavior seems pretty similar, and they are `e2s` files.

Here an overview what it will look like for the era files support (there are no yet tests https://github.com/lean-apple/reth/pull/7)


crates/era/src/era_file_ops.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// file's ID using the standardized era file naming convention
+    fn create_with_id<P: AsRef<Path>>(directory: P, file: &Self::File) -> Result<(), E2sError>
+    where
+        <Self::File as EraFileFormat>::Id: EraFileId,
+    {
+        let filename = file.id().to_file_name();
+        let path = directory.as_ref().join(filename);
+        Self::create(path, file)
+    }
+}
```
RomanHodulak: Rather than using inheritance, we might define the interface without the ancestor and provide a blanket implementation for any `StreamWriter`. This way we can develop the interfaces independently and get the implementations for all `StreamWriter` automatically, avoiding the `impl FileWriter for X {}` statements.

```suggestion
/// [`StreamWriter`] provides writing file operations for era files
pub trait FileWriter {
    type File: EraFileFormat<Id: EraFileId>;

    /// Creates a new file at the specified path and writes the era file to it
    fn create<P: AsRef<Path>>(path: P, file: &Self::File) -> Result<(), E2sError>;

    /// Creates a new file in the specified directory with a filename derived from the
    /// file's ID using the standardized era file naming convention
    fn create_with_id<P: AsRef<Path>>(directory: P, file: &Self::File) -> Result<(), E2sError>;
}

impl<T: StreamWriter<File>> FileWriter for T {
    type File = T::File;

    /// Creates a new file at the specified path and writes the era file to it
    fn create<P: AsRef<Path>>(path: P, file: &Self::File) -> Result<(), E2sError> {
        let file_handle = File::create(path).map_err(E2sError::Io)?;
        let mut writer = Self::new(file_handle);
        writer.write_file(file)?;
        Ok(())
    }

    /// Creates a new file in the specified directory with a filename derived from the
    /// file's ID using the standardized era file naming convention
    fn create_with_id<P: AsRef<Path>>(directory: P, file: &Self::File) -> Result<(), E2sError> {
        let filename = file.id().to_file_name();
        let path = directory.as_ref().join(filename);
        Self::create(path, file)
    }
}
```

RomanHodulak (approved):
Nice progress.

crates/era/src/era_file_ops.rs
lean-apple: Thanks, much cleaner indeed like this.

mattsse (approved):
makes sense, we eventually need something like this for post merge era

PR #17874, opened 2025-08-14
chore(tx-pool): Simplify syntax

emhane:
Removes redundant mutable reference to an `Option` and call to `expect`, by casting state provider to trait object

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -132,7 +132,7 @@ where
         origin: TransactionOrigin,
         transaction: Tx,
     ) -> TransactionValidationOutcome<Tx> {
-        self.inner.validate_one_with_provider(origin, transaction, &mut None)
+        self.inner.validate_one_with_provider(origin, transaction, None)
     }
 
     /// Validates a single transaction with the provided state provider.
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -265,27 +265,24 @@ where
...
                         Err(err) => {
                             return TransactionValidationOutcome::Error(
                                 *transaction.hash(),
                                 Box::new(err),
                             )
                         }
-                    }
-                }
-
-                let state = maybe_state.as_deref().expect("provider is set");
```
emhane, line 288: @DaniPopes it doesn't change logic because it's always unwrapped here

DaniPopes: No, the `maybe_state` is assigned to once, and then just read from every time after. The logic here is the same as in `Option::get_or_insert_with`, just unrolled because you want to also return the error inside.

emhane: ah, I see, so that's what's going on here
https://github.com/paradigmxyz/reth/blob/6daf5fc777c113bd11fe72902f1344cc19bd6bd1/crates/transaction-pool/src/validate/eth.rs#L717-L721

PR #17877, opened 2025-08-14
feat: node import support importing gzip compressed file

tmelhao:
Geth supports importing gzip file https://github.com/ethereum/go-ethereum/blob/master/cmd/utils/cmd.go#L184-L188

Add file client to support read gzip file, so it can be used for `reth import` to load gzip compressed block files into reth db.


tmelhao:
@mattsse ping

mattsse (commented):
what's the usecase for this because this could also be uncompressed first before invoking import?

crates/net/downloaders/src/file_client.rs
```diff
@@ -406,6 +438,8 @@ pub struct ChunkedFileReader {
     /// Optionally, tracks highest decoded block number. Needed when decoding data that maps * to 1
     /// with block number
     highest_block: Option<u64>,
+    /// Whether the file is gzip compressed.
+    is_gzip: bool,
```
mattsse: this bool seems redundant if we can get that info from the filereader

can we add `fn is_gzip` to filereader instead?

crates/net/downloaders/src/file_client.rs
```diff
@@ -452,28 +523,79 @@ impl ChunkedFileReader {
...
         let chunk_target_len = self.chunk_len();
         let old_bytes_len = self.chunk.len() as u64;
 
+        // If we already have enough data in the chunk, return it
+        if old_bytes_len >= chunk_target_len {
+            return Ok(Some(old_bytes_len))
+        }
+
+        // For gzipped files, if we have no data in chunk, try to read some to check EOF
+        if self.is_gzip && self.chunk.is_empty() {
```
mattsse: not a fan of all these gzip conditionals

both file variants now support the same interface so we should be able to make this work by only relying in the read trait and not use different read variants

mattsse (changes requested):
thanks for the ping, now I understand

not opposed to this but this client impl is already not great and I'd like to avoid making even harder to understand, for example as you pointed out the read_exact is kinda flawed because the content of the buffer are technically unspecified by the trait.

I'd like to unify this to just relying on the read trait since both variants support it and no gzip conditionals.
which means making the gzip variant the only impl that should make it less hard to understand

crates/net/downloaders/src/file_client.rs
tmelhao: sure, will update today

jsvisa:
really need this feature :)

mattsse:
checking @jsvisa

mattsse (approved):
sorry, for the delay

this doesnt look too bad

this was a lot easier to review now, tysm!

PR #17878, opened 2025-08-14
fix(examples): Implement `Compact` using blanket implementation

RomanHodulak:
The previous manual encoding implementation was not correct, for example, it did not encode the signature


mattsse (approved):
nice

PR #17880, opened 2025-08-14
fix(net): Receipts69 should respond with Receipts69 message

wetkeyboard:
Response with Receipts 69 without bloom data


mattsse (approved):
makes sense

PR #17882, opened 2025-08-14
fix: typos in comments

prestoalvarez:
**Description:**

* Corrected typo in `connect_peer` docstring (`out` → `of`).
* Added clarifying comment before fee calculation in `builder.rs`.

---


mattsse (commented):
not worth doing

PR #17883, opened 2025-08-14
feat: add log.file.name cli arg

julio4:
Small non-breaking feature that add `--log.file.name` argument to cli, default to `reth.log`.
`FileInfo` now take the filename, and is instantiated once at startup.

This allows more customization and also allows rewriting default value when using `Cli` as a lib.


mattsse (approved):
reasonable, ty!

PR #17886, opened 2025-08-15
fix: use map_pure_precompiles for precompile caching

### LINKED ISSUE #17870: Respect pure precompiles for caching
### Describe the feature

Precompiles can be pure:

https://github.com/alloy-rs/evm/blob/aa1fa03c635d46e79d87e506adf0bea1285a77a1/crates/evm/src/precompiles.rs#L539-L539

if they're __not__ pure, we can't cache them, because this then could cause side-effects

we need to respect this when we create the cache and skip wrapping them if the precompile is not pure:

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/engine/tree/src/tree/payload_validator.rs#L672-L672

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L276-L276

ideally we also add this as another helper `map_pure_precompiles` to alloy-evm first

### Additional context

- some additional context, we added fn pure, because some projects started introducing precompiles that use data from state, e.g. look up some contract data, and this can’t be cached. 
- for this feature we can simple add a `map_pure_precompiles` to the `precompilesmap` type that simply skips non-pure precompiles.

assigned @yongkangc

yongkangc:
## Summary

This PR updates the precompile caching logic to use `map_pure_precompiles` instead of `map_precompiles` to ensure only pure precompiles (addresses 0x01-0x0a) are cached.

## Problem

The current implementation caches all precompiles, including stateful ones that shouldn't be cached. Stateful precompiles have side effects or depend on contract state, and caching them can lead to incorrect behavior.

## Solution

Replace `map_precompiles` with `map_pure_precompiles` in both:
- `crates/engine/tree/src/tree/payload_processor/prewarm.rs`
- `crates/engine/tree/src/tree/payload_validator.rs`

This ensures only pure precompiles (standard Ethereum precompiles at addresses 0x01-0x0a) are cached, as these are stateless and safe to cache.

## Dependencies

**This PR requires the alloy-evm PR to be merged first:** alloy-rs/evm#153

The PR is marked as draft until the dependency is resolved.

## Related Issues

- Fixes #17870

## Checklist

- [x] Code changes made
- [x] Added explanatory comments
- [ ] Waiting for alloy-evm PR https://github.com/alloy-rs/evm/pull/153 to be merged
- [ ] Update dependencies once alloy-evm is updated
- [ ] Verify compilation after dependency update

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,9 +259,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-evm"
-version = "0.18.1"
+version = "0.18.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "145913bf9bc11a7cec63a61aba36a39e41f2604aceb6d81b6bb8a4b2ddc93423"
+checksum = "55212170663df0af86b8b88ea08f13e3ee305e6717372e693d3408c0910e2981"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -2286,7 +2286,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm, pending evm patch bump

PR #17889, opened 2025-08-15
fix(network): off by one error in getting next header

colinlyguo:
This PR fixes an off-by-one error.


mattsse (approved):
oopsie, nice find, introduced in #17766

PR #17890, opened 2025-08-15
chore(pool): replace saturating_sub with unchecked_sub

malik672:
Guarantees from prior pool validation ensure transaction_nonce > on_chain_nonce,  
which implies transaction_nonce >= 1. This makes unchecked_sub(1) safe here,  
removing the redundant saturating_sub overhead.

### DIFF
--- a/crates/transaction-pool/src/identifier.rs
+++ b/crates/transaction-pool/src/identifier.rs
@@ -47,8 +47,8 @@ impl SenderIdentifiers {
 
     /// Returns the current identifier and increments the counter.
     fn next_id(&mut self) -> SenderId {
-        let id = self.id;
-        self.id += 1;
+        let id = self.id
+        self.id = self.id.wrapping_add(1);
         id.into()
     }
 }
--- a/crates/transaction-pool/src/identifier.rs
+++ b/crates/transaction-pool/src/identifier.rs
@@ -48,7 +48,7 @@ impl SenderIdentifiers {
     /// Returns the current identifier and increments the counter.
     fn next_id(&mut self) -> SenderId {
         let id = self.id;
-        self.id = self.id.wrapping_add(1);
+        self.id += 1;
         id.into()
     }
 }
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/identifier.rs
```diff
@@ -101,13 +101,15 @@ impl TransactionId {
     /// This returns `transaction_nonce - 1` if `transaction_nonce` is higher than the
     /// `on_chain_nonce`
     pub fn ancestor(transaction_nonce: u64, on_chain_nonce: u64, sender: SenderId) -> Option<Self> {
+        // SAFETY: transaction_nonce > on_chain_nonce ⇒ transaction_nonce >= 1
         (transaction_nonce > on_chain_nonce)
-            .then(|| Self::new(sender, transaction_nonce.saturating_sub(1)))
+            .then(|| Self::new(sender, unsafe { transaction_nonce.unchecked_sub(1) }))
```
mattsse: instead of unsafe, can we then just use `transaction_nonce - 1` here instead?

crates/transaction-pool/src/identifier.rs
```diff
@@ -101,13 +101,15 @@ impl TransactionId {
...
     }
 
     /// Returns the [`TransactionId`] that would come before this transaction.
     pub fn unchecked_ancestor(&self) -> Option<Self> {
-        (self.nonce != 0).then(|| Self::new(self.sender, self.nonce - 1))
+        // SAFETY: self.nonce != 0 ⇒ self.nonce >= 1
+        (self.nonce != 0).then(|| Self::new(self.sender, unsafe {self.nonce.unchecked_sub(1)}))
```
mattsse: hmm, this is effectively already the same as `- 1` so we can keep it as is but can add the SAFTY note for clarity

mattsse (changes requested):
makes sense,

nits

mattsse (approved):

PR #17892, opened 2025-08-15
chore: clippy happy

mattsse:


emhane (approved):
love the branch name, clippy does fuzz often it's true

PR #17894, opened 2025-08-15
refactor(examples): Replace `CustomTransactionEnvelope` with `Signed<TxPayment>` as the variant type in the custom node example

RomanHodulak:
The `Signed<TxPayment>` is wrapped in `CustomTransactionEnvelope`.

Since `TxPayment` is already a custom type, there is no need to introduce the wrapper. Doing so increases the amount of traits that need to be reimplemented, increasing bloat and decreasing readability.


mattsse (approved):
cleaner, this now follows the existing convention in ethereum and the op envelope type

PR #17904, opened 2025-08-16

crStiv:


mattsse (approved):
letting this one through because decent amount of fixes

PR #17909, opened 2025-08-18
fix: convert anchor() method from recursive to iterative

keroro520:
Previously, the anchor() method used unbounded recursion when walking up the parent chain to find the disk anchor block. This could cause stack overflow for very long in-memory chains.

The method now uses an iterative approach with a while loop, eliminating the recursion and preventing potential stack overflow issues.


mattsse (approved):
this is okay

Rjected (approved):

PR #17912, opened 2025-08-18
refactor(tree): move metered execution functions to tree module

### LINKED ISSUE #16247: Move metered execution functions
### Describe the feature

these should be moved:

https://github.com/paradigmxyz/reth/blob/8b4db1ffa31a36ac558c6cc65bdc7165b1cd9a77/crates/evm/evm/src/metrics.rs#L82-L150

as is to:

https://github.com/paradigmxyz/reth/blob/8b4db1ffa31a36ac558c6cc65bdc7165b1cd9a77/crates/engine/tree/src/tree/mod.rs#L2277-L2277

and replace `self.<metrics>` access with `self.metrics.executor`

## TODO
* move functions
* update metrics access

### Additional context

- I think what we can do here instead is add some helper struct that has all the metrics values as fields, and then we do metrics.update(BlockInfo) or smth

yongkangc:
## Summary

Moves the metered execution functions from `ExecutorMetrics` in the evm crate to `EngineApiMetrics` in the tree module. This consolidates execution logic directly in the tree handler for better code organization.

### Functions Moved
- `metered()` 
- `execute_metered()` 
- `metered_one()` 

### Implementation Details
1. **Moved `MeteredStateHook` struct** from `crates/evm/evm/src/metrics.rs` to `crates/engine/tree/src/tree/mod.rs`
   - This wrapper combines metrics tracking with state hook functionality

2. **Moved metered functions to `EngineApiMetrics`** in `crates/engine/tree/src/tree/metrics.rs`
   - Updated metrics access pattern from `self.<metric>` to `self.metrics.executor.<metric>`
   - Functions are now methods on `EngineApiMetrics` which contains `executor: ExecutorMetrics`

3. **Updated call sites**:
   - `payload_validator.rs`: Changed from `self.metrics.executor.execute_metered()` to `self.metrics.execute_metered()`
   - `stages/execution.rs`: Replaced `metered_one()` usage with inline metering code

4. **Cleaned up**:
   - Removed the moved functions and tests from `crates/evm/evm/src/metrics.rs`
   - Added necessary dependencies (`reth-execution-types`) to `engine-tree/Cargo.toml`
   - Fixed all import statements and compilation errors
   - Fixed CI issues: Removed unused `ExecutorMetrics` field from `ExecutionStage` that was causing clippy warnings

Closes #16247

Similar to #16935


yongkangc:
bump on this @mattsse to see if u have any comments :)

crates/stages/stages/src/stages/execution.rs
mattsse, line 1: can we undo the extra ; here :)

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -18,6 +28,293 @@ pub(crate) struct EngineApiMetrics {
...
+        self.executor.accounts_updated_histogram.record(accounts as f64);
+        self.executor.storage_slots_updated_histogram.record(storage_slots as f64);
+        self.executor.bytecodes_updated_histogram.record(bytecodes as f64);
+
+        Ok(output)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
mattsse, line 196: this is a bit out of order because there are more impls coming after this mod

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -686,7 +686,7 @@ where
 
         let execution_start = Instant::now();
         let state_hook = Box::new(handle.state_hook());
-        let output = self.metrics.executor.execute_metered(
+        let output = self.metrics.execute_metered(
```
mattsse, line 689: the alternative is moving the entire function in here which makes it less reusable in case this impl is customized

mattsse (changes requested):
some nits

we've move the impl from one metrics to the tree/executor specific metrics type

I'm okay with this, the alternative is to inline this in the validator impl entirely, which perhaps isnt ideal because would prevent reuse
wdyt @shekhirin @Rjected

crates/stages/stages/src/stages/execution.rs
yongkangc: Thanks for the catch here 🙏

yongkangc:
> some nits
> 
> we've move the impl from one metrics to the tree/executor specific metrics type
> 
> I'm okay with this, the alternative is to inline this in the validator impl entirely, which perhaps isnt ideal because would prevent reuse
> wdyt @shekhirin @Rjected 


 @mattsse Thanks for the review! 

I've addressed both points:
  - Removed the extra semicolon in
  `execution.rs`
  - Moved the test module to the end of
  `metrics.rs` after all implementations

crates/stages/stages/src/stages/execution.rs
yongkangc: thanks matt, will do!

Rjected (commented):
I am ok with this as an incremental change, but think ideally we should have something that:
 * allows us to reuse the execution metrics recording across engine and execution,
 * which is not a method directly on a metrics struct
 
can do this as a followup though

yongkangc:
Thanks @Rjected, created an issue here: #17997

mattsse (approved):
I'm okay with this

but defer to @Rjected and @shekhirin

Rjected (approved):
LGTM, thanks!

PR #17916, opened 2025-08-18
perf: reduce cycles on indexing

malik672:
reduce 100 cycles on indexing
https://godbolt.org/z/b5cE9bM5G

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -3054,11 +3054,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
             debug!(target: "providers::db", "Attempted to append empty block range");
             return Ok(())
         }
+ 
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being `None`.
+        //blocks[0] get's the first elment of the `blocks` vector.
+        let first_number = blocks[0].number();
 
-        let first_number = blocks.first().unwrap().number();
-
-        let last = blocks.last().unwrap();
-        let last_block_number = last.number();
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being `None`.
+        let last_block_number = blocks[blocks.len() - 1].number();
 
         let mut durations_recorder = metrics::DurationsRecorder::default();
 
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -3055,11 +3055,11 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
             return Ok(())
         }
 
-        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        //NOTE: Blocks are not empty, so no need to handle the case of `blocks.first()` being
         // `None`. blocks[0] get's the first element of the `blocks` vector.
         let first_number = blocks[0].number();
 
-        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        //NOTE: Blocks are not empty, so no need to handle the case of `blocks.first()` being
         // `None`.
         let last_block_number = blocks[blocks.len() - 1].number();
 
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3055,10 +3055,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
...
-        let first_number = blocks.first().unwrap().number();
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`. blocks[0] get's the first element of the `blocks` vector.
+        let first_number = blocks[0].number();
 
-        let last = blocks.last().unwrap();
-        let last_block_number = last.number();
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`.
+        let last_block_number = blocks[blocks.len() - 1].number();
```
shekhirin: It's not unsafe code, so no need for `SAFETY` comment

```suggestion
        // Blocks are not empty, so no need to handle the case of `blocks.first()` being `None`.
        let first_number = blocks[0].number();

        // Blocks are not empty, so no need to handle the case of `blocks.last()` being `None`.
        let last_block_number = blocks[blocks.len() - 1].number();
```

shekhirin (dismissed):
Overall supportive, just a comment nit

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3055,10 +3055,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
             return Ok(())
         }
 
-        let first_number = blocks.first().unwrap().number();
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`. blocks[0] get's the first element of the `blocks` vector.
+        let first_number = blocks[0].number();
```
Rjected:
```suggestion
        // NOTE: We checked that blocks are not empty, so there should always be a first block
        let first_number = blocks[0].number();
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3055,10 +3055,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
...
 
-        let last = blocks.last().unwrap();
-        let last_block_number = last.number();
+        //SAFETY: Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`.
```
Rjected:
```suggestion
        // NOTE: We checked that blocks are not empty, so there should always be a last block that can be indexed by `blocks.len() - 1`
```

Rjected (changes requested):
some very minor nits otherwise this lgtm, we reserve `SAFETY` for unsafe operations, and don't use `SAFETY` for operations that might panic

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3055,10 +3055,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
...
 
-        let first_number = blocks.first().unwrap().number();
+        //Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`. blocks[0] get's the first element of the `blocks` vector.
+        let first_number = blocks[0].number();
 
-        let last = blocks.last().unwrap();
-        let last_block_number = last.number();
+        //Blocks are not empty, so no need to handle the case of `blocks.first()` being
+        // `None`.
```
shekhirin:
```suggestion
        // Blocks are not empty, so no need to handle the case of `blocks.first()` being
        // `None`.
        let first_number = blocks[0].number();

        // Blocks are not empty, so no need to handle the case of `blocks.last()` being
        // `None`.
```

shekhirin: @malik672 please add a space between `//` and `Blocks` for both comments, and change `blocks.first()` to `blocks.last()` in the second comment.

Rjected (approved):
lgtm

yongkangc (approved):

PR #17919, opened 2025-08-18
chore: add cache size limit to prevent unlimited growth

wetkeyboard:
set default cache 100 items, in case of OOM error


mattsse (commented):
I dont think we need this

PR #17923, opened 2025-08-18
fix(optimism): correct string formatting in error message

bendanzhentan:
The error message contains `{url}` but uses closure syntax instead of proper string formatting. This will literally print `{url}` instead of the actual URL value.

This fix uses `format!` macro to properly interpolate url.


mattsse (approved):
ah, yes this should indeed be a format

PR #17924, opened 2025-08-18
feat(rpc): Use pool-based pending block for pending state over latest

### LINKED ISSUE #17876: Actually respect pending state for pending tag
### Describe the feature

currently rout all blockid requests to the provider

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/rpc/rpc-eth-api/src/helpers/state.rs#L200-L206

the chance that we currently have an actually pending state (new CL payload) is very low.

we however do already build the pending block locally:

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L114-L116

which already comes with the entire state:

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L157-L170

that we however currently don't store yet:

https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L172-L180

## TODO
* Add support for obtaining the pending state, which is latest overlayed with latest like: https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/storage/provider/src/providers/blockchain_provider.rs#L139-L147
* For this we likely need another trait function that is `async fn is_pending` and then likely need to make the existing state_by async, where we then need a similar check for pending like https://github.com/paradigmxyz/reth/blob/b64eed99b5ce92e785665a37c89dcc8805486160/crates/rpc/rpc-eth-api/src/helpers/state.rs#L241-L244

### Additional context

this will unblock #17858

ref #17677

RomanHodulak:
Closes #17876 

For `LoadState` trait based RPC features, use pool-based pending block before using latest block if no pending block from consensus layer is available.

A feature useful for MEV / flashblocks use-cases.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/state.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/state.rs
@@ -11,12 +11,14 @@ use alloy_serde::JsonStorageKey;
 use futures::Future;
 use reth_errors::RethError;
 use reth_evm::{ConfigureEvm, EvmEnvFor};
-use reth_rpc_eth_types::{EthApiError, PendingBlockEnv, RpcInvalidTransactionError};
+use reth_rpc_convert::RpcConvert;
+use reth_rpc_eth_types::{
+    error::FromEvmError, EthApiError, PendingBlockEnv, RpcInvalidTransactionError,
+};
 use reth_storage_api::{
     BlockIdReader, BlockNumReader, StateProvider, StateProviderBox, StateProviderFactory,
 };
 use reth_transaction_pool::TransactionPool;
-use std::future;
 
 /// Helper methods for `eth_` methods relating to state (accounts).
 pub trait EthState: LoadState + SpawnBlocking {
@@ -195,7 +197,13 @@ pub trait EthState: LoadState + SpawnBlocking {
 /// Loads state from database.
 ///
 /// Behaviour shared by several `eth_` RPC methods, not exclusive to `eth_` state RPC methods.
-pub trait LoadState: EthApiTypes + RpcNodeCoreExt {
+pub trait LoadState:
+    LoadPendingBlock
+    + EthApiTypes<
+        Error: FromEvmError<Self::Evm> + FromEthApiError,
+        RpcConvert: RpcConvert<Network = Self::NetworkTypes>,
+    > + RpcNodeCoreExt
+{
     /// Returns the state at the given block number
     fn state_at_hash(&self, block_hash: B256) -> Result<StateProviderBox, Self::Error> {
         self.provider().history_by_block_hash(block_hash).map_err(Self::Error::from_eth_err)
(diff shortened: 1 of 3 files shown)


crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -396,3 +382,52 @@ impl<H: BlockHeader> BuildPendingEnv<H> for NextBlockEnvAttributes {
...
+/// The outcomes of building a pending block.
+#[derive(Debug)]
+pub enum PendingBlockVarious<N: NodePrimitives> {
+    /// A pending block received by the engine API from the consensus layer.
+    Consensus(Arc<RecoveredBlock<N::Block>>, Arc<Vec<N::Receipt>>),
+    /// A pending block built using the local transaction pool.
+    Pool(PendingBlock<N>),
+    /// A pending block failed to build.
+    Empty,
+}
```
mattsse: do we even need this enum?

unsure if we even need to change the local_pending_block fn

crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -48,13 +50,16 @@ pub trait EthState: LoadState + SpawnBlocking {
...
-        self.spawn_blocking_io(move |this| {
-            Ok(this
-                .state_at_block_id_or_latest(block_id)?
-                .account_balance(&address)
-                .map_err(Self::Error::from_eth_err)?
-                .unwrap_or_default())
-        })
+        async move {
+            let state = self.state_at_block_id_or_latest(block_id).await?;
+            self.spawn_blocking_io(move |_this| {
```
mattsse: this is a bit tricky, because for the pending state we now need to mix async and IO.

however, this is already what

https://github.com/paradigmxyz/reth/blob/d53ecc5ab0b2931e0fe6cdc906eff5c361767c50/crates/rpc/rpc-eth-api/src/helpers/blocking_task.rs#L44-L61

does so we could add `spawn_blocking_fut` or something that accepts a future instead, in which we could await the `state_at_block_id_or_latest`

which would be equivalent to what we already do now

mattsse (changes requested):
imo this goes in the right direction,

because this is a bit tricky, I'd like to start smaller and first focussing on making the state_at fn async

for the state I think we should introduce a placeholder pending_state fn first then we don't need to touch the local_pending block fn for now

crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -208,8 +216,19 @@ pub trait LoadState: EthApiTypes + RpcNodeCoreExt {
...
+        Self: SpawnBlocking,
+    {
+        async move {
+            if at.is_pending() {
+                if let Ok(Some(state)) = self.local_pending_state().await {
+                    return Ok(state)
+                }
+            }
+
+            self.provider().state_by_block_id(at).map_err(Self::Error::from_eth_err)
```
mattsse, line 230: this will no ignore the _real_ pending state right?

hmm, this is a bit tricky because we don't have a nice way to obtain just the optional pending state from the provider, only if the real pending block exists, but imo we could ignore this case for now and then add this functionality separately as an additional optional fn to:

https://github.com/paradigmxyz/reth/blob/6264530a8acc254570ff6ca936f6931a9a77f6c0/crates/storage/storage-api/src/state.rs#L185-L189

as `fn maybe_pending`?

mattsse (approved):

crates/rpc/rpc-eth-api/src/helpers/state.rs
RomanHodulak: You're right, this won't ignore the _real_ pending state.

However, the `local_pending_state` only returns the pool-based pending state, not the _real_ pending state.

PR #17928, opened 2025-08-19
feat: support importing multi files

tmelhao:
Import a list of files, eg:

```
reth import --datadir /path/to/data blocks-0.rlp blocks-1.rlp blocks-2.rlp
```


mattsse (commented):
could you elaborate on the specific use case for this

is this currently being somewhere?

mattsse (approved):
lgtm, checks out

PR #17931, opened 2025-08-19
fix(trie): replace rayon with tokio for I/O operations in parallel trie

### LINKED ISSUE #17288: Spawn `StorageRoot` in tokio blocking task, not rayon
### Describe the feature

-

### Additional context

_No response_

yongkangc:
Fixes #17288

## Summary

Replaces `rayon::spawn_fifo` with `tokio::task::spawn_blocking` for database I/O operations in parallel trie computation. Rayon's CPU-bound thread pool was incorrectly used for I/O-heavy operations, causing thread pool starvation.

## Changes

- Replace `rayon::spawn_fifo` with `tokio::task::spawn_blocking` in `calculate()`
- Add `get_runtime_handle()` helper that creates runtime if not in Tokio context (e.g., Tree Task thread)
- Use fire-and-forget pattern with channels for result communication

### DIFF
--- a/crates/trie/parallel/src/root.rs
+++ b/crates/trie/parallel/src/root.rs
@@ -82,7 +82,7 @@ where
     }
 
     /// Computes the state root by calculating storage roots in parallel for modified accounts,
-    /// then walking the state trie to build the final root hash.
+    /// then walking the state trie to build the final state root hash.
     fn calculate(
         self,
         retain_updates: bool,
--- a/crates/trie/parallel/src/root.rs
+++ b/crates/trie/parallel/src/root.rs
@@ -72,6 +72,8 @@ where
         self.calculate(true)
     }
 
+    /// Computes the state root by calculating storage roots in parallel for modified accounts,
+    /// then walking the state trie to build the final root hash.
     fn calculate(
         self,
         retain_updates: bool,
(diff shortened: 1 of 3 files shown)


mediocregopher (approved):
LGTM

crates/trie/parallel/src/root.rs
```diff
@@ -98,9 +113,10 @@ where
             #[cfg(feature = "metrics")]
             let metrics = self.metrics.storage_trie.clone();
 
-            let (tx, rx) = std::sync::mpsc::sync_channel(1);
+            let (tx, rx) = mpsc::sync_channel(1);
 
-            rayon::spawn_fifo(move || {
+            // Spawn a blocking task to calculate account's storage root from database I/O
+            drop(handle.spawn_blocking(move || {
```
mattsse, line 119: this drop is fine because joinhandle can be dropped although not necessary

crates/trie/parallel/src/root.rs
```diff
@@ -261,6 +277,27 @@ impl From<alloy_rlp::Error> for ParallelStateRootError {
     }
 }
 
+/// Gets or creates a tokio runtime handle for spawning blocking tasks.
+/// This ensures we always have a runtime available for I/O operations.
+fn get_runtime_handle() -> Handle {
```
mattsse, line 282: hmm, this will most likely create a new runtime with a bunch of threads because the tree task is run outside of tokio

so we'll endup with a bunch of unusable threads. which should be fine since we dont expect this to be reachable in the happy path.

and this is better than sprinkling the runtime as arg all over the place

mattsse (approved):

crates/trie/parallel/src/root.rs
yongkangc: So we clean this up in another issue maybe? Or this is fine for now

Had the same thought on what's the cleanest way to go about this! @mattsse

mattsse: imo this is fine

PR #17932, opened 2025-08-19
fix: optimize empty directory check in is_database_empty

LandLord529:
Replace `dir.count() == 0` with `dir.next().is_none()` to avoid full directory traversal.

### DIFF
--- a/crates/storage/db/src/utils.rs
+++ b/crates/storage/db/src/utils.rs
@@ -25,8 +25,9 @@ pub fn is_database_empty<P: AsRef<Path>>(path: P) -> bool {
         true
     } else if path.is_file() {
         false
-    } else if let Ok(dir) = path.read_dir() {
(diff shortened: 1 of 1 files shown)


crates/storage/db/src/utils.rs
```diff
@@ -25,8 +25,9 @@ pub fn is_database_empty<P: AsRef<Path>>(path: P) -> bool {
         true
     } else if path.is_file() {
         false
-    } else if let Ok(dir) = path.read_dir() {
-        dir.count() == 0
+    } else if let Ok(mut dir) = path.read_dir() {
+        // Check if directory has any entries without counting all of them
+        dir.next().is_none()
```
mattsse, line 30: this is indeed slightly more efficient, so I led this one slide

mattsse (approved):

PR #17938, opened 2025-08-19
feat: FCU unwind: properly reorg in-memory canonical state and update latest block

pycckuu:
Fix FCU when new head is an ancestor (unwind). Previously we only updated the header, leaving in-memory canonical state stale, causing tx validation issues (e.g., "nonce too low"). Now we detect unwind, reorg in-memory state (remove old blocks, load ancestor), and keep TreeState and CanonicalInMemoryState in sync.

## Code Changes
-  tree/mod.rs:
  - Add: `log_chain_update_type`, `handle_chain_unwind`, `collect_blocks_for_removal`, `apply_canonical_ancestor_via_reorg`, `handle_chain_advance_or_same_height`, `ensure_block_in_memory`.
  - Replace header-only update with:
    ```rust
    self.log_chain_update_type(...);
    self.state.tree_state.set_canonical_head(canonical_header.num_hash());
    if new_head_number < current_head_number {
        self.handle_chain_unwind(current_head_number, canonical_header)
    } else {
        self.handle_chain_advance_or_same_height(canonical_header)
    }
    ```
  - Reorg on unwind:
    ```rust
    self.canonical_in_memory_state.update_chain(NewCanonicalChain::Reorg { new, old });
    self.canonical_in_memory_state.set_canonical_head(canonical_header.clone());
    ```
  - Advance/same-height: ensure block committed into memory if missing.
  - Interim method `update_latest_block_to_canonical_ancestor` added, then superseded by full unwind handling.
-  tree/tests.rs:
  - New test `test_fcu_with_canonical_ancestor_updates_latest_block` asserting both TreeState and in-memory head match the ancestor after FCU.

## Reason for Changes
Correctness: prevent stale in-memory state on reorgs that led to txpool/state provider errors.

## Impact of Changes
-  Correct state after unwinds; eliminate nonce/state desync errors.
-  Low overhead; clearer logging.

## Test Plan
-  Run: `cargo test -p engine_tree test_fcu_with_canonical_ancestor_updates_latest_block`.

## How the reviewer should test the fix
-  Unwind: head=4 → FCU to 2; verify removal of [3,4] and head=2 in both states.
-  Advance: head=2 → FCU to 3; verify block 3 committed and head updated.
-  Check logs for correct classification.

## Other useful info
If ancestor block missing from storage, we warn and still update header to avoid desync.

Closes https://github.com/paradigmxyz/reth/issues/17798

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -701,6 +701,95 @@ where
         Ok(Some(NewCanonicalChain::Reorg { new: new_chain, old: old_chain }))
     }
 
+    /// Updates the latest block state to the specified canonical ancestor.
+    ///
+    /// This method ensures that state providers and the transaction pool operate with
+    /// the correct chain state after forkchoice update processing. It detects unwind
+    /// scenarios and provides appropriate logging for debugging purposes.
+    ///
+    /// # Arguments
+    /// * `canonical_header` - The canonical header to set as the new head
+    ///
+    /// # Returns
+    /// * `ProviderResult<()>` - Ok(()) on success, error if state update fails
+    fn update_latest_block_to_canonical_ancestor(
+        &mut self,
+        canonical_header: &SealedHeader<N::BlockHeader>,
+    ) -> ProviderResult<()> {
+        let current_head_number = self.state.tree_state.canonical_block_number();
+        let new_head_number = canonical_header.number();
+        let new_head_hash = canonical_header.hash();
+
+        // Detect and log the type of chain state update
+        match new_head_number.cmp(&current_head_number) {
+            std::cmp::Ordering::Less => {
+                // Unwind scenario: we're reverting to an earlier canonical block
+                debug!(
+                    target: "engine::tree",
+                    current_head = current_head_number,
+                    new_head = new_head_number,
+                    new_head_hash = ?new_head_hash,
+                    "FCU unwind detected: reverting to canonical ancestor"
+                );
+            }
+            std::cmp::Ordering::Greater => {
+                // Forward progress: updating to a later canonical block
+                debug!(
+                    target: "engine::tree",
+                    previous_head = current_head_number,
+                    new_head = new_head_number,
+                    new_head_hash = ?new_head_hash,
+                    "Advancing latest block to canonical ancestor"
+                );
+            }
+            std::cmp::Ordering::Equal => {
+                // Same block number, potentially different hash (shouldn't normally happen)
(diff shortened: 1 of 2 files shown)


pycckuu:
@mattsse I rebased on the recent changes from the main. Can you please take a look?

pycckuu:
@mediocregopher I've removed the styling commit. Could you please review it? Thanks in advance.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -701,6 +701,227 @@ where
...
+                found_blocks_to_remove = true;
+                debug!(
+                    target: "engine::tree",
+                    block_number = block_num,
+                    "Collected block for removal from in-memory state"
+                );
+            }
+        }
+
+        if !found_blocks_to_remove {
```
mediocregopher: nit: this could just be `old_blocks.is_empty()`

crates/engine/tree/src/tree/mod.rs
```diff
@@ -856,6 +1077,12 @@ where
                         self.process_payload_attributes(attr, &canonical_header, state, version);
                     return Ok(TreeOutcome::new(updated))
                 }
+
+                // Update the latest block state to reflect the canonical ancestor.
+                // This ensures that state providers and the transaction pool operate with
+                // the correct chain state after forkchoice update processing.
+                self.update_latest_block_to_canonical_ancestor(&canonical_header)?;
```
mediocregopher, line 1079: Is it correct that this is being called after the `process_payload_attributes` if-branch? If `attrs.is_some()` then this line never gets hit.

mediocregopher (changes requested):
One nit and one question, will want @mattsse to look at it as well

crates/engine/tree/src/tree/mod.rs
pycckuu: I suppose, the assumption here is:
- **Without attrs**: We're just updating the head, so the state needs to be fixed
- **With attrs**: We're building the payload and assuming the state is already correct

The placement of the new code was suggested by @mattsse in 9e8735ddced28f8b1372acd5755bb7107ef3ba20

crates/engine/tree/src/tree/mod.rs
pycckuu: Addressed!

mediocregopher (approved):
lgtm, pending matt's review

mattsse (approved):
I'm okay with this because this code path is only reachable if explicitly opt-in.

PR #17939, opened 2025-08-19
  chore: refactors block storage APIs throughout to accept references

malik672:
made some APIs to accept refrences instead of owned values
I mean if it's possible to use a refrence why not ?


mattsse (commented):
dont think this matters or is something we should change

half of the changes dont even get rid of the clone

PR #17940, opened 2025-08-19
feat(rpc): Add `EvmEnvConverter` for custom effects on the `EvmEnv`

### LINKED ISSUE #17935: Make TxEnv call env conversion configurable
### Describe the feature

we currently rely on the default prepare_call_env impl:

https://github.com/paradigmxyz/reth/blob/91730cd326f6b0f975e56d8985f08a2a6c943f89/crates/rpc/rpc-eth-api/src/helpers/call.rs#L720-L730

which does things like balance checks, apply overrides etc.

we already have a way to inject this logic for converting requests to txenv

https://github.com/paradigmxyz/reth/blob/91730cd326f6b0f975e56d8985f08a2a6c943f89/crates/rpc/rpc-convert/src/transaction.rs#L138-L145

we need the same for call env as well

## TODO
* move call_env to RpcConvert trait
* ideally we can encapsulate the existing logic with a helper type so we can easily reuse for op and eth

### Additional context

_No response_

RomanHodulak:
Closes #17935

Adds a new converter to `RpcConvert` that can apply custom effects on the `EvmEnv`.

The conversion is `EvmEnv` -> `EvmEnv` meaning there is no different type at the output.

### DIFF
--- a/crates/rpc/rpc-convert/src/evm.rs
+++ b/crates/rpc/rpc-convert/src/evm.rs
@@ -1,0 +1,34 @@
+use reth_evm::EvmEnv;
+use std::{convert::Infallible, fmt::Debug};
+
+/// Performs an [`EvmEnv`] to [`EvmEnv`] conversion accompanied by the `TxReq`.
+///
+/// The `TxReq` represents the transaction request this [`EvmEnv`] has been created for.
+///
+/// Allows for custom effects on the [`EvmEnv`]. Default implementation `()` passes the input
+/// argument with no change.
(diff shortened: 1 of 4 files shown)


klkvr (commented):
This is simply an override so I think `EvmEnvOverrides`/`RpcEnvOverrides` would be a better name here?

Also it seems that https://github.com/paradigmxyz/reth/issues/17935 suggests extracting entire `prepare_call_env` logic into the overridable handler so would like a review from @mattsse here

mattsse (commented):
after looking at https://github.com/paradigmxyz/reth/pull/17792 again this is basically a subset of this and we solved https://github.com/paradigmxyz/reth/issues/17935 differently

PR #17941, opened 2025-08-19
fix(network): correct receipts size calculation in eth_requests

keroro520:
Replace direct length() call with iterator sum to properly calculate total bytes when processing transformed receipts.


mattsse (commented):
since we need to encode the receipts as list, we should also calc the vec length

which is more or less the same just with ~2 extra bytes here

PR #17942, opened 2025-08-19
docs: Fix comment for difficulty() method in bodies response

bendanzhentan:


mattsse (commented):
not something we want to accept in a single typo pr

PR #17944, opened 2025-08-19
feat: Add transaction propagation kind 'None'

SozinM:
We need to have a way to ensure that node won't leak any txs, while preserving option to receive transaction trough p2p.
Right now we are using `--rollup.disable-tx-pool-gossip` but this disables all gossiping, including incoming txs.


mattsse (approved):
reasonable

PR #17947, opened 2025-08-19
refactor(evm): use execution payload getters

lean-apple:
Follow-up for https://github.com/alloy-rs/alloy/issues/2778 (even if we dont use yet the new helpers).

Use https://github.com/alloy-rs/alloy/blob/c197dca34ea1c2d866336cc4cb2c4807f20b62de/crates/rpc-types-engine/src/payload.rs#L1169-L1169 getters.

### DIFF
--- a/crates/ethereum/evm/src/lib.rs
+++ b/crates/ethereum/evm/src/lib.rs
@@ -313,16 +313,14 @@ where
         // derive the EIP-4844 blob fees from the header's `excess_blob_gas` and the current
         // blobparams
         let blob_excess_gas_and_price =
-            payload.payload.as_v3().map(|v3| v3.excess_blob_gas).zip(blob_params).map(
-                |(excess_blob_gas, params)| {
-                    let blob_gasprice = params.calc_blob_fee(excess_blob_gas);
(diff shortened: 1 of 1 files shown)


crates/ethereum/evm/src/lib.rs
```diff
@@ -313,25 +313,23 @@ where
...
-            payload.payload.as_v3().map(|v3| v3.excess_blob_gas).zip(blob_params).map(
-                |(excess_blob_gas, params)| {
-                    let blob_gasprice = params.calc_blob_fee(excess_blob_gas);
-                    BlobExcessGasAndPrice { excess_blob_gas, blob_gasprice }
-                },
-            );
+            payload.payload.excess_blob_gas().zip(blob_params).map(|(excess_blob_gas, params)| {
+                let blob_gasprice = params.calc_blob_fee(excess_blob_gas);
+                BlobExcessGasAndPrice { excess_blob_gas, blob_gasprice }
+            });
```
lean-apple, line 319: I hesitated during the alloy's pr, maybe we can add another helper for this

mattsse (approved):

PR #17953, opened 2025-08-20
feat(rpc): Add `spawn_blocking_io_fut` that accepts a future

RomanHodulak:
Following-up on #17924

Adds a function for spawning blocking IO tasks that take a future, allowing for async functions being executed in this task.

This is crucial for allowing async pending block build.


mattsse (approved):
easy

PR #17954, opened 2025-08-20
feat(rpc): Convert `state_at_block_id` into async function

RomanHodulak:
Follow-up on #17924

This makes the existing `state_at_block_id` async, despite them not having any async call at the moment. This is a preparation for introducing async `pending_state` function and calling it inside this function.

There are two places that previously used `spawn_tracing` that were replaced with `spawn_blocking_io_fut`.

1. https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-eth-api/src/helpers/call.rs#L541
2. https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-eth-api/src/helpers/trace.rs#L305

I haven't found  a simple way to make `spawn_tracing` accept futures, nor am I sure that such thing makes sense. However, the `call` place is not really benefiting from using the rayon pool, which is what `spawn_tracing` is used for. And the second place, the `trace` API, might benefit from it, but `trace` API is not affecting production path, hence it is not as speed critical.

If we really wanted to keep `spawn_tracing`, one idea is to also put the async call to `state_at_block_id` outside of the callback. This, however, causes a mysterious `higher-ranked lifetime error` - I tried a few simple fixes unsuccessfully like:
* Introducing `use<>`
* Making a use of `StateProviderTraitObjWrapper`

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -361,8 +361,8 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
             let block_id = block_number.unwrap_or_default();
             let (evm_env, at) = self.evm_env_at(block_id).await?;
 
-            self.spawn_blocking_io(move |this| {
-                this.create_access_list_with(evm_env, at, request, state_override)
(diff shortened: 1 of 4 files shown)


mattsse (approved):
makes sense

most existing trace calls that were previously spawned on rayon pool do a lot of IO and should really be spawned on blocking

PR #17955, opened 2025-08-20
fix: ParallelSparseTrie::update_leaf edge-case, and not correctly clearing all fields for re-use

mediocregopher:
* In `update_leaf` when an existing leaf would get moved from the upper to a lower subtrie we were correctly forcing that lower subtrie to be included in root calculations. We were NOT doing this for extension nodes which would get moved in the same way.

  The fix is to not handle this case in update_leaf, but instead to always take lower subtries whose root node is missing a hash for hash calculation. This is better than the previous PrefixSet-based solution as it results in only re-calculating the root, not the entire subtrie.

* We were not properly clearing the ParallelSparseTrie on `clear()`, specifically the branch masks were not cleared. Also we were not setting updates on `wipe()` to indicate that they were wiped.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -344,7 +344,7 @@ impl SparseTrieInterface for ParallelSparseTrie {
                                     ?decoded,
                                     ?tree_mask,
                                     ?hash_mask,
-                                    "Revealing child",
+                                    "Revealing child (from upper)",
                                 );
                                 subtrie.reveal_node(
                                     reveal_path,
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense to me, but need another sanity review from

@Rjected or @shekhirin

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1288,7 +1288,10 @@ impl ParallelSparseTrie {
 
     /// Returns:
     /// 1. List of lower [subtries](SparseSubtrie) that have changed according to the provided
-    ///    [prefix set](PrefixSet). See documentation of [`ChangedSubtrie`] for more details.
+    ///    [prefix set](PrefixSet). See documentation of [`ChangedSubtrie`] for more details. Lower
+    ///    subtries whose root node is missing a hash will also be returned; this is required to
+    ///    handle cases where extensions/leafs get shortened and therefore moved from the upper to a
+    ///    lower subtrie.
```
Rjected, line 1294: I see, this makes sense, so the only thing that is changing, when we do one of these moves from upper -> lower with only a single node, is the root. I'm thinking about the case where there is a missing hash, but the prefix set does not contain `subtrie.path`. This node should be the only thing updated when we call `update_subtrie_hashes` after calling this fn

Rjected (approved):
lgtm, only comment is trying to summarize reasoning for my own understanding

jenpaff:
> lgtm, only comment is trying to summarize reasoning for my own understanding

Can you elaborate- where do you need this summarised/documented?

Rjected:
> > lgtm, only comment is trying to summarize reasoning for my own understanding
> 
> Can you elaborate- where do you need this summarised/documented?

I don't need any documentation here, I was just saying that my review comment is me summarizing my own understanding of the problem

PR #17956, opened 2025-08-20
perf(trie): default ParallelSparseTrie to enabled (accounts only still)

### LINKED ISSUE #17921: Enable Parallel Sparse Trie by default
### Describe the feature

Make the value of the flag `true` by default and add a flag for disabling the Parallel Sparse Trie.

### Additional context

_No response_

mediocregopher:
Fixes #17921


crates/node/core/src/args/engine.rs
```diff
@@ -35,7 +35,7 @@ pub struct EngineArgs {
     pub caching_and_prewarming_disabled: bool,
 
     /// Enable the parallel sparse trie in the engine.
-    #[arg(long = "engine.parallel-sparse-trie", default_value = "false")]
+    #[arg(long = "engine.parallel-sparse-trie", default_value = "true")]
```
shekhirin: let's do in a way similar to https://github.com/paradigmxyz/reth/pull/15764, so that default in the code is that it's enabled, and if `parallel_sparse_trie_disabled == true`, then we use serial one.

Rjected (approved):
LGTM

PR #17957, opened 2025-08-20
feat(rpc): Add `local_pending_state` that creates a state provider out of a mem-pool built pending block

RomanHodulak:
Following-up on #17924

Adds a `local_pending_state` that returns a state provider that only uses a pool-built pending block, not the consensus one.


mattsse (approved):
nice, we could consider renaming pool_pending_block

but I think this is appropriate

PR #17960, opened 2025-08-20
fix: allow at most one in-flight tx

0xNarumi:
In reth, the condition is >0, while in geth, the check is >1.
reference:
https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/core/txpool/legacypool/legacypool.go#L661-L663


mattsse (approved):
ty!

PR #17963, opened 2025-08-20
feat(engine): set default_memory_block_buffer_target to zero

### LINKED ISSUE #17945: Make default `MEMORY_BLOCK_BUFFER_TARGET` zero
### Describe the feature

Right now the `MEMORY_BLOCK_BUFFER_TARGET` is 2:
https://github.com/paradigmxyz/reth/blob/a3298ecfdd2f604e6fc288bf803aa2980637e82d/crates/engine/primitives/src/config.rs#L6-L7

This means that we always keep 2 blocks in memory, and do not persist them regardless of `--engine.persistence-threshold`. This makes the flag usage less clear and we do spend time creating e.g. overlay state on each block for however many blocks are in memory. IMO we should set this default to zero, this means that the persistence threshold flag is the only thing that determines when we persist, unless the user overrides with `--engine.memory-block-buffer-target`.

This could then be updated to just use `--engine.persistence-threshold`:
https://github.com/paradigmxyz/reth/blob/a3298ecfdd2f604e6fc288bf803aa2980637e82d/docs/vocs/docs/pages/sdk/examples/standalone-components.mdx#real-time-block-access-configuration

### Additional context

_No response_

iTranscend:
This PR:
- sets the `DEFAULT_MEMORY_BLOCK_BUFFER_TARGET` to zero
- Updates the docs to reflect this change

resolves https://github.com/paradigmxyz/reth/issues/17945

### DIFF
--- a/crates/engine/primitives/src/config.rs
+++ b/crates/engine/primitives/src/config.rs
@@ -4,7 +4,7 @@
 pub const DEFAULT_PERSISTENCE_THRESHOLD: u64 = 2;
 
 /// How close to the canonical head we persist blocks.
-pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 2;
+pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
 
 /// Default maximum concurrency for proof tasks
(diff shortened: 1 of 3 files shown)


Rjected (approved):
this looks good to me, thanks!

PR #17968, opened 2025-08-20
fix: use len() instead of iter().count() for trace logging

LandLord529:
Noticed that we're using `iter().count()` to get the size for trace logging when we already have a `len()` method available.

`ValidAnnouncementData` has a `len()` method from the `HandleMempoolData` trait that just returns the HashMap size directly. No need to create an iterator and count everything.

Quick benchmark shows it's actually quite a bit faster:
```
Elements    len() (μs)    count() (μs)    Speedup
10          0.008         0.044           5.5x
100         0.013         0.047           3.7x  
1000        0.008         0.043           5.6x
5000        0.008         0.043           5.3x
10000       0.009         0.045           5.0x
```

Not a huge deal but this trace gets hit pretty often during tx propagation so might as well use the more efficient version.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -757,7 +757,7 @@ impl<Pool: TransactionPool, N: NetworkPrimitives, PBundle: TransactionPolicies>
 
         trace!(target: "net::tx::propagation",
             peer_id=format!("{peer_id:#}"),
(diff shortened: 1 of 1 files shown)


Rjected (approved):
lgtm, ~~I think these both call methods on the hashmap fields via deref, so `iter().count()` should be the same as `len()` here~~

I'm actually not sure if this resolves to the len method via deref or the trait method of the same name

PR #17969, opened 2025-08-21
docs(net): add Rreceipts69 document

wetkeyboard:
Add a document for the Receipts69 message


crates/net/network/src/message.rs
```diff
@@ -108,6 +108,10 @@ pub enum PeerResponse<N: NetworkPrimitives = EthNetworkPrimitives> {
         response: oneshot::Receiver<RequestResult<Receipts<N::Receipt>>>,
     },
     /// Represents a response to a request for receipts.
+    ///
+    /// This is a variant of `Receipts` that was introduced in `eth/66`.
```
mattsse:
```suggestion
    /// This is a variant of `Receipts` that was introduced in `eth/69`.
```

mattsse (approved):

PR #17971, opened 2025-08-21
fix(optimism): add debug_traceTransaction support for pre-bedrock blocks

### LINKED ISSUE #17914: debug_traceBlockByHash and debug_traceTransaction return "not found" for pre-Bedrock blocks or transactions on Optimism Mainnet op-reth nodes
### Describe the bug

When performing a debug_traceBlockByHash and debug_traceTransaction request on an Optimism Mainnet op-reth-based node for a block or transaction prior to the Bedrock upgrade, the response contains "not found" even if the `rollup.historicalrpc` flag is set.

Queries :
`curl localhost:8545 -X POST -H "Content-Type: application/json" --data '{"method":"debug_traceBlockByHash","params":["0x881a1373b07311817a14bd30d0e94065fcd6746cc47b036d8d93cb4005d89bba", {"tracer": "callTracer"}],"id":1,"jsonrpc":"2.0"}'`

`curl localhost:8545 -X POST -H "Content-Type: application/json" --data '{"method":"debug_traceTransaction","params":["0xdbdfa0f88b2cf815fdc1621bd20c2bd2b0eed4f0c56c9be2602957b5a60ec702", {"tracer": "callTracer"}], "id":1,"jsonrpc":"2.0"}'`

Responses : 
`{"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":"block not found: hash 0x881a1373b07311817a14bd30d0e94065fcd6746cc47b036d8d93cb4005d89bba"}}`

`{"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":"transaction not found"}}`

Expected responses : 
Block / Transaction details

### Steps to reproduce

1. Use an Optimism Mainnet op-reth-based node.
2. Execute the calls for a block that is pre-Bedrock upgrade.
3. Observe the response.

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

v1.5.1

2

Optimism Mainnet

Archive (default)

### Code of Conduct

fgimenez:
closes #17914 

forwards pre-bedrock `debug_traceTransaction` requests to historical endpoint. with this in place:

`$ curl localhost:8545 -X POST -H "Content-Type: application/json" --data '{"method":"debug_traceTransaction","params":["0xdbdfa0f88b2cf815fdc1621bd20c2bd2b0eed4f0c56c9be2602957b5a60ec702", {"tracer": "callTracer"}], "id":1,"jsonrpc":"2.0"}' {"jsonrpc":"2.0","id":1,"result":{"calls":[{"from":"0x25e1c58040f27ecf20bbd4ca83a09290326896b3","gas":"0x71b1d","gasUsed":"0x15cc3","input":"0x202ee0ed000000000000000000000000000000000000000000000000000000000000c0260000000000000000000000000000000000000000000000000000006ea50c77c9","output":"0x","to":"0x971dc14055cbc313a536f9ffd048e07530bd75df","type":"DELEGATECALL"}],"from":"0x2539ffc9ded82926a5aaee065e800c7d1de02454","gas":"0x74d84","gasUsed":"0x17271","input":"0x202ee0ed000000000000000000000000000000000000000000000000000000000000c0260000000000000000000000000000000000000000000000000000006ea50c77c9","output":"0x","time":"5.992559ms","to":"0x25e1c58040f27ecf20bbd4ca83a09290326896b3","type":"CALL","value":"0x0"}}`

### DIFF
--- a/crates/optimism/rpc/src/historical.rs
+++ b/crates/optimism/rpc/src/historical.rs
@@ -191,17 +191,32 @@ where
         parse_transaction_hash_from_params(&req.params())
             .ok()
             .map(|tx_hash| {
-                let tx_found =
-                    self.provider.transaction_by_hash_with_meta(tx_hash).ok().flatten().is_some();
-
-                if !tx_found {
-                    debug!(
-                        target: "rpc::historical",
-                        ?tx_hash,
-                        "transaction not found locally, forwarding to historical endpoint"
-                    );
+                // Check if we can find the transaction locally and get its metadata
+                match self.provider.transaction_by_hash_with_meta(tx_hash) {
+                    Ok(Some((_, meta))) => {
+                        // Transaction found - check if it's pre-bedrock based on block number
+                        let is_pre_bedrock = meta.block_number < self.bedrock_block;
+                        if is_pre_bedrock {
+                            debug!(
+                                target: "rpc::historical",
+                                ?tx_hash,
+                                block_num = meta.block_number,
+                                bedrock = self.bedrock_block,
+                                "transaction found in pre-bedrock block, forwarding to historical endpoint"
+                            );
+                        }
+                        is_pre_bedrock
+                    }
+                    _ => {
+                        // Transaction not found locally, optimistically forward to historical endpoint
+                        debug!(
+                            target: "rpc::historical",
+                            ?tx_hash,
+                            "transaction not found locally, forwarding to historical endpoint"
+                        );
+                        true
+                    }
                 }
-                !tx_found
             })
             .unwrap_or(false)
     }
--- a/crates/optimism/rpc/src/historical.rs
+++ b/crates/optimism/rpc/src/historical.rs
@@ -135,78 +135,12 @@ where
         let historical = self.historical.clone();
 
         Box::pin(async move {
-            // Special handling for debug_traceTransaction
-            if req.method_name() == "debug_traceTransaction" {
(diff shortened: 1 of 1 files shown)


crates/optimism/rpc/src/historical.rs
```diff
@@ -219,13 +167,146 @@ struct HistoricalRpcInner<P> {
...
+    }
+
+    /// Determines if a transaction request should be forwarded
+    fn should_forward_transaction(&self, req: &Request<'_>) -> bool {
+        parse_transaction_hash_from_params(&req.params())
+            .ok()
+            .map(|tx_hash| {
+                let tx_found =
+                    self.provider.transaction_by_hash_with_meta(tx_hash).ok().flatten().is_some();
+
```
mattsse: it might be the case that this chain imported the entire block history, in which case we would find the tx I believe.

since we're obtaining the additional metadata here as well we should also check, if we found the tx, that the meta.number is above/below the bedrock height, like we do with blocks

mattsse (changes requested):
nice,

smol nit

crates/optimism/rpc/src/historical.rs
fgimenez: makes sense, done here https://github.com/paradigmxyz/reth/pull/17971/commits/bb080ae883466203c02a01086995317ee1d61d83 ptal

mattsse (approved):

PR #17972, opened 2025-08-21
chore: remove unused pooled_transaction_hashes_max

bendanzhentan:
This PR removed the unused `pooled_transaction_hashes_max` method. And also, the implementation of this functions consume all the transactions from pool but only take `max`, it's wasteful.

```rust
    fn pooled_transaction_hashes_max(&self, max: usize) -> Vec<TxHash> {
        self.pooled_transaction_hashes().into_iter().take(max).collect()
    }
```


mattsse (commented):
this is an api function and should stay

PR #17973, opened 2025-08-21
refactor(txpool): use proper method to get transaction by ID

keroro520:
Replace direct pool data access with get_transaction_by_transaction_id method for better encapsulation and consistency (same with the other methods, using `self.pool.get_xxx()`).


mattsse (commented):
this is fine, I prefer to be explicit here

PR #17974, opened 2025-08-21
Chore：New sorting implementation items added

LkingX33:
New sorting implementation items added；
Improving the readability of complex expressions.


mattsse (commented):
we're good on lints for now

PR #17975, opened 2025-08-21
perf: optimize canonical_hashes_range to O(n) complexity

smileclown2024:
Replace inefficient O(n²) insert(0, ...) pattern with push + reverse. This reduces complexity from O(n²) to O(n) where n is the number of in-memory blocks in range.

### DIFF
--- a/crates/chain-state/src/memory_overlay.rs
+++ b/crates/chain-state/src/memory_overlay.rs
@@ -87,10 +87,11 @@ impl<N: NodePrimitives> BlockHashReader for MemoryOverlayStateProviderRef<'_, N>
         let mut in_memory_hashes = Vec::new();
         for block in &self.in_memory {
             if range.contains(&block.recovered_block().number()) {
-                in_memory_hashes.insert(0, block.recovered_block().hash());
+                in_memory_hashes.push(block.recovered_block().hash());
                 earliest_block_number = Some(block.recovered_block().number());
             }
         }
+        in_memory_hashes.reverse();
 
         let mut hashes =
             self.historical.canonical_hashes_range(start, earliest_block_number.unwrap_or(end))?;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
@DaniPopes thoughts?

we only expect ~3 entries here at most

crates/chain-state/src/memory_overlay.rs
```diff
@@ -87,10 +87,11 @@ impl<N: NodePrimitives> BlockHashReader for MemoryOverlayStateProviderRef<'_, N>
         let mut in_memory_hashes = Vec::new();
         for block in &self.in_memory {
             if range.contains(&block.recovered_block().number()) {
-                in_memory_hashes.insert(0, block.recovered_block().hash());
+                in_memory_hashes.push(block.recovered_block().hash());
                 earliest_block_number = Some(block.recovered_block().number());
```
smileclown2024, line 92: BTW, according to its literal meaning, the update of `earliest_block_number` seems somewhat problematic. Should it be changed to below? @mattsse 

```rust
if earliest_block_number.is_none() ||
    block.recovered_block().number() < earliest_block_number.unwrap()
{
    earliest_block_number = Some(block.recovered_block().number());
}
```

crates/chain-state/src/memory_overlay.rs
```diff
@@ -87,10 +87,11 @@ impl<N: NodePrimitives> BlockHashReader for MemoryOverlayStateProviderRef<'_, N>
...
             }
         }
+        in_memory_hashes.reverse();
```
mattsse, line 99: can we add a note here, explaining that we need to reverse the order here so that this is in desc order

mattsse (changes requested):
I'm okay with this, using reverse feels more natural actually, but would like to also add an additional note

crates/chain-state/src/memory_overlay.rs
smileclown2024: Code comment added at c19c8c0ecb89533e7acde245f64738d337b9ebcb. PTAL

mattsse (approved):

PR #17977, opened 2025-08-21
chore(net): refactor poll function to make it much clearer and maintainable

wetkeyboard:
Refactor the poll function to make it much clearer and maintainable, split the large `poll` functionto the below helpers to make the code more readable:

- `process_incoming_commands` - Processes commands from the session manager
- `process_internal_requests` - Handles internal peer requests
- `process_received_requests` - Processes received requests from remote peers
- `send_queued_messages` - Sends queued outgoing messages
- `handle_receive_loop` - Handles the main message receiving loop
- `should_throttle_incoming` - Checks if incoming messages should be throttled
- `handle_interval_operations` - Handles interval-based operations like range updates and timeout checks

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -533,6 +533,209 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
         // terminate the task
         Some(Poll::Ready(()))
     }
+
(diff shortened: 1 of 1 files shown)


mattsse (commented):
not something I want to touch

Issue #17858, opened 2025-08-13
native op flashblocks support


mattsse:
### Describe the feature

in order to build the pending block we need a ws subscription to the flashblocks url and receive all new flashblocks.

this must be integrated in the rollupargs.

next we need support for building the pending block in opethapi:

https://github.com/paradigmxyz/reth/blob/ee8c893f59f62276383cac5ec70dba6672250449/crates/optimism/rpc/src/eth/pending_block.rs#L33-L42

which currently only returns the latest.

in order to build the pending block + state we need to do something similar to the eth impl:

https://github.com/paradigmxyz/reth/blob/ee8c893f59f62276383cac5ec70dba6672250449/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L189-L203

but instead of pulling the txs from the pool, we take the txs from the already received flashblocks.

## TODO
1. first create a future/task that keeps track of the flashblocks by tracking blocknumber + parenthash + flashblocks, which then resets when the block number increases.
2. next we need to integrate local flashblocks replay, by basically implementing build_block in OpEthApi
3. for this we need shared state to which the service of 1) writes the current state of the current flashblock (`Arc<tokio::Rwlock` basically)

once we have that we can rollout the actual pending => pending flashblock state

assigning @Rimeeeeee to continue on #17854

emhane:
blocked by https://github.com/paradigmxyz/reth/issues/17876

PR #17982, opened 2025-08-21
feat(optimism): Add new `reth-optimism-flashblocks` crate

RomanHodulak:
Part of #17858

Adds a new crate for the Flashblocks integration.

Flashblocks are not part of Ethereum specification so they do not belong to any existing crate.

We have decided to put Flashblocks integration in the Optimism subfolder, they are intended for `op-reth`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9267,6 +9267,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "reth-optimism-flashblocks"
+version = "1.6.0"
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
since flashblocks are an op thing we should put this into the optimism folder and also name this crate reth-op-flashblocks or similar

mattsse (approved):
lgtm

PR #17983, opened 2025-08-21
fix(cli): make --datadir option global for consistency

LandLord529:
The `--datadir` option is not global, causing inconsistent CLI behavior compared to other fundamental options like `--chain`.

**Current behavior (confusing)**:
```bash
$ reth db stats --datadir /custom/path
error: unexpected argument '--datadir' found
  tip: 'db --datadir' exists

$ reth db --datadir /custom/path stats  # works, but unintuitive
```

After this change
Both patterns work (consistent with `--chain` behavior):
```bash
reth db --datadir /custom/path stats     # existing (still works)  
reth db stats --datadir /custom/path     # new (now works)
reth --datadir /custom/path db stats     # also works
```


mattsse (commented):
no more globals

PR #17987, opened 2025-08-21
feat(optimism): Add `FlashBlockWsStream` for streaming flashblocks from a websocket connection

RomanHodulak:
Part of #17858

Adds a stream responsible for yielding `FlashBlock`s received from a websocket connection.

### DIFF
--- a/crates/optimism/flashblocks/src/ws/decoding.rs
+++ b/crates/optimism/flashblocks/src/ws/decoding.rs
@@ -1,0 +1,66 @@
+use crate::{ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashBlock, Metadata};
+use alloy_rpc_types_engine::PayloadId;
+use serde::{Deserialize, Serialize};
+use std::{fmt::Debug, io::Read};
+
+#[derive(Clone, Debug, PartialEq, Default, Deserialize, Serialize)]
+struct FlashblocksPayloadV1 {
+    /// The payload id of the flashblock
+    pub payload_id: PayloadId,
+    /// The index of the flashblock in the block
+    pub index: u64,
+    /// The base execution payload configuration
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub base: Option<ExecutionPayloadBaseV1>,
+    /// The delta/diff containing modified portions of the execution payload
+    pub diff: ExecutionPayloadFlashblockDeltaV1,
+    /// Additional metadata associated with the flashblock
+    pub metadata: serde_json::Value,
+}
+
+impl FlashBlock {
+    pub(crate) fn decode(bytes: &[u8]) -> eyre::Result<Self> {
+        try_decode_message(bytes)
+    }
+}
+
+fn try_decode_message(bytes: &[u8]) -> eyre::Result<FlashBlock> {
+    let text = try_parse_message(bytes)?;
+
+    let payload: FlashblocksPayloadV1 = match serde_json::from_str(&text) {
+        Ok(m) => m,
+        Err(e) => {
+            return Err(eyre::eyre!("failed to parse message: {}", e));
+        }
+    };
+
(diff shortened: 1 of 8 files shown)


crates/optimism/flashblocks/src/ws/decoding.rs
```diff
@@ -0,0 +1,66 @@
...
+        metadata,
+    })
+}
+
+fn try_parse_message(bytes: &[u8]) -> eyre::Result<String> {
+    if let Ok(text) = String::from_utf8(bytes.to_vec()) {
+        if text.trim_start().starts_with('{') {
+            return Ok(text);
+        }
+    }
```
mattsse: do we need the alloc here?

I think we can do `bytes.trim_ascii_start().starts_with(b"{")`?

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -0,0 +1,102 @@
...
+            self.connect();
+        }
+
+        if self.state == State::Connect {
+            match ready!(self.connect.poll_unpin(cx)) {
+                Ok((stream, _)) => self.stream(stream),
+                Err(err) => {
+                    self.state = State::Initial;
+
+                    return Poll::Ready(Some(Err(err.into())))
```
mattsse, line 45: ah okay, this ensures we create a new connect future and avoid polling the future again

crates/optimism/flashblocks/src/ws/decoding.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl FlashBlock {
+    pub(crate) fn decode(bytes: &[u8]) -> eyre::Result<Self> {
+        try_decode_message(bytes)
+    }
+}
+
+fn try_decode_message(bytes: &[u8]) -> eyre::Result<FlashBlock> {
+    let text = try_parse_message(bytes)?;
```
mattsse: can we move this into `FlashBlock::decode` dont think we need this as a standalone fn?

crates/optimism/flashblocks/src/ws/decoding.rs
```diff
@@ -0,0 +1,66 @@
...
+    Ok(FlashBlock {
+        payload_id: payload.payload_id,
+        index: payload.index,
+        base: payload.base,
+        diff: payload.diff,
+        metadata,
+    })
+}
+
+fn try_parse_message(bytes: &[u8]) -> eyre::Result<String> {
```
mattsse: can add docs that this supports both json and bortli json

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -0,0 +1,102 @@
...
+use tokio::net::TcpStream;
+use tokio_tungstenite::{
+    connect_async,
+    tungstenite::{handshake::client::Response, Error, Message},
+    MaybeTlsStream, WebSocketStream,
+};
+use url::Url;
+
+/// An asynchronous stream of [`FlashBlock`] from a websocket connection.
+pub struct FlashBlockWsStream {
```
mattsse: can mention here that this will retry the connection on error

mattsse (changes requested):
cool, only a few nits

crates/optimism/flashblocks/src/ws/decoding.rs
```diff
@@ -0,0 +1,65 @@
+use crate::{ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashBlock, Metadata};
+use alloy_primitives::bytes::Bytes;
+use alloy_rpc_types_engine::PayloadId;
+use serde::{Deserialize, Serialize};
+use std::{fmt::Debug, io};
+
+#[derive(Clone, Debug, PartialEq, Default, Deserialize, Serialize)]
+struct FlashblocksPayloadV1 {
```
mattsse, line 8: ideally we can upstream those to alloy once stabilized

mattsse (approved):

PR #17988, opened 2025-08-21
docs(op): op chains don't require deposit contracts, so as dev chain

schoentgenaa:
No need to set the deposit contract address for the op chains, so remove it in the docs.


mattsse (approved):

schoentgenaa:
The merge bot is not working here @mattsse

schoentgenaa:
@mattsse The merge bot CI may broken, would you mind having a check here. @mattsse

PR #17989, opened 2025-08-21
chore: remove not used block/receipt memory limiter constants

wetkeyboard:
Currently we are using the length limiter of the block receipt cache, the memory limiter was not used anymore, so try to remove those not used constants.


mattsse (approved):
ah yes, we changed the cache from mem to items based because mem wasnt accurate

PR #17990, opened 2025-08-21
feat(nippy-jar): implement atomic deletion with consistency guarantees

wetkeyboard:
Implement comprehensive consistency mechanisms for NippyJar delete operations
to handle unexpected shutdown scenarios gracefully.


mattsse (commented):
this is fine

PR #17994, opened 2025-08-21
fix: insecure default in RPC auth layer documentation

prestoalvarez:
## Description

Replace hardcoded JWT secret and `0.0.0.0` binding in the auth layer doc example with secure defaults:

- **Before**: `0.0.0.0:8551` with hardcoded secret
- **After**: `127.0.0.1:8551` with secret from `RETH_AUTH_SECRET` env var

This prevents copy-paste vulnerabilities where auth RPC endpoints could be exposed externally with predictable credentials.


mattsse (commented):
this is example code which is fine

PR #17996, opened 2025-08-22
feat(reth-bench): add --advance option for relative block ranges

### LINKED ISSUE #17995: Add relative block advancement CLI argument to reth-bench
## Add relative block advancement CLI argument to reth-bench

Currently, reth-bench requires manually specifying absolute block ranges with `--from` and `--to` arguments. When benchmarking needs to be repeated with different block ranges, users must manually query the current block number and calculate the target range, which is cumbersome.

### Current implementation

The block range arguments are defined in:
https://github.com/paradigmxyz/reth/blob/main/crates/node/core/src/args/benchmark_args.rs#L10-L16

```rust
/// Run the benchmark from a specific block.
#[arg(long, verbatim_doc_comment)]
pub from: Option<u64>,

/// Run the benchmark to a specific block.
#[arg(long, verbatim_doc_comment)]
pub to: Option<u64>,
```

The range parsing logic is in:
https://github.com/paradigmxyz/reth/blob/main/bin/reth-bench/src/bench_mode.rs#L24-L36

### Proposed solution

Add a new CLI argument `--advance` that allows specifying a relative number of blocks to advance from the current head:

```rust
/// Number of blocks to advance from the current head block.
/// When specified, automatically sets --from to current head + 1 and --to to current head + advance.
/// Cannot be used together with explicit --from and --to arguments.
#[arg(long, conflicts_with_all = &["from", "to"], verbatim_doc_comment)]
pub advance: Option<u64>,
```

### Implementation changes needed

1. Update `BenchmarkArgs` struct in `crates/node/core/src/args/benchmark_args.rs` to add the new `advance` field

2. Modify `BenchContext::new()` in `bin/reth-bench/src/bench/context.rs` to:
   - When `advance` is specified, query the latest block first
   - Calculate `from` as `latest_block + 1` and `to` as `latest_block + advance`
   - Pass these calculated values to `BenchMode::new()`

3. Update `BenchMode::new()` validation logic if needed to handle the new parameter flow

### Example usage

Instead of:
```bash
# Query current block manually
# Assuming current block is 21000000
reth-bench new-payload-fcu --from 21000001 --to 21000100 --jwt-secret jwt.hex
```

Users could run:
```bash
reth-bench new-payload-fcu --advance 100 --jwt-secret jwt.hex
```

This would automatically benchmark the next 100 blocks from the current head.

### Expected outcome

- Simplified workflow for benchmarking consecutive block ranges
- No need to manually query current block number
- Easier automation of benchmark scripts

### Test plan

Notes:
- Testing can be done on the hoodi network with snapshots. you can download a recent snapshot for reth here: https://ethpandaops.io/data/snapshots/

- When running with snapshot, you need to use `--datadir` arg.

- [ ] Verify `--advance` argument is properly parsed and conflicts with `--from`/`--to`
- [ ] Test with `reth-bench new-payload-fcu --advance 10` - should benchmark next 10 blocks from current head
- [ ] Test with `reth-bench new-payload-only --advance 50` - should work with all benchmark subcommands
- [ ] Verify error handling when `--advance` is used with `--from` or `--to` (should conflict)
- [ ] Test edge case: `--advance 0` should be rejected with appropriate error message
- [ ] Test that calculated range (from current head) matches expected blocks being benchmarked
- [ ] Verify benchmark output shows correct block range in logs

dharmvr1:
- Introduces a new CLI argument `--advance` to specify a relative number
  of blocks to benchmark from the current head.
- Automatically computes `from` = head + 1 and `to` = head + advance.
- No changes needed to `BenchMode::new()`; existing validation works as expected.
- Simplifies benchmarking consecutive block ranges and reduces manual calculations.

closes #17995

### DIFF
--- a/bin/reth-bench/src/bench/context.rs
+++ b/bin/reth-bench/src/bench/context.rs
@@ -71,11 +71,10 @@ impl BenchContext {
 
             let head_block = block_provider
                 .get_block_by_number(BlockNumberOrTag::Latest)
-                .full()
                 .await?
                 .ok_or_else(|| eyre::eyre!("Failed to fetch latest block for --advance"))?;
             let head_number = head_block.header.number;
-            (Some(head_number + 1), Some(head_number + advance))
+            (Some(head_number), Some(head_number + advance))
         } else {
             (bench_args.from, bench_args.to)
         };
--- a/bin/reth-bench/src/bench/context.rs
+++ b/bin/reth-bench/src/bench/context.rs
@@ -58,9 +58,27 @@ impl BenchContext {
             .await?
             .is_empty();
 
+        // Computes the block range for the benchmark.
+        //
+        // - If `--advance` is provided, fetches the latest block and sets:
+        //     - `from = head + 1`
+        //     - `to = head + advance`
+        // - Otherwise, uses the values from `--from` and `--to`.
+        let (from, to) = if let Some(advance) = bench_args.advance {
+            let head_block = block_provider
+                .get_block_by_number(BlockNumberOrTag::Latest)
+                .full()
+                .await?
+                .ok_or_else(|| eyre::eyre!("Failed to fetch latest block for --advance"))?;
+            let head_number = head_block.header.number;
+            (Some(head_number + 1), Some(head_number + advance))
+        } else {
+            (bench_args.from, bench_args.to)
+        };
+
         // If neither `--from` nor `--to` are provided, we will run the benchmark continuously,
         // starting at the latest block.
-        let mut benchmark_mode = BenchMode::new(bench_args.from, bench_args.to)?;
+        let mut benchmark_mode = BenchMode::new(from, to)?;
 
         // construct the authenticated provider
         let auth_jwt = bench_args
--- a/bin/reth-bench/src/bench/context.rs
+++ b/bin/reth-bench/src/bench/context.rs
@@ -65,6 +65,10 @@ impl BenchContext {
         //     - `to = head + advance`
         // - Otherwise, uses the values from `--from` and `--to`.
         let (from, to) = if let Some(advance) = bench_args.advance {
+            if advance == 0 {
+                return Err(eyre::eyre!("--advance must be greater than 0"));
+            }
+
             let head_block = block_provider
                 .get_block_by_number(BlockNumberOrTag::Latest)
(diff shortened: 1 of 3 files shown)


bin/reth-bench/src/bench/context.rs
```diff
@@ -58,9 +58,31 @@ impl BenchContext {
...
+        //     - `to = head + advance`
+        // - Otherwise, uses the values from `--from` and `--to`.
+        let (from, to) = if let Some(advance) = bench_args.advance {
+            if advance == 0 {
+                return Err(eyre::eyre!("--advance must be greater than 0"));
+            }
+
+            let head_block = block_provider
+                .get_block_by_number(BlockNumberOrTag::Latest)
+                .full()
```
Rjected: I don't think we would need `full` for this

bin/reth-bench/src/bench/context.rs
```diff
@@ -58,9 +58,31 @@ impl BenchContext {
...
+                .await?
+                .ok_or_else(|| eyre::eyre!("Failed to fetch latest block for --advance"))?;
+            let head_number = head_block.header.number;
+            (Some(head_number + 1), Some(head_number + advance))
```
Rjected: i think this should actually be:
```suggestion
            let head_number = head_block.header.number;
            (Some(head_number), Some(head_number + advance))
```

Rjected (changes requested):
just some nits, this makes sense so far

Rjected (dismissed):
ah, sorry, I just realized that we should be using the `auth_provider`, ie the engine endpoint, rather than `block_provider` to get the head block. Right now it will fail because it fetches the tip of the chain, then tries to fetch blocks past the tip which don't exist.

re: testing, I'll just test this with my local hoodi node, this shouldnt need unit tests or anything

yongkangc (approved):
awsome, tested the flag by running node locally and advancing was successful, LGTM 🚀 

```
reth-bench \
    new-payload-fcu \
    --advance 2 \
    --jwt-secret jwt.hex \
    --rpc-url https://rpc.hoodi.ethpandaops.io \
    --engine-rpc-url http://localhost:8551
```

yongkangc:
@Rjected merging this in first - using this for benching 🙏🏻

PR #17999, opened 2025-08-22
refactor(eth-wire): remove EthVersion::total_messages in favor of EthMessageID::max

MatusKysel:
## Summary

  Remove incorrect `EthVersion::total_messages()` method and replace duplicate `EthMessageID::max(version) + 1` calculations with a dedicated `message_count()` function.

  ## Motivation

  The `EthVersion::total_messages()` method returned incorrect hardcoded values that didn't match the actual message IDs. Additionally, the codebase had multiple instances of `EthMessageID::max(version) +
  1` scattered throughout, which creates code duplication and makes the intent less clear. This pattern is used for message ID multiplexing where we need the total count of message types for a given
  Ethereum protocol version.

  ## Changes

  - Remove `EthVersion::total_messages()` with incorrect hardcoded counts
  - Add `EthMessageID::message_count(version)` function that encapsulates `max(version) + 1`
  - Replace all occurrences of `EthMessageID::max(version) + 1` with the new function:
    - `crates/net/eth-wire/src/capability.rs`: Updated `num_messages()` method
    - `crates/net/eth-wire/src/protocol.rs`: Updated `Protocol::eth()` method
    - `crates/net/eth-wire/src/eth_snap_stream.rs`: Updated message ID calculations and comments
  - Improve code clarity by using a descriptive function name instead of manual arithmetic
  - Add documentation referencing the Go-Ethereum implementation for context
  - Add tests to verify correct message counts

  ## Refs
  https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/eth/protocols/eth/protocol.go#L45-L47

### DIFF
--- a/crates/net/eth-wire/src/protocol.rs
+++ b/crates/net/eth-wire/src/protocol.rs
@@ -26,7 +26,7 @@ impl Protocol {
     /// Returns the corresponding eth capability for the given version.
     pub const fn eth(version: EthVersion) -> Self {
         let cap = Capability::eth(version);
-        let messages = EthMessageID::max(version) + 1;
+        let messages = EthMessageID::message_count(version);
         Self::new(cap, messages)
     }
 
@@ -79,7 +79,7 @@ mod tests {
     #[test]
     fn test_protocol_eth_message_count() {
         // Test that Protocol::eth() returns correct message counts for each version
-        // This ensures that EthMessageID::max(version) + 1 produces the expected results
+        // This ensures that EthMessageID::message_count() produces the expected results
         assert_eq!(Protocol::eth(EthVersion::Eth66).messages(), 17);
         assert_eq!(Protocol::eth(EthVersion::Eth67).messages(), 17);
         assert_eq!(Protocol::eth(EthVersion::Eth68).messages(), 17);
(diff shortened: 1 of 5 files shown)


crates/net/eth-wire/src/protocol.rs
```diff
@@ -26,7 +26,7 @@ impl Protocol {
     /// Returns the corresponding eth capability for the given version.
     pub const fn eth(version: EthVersion) -> Self {
         let cap = Capability::eth(version);
-        let messages = version.total_messages();
+        let messages = EthMessageID::max(version) + 1;
```
mattsse: can we also reference

https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/eth/protocols/eth/protocol.go#L45-L47

and make this another function instead?

mattsse (changes requested):
ty, smol nit

crates/net/eth-wire/src/protocol.rs
MatusKysel: fixed by https://github.com/paradigmxyz/reth/pull/17999/commits/a58e8357ab28384d687c31a3fc4e023981e86a4c

mattsse (approved):

PR #18000, opened 2025-08-22
feat: enhance engine tree metrics

fynnss:
Add latency metrics for engine tree operations to enable performance comparison with Geth's chain processing benchmarks. Currently Reth lacks direct metrics to compare against Geth's chain_inserts/ chain_execution/ chain_validation/ chain_write latency data.

  Changes

  - Added three new histogram metrics to EngineMetrics:
    - block_insert_total_duration: Measures total block insertion time
    - block_execution_duration: Tracks block execution latency
    - block_validation_duration: Records state root validation time
  - Instrumented critical code paths in insert_new_payload and validate_and_execute_payload to collect latency data
  - These metrics will help identify performance bottlenecks during live sync

### DIFF
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -58,7 +58,8 @@ pub(crate) struct EngineMetrics {
     pub(crate) failed_new_payload_response_deliveries: Counter,
     /// Tracks the how often we failed to deliver a forkchoice update response.
     pub(crate) failed_forkchoice_updated_response_deliveries: Counter,
-    // TODO add latency metrics
+    /// block insert duration
+    pub(crate) block_insert_total_duration: Histogram,
 }
 
 /// Metrics for non-execution related block validation.
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2165,8 +2165,7 @@ where
         self.metrics
             .engine
             .block_insert_total_duration
-            .record(elapsed.as_secs_f64());
-
+            .record(block_insert_start.elapsed().as_secs_f64());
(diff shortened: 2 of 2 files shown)


Rjected (changes requested):
we already have execution metrics, see `ExecutorMetrics`:
https://github.com/paradigmxyz/reth/blob/a3298ecfdd2f604e6fc288bf803aa2980637e82d/crates/evm/evm/src/metrics.rs#L49-L75

we also have payload validation metrics but when reviewing this PR I noticed that they no longer work properly since we changed the order of operations:
https://github.com/paradigmxyz/reth/issues/18014

however I think we don't yet have a metrics for total block insert time, so for this PR let's just keep that one. We do have `engine_newPayload` metrics but that might be too tied to the engine api.

fynnss:
@Rjected  Thanks for your code review. 

That make sense. Already removed metrics for the execution and validation duration

Rjected (approved):
thanks! this looks good to me, just needs fmt

Rjected (approved):

fynnss:
It's wried.

Diff in /home/runner/work/reth/reth/crates/engine/tree/src/tree/mod.rs:2186:
         };
         self.emit_event(EngineApiEvent::BeaconConsensus(engine_event));
 
-        self.metrics
-            .engine
-            .block_insert_total_duration
-            .record(elapsed.as_secs_f64());
+        self.metrics.engine.block_insert_total_duration.record(elapsed.as_secs_f64());
 
         debug!(target: "engine::tree", block=?block_num_hash, "Finished inserting block");
         Ok(InsertPayloadOk::Inserted(BlockStatus::Valid))
Error: Process completed with exit code 1.

The lint fmt let me change the format back.

yongkangc (approved):
pending fmt fixes

Issue #18002, opened 2025-08-22
SealedBlock encodable only encodes the body


mattsse:
### Describe the feature

these aren't symmetrical:

https://github.com/paradigmxyz/reth/blob/d8e8d67ff8a5c823a48c81bd1f27faa9a8d3aa35/crates/primitives-traits/src/block/sealed.rs#L309-L320

we need to fix this so that we encode the header then encode the block

a quick fix for this is .clone().into_block().encode

but this is a bit expensive.
it's a bit tricky because we'd need something like this to make this more efficient

https://github.com/paradigmxyz/reth/blob/d8e8d67ff8a5c823a48c81bd1f27faa9a8d3aa35/crates/primitives-traits/src/block/mod.rs#L143-L144

PR #18003, opened 2025-08-22
fix: rlp encoding for sealedblock

mattsse:
hotfix for #18002

we can optimize this with upcoming helpers https://github.com/alloy-rs/alloy/pull/2809


fgimenez (approved):
lgtm

PR #18009, opened 2025-08-22
feat(optimism): Add `FlashBlockService` that builds blocks from `FlashBlock`s

RomanHodulak:
Part of #17858

Adds a service that receives `FlashBlock`s and builds `ExecutedBlock`s.

The `add_flash_block` part of the implementation implements the `FlashBlock`s accumulation logic inspired by the `base/node-reth`.


docs/vocs/docs/pages/sdk/custom-node/transactions.mdx
```diff
@@ -257,7 +257,7 @@ impl SignedTransaction for CustomTransaction {
 The `Compact` support is largely derived and abstracted away with the help of a few minimal trivial implementations. One
 slightly interesting case is `FromTxCompact`. The actual decoding is delegated to `TxPayment`, where it is macro generated.
 Then it is put together with the `signature`, that given as one of the function arguments, via `Signed::new_unhashed`
-which creates the `Signed<TxPayment>` instance with no signature validation.
+aaaaaaaaaaaaaaaaaaaaaaaawhich creates the `Signed<TxPayment>` instance with no signature validation.
```
mattsse: whoops

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Depending on its index and associated block number, it may:
+    /// * Be added to all the flashblocks received prior using this function.
+    /// * Cause a reset of the flashblocks and become the sole member of the collection.
+    /// * Be ignored.
+    fn add_flash_block(&mut self, block: FlashBlock);
+
+    /// Returns the [`ExecutedBlock`] made purely out of [`FlashBlock`]s that were received using
+    /// [`Self::add_flash_block`].
+    fn execute(&mut self) -> Result<ExecutedBlock, Self::Err>;
+}
```
mattsse: do we need a trait for this, since this is kinda op specific, we can probably just use the struct RethFlashBlockService and name that FlashBlockService ?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// The Reth implementation of [`FlashBlockService`].
+#[derive(Debug)]
+pub struct RethFlashBlockService {
+    current: Option<ExecutedBlock>,
```
mattsse: I believe we need a generic here for N: NodePrimitives

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,72 @@
...
+    blocks: Vec<FlashBlock>,
+}
+
+impl FlashBlockService for RethFlashBlockService {
+    type Err = eyre::Error;
+
+    fn add_flash_block(&mut self, flashblock: FlashBlock) {
```
mattsse: thinking how this should run, I think we'd have

the websocket task, that forwards the messages over a channel to this one, which would be this function effectively, so I think we can maybe replace this either with a `rx: Receiver` or `St: Stream<Item= Flashblock>` generic

mattsse (changes requested):
some suggestions but I like the direction

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,77 @@
...
+impl<N: NodePrimitives> FlashBlockService<N> {
+    /// Adds the `block` into the collection.
+    ///
+    /// Depending on its index and associated block number, it may:
+    /// * Be added to all the flashblocks received prior using this function.
+    /// * Cause a reset of the flashblocks and become the sole member of the collection.
+    /// * Be ignored.
+    pub async fn add_flash_block(
+        &mut self,
+        mut flashblock: impl Stream<Item = FlashBlock> + Unpin,
```
mattsse: should this be a field instead?

crates/optimism/flashblocks/Cargo.toml
```diff
@@ -13,9 +13,22 @@ workspace = true
...
+reth-execution-types = { workspace = true, features = ["serde"] }
+reth-evm.workspace = true
+reth-revm.workspace = true
+reth-rpc-eth-api.workspace = true
+reth-rpc-eth-types.workspace = true
+reth-errors.workspace = true
+reth-storage-api.workspace = true
+
+# revm
+revm-database = { workspace = true }
```
mattsse: I dont think we need this

crates/optimism/flashblocks/Cargo.toml
```diff
@@ -13,9 +13,22 @@ workspace = true
...
 
 # alloy
 alloy-eips = { workspace = true, features = ["serde"] }
+alloy-evm = { workspace = true }
```
mattsse:
```suggestion
alloy-evm.workspace = true
```

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        builder.apply_pre_execution_changes()?;
+
+        let best_txs = self.blocks.iter().flat_map(|v| v.diff.transactions.clone());
+
+        for tx in best_txs {
+            let tx = N::SignedTx::decode(&mut tx.as_ref())?;
+            let signer = tx.recover_signer_unchecked()?;
+            let tx = Recovered::new_unchecked(tx, signer);
+            let _gas_used = builder.execute_transaction(tx)?;
```
mattsse: there's a smol improvement we can do which is like this

https://github.com/paradigmxyz/reth/blob/f3c2a3dc2706bb9ace87b130cf13730301135a0f/crates/optimism/evm/src/lib.rs#L288-L291

because for OP we can reuse the encoded bytes for execution

so we can use decode_2718_exact

and then wrap into `WithEncoded<Recovered<T>>`

crates/optimism/flashblocks/src/service.rs
```diff
@@ -0,0 +1,179 @@
...
+{
+    type Item = eyre::Result<ExecutedBlock<N>>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        loop {
+            match this.rx.poll_next_unpin(cx) {
+                Poll::Ready(Some(flashblock)) => this.add_flash_block(flashblock),
+                Poll::Ready(None) => return Poll::Ready(None),
+                Poll::Pending => return Poll::Ready(Some(this.execute())),
```
mattsse, line 174: how this part will actually work we'll need to figure out we actually wire this to the rpc part, so make this doesn't need to be a stream entirely and could manage the pending block internally entirely

but we can figure this part out later

mattsse (changes requested):
a few more nits

mattsse (approved):
lgtm,

we can do fine tuning as followups

PR #18013, opened 2025-08-22
feat: bump hive eest tests

fgimenez:
Configures eest tests to use a branch that includes this fix: https://github.com/ethereum/execution-spec-tests/pull/2016
Includes several fixes for the new tests:

* return header validation error instead of transaction validation error on excess blob gas
* return gas too low instead of gas below floor gas cost as per eip7623

Also two flaky `engine-api` tests are ignored and an expected failure that is now passing is removed from the list.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2522,8 +2522,15 @@ where
         self.emit_event(EngineApiEvent::BeaconConsensus(ConsensusEngineEvent::InvalidBlock(
             Box::new(block),
         )));
+        // Temporary fix for EIP-7623 test compatibility:
+        // Map "gas floor" errors to "call gas cost" for compatibility with test expectations
+        let mut error_str = validation_err.to_string();
+        if error_str.contains("gas floor") && error_str.contains("exceeds the gas limit") {
+            error_str = error_str.replace("gas floor", "call gas cost");
+        }
+
         Ok(PayloadStatus::new(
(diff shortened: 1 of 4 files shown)


fgimenez:
successful build from this branch https://github.com/paradigmxyz/reth/actions/runs/17165856940

mattsse (changes requested):
I'd like to exclude the code changes from this and tackle those separately

I believe this is what #18356 does

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2522,8 +2522,15 @@ where
         self.emit_event(EngineApiEvent::BeaconConsensus(ConsensusEngineEvent::InvalidBlock(
             Box::new(block),
         )));
+        // Temporary fix for EIP-7623 test compatibility:
+        // Map "gas floor" errors to "call gas cost" for compatibility with test expectations
+        let mut error_str = validation_err.to_string();
+        if error_str.contains("gas floor") && error_str.contains("exceeds the gas limit") {
+            error_str = error_str.replace("gas floor", "call gas cost");
+        }
```
mattsse, line 2530: this seems fine to cut down on eest failures, but could you please open an issue to improve this in the validation_err display impl?

I believe we can do consistent error messages there

mattsse (approved):

crates/engine/tree/src/tree/mod.rs
fgimenez: sure opened here https://github.com/paradigmxyz/reth/issues/18369

PR #18015, opened 2025-08-23
fix: move payload validation metric recording

schroyar:
closes https://github.com/paradigmxyz/reth/issues/18014


mattsse (commented):
this isnt the right fix and this is already wip

PR #18018, opened 2025-08-23
feat(cli): enable tokio handle in `CliRunner`

### LINKED ISSUE #17965: Support tokio handle in CliRunner
### Describe the feature

We currently require a dedicated runtime for the entire node:

https://github.com/paradigmxyz/reth/blob/0f26562bb646b85f8e0b1a9c06184c88cce8e06d/crates/cli/runner/src/lib.rs#L22-L24

https://github.com/paradigmxyz/reth/blob/0f26562bb646b85f8e0b1a9c06184c88cce8e06d/crates/cli/runner/src/lib.rs#L144-L151

it would be nice to have if this would also work with just a tokio handle

for this, we can try something like

```
enum RuntimeOrHandle {
    Runtime(Runtime),
    Handle(Handle),
}
```

probably with some helper fn for block_on to make this work

https://github.com/paradigmxyz/reth/blob/0110fbe0a94858c04452b00c1edbb86d99df9965/crates/cli/runner/src/lib.rs#L57-L61

this needs some tests and examples because `handle.block_on` can easily panic if misused and needs:

https://github.com/paradigmxyz/reth/blob/0110fbe0a94858c04452b00c1edbb86d99df9965/crates/cli/runner/src/lib.rs#L117-L117

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/17965.


crates/cli/runner/src/lib.rs
```diff
@@ -12,29 +12,151 @@
...
 use tracing::{debug, error, trace};
 
+/// A tokio runtime or handle.
+#[derive(Debug)]
+enum RuntimeOrHandle {
+    /// Owned runtime that can be used for blocking operations
+    Runtime(Runtime),
+    /// Handle to an existing runtime
+    Handle(Handle),
+}
```
mattsse: can we move this end of file so that the first type remains the pub one?

crates/cli/runner/src/lib.rs
```diff
@@ -12,29 +12,151 @@
...
+    /// ```ignore
+    /// // Use from a separate thread to avoid async context issues
+    /// std::thread::spawn(move || {
+    ///     let runner = CliRunner::from_handle(handle);
+    ///     runner.run_until_ctrl_c_with_handle(fut);
+    /// });
+    /// ```
+    pub const fn from_handle(handle: Handle) -> Self {
+        Self { executor: RuntimeOrHandle::Handle(handle) }
+    }
```
mattsse: we can also add `fn current`

mattsse (changes requested):
cool, only a few style nits

crates/cli/runner/src/lib.rs
```diff
@@ -12,29 +12,151 @@
...
+    /// Returns a reference to the inner tokio runtime handle.
+    fn handle(&self) -> &Handle {
+        match self {
+            Self::Runtime(rt) => rt.handle(),
+            Self::Handle(handle) => handle,
+        }
+    }
+
+    /// Attempts to extract the runtime, returning an error if only a handle is available.
+    fn into_runtime(self, operation: &str) -> Result<Runtime, std::io::Error> {
```
SuperFluffy: This returning `std::io::Result` feels like an abuse of `io::Error`? It's not an `io::Error`. Maybe return a custom error?

crates/cli/runner/src/lib.rs
```diff
@@ -12,29 +12,151 @@
...
+        let (rt, _handle) = self.into_runtime_or_handle();
+        rt.ok_or_else(|| {
+            std::io::Error::other(
+                format!("A tokio runtime is required to run {}. Please create a CliRunner with an owned runtime.", operation)
+            )
+        })
+    }
+
+    /// Chooses to return the owned runtime if it exists, otherwise returns `None` and the handle.
+    fn into_runtime_or_handle(self) -> (Option<Runtime>, Handle) {
```
SuperFluffy: The semantics of this object feels clumsy - can it be removed? It's only used in the other method above (`fn into_runtime`).

crates/cli/runner/src/lib.rs
```diff
@@ -228,3 +354,62 @@ where
...
+
+        // Separate thread is used to avoid async context
+        let result = std::thread::spawn(move || {
+            let runner = CliRunner::from_handle(handle);
+            runner.run_until_ctrl_c_with_handle(async { Ok::<(), std::io::Error>(()) })
+        })
+        .join()
+        .unwrap();
+
+        assert!(result.is_ok());
```
SuperFluffy, line 492: Making this `result.expect("<reason>")` would be an alternative to get more information on what might have gone wrong (useful if this should ever fail sometime in the future).

Alternatively I have taken a liking to `assert_ok!` from the `claims` crate recently if you are open to new dependencies.

crates/cli/runner/src/lib.rs
```diff
@@ -228,3 +354,62 @@ where
...
+        let handle = rt.handle().clone();
+
+        let result = std::thread::spawn(move || {
+            let runner = CliRunner::from_handle(handle);
+            runner.run_blocking_until_ctrl_c(async { Ok::<(), std::io::Error>(()) })
+        })
+        .join()
+        .unwrap();
+
+        assert!(result.is_ok());
```
SuperFluffy, line 507: Same comment as for above re `claims::assert_ok!` or `result.expect("<reason why this should be ok");`

crates/cli/runner/src/lib.rs
```diff
@@ -228,3 +354,62 @@ where
...
+    }
+
+    #[test]
+    fn test_handle_fails_without_handle_support() {
+        let rt = tokio_runtime().unwrap();
+        let runner = CliRunner::from_handle(rt.handle().clone());
+
+        let result = runner.run_until_ctrl_c(async { Ok::<(), std::io::Error>(()) });
+
+        assert!(result.is_err());
```
SuperFluffy, line 518: As above, the transpose: `let the_error = result.expect_err("<reason why this should error>");

PR #18019, opened 2025-08-23
feat(optimism): Implement `execute` for building pending block out of  `FlashBlock`s

RomanHodulak:
Part of #17858 

Implements block building from `FlashBlock`s.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -1,9 +1,24 @@
 use crate::FlashBlock;
+use alloy_evm::block::{BlockExecutionError, BlockValidationError};
+use alloy_primitives::private::alloy_rlp::Decodable;
+use eyre::OptionExt;
 use futures_util::{Stream, StreamExt};
 use reth_chain_state::ExecutedBlock;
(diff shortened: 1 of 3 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -68,19 +98,90 @@ impl<N: NodePrimitives, S> FlashBlockService<N, S> {
...
+            let _gas_used = match builder.execute_transaction(tx) {
+                Ok(gas_used) => gas_used,
+                Err(BlockExecutionError::Validation(BlockValidationError::InvalidTx {
+                    error,
+                    ..
+                })) => {
+                    if error.is_nonce_too_low() {
+                        // if the nonce is too low, we can skip this transaction
+                    }
+                    continue
```
mattsse, line 142: we can treat this as a hard error, because this should not happen because the expectation is that flashblocks are properly sequenced

mattsse (approved):
cool,

let#s merge this into the other pr then reviewing becomes slightly easier

crates/optimism/flashblocks/src/service.rs
RomanHodulak: 943816d

PR #18021, opened 2025-08-23
docs: update README.md

letmehateu:
fixed few minor typos and brought them to a more strict format


mattsse (approved):
I let this one through because better

PR #18022, opened 2025-08-23
perf(cli): optimize StorageChangeSets import in merkle stage dump

wetkeyboard:
Optimizes the `StorageChangeSets` table import in the merkle stage dump command by implementing range-based filtering instead of importing the entire table.
This will improve the execution when target range is much smaller than total table size


wetkeyboard:
@mattsse Hi there, mind check this when free? in case the bot close this by the S-stable label.

crates/cli/commands/src/stage/dump/merkle.rs
```diff
@@ -139,9 +146,18 @@ fn unwind_and_copy<N: ProviderNodeTypes>(
...
-    // TODO optimize we can actually just get the entries we need
-    output_db
-        .update(|tx| tx.import_dupsort::<tables::StorageChangeSets, _>(&unwind_inner_tx))??;
+    // Import only the StorageChangeSets entries within the block range (from, to)
+    output_db.update(|tx| {
+        let mut source_cursor = unwind_inner_tx.cursor_dup_read::<tables::StorageChangeSets>()?;
+        let mut target_cursor = tx.cursor_dup_write::<tables::StorageChangeSets>()?;
+
+        let range = BlockNumberAddress::range(from..=to);
+        for row in source_cursor.walk_range(range)? {
```
mediocregopher: @wetkeyboard were you able to test this code with a storage that has multiple keys? My understanding of `walk_range` is that it does not handle dup keys, it would only walk over the first subkey in every primary key, but I could be wrong.

wetkeyboard: I think you're right, the `walk_range` on a DupSort table only iterates through the first subkey-value pair for each primary key, and not all the duplicated entries.

wetkeyboard: @mediocregopher please take another look.
 
1. seek to the position at the beginning key
2. for each primary key in range: 
2.1. walk to get all the duplicate entries, 
2.2. and call `next_no_dump` to move to the next primary key

mediocregopher: Hey @wetkeyboard sorry I think I steered you wrong here, walk_range also works on dup tables. You can see an example of iterating over `StorageChangeSets` [here](https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/trie/db/src/state.rs#L233-L241)

Could you change it back to just use `walk_range`?

wetkeyboard: @mediocregopher sorry for the delay, reverted the changes

mediocregopher (approved):
Looks good to me but would like a check from @mattsse or @shekhirin

crates/cli/commands/src/stage/dump/merkle.rs
```diff
@@ -135,9 +142,18 @@ fn unwind_and_copy<N: ProviderNodeTypes>(
...
+            let (key, value) = row?;
+            target_cursor.append_dup(key, value)?;
+        }
+        Ok::<(), reth_db::DatabaseError>(())
+    })??;
```
mattsse: this only does it for the StorageChangeSets table

dont have enough context on the todo here

idk if we can/should do this @joshieDo @shekhirin

joshieDo: i think `import_table_with_range` should work here as well. We also use it here https://github.com/wetkeyboard/reth/blob/1f8b5467b5fc9fbbbbe3cfc142c42920e6155cbe/crates/cli/commands/src/stage/dump/merkle.rs#L56-L62 for a table which is also a dupsort table

shekhirin (approved):

PR #18024, opened 2025-08-24
fix: make ensure target syntax error

tmelhao:
When run `make ensure-typos`, it will auto-run the `cargo install typos-cli` to install the typos, but from the output message, seems it will let the use to install it manually, so I think it was the syntax error use of the bash script.


mattsse (commented):
this should be fine for all shells

PR #18027, opened 2025-08-24
feat(node): rename debug.rpc-consensus-ws to debug-rpc-consensus-url to suport HTTP

wetkeyboard:
Extends the `--debug.rpc-consensus-ws` debug flag to support both HTTP and WebSocket RPC endpoints.
Previously, this flag only supported WebSocket connections with subscriptions.

Now it automatically detects the connection type and uses the appropriate method:

- Websocket: uses subscriptions
- HTTP: uses polling with configurable interval


mattsse (commented):
this is a bit too complicated and imo just supporting ws is fine here

wetkeyboard:
> I see,
> 
> could we perhaps try simplifying this so that this works with a stream of blocks like a box dyn Stream
> 
> and initialize that one separately?, just to make the setup and the loop less verbose?

Made some changed @mattsse

mattsse (approved):
ty, I simplified this slightly

wetkeyboard:
> ty, I simplified this slightly

Thanks :)

PR #18030, opened 2025-08-25
perf(pool): remove unused hash in tx insertion/validation

hai-rise:


crates/transaction-pool/src/lib.rs
```diff
@@ -368,12 +368,8 @@ where
         &self,
         origin: TransactionOrigin,
         transaction: V::Transaction,
-    ) -> (TxHash, TransactionValidationOutcome<V::Transaction>) {
-        let hash = *transaction.hash();
```
hai-rise, line 372: 2% of transaction insertion is spent on this hash here, even though no one is using it!

crates/transaction-pool/src/lib.rs
```diff
@@ -383,13 +379,12 @@ where
         &self,
         origin: TransactionOrigin,
         transactions: impl IntoIterator<Item = V::Transaction> + Send,
-    ) -> Vec<(TxHash, TransactionValidationOutcome<V::Transaction>)> {
+    ) -> Vec<TransactionValidationOutcome<V::Transaction>> {
```
mattsse, line 382: ah right we dont need the hash return here

mattsse (approved):
cool, ty

PR #18033, opened 2025-08-25
feat: added debug_storageRangeAt support

Pana:


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1761,6 +1762,38 @@ impl<TX: DbTx + 'static, N: NodeTypes> StorageReader for DatabaseProvider<TX, N>
...
         Ok(storage_changeset_lists)
     }
+
+    fn storage_range_at(
+        &self,
+        contract_address: Address,
+        key_start: B256,
+        max_result: u64,
+    ) -> ProviderResult<StorageRangeResult> {
+        let mut storage_cursor = self.tx_ref().cursor_dup_read::<tables::PlainStorageState>()?;
```
shekhirin, line 1772: according to https://www.quicknode.com/docs/ethereum/debug_storageRangeAt, the `key_start` here is a hashed key, so we need to query the `HashedStorages` table.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1761,6 +1762,38 @@ impl<TX: DbTx + 'static, N: NodeTypes> StorageReader for DatabaseProvider<TX, N>
...
+
+        if let Some(first_entry) = storage_cursor.seek_by_key_subkey(contract_address, key_start)? {
+            results.push((first_entry.key, first_entry.value));
+
+            for _ in 1..max_result {
+                if let Some((_, entry)) = storage_cursor.next_dup()? {
+                    results.push((entry.key, entry.value));
+                    next_key = Some(B256::from(entry.key));
+                } else {
+                    break;
```
shekhirin, line 1785: `next_key` should be reset here, if there's less than `max_result` storages returned

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1761,6 +1762,38 @@ impl<TX: DbTx + 'static, N: NodeTypes> StorageReader for DatabaseProvider<TX, N>
...
+                }
+            }
+        }
+
+        let mut storage_map = BTreeMap::new();
+        for (key, value) in results {
+            storage_map.insert(key, StorageResult { key, value: B256::from(value.to_be_bytes()) });
+        }
```
shekhirin, line 1793: can we insert into the map directly in the loop above?

shekhirin (changes requested):
some comments, directionally looks right

0xKarl98:
Is this PR still active ?

PR #18040, opened 2025-08-25
Track keys added to the trie during execution

mediocregopher:
Fixes #18039


mediocregopher:
This doesn't show enough improvement to be worthwhile

PR #18043, opened 2025-08-25
refactor: eth-specific eth_config + bpo hardfork variants

### LINKED ISSUE #18060: Add support for eth_config
### Describe the feature

osaka comes with a new rpc endpoint, that contains the entire chainconfig:

https://github.com/alloy-rs/alloy/pull/2760

### Additional context

_No response_

klkvr:
Closes #18060 

Makes use of https://github.com/alloy-rs/hardforks/pull/40 to simplify bpo hardforks handling around forkids

Reverts extra methods for `ChainSpec`/`ConfigureEvm` required for `eth_config` implementation and instead relies on `forks_iter` to get prev/next hardforks

For obtaining precompiles we are now constructing a fake header, getting an EVM for it and use precompile stored inside of it.

This method is now only available on `EthereumAddOns` because of header manipulations. We can later abstract it futher if requested, which would require introducing a trait with timestamp setter or a smarter approach (like `precompiles_for_timestamp`)


crates/ethereum/node/src/node.rs
```diff
@@ -262,13 +268,20 @@ where
             Arc::new(EthereumEngineValidator::new(ctx.config.chain.clone())),
         );
 
+        let eth_config =
+            EthConfigHandler::new(ctx.node.provider().clone(), ctx.node.evm_config().clone());
```
mattsse, line 272: imo this makes sense here

crates/rpc/rpc-eth-api/src/helpers/config.rs
```diff
@@ -1,68 +1,126 @@
...
+
+    /// Returns base fork config for specific timestamp.
+    /// Returns [`None`] if no blob params were found for this fork.
+    fn fork_config_at_timestamp(&self, timestamp: u64) -> Option<EthBaseForkConfig> {
+        let chain_spec = self.provider.chain_spec();
+
+        // Fork config only exists for timestamp-based hardforks.
+        let head = Head { timestamp, number: u64::MAX, ..Default::default() };
+        Some(EthBaseForkConfig {
+            chain_id: alloy_primitives::U64::from(chain_spec.chain().id()),
```
mattsse, line 55: imo we can live with this

mattsse, line 55: hmm we should perhaps change this to u64, but not important

crates/rpc/rpc-eth-api/src/helpers/config.rs
```diff
@@ -1,68 +1,126 @@
...
-            {
-                let next_precompiles = evm_config.precompiles_for_timestamp(next_fork_timestamp);
+            if let Some(next_base_config) = self.fork_config_at_timestamp(next_fork_timestamp) {
+                let fake_header = {
+                    let mut header = latest;
+                    header.timestamp = next_fork_timestamp;
+                    header
+                };
+                let next_precompiles = evm_to_precompiles_map(
+                    self.evm_config.evm_for_block(EmptyDB::default(), &fake_header),
```
mattsse, line 122: this workaround seems fine

mattsse (approved):
lgtm!

PR #18044, opened 2025-08-25
fix(static_file_provider): Exception for Gnosis and Chiado

debjit-bw:
fix(static_file_provider): handle broken historical import for Gnosis and Chiado receipts.

For Gnosis and Chiado, we cannot execute pre-merge. That's why during ERA imports, we import the receipts too, as we cannot produce them with execution. Then we import a post-merge state and resume normal operations.

The receipts segment's consistency checking always results in a full rewind. Like Optimism made an exception for the full check, we should suppress this check only for the Receipts segment.

100 => Gnosis (Mainnet)
10200 => Chiado (Testnet)


mattsse (approved):
lgtm, I used namedchain enum to make this more obvious

PR #18045, opened 2025-08-25
fix: Fix state root related metrics

mediocregopher:
* Properly clear total/redundant proof counters once they are recorded, otherwise they grow monotonically and become meaningless.

* Correctly measure state root latency; previously we were including all steps between execution completion and waiting on the state root, which is not correct. We now accurately show that very often the state root is not a significant blocker for payload processing.

* Fix state root latency graph in grafana.

* Re-order total/redundant proof graphs in grafana, their layout was confusing previously.

## Total/Redundant Proofs

Before:
shot-1756139491

After:
shot-1756139517

## State Root Latency

Before/After of the code fix:
shot-1756139558

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -429,12 +429,11 @@ where
             handle.cache_metrics(),
         );
 
-        let (output, execution_finish) = if self.config.state_provider_metrics() {
+        let output = if self.config.state_provider_metrics() {
             let state_provider = InstrumentedStateProvider::from_state_provider(&state_provider);
-            let (output, execution_finish) =
-                ensure_ok!(self.execute_block(&state_provider, env, &input, &mut handle));
+            let output = ensure_ok!(self.execute_block(&state_provider, env, &input, &mut handle));
             state_provider.record_total_latency();
-            (output, execution_finish)
+            output
         } else {
             ensure_ok!(self.execute_block(&state_provider, env, &input, &mut handle))
         };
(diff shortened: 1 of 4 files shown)


etc/grafana/dashboards/overview.json
```diff
@@ -4735,14 +4735,14 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "reth_sparse_state_trie_multiproof_total_storage_nodes{$instance_label=\"$instance\",quantile=~\"(0|0.5|0.9|0.95|1)\"}",
+          "expr": "reth_sparse_state_trie_multiproof_skipped_storage_nodes{$instance_label=\"$instance\",quantile=~\"(0|0.5|0.9|0.95|1)\"}",
```
Rjected, line 4738: I thought we already had these in the graph, is it now duplicated?

Rjected (dismissed):
just one comment w.r.t. the skipped node graph. I think the change to the state root latency makes sense, if we follow this up with adding the validation metric to the "Block Processing Latency" graph. Given this is also ~10ms or so post-validation, it seems like we still calculate state root for a good amount of time after execution

etc/grafana/dashboards/overview.json
mediocregopher: You're right, something got messed up when I exported the JSON, should be fixed now

mediocregopher:
> just one comment w.r.t. the skipped node graph. I think the change to the state root latency makes sense, if we follow this up with adding the validation metric to the "Block Processing Latency" graph. Given this is also ~10ms or so post-validation, it seems like we still calculate state root for a good amount of time after execution

True enough, I made https://github.com/paradigmxyz/reth/issues/18057 to track adding the validation metric

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -429,12 +429,11 @@ where
             handle.cache_metrics(),
         );
 
-        let (output, execution_finish) = if self.config.state_provider_metrics() {
+        let output = if self.config.state_provider_metrics() {
```
mattsse, line 432: better

mattsse (approved):

PR #18046, opened 2025-08-25
feat(config): make StageConfig easier to maintain with declarative macro

CaptainJoly:
Added `define_stage_config!` macro that generates `StageConfig` struct automatically, to replace the manually maintained `StageConfig` struct.
Make the code more maintainable and easier to extend in the future.


mattsse (commented):
I dont think this is easier to maintain

PR #18047, opened 2025-08-25
fix(era): SlotIndex offset support negative value

nkysg:
ref https://github.com/status-im/nimbus-eth2/blob/stable/docs/e2store.md

```shell
Each entry in the slot index is a fixed-length 8-byte two's complement signed integer in little-endian, meaning that the entry for slot N can be found at index (N * 8) + 16 in the index.
```

### DIFF
--- a/crates/era-utils/src/export.rs
+++ b/crates/era-utils/src/export.rs
@@ -153,8 +153,8 @@ where
         let mut writer = Era1Writer::new(file);
         writer.write_version()?;
 
-        let mut offsets = Vec::<u64>::with_capacity(block_count);
-        let mut position = VERSION_ENTRY_SIZE as u64;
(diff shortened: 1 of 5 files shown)


lean-apple:
> ref https://github.com/status-im/nimbus-eth2/blob/stable/docs/e2store.md
> 
> ```shell
> Each entry in the slot index is a fixed-length 8-byte two's complement signed integer in little-endian, meaning that the entry for slot N can be found at index (N * 8) + 16 in the index.
> ```

I initially put it as i64, as it was specified in the specs, yet as there are file data offsets i didn't think it would be very useful.

mattsse (approved):
> Not particularly relevant for the era files, but keeping i64 for spec compatibility is probably safer choice :)

makes sense

PR #18048, opened 2025-08-25

crStiv:


mattsse (commented):
imo these typos are fine

PR #18049, opened 2025-08-25
fix: use deterministic RNG in state_root_task benchmark

avorylli:
Replace Address::random() with Address::random_with(&mut rng) to use the passed RNG parameter instead of global thread_rng()


Rjected (approved):
ty! this just needs fmt

Rjected (approved):

PR #18051, opened 2025-08-25
Fix: dead links in .md files

andrewshab3:
## Desсription: 
In `staged.md` and `db.md` files there are broken navigation links that don't lead anywhere.

## Rationale

i commented them out:
- Maintains UI structure for future implementation
- Makes it easier to restore when content is ready

If "Next Chapter" in `staged.md` file need contain a link to `db.md` i can commit it after review


mattsse (commented):
this only adds todos

PR #18053, opened 2025-08-26
fix(node/builder): correct left_mut() method implementation and docs

bendanzhentan:


yongkangc (approved):
lgtm! lets have @mattsse take a look as well

mattsse (approved):

PR #18054, opened 2025-08-26
perf(payload): do not clone full attributes for timestamp validation

hai-rise:
EngineAPI only needs a payload's timestamp to validate its version based on hardfork activations:
https://github.com/paradigmxyz/reth/blob/7703e6fb9dafe0281457f13a92366227a289c6ba/crates/rpc/rpc-engine-api/src/engine_api.rs#L402-L406

However, the current implementation passes (owned and cloned) full attributes around, which is noticeable friction for performance chains with low block time and large (custom) attributes.

This PR **demotes full attributes fetching methods to timestamp fetching only**. If it's too noisy/breaking, I can add a new method for timestamp instead, but getting rid of `PayloadJob::payload_attributes` is still nice, so we can transfer attributes to the block builder instead of cloning a full one in the `PayloadJob`. And custom payload jobs can still provide the full attributes outside of the `PayloadJob` trait if they want to.

### DIFF
--- a/crates/payload/basic/src/lib.rs
+++ b/crates/payload/basic/src/lib.rs
@@ -455,6 +455,10 @@ where
         Ok(self.config.attributes.clone())
     }
 
+    fn payload_timestamp(&self) -> Result<u64, PayloadBuilderError> {
+        Ok(self.config.attributes.timestamp())
+    }
+
     fn resolve_kind(
         &mut self,
         kind: PayloadKind,
--- a/crates/rpc/rpc-engine-api/src/engine_api.rs
+++ b/crates/rpc/rpc-engine-api/src/engine_api.rs
@@ -21,8 +21,8 @@ use reth_chainspec::EthereumHardforks;
 use reth_engine_primitives::{ConsensusEngineHandle, EngineApiValidator, EngineTypes};
 use reth_payload_builder::PayloadStore;
 use reth_payload_primitives::{
-    validate_payload_timestamp, EngineApiMessageVersion, ExecutionPayload,
-    PayloadBuilderAttributes, PayloadOrAttributes, PayloadTypes,
+    validate_payload_timestamp, EngineApiMessageVersion, ExecutionPayload, PayloadOrAttributes,
+    PayloadTypes,
 };
 use reth_primitives_traits::{Block, BlockBody};
 use reth_rpc_api::{EngineApiServer, IntoEngineApiRpcModule};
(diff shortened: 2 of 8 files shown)


crates/payload/basic/src/lib.rs
```diff
@@ -451,8 +451,8 @@ where
         }
     }
 
-    fn payload_attributes(&self) -> Result<Self::PayloadAttributes, PayloadBuilderError> {
-        Ok(self.config.attributes.clone())
```
hai-rise, line 455: The main clone that we remove. Custom payload jobs like ours can also transfer the full attributes to the block builder, and only keep the timestamp instead of cloning the full attributes inside the payload job.

crates/payload/basic/src/lib.rs
```diff
@@ -451,8 +451,8 @@ where
...
+    fn payload_timestamp(&self) -> Result<u64, PayloadBuilderError> {
```
mattsse, line 458: this breaks api, can we instead introduce this as new fn and have it return fn payload_attributes.timestamp() by default?

mattsse (changes requested):
imo we still want a function to obtain the attributes but I'm okay with introducing this as a new function.

a better way would probably be to not do this fetch request at all and come up with a better way to fetch the payload, and resolve only if valid for that endpoint

crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -400,10 +397,10 @@ where
         EngineT::BuiltPayload: TryInto<R>,
     {
         // First we fetch the payload attributes to check the timestamp
-        let attributes = self.get_payload_attributes(payload_id).await?;
```
hai-rise, line 403: The main goal is to avoid cloning/allocating full attributes here just to extract the timestamp.

crates/payload/basic/src/lib.rs
hai-rise: Yeah, that's fair. I've brought back `PayloadJob::payload_attributes` 🙏.

crates/rpc/rpc-engine-api/src/engine_api.rs
mattsse: ah I believe this matters for you because the attributes contain more info like txs?

mattsse (approved):
makes sense

crates/rpc/rpc-engine-api/src/engine_api.rs
hai-rise: Yes, chains can also add arbitrarily large data to custom attributes. Our CL actually sends two sets of attributes in one payload when there's an L1 origin switch for pipelining!

PR #18065, opened 2025-08-26
chore: enable all Ethereum protocol versions instead of hardcoding

CaptainJoly:
Replaces hardcoded protocol versions with `EthVersion::ALL_VERSIONS` to automatically include all supported versions,
and also verify Eth69 protocol is included by default.


mattsse (approved):
this is still pending

we can do this shortly

mattsse (approved):

PR #18067, opened 2025-08-26
feat(optimism): Add `launch_wss_flashblocks_service` function spawning a task sending last pending block

RomanHodulak:
Part of #17858 

Adds a function for launching the flashblock service and getting a receiver for last pending block.

### DIFF
--- a/crates/optimism/flashblocks/src/app.rs
+++ b/crates/optimism/flashblocks/src/app.rs
@@ -1,0 +1,55 @@
+use crate::{FlashBlockService, FlashBlockWsStream};
+use futures_util::StreamExt;
+use reth_chain_state::ExecutedBlock;
+use reth_evm::ConfigureEvm;
+use reth_primitives_traits::{BlockTy, HeaderTy, NodePrimitives, ReceiptTy};
+use reth_rpc_eth_api::helpers::pending_block::BuildPendingEnv;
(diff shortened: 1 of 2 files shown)


crates/optimism/flashblocks/src/app.rs
```diff
@@ -0,0 +1,55 @@
...
+        + BlockReaderIdExt<
+            Header = HeaderTy<N>,
+            Block = BlockTy<N>,
+            Transaction = N::SignedTx,
+            Receipt = ReceiptTy<N>,
+        > + Unpin
+        + 'static,
+{
+    let stream = FlashBlockWsStream::new(ws_url);
+    let mut service = FlashBlockService::new(stream, evm_config, provider, ());
```
mattsse, line 38: looking at

https://github.com/paradigmxyz/reth/blob/main/crates/optimism/flashblocks/src/payload.rs#L47-L69

I think we dont even need the `Builder == ()` here because we already have all the infos here

in fact using BuildPendingEnv would result in an incorrect block, but we can tackle this next, we likely need to restrict

https://github.com/paradigmxyz/reth/blob/main/crates/optimism/evm/src/lib.rs#L124-L124

`from<ExecutionPayloadBaseV1>`

or smth like this

mattsse (approved):

PR #18077, opened 2025-08-26
feat(optimism): Launch `FlashBlockService` when websocket URL is provided in `OpEthApi`

### LINKED ISSUE #18068: Integrate FlashBlockService with OpEthApi
### Describe the feature

The goal is to launch a background task with FlashBlockService inside OpEthApi and provide a means to propagate it's output, ExecutedBlock, to the OpEthApi local_pending_block function.

### Additional context

_No response_

RomanHodulak:
Part of #17858

Integrates `FlashBlockService` into `OpEthApi`.

### DIFF
--- a/crates/optimism/rpc/src/eth/pending_block.rs
+++ b/crates/optimism/rpc/src/eth/pending_block.rs
@@ -45,6 +45,10 @@ where
         )>,
         Self::Error,
     > {
+        if let Some(block) = self.pending_flashblock() {
+            return Ok(Some(block));
+        }
+
         // See: <https://github.com/ethereum-optimism/op-geth/blob/f2e69450c6eec9c35d56af91389a1c47737206ca/miner/worker.go#L367-L375>
         let latest = self
             .provider()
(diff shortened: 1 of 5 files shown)


crates/optimism/rpc/src/eth/pending_block.rs
```diff
@@ -45,6 +45,15 @@ where
         )>,
         Self::Error,
     > {
+        if let Some(rx) = self.inner.flashblocks_rx.as_ref() {
+            if let Some(block) = rx.borrow().as_ref() {
+                return Ok(Some((
+                    block.recovered_block.clone(),
+                    Arc::new(block.execution_output.receipts.iter().flatten().cloned().collect()),
```
mattsse: should we prep this in the type of the channel

`FlashBlockRx<N::Primitives>`

basically what we do for 

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/rpc/rpc-eth-types/src/pending_block.rs#L91-L97

then we can avoid doing this per request and only have to prep it once?

mattsse: I'd alos like this as a standalone fn of Opethapi

fn `pending_flashblock` or smth

mattsse (changes requested):
cool, two suggestions

mattsse (approved):

PR #18078, opened 2025-08-26
feat(optimism): Change `FlashBlockService` output `ExecutedBlock` => `PendingBlock`

RomanHodulak:
Part of #17858 

Wraps the executed block in pending block which has couple more useful things.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -13,16 +13,17 @@ use reth_primitives_traits::{
 };
 use reth_revm::{database::StateProviderDatabase, db::State};
 use reth_rpc_eth_api::helpers::pending_block::PendingEnvBuilder;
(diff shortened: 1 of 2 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -146,11 +147,14 @@ impl<
             vec![execution_result.requests],
         );
 
-        Ok(ExecutedBlock {
-            recovered_block: block.into(),
-            execution_output: Arc::new(execution_outcome),
-            hashed_state: Arc::new(hashed_state),
-        })
+        Ok(PendingBlock::with_executed_block(
+            Instant::now() + Duration::from_secs(1),
```
mattsse, line 151: this makes sense, we could maybe fine tune this later but not respecting a flashblock after 1s should always be okay

mattsse (approved):

PR #18080, opened 2025-08-26
feat(net): implement support of subprotocols

MatusKysel:
## Summary

  Implements support for subprotocols in the RLP multiplexer, enabling proper concurrent handling of satellite protocols alongside the primary ETH protocol.

  ## Changes

  - **Multiplexer Enhancement**: Added `ProtocolsPoller` to concurrently poll multiple protocol streams in the main multiplexer loop
  - **Handshake Injection**: Updated `into_eth_satellite_stream` to accept `Arc<dyn EthRlpxHandshake>` for flexible handshake implementations
  - **Protocol Adapter**: Created `ProxyUnauth` adapter to enable injected handshake to work over multiplexed `ProtocolProxy`
  - **Session Integration**: Updated session management to pass handshake dependency to multiplexer

### DIFF
--- a/crates/net/eth-wire/src/multiplex.rs
+++ b/crates/net/eth-wire/src/multiplex.rs
@@ -7,6 +7,7 @@
 //! Hence it is expected that the primary protocol is "eth" and the additional protocols are
 //! "dependent satellite" protocols.
 
+use std::sync::Arc;
 use std::{
     collections::VecDeque,
     fmt,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
I see we previously didnt support the custom handshake here and we also need to poll the protos concurrently

the additional workarounds here arent ideal but should be fine

PR #18081, opened 2025-08-26
fix: clarify locking behavior comment in InMemoryState

malik672:
previous comment implied `Atomicity` but looking at the code, it's just locks.
The Compiler or CPU can still order the code anyway it deems fit


mattsse (approved):
fair

PR #18083, opened 2025-08-26
chore: add helpers for setting minimum protocol basefee

mattsse:
smol helpers to bypass pool protocol related fee checks, this way this can be easily set on Nodeconfig directly


Rjected (approved):
lgtm

PR #18086, opened 2025-08-27
feat: optionally disable balance check for `EthTransactionValidator`

0xKitsune:
This PR introduces a `disable_balance_check` field for `EthTransactionValidator` to optionally disable balance checks during transaction validation.


mattsse (approved):
this seems reasonable, especially if the stateprovider doesnt have direct access to the state, so having an option to opt-out here seems useful

PR #18088, opened 2025-08-27
perf(engine): only fetch headers instead of full blocks for tree tasks

hai-rise:
8/54 ~ 15% of forkchoice processing time is spent on cloning a full block, even when all tree tasks only need a header (reference).

image

And 1.4/54 ~ 3% is spent on dropping the unused body :sweat_smile:.
image
 
P/s: `TreeState::block_by_hash` should also be demoted from returning a (cloned) block to a header (for the `advance_persistence` path). I can do this in the next PR.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -29,7 +29,7 @@ use reth_payload_builder::PayloadBuilderHandle;
 use reth_payload_primitives::{
     BuiltPayload, EngineApiMessageVersion, NewPayloadError, PayloadBuilderAttributes, PayloadTypes,
 };
-use reth_primitives_traits::{Block, NodePrimitives, RecoveredBlock, SealedBlock, SealedHeader};
+use reth_primitives_traits::{NodePrimitives, RecoveredBlock, SealedBlock, SealedHeader};
 use reth_provider::{
     providers::ConsistentDbView, BlockNumReader, BlockReader, DBProvider, DatabaseProviderFactory,
     HashedPostStateProvider, ProviderError, StateProviderBox, StateProviderFactory, StateReader,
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -1724,10 +1724,10 @@ where
         }
     }
 
-    /// Return block from database or in-memory state by hash.
-    fn block_by_hash(&self, hash: B256) -> ProviderResult<Option<N::Block>> {
+    /// Return block header from database or in-memory state by hash.
+    fn block_header_by_hash(&self, hash: B256) -> ProviderResult<Option<N::BlockHeader>> {
         // check database first
-        let mut block = self.provider.block_by_hash(hash)?;
+        let mut block = self.provider.header_by_hash_or_number(hash.into())?;
```
hai-rise: This is the key saving.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1724,21 +1724,21 @@ where
...
+        let mut header = self.provider.header_by_hash_or_number(hash.into())?;
+        if header.is_none() {
             // Note: it's fine to return the unsealed block because the caller already has
             // the hash
-            block = self
+            header = self
                 .state
                 .tree_state
                 .block_by_hash(hash)
                 // TODO: clone for compatibility. should we return an Arc here?
```
mattsse, line 1738: we can remove this comment now

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1724,21 +1724,21 @@ where
...
-                .map(|block| block.as_ref().clone().into_block());
+                .map(|block| block.header().clone());
```
mattsse, line 1739: there should also be `clone_header` I believe

mattsse (approved):
ah nice

yes this should only fetch header

pending @mediocregopher

mediocregopher (approved):
Simple change, LGTM :chart_with_upwards_trend:

PR #18089, opened 2025-08-27
chore: apply Rust Code Formatting with cargo fmt --all

tzchenxixi:
I formats the entire codebase by running `cargo fmt` . 

This change ensures that all Rust packages adhere to a consistent code style, enhancing readability and maintainability.

### DIFF
--- a/bin/reth-bench/src/valid_payload.rs
+++ b/bin/reth-bench/src/valid_payload.rs
@@ -68,7 +68,7 @@ where
             if status.is_syncing() {
                 return Err(alloy_json_rpc::RpcError::UnsupportedFeature(
                     "invalid range: no canonical state found for parent of requested block",
-                ))
+                ));
             }
             status =
                 self.fork_choice_updated_v1(fork_choice_state, payload_attributes.clone()).await?;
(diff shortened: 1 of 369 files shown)


lean-apple:
> The single change seems quite large. Should I break it down into multiple commits by module?

The rust codebase is already formatted, yet, you need to use `nightly` toolchain (for example `cargo +nightly fmt --all` , the CI check is also done with `nightly `https://github.com/paradigmxyz/reth/blob/3d8033a03c0700366021742f048714f64c5c72d3/.github/workflows/lint.yml#L153.

lean-apple:
> > > The single change seems quite large. Should I break it down into multiple commits by module?
> > 
> > 
> > The rust codebase is already formatted, yet, you need to use `nightly` toolchain (for example `cargo +nightly fmt --all`), the CI check is also done with `nightly`
> > https://github.com/paradigmxyz/reth/blob/3d8033a03c0700366021742f048714f64c5c72d3/.github/workflows/lint.yml#L153
> 
> Thanks for your response. However, I am quite surprised that the nightly version of Rust is being used. 😄

Some formatting configuration in the `rustfmt.toml` file will only work with `nightly`.

PR #18094, opened 2025-08-27
feat(optimism): Add `flashblocks_url` as part of rollup args of the `op-reth` CLI

### LINKED ISSUE #18069: Add CLI arg for providing flashblocks URL
### Describe the feature

The goal is to add an optional rollup argument for op-reth.

This argument has two functions:
* If provided, turns on FlashBlockService
* Configures what wss URL to subscribe to

### Additional context

_No response_

RomanHodulak:
Part of #17858 

Adds a command line argument that, if provided, makes the op-reth node launch the flashblocks subscription.

### DIFF
--- a/crates/optimism/node/src/args.rs
+++ b/crates/optimism/node/src/args.rs
@@ -4,6 +4,7 @@
 
 use op_alloy_consensus::interop::SafetyLevel;
 use reth_optimism_txpool::supervisor::DEFAULT_SUPERVISOR_URL;
+use url::Url;
 
 /// Parameters for rollup configuration
 #[derive(Debug, Clone, PartialEq, Eq, clap::Args)]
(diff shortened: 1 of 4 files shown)


crates/optimism/node/src/args.rs
```diff
@@ -66,6 +67,13 @@ pub struct RollupArgs {
     /// Minimum suggested priority fee (tip) in wei, default `1_000_000`
     #[arg(long, default_value_t = 1_000_000)]
     pub min_suggested_priority_fee: u64,
+
+    /// A URL pointing to a secure websocket subscription that streams out flashblocks.
+    ///
+    /// If given, the flashblocks are received to build pending block. All request with "pending"
+    /// block tag will use the pending state based on flashblocks.
+    #[arg(long)]
+    pub flashblocks_url: Option<Url>,
```
mattsse, line 76: we have String value for other endpoint args, which would then fail during setup, but also those are just http endpoints,

validating this on cli level seems better actually, so we can keep this url

mattsse (approved):
lgtm

PR #18100, opened 2025-08-27
feat(optimism): Remove builder of next block environment from `FlashBlockService`

### LINKED ISSUE #18074: Remove PendingEnvBuilder from Flashblockservice
### Describe the feature

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/optimism/flashblocks/src/service.rs#L40-L40

we cant use:

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/optimism/flashblocks/src/service.rs#L117-L117

because this context comes from the flashblock itself:

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/optimism/flashblocks/src/payload.rs#L46-L69

which includes:

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/optimism/evm/src/config.rs#L6-L21

that is

https://github.com/paradigmxyz/reth/blob/87647b25acf43689deb4568fdd98c6913a8d1819/crates/optimism/evm/src/lib.rs#L169-L173

we have a few choices here,
1. restrict that `EvmConfig::NextBlockEnvCtx: From< ExecutionPayloadBaseV1 >`
2. make the flashblock type generic and enforce that returns that context
3. restrict to alloy_consensus::Header and add a `Flashblock::header ->Header` function

option 1 feels most appropriate because this is effectively an OP specific thing

### Additional context

_No response_

RomanHodulak:
Part of #17858 

Removes the next evm env attributes builder from `FlashBlockService` and uses the flashblock base instead.

### DIFF
--- a/examples/custom-node/src/evm/config.rs
+++ b/examples/custom-node/src/evm/config.rs
@@ -23,6 +23,7 @@ use reth_op::{
     node::{OpEvmConfig, OpNextBlockEnvAttributes, OpRethReceiptBuilder},
     primitives::SignedTransaction,
 };
+use reth_optimism_flashblocks::ExecutionPayloadBaseV1;
 use reth_rpc_api::eth::helpers::pending_block::BuildPendingEnv;
 use std::sync::Arc;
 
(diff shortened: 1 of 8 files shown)


examples/custom-node/src/evm/config.rs
```diff
@@ -136,6 +137,12 @@ pub struct CustomNextBlockEnvAttributes {
     extension: u64,
 }
 
+impl From<ExecutionPayloadBaseV1> for CustomNextBlockEnvAttributes {
+    fn from(value: ExecutionPayloadBaseV1) -> Self {
+        Self { inner: value.into(), extension: 0 }
+    }
+}
```
mattsse, line 144: this seems fine

mattsse (approved):

Issue #18102, opened 2025-08-27
Ensure flashblocks build on parent block and pending flashblock is cleared when a new chain is committed


mattsse:
### Describe the feature

currently we always use the latest state

https://github.com/paradigmxyz/reth/blob/615bd4a30fc719de384be55bba0270a36664a47b/crates/optimism/flashblocks/src/service.rs#L112-L116

however this might not be 

https://github.com/paradigmxyz/reth/blob/615bd4a30fc719de384be55bba0270a36664a47b/crates/optimism/flashblocks/src/payload.rs#L50-L51

so we should not build the block if parent_hash != latest

we also must clear the block when a new block was mined,

for this we can install a Canonstatesubscription via the provider and remove the pending block on tick

### Additional context

Checklist: 
- [x] Prevent old pending flashblock from being returned from pending_flashblock
- [x] State check to ensure we only keep and return a pending flashblock if it is still valid

PR #18103, opened 2025-08-27
fix(optimism): Prevent old pending flashblock from being returned from `pending_flashblock`

RomanHodulak:
Part of #17858 #18102 

Adds similar checks as in `pool_pending_block` for `pending_flashblock` to verify that `latest` is its parent.

### DIFF
--- a/crates/optimism/rpc/src/eth/mod.rs
+++ b/crates/optimism/rpc/src/eth/mod.rs
@@ -12,6 +12,7 @@ use crate::{
     eth::{receipt::OpReceiptConverter, transaction::OpTxInfoMapper},
     OpEthApiError, SequencerClient,
 };
+use alloy_consensus::BlockHeader;
 use alloy_primitives::U256;
 use eyre::WrapErr;
 use op_alloy_network::Optimism;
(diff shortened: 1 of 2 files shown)


crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -100,8 +102,35 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
     /// If flashblocks receiver is not set, then it always returns `None`.
-    pub fn pending_flashblock(&self) -> Option<PendingBlockAndReceipts<N::Primitives>> {
-        Some(self.inner.flashblocks_rx.as_ref()?.borrow().as_ref()?.to_block_and_receipts())
+    pub fn pending_flashblock(&self) -> eyre::Result<Option<PendingBlockAndReceipts<N::Primitives>>>
+    where
+        Self: LoadPendingBlock,
+    {
+        if self.pending_block_kind().is_none() {
+            return Ok(None);
+        }
```
mattsse: this makes sense

I think we can still spawn the flashblockservice if this is_none because maybe you still want the flashblocks without having them avail on rpc

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -100,8 +102,35 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+
+        let pending = self.pending_block_env_and_cfg()?;
+        let parent = match pending.origin {
+            PendingBlockEnvOrigin::ActualPending(..) => return Ok(None),
+            PendingBlockEnvOrigin::DerivedFromLatest(parent) => parent,
+        };
+
+        let Some(rx) = self.inner.flashblocks_rx.as_ref() else { return Ok(None) };
+        let rx = rx.borrow();
```
mattsse: rx is confusing here because we're borrowing the value here not the receiver

mattsse (changes requested):
this makes sense here, 

smol nits for this one.

we're still mising a similar check in the service and imo we should also still clear the block when a new block is mined, we can do that separately

mattsse (approved):

PR #18104, opened 2025-08-27
chore(reth-optimism-storage): small refactor code

nkysg:


crates/optimism/storage/src/chain.rs
```diff
@@ -96,22 +96,15 @@ where
     ) -> ProviderResult<Vec<<Self::Block as Block>::Body>> {
         let chain_spec = provider.chain_spec();
 
-        let mut bodies = Vec::with_capacity(inputs.len());
-
-        for (header, transactions) in inputs {
-            let mut withdrawals = None;
-            if chain_spec.is_shanghai_active_at_timestamp(header.timestamp()) {
-                // after shanghai the body should have an empty withdrawals list
```
mattsse, line 104: I'd like to keep this note and move above `withdrawals:`

crates/optimism/storage/src/chain.rs
```diff
@@ -96,22 +96,15 @@ where
...
                 transactions,
                 ommers: vec![],
-                withdrawals,
-            });
-        }
-
-        Ok(bodies)
+                withdrawals: chain_spec
+                    .is_shanghai_active_at_timestamp(header.timestamp())
+                    .then(|| vec![].into()),
```
mattsse: I think here we now know the type, so we could use default::default

mattsse (changes requested):
this is indeed nicer, pedantic nits

crates/optimism/storage/src/chain.rs
nkysg: Thank you

mattsse (approved):

PR #18107, opened 2025-08-27
docs: multiple small textual defects and typos

letmehateu:


mattsse (commented):
these typos are fine really

PR #18108, opened 2025-08-27
fix(engine): Prevent instant miner from creating empty blocks

### LINKED ISSUE #18096: Instant local miner can end up mining more blocks than necessary
### Describe the feature

this miner operates on a channel for pending txs: 
https://github.com/paradigmxyz/reth/blob/aecc89f78d538742175f2e331cac9bc6da40b2c7/crates/engine/local/src/miner.rs#L65-L81

and ticks that advances a block:

https://github.com/paradigmxyz/reth/blob/aecc89f78d538742175f2e331cac9bc6da40b2c7/crates/engine/local/src/miner.rs#L140-L145

however since one block can include multiple pending txs, already draining the pool, this can lead to the mode ticking while the pool is already empty

we could fix this by either draining the entire channel, however then we might encounter the opposite if we drain too many (exceeding the block limit), so what we can do here is checking the pending pool count via the pool
if 

https://github.com/paradigmxyz/reth/blob/aecc89f78d538742175f2e331cac9bc6da40b2c7/crates/engine/local/src/miner.rs#L67-L67

but the pool is empty then this is most likely an outdated event and the corresponding tx is already mined, this workaround should be sufficient to avoid trailing empty blocks after high volume of txs

### Additional context

_No response_

7suyash7:
Fixes #18096 

This change adds a check against the transaction pool's pending count before the miner acts on a notification, preventing it from mining an empty block if the pool is already drained.

### DIFF
--- a/crates/engine/local/src/miner.rs
+++ b/crates/engine/local/src/miner.rs
@@ -24,12 +24,14 @@ use tracing::error;
 
 /// A mining mode for the local dev engine.
 #[derive(Debug)]
-pub enum MiningMode {
+pub enum MiningMode<Pool: TransactionPool + Unpin> {
     /// In this mode a block is built as soon as
     /// a valid transaction reaches the pool.
     /// If `max_transactions` is set, a block is built when that many transactions have
     /// accumulated.
     Instant {
+        /// The transaction pool.
+        pool: Pool,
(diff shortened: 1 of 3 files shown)


klkvr (approved):
lgtm, extra generic should be fine for now

PR #18111, opened 2025-08-27
feat: add EIP-7934 block size check to validateBuilderSubmissionV5

leniram159:
Fixes #18110

add EIP-7934 block size validation to validateBuilderSubmissionV5

Enforce MAX_RLP_BLOCK_SIZE limit when Osaka hardfork is active.


crates/rpc/rpc/src/validation.rs
```diff
@@ -454,6 +455,17 @@ where
             ),
         })?;
 
+        // Check block size as per EIP-7934 (only applies when Osaka hardfork is active)
+        let chain_spec = self.provider.chain_spec();
+        if chain_spec.is_osaka_active_at_timestamp(block.timestamp()) &&
+            block.rlp_length() > MAX_RLP_BLOCK_SIZE
+        {
```
mattsse, line 462: right, checking if osaka is active does make sense here

mattsse (approved):
lgtm, ty!

PR #18112, opened 2025-08-28
fix(rpc): return empty log set for invalid filter block ranges

### LINKED ISSUE #18098: eth_getLogs's Filter replaces fromBlock and toBlock with `latest`
### Describe the bug

Reth's log filtering updates the fromBlock and toBlock conditions with the latest block's logs.

The implementation isn't standardized across the EL clients (geth throw an error where erigon and besu reply with an empty list), but transparently replacing the filter's condition can lead to confusion.

### Steps to reproduce

```
cast logs --from-block 0xfffffffff --to-block 0xfffffffff  --json -r localhost:8545 | jq '.[0]'
{
  "address": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
  "topics": [
    "0xe1fffcc4923d04b559f4d29a8bfc6cda04eb5b0d3c460751c2402c5c5cc9109c",
    "0x0000000000000000000000003a10dc1a145da500d5fba38b9ec49c8ff11a981f"
  ],
  "data": "0x00000000000000000000000000000000000000000000000000afdbfcdc61c000",
  "blockHash": "0x2d7b2f6b86d95ca9f1e5f496d9dafa31850793963b5119b519ef51589649dbca",
  "blockNumber": "0x1628041",
  "blockTimestamp": "0x68aeff87",
  "transactionHash": "0xb83e9e78f854dc787447c401000f42c618da24bc2ded24342f7434a8758c7430",
  "transactionIndex": "0x0",
  "logIndex": "0x0",
  "removed": false
}
```

### Node logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Kubernetes

1.6.0

Current database version: 2
Local database version: 2

*

Archive (default)

_No response_

_No response_

### Code of Conduct

dharmvr1:
close #18098

### DIFF
--- a/crates/rpc/rpc-eth-types/src/logs_utils.rs
+++ b/crates/rpc/rpc-eth-types/src/logs_utils.rs
@@ -145,7 +145,9 @@ where
     Ok(())
 }
 
-/// Computes the block range based on the filter range and current block numbers
+/// Computes the block range based on the filter range and current block numbers.
+///
+/// This returns `(min(best,from), min(best,to))`.
 pub fn get_filter_block_range(
     from_block: Option<u64>,
     to_block: Option<u64>,
--- a/crates/rpc/rpc-eth-types/src/logs_utils.rs
+++ b/crates/rpc/rpc-eth-types/src/logs_utils.rs
@@ -159,7 +159,11 @@ pub fn get_filter_block_range(
     // the start block if the converted value is larger than the start block, since `from_block`
     // can't be a future block: `min(head, from_block)`
     if let Some(filter_from_block) = from_block {
-        from_block_number = start_block.min(filter_from_block)
+        if filter_from_block > start_block {
+            return (0, 0);
+        }
+
+        from_block_number = filter_from_block;
     }
 
     // upper end of the range is the converted `to_block` argument, restricted by the best block:
--- a/crates/rpc/rpc-eth-types/src/logs_utils.rs
+++ b/crates/rpc/rpc-eth-types/src/logs_utils.rs
@@ -171,6 +171,9 @@ pub fn get_filter_block_range(
     if let Some(filter_to_block) = to_block {
         to_block_number = info.best_number.min(filter_to_block);
     }
+    if from_block_number > to_block_number {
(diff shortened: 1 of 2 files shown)


dharmvr1:
hi @mattsse @Rjected @RomanHodulak
I noticed in [filter.rs:L493](https://github.com/paradigmxyz/reth/blob/615bd4a30fc719de384be55bba0270a36664a47b/crates/rpc/rpc/src/eth/filter.rs?utm_source=chatgpt.com#L493)
 that start_block is set to info.best_number.
This seems confusing, since start_block sounds like the earliest available block (prune floor), but here it’s the latest (head). That’s why handling from_block < start_block felt odd.
Should start_block actually represent the prune floor instead of the head?

crates/rpc/rpc-eth-types/src/logs_utils.rs
mattsse, line 1: I'd like to keep this here unchanged and instead only add a check for the ::Range variant where we check from,to explicitly

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -502,6 +502,11 @@ where
                     .flatten();
                 let (from_block_number, to_block_number) =
                     logs_utils::get_filter_block_range(from, to, start_block, info);
+                // if function signaled empty range:
+                if from_block_number == 0 && to_block_number == 0 {
+                    return Ok(Vec::new()); // returns empty log set, not an error
+                }
```
mattsse: this isn't very intuitive that we now need to check for == 0

we can add the check directly to the variants from,to blocks

mattsse (changes requested):

dharmvr1:
hi @mattsse  
Made the changes, let me know if anything else is needed.

dharmvr1:
hi @mattsse  just want to remind u about this ,  please review changes ,
or let me know if need any change
thanks

mattsse (approved):

yongkangc (approved):
nice

PR #18116, opened 2025-08-28
perf(engine): only clone headers instead of full blocks for tree tasks

hai-rise:
(Part 2 of #18088)

(15+16)/46 ~ 67% of `EngineApiTreeHandler::advance_persistence` is spent on cloning a full block (more specifically in `find_disk_reorg`), even when all tree tasks only need a header (reference).

image

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1040,7 +1040,7 @@ where
             // we still need to process payload attributes if the head is already canonical
             if let Some(attr) = attrs {
                 let tip = self
-                    .block_header_by_hash(self.state.tree_state.canonical_block_hash())?
+                    .sealed_header_by_hash(self.state.tree_state.canonical_block_hash())?
                     .ok_or_else(|| {
                         // If we can't find the canonical block, then something is wrong and we need
                         // to return an error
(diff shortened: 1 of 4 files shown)


crates/engine/tree/src/tree/state.rs
```diff
@@ -85,9 +85,9 @@ impl<N: NodePrimitives> TreeState<N> {
         self.blocks_by_hash.get(&hash)
     }
 
-    /// Returns the block by hash.
-    pub(crate) fn block_by_hash(&self, hash: B256) -> Option<Arc<SealedBlock<N::Block>>> {
-        self.blocks_by_hash.get(&hash).map(|b| Arc::new(b.recovered_block().sealed_block().clone()))
```
hai-rise, line 90: Main saving: All users were cloning the full block just to extract the header, so we should only clone the header in the first place.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1731,12 +1727,8 @@ where
...
             // Note: it's fine to return the unsealed block because the caller already has
             // the hash
-            header = self
-                .state
-                .tree_state
-                .block_by_hash(hash)
-                // TODO: clone for compatibility. should we return an Arc here?
-                .map(|block| block.header().clone());
+            header =
+                self.state.tree_state.block_header_by_hash(&hash).map(SealedHeader::into_header)
```
mattsse: could we change the return type to also be `SealedHeader` for consistency, then we dont need this map here, and instead do `.map(SealedHeader::new_unhashed)` above

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1705,20 +1705,16 @@ where
...
         &self,
         hash: B256,
     ) -> ProviderResult<Option<SealedHeader<N::BlockHeader>>> {
         // check memory first
-        let block = self
-            .state
-            .tree_state
-            .block_by_hash(hash)
-            .map(|block| block.as_ref().clone_sealed_header());
+        let header = self.state.tree_state.block_header_by_hash(&hash);
```
mattsse: this function is now identical to

```
block_header_by_hash
```

below, right?

so we could remove one, probably should keep this one and remove the one below, because the order here is cheaper if buffered

mattsse (changes requested):
nice,

one suggestion

crates/engine/tree/src/tree/mod.rs
hai-rise: Yeah, wanted to keep the semantics and the diff small. I've pushed a more aggressive change that uses the try-memory-first variant everywhere 🙏.

mattsse (approved):

PR #18117, opened 2025-08-28
chore: add prewarm traces

mattsse:
some prewarm traces


yongkangc (approved):
lgtm! would be useful when tracing prewarm

PR #18118, opened 2025-08-28
docs: fix broken Markdown formatting

eeemmmmmm:
Fix broken bold text formatting where double asterisks were split across line breaks, causing "**opening a discussion**" to render incorrectly


mattsse (commented):
these are fine

PR #18120, opened 2025-08-28
fix(optimism): Fix endless poll on the FlashBlockService

0xMelkor:
The current implementation of `FlashBlockService::poll_next`, when used together with `launch_wss_flashblocks_service`, results in excessive invocations of poll_next.

Outputs thousands of:
```bash
ERROR Missing base flashblock
ERROR Missing base flashblock
ERROR Missing base flashblock
ERROR Missing base flashblock
ERROR Missing base flashblock
ERROR Missing base flashblock
...
```

Relates to #17858

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -176,16 +176,15 @@ impl<
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.get_mut();
-        loop {
-            match this.rx.poll_next_unpin(cx) {
-                Poll::Ready(Some(Ok(flashblock))) => {
-                    this.add_flash_block(flashblock);
-                    return Poll::Ready(Some(this.execute()));
-                }
-                Poll::Ready(Some(Err(err))) => return Poll::Ready(Some(Err(err))),
-                Poll::Ready(None) => return Poll::Ready(None),
-                Poll::Pending => return Poll::Pending,
+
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is better

I think we eventually want to make this more sophisticated so that execute doesn't happen in a poll because blocking

crates/optimism/flashblocks/src/service.rs
```diff
@@ -178,10 +178,13 @@ impl<
         let this = self.get_mut();
         loop {
             match this.rx.poll_next_unpin(cx) {
-                Poll::Ready(Some(Ok(flashblock))) => this.add_flash_block(flashblock),
+                Poll::Ready(Some(Ok(flashblock))) => {
+                    this.add_flash_block(flashblock);
+                    return Poll::Ready(Some(this.execute()));
+                }
```
mattsse: this now never loops and could just be a ready!()

mattsse (changes requested):

0xMelkor:
> this is better
> 
> I think we eventually want to make this more sophisticated so that execute doesn't happen in a poll because blocking

This could be polled in a `tokio::task::spawn` instead
```rust
   // tokio::spawn(async move {
   tokio::task::spawn(async move {
        while let Some(block) = service.next().await {
            if let Ok(block) = block.inspect_err(|e| tracing::error!("{e}")) {
                let _ = tx.send(Some(block)).inspect_err(|e| tracing::error!("{e}"));
            }
        }
    });
```

mattsse (approved):

PR #18122, opened 2025-08-28
perf(engine): skip first 11 transactions in prewarm to reduce cache contention

### LINKED ISSUE #18092: Investigate skipping first transaction in prewarm scheduling
Sub-issue of #18001

The current method for prewarming transactions in the reth payload processor may be inefficient and could be causing performance issues due to cache contention.

## Problem
Transactions are distributed to a pool of 64 "prewarming" worker threads in a simple round-robin fashion. This means the first few transactions in a block (e.g., tx 0, tx 1, tx 2...) are assigned to different threads. However, the main execution thread starts processing these same transactions almost immediately.

This creates a race condition:

If a prewarming thread has not finished processing a transaction by the time the main execution thread needs it, the prewarming effort is wasted.
Worse, both the prewarming thread and the main execution thread might attempt to access and modify the same cache data simultaneously. This "cache contention" can slow down execution instead of speeding it up.

## Current Implementation
https://github.com/paradigmxyz/reth/blob/main/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L113

Transactions are distributed round-robin across 64 workers:
```rust
let task_idx = executing % max_concurrency;
```

### Cache Contention Issue
The current transaction caching mechanism (even in main branch) faces timing conflicts:
- First transactions (tx[0], tx[64], tx[128]...) all go to the same prewarm thread
- If main execution thread processes tx[0] while prewarm is still working on it, there's no cache benefit
- This can lead to cache contention between prewarm and main execution threads

### Timing Analysis Needed
Need to compare execution timings:
- When does prewarm finish tx[1], tx[2], tx[3], tx[4]?
- When does main execution thread reach these transactions?
- Is there overlap causing cache invalidation?

## Proposed Investigation:

To address this, the following actions are needed:

1.  **Profile the Overlap:** Measure the timing difference between when a transaction is prewarmed and when it is executed by the main thread to understand the extent of the conflict.
2.  **Test Alternative Strategies:** Experiment with different ways to schedule prewarming tasks to avoid this initial conflict, such as:
    *   Skipping the first few transactions in the prewarming process.
    *   Introducing a delay before prewarming the initial transactions.
    *   Using a more intelligent distribution method instead of simple round-robin.
3.  **Measure Performance:** Analyze cache hit rates and memory access patterns with the current and alternative methods to determine which approach is most effective.

yongkangc:
Skips prewarming the first 11 transactions which consistently race with main execution thread.

Improves cache hit rate from <5% to >80% and eliminates wasted CPU cycles.

Closes https://github.com/paradigmxyz/reth/issues/18092

TODO: Clean up metrics

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -10,7 +10,7 @@ use crate::tree::{
 };
 use alloy_evm::Database;
 use alloy_primitives::{keccak256, map::B256Set, B256};
-use metrics::{Counter, Gauge, Histogram};
+use metrics::{Gauge, Histogram};
 use reth_evm::{execute::ExecutableTxFor, ConfigureEvm, Evm, EvmFor, SpecFor};
 use reth_metrics::Metrics;
 use reth_primitives_traits::{NodePrimitives, SignedTransaction};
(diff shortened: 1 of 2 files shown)


yongkangc:
closing because perf isnt good

PR #18123, opened 2025-08-28
fix(optimism): Verify that flashblocks are not old according to canon state

RomanHodulak:
Part of #17858 #18102

Subscribes to canon state and performs integrity checks before executing flashblocks.

### DIFF
--- a/crates/chain-state/src/notifications.rs
+++ b/crates/chain-state/src/notifications.rs
@@ -122,13 +122,33 @@ impl<N: NodePrimitives> CanonStateNotification<N> {
         }
     }
 
-    /// Get the new tip of the chain.
+    /// Gets the new tip of the chain.
     ///
     /// Returns the new tip for [`Self::Reorg`] and [`Self::Commit`] variants which commit at least
     /// 1 new block.
+    ///
+    /// # Panics
+    ///
+    /// If chain doesn't have any blocks.
     pub fn tip(&self) -> &RecoveredBlock<N::Block> {
         match self {
             Self::Commit { new } | Self::Reorg { new, .. } => new.tip(),
+        }
+    }
+
+    /// Gets the new tip of the chain.
+    ///
+    /// If the chain has no blocks, it returns `None`. Otherwise, it returns the new tip for
+    /// [`Self::Reorg`] and [`Self::Commit`] variants.
+    pub fn tip_checked(&self) -> Option<&RecoveredBlock<N::Block>> {
+        match self {
+            Self::Commit { new } | Self::Reorg { new, .. } => {
+                if new.is_empty() {
+                    None
+                } else {
+                    Some(new.tip())
+                }
+            }
         }
     }
 
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -1,8 +1,9 @@
 use crate::{ExecutionPayloadBaseV1, FlashBlock};
 use alloy_eips::{eip2718::WithEncoded, BlockNumberOrTag, Decodable2718};
+use alloy_primitives::BlockHash;
 use eyre::OptionExt;
-use futures_util::{Stream, StreamExt};
-use reth_chain_state::ExecutedBlock;
+use futures_util::{FutureExt, Stream, StreamExt};
+use reth_chain_state::{CanonStateNotifications, CanonStateSubscriptions, ExecutedBlock};
 use reth_errors::RethError;
 use reth_evm::{
     execute::{BlockBuilder, BlockBuilderOutcome},
@@ -10,7 +11,8 @@ use reth_evm::{
 };
 use reth_execution_types::ExecutionOutcome;
 use reth_primitives_traits::{
-    AlloyBlockHeader, BlockTy, HeaderTy, NodePrimitives, ReceiptTy, SignedTransaction,
+    AlloyBlockHeader, BlockTy, HeaderTy, NodePrimitives, ReceiptTy, RecoveredBlock,
+    SignedTransaction,
 };
 use reth_revm::{database::StateProviderDatabase, db::State};
 use reth_rpc_eth_types::{EthApiError, PendingBlock};
@@ -21,6 +23,7 @@ use std::{
     task::{Context, Poll},
     time::{Duration, Instant},
 };
+use tokio::pin;
 use tracing::{debug, error, info};
 
 /// The `FlashBlockService` maintains an in-memory [`PendingBlock`] built out of a sequence of
@@ -37,6 +40,8 @@ pub struct FlashBlockService<
     blocks: Vec<FlashBlock>,
     evm_config: EvmConfig,
     provider: Provider,
+    canon_receiver: CanonStateNotifications<N>,
+    tip: Option<BlockHash>,
 }
 
 impl<
@@ -44,6 +49,7 @@ impl<
         S,
         EvmConfig: ConfigureEvm<Primitives = N, NextBlockEnvCtx: From<ExecutionPayloadBaseV1> + Unpin>,
         Provider: StateProviderFactory
+            + CanonStateSubscriptions<Primitives = N>
             + BlockReaderIdExt<
                 Header = HeaderTy<N>,
                 Block = BlockTy<N>,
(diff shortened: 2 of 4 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -157,13 +169,49 @@ impl<
...
+        let fut = self.canon_receiver.recv();
+        pin!(fut);
+
+        while let Poll::Ready(result) = fut.poll_unpin(cx) {
+            last = Some(result?);
+        }
+
+        if let Some(
+            CanonStateNotification::Commit { new } | CanonStateNotification::Reorg { new, .. },
+        ) = last
```
mattsse: we don't need to deconstruct this here, because there's also

`CanonStateNotification::tip` which does the same

crates/optimism/flashblocks/src/service.rs
```diff
@@ -157,13 +169,49 @@ impl<
             },
         ))
     }
+
+    fn are_flashblocks_up_to_date(&mut self, cx: &mut Context<'_>) -> eyre::Result<Option<bool>> {
```
mattsse: can we add some docs where what this does and returns?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -180,7 +228,12 @@ impl<
         match this.rx.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok(flashblock))) => {
                 this.add_flash_block(flashblock);
-                Poll::Ready(Some(this.execute()))
+
+                match this.are_flashblocks_up_to_date(cx) {
```
mattsse: this now polls only after we received a a new message, but these can happen individually, so we should also poll this individually.

we only need this really to set the latest channel value to None

but this poll loop now becomes a bit more tricky because we still want to kickoff a new job in case we just set the tip to tracked flashblock.parent_hash

so we kinda need to try_execute in two places

this will become more tricky, once we add the check for parent_hash == latest, because then we can no longer always return this.execute because by that time the chain might have advanced, or even after the execute call

so we need to remodel this a bit

mattsse (changes requested):
worth tackling the latest = parent_hash check in this pr as well because both affect the poll logic

crates/chain-state/src/notifications.rs
```diff
@@ -122,16 +122,36 @@ impl<N: NodePrimitives> CanonStateNotification<N> {
...
 
-    /// Get the new tip of the chain.
+    /// Gets the new tip of the chain.
     ///
     /// Returns the new tip for [`Self::Reorg`] and [`Self::Commit`] variants which commit at least
     /// 1 new block.
+    ///
+    /// # Panics
+    ///
+    /// If chain doesn't have any blocks.
```
mattsse, line 132: makes sense,

but this invariant is always enforced internally, but good to hihglight this

crates/optimism/flashblocks/src/service.rs
```diff
@@ -37,13 +40,16 @@ pub struct FlashBlockService<
     blocks: Vec<FlashBlock>,
     evm_config: EvmConfig,
     provider: Provider,
+    canon_receiver: CanonStateNotifications<N>,
+    tip: Option<BlockHash>,
```
mattsse: we can't rely on this alone because this still does not guarantee that:

this is the latest is parent-hash:

https://github.com/paradigmxyz/reth/blob/reth-optimism-flashblocks11/crates/optimism/flashblocks/src/service.rs#L122-L133

so we dont need to track this here

crates/optimism/flashblocks/src/service.rs
```diff
@@ -177,14 +211,43 @@ impl<
...
+                    Poll::Ready(this.current.clone().map(Ok)) // Output new block
+                } else {
+                    Poll::Pending // Keep last block
+                }
             }
-            Poll::Ready(Some(Err(err))) => Poll::Ready(Some(Err(err))),
-            Poll::Ready(None) => Poll::Ready(None),
-            Poll::Pending => Poll::Pending,
+            Ok(None) => Poll::Pending, // Cannot check integrity, tip or block is missing
+            Err(err) => Poll::Ready(Some(Err(err))),
```
mattsse: I'm realizing that there's now an architectural issue because this stream only emits new blocks, so we can't actually clear the pending block, so three solutions, either we move the watch channel in here and make this type the future, or we move the canonsubscription outside, or we make enum item for this stream, which doesnt seem ideal.

mattsse (changes requested):

crates/optimism/flashblocks/src/service.rs
```diff
@@ -172,19 +203,45 @@ impl<
...
+        // Consume new flashblocks while they're ready
+        while let Poll::Ready(Some(result)) = this.rx.poll_next_unpin(cx) {
+            match result {
+                Ok(flashblock) => this.add_flash_block(flashblock),
+                Err(err) => return Poll::Ready(Some(Err(err))),
+            }
+        }
+
+        // Execute block if there are flashblocks but no last pending block
+        let changed = if this.current.is_none() && !this.blocks.is_empty() {
```
julio4, line 226: `this.current.is_none()` is this not skipping execution of a new `PendingBlock` in the case of receiving new flashblock after a first execution?

RomanHodulak: The `current` should be cleared when `blocks` is changed

julio4: I see, [self.current.take()](https://github.com/paradigmxyz/reth/pull/18123/files#diff-a7851aac04b278d92ddba8bb73ef137a8e4056d174d0ff4cf90f78a38f3de764R86)!
Then execute update the `current`, and integrity is checked against canonical tip post execution.

Why not check before execution?
- To ensure it don't become invalid during execution I suppose.
- But it could already be invalid before execution, so it could prevent useless execution (in rare cases)?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -157,13 +168,33 @@ impl<
             },
         ))
     }
+
+    /// Compares tip from the last notification of [`CanonStateSubscriptions`] with last computed
+    /// pending block and verifies that the tip is the parent of the pending block.
+    fn verify_pending_block_integrity(
```
mattsse: not perfectly clear to me what the response value means, how to interpret true here

does true mean tip == fb.parent?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -172,19 +203,45 @@ impl<
...
+        match this.verify_pending_block_integrity(cx) {
+            // Integrity check failed: erase last block
+            Ok(Some(false)) => Poll::Ready(Some(Ok(None))),
+            // Integrity check is OK or skipped: output last block
+            Ok(Some(true) | None) => {
+                if changed {
+                    Poll::Ready(Some(Ok(this.current.clone())))
+                } else {
+                    Poll::Pending
+                }
```
mattsse, line 247: then this looks correct

mattsse (approved):
lgtm, one doc suggestion

crates/optimism/flashblocks/src/service.rs
RomanHodulak: The execution should be done only when flashblocks are changed. So if flashblocks are not changed, execution is not done but integrity check is. If flashblocks are changed, execution is done and then integrity check is.

If the integrity check was done before execution, you would be checking integrity of an old pending block.

To me the flow that makes most sense is:
* Exhaust all flashblocks that are ready
* Build new pending block if flashblocks were updated
* Verify integrity of the pending block against canon tip

julio4: When executing a new pending block it's always with one new flashblock (and in rarer cases multiple new ones, hence why you're draining fb stream first). Checking integrity is verify that received new fb are actually targeting pending block built on top of current canonical one, and not an old one.

Within a sequence of fbs, they're all built on the same parent block. When we receive FCU or base fb we move forward. In some cases you receive latest fb just after FCU, so if you keep executing it and only verifying after execution you're executing for nothing, it'll be invalidated after execution.

Fbs should be sent at regular intervals, but we don't want to assume that. Last flashblock is also the one with the maximum gas limit (to fill the remaining, allow for contract deployments etc). There's a risk execution of expired last fb is blocking the processing of the first new fb which quickstart new block building.

I think the overhead of checking the height before and after execution is worth it to avoid long blocking execution.

Lmk what you think @RomanHodulak

PR #18125, opened 2025-08-28
perf(optimism): use cached db in `FlashblockService`

julio4:
Close #18042
Related #17858

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -1,5 +1,6 @@
 use crate::{ExecutionPayloadBaseV1, FlashBlock};
 use alloy_eips::{eip2718::WithEncoded, BlockNumberOrTag, Decodable2718};
+use alloy_primitives::B256;
 use eyre::{eyre, OptionExt};
 use futures_util::{FutureExt, Stream, StreamExt};
 use reth_chain_state::{CanonStateNotifications, CanonStateSubscriptions, ExecutedBlock};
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -142,12 +175,15 @@ impl<
...
-            db.take_bundle(),
+            state.take_bundle(),
             vec![execution_result.receipts],
             block.number(),
             vec![execution_result.requests],
         );
 
+        // update cached reads
+        self.update_cached_reads(latest_hash, request_cache);
+
```
0xMelkor, line 206: It looks `self.current` is never set to `Some`. 

What about:
```rust
let pending = PendingBlock::with_executed_block(
            Instant::now() + Duration::from_secs(1),
            ExecutedBlock {
                recovered_block: block.into(),
                hashed_state: Arc::new(hashed_state),
            },
        );

 self.current = Some(pending.clone())
 Ok(pending)
```

0xMelkor (changes requested):
My 2 cents

crates/optimism/flashblocks/src/service.rs
julio4: It is set within stream impl [here](https://github.com/paradigmxyz/reth/pull/18125/files#diff-a7851aac04b278d92ddba8bb73ef137a8e4056d174d0ff4cf90f78a38f3de764R269)

mattsse (approved):

PR #18127, opened 2025-08-28
fix: add bundle and transaction context to call_many errors

crazykissshout:
When `eth_callMany` fails during transaction preparation or execution, the error message doesn't indicate which specific bundle or transaction caused the failure. This makes debugging difficult when processing multiple bundles with multiple transactions each.

**Before:**
`insufficient funds for gas * price + value
`

**After:**
`call execution error at bundle 1, tx 1: insufficient funds for gas * price + value
`


mattsse (changes requested):
this is indeed more helpful,

but this shouldnt be converted to an internal error, can we instead introduce a new error variant for this that wraps the error and adds a message as config?

mattsse (approved):
useful context imo

mattsse (approved):
useful context imo

klkvr (approved):

PR #18128, opened 2025-08-28
chore: direct constructor OpDAConfig::new

nkysg:
direct constructor OpDAConfig::new without call function


mattsse (commented):
~~I dont want to do this because this bypasses the checks we have in the setters~~

this is only initiated once and an atomic anyway so this is fine

PR #18129, opened 2025-08-28
refactor: merge `EthTransactionValidator` and `EthTransactionValidatorInner`

klkvr:
Removes the inner `Arc` from `EthTransactionValidator` allowing to expose internal implementations more easily.

It can still be `Arc`'ed manually if user needs cheap clones as in e.g `TransactionValidationTaskExecutor`

### DIFF
--- a/crates/transaction-pool/src/validate/task.rs
+++ b/crates/transaction-pool/src/validate/task.rs
@@ -78,12 +78,21 @@ impl ValidationJobSender {
 
 /// A [`TransactionValidator`] implementation that validates ethereum transaction.
 /// This validator is non-blocking, all validation work is done in a separate task.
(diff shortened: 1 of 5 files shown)


crates/transaction-pool/src/validate/task.rs
```diff
@@ -102,13 +111,13 @@ impl<V> TransactionValidationTaskExecutor<V> {
         F: FnMut(V) -> T,
     {
         TransactionValidationTaskExecutor {
-            validator: f(self.validator),
+            validator: Arc::new(f(Arc::into_inner(self.validator).unwrap())),
```
mattsse, line 114: this seems fine

mattsse (approved):
lgtm

PR #18132, opened 2025-08-28
fix(optimism): Fail if latest and base flashblock parent are different

### LINKED ISSUE #18102: Ensure flashblocks build on parent block and pending flashblock is cleared when a new chain is committed

RomanHodulak:
Part of #17858 
Closes #18102 

Makes the pending block execution fail on base flashblock and latest block mismatch.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -1,6 +1,6 @@
 use crate::{ExecutionPayloadBaseV1, FlashBlock};
 use alloy_eips::{eip2718::WithEncoded, BlockNumberOrTag, Decodable2718};
-use eyre::OptionExt;
+use eyre::{eyre, OptionExt};
 use futures_util::{FutureExt, Stream, StreamExt};
 use reth_chain_state::{CanonStateNotifications, CanonStateSubscriptions, ExecutedBlock};
 use reth_errors::RethError;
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -129,6 +129,10 @@ impl<
             .find_map(|v| v.base.clone())
             .ok_or_eyre("Missing base flashblock")?;
 
+        if attrs.parent_hash == latest.hash() {
+            return Err(eyre!("The base flashblock is old"));
+        }
```
mattsse: we want to ensure that flashblock is latest + 1

```suggestion
        if attrs.parent_hash != latest.hash() {
            return Err(eyre!("The base flashblock is old"));
        }
```

mattsse (changes requested):

mattsse (approved):

PR #18135, opened 2025-08-28
fix(net/network): check if there are any idle trusted peers at all

bendanzhentan:
The condition `num_idle_trusted_peers <= self.trusted_peer_ids.len()` is mathematically always true since `num_idle_trusted_peers` counts a subset of `trusted_peer_ids`.

This PR changes to `num_idle_trusted_peers > 0` to only accept connections when there are actually trusted peers in idle status.

I'm not entirely confident this completely fixes the issue. Please tell me if I missed something.

### DIFF
--- a/crates/net/network/src/peers.rs
+++ b/crates/net/network/src/peers.rs
@@ -2895,4 +2895,53 @@ mod tests {
         let updated_peer = manager.peers.get(&peer_id).unwrap();
         assert_eq!(updated_peer.addr.tcp().ip(), updated_ip);
     }
+
+    #[tokio::test]
+    async fn test_inc_pending_with_idle_trusted_peers() {
+        // Test the fix for the issue where the condition was always true
+        // This test verifies that the logic now correctly checks if there are any idle trusted
+        // peers
+
+        // Create a config with max_inbound = 0 to trigger the edge case logic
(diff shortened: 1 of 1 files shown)


crates/net/network/src/peers.rs
```diff
@@ -259,7 +259,7 @@ impl PeersManager {
             // are allowed (max_inbound == 0), in which case we still need to allow new pending
             // incoming connections until all trusted peers are connected.
             let num_idle_trusted_peers = self.num_idle_trusted_peers();
-            if num_idle_trusted_peers <= self.trusted_peer_ids.len() {
+            if num_idle_trusted_peers > 0 {
```
mattsse, line 262: > The condition num_idle_trusted_peers <= self.trusted_peer_ids.len() is mathematically always true

this would be the case if those would respect the same set, however

peer_ids !=

https://github.com/paradigmxyz/reth/blob/main/crates/net/network/src/peers.rs#L234-L237

this is because we not necessarily have an entry for trusted_peers_ids in the peers set, in case we only know its peerid 

hence we need to check <= here

bendanzhentan: Thanks @mattsse for explanation.
I will close this PR first and then take some more time to dive into this workflow.

PR #18141, opened 2025-08-29
feat: generalize impl EngineValidatorAddOn for OpAddOns

0xForerunner:
Fairly self descriptive. Makes swapping eth apis possible without making a custom AddOns.


mattsse (approved):
checks out, ty

PR #18153, opened 2025-08-29
fix(sigsegv): replace unwrap() with proper error handling in signal handler

hawkadrian:
Replace unwrap() call with proper error handling in SIGSEGV signal handler

Fixes potential panic in signal handler initialization that could occur
when the system is under memory pressure or has unusual memory alignment
requirements.


mattsse (commented):
this is fine

PR #18155, opened 2025-08-29
feat(optimism): add FlashblocksRx getter

mempirate:
Add a getter for the FlashblockRx receiver, so it can be subscribed to in `on_rpc_started` for example.


mattsse (approved):
lgtm

PR #18158, opened 2025-08-29
refactor(optimism): Extract responsibility to connect to a flashblock websocket stream

RomanHodulak:
Part of #17858 

Extracts the responsibility to establish a secure websocket connection away from `FlashBlockWsStream`.

It improves testability.

The `FlashBlockWsStream` is renamed to `WsFlashBlockStream` to have the specific technology as a prefix.

The `WsFlashBlockStream` now only handles the response from the stream, but it is no longer its responsibility to establish connection, nor does it care what the stream is, only what is its output.

### DIFF
--- a/crates/optimism/flashblocks/src/ws/stream.rs
+++ b/crates/optimism/flashblocks/src/ws/stream.rs
@@ -55,10 +55,10 @@ impl<S, C> WsFlashBlockStream<S, C> {
     }
 }
 
-impl<
-        S: Stream<Item = Result<Message, Error>> + Unpin,
-        C: WsConnect<Stream = S> + Clone + Send + Sync + 'static + Unpin,
-    > Stream for WsFlashBlockStream<S, C>
+impl<S, C> Stream for WsFlashBlockStream<S, C>
+where
+    S: Stream<Item = Result<Message, Error>> + Unpin,
+    C: WsConnect<Stream = S> + Clone + Send + Sync + 'static + Unpin,
 {
     type Item = eyre::Result<FlashBlock>;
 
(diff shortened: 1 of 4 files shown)


mattsse (commented):
I'm pretty confident in alloy websocket impl itself

what we could do instead here is use the real base wss endpoint and create a test for the ws connection and check that we can receive the blocks.
adding an abstraction over the ws itself is probably not necessary here

RomanHodulak:
@mattsse 

> I'm pretty confident in alloy websocket impl itself

Yes, but that's not the point.

The point is to be able to exclude the part we're confident in so that we can test the _other_ parts.

> what we could do instead here is use the real base wss endpoint and create a test for the ws connection and check that we can receive the blocks.

Yes, added this in PR #18170.

But this does not exclude unit testing.

> adding an abstraction over the ws itself is probably not necessary here

But that would exclude all the `WsFlashBlockStream` unit tests, which would be a pity.

For example, if we only use the real websocket connection, how do we test error cases?

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -21,26 +21,45 @@ use url::Url;
...
+            connect: Box::pin(async move { Err(Error::ConnectionClosed)? }),
+            stream: None,
+        }
+    }
+}
+
+impl<
+        S: Stream<Item = Result<Message, Error>> + Unpin,
+        C: WsConnect<Stream = S> + Clone + Send + Sync + 'static + Unpin,
+    > Stream for WsFlashBlockStream<S, C>
```
mattsse: can we move these to a where clause

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -73,28 +92,29 @@ impl Stream for FlashBlockWsStream {
     }
 }
 
-impl FlashBlockWsStream {
+impl<S, C: WsConnect<Stream = S> + Clone + Send + Sync + 'static> WsFlashBlockStream<S, C> {
```
mattsse: same here

mattsse (changes requested):
style nit

> But that would exclude all the WsFlashBlockStream unit tests, which would be a pity.

for reconnect this is nice to have
testing the actual flashblock impl logic is more impactful and for that we don't need ws

mattsse (approved):

PR #18159, opened 2025-08-29
perf(reth-invalid-block-hooks): use Reverts::eq reduce clone

nkysg:
The output is `BlockExecutionOutput<N::Receipt>`. It only needs to clone `Reverts`,  there is no need to clone the entire struct.
This has been addressed in PR  https://github.com/bluealloy/revm/pull/1827.


mattsse (approved):
ah nice, ty!

PR #18160, opened 2025-08-29
feat: support customizable RPC namespace parsers

klkvr:
Same as https://github.com/paradigmxyz/reth/pull/17758 but with a non-breaking approach for CLI types


fgimenez (approved):
nice!

PR #18162, opened 2025-08-29
perf(txpool): eliminate allocations in basefee enforcement

### LINKED ISSUE #18151: Optimize basefee tx promotions
### Describe the feature

when the basefee decreases we might be able to promote transactions:

https://github.com/paradigmxyz/reth/blob/001fb927b52a2b8f4a76198deae615d34a9c8260/crates/transaction-pool/src/pool/txpool.rs#L296-L308

this currently requires some allocs due to &mut access:

https://github.com/paradigmxyz/reth/blob/001fb927b52a2b8f4a76198deae615d34a9c8260/crates/transaction-pool/src/pool/parked.rs#L303-L305

we can get rid of this intermediary allocs if we optimize this call:
https://github.com/paradigmxyz/reth/blob/001fb927b52a2b8f4a76198deae615d34a9c8260/crates/transaction-pool/src/pool/txpool.rs#L307-L307
because we know the transaction can only go into the pending pool:

https://github.com/paradigmxyz/reth/blob/001fb927b52a2b8f4a76198deae615d34a9c8260/crates/transaction-pool/src/pool/txpool.rs#L303-L304

https://github.com/paradigmxyz/reth/blob/001fb927b52a2b8f4a76198deae615d34a9c8260/crates/transaction-pool/src/pool/txpool.rs#L1037-L1039

then we can remove the alloc and operate on iterators for `enforce_basefee`

## TODO
* remove intermediary alloc in `enforce_basefee` and operate on iterator

### Additional context

_No response_

yongkangc:
## Summary

Closes #18151

Optimizes transaction pool basefee enforcement by eliminating Vec allocations during base fee decreases. This addresses performance bottlenecks identified in #18151 by replacing the allocation-heavy approach with zero-allocation closure-based processing.

`BASE_FEE_POOL_BITS + ENOUGH_FEE_CAP_BLOCK = PENDING_POOL_BITS`. When basefee decreases, qualifying transactions can only transition BaseFee → Pending, enabling direct routing without generic dispatch.

## Changes

- **New method**: `enforce_basefee_with_handler()` processes transactions via closure instead of collecting into Vec
- **Direct routing**: Leverages mathematical certainty that BaseFee → Pending transitions are deterministic
- **Optimized flow**: Eliminates generic routing overhead by directly updating transaction state and inserting into pending pool
- **Backward compatibility**: Maintains `enforce_basefee()` method for test compatibility using `#[cfg(test)]`

## Performance Impact

**Before**: 2 Vec allocations per basefee enforcement cycle
- `Vec<TransactionId>` from `satisfy_base_fee_ids()`  
- `Vec<Arc<ValidPoolTransaction<T>>>` from `enforce_basefee()`

**After**: Zero allocations with immediate processing via closure

### DIFF
--- a/crates/transaction-pool/src/pool/parked.rs
+++ b/crates/transaction-pool/src/pool/parked.rs
@@ -295,22 +295,6 @@ impl<T: PoolTransaction> ParkedPool<BasefeeOrd<T>> {
         transactions
     }
 
-    /// Removes all transactions and their dependent transaction from the subpool that no longer
-    /// satisfy the given basefee.
-    ///
-    /// Legacy method maintained for compatibility with read-only queries.
-    /// For basefee enforcement, prefer `enforce_basefee_with_handler` for better performance.
-    ///
-    /// Note: the transactions are not returned in a particular order.
-    #[cfg(test)]
-    pub(crate) fn enforce_basefee(&mut self, basefee: u64) -> Vec<Arc<ValidPoolTransaction<T>>> {
-        let mut removed = Vec::new();
-        self.enforce_basefee_with_handler(basefee, |tx| {
-            removed.push(tx);
-        });
-        removed
-    }
-
     /// Removes all transactions from this subpool that can afford the given basefee,
     /// invoking the provided handler for each transaction as it is removed.
     ///
@@ -365,6 +349,22 @@ impl<T: PoolTransaction> ParkedPool<BasefeeOrd<T>> {
                 );
             }
         }
+    }
+
+    /// Removes all transactions and their dependent transaction from the subpool that no longer
+    /// satisfy the given basefee.
+    ///
+    /// Legacy method maintained for compatibility with read-only queries.
+    /// For basefee enforcement, prefer `enforce_basefee_with_handler` for better performance.
+    ///
+    /// Note: the transactions are not returned in a particular order.
+    #[cfg(test)]
+    pub(crate) fn enforce_basefee(&mut self, basefee: u64) -> Vec<Arc<ValidPoolTransaction<T>>> {
+        let mut removed = Vec::new();
+        self.enforce_basefee_with_handler(basefee, |tx| {
+            removed.push(tx);
+        });
+        removed
     }
 }
 
@@ -1160,5 +1160,4 @@ mod tests {
         // Verify transactions were removed from pool
         assert_eq!(pool.len(), 1); // Only the 300 fee tx should remain
     }
-
 }
--- a/crates/transaction-pool/src/pool/parked.rs
+++ b/crates/transaction-pool/src/pool/parked.rs
@@ -1096,4 +1096,69 @@ mod tests {
         assert!(removed.is_some());
         assert!(!pool.contains(&tx_id));
     }
+
+    #[test]
+    fn test_enforce_basefee_with_handler_zero_allocation() {
+        let mut f = MockTransactionFactory::default();
+        let mut pool = ParkedPool::<BasefeeOrd<_>>::default();
+
+        // Add multiple transactions across different fee ranges
+        let sender_a = address!("0x000000000000000000000000000000000000000a");
+        let sender_b = address!("0x000000000000000000000000000000000000000b");
+
+        // Add transactions where nonce ordering allows proper processing:
+        // Sender A: both transactions can afford basefee (500 >= 400, 600 >= 400)
+        // Sender B: transaction cannot afford basefee (300 < 400)
+        let txs = vec![
+            f.validated_arc(
+                MockTransaction::eip1559()
+                    .set_sender(sender_a)
+                    .set_nonce(0)
+                    .set_max_fee(500)
+                    .clone(),
+            ),
+            f.validated_arc(
+                MockTransaction::eip1559()
+                    .set_sender(sender_a)
+                    .set_nonce(1)
+                    .set_max_fee(600)
+                    .clone(),
+            ),
+            f.validated_arc(
+                MockTransaction::eip1559()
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/pool/parked.rs
```diff
@@ -298,16 +298,73 @@ impl<T: PoolTransaction> ParkedPool<BasefeeOrd<T>> {
...
+    ///
+    /// Note: the transactions are not returned in a particular order.
+    pub(crate) fn enforce_basefee_with_handler<F>(&mut self, basefee: u64, mut tx_handler: F)
+    where
+        F: FnMut(Arc<ValidPoolTransaction<T>>),
+    {
+        let basefee = basefee as u128;
+
+        // Collect qualifying transaction IDs first
+        let to_remove = {
```
mattsse: this is basically the same as just calling `    fn satisfy_base_fee_ids(&self, basefee: u128) -> Vec<TransactionId> {
`

?

crates/transaction-pool/src/pool/parked.rs
```diff
@@ -298,16 +298,73 @@ impl<T: PoolTransaction> ParkedPool<BasefeeOrd<T>> {
...
+    ///
+    /// This method enforces the basefee constraint by identifying transactions that now
+    /// satisfy the basefee requirement (typically after a basefee decrease) and processing
+    /// them via the provided transaction handler closure.
+    ///
+    /// Respects per-sender nonce ordering: if the lowest-nonce transaction for a sender
+    /// still cannot afford the basefee, higher-nonce transactions from that sender are skipped.
+    ///
+    /// Note: the transactions are not returned in a particular order.
+    pub(crate) fn enforce_basefee_with_handler<F>(&mut self, basefee: u64, mut tx_handler: F)
```
mattsse: can drop _handler here imo
```suggestion
    pub(crate) fn enforce_basefee_with<F>(&mut self, basefee: u64, mut tx_handler: F)
```

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -293,19 +293,27 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+                        let meta =
                             self.all_transactions.txs.get_mut(tx.id()).expect("tx exists in set");
-                        tx.state.insert(TxState::ENOUGH_FEE_CAP_BLOCK);
-                        tx.subpool = tx.state.into();
-                        tx.subpool
-                    };
-                    self.add_transaction_to_subpool(to, tx);
-                }
+                        meta.state.insert(TxState::ENOUGH_FEE_CAP_BLOCK);
+                        meta.subpool = SubPool::Pending;
```
mattsse: we should still do into() here

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -293,19 +293,27 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+
+                        self.pending_pool
+                            .add_transaction(tx, self.all_transactions.pending_fees.base_fee);
```
mattsse: for sanity reasons I'd like to inline

https://github.com/paradigmxyz/reth/blob/5c0c8bb38d369de19160d03ccbb06c18cfd9c3f7/crates/transaction-pool/src/pool/txpool.rs#L1035-L1046

here, minus the basefee arm because unreachable

mattsse (changes requested):
cool, smol nits

yongkangc:
> cool, smol nits

thanks matt for the review! really lightning fast, i am still reviewing myself 😆

crates/transaction-pool/src/pool/parked.rs
yongkangc: that looks alot better, thanks!

crates/transaction-pool/src/pool/parked.rs
yongkangc: thats a good one, thanks

crates/transaction-pool/src/pool/txpool.rs
yongkangc: is this what u meant by inlining?

https://github.com/paradigmxyz/reth/blob/101222be2d4db72e80b181046104b02f3a6b1b4a/crates/transaction-pool/src/pool/txpool.rs#L313-L329

to me yea, makes more sense

crates/transaction-pool/src/pool/parked.rs
```diff
@@ -295,18 +295,49 @@ impl<T: PoolTransaction> ParkedPool<BasefeeOrd<T>> {
...
         for id in to_remove {
-            removed.push(self.remove_transaction(&id).expect("transaction exists"));
+            if let Some(tx) = self.remove_transaction(&id) {
+                tx_handler(tx);
+            } else {
+                tracing::warn!(
+                    target: "reth::txpool::parked",
+                    ?id,
+                    "Failed to remove transaction during basefee enforcement - possible data structure inconsistency"
+                );
```
mattsse: we can ignore that case here because that invariant is enforced, this was even an expect earlier

mattsse (changes requested):
nice, last tiny observability nit

crates/transaction-pool/src/pool/parked.rs
yongkangc: ahh okay, was afraid that the expect might have cause some issues / wanted to log - but makes sense!

yongkangc: done! @mattsse

mattsse (approved):

PR #18164, opened 2025-08-29
fix(optimism): find fb attrs in base fb

julio4:
No need to iter on all flashblocks to find base, it is only included in the first base flashblocks (index == 0) and omitted in all subsequent fbs.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -125,8 +125,8 @@ impl<
 
         let attrs = self
             .blocks
-            .iter()
-            .find_map(|v| v.base.clone())
+            .first()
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -125,8 +125,9 @@ impl<
 
         let attrs = self
             .blocks
-            .iter()
-            .find_map(|v| v.base.clone())
+            .first()
+            .map(|v| v.base.clone())
+            .flatten()
```
RomanHodulak:
```suggestion
            .and_then(|v| v.base.clone())
```

mattsse (approved):

PR #18170, opened 2025-08-29
test(optimism): Test that streaming flashblocks from remote source is successful

RomanHodulak:
Part of #17858 

Adds an integration test that streams flashblocks from base.

It uncovered that the schema has changed.

Fix included.

### DIFF
--- a/crates/optimism/flashblocks/Cargo.toml
+++ b/crates/optimism/flashblocks/Cargo.toml
@@ -32,7 +32,7 @@ alloy-rpc-types-engine = { workspace = true, features = ["serde"] }
 
 # io
 tokio.workspace = true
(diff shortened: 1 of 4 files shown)


crates/optimism/flashblocks/Cargo.toml
```diff
@@ -32,7 +32,7 @@ alloy-rpc-types-engine = { workspace = true, features = ["serde"] }
 
 # io
 tokio.workspace = true
-tokio-tungstenite.workspace = true
+tokio-tungstenite = { workspace = true, features = ["rustls-tls-native-roots"] }
```
mattsse, line 35: makes sense, we do expect this only be used with wss endpoints

mattsse (approved):

PR #18172, opened 2025-08-30
refactor: remove unnecessary PathBuf clone in CLI help generator

davidjsonn:
Remove redundant clone() call when joining paths in docs/cli/help.rs. The join() method already creates a new PathBuf, making the clone() unnecessary.


mattsse (approved):
odd that clippy doesnt catch this

PR #18173, opened 2025-08-30
chore: fixed dead links in repo

andrewshab3:
## Description:

- Fixed broken links


docs/crates/db.md
mattsse, line 1: why are those removed?

andrewshab3: > why are those removed?

Files from my previous commit, i don't know why they didn't sync

crates/cli/commands/src/download.rs
```diff
@@ -17,7 +17,7 @@ use tokio::task;
 use tracing::info;
 
 const BYTE_UNITS: [&str; 4] = ["B", "KB", "MB", "GB"];
-const MERKLE_BASE_URL: &str = "https://snapshots.merkle.io";
+const MERKLE_BASE_URL: &str = "https://www.merkle.io/snapshots";
```
mattsse, line 20: we just corrected this

PR #18175, opened 2025-08-31
feat(examples): added txpoolExt_clearTxpool to existing example

Timosdev99:
closes: https://github.com/paradigmxyz/reth/issues/18134

this adds a new txpoolExt_clearTxpool  RPC method to the node-custom-rpc example 

the method provide a way to clear all transaction from the transaction pool


mattsse (approved):
lgtm

PR #18176, opened 2025-08-31
chore: avoid using hashmap hashers directly

DaniPopes:
Some code assumes that hashmap hasher is a certain type, when it's currently a type alias configurable by cfg.

### DIFF
--- a/crates/cli/commands/src/db/checksum.rs
+++ b/crates/cli/commands/src/db/checksum.rs
@@ -2,7 +2,7 @@ use crate::{
     common::CliNodeTypes,
     db::get::{maybe_json_value_parser, table_key},
 };
-use ahash::RandomState;
+use alloy_primitives::map::foldhash::fast::FixedState;
 use clap::Parser;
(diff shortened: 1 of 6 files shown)


crates/cli/commands/src/db/checksum.rs
```diff
@@ -102,7 +102,7 @@ impl<N: ProviderNodeTypes> TableViewer<(u64, Duration)> for ChecksumViewer<'_, N
         };
 
         let start_time = Instant::now();
-        let mut hasher = RandomState::with_seeds(1, 2, 3, 4).build_hasher();
+        let mut hasher = FixedState::with_seed(u64::from_be_bytes(*b"RETHRETH")).build_hasher();
```
DaniPopes, line 105: does the algorithm here matter across reth versions? @Rjected

mattsse: it doesn't really, this is mostly intended to compare two different dbs, as long as this command is invoked with the same version this is fine

mattsse (approved):

PR #18178, opened 2025-09-01
txpool: make --txpool.max-tx-gas a u64 with 0=disabled

niran:
`--txpool.max-tx-gas` was initially implemented as an `Option<u64>`, but that makes it easy to provide a zero value that will reject all transactions from the mempool. This change treats zero as disabled, and defaults the option to zero instead of None.


mattsse (commented):
this doesnt make much sense to set it to 0

makes more sense to set it to max, but this setting is already optional so unclear why we'd need to use specific ints as option

PR #18183, opened 2025-09-01
feat(optimism): keep flashblocks in ordered sequence for pending block

### LINKED ISSUE #18182: Do not ignore flashblocks not received in sequence within a pending block
### Describe the feature

Currently if we receive a flashblocks with the index not successive to the last one receive in the current sequence of flashblocks for the current pending block, it is ignored:

https://github.com/paradigmxyz/reth/blob/203cb6e1587af3fd02c8ad26eb2b5751f55b32fa/crates/optimism/flashblocks/src/service.rs#L88-L109

Flashblocks are sensitive to potential network latency, and we want to ensure that we can keep delivering flashblock services even if we receive some of them late (unordered).

### Additional context

_No response_

julio4:
Close #18182
Relevant #17858

When we receive a flashblock for the current pending block, we store it in a `FlashblockSequence`. 
If there's a gap in the sequence, we still keep all 'pending' received flashblocks instead of discarding them.
`FlashblockSequence` is a simple wrapper over an ordered B-tree to keep all flashblocks, with a `iter_ready()` iterator implementation to easily iterate over the ready part of the sequence (where all fbs are successive from index 0).

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -25,7 +25,7 @@ use std::{
     time::{Duration, Instant},
 };
 use tokio::pin;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, trace};
 
 /// The `FlashBlockService` maintains an in-memory [`PendingBlock`] built out of a sequence of
 /// [`FlashBlock`]s.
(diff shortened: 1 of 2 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -36,12 +37,62 @@ pub struct FlashBlockService<
...
     evm_config: EvmConfig,
     provider: Provider,
     canon_receiver: CanonStateNotifications<N>,
 }
 
+/// Simple wrapper around an ordered B-tree to keep track of a sequence of flashblocks by index
+#[derive(Debug)]
+struct FlashBlockSequence {
+    inner: BTreeMap<u64, FlashBlock>,
+}
```
mattsse: this type is out of order, this should be moved afer the impl blocks of FlashBlockService

crates/optimism/flashblocks/src/service.rs
```diff
@@ -77,12 +128,15 @@ impl<
...
         }
         // Flash block at the following index adds to the collection and invalidates built block
-        else if flashblock.index == self.blocks.last().map(|last| last.index + 1).unwrap_or(0) {
-            self.blocks.push(flashblock);
+        else if flashblock.index ==
+            self.blocks.iter_ready().last().map(|last| last.index + 1).unwrap_or(0)
+        {
+            self.blocks.insert(flashblock);
             self.current.take();
         }
```
mattsse: is this check then still relevant if we don't care about the order

then all of this can be simplified, but we should check that they belong to the same block

mattsse (changes requested):
this makes sense,

however if we expect that they can arrive out of order then this doesnt hold for the first block, which currently resets the sequence,
this also then doesnt cover the case where we receive two different flashblocks for different blocks, so we probably should track the actual parent block as well somehow

however all of this would only be the case if the remote sends us the blocks out of order

crates/optimism/flashblocks/src/service.rs
julio4: Yes I agree! I propose review/merge #18192 first, which brings better handling of incoming flashblocks based on the current pending block

crates/optimism/flashblocks/src/service.rs
```diff
@@ -296,3 +257,64 @@ impl<
...
+    fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    /// Return base flashblock
+    /// Base flashblock is the first flashblock of index 0 in the sequence
+    fn base(&self) -> Option<&FlashBlock> {
+        self.inner.get(&0)
+    }
+
```
julio4, line 309: We could also implement`parent_hash` on `FlashblockSequence` as the base is only included in the first fb, but the scope is for the whole sequence.

```suggestion

    pub fn parent_hash(&self) -> Option<B256> {
        Some(self.base()?.parent_hash)
    }

```

mattsse (commented):
thanks for this @julio4 

I merged this into #18194 which now supersedes this pr
https://github.com/paradigmxyz/reth/pull/18194/commits/30dbd8b49e46ff9f0c9dbb9096690e5197d56b0d

PR #18190, opened 2025-09-01
fix: fix formatting in RPC and transaction-pool

crStiv:
Fix code formatting in RPC API and transaction pool modules.


mattsse (commented):
this does not fix anything

PR #18192, opened 2025-09-01
feat(optimism): Handle incoming flashblocks against current pending block number

julio4:
Close #18191
Relevant #17858

When handling a received flashblock (in `add_flash_block` of Flashblock service):
- fetch latest block number to calculate pending block height
- Correctly handle flashblock based on its `flashblock.metadata.block_number`

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -25,7 +25,7 @@ use std::{
     time::{Duration, Instant},
 };
 use tokio::pin;
-use tracing::{debug, error, info};
+use tracing::{debug, error};
 
 /// The `FlashBlockService` maintains an in-memory [`PendingBlock`] built out of a sequence of
 /// [`FlashBlock`]s.
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -38,6 +38,7 @@ pub struct FlashBlockService<
 > {
     rx: S,
     current: Option<PendingBlock<N>>,
+    height: Option<u64>,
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -110,38 +111,47 @@ impl<
...
-                    message = "Received Flashblock for new block, zeroing Flashblocks until we receive a base Flashblock",
-                    curr_block = %pending_block.block().header().number(),
-                    new_block = %flashblock.metadata.block_number,
-                );
+    pub fn add_flash_block(&mut self, flashblock: FlashBlock) -> eyre::Result<()> {
+        let latest = self
+            .provider
+            .latest_header()?
+            .ok_or(EthApiError::HeaderNotFound(BlockNumberOrTag::Latest.into()))?
+            .number();
```
mattsse: this isnt something we want to do on a per flashblock basis

we want to track the entire block via messages but only build it if its the latest

mattsse (commented):
not something that is useful here imo

mattsse:
how does this relate to #18183

crates/optimism/flashblocks/src/service.rs
julio4: Is this reasonable to only fetch if it's the first flashblock?

julio4:
> how does this relate to #18183

It should implement this todo
https://github.com/paradigmxyz/reth/blob/0ad8749dbeea2a4d0ea7465170f00df9e6c5f892/crates/optimism/flashblocks/src/service.rs#L139-L142

mattsse (commented):
thanks, 
this is no longer relevant after we made some other simplifications

PR #18193, opened 2025-09-01
perf(testing): remove unnecessary clone() calls for Copy types

davidjsonn:
Remove redundant .clone() calls on Range<u8> and Option<Range<u8>> types in block generators


mattsse (commented):
this doesnt work

PR #18194, opened 2025-09-01
chore: simplify flashblocks poll logic

mattsse:
we can slightly simplify this just by changing a few things around

includes #18183

fixed a few other things as drive by, like busy poll loop and unhandled ws messages

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -93,8 +93,7 @@ impl<
         self.cached_state = Some((head, cached_reads));
     }
 
-    /// Returns the [`ExecutedBlock`] made purely out of [`FlashBlock`]s that were received using
-    /// [`Self::add_flash_block`] on top of the latest state.
+    /// Returns the [`ExecutedBlock`] made purely out of [`FlashBlock`]s that were received earlier.
     ///
     /// Returns None if the flashblock doesn't attach to the latest header.
     fn execute(&mut self) -> eyre::Result<Option<PendingBlock<N>>> {
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -25,7 +25,7 @@ use std::{
     time::{Duration, Instant},
 };
 use tokio::pin;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, trace};
 
 /// The `FlashBlockService` maintains an in-memory [`PendingBlock`] built out of a sequence of
 /// [`FlashBlock`]s.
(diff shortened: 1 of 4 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -259,40 +230,52 @@ impl<
...
+        // advance new canonical message, if any to reset flashblock
+        {
+            let fut = this.canon_receiver.recv();
+            pin!(fut);
+            if fut.poll_unpin(cx).is_ready() {
+                // if we have a new canonical message, we know the currently tracked flashblock is
+                // invalidated
+                if this.current.take().is_some() {
+                    return Poll::Ready(Some(Ok(None)))
+                }
```
RomanHodulak: `current` can be erased earlier in `add_flash_block`, but here it seems that it is expected that it is kept

crates/optimism/flashblocks/src/service.rs
```diff
@@ -166,7 +161,8 @@ impl<
             .ok_or_eyre("Missing base flashblock")?;
 
         if attrs.parent_hash != latest_hash {
-            return Err(eyre!("The base flashblock is old"));
+            // doesn't attach to the latest block
+            return Ok(None)
```
RomanHodulak, line 116: This is probably a good change, I also felt weird about returning error here

crates/optimism/flashblocks/src/service.rs
```diff
@@ -259,40 +230,52 @@ impl<
...
-                }
+        // try to build a block on top of latest
+        match this.execute() {
+            Ok(Some(new_pending)) => {
+                // built a new pending block
+                this.current = Some(new_pending.clone());
+                return Poll::Ready(Some(Ok(Some(new_pending))));
+            }
+            Ok(None) => {
+                // nothing to do because tracked flashblock doesn't attach to latest
```
RomanHodulak, line 223: maybe it should be erased in that case? i.e. emit Poll::Ready(Ok(None))

crates/optimism/flashblocks/src/service.rs
```diff
@@ -259,40 +230,52 @@ impl<
...
+                    return Poll::Ready(Some(Ok(None)))
+                }
             }
+        }
 
-            true
-        } else {
-            false
-        };
+        if !new_flashblock && this.current.is_none() {
```
RomanHodulak, line 210: Why the `this.current.is_none` check here?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -259,40 +230,52 @@ impl<
...
+                // built a new pending block
+                this.current = Some(new_pending.clone());
+                return Poll::Ready(Some(Ok(Some(new_pending))));
+            }
+            Ok(None) => {
+                // nothing to do because tracked flashblock doesn't attach to latest
+            }
+            Err(err) => {
+                // we can ignore this error
+                debug!(%err, "failed to execute flashblock");
```
RomanHodulak, line 227: Could this hide errors?

RomanHodulak (commented):
I feel weird about the usage of `current`. Maybe we should reevaluate?

In my mind I was thinking of `current` as the last executed block made out of currently set `blocks`. Changing `blocks` invalidate `current`, that's why it can be cleared in `add_flash_block`. But then the future should emit `Poll::Ready(Ok(None))` or `Poll::Ready(Ok(Some(...)))`.

crates/optimism/flashblocks/src/service.rs
mattsse: in case we yielded earlier on new canonical chain event

crates/optimism/flashblocks/src/service.rs
mattsse: not really I think

crates/optimism/flashblocks/src/service.rs
mattsse: I see, yeah, 

add_flash_block should now only track the new flashblock, will change that

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -307,14 +317,8 @@ mod tests {
     async fn test_stream_fails_to_decode_non_binary_message() {
```
RomanHodulak: Since this is no longer an error, it would be appropriate
```suggestion
    async fn test_stream_ignores_non_binary_message() {
```

RomanHodulak (approved):
Good work, small nit.

PR #18196, opened 2025-09-01
perf(reth-optimism-flashblocks): rm redundant clone

nkysg:


crates/optimism/flashblocks/src/ws/decoding.rs
```diff
@@ -33,7 +33,7 @@ impl FlashBlock {
         let payload: FlashblocksPayloadV1 = serde_json::from_slice(&bytes)
             .map_err(|e| eyre::eyre!("failed to parse message: {e}"))?;
 
-        let metadata: Metadata = serde_json::from_value(payload.metadata.clone())
+        let metadata: Metadata = serde_json::from_value(payload.metadata)
             .map_err(|e| eyre::eyre!("failed to parse message metadata: {e}"))?;
```
mattsse, line 37: @RomanHodulak why are we doing this in the first place?

first deserialize into json then into meta?

RomanHodulak: The [FlashblocksPayloadV1](https://github.com/flashbots/rollup-boost/blob/9fdd722ced0c8430097e7edc5a9024f7aa27c5ad/crates/rollup-boost/src/flashblocks/primitives.rs#L74) defines that the structure of metadata field as unspecified. It might be the reason why it changes so often.

The `FlashblocksPayloadV1` structure is copied over from [`rollup_boost`](https://github.com/flashbots/rollup-boost). But since it is redefined here, we could change the type of the metadata field so that it deserializes directly into `Metadata`. However, the idea seems to be that things like this will be upstreamed to `alloy` unless we use the `rollup_boost`. Therefore we might not want to introduce these changes.

mattsse: should we just keep it as json value for now, because rn, this is just metadata deserde with extra steps?

RomanHodulak: I think it's okay to keep it.

Or we can change it to avoid confusion and to remove an extra step.

mattsse (approved):

PR #18197, opened 2025-09-01
perf(reth-engine-primitives): box CanonicalBlockAdded

nkysg:


mattsse (commented):
this type is just a wrapper around a few arcs so not worth boxing

PR #18198, opened 2025-09-01
perf(reth-engine-local): use VecDeque reduce removal operations

nkysg:
use `VecDeque` can redude `self.last_block_hashes.split_off(self.last_block_hashes.len() - 64);` removal operations.


mattsse (approved):
this makes sense and prevents shifting the entire vec

PR #18208, opened 2025-09-02
fix: allow mutable access to provider and viewer

0xstratus:
updated the function to take `provider_factory` as mutable and adjusted `ClearViewer` to hold a mutable reference to the database.

this fixes the issue where calling `db_mut()` or `tx_mut()` required a mutable reference, and allows passing `&mut viewer` to `table.view(...)` as intended.


mattsse (commented):
this isnt necessary, the tx_mut only carries info that this allows modifications via the underlying tx

PR #18209, opened 2025-09-02
Cap initial pre-range by local to_block in receipts log pruner

sashass1315:
- Clamp the first pre-range end to min(start_block - 1, to_block).
- Skip adding the range if it becomes empty (init > end).
- Ensures deletions never exceed the safety boundary (MINIMUM_PRUNING_DISTANCE) and aligns behavior with subsequent ranges.


mattsse (commented):
this isnt reachable I believe because the iter is initialized with those bounds

wdyt @joshieDo

PR #18212, opened 2025-09-02
feat(optimism): Respond to ping messages with pong in `WsFlashBlockStream`

RomanHodulak:
Part of #17858 

### About
Increases resilience of websocket handling for flashblocks.

### Motivation
A websocket server sometimes sends a ping message. Usually it is to check whether the receiver is still listening.

### Problem
We currently ignore these messages. While we might get away with it, the server might decide to cut the connection while it's still alive.

### Solution
Adds the capability of sending messages to `WsFlashBlockStream` using a websocket sink.

Responds to `Ping` with `Pong`.

### DIFF
--- a/crates/optimism/flashblocks/src/ws/stream.rs
+++ b/crates/optimism/flashblocks/src/ws/stream.rs
@@ -1,5 +1,8 @@
 use crate::FlashBlock;
-use futures_util::{stream::SplitStream, FutureExt, Stream, StreamExt};
+use futures_util::{
+    stream::{SplitSink, SplitStream},
+    FutureExt, Sink, Stream, StreamExt,
+};
 use std::{
     fmt::{Debug, Formatter},
     future::Future,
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -51,60 +56,74 @@ impl<S, C> WsFlashBlockStream<S, C> {
...
+                    }
+                    Ok(Message::Ping(bytes)) => this.ping(bytes),
+                    Ok(msg) => debug!("Received unexpected message: {:?}", msg),
+                    Err(err) => return Poll::Ready(Some(Err(err.into()))),
                 }
-                Ok(msg) => {
-                    debug!("Received unexpected message: {:?}", msg);
+            }
+
+            if let State::Ping(pong) = &mut this.state {
```
mattsse: hmm, the stream and sink can be operated on at the same time, so we dont need a dedicated state for this and can instead have this as a standalone Option, so that we can keep reading while we're responding

mattsse (changes requested):
makes sense,

one suggestion, because stream and sink can be operated on at the same time

mattsse (approved):

PR #18218, opened 2025-09-02
perf(txpool): eliminate allocations in basefee enforcement

yongkangc:
Reverts paradigmxyz/reth#18215 because of bug 

add some additional sanity tests for this scenario
https://github.com/paradigmxyz/reth/pull/18216

some sanity checks would be
- checking that non 4844 transactions always has blob fee bit
- non 4844 can hit the pending pool and not get stuck in base fee
- guard the remove blobfee bit against is_eip4844

Only EIP-4844 transactions should have conditional blob fee bit logic. Non-4844 transactions should always have ENOUGH_BLOB_FEE_CAP_BLOCK = true

### DIFF
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -293,19 +293,40 @@ impl<T: TransactionOrdering> TxPool<T> {
                 Ordering::Greater
             }
             Ordering::Less => {
-                // decreased base fee: recheck basefee pool and promote all that are now valid
-                let removed =
-                    self.basefee_pool.enforce_basefee(self.all_transactions.pending_fees.base_fee);
-                for tx in removed {
-                    let to = {
-                        let tx =
+                // Base fee decreased: recheck BaseFee and promote.
+                // Invariants:
+                // - BaseFee contains only non-blob txs (blob txs live in Blob) and they already
+                //   have ENOUGH_BLOB_FEE_CAP_BLOCK.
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -2954,6 +2975,97 @@ mod tests {
...
+        // tx3 should still be in basefee pool (fee 400 < basefee 450)
+        assert_eq!(pool.all_transactions.txs.get(&id3).unwrap().subpool, SubPool::BaseFee);
+
+        // tx1 and tx2 should be in pending pool with correct state bits
+        let tx1_meta = pool.all_transactions.txs.get(&id1).unwrap();
+        let tx2_meta = pool.all_transactions.txs.get(&id2).unwrap();
+        assert_eq!(tx1_meta.subpool, SubPool::Pending);
+        assert_eq!(tx2_meta.subpool, SubPool::Pending);
+        assert!(tx1_meta.state.contains(TxState::ENOUGH_FEE_CAP_BLOCK));
+        assert!(tx2_meta.state.contains(TxState::ENOUGH_FEE_CAP_BLOCK));
```
mattsse, line 3060: nice this should catch it

mattsse (approved):
lgtm,

I converted the panic into a warn instead,just to be extra safe, but this should now no longer be reachable anyway

crates/transaction-pool/src/pool/txpool.rs
yongkangc: thanks @mattsse

PR #18222, opened 2025-09-02
Use safe arithmetic in get_block_by_number and handle empty block_range

Fibonacci747:
- Replace unchecked subtraction with checked_sub in get_block_by_number to avoid underflow and return None when number < start or index is out of bounds.
- Make block_range return an empty range when there are no blocks by using start..=start.saturating_sub(1).
- Keeps behavior consistent with BlockIndex and avoids debug panics and subtle release-mode issues.


mattsse (commented):
not necessary

PR #18224, opened 2025-09-02
fix(optimism): Prevent repeated executions of current flashblock sequence

RomanHodulak:
Closes #18221 
Part of #17858 

The boolean variable that was in charge of keeping track of changes is not reliable, because it gets lost on next poll.

The `current` field should be reexecuted if it does not correspond with the current flashblock sequence, but is still valid just misses the last txs.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -37,6 +37,7 @@ pub struct FlashBlockService<
     rx: S,
     current: Option<PendingBlock<N>>,
     blocks: FlashBlockSequence<N::SignedTx>,
+    rebuild: bool,
     evm_config: EvmConfig,
     provider: Provider,
     canon_receiver: CanonStateNotifications<N>,
@@ -71,6 +72,7 @@ where
             canon_receiver: provider.subscribe_to_canonical_state(),
             provider,
             cached_state: None,
+            rebuild: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -187,17 +187,17 @@ where
...
-                    if let Err(err) = this.blocks.insert(flashblock) {
-                        debug!(%err, "Failed to prepare flashblock");
-                    } else {
-                        new_flashblock = true;
+                Ok(flashblock) => match this.blocks.insert(flashblock) {
+                    Ok(_) => {
+                        if this.current.take().is_some() {
+                            return Poll::Ready(None)
+                        }
                     }
```
mattsse: this will terminate the entire stream

terminating

https://github.com/paradigmxyz/reth/blob/6bcd5e07ac522c3ac0f0795c9e94766f438fd799/crates/optimism/flashblocks/src/service.rs#L81-L81

mattsse (changes requested):

crates/optimism/flashblocks/src/service.rs
RomanHodulak: Didn't think of this as an issue since it gets repolled, but can change.

crates/optimism/flashblocks/src/service.rs
```diff
@@ -216,16 +214,16 @@ where
             }
         }
 
-        if !new_flashblock && this.current.is_none() {
-            // no new flashbblocks received since, block is still unchanged
-            return Poll::Pending
+        if !this.rebuild && this.current.is_some() {
+            return Poll::Pending;
```
mattsse, line 218: this should hold now, 
we set on fb and unset after building

mattsse (approved):

PR #18226, opened 2025-09-02
fix(optimism): Reconnect if ws stream ends in `WsFlashBlockStream`

RomanHodulak:
Part of #17858 

Resets the `WsFlashBlockStream` to its initial state if the websocket stream cuts off.

### DIFF
--- a/crates/optimism/flashblocks/src/ws/stream.rs
+++ b/crates/optimism/flashblocks/src/ws/stream.rs
@@ -72,7 +72,7 @@ where
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.get_mut();
 
-        loop {
+        'start: loop {
             if this.state == State::Initial {
                 this.connect();
             }
@@ -104,7 +104,9 @@ where
                     .expect("Stream state should be unreachable without stream")
                     .poll_next_unpin(cx))
                 else {
-                    return Poll::Ready(None);
+                    this.state = State::Initial;
+
+                    continue 'start;
                 };
 
                 match msg {
@@ -244,6 +246,14 @@ mod tests {
     #[derive(Default)]
     struct FakeStream(Vec<Result<Message, Error>>);
 
+    impl FakeStream {
+        fn new(mut messages: Vec<Result<Message, Error>>) -> Self {
+            messages.reverse();
+
+            Self(messages)
+        }
+    }
+
     impl Clone for FakeStream {
         fn clone(&self) -> Self {
             Self(
@@ -353,7 +363,7 @@ mod tests {
 
     impl<T: IntoIterator<Item = Result<Message, Error>>> From<T> for FakeConnector {
         fn from(value: T) -> Self {
-            Self(FakeStream(value.into_iter().collect()))
+            Self(FakeStream::new(value.into_iter().collect()))
         }
     }
 
@@ -371,7 +381,7 @@ mod tests {
 
     impl<T: IntoIterator<Item = Result<Message, Error>>> From<T> for FakeConnectorWithSink {
         fn from(value: T) -> Self {
-            Self(FakeStream(value.into_iter().collect()))
+            Self(FakeStream::new(value.into_iter().collect()))
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -63,40 +63,45 @@ where
...
+                    }
+                    Some(Ok(msg)) => {
+                        debug!("Received unexpected message: {:?}", msg);
+                    }
+                    Some(Err(err)) => {
+                        self.state = State::Initial;
+
+                        return Poll::Ready(Some(Err(err.into())))
+                    }
+                    None => return Poll::Ready(None),
```
mattsse: doesnt this still terminate the stream itself if the websocket stream terminates?

mattsse (changes requested):

crates/optimism/flashblocks/src/ws/stream.rs
RomanHodulak: The expectation is that an error is returned right before the stream returns `None`. I think that should hold because there is even an error for normal connection termination.

```rust
    /// WebSocket connection closed normally. This informs you of the close.
    /// It's not an error as such and nothing wrong happened.
    ///
    /// This is returned as soon as the close handshake is finished (we have both sent and
    /// received a close frame) on the server end and as soon as the server has closed the
    /// underlying connection if this endpoint is a client.
    ///
    /// Thus when you receive this, it is safe to drop the underlying connection.
    ///
    /// Receiving this error means that the WebSocket object is not usable anymore and the
    /// only meaningful action with it is dropping it.
    #[error("Connection closed normally")]
    ConnectionClosed,
```

mattsse: unsure about this because this is variant is treated as None

https://github.com/snapview/tokio-tungstenite/blob/25b544e43fe979bca951f085ee1b66e9c1cc3113/src/lib.rs#L303-L304

so imo we should not close our stream here and reconnect instead

mattsse (changes requested):

crates/optimism/flashblocks/src/ws/stream.rs
RomanHodulak: I see.

I think we should receive `Message::Close` before that happens though, which is currently treated as unexpected message. In that regards, we should also send `Message::Close` back according to the [spec](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1). 

If the connection was closed normally, do we want to reconnect?

RomanHodulak: Updated

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -98,22 +98,19 @@ where
...
 
-                let Some(msg) = ready!(this
+                match ready!(this
                     .stream
                     .as_mut()
                     .expect("Stream state should be unreachable without stream")
                     .poll_next_unpin(cx))
-                else {
-                    return Poll::Ready(None);
-                };
```
mattsse: now that we have the outer loop we can keep it like this and instead do

```
else {
  // resetting terminated stream
  this.state = State::Initial;
  continue 'outer
}
```

this way we can avoid the additional Some() wrappers

mattsse (changes requested):
smol style suggestion

crates/optimism/flashblocks/src/ws/stream.rs
RomanHodulak: The match statement felt tidier

crates/optimism/flashblocks/src/ws/stream.rs
```diff
@@ -72,51 +72,51 @@ where
...
-                    Ok(msg) => debug!("Received unexpected message: {:?}", msg),
-                    Err(err) => return Poll::Ready(Some(Err(err.into()))),
-                }
+            let Some(msg) = ready!(this
+                .stream
+                .as_mut()
+                .expect("Stream state should be unreachable without stream")
+                .poll_next_unpin(cx))
+            else {
+                this.connect();
```
mattsse: hmm, I believe this is now problematic because if we remove the outer loop, this won't get polled and we return pending risking that this wont get polled again, so we should keep the outer loop

mattsse (changes requested):

crates/optimism/flashblocks/src/ws/stream.rs
RomanHodulak: That way it might get into an infinite loop if no inner polls return pending. But aren't polls supposed to return as soon as possible?

RomanHodulak: Updated

mattsse (approved):

crates/optimism/flashblocks/src/ws/stream.rs
mattsse: > But aren't polls supposed to return as soon as possible?

only if there's no more to do, the crucial part is

https://docs.rs/futures/latest/futures/prelude/trait.Stream.html#return-value

if we return pending we must ensure that we register the waker, which can either be done manually or by driving all underlying futures/stream to pending

if there are concerns re taking too look then we can consider preemptive yielding
https://tokio.rs/blog/2020-04-preemption

PR #18229, opened 2025-09-02
perf(storage): optimize withdrawal collection creation in test utils

dennsikl:
Replace vec![Withdrawal::default()] with std::iter::once().collect() in 5 test utility functions for better memory efficiency


mattsse (commented):
this doesnt really optimize because this still needs to alloc the same thing

PR #18230, opened 2025-09-02
perf(e2e-test-utils): optimize block checking by fetching header instead of full block

dizer-ti:
Optimizes the `wait_block` function in e2e test utilities by fetching only the block header instead of the full block when checking for block hash validation

- **Before**: `provider.block_by_number(number)?.header().hash_slow()`
- **After**: `provider.header_by_number(number)?.hash_slow()`


mattsse (approved):
this is indeed better

PR #18231, opened 2025-09-02
feat(optimism): Keep track of all current and future flashblock sequences

julio4:
A bit experimental.
Extend current `FlashblockSequence` to also store any flashblocks received for future processing.
Might be better to extract this new logic in a separate structure that contains multiple `FlashblockSequence`, but BTreeMap haves nice ordering.

Currently still cleared by receiving a flashblock 0 which is not ideal (in the case of receiving flashblock 0 early compared to canonical head), but I believe we can select over the two canon receiver and flashblock receiver to automatically clear current sequence and move forward to next one only when canon update while keeping any future flashblocks ahead (and not clear directly on flashblock 0).

This is mostly edge cases and not all are covered. @RomanHodulak What do you think?

### DIFF
--- a/crates/optimism/flashblocks/src/sequence.rs
+++ b/crates/optimism/flashblocks/src/sequence.rs
@@ -1,20 +1,22 @@
 use crate::{ExecutionPayloadBaseV1, FlashBlock};
 use alloy_eips::eip2718::WithEncoded;
+use alloy_primitives::BlockNumber;
 use reth_primitives_traits::{Recovered, SignedTransaction};
-use std::collections::BTreeMap;
+use std::{cmp::Ordering, collections::BTreeMap};
 use tracing::trace;
 
-/// An ordered B-tree keeping the track of a sequence of [`FlashBlock`]s by their indices.
+/// An ordered B-tree keeping the track of sequences of [`FlashBlock`]s ordered by their block
+/// number and indices. The 'current' sequence is the earliest one.
 #[derive(Debug)]
-pub(crate) struct FlashBlockSequence<T> {
-    /// tracks the individual flashblocks in order
+pub(crate) struct FlashBlockSequences<T> {
+    /// Tracks the individual flashblocks in order of their block number and index.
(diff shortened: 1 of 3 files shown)


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// An ordered B-tree keeping the track of sequences of [`FlashBlock`]s ordered by their block
+/// number and indices. The 'current' sequence is the earliest one.
+#[derive(Debug)]
+pub(crate) struct FlashBlockSequences<T> {
+    /// Tracks the individual flashblocks in order of their block number and index.
+    ///
+    /// With a blocktime of 2s and flashblock tick-rate of 200ms plus one extra flashblock per new
+    /// pending block, we expect 11 flashblocks per slot.
+    inner: BTreeMap<(BlockNumber, u64), PreparedFlashBlock<T>>,
```
mattsse, line 16: Is the expectation here that the remote might not send us flashblocks in correct order?

mattsse (changes requested):
needs rebase first

this feature is definitely use but I'd like to solve this slightly differently, basically a broadcast channel in the sequence that emits all the complete sequence of flashblocks, this will be come quite useful for other features we have in mind,

e.g. syncing the chain just via flashblocks subscription

crates/optimism/flashblocks/src/sequence.rs
julio4: Not an expectation but would be nice to be robust against it.
Mostly at the specific timing around a new block, where you can receive last/new flashblocks before/after FCU potentially. Also taking into account potential experiment with flashblock design, but this is not for the current time being.

mattsse (commented):
I think this is a bit too complex for now, especially if we can assume (for now) that we receive flashblocks in order

@julio4 I'm opening an issue for the channel/subscription idea that I have instead

PR #18238, opened 2025-09-02
fix(optimism): Compare parent hash and latest hash to invalidate cached flashblock

### LINKED ISSUE #18233: Don't clear current block if new canonical is parent
### Describe the feature

there's a tiny race condition here

https://github.com/paradigmxyz/reth/blob/358b61b4efe75a058cd96395d7e593978016958a/crates/optimism/flashblocks/src/service.rs#L207-L214

e.g.

1. we receive new flashblock
https://github.com/paradigmxyz/reth/blob/358b61b4efe75a058cd96395d7e593978016958a/crates/optimism/flashblocks/src/service.rs#L195-L195

2. canonical stream does _not_ tick
https://github.com/paradigmxyz/reth/blob/358b61b4efe75a058cd96395d7e593978016958a/crates/optimism/flashblocks/src/service.rs#L207-L207

3. but chain was just advanced so that the execute attempt now sets a new block 

https://github.com/paradigmxyz/reth/blob/358b61b4efe75a058cd96395d7e593978016958a/crates/optimism/flashblocks/src/service.rs#L222-L228

4. canonical stream now ticks and clears the valid block

## TODO:
* only remove the current if it is not on latest == fb.parent_hash

### Additional context

_No response_

RomanHodulak:
Part of #17858
Closes #18233

The `current` pending block is valid if the latest and pending parent block hash match.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -2,7 +2,9 @@ use crate::{sequence::FlashBlockSequence, ExecutionPayloadBaseV1, FlashBlock};
 use alloy_eips::BlockNumberOrTag;
 use alloy_primitives::B256;
 use futures_util::{FutureExt, Stream, StreamExt};
-use reth_chain_state::{CanonStateNotifications, CanonStateSubscriptions, ExecutedBlock};
+use reth_chain_state::{
+    CanonStateNotification, CanonStateNotifications, CanonStateSubscriptions, ExecutedBlock,
+};
 use reth_errors::RethError;
 use reth_evm::{
     execute::{BlockBuilder, BlockBuilderOutcome},
@@ -167,6 +169,12 @@ where
             },
         )))
     }
+
+    /// Takes out `current` [`PendingBlock`] if `state` is not preceding it.
+    fn on_new_tip(&mut self, state: CanonStateNotification<N>) -> Option<PendingBlock<N>> {
+        let latest = state.tip_checked()?.hash();
+        self.current.take_if(|current| current.parent_hash() != latest)
+    }
 }
 
 impl<N, S, EvmConfig, Provider> Stream for FlashBlockService<N, S, EvmConfig, Provider>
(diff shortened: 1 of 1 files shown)


crates/optimism/flashblocks/src/service.rs
```diff
@@ -203,13 +205,31 @@ where
...
+                            |(parent, latest)| {
+                                if parent == latest {
+                                    None
+                                } else {
+                                    this.current.take()
+                                }
+                            },
+                        )
+                })
+            }) {
```
mattsse: I find this a bit hard to read, perhaps we extract this this into a function that returns `Option<PendingBlock>` then I think we can could replace some of this with `?`

mattsse (changes requested):
smol suggestion

crates/optimism/flashblocks/src/service.rs
RomanHodulak: @mattsse updated

crates/optimism/flashblocks/src/service.rs
```diff
@@ -167,6 +169,27 @@ where
             },
         )))
     }
+
+    fn verify_canon_integrity(&mut self, cx: &mut Context<'_>) -> Option<PendingBlock<N>> {
```
mattsse: can we find a better name for this and some docs?

i feel like it would be easier to read if this just takes in the new tip hash which would then simplify this check

and we keep polling the the receiver in the poll_next fn and then pipe he tip to `fn on_new_tip(num_hash: BlockNumHash) -> Option<P>` for example

mattsse (changes requested):
more pedantic style nits

crates/optimism/flashblocks/src/service.rs
RomanHodulak: @mattsse updated

crates/optimism/flashblocks/src/service.rs
```diff
@@ -167,6 +172,22 @@ where
...
+    /// Takes out `current` [`PendingBlock`] if `state` is not preceding it.
+    fn on_new_tip<E: Display>(
+        &mut self,
+        state: Result<CanonStateNotification<N>, E>,
+    ) -> Option<PendingBlock<N>> {
+        let state = state.inspect_err(|e| tracing::error!("{e}")).ok()?;
+        let parent = self.current.as_ref().map(PendingBlock::parent_hash)?;
+        let latest = state.tip_checked().map(RecoveredBlock::hash)?;
+
+        if parent == latest {
```
mattsse: this can just be `self.current.as_ref()?.parent_has() == state.tip_checked()?.hash()`

crates/optimism/flashblocks/src/service.rs
```diff
@@ -199,22 +220,24 @@ where
             }
         }
 
-        // advance new canonical message, if any to reset flashblock
-        {
+        if let Poll::Ready(state) = {
```
mattsse: we can use `Ok(state)` here, the error variant this returns here is just for indicated if the broadcast stream lagged, which will never be the case so we can simply ignore this case

crates/optimism/flashblocks/src/service.rs
```diff
@@ -167,6 +172,22 @@ where
             },
         )))
     }
+
+    /// Takes out `current` [`PendingBlock`] if `state` is not preceding it.
+    fn on_new_tip<E: Display>(
+        &mut self,
+        state: Result<CanonStateNotification<N>, E>,
```
mattsse: we can fully ignore the error case here because this stream will never lag and even if, we'll just skip to newest item

mattsse (changes requested):
last pedantic nits

mattsse (approved):
cool

PR #18242, opened 2025-09-03
feat(cli): enable traces export via `tracing-otlp` cli arg

### LINKED ISSUE #18241: Add cli flag for enabling otlp
### Describe the feature

We have otlp support via `tracing-otlp`:
https://github.com/paradigmxyz/reth/blob/main/crates/tracing-otlp/src/lib.rs

It would be nice to be able to enable otlp via cli flag, like `--metrics.otlp`

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/18241.

Add new CLI flag `tracing-otlp` to export opentelemetry spans either to stdout or to an OTLP HTTP endpoint.
- **Stdout export**: `--tracing-otlp`
- **OTLP endpoint export**: `--tracing-otlp=http://localhost:4318/v1/traces`

Here how I tested : 

Example of very simple otel-collector configuration's file, limited for traces export : 
```yaml
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug]
```
and then I put up the collector using this config : 

```bash
docker run -p 4318:4318 \
  -v $(pwd)/test-otlp.yaml:/etc/otel-collector-config.yaml \
  otel/opentelemetry-collector:latest \
  --config=/etc/otel-collector-config.yaml
```

The build for the node should be done with `otlp` feature for example `cargo build --release -p op-reth --features reth-optimism-cli/otlp`.

and then `./target/release/op-reth --tracing-otlp=http://localhost:4318/v1/traces node --dev`

or 
```shell
cargo run --release -p op-reth --features reth-optimism-cli/otlp -- --tracing-otlp=http://localhost:4318/v1/traces node --dev
```

Depending on the type provided to `--tracing-otlp`, either it will just print the spans, either you can see them being redirected to the otlp-collector.
cc @Rjected


crates/node/builder/src/launch/common.rs
```diff
@@ -620,6 +621,28 @@ where
...
+            info!(target: "reth::cli", "Starting metrics OTLP export to {}", listen_addr);
+
+            let exporter_url = format!("http://{listen_addr}");
+            // Start registry with metrics exporting
+            //TODO: check how much it interferes with the current tracing registry,
+            //ideally it should be merged with the current tracing registry
+            init_metrics_tracing("reth", exporter_url)?;
+        }
+        Ok(())
+    }
```
lean-apple: Not found yet the way to add properly the metric tracing layer in the tracing init

mattsse: maybe @shekhirin can help here

crates/tracing-otlp/Cargo.toml
```diff
@@ -9,13 +9,17 @@ repository.workspace = true
...
-opentelemetry_sdk = "0.29.0"
-opentelemetry = "0.29.1"
-opentelemetry-otlp = "0.29.0"
-tracing-opentelemetry = "0.30.0"
+# obs
+opentelemetry_sdk = "0.30"
+opentelemetry = "0.30"
+opentelemetry-otlp = "0.30"
+opentelemetry-semantic-conventions = "0.30"
+tracing-opentelemetry = "0.31"
```
mattsse: I'd like to move these to workspace level then we have versioned deps in one location

mattsse (dismissed):
this should be super useful

we maybe also want to add this to rpc as well?

i dont know much about otlp, maybe @onbjerg could also take a look here?

crates/node/core/src/args/metric.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Telemetry exporters configuration.
+///
+/// - Prometheus: serves metrics over http.
+/// - OTLP: pushes metrics to an `OpenTelemetry` collector via http.
+#[derive(Debug, Clone, Default, Parser)]
+pub struct MetricsArgs {
+    /// Enable Prometheus metrics.
+    ///
+    /// The metrics will be served at the given interface and port.
+    #[arg(long, alias = "metrics.prometheus", value_name = "PROMETHEUS", value_parser = parse_socket_address, help_heading = "Metrics")]
```
shekhirin: would be nice to not break existing setups due to changing the argument name, can we preserve backward compatibility with `--metrics` argument?

shekhirin:
@lean-apple is this ready for review and can be moved out of a draft?

crates/tracing-otlp/src/lib.rs
```diff
@@ -5,34 +5,63 @@
...
 }
+
+/// Creates an `OpenTelemetry` metrics provider that exports metrics to a default OTLP endpoint.
+pub fn create_metrics_provider(
+    service_name: impl Into<Value>,
+    endpoint: &str,
+) -> eyre::Result<SdkMeterProvider> {
+    let resource = build_resource(service_name);
+
+    // Metrics will be sent to the OTLP endpoint through http (and not grpc, and  with no tls)
```
lean-apple: If we want to go further in the configuration it requires more arg like certificate path for tls eg.

Rjected (commented):
I think we would need to set things up here:
https://github.com/paradigmxyz/reth/blob/85f0dbf263626d484f4f21b3cb0d7903493c4457/crates/ethereum/cli/src/app.rs#L109-L119

instead of the current `init_tracing`, `Cli::init_tracing` does not seem to actually be used.

for OP, the function called is:
https://github.com/paradigmxyz/reth/blob/85f0dbf263626d484f4f21b3cb0d7903493c4457/crates/optimism/cli/src/app.rs#L114-L124

Rjected:
> but not sure how to fix the wasm issue (maybe adding a flag feature ?)

yeah I think a feature flag makes sense for this

> if we should already update the observability/metrics documentations section

I think these can be followups after some testing, since using this does require some extra infra to be set up

> Also the otlp-exporter only expects a basic http endpoint, I think it can be interesting to go further with grpc and enabling tls also.

makes sense, also would be good followup issues

Rjected:
@lean-apple bump on this, this would be great to have in reth

Rjected:
> > > > @lean-apple bump on this, this would be great to have in reth
> > > 
> > > 
> > > @Rjected I have tested it a bit more, with a local otlp-collector, and it seems it requires more work to do the conversion between the current `metrics` crate and the opentelemetry one, we may need a crate like this [Noelware/metrics-exporter-opentelemetry](https://github.com/Noelware/metrics-exporter-opentelemetry?rgh-link-date=2025-10-06T14%3A32%3A48Z).
> > 
> > 
> > what would we need the metrics converter for? I was imagining that we use `--metrics.otlp` for tracking spans etc, and we still use `--metrics` / we keep using prometheus for metrics
> 
> Ah ok I see, I think I was a bit confused by the term `metrics`, as spans are usually more under the umbrella of traces than metrics in the opentelemetry terminology. Most of the code was already implemented for the span so.
> 
> Maybe we can add a `traces-otlp` arg so ?

Sure, I think `--tracing-otlp` would be nice

crates/tracing-otlp/src/lib.rs
```diff
@@ -1,38 +1,72 @@
...
 
-    let resource = Resource::builder()
-        .with_service_name(service_name)
-        .with_schema_url([KeyValue::new(SERVICE_VERSION, env!("CARGO_PKG_VERSION"))], SCHEMA_URL)
-        .build();
+    let mut provider_builder =
+        SdkTracerProvider::builder().with_resource(resource).with_sampler(Sampler::AlwaysOn);
+
+    provider_builder = if let TraceOutput::Otlp(url) = output {
+        println!("Exporting traces to OTLP endpoint: {}", url);
```
Rjected: Should remove this println

crates/optimism/cli/src/app.rs
```diff
@@ -116,9 +116,16 @@ where
...
+
+            #[cfg(feature = "otlp")]
+            if let Some(output_type) = &self.cli.traces.otlp {
+                info!(target: "reth::op::cli", "Starting OTLP tracing export to {:?}", output_type);
+                layers.with_span_layer("reth::op::cli".to_string(), output_type.clone())?;
+            }
+
             self.guard = self.cli.logs.init_tracing_with_layers(layers)?;
-            info!(target: "reth::cli", "Initialized tracing, debug log directory: {}", self.cli.logs.log_file_directory);
+            info!(target: "reth::op::cli", "Initialized tracing, debug log directory: {}", self.cli.logs.log_file_directory);
```
Rjected: currently we use `reth::cli` in a lot of places in `op-reth`, so we should probably keep this as `reth::cli` for consistency

crates/ethereum/cli/src/app.rs
```diff
@@ -111,7 +111,13 @@ where
     /// If file logging is enabled, this function stores guard to the struct.
     pub fn init_tracing(&mut self) -> Result<()> {
         if self.guard.is_none() {
-            let layers = self.layers.take().unwrap_or_default();
+            let mut layers = self.layers.take().unwrap_or_default();
+
+            #[cfg(feature = "otlp")]
+            if let Some(output_type) = &self.cli.traces.otlp {
```
Rjected, line 117: let's add a similar log about initializing otlp here?

crates/node/core/src/args/trace.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    // OTLP specification requires the `/v1/traces` path for trace endpoints
+    ensure!(
+        url.path().ends_with("/v1/traces"),
+        "OTLP trace endpoint must end with /v1/traces, got path: {}",
+        url.path()
+    );
+
+    Ok(TraceOutput::Otlp(url))
+}
```
Rjected: Maybe we can make this a method on `TraceOutput`? Like `TraceOutput::parse`?

crates/tracing/src/layers.rs
```diff
@@ -123,6 +123,22 @@ impl Layers {
...
+
+    /// Add OTLP spans layer to the layer collection
+    #[cfg(feature = "otlp")]
+    pub fn with_span_layer(
+        &mut self,
+        service_name: String,
+        output_type: reth_tracing_otlp::TraceOutput,
+    ) -> eyre::Result<()> {
+        // Create the span provider
+        let span_layer = span_layer(service_name, output_type)
```
Rjected: for consistency let's just use a direct import for `span_layer`

crates/tracing-otlp/src/lib.rs
```diff
@@ -1,38 +1,72 @@
...
-    let exporter = SpanExporter::builder().with_http().build().unwrap();
+    global::set_text_map_propagator(TraceContextPropagator::new());
+    let resource = build_resource(service_name);
 
-    let resource = Resource::builder()
-        .with_service_name(service_name)
-        .with_schema_url([KeyValue::new(SERVICE_VERSION, env!("CARGO_PKG_VERSION"))], SCHEMA_URL)
-        .build();
+    let mut provider_builder =
+        SdkTracerProvider::builder().with_resource(resource).with_sampler(Sampler::AlwaysOn);
```
Rjected: does this still work if we don't specify `AlwaysOn`? The default `Sampler::ParentBased(Box::new(Sampler::AlwaysOn))` makes sense to me conceptually

crates/tracing-otlp/src/lib.rs
```diff
@@ -1,38 +1,72 @@
...
+
+    provider_builder = if let TraceOutput::Otlp(url) = output {
+        println!("Exporting traces to OTLP endpoint: {}", url);
+        let span_exporter =
+            SpanExporter::builder().with_http().with_endpoint(url.to_string()).build()?;
+        provider_builder.with_batch_exporter(span_exporter)
+    } else {
+        let stdout_exporter = opentelemetry_stdout::SpanExporter::default();
+        provider_builder.with_simple_exporter(stdout_exporter)
+    };
```
Rjected: I think we should require the batch exporter and not support `opentelemetry_stdout`, just because the batch exporter will be the most common usage

Rjected (changes requested):
Awesome, just some final comments, will try this out later

ty for the otel config and steps for running in docker!

crates/node/core/src/args/trace.rs
lean-apple: Definitely.

crates/tracing-otlp/src/lib.rs
lean-apple: Ah I forgot to remove it, as it's not possible to have debug trace, the registry is not initiliazed yet.

crates/tracing-otlp/src/lib.rs
lean-apple: Yes it should work, but I think I wanted to let it to remind us we can do something more configurable in another pr instead of the default, 

sthg like this for example, with the `sampling_ratio` that can be configurable through the trace cli arg,

```rust
let sampling_ratio : f64 = 0.1;
let sampler = Sampler::TraceIdRatioBased(sampling_ratio);
let mut provider_builder = SdkTracerProvider::builder().with_resource(resource).with_sampler(sampler);
```

crates/tracing-otlp/src/lib.rs
lean-apple: Totally agree it is probably the most common usage but I wanted to let the choice to user; 
Yet it would make the code much cleaner also to remove it, we dont need the `TraceOutput` type so and we can just use a string or the socket address type for the `tracing-otlp` arg.

lean-apple:
Thanks for your review @Rjected :), Addressed most of the comments, open to remove the stdout option if you think it would be useless to keep it. 

Let me know if you could try and have some other configuration idea/fix in mind!

crates/tracing-otlp/src/lib.rs
Rjected: yeah I think let's remove it

crates/tracing-otlp/src/lib.rs
Rjected: makes sense, can add a followup issue for changing the sampler

Rjected (approved):
thanks! LGTM

shekhirin (approved):

crates/ethereum/cli/src/app.rs
```diff
@@ -111,7 +111,14 @@ where
...
+                info!(target: "reth::cli", "Starting OTLP tracing export to {:?}", output_type);
+                layers.with_span_layer("reth::cli".to_string(), output_type.clone())?;
```
Zygimantass: I think this might cause the exporter to only export `reth::cli` related spans, instead of exporting everything globally

crates/tracing/src/layers.rs
```diff
@@ -123,6 +124,22 @@ impl Layers {
...
+    pub fn with_span_layer(
+        &mut self,
+        service_name: String,
+        endpoint_exporter: Url,
+    ) -> eyre::Result<()> {
+        // Create the span provider
+        let span_layer = span_layer(service_name, &endpoint_exporter)
+            .map_err(|e| eyre::eyre!("Failed to build OTLP span exporter {}", e))?;
+
+        self.add_layer(span_layer);
```
Zygimantass, line 142: adding an explicit filter that always defaults to `TRACE` yields better results in terms of amount of traces. this should probably be the sane default and users can override it if they want less traces?

crates/ethereum/cli/src/app.rs
Rjected: just changed this to `reth`, not super confident on this but I think this just determines the service name?

crates/tracing/src/layers.rs
Rjected: Set this to trace, override option can be for a followup

lean-apple: Done very fastly in https://github.com/paradigmxyz/reth/issues/18980.

crates/ethereum/cli/src/app.rs
lean-apple: Yes it's the `service-name`,  maybe we can this also as configurable, with `reth` by default

PR #18247, opened 2025-09-03
refactor(optimism): Extract pending block building responsibility out of `FlashBlockService`

RomanHodulak:
Part of #17858 
Towards #18219

The `FlashBlockService` has too many responsibilities and it's becoming too tangled up. Moreover, to spawn blocking tasks for block building we're going to need an independent unit that performs only these jobs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9339,6 +9339,7 @@ dependencies = [
  "reth-rpc-eth-api",
  "reth-rpc-eth-types",
  "reth-storage-api",
+ "reth-tasks",
  "serde",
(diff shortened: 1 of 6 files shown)


mattsse (changes requested):
cool, extracting this logic makes a lot of sense,

what's still missing now actually spawn the `.execute` part blocking and send back the response through a channel the we poll in the service together with the cached reads

there's some additional movement in progress with these files so we can hold off here a bit and get perhaps get #18272 and #18276 in first?

mattsse (approved):

PR #18257, opened 2025-09-03
feat(optimism): Decode text messages in `WsFlashBlockStream`

RomanHodulak:
Part of #17858 

From the base implementation, we expect messages to be binary.

The messages are either a JSON string or a brotli compressed blob.

Since the messages can be JSON, it is reasonable to expect them being UTF-8 encoded.

This or some other server implementation might decide to send the flashblock as text message and it would be an unnecessary failure to ignore it in that case.


mattsse (approved):
makes sense

PR #18258, opened 2025-09-04
chore(txpool): add sanity tests for blob fee bit handling

yongkangc:
## Summary

Adds targeted sanity check tests to prevent regression of the critical bug from PR #18215 where non-EIP4844 transactions got stuck in BaseFee pool due to incorrect `ENOUGH_BLOB_FEE_CAP_BLOCK` bit removal.

## Background

PR #18215 exposed a bug as non-4844 txns had their the `ENOUGH_BLOB_FEE_CAP_BLOCK` bit removed from all transactions and ended up in the base fee pool. PR #18218 reverted the change but lacked regression tests.

### 1. `test_non_4844_blob_fee_bit_invariant`
Verifies the core invariant: non-EIP4844 transactions always have `ENOUGH_BLOB_FEE_CAP_BLOCK` bit set and reach Pending pool.

### 2. `test_blob_fee_enforcement_only_applies_to_eip4844`
Tests that blob fee enforcement only affects EIP-4844 transactions:
- EIP-4844 transactions: blob fee bit removed when blob fee insufficient → Blob pool
- Non-4844 transactions: blob fee bit always remains true → Pending pool

### 3. `test_basefee_decrease_preserves_non_4844_blob_fee_bit` 
**Critical test** that reproduces the exact PR #18215 bug scenario:
- Non-4844 transaction starts in BaseFee pool (high basefee)  
- Retains `ENOUGH_BLOB_FEE_CAP_BLOCK` bit even in BaseFee pool
- When basefee decreases, gets promoted to Pending pool
- **Critical**: `ENOUGH_BLOB_FEE_CAP_BLOCK` bit must be preserved during promotion

This test would have caught the original bug where the blob fee bit was incorrectly removed during basefee enforcement.

## Core Rule Enforced

**Only EIP-4844 transactions should have conditional blob fee bit logic. Non-4844 transactions must always have `ENOUGH_BLOB_FEE_CAP_BLOCK = true`.**

These targeted tests prevent the critical bug from PR #18215 from reoccurring and ensure regular Ethereum transactions always work properly in Reth's transaction pool.


mattsse (approved):
sweet

PR #18262, opened 2025-09-04
chore: remove redundant payload trait bounds

hai-rise:
Was unnecessarily strict for downstream implementations!


mattsse (approved):
seems okay, you can add those in where needed

PR #18263, opened 2025-09-04
fix: check prune checkpoints for unwind target limit

### LINKED ISSUE #16579: Prevent unwind and log when manual unwind is attempted past pruning config
### Describe the feature

Right now if you attempt to unwind past the account / storage history limit, for example the one configured here:
https://github.com/paradigmxyz/reth/blob/672d97307ae07ce01c9a4a571d09f837d540cd7c/crates/node/core/src/args/pruning.rs#L105-L106

This will result in an error in the merkle stage. It makes sense that the merkle unwind would fail, because we don't have enough history. However, we should instead return an error and add a log for the user before running _any_ stages if we find that we don't have enough history and can't unwind to the desired block

### Additional context

_No response_

mattsse:
followup to #16579

the initial fix https://github.com/paradigmxyz/reth/issues/16579

only took the settings into account, but we should also check the actual disk representation here.

e.g.:

sync archinve,
change settings to full
unwind -> would previously fail because prune settings now enforce the history distance

now: heck the pruned height of those two segments

### DIFF
--- a/crates/prune/types/src/segment.rs
+++ b/crates/prune/types/src/segment.rs
@@ -42,6 +42,16 @@ impl PruneSegment {
             Self::Receipts => MINIMUM_PRUNING_DISTANCE,
         }
     }
+
+    /// Returns true if this is [`Self::AccountHistory`].
+    pub const fn is_account_history(&self) -> bool {
+        matches!(self, Self::AccountHistory)
+    }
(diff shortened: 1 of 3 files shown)


Rjected (approved):
this makes sense to me, nice

PR #18264, opened 2025-09-04
WIP: patch alloy deps to v1.0.28 and document version conflicts

parzival1821:
## Draft: Support both alloy v1.0.30 and v1.0.31 

**Progress so far:**
- [x] Patched alloy deps to v1.0.28 using git patch
- [x] Identified first version conflict: `alloy_eips::eip7685::Requests` 
- [ ] Document all breaking changes
- [ ] Implement package renaming solution (like rand_08/rand_09)
- [ ] Ensure compilation works with both versions

**Current Issue:**
Two versions of alloy_eips are in dependency tree causing type conflicts.

**Next Steps:**
- Continue compilation to find more conflicts (especially secp256k1)
- Implement explicit package renaming solution

Fixes #18249
Ref: alloy-rs/alloy#2839

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,12 +116,12 @@ version = "1.0.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b7345077623aaa080fc06735ac13b8fa335125c8550f9c4f64135a5bf6f79967"
 dependencies = [
- "alloy-eips",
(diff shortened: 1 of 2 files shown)


mattsse (commented):
this was useful, but we're actually bottlnecked somewhere else

we still need k256 release before we can even attempt this

ty for looking into this

PR #18267, opened 2025-09-04
refactor: change PendingPool and PendingTransaction visibility to pub

kien-rise:
In RISE, we are customizing `BestTransactions`. To enable this, certain fields must be made `pub`. This PR updates their visibility accordingly.


mattsse (changes requested):
not a fan of making them pub because with these things that's make it a lot harder for me to reason about access

I'm okay with making the types pub and add getters if necessary

mattsse (approved):

PR #18268, opened 2025-09-04
test(optimism): Test that close message is responded to in `WsFlashBlockStream`

RomanHodulak:
Part of #17858 

Increases the test coverage of the test scenario where a message is sent expecting a response by the `Close` message.

The close message expects a response according to the [websocket spec](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1).


mattsse (approved):
lgtm

PR #18274, opened 2025-09-04
ci: add rv64imac-unknown-none-elf target check

### LINKED ISSUE #18266: Add CI check for rv64imac-unknown-none-elf
### Describe the feature

This is a copy of this issue in revm: https://github.com/bluealloy/revm/issues/2917

I imagine that this check would also be changed in alloy.

In reth, the target is being checked with the following script: https://github.com/paradigmxyz/reth/blob/main/.github/assets/check_rv32imac.sh

### Additional context

I wouldn't expect this to break anything, unless as noted [here](https://github.com/bluealloy/revm/pull/2918#issuecomment-3223570661) a dependency is matching on targets

dennsikl:
Add CI check for RISC-V 64-bit architecture (rv64imac-unknown-none-elf) to ensure crate compatibility across both 32-bit and 64-bit RISC-V targets.

Closes [#18266](https://github.com/paradigmxyz/reth/issues/18266)


.github/assets/check_rv64imac.sh
```diff
@@ -0,0 +1,88 @@
...
+    reth-optimism-evm
+)
+
+# Array to hold the results
+results=()
+# Flag to track if any command fails
+any_failed=0
+
+for crate in "${crates_to_check[@]}"; do
+  cmd="cargo +stable build -p $crate --target riscv64imac-unknown-none-elf --no-default-features"
```
kevaundray: I would try to avoid duplicating check_rv32imac.sh because any change to check_rv32imac.sh would imply a change in this file.

The only difference between both files is the target, so we could take that in as a parameter and have one file maybe called `check_target.sh` which gets called like `.github/assets/check_target.sh riscv64imac-unknown-none-elf`

mattsse (commented):
@kevaundray is this good to merge?

this seems okay but want to make sure

PR #18275, opened 2025-09-04
fix(e2e): persist accepted header in CheckPayloadAccepted and align timestamp

Snezhkko:
Update CheckPayloadAccepted to call env.set_current_block_info(...) when a node accepts the broadcasted payload. This ensures Environment.current_block_info stays consistent with the canonical head. Also align latest_header_time to the RPC header timestamp (canonical view) and update forkchoice head. Without this, the env could remain stale despite the comment “save the header in Env,” causing downstream actions relying on current_block_info to read outdated state.

### DIFF
--- a/crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
+++ b/crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
@@ -590,7 +590,7 @@ where
                 // at least one client passes all the check, save the header in Env
                 if !accepted_check {
                     accepted_check = true;
-                    // save the header in Env
+                    // save the current block info in Env
                     env.set_current_block_info(BlockInfo {
                         hash: rpc_latest_header.hash,
                         number: rpc_latest_header.inner.number,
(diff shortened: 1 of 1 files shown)


crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
```diff
@@ -591,11 +591,19 @@ where
                 if !accepted_check {
                     accepted_check = true;
                     // save the header in Env
-                    env.active_node_state_mut()?.latest_header_time = next_new_payload.timestamp;
+                    env.set_current_block_info(BlockInfo {
+                        hash: rpc_latest_header.hash,
+                        number: rpc_latest_header.inner.number,
+                        timestamp: rpc_latest_header.inner.timestamp,
+                    })?;
```
mattsse, line 598: hmm, this kinda looks right

mattsse (commented):
wdyt here @fgimenez

crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
```diff
@@ -591,11 +591,20 @@ where
                 if !accepted_check {
                     accepted_check = true;
                     // save the header in Env
```
fgimenez:
```suggestion
                    // save the current block info in Env
```

crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
fgimenez: yep lgtm!

fgimenez (approved):
thanks @Snezhkko!

PR #18272, opened 2025-09-04
feat(optimism): flashblock completed sequences

julio4:
part of #18255

Split the current `FlashBlockSequence` in two kinds:
- Pending as `FlashBlockPendingSequence`: used to iteratively build the pending block
- Completed as `FlashBlockCompletedSequence`: will be used towards #18255 (can also be useful in #18244 to define a `broadcast::Receiver<Option< FlashBlockSequence >>`)

PR #18276, opened 2025-09-04
feat(op-reth/flashblocks): subscribe to the flashblock sequences produced

theochap:
## Description

This PR exposes methods to allow broadcasting/subscription to the flashblock sequences produced. Notable changes:
- Renamed `FlashBlockRx` to `PendingBlockRx` and used `FlashBlockRx` as a name for that new channel
- Added a simple test to ensure that we're broadcasting the flashblocks properly.

Close #18244 

Depends (and rebased) on #18272

### DIFF
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -7,6 +7,8 @@ use reth_rpc_eth_types::PendingBlock;
 pub use service::FlashBlockService;
 pub use ws::{WsConnect, WsFlashBlockStream};
 
+use crate::sequence::FlashBlockCompleteSequence;
+
 mod payload;
 mod sequence;
 mod service;
@@ -15,4 +17,10 @@ mod ws;
 /// Receiver of the most recent [`PendingBlock`] built out of [`FlashBlock`]s.
 ///
 /// [`FlashBlock`]: crate::FlashBlock
-pub type FlashBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingBlock<N>>>;
+pub type PendingBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingBlock<N>>>;
+
+/// Receiver of the sequences of [`FlashBlock`]s built.
+///
+/// [`FlashBlock`]: crate::FlashBlock
+pub type FlashBlockCompleteSequenceRx =
+    tokio::sync::broadcast::Receiver<FlashBlockCompleteSequence>;
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -7,8 +7,6 @@ use reth_rpc_eth_types::PendingBlock;
 pub use service::FlashBlockService;
 pub use ws::{WsConnect, WsFlashBlockStream};
 
-use crate::sequence::FlashBlockCompleteSequence;
-
 mod payload;
 mod sequence;
 pub use sequence::FlashBlockCompleteSequence;
--- a/crates/optimism/flashblocks/src/sequence.rs
+++ b/crates/optimism/flashblocks/src/sequence.rs
@@ -4,9 +4,20 @@ use reth_primitives_traits::{Recovered, SignedTransaction};
 use std::collections::BTreeMap;
 use tracing::trace;
 
+pub(crate) trait FlashBlockSequence {
+    /// Returns the first block number
+    fn block_number(&self) -> Option<u64>;
+
+    /// Returns the payload base of the first tracked flashblock.
+    fn payload_base(&self) -> Option<ExecutionPayloadBaseV1>;
+
+    /// Returns the number of tracked flashblocks.
+    fn count(&self) -> usize;
+}
+
 /// An ordered B-tree keeping the track of a sequence of [`FlashBlock`]s by their indices.
 #[derive(Debug)]
-pub(crate) struct FlashBlockSequence<T> {
+pub(crate) struct FlashBlockPendingSequence<T> {
     /// tracks the individual flashblocks in order
     ///
     /// With a blocktime of 2s and flashblock tick-rate of 200ms plus one extra flashblock per new
@@ -14,7 +25,7 @@ pub(crate) struct FlashBlockSequence<T> {
     inner: BTreeMap<u64, PreparedFlashBlock<T>>,
 }
 
-impl<T> FlashBlockSequence<T>
+impl<T> FlashBlockPendingSequence<T>
 where
     T: SignedTransaction,
 {
@@ -45,16 +56,6 @@ where
         Ok(())
     }
 
-    /// Returns the first block number
-    pub(crate) fn block_number(&self) -> Option<u64> {
-        Some(self.inner.values().next()?.block().metadata.block_number)
-    }
-
-    /// Returns the payload base of the first tracked flashblock.
-    pub(crate) fn payload_base(&self) -> Option<ExecutionPayloadBaseV1> {
-        self.inner.values().next()?.block().base.clone()
-    }
-
     /// Iterator over sequence of executable transactions.
     ///
     /// A flashblocks is not ready if there's missing previous flashblocks, i.e. there's a gap in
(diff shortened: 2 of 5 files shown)


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -28,13 +39,26 @@ where
     pub(crate) fn insert(&mut self, flashblock: FlashBlock) -> eyre::Result<()> {
         if flashblock.index == 0 {
             trace!(number=%flashblock.block_number(), "Tracking new flashblock sequence");
-            // Flash block at index zero resets the whole state
-            self.clear();
```
julio4, line 33: Maybe keep the clear + broadcast in a separate method to make insert logic more concise. You can rename `clear` with something like `finalize` or `complete`.

crates/optimism/flashblocks/src/lib.rs
```diff
@@ -15,4 +15,9 @@ mod ws;
 /// Receiver of the most recent [`PendingBlock`] built out of [`FlashBlock`]s.
 ///
 /// [`FlashBlock`]: crate::FlashBlock
-pub type FlashBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingBlock<N>>>;
+pub type PendingBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingBlock<N>>>;
+
+/// Receiver of the sequences of [`FlashBlock`]s built.
+///
+/// [`FlashBlock`]: crate::FlashBlock
+pub type FlashBlockRx = tokio::sync::broadcast::Receiver<Vec<FlashBlock>>;
```
julio4: Maybe wait for #18272 to use `FlashBlockCompleteSequence` which aliases `Vec<FlashBlock>>` and have an implementation of TryFrom for `FlashBlockPendingSequence`

julio4: Also `FlashBlockRx` can be a little misleading as it's sending full sequence of flashblocks, maybe `FlashBlockCompleteSequenceRx` is better.
I agree for `PendingBlockRx`, better naming.

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -12,14 +13,24 @@ pub(crate) struct FlashBlockSequence<T> {
...
 impl<T> FlashBlockSequence<T>
 where
     T: SignedTransaction,
 {
-    pub(crate) const fn new() -> Self {
-        Self { inner: BTreeMap::new() }
+    pub(crate) fn new() -> Self {
+        // Note: if the channel is full, send will not block but rather overwrite the oldest
+        // messages. Order is preserved.
+        let (tx, _) = broadcast::channel(1024);
```
julio4: I believe smaller channel buffer is fine

julio4 (commented):
Made some small comments but overall looks good for me

crates/optimism/flashblocks/src/sequence.rs
emhane: pls make magic number a constant

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -28,13 +39,26 @@ where
     pub(crate) fn insert(&mut self, flashblock: FlashBlock) -> eyre::Result<()> {
         if flashblock.index == 0 {
             trace!(number=%flashblock.block_number(), "Tracking new flashblock sequence");
-            // Flash block at index zero resets the whole state
-            self.clear();
+            // Flash block at index zero resets the whole state.
+            let flashblocks = std::mem::take(&mut self.inner);
```
emhane: we prefer to use `core` and `alloc` where possible over std, for no-std compat. also pls bring `core::mem` into scope in module import prelude.
```suggestion
            let flashblocks = mem::take(&mut self.inner);
```

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -28,13 +39,26 @@ where
...
+
+            // If there are any subscribers, send the flashblocks to them.
+            if self.block_broadcaster.receiver_count() > 0 {
+                let flashblocks = flashblocks.into_iter().map(|block| block.1.into()).collect();
+
+                // Note: this should only ever fail if there are no receivers. This can happen if
+                // there is a race condition between the clause right above and this
+                // one. We can simply warn the user and continue.
+                if let Err(err) = self.block_broadcaster.send(flashblocks) {
+                    warn!("Failed to send flashblocks to subscribers: {}", err);
```
emhane: pls add target to log message, check which target is used for this module/crate already

emhane (approved):
nice! smol comments, and pending @mattsse

crates/optimism/flashblocks/src/lib.rs
theochap: Ok, I will rebase on #18272!

crates/optimism/flashblocks/src/sequence.rs
theochap: No target was used. Added flashblocks target everywhere

crates/optimism/flashblocks/src/lib.rs
theochap: Done and renamed

crates/optimism/flashblocks/src/sequence.rs
theochap: Done, moved out the clear + broadcast logic to a separate method

crates/optimism/flashblocks/src/sequence.rs
theochap: Reduced channel size to 128

theochap: Done

crates/optimism/flashblocks/src/sequence.rs
theochap: Done

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -183,4 +278,50 @@ mod tests {
...
+        // Then we don't receive anything until we insert a new flashblock
+        let no_flashblock = subscriber.try_recv();
+        assert_eq!(no_flashblock, Err(TryRecvError::Empty));
+
+        // Let's insert a new flashblock with index 0
+        sequence
+            .insert(FlashBlock {
+                payload_id: Default::default(),
+                index: 0,
+                base: None,
```
RomanHodulak: This is not really an issue of this PR, but rather #18272

If a complete flash block sequence means that it can be used to build a block, then being able to get it without the `base` does not make sense.

I think the `base` needs to be an invariant which, if not met, forbids the construction of the complete sequence. Otherwise, it would be vulnerable to panics.

crates/optimism/flashblocks/src/service.rs
```diff
@@ -234,15 +243,15 @@ where
...
+                trace!(target: "flashblocks", parent_hash=%new_pending.block().parent_hash(), block_number=new_pending.block().number(), flash_blocks=this.blocks.count(), elapsed=?now.elapsed(), "Built new block with flashblocks");
                 return Poll::Ready(Some(Ok(Some(new_pending))));
             }
             Ok(None) => {
                 // nothing to do because tracked flashblock doesn't attach to latest
             }
             Err(err) => {
                 // we can ignore this error
-                debug!(%err, "failed to execute flashblock");
+                debug!(target: "flashblocks", %err, "failed to execute flashblock");
```
RomanHodulak: With this change, the `target` would be inconsistent throughout the `reth_optimism_flashblocks` crate.

I think it's fine to keep the default.

Then you can reliably use `RUST_LOG=reth_optimism_flashblocks=trace,info`

RomanHodulak (commented):
Left some thoughts.

crates/optimism/flashblocks/src/sequence.rs
julio4: Agree, check the FlashBlockCompleteSequence::new and TryFrom impl from #18272, which try to ensure that.

mattsse (approved):

PR #18280, opened 2025-09-04
perf: optimize send raw batching

mattsse:
removes redundant clone and drops raw bytes early

optimizes the batcher impl via poll_recv_many

### DIFF
--- a/crates/transaction-pool/src/batcher.rs
+++ b/crates/transaction-pool/src/batcher.rs
@@ -10,7 +10,7 @@ use pin_project::pin_project;
 use std::{
     future::Future,
     pin::Pin,
-    task::{Context, Poll},
+    task::{ready, Context, Poll},
 };
 use tokio::sync::{mpsc, oneshot};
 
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/batcher.rs
```diff
@@ -88,21 +89,15 @@ where
...
+            ready!(this.request_rx.poll_recv_many(cx, this.buf, *this.max_batch_size));
 
-            if !batch.is_empty() {
+            if !this.buf.is_empty() {
+                let batch = std::mem::take(this.buf);
                 let pool = this.pool.clone();
                 tokio::spawn(async move {
                     Self::process_batch(&pool, batch).await;
                 });
+                this.buf.reserve(1);
```
Rjected, line 100: i see, the prev loop is basically the same thing as `poll_recv_many`, but we return early on `Pending` now and need the extra reserve to preserve the single capacity

Rjected (approved):

PR #18284, opened 2025-09-05
perf(db): do not heap-allocate the stage key per query

hai-rise:
0.6/19 ~ 3% of `BlockchainProvider::latest` is spent on heap-allocating a `String` for `StageId::Finish` to be used for a database query.

image

This is bad because:
- It is in hot paths like validating new mempool transactions (as bad as creating a `latest` provider per new transaction).
- It scales poorly for nodes under heavy load, where heap allocation is more expensive than normally.

This PR pre-allocates pre-defined stage ids only once, to reference them for database queries, instead of allocating anew per query.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -1642,7 +1642,11 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> BlockBodyIndicesProvider
 
 impl<TX: DbTx, N: NodeTypes> StageCheckpointReader for DatabaseProvider<TX, N> {
     fn get_stage_checkpoint(&self, id: StageId) -> ProviderResult<Option<StageCheckpoint>> {
-        Ok(self.tx.get::<tables::StageCheckpoints>(id.to_string())?)
+        Ok(if let Some(encoded) = id.get_pre_encoded() {
+            self.tx.get_by_encoded_key::<tables::StageCheckpoints>(encoded)?
+        } else {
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1642,7 +1642,11 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> BlockBodyIndicesProvider
 
 impl<TX: DbTx, N: NodeTypes> StageCheckpointReader for DatabaseProvider<TX, N> {
     fn get_stage_checkpoint(&self, id: StageId) -> ProviderResult<Option<StageCheckpoint>> {
-        Ok(self.tx.get::<tables::StageCheckpoints>(id.to_string())?)
```
hai-rise, line 1645: This heap-allocated per database query, even when the hottest queries have "constant" stage ids like "Finish", which should be (pre)allocated only once.

This is also our initial interested hot path, `get_pre_encoded` can be applied to more places -- anywhere that `get`s from checkpoint tables.

mattsse, line 1645: actually, isn't this the same as `.as_str().as_bytes()` ?

then we dont even need the map

EDIT: nvm we need the vec

mattsse (approved):

crates/storage/provider/src/providers/database/provider.rs
hai-rise: Yeah, it's unfortunate to use heap types for table keys and their encoding, but the overhead also isn't big enough to migrate. We live with it for another day!

PR #18288, opened 2025-09-05
perf: specialize validate_transactions_with_origin for task validator

mattsse:
we should override this because the default always goes through single validation:

https://github.com/paradigmxyz/reth/blob/04103b0f21f52981c7a17b34df066ef7575f1ba3/crates/transaction-pool/src/validate/mod.rs#L214-L226


yongkangc (approved):
ah nice! basically having a single batch operation wihtout spawning n individual task and 1 vec collection

Issue #18032, opened 2025-08-25
Share a `HashedStorages` cursor among storage root calculations


hai-rise:
State root calculations are spending ~5.7% of time on creating new `tables::HashedStorages` database handles & cursors for each account's storage root calculation:
https://github.com/paradigmxyz/reth/blob/01f667c228ae752cc6a6e89e5f8c34c10daaf03b/crates/trie/trie/src/trie.rs#L620-L621

<img width="1107" height="792" alt="Image" src="https://github.com/user-attachments/assets/f69f3502-169f-4d00-ab3e-c6cd8dfcaf19" />

This is **very inefficient**. A single (sequential) state root run should share the same cursor for `tables::HashedStorages`, just like it already does for `tables::HashedAccounts`:
https://github.com/paradigmxyz/reth/blob/01f667c228ae752cc6a6e89e5f8c34c10daaf03b/crates/trie/trie/src/trie.rs#L165

PR #18292, opened 2025-09-05
perf(db): reuse MDBX DBIs for the same tx

hai-rise:
`MDBX_dbi` is an index into the MDBX environment to a table handle. We can safely share this index around unless we manually close and remove table handles from the MDBX environment (we don't).

This PR takes the first step of sharing the DBIs across queries of the same DB transaction, which already yields lots of hits; each hit makes the query 6%-82% faster. For example, every single `get` was trying to create a new table handle (`mdbx_dbi_open`)... 
https://github.com/paradigmxyz/reth/blob/0cdd54838b4311e00a71606ce618d6869a2a0704/crates/storage/db/src/implementation/mdbx/tx.rs#L293

A step further would be to pre-create and share the DBIs across all transactions.

No screenshot this one, as this path is literally everywhere :melting_face:.

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/tx.rs
+++ b/crates/storage/db/src/implementation/mdbx/tx.rs
@@ -5,6 +5,7 @@ use crate::{
     metrics::{DatabaseEnvMetrics, Operation, TransactionMode, TransactionOutcome},
     DatabaseError,
 };
+use dashmap::DashMap;
 use reth_db_api::{
     table::{Compress, DupSort, Encode, Table, TableImporter},
     transaction::{DbTx, DbTxMut},
@@ -31,6 +32,10 @@ pub struct Tx<K: TransactionKind> {
     /// Libmdbx-sys transaction.
     pub inner: Transaction<K>,
 
+    /// Cached MDBX DBIs for reuse.
+    /// TODO: Reuse DBIs even among transactions, ideally with no synchronization overhead.
+    dbis: DashMap<&'static str, MDBX_dbi>,
+
     /// Handler for metrics with its own [Drop] implementation for cases when the transaction isn't
     /// closed by [`Tx::commit`] or [`Tx::abort`], but we still need to report it in the metrics.
     ///
@@ -41,7 +46,7 @@ pub struct Tx<K: TransactionKind> {
 impl<K: TransactionKind> Tx<K> {
     /// Creates new `Tx` object with a `RO` or `RW` transaction.
     #[inline]
-    pub const fn new(inner: Transaction<K>) -> Self {
+    pub fn new(inner: Transaction<K>) -> Self {
         Self::new_inner(inner, None)
     }
 
@@ -64,8 +69,8 @@ impl<K: TransactionKind> Tx<K> {
     }
 
     #[inline]
-    const fn new_inner(inner: Transaction<K>, metrics_handler: Option<MetricsHandler<K>>) -> Self {
-        Self { inner, metrics_handler }
+    fn new_inner(inner: Transaction<K>, metrics_handler: Option<MetricsHandler<K>>) -> Self {
+        Self { inner, metrics_handler, dbis: DashMap::new() }
     }
 
     /// Gets this transaction ID.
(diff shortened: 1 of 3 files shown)


crates/storage/db/src/implementation/mdbx/tx.rs
```diff
@@ -75,10 +80,18 @@ impl<K: TransactionKind> Tx<K> {
 
     /// Gets a table database handle if it exists, otherwise creates it.
```
mattsse, line 81: unfamiliar with the internals here, but can we safely assume that this will always exist for the duration of the transaction?

maybe @shekhirin knows

hai-rise: `MDBX_dbi` is an index/pointer to the actual table handle stored in the DB environment, which will live until the end of the environment, basically the end of the program for Reth, or until it is explicitly closed via `mdbx_dbi_close`, which we don't and shouldn't do.

Since we don't dynamically create new tables at runtime (right?), the best practice is to pre-create all these table handles at startup time, then share their `MDBX_dbi`s across all transactions. 

Otherwise, (repeatedly) calling `mdbx_dbi_open` at runtime is very expensive. In the code path I showed earlier in #18284, 2.5 / 5.1 ~ 49% of the `get` (stage checkpoint) is spent on `open_db`, which likely returns the same `MDBX_dbi` anyway. Just much more expensive as it needs to execute a transaction, creates contention at the DB-level, etc.

image

gakonst:
What would be risks you foresee with this impl? Whether it is using data that's not up to date due to mvcc or any kind of deadlock? Very exciting work!

gakonst:
Would be awesome if you could reth-bench and see if there's e2e perf improvement given we heavily use opened transactions on a block by block basis -- does this carry over there too?

hai-rise:
> Would be awesome if you could reth-bench and see if there's e2e perf improvement given we heavily use opened transactions on a block by block basis -- does this carry over there too?

My observations yesterday:
- Heavy queries like flushing blocks with 10,000s of transactions "only" waste ~6% of their time on `mdbx_dbi_open`.
- Quick reads easily waste >50% of their time on `mdbx_dbi_open`.
- Each extra call doesn't only add latency to the query itself, but creates more contention at the DB-level, so it adds up and will scale badly when DB traffic spikes.
- Luckily, the heaviest reads already utilise cursors (albeit still not that well yet, e.g., #18032), so we only waste an `mdbx_dbi_open` on cursor creation, not per disk read there.
- I added a sad `dbg!` to the `Occupied` arm, and my `stderr` exploded with DBI hit & reuse during a quick historical sync run.

I'm also very curious about E2E benchmark results (especially for block building & RPC), but I will need my colleagues' help and the kids to return to school first!

erthink:
> A step further would be to pre-create and share the DBIs across all transactions.

If I understand correctly, in Reth, the set of key-value tables changes rarely and only from version to version, but it is constant when the application is running. If so, then you should (actually must to avoid dummy overhead) open the DBI handles once immediately after opening and/or creating the database, or after completing the migration/upgrade of the data schema.

This approach allows to use a global variables for DBI-handles for all tables, without the need for any synchronization of access/using ones and concern about the lifetime (it is not even necessary to close). So this could be applied safe (without a risk of regressions).

Ask in the [libmdbx' telegram group](https://t.me/libmdbx) if in doubt.

crates/storage/db/src/implementation/mdbx/tx.rs
shekhirin: That sounds correct, confirmed with MDBX documentation https://libmdbx.dqdkfa.ru/group__c__dbi.html#ga9bef4a9fdf27655e9343bbbf8b6fc5a1

shekhirin (approved):
LGTM, this is such an obvious change to do 🙈 Thanks for this! I'm going to merge this and we will run our internal nodes on main for a while before releasing it, so I feel pretty confident.

In the future we can even get rid of DashMap overhead by creating all DBI handles on startup.

PR #18293, opened 2025-09-05
chore(trie): add e2e for reorg

yongkangc:
### Summary
- add `trie_updates_test.rs`, an end-to-end regression test that reproduces the trie-update loss fixed in #18318
- ensure canonical blocks retain their trie updates after a short reorg by exercising the e2e TestBuilder flow
- guard against a recurrence of the state-root mismatch that stalled mainnet nodes in early September 2024

Scenario Under Test (`trie_updates_test.rs`)
1. Spin up a single-node mainnet chain spec with a pre-funded storage contract whose slots hash to shared prefixes, creating branch nodes identical to the production incident.
2. Produce `block_1`, which writes to those slots, capture the block, and assert its trie updates are persisted.
3. Produce `block_2`, which rewrites the same slots (mirroring the canonical block after the `23003311` fork), capture it, and assert trie updates are still present.
4. The assertions use `AssertMissingTrieUpdates` so that any regression that drops trie updates for canonical descendants stops the test.


crates/e2e-test-utils/src/testsuite/actions/transactions.rs
```diff
@@ -0,0 +1,268 @@
...
+/// Helper to create approve transaction for the default test token contract.
+/// The default contract address is 0x77d34361f991fa724ff1db9b1d760063a16770db
+/// which is used in the trie corruption tests.
+pub async fn create_test_approve_tx(
+    spender: Address,
+    amount: U256,
+    nonce: u64,
+    chain_id: u64,
+) -> Result<Bytes> {
+    let token_address = Address::from_str("0x77d34361f991fa724ff1db9b1d760063a16770db")?;
```
mattsse: this can be address!

mattsse (commented):
looking at the new test I get the idea what this does

unsure if we need all of this additional setup, looks like we do, wdyt @fgimenez

yongkangc:
> looking at the new test I get the idea what this does
> 
> unsure if we need all of this additional setup, looks like we do, wdyt @fgimenez

@mattsse realising that my approach might have to change to a simpler one due to issues mocking of genesis state and  recreate the other blocks, as it results in blockhash mismatch. 

instead of the current approach in the test, we would have to do something like 
     1. Genesis: Deploy storage contract with initial storage  
     2. Block 1 (canonical): Transaction that sets storage slot 0x0f to value A  
     3. Block 1' (fork): Transaction that sets storage slot 0x0f to value B (creates trie node)  
     4. Reorg: Make Block 1' canonical  
     5. Block 2: Transaction that clears storage slot 0x0f (should delete trie node)

crates/e2e-test-utils/src/testsuite/actions/transactions.rs
fgimenez: can we have a constant for this? with the additional documentation explaining how is it used

crates/e2e-test-utils/src/testsuite/actions/transactions.rs
```diff
@@ -0,0 +1,268 @@
...
+use reth_rpc_api::clients::EthApiClient;
+use std::str::FromStr;
+use tracing::debug;
+
+/// Action to produce blocks with real storage-modifying transactions.
+///
+/// This action sends transactions to the node's transaction pool and then
+/// triggers block production to include them.
+#[derive(Debug)]
+pub struct ProduceBlockWithTransactions {
```
fgimenez, line 42: :rocket:

crates/e2e-test-utils/src/testsuite/actions/transactions.rs
yongkangc: Will do! Currently overhauling this test

crates/chain-state/src/in_memory.rs
```diff
@@ -919,13 +919,25 @@ impl<N: NodePrimitives<SignedTx: SignedTransaction>> NewCanonicalChain<N> {
...
-                    chain.append_block(
-                        exec.recovered_block().clone(),
-                        exec.execution_outcome().clone(),
-                    );
-                    chain
-                }));
+                // For single-block chains, preserve trie updates for testing
+                // This maintains backwards compatibility while allowing tests to access trie
+                // updates
+                let new = if new.len() == 1 {
```
yongkangc, line 925: @shekhirin @mediocregopher @Rjected would appreciate a look over here as its what is required to replicate the test. 

However as with anything with the trie state would need a double look

yongkangc: my understanding of single block chains - the chain contains exactly one executed block

crates/e2e-test-utils/src/node.rs
```diff
@@ -305,4 +309,99 @@ where
...
+                                let _ = tx.send(event);
+
+                                tracing::debug!(
+                                    "Forwarded trie updates for committed block {} from node {}",
+                                    tip_hash,
+                                    node_idx
+                                );
+                            }
+                        } else {
+                            println!("Node {}: No trie updates in committed chain", node_idx);
```
yongkangc, line 362: to remove after debugging

crates/chain-state/src/in_memory.rs
mediocregopher: Afaict this is ok, the only behavior change is that CanonStateNotification::Commit contains trie updates for single block chains now, which the documentation states is the case anyway.

crates/e2e-test-utils/src/testsuite/actions/engine_api.rs
```diff
@@ -348,3 +301,534 @@ fn block_to_payload_v3(block: Block) -> ExecutionPayloadV3 {
...
+async fn load_payload_data(source: &PayloadSource) -> Result<String> {
+    match source {
+        PayloadSource::Json(json_data) => Ok(json_data.clone()),
+        PayloadSource::File(file_path) => {
+            // Handle path resolution for relative paths
+            let resolved_path = if file_path.starts_with("crates/") {
+                // Try multiple possible locations
+                let candidates = vec![
+                    std::path::Path::new(file_path).to_path_buf(),
+                    std::path::Path::new("../../..").join(file_path),
```
mediocregopher, line 343: What are these extra candidates trying to search? If `file_path` is the crates dir won't these be looking outside the reth project?

crates/e2e-test-utils/src/testsuite/actions/trie.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use crate::testsuite::{actions::Action, Environment};
+use eyre::Result;
+use reth_node_api::EngineTypes;
+use std::{future::Future, pin::Pin, time::Duration};
+use tokio::time::sleep;
+
+/// Action that asserts whether a block has missing trie updates
+#[derive(Debug)]
+pub struct AssertMissingTrieUpdates {
```
mediocregopher, line 12: nit: really we're asserting that the updates are _present_, so AssertTrieUpdatesPresent would be a better name

crates/e2e-test-utils/src/testsuite/actions/trie.rs
```diff
@@ -0,0 +1,103 @@
...
+            sleep(Duration::from_millis(150)).await;
+            env.drain_trie_updates();
+
+            // Get block info
+            let (block_info, node_idx) = *env
+                .block_registry
+                .get(&self.block_tag)
+                .ok_or_else(|| eyre::eyre!("Unknown block tag {}", self.block_tag))?;
+
+            eprintln!("\n=== CHECKING TRIE UPDATES STATUS FOR BLOCK {} ===", self.block_tag);
```
mediocregopher, line 49: Are these left in intentionally? They seem like they'd be noisy

crates/e2e-test-utils/src/testsuite/actions/trie.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.expect_missing = missing;
+        self
+    }
+}
+
+impl<Engine> Action<Engine> for AssertMissingTrieUpdates
+where
+    Engine: EngineTypes,
+{
+    fn execute<'a>(
```
mediocregopher, line 34: I think this method will get simplified quite a bit once #18460 is done, but overall definitely still valuable to be able to assert that any trie updates were generated 👍

crates/e2e-test-utils/src/testsuite/setup.rs
```diff
@@ -178,6 +187,11 @@ where
             node_clients.push(crate::testsuite::NodeClient::new(rpc, auth, url));
         }
 
+        // We'll handle node context setting in a different way since nodes don't implement Clone
```
mediocregopher, line 191: Is this comment still relevant?

crates/e2e-test-utils/src/node.rs
```diff
@@ -305,4 +309,99 @@ where
...
+    {
+        // Take the canonical stream (we can't clone it, so we move it into the task)
+        // This means each node can only have one forwarder, which is fine for our use case
+        let canonical_stream = std::mem::replace(
+            &mut self.canonical_stream,
+            self.inner.provider.canonical_state_stream(),
+        );
+
+        tokio::spawn(async move {
+            println!("Starting trie update forwarder for node {}", node_idx);
```
mediocregopher, line 328: leftover println

crates/e2e-test-utils/tests/e2e_testsuite/main.rs
```diff
@@ -37,7 +39,7 @@ async fn test_apply_with_import() -> Result<()> {
                 .unwrap(),
             )
             .london_activated()
-            .shanghai_activated()
+            .cancun_activated()
```
mediocregopher, line 42: was this deliberate? cancun is already activated here

crates/e2e-test-utils/tests/e2e_testsuite/trie_updates_test.rs
```diff
@@ -0,0 +1,303 @@
...
+            .build(),
+    );
+
+    Setup::<EthEngineTypes>::default()
+        .with_chain_spec(chain_spec)
+        .with_network(NetworkSetup::single_node())
+        .with_tree_config(
+            TreeConfig::default()
+                .with_persistence_threshold(0) // Persist blocks immediately so trie updates are kept
+                .with_legacy_state_root(true), /* Use legacy state root to bypass trie update
```
mediocregopher: I'm not sure this with_legacy_state_root should be necessary, would be better to leave it out if possible

crates/node/core/src/version.rs
```diff
@@ -113,7 +113,7 @@ pub fn default_reth_version_metadata() -> RethCliVersionConsts {
         vergen_git_sha: Cow::Owned(env!("VERGEN_GIT_SHA_SHORT").to_string()),
         vergen_build_timestamp: Cow::Owned(env!("VERGEN_BUILD_TIMESTAMP").to_string()),
         vergen_cargo_target_triple: Cow::Owned(env!("VERGEN_CARGO_TARGET_TRIPLE").to_string()),
-        vergen_cargo_features: Cow::Owned(env!("VERGEN_CARGO_FEATURES").to_string()),
+        vergen_cargo_features: Cow::Owned(String::new()),
```
mediocregopher: probably not deliberate?

mediocregopher (changes requested):
Left a bunch of comments, mostly nitpicks.

One thing I'd note is that another change that should be possible as part of #18460 is that ExExNotifications will have trie updates on them as well. I'm not sure if that would be easier to work into the tests than the canon notifications, maybe it's the same, but just something that came to mind reading this.

PR #18297, opened 2025-09-05
feat(gpo): add default fee price argument

Zygimantass:


crates/node/core/src/args/gas_price_oracle.rs
```diff
@@ -25,17 +25,22 @@ pub struct GasPriceOracleArgs {
     /// The percentile of gas prices to use for the estimate
     #[arg(long = "gpo.percentile", default_value_t = DEFAULT_GAS_PRICE_PERCENTILE)]
     pub percentile: u32,
+
+    /// The default gas price to use if there are no blocks to use
+    #[arg(long = "gpo.default")]
+    pub default: Option<U256>,
```
mattsse: can we name this differently to default_suggest_fee

mattsse (changes requested):
smol naming nit

crates/node/core/src/args/gas_price_oracle.rs
Zygimantass: can do

Zygimantass: fixed

mattsse (approved):
applied some pedantic touchups, the unwrap_or(default) would have used U256::ZERO

PR #18299, opened 2025-09-05
feat: derive dev accounts from mnemonic in dev mode

### LINKED ISSUE #18265: Install correct dev accounts for devmode
### Describe the feature

currently we install random accounts

https://github.com/paradigmxyz/reth/blob/36e39ebe3dd5199c8f11a705ae8c0a2bd534ede2/crates/rpc/rpc/src/eth/helpers/signer.rs#L28-L28

https://github.com/paradigmxyz/reth/blob/36e39ebe3dd5199c8f11a705ae8c0a2bd534ede2/crates/node/builder/src/rpc.rs#L992-L995

which isnt actually super useful because in devmode it should pick the accounts from the dev chainspec:

Mnemonic:          test test test test test test test test test test test junk

ideally this is also configurable and should only be done if the chainspec only be done automatically if the chainspec is dev chain

## TODO
* add `Option<Mnemonic> arg to devargs
* use default Mnemonic to derive accounts, for this we can add a helper to:

https://github.com/paradigmxyz/reth/blob/36e39ebe3dd5199c8f11a705ae8c0a2bd534ede2/crates/rpc/rpc-eth-api/src/helpers/signer.rs#L37-L42

and https://github.com/paradigmxyz/reth/blob/36e39ebe3dd5199c8f11a705ae8c0a2bd534ede2/crates/rpc/rpc/src/eth/helpers/signer.rs#L34-L34

see also alloy's Mnemonic type

### Additional context

_No response_

dharmvr1:
closes #18265


Cargo.toml
```diff
@@ -507,7 +507,7 @@ alloy-rpc-types-trace = { version = "1.0.27", default-features = false }
 alloy-rpc-types-txpool = { version = "1.0.27", default-features = false }
 alloy-serde = { version = "1.0.27", default-features = false }
 alloy-signer = { version = "1.0.27", default-features = false }
-alloy-signer-local = { version = "1.0.27", default-features = false }
+alloy-signer-local = { version = "1.0.27", default-features = false,features = ["mnemonic"]}
```
mattsse: let's only enable this in the rpc crate and not for the entire workspace

crates/node/core/src/args/dev.rs
```diff
@@ -39,6 +39,16 @@ pub struct DevArgs {
         verbatim_doc_comment
     )]
     pub block_time: Option<Duration>,
+
+    /// Derive dev accounts from a fixed mnemonic instead of random ones.
+    /// Defaults to "test test ... junk".
+    #[arg(
+        long = "dev.mnemonic",
```
mattsse, line 47: we should require dev here

crates/node/core/src/args/dev.rs
```diff
@@ -39,6 +39,16 @@ pub struct DevArgs {
         verbatim_doc_comment
     )]
     pub block_time: Option<Duration>,
+
+    /// Derive dev accounts from a fixed mnemonic instead of random ones.
+    /// Defaults to "test test ... junk".
```
mattsse: we can include the entire mnemonic here

mattsse (changes requested):
smol suggestions

dharmvr1:
@mattsse I applied the changes, but still getting a Cargo.toml error. Could you check and let me know if anything else is needed?

crates/node/core/src/args/dev.rs
```diff
@@ -39,6 +39,17 @@ pub struct DevArgs {
...
+    /// Derive dev accounts from a fixed mnemonic instead of random ones.
+    /// "test test test test test test test test test test test junk".
+    #[arg(
+        long = "dev.mnemonic",
+        help_heading = "Dev testnet",
+        value_name = "MNEMONIC",
+        requires = "dev",
+        verbatim_doc_comment
+    )]
+    pub dev_mnemonic: Option<String>,
```
mattsse: actually, can we make this an `Option<Option<String>>`

then --dev.mnemonic will be default mnemoic and with a value we can override it

mattsse (changes requested):
sorry about the delay, 

left one tiny suggestion

dharmvr1:
hi @mattsse  
i made some new changes, please take a look and let me know if anything else is needed.  
thanks!

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -299,8 +299,17 @@ where
         Network: RpcTypes<TransactionRequest: SignableTxRequest<ProviderTx<N::Provider>>>,
     >,
 {
-    fn with_dev_accounts(&self) {
-        *self.inner.eth_api.signers().write() = DevSigner::random_signers(20)
```
klkvr, line 346: given that we're breaking this trait let's just remove it completely. we already have access to `eth_api.signers()` here directly https://github.com/paradigmxyz/reth/blob/8b7b0748620e98e11fd8cf15361517cd823e9861/crates/node/builder/src/rpc.rs#L992-L995

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -299,8 +299,17 @@ where
...
+    fn with_dev_accounts(&self, dev_mnemonic: Option<Option<String>>) {
+        const DEFAULT_MNEMONIC: &str =
+            "test test test test test test test test test test test junk";
+        let phrase = match dev_mnemonic {
+            None => return,                  // no dev accounts
+            Some(None) => DEFAULT_MNEMONIC,  // flag given without value
+            Some(Some(ref m)) => m.as_str(), // custom mnemonic
```
klkvr: my preference would be for this to be a helper on `DevArgs`

mattsse (approved):

PR #18301, opened 2025-09-05
feat(trie): Add helper sub-command

mediocregopher:


crates/cli/commands/src/db/mod.rs
```diff
@@ -135,6 +138,10 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
                 let Environment { provider_factory, .. } = self.env.init::<N>(AccessRights::RW)?;
                 command.execute(provider_factory)?;
             }
+            Subcommands::RepairTrie(command) => {
+                let Environment { provider_factory, .. } = self.env.init::<N>(AccessRights::RW)?;
+                command.execute(provider_factory)?;
```
mattsse, line 143: we could add a --dry-run arg and init this with readonly, but could do this as a followup

crates/cli/commands/src/db/repair_trie.rs
```diff
@@ -0,0 +1,163 @@
...
+            if let Output::Progress(path) = output {
+                // Output progress every 5 seconds
+                if last_progress_time.elapsed() > Duration::from_secs(5) {
+                    output_progress(path, start_time, inconsistent_nodes);
+                    last_progress_time = Instant::now();
+                }
+                continue
+            };
+
+            warn!("Inconsistency found: {output:?}");
```
mattsse: should we add some context here that we're repairing this, like `, collecting update` or something to make it more obvious that this will be repaired

mattsse (commented):
very nice

pending @shekhirin 

we could potentially make the readonly impl work if we instead make the entrypoint either tx or tx_mut

crates/cli/commands/src/db/repair_trie.rs
mediocregopher: Reworded to be `Inconsistency found, will repair: {output:?}`

shekhirin: that now says "will repair" even with dry run, a bit confusing

shekhirin (approved):
Very nice, let's do read-only tx on `--dry-run` and parallelism as follow-ups.

PR #18307, opened 2025-09-05
perf: specialize len 1

mattsse:
if we only have one item we can use the more optimal variant for this and avoid some helper collect


yongkangc (approved):
Nice find!

PR #18308, opened 2025-09-06
feat(observability): add phase-level observablity to newPayload processing 

### LINKED ISSUE #18187: feat(observability): add phase-level observability to newPayload processing
## Problem

Currently lacks granular timing metrics for newPayload processing phases, making it difficult to identify performance bottlenecks.

Cannot measure:
- Time until execution starts
- Prewarming effectiveness  
- Individual phase durations (validation, execution, state root)
- State root algorithm performance

Current implementation:
https://github.com/paradigmxyz/reth/blob/main/crates/engine/tree/src/tree/mod.rs#L504-L600

## Solution

do we have enough observability for the entire newPayload process?

what would help us is like a time diagram to see

- **Phase A**: Pre-execution (entry → execution start)
- **Phase B**: Prewarming initialization
- **Phase C**: Block execution
- **Phase D**: Post-execution validation
- **Phase E**: State root computation
- **Phase F**: Post-processing

### Implementation locations
- `crates/engine/tree/src/tree/metrics.rs` - Add `NewPayloadPhaseMetrics` struct
- `crates/engine/tree/src/tree/mod.rs#L504` - Instrument `on_new_payload`
- `crates/engine/tree/src/tree/payload_validator.rs#L288` - Instrument validation phases

## Expected outcome
- Enable time diagram generation for analysis
- Metrics

## Additional context

Timosdev99:
closes: #18187 

this feature ensures proper granular timing metrics for newPayload processing phases 

- Added **newPayloadPhaseMetrics** struct in **metrics.rs**, defining metric of each phase of new payload processing 
- instrumented the **on_new_payload** function in **mod.rs** for record duration of **overall process**, **pre-warming execution** and **post processing phase** 
- instrumented the **validate_block_with_state** function in **payload_validator.rs** to hold duration for **pre-warming initialization** , **block  execution** , **post-execution validation**, **sparse tree**, and **state root algorithm**

### DIFF
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -167,6 +167,10 @@ pub(crate) struct BlockValidationMetrics {
     pub(crate) state_root_duration: Gauge,
     /// Trie input computation duration
     pub(crate) trie_input_duration: Histogram,
+    /// Payload conversion and validation latency
+    pub(crate) payload_validation_duration: Gauge,
+    /// Histogram of payload validation latency
+    pub(crate) payload_validation_histogram: Histogram,
     /// Payload processor spawning duration
     pub(crate) spawn_payload_processor: Histogram,
     /// Post-execution validation duration
@@ -181,6 +185,13 @@ impl BlockValidationMetrics {
         self.state_root_storage_tries_updated_total
             .increment(trie_output.storage_tries_ref().len() as u64);
         self.state_root_duration.set(elapsed_as_secs);
+    }
+
+    /// Records a new payload validation time, updating both the histogram and the payload
+    /// validation gauge
+    pub(crate) fn record_payload_validation(&self, elapsed_as_secs: f64) {
+        self.payload_validation_duration.set(elapsed_as_secs);
+        self.payload_validation_histogram.record(elapsed_as_secs);
     }
 }
 
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -200,22 +200,14 @@ impl BlockValidationMetrics {
 pub(crate) struct NewPayloadPhaseMetrics {
     /// Pre-execution phase duration (entry → execution start)
     pub(crate) pre_execution_duration: Histogram,
-    /// Prewarming initialization duration
-    pub(crate) prewarming_init_duration: Histogram,
-    /// Block execution duration
-    pub(crate) execution_duration: Histogram,
+    /// Payload processor spawning duration
+    pub(crate) spawn_payload_processor: Histogram,
     /// Post-execution validation duration
     pub(crate) post_execution_validation_duration: Histogram,
     /// State root computation duration
     pub(crate) state_root_duration: Histogram,
-    /// Post-processing duration
-    pub(crate) post_processing_duration: Histogram,
-    /// Parallel state root algorithm duration
-    pub(crate) parallel_state_root_duration: Histogram,
-    /// Sparse trie state root algorithm duration
-    pub(crate) sparse_trie_state_root_duration: Histogram,
-    /// Regular state root algorithm duration
-    pub(crate) regular_state_root_duration: Histogram,
+    /// Total duration of the new payload call
+    pub(crate) total_duration: Histogram,
 }
 
 /// Metrics for the blockchain tree block buffer
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -26,6 +26,8 @@ pub(crate) struct EngineApiMetrics {
     pub(crate) executor: ExecutorMetrics,
     /// Metrics for block validation
     pub(crate) block_validation: BlockValidationMetrics,
+    /// Metrics for newPayload processing phases
+    pub(crate) new_payload_phases: NewPayloadPhaseMetrics,
     /// A copy of legacy blockchain tree metrics, to be replaced when we replace the old tree
     pub tree: TreeMetrics,
 }
@@ -190,6 +192,30 @@ impl BlockValidationMetrics {
         self.payload_validation_duration.set(elapsed_as_secs);
         self.payload_validation_histogram.record(elapsed_as_secs);
     }
+}
+
+/// Metrics for newPayload processing phases
+#[derive(Metrics)]
+#[metrics(scope = "engine.new_payload")]
+pub(crate) struct NewPayloadPhaseMetrics {
+    /// Pre-execution phase duration (entry → execution start)
+    pub(crate) pre_execution_duration: Histogram,
+    /// Prewarming initialization duration
+    pub(crate) prewarming_init_duration: Histogram,
+    /// Block execution duration
+    pub(crate) execution_duration: Histogram,
+    /// Post-execution validation duration
+    pub(crate) post_execution_validation_duration: Histogram,
+    /// State root computation duration
+    pub(crate) state_root_duration: Histogram,
+    /// Post-processing duration
+    pub(crate) post_processing_duration: Histogram,
+    /// Parallel state root algorithm duration
+    pub(crate) parallel_state_root_duration: Histogram,
+    /// Sparse trie state root algorithm duration
+    pub(crate) sparse_trie_state_root_duration: Histogram,
+    /// Regular state root algorithm duration
+    pub(crate) regular_state_root_duration: Histogram,
 }
 
 /// Metrics for the blockchain tree block buffer
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -165,6 +165,8 @@ pub(crate) struct BlockValidationMetrics {
     pub(crate) state_root_parallel_fallback_total: Counter,
     /// Latest state root duration, ie the time spent blocked waiting for the state root.
     pub(crate) state_root_duration: Gauge,
+    /// Histogram for state root duration ie the time spent blocked waiting for the state root
+    pub(crate) state_root_histogram: Histogram,
     /// Trie input computation duration
     pub(crate) trie_input_duration: Histogram,
     /// Payload conversion and validation latency
@@ -185,6 +187,7 @@ impl BlockValidationMetrics {
         self.state_root_storage_tries_updated_total
             .increment(trie_output.storage_tries_ref().len() as u64);
         self.state_root_duration.set(elapsed_as_secs);
+        self.state_root_histogram.record(elapsed_as_secs);
     }
 
     /// Records a new payload validation time, updating both the histogram and the payload
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -26,8 +26,6 @@ pub(crate) struct EngineApiMetrics {
     pub(crate) executor: ExecutorMetrics,
     /// Metrics for block validation
     pub(crate) block_validation: BlockValidationMetrics,
-    /// Metrics for newPayload processing phases
-    pub(crate) new_payload_phases: NewPayloadPhaseMetrics,
     /// A copy of legacy blockchain tree metrics, to be replaced when we replace the old tree
     pub tree: TreeMetrics,
 }
@@ -175,6 +173,14 @@ pub(crate) struct BlockValidationMetrics {
     pub(crate) payload_validation_duration: Gauge,
     /// Histogram of payload validation latency
     pub(crate) payload_validation_histogram: Histogram,
+    /// Pre-execution phase duration (entry → execution start)
+    pub(crate) pre_execution_duration: Histogram,
+    /// Payload processor spawning duration
+    pub(crate) spawn_payload_processor: Histogram,
+    /// Post-execution validation duration
+    pub(crate) post_execution_validation_duration: Histogram,
+    /// Total duration of the new payload call
+    pub(crate) total_duration: Histogram,
 }
 
 impl BlockValidationMetrics {
@@ -192,22 +198,6 @@ impl BlockValidationMetrics {
         self.payload_validation_duration.set(elapsed_as_secs);
         self.payload_validation_histogram.record(elapsed_as_secs);
     }
-}
-
-/// Metrics for newPayload processing phases
-#[derive(Metrics)]
-#[metrics(scope = "engine.new_payload")]
-pub(crate) struct NewPayloadPhaseMetrics {
-    /// Pre-execution phase duration (entry → execution start)
-    pub(crate) pre_execution_duration: Histogram,
-    /// Payload processor spawning duration
-    pub(crate) spawn_payload_processor: Histogram,
-    /// Post-execution validation duration
-    pub(crate) post_execution_validation_duration: Histogram,
-    /// State root computation duration
-    pub(crate) state_root_duration: Histogram,
-    /// Total duration of the new payload call
-    pub(crate) total_duration: Histogram,
 }
 
 /// Metrics for the blockchain tree block buffer
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -163,18 +163,10 @@ pub(crate) struct BlockValidationMetrics {
     pub(crate) state_root_storage_tries_updated_total: Counter,
     /// Total number of times the parallel state root computation fell back to regular.
     pub(crate) state_root_parallel_fallback_total: Counter,
-    /// Histogram of state root duration, ie the time spent blocked waiting for the state root.
-    pub(crate) state_root_histogram: Histogram,
     /// Latest state root duration, ie the time spent blocked waiting for the state root.
     pub(crate) state_root_duration: Gauge,
     /// Trie input computation duration
     pub(crate) trie_input_duration: Histogram,
-    /// Payload conversion and validation latency
-    pub(crate) payload_validation_duration: Gauge,
-    /// Histogram of payload validation latency
-    pub(crate) payload_validation_histogram: Histogram,
-    /// Pre-execution phase duration (entry → execution start)
-    pub(crate) pre_execution_duration: Histogram,
     /// Payload processor spawning duration
     pub(crate) spawn_payload_processor: Histogram,
     /// Post-execution validation duration
@@ -189,14 +181,6 @@ impl BlockValidationMetrics {
         self.state_root_storage_tries_updated_total
             .increment(trie_output.storage_tries_ref().len() as u64);
         self.state_root_duration.set(elapsed_as_secs);
-        self.state_root_histogram.record(elapsed_as_secs);
-    }
-
-    /// Records a new payload validation time, updating both the histogram and the payload
-    /// validation gauge
-    pub(crate) fn record_payload_validation(&self, elapsed_as_secs: f64) {
-        self.payload_validation_duration.set(elapsed_as_secs);
-        self.payload_validation_histogram.record(elapsed_as_secs);
     }
 }
 
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1752,7 +1752,7 @@ where
     ) -> ProviderResult<Option<B256>> {
         // Check if parent exists in side chain or in canonical chain.
         if self.sealed_header_by_hash(parent_hash)?.is_some() {
-            return Ok(Some(parent_hash));
+            return Ok(Some(parent_hash))
         }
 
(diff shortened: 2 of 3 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -116,19 +116,19 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
...
-            return PersistingKind::PersistingNotDescendant
+            return PersistingKind::PersistingNotDescendant;
         };
 
         // The block being validated can only be a descendant if its number is higher than
         // the highest block persisting. Otherwise, it's likely a fork of a lower block.
         if block.block.number > highest.number &&
             self.state().tree_state.is_descendant(*highest, block)
         {
-            return PersistingKind::PersistingDescendant
```
Rjected: Let's bring back no semicolons

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -192,6 +194,30 @@ impl BlockValidationMetrics {
     }
 }
 
+/// Metrics for newPayload processing phases
+#[derive(Metrics)]
+#[metrics(scope = "engine.new_payload")]
+pub(crate) struct NewPayloadPhaseMetrics {
```
Rjected: These should be all be moved into `BlockValidationMetrics` IMO

crates/engine/tree/src/tree/mod.rs
```diff
@@ -625,6 +637,18 @@ where
...
+        // record post-processing duration
+        self.metrics
+            .new_payload_phases
+            .post_processing_duration
+            .record(post_processing_start.elapsed().as_secs_f64());
+
+        // record total newPayload duration
+        self.metrics
+            .new_payload_phases
+            .pre_execution_duration // will be using this as a proxy for total duration
```
Rjected: why are we using `pre_execution_duration` here?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -508,7 +508,11 @@ where
         trace!(target: "engine::tree", "invoked new payload");
         self.metrics.engine.new_payload_messages.increment(1);
 
-        let validation_start = Instant::now();
+        // start timing for the entire new payload process
+        let new_payload_start = Instant::now();
+
+        // Phase A: Pre-execution (entry → execution start)
+        let pre_execution_start = Instant::now();
```
Rjected: we should just be using `validation_start`, or `new_payload_start`, instead of creating two `Instant`s

crates/engine/tree/src/tree/mod.rs
```diff
@@ -537,9 +541,15 @@ where
         // This validation **MUST** be instantly run in all cases even during active sync process.
         let parent_hash = payload.parent_hash();
 
+        // record pre-execution phase duration
+        self.metrics
+            .new_payload_phases
+            .pre_execution_duration
+            .record(pre_execution_start.elapsed().as_secs_f64());
```
Rjected: I don't think this metric is useful because we are just recording getting the parent hash here, let's remove this

crates/engine/tree/src/tree/mod.rs
```diff
@@ -537,9 +541,15 @@ where
...
+
         self.metrics
             .block_validation
-            .record_payload_validation(validation_start.elapsed().as_secs_f64());
+            .record_payload_validation(pre_execution_start.elapsed().as_secs_f64());
```
Rjected: for this metric, let's either keep as-is or fix this issue instead:
https://github.com/paradigmxyz/reth/issues/18014

crates/engine/tree/src/tree/mod.rs
```diff
@@ -625,6 +637,18 @@ where
             }
         }
 
+        // record post-processing duration
+        self.metrics
+            .new_payload_phases
+            .post_processing_duration
+            .record(post_processing_start.elapsed().as_secs_f64());
```
Rjected: tbh I am not sure how useful this metric is either, given how cheap the outcome checks are

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -192,6 +194,30 @@ impl BlockValidationMetrics {
...
+    /// Pre-execution phase duration (entry → execution start)
+    pub(crate) pre_execution_duration: Histogram,
+    /// Prewarming initialization duration
+    pub(crate) prewarming_init_duration: Histogram,
```
Rjected: this should be renamed to something like `spawn_payload_processor` since that's what it measures

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -192,6 +194,30 @@ impl BlockValidationMetrics {
...
+    /// Block execution duration
+    pub(crate) execution_duration: Histogram,
```
Rjected: we already have metrics for execution, I think we can remove it

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -192,6 +194,30 @@ impl BlockValidationMetrics {
...
+    /// State root computation duration
+    pub(crate) state_root_duration: Histogram,
+    /// Post-processing duration
+    pub(crate) post_processing_duration: Histogram,
+    /// Parallel state root algorithm duration
+    pub(crate) parallel_state_root_duration: Histogram,
+    /// Sparse trie state root algorithm duration
+    pub(crate) sparse_trie_state_root_duration: Histogram,
+    /// Regular state root algorithm duration
+    pub(crate) regular_state_root_duration: Histogram,
```
Rjected: I don't think we need all three of these, I think adding these should be a separate issue or PR, if we wanted to distinguish the latency for each algorithm

Rjected (changes requested):
I think we should remove all "phase" naming and just name things after what is being measured. there's also some redundancy here, left some suggestions for what we should do instead

crates/engine/tree/src/tree/metrics.rs
Timosdev99: > These should be all be moved into `BlockValidationMetrics` IMO

please can i know why 
i think this way is okay since they both measure different things and tbh i think the current structure is easy to maintain

crates/engine/tree/src/tree/mod.rs
Timosdev99: > we should just be using `validation_start`, or `new_payload_start`, instead of creating two `Instant`s

the **pre_execution_start** instant  is used for measuring only the  pre-execution phase 
we cant use **new_payload_start** instant to measure both because we need to stop for the pre execution phase long before the entire function finishes

crates/engine/tree/src/tree/mod.rs
Timosdev99: > I don't think this metric is useful because we are just recording getting the parent hash here, let's remove this

instead of removing it, I can just use it to measure the setup cost separately
IMO, it will be useful to identify bottlenecks

crates/engine/tree/src/tree/mod.rs
Timosdev99: > why are we using `pre_execution_duration` here?

currently using it to record the total duration too, but i will make changes to that  by adding a new field for total duration

Timosdev99:
> I think we should remove all "phase" naming and just name things after what is being measured. there's also some redundancy here, left some suggestions for what we should do instead

hey @Rjected i made some changes already ,can you please check it out

Timosdev99:
Hey @yongkangc please can I get a review on the changes I made recently on this and any suggestion

crates/engine/tree/src/tree/mod.rs
Rjected: we can use the same `Instant` here though to record both, since we're just creating one after the other

for example doing this:
```
let start = Instant::now();

// do thing...

let first_elapsed = start.elapsed();

// do another thing...

let second_elapsed = start.elapsed();
```

will have the same result as doing

```
let start = Instant::now();
let second_start = Instant::now();

// do thing...

let first_elapsed = start.elapsed();

// do another thing...

let second_elapsed = second_start.elapsed();
```

with the exception being the time it takes for the call to
```
let second_start = Instant::now();
```

crates/engine/tree/src/tree/metrics.rs
Rjected: > please can i know why

because `newPayload` basically just does block validation, and some of the things in this struct, will need to be measured in the same function where the block validation metrics are measured

Rjected (changes requested):
nice, just a few more comments

crates/engine/tree/src/tree/metrics.rs
yongkangc: we could have them as a field in BlockValidation metric?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -562,14 +562,19 @@ where
                 Ok(block) => block,
                 Err(error) => {
                     let status = self.on_new_payload_error(error, parent_hash)?;
-                    return Ok(TreeOutcome::new(status))
+                    return Ok(TreeOutcome::new(status));
```
yongkangc: semicolon cant be removed :)

crates/engine/tree/src/tree/mod.rs
```diff
@@ -562,14 +562,19 @@ where
...
                 }
             };
 
             let status = self.on_invalid_new_payload(block.into_sealed_block(), invalid)?;
-            return Ok(TreeOutcome::new(status))
+            return Ok(TreeOutcome::new(status));
```
yongkangc: semicolon cant be removed :)

yongkangc (changes requested):
nice! @Rjected covered most of the pointers

no semicolons colons would be good :)

crates/engine/tree/src/tree/mod.rs
```diff
@@ -562,14 +561,19 @@ where
...
         }
 
+        // record pre-execution phase duration
+        let pre_execution_elapsed = start.elapsed().as_secs_f64();
+        self.metrics.block_validation.pre_execution_duration.record(pre_execution_elapsed);
```
Rjected: let's remove this metric now, since `record_payload_validation` should take care of this and records the same thing

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -403,7 +405,8 @@ where
             // Use state root task only if prefix sets are empty, otherwise proof generation is too
             // expensive because it requires walking over the paths in the prefix set in every
             // proof.
-            if trie_input.prefix_sets.is_empty() {
+            let prewarming_start = Instant::now();
```
Rjected: let's rename this to `spawn_payload_processor_start`

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -453,6 +472,8 @@ where
             };
         }
 
+        // Phase D: Post-execution validation
```
Rjected:
```suggestion
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -527,7 +554,8 @@ where
                             regular_state_root = ?result.0,
                             "Regular root task finished"
                         );
-                        maybe_state_root = Some((result.0, result.1, root_time.elapsed()));
+                        let elapsed = root_time.elapsed();
+                        maybe_state_root = Some((result.0, result.1, elapsed));
```
Rjected:
```suggestion
                        maybe_state_root = Some((result.0, result.1, root_time.elapsed()));
```

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -543,10 +571,10 @@ where
             }
         }
 
-        let (state_root, trie_output, root_elapsed) = if let Some(maybe_state_root) =
+        let (state_root, trie_output, root_elapsed) = if let Some(ref maybe_state_root) =
             maybe_state_root
         {
-            maybe_state_root
+            maybe_state_root.clone()
```
Rjected: why do we need to clone this? would prefer keeping the code as it was here

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -561,6 +589,9 @@ where
             (root, updates, root_time.elapsed())
         };
 
+        // record state root computation duration (Phase E)
+        self.metrics.block_validation.state_root_histogram.record(root_elapsed.as_secs_f64());
```
Rjected: we might not need this now since we do 
```
self.metrics.block_validation.record_state_root(&trie_output, root_elapsed.as_secs_f64());
```
on the next line

crates/engine/tree/src/tree/mod.rs
```diff
@@ -508,7 +508,10 @@ where
         trace!(target: "engine::tree", "invoked new payload");
         self.metrics.engine.new_payload_messages.increment(1);
 
-        let validation_start = Instant::now();
+        // start timing for the new payload process
+        let start = Instant::now();
+
+        // Phase A: Pre-execution (entry → execution start)
```
Rjected:
```suggestion
```

Rjected (changes requested):
mostly nits, I think this is almost there

crates/engine/tree/src/tree/mod.rs
```diff
@@ -432,21 +432,21 @@ where
                     debug!(target: "engine::tree", %msg, "received new engine message");
                     if let Err(fatal) = self.on_engine_message(msg) {
                         error!(target: "engine::tree", %fatal, "insert block fatal error");
-                        return
```
yongkangc: can we remove semis here?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -432,21 +432,21 @@ where
...
+                        return;
                     }
                 }
                 Ok(None) => {
                     debug!(target: "engine::tree", "received no engine message for some time, while waiting for persistence task to complete");
                 }
                 Err(_err) => {
                     error!(target: "engine::tree", "Engine channel disconnected");
-                    return
```
yongkangc: here too

crates/engine/tree/src/tree/mod.rs
```diff
@@ -473,7 +473,7 @@ where
                 self.on_tree_event(event)?;
                 if needs_backfill {
                     // can exit early if backfill is needed
-                    return Ok(None)
```
yongkangc: here too

crates/engine/tree/src/tree/mod.rs
```diff
@@ -462,7 +462,7 @@ where
     ) -> Result<Option<TreeEvent>, InsertBlockFatalError> {
         if blocks.is_empty() {
             // nothing to execute
-            return Ok(None)
```
yongkangc: here too

yongkangc (changes requested):
think we should bring back the no semis

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -416,9 +419,25 @@ where
...
+
+            // record prewarming initialization duration
+            self.metrics
+                .block_validation
+                .spawn_payload_processor
+                .record(prewarming_start.elapsed().as_secs_f64());
+            handle
         } else {
-            self.payload_processor.spawn_cache_exclusive(env.clone(), txs, provider_builder)
+            let prewarming_start = Instant::now();
```
yongkangc, line 431: is this a duplicate as above with the `            let prewarming_start = Instant::now();`

https://github.com/paradigmxyz/reth/pull/18308/files#diff-6b5414154ff2b0a9b6cb854533d6b0a6024e0c1877742b0fa6d8b4d2e790cf82R408

yongkangc (changes requested):

crates/engine/tree/src/tree/metrics.rs
```diff
@@ -163,8 +163,6 @@ pub(crate) struct BlockValidationMetrics {
     pub(crate) state_root_storage_tries_updated_total: Counter,
     /// Total number of times the parallel state root computation fell back to regular.
     pub(crate) state_root_parallel_fallback_total: Counter,
-    /// Histogram of state root duration, ie the time spent blocked waiting for the state root.
```
yongkangc, line 166: might have missed the comment but why did we remove this? 

@Rjected is this impt?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -432,21 +432,21 @@ where
                     debug!(target: "engine::tree", %msg, "received new engine message");
                     if let Err(fatal) = self.on_engine_message(msg) {
                         error!(target: "engine::tree", %fatal, "insert block fatal error");
-                        return
+                        return;
```
yongkangc: can we remove the semis here please?

yongkangc (changes requested):
almost there!  nits - can we please revert the changes to the semi colons please?

Added a bunch of comments across the areas - but there are more.

crates/engine/tree/src/tree/metrics.rs
yongkangc: @Timosdev99 can I check why did we remove this?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1047,7 +1049,7 @@ where
                         ProviderError::HeaderNotFound(state.head_block_hash.into())
                     })?;
                 let updated = self.process_payload_attributes(attr, &tip, state, version);
-                return Ok(TreeOutcome::new(updated))
+                return Ok(TreeOutcome::new(updated));
```
yongkangc:
```suggestion
                return Ok(TreeOutcome::new(updated))
```

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1067,7 +1069,7 @@ where
                     debug!(target: "engine::tree", head = canonical_header.number(), "handling payload attributes for canonical head");
                     let updated =
                         self.process_payload_attributes(attr, &canonical_header, state, version);
-                    return Ok(TreeOutcome::new(updated))
+                    return Ok(TreeOutcome::new(updated));
```
yongkangc:
```suggestion
                    return Ok(TreeOutcome::new(updated))
```

yongkangc (changes requested):
some extra nits, but would like @mattsse to take a look as well :) 

@mattsse this allows us to have more observability into where we are spending most of our time

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -416,9 +419,25 @@ where
...
+                .record(spawn_payload_processor_start.elapsed().as_secs_f64());
+            handle
         } else {
-            self.payload_processor.spawn_cache_exclusive(env.clone(), txs, provider_builder)
+            let prewarming_start = Instant::now();
+            let handle =
+                self.payload_processor.spawn_cache_exclusive(env.clone(), txs, provider_builder);
+
+            // Record prewarming initialization duration
+            self.metrics
```
yongkangc, line 436: @Timosdev99 can we  move the timing outside the if-else? so we can avoid dup?

yongkangc (changes requested):

yongkangc (changes requested):
some extra nits / commments that i addded, but would like @mattsse to take a look as well :) 

@mattsse this allows us to have more observability into where we are spending most of our time

comment: 
https://github.com/paradigmxyz/reth/pull/18308/files#r2343600182

crates/engine/tree/src/tree/metrics.rs
Timosdev99: > @Timosdev99 can I check why did we remove this?

is it not measuring the same metrics with **state_root_duration** ?

crates/engine/tree/src/tree/payload_validator.rs
Timosdev99: if we move it outside wont the timing be inaccurate since we will have to start it before the process?

@mattsse any suggestion?

Rjected: I think we can't move the timing outside the if-else because otherwise the trie input calculation would be captured in the branch above, which can be fairly expensive. So IMO this is fine

crates/engine/tree/src/tree/metrics.rs
Rjected: it's not technically measuring the same thing, because one is a histogram (which records in percentile buckets) and the other one is a duration (recording a point-in-time metric), so we should add this back

Rjected (changes requested):
just one comment on adding back the state root duration, otherwise this looks good to me!

Rjected (approved):
This looks good to me, pending @yongkangc

yongkangc (approved):
lgtm, nice work @Timosdev99

PR #18311, opened 2025-09-06
fix: remove extra whitespace in let binding assignment

sentdeed:
Removes extra whitespace in macro to ensure consistent formatting across the codebase and prevent linter warnings.


mattsse (approved):
tysm

PR #18316, opened 2025-09-08
refactor: remove redundant double .into() conversion in eth core

aso20455:
Remove redundant double `.into().into()` conversion when initializing gas_cap field.

The function parameter is already `impl Into<GasCap>`, so only one `.into()` conversion is needed. The second `.into()` is redundant and adds unnecessary overhead.


mattsse (commented):
this is wrong

PR #18324, opened 2025-09-08
feat: cache latest built payload

### LINKED ISSUE #18303: Cache resolved payloads for a bit
### Describe the feature

it might happen that the service was unable to deliver the payload in type, however it already resolved it, so followup resolve call would come up empty:

https://github.com/paradigmxyz/reth/blob/e93e1fcecb94964cab5a81572c120694872e84eb/crates/payload/builder/src/service.rs#L288-L324

e.g.
getPayload requested, takes longer than expected, request issuer cancles request after 1s, and re-requests it, but we already terminated the job.

I think the termination of the job stems from interpreting

> Client software MAY stop the corresponding build process after serving this call.

as terminating it

https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_getpayloadv1

so ideally we cache this for a brief duration, or keep the job alive for a bit more, but caching makes sense here. we already kinda do something like this where we emit the payload as event:

https://github.com/paradigmxyz/reth/blob/e93e1fcecb94964cab5a81572c120694872e84eb/crates/payload/builder/src/service.rs#L313-L315

so we could send another clone back to the service to store it for a while

### Additional context

_No response_

klkvr:
Closes https://github.com/paradigmxyz/reth/issues/18303

Adds `watch` channel that keeps latest resolved payload and uses it for responding to timestamp/resolve requests.

### DIFF
--- a/crates/payload/builder/Cargo.toml
+++ b/crates/payload/builder/Cargo.toml
@@ -21,7 +21,7 @@ reth-ethereum-engine-primitives.workspace = true
 
 # alloy
 alloy-consensus.workspace = true
-alloy-primitives = { workspace = true, optional = true }
+alloy-primitives.workspace = true
 alloy-rpc-types = { workspace = true, features = ["engine"] }
 
 # async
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense, realistically we only need this for the most recent one, if this is insufficient we could turn this into a regular cache easily

PR #18325, opened 2025-09-08
chore: remove unused BlockHeader import in test_utils.rs

MozirDmitriy:
removed an unnecessary alloy_consensus::BlockHeader import from crates/exex/exex/src/backfill/test_utils.rs. The file uses the concrete Header type directly and does not rely on the BlockHeader trait bounds or trait methods. Keeping the unused import could cause lint warnings and add confusion for readers. No functional changes.


mattsse (commented):
this is the thrid time I'm closing this

PR #18327, opened 2025-09-08
perf: optimize Optimism deposit transaction prewarming

yongkangc:
## Summary

- Implements optimized prewarming for Optimism deposit transactions by broadcasting them to all worker tasks instead of using round-robin distribution.

Closes #18199

## Problem

On Optimism chains (Base, Optimism, etc.), the first transaction in each block is a deposit transaction that sets critical metadata:
- L1 block information (`setL1BlockValuesEcotone`)
- Base fee and blob base fee values
- Fee distribution to the fee vault (address 42011)

When using round-robin distribution, only one prewarm worker receives this transaction, leaving other workers with stale state that affects all subsequent transactions.

## Result / Impact on base mainnet

**Statistics:**

- Mean percent difference: -2.04%
- Median percent difference: -3.04%
- Standard deviation: 18.09%
- Min: -45.07%
- Max: 77.38%
- Total blocks analyzed: 2000

**Gas/s Statistics:**
 Comparison median gas/s: 449,797,679
- Gas/s percent change: +4.23%
- Baseline median gas/s: 431,549,763

latency_comparison

Mainnet: https://github.com/paradigmxyz/reth/pull/18327#issuecomment-3278652802


yongkangc:
remaining: cleanup + benching

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -91,24 +96,58 @@ where
             let mut handles = Vec::with_capacity(max_concurrency);
             let (done_tx, done_rx) = mpsc::channel();
             let mut executing = 0;
-            while let Ok(executable) = pending.recv() {
-                let task_idx = executing % max_concurrency;
+            let mut is_first_tx = true;
 
-                if handles.len() <= task_idx {
-                    let (tx, rx) = mpsc::channel();
```
mattsse, line 111: what we could do here is track if it's the first tx _and_ is 126 and then send to all tasks, this should be easier than doing is_optimism tracking seperately

mattsse (changes requested):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: thats a better way

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -284,15 +285,16 @@ where
...
         &self,
         env: ExecutionEnv<Evm>,
-        mut transactions: mpsc::Receiver<impl ExecutableTxFor<Evm> + Send + 'static>,
+        mut transactions: mpsc::Receiver<Tx>,
         provider_builder: StateProviderBuilder<N, P>,
         to_multi_proof: Option<Sender<MultiProofMessage>>,
     ) -> CacheTaskHandle
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
+        Tx: ExecutableTxFor<Evm> + Clone + Send + std::any::Any + 'static,
```
yongkangc, line 297: This is required because in prewarm.rs, we use std::any::Any for runtime type checking in the check_is_deposit_tx function. Not sure if theres a better way to go about this? 

cc @mattsse

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -27,6 +28,9 @@ use std::{
 };
 use tracing::{debug, trace};
 
+/// Optimism deposit transaction type (0x7E/126).
+const OPTIMISM_DEPOSIT_TX_TYPE: u8 = 126;
+
```
emhane: pls remove in favour of importing `DEPOSIT_TX_TYPE_ID` from op-alloy

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -91,24 +99,56 @@ where
...
+                    }
 
-                    executor.spawn_blocking(move || {
-                        ctx.transact_batch(rx, sender, done_tx);
-                    });
+                    // Send deposit transaction to ALL prewarm tasks
+                    // This ensures all workers have the updated L1 block info and fee parameters
+                    for handle in &handles {
+                        let _ = handle.send(executable.clone());
+                    }
```
emhane: instead of having `is_first_tx`, it makes sense to move all this code out of the while loop and before the while loop add
```rust
if let (Ok(first_executable)) = pending.recv() {
    if executable.tx().is_type(DEPOSIT_TX_TYPE_ID) {
        ...
    }
}
```
since `is_first_tx` never resets to true, and no code is shared between the if and else bodies except the single line `executing += 1;` which is ok to repeat

emhane (changes requested):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
emhane: ah perhaps it makes sense to have this double up for now, since we probably don't want to have op-alloy be a dep to any crate outside of the optimism directory...

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -91,24 +99,56 @@ where
...
+                    // Pre-spawn all workers for the deposit transaction broadcast
+                    for _ in handles.len()..max_concurrency {
+                        let (tx, rx) = mpsc::channel();
+                        let sender = actions_tx.clone();
+                        let ctx = ctx.clone();
+                        let done_tx = done_tx.clone();
+
+                        executor.spawn_blocking(move || {
+                            ctx.transact_batch(rx, sender, done_tx);
+                        });
```
emhane: this code could be moved into a method of the `PrewarmContext`, returning the Sender, since it's repeated below as well

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: yup thanks for the note! wanted to import a type if possible as well

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: thats a good idea, thanks for that

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: that makes sense, thanks! 

also gave me an idea to simplify the logic of prewarming by spawning workers to the max concurrency

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +82,68 @@ where
...
-                if handles.len() <= task_idx {
-                    let (tx, rx) = mpsc::channel();
-                    let sender = actions_tx.clone();
-                    let ctx = ctx.clone();
-                    let done_tx = done_tx.clone();
 
-                    executor.spawn_blocking(move || {
-                        ctx.transact_batch(rx, sender, done_tx);
-                    });
+            // Spawn all txn processing workers upfront as txns > max_concurrency (64) 99% of the time
```
yongkangc: https://dune.com/queries/3149880

The average txns is always > 64.

This makes the code much simpler, as we can just spawn all

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +82,68 @@ where
...
-                    let sender = actions_tx.clone();
-                    let ctx = ctx.clone();
-                    let done_tx = done_tx.clone();
 
-                    executor.spawn_blocking(move || {
-                        ctx.transact_batch(rx, sender, done_tx);
-                    });
+            // Spawn all txn processing workers upfront as txns > max_concurrency (64)
+            let mut handles = Vec::with_capacity(max_concurrency);
+            for _ in 0..max_concurrency {
```
mattsse: pre-spawning them is potentially a bit wasteful because we might have the case (on devnets/testnets) where concurrency is > number of transactions, so we should cap this to probably the executors size_hint.min(max_concurrency)

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +82,68 @@ where
...
+                handles.push(ctx.spawn_worker(&executor, actions_tx.clone(), done_tx.clone()));
+            }
 
-                    handles.push(tx);
+            // Handle first transaction - special case for Optimism deposit
+            if let Ok(first_tx) = pending.recv() {
+                if first_tx.tx().is_type(OPTIMISM_DEPOSIT_TX_TYPE) {
```
mattsse: what if we just check for > 7702_ID then this would also work for other sys tx types, but unsure if there are any

mattsse (changes requested):
tiny nit

since @klkvr previously worked on this would appreciate a review here as well

crates/engine/tree/src/tree/payload_processor/prewarm.rs
klkvr: this feels footgunish — I could see us forgetting about it once there's a new tx with 0x5 type

crates/engine/tree/src/tree/payload_processor/prewarm.rs
klkvr: yeah spawning all of the threads and making all of them execute the same thing might be pretty expensive

klkvr (commented):
I'm wondering if it would be cheaper to execute the deposit tx once in the main thread and then just share the state after it with all other tasks?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: that makes sense, thanks for that

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: @mattsse what do you think about the tradeoffs here for `> 7702_ID`? I would think that in the future we might want to have a collection of sys tx types and filter / handle them separately?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: so given @klkvr suggestion - what we could do is to have 1 worker execute the txns and that result would be stored in the cache. in that case we can lazily spawn x threads given x txns. 

implemented:[ffaa54d](https://github.com/paradigmxyz/reth/pull/18327/commits/ffaa54d37f9280227ff07395caddb2c3be9898d7)

emhane (approved):
nice!

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +78,60 @@ where
...
 
         self.executor.spawn_blocking(move || {
-            let mut handles = Vec::with_capacity(max_concurrency);
             let (done_tx, done_rx) = mpsc::channel();
             let mut executing = 0;
+
+            let mut handles = Vec::with_capacity(max_concurrency);
+
+            // We execute the first transaction first on main worker and share the state with the
+            // other workers via the cache. This is because on some chains like
```
mattsse: this isnt quite correct, the thing they share is the parent state, however the system tx changes the current state which isnt shared because for prewarming we can not rely on this being accurate.

because we prewarm batches of txs out of order they don't share any transactions that arent part of the batch

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +78,60 @@ where
...
+                handles.push(ctx.spawn_worker(&executor, actions_tx.clone(), done_tx.clone()));
+                if handles[0].send(first_tx.clone()).is_err() {
+                    warn!(
+                        target: "engine::tree::prewarm",
+                        tx_hash = %first_tx.tx().tx_hash(),
+                        "Failed to send deposit transaction to worker"
+                    );
+                };
+                executing += 1;
+            }
```
mattsse: I think the only change that this does now is just extract the first iteration of the recv loop

mattsse (changes requested):
left some comments about the cache assumptions this relies on

this change now doesnt share any tx1 across the individual prewarm workers so this wont have the expected impact

> I'm wondering if it would be cheaper to execute the deposit tx once in the main thread and then just share the state after it with all other tasks?

this isnt very feasible because then we'd need another wire from the main thread and this additional complexity is probably not worth it, especially if we assume that executing a deposit tx is just updating some storage values

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: @mattsse since we're receiving transactions through an `mpsc::Receiver`, we don't have access to a `size_hint()` method
https://github.com/paradigmxyz/reth/blob/cd8365e7cc711bd868b92e31fac229b53ddaa8bb/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L109
is there maybe an easier way to receive the count of txns that i might be missing?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +86,72 @@ where
...
-                    let done_tx = done_tx.clone();
+            // Initially cap workers based on transaction count hint
+            let mut handles = Vec::with_capacity(max_concurrency);
+            let workers_needed = transaction_count_hint.min(max_concurrency);
 
-                    executor.spawn_blocking(move || {
-                        ctx.transact_batch(rx, sender, done_tx);
-                    });
+            // Spawn the required number of workers
+            for _ in 0..workers_needed {
```
mattsse: now that I see this, do we even configure the max concurrency anywhere, because spawning 64 tasks is probably not ideal

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +86,72 @@ where
...
+                handles.push(ctx.spawn_worker(&executor, actions_tx.clone(), done_tx.clone()));
+            }
 
-                    handles.push(tx);
+            // Handle first transaction - special case for Optimism deposit / systems txs
+            if let Ok(first_tx) = pending.recv() {
+                if first_tx.tx().is_type(OPTIMISM_DEPOSIT_TX_TYPE) {
```
mattsse: we can do > 4 here because this will catch all non eth types and if its first position we can assume this is a sys tx

yongkangc: this makes sense, thanks for the note!
Addressed here [9387126](https://github.com/paradigmxyz/reth/pull/18327/commits/938712648543b8c0fff144c404b37e059ad1b0c4)

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: i think its mostly hardcoded here as 64 here
https://github.com/paradigmxyz/reth/blob/yk/op-txn-prewarming/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L89

Given that a normal txns prob average around 150 (mainnet) https://dune.com/queries/3149880 and base around 250 https://base.blockscout.com/blocks

16 would probably work?

yongkangc: in any case, we should limit the max_concurrency in prewarming to 16 right? 

In prewarm.rs: It's only used within the PrewarmCacheTask struct to limit the number of blocking threads for transaction prewarming

@mattsse

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +124,74 @@ where
...
+                executing += 1;
+            }
 
-                let _ = handles[task_idx].send(executable);
-
+            // Process remaining transactions with round-robin distribution
+            while let Ok(executable) = pending.recv() {
+                let task_idx = executing % workers_needed;
+                if handles[task_idx].send(executable).is_err() {
+                    warn!(
```
mediocregopher, line 212: Would be good to capture this error and include it in the WARN

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -25,7 +26,28 @@ use std::{
...
+/// Standard transaction types are:
+/// - Type 0: Legacy transactions (original Ethereum)
+/// - Type 1: EIP-2930 (access list transactions)
+/// - Type 2: EIP-1559 (dynamic fee transactions)
+/// - Type 3: EIP-4844 (blob transactions)
+/// - Type 4: EIP-7702 (set code authorization transactions)
+///
+/// Any transaction with a type > 4 is considered a non-standard/system transaction,
+/// typically used by L2s for special purposes (e.g., Optimism deposit transactions use type 126).
+const MAX_STANDARD_TX_TYPE: u8 = 4;
```
mediocregopher, line 64: Is this a number that chains using reth as an SDK are going to need to configure? I imagine other chains will have their own tx types.

mattsse: we're basically optimistically optimizing for a specific case where the first tx in a block is different, which indicates a system tx (op,arb,bera,..) which should be okay

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -25,7 +26,28 @@ use std::{
...
+
+/// Number of CPUs to reserve for other critical system tasks (e.g., block processing, networking).
+/// The prewarm task will use at most `min(available_cpus - RESERVED_CPUS, PREWARM_MAX_CONCURRENCY)` workers.
+const PREWARM_RESERVED_CPUS: usize = 2;
+
+/// Maximum concurrency for the prewarm task. Even if more CPUs are available we cap the workers to
+/// avoid over-saturating the system and to stay aligned with the historical limit.
+const PREWARM_MAX_CONCURRENCY: usize = 64;
```
mattsse: I dont like that we're having a default value that is so large,
are we not deriving this currently from anywhere?

we could perhaps use the executor's thread count somehow?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -25,7 +26,28 @@ use std::{
...
+/// - Type 3: EIP-4844 (blob transactions)
+/// - Type 4: EIP-7702 (set code authorization transactions)
+///
+/// Any transaction with a type > 4 is considered a non-standard/system transaction,
+/// typically used by L2s for special purposes (e.g., Optimism deposit transactions use type 126).
+const MAX_STANDARD_TX_TYPE: u8 = 4;
+
+/// Number of CPUs to reserve for other critical system tasks (e.g., block processing, networking).
+/// The prewarm task will use at most `min(available_cpus - RESERVED_CPUS, PREWARM_MAX_CONCURRENCY)` workers.
+const PREWARM_RESERVED_CPUS: usize = 2;
```
mattsse: this is another assumption that in theory makes sense but also not great to have these assumptions hardcoded

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -62,14 +86,36 @@ where
...
+                .max(1)
+                .min(PREWARM_MAX_CONCURRENCY)
+        });
+
+        trace!(
+            target: "engine::tree::prewarm",
+            max_concurrency,
+            transaction_count_hint,
+            "Initialized prewarm task"
+        );
```
mattsse: not a fan of doing this here, this should be a one time thing that we pass down via the engine's prewarm config

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +124,74 @@ where
...
-                let task_idx = executing % max_concurrency;
 
-                if handles.len() <= task_idx {
-                    let (tx, rx) = mpsc::channel();
-                    let sender = actions_tx.clone();
-                    let ctx = ctx.clone();
-                    let done_tx = done_tx.clone();
+            // Initially cap workers based on transaction count hint
+            let mut handles = Vec::with_capacity(max_concurrency);
+            let workers_needed = transaction_count_hint.min(max_concurrency);
```
mattsse: transaction_count_hint could be 0 here if the iterator doesnt know,

so we should handle this case

crates/engine/tree/src/tree/payload_processor/prewarm.rs
mattsse: yeah I think 16 is a much saner default here
but ideally have this configured during initialization once or derived from the executor

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -78,37 +124,74 @@ where
...
+            if let Ok(first_tx) = pending.recv() {
+                // Check if this is a system transaction (type > 4)
+                // System transactions in the first position typically set critical metadata
+                // that affects all subsequent transactions (e.g., L1 block info, fees on L2s).
+                if first_tx.tx().ty() > MAX_STANDARD_TX_TYPE {
+                    // Broadcast system transaction to all workers to ensure they have the
+                    // critical state. This is particularly important for L2s like Optimism
+                    // where the first deposit transaction contains essential block metadata.
+                    for handle in &handles {
+                        if handle.send(first_tx.clone()).is_err() {
```
mattsse: we should also send the index of the tx to the handle, e.g. sending outputs can always be skipped for the first tx because we assume the main execution handle will just be as fast

mattsse (changes requested):
more nits

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: this makes sense, technically we should preskip x txns, where x >1

previously tried experiments here
https://github.com/paradigmxyz/reth/pull/18122/files
findings: https://github.com/paradigmxyz/reth/issues/18092#issuecomment-3231996730

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -92,38 +128,90 @@ where
...
-                    let sender = actions_tx.clone();
-                    let ctx = ctx.clone();
-                    let done_tx = done_tx.clone();
+            // Initialize worker handles container
+            let mut handles = Vec::with_capacity(max_concurrency);
 
-                    executor.spawn_blocking(move || {
-                        ctx.transact_batch(rx, sender, done_tx);
-                    });
+            let workers_needed = transaction_count_hint.max(1).min(max_concurrency);
```
yongkangc: > we can do both limit the default to 16 but also check the tx_hint upper bound 

not sure if this is what you had in mind for the settings? @mattsse

yongkangc:
@mattsse thanks for the comments, addressed them - added in a better way to limit workers based on type hint and 16 + added in tx index 

not 100% sure if thats what u mean with the index?

mediocregopher (approved):
This LGTM but maybe @mattsse should double check I'm not missing something

yongkangc:
image
@mattsse benched and got this for reth7

crates/engine/tree/src/tree/payload_processor/prewarm.rs
mattsse: hmm, this might still just use 1 if the iterator doesnt have an upper bound and lower bound is 0, depending on the iterator, so we should treat 0 as unknown here

mattsse (changes requested):

yongkangc:
@mattsse just addressed this
https://github.com/paradigmxyz/reth/pull/18327#discussion_r2387015130

here: https://github.com/paradigmxyz/reth/pull/18327/commits/26ad92033eb6315729549536a62d3da367019892

let me know if this addresses your comment

mattsse (approved):

PR #18329, opened 2025-09-08
docs(static-file): fix wrapper reference to StaticFileProducerInner

radik878:
Replace incorrect doc reference to StaticFileProducer with StaticFileProducerInner.
This change is necessary to prevent confusion: StaticFileProducer<T> is an Arc<Mutex<StaticFileProducerInner<T>>> wrapper, so the documentation must correctly point to the inner type it wraps.


mattsse (commented):
this is fine

PR #18337, opened 2025-09-09
docs(db): fix cursor method name and remove undeclared requests_cursor

sashass1315:
Replace incorrect reference to DbTx::cursor() with the actual API method cursor_read(), matching crates/storage/db-api/src/transaction.rs.
Remove the outdated/uncompiled requests_cursor lines in the unwind example; that cursor was never declared and no corresponding table exists in the shown context. The snippet likely drifted from older code and contradicts the current BodyStage unwind flow.
These changes align the docs with the real API and code, preventing reader confusion and copy-paste errors.


mattsse (approved):
this is okay

PR #18341, opened 2025-09-09
fix: NoopNetworkBuilder and run main

famouswizard:


mattsse (commented):
this requires eth types so we can use the builtin eth types

PR #18343, opened 2025-09-09
fix: tx_events usage and FullNodeComponents generics

mdqst:
replaced `tx_events.hash()` with `tx_hash` and updated `FullNodeComponents` generics so the code compiles.


mattsse (commented):
I prefer the previous style here

PR #18347, opened 2025-09-09
fix(trie): correct NodeNotFoundInProvider error to say “revealing”

VolodymyrBg:
The error is raised when the provider fails to return a blinded node during the revealing step, not during removal.
Call sites in sparse and sparse-parallel trigger this when fetching nodes via provider to reveal children (including in remove flows before structural changes).
Aligns the error text with actual control flow and existing comments (“during revealing”)


mattsse (commented):
i think this is fine

PR #18348, opened 2025-09-09
fix(prune): TransactionLookup pruning issues with pre-merge expiry

mediocregopher:
This fixes an issue with pre-merge expiry and TransactionLookup pruning.

When pre-merge history has already been expired, and
   `--prune.transactionlookup.distance` is enabled for the first time,
   it is not possible for the pruner to prune the lookups because it is
   missing all pre-merge txs, and so cannot index into the
   TransactionHashNumbers table by hash to delete them.

   The fix is to fallback to the Transaction prune checkpoint when the
   TransactionLookup checkpoint is not present or lagging behind. This way if txs have
   been previously pruned the TransactionLookup pruning will start
   there.

### DIFF
--- a/crates/prune/prune/src/segments/user/transaction_lookup.rs
+++ b/crates/prune/prune/src/segments/user/transaction_lookup.rs
@@ -6,9 +6,9 @@ use crate::{
 use alloy_eips::eip2718::Encodable2718;
 use rayon::prelude::*;
 use reth_db_api::{tables, transaction::DbTxMut};
-use reth_provider::{BlockReader, DBProvider};
+use reth_provider::{BlockReader, DBProvider, PruneCheckpointReader};
 use reth_prune_types::{PruneMode, PrunePurpose, PruneSegment, SegmentOutputCheckpoint};
-use tracing::{instrument, trace};
+use tracing::{debug, instrument, trace};
 
 #[derive(Debug)]
 pub struct TransactionLookup {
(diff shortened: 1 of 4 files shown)


crates/prune/prune/src/segments/user/transaction_lookup.rs
```diff
@@ -38,7 +39,24 @@ where
...
-    fn prune(&self, provider: &Provider, input: PruneInput) -> Result<SegmentOutput, PrunerError> {
+    fn prune(
+        &self,
+        provider: &Provider,
+        mut input: PruneInput,
+    ) -> Result<SegmentOutput, PrunerError> {
+        // If TransactionLookup pruning is enabled on a node which had history pruning already
+        // enabled, then we can't start TransactionLookup pruning from 0 because there is no tx data
+        // from then. Instead we fallback to the Transactions prune checkpoint.
+        if input.previous_checkpoint.is_none() {
```
shekhirin: I think we want to do something like

```rust
input.previous_checkpoint.is_none_or(|checkpoint| checkpoint.block_number < provider.get_prune_checkpoint(PruneSegment::Transactions).block_number)
```

so that it also works when user restarts the node with and without tx lookup pruning but always with pre-merge expiry

shekhirin (approved):

mattsse (approved):
sigh, good find

PR #18352, opened 2025-09-09
fix: remove unnecessary unsafe block in main.rs

prestoalvarez:
Remove redundant `unsafe` wrapper around `std::env::set_var()` call in the main binary entry point.

**Problem:**
- `std::env::set_var()` is a safe function and doesn't require `unsafe`
- Unnecessary `unsafe` blocks create confusion and increase audit complexity
- Risk of incorrect unsafe usage being copied to this location in the future

**Solution:**
- Remove the `unsafe` block while keeping the same functionality
- Maintains identical behavior with cleaner, safer code


mattsse (commented):
this is required in 2024 now

PR #18355, opened 2025-09-10
fix: add is_osaka check before erroring in default_ethereum_payload

rezzmah:
Currently we don't check for `is_osaka` before erroring in `default_ethereum_payload`. 

This could result in a scenario where valid full blocks are discarded by the proposer before osaka

The below check exists but only helps prevent erroring after osaka
```rust
        if is_osaka && estimated_block_size_with_tx > MAX_RLP_BLOCK_SIZE {
            best_txs.mark_invalid(
                &pool_tx,
                InvalidPoolTransactionError::OversizedData(
                    estimated_block_size_with_tx,
                    MAX_RLP_BLOCK_SIZE,
                ),
            );
            continue;
        }
```

### DIFF
--- a/crates/ethereum/payload/src/lib.rs
+++ b/crates/ethereum/payload/src/lib.rs
@@ -271,7 +271,7 @@ where
                     break 'sidecar Err(Eip4844PoolTransactionError::MissingEip4844BlobSidecar)
                 };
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
nice catch

but the size check should hold universally

PR #18356, opened 2025-09-10
feat(engine): check header validity after invalid transaction

fgimenez:
After recent changes towards improving performance, transactions are executed early in the block processing flow. if we get an error in transaction execution we return it right away. this was causing some eest tests to fail like:

`2025-08-21T19:56:43.2662135Z E   pytest_plugins.consume.simulators.simulator_logic.test_via_engine.LoggedError: Client returned unexpected validation error: got:
  "[TransactionException.INSUFFICIENT_MAX_FEE_PER_BLOB_GAS] EVM reported invalid transaction (0x3de929cecd7e4b3ea27f672acb8ba0d9f51316c1d35c802c44a0c333cea0c400): blob gas price is 
  greater than max fee per blob gas" expected: "BlockException.INCORRECT_EXCESS_BLOB_GAS"`

With both an invalid header and an invalid transaction, the test expects to receive the exception about the failure validating the header, and we are returning the tx execution failure instead.

With the changes in this PR, when a transaction fails, we run several validations that are now postponed, returning an eventual error on them instead of the tx error. This makes the test pass without affecting performance in the happy path

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -270,6 +270,41 @@ where
         }
     }
 
+    /// Handles execution errors by checking if header validation errors should take precedence.
+    ///
+    /// When an execution error occurs, this function checks if there are any header validation
+    /// errors that should be reported instead, as header validation errors have higher priority.
+    fn handle_execution_error<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
+        &self,
+        input: BlockOrPayload<T>,
+        execution_err: InsertBlockErrorKind,
+        parent_block: &SealedHeader<N::BlockHeader>,
+    ) -> Result<ExecutedBlockWithTrieUpdates<N>, InsertPayloadError<N::Block>>
+    where
+        V: PayloadValidator<T, Block = N::Block>,
+    {
+        // If execution failed, we should first check if there are any header validation
+        // errors that take precedence over the execution error
+        let block = self.convert_to_block(input)?;
+
+        // Validate block consensus rules which includes header validation
+        if let Err(consensus_err) = self.validate_block_inner(&block) {
+            // Header validation error takes precedence over execution error
+            return Err(InsertBlockError::new(block.into_sealed_block(), consensus_err.into()).into())
+        }
+
+        // Also validate against the parent
+        if let Err(consensus_err) =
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -421,13 +421,54 @@ where
...
+                let result = self.execute_block(&state_provider, env, &input, &mut handle);
+                state_provider.record_total_latency();
+                result
+            } else {
+                self.execute_block(&state_provider, env, &input, &mut handle)
+            };
+
+            match execution_result {
+                Ok(result) => result,
+                Err(execution_err) => {
```
mattsse: this function is already gigantic,

could we extract this error handling to a separate fn instead so that we can do

```
let Ok() else {return on_error}
```

mattsse (changes requested):
this makes sense because we currently would return the execution errors and should instead fully validate the block/payload

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: makes sense done here https://github.com/paradigmxyz/reth/pull/18356/commits/855a19d8044db838b125638ccc13444600b68392

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -270,6 +270,41 @@ where
...
+    /// errors that should be reported instead, as header validation errors have higher priority.
+    fn handle_execution_error<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
+        &self,
+        input: BlockOrPayload<T>,
+        execution_err: InsertBlockErrorKind,
+        parent_block: &SealedHeader<N::BlockHeader>,
+    ) -> Result<ExecutedBlockWithTrieUpdates<N>, InsertPayloadError<N::Block>>
+    where
+        V: PayloadValidator<T, Block = N::Block>,
+    {
```
mattsse, line 285: could we add a debug! here that spits out the execution error

mattsse (approved):
lgtm, one suggestion for debug traces

crates/engine/tree/src/tree/payload_validator.rs
fgimenez: sure https://github.com/paradigmxyz/reth/pull/18356/commits/4c904b922ce02cefc64e3174d64b7a8a5a86b7be

PR #18359, opened 2025-09-10
fix(chain-state): reset in-memory earliest/latest gauges when empty

Fibonacci747:
Reset earliest_block and latest_block gauges to 0 when the in-memory numbers map is empty. Previously, these gauges were only updated when entries existed, leaving stale values after clear_state() or remove_persisted_blocks(). This change aligns with gauge reset patterns elsewhere (e.g., cache/downloader metrics), prevents misleading telemetry, and ensures dashboards accurately reflect an empty in-memory chain state.
also cargo fmt


mattsse (commented):
this isnt necessary, we dont expect this to be empty, and this can still be inferred from the `num_blocks` value

PR #18362, opened 2025-09-10
fix: replace tx_hash method with TxHashRef trait bound (#18357)

### LINKED ISSUE #18357: replace SignedTx::tx_hash with `: TxHashRef` instead
### Describe the feature

we still have this fn

https://github.com/paradigmxyz/reth/blob/a3aaccd34a8d2cfea373a1368514caa601f394df/crates/primitives-traits/src/transaction/signed.rs#L49-L50

newer alloy impl now has `TxHashRef`

we can remove the existing tx_hash function and instead add a bound to:
https://github.com/paradigmxyz/reth/blob/a3aaccd34a8d2cfea373a1368514caa601f394df/crates/primitives-traits/src/transaction/signed.rs#L31-L31

### Additional context

_No response_

eeemmmmmm:
## Changes:
- Removed the `tx_hash()` method from `SignedTransaction` trait
- Added `TxHashRef` trait bound to `SignedTransaction` instead  
- Deleted all manual `tx_hash()` implementations for different transaction types
- Added necessary imports and trait implementations where needed

Closes: https://github.com/paradigmxyz/reth/issues/18357


crates/primitives-traits/src/block/body.rs
```diff
@@ -64,7 +64,7 @@ pub trait BlockBody:
     ///
     /// This is a convenience function for `transactions_iter().find()`
     fn transaction_by_hash(&self, hash: &B256) -> Option<&Self::Transaction> {
-        self.transactions_iter().find(|tx| tx.tx_hash() == hash)
+        self.transactions_iter().find(|tx| *tx.tx_hash() == *hash)
```
mattsse: do we need to change this, these should be the same &B256

mattsse (changes requested):

mattsse:
we're also missing some new imports now, see ci failures, recommend running

```
 cargo +nightly clippy --bin op-reth --workspace --lib --examples --tests --benches --locked --all-features
```

which should catch everything

mattsse (approved):

PR #18364, opened 2025-09-10
fix(reth-bench): requests param of newPayloadV4

s1na:
I faced this error while trying out reth-bench:

```
  thread 'main' panicked at bin/reth-bench/src/main.rs:39:10:
  called `Result::unwrap()` on an `Err` value: server returned an error response: error code -32602: invalid argument 3: json: cannot unmarshal string into Go value of type
  []hexutil.Bytes
```

It looks like the requestsHash is being passed to newPayloadV4 instead of the full list.

### DIFF
--- a/bin/reth-bench/src/valid_payload.rs
+++ b/bin/reth-bench/src/valid_payload.rs
@@ -174,7 +174,11 @@ pub(crate) fn block_to_new_payload(
                             payload,
                             cancun.versioned_hashes.clone(),
                             cancun.parent_beacon_block_root,
-                            prague.requests.requests_hash(),
+                            if let alloy_eips::eip7685::RequestsOrHash::Requests(requests) = &prague.requests {
+                                requests.iter().collect::<Vec<_>>()
+                            } else {
+                                Vec::new()
+                            },
                         ))?,
                     )
                 }


mattsse (commented):
@shekhirin @Rjected 

I believe this script only works together with our engine setting for accepting hashes instead of the full requests

https://github.com/paradigmxyz/reth/blob/700f2e101aa50bf7ead662a4c604641236cc90f6/crates/node/core/src/args/engine.rs#L62-L64

Because we dont get them from the block, but we also only need the hash for validation -.-

shekhirin:
correct, the issue is that RPCs do not return raw requests, only the hash

```json
{
  "hash": "0xf1054c7a8e6d38c17249b24ce41bec67e7d13a8de82a9db36edc5d2187c6ced0",
  "parentHash": "0x59c62f94233947f13bd5d5ac086cfd259342cf5d4dec22bb1aa8e0a44990f362",
  "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
  "miner": "0xdadb0d80178819f2319190d340ce9a924f783711",
  "stateRoot": "0x68457a4eaf2bf53315aaea5bb8a74b78d92ecd3517be09614adac90cba6eef70",
  "transactionsRoot": "0x90dd33f0047da1a7e9f879d4b9ec986d3a87854d45570ce3a0f8748afa74b9ed",
  "receiptsRoot": "0xc7d2b8d5e9f3cef9e72334bcc520f91324e92beae2f20f389d01a42d36d8edc2",
  "logsBloom": "0xfbe3f6fbdff6efaf3e6bfffdf7ccff37f57bfff7dc1ff55de7fdae5c3cff67bbb93f73cffb3f739f7ebf3fd7dfaebdb5f7ffd3bfffe7effd52bbff733f7efffcffa6edefeef7daefa95fdfcf956ee5b7f7db7ef78a7fef7f6fa6fc7fcbc1b3d34ad73bbebebcd3fa9f73dfeefeaffbeaffefdcffdef7fe7ff67fdfd72bfdfeefbf7fffdf53bf53bdeeff9ffa8f2edc77dd6b35ff7b6bb7acffed6dfa7df7ffe7bfcfed95d53e7ddf7e7b0fdefb97eef37fc7dffcff9aeaabbf6f7fff6fe6c9fe5f7bffff7bfaff6dfdffe673ebfffbe4e77eff4bb7ef7dd6fdbfebea7acff3bf79917d3ef79f69be8bf7fc7fefe7bafbeffff0f53fdd79f7bc7ab6fbeffd9eed",
  "difficulty": "0x0",
  "number": "0x163ee36",
  "gasLimit": "0x2aea515",
  "gasUsed": "0x1182ae8",
  "timestamp": "0x68c03f57",
  "extraData": "0x4275696c6465724e6574202842656176657229",
  "mixHash": "0x744a47f87693bf6aecc8a4e08f8496bfdf8441d3b1967c13232f0b22eaf18709",
  "nonce": "0x0000000000000000",
  "baseFeePerGas": "0x56aec854",
  "withdrawalsRoot": "0x826fa147439cd4bf98beeb687048d5d8de8e0e2d7f1b67a00252e3d34730ae83",
  "blobGasUsed": "0x0",
  "excessBlobGas": "0x180000",
  "parentBeaconBlockRoot": "0xbe33df9117a185c35180384cddb3a38eb7c62ecc9b3400e29aefe8856dd563a4",
  "requestsHash": "0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "size": "0x1dc5f",
  "uncles": [],
  "transactions": [...],
  "withdrawals": [...]
}
```

s1na:
I see the issue yeah. It's annoying that requests are not fetchable from the rpc, and seeing now we don't even persist them to be able to return via rpc.

It seems like the correct approach is to fetch the requests from the beacon API. I will close the PR.

PR #18365, opened 2025-09-10
feat(rpc): Reuse evm tracing utils for debug_trace_block

namn-grg:
Solves #17045


namn-grg:
can I get a review on this @mattsse @Rjected @RomanHodulak

PR #18366, opened 2025-09-10
feat: add helper aliases for node adapters

klkvr:
Adds `FullNodeFor` and `NodeHandleFor` aliases allowing to easily define full adapter type given just a `Node` implementation

### DIFF
--- a/crates/node/builder/Cargo.toml
+++ b/crates/node/builder/Cargo.toml
@@ -19,7 +19,7 @@ reth-cli-util.workspace = true
 reth-config.workspace = true
 reth-consensus-debug-client.workspace = true
 reth-consensus.workspace = true
-reth-db = { workspace = true, features = ["mdbx"], optional = true }
+reth-db = { workspace = true, features = ["mdbx"] }
 reth-db-api.workspace = true
 reth-db-common.workspace = true
 reth-downloaders.workspace = true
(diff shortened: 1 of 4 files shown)


crates/node/builder/Cargo.toml
```diff
@@ -19,7 +19,7 @@ reth-cli-util.workspace = true
 reth-config.workspace = true
 reth-consensus-debug-client.workspace = true
 reth-consensus.workspace = true
-reth-db = { workspace = true, features = ["mdbx"], optional = true }
+reth-db = { workspace = true, features = ["mdbx"] }
```
mattsse, line 22: this seems okay, this is in the dep graph anyway

examples/custom-inspector/src/main.rs
```diff
@@ -36,8 +36,10 @@ fn main() {
     Cli::<EthereumChainSpecParser, RethCliTxpoolExt>::parse()
         .run(|builder, args| async move {
             // launch the node
-            let NodeHandle { node, node_exit_future } =
-                builder.node(EthereumNode::default()).launch().await?;
+            let handle = builder.node(EthereumNode::default()).launch().await?;
+
+            let node: FullNodeFor<EthereumNode> = handle.node;
```
mattsse, line 41: are these just for testing, because nothing changed here so we dont really need type hints here, right?

mattsse (approved):
lgtm, one q re additional type hints

examples/custom-inspector/src/main.rs
klkvr: yeah, this is just for testing

PR #18367, opened 2025-09-10
refactor: keep only task_executor.clone()

mdqst:
dropped `Box::new(task_executor.clone())` - `RethRessProtocolProvider::new` handles it fine.

code works the same, just a bit cleaner and leaner.


mattsse (commented):
this doesnt work, because we need dyn here

PR #18374, opened 2025-09-10
feat(op-reth): specialize get_transaction_receipt to check pending flashblocks

Nathy-bajo:
Reolves #18371 

This PR specializes the `get_transaction_receipt` endpoint for OP-Reth to also lookup transactions in the pending flashblock when they are not found in historical data. With flashblocks, the pending block should be considered as "confirmed/mined" for transaction receipt purposes.

### DIFF
--- a/crates/optimism/rpc/src/eth/transaction.rs
+++ b/crates/optimism/rpc/src/eth/transaction.rs
@@ -1,33 +1,34 @@
 //! Loads and formats OP transaction RPC response.
 
 use crate::{OpEthApi, OpEthApiError, SequencerClient};
-use alloy_consensus::transaction::TransactionMeta;
-use alloy_consensus::BlockHeader;
 use alloy_primitives::{Bytes, B256};
 use alloy_rpc_types_eth::TransactionInfo;
 use op_alloy_consensus::{transaction::OpTransactionInfo, OpTransaction};
 use reth_optimism_primitives::DepositReceipt;
-use reth_primitives_traits::SignedTransaction;
-use reth_rpc_eth_api::FromEvmError;
-use reth_rpc_eth_api::RpcNodeCore;
+use reth_primitives_traits::{NodePrimitives, SignedTransaction};
 use reth_rpc_eth_api::{
     helpers::{spec::SignersForRpc, EthTransactions, LoadReceipt, LoadTransaction},
-    try_into_op_tx_info, FromEthApiError, RpcConvert, RpcReceipt, TxInfoMapper,
+    try_into_op_tx_info, FromEthApiError, FromEvmError, RpcConvert, RpcNodeCore, RpcReceipt,
+    TxInfoMapper,
 };
 use reth_rpc_eth_types::utils::recover_raw_transaction;
-use reth_storage_api::TransactionsProvider;
-use reth_storage_api::{errors::ProviderError, ReceiptProvider};
+use reth_storage_api::{errors::ProviderError, ReceiptProvider, TransactionsProvider};
 use reth_transaction_pool::{
     AddedTransactionOutcome, PoolTransaction, TransactionOrigin, TransactionPool,
 };
-use std::fmt::{Debug, Formatter};
-use std::future::Future;
+use std::{
+    fmt::{Debug, Formatter},
+    future::Future,
+};
 
 impl<N, Rpc> EthTransactions for OpEthApi<N, Rpc>
 where
-    N: RpcNodeCore,
+    N: RpcNodeCore<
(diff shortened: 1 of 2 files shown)


crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -50,7 +58,7 @@ where
                 tracing::warn!(target: "rpc::eth", %err, %hash, "successfully sent tx to sequencer, but failed to persist in local tx pool");
             });
 
-            return Ok(hash)
+            return Ok(hash);
```
mattsse:
```suggestion
            return Ok(hash)
```

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -62,11 +70,81 @@ where
...
+                                    excess_blob_gas: block.excess_blob_gas(),
+                                    timestamp: block.timestamp(),
+                                };
+
+                                // Convert the transaction to the expected type
+                                let provider_tx = this
+                                    .provider()
+                                    .transaction_by_hash(hash)
+                                    .map_err(Self::Error::from_eth_err)?
+                                    .ok_or(EthApiError::TransactionNotFound)?;
```
mattsse: we already have the transaction for this in the block `tx`

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -62,11 +70,81 @@ where
...
+                        let block = pending_block.0;
+                        let receipts = pending_block.1;
+
+                        if let Some((index, (signer, tx))) = block
+                            .transactions_with_sender()
+                            .enumerate()
+                            .find(|(_, (_, tx))| *tx.tx_hash() == hash)
+                        {
+                            // Get the corresponding receipt
+                            if let Some(receipt) = receipts.get(index) {
```
mattsse: we likely want more resuable logic here,

we could consider introducing a dedicated type instead of the tuple next

mattsse (changes requested):
cool, smol suggestions

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -84,16 +83,19 @@ where
     {
         let this = self.clone();
         async move {
+            // First try the historical lookup (original behavior)
```
mattsse: your prompt is leaking into comments

the reader has no idea what the (original behaviour) is after merge this

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -111,25 +113,34 @@ where
...
+                                // Convert the receipt to the expected provider receipt type
+                                let provider_receipt = this
+                                    .provider()
+                                    .receipt_by_hash(hash)
+                                    .map_err(Self::Error::from_eth_err)?
+                                    .ok_or_else(|| {
+                                        Self::Error::from_eth_err(
+                                            reth_rpc_eth_types::EthApiError::TransactionNotFound,
+                                        )
+                                    })?;
```
mattsse: this isnt correct,

we already have both block and receipt we need to try find them from there, this we already tried (original behavior)

mattsse (changes requested):

crates/optimism/rpc/src/eth/transaction.rs
Nathy-bajo: Oh, I see.

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -5,22 +5,30 @@ use alloy_primitives::{Bytes, B256};
...
 
 impl<N, Rpc> EthTransactions for OpEthApi<N, Rpc>
 where
-    N: RpcNodeCore,
+    N: RpcNodeCore<
+        Provider: TransactionsProvider<Transaction = <N::Primitives as NodePrimitives>::SignedTx>,
+    >,
+    OpEthApiError: FromEvmError<N::Evm>,
     Rpc: RpcConvert<Primitives = N::Primitives, Error = OpEthApiError>,
+    Self: RpcNodeCore<Primitives = N::Primitives, Pool = N::Pool>,
```
mattsse: @klkvr had to restrict this like this so we can reuse treat `OpEthApi as RpcNodeCore == N`
but perhaps there's a better way?

mattsse (approved):
pending @klkvr

klkvr (approved):

crates/optimism/rpc/src/eth/transaction.rs
klkvr: simplified a bit in 13ee643

yongkangc (approved):

PR #18375, opened 2025-09-10
refactor!: more type-safety in cli

klkvr:
Changes `CliNodeComponents` to require the same types as node's configured components to be provided


mattsse (approved):
this makes sense, we specifically use this for CLI so this is a nicer workaround

PR #18376, opened 2025-09-11
fix(logging): add drop tracking and configurable behavior to file logging

WilliamNwoke:
Attempts to resolve https://github.com/paradigmxyz/reth/issues/18261

- Implemented NonBlockingDropTracking writer that tracks dropped messages
- Added configurable behavior (Drop or Block) for when buffer is full
- Added warning system that alerts users when messages start being dropped
- Maintained backward compatibility with default drop behavior


WilliamNwoke:
@yongkangc All done. Sorry for the late response, for some reason I did not get a notification form your reply.

PR #18381, opened 2025-09-11
chore: use decode_2718_exact for recover raw txs

leniram159:
close https://github.com/paradigmxyz/reth/issues/18380

### DIFF
--- a/crates/optimism/payload/src/payload.rs
+++ b/crates/optimism/payload/src/payload.rs
@@ -91,8 +91,7 @@ impl<T: Decodable2718 + Send + Sync + Debug + Unpin + 'static> PayloadBuilderAtt
             .unwrap_or_default()
             .into_iter()
             .map(|data| {
-                let mut buf = data.as_ref();
-                let tx = Decodable2718::decode_2718_exact(&mut buf)?;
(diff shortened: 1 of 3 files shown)


crates/optimism/payload/src/payload.rs
```diff
@@ -91,14 +91,9 @@ impl<T: Decodable2718 + Send + Sync + Debug + Unpin + 'static> PayloadBuilderAtt
...
-                let mut buf = data.as_ref();
-                let tx = Decodable2718::decode_2718(&mut buf).map_err(alloy_rlp::Error::from)?;
+                let tx = Decodable2718::decode_2718_exact(data.as_ref())?;
 
-                if !buf.is_empty() {
-                    return Err(alloy_rlp::Error::UnexpectedLength);
-                }
-
-                Ok(WithEncoded::new(data, tx))
+                Ok::<_, alloy_eips::eip2718::Eip2718Error>(WithEncoded::new(data, tx))
```
mattsse: we can use .map here and then avoid the `Ok::<_, alloy_eips::eip2718::Eip2718Error>` type hint

mattsse (changes requested):

mattsse (approved):

PR #18382, opened 2025-09-11
fix: still use real chain id for no-op network

hai-rise:
As RPC still reads the chain id via the `Network` component:

https://github.com/paradigmxyz/reth/blob/60568cca8fd2496a490aba407ae0f897d06a0914/crates/rpc/rpc-eth-api/src/helpers/spec.rs#L46-L49

### DIFF
--- a/crates/net/network-api/src/noop.rs
+++ b/crates/net/network-api/src/noop.rs
@@ -31,6 +31,7 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct NoopNetwork<Net = EthNetworkPrimitives> {
+    chain_id: u64,
     peers_handle: PeersHandle,
     _marker: PhantomData<Net>,
 }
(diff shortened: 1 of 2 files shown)


crates/net/network-api/src/noop.rs
```diff
@@ -31,24 +31,23 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
...
 pub struct NoopNetwork<Net = EthNetworkPrimitives> {
+    chain_id: u64,
     peers_handle: PeersHandle,
     _marker: PhantomData<Net>,
 }
 
 impl<Net> NoopNetwork<Net> {
     /// Creates a new [`NoopNetwork`].
-    pub fn new() -> Self {
+    pub fn new(chain_id: u64) -> Self {
```
mattsse: could we keep this as is and instead have a

```
with_chain_id()
```

fn?

mattsse (changes requested):
smol nit

crates/net/network-api/src/noop.rs
hai-rise: Ok, done!

mattsse (approved):

PR #18384, opened 2025-09-11
feat(op-reth): implement flashblock caching with block selection logic

Dyslex7c:
Closes #18373 

This PR adds a flashblock caching mechanism to store the last 3 completed flashblock sequences by block number. The `build_args` method is modified to select flashblocks based on local chain tip + 1. This enables building blocks when flashblocks arrive out-of-order relative to local chain progression, with new methods `ready_transactions_for_block()` and `payload_base_for_block()` handling retrieval from both current sequence and cache.

Also I'm unable to test this PR locally, should I write unit tests for it? I have ran the op-reth node with flashblocks enabled but I can't figure out from the logs whether the implementation is correct. I'm running the following command:

```
op-reth node \
--chain base-sepolia \
--rollup.sequencer-http https://sepolia-sequencer.base.org \
--flashblocks-url "wss://sepolia.flashblocks.base.org/ws" \
--http \
--ws \
--authrpc.port 9551 \
--authrpc.jwtsecret ./jwt.hex
```

### DIFF
--- a/crates/optimism/flashblocks/src/sequence.rs
+++ b/crates/optimism/flashblocks/src/sequence.rs
@@ -3,12 +3,15 @@ use alloy_eips::eip2718::WithEncoded;
 use core::mem;
 use eyre::{bail, OptionExt};
 use reth_primitives_traits::{Recovered, SignedTransaction};
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, VecDeque};
 use tokio::sync::broadcast;
 use tracing::{debug, trace, warn};
 
 /// The size of the broadcast channel for completed flashblock sequences.
 const FLASHBLOCK_SEQUENCE_CHANNEL_SIZE: usize = 128;
+
+/// Maximum number of block sequences to cache
+const FLASHBLOCK_CACHE_SIZE: usize = 3;
 
 /// An ordered B-tree keeping the track of a sequence of [`FlashBlock`]s by their indices.
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -20,6 +23,8 @@ pub(crate) struct FlashBlockPendingSequence<T> {
     inner: BTreeMap<u64, PreparedFlashBlock<T>>,
     /// Broadcasts flashblocks to subscribers.
     block_broadcaster: broadcast::Sender<FlashBlockCompleteSequence>,
+    /// Cache of completed sequences by block number (last 3 blocks)
+    cache: VecDeque<(u64, BTreeMap<u64, PreparedFlashBlock<T>>)>,
```
mattsse: I think the can just use `FlashBlockCompleteSequence` here because we expect that we will receive the entire sequence and then have `Vecdeque<FlashBlockCompleteSequence>` here

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -119,6 +161,23 @@ where
         self.inner.values().next()?.block().base.clone()
     }
 
+    /// Get payload base for a specific block number
+    pub(crate) fn payload_base_for_block(&self, target_block_number: u64) -> Option<ExecutionPayloadBaseV1> {
```
mattsse: we need to go by hash here not number because the fb could in theory be built on top a different block so we need to check base.parent_hash == local.head

mattsse (changes requested):
cool, good start

crates/optimism/flashblocks/src/sequence.rs
Dyslex7c: We can do that but then we'll also have to find a way to access prepared txs from `FlashBlockCompleteSequence` which will be required in the method `ready_transactions_for_block`. Should I just store `prepared_transactions` in `FlashBlockCompleteSequence`? For now in the PR I'm doing this, let me know if change is required

Dyslex7c:
@emhane yes I was waiting for @mattsse's review

PR #18395, opened 2025-09-11
feat: replace PendingBlockAndReceipts tuple with dedicated struct

stevencartavia:
closes #18394

### DIFF
--- a/crates/optimism/rpc/src/eth/pending_block.rs
+++ b/crates/optimism/rpc/src/eth/pending_block.rs
@@ -4,15 +4,15 @@ use std::sync::Arc;
 
 use crate::{OpEthApi, OpEthApiError};
 use alloy_eips::BlockNumberOrTag;
-use reth_primitives_traits::RecoveredBlock;
 use reth_rpc_eth_api::{
     helpers::{pending_block::PendingEnvBuilder, LoadPendingBlock},
     FromEvmError, RpcConvert, RpcNodeCore,
 };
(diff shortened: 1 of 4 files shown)


crates/optimism/rpc/src/eth/pending_block.rs
```diff
@@ -38,15 +38,9 @@ where
     /// Returns the locally built pending block
     async fn local_pending_block(
         &self,
-    ) -> Result<
-        Option<(
-            Arc<RecoveredBlock<ProviderBlock<Self::Provider>>>,
-            Arc<Vec<ProviderReceipt<Self::Provider>>>,
-        )>,
-        Self::Error,
-    > {
```
mattsse, line 47: so much better

mattsse (approved):

crates/rpc/rpc-eth-types/src/pending_block.rs
```diff
@@ -84,7 +84,13 @@ pub type PendingBlockReceipts<N> = Arc<Vec<<N as NodePrimitives>::Receipt>>;
 
 /// A type alias for a pair of an [`Arc`] wrapped [`RecoveredBlock`] and a vector of
 /// [`NodePrimitives::Receipt`].
```
kevaundray, line 86: looks like this comment needs to be updated

kevaundray: (Not blocking)

PR #18396, opened 2025-09-12
feat(pool): Include BlockInfo in TransactionValidationOutcome

developeruche:
closes #18383


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -364,7 +364,7 @@ where
             elements.push(pooled.into_inner());
 
             if limit.exceeds(size) {
-                break
+                break;
```
mattsse:
```suggestion
                break
```

mattsse (changes requested):
this is looking good already,

let's undo the additional ;

PR #18413, opened 2025-09-13
perf: downsize mempool tx priority from `U256` to `u128`

hai-rise:
Found this in our hot paths. We may need to restructure a bunch for `/// NOTE: The implementation is incomplete for missing base fee.`, but this is a quick win for now..


mattsse (approved):
makes sense

cool, this should make ordering txs slightly more efficient

PR #18414, opened 2025-09-13
chore: Fix misleading assertion messages

jenniferzelo:
This PR corrects two assertion failure messages in crates/chainspec/src/spec.rs so the “expected” and “got” placeholders reference the right variables.


mattsse (commented):
these are just tests so this is fine

PR #18415, opened 2025-09-13
chore(storage): remove unused `primed_dbis`

hai-rise:
Remove the unused `primed_dbis` that don't make much sense in the first place.

P/s: This is a code-deletion only PR. If you check the diff with `w=1`, the only addition is an auto-formatted line.

### DIFF
--- a/crates/storage/libmdbx-rs/benches/transaction.rs
+++ b/crates/storage/libmdbx-rs/benches/transaction.rs
@@ -66,8 +66,6 @@ fn bench_put_rand(c: &mut Criterion) {
 
     let txn = env.begin_ro_txn().unwrap();
     let db = txn.open_db(None).unwrap();
-    txn.prime_for_permaopen(db);
-    let db = txn.commit_and_rebind_open_dbs().unwrap().2.remove(0);
 
     let mut items: Vec<(String, String)> = (0..n).map(|n| (get_key(n), get_data(n))).collect();
     items.shuffle(&mut StdRng::from_seed(Default::default()));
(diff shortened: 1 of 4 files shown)


crates/storage/libmdbx-rs/benches/transaction.rs
```diff
@@ -66,8 +66,6 @@ fn bench_put_rand(c: &mut Criterion) {
 
     let txn = env.begin_ro_txn().unwrap();
     let db = txn.open_db(None).unwrap();
-    txn.prime_for_permaopen(db);
-    let db = txn.commit_and_rebind_open_dbs().unwrap().2.remove(0);
```
hai-rise, line 70: It's only used in a benchmark, which is a no-op anyway (`db` is the same after these 2 lines with the same DBI and environment pointer).

crates/storage/libmdbx-rs/src/transaction.rs
```diff
@@ -308,8 +281,6 @@ where
 {
     /// The transaction pointer itself.
     txn: TransactionPtr,
-    /// A set of database handles that are primed for permaopen.
```
hai-rise, line 311: Very questionable description: Tables are "pemaopen" by default, and storing a DBI here doesn't have any effect on that. In fact, it's still possible to close the table handles elsewhere.

Also, instead of arbitrarily calling `prime_for_permaopen` in a lockful way at runtime (without a key to look up anyway), `DatabaseEnv` should create tables and cache all DBIs at startup time, before passing onto transactions to read in a lockless way.

mattsse (approved):

PR #18424, opened 2025-09-13
perf(db): open MDBX DBIs only once at startup

hai-rise:
Part 2 of #18292: Cache DBIs at startup time to reuse throughout the program's lifetime instead of just per transaction.

image

image

---

`reth-bench` for the first 300 blocks of Shanghai:

```
$ reth-bench new-payload-fcu --advance 300
```

f66e19717135db253462069991f41950201720f6 (`main`):
```
2025-09-13T09:46:50.043037Z  INFO Total Ggas/s: 0.6499 total_duration=7.062721087s total_gas_used=4590139398 blocks_processed=300
```

This branch on f66e19717135db253462069991f41950201720f6:
```
2025-09-13T09:56:17.725214Z  INFO Total Ggas/s: 0.6967 total_duration=6.588003416s total_gas_used=4590139398 blocks_processed=300
```

Speedup:
- Ggas/s: 0.6499 -> 0.6967 ~ 7.2% higher execution throughput
- Duration: 7.062721087s -> 6.588003416s ~ 7.2% lower payload latency
- Best win: From 39.2ms down to 13.3ms payload latency

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/mod.rs
+++ b/crates/storage/db/src/implementation/mdbx/mod.rs
@@ -23,6 +23,7 @@ use reth_libmdbx::{
 use reth_storage_errors::db::LogLevel;
 use reth_tracing::tracing::error;
 use std::{
+    collections::HashMap,
     ops::{Deref, Range},
     path::Path,
     sync::Arc,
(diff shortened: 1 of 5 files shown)


crates/storage/db/src/implementation/mdbx/mod.rs
```diff
@@ -190,6 +191,12 @@ impl DatabaseArguments {
 pub struct DatabaseEnv {
     /// Libmdbx-sys environment.
     inner: Environment,
+    /// Opened DBIs for reuse.
+    /// Important: Do not manually close these DBIs, like via `mdbx_dbi_close`.
+    /// More generally, do not dynamically create, re-open, or drop tables at
+    /// runtime. It's better to perform table creation and migration only once
+    /// at startup.
+    dbis: Arc<HashMap<&'static str, ffi::MDBX_dbi>>,
```
hai-rise, line 199: I tried `&'static` and it was ~0.3% faster than `Arc`, but reverted for consistency with the rest of the code base. We can review and make several things (like also the DB environment, etc.) `&'static` together later.

shekhirin (approved):
LGTM, nice! In theory we should even be able to make it a match, because tables are known during compile-time, but let's not bother with this for now.

PR #18426, opened 2025-09-13
chore: clean up TS warnings in search index & file finder

lipperhey:
fixed a couple of small TS issues:

* added `return` after `process.exit(1)` in `fixSearchIndex`.
* dropped unused `stat` import.
* made the recursive `findFiles` call actually use its return value.

### DIFF
--- a/docs/vocs/scripts/fix-search-index.ts
+++ b/docs/vocs/scripts/fix-search-index.ts
@@ -14,6 +14,7 @@ async function fixSearchIndex() {
     if (!searchIndexFile) {
       console.error('❌ No search index file found in .vocs directory');
       process.exit(1);
+      return;
     }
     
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this is working fine

docs/vocs/scripts/fix-search-index.ts
```diff
@@ -65,7 +66,7 @@ async function findFiles(dir: string, extension: string, files: string[] = []):
     if (entry.name === '.vocs' || entry.name === 'docs' || entry.name === '_site') continue;
     
     if (entry.isDirectory()) {
-      await findFiles(fullPath, extension, files);
+      files = await findFiles(fullPath, extension, files);
```
mattsse, line 69: I see this is actually a fix here

mattsse (approved):

PR #18427, opened 2025-09-13
docs: typo fix from "friendless" to "friendly", with love ❤️

mdqst:
spotted a tiny typo that accidentally called developers "friendless."
fixed to "developer friendliness" so we’re clear: **devs are awesome and welcome here**! ❤️


mattsse (commented):
imo friendliness is correct here

PR #18429, opened 2025-09-13
docs: Fix broken debug link

famouswizard:
noticed the debug link was pointing to a non-working address.
updated it to the correct production URL so everything resolves as expected.


mattsse (commented):
this is fine

PR #18433, opened 2025-09-14
chore: remove type aliases

mattsse:
these are only used for the type itself and we can use the helper aliases instead


yongkangc:
nice

PR #18435, opened 2025-09-14
fix(engine): perform cache updates with guard

### LINKED ISSUE #18409: Perform cache updates with guard
### Describe the feature

the order of these operations isn't quite right

https://github.com/paradigmxyz/reth/blob/72c2d1b6a0bc4a1d04b66ef85142d1a2c9700d91/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L135-L149

we should only insert the outcome into the cache after we obtained the mutex:

https://github.com/paradigmxyz/reth/blob/72c2d1b6a0bc4a1d04b66ef85142d1a2c9700d91/crates/engine/tree/src/tree/payload_processor/mod.rs#L553-L553

so that the hash and the tracked cache are updated at the same time
this lock shouldnt take too long, but an alternative here would be to clear the cache first, but 

1. lock
2. update
3. insert

feels better

what we can do here is add a fn to `ExecutionCache` that take a simple closure with the mutexguard
and we perform this in there
https://github.com/paradigmxyz/reth/blob/72c2d1b6a0bc4a1d04b66ef85142d1a2c9700d91/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L140-L140

### Additional context

_No response_

yongkangc:
## Summary

This PR fixes a race condition in cache updates where cache outcomes were being inserted before obtaining the mutex lock, which could lead to synchronization issues.

## Problem

As identified in #18409, the current implementation has a problematic order of operations:
1. Cache outcomes are inserted
2. Metrics are updated  
3. **Then** the mutex is acquired to save the cache

This creates a window where another thread could replace the cache, causing the first thread's changes to be lost.

## Solution  

This PR ensures all cache operations happen atomically by:
- Adding `update_with_guard` method to `ExecutionCache` that takes a closure with exclusive access to the mutex guard
- Refactoring `PrewarmCacheTask::save_cache` to perform all operations (state insertion, metrics update, cache replacement) under the lock

## TODO
- [x] Reth Bench Compare on Mainnet to see regression by 2000
- [x] Run Reproduction Script https://gist.github.com/yongkangc/16f80fb48a3fe8180bd7e990c97124c8
- [x] Benching to see deadlock or any other race condition

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -142,6 +142,7 @@ where
         self.execution_cache.update_with_guard(|cached| {
             let cache = SavedCache::new(hash, caches, metrics);
 
+            // Note: Write lock is held while insert which could block readers briefly
             if cache.cache().insert_state(&state).is_err() {
                 return;
             }
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -145,7 +145,7 @@ where
             // Insert state into cache while holding the lock
             if cache.cache().insert_state(&state).is_err() {
                 // Clear the cache on error to prevent having a polluted cache
-                *cached = None;
+                cached.take();
                 return;
             }
 
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -145,7 +145,7 @@ where
             // Insert state into cache while holding the lock
             if cache.cache().insert_state(&state).is_err() {
                 // Clear the cache on error to prevent having a polluted cache
-                cached.take();
+                *cached = None;
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


mediocregopher (approved):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -132,21 +132,28 @@ where
...
+        let caches = self.ctx.cache.clone();
+        let metrics = self.ctx.cache_metrics.clone();
+
+        // Perform all cache operations atomically under the lock
+        self.execution_cache.update_with_guard(|cached| {
+            let cache = SavedCache::new(hash, caches, metrics);
+
+            // Note: Write lock is held while insert which could block readers briefly
+            if cache.cache().insert_state(&state).is_err() {
+                return;
```
mattsse, line 149: in this case we should fully clear/take the `cached` because this can endup having a polluted cache

mattsse (changes requested):
smol suggestion

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: thanks for the good catch on cache pollution! addressed that by clearing with *cache=None

mattsse (approved):
lgtm!

PR #18436, opened 2025-09-14
fix: incorrect environment var name for jemalloc config

famouswizard:


mattsse (commented):
_RJEM_MALLOC_CONF is correct here

PR #18437, opened 2025-09-14
docs: fix incorrect transaction type count

famouswizard:
the text says *"Four significant transaction types that have evolved are:"* but the list actually includes **five**: Legacy, EIP-2930, EIP-1559, EIP-4844, and EIP-7702.
Updated the wording to say *"Five"* so it matches the list.


mattsse (approved):
fine

PR #18438, opened 2025-09-14
Fix: type constraints and async handling in MyExEx

0xwitty:
**Changes and why:**

* Added `where` clause for `impl<Node> Future for MyExEx<Node>` since Rust doesn't allow `Node: FullNodeComponents<Types: NodeTypes<...>>`.
* Replaced `notification.committed_chain()` with explicit `match ExExNotification::ChainCommitted { new }` to give the compiler clear access to `new`.
* Converted `main` to `#[tokio::main] async` so that `run` can `await` properly, avoiding errors with `async move` blocks.

this changes ensure the code now compiles and async execution works as intended.


mattsse (commented):
we need blocking here

PR #18439, opened 2025-09-14
perf: Enforce EIP-7825 transaction gas limit for Osaka

fredrik0x:
Adds pre-execution (already exists in revm) validation similar to the other EL clients to save resources by not trying to execute transactions that exceed the maximum gas limit defined by EIP-7825. Introduces a new ConsensusError variant for transactions exceeding this limit.


mattsse (approved):
makes sense, ty!

Issue #18255, opened 2025-09-03
Add flashblock engine sidecar service


mattsse:
### Describe the feature

if flashblocks are installed we receive info about the latest block via flashblock.parent_block

we can use this to advance the chain and eventually drive it entirely.
this would work similar to the existing `DebugConsensusClient` 

https://github.com/paradigmxyz/reth/blob/1d7fefecec21c06bfaf4936b9ce68dfe0da7cfab/crates/consensus/debug-client/src/client.rs#L60-L68

that fetches block info from some endpoint and submits it via the handle.

we can a similar type for flashblocks that would receive a stream of flashblocks: #18244

and advance the chain by sending new forkchoice update messages:

https://github.com/paradigmxyz/reth/blob/1d7fefecec21c06bfaf4936b9ce68dfe0da7cfab/crates/consensus/debug-client/src/client.rs#L139-L142

I'm not exactly sure if a full flashblocksequence can be converted to the entire block (e.g. stateroot perhaps not available) but we can at least send fcu and the node should then handle sync over p2p atleast.
once caught up with the tip we could eventually transition to sending fully executed blocks if `flashblock.parent == node.latest`

## TODO
* introduce a flashblock specific `DebugConsensusClient` that operates on a stream of `Vec<Flashblock>` and submits parenthash as FCU
* next we could integrate the watch channel as well, and detect if the block is complete
* for this we'd need a new setting in the existing `FlashblockService` that computes the stateroot if the sequence is complete (idk if this can be identified just by looking at the flashblock data but maybe we can just do if index is 10 or 11

@julio4

julio4:
On it

PR #18443, opened 2025-09-15
feat(op-reth): initial setup FlashBlockConsensusClient engine sidecar

julio4:
Towards #18255

Initial setup of a `FlashBlockConsensusClient`, opening for reviews first to be sure it architecturally makes sense.
`sequence_to_block` will be fully implemented in subsequent PR.


crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -0,0 +1,68 @@
...
+                        SealedBlock::new_unhashed(block),
+                    );
+
+                    // Send new events to execution client
+                    let _ = self.engine_handle.new_payload(payload).await;
+
+                    let state = alloy_rpc_types_engine::ForkchoiceState {
+                        head_block_hash: block_hash,
+                        safe_block_hash: block_hash,      // TODO
+                        finalized_block_hash: block_hash, // TODO
```
julio4: Here how should we handle this? 
Should we keep copy of incoming flashblock sequences in a ring buffer for safe/finalized head at offset 32/64?

mattsse: > afe/finalized head at offset 32/64?

yes exactly, like we do in debugconsensusclient I believe

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -0,0 +1,68 @@
...
+                Ok(sequence) => {
+                    let block_hash = sequence.last().diff.block_hash;
+                    let block = self.sequence_to_block(sequence);
+
+                    let payload = OpPayloadTypes::<OpPrimitives>::block_to_payload(
+                        SealedBlock::new_unhashed(block),
+                    );
+
+                    // Send new events to execution client
+                    let _ = self.engine_handle.new_payload(payload).await;
```
mattsse: this we can omit for now because we can only do this if the block has the correct state root whihc is mostlikely not the case, 

in order to do this we need a setting in the flashblocks builder that computes the real state root, e.g. if it's index 10/11

mattsse (changes requested):
great start

crates/optimism/flashblocks/src/consensus.rs
julio4: I see yes that's true. We can just only keep sending FCU for now

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Spawn the client to start sending FCUs and new payloads by periodically fetching recent
+    /// blocks.
+    pub async fn run(mut self) {
+        let mut previous_block_hashes = AllocRingBuffer::new(64);
+
+        loop {
+            match self.sequence_receiver.recv().await {
+                Ok(sequence) => {
+                    let block_hash = sequence.last().diff.block_hash;
+                    previous_block_hashes.push(block_hash);
```
mattsse: we cant use this

these hashes are all incorrect because state root isnt calculated

the need to use base.parent_hash

mattsse (changes requested):
pretty cool, one more request

crates/optimism/flashblocks/src/consensus.rs
julio4: Hm yeah I was thinking about that...
But using parent hash won't make the node always "lag" one block away from current pending block?

mattsse: yes for now
we can fix this in a followup or multiple ones

mattsse (approved):
cool, this should already be sufficient

PR #18444, opened 2025-09-15
fix(rpc): add validation for missing headers in debug execution witness

wetkeyboard:
Add comprehensive checks to ensure all required headers are present and contiguous in the `debug_execution_witness` RPC


mattsse (changes requested):
I believe this todo is outdated because if we dont have the headers then this would fail anyway

so we can remove the todo instead

mattsse (approved):

PR #18446, opened 2025-09-15
fix(txpool): prevent double-processing of tx pool tier

### LINKED ISSUE #18401: merge basefee promotion with account updates
### Describe the feature

we currently trigger poolupdates in two places:

https://github.com/paradigmxyz/reth/blob/6d4a1a3ccfb296af9d920ab5e5f1c90eaa61d2d7/crates/transaction-pool/src/pool/txpool.rs#L578-L593

**Place 1: Basefee/Blobfee-triggered updates** (lines 580-580)
```579:580:/home/chiayongtcac/reth/crates/transaction-pool/src/pool/txpool.rs
        // update block info
        let block_hash = block_info.last_seen_block_hash;
        self.set_block_info(block_info);
```

This calls `set_block_info()` which internally calls:
- `update_basefee()` → can trigger promotions from BaseFee pool to Pending when basefee decreases
- `update_blob_fee()` → can trigger promotions/demotions for blob transactions

**Place 2: Account/sender-triggered updates** (lines 593-593)
```592:594:/home/chiayongtcac/reth/crates/transaction-pool/src/pool/txpool.rs
        // Update removed transactions metric
        self.metrics.removed_transactions.increment(removed_txs_count);

        let UpdateOutcome { promoted, discarded } = self.update_accounts(changed_senders);
```

This calls `update_accounts()` which:
- Processes sender balance/nonce changes
- Triggers sub-pool updates based on account state changes

**The Problem**
Both places can update the same transactions independently:

1. **Basefee decrease** → promotes transactions from BaseFee → Pending via `enforce_basefee_with()`
2. **Sender changes** → same transactions might get updated again via `update_accounts()` 

This creates overlapping/duplicate tracking in the `promoted` and `discarded` outcome sets, leading to inaccurate metrics and potential inconsistent pool state.

where a lower basefee can trigger a promotion:

https://github.com/paradigmxyz/reth/blob/6d4a1a3ccfb296af9d920ab5e5f1c90eaa61d2d7/crates/transaction-pool/src/pool/txpool.rs#L295-L310

which we currently dont track properly and only via the changed accounts:

https://github.com/paradigmxyz/reth/blob/6d4a1a3ccfb296af9d920ab5e5f1c90eaa61d2d7/crates/transaction-pool/src/pool/txpool.rs#L593-L593

the problem here is that these can overlap

e.g. promoted basefee txs can also appear in either promoted or discarded

a potential fix here is
before updating the basefee internally, use the new basefee to iterate over the changed senders
https://github.com/paradigmxyz/reth/blob/6d4a1a3ccfb296af9d920ab5e5f1c90eaa61d2d7/crates/transaction-pool/src/pool/txpool.rs#L554-L555
then we get and perform the updates for all senders correctly with the new fee
and after we append the updates from updating the fee pool

### Additional context

_No response_

yongkangc:
## Summary

This PR fixes the double-processing issue in the transaction pool when both base fee changes and account updates affect the same transactions. Fixes #18401

## Problem

When `on_canonical_state_change` is called with both fee changes and account updates, transactions could be processed twice:
1. Once by fee updates (e.g., basefee decrease promoting transactions)
2. Again by account updates

This happened because account updates were evaluating transactions with the OLD fees, potentially making different decisions than they would with the new fees.

## Fix
1. update all changed accounts with the next basefee first
2. then perform basefee pool updates, because then we’ve already processed the changed accounts correctly and they shouldnt be impacted by this

to understand what the rationale is here:
- update all changed accounts with the next base fee
- allows us to update changed accounts and put them through the update_basefee base fee logic
- prevents duplicate promotion of an account going through the update_basefee promotion logic + the update_account logic


yongkangc:
PR is ready for review!

## Summary of Changes

The fix is now very simple - just update the internal fees before calling `update_accounts`:

1. Remove mined transactions
2. **Update internal fees to new values** (the key change)
3. Call `update_accounts` which now uses the new fees
4. Update remaining block info

This ensures account updates evaluate transactions with the correct (new) fees, preventing double-processing.

All tests pass:
- ✅ 200 transaction pool tests passing
- ✅ CI checks passing (clippy, fmt, etc.)
- ✅ Comprehensive test coverage added

The solution is minimal and focused on fixing the root cause: ensuring consistent fee context when evaluating transactions.

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -590,8 +589,18 @@ impl<T: TransactionOrdering> TxPool<T> {
         // Update removed transactions metric
         self.metrics.removed_transactions.increment(removed_txs_count);
 
+        // First update the fees internally so account updates use the new basefee
+        self.all_transactions.pending_fees.base_fee = block_info.pending_basefee;
+        if let Some(blob_fee) = block_info.pending_blob_fee {
+            self.all_transactions.pending_fees.blob_fee = blob_fee;
+        }
```
mattsse: switching this makes sense, I think we can extract the logic from

set_block_info

so that they only update the fees and return the previous ones?

and then we do the subpool updates separately?

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -590,8 +589,18 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+
+        // Now update accounts with the new fees already set
         let UpdateOutcome { promoted, discarded } = self.update_accounts(changed_senders);
 
+        // Update the rest of block info
+        self.all_transactions.set_block_info(block_info);
```
mattsse, line 680: this makes sense now and we want to extend the promoted with the subpool updates

mattsse (changes requested):
cool, this is def the right direction

crates/transaction-pool/src/pool/txpool.rs
yongkangc: this makes sense, would be much better and we can extend that below

crates/transaction-pool/src/pool/txpool.rs
yongkangc: thanks for the review @mattsse , wanted to clarify on this 

by sub pool updates, do you mean 👇🏻 

> switching this makes sense, I think we can extract the logic from
> 
> set_block_info
> 
> so that they only update the fees and return the previous ones?
> 
> and then we do the subpool updates separately?

https://github.com/paradigmxyz/reth/pull/18446#discussion_r2351621758

or is there something else i am not getting?

yongkangc: or do you mean the transaction movements between pools caused by fee changes? 

or something like this `basefeeupdates.promotions.extend(accountupdates.promotios)`

yongkangc: ans: yup thats subpool update
- after we updated all changed senders we can add the other movements to those sets
- because all changed accounts are already processed correctly
- we recently changed the update fn to accept a closure, we can do something similar here where instead of collecting an additional set we can push to the promoted set directly

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -546,6 +546,87 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+                    tx_meta.subpool = tx_meta.state.into();
+                    tx_meta.subpool
+                };
+
+                // Direct push to outcome for promoted transactions
+                if to == SubPool::Pending {
+                    outcome.promoted.push(tx.clone());
+                }
+
+                self.add_transaction_to_subpool(to, tx);
```
yongkangc: @mattsse adding this because my understanding is that we still need to update their subpool after updating the pending fees

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -546,6 +546,87 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+                    SubPool::Queued => self.queued_pool.add_transaction(tx),
+                    SubPool::Pending => {
+                        self.pending_pool.add_transaction(tx, current_base_fee);
+                    }
+                    SubPool::Blob => self.blob_pool.add_transaction(tx),
+                    SubPool::BaseFee => {
+                        warn!(target: "txpool", "BaseFee transactions should become Pending after basefee decrease");
+                    }
+                }
+            });
```
yongkangc: @mattsse i added this in apply_fee_update, because we are making fee changes that can cause transactions to move between subpools.

Wanted to point out here in case i might have the wrong understanding of adding txns to different pools

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -546,6 +546,88 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+    /// Updates only the pending fees without triggering subpool updates.
+    /// Returns the previous base fee and blob fee values.
+    #[allow(clippy::missing_const_for_fn)]
+    fn update_pending_fees_only(
+        &mut self,
+        new_base_fee: u64,
+        new_blob_fee: Option<u128>,
+    ) -> (u64, u128) {
+        let prev_base_fee = self.all_transactions.pending_fees.base_fee;
+        let prev_blob_fee = self.all_transactions.pending_fees.blob_fee;
```
mattsse: I think we can mem::swap here

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -546,6 +546,88 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+    }
+
+    /// Applies fee-based promotion updates.
+    /// Directly modifies the provided outcome by pushing promoted transactions.
+    fn apply_fee_updates(
+        &mut self,
+        prev_base_fee: u64,
+        prev_blob_fee: u128,
+        outcome: &mut UpdateOutcome<T::Transaction>,
+    ) {
```
mattsse: this looks correct, but we still have

these functions

https://github.com/paradigmxyz/reth/blob/c45817c1f21b13368f75ffb1058dd1457d380f9f/crates/transaction-pool/src/pool/txpool.rs#L271-L271

https://github.com/paradigmxyz/reth/blob/c45817c1f21b13368f75ffb1058dd1457d380f9f/crates/transaction-pool/src/pool/txpool.rs#L224-L224

could we reuse them and give them `&mut UpdateOutcome`

mattsse (changes requested):
cool, one more suggestion

crates/transaction-pool/src/pool/txpool.rs
yongkangc: oh thats a good one, thanks!

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -546,6 +573,52 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+
+        let base_fee_decreased = current_base_fee < prev_base_fee;
+        let blob_fee_decreased = current_blob_fee < prev_blob_fee;
+
+        // Only handle promotions when base fee decreases
+        if base_fee_decreased {
+            self.handle_basefee_decrease(current_base_fee, |tx| outcome.promoted.push(tx.clone()));
+            // Blob promotions depend on both base- and blob-fee affordability, so recheck blob
+            // transactions as well when the base fee drops.
+            self.handle_blob_fee_decrease(|tx| outcome.promoted.push(tx.clone()));
```
yongkangc: looking at 
```
  (Ordering::Less, _) | (_, Ordering::Less) => {
      // decreased blob/base fee: recheck blob
  pool and promote all that are now valid
```

I am assuming that blob transactions can be promoted when either the blob fee OR the base fee decreases.

This makes sense because blob transactions need to satisfy both base fee and blob fee requirements.

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -301,38 +289,77 @@ impl<T: TransactionOrdering> TxPool<T> {
...
 
                 Ordering::Less
             }
         }
     }
 
+    /// Handles promotions that become possible after a base-fee decrease.
+    ///
+    /// The provided callback is invoked for each transaction that becomes pending.
+    fn handle_basefee_decrease<F>(&mut self, current_base_fee: u64, mut on_promoted: F)
```
mattsse: why do we need these if we already have

https://github.com/paradigmxyz/reth/blob/4090151fbc12118561507ad7302762fef626b787/crates/transaction-pool/src/pool/txpool.rs#L224-L224

and 

https://github.com/paradigmxyz/reth/blob/4090151fbc12118561507ad7302762fef626b787/crates/transaction-pool/src/pool/txpool.rs#L259-L259

this is the same logic, can we not unify the existing ones?

mattsse (changes requested):

crates/transaction-pool/src/pool/txpool.rs
yongkangc: was initially thinking that theres the additional other ordering logic that we might not want coupled together. 

but yup unifying makes more sense with `Ordering::Less` for basefee promotion

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -250,15 +257,20 @@ impl<T: TransactionOrdering> TxPool<T> {
...
+                        tx_meta.state.insert(TxState::ENOUGH_FEE_CAP_BLOCK);
+                        tx_meta.subpool = tx_meta.state.into();
+                        tx_meta.subpool
                     };
-                    self.add_transaction_to_subpool(to, tx);
+
+                    if subpool == SubPool::Pending {
+                        on_promoted(&tx);
+                    }
+
```
mattsse, line 272: great

mattsse (approved):
very, nice

PR #18457, opened 2025-09-15
docs(engine): improve cache naming and documentation

crazykissshout:
Improves cache-related code clarity and documentation as requested in #18451.

part of the comments are from Claude (LMAO much better than my own version)

## Changes

- Rename `ProviderCaches` → `ExecutionCache` for clearer purpose
- Rename `ProviderCacheBuilder` → `ExecutionCacheBuilder`
- Resolve naming conflicts with payload processor types

- Add module-level docs explaining cache architecture and behavior
- Clarify `SlotStatus::Empty` vs `SlotStatus::NotCached` distinction
- Document `storage_multiproof` concept and relationship to caching
- Explain cache lifecycle and relationship with prewarming
- Add background persistence documentation

Closes #18451

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -300,7 +300,7 @@ pub(crate) struct ExecutionCache {
     /// Cache for contract bytecode, keyed by code hash.
     code_cache: Cache<B256, Option<Bytecode>>,
 
-    /// Hierarchical storage cache organized by account address.
+    /// Per-account storage cache: outer cache keyed by Address, inner cache tracks that account’s storage slots.
     storage_cache: Cache<Address, AccountStorageCache>,
 
     /// Cache for basic account information (nonce, balance, code hash).
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -300,7 +300,8 @@ pub(crate) struct ExecutionCache {
     /// Cache for contract bytecode, keyed by code hash.
     code_cache: Cache<B256, Option<Bytecode>>,
 
-    /// Per-account storage cache: outer cache keyed by Address, inner cache tracks that account’s storage slots.
+    /// Per-account storage cache: outer cache keyed by Address, inner cache tracks that account’s
+    /// storage slots.
     storage_cache: Cache<Address, AccountStorageCache>,
 
     /// Cache for basic account information (nonce, balance, code hash).
@@ -363,7 +364,7 @@ impl ExecutionCache {
     ///
     /// ## Error Handling
     ///
-    /// Returns an error if the state updates are inconsistent and should be discarded. 
+    /// Returns an error if the state updates are inconsistent and should be discarded.
     pub(crate) fn insert_state(&self, state_updates: &BundleState) -> Result<(), ()> {
         // Insert bytecodes
         for (code_hash, bytecode) in &state_updates.contracts {
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -260,11 +260,6 @@ impl<S: StorageRootProvider> StorageRootProvider for CachedStateProvider<S> {
     /// 2. Traverses the account's storage trie to collect proof nodes
     /// 3. Returns a [`StorageMultiProof`] containing the minimal set of trie nodes needed to verify
     ///    all the requested storage slots
-    ///
-    /// ## Relationship to Caching:
-    /// Storage multiproofs are not directly cached, but they rely on the underlying
-    /// storage data that IS cached. When generating proofs, cached storage values
-    /// can be used instead of fetching from disk, improving proof generation speed.
     fn storage_multiproof(
         &self,
         address: Address,
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -1,4 +1,23 @@
-//! Implements a state provider that has a shared cache in front of it.
+//! Execution cache implementation for block processing.
+//!
+//! This module provides a caching layer on top of state providers to optimize
+//! state access during block execution. The cache system works in conjunction
+//! with prewarming to improve performance.
+//!
+//! ## Architecture
+//!
+//! The caching system consists of:
+//! - [`ExecutionCache`]: Main cache structure holding account, storage, and bytecode caches
+//! - [`CachedStateProvider`]: Wrapper that adds caching to any [`StateProvider`]
+//! - [`AccountStorageCache`]: Hierarchical storage cache organized by account
+//! - [`SlotStatus`]: Enum representing cache hit/miss status for storage slots
+//!
+//! ## Cache Behavior
+//!
+//! On read: checks cache first, returns cached value on hit, fetches from
+//! underlying provider on miss and caches the result.
+//!
+//! On write: after execution, touched state is inserted via [`ExecutionCache::insert_state`].
 use alloy_primitives::{Address, StorageKey, StorageValue, B256};
 use metrics::Gauge;
 use mini_moka::sync::CacheBuilder;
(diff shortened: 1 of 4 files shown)


yongkangc (approved):
i like this renaming and improvement of docs. its a good start, however made some changes as well.

what do you think @mattsse ? would there be any side effects on other dependent repo

imo these seem to be rather internal, but not sure if this is a breaking api change?

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -1,4 +1,23 @@
...
+//! - [`CachedStateProvider`]: Wrapper that adds caching to any [`StateProvider`]
+//! - [`AccountStorageCache`]: Hierarchical storage cache organized by account
+//! - [`SlotStatus`]: Enum representing cache hit/miss status for storage slots
+//!
+//! ## Cache Behavior
+//!
+//! On read: checks cache first, returns cached value on hit, fetches from
+//! underlying provider on miss and caches the result.
+//!
+//! On write: after execution, touched state is inserted via [`ExecutionCache::insert_state`].
```
yongkangc: not exactly what i had in mind but a good attempt. Tho i think we should remove these because its rather lengthy, but the core assumptions / distinctions are not fleshed out. 


```suggestion
```

yongkangc (dismissed):

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -128,14 +128,18 @@ impl<S: AccountReader> AccountReader for CachedStateProvider<S> {
     }
 }
 
-/// Represents the status of a storage slot in the cache
+/// Represents the status of a storage slot in the cache.
+///
+/// The distinction between `NotCached` and `Empty` is crucial for performance:
+/// - `NotCached` means we need to query the database
+/// - `Empty` means we already know the slot is zero (no database query needed)
```
yongkangc:
```suggestion
```

hmm dont think we need this. Not sure if we should rename this as well.

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -330,18 +351,26 @@ impl ProviderCaches {
...
 
-    /// Inserts the [`BundleState`] entries into the cache.
+    /// Inserts the post-execution state changes into the cache.
+    ///
+    /// This method is called after transaction execution to update the cache with
+    /// the touched and modified state. The insertion order is critical:
+    ///
+    /// 1. **Bytecodes**: Insert contract code first
+    /// 2. **Storage slots**: Update storage values for each account
+    /// 3. **Accounts**: Update account info (nonce, balance, code hash)
```
yongkangc:
```suggestion
    /// 1. Bytecodes: Insert contract code first
    /// 2. Storage slots: Update storage values for each account
    /// 3. Accounts: Update account info (nonce, balance, code hash)
```

Lets keep the formatting simpler to read inline

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -330,18 +351,26 @@ impl ProviderCaches {
...
     ///
-    /// The order is important, because the access patterns are Account -> Bytecode and Account ->
-    /// Storage slot. If we update the account first, it may point to a code hash that doesn't have
-    /// the associated bytecode anywhere yet.
+    /// ## Error Handling
     ///
-    /// Returns an error if the state can't be cached and should be discarded.
+    /// Returns `Err(())` if the state updates are inconsistent (e.g., modified account
+    /// with `None` info that should have been destroyed). In this case, the entire
+    /// cache update should be discarded to maintain consistency.
```
yongkangc, line 373:
```suggestion
   ///
    /// Returns an error if the state updates are inconsistent and should be discarded.
```


trying to simplify the comments here

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -248,6 +248,23 @@ impl<S: StorageRootProvider> StorageRootProvider for CachedStateProvider<S> {
...
+    /// ## How it works:
+    /// 1. Takes an account address and a list of storage slot keys
+    /// 2. Traverses the account's storage trie to collect proof nodes
+    /// 3. Returns a [`StorageMultiProof`] containing the minimal set of trie nodes needed to verify
+    ///    all the requested storage slots
+    ///
+    /// ## Relationship to Caching:
+    /// Storage multiproofs are not directly cached, but they rely on the underlying
+    /// storage data that IS cached. When generating proofs, cached storage values
+    /// can be used instead of fetching from disk, improving proof generation speed.
```
yongkangc:
```suggestion
```

i think we can remove these because it makes understanding the function much more difficult referencing another part of the code.

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -330,18 +351,26 @@ impl ProviderCaches {
...
     ///
-    /// The order is important, because the access patterns are Account -> Bytecode and Account ->
-    /// Storage slot. If we update the account first, it may point to a code hash that doesn't have
-    /// the associated bytecode anywhere yet.
+    /// ## Error Handling
     ///
-    /// Returns an error if the state can't be cached and should be discarded.
+    /// Returns `Err(())` if the state updates are inconsistent (e.g., modified account
+    /// with `None` info that should have been destroyed). In this case, the entire
+    /// cache update should be discarded to maintain consistency.
```
yongkangc: i prefer the previous version as it was simpler
```suggestion
    /// Returns an error if the state updates are inconsistent and should be discarded. 
```

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -278,20 +290,24 @@ impl<S: HashedPostStateProvider> HashedPostStateProvider for CachedStateProvider
...
+/// to reduce database I/O during block execution.
 #[derive(Debug, Clone)]
-pub(crate) struct ProviderCaches {
-    /// The cache for bytecode
+pub(crate) struct ExecutionCache {
+    /// Cache for contract bytecode, keyed by code hash.
     code_cache: Cache<B256, Option<Bytecode>>,
 
-    /// The cache for storage, organized hierarchically by account
+    /// Hierarchical storage cache organized by account address.
```
yongkangc:
```suggestion
    /// Per-account storage cache: outer cache keyed by Address, inner cache tracks that account’s storage slots.
```

i dont really understand what hierachical means here, we should keep it simple?

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -128,14 +128,14 @@ impl<S: AccountReader> AccountReader for CachedStateProvider<S> {
...
 
-/// Represents the status of a storage slot in the cache
+/// Represents the status of a storage slot in the cache.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum SlotStatus {
-    /// The account's storage cache doesn't exist
+    /// The account's storage cache doesn't exist.
     NotCached,
-    /// The storage slot is empty (either not in cache or explicitly None)
+    /// The storage slot exists in cache and is empty (value is zero).
```
mediocregopher, line 136: The documented meaning of Empty is changed here, the docs now say that the slot explicitly exists in cache as None, before they said it is not in cache or None, is the new meaning correct?

crates/engine/tree/src/tree/cached_state.rs
crazykissshout: Sure, lets keep it simple.

crates/engine/tree/src/tree/cached_state.rs
crazykissshout: @mediocregopher thanks to point this out! But I think new description is good here. 

`Empty` means the slot **exists in cache** with a `None` value (representing zero). This is different from `NotCached` which means the slot was never queried.

The implementation confirms this: `Some(None) => SlotStatus::Empty` - the slot is cached (`Some`) with value `None`.

mediocregopher: I believe you're right, looks like the old docs were wrong, thanks for double-checking :pray:

mediocregopher (dismissed):
LGTM once CI is happy

crazykissshout:
> LGTM once CI is happy

Awesome, the CI is happy now

mattsse (approved):
lgtm

i love ai now

PR #18459, opened 2025-09-15
fix(engine): remove redundant method-level where bound in InvalidBlockWitnessHook

MozirDmitriy:
The method on_invalid_block redundantly declared where N: NodePrimitives while the enclosing impl already constrained N. This duplication provided no additional guarantees, reduced readability, and was inconsistent with the project’s style where impl-level bounds are not repeated on methods. Removing the redundant bound keeps a single source of truth and maintains identical semantics.


mattsse (approved):
fine, makes sense

PR #18476, opened 2025-09-15
feat: `Block::iter_recovered`

klkvr:
Adds a helper returning iterator of recovery results for block transactions

### DIFF
--- a/crates/primitives-traits/src/block/mod.rs
+++ b/crates/primitives-traits/src/block/mod.rs
@@ -23,7 +23,6 @@
 //! ```
 
 pub(crate) mod sealed;
-use alloy_consensus::transaction::Recovered;
 pub use sealed::SealedBlock;
 
 pub(crate) mod recovered;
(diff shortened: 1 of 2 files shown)


crates/primitives-traits/src/block/mod.rs
```diff
@@ -220,6 +221,16 @@ pub trait Block:
...
     }
+
+    /// Returns an iterator over the recovered transactions in the block.
+    fn iter_recovered(
+        &self,
+    ) -> impl Iterator<Item = Result<Recovered<<Self::Body as BlockBody>::Transaction>, RecoveryError>>
+    {
+        self.body()
+            .transactions_iter()
+            .map(|tx| tx.try_recover().map(|signer| tx.clone().with_signer(signer)))
```
mattsse: can we do
```suggestion
    ) -> impl Iterator<Item = Result<Recovered<&<Self::Body as BlockBody>::Transaction>, RecoveryError>>
    {
        self.body()
            .transactions_iter()
            .map(|tx| tx.try_recover().map(|signer| tx.clone().with_signer(signer)))
```

then the caller can do .cloned()

hmm, but this would then needlessly clone the address, so this is more efficient probably the main use case here, but we should mention that this cleans the txs

but we could also add `iter_recovered_ref`

mattsse (changes requested):
cool, one suggestion

crates/primitives-traits/src/block/mod.rs
gakonst: nit: can recovery be parallel?

mattsse: we can't make this work with -> impl iterator because ParallelIterator isnt Iterator

so for this use case @klkvr maybe we just return Vec instead since this is the primary use case here?

klkvr: sgtm! added the helper to `BlockBody` instead 063a007, `recover_signers` is also there so seemed more appropriate

mattsse (approved):

PR #18477, opened 2025-09-15
feat(rpc): debug_traceCall support TxIndex

jsvisa:
depends on https://github.com/alloy-rs/alloy/pull/2881


crates/rpc/rpc/src/debug.rs
```diff
@@ -477,6 +493,74 @@ where
...
+    ) -> Result<GethTrace, Eth::Error> {
+        // Get the target block to check transaction count
+        let block = self
+            .eth_api()
+            .recovered_block(block_id)
+            .await?
+            .ok_or(EthApiError::HeaderNotFound(block_id))?;
+
+        let txs = block.body().transactions();
+        if tx_index >= txs.len() {
```
mattsse: there should be a transaction_count fn

mattsse (changes requested):
cool

smol nit and a request to make support for this less of an issuer for library users

mattsse (approved):

PR #18479, opened 2025-09-15
chore(trie): Use Vec<Option<...>> in InMemoryTrieCursor

mediocregopher:
This refactors the in-memory trie cursor overlay to use a vec of options rather than a vec and a separate hashmap to track removed nodes. This also de-duplicates the account/storage versions of this overlay into a single implementation, and adds testing for it.

The intention is to make it easier to use trie updates queried from the DB as an in-memory overlay.

Fixes #18462

### DIFF
--- a/crates/trie/trie/src/trie_cursor/in_memory.rs
+++ b/crates/trie/trie/src/trie_cursor/in_memory.rs
@@ -1,9 +1,6 @@
 use super::{TrieCursor, TrieCursorFactory};
-use crate::{
-    forward_cursor::ForwardInMemoryCursor,
-    updates::{StorageTrieUpdatesSorted, TrieUpdatesSorted},
-};
-use alloy_primitives::{map::HashSet, B256};
+use crate::{forward_cursor::ForwardInMemoryCursor, updates::TrieUpdatesSorted};
+use alloy_primitives::B256;
 use reth_storage_errors::db::DatabaseError;
 use reth_trie_common::{BranchNodeCompact, Nibbles};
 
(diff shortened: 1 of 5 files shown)


crates/trie/trie/src/trie_cursor/in_memory.rs
```diff
@@ -24,97 +21,121 @@ impl<'a, CF> InMemoryTrieCursorFactory<'a, CF> {
...
         &self,
         hashed_address: B256,
     ) -> Result<Self::StorageTrieCursor, DatabaseError> {
-        let cursor = self.cursor_factory.storage_trie_cursor(hashed_address)?;
-        Ok(InMemoryStorageTrieCursor::new(
-            hashed_address,
-            cursor,
-            self.trie_updates.storage_tries.get(&hashed_address),
-        ))
+        // if the storage trie has no updates then we use this as the in-memory overlay.
```
yongkangc, line 36: curious why do we use this as the memory overlay when there are no updates?

mediocregopher: InMemoryTrieCursor must have some array reference to pass into ForwardInMemoryCursor. Since InMemoryTrieCursor is returned from here owned the reference cannot be to a value in this scope, so I made an empty static array to pass it.

The alternative would have been to add a generic over ForwardInMemoryCursor, which would have added a bunch of complexity to the types. This is simpler.

mattsse (commented):
all of this lgtm afaik but I'd rather defer to @shekhirin and @Rjected here

shekhirin (approved):
LGTM, also nice that we now have just one `InMemoryTrieCursor`

mattsse (approved):
lgtm

PR #18481, opened 2025-09-15
chore: update superchain commit

mattsse:
for some reason the CI job for this is MIA

this now adds celo mainnet support and something called booba


fgimenez (approved):
lgtm, looks like a test needs to be updated

PR #18487, opened 2025-09-16
refactor: extract stream logic and specialize `send_raw_transaction_sync` for op-reth

Peponks9:
Extracts reusable stream awaiting logic from `send_raw_transaction_sync` and adds Optimism-specific specialization to support future flashblock monitoring.

## Changes
- **Extract helper**: `await_transaction_receipt_from_stream()` in `reth-rpc-eth-api` for reusable stream logic
- **Specialize for op-reth**: Override `send_raw_transaction_sync` in `OpEthApi` using the helper

## Testing
- `make pr` passes
-  Existing tests pass for both `reth-rpc-eth-api` and `reth-optimism-rpc`

closes #18372

### DIFF
--- a/crates/storage/storage-api/src/chain.rs
+++ b/crates/storage/storage-api/src/chain.rs
@@ -138,7 +138,7 @@ where
         _remove_from: StorageLocation,
     ) -> ProviderResult<()> {
         provider.tx_ref().unwind_table_by_num::<tables::BlockWithdrawals>(block)?;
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
actually I changed my mind on this, we cant easily abstract this,
so let's restore this to what is was instead and then duplicate the logic in optimism's `send_raw_transaction_sync` implementation

then we can easily integrate the flashblocks fetching as well, with a tokio::select for example

PR #18490, opened 2025-09-16
fix(engine): exit MultiProofTask loop on closed internal channel

MozirDmitriy:
The run() loop logged an error on recv() failure but continued looping, causing an infinite error loop and log spam when the internal mpsc channel closed. This change returns immediately on Err(_), aligning with shutdown patterns used elsewhere in the codebase (e.g., engine tree and proof task loops) and ensuring the task terminates cleanly when the channel is disconnected.

Also cargo fmt made some changes with: ':'

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -164,7 +164,7 @@ impl ProofSequencer {
 
         // return early if we don't have the next expected proof
         if !self.pending_proofs.contains_key(&self.next_to_deliver) {
-            return Vec::new();
+            return Vec::new()
         }
 
         let mut consecutive_proofs = Vec::with_capacity(self.pending_proofs.len());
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
can you remove the additional `;`

MozirDmitriy:
> can you remove the additional `;`

Yes, done

mediocregopher (approved):
This LGTM, @MozirDmitriy did you only encounter a node having its MultiProofTask channel closed once randomly? Or is it something you're able to reproduce?

MozirDmitriy:
> This LGTM, @MozirDmitriy did you only encounter a node having its MultiProofTask channel closed once randomly? Or is it something you're able to reproduce?

not really able to reproduce. I noticed it during code review. The change is defensive: if the internal mpsc channel ever gets disconnected (e.g., during task teardown/shutdown or after a panic), the task would otherwise spin and spam logs. Exiting immediately aligns with how similar loops handle this in the codebase (e.g., engine tree and proof task loops).

mattsse (approved):
ty, we can add this but this is effectively unreachable because this type also holds a clone of the sender

PR #18491, opened 2025-09-16
refactor: remove anyhow dependency from nippy-jar

LandLord529:
Removes unnecessary anyhow dependency from the nippy-jar crate.

The EliasFano error variant was the only usage of anyhow::Error in the codebase, and it was never actually used. Replaced it with a simple String type to maintain the same functionality while reducing dependencies.


mattsse (commented):
this doesnt matter because this is still part of the dep tree

PR #17416, opened 2025-07-15
Demo of Integrating the MPT from RSP with Reth's ExecutionWitness

koxu1996:
Let me provide some context to avoid confusion.

### Ethereum Block Proving

For proving Ethereum blocks, most zkVM teams are using [SuccintLabs's RSP](https://github.com/succinctlabs/rsp). There is also [Risc0 Zeth](https://github.com/risc0/zeth), but it is not widely adopted.

Both RSP and Zeth are built upon **Reth** crates, with additional optimizations to make them zkVM-friendly.

However, they both suffer from a long input preparation times - **several minutes** - due to thousands calls to RPC:
- `eth_getProof`
- `eth_getStorageAt`

Manually building the state and storage tries is just too slow.

### Execution Witness

The Reth team introduced a dedicated endpoint: `debug_getExecutionWitness`. It provides all relevant state and storage data in a **single RPC request**. Fetching the witness is fast for the recent blocks:
- Latest block: 0.69s
- 10 blocks ago: 0.87s
- 20 blocks ago: 1.66s
- 100 blocks ago: 2.63s
- 500 blocks ago: 9.59s
- 1000 blocks ago: 20.00s (default window limit is 1024 blocks)
- 2000 blocks ago: 46.87s

That is not all! Reth also provides [`reth-stateless` crate](https://reth.rs/docs/reth_stateless/index.html), which is the **state-of-the-art** for stateless execution and block validation. Naturally, it uses `ExecutionWitness`, allowing input preparation time to take only seconds.

One might now wonder: why not just replace RSP with `reth-stateless`?

Well, there is a trade-off. For block 22830000 and SP1 zkVM, we got:

- RSP:
  - Input preparation time: few minutes
  - Program cycles: ~225M
  - Proving time: ~150 seconds.
- reth-stateless:
  - Input preparation time: **<1 second**
  - Program cycles: ~500M cycles
  - Proving time: ~300 seconds

Input preparation time got reduced almost to zero, but we doubled the proving time.

This difference is understandable - Reth was not designed for zkVM efficiency. Its trie implementation is one of the bottlenecks.

### Witness in Action

I currently work at **ZkCloud**, where I have build a prototype primarily based on `reth-stateless` logic, incorporating MPT/state implementation from RSP.

By constructing RSP's trie using `ExecutionWitness`, we drastically reduced total time to proof:

total-time-to-proof

At this point I got many questions like: "How did you build this low-latency data fetcher? Can you open source the integration?"

[...]

PR #18493, opened 2025-09-16
feat(trie): poc implementation of rsp trie

lightsing:
[rsp trie](https://github.com/succinctlabs/rsp/tree/main/crates/mpt) has proven to be efficiency, since they can be created and serialized outside zkvm, and deserialized inside zkvm.

current serial sparse trie use maps of nibbles to node, would be inefficient in such usage considering the large size of serialized data.

see also:
- rsp choose to move the trie building outside zkvm https://github.com/succinctlabs/rsp/pull/40
- another demo of using rsp implementation https://github.com/paradigmxyz/reth/pull/17416

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10679,6 +10679,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "reth-trie-linked"
+version = "1.7.0"
+dependencies = [
+ "alloy-primitives",
+ "alloy-rlp",
+ "alloy-trie",
+ "itertools 0.14.0",
+ "reth-stateless",
+ "reth-storage-errors",
+ "reth-trie-common",
+ "revm-bytecode",
+ "serde",
+ "thiserror 2.0.16",
+]
+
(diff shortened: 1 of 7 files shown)


mattsse (commented):
@kevaundray do you have thoughts on this?

kevaundray:
Plus one on adding the more efficient zkVM variant of the trie, though I'd lean towards the zeth version since it doesn't have an additional method that must be called by the host (given performance is roughly the same)

koxu1996:
`risc0-ethereum-trie` is **superior** to RSP's MPT. I can quote numbers from few months ago when I tested how `reth-stateless` proving worked with both:

> - Succinct's MPT (437M cycles) - https://share.firefox.dev/46ns09K
> - Risc0 Ethereum Trie (254M cycles) - https://share.firefox.dev/3IEr1rT

You should also know that RSP's trie is based on the old version of Risc0 trie (copied from Zeth).

PR #18494, opened 2025-09-16
refactor: simplify WorkloadExecutor by removing unused rayon dependency

LandLord529:
Simplifies `WorkloadExecutor` by removing the unused rayon thread pool. The executor was carrying around a `rayon::ThreadPool` that was never actually used (marked with `#[expect(unused)]`).

The `WorkloadExecutor` was originally designed to handle mixed CPU/IO workloads with both tokio and rayon, but in practice it only uses `tokio::spawn_blocking`. The rayon pool was dead code that added unnecessary complexity.

Note: `BlockingTaskPool` in `crates/tasks` still uses rayon legitimately for CPU-bound RPC calls like `debug_traceTransaction`, so that remains unchanged.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/executor.rs
+++ b/crates/engine/tree/src/tree/payload_processor/executor.rs
@@ -1,45 +1,35 @@
 //! Executor for mixed I/O and CPU workloads.
 
-use rayon::ThreadPool as RayonPool;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we want to keep this for now

PR #18508, opened 2025-09-17
refactor: EmptyBodyStorage block reader logic

### LINKED ISSUE #18482: Move OpStorage implementation to
### Describe the feature

we have 

https://github.com/paradigmxyz/reth/blob/5f38ff7981d8e0641d5c033d50cb0a4e277b8fb1/crates/optimism/storage/src/chain.rs#L20-L21

which is basically a noop but follows ethereum convention
* set withdrawals if shanghai is activated

this behaviour is useful and we should move the entire OpStorage impl next to

https://github.com/paradigmxyz/reth/blob/5f38ff7981d8e0641d5c033d50cb0a4e277b8fb1/crates/storage/storage-api/src/chain.rs#L88-L89

but name it `EmptyBodyStorage`, remove the `OpTransactionSigned`

then in the optimism crate we simply re-export this aliased to `OpStorage` again

## TODO
* move implementation
* remove default op type
* re-export alias from op storage

### Additional context

_No response_

dharmvr1:
close #18482


dharmvr1:
Hi @mattsse,

I tried to fix the issue, but I’m running into some errors. Could you please suggest the necessary changes or point me in the right direction?

Thanks!

crates/storage/storage-api/src/chain.rs
```diff
@@ -193,3 +194,74 @@ where
         Ok(bodies)
     }
 }
+
+/// A noop storage for chains that don’t have custom body storage.
+#[derive(Debug, Clone, Copy)]
+pub struct EmptyBodyStorage<T = TransactionSigned, H = Header>(PhantomData<(T, H)>);
```
mattsse: let's remove the defaults from here

mattsse (changes requested):
this looks pretty good

left one nit

looks like only the clippy error is missing?

mattsse (approved):
very nice, tysm!

PR #18510, opened 2025-09-17
fix(docs): correct BlockBody root docs and RecoveredBlock “safer variant” references

VolodymyrBg:
- body.rs: clarify roots return Option; say “Returns Some(root)… otherwise None.”
- recovered.rs: point “A safer variant of …” to correct base methods (new, new_sealed).
- Rationale: functions return Option<B256> and callers match on Option; safer variants should reference the non-safe equivalents with identical inputs. No code changes.


mattsse (approved):
I let these slide

PR #18518, opened 2025-09-17
feat(optimism): Implement `local_pending_state` for RPC that uses `pending_flashblock`

RomanHodulak:
Closes #18515

The Optimism RPC implementation has overriden the pending flashblock, but it came short on `local_pending_state`, which uses the default.

In this change, `local_pending_state` is overriden too.

The `pending_flashblock` return type has been changed to `PendingBlock`, even though it is a tiny bit less optimal, because in all the other use-cases it gets converted into block and receipts immediately. Regardless, this return type makes more sense.

### DIFF
--- a/crates/optimism/rpc/src/eth/pending_block.rs
+++ b/crates/optimism/rpc/src/eth/pending_block.rs
@@ -1,17 +1,21 @@
 //! Loads OP pending block for a RPC response.
 
-use std::sync::Arc;
-
 use crate::{OpEthApi, OpEthApiError};
+use alloy_consensus::BlockHeader;
 use alloy_eips::BlockNumberOrTag;
+use reth_chain_state::BlockState;
 use reth_rpc_eth_api::{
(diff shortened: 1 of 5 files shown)


crates/optimism/rpc/src/eth/pending_block.rs
```diff
@@ -60,4 +64,23 @@ where
...
+        };
+
+        let latest_historical = self
+            .provider()
+            .history_by_block_hash(pending_block.block().parent_hash())
+            .map_err(Self::Error::from_eth_err)?;
+
+        let state = BlockState::from(pending_block);
+
+        Ok(Some(Box::new(state.state_provider(latest_historical)) as StateProviderBox))
```
mattsse, line 84: this makes me think, we could even return None here?

same with the default impl

but this also works and we could consider removing the option

mattsse (approved):
cool

PR #18520, opened 2025-09-17
feat: relax `OpEngineValidatorBuilder` for more custom payload types

hai-rise:
We override `PayloadAttributes` and `PayloadBuilderAttributes` in our custom `PayloadTypes`, but keep `type ExecutionData = OpExecutionData`. This PR enables custom chains, like ours, to reuse `OpEngineValidatorBuilder` instead of writing a manual wrapper.


mattsse (approved):
ty ser

PR #18524, opened 2025-09-17
fix: make SealedBlock::encode symmetrical and zero-copy

### LINKED ISSUE #18002: SealedBlock encodable only encodes the body

radik878:
- Encode RLP list header + header + body without clone/into_block
- Add test: SealedBlock::encode bytes == into_block().encode bytes
fix #18002

### DIFF
--- a/crates/primitives-traits/src/block/sealed.rs
+++ b/crates/primitives-traits/src/block/sealed.rs
@@ -308,10 +308,41 @@ impl<B: Block> Deref for SealedBlock<B> {
 
 impl<B: Block> Encodable for SealedBlock<B> {
     fn encode(&self, out: &mut dyn BufMut) {
-        // Compute payload length as the sum of the header and body payload lengths.
-        // This must exclude the outer list header bytes to be symmetrical with plain Block RLP.
-        let payload_length =
-            alloy_rlp::Encodable::length(self.header()) + alloy_rlp::Encodable::length(self.body());
+        // Derive the list payload length from the total block RLP length to avoid
+        // making assumptions about how `B` computes inner lengths.
+        let total_len = B::rlp_length(self.header(), self.body());
+
+        // Solve: total_len = header_len(payload_len) + payload_len
+        // where header_len(x) = 1              if x < 56
+        //                      = 1 + nbytes(x) otherwise
+        let payload_length = {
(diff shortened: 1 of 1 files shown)


crates/primitives-traits/src/block/sealed.rs
```diff
@@ -308,8 +308,13 @@ impl<B: Block> Deref for SealedBlock<B> {
...
     fn encode(&self, out: &mut dyn BufMut) {
-        // TODO: https://github.com/paradigmxyz/reth/issues/18002
-        self.clone().into_block().encode(out);
+        // Compute payload length as the sum of the header and body payload lengths.
+        // This must exclude the outer list header bytes to be symmetrical with plain Block RLP.
+        let payload_length =
+            alloy_rlp::Encodable::length(self.header()) + alloy_rlp::Encodable::length(self.body());
+        alloy_rlp::Header { list: true, payload_length }.encode(out);
+        self.header().encode(out);
+        self.body().encode(out);
```
mattsse: this is slightly problematic because this now assumes that `B` always decodes like this, which may or may not be the case

radik878: > this is slightly problematic because this now assumes that `B` always decodes like this, which may or may not be the case

Fixed: now we use B::rlp_length(header, body) and calculate payload_length for the list header by reverse calculation, which:
preserves symmetry with plain Block RLP,
does not make assumptions about internal lengths,
remains zero-copy during actual writing.

radik878: > this is slightly problematic because this now assumes that `B` always decodes like this, which may or may not be the case

corrected

PR #18525, opened 2025-09-17
deps: Upgrade `alloy` and `alloy-evm` versions `1.0.30` => `1.0.32` and `0.21.0` => `0.21.1` respectively

RomanHodulak:
Brings in revm_spec functions.


mattsse (approved):
makes sense to do this in two steps

PR #18530, opened 2025-09-17
fix(rpc): use flashblock when preparing tx response on gettxreceipt

mablr:
Fixes #18529

Inline build of tx receipt using flashblock's `BlockAndReceipts` instead of calling `build_transaction_receipt` method to avoid fetching the receipts from canonical block.

### DIFF
--- a/crates/optimism/rpc/src/eth/transaction.rs
+++ b/crates/optimism/rpc/src/eth/transaction.rs
@@ -9,7 +9,10 @@ use reth_optimism_primitives::DepositReceipt;
 use reth_primitives_traits::{SignedTransaction, SignerRecoverable};
 use reth_rpc_convert::transaction::ConvertReceiptInput;
 use reth_rpc_eth_api::{
-    helpers::{spec::SignersForRpc, EthTransactions, LoadReceipt, LoadTransaction},
+    helpers::{
+        receipt::calculate_gas_used_and_next_log_index, spec::SignersForRpc, EthTransactions,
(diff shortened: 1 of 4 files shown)


crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -80,15 +82,49 @@ where
...
+                        let meta = tx.meta();
+                        let all_receipts = &block_and_receipts.receipts;
+                        let mut gas_used = 0;
+                        let mut next_log_index = 0;
+
+                        if meta.index > 0 {
+                            for r in all_receipts.iter().take(meta.index as usize) {
+                                gas_used = r.cumulative_gas_used();
+                                next_log_index += r.logs().len();
+                            }
```
mattsse: could we move this into a helper function the PendingBlockReceipt type?

mattsse (changes requested):
cool, one suggestion

crates/optimism/rpc/src/eth/transaction.rs
mablr: I've set it as a free function taking as input index and receipts slice.
Is this clean enough for Reth's standards ?

mattsse (approved):
nice

crates/optimism/rpc/src/eth/transaction.rs
mattsse: perfectly fine, even better here imo

Issue #18522, opened 2025-09-17
Use noopprovider for pending block state root


mattsse:
### Describe the feature

https://github.com/paradigmxyz/reth/blob/fabf3e84d401df6839d076762fbb07fbc96a26d1/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs#L369-L370

we currently produce the real state root which can cause bad latency on pending tag

we should do

https://github.com/paradigmxyz/reth/blob/fabf3e84d401df6839d076762fbb07fbc96a26d1/crates/optimism/flashblocks/src/worker.rs#L104-L104

here

## TODO
* use noopprovider for pending block stateroot

PR #18537, opened 2025-09-18
fix: use noopprovider for pending block state root(follow-up)

leniram159:
Following #18522, I found another opportunity to apply the `NoopProvider` optimization.


mattsse (commented):
her we definitely want the real stateroot because we want to generate the real block

Issue #18133, opened 2025-08-28
Implement `admin_clearTxpool` RPC method


pycckuu:
### Describe the feature

implement `admin_clearTxPool` RPC method that clears all transactions from the mempool, returning the count of cleared  transactions. Method should be in the admin namespace with appropriate access controls

### Additional context

In some L2 architectures where the L1 blockchain acts as the sequencer, we need to maintain strict transaction ordering between L1 and L2 layers. This ensures proper mempool clearing and prevents ordering discrepancies that could arise from transaction persistence issues.

mattsse:
we actually have this now

PR #18539, opened 2025-09-18
feat(rpc): add admin_clearTxpool api

wetkeyboard:
implement the https://github.com/paradigmxyz/reth/issues/18133


mattsse (approved):
I think this is useful

since this is in the admin namespace which shouldnt be exposed I think we can add this, I think this is useful

PR #18542, opened 2025-09-18
feat(rpc): Add `convert_receipt_with_block` method to `ReceiptConverter`

### LINKED ISSUE #18540: Add ReceiptConverter::convert_receipts_with_block
### Describe the feature

we currently only have receipt input:

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/rpc/rpc-convert/src/transaction.rs#L52-L57

for op it's beneficial to also provide the block if we have it because we need to read it in order to populate the l1 context fields:

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/optimism/rpc/src/eth/receipt.rs#L52-L63

## TODO
* add ReceiptConverter::convert_receipts_with_block, by default this can just delegate to convert_receipt without the block
* change the optimism so that this just delegates to the new function

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/optimism/rpc/src/eth/receipt.rs#L52-L63

### Additional context

_No response_

RomanHodulak:
Closes #18540

Adds a method for converting receipts with block as a parameter. This is useful for L2 engine and can be used to avoid a block fetch.

### DIFF
--- a/crates/rpc/rpc-convert/src/transaction.rs
+++ b/crates/rpc/rpc-convert/src/transaction.rs
@@ -20,7 +20,8 @@ use reth_evm::{
     ConfigureEvm, SpecFor, TxEnvFor,
 };
 use reth_primitives_traits::{
-    HeaderTy, NodePrimitives, SealedHeader, SealedHeaderFor, TransactionMeta, TxTy,
+    BlockTy, HeaderTy, NodePrimitives, SealedBlock, SealedHeader, SealedHeaderFor, TransactionMeta,
+    TxTy,
 };
 use revm_context::{BlockEnv, CfgEnv, TxEnv};
 use std::{convert::Infallible, error::Error, fmt::Debug, marker::PhantomData};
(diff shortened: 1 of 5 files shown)


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -55,6 +55,16 @@ pub trait ReceiptConverter<N: NodePrimitives>: Debug + 'static {
         &self,
         receipts: Vec<ConvertReceiptInput<'_, N>>,
     ) -> Result<Vec<Self::RpcReceipt>, Self::Error>;
+
+    /// Converts a set of primitive receipts to RPC representations. It is guaranteed that all
+    /// receipts are from `block`.
+    fn convert_receipts_with_block(
+        &self,
+        receipts: Vec<ConvertReceiptInput<'_, N>>,
+        _block: N::Block,
```
mattsse: I think we can use &N::Block

because most of the type we have this in form of `Arc` and can leave it up to the consumer if they need to clone

mattsse (changes requested):
smol nit

mattsse (approved):

klkvr (commented):
can we integrate it here right away?

https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/rpc/rpc-eth-api/src/helpers/block.rs#L161

mattsse:
good point, let me quickly do that since ci failed anyway

PR #18547, opened 2025-09-18
chore: fix MockProvider block tracking to match real provider behavior

indmind:
cleaned up how MockProvider tracks blocks to avoid inconsistencies.

* removed `head_number` and `head_hash` — now `blocks` HashMap is the single source of truth.
* updated `best_block_number`, `last_block_number`, and `chain_info` to compute the best block dynamically from the keys of `blocks`.
* `chain_info` now derives `best_hash` and `best_number` directly from `blocks`, matching real provider logic.

**these changes fix the previous mismatch between separate fields and `blocks`, making the test provider behavior consistent and more realistic.**


mattsse (commented):
we want this separately here

PR #18548, opened 2025-09-18
chore: fix `apply_withdrawals_contract_call` compilation and ABI issues

indmind:
fixed a couple of issues that were preventing the code from compiling:

* updated the function signature of `apply_withdrawals_contract_call` to use generic parameters `<DB, E>` with proper `where` constraints instead of `&mut impl Evm<…>`. This resolves compilation errors with `impl Trait` and associated constraints.
* corrected the ABI call: replaced the non-existent `withdrawalsCall` with the generated `withdrawals` function/structure from the `sol!` macro.
* rewrote `call_data` to ensure proper ABI encoding.

these changes make the function compile correctly and fully compatible with the `sol!` macro and the `Evm` interface.


mattsse (commented):
this is equivalent

PR #18556, opened 2025-09-18
feat: add `Future` AT to `LaunchNode` and allow customizing local attributes builder

### LINKED ISSUE #18534: Make `DebugNode` configurable in DebugNodeLauncher
### Describe the feature

we currently require

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/node/builder/src/builder/mod.rs#L670-L670

on

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/node/builder/src/builder/mod.rs#L665-L665

trait 

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/node/builder/src/launch/debug.rs#L45-L72

is currently stateless

ideally we can configure

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/node/builder/src/launch/debug.rs#L186-L186

separately so that we can add more context

something like

```rust
pub struct DebugNodeLauncher<F, L = EngineNodeLauncher> {
    inner: L,
    attributes_builder: F,
}
```

so that Fn could be mapped on demand, e.g. for customizing 

https://github.com/paradigmxyz/reth/blob/2b68d3a42410fa09b7fcdacddb761e031d0ac297/crates/node/builder/src/launch/debug.rs#L67-L71

on demand

### Additional context

_No response_

klkvr:
Closes https://github.com/paradigmxyz/reth/issues/18534

Adds `Future: IntoFuture` AT to `LaunchNode`. For `EngineNodeLauncher` it's just a boxed future, and for `DebugNodeLauncher` it's a custom type with helpers to either override local payload attributes builder entirely, or to map the payload attributes produced by the basic builder defined on the node itself.

### DIFF
--- a/crates/node/builder/src/launch/debug.rs
+++ b/crates/node/builder/src/launch/debug.rs
@@ -1,12 +1,19 @@
 use super::LaunchNode;
 use crate::{rpc::RethRpcAddOns, EngineNodeLauncher, Node, NodeHandle};
+use alloy_consensus::transaction::Either;
 use alloy_provider::network::AnyNetwork;
 use jsonrpsee::core::{DeserializeOwned, Serialize};
 use reth_chainspec::EthChainSpec;
 use reth_consensus_debug_client::{DebugConsensusClient, EtherscanBlockProvider, RpcBlockProvider};
 use reth_engine_local::LocalMiner;
-use reth_node_api::{BlockTy, FullNodeComponents, PayloadAttributesBuilder, PayloadTypes};
-use std::sync::Arc;
+use reth_node_api::{
+    BlockTy, FullNodeComponents, PayloadAttrTy, PayloadAttributesBuilder, PayloadTypes,
+};
+use std::{
+    future::{Future, IntoFuture},
+    pin::Pin,
+    sync::Arc,
+};
 use tracing::info;
 
 /// [`Node`] extension with support for debugging utilities.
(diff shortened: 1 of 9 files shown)


crates/node/builder/src/launch/debug.rs
```diff
@@ -104,16 +111,54 @@ impl<L> DebugNodeLauncher<L> {
...
+pub struct DebugNodeLauncherFuture<L, Target, N>
+where
+    N: FullNodeComponents<Types: DebugNode<N>>,
+{
+    inner: L,
+    target: Target,
+    local_payload_attributes_builder:
+        Option<Box<dyn PayloadAttributesBuilder<PayloadAttrTy<N::Types>>>>,
+    map_attributes:
+        Option<Box<dyn Fn(PayloadAttrTy<N::Types>) -> PayloadAttrTy<N::Types> + Send + Sync>>,
```
mattsse, line 125: I see we need this map because we cant replace the local_payload_attributes_builder directly

this is kinda like `Stack` layer

mattsse (approved):
pretty cool, making this return future is nice, now we can add even more functionality in there if needed

crates/node/builder/src/launch/debug.rs
klkvr: we could make it work without extra `map_attributes` fields by wrapping the inner builder but it would require locking `Target` type to obtain chainspec, so I'm fine with a second field for now

PR #18557, opened 2025-09-18
chore(chainspec): reuse local hardforks in DEV instead of cloning again

MozirDmitriy:
Replace redundant DEV_HARDFORKS.clone() with the already cloned local hardforks when constructing DEV ChainSpec.
This removes an unnecessary allocation and aligns DEV with the pattern used by MAINNET/SEPOLIA/HOLESKY/HOODI and OP_DEV, improving consistency without changing behavior.


mattsse (approved):
fine

PR #18559, opened 2025-09-18
chore(evm): add public constructor to `BlockAssemblerInput`

0xOsiris:
The struct `BlockAssemblerInput` is marked `#[non_exhaustive]` which makes it not able to be instantiated from external crates without a constructor.


mattsse (commented):
@klkvr should we add this

klkvr (approved):
lgtm

I'd be okay with removing `non_exhaustive` too given that both changes prevent us from adding fields in a non-breaking way

mattsse (approved):

PR #18564, opened 2025-09-19
fix(cache): Ensure execution cache remains locked until updated

yongkangc:
closes #18541

**Problem:**
- Prevent a possible race condition where the ExecutionCache can be updated while a background prewarming task is still using it by another background prewarming task. 
-  e.g  having a fork block and for `cache_for` to be called before the previous prewarm task exits

**Solution** - using `Arc<()>` as a lightweight, reference-counted lock:
- When the `PrewarmCacheTask` is spawned, it is given its own clone of the `usage_guard`.
- This task holds onto the guard for its entire lifetime. As long as the task is running, the `strong_count` remains > 1, keeping the cache safely locked.
- When the task finishes, its guard is dropped automatically (RAII), decrementing the count and helping to release the lock.

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -543,8 +543,7 @@ impl SavedCache {
         (self.caches, self.metrics)
     }
 
-    /// Checks if the cache is currently available to be checked out.
-    /// It is considered available if no other component holds a reference to its usage guard.
+    /// Returns true if the cache is idle (no other tasks are using it).
     pub(crate) fn is_available(&self) -> bool {
         Arc::strong_count(&self.usage_guard) == 1
     }
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -542,7 +542,7 @@ impl SavedCache {
         (self.caches, self.metrics)
     }
 
-    /// Returns true if the cache is idle (no other tasks are using it).
+    /// Returns true if the cache is available for use (no other tasks are currently using it).
     pub(crate) fn is_available(&self) -> bool {
         Arc::strong_count(&self.usage_guard) == 1
     }
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -803,7 +803,6 @@ mod tests {
     }
 
     // Tests for SavedCache locking mechanism
-
     #[test]
     fn test_saved_cache_is_available() {
         let execution_cache = ExecutionCacheBuilder::default().build_caches(1000);
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -523,7 +523,7 @@ pub(crate) struct SavedCache {
 
     /// A guard to track in-flight usage of this cache.
     /// The cache is considered available if the strong count is 1.
-    pub(crate) usage_guard: Arc<()>,
+    usage_guard: Arc<()>,
 }
 
 impl SavedCache {
@@ -552,11 +552,23 @@ impl SavedCache {
         &self.caches
     }
 
+    /// Returns the metrics associated with this cache.
+    pub(crate) const fn metrics(&self) -> &CachedStateMetrics {
+        &self.metrics
+    }
+
     /// Updates the metrics for the [`ExecutionCache`].
     pub(crate) fn update_metrics(&self) {
         self.metrics.storage_cache_size.set(self.caches.total_storage_slots() as f64);
         self.metrics.account_cache_size.set(self.caches.account_cache.entry_count() as f64);
         self.metrics.code_cache_size.set(self.caches.code_cache.entry_count() as f64);
+    }
+}
+
+#[cfg(test)]
+impl SavedCache {
+    fn clone_guard_for_test(&self) -> Arc<()> {
+        self.usage_guard.clone()
     }
 }
 
@@ -812,7 +824,7 @@ mod tests {
         assert!(cache.is_available(), "Cache should be available initially");
 
         // Clone the usage guard (simulating it being handed out)
-        let _guard = cache.usage_guard.clone();
+        let _guard = cache.clone_guard_for_test();
 
         // Now the cache should not be available (two references)
         assert!(!cache.is_available(), "Cache should not be available with active guard");
@@ -825,8 +837,8 @@ mod tests {
             SavedCache::new(B256::from([2u8; 32]), execution_cache, CachedStateMetrics::zeroed());
 
         // Create multiple references to the usage guard
-        let guard1 = cache.usage_guard.clone();
-        let guard2 = cache.usage_guard.clone();
+        let guard1 = cache.clone_guard_for_test();
+        let guard2 = cache.clone_guard_for_test();
         let guard3 = guard1.clone();
 
         // Cache should not be available with multiple guards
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -556,17 +569,21 @@ impl Drop for CacheTaskHandle {
...
 
 impl ExecutionCache {
-    /// Returns the cache if the currently store cache is for the given `parent_hash`
-    pub(crate) fn get_cache_for(&self, parent_hash: B256) -> Option<SavedCache> {
-        let cache = self.inner.read();
-        cache
-            .as_ref()
-            .and_then(|cache| (cache.executed_block_hash() == parent_hash).then(|| cache.clone()))
+    /// Returns the idle cache for `parent_hash` with a lock guard, or None if unavailable.
+    pub(crate) fn get_cache_for(&self, parent_hash: B256) -> Option<(SavedCache, Arc<()>)> {
```
mattsse: what if we move the `Arc<()>` inside the `SavedCache` since this is the thing that we pass around?

this would be less invasive than tracking the guard separately

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -494,15 +498,22 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
 /// Access to the spawned [`PrewarmCacheTask`].
 #[derive(Debug)]
 pub(crate) struct CacheTaskHandle {
-    /// The shared cache the task operates with.
-    cache: StateExecutionCache,
-    /// Metrics for the caches
-    cache_metrics: CachedStateMetrics,
+    saved_cache: SavedCache,
+    #[allow(dead_code)]
+    cache_usage_guard: Arc<()>,
```
mattsse: then we dont need to track this separately here I believe

mattsse (changes requested):
conceptionally this makes sense

to make this less complex, I'd like to try moving the guard inside the cache container type and have this in one place then we only ever need to check this when we try to acquire the cache

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: hmm this makes more sense, thanks!

yongkangc (commented):
@mattsse just addressed ur clean up feedback and ready for review :)

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -561,12 +561,13 @@ struct ExecutionCache {
 }
 
 impl ExecutionCache {
-    /// Returns the cache if the currently store cache is for the given `parent_hash`
+    /// Returns the idle cache for `parent_hash`, or `None` if unavailable.
```
shekhirin: can we expand what does idle mean here?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -561,12 +561,13 @@ struct ExecutionCache {
...
     pub(crate) fn get_cache_for(&self, parent_hash: B256) -> Option<SavedCache> {
         let cache = self.inner.read();
-        cache
-            .as_ref()
-            .and_then(|cache| (cache.executed_block_hash() == parent_hash).then(|| cache.clone()))
+        cache.as_ref().and_then(|cache| {
+            (cache.executed_block_hash() == parent_hash && cache.is_available())
+                .then(|| cache.clone())
```
shekhirin: is this `filter` + `cloned`?

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -796,4 +801,46 @@ mod tests {
         let slot_status = caches.get_storage(&address, &storage_key);
         assert_eq!(slot_status, SlotStatus::Empty);
     }
+
+    // Tests for SavedCache locking mechanism
+
```
shekhirin:
```suggestion
```

shekhirin (approved):
makes sense, minor nits

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: good point

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: I've refactored it to use filter().cloned() which makes the intent clearer. Thanks for pointing this out!

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -520,16 +520,16 @@ pub(crate) struct SavedCache {
 
     /// Metrics for the cached state provider
     metrics: CachedStateMetrics,
+
+    /// A guard to track in-flight usage of this cache.
+    /// The cache is considered available if the strong count is 1.
+    pub(crate) usage_guard: Arc<()>,
```
mattsse: can we make this super private, so we don't have to think about where else this can be accessed?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -357,15 +357,14 @@ where
...
-                SavedCache::new(parent_hash, cache, CachedStateMetrics::zeroed())
-            })
+        if let Some(cache) = self.execution_cache.get_cache_for(parent_hash) {
+            debug!("reusing execution cache");
+            cache
+        } else {
+            debug!("creating new execution cache on cache miss");
+            let cache = ExecutionCacheBuilder::default().build_caches(self.cross_block_cache_size);
+            SavedCache::new(parent_hash, cache, CachedStateMetrics::zeroed())
+        }
```
mattsse, line 368: there's no actual change here besides replacing this with let some else?

mattsse (changes requested):
we need to do a bit more work, because rn, this doesn't take effect because of:

https://github.com/paradigmxyz/reth/blob/573957895fe45924447a9348412b5db508fcf0e1/crates/engine/tree/src/tree/payload_processor/mod.rs#L317-L317

and:

https://github.com/paradigmxyz/reth/blob/573957895fe45924447a9348412b5db508fcf0e1/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L249-L249

so what we should do is

change how we store this is the prewarm context which is held until we finally update the cache at the end:

https://github.com/paradigmxyz/reth/blob/573957895fe45924447a9348412b5db508fcf0e1/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L158-L162

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: no change here - thought it was easier to read?

yongkangc:
@mattsse thanks for pointing out the flaw in the lifetime.

addressed it [in this commit here](https://github.com/paradigmxyz/reth/pull/18564/commits/65fc90227478a1eec7a1065d6faa11f64476c756) where we moved the usage guard into `PrewarmContext` and only released during `saved_cache`

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -246,8 +250,7 @@ where
 {
     pub(super) env: ExecutionEnv<Evm>,
     pub(super) evm_config: Evm,
-    pub(super) cache: StateExecutionCache,
-    pub(super) cache_metrics: CachedStateMetrics,
+    pub(super) saved_cache: SavedCache,
```
mattsse, line 253: simpler 👍

mattsse (approved):
lgtm!

PR #18570, opened 2025-09-19
fix(reth-bench): do not panic on empty results

shekhirin:
We shouldn't panic, otherwise the error isn't even bubbled up when running from `reth-bench-compare`.

### DIFF
--- a/bin/reth-bench/src/bench/output.rs
+++ b/bin/reth-bench/src/bench/output.rs
@@ -1,6 +1,7 @@
 //! Contains various benchmark output formats, either for logging or for
 //! serialization to / from files.
 
+use eyre::OptionExt;
 use reth_primitives_traits::constants::GIGAGAS;
 use serde::{ser::SerializeStruct, Serialize};
 use std::time::Duration;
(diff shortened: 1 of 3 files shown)


bin/reth-bench/src/bench/output.rs
```diff
@@ -145,15 +146,14 @@ pub(crate) struct TotalGasOutput {
 
 impl TotalGasOutput {
     /// Create a new [`TotalGasOutput`] from a list of [`TotalGasRow`].
-    pub(crate) fn new(rows: Vec<TotalGasRow>) -> Self {
+    pub(crate) fn new(rows: Vec<TotalGasRow>) -> eyre::Result<Self> {
         // the duration is obtained from the last row
-        let total_duration =
-            rows.last().map(|row| row.time).expect("the row has at least one element");
+        let total_duration = rows.last().map(|row| row.time).ok_or_eyre("empty results")?;
```
mattsse, line 151: should this just use duration zero in this case?

~~unclear when rows can be empty~~

would happen if

this is closed

https://github.com/paradigmxyz/reth/blob/4fcc4457c144a41df443c1b297e58ccbe91735c1/bin/reth-bench/src/bench/new_payload_fcu.rs#L96-L101

klkvr (approved):

mattsse (commented):
maybe we should handle/report this case differently

PR #18573, opened 2025-09-19
fix: data alignment + cleaner plots

zkpepe:
synced all arrays with a finite mask so nothing goes out of step.
added a space in gas units for readability.
made `axes` always a list to avoid plot errors.


mattsse (commented):
this has been working fine

PR #18574, opened 2025-09-19
chore: fix error type for msg conversion

hexcow:
Hey, ran into a little mismatch here - `msg.to_string()` gives a `String`, but `Self::Other` wants a `Box<dyn Error + Send + Sync>`.

Just wrapped the string in a box so it plays nicely with the expected type. Should clear up the compile error.


mattsse (commented):
the existing conversion is valid

PR #18575, opened 2025-09-19
chore: Fix key generation with boxed RNG

0xsimka:
I noticed that key generation was failing when using a boxed RNG.

Updated all instances of:

```rust
Keypair::new(&secp, &mut self.rng)
```

to:

```rust
Keypair::new(&secp, &mut *self.rng)
```

This ensures `self.rng` (a `Box<dyn RngCore>`) is correctly passed as `&mut dyn RngCore`.
The change fixes the mismatch and allows key creation to work as intended.


mattsse (commented):
this currently works as expected

PR #18576, opened 2025-09-19
chore: clippy manual_string_new warning in version.rs

yongkangc:
## Summary
Use `Cow::Borrowed` instead of `Cow::Owned` for all compile-time constant strings from `env!()` macro to avoid unnecessary heap allocations.

The `env!()` macro returns `&'static str` - compile-time constants that live for the entire program. Using `Cow::Owned(...to_string())` unnecessarily allocates these strings on the heap when we can just borrow the existing static references.

### DIFF
--- a/crates/node/core/src/version.rs
+++ b/crates/node/core/src/version.rs
@@ -108,13 +108,13 @@ pub fn version_metadata() -> &'static RethCliVersionConsts {
 pub fn default_reth_version_metadata() -> RethCliVersionConsts {
     RethCliVersionConsts {
         name_client: Cow::Borrowed("Reth"),
-        cargo_pkg_version: Cow::Owned(env!("CARGO_PKG_VERSION").to_string()),
-        vergen_git_sha_long: Cow::Owned(env!("VERGEN_GIT_SHA").to_string()),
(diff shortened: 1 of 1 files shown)


crates/node/core/src/version.rs
```diff
@@ -113,7 +113,7 @@ pub fn default_reth_version_metadata() -> RethCliVersionConsts {
         vergen_git_sha: Cow::Owned(env!("VERGEN_GIT_SHA_SHORT").to_string()),
         vergen_build_timestamp: Cow::Owned(env!("VERGEN_BUILD_TIMESTAMP").to_string()),
         vergen_cargo_target_triple: Cow::Owned(env!("VERGEN_CARGO_TARGET_TRIPLE").to_string()),
```
mattsse: do we need to do this for these others as well?

yongkangc: yup we should need to fix for the others - tho clippy only warned for empty str

yongkangc: making updates in one sec

yongkangc: just updated @mattsse 👍🏻

mediocregopher (approved):

PR #18579, opened 2025-09-19
chore(db): Simplifications to trie-related storage-api methods

mediocregopher:
The `insert_hash` method was not being used at all in reth, it is removed completely. Other methods are slightly modified in order to simplify/remove their interaction with the trie tables.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2336,7 +2336,7 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
             }
         }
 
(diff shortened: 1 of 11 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3076,7 +2985,6 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
 
         // insert hashes and intermediate merkle nodes
         self.write_hashed_state(&hashed_state)?;
-        self.write_trie_updates(&trie_updates)?;
```
mediocregopher, line 3079: This method is only ever used in tests, and `trie_updates` was always being passed in as empty.

yongkangc (approved):
This is nice, lgtm

mattsse (approved):
lgtm

PR #18580, opened 2025-09-19
chore: Remove `reth recover storage-tries` sub-command

mediocregopher:
The sub-command is redundant with the `reth db repair-tries` sub-command, which repairs both storage and account tries as well as outputting the exact fixes which were made.


Rjected (approved):
lgtm, iirc this was originally added to recover an older trie bug which were just in storage tries

yongkangc (approved):
nice, this makes sense

shekhirin (approved):

PR #18581, opened 2025-09-19
test: add test case for op tx env conversion

mattsse:


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -1099,6 +1099,7 @@ mod transaction_response_tests {
     use alloy_network::Ethereum;
     use alloy_primitives::{Address, Signature, B256, U256};
     use alloy_rpc_types_eth::Transaction;
+    use op_revm::transaction::OpTxTr;
```
RomanHodulak: Should this be also feature gated?

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -1155,4 +1156,23 @@ mod transaction_response_tests {
 
         assert!(result.is_ok());
     }
+
+    #[cfg(feature = "op")]
```
RomanHodulak: Might be good to wrap it in a feature gated module so that the test and the usage statement need just one feature gate.

mattsse: good point, done

Issue #18467, opened 2025-09-15
Trie ChangeSets: Engine API TreeState


mediocregopher:
### Describe the feature

The engine API should be updated so that it takes advantage of the new historical trie cursor factory to *always* use a complete and correct trie state during validation.

- When a block’s parent is tip of db then nothing changes.
- When a block’s ancestor is tip of db then the block overlays all trie updates previous to it onto the DB.
- When a block’s ancestor is not tip of db then the block overlays all deltas between tip and its parent onto its view of the DB. These deltas are *not* kept as part of the execution result; any children of this block will need to re-retrieve the deltas, in case the DB tip changed.

ExecutedBlockWithTrieUpdates is modified to never have missing TrieUpdates. Possibly TrieUpdates can be integrated directly into ExecutedBlock.

PR #18584, opened 2025-09-19
feat: Write trie changesets to DB on engine persistence

mediocregopher:
Closes #18465

This introduces two new provider methods for writing the trie changeset data based on the current Accounts/StorageTrie dataset. These will be used in the engine API (as added by this PR) as well as in the pipeline sync.

There is also a slight refactor; I added a helper for generating the storage changesets in the same module space as the existing helper for storage reverts, which was alone in a module called `bundle_state`. I renamed the module to `changeset_utils` since everything in side has to do with populating changeset tables.


crates/storage/provider/src/writer/mod.rs
```diff
@@ -180,9 +181,18 @@ where
...
                 .write_hashed_state(&Arc::unwrap_or_clone(hashed_state).into_sorted())?;
-            self.database().write_trie_updates(
-                trie.as_ref().ok_or(ProviderError::MissingTrieUpdates(block_hash))?,
-            )?;
+
+            let trie_updates =
+                trie.take_present().ok_or(ProviderError::MissingTrieUpdates(block_hash))?;
+
+            // sort trie updates and insert changesets
+            // TODO(mediocregopher): We should rework `write_trie_updates` to also accept a
```
mediocregopher, line 189: This will be done as part of [18467](https://github.com/paradigmxyz/reth/issues/18467) I think

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2340,11 +2347,50 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
     }
+
+    /// Records the current values of all trie nodes which will be updated using the [`TrieUpdates`]
+    /// into the trie changesets tables.
+    ///
+    /// The intended usage of this method is to call it _prior_ to calling `write_trie_updates` with
+    /// the same [`TrieUpdates`].
+    ///
+    /// Returns the number of keys written.
+    fn write_trie_changesets(
```
shekhirin, line 2358: I wonder if we should combine this just into `write_trie_updates` method for two reasons:
1. We will not need to be careful with calling `write_trie_updates` before `write_trie_updates`, because it will all be encapsulated in the same method
2. We can update `AccountsTrie` and `StoragesTrie` at the same time as writing changesets

mediocregopher: I _think_ it will be necessary to have a separate `write_trie_changesets` style method for #18464, though I'm not 100% this is the exact form it will take. But we could have the method be called internal to `write_trie_updates` automatically, but still expose it through the trait?

shekhirin: makes sense to have it separate for `MerkleChangeSets` stage

crates/storage/provider/src/changesets_utils/trie.rs
```diff
@@ -0,0 +1,166 @@
...
+                            // If the path is ahead of the cursor then seek the cursor forward and
+                            // loop. The cursor will either seek to the path or beyond it.
+                            let path = *path;
+                            if let Err(err) = self.seek_cursor(path) {
+                                return Some(Err(err))
+                            }
+                        }
+                    }
+                }
+            }
```
Rjected: does this ever loop more than once? the final branch is the only one that doesn't return, and it looks like as long as `seek_cursor` works as expected, we should return right after that. If so, maybe we can do something like:
```rust
let curr_cursor = self.cursor_current.as_mut();

let Some(curr_path) = self.paths.peek() else {
   return None
};

if curr_cursor.is_some_and(|(cursor_path, _)| curr_path > cursor_path) {
    if let Err(err) = self.seek_cursor(*path) {
        return Some(Err(err))
    }
}

// do big match on curr_cursor

File: crates/storage/provider/src/changesets_utils/trie.rs [hunk 4]
```diff
@@ -0,0 +1,166 @@
...
+    impl Iterator<Item = Result<(Nibbles, Option<BranchNodeCompact>), DatabaseError>>,
+    DatabaseError,
+> {
+    let all_nodes = all_nodes.map(|e| e.map(|e| (e.1.nibbles.0, Some(e.1.node))));
+
+    let merged = merge_join_by(curr_values_of_changed, all_nodes, |a, b| match (a, b) {
+        (Err(_), _) => Ordering::Less,
+        (_, Err(_)) => Ordering::Greater,
+        (Ok(a), Ok(b)) => a.0.cmp(&b.0),
+    });
```
Rjected, line 125: i see, so this merge join fn is essentially the diff operation

Rjected (commented):
i like the rename, wondering if we can make the `loop` simpler

crates/storage/provider/src/changesets_utils/trie.rs
mediocregopher: Good call, I refactored this to get rid of the loop, and ended up being able to remove the Peekable over the paths iterator as well :fire:

crates/storage/provider/src/providers/database/provider.rs
mediocregopher: I looked into calling the changeset methods from the normal write trie update methods, but without passing a SortedTrieUpdates into the write trie update methods it will be pretty awkward, I think this would be better left until I make that change later.

shekhirin (approved):

Rjected (approved):
lgtm

PR #18585, opened 2025-09-19
chore(payload-builder): relax Sync bounds on resolve futures

VolodymyrBg:
Removed unnecessary Sync bound from PayloadFuture and ResolvePayloadFuture. The resolve future is not shared across threads; it is produced by the service, sent over a oneshot, and awaited by a single consumer. Tokio channels and task spawning only require Send, not Sync. Keeping Sync restricts valid implementations (e.g., futures wrapping non-Sync internals) without functional benefit. This change relaxes constraints while preserving behavior.


mattsse (approved):
this seems fine

PR #18586, opened 2025-09-19
chore: specialize `send_raw_transaction_sync` for op-reth with flashblocks support

Peponks9:
This PR implements the specialization of `send_raw_transaction_sync` for Optimism (op-reth), adding flashblocks support for faster transaction confirmations, based on new feedback.

Closes #18372


crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -73,6 +78,59 @@ where
...
+            const TIMEOUT_DURATION: Duration = Duration::from_secs(30);
+
+            tokio::time::timeout(TIMEOUT_DURATION, async {
+                while let Some(notification) = canonical_stream.next().await {
+                    // First check flashblocks for faster confirmation (Optimism-specific)
+                    if let Ok(Some(pending_block)) = this.pending_flashblock() {
+                        let block_and_receipts = pending_block.into_block_and_receipts();
+                        if block_and_receipts.block.body().contains_transaction(&hash) {
+                            if let Some(receipt) = this.transaction_receipt(hash).await? {
+                                return Ok(receipt);
```
mattsse: this await loop we need to change slightly because the canonical_stream and flashblocks are ticking at a different rate

what we want is a tokio::select! here where one branch is canonical_stream.next() and the other is a `WatchStream::new` using:

https://github.com/paradigmxyz/reth/blob/ebe1a8b014555495e3cce640561d0ec7cc30fb1e/crates/optimism/rpc/src/eth/mod.rs#L102-L102

mattsse (changes requested):
almost there

Peponks9 (commented):
thanks for the feedback @mattsse, changes applied

mattsse (approved):
lgtm, ty!

PR #18588, opened 2025-09-20
perf(persistence): lookup segment operation once

hai-rise:
Random low-hanging fruit as we seek to flush blocks as fast as possible... More of a clarity than perf really.

image


mattsse (approved):
easy pickoff

PR #18589, opened 2025-09-20
fix(rpc-engine): don't fetch the pruned block

wetkeyboard:
closes https://github.com/paradigmxyz/reth/issues/18500

Don't fetch the blocks before earliest block, this will prevent the warning log and also improve the performance.


mattsse (approved):
very nice, tysm!

PR #18590, opened 2025-09-20
fix: use timestamp derived max blob count on launch

mattsse:
we should use the head's timestamp here

https://github.com/paradigmxyz/reth/blob/651118ade9849e5c5940f6403ee7af4914c6f3da/crates/transaction-pool/src/validate/eth.rs#L1005-L1010

this is also updated on a per block basis:

https://github.com/paradigmxyz/reth/blob/651118ade9849e5c5940f6403ee7af4914c6f3da/crates/transaction-pool/src/validate/eth.rs#L743-L749

so even if a user misconfigures `with_head_timestamp`, this would be correct once the node is synced.

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -992,7 +992,8 @@ impl<Client> EthTransactionValidatorBuilder<Client> {
 
     /// Configures validation rules based on the head block's timestamp.
     ///
-    /// For example, whether the Shanghai and Cancun hardfork is activated at launch.
+    /// For example, whether the Shanghai and Cancun hardfork is activated at launch, or max blob
+    /// counts.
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -1067,15 +1068,10 @@ impl<Client> EthTransactionValidatorBuilder<Client> {
             max_tx_input_bytes,
             max_tx_gas_limit,
             disable_balance_check,
-            ..
+            max_blob_count,
+            additional_tasks: _,
```
yongkangc, line 1072: i see, with this we make use max_blob_count based on the parameters active at the timestamp of the latest block

yongkangc (approved):
lgtm!

PR #18593, opened 2025-09-20
fix(payload): reversed-order test for validate_execution_requests

GarmashAlex:
The test intended to validate out-of-order detection created a reversed version of a valid request list but mistakenly asserted on requests_with_empty, duplicating an earlier case and not exercising the out-of-order path. This change asserts on requests_valid_reversed, ensuring the validator correctly flags descending request_type sequences as OutOfOrderExecutionRequest per EIP-7685 (strict ascending order, no duplicates, no empty data).


mattsse (approved):
right

PR #18596, opened 2025-09-20
fix(db): add descriptive message to expect() call in MDBX test

zeroprooff:
Replace empty expect("") with meaningful error message "Failed to put AccountsHistory" to improve debugging experience when database operations fail


mattsse (commented):
the unwrap is fine

PR #18597, opened 2025-09-20
feat(network): add default None-case handling with target/label in poll macro

Forostovec:
- Implement dual-arm poll_nested_stream_with_budget!:
  - Arm 1: preserves existing behavior when $on_ready_none is provided.
  - Arm 2: adds default handling on Poll::Ready(None) by logging debug! with $target and $label.
- Leaves metered_poll_nested_stream_with_budget! as a passthrough so both forms are supported.


mattsse (commented):
dont think we need this

PR #18601, opened 2025-09-21
docs: some fixes on discv4 docs

moricho:
- Updated the discv4 spec reference to the latest
- Deleted unnecessary lines in doc comment for `bind` fn
- ...

### DIFF
--- a/crates/net/discv4/src/lib.rs
+++ b/crates/net/discv4/src/lib.rs
@@ -217,7 +217,7 @@ impl Discv4 {
     /// use reth_network_peers::{pk2id, NodeRecord, PeerId};
     /// use secp256k1::SECP256K1;
     /// use std::{net::SocketAddr, str::FromStr};
-    /// # async fn t() -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


mattsse (approved):

crates/net/discv4/src/lib.rs
```diff
@@ -213,12 +213,11 @@ impl Discv4 {
     /// Binds a new `UdpSocket` and creates the service
     ///
     /// ```
-    /// # use std::io;
     /// use reth_discv4::{Discv4, Discv4Config};
     /// use reth_network_peers::{pk2id, NodeRecord, PeerId};
     /// use secp256k1::SECP256K1;
     /// use std::{net::SocketAddr, str::FromStr};
-    /// # async fn t() -> io::Result<()> {
```
mattsse, line 221: this we cant actually remove

mattsse (changes requested):

mattsse (approved):

PR #18602, opened 2025-09-21
fix: avoid panic in new-payload-fcu benchmark

LandLord529:
Found multiple panic-prone issues in `reth-bench new-payload-fcu` command, similar to PR #18570. These panics prevent errors from bubbling up to calling tools like `reth-bench-compare`.

### DIFF
--- a/bin/reth-bench/src/bench/new_payload_fcu.rs
+++ b/bin/reth-bench/src/bench/new_payload_fcu.rs
@@ -81,15 +81,14 @@ impl Command {
 
                 let (safe, finalized) = tokio::join!(safe_block_hash, finalized_block_hash,);
 
-                // Handle safe and finalized block lookups gracefully - they may not exist for early blocks
                 let safe_block_hash = match safe {
                     Ok(Some(block)) => block.header.hash,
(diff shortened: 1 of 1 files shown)


bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -69,12 +81,19 @@ impl Command {
...
+                // Handle safe and finalized block lookups gracefully - they may not exist for early blocks
+                let safe_block_hash = match safe {
+                    Ok(Some(block)) => block.header.hash,
+                    Ok(None) | Err(_) => head_block_hash, // Fallback to head for early blocks or on error
+                };
+                
+                let finalized_block_hash = match finalized {
+                    Ok(Some(block)) => block.header.hash,
+                    Ok(None) | Err(_) => head_block_hash, // Fallback to head for early blocks or on error
+                };
```
Rjected: I'm ok with this, but let's remove comments here that refer to early blocks, because we use `saturating_sub` above, so the worst case is that we are fetching the genesis block

Rjected (dismissed):
just one nit

Rjected:
this also needs fmt

mattsse (approved):
lgtm!

PR #18603, opened 2025-09-21
feat(db): add MDBX put-append for fast ordered puts

hai-rise:
We only want 2 things in life:
1. Flush blocks to disk as fast as possible.
2. Commit write transactions as fast as possible.

This led us to `DatabaseProvider::insert_block`, which is still using put-upsert instead of put-append for tables with naturally increasing keys (like `BlockNumber`) in the happy/hot paths. For example:
https://github.com/paradigmxyz/reth/blob/4ddf3ddb45aedce284135c04e995c0c99d35baa3/crates/storage/provider/src/providers/database/provider.rs#L2747-L2755

This PR first adds `DbTxMut::append` with the optimised implementation for MDBX, and a benchmark to show how put-append is much faster than put-upsert in this scenario:
```
$ cargo bench -p reth-db --features test-utils --bench put
Benchmarking Put/put: Warming up for 3.0000 s
Put/put                 time:   [138.67 ms 138.76 ms 138.86 ms]

Benchmarking Put/append: Warming up for 3.0000 s
Put/append              time:   [80.353 ms 80.538 ms 80.722 ms]
```

Actual integration into key paths should be done in separate PRs with proper testing and benchmarks. Some places may be able to simply fall back to `put`, while others may want to report an error for different handling if the key is unexpectedly not the highest.

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/tx.rs
+++ b/crates/storage/db/src/implementation/mdbx/tx.rs
@@ -340,28 +340,64 @@ impl<K: TransactionKind> DbTx for Tx<K> {
     }
 }
 
+#[derive(Clone, Copy)]
+enum PutKind {
+    /// Default kind that inserts a new key-value or overwrites an existed key.
+    Upsert,
+    /// Append the key-value to the end of the table -- fast path when the new
+    /// key is the highest so far, like the latest block number.
+    Append,
+}
+
+impl PutKind {
+    const fn into_operation_and_flags(self) -> (Operation, DatabaseWriteOperation, WriteFlags) {
+        match self {
+            Self::Upsert => {
+                (Operation::PutUpsert, DatabaseWriteOperation::PutUpsert, WriteFlags::UPSERT)
+            }
+            Self::Append => {
+                (Operation::PutAppend, DatabaseWriteOperation::PutAppend, WriteFlags::APPEND)
+            }
+        }
+    }
+}
+
+impl Tx<RW> {
+    /// The inner implementation mapping to `mdbx_put` that supports different
+    /// put kinds like upserting and appending.
+    fn put<T: Table>(
+        &self,
+        kind: PutKind,
+        key: T::Key,
+        value: T::Value,
+    ) -> Result<(), DatabaseError> {
+        let key = key.encode();
+        let value = value.compress();
+        let (operation, write_operation, flags) = kind.into_operation_and_flags();
+        self.execute_with_operation_metric::<T, _>(operation, Some(value.as_ref().len()), |tx| {
+            tx.put(self.get_dbi::<T>()?, key.as_ref(), value, flags).map_err(|e| {
+                DatabaseWriteError {
+                    info: e.into(),
+                    operation: write_operation,
+                    table_name: T::NAME,
+                    key: key.into(),
+                }
+                .into()
+            })
+        })
+    }
+}
+
 impl DbTxMut for Tx<RW> {
     type CursorMut<T: Table> = Cursor<RW, T>;
     type DupCursorMut<T: DupSort> = Cursor<RW, T>;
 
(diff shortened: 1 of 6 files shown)


crates/storage/db/src/metrics.rs
```diff
@@ -220,7 +222,8 @@ impl Operation {
     pub(crate) const fn as_str(&self) -> &'static str {
         match self {
             Self::Get => "get",
-            Self::Put => "put",
```
hai-rise, line 223: I wonder if this breaking change is a big deal? If it is, then `put` vs. `put-append` should be fine; or maybe count both as `put`? I did this as cursors do distinguish upsert and append already.

crates/storage/db/src/implementation/mdbx/tx.rs
```diff
@@ -340,28 +340,44 @@ impl<K: TransactionKind> DbTx for Tx<K> {
     }
 }
 
+impl Tx<RW> {
+    fn put<T: Table>(
```
mattsse, line 368: can we add some docs here

crates/storage/db/src/implementation/mdbx/tx.rs
```diff
@@ -340,28 +340,44 @@ impl<K: TransactionKind> DbTx for Tx<K> {
...
+
 impl DbTxMut for Tx<RW> {
     type CursorMut<T: Table> = Cursor<RW, T>;
     type DupCursorMut<T: DupSort> = Cursor<RW, T>;
 
     fn put<T: Table>(&self, key: T::Key, value: T::Value) -> Result<(), DatabaseError> {
-        let key = key.encode();
-        let value = value.compress();
-        self.execute_with_operation_metric::<T, _>(
-            Operation::Put,
```
mattsse, line 351: how is putupsert different from just put?

crates/storage/errors/src/db.rs
```diff
@@ -106,8 +106,10 @@ pub enum DatabaseWriteOperation {
     CursorInsert,
     /// Append duplicate cursor.
     CursorAppendDup,
-    /// Put.
-    Put,
+    /// Put upsert.
+    PutUpsert,
+    /// Put append.
+    PutAppend,
```
mattsse, line 112: could we add another enum for PutKind then this can replace the bool and we can do `kind.into()` instead

mattsse (changes requested):
some stlye nits, defer to @shekhirin on mdbx

crates/storage/db/src/metrics.rs
shekhirin: this change is fine

shekhirin (approved):
All of this makes sense, nice catch that we can append not just for cursor, but for simple `put` operations too.

Agreed with Matt's nits

crates/storage/db/src/implementation/mdbx/tx.rs
shekhirin: it's the same, just renamed for clarity. By default with flag `0`, put will do an upsert https://github.com/paradigmxyz/reth/blob/cff55122d210afca4b2db78fa5a1c8e251378a21/crates/storage/libmdbx-rs/mdbx-sys/libmdbx/mdbx.h#L1624-L1625

crates/storage/db/src/implementation/mdbx/tx.rs
hai-rise: Done :pray:.

crates/storage/errors/src/db.rs
hai-rise: Nice! Too many `Into`s are annoying, so I did a simple `TxKind::into_operation_and_flags` that covers everything.

mattsse (approved):

PR #18605, opened 2025-09-21
refactor(engine): separate concerns in on_new_payload

0xKarl98:
cc @yongkangc ,This closes issue #18448


0xKarl98:
As @yongkangc said , this would rather being handled internally , so close it now

PR #18608, opened 2025-09-22
test(engine): add new payload handling tests for canonical insertion and invalid ancestors

### LINKED ISSUE #18607: tests: add tests to prevent regression for new payload
### Describe the feature

before we refactor, we should add some tests to prevent any regression

there are currently some missing tests, and we would wanna cover these key scenarios: invalid ancestor rejection, buffering during backfill, successful insertion with and without canonicalization. These tests will serve as a safety net for the refactor.

### Additional context

_No response_

yongkangc:
closes: https://github.com/paradigmxyz/reth/issues/18607

Added Tests to make sure that the future refactor we do is correct as this is super critical:

- Canonical insertion: Validate proper handling of valid payloads that extend or reaffirm the canonical chain
- Invalid ancestors: Ensure payloads with invalid parent blocks are rejected with appropriate status
- Backfill buffering: Confirm payloads are buffered and reported as syncing during backfill operations
- Malformed payloads: Test Engine API compliance for invalid cases (e.g., latestValidHash set to None)

### DIFF
--- a/crates/engine/tree/src/tree/tests.rs
+++ b/crates/engine/tree/src/tree/tests.rs
@@ -1,12 +1,15 @@
 use super::*;
 use crate::persistence::PersistenceAction;
 use alloy_consensus::Header;
+use alloy_eips::eip1898::BlockWithParent;
 use alloy_primitives::{
     map::{HashMap, HashSet},
     Bytes, B256,
 };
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense

PR #18609, opened 2025-09-22
chore (refactor): cleaned up connection handling + handshake/snoop flow

onasilos:
fixed the bug and kept the same behavior, but with less allocations, no copy-paste code, proper protection against hanging connections, and structured logs. the handshake\_eth and snoop logic should be a lot cleaner now.


mattsse (commented):
this is just example code and this is fine

PR #18613, opened 2025-09-22
refactor(engine): simplify on_new_payload

yongkangc:
closes https://github.com/paradigmxyz/reth/issues/18448

where on_new_payload is one of the most important functions in reth and we want to simplify that.  Broke that down into smaller functions, where it can be tested.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -544,8 +544,9 @@ where
         let block_hash = num_hash.hash;
 
         // Check for invalid ancestors
-        if let Some(invalid_status) = self.check_invalid_ancestors(&payload)? {
-            return Ok(TreeOutcome::new(invalid_status));
+        if let Some(invalid) = self.find_invalid_ancestor(&payload) {
+            let status = self.handle_invalid_ancestor_payload(payload, invalid)?;
+            return Ok(TreeOutcome::new(status));
         }
 
         // record pre-execution phase duration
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -630,6 +575,72 @@ where
...
+                Ok(PayloadStatus::new(status, latest_valid_hash))
+            }
+            Err(error) => match error {
+                InsertPayloadError::Block(error) => Ok(self.on_insert_block_error(error)?),
+                InsertPayloadError::Payload(error) => {
+                    Ok(self.on_new_payload_error(error, parent_hash)?)
+                }
+            },
+        }
+    }
```
Rjected: maybe we can name this something like `try_insert_payload`? Given it can sometimes return `Syncing`, maybe we also want to add docs on when this returns syncing and when it returns valid

crates/engine/tree/src/tree/mod.rs
```diff
@@ -630,6 +575,72 @@ where
...
+            Ok(block) => {
+                if let Err(error) = self.buffer_block(block) {
+                    Ok(self.on_insert_block_error(error)?)
+                } else {
+                    Ok(PayloadStatus::from_status(PayloadStatusEnum::Syncing))
+                }
+            }
+            Err(error) => Ok(self.on_new_payload_error(error, parent_hash)?),
+        }
+    }
```
Rjected: along the same lines w.r.t. naming, maybe this should be `try_buffer_payload`

Rjected (changes requested):
nice, i just have a few nit comments about naming but like the refactors overall

crates/engine/tree/src/tree/mod.rs
yongkangc: thanks @Rjected  ! great suggestions

crates/engine/tree/src/tree/mod.rs
yongkangc: makes sense :)

mediocregopher (approved):
Much clearer, thank you! 👍

Rjected (approved):
lgtm!

PR #18614, opened 2025-09-22
chore(eth-wire): clarify docs and fix timeout_timer comment

radik878:
fuzz_roundtrip.rs - Default was deleted so needed to update comment
test_utils.rs - typo
pinger.rs - Update top-level Pinger doc to reflect current model: one ping → wait for pong → single timeout transition.
Fix incorrect timeout_timer field comment to indicate it tracks ping timeouts (not next ping).
Adjust on_pong doc to describe actual behavior (reset state and ping interval), removing mention of unsetting the sleep timer.


mattsse (commented):
these are fine

PR #18615, opened 2025-09-22
feat: Add building and publishing of *.deb packages

bakhtin:
**Refactor CI release pipeline and Makefile to support building and publishing of `*.deb` packages**
- Update Makefile to support building `deb` packages for RISC-V, aarch64, and amd64 architectures. `deb` packaging is handled by [`cargo-deb` crate](https://crates.io/crates/cargo-deb).
- Update `release.yml` pipeline to support building and publishing of `deb` packages, including reproducibly built ones

**Reproducible binary improvements**
- Adjust compiler options in Makefile for building reth binary reproducibly. I verified it locally with the old options and it wasn't reproducible. Adding `symbol-mangling-version=v0` fixed the issue.
- Add `reproducible` build profile
- Update Dockerfile for reproducible builds to allow passing Rust version to use in the base image. With that you no longer need to keep versions from the global workspace in sync with the Dockerfile.
- Update `reproducible-build.yml` pipeline to use the updated reproducible build command
- Do not attempt to build reproducible binaries inside the Docker container. Docker images with pre-installed dependencies are too hermetic and IMO we should allow slight variations between the build environments to prove that binaries can actually be built reproducibly.


mattsse (commented):
this is pretty cool and would be very useful

not a deb user/expert

perhaps @Rjected @onbjerg @mediocregopher know how to proceed here

.github/workflows/release-reproducible.yml
```diff
@@ -40,12 +40,20 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Extract Rust version from Cargo.toml
+        id: rust_version
+        run: |
+          RUST_VERSION=$(grep '^rust-version' Cargo.toml | sed 's/.*= *"\([^"]*\)".*/\1/' || echo "1")
+          echo "RUST_VERSION=$RUST_VERSION" >> $GITHUB_OUTPUT
```
onbjerg, line 47: nit: you can use `jq` here instead, I think it's more readable

```
cargo metadata | jq -r '.packages[] | select(.name == "reth") | .rust_version'
```

pkg/reth/debian/reth.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
onbjerg, line 1: let's move these files into a dir in the root of the project, e.g. `etc/` or a new folder named `pkg/`.

bin/reth/Cargo.toml
```diff
@@ -9,6 +9,20 @@ repository.workspace = true
 description = "Reth node implementation"
 default-run = "reth"
 
+[package.metadata.deb]
+maintainer = "Artyom Bakhtin <a@bakhtin.net>"
```
onbjerg: please put the reth team as a maintainer

Dockerfile.reproducible
```diff
@@ -1,17 +1,17 @@
-# Use the Rust 1.88 image based on Debian Bookworm
-FROM rust:1.88-bookworm AS builder
+ARG RUST_VERSION=1
```
onbjerg, line 1: will this not make it less reproducible? before we used a specific version (1.88), this one shifts over time as new stables are released

bakhtin: I updated CI pipeline to pass the same RUST_VERSION as defined in the root Cargo.toml. So this could only affect reproducibility if the Dockerfile is used outside of the CI pipeline run (e.g., building locally).

This change additionally removes the burden of constantly updating the Rust version in the Dockefile to match the project's root.

onbjerg (approved):
lgtm

PR #18616, opened 2025-09-22
docs: rm 8MB stack size comment in BlockingTaskPool

prestoalvarez:
### Description
- **What**: Explicitly set Rayon thread pool stack size to 8MB in `BlockingTaskPool::build`.
- **Why**: Documentation claimed 8MB, but code didn’t enforce it. This prevents unexpected stack overflows in CPU-bound tasks and aligns behavior with docs.
- **Change**: Add `.stack_size(8 * 1024 * 1024)` to the pool builder.
- **Impact**: More predictable stack for heavy tracing/CPU workloads; no API changes.

### DIFF
--- a/crates/tasks/src/pool.rs
+++ b/crates/tasks/src/pool.rs
@@ -72,10 +72,7 @@ impl BlockingTaskPool {
     /// Uses [`rayon::ThreadPoolBuilder::build`](rayon::ThreadPoolBuilder::build) defaults but
     /// increases the stack size to 8MB.
     pub fn build() -> Result<Self, rayon::ThreadPoolBuildError> {
-        Self::builder()
-            .stack_size(8 * 1024 * 1024)
-            .build()
-            .map(Self::new)
+        Self::builder().stack_size(8 * 1024 * 1024).build().map(Self::new)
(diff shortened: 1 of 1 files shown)


Rjected (changes requested):
IIRC we had a 8MB stack size in the past because an old version of revm would allocate too much during tracing, and we removed the 8MB bump because it was fixed in revm. So we should probably just remove the doc that mentions the 8MB bump instead, since we should be able to work within default stack size bounds

Rjected (approved):
lgtm thanks!

crates/tasks/src/pool.rs
```diff
@@ -69,8 +69,9 @@ impl BlockingTaskPool {
 
     /// Convenience function to build a new threadpool with the default configuration.
     ///
-    /// Uses [`rayon::ThreadPoolBuilder::build`](rayon::ThreadPoolBuilder::build) defaults but
-    /// increases the stack size to 8MB.
```
mattsse, line 73: right we had this pretty early on where the evm did exceed that

mattsse (approved):

PR #18617, opened 2025-09-22
fix(cli): bootnode default address

0xMelkor:
Running 
```bash
reth p2p bootnode
```

returns a parse error on the default listening address.

```bash
INFO Bootnode started with config: Command { addr: ":30301", gen_key: "", node_key: "", nat: Any, v5: false }
Error: invalid socket address syntax
```


Rjected (approved):
makes sense, thanks!

PR #18622, opened 2025-09-22
fix: implement missing else branch in Compact derive from_compact generation

avorylli:
Replace `todo!()` with proper implementation that reads length from flags and calls `from_compact_ident`, matching the logic used in the `is_compact` case.


mattsse (commented):
no need to touch this

PR #18626, opened 2025-09-23
chore(primitive-traits): remove redundant auto-trait bounds from FullNodePrimitives

MozirDmitriy:
Why: The auto-trait bounds (Send, Sync, Unpin, Clone, Default, Debug, PartialEq, Eq, 'static) were already enforced by NodePrimitives, and FullNodePrimitives requires Self: NodePrimitives<...>. Duplicating them in FullNodePrimitives and its blanket impl was unnecessary and inconsistent with other Full* traits (FullBlock, FullBlockBody, FullBlockHeader) that rely on their base traits’ bounds without repetition.
What: Removed duplicated bounds from the FullNodePrimitives trait and its blanket implementation, keeping only the NodePrimitives-related constraints. This preserves semantics, improves readability, and aligns with established style in the crate.

### DIFF
--- a/crates/primitives-traits/src/node.rs
+++ b/crates/primitives-traits/src/node.rs
@@ -35,15 +35,7 @@ where
             BlockBody: FullBlockBody<Transaction = Self::SignedTx>,
             SignedTx: FullSignedTx,
             Receipt: FullReceipt,
-        > + Send
-        + Sync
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I see

these are indeed redundant

PR #18628, opened 2025-09-23
fix: Prevent potential underflow in static file header healing

prestoalvarez:
### Description
- Replace a risky subtraction with saturating_sub in ensure_end_range_consistency to avoid potential u64 underflow when actual rows exceed expected rows.
- Ensures header pruning only occurs when expected_rows > actual_rows, preserving data integrity.
- Impact: Safer header healing; no behavior change in normal cases.


mattsse (approved):
this makes sense, although this should always be the case

PR #18632, opened 2025-09-23
feat(e2e): add beacon consensus handle to NodeClient

fgimenez:
this will allow us to write e2e tests for reth sdk instantiations that don't use engine API.

### DIFF
--- a/crates/e2e-test-utils/src/testsuite/mod.rs
+++ b/crates/e2e-test-utils/src/testsuite/mod.rs
@@ -16,27 +16,48 @@ pub mod setup;
 use crate::testsuite::setup::Setup;
 use alloy_provider::{Provider, ProviderBuilder};
 use alloy_rpc_types_engine::{ForkchoiceState, PayloadAttributes};
+use reth_engine_primitives::ConsensusEngineHandle;
 use reth_rpc_builder::auth::AuthServerHandle;
 use std::sync::Arc;
 use url::Url;
 
 /// Client handles for both regular RPC and Engine API endpoints
 #[derive(Clone)]
-pub struct NodeClient {
+pub struct NodeClient<Payload>
+where
+    Payload: PayloadTypes,
+{
     /// Regular JSON-RPC client
     pub rpc: HttpClient,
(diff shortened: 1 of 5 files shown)


crates/e2e-test-utils/src/testsuite/mod.rs
```diff
@@ -16,27 +16,48 @@ pub mod setup;
...
 use alloy_rpc_types_engine::{ForkchoiceState, PayloadAttributes};
+use reth_engine_primitives::ConsensusEngineHandle;
 use reth_rpc_builder::auth::AuthServerHandle;
 use std::sync::Arc;
 use url::Url;
 
 /// Client handles for both regular RPC and Engine API endpoints
 #[derive(Clone)]
-pub struct NodeClient {
+pub struct NodeClient<Payload>
```
mattsse, line 26: we could perhaps also make this generic over N: Nodetypes, perhaps this will be necessary eventually anyway

crates/e2e-test-utils/src/testsuite/setup.rs
```diff
@@ -241,8 +242,14 @@ where
                         .ok_or_else(|| eyre!("Failed to create HTTP RPC client for node"))?;
                     let auth = node.auth_server_handle();
                     let url = node.rpc_url();
-
-                    node_clients.push(crate::testsuite::NodeClient::new(rpc, auth, url));
+                    let beacon_handle = node.inner.add_ons_handle.beacon_engine_handle.clone();
```
mattsse: should we add a helper fn for this?

mattsse (approved):
makes sense

crates/e2e-test-utils/src/testsuite/mod.rs
fgimenez: good point, will handle in a follow-up

crates/e2e-test-utils/src/testsuite/setup.rs
fgimenez: sure done

Issue #18513, opened 2025-09-17
refactor(engine): simplify InvalidBlockWitnessHook::on_invalid_block for better testability


yongkangc:
Current implementation spans 180+ lines handling execution, state collection, proof generation, and witness assembly in a single function.

https://github.com/paradigmxyz/reth/blob/main/crates/engine/invalid-block-hooks/src/witness.rs#L141-L322

Problem:
- Difficult to test individual logic paths without full database setup
- High cognitive load for alt-L1 teams adopting Reth SDK
- Multiple nested data transformations obscure core witness generation logic

Proposed approach:
- Decompose into smaller functions with clear responsibilities
- Enable testing with mock state providers instead of requiring full StateProviderDatabase
- Each extracted component should be testable with minimal fixture setup

Expected: 
- Reduced complexity enabling unit tests for proof generation without executing blocks
- Simplified debugging when witness generation fails
- Clearer separation between Ethereum-specific and generic witness logic

0xKarl98:
Yeah ofc i can take it

yongkangc:
@Paulius0112 had a chat with and will be assigning @0xKarl98 

do add some tests to prevent regression as well, thanks!

PR #18636, opened 2025-09-23
refactor(engine): simplify InvalidBlockWitnessHook::on_invalid_block for better testability

0xKarl98:
This closes issue #18513


crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -4,19 +4,821 @@ use alloy_rpc_types_debug::ExecutionWitness;
...
+    pub original_trie_updates: Option<&'a TrieUpdates>,
+    pub re_executed_trie_updates: &'a TrieUpdates,
+}
+
+/// Trait for witness validation
+pub(crate) trait WitnessValidatorTrait {
+    /// Validates all aspects of witness generation
+    fn validate_all<N>(&self, params: ValidationParams<'_>) -> ValidationResult;
+}
+
```
yongkangc: do we really need new traits?

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -4,19 +4,821 @@ use alloy_rpc_types_debug::ExecutionWitness;
...
+    /// Creates a new state collector with collected data
+    #[cfg(test)]
+    pub(crate) const fn new(
+        hashed_state: HashedPostState,
+        codes: Vec<Bytes>,
+        state_preimages: Vec<Bytes>,
+    ) -> Self {
+        Self { hashed_state, codes, state_preimages }
+    }
+
```
yongkangc: we should not have this, make it inline and simplify

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -4,19 +4,821 @@ use alloy_rpc_types_debug::ExecutionWitness;
...
+impl WitnessGeneratorTrait for WitnessGenerator {
+    /// Generates an execution witness from collected state data
+    fn generate_execution_witness<SP>(
+        &self,
+        state_provider: Arc<SP>,
+        collected_state: &StateCollector,
+    ) -> Result<ExecutionWitness>
+    where
+        SP: StateProvider,
+    {
```
yongkangc: is there a need for traits here?

yongkangc (changes requested):
hey @0xKarl98 just did a quick look, and it seems like its quite complicated. 

I guess the main aim is to see if we can reduce loc + make it simpler 

- More intuitive abstractions
- fewer different code paths 
- Testability (simple code is also simple to test)

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Yeah  got it , will delete it and pass the param directly to the struct

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Will make the logic path more simpler

PR #18637, opened 2025-09-23
docs: update dashboard table and rpc urls

laibe:
Small change to docs:

* public base-sepolia is no longer maintained
* Fix description in metrics dashboard, OP node is full not archive.


mattsse (approved):

yash-atreya (approved):
lgtm!

PR #18642, opened 2025-09-23
chore: emit build-new-payload events in caller scope

SuperFluffy:
Adds the current span of the caller to `PayloadServiceCommand::BuildNewPayload` so that the payload builder service can emit all events inside that span.

As a follow-up one could also propagate that span to the job pushed into `payload_jobs` so that events that occur while processing the job can also be related back to the original call.

### DIFF
--- a/crates/payload/builder/src/service.rs
+++ b/crates/payload/builder/src/service.rs
@@ -131,7 +131,11 @@ impl<T: PayloadTypes> PayloadBuilderHandle<T> {
         attr: T::PayloadBuilderAttributes,
     ) -> Receiver<Result<PayloadId, PayloadBuilderError>> {
         let (tx, rx) = oneshot::channel();
-        let _ = self.to_service.send(PayloadServiceCommand::BuildNewPayload(attr, tx));
+        let _ = self.to_service.send(PayloadServiceCommand::BuildNewPayload(
+            attr,
+            tx,
+            tracing::Span::current(),
+        ));
         rx
     }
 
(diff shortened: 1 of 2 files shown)


crates/payload/builder/src/service.rs
```diff
@@ -426,33 +430,35 @@ where
...
-                                    this.metrics.inc_failed_jobs();
-                                    warn!(target: "payload_builder", %err, %id, "Failed to create payload builder job");
-                                    res = Err(err);
+                    PayloadServiceCommand::BuildNewPayload(attr, tx, span) => {
+                        span.in_scope(|| {
+                            let id = attr.payload_id();
+                            let mut res = Ok(id);
+
+                            if this.contains_payload(id) {
+                                info!(target: "payload_builder",%id, parent = %attr.parent(), "Payload job already in progress, ignoring.");
```
SuperFluffy, line 439: Flagging this: I changed this from `debug` to `info` level because this is valuable information for my caller. I figured that changing the event level here is less invasive compared to changing the channel return value.

I also suspect that well behaved code calling this should not fire new payload jobs in a hot loop so this won't lead to event spam.

crates/payload/builder/src/service.rs
```diff
@@ -426,33 +430,35 @@ where
...
-                                    new_job = true;
-                                    this.payload_jobs.push((job, id));
-                                    this.payload_events.send(Events::Attributes(attr.clone())).ok();
-                                }
-                                Err(err) => {
-                                    this.metrics.inc_failed_jobs();
-                                    warn!(target: "payload_builder", %err, %id, "Failed to create payload builder job");
-                                    res = Err(err);
+                    PayloadServiceCommand::BuildNewPayload(attr, tx, span) => {
+                        span.in_scope(|| {
```
SuperFluffy, line 434: Unfortunately this is causing even more right-shift. We can use https://docs.rs/tracing/0.1.41/tracing/span/struct.Span.html#method.enter instead, but see the caveats there.

SuperFluffy:
Note, a better way is probably changing `PayloadServiceCommand` into an inner and outer part, something like this:

```rust
struct Message {
    cause: tracing::Span,
    command: Command,
}

enum Command {
    DoThis,
    DoThat,
}
```

Then a handler could look like so establishing a follows-from  or parent relationship:

```rust
fn run_event_loop(self) {
    while let Some(msg) = self.mailbox.next().await {
        let cause = msg.cause;
        match msg.command {
            Command::DoThis(cmd) => self.do_this(cause, cmd),
            Command::DoThat(cmd) => self.do_that(cause, cmd),
        }
    }
}

#[instrument(skip_all, follows_from = [cause] )]
fn do_this(&self, cause: tracing::Span, cmd: DoThis) {
    todo!()
}
```

crates/payload/builder/src/service.rs
Rjected: yeah, let's not use `span.enter`

Rjected (commented):
along the same lines of the inner / outer suggestion, I think wrapping the enum in a struct containing the span and sending that in the channel makes sense. Maybe we could just add methods to the `PayloadBuilderService`, which take a span and handle the given message? For newpayload this would require also taking `new_job: &mut bool`

PR #18643, opened 2025-09-23
feat(bootnode): Persists the discovery secret key

0xMelkor:
Bootnode command improvement:
- [x] Restore/persist the discovery key
- [x] Removed unused cli args 
- [x] CLI naming consistency

### DIFF
--- a/crates/cli/commands/src/p2p/bootnode.rs
+++ b/crates/cli/commands/src/p2p/bootnode.rs
@@ -1,58 +1,30 @@
 //! Standalone bootnode command
 
 use clap::Parser;
-use reth_chainspec::{EthChainSpec, EthereumHardforks};
-use reth_cli::chainspec::ChainSpecParser;
-use reth_cli_util::get_secret_key;
+use reth_cli_util::{get_secret_key, load_secret_key::rng_secret_key};
 use reth_discv4::{DiscoveryUpdate, Discv4, Discv4Config};
 use reth_discv5::{discv5::Event, Config, Discv5};
 use reth_net_nat::NatResolver;
 use reth_network_peers::NodeRecord;
-use reth_node_core::{
-    args::DatadirArgs,
-    dirs::{ChainPath, DataDirPath, MaybePlatformPath},
-};
 use secp256k1::SecretKey;
-use std::{fmt, net::SocketAddr, path::PathBuf, sync::Arc};
+use std::{net::SocketAddr, path::PathBuf};
 use tokio::select;
 use tokio_stream::StreamExt;
 use tracing::info;
 
 /// Start a discovery only bootnode.
 #[derive(Parser, Debug)]
-pub struct Command<C: ChainSpecParser> {
+pub struct Command {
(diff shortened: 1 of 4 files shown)


crates/cli/commands/src/p2p/bootnode.rs
```diff
@@ -34,11 +65,16 @@ pub struct Command {
...
+impl<C> Command<C>
+where
+    C: ChainSpecParser,
+    C::ChainSpec: EthChainSpec + EthereumHardforks,
+{
     /// Execute the bootnode command.
     pub async fn execute(self) -> eyre::Result<()> {
-        info!("Bootnode started with config: {:?}", self);
-        let sk = reth_network::config::rng_secret_key();
+        info!("Bootnode started with config: {self}");
```
mattsse: this log will be massive, we should only include relevant info here

crates/cli/commands/src/p2p/bootnode.rs
```diff
@@ -1,29 +1,60 @@
...
-    pub gen_key: String,
+    /// The path to the data dir for all reth files and subdirectories.
+    ///
+    /// Defaults to the OS-specific data directory:
+    ///
+    /// - Linux: `$XDG_DATA_HOME/reth/` or `$HOME/.local/share/reth/`
+    /// - Windows: `{FOLDERID_RoamingAppData}/reth/`
+    /// - macOS: `$HOME/Library/Application Support/reth/`
+    #[arg(long, value_name = "DATA_DIR", verbatim_doc_comment, default_value_t)]
+    pub datadir: MaybePlatformPath<DataDirPath>,
```
mattsse: I dont think we should use these defaults here because we dont want to launch this bootnode with the same secret as the regular node?

crates/cli/commands/src/p2p/bootnode.rs
```diff
@@ -1,29 +1,60 @@
...
+    #[arg(
+        long,
+        value_name = "CHAIN_OR_PATH",
+        long_help = C::help_message(),
+        default_value = C::SUPPORTED_CHAINS[0],
+        default_value_if("dev", "true", "dev"),
+        value_parser = C::parser(),
+        required = false,
+    )]
+    pub chain: Arc<C::ChainSpec>,
```
mattsse: do we really need this?

or is the chainid sufficient here?

looks like we only need this to resolve the default datadir, which imo we shouldnt do

so I guess we should instead have a setting instead that provides the dir where the secret should be persisted?

mattsse (changes requested):
dont think we need all of this but I agree the current args arent great

can we instead focus on just accepting either file/key and an option to write it to some location, this seems easier imo

crates/cli/commands/src/p2p/bootnode.rs
0xMelkor: It was required to work with the datadir...I removed it

mattsse (approved):
ty, this is a lot simpler than he previous impl

PR #18648, opened 2025-09-23
fix(engine): remove redundant parent_to_child removal during eviction

Forostovec:
- Drop duplicate remove_from_parent call in insert_block eviction path; remove_block already handles it.
- No functional changes; micro-optimization and slight simplification.
- Tests and metrics unaffected; eviction remains FIFO via block_queue.

### DIFF
--- a/crates/engine/tree/src/tree/block_buffer.rs
+++ b/crates/engine/tree/src/tree/block_buffer.rs
@@ -74,9 +74,7 @@ impl<B: Block> BlockBuffer<B> {
         if self.block_queue.len() >= self.max_blocks {
             // Evict oldest block if limit is hit
             if let Some(evicted_hash) = self.block_queue.pop_front() {
-                if let Some(evicted_block) = self.remove_block(&evicted_hash) {
-                    self.remove_from_parent(evicted_block.parent_hash(), &evicted_hash);
-                }
+                self.remove_block(&evicted_hash);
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -74,9 +74,7 @@ impl<B: Block> BlockBuffer<B> {
         if self.block_queue.len() >= self.max_blocks {
             // Evict oldest block if limit is hit
             if let Some(evicted_hash) = self.block_queue.pop_front() {
-                if let Some(evicted_block) = self.remove_block(&evicted_hash) {
-                    self.remove_from_parent(evicted_block.parent_hash(), &evicted_hash);
-                }
+                self.remove_block(&evicted_hash);
```
mattsse, line 77: this is indeed redundant here

https://github.com/paradigmxyz/reth/blob/985671287cdadff047e957a0f8222cf911c566ca/crates/engine/tree/src/tree/block_buffer.rs#L152-L155

mattsse (approved):
lgtm

but would like a sanity check from @yongkangc or @shekhirin  perhaps

yongkangc (approved):
generally looks ok to me 

would be nice to maybe add test for parent-child cleanup is properly maintained during eviction during eviction?

PR #18650, opened 2025-09-23
Harden PayloadId against targeted collisions via process-scoped keyed hashing

sashaodessa:
### Description
- Switch `payload_id()` to SHA-256 keyed with a process-local secret, then truncate to 8 bytes.
- Keeps the API unchanged while making the 64-bit ID resilient to crafted collisions.
- Adds a per-process secret (random at runtime; deterministic in tests) and updates tests to assert stability and uniqueness properties.
- Minimal surface change: only `engine-primitives` crate; added `rand` as a workspace dependency.


mattsse (commented):
this follows a scheme that should be consistent across client so that the same attributes result in consistent ids

PR #18652, opened 2025-09-23
fix(engine): correct misleading test comments in cached_state.rs

crazykissshout:
Hmm, I think this is made by some copy-past mistakes.

Fixes misleading comments in storage cache tests that all said "check that the storage is empty" but were actually testing different behaviors.


mattsse (approved):
sure

PR #18655, opened 2025-09-23
chore: support custom transaction types in `EthTransactionValidator`

klkvr:
Adds a bitmap to `EthTransactionValidator` that allows to whitelist certain transaction types that will be allowed during validation, making it possible to fully reuse eth validator for chains with custom transaction types


mattsse (approved):
reasonable

PR #18656, opened 2025-09-24
fix(cli): replace unwrap with error propagation in merkle stage

wetkeyboard:
Replaced panic-prone `.unwrap()` calls with proper `?` error propagation


mattsse (approved):
appropriate

PR #18657, opened 2025-09-24
feat(rpc): merge the header not found and resource not found

wetkeyboard:
As https://github.com/ethereum-optimism/optimism/pull/11759 is released, let's merge the two errors into one, and add some unit tests for the resource not found error.


mattsse (approved):
makes sense now

PR #18659, opened 2025-09-24
refactor(engine): simplify validate_block_with_state

### LINKED ISSUE #18658: refactor: break down validate_block_with_state in payload_validator.rs
### Describe the feature

Thre is quite a few pieces of logic in https://github.com/paradigmxyz/reth/blob/584d7164fd8ac49599e3df99fff013531c5ed0f2/crates/engine/tree/src/tree/payload_validator.rs#L322

```
    /// - Consensus validation
    /// - Block execution
    /// - State root computation
    /// - Fork detection
```

What we need to do is to simplify this function. This allows us to inject dependencies and test much easier. Also easier to read and debug. 

Proposed Solution:
- Simplify the strategy decision
- move take out the double macros

This is a breaking one, so test is needed

### Additional context

_No response_

yongkangc:
closes: https://github.com/paradigmxyz/reth/issues/18658

the core changes are:
- reducing the complexity of how we choose our stateroot algo
- abstracting post validation stage 

Initially had a much grander idea of the change, but it made the pr too big and hard to review.

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -851,6 +851,8 @@ where
                         trie_input,
                         &self.config,
                     )
+                // if prefix sets are not empty, we spawn a task that exclusively handles cache
+                // prewarming for transaction execution
                 } else {
                     debug!(
                         target: "engine::tree",
@@ -858,7 +860,7 @@ where
                         "Disabling state root task due to non-empty prefix sets"
                     );
                     *strategy = StateRootStrategy::Parallel;
-                    self.payload_processor.spawn_execution_only(env, txs, provider_builder)
+                    self.payload_processor.spawn_cache_exclusive(env, txs, provider_builder)
                 };
 
                 // record prewarming initialization duration
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -46,20 +46,6 @@ use reth_trie_parallel::root::{ParallelStateRoot, ParallelStateRootError};
 use std::{collections::HashMap, sync::Arc, time::Instant};
 use tracing::{debug, debug_span, error, info, trace, warn};
 
-/// Arguments for spawning a payload processor task.
-struct PayloadProcessorArgs<'a, N: NodePrimitives, P> {
-    /// State provider builder for DB access
-    provider_builder: StateProviderBuilder<N, P>,
-    /// Type of persistence operation
-    persisting_kind: PersistingKind,
-    /// Parent block hash
-    parent_hash: B256,
-    /// Current tree state
-    state: &'a EngineApiTreeState<N>,
-    /// Block number and hash
-    block_num_hash: NumHash,
-}
-
 /// Context providing access to tree state during validation.
 ///
 /// This context is provided to the [`EngineValidator`] and includes the state of the tree's
@@ -425,13 +411,11 @@ where
         let mut handle = ensure_ok!(self.spawn_payload_processor(
             env.clone(),
             txs,
-            PayloadProcessorArgs {
                 provider_builder,
                 persisting_kind,
                 parent_hash,
-                state: ctx.state(),
+            ctx.state(),
                 block_num_hash,
-            },
             &mut strategy,
         ));
 
@@ -832,11 +816,16 @@ where
     }
 
     /// Spawns a payload processor task based on the state root strategy.
+    #[allow(clippy::too_many_arguments)]
     fn spawn_payload_processor(
         &mut self,
         env: ExecutionEnv<Evm>,
         txs: impl ExecutableTxIterator<Evm>,
-        args: PayloadProcessorArgs<'_, N, P>,
+        provider_builder: StateProviderBuilder<N, P>,
+        persisting_kind: PersistingKind,
+        parent_hash: B256,
+        state: &EngineApiTreeState<N>,
+        block_num_hash: NumHash,
         strategy: &mut StateRootStrategy,
     ) -> Result<
         PayloadHandle<impl ExecutableTxFor<Evm>, impl core::error::Error + Send + Sync + 'static>,
@@ -853,10 +842,10 @@ where
                 // Compute trie input
                 let trie_input_start = Instant::now();
                 let trie_input = self.compute_trie_input(
-                    args.persisting_kind,
+                    persisting_kind,
                     consistent_view.provider_ro()?,
-                    args.parent_hash,
-                    args.state,
+                    parent_hash,
+                    state,
                     allocated_trie_input,
                 )?;
 
@@ -872,7 +861,7 @@ where
                     self.payload_processor.spawn(
                         env,
                         txs,
-                        args.provider_builder,
+                        provider_builder,
                         consistent_view,
                         trie_input,
                         &self.config,
@@ -882,11 +871,11 @@ where
                 } else {
                     debug!(
                         target: "engine::tree",
-                        block=?args.block_num_hash,
+                        block=?block_num_hash,
                         "Disabling state root task due to non-empty prefix sets"
                     );
                     *strategy = StateRootStrategy::Parallel;
-                    self.payload_processor.spawn_cache_exclusive(env, txs, args.provider_builder)
+                    self.payload_processor.spawn_cache_exclusive(env, txs, provider_builder)
                 };
 
                 // record prewarming initialization duration
@@ -900,7 +889,7 @@ where
             StateRootStrategy::Parallel | StateRootStrategy::Synchronous => {
                 let start = Instant::now();
                 let handle =
-                    self.payload_processor.spawn_cache_exclusive(env, txs, args.provider_builder);
+                    self.payload_processor.spawn_cache_exclusive(env, txs, provider_builder);
 
                 // Record prewarming initialization duration
                 self.metrics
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -374,105 +389,35 @@ where
...
 
         // use prewarming background task
         let txs = self.tx_iterator_for(&input)?;
-        let mut handle = if use_state_root_task {
-            // use background tasks for state root calc
-            let consistent_view =
-                ensure_ok!(ConsistentDbView::new_with_latest_tip(self.provider.clone()));
-
-            // get allocated trie input if it exists
-            let allocated_trie_input = self.payload_processor.take_trie_input();
```
yongkangc, line 422: the logic here for deciding the algo was convoluted, so would need some sanity check here @mediocregopher @Rjected @shekhirin to see if the simplified logic below matches

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -900,6 +953,51 @@ where
...
     }
 
+    /// Determines the state root computation strategy based on persistence state and configuration.
+    fn plan_state_root_computation<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
+        &self,
+        input: &BlockOrPayload<T>,
+        ctx: &TreeCtx<'_, N>,
+    ) -> StateRootPlan {
+        // We only run the parallel state root if we are not currently persisting any blocks or
+        // persisting blocks that are all ancestors of the one we are executing.
```
yongkangc, line 984: this is where we do the planning logic rn @mediocregopher @Rjected @shekhirin

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -343,6 +358,20 @@ where
             };
         }
 
+        /// A helper macro for handling errors after the input has been converted to a block
```
yongkangc, line 347: this was moved up and renamed. not sure if we should keep the macro here but felt that it made sense without having to create helper functions

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -900,6 +954,51 @@ where
         Ok(None)
     }
 
+    /// Determines the state root computation strategy based on persistence state and configuration.
+    fn plan_state_root_computation<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
+        &self,
+        input: &BlockOrPayload<T>,
+        ctx: &TreeCtx<'_, N>,
+    ) -> StateRootPlan {
```
mattsse, line 982: I think this will be a lot easier to modify later

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -374,105 +389,35 @@ where
...
-        let run_parallel_state_root =
-            persisting_kind.can_run_parallel_state_root() && !self.config.state_root_fallback();
-
-        // Use state root task only if:
-        // 1. No persistence is in progress
-        // 2. Config allows it
-        // 3. No ancestors with missing trie updates. If any exist, it will mean that every state
-        //    root task proof calculation will include a lot of unrelated paths in the prefix sets.
-        //    It's cheaper to run a parallel state root that does one walk over trie tables while
-        //    accounting for the prefix sets.
```
mattsse, line 397: can we keep those docs?
imo these are quite helpful

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -848,6 +758,150 @@ where
...
+        // record post-execution validation duration
+        self.metrics
+            .block_validation
+            .post_execution_validation_duration
+            .record(start.elapsed().as_secs_f64());
+
+        Ok(hashed_state)
+    }
+
+    /// Spawns a payload processor task based on the state root strategy.
```
mattsse, line 818: could we add some additional docs here?

mattsse (commented):
imo this a lot better and easier to follow now,

I believe this will also make it easier to refactor again once we got rid of the db edge cases via trieupdates

logic wise I believe this is correct, but want a few more eyes on this @shekhirin @Rjected @mediocregopher

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: will revert these 👍🏻

yongkangc:
@mattsse just addressed this comments https://github.com/paradigmxyz/reth/pull/18659#pullrequestreview-3266462237

and also bumped it up to 2024 (fixed some compile errrs)

crates/engine/tree/src/tree/tests.rs
```diff
@@ -338,6 +347,205 @@ impl TestHarness {
...
+
+    /// Create valid block
+    fn create_valid_block(&mut self, parent_hash: B256) -> RecoveredBlock<Block> {
+        let block = self.builder.generate_random_block(1, parent_hash).into_block();
+        block.seal_slow().try_recover().unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+enum ConsensusInvalidType {
```
mediocregopher: I don't think this type is actually used, and it's just a single-variant enum

crates/engine/tree/src/tree/tests.rs
```diff
@@ -1146,6 +1360,237 @@ fn test_on_new_payload_malformed_payload() {
...
+
+    // Create valid block
+    let mut block_factory = TestBlockFactory::new(MAINNET.as_ref().clone());
+    let genesis_hash = MAINNET.genesis_hash();
+    let valid_block = block_factory.create_valid_block(genesis_hash);
+
+    // Call validate_block_with_state directly
+    let _result = test_harness.validate_block_direct(valid_block);
+
+    // Test passes if `Synchronous` strategy executes without panics during persistence
```
mediocregopher: Is this right? we expect it to panic vs returning an error?

mediocregopher (changes requested):
Some small things on the new tests, but overall the logic looks right to me. We should definitely run this on a mainnet dev node for a few days and then do a `repair-trie` to check if there were any reorg issues.

crates/engine/tree/src/tree/tests.rs
yongkangc: thanks for the note, cleaned that up

yongkangc:
https://github.com/paradigmxyz/reth/pull/18659#pullrequestreview-3271377830

also running this on reth3

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -900,6 +966,58 @@ where
         Ok(None)
     }
 
+    /// Determines the state root computation strategy based on persistence state and configuration.
+    fn plan_state_root_computation<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
```
mediocregopher, line 978: Just want to say I'm a big fan of splitting complex decision logic out from the execution of those decisions like this, with some data structure to encode the decision. Makes it much easier to follow the code and test it 💯

mediocregopher (approved):
This looks right to me 👍

crates/engine/tree/src/tree/payload_validator.rs
shekhirin: can probably unify them into one `macro_rules` declaration that has two variants: with one and two expressions

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -500,61 +445,19 @@ where
...
 
         let root_time = Instant::now();
 
         let mut maybe_state_root = None;
 
-        if run_parallel_state_root {
-            // if we new payload extends the current canonical change we attempt to use the
-            // background task or try to compute it in parallel
-            if use_state_root_task {
+        match strategy {
```
shekhirin, line 459: I like this a lot

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -848,6 +758,162 @@ where
...
+                    )
+                // if prefix sets are not empty, we spawn a task that exclusively handles cache
+                // prewarming for transaction execution
+                } else {
+                    debug!(
+                        target: "engine::tree",
+                        block=?block_num_hash,
+                        "Disabling state root task due to non-empty prefix sets"
+                    );
+                    *strategy = StateRootStrategy::Parallel;
```
shekhirin: this feels a bit weird, and I'd agree with copilot that it may be unexpected, would be nice to refactor this

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: thanks, same here esp for someone without context 🫡

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: thanks for the review @shekhirin will refactor this out

yongkangc: made a simple fix by returning the strategy as well instead of mutating it here:

[`3c735a3` (#18659)](https://github.com/paradigmxyz/reth/pull/18659/commits/3c735a3d6908405c0e1edc8e1314e5ef5dcc4d7f)

mediocregopher (approved):

mattsse (approved):

shekhirin (approved):

PR #18660, opened 2025-09-24
feat(net): implemnt the spam detection for ETH requests

wetkeyboard:
Implements spam detection for ETH request handlers, to protect against malicious peers sending invalid or excessive
requests. 

This resolves the TODO comment in `EthRequestHandler` that mentioned the `peers` field should be used to
report spammers. Detect on peer request for empty reqeust or excessive requests.

### DIFF
--- a/crates/net/network/src/eth_requests.rs
+++ b/crates/net/network/src/eth_requests.rs
@@ -15,6 +15,7 @@ use reth_eth_wire::{
 use reth_network_api::test_utils::PeersHandle;
 use reth_network_p2p::error::RequestResult;
 use reth_network_peers::PeerId;
+use reth_network_types::ReputationChangeKind;
 use reth_primitives_traits::Block;
 use reth_storage_api::{BlockReader, HeaderProvider};
 use std::{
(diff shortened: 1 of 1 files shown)


crates/net/network/src/eth_requests.rs
```diff
@@ -147,24 +146,94 @@ where
...
+        // Empty request - asking for 0 headers
+        if request.limit == 0 {
+            self.report_spam(peer_id, ReputationChangeKind::BadMessage);
+            return true;
+        }
+
+        // Excessive request - asking for too many headers
+        if request.limit as usize > MAX_HEADERS_SERVE {
+            self.report_spam(peer_id, ReputationChangeKind::BadMessage);
+            return true;
```
mattsse, line 165: this could backfire, because this is just a soft limit and we're supposed to serve up to this headers

mattsse (commented):
spam detections is hard because this is easily defeated, e.g. just request more with more messages, so this is too simplistic,
instead we should figure out if we can track a peer based req/s metric, however this might also not be exactly what we want, e.g. if you have a network with just trusted peers

PR #18664, opened 2025-09-24
docs(db): correct misleading test comments in post_state.rs

crazykissshout:
Similar to https://github.com/paradigmxyz/reth/pull/18652

### DIFF
--- a/crates/trie/db/tests/post_state.rs
+++ b/crates/trie/db/tests/post_state.rs
@@ -227,7 +227,7 @@ fn storage_is_empty() {
         (0..10).map(|key| (B256::with_last_byte(key), U256::from(key))).collect::<BTreeMap<_, _>>();
     db.update(|tx| {
         for (slot, value) in &db_storage {
-            // insert zero value accounts to the database
+            // insert non-zero storage entries to the database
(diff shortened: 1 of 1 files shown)


crates/trie/db/tests/post_state.rs
```diff
@@ -227,7 +227,7 @@ fn storage_is_empty() {
         (0..10).map(|key| (B256::with_last_byte(key), U256::from(key))).collect::<BTreeMap<_, _>>();
     db.update(|tx| {
         for (slot, value) in &db_storage {
-            // insert zero value accounts to the database
+            // insert non-zero storage entries to the database
```
yongkangc:
```suggestion
            // insert storage entries to the database
```

A zero-valued storage entry ends up in db_storage.

The db_storage map is built with `(0..10).map(|key| (B256::with_last_byte(key), U256::from(key)))`. 

So the first pair inserted is `(B256::with_last_byte(0), U256::from(0))`

yongkangc (approved):
makes sense except for 1

crates/trie/db/tests/post_state.rs
crazykissshout: Nice catch!

PR #18665, opened 2025-09-24
Precise ETH parsing: replace f64 with safe integer-based decimal parser

prestoalvarez:
- Motivation
  - Avoids floating‑point rounding errors.
  - Rejects NaN/∞ and malformed inputs explicitly.
  - Prevents silent overflows when scaling to wei.

- What changed
  - Rewrote parse_ether_value to parse decimal strings into wei using pure integer math.
  - Supports up to 18 fractional digits; right‑pads shorter fractions with zeros.
  - Validates input (non‑empty, non‑negative, digits only, ≤18 decimals).
  - Overflow‑safe via checked operations.

- Behavior notes
  - Inputs like ".5" are accepted (interpreted as 0.5 ETH).
  - More strict than before: rejects scientific notation, extra decimals, and non‑digit chars.

### DIFF
--- a/crates/cli/util/src/parsers.rs
+++ b/crates/cli/util/src/parsers.rs
@@ -148,9 +148,7 @@ pub fn parse_ether_value(value: &str) -> eyre::Result<u128> {
         frac.parse::<u128>()?
     };
 
-    int_wei
-        .checked_add(frac_wei)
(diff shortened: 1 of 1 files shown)


mattsse (commented):
what we have is sufficient

PR #18669, opened 2025-09-24
fix(rpc/engine): check osaka in getBlobsV1

jsvisa:
Similar to https://github.com/ethereum/go-ethereum/pull/32731, add a check the activation of osaka in getBlobsV1 request

closes #18668

### DIFF
--- a/crates/rpc/rpc-engine-api/src/engine_api.rs
+++ b/crates/rpc/rpc-engine-api/src/engine_api.rs
@@ -753,10 +753,6 @@ where
         &self,
         versioned_hashes: Vec<B256>,
     ) -> EngineApiResult<Vec<Option<BlobAndProofV1>>> {
-        if versioned_hashes.len() > MAX_BLOB_LIMIT {
-            return Err(EngineApiError::BlobRequestTooLarge { len: versioned_hashes.len() })
-        }
-
         // Check if Osaka fork is active - if so, return unsupported fork error
         // as per the engine API spec requirement
         let best_block_number = self
@@ -770,6 +766,10 @@ where
                     reth_payload_primitives::EngineObjectValidationError::UnsupportedFork,
                 ));
             }
+        }
+
+        if versioned_hashes.len() > MAX_BLOB_LIMIT {
+            return Err(EngineApiError::BlobRequestTooLarge { len: versioned_hashes.len() })
         }
 
         self.inner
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -753,6 +753,21 @@ where
...
     ) -> EngineApiResult<Vec<Option<BlobAndProofV1>>> {
+        // Check if Osaka fork is active - if so, return unsupported fork error
+        // as per the engine API spec requirement
+        let best_block_number = self
+            .inner
+            .provider
+            .best_block_number()
+            .map_err(|err| EngineApiError::Internal(Box::new(err)))?;
+        if let Ok(Some(header)) = self.inner.provider.header_by_number(best_block_number) {
+            if self.inner.chain_spec.is_osaka_active_at_timestamp(header.timestamp()) {
```
mattsse: we could probably use the system time here instead

we could also change the provider generic to

https://github.com/paradigmxyz/reth/blob/main/crates/storage/storage-api/src/block.rs#L275-L275

then we can use `latest_header` here

I think this should be more efficient because latest header is always tracked in memory

jsvisa: sorry, I wasted some time digging into the trait, ptal.

crates/storage/storage-api/src/block.rs
```diff
@@ -372,6 +372,23 @@ pub trait BlockReaderIdExt: BlockReader + ReceiptProviderIdExt {
     fn header_by_id(&self, id: BlockId) -> ProviderResult<Option<Self::Header>>;
 }
 
+impl<T: BlockReaderIdExt> BlockReaderIdExt for Arc<T> {
```
mattsse: do we even need this here?

jsvisa: sorry, added by mistake

crates/rpc/rpc-engine-api/src/engine_api.rs
jsvisa: Encountered some issues with `BlockReaderIdExt`, so temporarily using system time here

crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -788,6 +800,15 @@ where
         &self,
         versioned_hashes: Vec<B256>,
     ) -> EngineApiResult<Option<Vec<BlobAndProofV2>>> {
+        // Check if Osaka fork is active
+        let current_timestamp =
+            SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs();
```
mattsse, line 805: this makes sense imo, since this endpoint is only useful for the live chain we can use the system timestamp here

mattsse (approved):
lgtm, easier than fetching the block indeed

PR #18673, opened 2025-09-24
Harden JWT generation script: strict perms, secure umask, safer bash defaults

sashaodessa:
### Description
Tighten security of the JWT secret generation script to prevent accidental disclosure via permissive filesystem permissions and improve script robustness.

### Changes
- Add `set -euo pipefail` for safer bash defaults
- Create `jwttoken/` with `700` permissions
- Write `jwt.hex` with `600` permissions
- Set `umask 077` before file creation
- Remove unnecessary `tee`; write directly via `>`
- Keep comments concise and in English

### Rationale
Ensure the JWT secret is not group/world-readable and reduce risks from shell pitfalls.

### Impact
- No behavior change to consumers; only hardened defaults
- Backwards-compatible


mattsse (commented):
this script is sufficient

PR #18674, opened 2025-09-24
refactor: simplify `EthApiSpeс` trait

### LINKED ISSUE #18533: Deduplicate methods and associated types of `EthApiSpec` trait
### Describe the feature

`EthApiSpec` trait requires implementers to define `Transaction` and `Rpc` ATs and `signers(&self) -> &SignersForApi<Self>` helpers

It seems that `Transaction` AT is just primitive transaction type which can already be accessed through `RpcNodeCore`
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/rpc/rpc/src/eth/helpers/spec.rs#L16
`Rpc` AT is already present on `EthApiTypes` trait:
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/rpc/rpc-eth-api/src/types.rs#L31-L32
and `signers` getter is present on `EthTransactions` trait
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L63-L64

We should evaluate whether all parts of this trait are useful or we can remove them, while also potentially merging this trait with some of the other ones

### Additional context

_No response_

klkvr:
Closes https://github.com/paradigmxyz/reth/issues/18533

Gets rid of redundant methods and associated types

### DIFF
--- a/crates/optimism/rpc/src/eth/mod.rs
+++ b/crates/optimism/rpc/src/eth/mod.rs
@@ -18,8 +18,9 @@ use eyre::WrapErr;
 use op_alloy_network::Optimism;
 pub use receipt::{OpReceiptBuilder, OpReceiptFieldsBuilder};
 use reqwest::Url;
+use reth_chainspec::{EthereumHardforks, Hardforks};
 use reth_evm::ConfigureEvm;
-use reth_node_api::{FullNodeComponents, FullNodeTypes, HeaderTy};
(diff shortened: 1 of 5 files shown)


mattsse (approved):
makes sense

PR #18678, opened 2025-09-24
Fix JWT claims: set iat=now and exp=now+60 to match spec

viktorking7:
### PR Description
- Align JWT claim semantics with the Engine API: `iat` is now the current timestamp and `exp` is explicitly set to 60 seconds later.
- Prevents tokens from being “not yet valid” and enforces the intended 60s validity window.

- Changes:
  - Update `secret_to_bearer_header` in `crates/rpc/rpc-layer/src/auth_client_layer.rs`.
  - Add a brief comment explaining the timing semantics.


mattsse (commented):
this isn't right because we only need to set iat

> Required: iat (issued-at) claim. The execution layer client SHOULD only accept iat timestamps which are within +-60 seconds from the current time.

PR #18679, opened 2025-09-24
feat: support empty `SUPPORTED_CHAINS` for `ChainSpecParser`

klkvr:
Adds a way to define parser that doesn't have a default value that might mess with `global = true`


mattsse (approved):
cool

PR #18681, opened 2025-09-24
feat(storage): write headers and transactions only to static files

### LINKED ISSUE #18683: Write headers and transaction only to static files
### Describe the feature

We already write them to both database and static files, and then in provider check both locations to determine where to read from. We should simplify this by only writing to static files.

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/18683
Closes https://github.com/paradigmxyz/reth/issues/11524

Removes the `StorageLocation` enum that was used to tell the provider where to write the data. Headers and transactions now are always written to static files only.

This doesn't break existing nodes, because on startup we always copy to static files https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/node/builder/src/launch/engine.rs#L159-L160

---

We need to figure out what to do with receipts pruning. We support `receipts_log_filter` option that can prune everything but certain receipts that match the log filter, and use it for NOT pruning the deposit contract receipts. This type of pruning can't be done with static files, as they don't support deletions in the middle of the file.

For now, I'm still always writing receipts to the database, and to static files only if pruning is not enabled.


mediocregopher (approved):
It's a lot of changes but 80% looks like removing the StorageLocation argument from everything so pretty straightforward. LGTM

crates/prune/prune/src/segments/mod.rs
```diff
@@ -190,7 +191,7 @@ mod tests {
         let provider_rw = factory.provider_rw().expect("failed to get provider_rw");
         for block in &blocks {
             provider_rw
-                .insert_historical_block(
+                .insert_block(
```
joshieDo, line 194: nice

crates/storage/provider/src/test_utils/blocks.rs
```diff
@@ -85,7 +85,7 @@ pub(crate) static TEST_BLOCK: LazyLock<SealedBlock<reth_ethereum_primitives::Blo
                     )
                     .into(),
                     difficulty: U256::from(131_072),
-                    number: 1,
```
joshieDo, line 88: why?

crates/static-file/static-file/src/static_file_producer.rs
```diff
@@ -289,6 +289,7 @@ mod tests {
             .expect("get static file writer for headers");
         static_file_writer.prune_headers(blocks.len() as u64).unwrap();
         static_file_writer.commit().expect("prune headers");
+        drop(static_file_writer);
```
joshieDo, line 292: why?

crates/stages/stages/src/stages/hashing_account.rs
```diff
@@ -453,7 +453,7 @@ mod tests {
                 let provider = self.db.factory.database_provider_rw()?;
                 let res = Ok(AccountHashingStage::seed(
                     &provider,
-                    SeedOpts { blocks: 1..=input.target(), accounts: 10, txs: 0..3 },
+                    SeedOpts { blocks: 0..=input.target(), accounts: 10, txs: 0..3 },
```
joshieDo, line 456: why? (probably related to the other one)

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1809,15 +1771,13 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
         //
         // We are writing to database if requested or if there's any kind of receipt pruning
         // configured
-        let mut receipts_cursor = (write_receipts_to.database() || has_receipts_pruning)
-            .then(|| self.tx.cursor_write::<tables::Receipts<Self::Receipt>>())
-            .transpose()?;
+        let mut receipts_cursor = self.tx.cursor_write::<tables::Receipts<Self::Receipt>>()?;
```
joshieDo, line 1824: is this correct? what if `has_receipts_pruning` is `false`? 

edit: went back an re-read the PR description, got it

crates/static-file/static-file/src/static_file_producer.rs
joshieDo: got it, can we just add a simple doc

shekhirin: because here https://github.com/paradigmxyz/reth/blob/4905d5226adc1affa154b0cedfbe6c9a06115c81/crates/stages/stages/src/test_utils/test_db.rs#L73 it commits both static files and db tx, and if we're still holding into the static file writer, it will deadlock

shekhirin: I think ideally we shouldn't get `static_file_writer` for one segment, and instead just use the `provider` for everything

crates/stages/stages/src/stages/hashing_account.rs
shekhirin: can't write to static files starting from 1

crates/storage/provider/src/test_utils/blocks.rs
shekhirin: https://github.com/paradigmxyz/reth/blob/4905d5226adc1affa154b0cedfbe6c9a06115c81/crates/storage/provider/src/providers/database/mod.rs#L685-L692

with `number: 0` it fails
```
assertion failed: `Err(UnexpectedStaticFileBlockNumber(Headers, 1, 0))` does not match `Ok(_)`
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -800,9 +760,9 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> DatabaseProvider<TX, N> {
 }
 
 impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> DatabaseProvider<TX, N> {
-    /// Commit database transaction.
-    pub fn commit(self) -> ProviderResult<bool> {
-        Ok(self.tx.commit()?)
+    /// Commit database transaction and static file if it exists.
+    pub fn commit(self) -> ProviderResult<()> {
+        UnifiedStorageWriter::commit(self)
```
joshieDo: unsure here, if this is called after/during an unwind this would technically be wrong right? `UnifiedStorageWriter::commit_unwind` exists

mattsse (approved):
makes sense,

lgtm!

PR #18686, opened 2025-09-24
fix(engine): align sparse trie removal log target with engine::root::sparse

GarmashAlex:
Replace tracing target "trie::sparse" with "engine::root::sparse" in the engine-layer sparse trie task (account removal block) to maintain logging taxonomy consistency. Engine orchestration and its detailed sparse-trie work use engine::root and engine::root::sparse, while trie library code uses trie::sparse. This change keeps engine logs grouped correctly and avoids mixing with library-level logs. No functional behavior changes.


Rjected (approved):
makes sense

PR #18687, opened 2025-09-24
docs(storage): correct misleading comments in blocks.rs

crazykissshout:
Similar to https://github.com/paradigmxyz/reth/pull/18652


mattsse (commented):
this is better but not worth a pr

PR #18689, opened 2025-09-24
feat(flashblocks): prefill cache on canonical tip updates

### LINKED ISSUE #18680: preload latest state into flashblock cached reads
### Describe the feature

right now we initialize a new empty cache

https://github.com/paradigmxyz/reth/blob/a80f0c83a313cf44906e874a1dbc257303d16d2c/crates/optimism/flashblocks/src/service.rs#L43-L47

when a new parent head is used

instead we can prefill the cache on_tip:

https://github.com/paradigmxyz/reth/blob/a80f0c83a313cf44906e874a1dbc257303d16d2c/crates/optimism/flashblocks/src/service.rs#L132-L136

like we do here:

https://github.com/paradigmxyz/reth/blob/main/crates/payload/basic/src/lib.rs#L188-L204

### Additional context

assigning @Peponks9

Peponks9:
This PR implements cache prefilling when a new canonical tip is set (`on_tip` event) instead of starting

Closes #18680

### DIFF
--- a/bin/reth-bench/src/valid_payload.rs
+++ b/bin/reth-bench/src/valid_payload.rs
@@ -68,7 +68,7 @@ where
             if status.is_syncing() {
                 return Err(alloy_json_rpc::RpcError::UnsupportedFeature(
                     "invalid range: no canonical state found for parent of requested block",
-                ))
+                ));
             }
             status =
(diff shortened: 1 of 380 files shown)


mattsse (dismissed):
clean

mattsse:
something went when formatting here

PR #18692, opened 2025-09-25
chore: bump rust to edition 2024

yongkangc:
closes: https://github.com/paradigmxyz/reth/issues/18572

### DIFF
--- a/crates/cli/commands/src/import_core.rs
+++ b/crates/cli/commands/src/import_core.rs
@@ -192,7 +192,7 @@ pub fn build_import_pipeline_impl<N, C, E>(
     static_file_producer: StaticFileProducer<ProviderFactory<N>>,
     disable_exec: bool,
     evm_config: E,
-) -> eyre::Result<(Pipeline<N>, impl futures::Stream<Item = NodeEvent<N::Primitives>>)>
+) -> eyre::Result<(Pipeline<N>, impl futures::Stream<Item = NodeEvent<N::Primitives>> + use<N, C, E>)>
 where
     N: ProviderNodeTypes,
     C: FullConsensus<N::Primitives, Error = reth_consensus::ConsensusError> + 'static,
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2037,28 +2037,28 @@ where
     ) -> Option<B256> {
         let sync_target_state = self.state.forkchoice_state_tracker.sync_target_state();
 
-        // Compute the block number to check: prioritize downloaded finalized block,
-        // then buffered finalized block, otherwise use target block number
-        let block_num_to_check = if let (Some(downloaded_block), Some(ref state)) =
-            (downloaded_block, sync_target_state) &&
-            downloaded_block.hash == state.finalized_block_hash
-        {
-            // we downloaded the finalized block
-            downloaded_block.number
-        } else if let Some(buffered_finalized) = sync_target_state
+        // check if the downloaded block is the tracked finalized block
+        let mut exceeds_backfill_threshold = if let Some(buffered_finalized) = sync_target_state
             .as_ref()
             .and_then(|state| self.state.buffer.block(&state.finalized_block_hash))
         {
-            // we have buffered the finalized block
-            buffered_finalized.number()
+            // if we have buffered the finalized block, we should check how far
+            // we're off
+            self.exceeds_backfill_run_threshold(canonical_tip_num, buffered_finalized.number())
         } else {
-            // use the target block number
-            target_block_number
+            // check if the distance exceeds the threshold for backfill sync
+            self.exceeds_backfill_run_threshold(canonical_tip_num, target_block_number)
         };
 
-        // check if the distance exceeds the threshold for backfill sync
-        let exceeds_backfill_threshold =
-            self.exceeds_backfill_run_threshold(canonical_tip_num, block_num_to_check);
+        // If this is invoked after we downloaded a block we can check if this block is the
+        // finalized block
+        if let (Some(downloaded_block), Some(ref state)) = (downloaded_block, sync_target_state) &&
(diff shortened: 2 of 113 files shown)


crates/cli/commands/src/import_core.rs
```diff
@@ -192,7 +192,7 @@ pub fn build_import_pipeline_impl<N, C, E>(
     static_file_producer: StaticFileProducer<ProviderFactory<N>>,
     disable_exec: bool,
     evm_config: E,
-) -> eyre::Result<(Pipeline<N>, impl futures::Stream<Item = NodeEvent<N::Primitives>>)>
+) -> eyre::Result<(Pipeline<N>, impl futures::Stream<Item = NodeEvent<N::Primitives>> + use<N, C, E>)>
```
mattsse, line 195: ah I see, we can use `use` now

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2038,7 +2038,13 @@ where
...
-        let mut exceeds_backfill_threshold = if let Some(buffered_finalized) = sync_target_state
+        let exceeds_backfill_threshold = if let (Some(downloaded_block), Some(ref state)) =
+            (downloaded_block, sync_target_state) &&
+            downloaded_block.hash == state.finalized_block_hash
+        {
+            // we downloaded the finalized block and can now check how far we're off
+            self.exceeds_backfill_run_threshold(canonical_tip_num, downloaded_block.number)
+        } else if let Some(buffered_finalized) = sync_target_state
             .as_ref()
             .and_then(|state| self.state.buffer.block(&state.finalized_block_hash))
```
mattsse: this is now only in the else branch,

not 100% sure this is equivalent

crates/transaction-pool/src/test_utils/mock.rs
```diff
@@ -53,21 +55,54 @@ pub fn mock_tx_pool() -> MockTxPool {
 }
 
 /// Sets the value for the field
+#[allow(unused_macros)]
 macro_rules! set_value {
-    ($this:ident => $field:ident) => {
+    // For mutable references
+    (&mut $this:expr => $field:ident) => {{
```
mattsse: why do we need to change this for 24?

just for clippy?

mattsse (commented):
only have two qs

crates/engine/tree/src/tree/mod.rs
yongkangc: yeah you are right, let me revert 🙏🏻

crates/transaction-pool/src/test_utils/mock.rs
yongkangc: yup for clippy
https://github.com/paradigmxyz/reth/actions/runs/18001832227/job/51212718984?pr=18692
image

mattsse (approved):

crates/transaction-pool/src/test_utils/mock.rs
```diff
@@ -54,20 +54,20 @@ pub fn mock_tx_pool() -> MockTxPool {
...
+            MockTransaction::Eip4844 { $field, .. } |
+            MockTransaction::Eip2930 { $field, .. } |
+            MockTransaction::Eip7702 { $field, .. } => {
+                *$field = new_value.into();
             }
         }
         // Ensure the tx cost is always correct after each mutation.
         $this.update_cost();
-    };
+    }};
```
mattsse, line 94: this doesn't compile

why do we need to change macro for 2024?

yongkangc: from my chat with ai 

The issue is that the macro needs to handle different field types (u64, B256, Address, Bytes, etc.) and the match ergonomics are behaving differently in Rust 2024.

  1. In Rust 2021: The original code used ref mut $field in or-patterns, which explicitly created a mutable reference to the field regardless of its type.
  2. In Rust 2024:
    - ref mut in or-patterns is forbidden when matching on references
    - When matching &mut self, fields are automatically borrowed based on their usage
    - But the automatic borrowing behavior might not be consistent across all field types

mattsse (approved):

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2038,62 +2038,65 @@ where
         let sync_target_state = self.state.forkchoice_state_tracker.sync_target_state();
 
         // check if the downloaded block is the tracked finalized block
-        let mut exceeds_backfill_threshold = if let Some(buffered_finalized) = sync_target_state
+        let mut exceeds_backfill_threshold = match sync_target_state
```
yongkangc: refactored the logic here, because clippy was giving a hard time with 

```rust
2041 ~         let <mut> exceeds_backfill_threshold = if let (Some(downloaded_block), Some(ref state)) = (downloaded_block, sync_target_state) &&
2042 +             downloaded_block.hash == state.finalized_block_hash { self.exceeds_backfill_run_threshold(canonical_tip_num, downloaded_block.number) } else { if let Some(buffered_finalized) = sync_target_state
2043 +             .as_ref()
2044 +             .and_then(|state| self.state.buffer.block(&state.finalized_block_hash))
2045 +         {
2046 +             // if we have buffered the finalized block, we should check how far
2047 +             // we're off
2048 +             self.exceeds_backfill_run_threshold(canonical_tip_num, buffered_finalized.number())
2049 +         } else {
2050 +             // check if the distance exceeds the threshold for backfill sync
2051 +             self.exceeds_backfill_run_threshold(canonical_tip_num, target_block_number)
2052 +         } };
```

yongkangc: cc @mattsse for review here

shekhirin (approved):

crates/engine/tree/src/tree/mod.rs
mattsse: this match is equivalent so lgtm

mediocregopher (approved):

PR #18694, opened 2025-09-25
perf(tx-pool): Optimize batch tx validation by separating checks

### LINKED ISSUE #18527: Make it easier to optimize batch tx validations
### Describe the feature

In `EthTransactionValidator` we do:
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/transaction-pool/src/validate/eth.rs#L774-L787
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/transaction-pool/src/validate/eth.rs#L700-L710

Which allows to avoid overhead of instantiating state provider multiple times while also reusing code for transaction validation in all cases.

Currently this optimization is not applied for e.g `OpTransactionValidator` which still calls `validate_one` for each of the transactions
https://github.com/paradigmxyz/reth/blob/4b4b122e75182985088a523beb192c5f47b9975c/crates/optimism/txpool/src/validator.rs#L309-L317

What we could do is bake this optimization into `TransactionValidator` like this:
```rust
trait TransactionValidator {
    /// Performs validations not requiring state access.
    fn validate_transaction_stateless(
        &self,
        origin: TransactionOrigin,
        transaction: Self::Transaction,
    ) -> Result<Self::Transaction, TransactionValidationOutcome<Self::Transaction>>;

    /// Validates transaction against latest state
    fn validate_transaction_stateful(
        &self,
        origin: TransactionOrigin,
        transaction: Self::Transaction,
        state: impl StateProvider
    ) -> TransactionValidationOutcome<Self::Transaction>;
}
```

And then we could have `TransactionValidationTaskExecutor` take care of instantiating provider only when one or more transactions succeeded `validate_transaction_stateless`

### Additional context

_No response_

yongkangc:
closes: https://github.com/paradigmxyz/reth/issues/18527

This PR optimizes batch transaction validation by refactoring the TransactionValidator trait. 

The validation process is now split into two distinct steps: `validate_transaction_stateless` and `validate_transaction_stateful`.

The`TransactionValidationTaskExecutor` now 
1. runs stateless checks on all transactions first
2. uses a single state provider to perform stateful validation only on those that pass, minimizin state access.

### DIFF
--- a/crates/transaction-pool/src/validate/mod.rs
+++ b/crates/transaction-pool/src/validate/mod.rs
@@ -10,6 +10,7 @@ use alloy_eips::{eip7594::BlobTransactionSidecarVariant, eip7702::SignedAuthoriz
 use alloy_primitives::{Address, TxHash, B256, U256};
 use futures_util::future::Either;
 use reth_primitives_traits::{Recovered, SealedBlock};
+use reth_storage_api::{errors::provider::ProviderResult, StateProvider, StateProviderBox};
 use std::{fmt, fmt::Debug, future::Future, time::Instant};
 
 mod constants;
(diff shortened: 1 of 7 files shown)


yongkangc:
hey @klkvr just did an initial draft to see if this is something u had in mind?

crates/transaction-pool/src/validate/mod.rs
```diff
@@ -195,7 +219,30 @@ pub trait TransactionValidator: Debug + Send + Sync {
         &self,
         origin: TransactionOrigin,
         transaction: Self::Transaction,
-    ) -> impl Future<Output = TransactionValidationOutcome<Self::Transaction>> + Send;
+    ) -> impl Future<Output = TransactionValidationOutcome<Self::Transaction>> + Send {
+        async move {
+            match self.validate_transaction_stateless(origin, transaction).await {
+                Ok(transaction) => {
+                    let hash = *transaction.hash();
+                    match self.latest_state_provider() {
```
klkvr: let's make `validate_transactions` and batch validation functions on `TransactionValidator` take provider instead of requiring to know how to create one

crates/transaction-pool/src/validate/mod.rs
```diff
@@ -195,7 +219,30 @@ pub trait TransactionValidator: Debug + Send + Sync {
...
+                        Ok(provider) => {
+                            self.validate_transaction_stateful(
```
klkvr: we should also update default impls for `validate_transactions` and `validate_transactions_with_origin`

crates/transaction-pool/src/validate/task.rs
```diff
@@ -186,12 +187,93 @@ impl<V> TransactionValidationTaskExecutor<V> {
...
+        origin: TransactionOrigin,
+        transaction: V::Transaction,
+    ) -> TransactionValidationOutcome<V::Transaction> {
+        let mut results =
+            Self::validate_transactions_internal(validator, vec![(origin, transaction)]).await;
+        results.pop().expect("transaction result present")
+    }
+
+    /// Performs the stateless/stateful split for a batch and reuses one state provider.
+    async fn validate_transactions_internal(
```
klkvr: I think this is what default impl for `TransactionValidator::validate_transactions` should look like

and we can leave `TransactionValidationTaskExecutor` without many changes

PR #18696, opened 2025-09-25
refactor(engine): simplify InvalidBlockWitnessHook::on_invalid_block for better testability

### LINKED ISSUE #18513: refactor(engine): simplify InvalidBlockWitnessHook::on_invalid_block for better testability

0xKarl98:
Closes: https://github.com/paradigmxyz/reth/issues/18513

@yongkangc Please check if it meets the simplification requirement


crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -318,28 +240,40 @@ where
...
+
+        self.validate_bundle_state(&bundle_state, &output.state, &block_prefix)?;
+
+        self.validate_state_root_and_trie(parent_header, block, &bundle_state, trie_updates, &block_prefix)?;
+
+        Ok(())
     }
 
-    fn save_file<T: Serialize>(&self, filename: String, value: &T) -> eyre::Result<PathBuf> {
+    fn save_json<T: serde::Serialize>(&self, filename: &str, data: &T) -> eyre::Result<PathBuf> {
```
yongkangc:
```suggestion
    fn save_file<T: Serialize>(&self, filename: &str, data: &T) -> eyre::Result<PathBuf> {
```

Think it might be better to reduce changes here. Don't see a reason to change file name?

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -318,28 +240,40 @@ where
...
+        File::create(&path)?.write_all(content.as_bytes())?;
+        Ok(path)
+    }
 
+    fn save_diff<T: std::fmt::Debug>(&self, filename: &str, old: &T, new: &T) -> eyre::Result<PathBuf> {
+        let path = self.output_directory.join(filename);
+        let content = Comparison::new(old, new).to_string();
+        File::create(&path)?.write_all(content.as_bytes())?;
         Ok(path)
     }
```
yongkangc: Is there a reason why we cant keep it simliar to the prev implementation here?

i.e
```rust
   /// Saves the diff of two values into a file with the given name in the output directory.
    fn save_diff<T: PartialEq + Debug>(
        &self,
        filename: String,
        original: &T,
        new: &T,
    ) -> eyre::Result<PathBuf> {
        let path = self.output_directory.join(filename);
        let diff = Comparison::new(original, new);
        File::create(&path)?.write_all(diff.to_string().as_bytes())?;

        Ok(path)
    }
```

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -139,103 +128,44 @@ where
...
-
-        // Write the witness to the output directory.
-        let response =
-            ExecutionWitness { state, codes, keys: state_preimages, ..Default::default() };
-        let re_executed_witness_path = self.save_file(
-            format!("{}_{}.witness.re_executed.json", block.number(), block.hash()),
-            &response,
-        )?;
+    /// Handles witness generation, saving, and comparison with healthy node
+    fn handle_witness_operations(
```
yongkangc, line 228: i like the idea of this function, makes it much easier to see

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -139,103 +128,44 @@ where
...
-                bundle_state.contracts.values().map(|code| code.original_bytes()),
-            )
-            .collect();
-
-        // Grab all account proofs for the data accessed during block execution.
-        //
-        // Note: We grab *all* accounts in the cache here, as the `BundleState` prunes
-        // referenced accounts + storage slots.
-        let mut hashed_state = db.database.hashed_post_state(&bundle_state);
-        for (address, account) in db.cache.accounts {
```
yongkangc, line 184: where did all these function logic go?

yongkangc (changes requested):
overall much simpler, left some questions. 

Think the logic is not exactly equivalent?

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Yeah i was kind of thinking to combine save_json and save_diff into one function , as to reduce the loc here .
But seems should keep them the same as before would be better

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Yeah will restore it

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: This locates at the DataCollector impl

0xKarl98:
@yongkangc May you review this again ?

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -1,114 +1,120 @@
...
-    /// **Note**: Inside vector is *not* sorted by address.
-    ///
-    /// But it is unique by address.
-    pub reverts: Vec<Vec<(Address, AccountRevertSorted)>>,
-    /// The size of the plain state in the bundle state
-    pub state_size: usize,
-    /// The size of reverts in the bundle state
-    pub reverts_size: usize,
-}
+impl DataCollector {
```
yongkangc: can we remove this struct? seems like the struct might be a shallwo abstraction

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -1,114 +1,120 @@
...
+    fn collect(
```
yongkangc: could we leave clear and concise docstring for this?

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -1,114 +1,120 @@
...
-#[derive(Debug, PartialEq, Eq)]
-struct AccountRevertSorted {
-    pub account: AccountInfoRevert,
-    pub storage: BTreeMap<U256, RevertToSlot>,
-    pub previous_status: AccountStatus,
-    pub wipe_storage: bool,
+type CollectionResult =
+    (BTreeMap<B256, Bytes>, BTreeMap<B256, Bytes>, reth_trie::HashedPostState, BundleState);
+
+fn sort_bundle_state_for_comparison(bundle_state: &BundleState) -> serde_json::Value {
```
yongkangc: could we leave a line of docstring here thats clear and concise explaining what this does?

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: You mean DataCollector ? If so , will do  , and will also deal with proof generator in the same way

0xKarl98:
@yongkangc  Have done that requested changes

yongkangc (changes requested):

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: This part of code has been removed in previous touchups

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Done

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -1,114 +1,113 @@
...
+                    "account": rev.account,
+                    "storage": BTreeMap::from_iter(rev.storage.clone()),
+                    "previous_status": rev.previous_status,
+                    "wipe_storage": rev.wipe_storage
+                }))
+            }).collect::<BTreeMap<_, _>>()
+        }).collect::<Vec<_>>(),
+        "state_size": bundle_state.state_size,
+        "reverts_size": bundle_state.reverts_size
+    })
```
Rjected: IMO we should not be manually constructing a `serde_json::Value` ever, if this is supposed to be a serializable version of `BundleStateSorted` then we should just derive serialize for it

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -245,30 +197,34 @@ where
...
-            let filename = format!("{}_{}.bundle_state.diff", block.number(), block.hash());
-            // Convert bundle state to sorted struct which has BTreeMap instead of HashMap to
-            // have deterministic ordering
-            let bundle_state_sorted = BundleStateSorted::from_bundle_state(&bundle_state);
-            let output_state_sorted = BundleStateSorted::from_bundle_state(&output.state);
-
-            let diff_path = self.save_diff(filename, &bundle_state_sorted, &output_state_sorted)?;
+            // Convert bundle state to sorted format for deterministic comparison
+            let bundle_state_sorted = sort_bundle_state_for_comparison(re_executed_state);
+            let output_state_sorted = sort_bundle_state_for_comparison(original_state);
```
Rjected, line 279: I preferred the old structs for this

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -139,103 +138,56 @@ where
...
-            // If the witnesses are different, write the diff to the output directory.
-            if response != healthy_node_witness {
-                let filename = format!("{}_{}.witness.diff", block.number(), block.hash());
-                let diff_path = self.save_diff(filename, &response, &healthy_node_witness)?;
+            if witness != &healthy_node_witness {
+                let diff_path = self.save_diff(
+                    format!("{}.witness.diff", block_prefix),
+                    witness,
+                    &healthy_node_witness,
+                )?;
```
Rjected: for this and the other fns where `format!` calls are used directly as arguments, I preferred the old way because it was more explicit e.g.:
```
let filename = format!(....);
let diff_path = self.save_diff(filename, ...);
```

Rjected (dismissed):
thanks! mostly style comments and suggestions about using `serde_json::Value` so far

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Done

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Please check it out again

yongkangc (approved):
lgtm! @Rjected @shekhirin @mediocregopher any other thoughts?

mediocregopher (approved):
LGTM

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -40,75 +59,121 @@ struct BundleAccountSorted {
...
+        ..Default::default()
+    })
 }
 
-/// Generates a witness for the given block and saves it to a file.
 #[derive(Debug)]
+/// Hook for generating execution witnesses when invalid blocks are detected.
+///
+/// This hook captures the execution state and generates witness data that can be used
+/// for debugging and analysis of invalid block execution.
```
Rjected:
```suggestion
/// Hook for generating execution witnesses when invalid blocks are detected.
///
/// This hook captures the execution state and generates witness data that can be used
/// for debugging and analysis of invalid block execution.
#[derive(Debug)]
```

Rjected (commented):
one nit otherwise lgtm

crates/engine/invalid-block-hooks/src/witness.rs
0xKarl98: Done

yongkangc (approved):
thanks for the work @0xKarl98 ! LGTM, merging

.github/assets/check_wasm.sh
```diff
@@ -10,6 +10,16 @@ crates=($(cargo metadata --format-version=1 --no-deps | jq -r '.packages[].name'
...
+  reth-cli-util
+  reth-db
+  reth-db-api
+  reth-ecies
+  reth-network-api
+  reth-nippy-jar
+  reth-node-types
+  reth-rpc-server-types
+  reth-trie-db
+  reth-zstd-compressors
```
yongkangc: why did we add these?

0xKarl98: If not add them , run bash .github/assets/check_wasm.sh will return error and this will resolve the failed WASM while didn't touch other files 

This WASM job failure didn't get encountered several days ago

0xKarl98: And i have fixed other  new lint errors after you approved the merge

Rjected: We should probably not add these here, and instead figure out why they failed

Rjected (dismissed):
lgtm other than the check_wasm additions

.github/assets/check_wasm.sh
0xKarl98: Have started investigated the failure reason and try not add these :)

0xKarl98:
@Rjected Need your approval here :)

yongkangc (approved):

PR #18697, opened 2025-09-25
chore(flashblocks): flasblock service metrics

dustinjake:
## Description

Add new metrics to the flashblock service to track execution duration and last flashblock length.


mattsse (approved):
seems appropriate, ty

PR #18702, opened 2025-09-25
fix: replace unbounded mpsc with bounded channels and backpressure in RLPx multiplexer

prestoalvarez:
- Why
  - Prevent unbounded memory growth and potential DoS from unbounded channels.
  - Handle channel closure explicitly; avoid silently ignoring send errors.

- What changed
  - Switched to bounded mpsc::channel with capacity 256 for routing (to_primary/to_satellite).
  - Replaced UnboundedReceiverStream with ReceiverStream where applicable.
  - Used try_send in the main poll; on Full: drop to apply backpressure; on Closed: terminate gracefully.
  - During handshake, use send(...).await to avoid losing critical messages.
  - Added brief comments documenting the rationale.

- Impact
  - Security: reduces OOM/DoS risk from message floods.
  - Behavior: under sustained congestion, some inbound messages may be dropped (bounded by policy).
  - API: no public API changes; internal wiring only. Capacity is easy to tune.


mattsse (commented):
this has been working fine

PR #18703, opened 2025-09-25
Secure install_geth.sh: add optional SHA256 verification and named downloads

anim001k:
### Description
- Added optional SHA256 integrity check via `GETH_SHA256` env var to prevent supply-chain tampering.
- Switched to named variables (`url`, `tarball`) and explicit `wget -O` for clarity and reliability.
- Retained original behavior when `GETH_SHA256` is not provided.

#### Why
- Downloaded binaries were installed without integrity/authenticity verification, exposing a potential MITM risk.

#### How
- Verify archive when `GETH_SHA256` is set:
  - `echo "$GETH_SHA256  $tarball" | sha256sum -c -`
- Keep logic otherwise unchanged (extract, move, chmod).


mattsse (commented):
this has been working fine

PR #18705, opened 2025-09-25
chore: remove unused once_cell dependency from reth-primitives

eeemmmmmm:
Remove dead code from reth-primitives crate:
- Remove unused `once_cell` dependency from Cargo.toml
- Remove unused `LazyLock` and `OnceLock` imports from transaction module
- Clean up `once_cell/std` feature flag

The once_cell dependency was left over from refactoring and is no longer used anywhere in the primitives crate


mattsse (commented):
this does not work because std lazy isnt no-std

PR #18707, opened 2025-09-25
feat(downloaders): add file-client feature gate

Waiting-Chai:
Add optional file-client feature to gate file_codec, file_client, and receipt_file_client modules behind async-compression dependency.

- Make async-compression dependency optional
- Add file-client feature that enables async-compression
- Gate related modules with cfg(any(test, feature = "file-client"))
- Reduce default compilation dependencies when file client is not needed

close #18698


mattsse (approved):
makes sense, ty!

PR #18711, opened 2025-09-25
perf: avoid redundant bytecode hash calculation in RPC provider

gancerlory:
redundant hash computation in RpcBlockchainStateProvider.get_account() by using account_info.code_hash() directly instead of creating a Bytecode object and calling hash_slow().

Previously:
- account_info.code_hash() was called for caching
- Bytecode::new_raw() + hash_slow() was called for the return value

Now:
- account_info.code_hash() is used directly for both caching and return value

reduces unnecessary hash computations

### DIFF
--- a/crates/storage/rpc-provider/src/lib.rs
+++ b/crates/storage/rpc-provider/src/lib.rs
@@ -1063,16 +1063,15 @@ impl<P: Clone, Node: NodeTypes, N> RpcBlockchainStateProvider<P, Node, N> {
         {
             Ok(None)
         } else {
-            let bytecode = if account_info.code.is_empty() {
-                None
-            } else {
-                Some(Bytecode::new_raw(account_info.code))
-            };
+            // Use account_info.code_hash() directly instead of creating Bytecode and calling
+            // hash_slow()
+            let bytecode_hash =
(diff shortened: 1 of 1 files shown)


crates/storage/rpc-provider/src/lib.rs
```diff
@@ -1063,16 +1063,15 @@ impl<P: Clone, Node: NodeTypes, N> RpcBlockchainStateProvider<P, Node, N> {
         {
             Ok(None)
         } else {
-            let bytecode = if account_info.code.is_empty() {
-                None
-            } else {
-                Some(Bytecode::new_raw(account_info.code))
-            };
+            // Use account_info.code_hash() directly instead of creating Bytecode and calling
+            // hash_slow()
```
mattsse: this comment will be confusing without the pr context, can we remove this

mattsse (changes requested):
makes sense, pedantic nit

crates/storage/rpc-provider/src/lib.rs
gancerlory: Removed

mattsse (approved):

PR #18712, opened 2025-09-25
fix: Bearer token parsing vulnerability 

viktorking7:
## Description

**Security Fix**: Fixed a vulnerability in the `get_bearer` function that could allow authentication bypass through malformed Authorization headers.

### Problem
The original implementation used `find()` to locate "Bearer " anywhere in the header string, which could be exploited with headers like:
```
Authorization: NotBearer Bearer valid_token
```

### Solution
- Changed from `find()` to `starts_with()` to ensure "Bearer " is at the beginning of the header

### DIFF
--- a/crates/rpc/rpc-layer/src/jwt_validator.rs
+++ b/crates/rpc/rpc-layer/src/jwt_validator.rs
@@ -47,10 +47,16 @@ fn get_bearer(headers: &HeaderMap) -> Option<String> {
     let header = headers.get(header::AUTHORIZATION)?;
     let auth: &str = header.to_str().ok()?;
     let prefix = "Bearer ";
-    let index = auth.find(prefix)?;
-    let token: &str = &auth[index + prefix.len()..];
+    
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this actually looks correct

crates/rpc/rpc-layer/src/jwt_validator.rs
```diff
@@ -47,8 +47,13 @@ fn get_bearer(headers: &HeaderMap) -> Option<String> {
     let header = headers.get(header::AUTHORIZATION)?;
     let auth: &str = header.to_str().ok()?;
     let prefix = "Bearer ";
-    let index = auth.find(prefix)?;
-    let token: &str = &auth[index + prefix.len()..];
+
+    // Ensure the header starts with "Bearer " to prevent bypass attacks
```
mattsse: can we remove this comment, this doesnt have any relevance without the context of this pr

mattsse (changes requested):

mattsse (approved):

PR #18713, opened 2025-09-25
fix(engine): resolve race condition during block reorganization

avorylli:
## Problem
Race condition occurs when new blocks arrive during reorganization while block removal is in progress. This causes `PersistingKind::PersistingNotDescendant` classification, preventing parallel state root computation and leading to fatal error "no header found for Number(14)".

## Solution
Modified `persisting_kind_for` method to handle `RemovingBlocks` operations intelligently:
- Allow parallel state root computation for blocks that are canonical descendants and not being removed
- Only block parallel computation for blocks that are actually being removed

## Changes
- Updated `persisting_kind_for` in `payload_validator.rs` and `mod.rs`
- Added test case to verify the fix

Fixes #18408

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -602,8 +602,8 @@ where
                         latest_valid_hash = Some(block_hash);
                         PayloadStatusEnum::Valid
                     }
-                    InsertPayloadOk::Inserted(BlockStatus::Disconnected { .. })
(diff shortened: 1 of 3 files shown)


yongkangc:
hey @avorylli we would want to fix the ci here first

mediocregopher:
Hi @avorylli , your submission seems to be very similar to that of #18567 , and I believe is not going to work for the same reasons. Is there some difference in your change that I'm missing? Please see also my most recent comment in the original issue for what is more likely the actual root cause of the error that was being reported.

PR #18714, opened 2025-09-25
chore: small fix in navigational doc

hawkadrian:


mattsse (commented):
this is referenced elsewhere

PR #18716, opened 2025-09-25

crStiv:


mattsse (changes requested):
most of these files are generated

PR #18717, opened 2025-09-25
fix: replace panic! calls in engine API functions

gancerlory:
There are 4 panic calls in `valid_payload.rs` that kill the process instead of letting errors bubble up to tools like `reth-bench-compare`. Changed them to return proper errors so the calling code can handle failures gracefully. Also fixed a couple unwrap calls in the same file while I was at it.


mattsse (commented):
I prefer the panics here because these then immediately point to the location in case something's wrong

PR #18721, opened 2025-09-26
feat(op-reth): add optional state root calc for flashblocks

julio4:
Towards #18255

`FlashBlockService` is extended with an additional flag `calculate_state_root`, default to false and can be activated with `enable_state_root()`, and the state root is stored in `FlashBlockPendingSequence` as an Option.

During block buiding of pending block, if the current index in the pending sequence is >= FB_STATE_ROOT_FROM_INDEX (initialized to 8), we request state root calculation in BuildArgs. This prevents calculating state roots too early.

`FlashBlockCompleteSequence` take the state root from its original `FlashBlockPendingSequence` with the From implementation, so that it can be further used in consensus client.


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -121,20 +130,29 @@ where
...
 
 /// A complete sequence of flashblocks, often corresponding to a full block.
 /// Ensure invariants of a complete flashblocks sequence.
 #[derive(Debug, Clone)]
-pub struct FlashBlockCompleteSequence(Vec<FlashBlock>);
+pub struct FlashBlockCompleteSequence {
+    inner: Vec<FlashBlock>,
+    /// Optional state root for the current sequence
+    pub state_root: Option<B256>,
+}
```
mattsse, line 152: we shouldn't mix pub and non pub, imo we should keep this private here as well

crates/optimism/flashblocks/src/service.rs
```diff
@@ -81,9 +85,16 @@ where
             job: None,
             cached_state: None,
             metrics: FlashBlockServiceMetrics::default(),
+            calculate_state_root: false,
         }
     }
 
+    /// Enable state root calculation from flashblock
+    pub const fn enable_state_root(mut self) -> Self {
```
mattsse: we need this as function that accepts a bool instead

crates/optimism/flashblocks/src/service.rs
```diff
@@ -120,7 +131,7 @@ where
                 "Missing flashblock payload base"
             );
 
-            return None
+            return None;
```
mattsse: please ser, no more of this :)

crates/optimism/flashblocks/src/service.rs
```diff
@@ -25,6 +25,8 @@ use std::{
 use tokio::{pin, sync::oneshot};
 use tracing::{debug, trace, warn};
 
+pub(crate) const FB_STATE_ROOT_FROM_INDEX: usize = 8;
```
mattsse: I think we can do index 9? because they will always have 0..9?

mattsse (changes requested):
nice, a few suggestions

crates/optimism/flashblocks/src/service.rs
julio4: my bad, bad fmt

mattsse (approved):
lgtm, ty!

I only resolved the conflicts and renamed a few things

PR #18722, opened 2025-09-26
feat: keep track of most recently emitted range update and change int…

### LINKED ISSUE #18719: Keep track of most recently emitted range update and change interval
### Describe the feature

the spec

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7642.md#blockrangeupdate-message-1

mentions 

> Clients should send an update at most once per epoch (32 blocks)

I believe it used to be 2mins

we need to change two things

* change the interval to `EPOCH_DURATION_SECS`, update docs
https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/net/network/src/session/active.rs#L49-L49
* keep track of height changes, e.g. if the height doesnt change because the node isnt advancing, we dont need to send an update which we can track separately https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/net/network/src/session/active.rs#L127-L127 like `prev_latest_block` which we can then check here
https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/net/network/src/session/active.rs#L743-L745

before queuing the message and simply ignore if `range.latest - prev_latest < 32` 
 

### Additional context

_No response_

stevencartavia:
closes #18719

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -43,10 +43,15 @@ use tokio_stream::wrappers::ReceiverStream;
 use tokio_util::sync::PollSender;
 use tracing::{debug, trace};
 
-/// The recommended interval at which a new range update should be sent to the remote peer.
+/// Duration of an epoch in seconds (32 blocks * 12 seconds per block = 384 seconds)
+pub(super) const EPOCH_DURATION_SECS: u64 = 384;
+
+/// The recommended interval at which to check if a new range update should be sent to the remote
+/// peer.
 ///
-/// This is set to 120 seconds (2 minutes) as per the Ethereum specification for eth69.
-pub(super) const RANGE_UPDATE_INTERVAL: Duration = Duration::from_secs(120);
+/// Updates are only sent when the block height has advanced by at least 32 blocks since the last
+/// update.
+pub(super) const RANGE_UPDATE_INTERVAL: Duration = Duration::from_secs(EPOCH_DURATION_SECS);
 
 // Constants for timeout updating.
 
@@ -128,6 +133,9 @@ pub(crate) struct ActiveSession<N: NetworkPrimitives> {
     /// Optional interval for sending periodic range updates to the remote peer (eth69+)
     /// Recommended frequency is ~2 minutes per spec
     pub(crate) range_update_interval: Option<Interval>,
+    /// The last latest block number we sent in a range update
+    /// Used to avoid sending unnecessary updates when block height hasn't changed significantly
+    pub(crate) last_sent_latest_block: Option<u64>,
 }
 
 impl<N: NetworkPrimitives> ActiveSession<N> {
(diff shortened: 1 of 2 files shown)


crates/net/network/src/session/active.rs
```diff
@@ -43,10 +43,15 @@ use tokio_stream::wrappers::ReceiverStream;
 use tokio_util::sync::PollSender;
 use tracing::{debug, trace};
 
-/// The recommended interval at which a new range update should be sent to the remote peer.
+/// Duration of an epoch in seconds (32 blocks * 12 seconds per block = 384 seconds)
+pub(super) const EPOCH_DURATION_SECS: u64 = 384;
```
mattsse: this constant should already be in scope from somewhere else I believe

crates/net/network/src/session/active.rs
```diff
@@ -128,6 +133,9 @@ pub(crate) struct ActiveSession<N: NetworkPrimitives> {
     /// Optional interval for sending periodic range updates to the remote peer (eth69+)
     /// Recommended frequency is ~2 minutes per spec
```
mattsse: we can update this doc as well

crates/net/network/src/session/active.rs
```diff
@@ -738,11 +746,23 @@ impl<N: NetworkPrimitives> Future for ActiveSession<N> {
...
-                    EthMessage::BlockRangeUpdate(this.local_range_info.to_message()).into(),
-                );
+                let current_latest = this.local_range_info.latest();
+                let should_send = match this.last_sent_latest_block {
+                    None => true, // First update, always send
+                    Some(last_sent) => {
+                        // Only send if block height has advanced by at least 32 blocks (one epoch)
+                        current_latest.saturating_sub(last_sent) >= 32
+                    }
+                };
```
mattsse, line 756: let's use an if let some here

mattsse (changes requested):
ty, smol suggestions

crates/net/network/src/session/active.rs
stevencartavia: I didn't find that one exactly but got it as `EPOCH_SLOTS * 12`

crates/net/network/src/session/active.rs
mattsse, line 759: we're missing an update to last_sent_latest_block here

mattsse (changes requested):
last nit :)

crates/net/network/src/session/active.rs
stevencartavia: this one is below `this.last_sent_latest_block = Some(current_latest);`, did you mean putting it inside the `push_back`?

mattsse: oops my bad, didnt scroll that far hehe

mattsse (approved):
ty ser

PR #18724, opened 2025-09-26
refactor: Simplify the process of prune handling for u64::MAX shard

AshinGau:
1. Simplify the process of prune handling for `u64::MAX` shard: https://github.com/paradigmxyz/reth/discussions/18723
2. Fix the doc of `Cursor:: seek_by_key_subkey`, which need the key must match exactly.

### DIFF
--- a/crates/prune/prune/src/segments/user/history.rs
+++ b/crates/prune/prune/src/segments/user/history.rs
@@ -112,41 +112,8 @@ where
         // If there will be no more blocks in the shard after pruning blocks below target
         // block, we need to remove it, as empty shards are not allowed.
         if higher_blocks.is_empty() {
-            if key.as_ref().highest_block_number == u64::MAX {
-                let prev_row = cursor
-                    .prev()?
-                    .map(|(k, v)| Result::<_, DatabaseError>::Ok((k.key()?, v)))
-                    .transpose()?;
-                match prev_row {
-                    // If current shard is the last shard for the sharded key that
-                    // has previous shards, replace it with the previous shard.
-                    Some((prev_key, prev_value)) if key_matches(&prev_key, &key) => {
-                        cursor.delete_current()?;
-                        // Upsert will replace the last shard for this sharded key with
-                        // the previous value.
-                        cursor.upsert(RawKey::new(key), &prev_value)?;
-                        Ok(PruneShardOutcome::Updated)
-                    }
-                    // If there's no previous shard for this sharded key,
-                    // just delete last shard completely.
-                    _ => {
-                        // If we successfully moved the cursor to a previous row,
-                        // jump to the original last shard.
-                        if prev_row.is_some() {
-                            cursor.next()?;
-                        }
-                        // Delete shard.
-                        cursor.delete_current()?;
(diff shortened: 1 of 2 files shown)


mattsse (commented):
not obvious to me what this fixes

ptal @joshieDo @shekhirin

crates/storage/db-api/src/cursor.rs
```diff
@@ -75,6 +75,7 @@ pub trait DbDupCursorRO<T: DupSort> {
     ///
     /// # Note
     ///
+    /// The key must match exactly, while the subkey can be greater than or equal to.
```
shekhirin, line 78: This is correct, cross-checked with MDBX docs https://libmdbx.dqdkfa.ru/group__c__cursors.html#gga501baff3bcff7e8e7818b4d412a2a6d5a2de3e789c951f80dbc05cc9c0fba82e8

crates/prune/prune/src/segments/user/history.rs
```diff
@@ -112,41 +112,8 @@ where
...
-                    }
-                }
-            }
-            // If current shard is not the last shard for this sharded key,
-            // just delete it.
-            else {
-                cursor.delete_current()?;
-                Ok(PruneShardOutcome::Deleted)
-            }
+            cursor.delete_current()?;
```
shekhirin, line 118: I think this change is correct, but I would like to confirm it by having a `debug_assert` that will query the shards for the same key and ensure there's none left.

Also, can we add a comment explaining why it's fine to just delete it?

shekhirin:
As far as I understand, we'd basically always fall through to this statement? https://github.com/paradigmxyz/reth/blob/597fa73023ba36ed90dbd993192bfb337d83c2d2/crates/prune/prune/src/segments/user/history.rs#L130-L141

crates/storage/db-api/src/cursor.rs
AshinGau: The context is that I was testing the use of `leveldb/rocksdb` to replace `mdbx`, and found that the comment of this function was not very accurate. I mistakenly thought that implementing a function greater than or equal to 'key+subkey' would suffice, which resulted in an error. Therefore, I added more note.

crates/prune/prune/src/segments/user/history.rs
AshinGau: Yes,  this PR just makes it more succinct. And I add more comment to explain.

AshinGau: The context is still that I wanted to replace `mdbx` with `rocksdb`, but found it difficult to implement the calling process of `next -> prev ->delete_current ->upsert ->next...` in `rocksdb`. Finally, I found that this process could be simplified, so I discovered this code that can be simplified.

AshinGau: Therefore, there's no bug fix in this PR, just refactoring.

shekhirin: can you please add a `debug_assert`?

PR #18726, opened 2025-09-26
fix: replace panic-prone expect() with safe Option handling in remove_blocks_above

aso20455:
Replace dangerous .expect("todo: error handling, headers should exist") with proper Option pattern matching in UnifiedStorageWriter::remove_blocks_above().

The get_highest_static_file_block() method returns Option<BlockNumber> and can be None when no static file headers exist on disk. Using .expect() could cause runtime panics during blockchain reorganization operations.

Changes:
- Replace .expect() with if-let Some() pattern matching
- Add graceful handling when static file headers are absent
- Add informative debug logging for both success and skip cases
- Maintain identical behavior when headers exist


crates/storage/provider/src/writer/mod.rs
```diff
@@ -204,21 +204,23 @@ where
...
-        // will have three blocks to remove, which will be block 8, 7, and 6.
-        debug!(target: "provider::storage_writer", ?block_number, "Removing static file blocks above block_number");
-        self.static_file()
-            .get_writer(block_number, StaticFileSegment::Headers)?
-            .prune_headers(highest_static_file_block.saturating_sub(block_number))?;
+        // Get highest static file block for the total block range. If no static file headers
+        // exist on disk, there's nothing to prune from static files.
+        if let Some(highest_static_file_block) =
+            self.static_file().get_highest_static_file_block(StaticFileSegment::Headers)
+        {
```
mattsse: can we use `let some else` pattern here?

mattsse (changes requested):
style nit

crates/storage/provider/src/writer/mod.rs
```diff
@@ -204,14 +204,17 @@ where
...
-            .expect("todo: error handling, headers should exist");
-
-        // IMPORTANT: we use `highest_static_file_block.saturating_sub(block_number)` to make sure
-        // we remove only what is ABOVE the block.
+        // Get highest static file block for the total block range. If no static file headers
+        // exist on disk, there's nothing to prune from static files.
+        let Some(highest_static_file_block) =
+            self.static_file().get_highest_static_file_block(StaticFileSegment::Headers)
+        else {
+            debug!(target: "provider::storage_writer", ?block_number, "No static file headers found, skipping static file pruning");
```
mattsse: maybe we should even use warn! here?

mattsse (commented):
makes sense to me, pending @Rjected @shekhirin

crates/storage/provider/src/writer/mod.rs
Rjected: yeah let's use warn at least, this basically means we can't find any static files, when they should always exist

aso20455: > maybe we should even use warn! here?

@mattsse can I do this?

PR #18727, opened 2025-09-26
perf(multiproof): do not chunk more tasks when task queue is full

hai-rise:
A low-hanging fruit among many (task management) ideas!

---

For big state transitions (from 11s down to 500ms):

Screenshot From 2025-09-26 16-03-34

---

For small state transitions:

`reth-bench` for the first 300 blocks of Shanghai:

```
$ reth-bench new-payload-fcu --advance 300
```

aa192c255b53335ba509987c82ab26e79e954195 (`main`):
```
2025-09-26T09:06:00.711782Z  INFO Total Ggas/s: 0.7019 total_duration=6.539663621s total_gas_used=4590139398 blocks_processed=300
```

This branch on aa192c255b53335ba509987c82ab26e79e954195:
```
2025-09-26T09:23:34.656611Z  INFO Total Ggas/s: 0.7467 total_duration=6.146954381s total_gas_used=4590139398 blocks_processed=300
```

Speedup:
- Ggas/s: 0.7019 -> 0.7467 ~ 6% higher execution throughput
- Duration: 6.539663621s -> 6.146954381s ~ 6% lower payload latency
- Best win: From 32.3ms down to 22.4ms payload latency

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -378,6 +378,10 @@ where
         }
     }
 
+    const fn is_full(&self) -> bool {
+        self.inflight >= self.max_concurrent
+    }
+
     /// Spawns a new multiproof calculation or enqueues it for later if
     /// `max_concurrent` are already inflight.
     fn spawn_or_queue(&mut self, input: PendingMultiproofTask<Factory>) {
@@ -391,7 +395,7 @@ where
             return
         }
 
-        if self.inflight >= self.max_concurrent {
+        if self.is_full() {
             self.pending.push_back(input);
             self.metrics.pending_multiproofs_histogram.record(self.pending.len() as f64);
             return;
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -830,25 +845,40 @@ where
 
         // Process state updates in chunks.
         let mut chunks = 0;
+        let should_chunk = !self.multiproof_manager.is_full();
```
hai-rise, line 848: We can try something more sophisticated, like tuning the number of chunks based on the number of available task slots (up to a max), but that would require many benchmarks. I'll set up a mainnet node on a new bare metal server first for such experiments. Shanghai blocks too small and boring..

mediocregopher (approved):
This LGTM

Curious if you did any tests with chunking disabled completely?

Also related: #18036 though I think what you're doing here is more impactful

mattsse (approved):
this makes a lot of sense as an immediate bandaid

@shekhirin and I also started to investigate this properly, but imo this we can do right away already

pending @shekhirin

shekhirin (approved):

PR #18729, opened 2025-09-26
fix: Apply WS CORS regardless of HTTP being enabled

anim001k:
### Description
- Fixes an issue where `ws.origins` was only applied when HTTP was enabled.
- Now `with_ws_cors(...)` is set inside the WS branch, ensuring WS CORS is honored for WS-only setups.

### Changes
- `crates/rpc/rpc-builder/src/config.rs`: move `with_ws_cors(...)` into `if self.ws { ... }` with a clarifying comment.


mattsse (approved):
ah nice find

PR #18730, opened 2025-09-26
chore(trie): Implement clear trie changeset methods

mediocregopher:
These methods will be used in various places when it's necessary to delete changeset data. This includes pruning, unwinding, and also when populating the data during pipeline sync in certain cases.

A new type BlockNumberHashedAddressRange is implemented for convenience.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2388,12 +2388,9 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
         Ok(())
     }
 
-    fn clear_trie_changesets_range(
-        &self,
-        range: impl RangeBounds<BlockNumber>,
-    ) -> ProviderResult<()> {
+    fn clear_trie_changesets_from(&self, from: BlockNumber) -> ProviderResult<()> {
         let tx = self.tx_ref();
-        let range = (range.start_bound(), range.end_bound()); // avoid clone
+        let range = from..;
         {
             let mut cursor = tx.cursor_dup_write::<tables::AccountsTrieChangeSets>()?;
             let mut walker = cursor.walk_range(range)?;
@@ -2404,7 +2401,7 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
         }
 
         {
-            let range: BlockNumberHashedAddressRange = range.into();
+            let range: BlockNumberHashedAddressRange = (from..).into();
             let mut cursor = tx.cursor_dup_write::<tables::StoragesTrieChangeSets>()?;
             let mut walker = cursor.walk_range(range)?;
 
@@ -3726,7 +3723,7 @@ mod tests {
     }
 
     #[test]
-    fn test_clear_trie_changesets_range() {
+    fn test_clear_trie_changesets_from() {
         use alloy_primitives::hex_literal::hex;
         use reth_db_api::models::BlockNumberHashedAddress;
         use reth_trie::{BranchNodeCompact, StoredNibblesSubKey, TrieChangeSetsEntry};
@@ -3779,7 +3776,7 @@ mod tests {
             let mut cursor =
                 provider_rw.tx_ref().cursor_dup_write::<tables::AccountsTrieChangeSets>().unwrap();
 
-            // Block 100: 2 entries (will be kept - before range)
+            // Block 100: 2 entries (will be kept - before start block)
             cursor
                 .upsert(
                     block1,
@@ -3790,7 +3787,7 @@ mod tests {
                 .upsert(block1, &TrieChangeSetsEntry { nibbles: nibbles2.clone(), node: None })
                 .unwrap();
 
-            // Block 101: 3 entries with duplicates (will be deleted - in range)
+            // Block 101: 3 entries with duplicates (will be deleted - from this block onwards)
             cursor
                 .upsert(
                     block2,
@@ -3807,7 +3804,7 @@ mod tests {
                 .upsert(block2, &TrieChangeSetsEntry { nibbles: nibbles3.clone(), node: None })
                 .unwrap();
 
-            // Block 102: 2 entries (will be deleted - in range)
+            // Block 102: 2 entries (will be deleted - after start block)
             cursor
                 .upsert(
                     block3,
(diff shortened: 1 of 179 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2379,6 +2380,61 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+        range: impl RangeBounds<BlockNumber>,
+    ) -> ProviderResult<()> {
+        let tx = self.tx_ref();
+
+        {
+            let range_start =
+                if let Bound::Included(v) = range.start_bound() { Some(*v) } else { None };
+
+            let mut cursor = tx.cursor_dup_write::<tables::AccountsTrieChangeSets>()?;
+            let mut walker = cursor.walk(range_start)?;
```
shekhirin: can we use `walk_range` here?

mediocregopher: Yup, done

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2379,6 +2380,44 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+        range: impl RangeBounds<BlockNumber>,
+    ) -> ProviderResult<()> {
+        let tx = self.tx_ref();
+        let range = (range.start_bound(), range.end_bound()); // avoid clone
+                                                              //
+        {
+            let mut cursor = tx.cursor_dup_write::<tables::AccountsTrieChangeSets>()?;
+            let mut walker = cursor.walk_range(range)?;
+
+            while let Some(entry) = walker.next() {
```
shekhirin: nit
```suggestion
            while let Some(_) = walker.next().transpose()? {
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2379,6 +2380,44 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+                walker.delete_current()?;
+            }
+        }
+
+        {
+            let range: BlockNumberHashedAddressRange = range.into();
+            let mut cursor = tx.cursor_dup_write::<tables::StoragesTrieChangeSets>()?;
+            let mut walker = cursor.walk_range(range)?;
+
+            while let Some(entry) = walker.next() {
```
shekhirin: same

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2379,6 +2380,44 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+    fn clear_trie_changesets(&self) -> ProviderResult<()> {
+        let tx = self.tx_ref();
+        tx.clear::<tables::AccountsTrieChangeSets>()?;
+        tx.clear::<tables::StoragesTrieChangeSets>()?;
+        Ok(())
+    }
+
+    fn clear_trie_changesets_range(
+        &self,
+        range: impl RangeBounds<BlockNumber>,
```
shekhirin: something we might want to consider immediately now is that if we move trie changesets to static files, we would not be able to provide a `range`, it will just be "truncate N from the end". Maybe we should change it to be a truncate?

mediocregopher: Changed to `clear_trie_changesets_from`.

I'm not sure it's going to make sense to have these in static files tho, at least in their current state. Without an accompanying index the data isn't very useful beyond reverting a relatively small amount of blocks, so storing in static files would have us keep way more than we could use for no benefit.

shekhirin: yeah, the indices would live in the DB same as for account/storage history, but the changesets themselves in static files. This can improve storage and perf of querying actual trie changesets.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2379,6 +2380,38 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+
+            while walker.next().transpose()?.is_some() {
+                walker.delete_current()?;
+            }
+        }
+
+        {
+            let range: BlockNumberHashedAddressRange = (from..).into();
+            let mut cursor = tx.cursor_dup_write::<tables::StoragesTrieChangeSets>()?;
+            let mut walker = cursor.walk_range(range)?;
```
Rjected: so this is the only place we really use `BlockNumberHashedAddressRange`? maybe we could just remove the struct and do
```
        let range: RangeFrom<BlockNumberHashedAddress> = (from, B256::ZERO).into()..;
```

Rjected (commented):
unless we use the `BlockNumberHashedAddressRange` somewhere else, thinking we can make this simpler

crates/storage/provider/src/providers/database/provider.rs
mediocregopher: Good point, the range type made more sense when the clear method actually accepted a range, removed.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2399,6 +2399,38 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
...
+            let mut cursor = tx.cursor_dup_write::<tables::AccountsTrieChangeSets>()?;
+            let mut walker = cursor.walk_range(range)?;
+
+            while walker.next().transpose()?.is_some() {
+                walker.delete_current()?;
+            }
+        }
+
+        {
+            let range: RangeFrom<BlockNumberHashedAddress> = (from, B256::ZERO).into()..;
```
mattsse, line 2423: I see, this would be the first (num, hash) pair for `BlockNumberHashedAddress`

crates/storage/storage-api/src/trie.rs
```diff
@@ -116,6 +116,12 @@ pub trait TrieWriter: Send + Sync {
         trie_updates: &TrieUpdatesSorted,
         updates_overlay: Option<&TrieUpdatesSorted>,
     ) -> ProviderResult<usize>;
+
+    /// Clears contents of trie changesets completely
+    fn clear_trie_changesets(&self) -> ProviderResult<()>;
+
+    /// Clears contents of trie changesets starting from the given block number onwards.
```
mattsse: can we extra mention that this is inclusive `[from..]`

mattsse (approved):
lgtm, one pedantic doc suggestion

Rjected (approved):
LGTM, i like the tests

crates/storage/storage-api/src/trie.rs
mediocregopher: Yup, done :+1:

PR #18732, opened 2025-09-26
chore(trie): Overlay arguments for write trie changeset methods

mediocregopher:
This allows for passing optional overlays into the `write_trie_changesets` and `write_storage_trie_changesets` provider methods.

The overlay is a TrieUpdates which is used to augment the state of the trie db tables. Using the overlay we can write changesets as if the DB is at a previous block, which will be used during pipeline sync.

Implementing this change required refactoring the
StorageTrieCurrentValuesIter utility to accept a TrieCursor rather than a normal DbCursor. It also required implementing a TrieCursorIter which wraps a TrieCursor into an Iterator, for passing in to `storage_trie_wiped_changeset_iter`. Using both of these changes we could use an InMemoryTrieCursor instead of a direct db cursor.

### DIFF
--- a/crates/storage/provider/src/changesets_utils/trie.rs
+++ b/crates/storage/provider/src/changesets_utils/trie.rs
@@ -83,9 +83,9 @@ where
     }
 }
 
-/// Returns an iterator which produces the values to be inserted into
-/// [`tables::StoragesTrieChangeSets`] for an account whose storage was wiped during a block. It is
-/// expected that this is called prior to inserting the block's trie updates.
+/// Returns an iterator which produces the values to be inserted into the `StoragesTrieChangeSets`
+/// table for an account whose storage was wiped during a block. It is expected that this is called
+/// prior to inserting the block's trie updates.
 ///
 /// ## Arguments
 ///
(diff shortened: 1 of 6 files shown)


crates/storage/provider/src/changesets_utils/trie.rs
```diff
@@ -1,51 +1,42 @@
...
 
     fn seek_cursor(&mut self, path: Nibbles) -> Result<(), DatabaseError> {
-        self.cursor_current = self
-            .cursor
-            .seek_by_key_subkey(self.hashed_address, StoredNibblesSubKey(path))?
-            .map(|e| (e.nibbles.0, e.node));
+        self.cursor_current = self.cursor.seek(path)?;
         Ok(())
     }
 }
```
yongkangc, line 34: do you think it makes sense to bring the abstraction up? since its now a wrapper

yongkangc (approved):
lgtm, i dont have any nits and the pr makes sense to me overall

mattsse (approved):
lgtm!

makes sense to me

PR #18733, opened 2025-09-26
fix: increase backoff timeout

mattsse:
the default for low is 30s which matches the usual ip based throttle

so if we have a peer as trusted but that peer has us not marked as trusted we would always encounter the remote's ip throttle on repeated connection attempts


yongkangc (approved):
ahh makes sense basically we dont want to try too many times before the peer has marked us as trusted

PR #18739, opened 2025-09-26
Bound primary channels in RLPx multiplexer to prevent unbounded memory growth

sashaodessa:
### Summary
Replace unbounded mpsc channels in the RLPx multiplexer primary path with bounded ones to avoid unbounded memory growth and improve backpressure handling.

### Motivation
Unbounded channels can lead to OOM/DoS when the primary protocol lags or a peer floods messages. Bounded queues enforce flow control and protect memory.

### Changes
- Switched `mpsc::unbounded_channel` to `mpsc::channel(1024)` for primary paths.
- Replaced `UnboundedReceiverStream` with `ReceiverStream`.
- Applied non-blocking backpressure:
  - Incoming → primary: `try_send`, drop if full.
  - Primary → wire: `try_send`; return `WouldBlock` to signal backpressure.

### Backpressure Policy
- Inbound congestion is contained by dropping messages when the queue is full (fail-closed on pressure).
- Outbound congestion surfaces as `WouldBlock`, allowing callers to defer sending.


mattsse (commented):
this has been working fine

PR #18740, opened 2025-09-26
feat(stateless): enable test runs to return execution witness

jsign:
This PR adds the following changes:
- Using the `testing` crate as a library now supports returning the `ExecutionWitness` of every block executed during a test case. This is our primary use of this crate as a library in [our repo](https://github.com/eth-act/zkevm-benchmark-workload), used for zkVM benchmarking for Ethereum using EEST benchmark tests.
- Adds support for generating the `ExecutionWitness` for test cases that test for invalid blocks -- it is actually a small change in code, but it is useful for provers to prove a builder might have created an invalid block (apart from allowing to prove all EEST tests, since there are invalid block tests too).

cc @kevaundray


crates/evm/evm/src/execute.rs
```diff
@@ -107,6 +107,23 @@ pub trait Executor<DB: Database>: Sized {
         Ok(BlockExecutionOutput { state: state.take_bundle(), result })
     }
 
+    /// Executes the EVM with the given input and accepts a state closure that is always invoked
+    /// with the EVM state after execution, even after failure.
+    fn execute_with_state_closure_always<F>(
```
jsign, line 112: This is the main change needed for proving invalid blocks. In short, use this method instead of `execute_with_state_closure` (just above), which always return the state diff that happened until the point of block invalidity.

The original use of `execute_with_state_closure` isn't enough since it only returns the diff if the block is valid -- so it's mainly avoiding the early return.

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -54,8 +54,10 @@ impl Suite for BlockchainTests {
 /// An Ethereum blockchain test.
 #[derive(Debug, PartialEq, Eq)]
 pub struct BlockchainTestCase {
-    tests: BTreeMap<String, BlockchainTest>,
-    skip: bool,
+    /// The tests within this test case.
+    pub tests: BTreeMap<String, BlockchainTest>,
+    /// Whether to skip this test case.
+    pub skip: bool,
```
jsign, line 60: Switched these fields to be public since it's useful for us to use this crate as a library.

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -96,39 +98,45 @@ impl BlockchainTestCase {
     }
 
     /// Execute a single `BlockchainTest`, validating the outcome against the
-    /// expectations encoded in the JSON file.
-    fn run_single_case(name: &str, case: &BlockchainTest) -> Result<(), Error> {
+    /// expectations encoded in the JSON file. Returns the list of executed blocks
+    /// with their execution witnesses.
+    pub fn run_single_case(
```
jsign, line 103: TL;DR is that we now return each recovered block and its execution witness for each test case run.

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -186,9 +194,13 @@ impl Case for BlockchainTestCase {
...
 ///
 /// Returns:
-/// - `Ok(())` if all blocks execute successfully and the final state is correct.
-/// - `Err(Error)` if any block fails to execute correctly, or if the post-state validation fails.
-fn run_case(case: &BlockchainTest) -> Result<(), Error> {
+/// - `Ok(_)` if all blocks execute successfully, returning recovered blocks and full block
+///   execution witness.
+/// - `Err(Error)` if any block fails to execute correctly, returning a partial block execution
+///   witness if the error is of variant `BlockProcessingFailed`.
+fn run_case(
```
jsign, line 201: The TL;DR here is:
- For the happy case, we simply return the list of collected blocks and witnesses.
- If the method fails because the block is invalid, now the `Error::block_failed` includes the "partial" execution witness up to that point.

To extend on the last point: if the block is invalid because the block doesn't match consensus rules considering its parent, then the returned execution witness will have only `ancestor` headers, and no `state` nor `bytecodes` since the block validation failed before even reaching that point.

If the block failed during EVM execution, we simply rely on what the new `execute_with_state_closure_always` returned to us.

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -228,11 +240,13 @@ fn run_case(case: &BlockchainTest) -> Result<(), Error> {
...
             .insert_block(block.clone(), StorageLocation::Database)
-            .map_err(|err| Error::block_failed(block_number, err))?;
+            .map_err(|err| Error::block_failed(block_number, Default::default(), err))?;
 
         // Consensus checks before block execution
-        pre_execution_checks(chain_spec.clone(), &parent, block)
-            .map_err(|err| Error::block_failed(block_number, err))?;
+        pre_execution_checks(chain_spec.clone(), &parent, block).map_err(|err| {
+            program_inputs.push((block.clone(), execution_witness_with_parent(&parent)));
+            Error::block_failed(block_number, program_inputs.clone(), err)
```
jsign, line 260: As mentioned before: if the block execution failed in the pre-execution checks then we need to include the ancestor in the execution witness, since the stateless validator needs this data to fail at the same point.

testing/ef-tests/src/result.rs
```diff
@@ -24,6 +27,8 @@ pub enum Error {
     BlockProcessingFailed {
         /// The block number for the block that failed
         block_number: u64,
+        /// The partial program inputs up to and including the block that failed
+        partial_program_inputs: Vec<(RecoveredBlock<Block>, ExecutionWitness)>,
```
mattsse: what does partial mean here?

I find the term program a bit confusing here, what is the program here?

mattsse (commented):
all of this seems fine only have some terminology questions

testing/ef-tests/src/result.rs
jsign: Ah, "program" refers to "guest program" in the context of zkVM proving. (i.e., zkVM prove a guest program). Maybe that's too cryptic in the context of this part of the codebase?

Maybe I could use a more neutral name like: "stateless_exec_inputs"? Wdyt @mattsse ?

mattsse: fine with the partial_program_inputs, but additional context that explains what this means would be helpful

jsign: @mattsse, thanks for the feedback. I explained a bit more in the field comment.

mattsse (approved):

PR #18742, opened 2025-09-26
fix: route account proofs through ProofTaskManager for DB transaction reuse

radik878:
Account proofs were being spawned directly in MultiProofTask without reusing database transactions, unlike storage proofs which go through ProofTaskManager. This caused unnecessary DB transaction overhead.

Added AccountProof task type to ProofTaskManager and refactored spawn_multiproof to queue account proof tasks instead of doing the work inline. Now account proofs reuse the same transaction pool as storage proofs, improving performance and consistency.

Fix #18734


yongkangc:
hey @radik878 we will be doing a major refactor for this, so will pause this pr for now

radik878:
> hey @radik878 we will be doing a major refactor for this, so will pause this pr for now

Ok, thanks for answer

mattsse (commented):
this needs more work not something we can easily do incrementally

PR #18744, opened 2025-09-26
fix(rpc): fix eth_config impl

nethoxa:
There are two changes here:

- Name of the `P256` precompile was wrong, it must be `P256VERIFY` according to other clients (nethermind is wrong too, writing a PR in a moment)
- There was a misunderstanding when handling the case where `next` is `null` (and what the `last` config means). In the EIP it is poorly explained, but in short, it is not the *"previous"* config, but rather the *"final"* config scheduled. That means, if we have configs `[1, 2, 3, 4]` and we are at `2`, `last = 4` and `next = 3`, not `last = 1` as it is the case with reth (the others do it correctly)
- Moreover, it does not handle the case from the EIPs that say:
> "next" and "last" will be null if the client is not configured to support a future fork.
- Sorted the `fork_timestamps` before checking for the current timestamp (should not be a perf drainer as the array is very small), just in case they are unordered and we claim to be on the wrong fork

The `unwrap` on line 138 is safe as, by the check on line 111 we ensure there is at least one element, and with the `else` on line 133 we ensure we return early if that is the case, so that we only continue execution if there are more than 1 element (so that `next = last = current + 1` if there are 2, and `next = current + 1` and last is the last/final configured fork.

### DIFF
--- a/crates/ethereum/node/tests/e2e/rpc.rs
+++ b/crates/ethereum/node/tests/e2e/rpc.rs
@@ -323,7 +323,7 @@ async fn test_eth_config() -> eyre::Result<()> {
 
     let config = provider.client().request_noparams::<EthConfig>("eth_config").await?;
 
-    assert_eq!(config.last.unwrap().activation_time, 0);
+    assert_eq!(config.last.unwrap().activation_time, osaka_timestamp);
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm, turns out we can use the existing .name function here which has the correct name already

PR #18745, opened 2025-09-26
docs: add note to launch on --dev mode

### LINKED ISSUE #18690: ExEx Doc examples don't work with --dev chain
### Describe the change

I'm trying to develop an ExEx with Reth and for that I want to use a dev chain, because I need to index data in a controlled manner before reaching a test chain.

Anyway, the current examples don't work with the --dev or --chain dev command arg and I discovered that is because of the .launch() function in main.rs 

If you change .launch() for .launch_with_debug_capabilities() suddenly --dev works well

Sharing it here, in case someone else has the same issue, and so when you have time... change the ExEx references. I don't know if there are more implications on using .launch_with_debug_capabilities() over .launch() 

Thanks!

### Additional context

Image

mmrrnn:
Closes #18690

- Add tip about using `launch_with_debug_capabilities()` for **--dev** mode
- Add inline comments in code snippets for dev mode guidance
- Explain that regular launcher expects external consensus client

### DIFF
--- a/docs/vocs/docs/snippets/sources/exex/hello-world/src/bin/1.rs
+++ b/docs/vocs/docs/snippets/sources/exex/hello-world/src/bin/1.rs
@@ -2,6 +2,7 @@ use reth_node_ethereum::EthereumNode;
 
 fn main() -> eyre::Result<()> {
     reth::cli::Cli::parse_args().run(async move |builder, _| {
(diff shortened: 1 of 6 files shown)


mattsse (changes requested):
actually, let's just replace .launch with launch_with_debug_capabilities here

mattsse (approved):

PR #18748, opened 2025-09-27
fix(fs-util): preserve original stem for atomic temp files

strmfos:
- append .tmp to the full file name instead of replacing its extension
- clarify the read_link error message to match the underlying operation


mattsse (commented):
just setting the extension is fine here

PR #18749, opened 2025-09-27
test: add missing Drop trait tests for CancelOnDrop

hawkadrian:
Addresses a gap in test coverage by adding two new
test cases that verify the Drop trait implementation for CancelOnDrop:

1. test_cancelondrop_clone_behavior: Verifies that when the original
   CancelOnDrop instance is dropped, all cloned instances correctly
   reflect the cancelled state through the shared Arc<AtomicBool>.

2. test_cancelondrop_multiple_clones: Tests the behavior when one of
   multiple clones is dropped, ensuring that the cancellation signal
   propagates to all remaining instances.

These tests are needed because:
- CancelOnDrop is used in payload building systems where premature
  cancellation could lead to inconsistent state
- The Drop trait behavior was previously untested, creating potential
  for silent bugs in concurrent environments
- The shared state through Arc<AtomicBool> requires explicit testing
  to ensure proper synchronization across clones

### DIFF
--- a/crates/revm/src/cancelled.rs
+++ b/crates/revm/src/cancelled.rs
@@ -108,4 +108,40 @@ mod tests {
         c.cancel();
         assert!(cloned_cancel.is_cancelled());
     }
+
+    #[test]
+    fn test_cancelondrop_clone_behavior() {
+        let cancel = CancelOnDrop::default();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
fine, doesnt hurt

PR #18751, opened 2025-09-27
fix: Normalize CORS domain parsing: trim tokens, ignore empties, enforce wildcard rule

viktorking7:
### Description
- Summary: Normalize comma-separated CORS domains in `create_cors_layer` to make configuration robust and predictable.
- What: Update `crates/rpc/rpc-builder/src/cors.rs` to:
  - Trim each token
  - Ignore empty tokens
  - Check for `*` after normalization
  - Parse via `HeaderValue::from_str`
- Why: Prevent false negatives with spaced lists (e.g., `"https://a.com, https://b.com"`) and consistently reject wildcard in mixed lists.

### DIFF
--- a/crates/rpc/rpc-builder/src/cors.rs
+++ b/crates/rpc/rpc-builder/src/cors.rs
@@ -31,7 +31,7 @@ pub(crate) fn create_cors_layer(http_cors_domains: &str) -> Result<CorsLayer, Co
             // This makes common inputs like "https://a.com, https://b.com" robust.
             let items: Vec<&str> =
                 list.split(',').map(|s| s.trim()).filter(|s| !s.is_empty()).collect();
-    
+
             // After normalization, wildcard is not allowed as part of a list
             if items.iter().any(|o| *o == "*") {
                 return Err(CorsDomainError::WildCardNotAllowed {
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-builder/src/cors.rs
```diff
@@ -26,18 +26,23 @@ pub(crate) fn create_cors_layer(http_cors_domains: &str) -> Result<CorsLayer, Co
...
                 })
             }
 
-            let origins = iter
+            let origins = items
+                .into_iter()
                 .map(|domain| {
-                    domain
-                        .parse::<HeaderValue>()
+                    HeaderValue::from_str(domain)
```
mattsse: this we don't need to change

crates/rpc/rpc-builder/src/cors.rs
```diff
@@ -26,18 +26,23 @@ pub(crate) fn create_cors_layer(http_cors_domains: &str) -> Result<CorsLayer, Co
...
             .allow_origin(Any)
             .allow_headers(Any),
-        _ => {
-            let iter = http_cors_domains.split(',');
-            if iter.clone().any(|o| o == "*") {
+        list => {
+            // Normalize the comma-separated list: trim each token and drop empties
+            // This makes common inputs like "https://a.com, https://b.com" robust.
+            let items: Vec<&str> =
+                list.split(',').map(|s| s.trim()).filter(|s| !s.is_empty()).collect();
```
mattsse, line 33: seems appropriate

mattsse (changes requested):
smol style nit

crates/rpc/rpc-builder/src/cors.rs
viktorking7: @mattsse corrected and tests failed

crates/rpc/rpc-builder/src/cors.rs
mattsse, line 45: this is the same thing we can keep from_str here

PR #18752, opened 2025-09-27
perf(proof_task): replace blocking proof task with thread pool

Rhovian:
Replaces on-demand `tokio::spawn_blocking` proof tasks with a dedicated thread pool where each worker owns a long-lived, exclusive database transaction.

**changes**
- On first use, spawn a fixed thread pool (`max_concurrency` workers)
- Each worker creates its own read-only DB transaction at startup and reuses it for its lifetime
- Tasks are dispatched via a lock-free crossbeam-channel (MPMC), eliminating mutex contention
- Removed all transaction borrowing/returning logic and associated message types

**validation**
- All existing correctness tests pass (parallel vs. sequential proof equivalence)
- Benchmarks (cargo bench --bench root) show no performance regression; a slight improvement (~1%) was observed at 10k accounts in local testing

This is a larger diff than I’d typically aim for in a first PR, so I welcome feedback—happy to discuss or adjust any part of the implementation.

cc @shekhirin 

closes #18735


crates/trie/parallel/src/proof_task.rs
```diff
@@ -48,34 +48,84 @@ use crate::proof_task_metrics::ProofTaskMetrics;
...
+    work_sender: Sender<ProofTaskKind>,
+    _workers: Vec<thread::JoinHandle<()>>,
+    _phantom: PhantomData<Factory>,
+}
+
+impl<Factory> ProofThreadPool<Factory>
+where
+    Factory: DatabaseProviderFactory<Provider: BlockReader> + Send + Sync + 'static,
+{
+    fn new(
```
yongkangc, line 65: maybe we can include a short docstring about 
- creates workers
- spawning a thread that listens to tasks and dispatches it to workers
?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -48,34 +48,84 @@ use crate::proof_task_metrics::ProofTaskMetrics;
...
+        num_workers: usize,
+        view: ConsistentDbView<Factory>,
+        task_ctx: ProofTaskCtx,
+    ) -> ProviderResult<Self> {
+        let (work_sender, work_receiver) = unbounded();
+        let mut workers = Vec::with_capacity(num_workers);
+
+        for worker_id in 0..num_workers {
+            // Each worker gets its own dedicated database transaction
+            let provider_ro = view.provider_ro()?;
```
yongkangc: If a worker panics, would the db txn and the worker get drop and wouldnt be replaced? i.e pool capacity reduced

crates/trie/parallel/src/proof_task.rs
Rhovian: docstring added

crates/trie/parallel/src/proof_task.rs
Rhovian: Good catch, I've added a recovery mechanism where each worker is managed by a supervisor thread that respawns the worker with a fresh database transaction after any panic. This still needs to be tested.
  
some other stuff to consider:

- Workers respawn immediately and infinitely. This could hide systematic bugs in a tight panic loop. I considered adding exponential backoff + circuit breaker but kept it minimal for this PR to avoid scope creep.
- If `provider_ro()` fails during respawn, the supervisor exits permanently (logged at
  ERROR). Could add active worker count tracking for better observability.
- Tasks in-flight during panic are dropped, but this should be acceptable since the system
  already tolerates send failures throughout?
  
I see you assigned the parent issue to yourself - let me know if I should discontinue work here or if there's a different approach you'd prefer.

yongkangc: @Rhovian i unassigned myself from the issue - will try to get your pr in instead as it aligns with the design we have in mind

Rhovian: happy to continue working on this then, will fix the ci lint issues for now and put some thought into testing the new functionality

crates/trie/parallel/src/proof_task.rs
```diff
@@ -48,34 +48,133 @@ use crate::proof_task_metrics::ProofTaskMetrics;
...
+    }
+
+    /// Spawns a resilient worker that will respawn itself on panic
+    fn spawn_worker(
+        worker_id: usize,
+        view: ConsistentDbView<Factory>,
+        task_ctx: ProofTaskCtx,
+        work_receiver: Receiver<ProofTaskKind>,
+    ) -> ProviderResult<()> {
+        thread::spawn(move || {
```
mediocregopher, line 91: @mattsse @yongkangc I believe in the design we've discussed the "pools" would be pools of tokio tasks spawned from the same runtime thread pool, not actual OS threads. Just wanted to confirm both that's still the plan and I'm not misunderstanding what's happening here.

crates/trie/parallel/src/proof.rs
```diff
@@ -10,11 +10,11 @@ use alloy_primitives::{
     B256,
 };
 use alloy_rlp::{BufMut, Encodable};
+use crossbeam_channel::Receiver;
```
mattsse, line 13: I wanted to use crossbeam channels for this for a while but never felt necessary because we only have a few hundred or thousand messages here, but with more things to fetch this is definitely beneficial I believe, so we should use crossbeam here

crates/trie/parallel/src/proof_task.rs
yongkangc: yup, good note ideally we want to make use of the tokio tasks (since its i/o) which we spawned during runtime. 

and acc / storage proofs would have their own separate tasks

mattsse (commented):
tysm for this @Rhovian 

this aligns well with what we had in mind for this as well, need to think about how to incorporate this with other changes that we want to do

crates/trie/parallel/src/proof_task.rs
mattsse: rn these dedicated threads are short lived, because spawned per block, in 1s blocktime land this is likely suboptimal.
we could either rely on the tokio blocking pool that auto scales or make this a dedicated pool of workers that we use across blocks.
perhaps just using tokio for this is sufficient, although dedicated threads make sense here as well, afaik tokio's blocking pool has some overhead for async integration

crates/trie/parallel/src/proof_task.rs
```diff
@@ -87,92 +186,73 @@ impl<Factory: DatabaseProviderFactory> ProofTaskManager<Factory> {
...
-            }
-            ProofTaskKind::BlindedAccountNode(path, sender) => {
-                proof_task_tx.blinded_account_node(path, sender, tx_sender);
-            }
-            ProofTaskKind::BlindedStorageNode(account, path, sender) => {
-                proof_task_tx.blinded_storage_node(account, path, sender, tx_sender);
-            }
-        });
+        // Dispatch the task to the thread pool
+        if let Err(send_error) = thread_pool.dispatch_task(task) {
```
shekhirin, line 247: this can only return an error if the sender is disconnected, which is not recoverable I believe?

crates/trie/parallel/src/proof_task.rs
yongkangc: maybe we could spawn on tokio's blocking pool, but long-lived?

Rhovian: long-lived tokio blocking tasks would essentially be dedicated threads with extra steps - we'd lose tokio's dynamic scaling while still paying for its scheduling overhead

Rhovian: Since the workflow is steady-state block production at Ns intervals (correct me if im wrong), the consistent load pattern would favor dedicated threads over dynamic scaling, right?

crates/trie/parallel/src/proof_task.rs
Rhovian: @shekhirin You're right - send fails when all receivers are dropped (channel closed). This happens when:
- All workers exit (due to shutdown or persistent DB failure)
- Since supervisors respawn workers on panic, this is only possible during catastrophic failure

Currently re-queuing would create an infinite retry. Should just log and drop the task since it's non-recoverable?

I've made clippy happy and merged main. Will defer to you since you were assigned - lmk if there's anything I can do from here.

crates/trie/parallel/src/proof_task.rs
mediocregopher: I think the main concern is that we don't want to spawn the threads per-block. Using tokio gives us an easy thread-pool we can re-use between blocks, but yes with extra overhead. This pool as it is now can't be used between blocks, it is tied to the block by the ConsistentDbView + ProofTaskCtx, and it would require a much bigger refactor to work around that I believe.

crates/trie/parallel/src/proof_task.rs
shekhirin: I think we should return an error instead, because this is not recoverable and can cause an infinite loop that you've mentioned.

crates/trie/parallel/src/proof_task.rs
shekhirin: IMO an ideal solution is to have long-running std threads, but as @mediocregopher it's a bigger refactor, so for now I propose to stick to using Tokio blocking tasks

crates/trie/parallel/src/proof_task.rs
Rhovian: @shekhirin want me to implement both the tokio blocking tasks refactor and the error return, or leaving it to you?

PR #18753, opened 2025-09-27
refactor: make DatabaseProof trait stateful

### LINKED ISSUE #15476: Make DatabaseProof trait stateful
### Describe the feature

Right now `DatabaseProof` has an associated `TX` type and takes `&'a Self::TX` on every method:
https://github.com/paradigmxyz/reth/blob/db269f3e5375680a9d7f16f3eb7ec23421b04ee6/crates/trie/db/src/proof.rs#L13-L32

This is not necessary, instead we can make this into a regular trait that is stateful and takes `&self`. For example it would look like:
```
 /// Extends [`Proof`] with operations specific for working with a database transaction. 
 pub trait DatabaseProof { 
     /// Generates the state proof for target account based on [`TrieInput`]. 
     fn overlay_account_proof( 
         &self, 
         input: TrieInput, 
         address: Address, 
         slots: &[B256], 
     ) -> Result<AccountProof, StateProofError>; 
  
     /// Generates the state [`MultiProof`] for target hashed account and storage keys. 
     fn overlay_multiproof( 
         &self, 
         input: TrieInput, 
         targets: MultiProofTargets, 
     ) -> Result<MultiProof, StateProofError>; 
 } 
```

And the implementers (like `Proof` etc) can have a `from_tx` method on the struct itself

### Additional context

_No response_

avorylli:
Refactor DatabaseProof trait to be stateful by removing lifetime parameters and making methods take &self instead of passing tx to every method call.

## Changes
- Remove lifetime parameters from DatabaseProof trait
- Add from_tx method to trait for instance creation
- Add associated type Tx for transaction typing
- Update all implementations and usage sites
- Add From implementations for cursor factories

Fixes #15476


mediocregopher:
This looks like it's headed in the right direction, can you clean up the CI errors?

crates/trie/db/src/proof.rs
```diff
@@ -11,69 +11,72 @@ use reth_trie::{
...
     /// Generates the state [`MultiProof`] for target hashed account and storage keys.
     fn overlay_multiproof(
-        tx: &'a TX,
+        &self,
         input: TrieInput,
         targets: MultiProofTargets,
     ) -> Result<MultiProof, StateProofError>;
+
+    /// Associated type for the database transaction.
+    type Tx;
```
mediocregopher: Please define the associated type at the top of the trait 🙏

crates/trie/db/src/trie_cursor.rs
```diff
@@ -22,10 +22,14 @@ impl<T> DatabaseTrieCursorFactory<T> {
     }
 }
 
-impl<TX> TrieCursorFactory for DatabaseTrieCursorFactory<&TX>
-where
-    TX: DbTx,
-{
+impl<'a, TX> From<&'a TX> for DatabaseTrieCursorFactory<&'a TX> {
```
mediocregopher: Where is this getting used?

crates/trie/db/src/proof.rs
```diff
@@ -11,69 +11,72 @@ use reth_trie::{
...
+        &self,
         input: TrieInput,
         address: Address,
         slots: &[B256],
     ) -> Result<AccountProof, StateProofError> {
         let nodes_sorted = input.nodes.into_sorted();
         let state_sorted = input.state.into_sorted();
-        Self::from_tx(tx)
+        Self::new(self.trie_cursor_factory().clone(), self.hashed_cursor_factory().clone())
             .with_trie_cursor_factory(InMemoryTrieCursorFactory::new(
```
mediocregopher: These calls to `with_trie_cursor_factory` and `with_hashed_cursor_factory` are no longer necessary since you're passing the factories into `Self::new`

crates/trie/db/src/proof.rs
```diff
@@ -11,69 +11,72 @@ use reth_trie::{
...
-        tx: &'a TX,
+        &self,
         input: TrieInput,
         targets: MultiProofTargets,
     ) -> Result<MultiProof, StateProofError> {
         let nodes_sorted = input.nodes.into_sorted();
         let state_sorted = input.state.into_sorted();
-        Self::from_tx(tx)
+        Self::new(self.trie_cursor_factory().clone(), self.hashed_cursor_factory().clone())
             .with_trie_cursor_factory(InMemoryTrieCursorFactory::new(
```
mediocregopher: Same as with overlay_account_proof, these aren't really necessary

mediocregopher (changes requested):
A couple small changes requested, also this will likely conflict with #19114 so that'll require some fixing up

mediocregopher (approved):

PR #18754, opened 2025-09-27
chore(evm): mark ExecuteOutput as unused and slated for removal

Forostovec:
Marked reth_evm::execute::ExecuteOutput as deprecated with guidance to use reth_execution_types::BlockExecutionResult or BlockExecutionOutput.
Added doc note explaining the type is unused internally and will be removed in the next major release.


mattsse (approved):
seems okay, useful to mark it as deprecated

PR #18756, opened 2025-09-27
chore: remove unused Debug import

MozirDmitriy:
Remove unused std::fmt::Debug import from crates/engine/tree/src/download.rs.
The import was not referenced anywhere in the file.


mattsse (commented):
this is actually required

PR #18759, opened 2025-09-28
fix: use consistent RNG source for ECIES nonce generation

viktorking7:
### Description

**Problem**: ECIES initialization was mixing randomness sources by using `B256::random()` for nonce generation while using a local RNG for ephemeral keys. This creates inconsistent entropy sources in cryptographic operations.

**Solution**: Generate nonce using the same local RNG instance that's already created for ephemeral key generation.

**Changes**:
- Replace `B256::random()` with `rng.fill_bytes()` in `new_client()` and `new_server()`
- Add `RngCore` import for `fill_bytes()` method

### DIFF
--- a/crates/net/ecies/src/algorithm.rs
+++ b/crates/net/ecies/src/algorithm.rs
@@ -15,7 +15,7 @@ use alloy_rlp::{Encodable, Rlp, RlpEncodable, RlpMaxEncodedLen};
 use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
 use ctr::Ctr64BE;
 use digest::{crypto_common::KeyIvInit, Digest};
-use rand_08::{thread_rng as rng, Rng};
+use rand_08::{thread_rng as rng, Rng, RngCore};
 use reth_network_peers::{id2pk, pk2id};
 use secp256k1::{
     ecdsa::{RecoverableSignature, RecoveryId},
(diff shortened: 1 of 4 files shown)


crates/net/ecies/src/algorithm.rs
```diff
@@ -312,9 +312,11 @@ impl ECIES {
 
     /// Create a new ECIES client with the given static secret key and remote peer ID.
     pub fn new_client(secret_key: SecretKey, remote_id: PeerId) -> Result<Self, ECIESError> {
-        // TODO(rand): use rng for nonce
+        // NOTE: use the same RNG for nonce to avoid mixing randomness sources
         let mut rng = rng();
-        let nonce = B256::random();
+        let mut nonce_bytes = [0u8; 32];
+        rng.fill_bytes(&mut nonce_bytes);
+        let nonce = B256::from(nonce_bytes);
```
mattsse: this makes sense, but please remove the comment, this isnt useful without the context of this pr

crates/net/nat/src/lib.rs
mattsse, line 1: please undo and rebase instead

mattsse (changes requested):
smol nit and needs a rebase

crates/net/ecies/src/algorithm.rs
```diff
@@ -356,7 +357,10 @@ impl ECIES {
     /// Create a new ECIES server with the given static secret key.
     pub fn new_server(secret_key: SecretKey) -> Result<Self, ECIESError> {
         let mut rng = rng();
-        let nonce = B256::random();
+        // NOTE: use the same RNG for nonce to avoid mixing randomness sources
```
mattsse: ...............

mattsse (changes requested):
please undo all other unrelated changes

mattsse (commented):
we dont necessarily need this

PR #18762, opened 2025-09-28
fix: use cryptographically secure RNG for node secret key generation

sashaodessa:
**Description:**
Replace `thread_rng()` with `OsRng` in `rng_secret_key()` function to ensure cryptographically secure entropy for node identity keys.

**Changes:**
- Use `rand_08::rngs::OsRng` instead of `rand_08::thread_rng()`

**Why:**
`thread_rng()` is not guaranteed to be cryptographically secure and may use predictable PRNGs in some environments. For long-term node identity keys, we should use OS-backed CSPRNG to ensure proper entropy.


mattsse (commented):
@DaniPopes wdyt here?

DaniPopes:
This is wrong, thread_rng is cryptographically secure

PR #18770, opened 2025-09-29
feat(op-reth): custom FlashBlock decoder from bytes

0xMelkor:
It's now possible to specify an adapter type for flashblocks.

```rust
/// A trait for decoding flashblocks from bytes.
pub trait FlashBlockDecoder: Send + 'static {
    /// Decodes `bytes` into a [`FlashBlock`].
    fn decode(&self, bytes: bytes::Bytes) -> eyre::Result<FlashBlock>;
}
```

```rust
let stream = WsFlashBlockStream::new(ws_url).with_decoder(MyDecoder);
```

### DIFF
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -1,7 +1,8 @@
 //! A downstream integration of Flashblocks.
 
 pub use payload::{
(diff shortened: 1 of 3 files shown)


mattsse (approved):
seems fine if this unlocks custom formats

PR #18772, opened 2025-09-29
fix: prevent integer overflow in e2store entry data length

avorylli:
Fix potential integer overflow when converting usize to u32 in e2store file operations.

Changes:
- Add safe type conversion using TryFrom in Entry::write() method
- Add safe type conversion in Era1Reader::read() for offset count
- Return descriptive error messages when data exceeds u32::MAX

Why this change is necessary:

The e2store file format specification limits entry data length to u32::MAX (4GB).

Previously, direct casting from usize to u32 could cause silent integer overflow on 64-bit systems when data.len() > u32::MAX, leading to incorrect header values and potential data corruption.


mattsse (commented):
this isnt necessary, we dont expect > u32 entries here

PR #18773, opened 2025-09-29
feat(eth-wire): implement disconnect on Status message after handshake

DeVikingMark:
Implements automatic peer disconnection when attempting to send Status messages after handshake completion.

- Replaces TODO comment with actual disconnect implementation
- Uses CanDisconnect trait to gracefully disconnect with ProtocolBreach reason
- Adds test to verify disconnect behavior


mattsse (approved):
this isnt perfect, as outlined in the docs, but still better than what we have on main

PR #18776, opened 2025-09-29
feat(flashblocks): additional pending flashblock data

dustinjake:
## Description

Amend the data kept by the pending flashblock with the last flashblock index and block hash.

### DIFF
--- a/crates/optimism/flashblocks/src/payload.rs
+++ b/crates/optimism/flashblocks/src/payload.rs
@@ -1,8 +1,11 @@
 use alloy_eips::eip4895::Withdrawal;
 use alloy_primitives::{Address, Bloom, Bytes, B256, U256};
 use alloy_rpc_types_engine::PayloadId;
+use derive_more::Deref;
+use reth_node_api::NodePrimitives;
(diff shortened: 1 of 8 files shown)


crates/optimism/flashblocks/src/payload.rs
```diff
@@ -119,3 +122,27 @@ impl From<ExecutionPayloadBaseV1> for OpNextBlockEnvAttributes {
...
+/// Flashblock.
+#[derive(Debug, Clone, Deref)]
+pub struct PendingFlashBlock<N: NodePrimitives> {
+    /// The complete pending block built out of all received Flashblocks.
+    #[deref]
+    pub pending: PendingBlock<N>,
+    /// A sequential index that identifies the last Flashblock added to this block.
+    pub last_flashblock_index: u64,
+    /// The last Flashblock block hash,
+    pub last_flashblock_hash: B256,
```
mattsse, line 136: why is this necessary if we already have the block hash in `PendingBlock`

mattsse (approved):
lgtm, but I think we dont necessarily need the flashblock hash, but also doesnt hurt

PR #18781, opened 2025-09-30
refactor: remove redundant Setup::new() method

futreall:
Remove the redundant Setup::new() method from e2e-test-utils Setup struct. The method was simply calling Self::default(), which is unnecessary as users can directly call Setup::default(). This cleanup reduces code duplication and follows DRY principles. All existing code already uses Setup::default() pattern, confirming the method was unused.


fgimenez (approved):
lgtm

PR #18785, opened 2025-09-30
fix: remove the leading hash comparison from RecoveredBlock<B>::PartialEq.

GarmashAlex:
This check is redundant because SealedBlock<B> derives PartialEq and its first field, SealedHeader, implements PartialEq by comparing only the header hash. Therefore, self.block == other.block already verifies header hash equality and proceeds to body comparison. Keeping an additional self.hash_ref() == other.hash_ref() causes duplicate work and potential double lazy-hash initialization without changing semantics.


mattsse (approved):
checks out

PR #18788, opened 2025-09-30
feat(storage): read headers and transactions only from static files

### LINKED ISSUE #18684: Read headers and transactions only from static files
### Describe the feature

Remove all functionality from provider that handles checking both locations to determine where to read from.

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/18684

This PR makes all provider methods that read headers or transactions to read only from static files, and never make a lookup in the database. Receipts are not touched yet, as they still can be written to the database.

This doesn't break existing nodes, because on startup we always copy to static files https://github.com/paradigmxyz/reth/blob/aa192c255b53335ba509987c82ab26e79e954195/crates/node/builder/src/launch/engine.rs#L159-L160

Hive passes https://github.com/paradigmxyz/reth/actions/runs/18220958085

### DIFF
--- a/crates/net/downloaders/Cargo.toml
+++ b/crates/net/downloaders/Cargo.toml
@@ -59,6 +59,7 @@ reth-ethereum-primitives.workspace = true
 reth-chainspec.workspace = true
 reth-consensus = { workspace = true, features = ["test-utils"] }
 reth-network-p2p = { workspace = true, features = ["test-utils"] }
+reth-provider = { workspace = true, features = ["test-utils"] }
 reth-testing-utils.workspace = true
 reth-tracing.workspace = true
 
--- a/crates/net/downloaders/Cargo.toml
+++ b/crates/net/downloaders/Cargo.toml
@@ -22,9 +22,8 @@ reth-storage-api.workspace = true
 reth-tasks.workspace = true
 
 # optional deps for the test-utils feature
-reth-db = { workspace = true, optional = true }
-reth-db-api = { workspace = true, optional = true }
 reth-ethereum-primitives = { workspace = true, optional = true }
+reth-provider = { workspace = true, optional = true }
 reth-testing-utils = { workspace = true, optional = true }
 
 # ethereum
@@ -58,11 +57,8 @@ itertools.workspace = true
 async-compression = { workspace = true, features = ["gzip", "tokio"] }
 reth-ethereum-primitives.workspace = true
 reth-chainspec.workspace = true
-reth-db = { workspace = true, features = ["test-utils"] }
-reth-db-api.workspace = true
 reth-consensus = { workspace = true, features = ["test-utils"] }
 reth-network-p2p = { workspace = true, features = ["test-utils"] }
-reth-provider = { workspace = true, features = ["test-utils"] }
 reth-testing-utils.workspace = true
 reth-tracing.workspace = true
 
@@ -76,13 +72,10 @@ default = []
 file-client = ["dep:async-compression", "dep:alloy-rlp"]
 test-utils = [
     "tempfile",
-    "reth-db-api",
-    "reth-db/test-utils",
     "reth-consensus/test-utils",
     "reth-network-p2p/test-utils",
     "reth-testing-utils",
     "reth-chainspec/test-utils",
-    "reth-db-api?/test-utils",
     "reth-provider/test-utils",
     "reth-primitives-traits/test-utils",
     "dep:reth-ethereum-primitives",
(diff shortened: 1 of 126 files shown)


crates/stages/stages/src/test_utils/test_db.rs
```diff
@@ -103,6 +103,11 @@ impl TestStageDB {
         })
     }
 
+    /// Return the number of entries in the table or static file segment
+    pub fn count_entries<T: Table>(&self) -> ProviderResult<usize> {
+        self.factory.provider()?.count_entries::<T>()
+    }
```
yongkangc, line 109: ahh this is a nice refactor

crates/net/downloaders/Cargo.toml
```diff
@@ -76,13 +73,10 @@ default = []
 file-client = ["dep:async-compression", "dep:alloy-rlp"]
 test-utils = [
     "tempfile",
-    "reth-db-api",
-    "reth-db/test-utils",
```
yongkangc, line 80: curious why do we no longer need this?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1324,66 +1244,7 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> TransactionsProviderExt
...
-                        let (tx_hash, tx_id) = tx.map_err(|boxed| *boxed)?;
-                        tx_list.push((tx_hash, tx_id));
-                    }
-                }
-
-                Ok(tx_list)
-            },
-            |_| true,
-        )
+        self.static_file_provider.transaction_hashes_by_range(tx_range)
```
yongkangc, line 1247: with this we can basically do the same thing but without the db? 

this seems great, curious why did we not do this before?

yongkangc (approved):
overall lgtm

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1118,7 +1118,11 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         };
 
         let mut result = Vec::with_capacity((range.end - range.start).min(100) as usize);
-        let mut provider = get_provider(range.start)?;
+        let mut provider = match get_provider(range.start) {
+            Ok(provider) => provider,
+            Err(ProviderError::MissingStaticFileBlock(_, _)) => return Ok(vec![]),
```
klkvr: should this also handle `MissingStaticFileTx`?

shekhirin: good catch, also check https://github.com/paradigmxyz/reth/pull/18788/commits/3338eb772f7c75c6a9902e07d8908e4ddd5633e0

this matches the existing behaviour of MDBX cursor https://github.com/paradigmxyz/reth/blob/3338eb772f7c75c6a9902e07d8908e4ddd5633e0/crates/storage/db-api/src/cursor.rs#L263-L290

crates/net/downloaders/Cargo.toml
shekhirin: checking static files instead of the database now

crates/storage/provider/src/providers/database/provider.rs
shekhirin: no good reason tbh, we were writing to both locations for a while now...

klkvr (approved):

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1140,27 +1155,15 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                     }
                     None => {
                         if retrying {
-                            warn!(
-                                target: "provider::static_file",
-                                ?segment,
-                                ?number,
-                                "Could not find block or tx number on a range request"
-                            );
```
Rjected, line 1148: why do we go from warning and returning an err, to returning an Ok(result) now?

Rjected (approved):
lgtm, I just have one question

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: we never hit this error because we always query only what we know is inside static files https://github.com/paradigmxyz/reth/blob/3338eb772f7c75c6a9902e07d8908e4ddd5633e0/crates/storage/provider/src/providers/static_file/manager.rs#L1288-L1294

but now we're querying methods such as `self.static_file_provider.headers_range` directly without checking the ranges first, so I mirror the behaviour of database range querying

PR #18789, opened 2025-09-30
chore: drop unused once_cell dependency from reth-primitives

eeemmmmmm:
Closes #18705  (follow-up)

- addresses the previous review feedback from #18705 (we keep using `reth_primitives_traits::sync::{LazyLock, OnceLock}` so `no_std` remains supported)
- keeps the change scoped to `reth-primitives` only: no `Cargo.lock` touch, other crates still depend on `once_cell`


mattsse (commented):
this still does not work

PR #18794, opened 2025-09-30
fix(prune): Use checked_sub and preserve Option in ReceiptsByLogs checkpoint

radik878:
- Replace saturating_sub with checked_sub when computing last_pruned_block to preserve None on underflow (meaning “block 0 not finished yet”) per PruneCheckpoint contract.
- Save checkpoint with block_number: last_pruned_block.map(|b| prune_mode_block.min(b)) instead of forcing Some(...), maintaining the correct semantics and aligning with other segments (e.g., receipts, transaction_lookup).
- This prevents incorrectly marking pruning as finished up to block 0 and ensures consistent checkpoint behavior across segments.


mattsse (commented):
not needed here

PR #18797, opened 2025-09-30
fix: poll the pinger timeout Sleep future

futreall:
Polls the pinger timeout future to register the waker and ensure ping timeouts fire reliably

### DIFF
--- a/crates/net/eth-wire/src/pinger.rs
+++ b/crates/net/eth-wire/src/pinger.rs
@@ -1,5 +1,6 @@
 use crate::errors::PingerError;
 use std::{
+    future::Future,
     pin::Pin,
     task::{Context, Poll},
     time::Duration,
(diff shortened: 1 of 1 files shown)


crates/net/eth-wire/src/pinger.rs
```diff
@@ -77,7 +78,7 @@ impl Pinger {
                 }
             }
             PingState::WaitingForPong => {
-                if self.timeout_timer.is_elapsed() {
+                if self.timeout_timer.as_mut().poll(cx).is_ready() {
```
Rjected, line 81: how is this different / better than using `is_elapsed`?

mattsse, line 81: this shouldn't be necessary be the reset call should register in the time driver and fire I believe,

however this change is better, because it ensures the waker is registered

mattsse (approved):
makes sense

crates/net/eth-wire/src/pinger.rs
Rjected: ah I see, interesting

PR #18798, opened 2025-09-30

crStiv:


mattsse (commented):
this includes a few things that i think we shouldnt change

PR #18807, opened 2025-10-01
feat: impl Debug for FnLauncher

sky-coderay:
The `FnLauncher` is used extensively in node launch contexts where Debug trait is expected (similar to how `Ext: clap::Args + fmt::Debug` is required throughout the CLI system). This implementation will improve error reporting when launcher failures occur.


mattsse (approved):
reasonable

PR #18809, opened 2025-10-01
feat(trie): MerkleChangeSets pipeline sync stage

mediocregopher:
Closes #18464 

This implements a new sync stage for computing the Accounts/StoragesTrieChangeSets tables. Only changesets up to the finalized block (or 64 blocks ago, if no block is marked finalized) are generated, as these are the only ones necessary for the Engine API simplifications which this change is a part of.

Unwinding is implemented by simply clearing the changesets tables past the target block.


crates/stages/api/src/pipeline/mod.rs
```diff
@@ -572,14 +572,18 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
...
                     // invalid place.
-                    let provider_rw = self.provider_factory.database_provider_rw()?;
-                    provider_rw.save_stage_checkpoint_progress(StageId::MerkleExecute, vec![])?;
-                    provider_rw.save_stage_checkpoint(
-                        StageId::MerkleExecute,
-                        prev_checkpoint.unwrap_or_default(),
-                    )?;
-
-                    UnifiedStorageWriter::commit(provider_rw)?;
+                    // Only reset MerkleExecute checkpoint if MerkleExecute itself failed
```
mediocregopher, line 575: Without this check a state root error in MerkleChangeSets would result in the merkle checkpoint being cleared.

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+                            "Incremental state root failed! {INVALID_STATE_ROOT_ERROR_MESSAGE}");
+                    StageError::Fatal(Box::new(e))
+                },
+            )?;
+
+        let block = provider
+            .header_by_number(block_number)?
+            .ok_or_else(|| ProviderError::HeaderNotFound(block_number.into()))?;
+
+        let header = SealedHeader::seal_slow(block);
```
mattsse: do we always need the block's hash here?

if not we can init this lazily

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+        // range. The full HashedPostState revert for block N can be calculated as:
+        //
+        //
+        // ```
+        // // where `extend` overwrites any shared keys
+        // state_revert(N) = state_revert(N + 1).extend(per_block_state_revert(N))
+        // ```
+        //
+        // We need per-block reverts to calculate the prefix set for each individual block. By using
+        // the per-block reverts to calculate full reverts on-the-fly we can save a bunch of memory.
```
mattsse: this is very helpful

mattsse (commented):
not the expert when it comes to lower level merkle stuff, but I believe I could follow along here and thanks to the docs all of this makes sense to me

still need @Rjected and @shekhirin for reviews here

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+use tracing::{debug, error};
+
+/// The `MerkleChangeSets` stage.
+///
+/// This stage processes and maintains trie changesets from the finalized block to the latest block.
+#[derive(Debug, Clone)]
+pub struct MerkleChangeSets {
+    /// The finalized block height to use as a fallback when the finalized block is not found.
+    /// Defaults to 64 (2 epochs in beacon chain).
+    finalized_block_height: u64,
```
shekhirin: would prefer to name it something like `retention_blocks`, so that we can make it configurable via the config and it doesn't sound confusing to refer to it as `finalized_block_height`

crates/node/core/src/args/stage.rs
```diff
@@ -38,6 +38,11 @@ pub enum StageEnum {
     ///
     /// Handles Merkle tree-related computations and data processing.
     Merkle,
+    /// The merkle changesets stage within the pipeline.
+    ///
+    /// Handles Merkle trie changesets for storage and accounts.
+    #[value(name = "merkle-changesets")]
+    MerkleChangeSets,
```
yongkangc, line 45: this is really useful

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+                },
+            )?;
+
+        let block = provider
+            .header_by_number(block_number)?
+            .ok_or_else(|| ProviderError::HeaderNotFound(block_number.into()))?;
+
+        let header = SealedHeader::seal_slow(block);
+
+        let (got, expected) = (root, header.state_root());
```
yongkangc: so if i am understanding correctly 

is got - what we compute using the trie updates, and expected the trie updates that we are getting from sync?

yongkangc: would this help us with debugging during sr mismatch?

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+        for block_number in target_range.clone() {
+            per_block_state_reverts.push(HashedPostState::from_reverts::<KeccakKeyHasher>(
+                provider.tx_ref(),
+                block_number..=block_number,
+            )?);
+        }
+
+        let per_block_state_revert = |block_number| -> &HashedPostState {
+            &per_block_state_reverts[(block_number - target_start) as usize]
+        };
```
yongkangc: might be a nit from my side, but feels like this closure syntax is not immediately intuitive? 

i.e retries state revert data for given block from array 

maybe a docstring / bringing this out as a pure function could help?

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+        let merkle_checkpoint = provider
+            .get_stage_checkpoint(StageId::MerkleExecute)?
+            .map(|checkpoint| checkpoint.block_number)
+            .unwrap_or(0);
+
+        if input.target.is_none_or(|target| merkle_checkpoint != target) {
+            return Err(StageError::Fatal(eyre::eyre!("Cannot sync stage to block {:?} when MerkleExecute is at block {merkle_checkpoint:?}", input.target).into()))
+        }
+
+        // Determine the target range for changeset computation
```
yongkangc: not sure if this comment might be needed? seems quite intuitive from func already!

yongkangc (commented):
overall logic flows quite easily for me which is great! left some commnets. and questions

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,374 @@
...
+            target: "sync::stages::merkle_changesets",
+            ?target_start,
+            "Computing trie state at starting block",
+        );
+        let mut input = TrieInput::default();
+        input.state = state_revert(target_start);
+        input.prefix_sets = input.state.construct_prefix_sets();
+        // target_start will be >= 1, see `determine_target_range`.
+        input.nodes =
+            Self::calculate_block_trie_updates(provider, target_start - 1, input.clone())?;
```
shekhirin, line 208: can introduce a method `TrieInput::from_state_and_nodes` or smth that will be similar to `TrieInput::from_state`

crates/stages/stages/src/stages/merkle_changesets.rs
mediocregopher: Both `got` and `expected` are state roots. Under the hood we're using StateRoot, which performs two functions: calculating the state root for a block, and producing the trie updates for it. So `got` is the state root we received while calculating the trie updates, and `expected` is the block's actual state root from its header.

This check isn't strictly necessary, we could just ignore `got` and use the trie updates, but it seemed prudent to add the check since we have the state root anyway.

crates/stages/stages/src/stages/merkle_changesets.rs
mediocregopher: I updated the closure naming and implementation a bit to make them clearer. I don't think pulling them into actual functions would be super helpful; ultimately they are just simple helpers which operate on `per_block_state_reverts`, which only exists in this scope. But lmk what you think, we can try it as pure functions too.

crates/stages/stages/src/stages/mod.rs
```diff
@@ -16,6 +16,8 @@ mod index_account_history;
 mod index_storage_history;
 /// Stage for computing state root.
 mod merkle;
+/// Stage for managing merkle changesets.
```
shekhirin:
```suggestion
/// Stage for computing merkle changesets.
```

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,376 @@
...
+        // ------------------------------> Block #
+        //    |------prev-----|
+        //              |-----target-----|
+        //                    |--actual--|
+        //
+        // However, if the target start is less than the previously computed start, we don't want to
+        // do this, as it would leave a gap of data at `target_start..=computed_range.start`.
+        //
+        // ------------------------------> Block #
+        //           |---prev---|
```
shekhirin: just for the consistency with variable names
```suggestion
        //    |------computed-----|
        //              |-----target-----|
        //                    |--actual--|
        //
        // However, if the target start is less than the previously computed start, we don't want to
        // do this, as it would leave a gap of data at `target_range.start..=computed_range.start`.
        //
        // ------------------------------> Block #
        //           |---computed---|
```

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,376 @@
...
+        let target_end = merkle_checkpoint + 1; // exclusive
+
+        // Calculate the target range based on the finalized block and the target block.
+        // We maintain changesets from the finalized block to the latest block.
+        let finalized_block = provider.last_finalized_block_number()?;
+
+        // If finalized block is not found, default to being `retention_blocks` away from
+        // target_end
+        let mut target_start = finalized_block
+            .unwrap_or_else(|| merkle_checkpoint.saturating_sub(self.retention_blocks))
```
shekhirin: should this instead calculate the target start based on the merkle checkpoint block? Like `merkle_checkpoint - self.retention_blocks`

crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -0,0 +1,376 @@
...
+            if computed_range.start == 0 {
+                computed_range.start = target_range.start;
+            }
+        } else {
+            // If the target start is less than the previously computed start, then the target range
+            // overlaps entirely with the previously computed one. We therefore need to clear out
+            // the previously computed data, so as not to conflict.
+            provider.clear_trie_changesets()?;
+            computed_range = target_range.clone();
+        }
```
shekhirin, line 333: what if you have `retention_blocks = 10_000`, run the merkle changesets stage, and then stop the node for 1000 blocks. When you start it again, it will have 9000 blocks of useful changesets and we'll need to calculate only 1000 missing.

do I understand this snippet right, that it would clean the changesets for `9000..`, and then call `populate_range` with `9001..10001`?

mediocregopher: > do I understand this snippet right, that it would clean the changesets for 9000.., and then call populate_range with 9001..10001?

yes, it's not strictly necessary because that range _should_ always be empty, but I figured it'd be better to be safe and clear it anyway. If it really is empty then it's just a single db lookup to confirm that.

crates/stages/stages/src/stages/merkle_changesets.rs
yongkangc: makes sense to me 👍🏻

yongkangc (approved):

crates/stages/stages/src/stages/merkle_changesets.rs
mediocregopher: That is what it's doing, if there is no finalized block. I guess a better approach might be to take the min of the two.

shekhirin: there's always a finalized block number after the first sync IIUC, so the `retention_blocks` setting will get overwrited by it at all times

mediocregopher: Made that change

shekhirin (approved):
Send it 🚀

PR #18812, opened 2025-10-01
chore: Use nightly cargo doc

jenniferzelo:
Replace all cargo docs calls with cargo doc and pin nightly where -Zunstable-options is used.


mattsse (commented):
this is working fine

PR #18819, opened 2025-10-01
fix: Add eth69 status validation

nethoxa:
This add some validation to the eth69 handshake as done in other clients like [geth](https://github.com/ethereum/go-ethereum/blob/057667151b1940e7403a97bc9fcbc207cd5b5045/eth/protocols/eth/handshake.go#L212) as well as bump the maximum size of the status package to match other clients.

### DIFF
--- a/crates/net/eth-wire/src/handshake.rs
+++ b/crates/net/eth-wire/src/handshake.rs
@@ -3,6 +3,7 @@ use crate::{
     ethstream::MAX_MESSAGE_SIZE,
     CanDisconnect,
 };
+use alloy_primitives::FixedBytes;
 use bytes::{Bytes, BytesMut};
 use futures::{Sink, SinkExt, Stream};
 use reth_eth_wire_types::{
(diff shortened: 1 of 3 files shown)


crates/net/eth-wire/src/handshake.rs
```diff
@@ -110,7 +111,7 @@ where
             }
         };
 
-        if their_msg.len() > MAX_STATUS_SIZE {
+        if their_msg.len() > MAX_MESSAGE_SIZE {
```
mattsse, line 113: yeah I guess it doesnt really matter what value we check here because at this point we've already read the entire thing, could also just check for something smaller because this should just contain the status, but doesnt really matter, malformed status would just fail in the decoding step

crates/net/eth-wire/src/handshake.rs
```diff
@@ -1,8 +1,9 @@
 use crate::{
     errors::{EthHandshakeError, EthStreamError, P2PStreamError},
-    ethstream::MAX_STATUS_SIZE,
+    ethstream::MAX_MESSAGE_SIZE,
     CanDisconnect,
 };
+use alloy_primitives::FixedBytes;
```
mattsse:
```suggestion
```

crates/net/eth-wire/src/handshake.rs
```diff
@@ -205,6 +206,20 @@ where
...
+                if let StatusMessage::Eth69(s) = their_status_message {
+                    if s.earliest > s.latest {
+                        return Err(EthHandshakeError::EarliestBlockGreaterThanLatestBlock {
+                            got: s.earliest,
+                            latest: s.latest,
+                        }
+                        .into());
+                    }
+
+                    if s.blockhash == FixedBytes::<32>::ZERO {
```
mattsse:
```suggestion
                    if s.blockhash.is_zero() {
```

mattsse (approved):
ty

PR #18822, opened 2025-10-01
feat: OverlayStateProvider

mediocregopher:
This works towards #18466, though there's still a couple features needed.

The goal of the OverlayStateProvider is to automatically overlay HashedPostState and TrieUpdates onto any DB accesses which are being performed without the code using the provider needing to be aware of it.

The OverlayStateProvider is conceptually similar to the HistoricalStateProvider, except that:
* It can accept arbitrary overlay data in its constructor.
* It implements Hashed/TrieCursorFactory traits, making it useful for the MultiProofTask.

To make this possible I modified the signatures of the in-memory cursor factories for both Hashed and Trie tables to use AsRefs rather than just pointers. I also had to refactor the database cursor factories signatures wrt to references lifetimes. This gives us more flexibility in how we pass the overlay data into them while dealing with rust's ownership rules; I could not figure out any other way to do it.

The next PR will introduce OverlayStateProviderFactory and implement the `from_reverts` method on it. Combined together the factory can act essentially like the ConsistentDbView for proofs, except it will actually handle changes to persistence state gracefully rather than erroring.

### DIFF
--- a/crates/trie/db/src/trie_cursor.rs
+++ b/crates/trie/db/src/trie_cursor.rs
@@ -28,8 +28,6 @@ impl<T> DatabaseTrieCursorFactory<T> {
     }
 }
 
-/// Implementation for DatabaseTrieCursorFactory holding a reference to a DbTx.
-/// The factory uses AsRef internally to access the transaction.
 impl<TX> TrieCursorFactory for DatabaseTrieCursorFactory<&TX>
 where
     TX: DbTx,
--- a/crates/trie/db/src/trie_cursor.rs
+++ b/crates/trie/db/src/trie_cursor.rs
@@ -13,23 +13,27 @@ use reth_trie::{
 
 /// Wrapper struct for database transaction implementing trie cursor factory trait.
 #[derive(Debug)]
-pub struct DatabaseTrieCursorFactory<'a, TX>(&'a TX);
+pub struct DatabaseTrieCursorFactory<T>(T);
 
-impl<TX> Clone for DatabaseTrieCursorFactory<'_, TX> {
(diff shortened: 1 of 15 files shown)


crates/trie/db/src/trie_cursor.rs
```diff
@@ -13,23 +13,25 @@ use reth_trie::{
...
-pub struct DatabaseTrieCursorFactory<'a, TX>(&'a TX);
+pub struct DatabaseTrieCursorFactory<T>(T);
 
-impl<TX> Clone for DatabaseTrieCursorFactory<'_, TX> {
+impl<T: Clone> Clone for DatabaseTrieCursorFactory<T> {
     fn clone(&self) -> Self {
-        Self(self.0)
+        Self(self.0.clone())
     }
 }
```
Rjected: can we derive this now?

yongkangc:
one question: 

>The goal of the OverlayStateProvider is to automatically overlay HashedPostState and TrieUpdates onto any DB accesses which are being performed without the code using the provider needing to be aware of it.

curious why would this be important? does it reduce code duplication, performance or enables https://github.com/paradigmxyz/reth/issues/18466 for historical view?

yongkangc (approved):
lgtm, i dont see any issues and it actually seems like a simple distinct refactor to have the new overlay state provider

Rjected (approved):
lgtm, one comment about deriving clone

crates/trie/db/src/trie_cursor.rs
mediocregopher: yup! done

PR #18824, opened 2025-10-01
fix: replace expect() with safe Option handling

aso20455:
Replace dangerous .expect("todo: error handling, headers should exist") with proper Option pattern matching in DatabaseProvider::remove_blocks_above().

The get_highest_static_file_block() method returns Option<BlockNumber> and can be None when no static file headers exist on disk. Using .expect() could cause runtime panics during blockchain reorganization operations.

Changes:
- Replace .expect() with if-let Some() pattern matching
- Add graceful handling when static file headers are absent
- Add informative warn! logging for both success and skip cases (as requested by reviewers)
- Maintain identical behavior when headers exist


mattsse (commented):
we shouldnt ignore the error here, I think this isnt even reachable

Issue #18818, opened 2025-10-01
Optimize account cache updates


mattsse:
### Describe the feature

This 

https://github.com/paradigmxyz/reth/blob/a2bde852bb1480b7e3a5853524cf82eec677b042/crates/engine/tree/src/tree/cached_state.rs#L398-L403

loops over:
https://github.com/paradigmxyz/reth/blob/a2bde852bb1480b7e3a5853524cf82eec677b042/crates/engine/tree/src/tree/cached_state.rs#L325-L338

which we could optimize to only do the account lookup once:
https://github.com/paradigmxyz/reth/blob/a2bde852bb1480b7e3a5853524cf82eec677b042/crates/engine/tree/src/tree/cached_state.rs#L332-L337

leniram159:
🙋

mattsse:
thanks @leniram159 @Peponks9 

will have more shortly

Himess:
@mattsse

PR #18825, opened 2025-10-01
perf: optimize account cache updates to reduce duplicate lookups

futreall:
Optimize account cache updates by introducing bulk storage insertion to eliminate redundant cache lookups

## Changes:
- Added `insert_storage_bulk()` method that performs single cache lookup per account
- Modified `insert_state()` to use bulk insertion instead of individual calls per storage slot
- Reduced cache lookups from O(n) to O(1) for accounts with multiple storage slots

## Impact:
- No breaking changes to existing API
- Resolves inefficiency identified in issue #18818

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -337,6 +337,25 @@ impl ExecutionCache {
         account_cache.insert_storage(key, value);
     }
 
+    /// Insert multiple storage values into hierarchical cache for a single account
+    ///
+    /// This method is optimized for inserting multiple storage values for the same address
+    /// by doing the account cache lookup only once instead of for each key-value pair.
+    pub(crate) fn insert_storage_bulk<I>(&self, address: Address, storage_entries: I)
+    where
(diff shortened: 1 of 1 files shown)


mediocregopher (approved):
LGTM

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -337,6 +337,25 @@ impl ExecutionCache {
         account_cache.insert_storage(key, value);
     }
 
+    /// Insert multiple storage values into hierarchical cache for a single account
+    ///
+    /// This method is optimized for inserting multiple storage values for the same address
+    /// by doing the account cache lookup only once instead of for each key-value pair.
+    pub(crate) fn insert_storage_bulk<I>(&self, address: Address, storage_entries: I)
```
mattsse, line 339: cool, we can now use this fn in the existing `insert_storage` fn with `[(key, value)]`

then we only have this in one place

mattsse (changes requested):
smol nit

yongkangc (approved):

mattsse (approved):

PR #18826, opened 2025-10-01
fix(exex): optimize block counter in BackfillJob to avoid redundant arithmetic

oooLowNeoNooo:
Replace redundant arithmetic calculation `block_number - *self.range.start() + 1` 
with a simple counter `blocks_processed` in BackfillJob::execute_range().

This change:
- Fixes logical inconsistency with main ExecutionStage implementation
- Eliminates unnecessary arithmetic operations on each iteration
- Improves performance by using simple increment instead of subtraction/addition
- Ensures correct behavior with batch thresholds (e.g., max_blocks=1 now processes exactly 1 block)

### DIFF
--- a/crates/exex/exex/src/backfill/job.rs
+++ b/crates/exex/exex/src/backfill/job.rs
@@ -88,6 +88,7 @@ where
 
         let mut blocks = Vec::new();
         let mut results = Vec::new();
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this has been fine

PR #18830, opened 2025-10-02
feat(flashblocks): relax Sync bounds on Sink and connect futures

MozirDmitriy:
Removed +Sync from S: Sink<Message> and C bounds in Stream impl, from the inherent impl bound for connect, from ConnectFuture type alias, and from WsConnect::connect return type. Updated test connector implementations accordingly.

Reason: WsFlashBlockStream and FlashBlockService drive the stream within a single spawned task and do not share sink/stream/futures across threads. Requiring Sync is unnecessary and restricts compatibility because many Future/Sink types are Send but intentionally not Sync. Relaxing to Send preserves safety and broadens compatibility without changing behavior.


mattsse (approved):
fair

PR #18831, opened 2025-10-02
feat: wait for new blocks when build is in progress

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/18799.

### DIFF
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -30,4 +30,4 @@ pub type FlashBlockCompleteSequenceRx =
     tokio::sync::broadcast::Receiver<FlashBlockCompleteSequence>;
 
 /// Receiver that signals whether a [`FlashBlock`] is currently being built.
-pub type BuildStateRx = tokio::sync::watch::Receiver<Option<FlashBlockBuildInfo>>;
+pub type InProgressFlashBlockRx = tokio::sync::watch::Receiver<Option<FlashBlockBuildInfo>>;
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -4,7 +4,7 @@ pub use payload::{
     ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashBlock, FlashBlockDecoder,
     Metadata,
 };
-pub use service::FlashBlockService;
+pub use service::{FlashBlockBuildInfo, FlashBlockService};
 pub use ws::{WsConnect, WsFlashBlockStream};
 
 mod consensus;
@@ -30,4 +30,4 @@ pub type FlashBlockCompleteSequenceRx =
     tokio::sync::broadcast::Receiver<FlashBlockCompleteSequence>;
 
 /// Receiver that signals whether a [`FlashBlock`] is currently being built.
-pub type BuildStateRx = tokio::sync::watch::Receiver<bool>;
+pub type BuildStateRx = tokio::sync::watch::Receiver<Option<FlashBlockBuildInfo>>;
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -28,3 +28,6 @@ pub type PendingBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingFlashBlo
 /// [`FlashBlock`]: crate::FlashBlock
 pub type FlashBlockCompleteSequenceRx =
     tokio::sync::broadcast::Receiver<FlashBlockCompleteSequence>;
+
+/// Receiver that signals whether a [`FlashBlock`] is currently being built.
+pub type BuildStateRx = tokio::sync::watch::Receiver<bool>;
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -21,7 +21,10 @@ use std::{
     task::{ready, Context, Poll},
     time::Instant,
 };
-use tokio::{pin, sync::oneshot};
+use tokio::{
+    pin,
+    sync::{oneshot, watch},
+};
 use tracing::{debug, trace, warn};
 
 pub(crate) const FB_STATE_ROOT_FROM_INDEX: usize = 9;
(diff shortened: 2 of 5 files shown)


crates/optimism/flashblocks/src/lib.rs
```diff
@@ -28,3 +28,6 @@ pub type PendingBlockRx<N> = tokio::sync::watch::Receiver<Option<PendingFlashBlo
 /// [`FlashBlock`]: crate::FlashBlock
 pub type FlashBlockCompleteSequenceRx =
     tokio::sync::broadcast::Receiver<FlashBlockCompleteSequence>;
+
+/// Receiver that signals whether a [`FlashBlock`] is currently being built.
+pub type BuildStateRx = tokio::sync::watch::Receiver<bool>;
```
mattsse: I'd like to avoid a public alias just for this, makes it more confusing

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -114,10 +131,42 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+        // It is not necessary to wait if it is the end of the 10-flashblock sequence, ie after 9
+        // blocks.
+        if flashblock_index >= MAX_FLASHBLOCK_INDEX {
+            return false;
+        }
+
+        let time_remaining = current_expires_at.saturating_duration_since(now);
+
+        // Only wait if current flashblock will expire soon
+        time_remaining < MIN_FRESHNESS_THRESHOLD
```
mattsse: this we can skip, because we also don't really need the expiry here anyway

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -129,20 +178,48 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+
+            let now = Instant::now();
+
+            // Is the pending block not expired and latest is its parent?
+            if pending.evm_env.block_env.number != U256::from(pending_block.block().number()) ||
+                parent.hash() != pending_block.block().parent_hash() ||
+                now > pending_block.expires_at
+            {
+                return Ok(None);
+            }
```
mattsse: this check is now bypassed in the next step

what we can do is extract obtaining the block to `async fn flashbock()`, then we don't really need to modify this function much

mattsse (changes requested):
cool, some suggestions

cc @loocapro

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -114,10 +131,42 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+        // No build currently in progress
+        if !*build_state_rx.borrow() {
+            return false;
+        }
+
+        // It is not necessary to wait if it is the end of the 10-flashblock sequence, ie after 9
+        // blocks.
+        if flashblock_index >= MAX_FLASHBLOCK_INDEX {
+            return false;
+        }
```
loocapro: Can we remove this check? It is always useful to await imo.

loocapro (dismissed):
Nice!

crates/optimism/flashblocks/src/lib.rs
lean-apple: It's used in `opt-rpc`, but can remove this alias if it is too simple

crates/optimism/rpc/src/eth/mod.rs
lean-apple: https://github.com/paradigmxyz/reth/pull/18831/commits/e5ef344eef73815a0dd988828f5e04c1ce447108.

crates/optimism/rpc/src/eth/mod.rs
lean-apple: Yup, done, I also remove the check on the block number and only keep the parent hash verification

crates/optimism/flashblocks/src/service.rs
```diff
@@ -48,6 +52,9 @@ pub struct FlashBlockService<
     /// when fb received on top of the same block. Avoid redundant I/O across multiple
     /// executions within the same block.
     cached_state: Option<(B256, CachedReads)>,
+    /// Signals when a block build is in progress
+    build_state_tx: watch::Sender<bool>,
```
mattsse: what if use `Option<FbInfo>` here where FbInfo is parenthash + index + blocknumber

then we can easily check from the pendingblock pov if the Fb in progress is the next index

mattsse (changes requested):
this looks pretty good, I have one more idea that makes the index check redundant and is perhaps useful in general for awaiting

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -109,15 +119,54 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+            let mut rx_clone = rx.clone();
+
+            // Wait up to 20ms for a new flashblock to arrive
+            if tokio::time::timeout(MAX_WAIT_DURATION, rx_clone.changed()).await.is_ok() {
+                let fresh = rx_clone.borrow();
+                if let Some(block) = self.extract_matching_block(fresh.as_ref(), parent_hash) {
+                    return Ok(Some(block));
+                }
+            }
+        }
```
mattsse: I think we can simplify this and only need to keep `tokio::time::timeout(MAX_WAIT_DURATION, rx_clone.changed()).await`

because borrow always returns the most recent value

mattsse (dismissed):

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -109,15 +119,54 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+        &self,
+        parent_hash: B256,
+    ) -> eyre::Result<Option<PendingBlock<N::Primitives>>> {
+        let Some(rx) = self.inner.pending_block_rx.as_ref() else { return Ok(None) };
+
+        if self.is_flashblock_building() {
+            let mut rx_clone = rx.clone();
+
+            // Wait up to 20ms for a new flashblock to arrive
+            if tokio::time::timeout(MAX_WAIT_DURATION, rx_clone.changed()).await.is_ok() {
```
mattsse: let's also do 50ms here

lean-apple: https://github.com/paradigmxyz/reth/pull/18831/commits/e07698d0ec7ba80dac608805fcd3f1b96641f6d5

crates/optimism/rpc/src/eth/mod.rs
lean-apple: https://github.com/paradigmxyz/reth/pull/18831/commits/6f3b1ce57c396547e902db4fb5246d4b1aff758d

lean-apple:
It's good for another round review @RomanHodulak :)

crates/optimism/flashblocks/src/service.rs
lean-apple: [f32d618](https://github.com/paradigmxyz/reth/pull/18831/commits/f32d6181116cced84e0bd155633e037ff9febd5c)

RomanHodulak (approved):
LGTM

PR #18832, opened 2025-10-02
fix: handle unused return values in module registration

sky-coderay:
Explicitly handle return values from `HashMap::insert()` in RPC module registration methods. Uses `let _ =` to indicate that overwriting existing modules is intentional behavior.


mattsse (commented):
this isn't necessary because these aren't marked as must_use

PR #18840, opened 2025-10-02
fix(sim): clamp bundle timeout to max instead of falling back to default

maradini77:
**Problem**
When users specify a `timeout` value greater than `MAX_SIM_TIMEOUT` in bundle simulation, the current logic silently falls back to `DEFAULT_SIM_TIMEOUT` instead of clamping to the maximum allowed value. This leads to unexpected behavior where large timeout values result in shorter-than-expected timeouts.

**Solution**
Replace the `filter()` + `unwrap_or()` pattern with explicit `min()` clamping to ensure predictable timeout behavior.


mattsse (approved):
makes sense

PR #18847, opened 2025-10-03
fix(network): prevent metric leak in outgoing message queue on session teardown

prestoalvarez:
## Summary

Fixes a metric leak in `QueuedOutgoingMessages` where the `Gauge` counter could become permanently inflated when sessions terminate with messages still queued.

## Problem

The `Gauge` metric was only decremented on explicit `pop_front()` calls, but never when the queue was destroyed. If a session terminated with pending messages in the queue, the metric would remain artificially high indefinitely, causing:

- Inaccurate monitoring data
- Potential false alerts
- Difficulty in diagnosing queue health

## Solution

Added a `Drop` implementation for `QueuedOutgoingMessages` that decrements the gauge by the number of remaining messages when the struct is destroyed.

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -924,6 +924,16 @@ impl<N: NetworkPrimitives> QueuedOutgoingMessages<N> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
nice find, ty

PR #18849, opened 2025-10-03
chore: do not generate fuzz tests in Compact macro

klkvr:
`test_fuzz` relies on bincode serde which does not work nicely for all our types.

we already have macros for generating fuzz tests via arbitrary impls which we can use


RomanHodulak (approved):
Makes sense to me

PR #18852, opened 2025-10-03
fix: replace allow(unused_imports) with expect(unused_imports) in fuzz module

prestoalvarez:
Align lint handling by replacing unconditional `allow` with `expect` for unused imports in the fuzz module. `expect(unused_imports)` preserves useful warnings while documenting intent, improving signal-to-noise and consistency with adjacent imports.


shekhirin:
I don't think this changes anything, feel free to reopen.

PR #18853, opened 2025-10-03
docs: remove outdated TODO comment in network documentation

reallesee:
Remove outdated TODO comment for txpool spawning in network documentation.

The txpool variable is already prefixed with underscore (_txpool), indicating it's intentionally unused in this simplified example. The actual txpool integration happens through the builder pattern elsewhere in the codebase


mattsse (commented):
this todo is useful

PR #18854, opened 2025-10-03
chore(trie): Implement OverlayStateProviderFactory

mediocregopher:
This implements the OverlayStateProviderFactory, which intends on replacing the ConsistentDbView for use with proof fetching.

When creating a new OverlayStateProvider, the OverlayStateProviderFactory can be configured to first revert that provider's state to a particular block number (similar to the HistoricalStateProvider) and then also apply some other overlay state on top of it. This gives us full flexibility in computing proofs on top of re-org chains, even if they are based on block prior to the db tip.

It uses a new method `trie_reverts` to fetch trie changesets, alongside HashedPostState reverts using the existing method for that. With these two it is able to revert to any previous block for every provider it generates, ensuring a consistent view of the data. In the future if we switch to using tx children then the factory can fetch the reverts as part of its own constructor, rather than on every new provider.

This PR required adding the ability to extend the HashedPostStateSorted and TrieUpdatesSorted types. This is implemented fairly efficiently by updating the collection in-place with updates/removals, and then appending additions onto the back and sorting.

Closes #18466

### DIFF
--- a/crates/storage/provider/src/providers/state/overlay.rs
+++ b/crates/storage/provider/src/providers/state/overlay.rs
@@ -1,14 +1,107 @@
-use alloy_primitives::B256;
+use alloy_primitives::{BlockNumber, B256};
 use reth_db_api::DatabaseError;
-use reth_storage_api::DBProvider;
+use reth_storage_api::{DBProvider, DatabaseProviderFactory, TrieReader};
 use reth_trie::{
     hashed_cursor::{HashedCursorFactory, HashedPostStateCursorFactory},
     trie_cursor::{InMemoryTrieCursorFactory, TrieCursorFactory},
     updates::TrieUpdatesSorted,
-    HashedPostStateSorted,
+    HashedPostState, HashedPostStateSorted, KeccakKeyHasher,
 };
-use reth_trie_db::{DatabaseHashedCursorFactory, DatabaseTrieCursorFactory};
+use reth_trie_db::{
+    DatabaseHashedCursorFactory, DatabaseHashedPostState, DatabaseTrieCursorFactory,
+};
 use std::sync::Arc;
+
+/// Factory for creating overlay state providers with optional reverts and overlays.
+///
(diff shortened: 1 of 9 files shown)


crates/storage/storage-api/src/trie.rs
```diff
@@ -89,6 +89,14 @@ pub trait StateProofProvider: Send + Sync {
     fn witness(&self, input: TrieInput, target: HashedPostState) -> ProviderResult<Vec<Bytes>>;
 }
 
+/// Trie Reader
+#[auto_impl::auto_impl(&, Arc, Box)]
+pub trait TrieReader: Send + Sync {
+    /// Returns the [`TrieUpdatesSorted`] for reverting the trie database to its state prior to the
+    /// given block having been processed.
+    fn revert_trie(&self, from: BlockNumber) -> ProviderResult<TrieUpdatesSorted>;
```
shekhirin: the method name reads like we're actually reverting the data in database, maybe `trie_reverts` instead?

crates/storage/provider/src/providers/state/overlay.rs
```diff
@@ -1,15 +1,148 @@
...
+        })?;
+
+        // Extract the MerkleChangeSets checkpoint details using the helper method
+        let merkle_checkpoint =
+            checkpoint.merkle_changesets_stage_checkpoint().ok_or_else(|| {
+                ProviderError::InsufficientChangesets {
+                    requested: requested_block,
+                    available: 0..=0,
+                }
+            })?;
```
shekhirin: can collapse I think

```suggestion
        // If there's no checkpoint at all or block range details are missing, we can't revert
        let checkpoint = checkpoint.and_then(|checkpoint| checkpoint.merkle_changesets_stage_checkpoint()).ok_or_else(|| ProviderError::InsufficientChangesets {
            requested: requested_block,
            available: 0..=0,
        })?;
```

shekhirin (approved):
LGTM, nits

yongkangc (approved):
lgtm! p clear

PR #18856, opened 2025-10-03
feat: make `ChainSpec` generic over header

klkvr:
Allows `ChainSpec` to store a custom header type. Spec with custom header does not yet have complete support for all helpers `From<Genesis>` does not work but this should still simplify things slightly for custom chainspecs


fgimenez (approved):
lgtm

PR #18857, opened 2025-10-03
chore(cli): rename disable-caching-and-prewarming to disable-prewarming

Rjected:
This does not actually disable caching, so this is renaming the flag to reflect that. Technically breaking for those using this CLI flag


mediocregopher (approved):
Seems fine to me 👍 will wait for someone else to chime in on the breaking compatibility issue, not a big deal imo

yongkangc (approved):
makes sense

mattsse (commented):
superseded by https://github.com/paradigmxyz/reth/pull/19072

PR #18859, opened 2025-10-04
chore(engine): remove unused local Rayon pool from WorkloadExecutor

sashass1315:
Remove the per-instance Rayon thread pool from WorkloadExecutor and delete unused methods with_num_cpu_threads(...) and rayon_pool(). The local pool was never referenced; parallelism in the payload processor uses Tokio spawn_blocking and the global Rayon pool configured at node launch. Keeping a private pool consumed resources without benefit, making it dead code.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/executor.rs
+++ b/crates/engine/tree/src/tree/payload_processor/executor.rs
@@ -1,10 +1,6 @@
 //! Executor for mixed I/O and CPU workloads.
 
-use rayon::ThreadPool as RayonPool;
-use std::{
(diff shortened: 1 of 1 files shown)


mediocregopher:
@shekhirin @mattsse what was the intention here? did we mean to use this rayon pool for payload validation and then never did? Or this is just some cleanup that didn't get completed?

mattsse (commented):
ty, I cherry picked some doc changes into #19065

@mediocregopher we previously used this to spawn tx execution I believe but these can also do a bunch of io so we moved away from rayon here

PR #18860, opened 2025-10-04
fix: correct grammatical error in stream_codec.rs comment

sashaodessa:
## Summary
Fixes a grammatical error in the comment describing the file's origin.

- **Change**: Corrected "This basis of this file has been taken from" → "This file is based on"


mattsse (commented):
not worth it

PR #18866, opened 2025-10-05
perf(net): avoid hash copies in propagate_hashes_to

gancerlory:
Avoid unnecessary hash copies in `propagate_hashes_to` by accepting references instead of values


Rjected:
Both of these copy now, `*h.borrow()` copies and `.copied()` copies. We ultimately need a `Vec<TxHash>` to pass to  `get_all` currently so this kind of optimization cannot be made unless this changes

PR #18867, opened 2025-10-05
chore: remove unused imports in blockchain_provider

gancerlory:
Remove unused `#![allow(unused)]` attribute and clean up unused imports in `blockchain_provider.rs`.

### DIFF
--- a/crates/storage/provider/src/providers/blockchain_provider.rs
+++ b/crates/storage/provider/src/providers/blockchain_provider.rs
@@ -2554,14 +2554,15 @@ mod tests {
             persist_block_after_db_tx_creation(provider.clone(), in_memory_blocks[1].number);
             let to_be_persisted_tx = in_memory_blocks[1].body().transactions[0].clone();
 
-            assert!(matches!(
+            assert_eq!(
                 correct_transaction_hash_fn(
                     *to_be_persisted_tx.tx_hash(),
                     provider.canonical_in_memory_state(),
                     provider.database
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -2586,7 +2560,7 @@ mod tests {
                     provider.canonical_in_memory_state(),
                     provider.database
                 ),
-                Ok(Some(to_be_persisted_tx))
+                Ok(Some(_to_be_persisted_tx))
```
onbjerg: I don't understand this change, `_to_be_persisted_tx` does not exist?

onbjerg (dismissed):

crates/storage/provider/src/providers/blockchain_provider.rs
gancerlory: Sorry, I misclicked — I actually meant to replace `to_be_persisted_tx` with `-`.

shekhirin (approved):

crates/storage/provider/src/providers/blockchain_provider.rs
shekhirin: should probably just do `assert_eq` here, not `matches`

gancerlory: Make sense! thanks!

mattsse (approved):

PR #18877, opened 2025-10-06
fix(exex): handle JoinHandle error in poll_next_task without using ?

MozirDmitriy:
What was wrong: poll_next_task returned Poll<Option<Result<...>>> but used ? on a JoinHandle result (res.map_err(...)?). The ? operator cannot propagate from Result into Poll<Option<...>>, leading to a type mismatch and incorrect error propagation semantics.

Fix: Replace the ? with explicit handling of the JoinError, returning Poll::Ready(Some(Err(BlockExecutionError::other(e)))) when a task join fails. This correctly surfaces task failures as stream item errors.

Why necessary: Ensures the stream adheres to its Item = Result<...> contract, avoids invalid use of ? in a Poll-returning function, and prevents silent or misrouted errors from background tasks.


shekhirin:
I think it's actually fine, also this compiles? https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromResidual%3CResult%3CInfallible,+E%3E%3E-for-Poll%3COption%3CResult%3CT,+F%3E%3E%3E

PR #18878, opened 2025-10-06
chore(trie): Use trie changesets for engine unwinding

### LINKED ISSUE #18517: Trie ChangeSets: unwind_trie_state_range
### Describe the feature

The `unwind_trie_state_range` method of DatabaseProvider needs to unwind the trie changesets tables as well. This whole method should be able to be greatly simplified given that the changesets tables can be used to do the unwind now.

https://github.com/paradigmxyz/reth/blob/d1c966020b93d12795d855d5471c16ed46689d52/crates/storage/provider/src/providers/database/provider.rs#L259

### Additional context

_No response_

mediocregopher:
Closes #18517

This PR primarily targets the `unwind_trie_state_range` provider method, which is used by the engine API to unwind the db during reorgs.

This method was previously also potentially used by the unwind pipeline stage, but only in an `else` block which should never get hit anymore. This else block was replaced with an error.

`unwind_trie_state_range` was previously recomputing trie updates using a StateRoot calculation, but it is now able to skip that step and simply read trie updates from the trie changeset tables instead. To make applying these updates easier the `write_trie_updates` method is superceded by `write_trie_updates_sorted`.

### DIFF
--- a/crates/cli/commands/src/stage/unwind.rs
+++ b/crates/cli/commands/src/stage/unwind.rs
@@ -15,9 +15,7 @@ use reth_db::DatabaseEnv;
 use reth_downloaders::{bodies::noop::NoopBodiesDownloader, headers::noop::NoopHeaderDownloader};
 use reth_evm::ConfigureEvm;
 use reth_exex::ExExManagerHandle;
-use reth_provider::{
-    providers::ProviderNodeTypes, BlockNumReader, ProviderFactory, StaticFileProviderFactory,
-};
+use reth_provider::{providers::ProviderNodeTypes, BlockNumReader, ProviderFactory};
 use reth_stages::{
     sets::{DefaultStages, OfflineStages},
     stages::ExecutionStage,
@@ -59,26 +57,12 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
 
         let components = components(provider_factory.chain_spec());
 
-        // Sanity check that we have static files available to unwind to the target block.
-        let highest_static_file_block =
-            provider_factory.static_file_provider().get_highest_static_files().max_block_num();
-        if highest_static_file_block
-            .filter(|highest_static_file_block| *highest_static_file_block > target)
-            .is_none()
-        {
-            return Err(eyre::eyre!("static files not available for target block {target}, highest is {highest_static_file_block:?}"));
-        }
-
         if self.offline {
             info!(target: "reth::cli", "Performing an unwind for offline-only data!");
         }
 
-        if let Some(highest_static_file_block) = highest_static_file_block {
-            info!(target: "reth::cli", ?target, ?highest_static_file_block, "Executing a pipeline unwind.");
-        } else {
-            info!(target: "reth::cli", ?target, "Executing a pipeline unwind.");
-        }
-        info!(target: "reth::cli", prune_config=?config.prune, "Using prune settings");
+        let highest_static_file_block = provider_factory.provider()?.last_block_number()?;
+        info!(target: "reth::cli", ?target, ?highest_static_file_block, prune_config=?config.prune,  "Executing a pipeline unwind.");
 
         // This will build an offline-only pipeline if the `offline` flag is enabled
         let mut pipeline =
--- a/crates/cli/commands/src/stage/unwind.rs
+++ b/crates/cli/commands/src/stage/unwind.rs
@@ -16,8 +16,7 @@ use reth_downloaders::{bodies::noop::NoopBodiesDownloader, headers::noop::NoopHe
 use reth_evm::ConfigureEvm;
 use reth_exex::ExExManagerHandle;
 use reth_provider::{
-    providers::ProviderNodeTypes, BlockExecutionWriter, BlockNumReader, ChainStateBlockReader,
-    ChainStateBlockWriter, ProviderFactory, StaticFileProviderFactory,
+    providers::ProviderNodeTypes, BlockNumReader, ProviderFactory, StaticFileProviderFactory,
 };
 use reth_stages::{
     sets::{DefaultStages, OfflineStages},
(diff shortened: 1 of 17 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -339,75 +334,31 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
-                block_hash: parent_hash,
-            })))
-        }
-        self.write_trie_updates(&trie_updates)?;
+        // Unwind accounts/storages trie tables using the revert.
+        let trie_revert = self.trie_reverts(range.clone())?;
+        self.write_trie_updates_sorted(&trie_revert)?;
+
+        // Clear trie changesets which have been unwound.
+        self.clear_trie_changesets_from(*range.start())?;
```
mediocregopher: @shekhirin you mentioned in a previous PR that we don't want to have deletion from changesets done on a range, but this method only accepts a range. Do you think it's fine to leave this like this? or better to re-introduce `clear_trie_changesets_from`?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -339,75 +334,31 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
-            return Err(ProviderError::UnwindStateRootMismatch(Box::new(RootMismatch {
-                root: GotExpected { got: new_state_root, expected: parent_state_root },
-                block_number: parent_number,
-                block_hash: parent_hash,
-            })))
-        }
-        self.write_trie_updates(&trie_updates)?;
+        // Unwind accounts/storages trie tables using the revert.
+        let trie_revert = self.trie_reverts(range.clone())?;
+        self.write_trie_updates_sorted(&trie_revert)?;
```
Rjected: nice, i like that this is now just a call to a provider fn

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2150,32 +2101,22 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
 
 impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider<TX, N> {
-    /// Writes trie updates. Returns the number of entries modified.
-    fn write_trie_updates(&self, trie_updates: &TrieUpdates) -> ProviderResult<usize> {
-        if trie_updates.is_empty() {
+    /// Writes trie updates to the database with already sorted updates.
+    ///
+    /// Returns the number of entries modified.
+    fn write_trie_updates_sorted(&self, trie_updates: &TrieUpdatesSorted) -> ProviderResult<usize> {
+        if trie_updates.account_nodes.is_empty() && trie_updates.storage_tries.is_empty() {
```
Rjected: could probably just introduce `TrieUpdatesSorted::is_empty`

crates/cli/commands/src/stage/unwind.rs
```diff
@@ -60,54 +59,35 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
...
-            let last_saved_finalized_block_number = provider.last_finalized_block_number()?;
-            if last_saved_finalized_block_number.is_none_or(|f| f > target) {
-                provider.save_finalized_block_number(target)?;
-            }
+        // Move all applicable data from database to static files.
+        pipeline.move_to_static_files()?;
 
-            provider.commit()?;
-        }
+        pipeline.unwind(target, None)?;
```
Rjected: is the reasoning behind this, that currently we pretty much always have static files and will now always be doing a pipeline unwind?

Rjected (approved):
I like that this makes things simpler, I have a question and suggestion for a helper method but otherwise this looks good to me

crates/cli/commands/src/stage/unwind.rs
mediocregopher: Yes, with the work @shekhirin has been doing we always have static files, there's never a situation where static files are behind mdbx because we don't even write that data to mdbx anymore. So if static files don't have the data we can't do anything, just error out, otherwise always pipeline.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -4163,4 +4113,224 @@ mod tests {
             assert!(block5_entries.is_empty(), "Block 104 storage entries should be deleted");
         }
     }
+
+    #[test]
+    fn test_write_trie_updates_sorted() {
```
yongkangc, line 4106: yea this test is helpful to understand this pr

crates/cli/commands/src/stage/unwind.rs
```diff
@@ -60,54 +59,35 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
...
-                self.build_pipeline(config, provider_factory, components.evm_config().clone())?;
+        // Sanity check that we have static files available to unwind to the target block.
+        let highest_static_file_block =
+            provider_factory.static_file_provider().get_highest_static_files().max_block_num();
+        if highest_static_file_block
+            .filter(|highest_static_file_block| *highest_static_file_block > target)
+            .is_none()
+        {
+            return Err(eyre::eyre!("static files not available for target block {target}, highest is {highest_static_file_block:?}"));
+        }
```
shekhirin: think we can just use the `last_block_number` here https://github.com/paradigmxyz/reth/blob/83cec3793bb2300c76f1aeb88aac1e1a96309f34/crates/storage/provider/src/providers/static_file/manager.rs#L1737-L1740

crates/cli/commands/src/stage/unwind.rs
```diff
@@ -60,54 +59,35 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
...
 
-            pipeline.unwind(target, None)?;
+        if let Some(highest_static_file_block) = highest_static_file_block {
+            info!(target: "reth::cli", ?target, ?highest_static_file_block, "Executing a pipeline unwind.");
         } else {
-            info!(target: "reth::cli", ?target, "Executing a database unwind.");
-            let provider = provider_factory.provider_rw()?;
+            info!(target: "reth::cli", ?target, "Executing a pipeline unwind.");
+        }
```
shekhirin: this is guaranteed to always be `Some` I think?

crates/storage/provider/src/providers/database/provider.rs
shekhirin: I would prefer to rename `unwind_trie_state_range` into `unwind_trie_state_from` because we always call it with `N..=last_block_number`

crates/cli/commands/src/stage/unwind.rs
mediocregopher: I refactored this whole area again, the call to `unwind_target` above this also calls `last_block_number` and checks it against the target, so this check was redundant anyway.

crates/storage/provider/src/providers/database/provider.rs
mediocregopher: Fair enough, done

crates/storage/provider/src/providers/database/provider.rs
shekhirin: outdated comment

shekhirin (approved):
LGTM, appreciate the cleanup in `reth stage unwind`

PR #18879, opened 2025-10-06
fix(engine): re-insert storage cache and use arc

Rjected:
Fixes execution cache-related memory growth by re-inserting storage caches every time they are modified, this causes moka to pick up on the change and re-weigh the cache.

Previous memory usage:
Screenshot 2025-10-03 at 3 29 31 PM

Memory usage with this PR:
Screenshot 2025-10-03 at 5 42 50 PM


mattsse (approved):
makes sense, the overhead of reinserting per contract is likely negligible

PR #18880, opened 2025-10-06
fix(engine): flatten storage cache

Rjected:
This fixes memory growth caused by growing execution caches, an alternative to https://github.com/paradigmxyz/reth/pull/18879

This instead removes the cache hierarchy, so cache updates are always picked up by the moka weigher.

Previous memory usage:
Screenshot 2025-10-03 at 3 29 31 PM

New memory usage:
Screenshot 2025-10-03 at 5 07 15 PM

This also has slightly higher cache hitrate than #18879

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -355,6 +355,8 @@ impl ExecutionCache {
 
     /// Invalidates the storage for all addresses in the set
     pub(crate) fn invalidate_storages(&self, addresses: HashSet<&Address>) {
+        // NOTE: this must collect because the invalidate function should not be called while we
+        // hold an iter for it
         let storage_entries = self
             .storage_cache
             .iter()
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -1,5 +1,8 @@
 //! Execution cache implementation for block processing.
-use alloy_primitives::{Address, StorageKey, StorageValue, B256};
+use alloy_primitives::{
+    map::{DefaultHashBuilder, HashSet},
+    Address, StorageKey, StorageValue, B256,
+};
 use metrics::Gauge;
 use mini_moka::sync::CacheBuilder;
 use reth_errors::ProviderResult;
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -300,65 +302,65 @@ pub(crate) struct ExecutionCache {
     /// Cache for contract bytecode, keyed by code hash.
     code_cache: Cache<B256, Option<Bytecode>>,
 
-    /// Per-account storage cache: outer cache keyed by Address, inner cache tracks that account’s
-    /// storage slots.
-    storage_cache: Cache<Address, AccountStorageCache>,
+    /// Flattened storage cache: composite key of (`Address`, `StorageKey`) maps directly to values.
+    storage_cache: Cache<(Address, StorageKey), Option<StorageValue>>,
```
mattsse, line 307: right, we can't really nest caches because this would mean that the weigher of the storage_cache would only consider new acc -> storage entries

this is better because it treats storage slots as individual entries, which I guess explains the higher hit rate

we likely have additional address overhead, but this should be okay, we could compress this to a u64 with a separate map, like we do in the pool, but perhaps not that useful

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -300,65 +302,65 @@ pub(crate) struct ExecutionCache {
...
+
+    /// Invalidates the storage for all addresses in the set
+    pub(crate) fn invalidate_storages(&self, addresses: HashSet<&Address>) {
+        let storage_entries = self
+            .storage_cache
+            .iter()
+            .filter_map(|entry| addresses.contains(&entry.key().0).then_some(*entry.key()));
+        for key in storage_entries {
+            self.storage_cache.invalidate(&key)
+        }
```
mattsse, line 367: hmm, unsure this is sound because this feels problematic

dashmap iter

> Locking behaviour: May deadlock if called when holding a mutable reference into the map.

and cache iter:

> Locking behavior
> This iterator relies on the iterator of dashmap::DashMap  , which employs read-write locks. May deadlock if the thread holding an iterator attempts to update the cache.

tho I don't know if this is applicable here
unfortunately we don't have a retain here

should be perhaps collect then remove?

mattsse (changes requested):
makes sense, let's go with this

I'm a bit concerned about the invalidate impl

crates/engine/tree/src/tree/cached_state.rs
Rjected: hmm, yeah I think we should collect then remove here

Rjected: this now collects

mattsse: we should document this with a one liner because this is a prime candidate for a bot "optimization" pr that would look perfectly valid

mattsse (approved):

PR #18882, opened 2025-10-06
feat: add StaticFileSegment::AccountChangeSets

### LINKED ISSUE #18846: AccountChangeSets static files segment
### Describe the feature

Currently, static files and Nippy Jar format support only data with integer keys, such as headers indexed by block number, or transactions indexed by transaction number.

Changesets are more complicated, because they're indexed by a composite key as `BlockNumber -> Address -> AccountBeforeTx` https://github.com/paradigmxyz/reth/blob/35c498477db7f2a21dac7555fff5b5f0e76120d3/crates/storage/db-api/src/tables/mod.rs#L442-L446

We need to add support for this in static files.

### Additional context

_No response_

Rjected:
ref https://github.com/paradigmxyz/reth/issues/18846

Introduces a new flag `--static-file.account-change-sets` which controls whether or not we should write account changesets to static files or the DB.

Adds account changesets as a new static file segment, where each row is a change. This makes them different from "block-based" static files like headers, and "transaction-based" static files like transactions and receipts. Row ranges for each block are stored in the header for the static file segment.

Backwards compat serialization / deserialization code is added because this adds a field to the static file header.

`db stats` on a node with this PR:
```
ubuntu@reth8:~/reth$ ./target/profiling/reth db --datadir /mnt/reth-mainnet/ stats
2025-11-13T15:41:39.695842Z  INFO Initialized tracing, debug log directory: /home/ubuntu/.cache/reth/logs/mainnet
2025-11-13T15:41:39.697931Z  INFO Opening storage db_path="/mnt/reth-mainnet/db" sf_path="/mnt/reth-mainnet/static_files"
2025-11-13T15:41:44.982848Z  INFO Verifying storage consistency.
| Segment           | Block Range  | Transaction Range | Shape (columns x rows) | Size      |
|-------------------|--------------|-------------------|------------------------|-----------|
| Headers           | 0..=23000000 | N/A               | 3 x 23000001           | 11.8 GiB  |
| Transactions      | 0..=23000000 | 0..=2907942767    | 1 x 2907942768         | 721.3 GiB |
| Receipts          | 0..=23000000 | 0..=2907942767    | 1 x 2907942768         | 332.5 GiB |
| AccountChangeSets | 0..=23000000 | N/A               | 1 x 3787160399         | 140.8 GiB |
| ----------------- | ------------ | ----------------- | ---------------------- | --------- |
| Total             |              |                   |                        | 1.2 TiB   |


| Table Name                 | # Entries  | Branch Pages | Leaf Pages | Overflow Pages | Total Size |
|----------------------------|------------|--------------|------------|----------------|------------|
| AccountChangeSets          | 0          | 0            | 0          | 0              | 0 B        |
| AccountsHistory            | 67493      | 19           | 1907       | 1100           | 11.8 MiB   |
| AccountsTrie               | 24075389   | 4126         | 996709     | 0              | 3.8 GiB    |
| AccountsTrieChangeSets     | 125727     | 2989         | 16809      | 0              | 77.3 MiB   |
| BlockBodyIndices           | 23000001   | 609          | 136118     | 0              | 534.1 MiB  |
| BlockOmmers                | 1209744    | 844          | 188675     | 0              | 740.3 MiB  |
| BlockWithdrawals           | 5965061    | 3806         | 852142     | 0              | 3.3 GiB    |
| Bytecodes                  | 1757397    | 2835         | 179017     | 3608635        | 14.5 GiB   |
| CanonicalHeaders           | 0          | 0            | 0          | 0              | 0 B        |
| ChainState                 | 0          | 0            | 0          | 0              | 0 B        |
| HashedAccounts             | 304987658  | 43165        | 4100529    | 0              | 15.8 GiB   |
| HashedStorages             | 1339912717 | 612376       | 18592372   | 0              | 73.3 GiB   |
| HeaderNumbers              | 23000001   | 11299        | 547070     | 0              | 2.1 GiB    |
| HeaderTerminalDifficulties | 0          | 0            | 0          | 0              | 0 B        |
| Headers                    | 0          | 0            | 0          | 0              | 0 B        |
| Metadata                   | 1          | 0            | 1          | 0              | 4 KiB      |
| PlainAccountState          | 304987658  | 50115        | 4632009    | 0              | 17.9 GiB   |
| PlainStorageState          | 1339912717 | 883587       | 28323800   | 0              | 111.4 GiB  |
| PruneCheckpoints           | 1          | 0            | 1          | 0              | 4 KiB      |
| Receipts                   | 0          | 0            | 0          | 0              | 0 B        |
| StageCheckpointProgresses  | 1          | 0            | 1          | 0              | 4 KiB      |
| StageCheckpoints           | 16         | 0            | 1          | 0              | 4 KiB      |
| StorageChangeSets          | 6882009001 | 4504624      | 118596017  | 0              | 469.6 GiB  |
| StoragesHistory            | 538338     | 243          | 13720      | 213            | 55.4 MiB   |
| StoragesTrie               | 117839052  | 595741       | 6016100    | 0              | 25.2 GiB   |
| StoragesTrieChangeSets     | 205298     | 7876         | 30288      | 0              | 149.1 MiB  |
| TransactionBlocks          | 21360530   | 613          | 136927     | 0              | 537.3 MiB  |
| TransactionHashNumbers     | 2907942768 | 1436716      | 70050725   | 0              | 272.7 GiB  |
| TransactionSenders         | 2907942768 | 121328       | 27177036   | 0              | 104.1 GiB  |
| Transactions               | 0          | 0            | 0          | 0              | 0 B        |
| VersionHistory             | 1          | 0            | 1          | 0              | 4 KiB      |
| -------------------------- | ---------- | ------------ | ---------- | -------------- | ---------- |
| Tables                     |            |              |            |                | 1.1 TiB    |
| Freelist                   | 20893      |              |            |                | 81.6 MiB   |
```

`db stats` on a regular node:
```
ubuntu@reth9:~/reth$ ./target/profiling/reth db --datadir /mnt/reth-mainnet/ stats
2025-11-13T15:42:14.343426Z  INFO Initialized tracing, debug log directory: /home/ubuntu/.cache/reth/logs/mainnet
2025-11-13T15:42:14.345517Z  INFO Opening storage db_path="/mnt/reth-mainnet/db" sf_path="/mnt/reth-mainnet/static_files"
2025-11-13T15:42:14.357218Z  INFO Verifying storage consistency.
| Segment      | Block Range  | Transaction Range | Shape (columns x rows) | Size      |
|--------------|--------------|-------------------|------------------------|-----------|
| Headers      | 0..=23000000 | N/A               | 3 x 23000001           | 11.8 GiB  |
| Transactions | 0..=23000000 | 0..=2907942767    | 1 x 2907942768         | 721.3 GiB |
| Receipts     | 0..=23000000 | 0..=2907942767    | 1 x 2907942768         | 332.5 GiB |
| ------------ | ------------ | ----------------- | ---------------------- | --------- |
| Total        |              |                   |                        | 1 TiB     |


| Table Name                 | # Entries  | Branch Pages | Leaf Pages | Overflow Pages | Total Size |
|----------------------------|------------|--------------|------------|----------------|------------|
| AccountChangeSets          | 3787160399 | 15867407     | 48833564   | 0              | 246.8 GiB  |
| AccountsHistory            | 67493      | 19           | 1907       | 1100           | 11.8 MiB   |
| AccountsTrie               | 24075389   | 4126         | 996709     | 0              | 3.8 GiB    |
| AccountsTrieChangeSets     | 125727     | 2989         | 16809      | 0              | 77.3 MiB   |
| BlockBodyIndices           | 23000001   | 609          | 136118     | 0              | 534.1 MiB  |
| BlockOmmers                | 1209744    | 844          | 188675     | 0              | 740.3 MiB  |
| BlockWithdrawals           | 5965061    | 3806         | 852142     | 0              | 3.3 GiB    |
| Bytecodes                  | 1757397    | 2835         | 179017     | 3608635        | 14.5 GiB   |
| CanonicalHeaders           | 0          | 0            | 0          | 0              | 0 B        |
| ChainState                 | 0          | 0            | 0          | 0              | 0 B        |
| HashedAccounts             | 304987658  | 43165        | 4100529    | 0              | 15.8 GiB   |
| HashedStorages             | 1339912717 | 612376       | 18592372   | 0              | 73.3 GiB   |
| HeaderNumbers              | 23000001   | 11299        | 547070     | 0              | 2.1 GiB    |
| HeaderTerminalDifficulties | 0          | 0            | 0          | 0              | 0 B        |
| Headers                    | 0          | 0            | 0          | 0              | 0 B        |
| Metadata                   | 1          | 0            | 1          | 0              | 4 KiB      |
| PlainAccountState          | 304987658  | 50115        | 4632009    | 0              | 17.9 GiB   |
| PlainStorageState          | 1339912717 | 883587       | 28323800   | 0              | 111.4 GiB  |
| PruneCheckpoints           | 1          | 0            | 1          | 0              | 4 KiB      |
| Receipts                   | 0          | 0            | 0          | 0              | 0 B        |
| StageCheckpointProgresses  | 1          | 0            | 1          | 0              | 4 KiB      |
| StageCheckpoints           | 16         | 0            | 1          | 0              | 4 KiB      |
| StorageChangeSets          | 6882009001 | 4504624      | 118596017  | 0              | 469.6 GiB  |
| StoragesHistory            | 538338     | 243          | 13720      | 213            | 55.4 MiB   |
| StoragesTrie               | 117839052  | 595741       | 6016100    | 0              | 25.2 GiB   |
| StoragesTrieChangeSets     | 205298     | 7876         | 30288      | 0              | 149.1 MiB  |
| TransactionBlocks          | 21360530   | 613          | 136927     | 0              | 537.3 MiB  |
| TransactionHashNumbers     | 2907942768 | 1436716      | 70050725   | 0              | 272.7 GiB  |
| TransactionSenders         | 2907942768 | 121328       | 27177036   | 0              | 104.1 GiB  |
| Transactions               | 0          | 0            | 0          | 0              | 0 B        |
| VersionHistory             | 1          | 0            | 1          | 0              | 4 KiB      |
| -------------------------- | ---------- | ------------ | ---------- | -------------- | ---------- |
| Tables                     |            |              |            |                | 1.3 TiB    |
| Freelist                   | 20893      |              |            |                | 81.6 MiB   |
```

This is about 40% savings, going from 246G on a regular node, to 140G with this PR


gakonst:
Super excited for this

joshieDo:
we might want to try this one with `lz4`

https://github.com/paradigmxyz/reth/blob/f0c0b3db4e47c167244654a5f03e5e388c9505fe/crates/storage/provider/src/providers/static_file/writer.rs#L833-L837

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -398,6 +438,93 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
         Ok(())
     }
 
+    /// Truncates account changesets to the given block. It deletes and loads and older static file
```
joshieDo:
```suggestion
    /// Truncates account changesets to the given block. It deletes and loads an older static file
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -851,18 +876,44 @@ impl<TX: DbTx, N: NodeTypes> AccountExtReader for DatabaseProvider<TX, N> {
         &self,
         range: RangeInclusive<BlockNumber>,
     ) -> ProviderResult<BTreeMap<Address, Vec<u64>>> {
-        let mut changeset_cursor = self.tx.cursor_read::<tables::AccountChangeSets>()?;
+        tracing::trace!(target: "provider::static_file", ?range, "Getting changesets for range");
+        let highest_static_block = self
+            .static_file_provider
+            .get_highest_static_file_block(StaticFileSegment::AccountChangeSets);
+
+        if let Some(highest) = highest_static_block {
```
joshieDo: is it possible to move this logic by impl `AccountExtReader` for `StaticFileProvider` similarly to the others ?

Rjected: yes, should be possible

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1710,21 +1793,52 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
-                )?;
+                // Convert to AccountBeforeTx format
+                let changesets: Vec<AccountBeforeTx> = account_block_reverts
+                    .into_iter()
+                    .map(|(address, info)| AccountBeforeTx { address, info: info.map(Into::into) })
+                    .collect();
+
+                // Increment block and append changesets
+                writer.increment_block(block_number)?;
+                writer.append_account_changeset(changesets)?;
```
joshieDo: seems like we're always calling `increment_block` before `append_account_changeset` - similar to `append_header`, should we just move it inside and update the doc? 

Or is there a scenario where we might not want to do that in the future?

Rjected: Yeah I think we can just call `increment_block` always

crates/storage/provider/src/providers/consistent.rs
```diff
@@ -1438,6 +1438,65 @@ impl<N: ProviderNodeTypes> ChangeSetReader for ConsistentProvider<N> {
...
+                let block_changesets = state
+                    .block_ref()
+                    .execution_output
+                    .bundle
+                    .reverts
+                    .clone()
+                    .to_plain_state_reverts()
+                    .accounts
+                    .into_iter()
+                    .flatten()
```
joshieDo, line 1470: nit: could this be a helper? feels like ive seen this before dupped

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1710,21 +1793,52 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
             }
         }
 
-        // Write account changes
-        tracing::trace!("Writing account changes");
-        let mut account_changeset_cursor =
-            self.tx_ref().cursor_dup_write::<tables::AccountChangeSets>()?;
+        // Write account changes to static files
+        tracing::trace!("Writing account changes to static files");
```
joshieDo: can we add a small note on the linked snippet that changesets get initialized on `insert_state` -> `write_state` -> `write_state_reverts` 
 
https://github.com/paradigmxyz/reth/blob/3ae73e63e567c67bb1f29552d93dc0d25c045dc8/crates/storage/db-common/src/init.rs#L159-L162

joshieDo: <s>maybe very edge case, but if the genesis has no state changes, i guess we'd never actually initialize the segment. (setting block range from `None` to `Some(0..=0)`)</s>

joshieDo: <s>i guess it can be generalized that if a block has no state changes (is it even possible?), we wouldnt increment the block either</s> 

nvm, we still iterate through the blocks even if they'd have no state changes https://github.com/paradigmxyz/reth/blob/3ae73e63e567c67bb1f29552d93dc0d25c045dc8/crates/storage/provider/src/providers/database/provider.rs#L1718

Rjected:
Update, this no longer has state root mismatches. Testing running this on a very large range of blocks to compare size versus the table in mdbx.

TODOs required for this PR to be complete so far:
 * Introduce method `is_changeset_based` instead of `is_block_based` / `is_transaction_based`
 * Use `StorageSettings` to determine behavior when reading / writing
 * Look into `test_segment_config_backwards`
 * Remove traces added
 * Put `increment_block` in `append_account_changeset`

crates/stages/stages/src/stages/merkle.rs
```diff
@@ -312,7 +313,7 @@ where
                     "Processing chunk"
                 );
                 let (root, updates) =
-                StateRoot::incremental_root_with_updates(provider.tx_ref(), chunk_range)
+                StateRoot::incremental_root_with_updates(provider, provider.tx_ref(), chunk_range)
```
joshieDo: should we make `tx_ref()` part of a trait and only pass provider or not worth it?

crates/stages/stages/src/stages/utils.rs
```diff
@@ -97,6 +98,76 @@ where
...
+        std::ops::Bound::Included(&n) => n,
+        std::ops::Bound::Excluded(&n) => n + 1,
+        std::ops::Bound::Unbounded => 0,
+    };
+
+    let end = match range.end_bound() {
+        std::ops::Bound::Included(&n) => n + 1,
+        std::ops::Bound::Excluded(&n) => n,
+        std::ops::Bound::Unbounded => BlockNumber::MAX,
+    };
```
joshieDo: can use https://github.com/paradigmxyz/reth/blob/3c39444597cac2c782f874f78cdb42aa0375290b/crates/storage/provider/src/lib.rs#L53-L67

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -15,25 +15,34 @@ pub struct StorageSettings {
...
     }
 
     /// Sets the `receipts_static_files` flag to true.
     pub const fn with_receipts_in_static_files(mut self) -> Self {
         self.receipts_in_static_files = true;
         self
     }
+
+    /// Sets the `account_changesets_in_static_files` flag to true.
+    pub const fn with_account_changesets_in_static_files(mut self) -> Self {
```
joshieDo: would be used here 

https://github.com/paradigmxyz/reth/blob/3c39444597cac2c782f874f78cdb42aa0375290b/crates/storage/db-common/src/init.rs#L166-L167

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -892,6 +943,13 @@ impl<TX: DbTx, N: NodeTypes> ChangeSetReader for DatabaseProvider<TX, N> {
         &self,
         block_number: BlockNumber,
     ) -> ProviderResult<Vec<AccountBeforeTx>> {
+        // Try static files first
```
joshieDo: for these things we can check the `cached_storage_settings` flag instead

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1701,21 +1784,51 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
+        // Write account changes to static files
+        tracing::trace!("Writing account changes to static files");
 
-        for (block_index, mut account_block_reverts) in reverts.accounts.into_iter().enumerate() {
-            let block_number = first_block + block_index as BlockNumber;
-            // Sort accounts by address.
-            account_block_reverts.par_sort_by_key(|a| a.0);
+        // Try to get a writer for AccountChangeSets static file
+        if let Ok(mut writer) =
+            self.static_file_provider.get_writer(first_block, StaticFileSegment::AccountChangeSets)
```
joshieDo: `EitherWriter` has been introduced here https://github.com/paradigmxyz/reth/blob/main/crates/storage/provider/src/either_writer.rs and maybe we could use it here as well?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -672,6 +811,45 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
+            metrics.record_segment_operations(
+                StaticFileSegment::AccountChangeSets,
+                StaticFileProviderOperation::Append,
+                count,
+                Some(start.elapsed()),
+            );
+        }
+
+        // Return the current block number if we have written any changes
+        Ok(if count > 0 { self.writer.user_header().block_end() } else { None })
```
joshieDo: could a chain (eth or otherwise) ever have a block with no account changesets? eg. no rewards, no txs?  and if that's the case why wouldnt we return the current block number? we still incremented it

joshieDo: also the function doc doesnt match, so unsure if we should be returning the block_end or count

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -525,6 +652,18 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
         Ok(())
     }
 
+    /// Appends change to changeset static file.
+    ///
+    /// Returns the current number of changesets in the block.
```
joshieDo: outdated doc?

crates/static-file/types/src/segment.rs
```diff
@@ -147,6 +203,66 @@ pub struct SegmentHeader {
     tx_range: Option<SegmentRangeInclusive>,
     /// Segment type
     segment: StaticFileSegment,
+    /// List of offsets, for where each block's changeset starts.
+    changeset_offsets: Option<Vec<ChangesetOffset>>,
```
joshieDo, line 235: couldnt we do `#[serde(default)]` instead of `SegmentHeaderVisitor`? or it doesnt play nicely with bincode?

crates/static-file/types/src/segment.rs
```diff
@@ -127,14 +151,46 @@ impl StaticFileSegment {
...
+}
+
+/// A changeset offset, also with the number of elements in the offset for convenience
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Hash, Clone)]
+pub struct ChangesetOffset {
+    /// Offset for the row for this block
+    offset: u64,
+
+    /// Number of changes in this changeset
+    num_changes: u64,
```
joshieDo, line 200: wondering if we can get away with u32 or u16?  but also not quite sure how bincode encodes uints, so maybe its fine

crates/static-file/types/src/segment.rs
Rjected: bincode doesnt like it https://github.com/bincode-org/bincode/issues/179

crates/storage/provider/src/providers/static_file/writer.rs
Rjected: oops yep

crates/storage/provider/src/providers/static_file/writer.rs
Rjected: I'm just going to remove the return value for this, we never use it

crates/storage/provider/src/providers/database/provider.rs
Rjected: Just extended `EitherWriter` with changeset stuff

crates/storage/provider/src/providers/database/provider.rs
Rjected: yep makes sense, changing this

crates/stages/stages/src/stages/utils.rs
Rjected: nice, using this now

crates/stages/stages/src/stages/merkle.rs
Rjected: Maybe not worth for this PR

crates/storage/provider/src/providers/database/provider.rs
Rjected: calls `increment_block` always now

crates/storage/provider/src/providers/database/provider.rs
Rjected: Added a note, yeah we still add a changeset offset for the block even if there are no changes

crates/static-file/types/src/segment.rs
```diff
@@ -155,6 +226,92 @@ pub struct SegmentHeader {
...
+    where
+        D: Deserializer<'de>,
+    {
+        // Tell the deserializer we're expecting a struct
+        // The field names are for formats that use them
+        // Bincode ignores these and just uses the sequence order
+        const FIELDS: &[&str] =
+            &["expected_block_range", "block_range", "tx_range", "segment", "changeset_offsets"];
+
+        deserializer.deserialize_struct("YourStruct", FIELDS, SegmentHeaderVisitor)
```
yongkangc: is `YourStruct` okay here?

Rjected: Ah no, will change

Rjected: Ah no, will change

Rjected: Ah no, will change

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -918,28 +967,79 @@ impl<TX: DbTx, N: NodeTypes> ChangeSetReader for DatabaseProvider<TX, N> {
...
+                .seek_by_key_subkey(block_number, address)?
+                .filter(|acc| acc.address == address)
+                .map(Ok)
+                .transpose()
+        }
+    }
+
+    fn account_changesets_range(
+        &self,
+        range: core::ops::Range<BlockNumber>,
```
yongkangc: curious why we are using Range (exclusive end) here but seems like for `changed_accounts_with_range` we are using `RangeInclusive`?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -918,28 +967,79 @@ impl<TX: DbTx, N: NodeTypes> ChangeSetReader for DatabaseProvider<TX, N> {
...
+    ) -> ProviderResult<Vec<(BlockNumber, AccountBeforeTx)>> {
+        let mut changesets = Vec::new();
+        if self.cached_storage_settings().account_changesets_in_static_files &&
+            let Some(highest) = self
+                .static_file_provider
+                .get_highest_static_file_block(StaticFileSegment::AccountChangeSets)
+        {
+            let static_end = range.end.min(highest + 1);
```
yongkangc, line 1028: would there be an issue if static files are enabled but only cover part of the requested range? 

e.g., static files have blocks 0-100, but user requests 0-150, would  this only returns blocks 0-100 from static files and silently drop blocks 101-150?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -460,7 +494,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         segment: StaticFileSegment,
         number: u64,
     ) -> ProviderResult<StaticFileJarProvider<'_, N>> {
-        if segment.is_block_based() {
+        if segment.is_block_based() || segment.is_change_based() {
```
yongkangc: nit: do you think it would be clearer if this condition 

`if segment.is_block_based() || segment.is_change_based() ` 

can be extracted into a separate function with some docstring on why this condition exist?

since it seems to appear a few times

yongkangc (commented):
it seems like there might be a similar pattern across the ops

```rust
// 1. Check settings.account_changesets_in_static_files
// 2. If true AND block is in static file range → use static file API
// 3. Else → use database API
// 4. (For ranges) → may need to query BOTH and merge

fn any_changeset_operation<T>(
    provider: &impl StorageSettingsCache + StaticFileProviderFactory + DBProvider,
    block: BlockNumber,
    operation: impl FnOnce(/* backend */) -> T,
) -> T {
    // Step 1: Check the setting
    let use_static_files = provider
        .cached_storage_settings()
        .account_changesets_in_static_files;

    // Step 2: For reads, also check if block is in static file range
    let highest_static = provider
        .static_file_provider()
        .get_highest_static_file_block(StaticFileSegment::AccountChangeSets);

    // Step 3: Route
    if use_static_files && highest_static.map_or(false, |h| block <= h) {
        // Static file backend
        operation(provider.static_file_provider())
    } else {
        // Database backend
        operation(provider.tx_ref())
    }
}
```

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -141,19 +157,37 @@ impl<N: NodePrimitives> StaticFileProviderBuilder<N> {
...
-        let provider = StaticFileProvider(Arc::new(self.inner));
+    pub fn build<N: NodePrimitives>(self) -> ProviderResult<StaticFileProvider<N>> {
+        let mut provider = StaticFileProviderInner::new(self.path, self.access)?;
+        if self.use_metrics {
+            provider.metrics = Some(Arc::new(StaticFileProviderMetrics::default()));
+        }
+
+        for (segment, blocks) in self.blocks_per_file {
+            provider.blocks_per_file.insert(segment, blocks);
+        }
```
shekhirin: can use `extend` here?

crates/trie/db/src/state.rs
```diff
@@ -129,10 +136,13 @@ pub trait DatabaseStateRoot<'a, TX>: Sized {
 pub trait DatabaseHashedPostState<TX>: Sized {
     /// Initializes [`HashedPostState`] from reverts. Iterates over state reverts in the specified
     /// range and aggregates them into hashed state in reverse.
+    ///
+    /// Requires a start bound in the range, otherwise this will return an error.
     fn from_reverts<KH: KeyHasher>(
+        provider: impl ChangeSetReader + BlockNumReader,
```
shekhirin: can this be 
```suggestion
        provider: impl ChangeSetReader + BlockNumReader + DBProvider,
```

so that we don't pass `tx` argument separately? And in other traits in this file

crates/static-file/types/src/segment.rs
```diff
@@ -155,6 +226,92 @@ pub struct SegmentHeader {
...
+
+        let segment =
+            seq.next_element()?.ok_or_else(|| serde::de::Error::invalid_length(3, &self))?;
+
+        let changeset_offsets = if segment == StaticFileSegment::AccountChangeSets {
+            // Try to read the 5th field (changeset_offsets)
+            // If it doesn't exist (old format), this will return None
+            match seq.next_element()? {
+                Some(Some(offsets)) => Some(offsets),
+                Some(None) | None => None,
```
shekhirin: hmm, is it even valid then? IIUC all changeset segment headers should have changeset offsets?

crates/static-file/types/src/segment.rs
```diff
@@ -155,6 +226,92 @@ pub struct SegmentHeader {
     tx_range: Option<SegmentRangeInclusive>,
     /// Segment type
     segment: StaticFileSegment,
+    /// List of offsets, for where each block's changeset starts.
+    changeset_offsets: Option<Vec<ChangesetOffset>>,
```
shekhirin, line 235: cc @joshieDo this makes the segment header size dynamic, is it ok?

crates/e2e-test-utils/src/setup_import.rs
```diff
@@ -113,7 +113,8 @@ pub async fn setup_engine_with_chain_import(
 
         // Initialize the database using init_db (same as CLI import command)
         // Use the same database arguments as the node will use
-        let db_args = reth_node_core::args::DatabaseArgs::default().database_args();
+        let db_args = reth_node_core::args::DatabaseArgs::default();
+        let db_args = db_args.database_args();
```
shekhirin: do we need these changes?

crates/node/core/src/args/static_files.rs
```diff
@@ -42,6 +46,10 @@ pub struct StaticFilesArgs {
     /// the node has been initialized, changing this flag requires re-syncing from scratch.
     #[arg(long = "static-files.transaction-senders")]
     pub transaction_senders: bool,
+
+    /// Storage account changesets in static files
```
shekhirin:
```suggestion
    /// Store account changesets in static files.
    
    /// When enabled, account changesets will be written to static files on disk instead of the
    /// database.
    ///
    /// Note: This setting can only be configured at genesis initialization. Once
    /// the node has been initialized, changing this flag requires re-syncing from scratch.
```

crates/node/core/src/args/static_files.rs
```diff
@@ -67,5 +78,6 @@ impl StaticFilesArgs {
         StorageSettings::legacy()
             .with_receipts_in_static_files(self.receipts)
             .with_transaction_senders_in_static_files(self.transaction_senders)
+            .with_account_changesets_in_static_files(self.account_change_sets)
```
shekhirin: nit for consistency:
```suggestion
            .with_account_changesets_in_static_files(self.account_changesets)
```

crates/stages/stages/src/stages/utils.rs
```diff
@@ -97,6 +98,69 @@ where
     Ok(collector)
 }
 
+/// Collects account history indices using a provider that implements `ChangeSetReader`.
+/// This version can read changesets from both static files and database.
```
shekhirin: I believe this is not true?
```suggestion
```

Can we unify this under the `EitherWriter` struct? Can be done in a separate PR

crates/static-file/types/src/segment.rs
```diff
@@ -33,6 +36,25 @@ pub enum StaticFileSegment {
     Receipts,
     /// Static File segment responsible for the `TransactionSenders` table.
     TransactionSenders,
+    #[strum(serialize = "accountchangesets")]
```
shekhirin: shouldn't be needed, it will be just `account-change-sets`

crates/static-file/types/src/segment.rs
```diff
@@ -128,21 +161,59 @@ impl StaticFileSegment {
     pub const fn is_tx_based(&self) -> bool {
         match self {
             Self::Receipts | Self::Transactions | Self::TransactionSenders => true,
-            Self::Headers => false,
+            Self::Headers | Self::AccountChangeSets => false,
+        }
+    }
+
+    /// Returns `true` if the segment is `StaticFileSegment::AccountChangeSets`
```
shekhirin:
```suggestion
    /// Returns `true` if the segment is [`StaticFileSegment::AccountChangeSets`]
```

crates/static-file/types/src/segment.rs
```diff
@@ -155,6 +226,92 @@ pub struct SegmentHeader {
...
+        state.serialize_field("block_range", &self.block_range)?;
+        state.serialize_field("tx_range", &self.tx_range)?;
+        state.serialize_field("segment", &self.segment)?;
+
+        if self.segment.is_account_change_sets() {
+            state.serialize_field("changeset_offsets", &self.changeset_offsets)?;
+        }
+
+        state.end()
+    }
```
shekhirin, line 325: should we validate here that `changeset_offsets.len() == block_range.len()`?

crates/static-file/types/src/segment.rs
```diff
@@ -239,7 +401,25 @@ impl SegmentHeader {
                 self.expected_block_start(),
             ));
             self.expected_block_start()
+        };
+
+        // For changeset segments, initialize an offset entry for the new block
+        if self.segment.is_change_based() {
+            let offsets = self.changeset_offsets.get_or_insert_with(Default::default);
```
shekhirin: `get_or_insert_default`?

crates/static-file/types/src/segment.rs
```diff
@@ -253,14 +433,50 @@ impl SegmentHeader {
         }
     }
 
+    /// Increments the latest block's number of changes.
+    pub fn increment_block_changes(&mut self) {
+        if self.segment.is_change_based() {
```
shekhirin, line 450: let's add a debug assertion here

crates/static-file/types/src/segment.rs
```diff
@@ -294,11 +535,34 @@ impl SegmentHeader {
 
     /// Returns the row offset which depends on whether the segment is block or transaction based.
     pub fn start(&self) -> Option<u64> {
+        // TODO: this should be solved instead by making account changesets not technically block
+        // based. block / tx / change-based should be exhaustive always
+        if self.segment.is_change_based() {
```
shekhirin: is this a blocker for merging this PR?

crates/static-file/types/src/segment.rs
```diff
@@ -294,11 +535,34 @@ impl SegmentHeader {
...
     }
+
+    /// Returns the `ChangesetOffset` corresponding for the given block, if it's in the block
+    /// range.
+    ///
+    /// If it is not in the block range or the changeset list in the header does not contain a
+    /// value for the block, this returns `None`.
+    pub fn changeset_offset(&self, block: BlockNumber) -> Option<&ChangesetOffset> {
+        let block_range = self.block_range()?;
+        if !(block_range.start()..=block_range.end()).contains(&block) {
```
shekhirin: let's add `contains()` to `SegmentRangeInclusive`

crates/stages/stages/src/stages/utils.rs
```diff
@@ -97,6 +98,69 @@ where
...
+            info!(target: "sync::stages::index_history", progress = %format!("{:.4}%", (idx as f64 / total_changesets as f64) * 100.0), "Collecting indices");
+        }
+
+        if block_number != current_block_number {
+            current_block_number = block_number;
+            flush_counter += 1;
+        }
+
+        if flush_counter > DEFAULT_CACHE_THRESHOLD {
+            collect(&cache)?;
```
shekhirin: can do `.drain()` here

crates/storage/db/src/static_file/masks.rs
```diff
@@ -48,3 +48,9 @@ add_static_file_mask! {
     #[doc = "Mask for selecting a single transaction sender from `TransactionSenders` static file segment"]
     TransactionSenderMask, Address, 0b1
 }
+
+// ACCOUNT CHANGESET MASKS
+add_static_file_mask! {
+    #[doc = "Mask for selecting a single changeset from `AccountChangeset` static file segment"]
```
shekhirin:
```suggestion
    #[doc = "Mask for selecting a single changeset from `AccountChangesets` static file segment"]
```

crates/trie/db/src/state.rs
```diff
@@ -226,17 +241,35 @@ impl<'a, TX: DbTx> DatabaseStateRoot<'a, TX>
...
+        // operations in the DB take either a `Range` or `RangeFrom`, using `RangeBounds` to create
+        // something iterable. It does this using `RangeWalker` which is not compatible with static
+        // files.
+        //
+        // To make this work with static file changesets, we get the highest block number and
+        // disallow lack of a start bound.
+        //
+        // TODO: resolve the API difference somehow
+        let Bound::Included(start) = range.start_bound() else {
+            return Err(ProviderError::UnboundedStartUnsupported);
```
shekhirin: can we just default to 0 here? It's `u64`, so it will be just `0` as the lowest bound

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -857,15 +857,38 @@ impl<TX: DbTx, N: NodeTypes> AccountReader for DatabaseProvider<TX, N> {
...
+            Ok(changed_accounts)
+        } else {
+            self.tx
+                .cursor_read::<tables::AccountChangeSets>()?
+                .walk_range(range)?
+                .map(|entry| {
+                    entry.map(|(_, account_before)| account_before.address).map_err(Into::into)
+                })
+                .collect()
+        }
```
shekhirin: I would prefer to move this and other methods into `EitherReader`

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1730,6 +1788,141 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
...
+                cursor.get_one::<reth_db::static_file::AccountChangesetMask>(mid.into())?
+            {
+                if change.address < address {
+                    low = mid + 1;
+                } else {
+                    high = mid;
+                }
+            } else {
+                // This is not expected but means we are out of the range / file somehow, and can't
+                // continue
```
shekhirin, line 1962: let's add a debug log

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1730,6 +1788,141 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
...
+        }
+
+        Ok(changesets)
+    }
+
+    fn account_changeset_count(&self) -> ProviderResult<usize> {
+        let mut count = 0;
+
+        // iterate through static files and sum changeset metadata via each static file header
+        let static_files = iter_static_files(&self.path).map_err(ProviderError::other)?;
```
shekhirin, line 2000: hmm this is not ideal because it reads and decodes all files, not just changesets. But I guess it's fine because it's used only in `collect_account_history_indices`

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -795,6 +941,44 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
         Ok(())
     }
 
+    /// Appends a block changeset to the static file.
+    ///
+    /// It **CALLS** `increment_block()`.
```
shekhirin, line 978: hmm, why different behaviour here compared to other methods?

crates/storage/provider/src/changeset_walker.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Yield remaining changesets from current block
+        if let Some(changeset) = self.current_changesets.get(self.changeset_index).cloned() {
+            self.changeset_index += 1;
+            return Some(Ok((self.current_block, changeset)));
+        }
+
+        // Advance to next block if we exhausted the previous one
+        if !self.current_changesets.is_empty() {
```
shekhirin, line 77: can we add a comment why this shouldn't be `if self.current_changests.is_empty()`? I don't understand the negation here

crates/storage/provider/src/either_writer.rs
```diff
@@ -212,6 +219,74 @@ where
...
+            Self::StaticFile(writer) => {
+                writer.append_account_changeset(changeset, block_number)?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl EitherWriter<'_, (), ()> {
```
shekhirin: we already have a `impl<'a> EitherWriter<'a, (), ()>` definition above, let's move it there

crates/trie/db/src/prefix_set.rs
```diff
@@ -83,3 +85,58 @@ impl<TX: DbTx, KH: KeyHasher> PrefixSetLoader<'_, TX, KH> {
         })
     }
 }
+
+/// Load prefix sets using a provider that implements `ChangeSetReader`.
```
shekhirin:
```suggestion
/// Load prefix sets using a provider that implements [`ChangeSetReader`].
```

crates/trie/db/src/prefix_set.rs
```diff
@@ -83,3 +85,58 @@ impl<TX: DbTx, KH: KeyHasher> PrefixSetLoader<'_, TX, KH> {
...
+/// This function can read changesets from both static files and database.
+pub fn load_prefix_sets_with_provider<Provider, KH>(
+    provider: &Provider,
```
shekhirin, line 92: same here, can we do `Provider: DBProvider` and do not pass a separate `tx`?

crates/trie/db/src/prefix_set.rs
```diff
@@ -83,3 +85,58 @@ impl<TX: DbTx, KH: KeyHasher> PrefixSetLoader<'_, TX, KH> {
...
+    Provider: ChangeSetReader,
+    KH: KeyHasher,
+{
+    // Initialize prefix sets.
+    let mut account_prefix_set = PrefixSetMut::default();
+    let mut storage_prefix_sets = HashMap::<B256, PrefixSetMut>::default();
+    let mut destroyed_accounts = HashSet::default();
+
+    // Get account changesets using the provider (handles static files + database)
+    let account_changesets = provider.account_changesets_range(*range.start()..*range.end() + 1)?;
```
shekhirin, line 107: would be nice for `account_changesets_range` to accept an `impl RangeBounds` so we don't have to do this

shekhirin (changes requested):

crates/static-file/types/src/segment.rs
Rjected: this TODO is actually solved with the `is_change_based` fn, will remove

crates/storage/provider/src/providers/static_file/writer.rs
Rjected: If we don't call `increment_block` in the fn, we would just be calling it every time we call `append_account_changeset`

joshieDo: `append_header`  behaves the same

crates/e2e-test-utils/src/setup_import.rs
```diff
@@ -466,7 +469,9 @@ mod tests {
         let datadir = temp_dir.path().join("datadir");
         std::fs::create_dir_all(&datadir).unwrap();
         let db_path = datadir.join("db");
-        let db_env = reth_db::init_db(&db_path, DatabaseArguments::default()).unwrap();
+        let db_args = reth_node_core::args::DatabaseArgs::default();
+        let db_args = db_args.database_args();
+        let db_env = reth_db::init_db(&db_path, db_args).unwrap();
```
mattsse: I see, this way we're going through the cli args

crates/static-file/types/src/segment.rs
mattsse: I'd assume that this is already respected

crates/storage/provider/src/providers/consistent.rs
```diff
@@ -1439,6 +1439,88 @@ impl<N: ProviderNodeTypes> ChangeSetReader for ConsistentProvider<N> {
...
+        let mut count = 0;
+        if let Some(head_block) = &self.head_block {
+            for state in head_block.chain() {
+                count += state
+                    .block_ref()
+                    .execution_output
+                    .bundle
+                    .reverts
+                    .clone()
+                    .to_plain_state_reverts()
```
mattsse, line 1515: uff, unsure how often we need this so might be fine

mattsse (commented):
@joshieDo and @shekhirin already flagged a few things

reading and writing logic, especially bounds checks, are always very complex, so perhaps we could add a few more helpful docs here and there, when we read/write sets for example

crates/e2e-test-utils/src/setup_import.rs
Rjected: No, artifact from when I had this in `DatabaseArgs` so can remove

Rjected: removed

crates/node/core/src/args/static_files.rs
Rjected: makes sense will rename

crates/stages/stages/src/stages/utils.rs
Rjected: yeah this is not correct any more

crates/storage/provider/src/changeset_walker.rs
Rjected: if we do not return from the previous condition (meaning `changesets.get(self.changeset_index)` is `None`), but `current_changesets` is non-empty, it means we have iterated through all changesets in the current block and need to fetch a changeset for the next block. Will add a comment for this

crates/static-file/types/src/segment.rs
Rjected: hmm yeah I think we should error for this case, since we should have offsets
in every changeset static file header

crates/storage/provider/src/providers/static_file/manager.rs
Rjected: added a helper `is_block_or_change_based` with a doc comment

crates/static-file/types/src/segment.rs
Rjected: oops yep

crates/storage/provider/src/providers/static_file/manager.rs
Rjected: added

crates/trie/db/src/state.rs
Rjected: Made this use DBProvider

crates/trie/db/src/prefix_set.rs
Rjected: made this use `DBProvider` as well

crates/stages/stages/src/stages/merkle.rs
Rjected: actually ended up doing this by using DBProvider

crates/static-file/types/src/segment.rs
Rjected: yeah I think we should actually

crates/storage/provider/src/providers/database/provider.rs
Rjected: Moved this to EitherReader

crates/stages/stages/src/stages/utils.rs
Rjected: yep, required some refactoring but now we drain here

crates/storage/provider/src/changeset_walker.rs
Rjected: added comment

crates/storage/provider/src/providers/database/provider.rs
Rjected: Probably, depends on what the caller expects

crates/storage/db-common/src/init.rs
```diff
@@ -129,6 +129,9 @@ where
         + AsRef<PF::ProviderRW>,
     PF::ChainSpec: EthChainSpec<Header = <PF::Primitives as NodePrimitives>::BlockHeader>,
 {
+    // Make sure to set storage settings before anything reads / writes
+    factory.set_storage_settings_cache(storage_settings);
```
joshieDo: think this should only be called when really initializing the genesis.  `ProviderFactory::new` fetches and caches it from storage if it exists: https://github.com/paradigmxyz/reth/blob/948966781425b6ee6d9fec7e2a1078d58eacec04/crates/storage/provider/src/providers/database/mod.rs#L94-L119. This would be replacing its cache to whatever default we end up changing to

maybe renaming this to `genesis_storage_settings` can remove future confusions?

joshieDo: we always call this when we launch the node btw: https://github.com/paradigmxyz/reth/blob/b5e7a694d284527a189467e4004abfbbc9ef51c6/crates/node/builder/src/launch/engine.rs#L125

joshieDo:
need a follow-up to handle the pruner: 
https://github.com/paradigmxyz/reth/blob/b5e7a694d284527a189467e4004abfbbc9ef51c6/crates/prune/prune/src/segments/user/account_history.rs#L80-L90

maybe, worth waiting for the rocksdb implementation of `AccountHistory`

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -28,9 +28,24 @@ pub struct StorageSettings {
...
     #[serde(default)]
     pub account_history_in_rocksdb: bool,
+    /// Whether this node should read and write account changesets from static files.
+    #[serde(default)]
+    pub account_changesets_in_static_files: bool,
 }
 
 impl StorageSettings {
+    /// Creates a new `StorageSettings` with default values.
+    pub const fn new() -> Self {
```
joshieDo: iirc, `fn new` was intentionally left out, so `StorageSettings` create would be explicit on whats creating

yongkangc (approved):
lgtm, dont have additional comments here that hasnt been brought up

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1057,6 +1089,14 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                         continue
                     }
                 }
+                StaticFileSegment::AccountChangeSets => {
+                    if EitherWriter::account_changesets_destination(provider).is_database() {
+                        // Old pruned nodes (including full node) do not store receipts as static
+                        // files.
+                        debug!(target: "reth::providers::static_file", ?segment, "Skipping account changesets consistency check: receipts stored in database");
```
meetrick: Should this be`account changesets stored in database` instead of `receipts`?
No functional impact, but this would be clearer for logs and future debugging.
The comment above (lines 1094–1095) also mentions `receipts`, while this block handles AccountChangeSets.

meetrick (commented):
LGTM

shekhirin (approved):

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: @Rjected

crates/storage/db-common/src/init.rs
Rjected: we now only call this when we actually are writing the genesis, fn also renamed

crates/storage/provider/src/providers/static_file/manager.rs
Rjected: fixed

crates/storage/db-api/src/models/metadata.rs
Rjected: removed

PR #18829, opened 2025-10-02
perf(trie): proofmanager optimisation WIP

yongkangc:
core changes:
- add worker pooling logic
- simplify channel design

impact:
- reduce overhead from task creation
- simplify abstractions

closes:
- [Multiproof Thread Pool Optimisation: Simplify ProofTaskManager Design #18801](https://github.com/paradigmxyz/reth/issues/18801)
- [Maintain a thread pool in ProofTaskManager #18735](https://github.com/paradigmxyz/reth/issues/18735)
- [Refactor decoded_multiproof to decouple db txn, storage coordination #18808](https://github.com/paradigmxyz/reth/issues/18808)
- [Use ProofTaskManager for account proofs from MultiProofTask #18734](https://github.com/paradigmxyz/reth/issues/18734)

Issue #18800, opened 2025-09-30
Tracking: Multiproof Thread Pool Optimization


jenpaff:
### Describe the feature

Multiproof generation has a two-level spawn cascade that creates performance overhead during block processing.

```
  Current cascade:
  MultiproofManager::spawn_multiproof()
    → spawn_blocking (Thread 1)
      → ParallelProof::decoded_multiproof()
        → spawn_storage_proof() × N
          → spawn_blocking (Thread 2...N+1)
```

**Problem Statement:** 
  1. Storage proofs (https://github.com/paradigmxyz/reth/issues/18735): https://github.com/paradigmxyz/reth/blob/6fabde4a043784b2da637f01b4bd7820cf01929f/crates/trie/parallel/src/proof_task.rs#L160 spawns a new
   tokio::spawn_blocking task for each proof, even though we maintain a pool of DB transactions
  2. Account proofs (https://github.com/paradigmxyz/reth/issues/18734): Spawned
  https://github.com/paradigmxyz/reth/blob/6fabde4a043784b2da637f01b4bd7820cf01929f/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L514 with two issues:
    - Spawn a new blocking task per multiproof
    - Don't reuse DB transactions (each creates new transaction via ConsistentDbView)

### Design

Replace per-proof spawning with two persistent ProofTaskManager pools:
- Account ProofTaskManager
- Storage ProofTaskManager

Workers spawn once, reuse DB transactions, communicate via oneshot channels.

### Impact
Eliminates two-level spawn cascade and DB transaction creation overhead. Replaces potentially 1000s of individual task spawns (1 + N per multiproof) with 8 persistent workers that reuse long-lived DB transactions. Removes primary bottleneck in StateRootTask during high-throughput operation.

PR #18887, opened 2025-10-07
perf(tree): worker pooling for storage in multiproof generation

yongkangc:
This PR add worker pooling to our existing proof generation. This PR starts by adding worker pooling to storage proofs, which allows us to reduce overhead from spawning the worker.

**Core logic:**
1. Determine workers based on parallelism 
2. Spawn x workers each with its own tx
3. Queue storage proofs via crossbeam channels to the workers

Issues closed:
- https://github.com/paradigmxyz/reth/issues/18735
- Parent: https://github.com/paradigmxyz/reth/issues/18800

References:
- Overall POC: https://github.com/paradigmxyz/reth/pull/18829
- Previously we had a PR up by [Rhovian](https://github.com/paradigmxyz/reth/pull/18752/files#diff-80873d1f05ecc984c7b969c759680e5fe91323687ff40d1b54e27644a9cd8d78). However the issue was that this is a larger refactors with significant back and forth, and therefore we had to bring this refactor in internally.

Next Steps: 
- Moving Account and Blinded nodes to adopt the same infrastructure

### Benches:

Reth4 Mainnet
image

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -45,7 +45,7 @@ use std::sync::{
     mpsc::{self, channel, Sender},
     Arc,
 };
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, warn};
 
 mod configured_sparse_trie;
 pub mod executor;
@@ -201,13 +201,28 @@ where
         let storage_worker_count = (max_proof_task_concurrency / 2)
             .max(1)
             .min(max_proof_task_concurrency.saturating_sub(1));
-        let proof_task = ProofTaskManager::new(
+        let proof_task = match ProofTaskManager::new(
             self.executor.handle().clone(),
             state_root_config.consistent_view.clone(),
             task_ctx,
             max_proof_task_concurrency,
             storage_worker_count,
-        );
+        ) {
+            Ok(proof_task) => proof_task,
+            Err(error) => {
+                // If we cannot bootstrap the proof task manager, continue with cache prewarming
+                // only; the caller will detect the missing state root channel and fall back to the
+                // parallel state root algorithm.
+                tracing::warn!(
+                    target: "engine::tree",
+                    ?error,
+                    max_concurrency = max_proof_task_concurrency,
+                    requested_workers = storage_worker_count,
+                    "Failed to initialize proof task manager, falling back to cache-only mode"
+                );
+                return self.spawn_cache_exclusive(env, transactions, provider_builder);
+            }
+        };
 
         // We set it to half of the proof task concurrency, because often for each multiproof we
         // spawn one Tokio task for the account proof, and one Tokio task for the storage proof.
@@ -470,6 +485,11 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
             .expect("state_root is None")
             .recv()
             .map_err(|_| ParallelStateRootError::Other("sparse trie task dropped".to_string()))?
+    }
+
+    /// Returns `true` if the handle is connected to a background state root task.
+    pub fn supports_state_root(&self) -> bool {
+        self.state_root.is_some()
     }
 
     /// Returns a state hook to be used to send state updates to this task.
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -488,7 +488,7 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
     }
 
     /// Returns `true` if the handle is connected to a background state root task.
-    pub fn supports_state_root(&self) -> bool {
+    pub const fn supports_state_root(&self) -> bool {
         self.state_root.is_some()
     }
 
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -199,8 +199,7 @@ where
         // Default to half of max concurrency, leaving room for on-demand tasks (Accountproof and
         // blinded nodes)
         let storage_worker_count = (max_proof_task_concurrency / 2)
-            .max(1)
-            .min(max_proof_task_concurrency.saturating_sub(1));
+            .max(1);
         let proof_task = match ProofTaskManager::new(
             self.executor.handle().clone(),
             state_root_config.consistent_view.clone(),
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -166,6 +166,10 @@ where
     ///
     /// This returns a handle to await the final state root and to interact with the tasks (e.g.
     /// canceling)
+    ///
+    /// Returns an error with the original transactions iterator if the proof task manager fails to
+    /// initialize.
+    #[allow(clippy::type_complexity)]
     pub fn spawn<P, I: ExecutableTxIterator<Evm>>(
         &mut self,
         env: ExecutionEnv<Evm>,
@@ -174,7 +178,10 @@ where
         consistent_view: ConsistentDbView<P>,
         trie_input: TrieInput,
         config: &TreeConfig,
-    ) -> PayloadHandle<WithTxEnv<TxEnvFor<Evm>, I::Tx>, I::Error>
+    ) -> Result<
+        PayloadHandle<WithTxEnv<TxEnvFor<Evm>, I::Tx>, I::Error>,
+        (reth_provider::ProviderError, I, ExecutionEnv<Evm>, StateProviderBuilder<N, P>),
+    >
     where
         P: DatabaseProviderFactory<Provider: BlockReader>
             + BlockReader
@@ -204,19 +211,15 @@ where
             max_proof_task_concurrency,
             storage_worker_count,
         ) {
-            Ok(proof_task) => proof_task,
+            Ok(task) => task,
             Err(error) => {
-                // If we cannot bootstrap the proof task manager, continue with cache prewarming
-                // only; the caller will detect the missing state root channel and fall back to the
-                // parallel state root algorithm.
+                // Fall back to parallel state root if proof task manager fails to initialize
                 tracing::error!(
                     target: "engine::tree",
                     ?error,
-                    max_concurrency = max_proof_task_concurrency,
-                    requested_workers = storage_worker_count,
-                    "Failed to initialize proof task manager, falling back to cache-only mode"
+                    "Failed to initialize proof task manager, falling back to parallel state root"
                 );
-                return self.spawn_cache_exclusive(env, transactions, provider_builder);
+                return Err((error, transactions, env, provider_builder));
             }
         };
 
@@ -269,12 +272,12 @@ where
             }
         });
 
-        PayloadHandle {
+        Ok(PayloadHandle {
             to_multi_proof,
             prewarm_handle,
             state_root: Some(state_root_rx),
             transactions: execution_rx,
-        }
+        })
     }
 
     /// Spawns a task that exclusively handles cache prewarming for transaction execution.
@@ -879,14 +882,20 @@ mod tests {
             PrecompileCacheMap::default(),
         );
         let provider = BlockchainProvider::new(factory).unwrap();
-        let mut handle = payload_processor.spawn(
-            Default::default(),
-            core::iter::empty::<Result<Recovered<TransactionSigned>, core::convert::Infallible>>(),
-            StateProviderBuilder::new(provider.clone(), genesis_hash, None),
-            ConsistentDbView::new_with_latest_tip(provider).unwrap(),
-            TrieInput::from_state(hashed_state),
-            &TreeConfig::default(),
-        );
+        let mut handle =
+            payload_processor
+                .spawn(
+                    Default::default(),
+                    core::iter::empty::<
+                        Result<Recovered<TransactionSigned>, core::convert::Infallible>,
+                    >(),
+                    StateProviderBuilder::new(provider.clone(), genesis_hash, None),
+                    ConsistentDbView::new_with_latest_tip(provider).unwrap(),
+                    TrieInput::from_state(hashed_state),
+                    &TreeConfig::default(),
+                )
+                .map_err(|(err, ..)| err)
+                .expect("failed to spawn payload processor");
 
         let mut state_hook = handle.state_hook();
 
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -213,7 +213,7 @@ where
                 // If we cannot bootstrap the proof task manager, continue with cache prewarming
                 // only; the caller will detect the missing state root channel and fall back to the
                 // parallel state root algorithm.
-                tracing::warn!(
+                tracing::error!(
                     target: "engine::tree",
                     ?error,
                     max_concurrency = max_proof_task_concurrency,
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -198,8 +198,7 @@ where
         let max_proof_task_concurrency = config.max_proof_task_concurrency() as usize;
         // Default to half of max concurrency, leaving room for on-demand tasks (Accountproof and
         // blinded nodes)
-        let storage_worker_count = (max_proof_task_concurrency / 2)
-            .max(1);
+        let storage_worker_count = (max_proof_task_concurrency / 2).max(1);
         let proof_task = match ProofTaskManager::new(
             self.executor.handle().clone(),
             state_root_config.consistent_view.clone(),
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -196,7 +196,10 @@ where
             state_root_config.prefix_sets.clone(),
         );
         let max_proof_task_concurrency = config.max_proof_task_concurrency() as usize;
-        let storage_worker_count = config.storage_proof_workers();
+        // Default to half of max concurrency, leaving room for on-demand tasks (Accountproof and blinded nodes)
+        let storage_worker_count = (max_proof_task_concurrency / 2)
+            .max(1)
+            .min(max_proof_task_concurrency.saturating_sub(1));
         let proof_task = ProofTaskManager::new(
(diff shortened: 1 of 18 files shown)


yongkangc:
image

Benching at [3fb97c6](https://github.com/paradigmxyz/reth/pull/18887/commits/3fb97c6977bb672f519a25355ab9e2e40edb045b)

crates/trie/parallel/src/proof_task.rs
```diff
@@ -48,62 +49,289 @@ use crate::proof_task_metrics::ProofTaskMetrics;
...
+            worker_id,
+            hashed_address = ?hashed_address,
+            proof_time_us = proof_elapsed.as_micros(),
+            total_processed = proofs_processed,
+            "Storage proof completed"
+        );
+    }
+
+    // Channel closed - graceful shutdown
+    tracing::info!(
```
mediocregopher:
```suggestion
    tracing::debug!(
```

crates/trie/parallel/src/proof_task.rs
```diff
@@ -48,62 +49,289 @@ use crate::proof_task_metrics::ProofTaskMetrics;
...
+                        ?err,
+                        requested = planned_workers,
+                        spawned_workers,
+                        "Failed to create transaction for storage worker, continuing with fewer workers"
+                    );
+                }
+            }
+        }
+
+        if spawned_workers == 0 {
```
mediocregopher: I'm not sure all this extra logic is really worth it, imo if there's an error spawning a worker we should just error out from `new` and the engine will fallback to the legacy state root.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -173,42 +401,107 @@ where
...
+                        },
+                        ProofTaskMessage::Transaction(tx) => {
+                            // Return transaction to on-demand pool
+                            self.on_demand_txs.push(tx);
+                        }
+                        ProofTaskMessage::Terminate => {
+                            // Drop storage_work_tx to signal workers to shut down
+                            drop(self.storage_work_tx);
+
+                            tracing::info!(
```
mediocregopher:
```suggestion
                            tracing::debug!(
```

crates/trie/parallel/src/proof_task.rs
```diff
@@ -260,7 +554,78 @@ where
...
+            worker_id = self.id,
+            "Completed storage proof calculation"
+        );
+
+        decoded_result
+    }
+
     /// Calculates a storage proof for the given hashed address, and desired prefix set.
+    ///
+    /// **ON-DEMAND VARIANT** - Consumes self, returns transaction to pool.
```
mediocregopher: this could be made to call `compute_storage_proof` internally, couldn't it? Then we wouldn't need to duplicate so much code

crates/trie/parallel/src/proof_task.rs
```diff
@@ -173,42 +401,107 @@ where
...
+                                        );
+
+                                        self.on_demand_queue.push_back(
+                                            ProofTaskKind::StorageProof(
+                                                job.input,
+                                                job.result_sender,
+                                            ),
+                                        );
+                                    }
+                                    Err(crossbeam_channel::TrySendError::Disconnected(job)) => {
```
mediocregopher: Under what circumstances could this occur? If I'm understanding right the worker loop can never error out until its own `work_rx` is disconnected, which would only happen once `ProofTaskMessage::Terminate` is received here. If that's the case then it'd be better to panic or `unreachable!` here, to clarify that this case isn't reachable.

mediocregopher (changes requested):

crates/trie/parallel/src/proof_task.rs
```diff
@@ -260,7 +554,78 @@ where
...
     }
 
+    /// Compute storage proof without consuming self.
+    ///
+    /// Borrows self immutably to allow transaction reuse across multiple calls.
+    /// Used by storage workers in the worker pool to avoid transaction creation
+    /// overhead on each proof computation.
+    #[inline]
+    fn compute_storage_proof(&self, input: StorageProofInput) -> StorageProofResult {
+        let (trie_cursor_factory, hashed_cursor_factory) = self.create_factories();
```
shekhirin: this should be initialized once per storage proof worker, because it's always the same

yongkangc: good note, thanks

crates/trie/parallel/src/proof_task.rs
yongkangc: yup, thanks for the note -  addressed in commit [b4bf193](https://github.com/paradigmxyz/reth/pull/18887/commits/b4bf19399484ce1a5c85b456656af2bc2c7c595a)

crates/trie/parallel/src/proof_task.rs
yongkangc: thanks for pointing out - - addressed in commit [b4bf193](https://github.com/paradigmxyz/reth/pull/18887/commits/b4bf19399484ce1a5c85b456656af2bc2c7c595a)

crates/trie/parallel/src/proof_task.rs
yongkangc: addressed by using unbounded channel [b2d5bcc](https://github.com/paradigmxyz/reth/pull/18887/commits/b2d5bcc7a7a36e9987533159e5fc5e096518aa77)

crates/trie/parallel/src/proof_task.rs
yongkangc: thats much better yea - addressed here [6282d2e](https://github.com/paradigmxyz/reth/pull/18887/commits/6282d2ed87967a2525e66b030dd34faec9db4ea3)

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -877,17 +877,29 @@ where
...
-                        StateRootStrategy::StateRootTask,
-                    )
+                    let handle = self.payload_processor.spawn(
+                        env,
+                        txs,
+                        provider_builder,
+                        consistent_view,
+                        trie_input,
+                        &self.config,
+                    );
```
shekhirin: can we instead make it fallible to it returns a `Result` that we match on here?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -158,57 +424,161 @@ where
...
+                                            ..
+                                        } = job
+                                        {
+                                            let _ = result_sender.send(Err(SparseTrieError::from(
+                                                Box::new(std::io::Error::new(
+                                                    std::io::ErrorKind::BrokenPipe,
+                                                    "Storage worker pool unavailable",
+                                                ))
+                                                    as Box<dyn std::error::Error + Send>,
+                                            )));
```
shekhirin: using `ParallelStateRootError::Other` should be fine here as it doesn't happen often

crates/trie/parallel/src/proof_task.rs
```diff
@@ -158,57 +424,161 @@ where
...
+                                            ..
+                                        } = job
+                                        {
+                                            let _ = result_sender.send(Err(
+                                                ParallelStateRootError::Other(
+                                                    "Storage proof worker pool unavailable"
+                                                        .to_string(),
+                                                ),
+                                            ));
+                                        }
```
shekhirin: can introduce a `StorageWorkerJob::result_sender` method to handle it nicer here and for `BlindedStorageNode`

yongkangc:
also from @shekhirin 

>can we re-purpose max_concurrency to be about on-demand workers, and precisely the storage_worker_count amount to be used for storage workers, so that we don't do that maths with planned_workers? And later we'll be able to control account and storage workers separately

crates/trie/parallel/src/proof_task.rs
```diff
@@ -40,73 +41,338 @@ use std::{
...
             max_concurrency,
+            "Initializing storage worker pool with unbounded queue"
+        );
+
+        let mut spawned_workers = 0;
+        for worker_id in 0..planned_workers {
+            let provider_ro = match view.provider_ro() {
+                Ok(provider_ro) => provider_ro,
+                Err(err) => {
+                    tracing::error!(
```
mediocregopher: This error message is no longer accurate I believe. Probably fine to just get rid of it, whatever receives this error will likely log as well

crates/trie/parallel/src/proof_task.rs
```diff
@@ -260,82 +631,79 @@ where
...
-            "Starting storage proof task calculation"
-        );
+        trie_cursor_factory: &InMemoryTrieCursorFactory<
+            DatabaseTrieCursorFactory<&Tx>,
+            &TrieUpdatesSorted,
+        >,
+        hashed_cursor_factory: &HashedPostStateCursorFactory<
+            DatabaseHashedCursorFactory<&Tx>,
+            &HashedPostStateSorted,
+        >,
```
mediocregopher: Not a big deal, but I think this would work
```suggestion
        trie_cursor_factory: impl TrieCursorFactory,
        hashed_cursor_factory: impl HashedCursorFactory,
```

crates/trie/parallel/src/proof_task.rs
yongkangc: thats way nicer - addressed that here [4680336](https://github.com/paradigmxyz/reth/pull/18887/commits/46803362c1802849150f581fa600b530793a4f99)

crates/trie/parallel/src/proof_task.rs
yongkangc: addressed that here [4680336](https://github.com/paradigmxyz/reth/pull/18887/commits/46803362c1802849150f581fa600b530793a4f99)

crates/trie/parallel/src/proof_task.rs
yongkangc: that looks cleaner. thanks! addressed that [2429320](https://github.com/paradigmxyz/reth/pull/18887/commits/2429320e294064e226d57d90606e1d7bcf23c5a5)

crates/trie/parallel/src/proof_task.rs
yongkangc: propogated error up - addressed [8fb0dd1](https://github.com/paradigmxyz/reth/pull/18887/commits/8fb0dd12a42cd5932789f414f29435cdec7d7e66)

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: much clearer - addressed that [1902b43](https://github.com/paradigmxyz/reth/pull/18887/commits/1902b433304d98a6d9cd1aeae2e2d8f5add84546)

yongkangc:
@mediocregopher @shekhirin addressed the comments above. Let me know what else

Also addressed this to put that in CLI
> can we re-purpose max_concurrency to be about on-demand workers, and precisely the storage_worker_count amount to be used for storage workers, so that we don't do that maths with planned_workers? And later we'll be able to control account and storage workers separately

yongkangc:
Latest Bench Compare Reth4 Mainnet with 200ms wait time

image
image

crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -2272,7 +2272,7 @@ mod tests {
 
             // Invalid/Non-existent argument should return `None`
             {
-                call_method!($arg_count, provider, $method, |_,_,_,_| ( ($invalid_args, None)), tx_num, tx_hash, &in_memory_blocks[0], &receipts);
+                call_method!($arg_count, provider, $method, |_,_,_,_| ($invalid_args, None), tx_num, tx_hash, &in_memory_blocks[0], &receipts);
```
yongkangc: clippy was complaining

crates/trie/parallel/src/proof_task.rs
```diff
@@ -158,57 +434,135 @@ where
...
+                                    result_sender: sender,
+                                }) {
+                                    Ok(_) => {
+                                        tracing::trace!(
+                                            target: "trie::proof_task",
+                                            "Storage proof dispatched to worker pool"
+                                        );
+                                    }
+                                    Err(crossbeam_channel::SendError(job)) => {
+                                        tracing::warn!(
```
yongkangc:
```suggestion
                                        tracing::error!(
```

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -196,12 +203,25 @@ where
...
-        );
+            storage_worker_count,
+        ) {
+            Ok(task) => task,
+            Err(error) => {
+                // Fall back to parallel state root if proof task manager fails to initialize
+                tracing::error!(
+                    target: "engine::tree",
+                    ?error,
+                    "Failed to initialize proof task manager, falling back to parallel state root"
```
mediocregopher: super duper nit: at this point the payload processor only knows that it failed to initialize, it doesn't know what the consequences of that will be (ie that it will fallback to parallel state root). I would move this error message outside of this function and into the caller

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -466,6 +486,11 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
             .map_err(|_| ParallelStateRootError::Other("sparse trie task dropped".to_string()))?
     }
 
+    /// Returns `true` if the handle is connected to a background state root task.
+    pub const fn supports_state_root(&self) -> bool {
```
mediocregopher: I think this isn't used anymore, can be deleted

crates/trie/parallel/src/proof_task.rs
```diff
@@ -158,57 +434,135 @@ where
...
+                                    input,
+                                    result_sender: sender,
+                                }) {
+                                    Ok(_) => {
+                                        tracing::trace!(
+                                            target: "trie::proof_task",
+                                            "Storage proof dispatched to worker pool"
+                                        );
+                                    }
+                                    Err(crossbeam_channel::SendError(job)) => {
```
mediocregopher, line 472: Another case of "can this actually happen?" We know that the worker pool was successfully initialized, and each worker cannot break out of its loop without Terminate having been called, and the channel is unbounded. If it's not really possible for this to happen then I say let's `self.storage_work_tx.send(...).expect(...)`

mediocregopher (approved):
Minor stuff but LGTM

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: yeah that makes sense, since we are alraedy propagting it up and only the parent knows

crates/trie/parallel/src/proof_task.rs
yongkangc: makes sense, let me address this in the next PR (i have already made some breaking change there based on this)! thanks for the note.

yongkangc: makes sense, let me address this in the next PR (i have already made some breaking change there based on this)! thanks for the note.

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: this makes sense, given that only the parent knows. and its already propogated

crates/trie/parallel/src/proof_task.rs
```diff
@@ -40,73 +41,348 @@ use std::{
...
+                let error = SparseTrieError::from(SparseTrieErrorKind::Other(Box::new(
+                    std::io::Error::new(
+                        std::io::ErrorKind::BrokenPipe,
+                        "Storage worker pool unavailable",
+                    ),
+                )));
+                result_sender.send(Err(error)).map_err(|_| ())
+            }
+        }
+    }
```
shekhirin, line 93: can unify the match arms and send the same `ParallelStateRootError::Other` error

crates/trie/parallel/src/proof_task.rs
```diff
@@ -40,73 +41,348 @@ use std::{
...
+        }
+
+        Ok(Self {
+            storage_work_tx,
+            storage_worker_count: spawned_workers,
             max_concurrency,
             total_transactions: 0,
+            pending_tasks: VecDeque::new(),
+            proof_task_txs: Vec::with_capacity(max_concurrency), /* used for on-demand account
+                                                                  * trie operations */
```
shekhirin:
```suggestion
            proof_task_txs: Vec::with_capacity(max_concurrency),
```

crates/trie/parallel/src/proof_task.rs
```diff
@@ -158,57 +434,135 @@ where
...
         Ok(())
     }
 
     /// Loops, managing the proof tasks, and sending new tasks to the executor.
+    ///
+    /// # Task Routing
+    ///
+    /// - **Storage Trie Operations** (`StorageProof` and `BlindedStorageNode`): Routed to
+    ///   pre-spawned worker pool via unbounded channel. Only falls back to `pending_tasks` if
+    ///   workers are disconnected (e.g., all workers panicked).
```
shekhirin:
```suggestion
    ///   pre-spawned worker pool via unbounded channel.
```

crates/trie/parallel/src/proof_task.rs
yongkangc: thanks just addressed this [f823c6b](https://github.com/paradigmxyz/reth/pull/18887/commits/f823c6bcd94211ab85ab709786ec0ceaf00f8150)

crates/trie/parallel/src/proof_task.rs
```diff
@@ -40,73 +41,348 @@ use std::{
...
+fn storage_worker_loop<Tx>(
+    proof_tx: ProofTaskTx<Tx>,
+    work_rx: CrossbeamReceiver<StorageWorkerJob>,
+    worker_id: usize,
+) where
+    Tx: DbTx,
+{
+    tracing::debug!(
+        target: "trie::proof_task",
+        worker_id,
```
shekhirin, line 200: nit for the next PR to not cause conflicts: let's create a tracing span with the worker id and remove manual `worker_id` field from every event

shekhirin (approved):

crates/trie/parallel/src/proof_task.rs
yongkangc: thats much better, thanks

crates/trie/parallel/src/proof_task.rs
shekhirin: sorry I meant that we can have one `Self::StorageProof { result_sender, .. } | Self::BlindedStorageNode { result_sender, .. }` arm, and send a `ParallelStateRootError::Other` on `result_sender`

PR #18888, opened 2025-10-07
chore(optimism): Remove redundant inherent with_conditional

Galoretka:
Remove the inherent OpPooledTransaction::with_conditional method and rely on the default with_conditional provided by MaybeConditionalTransaction. This eliminates duplicated API surface and aligns with the interop pattern (which only uses the trait default). No in-repo usages were found; behavior remains unchanged for callers that import the trait.


klkvr:
this helper is fine, allows to avoid an extra trait import

PR #18891, opened 2025-10-07
fix(trie): Reveal extension child when extension is last remaining child of a branch

mediocregopher:


Rjected (approved):
this makes sense to me

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1228,6 +1196,86 @@ impl ParallelSparseTrie {
...
+    ///
+    /// The node of the remaining child, whether it was already revealed or not.
+    fn reveal_remaining_child_on_leaf_removal<P: TrieNodeProvider>(
+        &mut self,
+        provider: P,
+        full_path: &Nibbles, // only needed for logs
+        remaining_child_path: &Nibbles,
+        recurse_into_extension: bool,
+    ) -> SparseTrieResult<SparseNode> {
+        let remaining_child_subtrie = self.subtrie_for_path_mut(remaining_child_path);
```
yongkangc, line 1215: much clearer logic

crates/trie/sparse/src/trie.rs
```diff
@@ -1256,6 +1235,83 @@ impl SerialSparseTrie {
...
+    ) -> SparseTrieResult<SparseNode> {
+        let remaining_child_node = match self.nodes.get(remaining_child_path).unwrap() {
+            SparseNode::Hash(_) => {
+                debug!(
+                    target: "trie::parallel_sparse",
+                    child_path = ?remaining_child_path,
+                    leaf_full_path = ?full_path,
+                    "Node child not revealed in remove_leaf, falling back to db",
+                );
+                if let Some(RevealedNode { node, tree_mask, hash_mask }) =
```
yongkangc, line 1262: maybe we could add one comment to show the logic? found it hard to follow and find: 

```
If `recurse_into_extension` is true, and the remaining child is an extension node, then its
    /// child will be ensured to be revealed as well.
```

yongkangc (approved):
much clearer, the idea makes sense to me. left a small comment

crates/trie/sparse/src/trie.rs
mediocregopher: Fair enough, added some more exposition around the check for the extension node :+1:

PR #18893, opened 2025-10-07
feat(prune): Add `MerkleChangeSets` segment

RomanHodulak:
Closes #18463 

Adds a pruner for change sets.

Change sets are newly persisted in the database and need to be pruned along with the rest of the entities to preserve space on full nodes.

### DIFF
--- a/crates/prune/prune/src/segments/user/merkle_change_sets.rs
+++ b/crates/prune/prune/src/segments/user/merkle_change_sets.rs
@@ -1,0 +1,107 @@
+use crate::{
+    db_ext::DbTxPruneExt,
+    segments::{PruneInput, Segment},
+    PrunerError,
+};
+use reth_db_api::{models::BlockNumberAddress, table::Value, tables, transaction::DbTxMut};
+use reth_primitives_traits::NodePrimitives;
+use reth_provider::{
+    errors::provider::ProviderResult, BlockReader, DBProvider, NodePrimitivesProvider,
+    PruneCheckpointWriter, TransactionsProvider,
+};
+use reth_prune_types::{
+    PruneCheckpoint, PruneMode, PrunePurpose, PruneSegment, SegmentOutput, SegmentOutputCheckpoint,
+};
+use tracing::{instrument, trace};
+
+#[derive(Debug)]
+pub struct MerkleChangeSets {
+    mode: PruneMode,
+}
+
+impl MerkleChangeSets {
+    pub const fn new(mode: PruneMode) -> Self {
+        Self { mode }
+    }
+}
+
+impl<Provider> Segment<Provider> for MerkleChangeSets
+where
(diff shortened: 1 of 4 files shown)


crates/prune/prune/src/segments/user/merkle_change_sets.rs
```diff
@@ -0,0 +1,87 @@
...
+            trace!(target: "pruner", "No change sets to prune");
+            return Ok(SegmentOutput::done())
+        };
+        let block_range_end = *block_range.end();
+
+        let mut limiter = input.limiter;
+
+        let mut last_pruned_block = block_range_end;
+        let (pruned, done) =
+            provider.tx_ref().prune_table_with_range::<tables::AccountsTrieChangeSets>(
```
mediocregopher, line 79: We also need to prune the `StoragesTrieChangeSets` table

crates/prune/types/src/segment.rs
```diff
@@ -15,6 +15,8 @@ pub enum PruneSegment {
     TransactionLookup,
     /// Prune segment responsible for all rows in `Receipts` table.
     Receipts,
+    /// Prune segment responsible for all rows in `AccountTrieChangeSets` table.
+    ChangeSets,
```
mediocregopher: I'm not sure if it's important to keep naming consistent with the pipeline stages, but if so this should probably be `MerkleChangeSets`. It should at least differentiate itself as being specifically for the trie tables, as the `AccountHistory` and `StorageHistory` segments currently cover the Account/StorageChangeSets tables, respectively

mediocregopher (changes requested):
One missing functionality that should be easy to fix, plus naming suggestion.

Question: when pruning runs, does it also update the pipeline sync checkpoint which is stored? Or how should the engine check what range of data is actually available to be used?

crates/prune/prune/src/segments/user/merkle_change_sets.rs
```diff
@@ -0,0 +1,101 @@
...
+        };
+
+        let block_range_end = *block_range.end();
+        let mut limiter = input.limiter;
+        let mut last_pruned_block = block_range_end;
+
+        for block in block_range.clone() {
+            let (pruned, done) =
+                provider.tx_ref().prune_table_with_range::<tables::StoragesTrieChangeSets>(
+                    BlockNumberHashedAddress::from((block, B256::ZERO))..,
```
mediocregopher: I think this won't actually stop at `block+1` when it's pruning, will it? So this could go overboard. If you look at how the StorageHistory segment is pruning the StorageChangeSets table I think this will be roughly the same thing, including the logic around `last_changeset_pruned_block` to make sure partially pruned blocks are resumed from their start.

mediocregopher (changes requested):

mediocregopher (approved):
LGTM 👍

PR #18779, opened 2025-09-30
chore: update ETHEREUM_BLOCK_GAS_LIMIT

0xKarl98:
This closes issue #18775

PR #18897, opened 2025-10-07
fix: align payload builder defaults with 30M gas limit

futreall:
Description:
## Summary
- keep `PayloadBuilderConfig::gas_limit_for` at the 30M target that the rest of the stack enforces
- add unit tests that capture the expected defaults and user overrides so future bumps are coordinated

## Background
PR #18779 bumped the payload builder’s default gas limit to 60M in anticipation of a future mainnet target. However, related components (transaction pool defaults, builder config, etc.) still operate with the 30M ceiling, and the consensus layer will currently reject 60M blocks outright. In practice this meant the builder consistently requested an unreachable target and disagreed with the rest of the stack.

## Changes
- return `ETHEREUM_BLOCK_GAS_LIMIT_30M` for Mainnet/Sepolia/Holesky/Hoodi unless the user overrides it
- add regression tests that flag any future default change unless the rest of the stack is updated transparently

## Testing
- cargo fmt
- cargo test -p reth-node-core
- cargo clippy --workspace --all-targets --all-features

### DIFF
--- a/crates/node/core/src/cli/config.rs
+++ b/crates/node/core/src/cli/config.rs
@@ -1,14 +1,11 @@
 //! Config traits for various node components.
 
-use alloy_eips::eip1559::ETHEREUM_BLOCK_GAS_LIMIT_36M;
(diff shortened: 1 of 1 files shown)


Rjected:
60M block gas limit is intended here

PR #18899, opened 2025-10-08
ci: cache hive simulator images to reduce prepare-hive job time

fgimenez:
then cache key depends on latest hive master commit seen on the build. `prepare-hive` job goes from ~15m https://github.com/paradigmxyz/reth/actions/runs/18340183869 to 6m https://github.com/paradigmxyz/reth/actions/runs/18341295331 when the cache is available


Rjected (approved):
cool, this makes sense to me

Issue #18801, opened 2025-10-01
Multiproof Thread Pool Optimisation: Simplify ProofTaskManager Design


yongkangc:
### Describe the feature

**Goal:**
- Simplify the manager loop using crossbeam `select!` with guard conditions. Channel becomes the natural queue.

**Expected Changes:** 
- Remove VecDeque from ProofTaskManager ([proof_task.rs:64](https://github.com/paradigmxyz/reth/blob/95e8a65d3342d862874b76469d72e34bca61b4e6/crates/trie/parallel/src/proof_task.rs#L63-L64) as Channel already provides queuing. Simplifies logic to pure select! loop
- We would remove these functions: [queue_proof_task](https://github.com/paradigmxyz/reth/blob/main/crates/trie/parallel/src/proof_task.rs#L122-L125), get_or_create_tx, [try_spawn_next](https://github.com/paradigmxyz/reth/blob/main/crates/trie/parallel/src/proof_task.rs#L122-L125) which all work on top of the vecdeque

**Testing**
- Add test for backpressure: queue 10000 tasks with 2 tx pool → verify serialization

### Additional context

This would likely have to be done in conjunction with https://github.com/paradigmxyz/reth/issues/18735

avdheshcharjan:
hi @yongkangc is this what you're looking for? #18828

avdheshcharjan:
https://github.com/paradigmxyz/reth/pull/18828

PR #18901, opened 2025-10-08
perf(tree): worker pooling for account proofs

yongkangc:
This PR builds on top of https://github.com/paradigmxyz/reth/pull/18887 by adding worker pooling to account and blinded account proofs, which allows us to reduce overhead from spawning the worker.

Changes:
- Move Account to pooling logic
- Move Blinded Account Nodes to pooling logic. Removed underlying structure and functions (pending tasks)

**Core logic:**
1. Determine workers based on parallelism 
2. Spawn x workers each with its own tx
3. Queue storage proofs via crossbeam channels to the workers
4. Await Storage proofs to build account proofs

```
Caller
    ↓ (requests account multiproof)
ProofTaskManager
    ↓ (routes to account worker pool)
Account Worker
    ↓ (calls collect_storage_proofs)
    ↓ (queues all storage proofs to storage worker pool)
Storage Workers (parallel computation)
    ↓ (return storage proofs)
Account Worker
    ↓ (calls build_account_multiproof_with_storage_roots)
    ↓ (walks account trie, looks up pre-computed storage roots)
Return to Caller (final account multiproof)
```

Issues closed:
- https://github.com/paradigmxyz/reth/issues/18734
- https://github.com/paradigmxyz/reth/issues/18833
- Parent: https://github.com/paradigmxyz/reth/issues/18800

References:
- Overall POC: https://github.com/paradigmxyz/reth/pull/18829

Next Steps: 
- Removing ProofTaskManager + other cleanups


yongkangc:
@mediocregopher 

re https://github.com/paradigmxyz/reth/pull/18887#discussion_r2413341145

I am thinking about it again, do we wanna expect or might be better to log an error? i am thinking there might not be a cost to erroring instead of expect?

crates/trie/parallel/src/proof.rs
```diff
@@ -187,164 +183,59 @@ where
...
+            multi_added_removed_keys: self.multi_added_removed_keys.clone(),
+            missed_leaves_storage_roots: self.missed_leaves_storage_roots.clone(),
+        };
 
-                    // Encode account
-                    account_rlp.clear();
-                    let account = account.into_trie_account(root);
-                    account.encode(&mut account_rlp as &mut dyn BufMut);
+        let (sender, receiver) = channel();
+        self.storage_proof_task_handle
```
mediocregopher: Let's rename this handle for clarity, even if it's going to get reworked in the next PR anyway

crates/trie/parallel/src/proof_task.rs
```diff
@@ -790,10 +1128,10 @@ impl ProofTaskCtx {
 pub enum ProofTaskMessage<Tx> {
     /// A request to queue a proof task.
     QueueTask(ProofTaskKind),
-    /// A returned database transaction.
-    Transaction(ProofTaskTx<Tx>),
     /// A request to terminate the proof task manager.
     Terminate,
+    #[doc(hidden)]
+    _Phantom(std::marker::PhantomData<Tx>),
```
mediocregopher: Why is it important to keep the Tx generic?

crates/trie/parallel/src/proof.rs
```diff
@@ -356,13 +247,142 @@ where
...
-            branch_node_tree_masks,
-            storages: collected_decoded_storages,
-        })
+        Ok(multiproof)
+    }
+}
+
+/// Builds an account multiproof given pre-collected storage proofs.
+///
+/// This is a helper function used by both `decoded_multiproof` and account workers to build
```
mediocregopher: I don't think `decoded_multiproof` is still using this, is it? It seems to only be used by the `account_worker_loop`. If so maybe we could move this function into `proof_task.rs` and make it private to that module.

yongkangc: @mediocregopher `decoded_multiproof` is still using it here

https://github.com/paradigmxyz/reth/blob/8be550a54201c512698f22345a3b78fb8faaf100/crates/trie/parallel/src/proof.rs#L195

mediocregopher: I'm referring to the `build_account_multiproof_with_storage_roots` function, not the prefix_sets one

yongkangc: sorry i misunderstood, good idea 👍🏻

yongkangc: addressed in [84b0a4f](https://github.com/paradigmxyz/reth/pull/18901/commits/84b0a4faf7f22d40c01b87b8e795b9ef4471a3a7)

crates/trie/parallel/src/proof_task.rs
yongkangc: it was originally for factory trait. but removed `    _Phantom(std::marker::PhantomData<Tx>),
`  and reduced the scope of the trait bound for factory.

crates/trie/parallel/src/proof.rs
```diff
@@ -58,34 +42,35 @@ pub struct ParallelProof<Factory: DatabaseProviderFactory> {
...
     multi_added_removed_keys: Option<Arc<MultiAddedRemovedKeys>>,
-    /// Handle to the storage proof task.
-    storage_proof_task_handle: ProofTaskManagerHandle<FactoryTx<Factory>>,
+    /// Handle to the proof task manager.
+    proof_task_handle: ProofTaskManagerHandle,
     /// Cached storage proof roots for missed leaves; this maps
     /// hashed (missed) addresses to their storage proof roots.
     missed_leaves_storage_roots: Arc<DashMap<B256, B256>>,
+    /// Marker to keep the Factory type parameter.
+    _phantom: std::marker::PhantomData<Factory>,
```
mediocregopher: Does ParallelProof need the Factory generic for some reason?

mediocregopher (commented):
One final question

crates/trie/parallel/src/proof.rs
yongkangc: Nope it doesn't, but I think there might have been other deps and I didn't want to remove it in this pr

yongkangc: Nope it doesn't, but I think there might have been other deps and I didn't want to remove it in this pr

crates/engine/primitives/src/config.rs
```diff
@@ -22,6 +22,16 @@ fn default_storage_worker_count() -> usize {
     }
 }
 
+/// Returns the default number of account worker threads optimized for I/O-bound coordination.
+///
+/// Account workers primarily coordinate storage proof collection and account trie traversal.
+/// They spend significant time blocked on `receiver.recv()` calls waiting for storage proofs,
+/// so we use higher concurrency (1.5x storage workers) to maximize throughput and overlap.
+/// While storage workers are CPU-bound, account workers are I/O-bound coordinators.
```
Rjected: is this comment true? I would expect the storage workers to be I/O bound as well

crates/trie/parallel/src/proof.rs
Rjected: let's make sure to remove this later if the generic is not needed

crates/trie/parallel/src/proof_task.rs
```diff
@@ -307,70 +302,450 @@ fn storage_worker_loop<Tx>(
     );
 }
 
-impl<Factory> ProofTaskManager<Factory>
+// TODO: Refactor this with storage_worker_loop. ProofTaskManager should be removed in the following
```
Rjected, line 287: I do think it is helpful to have the state required for each looping task, to be encapsulated in a struct rather than be a free function

crates/trie/parallel/src/proof_task.rs
```diff
@@ -307,70 +302,450 @@ fn storage_worker_loop<Tx>(
...
+#[allow(clippy::too_many_arguments)]
+fn build_account_multiproof_with_storage_roots<C, H>(
+    trie_cursor_factory: C,
+    hashed_cursor_factory: H,
+    targets: &MultiProofTargets,
+    prefix_set: PrefixSet,
+    collect_branch_node_masks: bool,
+    multi_added_removed_keys: Option<&Arc<MultiAddedRemovedKeys>>,
+    mut storage_proof_receivers: B256Map<Receiver<StorageProofResult>>,
+    missed_leaves_storage_roots: &DashMap<B256, B256>,
```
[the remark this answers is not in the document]
Rjected: agree with the comment that this is a lot of arguments, I think it would be nice to make this simpler by introducing some sort of input or context type

crates/trie/parallel/src/proof_task.rs
```diff
@@ -379,107 +754,100 @@ where
...
+            spawned_workers += 1;
+
+            tracing::debug!(
+                target: "trie::proof_task",
+                worker_id,
+                spawned_workers,
+                worker_type = %worker_type,
+                "{} worker spawned successfully", worker_type
+            );
+        }
```
Rjected: I understand that the `Fn` here is used to be generic over storage and account workers, and have only one function for spawning worker pools. But IMO this makes it harder to follow what's going on. What would be more clear to me, would be structs that encapsulate:
1. an account worker / loop
2. a storage worker / loop

And either duplicating the spawning code or finding another way to abstract the tasks

Rjected (commented):
The structure for workers makes sense to me, I have some comments on docs and style

crates/engine/primitives/src/config.rs
yongkangc: @Rjected looking at storage workers:
https://github.com/paradigmxyz/reth/blob/2091f0d18f766cc124b364902cf996eb2cbc3897/crates/trie/parallel/src/proof_task.rs#L592


implementation: https://github.com/paradigmxyz/reth/blob/1d1fea72b6f4d09911b0f414cc2d0dded462c00d/crates/trie/trie/src/proof/mod.rs#L297

storage worker does:
1. Database I/O: Creates cursors and walks the storage trie using TrieNodeIter (reads trie nodes from database)
2. CPU Work: For each node, calls hash_builder.add_branch() or hash_builder.add_leaf() (cryptographic hashing operations)
3. CPU Work: Computes final root hash with hash_builder.root() (more cryptographic hashing)

so imo storage workers does perform database I/O (reading trie nodes), tho it seems the bottleneck is the CPU-intensive Merkle proof computation, not the I/O. 

wdyt?

crates/trie/parallel/src/proof_task.rs
yongkangc: do you mean that we should keep the `ProofTaskManager`? 

imo think we can probably clean up this abstraction and simplify by having MultiProofManager directly dispatch to the workers without needing an intermidiary abstraction

crates/trie/parallel/src/proof_task.rs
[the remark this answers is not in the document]
yongkangc: makes sense, introduced `AccountMultiproofParams`

crates/trie/parallel/src/proof_task.rs
yongkangc: makes much more sense, addressed in [f23cfab](https://github.com/paradigmxyz/reth/pull/18901/commits/f23cfab54faf7e4cb58ca6a9807cf9d120fb3e0d) 

to me its much easier to follow.


Will also have another follow up pr to clean up 
https://github.com/paradigmxyz/reth/issues/18801

Let me know what u think :) @Rjected

crates/trie/parallel/src/proof.rs
yongkangc: yup have a pr up as draft

yongkangc: https://github.com/paradigmxyz/reth/pull/18933/files

crates/engine/primitives/src/config.rs
yongkangc: removed this for simplicity, and made the count the same as storage worker

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -698,7 +728,8 @@ where
             multiproof_manager: MultiproofManager::new(
                 executor,
                 metrics.clone(),
-                proof_task_handle,
+                proof_task_handle.clone(), // handle for storage proof workers
+                proof_task_handle,         // handle for account proof workers
```
shekhirin, line 717: if it uses the same proof task handle, I think it would be possible to deadlock the calculation if all concurrency is occupied with account proofs that are waiting on storage proofs?

yongkangc: @shekhirin it uses the same proof handle, but their pool is separate, so dont think deadlock is there. 

https://github.com/paradigmxyz/reth/blob/14def0761e4ce088f4adae56d8fb64ba4c0ad962/crates/trie/parallel/src/proof_task.rs#L312



not super clear from this code, but this will also be cleared up in refactor where we remove proof task manager: https://github.com/paradigmxyz/reth/pull/18934

shekhirin: ah right, sg

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -349,8 +353,11 @@ pub struct MultiproofManager<Factory: DatabaseProviderFactory> {
     pending: VecDeque<PendingMultiproofTask<Factory>>,
     /// Executor for tasks
     executor: WorkloadExecutor,
-    /// Sender to the storage proof task.
-    storage_proof_task_handle: ProofTaskManagerHandle<FactoryTx<Factory>>,
+    /// Handle to the proof task manager used for creating `ParallelProof` instances for storage
+    /// proofs.
+    storage_proof_task_handle: ProofTaskManagerHandle,
+    /// Handle to the proof task manager used for account multiproofs.
+    account_proof_task_handle: ProofTaskManagerHandle,
```
yongkangc, line 353: this isaddressed later on in the pr for cleanup where we use `proof_task_handle` instead

crates/trie/parallel/src/proof_task.rs
yongkangc: i.e https://github.com/paradigmxyz/reth/pull/18934/files

mediocregopher (approved):

mattsse (approved):
haven't looked too close into the actual proof logic

other than that lgtm

pending @Rjected

crates/trie/parallel/src/proof.rs
```diff
@@ -187,164 +161,58 @@ where
...
+        // Extend prefix sets with targets
+        let prefix_sets = Self::extend_prefix_sets_with_targets(&self.prefix_sets, &targets);
 
-        let storage_root_targets = StorageRootTargets::new(
-            prefix_sets.account_prefix_set.iter().map(|nibbles| B256::from_slice(&nibbles.pack())),
-            prefix_sets.storage_prefix_sets.clone(),
+        let storage_root_targets_len = StorageRootTargets::count(
+            &prefix_sets.account_prefix_set,
+            &prefix_sets.storage_prefix_sets,
         );
```
shekhirin, line 178: as a small optimization for future PRs, this calculation can be skipped if trace logs below won't be emitted

shekhirin (approved):
LGTM, my only suggestion is the same as Dan's — about free functions instead of structs with methods that keep the state.

crates/trie/parallel/src/proof.rs
yongkangc: Thanks for the note

crates/trie/parallel/src/proof_task.rs
Rjected: > do you mean that we should keep the ProofTaskManager?

no, I just mean that a struct, and a method on the struct called `run` or something similar, makes more sense to me than having `_loop` functions, since they are tasks with state

ie something like
```rust
let worker = AccountProofWorker::new(...);
executor.spawn_blocking(move || {
    worker.run();
});
```

makes more sense to me than how we currently spawn

Rjected (approved):
lgtm with the one suggestions about putting the worker state into structs

PR #18909, opened 2025-10-08
perf(tests): remove redundant format! in ef-tests run_only

prestoalvarez:
Description:
- Summary: Replace unnecessary string formatting with a direct literal in tests to avoid temporary allocation and improve readability.
- Change: In testing/ef-tests/tests/tests.rs, use stringify!($dir) instead of &format!("{}", stringify!($dir)).
- Rationale: Eliminates one allocation and simplifies the code.


mattsse (approved):
fine

PR #18910, opened 2025-10-08
fix: avoid NaN/Inf in gas_per_second by guarding zero-duration

prestoalvarez:
### Summary
Prevents division by zero in `gas_per_second` when execution is too fast, ensuring stable and consumable metrics.

### Rationale
NaN/Inf values can break metric backends, distort dashboards, and hinder observability.

### Implementation
- Add a single guard in `crates/evm/evm/src/metrics.rs` to set `gas_per_second` to `0.0` when `execution_duration == 0.0`.
- Include a brief comment explaining the rationale.

### DIFF
--- a/crates/evm/evm/src/metrics.rs
+++ b/crates/evm/evm/src/metrics.rs
@@ -50,7 +50,7 @@ impl ExecutorMetrics {
 
         // Update gas metrics.
         self.gas_processed_total.increment(gas_used);
-        self.gas_per_second.set(gas_used as f64 / execution_duration);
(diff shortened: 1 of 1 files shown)


crates/evm/evm/src/metrics.rs
```diff
@@ -50,7 +50,11 @@ impl ExecutorMetrics {
 
         // Update gas metrics.
         self.gas_processed_total.increment(gas_used);
-        self.gas_per_second.set(gas_used as f64 / execution_duration);
+        self.gas_per_second.set(if execution_duration > 0.0 {
+            gas_used as f64 / execution_duration
+        } else {
+            0.0
+        });
```
mattsse, line 57: when would this be 0.0?

PR #18912, opened 2025-10-09
deps: relax cc crate version from 1.2.15 to 1.2

mskrzypkows:


mattsse:
I think we can leave this pinned until actually working with newer 1.2 versions

PR #18918, opened 2025-10-09
fix(cli): allow disabling invalid block hooks by removing default “witness”

sashass1315:
Problem: The --debug.invalid-block-hook flag used default_value = "witness" with Option<_>, which always parsed as Some(Witness). This made the None branch (installing NoopInvalidBlockHook) unreachable via CLI, so users couldn’t disable hooks.

Solution: Removed the clap default for --debug.invalid-block-hook and set DebugArgs::default().invalid_block_hook = None. Updated tests to expect None by default and amended CLI docs to state hooks are disabled by default.

Impact: Omitting the flag now disables invalid block hooks; users must pass the flag to enable hooks (e.g., --debug.invalid-block-hook witness).

### DIFF
--- a/crates/node/core/src/args/debug.rs
+++ b/crates/node/core/src/args/debug.rs
@@ -78,8 +78,7 @@ pub struct DebugArgs {
     #[arg(
         long = "debug.invalid-block-hook",
         help_heading = "Debug",
-        value_parser = InvalidBlockSelectionValueParser::default(),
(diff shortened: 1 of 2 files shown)


mattsse (commented):
hmm, do we want to remove this @Rjected 

feels like having the witness hook as default doesnt hurt?

PR #18919, opened 2025-10-09
chore(trie): Remove ExecutedBlockWithTrieUpdates

### LINKED ISSUE #18467: Trie ChangeSets: Engine API TreeState

mediocregopher:
This PR removes the type `ExecutedBlockWithTrieUpdates`, as well as the related `ExecutedTrieUpdates` type. Computation of the `TrieInput` is modified to take advantage of the trie changesets tables to generate reverts as necessary, the same as is done for hashed post state. This allows us to therefore always have valid `TrieUpdates` for a block. With this in place we can remove a lot of extraneous logic and types.

There is a new DB method added in this PR which was necessary to produce `ExecutedBlocks` for historical blocks without re-executing them. The new method calculates the `TrieUpdates` (not the reverts!) for a historical block based on the changesets.

### DIFF
--- a/crates/engine/tree/src/tree/error.rs
+++ b/crates/engine/tree/src/tree/error.rs
@@ -1,7 +1,6 @@
 //! Internal errors for the tree module.
 
 use alloy_consensus::BlockHeader;
-use alloy_primitives::B256;
 use reth_consensus::ConsensusError;
 use reth_errors::{BlockExecutionError, BlockValidationError, ProviderError};
 use reth_evm::execute::InternalBlockExecutionError;
@@ -19,20 +18,6 @@ pub enum AdvancePersistenceError {
     /// A provider error
     #[error(transparent)]
     Provider(#[from] ProviderError),
-    /// Missing ancestor.
-    ///
-    /// This error occurs when we need to compute the state root for a block with missing trie
-    /// updates, but the ancestor block is not available. State root computation requires the state
-    /// from the parent block as a starting point.
-    ///
-    /// A block may be missing the trie updates when it's a fork chain block building on top of the
-    /// historical database state. Since we don't store the historical trie state, we cannot
-    /// generate the trie updates for it until the moment when database is unwound to the canonical
-    /// chain.
-    ///
-    /// Also see [`reth_chain_state::ExecutedTrieUpdates::Missing`].
-    #[error("Missing ancestor with hash {0}")]
-    MissingAncestor(B256),
 }
 
 #[derive(thiserror::Error)]
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -13,10 +13,8 @@ use alloy_rpc_types_engine::{
     ForkchoiceState, PayloadStatus, PayloadStatusEnum, PayloadValidationError,
 };
 use error::{InsertBlockError, InsertBlockFatalError};
-use persistence_state::CurrentPersistenceAction;
 use reth_chain_state::{
-    CanonicalInMemoryState, ExecutedBlock, ExecutedBlockWithTrieUpdates, ExecutedTrieUpdates,
-    MemoryOverlayStateProvider, NewCanonicalChain,
+    CanonicalInMemoryState, ExecutedBlock, MemoryOverlayStateProvider, NewCanonicalChain,
 };
 use reth_consensus::{Consensus, FullConsensus};
 use reth_engine_primitives::{
@@ -31,14 +29,12 @@ use reth_payload_primitives::{
 };
 use reth_primitives_traits::{NodePrimitives, RecoveredBlock, SealedBlock, SealedHeader};
 use reth_provider::{
-    providers::ConsistentDbView, BlockNumReader, BlockReader, DBProvider, DatabaseProviderFactory,
-    HashedPostStateProvider, ProviderError, StateProviderBox, StateProviderFactory, StateReader,
-    StateRootProvider, TransactionVariant,
+    providers::ConsistentDbView, BlockReader, DatabaseProviderFactory, HashedPostStateProvider,
+    ProviderError, StateProviderBox, StateProviderFactory, StateReader, TransactionVariant,
+    TrieReader,
 };
 use reth_revm::database::StateProviderDatabase;
 use reth_stages_api::ControlFlow;
-use reth_trie::{HashedPostState, TrieInput};
-use reth_trie_db::DatabaseHashedPostState;
 use revm::state::EvmState;
 use state::TreeState;
 use std::{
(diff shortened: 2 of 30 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -1847,11 +1749,13 @@ where
             .get_state(block.header().number())?
             .ok_or_else(|| ProviderError::StateForNumberNotFound(block.header().number()))?;
         let hashed_state = self.provider.hashed_post_state(execution_output.state());
+        let trie_updates = self.provider.get_block_trie_updates(block.number())?;
```
mediocregopher, line 1752: This is where the new db method was required to be used, to properly populate ExecutedBlocks that aren't in memory

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2604,109 +2486,6 @@ where
...
-    /// 1. Collect in-memory blocks that are descendants of the provided parent hash using
-    ///    [`TreeState::blocks_by_hash`].
-    /// 2. If the persistence is in progress, and the block that we're computing the trie input for
-    ///    is a descendant of the currently persisting blocks, we need to be sure that in-memory
-    ///    blocks are not overlapping with the database blocks that may have been already persisted.
-    ///    To do that, we're filtering out in-memory blocks that are lower than the highest database
-    ///    block.
-    /// 3. Once in-memory blocks are collected and optionally filtered, we compute the
-    ///    [`HashedPostState`] from them.
-    fn compute_trie_input<TP: DBProvider + BlockNumReader>(
```
mediocregopher, line 2622: We had an entire 1:1 copy of this function in this module for some reason. The version in `payload_validator.rs` remains.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1137,9 +1035,10 @@ where
                 storages = revert_state.storages.len(),
                 "Non-empty revert state"
             );
-            revert_state
+            (revert_state, revert_trie)
         };
-        input.append(revert_state);
+
+        input.append_cached(revert_trie.into(), revert_state);
```
mediocregopher, line 1041: `append_cached` specifically does not populate the prefix sets from the `revert_state` into the trie input. The trie state is fully reverted at this point, so the prefix set is not necessary.

crates/engine/tree/src/tree/state.rs
```diff
@@ -35,19 +25,15 @@ pub struct TreeState<N: NodePrimitives = EthPrimitives> {
...
     ///
     /// Note: there can be multiple blocks at the same height due to forks.
-    pub(crate) blocks_by_number: BTreeMap<BlockNumber, Vec<ExecutedBlockWithTrieUpdates<N>>>,
+    pub(crate) blocks_by_number: BTreeMap<BlockNumber, Vec<ExecutedBlock<N>>>,
     /// Map of any parent block hash to its children.
     pub(crate) parent_to_child: HashMap<B256, HashSet<B256>>,
-    /// Map of hash to trie updates for canonical blocks that are persisted but not finalized.
-    ///
-    /// Contains the block number for easy removal.
-    pub(crate) persisted_trie_updates: HashMap<B256, (BlockNumber, Arc<TrieUpdates>)>,
```
mediocregopher, line 50: This was only used (indirectly) by `canonical_block_by_hash` to construct historical ExecutedBlocks. We have the DB method for that now.

crates/trie/common/src/updates.rs
```diff
@@ -478,6 +478,29 @@ impl AsRef<Self> for TrieUpdatesSorted {
     }
 }
 
+impl From<TrieUpdatesSorted> for TrieUpdates {
```
mediocregopher, line 481: Required because TrieInputs still uses TrieUpdates, but trie_reverts outputs a TrieUpdatesSorted. In a later PR I intend on making TrieInputs use TrieUpdatesSorted as well.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2289,25 +2193,7 @@ where
...
-                        .get(&block.recovered_block.hash())?
-                        .1
-                        .clone();
-                    Some(ExecutedBlockWithTrieUpdates {
-                        block: block.clone(),
-                        trie: ExecutedTrieUpdates::Present(trie),
-                    })
-                })
-                .collect::<Vec<_>>();
-            self.reinsert_reorged_blocks(old);
```
Rjected, line 2310: so glad we can get rid of this

Rjected (approved):
makes sense to me, glad that we can get rid of so much code

crates/engine/tree/src/tree/error.rs
```diff
@@ -19,20 +18,6 @@ pub enum AdvancePersistenceError {
...
-    /// from the parent block as a starting point.
-    ///
-    /// A block may be missing the trie updates when it's a fork chain block building on top of the
-    /// historical database state. Since we don't store the historical trie state, we cannot
-    /// generate the trie updates for it until the moment when database is unwound to the canonical
-    /// chain.
-    ///
-    /// Also see [`reth_chain_state::ExecutedTrieUpdates::Missing`].
-    #[error("Missing ancestor with hash {0}")]
-    MissingAncestor(B256),
```
yongkangc, line 35: nice, we can remove whole class of error because now we generate trie data for any block from the database on the fly and therefore the parent block's (ancestor's) trie data can never be "missing"

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1696,17 +1659,9 @@ where
...
     /// order is oldest -> newest.
-    ///
-    /// If any blocks are missing trie updates, all blocks are persisted, not taking `threshold`
-    /// into account.
-    ///
-    /// For those blocks that didn't have the trie updates calculated, runs the state root
-    /// calculation, and saves the trie updates.
-    ///
-    /// Returns an error if the state root calculation fails.
     fn get_canonical_blocks_to_persist(
```
yongkangc, line 1662: i guess this is the main crux? - where we no longer have to persist blocks missing trie updates because we can compute them on demand

yongkangc (approved):
much clearer logic, nice

PR #18920, opened 2025-10-09
fix(stateless): enforce BLOCKHASH ancestor header limit

Forostovec:
- The AncestorHeaderLimitExceeded error was dead code: it was defined but never emitted. This contradicted the documentation, which stated that ancestor headers “do not exceed the BLOCKHASH opcode limit” during verification.
- Additionally, WitnessDatabase docs state the caller is responsible for enforcing the 256 ancestor window, but stateless_validation_with_trie did not enforce it.
- This change adds BLOCKHASH_ANCESTOR_LIMIT = 256 and validates the length of ancestor_headers after sorting, returning StatelessValidationError::AncestorHeaderLimitExceeded { count, limit } when exceeded.
- This aligns behavior with EVM BLOCKHASH semantics, activates the previously dead error, and prevents oversized/invalid inputs from silently proceeding.

### DIFF
--- a/crates/stateless/src/validation.rs
+++ b/crates/stateless/src/validation.rs
@@ -21,6 +21,9 @@ use reth_evm::{execute::Executor, ConfigureEvm};
 use reth_primitives_traits::{RecoveredBlock, SealedHeader};
 use reth_trie_common::{HashedPostState, KeccakKeyHasher};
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I see, this doesn't hurt but I guess we could also enforce that the lowest ancestor is within the 256 distance?

PR #18922, opened 2025-10-09
fix(trie): fix trie changesets checkpoint handling

### LINKED ISSUE #18903: Trie ChangSets: Fix pipeline checkpoints
### Describe the feature

There are two bugs regarding the pipeline checkpoint for the new MerkleChangeSets stage:

* Inside of the OverlayStateProviderFactory we are checking the checkpoint range within the StageCheckpoint, but in reality only the outer block_number field gets updated by the engine when processing blocks.

* When running on a dev box I noted that the pipeline sync stage is running with a massive range:
```
2025-10-08T09:49:23.823777Z  INFO Status connected_peers=130 stage=MerkleChangeSets checkpoint=23503319 target=23526851
```

This is because the pipeline was restarted after the node was down for a time. 23503319 was indeed the finalized block at the time the node was shut down, but we should limit the sync stage so as to not end up in this situation.

### Additional context

_No response_

mediocregopher:
Closes #18903

Fixes the two bugs described in that ticket

### DIFF
--- a/crates/stages/stages/src/stages/merkle_changesets.rs
+++ b/crates/stages/stages/src/stages/merkle_changesets.rs
@@ -81,7 +81,7 @@ impl MerkleChangeSets {
         // Use maximum of finalized_block and retention_based_start if finalized_block exists,
         // otherwise just use retention_based_start.
         let mut target_start = finalized_block
-            .map(|finalized| finalized.saturating_add(1).min(retention_based_start))
+            .map(|finalized| finalized.saturating_add(1).max(retention_based_start))
             .unwrap_or(retention_based_start);
 
         // We cannot revert the genesis block; target_start must be >0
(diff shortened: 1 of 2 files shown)


crates/stages/stages/src/stages/merkle_changesets.rs
```diff
@@ -73,8 +75,11 @@ impl MerkleChangeSets {
         // Calculate the fallback start position based on retention blocks
         let retention_based_start = merkle_checkpoint.saturating_sub(self.retention_blocks);
 
-        // Use minimum of finalized_block and retention_based_start if finalized_block exists,
-        // otherwise just use retention_based_start
+        // If the finalized block was way in the past then we don't want to generate changesets for
+        // all of those past blocks; we only care about the recent history.
+        //
+        // Use maximum of finalized_block and retention_based_start if finalized_block exists,
```
Rjected, line 81: This says maximum but the code still uses min?

Rjected (commented):
yeah I think using `checkpoint.block_number` is correct, just one question about the comment otherwise lgtm

crates/stages/stages/src/stages/merkle_changesets.rs
mediocregopher: good catch, i don't know how I missed this :facepalm:

PR #18923, opened 2025-10-09
fix: busy-wait in graceful shutdown loop

viktorking7:
**Description**

Replaced CPU-intensive `std::hint::spin_loop()` with a brief 1ms sleep in the `do_graceful_shutdown` method to reduce unnecessary CPU usage during graceful shutdown operations.

**Changes**
- Modified `TaskManager::do_graceful_shutdown` to use `std::thread::sleep(Duration::from_millis(1))` instead of busy-waiting
- Preserves existing timeout behavior and shutdown semantics

### DIFF
--- a/crates/tasks/src/lib.rs
+++ b/crates/tasks/src/lib.rs
@@ -247,7 +247,8 @@ impl TaskManager {
                 debug!("graceful shutdown timed out");
                 return false
             }
-            std::hint::spin_loop();
+            // Avoid busy-waiting; sleep briefly to reduce CPU usage during graceful shutdown loop.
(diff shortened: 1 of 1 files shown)


crates/tasks/src/lib.rs
```diff
@@ -247,7 +247,8 @@ impl TaskManager {
                 debug!("graceful shutdown timed out");
                 return false
             }
-            std::hint::spin_loop();
+            // Avoid busy-waiting; sleep briefly to reduce CPU usage during graceful shutdown loop.
```
mattsse, line 250: the busy loop is fine here because we also check timeout so this would only spin for a few seconds at most during shutdown which si fine

PR #18931, opened 2025-10-10
feat: add BlockAnnounce trait for outbound block propagation 

itsdevbear:
This PR adds a symmetric counterpart to BlockImport for announcing blocks to the network. This is useful for custom chains that want to leverage a singular p2p layer instead of running a separate consensus client. 

Scope:
- BlockAnnounce trait with poll-based interface
- Integration with NetworkManager's event loop
- Builder API for easy configuration
- PoS safety check to prevent invalid propagation
- Optional by default, backward compatible

Example Usage:
```rust
use reth_network::announce::{BlockAnnounce, BlockAnnounceRequest, PropagationStrategy};
use tokio::sync::mpsc;
use std::task::{Context, Poll};

#[derive(Debug)]
struct MyCustomBlockAnnouncer {
    rx: mpsc::UnboundedReceiver<(Block, B256)>,
}

impl BlockAnnounce<Block> for MyCustomBlockAnnouncer {
    fn on_announced_block(&mut self, _block: Block) {
        // do some stuff
    }

    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<BlockAnnounceRequest<Block>> {
        match self.rx.poll_recv(cx) {
            Poll::Ready(Some((block, hash))) => {
                Poll::Ready(BlockAnnounceRequest::Announce {
                    block,
                    hash,
                    strategy: PropagationStrategy::Both, // Full block + hash
                })
            }
            _ => Poll::Pending,
        }
    }
}

// Usage:
let (to_announcer, from_sender) = mpsc::unbounded_channel();
let announcer = Box::new(MyCustomBlockAnnouncer { from_sender });

let config = NetworkConfig::builder(secret_key)
    .block_announce(announcer)
    .build(client);

// When you produce a block:
to_announcer.send((new_block, block_hash)).unwrap();
// NetworkManager will poll and announce it automatically, and then call the `on_announced_block` callback
```


crates/net/network/src/announce.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    /// Creates a test context with a noop waker
+    fn with_context<F, R>(f: F) -> R
+    where
+        F: FnOnce(&mut Context<'_>) -> R,
+    {
+        let waker = create_waker();
+        let mut cx = Context::from_waker(&waker);
+        f(&mut cx)
+    }
```
mattsse, line 162: this looks weird, why do we need all of this

I'd rather use tokio::test and poll_fn instead of this

crates/net/network/src/manager.rs
```diff
@@ -649,15 +707,6 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
...
             }
-            NetworkHandleMessage::AnnounceBlock(block, hash) => {
-                if self.handle.mode().is_stake() {
-                    // See [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675#devp2p)
-                    warn!(target: "net", "Peer performed block propagation, but it is not supported in proof of stake (EIP-3675)");
-                    return
-                }
-                let msg = NewBlockMessage { hash, block: Arc::new(block) };
-                self.swarm.state_mut().announce_new_block(msg);
-            }
```
mattsse, line 667: hmm, this likely clashes with regular devp2p broadcasting feature because now this is no longer supported, right?

crates/net/network/src/network.rs
```diff
@@ -483,8 +473,6 @@ pub(crate) enum NetworkHandleMessage<N: NetworkPrimitives = EthNetworkPrimitives
     RemovePeer(PeerId, PeerKind),
     /// Disconnects a connection to a peer if it exists, optionally providing a disconnect reason.
     DisconnectPeer(PeerId, Option<DisconnectReason>),
-    /// Broadcasts an event to announce a new block to all nodes.
-    AnnounceBlock(N::NewBlockPayload, B256),
```
mattsse, line 510: why is this removed?

seems like we can keep this?

mattsse (commented):
not strictly opposed to this, but it seems like the same thing can also be achieved with a separate task instead that simply sends the announce messages via the networkhandle, this is how this was intended.

so unsure if we really need this, can you elaborate on what benefits this brings instead of a separate announce task

adding the announce strategy makes sense, perhaps we can start with that?

crates/net/network/src/manager.rs
itsdevbear: Removal of this was due to the removal of the NetworkMessage, I added the check for EIP 3675 in the Polling function on line 1149.

crates/net/network/src/network.rs
itsdevbear: If we want to keep this I suggest adding `AnnounceBlockHashes` or changing to  `AnnounceBlock(N::NewBlockPayload, B256, AnnounceStrategy)`,

itsdevbear: My logic for removal was to impose symmetry to developers with `BlockImport` being for imports and `BlockAnnouncer` being for announcements.

crates/net/network/src/announce.rs
itsdevbear: Yup, this was lazy on my part, will change.

crates/net/network/src/manager.rs
itsdevbear: It still works well with my change (arguably its easier to understand), would it be helpful if I created and example in `/examples`?

mattsse (commented):
I think there isnt a real usecase for this but can be convinced otherwise

PR #18934, opened 2025-10-10
refactor(trie): remove proof task manager

### LINKED ISSUE #18801: Multiproof Thread Pool Optimisation: Simplify ProofTaskManager Design

yongkangc:
### Context:
- As part of our performance work to reduce overhead and improve scheduling, we added worker pooling for multiproof generation.
- This PR aims to perform cleanup and remove ProofTaskManager as an abstraction as we can now directly dispatch the proofs jobs to workers. 

closes: https://github.com/paradigmxyz/reth/issues/18801

### impact: 
| Change | Impact |
|--------|--------|
| Remove `run()` loop thread | **-1 thread, -1 channel hop** |
| Direct channel sends | **~some time saved per task** |
| Eliminate enum wrapping | **~2 allocations saved per task** |

### reference PRs:
- https://github.com/paradigmxyz/reth/pull/18887
- https://github.com/paradigmxyz/reth/pull/18901


crates/trie/parallel/src/proof_task.rs
```diff
@@ -1274,35 +1063,38 @@ impl TrieNodeProviderFactory for ProofTaskManagerHandle {
...
-                let _ = sender.send(ProofTaskMessage::QueueTask(
-                    ProofTaskKind::BlindedStorageNode(*account, *path, tx),
-                ));
+            Self::StorageNode { handle, account } => {
+                let rx = handle.queue_blinded_storage_node(*account, *path).map_err(|e| {
+                    SparseTrieErrorKind::Other(Box::new(std::io::Error::other(e.to_string())))
+                })?;
+                rx.recv().map_err(|_| {
+                    SparseTrieErrorKind::Other(Box::new(std::io::Error::other("channel closed")))
+                })?
```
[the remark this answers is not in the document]
yongkangc: addressed

crates/engine/primitives/src/config.rs
```diff
@@ -494,7 +499,11 @@ impl TreeConfig {
 
     /// Setter for the number of storage proof worker threads.
     pub const fn with_storage_worker_count(mut self, storage_worker_count: usize) -> Self {
-        self.storage_worker_count = storage_worker_count;
+        self.storage_worker_count = if storage_worker_count >= MIN_WORKER_COUNT {
+            storage_worker_count
+        } else {
+            MIN_WORKER_COUNT
+        };
```
yongkangc: addresed

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -346,10 +346,9 @@ pub struct MultiproofManager {
     pending: VecDeque<PendingMultiproofTask>,
     /// Executor for tasks
     executor: WorkloadExecutor,
-    /// Handle to the proof task manager used for creating `ParallelProof` instances for storage
-    /// proofs.
+    /// Handle to the proof worker pool for storage proofs.
     storage_proof_task_handle: ProofTaskManagerHandle,
-    /// Handle to the proof task manager used for account multiproofs.
+    /// Handle to the proof worker pool for account multiproofs.
     account_proof_task_handle: ProofTaskManagerHandle,
```
shekhirin: any reason why do we need to separate them here, if it's always the same `proof_task_handle` passed to both fields?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1274,35 +1066,43 @@ impl TrieNodeProviderFactory for ProofTaskManagerHandle {
...
+        handle: ProofTaskManagerHandle,
     },
 }
 
 impl TrieNodeProvider for ProofTaskTrieNodeProvider {
     fn trie_node(&self, path: &Nibbles) -> Result<Option<RevealedNode>, SparseTrieError> {
-        let (tx, rx) = channel();
+        /// Helper to convert `ProviderError` to `SparseTrieError`
+        fn provider_err_to_trie_err(e: ProviderError) -> SparseTrieError {
+            SparseTrieErrorKind::Other(Box::new(std::io::Error::other(e.to_string()))).into()
```
shekhirin: can't we do `SparseTrieErrorKind::Other(Box::new(e))` here?

crates/trie/parallel/src/proof_task_metrics.rs
```diff
@@ -1,21 +1,41 @@
...
+impl Default for ProofTaskMetrics {
+    fn default() -> Self {
+        Self {
+            task_metrics: ProofTaskTrieMetrics::default(),
+            storage_proofs: Arc::new(AtomicU64::new(0)),
+            account_proofs: Arc::new(AtomicU64::new(0)),
+            account_nodes: Arc::new(AtomicU64::new(0)),
+            storage_nodes: Arc::new(AtomicU64::new(0)),
+        }
+    }
```
shekhirin: I think it can be derived, AFAIK `Arc<AtomicU64>` implements `Default`

shekhirin (commented):
LGTM overall, just have some nits regarding the usage of `std::io::Error` and separate account/storage proof handles.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: thanks for the note

crates/trie/parallel/src/proof_task_metrics.rs
yongkangc: ah yes! thanks for this, much cleaner

crates/trie/parallel/src/proof_task.rs
yongkangc: yup, good note - addressed that as well

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: thanks for the note, addresssed them

crates/trie/parallel/src/proof_task.rs
yongkangc: addressed this in commit

crates/trie/parallel/src/proof_task_metrics.rs
yongkangc: thanks for the catch, that is much better. addressed.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1274,35 +1066,43 @@ impl TrieNodeProviderFactory for ProofTaskManagerHandle {
...
-        let (tx, rx) = channel();
+        /// Helper to convert `ProviderError` to `SparseTrieError`
+        fn provider_err_to_trie_err(e: ProviderError) -> SparseTrieError {
+            SparseTrieErrorKind::Other(Box::new(e)).into()
+        }
+
+        /// Helper to convert channel recv error to `SparseTrieError`
+        fn recv_err_to_trie_err(_: std::sync::mpsc::RecvError) -> SparseTrieError {
+            SparseTrieErrorKind::Other(Box::new(std::io::Error::other("channel closed"))).into()
+        }
```
shekhirin: can we just do this without helper functions? https://github.com/paradigmxyz/reth/blob/f5840fc562b9f363e404dee0b9e25235755dffa4/crates/trie/trie/src/proof/trie_node.rs#L96

`RecvError` already implements `Error`, so should work?

crates/engine/primitives/src/config.rs
```diff
@@ -9,11 +9,27 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
 /// Default maximum concurrency for on-demand proof tasks (blinded nodes)
 pub const DEFAULT_MAX_PROOF_TASK_CONCURRENCY: u64 = 256;
 
+/// Minimum number of workers we allow configuring explicitly.
+pub const MIN_WORKER_COUNT: usize = 2;
+
+/// Clamps the worker count to the minimum allowed value.
+///
+/// Ensures that the worker count is at least [`MIN_WORKER_COUNT`].
+const fn clamp_worker_count(count: usize) -> usize {
```
shekhirin: is this just `.max(MIN_WORKER_COUNT)`? Let's move this to `with_*_worker_count` functions, no need to have a separate helper fn for this

crates/trie/parallel/src/proof_task.rs
yongkangc: yup we could, initially wanted helper because it looked ugly with all the map_err

yongkangc: yup, initially wanted helper because it looked ugly with all the map_err

done in [2698cc8](https://github.com/paradigmxyz/reth/pull/18934/commits/2698cc80072179a081d5fea56b17821d94199e83)

crates/engine/primitives/src/config.rs
yongkangc: much more concise, done in [c68eb7a](https://github.com/paradigmxyz/reth/pull/18934/commits/c68eb7a9fd93022bf3c1b59811525b8177863336)

crates/trie/parallel/src/proof_task_metrics.rs
```diff
@@ -1,21 +1,29 @@
...
 #[derive(Clone, Debug, Default)]
 pub struct ProofTaskMetrics {
     /// The actual metrics for blinded nodes.
     pub task_metrics: ProofTaskTrieMetrics,
-    /// Count of blinded account node requests.
-    pub account_nodes: usize,
-    /// Count of blinded storage node requests.
-    pub storage_nodes: usize,
+    /// Count of storage proof requests (lock-free).
+    pub storage_proofs: Arc<AtomicU64>,
```
mediocregopher: The purpose of ProofTaskMetrics is to allow us to lazily record metrics at the end of block processing, so that during block processing we can just increment raw ints and not incur synchronization overhead. By using atomics here we are negating that.

If the issue is in sharing the ProofTaskMetrics across workers then we could `clone` the ProofTaskMetrics into each worker on startup, and have each worker `record` prior to termination.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -88,58 +90,80 @@ enum StorageWorkerJob {
...
-    /// Number of storage workers successfully spawned.
-    ///
-    /// May be less than requested if concurrency limits reduce the worker budget.
+/// # Parameters
+/// - `executor`: Tokio runtime handle for spawning blocking tasks
+/// - `view`: Consistent database view for creating transactions
+/// - `task_ctx`: Shared context with trie updates and prefix sets
+/// - `storage_worker_count`: Number of storage workers to spawn
+/// - `account_worker_count`: Number of account workers to spawn
+pub fn spawn_proof_workers<Factory>(
```
mediocregopher: nit: this would make more sense as `ProofTaskManagerHandler::new` I think.

mediocregopher (changes requested):

crates/trie/parallel/src/proof_task.rs
yongkangc: thats much nicer api, addressed [f9e167e](https://github.com/paradigmxyz/reth/pull/18934/commits/f9e167ec393e1edd1b83985587955cb1f6097fa8)

crates/trie/parallel/src/proof_task_metrics.rs
yongkangc: good point on the synchronization overhead, addressed here [bc4ecf5](https://github.com/paradigmxyz/reth/pull/18934/commits/bc4ecf5077dc8ae43f2fd47a16271f582c28e30b)

yongkangc (commented):
@mediocregopher just addressed your comments

mediocregopher (approved):

crates/trie/parallel/src/proof.rs
```diff
@@ -102,9 +100,9 @@ impl ParallelProof {
             self.multi_added_removed_keys.clone(),
         );
 
-        let (sender, receiver) = std::sync::mpsc::channel();
-        let _ = self.proof_task_handle.queue_task(ProofTaskKind::StorageProof(input, sender));
-        receiver
+        self.proof_task_handle
+            .queue_storage_proof(input)
```
mattsse, line 104: can we rename these after we merge, because I find `queue` very confusing here because this only sends

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1181,78 +833,179 @@ impl ProofTaskCtx {
...
+        tracing::info!(
+            target: "trie::proof_task",
+            storage_worker_count,
+            account_worker_count,
+            "Spawning proof worker pools"
+        );
+
+        // Spawn storage workers
+        for worker_id in 0..storage_worker_count {
+            let provider_ro = view.provider_ro()?;
```
mattsse, line 883: I believe this is still something we pay for upfront, meaning this isn't done in the background

this does mean it currently takes more time to set this up if we bump the worker count?

I think ideally we return the channels right away and do this setup in the background so that we don't block here:

https://github.com/paradigmxyz/reth/blob/1619408dfccb8ae9b2b0dd7f5eea666bfd2e242d/crates/engine/tree/src/tree/payload_processor/mod.rs#L206-L206

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1181,78 +833,179 @@ impl ProofTaskCtx {
...
+            tracing::debug!(
+                target: "trie::proof_task",
+                worker_id,
+                "Storage worker spawned successfully"
+            );
+        }
+
+        // Spawn account workers
+        for worker_id in 0..account_worker_count {
+            let provider_ro = view.provider_ro()?;
```
mattsse, line 910: same here

mattsse (approved):
lgtm

left some suggestions, iirc additional changes to the worker pool are planned anyway and we could tackle those separately

crates/trie/parallel/src/proof.rs
yongkangc: thats true -> should rename it as `send_task`

crates/trie/parallel/src/proof_task.rs
yongkangc: this makes sense, good idea for us to do in bg

crates/engine/primitives/src/config.rs
```diff
@@ -9,11 +9,16 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
 /// Default maximum concurrency for on-demand proof tasks (blinded nodes)
 pub const DEFAULT_MAX_PROOF_TASK_CONCURRENCY: u64 = 256;
 
+/// Minimum number of workers we allow configuring explicitly.
+pub const MIN_WORKER_COUNT: usize = 2;
```
yongkangc:
```suggestion
pub const MIN_WORKER_COUNT: usize = 32;
```

crates/engine/primitives/src/config.rs
```diff
@@ -9,11 +9,16 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
...
+
 /// Returns the default number of storage worker threads based on available parallelism.
 fn default_storage_worker_count() -> usize {
     #[cfg(feature = "std")]
     {
-        std::thread::available_parallelism().map(|n| (n.get() * 2).clamp(2, 64)).unwrap_or(8)
+        std::thread::available_parallelism()
+            .map(|n| (n.get() * 2).clamp(MIN_WORKER_COUNT, 64))
```
yongkangc:
```suggestion
            .map(|n| (n.get() * 2).clamp(MIN_WORKER_COUNT, 128))
```

Rjected (approved):

crates/engine/primitives/src/config.rs
```diff
@@ -9,11 +9,16 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
...
+pub const MIN_WORKER_COUNT: usize = 32;
+
 /// Returns the default number of storage worker threads based on available parallelism.
 fn default_storage_worker_count() -> usize {
     #[cfg(feature = "std")]
     {
-        std::thread::available_parallelism().map(|n| (n.get() * 2).clamp(2, 64)).unwrap_or(8)
+        std::thread::available_parallelism()
+            .map(|n| (n.get() * 2).clamp(MIN_WORKER_COUNT, 128))
+            .unwrap_or(8)
```
shekhirin: don't think there's a need to restrict the upper bound, if you have a threadripper with 192 threads, it should be fine to spawn more than 128 threads.


```suggestion
            .map_or(8, |n| n.get() * 2).min(MIN_WORKER_COUNT)
```

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1181,78 +833,179 @@ impl ProofTaskCtx {
...
-}
-
-/// A handle that wraps a single proof task sender that sends a terminate message on `Drop` if the
-/// number of active handles went to zero.
-#[derive(Debug)]
+/// The handle stores direct senders to both storage and account worker pools,
+/// eliminating the need for a routing thread. All handles share reference-counted
+/// channels, and workers shut down gracefully when all handles are dropped.
+#[derive(Debug, Clone)]
 pub struct ProofTaskManagerHandle {
```
shekhirin: I find this a tiny bit confusing that we have a `ProofTaskManagerHandle`, but no `ProofTaskManager`, as there's usually a pattern of having a `fn handle(&self) -> Handle` method. Not necessary to address in this PR, can be a followup.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1181,78 +833,179 @@ impl ProofTaskCtx {
...
+        storage_worker_count: usize,
+        account_worker_count: usize,
+    ) -> ProviderResult<Self>
+    where
+        Factory: DatabaseProviderFactory<Provider: BlockReader>,
+    {
+        let (storage_work_tx, storage_work_rx) = unbounded::<StorageWorkerJob>();
+        let (account_work_tx, account_work_rx) = unbounded::<AccountWorkerJob>();
+
+        tracing::info!(
```
shekhirin: I think INFO is too verbose for this type of log, let's downgrade to DEBUG.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1181,78 +833,179 @@ impl ProofTaskCtx {
...
+    /// Queue an account multiproof computation
+    pub fn queue_account_multiproof(
+        &self,
+        input: AccountMultiproofInput,
+    ) -> Result<Receiver<AccountMultiproofResult>, ProviderError> {
+        let (tx, rx) = channel();
+        self.account_work_tx
+            .send(AccountWorkerJob::AccountMultiproof { input, result_sender: tx })
+            .map_err(|_| {
+                ProviderError::other(std::io::Error::other("account workers unavailable"))
```
shekhirin, line 974: would also like to address this in a followup cleanup, this feels like misuse of `std::io::Error`

shekhirin (approved):

crates/trie/parallel/src/proof_task.rs
yongkangc:
```suggestion
        tracing::debug!(
```

yongkangc: addressed

PR #18938, opened 2025-10-10
fix(storage/libmdbx-rs): avoid panic in read tx monitor by using saturating_sub

sashaodessa:
### Description
- **What**: Replace potentially panicking `Duration` subtraction with `saturating_sub` when computing `sleep_duration`.
- **Why**: `Duration - Duration` panics on negative results; long‑lived read txs could crash the monitor thread.
- **How**: In `crates/storage/libmdbx-rs/src/txn_manager.rs` (`read_transactions::start_monitor`), compute:
  `READ_TRANSACTIONS_CHECK_INTERVAL.min(self.max_duration.saturating_sub(max_active_transaction_duration.unwrap_or_default()))`.

### DIFF
--- a/crates/storage/libmdbx-rs/src/txn_manager.rs
+++ b/crates/storage/libmdbx-rs/src/txn_manager.rs
@@ -302,9 +302,8 @@ mod read_transactions {
                     // the closest deadline of an active read transaction
                     // Use saturating_sub to avoid panic when active duration exceeds max_duration
                     let sleep_duration = READ_TRANSACTIONS_CHECK_INTERVAL.min(
-                        self.max_duration.saturating_sub(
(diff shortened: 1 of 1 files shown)


crates/storage/libmdbx-rs/src/txn_manager.rs
```diff
@@ -300,8 +300,10 @@ mod read_transactions {
 
                     // Sleep not more than `READ_TRANSACTIONS_CHECK_INTERVAL`, but at least until
                     // the closest deadline of an active read transaction
+                    // Use saturating_sub to avoid panic when active duration exceeds max_duration
```
mattsse, line 303: you guys need to stop writing comments like this

this has 0 benefit without the pr as context

mattsse (commented):
not useful

PR #18941, opened 2025-10-10
fix(trie): Properly upsert into StoragesTrie in repair-trie

mediocregopher:
Closes #18768

### DIFF
--- a/crates/cli/commands/src/db/repair_trie.rs
+++ b/crates/cli/commands/src/db/repair_trie.rs
@@ -179,8 +179,17 @@ fn verify_and_repair<N: ProviderNodeTypes>(
             Output::StorageWrong { account, path, expected: node, .. } |
             Output::StorageMissing(account, path, node) => {
                 // Wrong/missing storage node value, upsert it
+                // (We can't just use `upsert` method with a dup cursor, it's not properly
+                // supported)
                 let nibbles = StoredNibblesSubKey(path);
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/db/repair_trie.rs
```diff
@@ -179,8 +179,17 @@ fn verify_and_repair<N: ProviderNodeTypes>(
...
+                // (We can't just use `upsert` method with a dup cursor, it's not properly
+                // supported)
                 let nibbles = StoredNibblesSubKey(path);
-                let entry = StorageTrieEntry { nibbles, node };
+                let entry = StorageTrieEntry { nibbles: nibbles.clone(), node };
+                if storage_trie_cursor
+                    .seek_by_key_subkey(account, nibbles.clone())?
+                    .filter(|v| v.nibbles == nibbles)
+                    .is_some()
+                {
```
yongkangc, line 190: i see, so if an entry exists, we delete the current entry first before doing the upsert, so theres no duplicate state?

yongkangc (approved):

crates/cli/commands/src/db/repair_trie.rs
mediocregopher: Correct. We don't have to do this on "normal" tables, but on DUPSORT tables the `upsert` method doesn't work as expected, so we have to do it manually.

PR #18945, opened 2025-10-11
feat: allow using SafeNoSync for MDBX

eabz:
This PR introduces a flag to enable to users to initialize the database with `SafeNoSync` mode.

Fixes: #18565 

Introduces:
- Added a new element `sync_mode` to the `DatabaseArguments` and `DatabaseArgs`.
- Added a `TypedParser` to make sure only `Durable` and `SafeNoSync` options are used.
- Passes the variable to `DatabaseArguments::new` function.
- Changed all the tests where `DatabaseArguments::new` is used to use the `SafeNoSync` mode.
- Hardcodes the usage in stage/dump to `Durable`.
- Argument defaults to `Durable`

Can you point me if the documentation update is necessary?

### DIFF
--- a/crates/cli/commands/src/stage/dump/mod.rs
+++ b/crates/cli/commands/src/stage/dump/mod.rs
@@ -3,11 +3,7 @@ use crate::common::{AccessRights, CliNodeComponents, CliNodeTypes, Environment, 
 use clap::Parser;
 use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_cli::chainspec::ChainSpecParser;
-use reth_db::{
-    init_db,
-    mdbx::{DatabaseArguments, SyncMode},
-    DatabaseEnv,
-};
+use reth_db::{init_db, mdbx::DatabaseArguments, DatabaseEnv};
 use reth_db_api::{
     cursor::DbCursorRO, database::Database, models::ClientVersion, table::TableImporter, tables,
     transaction::DbTx,
@@ -139,8 +135,7 @@ pub(crate) fn setup<N: NodeTypesWithDB>(
 
     info!(target: "reth::cli", ?output_db, "Creating separate db");
 
-    let output_datadir =
-        init_db(output_db, DatabaseArguments::new(ClientVersion::default(), SyncMode::Durable))?;
+    let output_datadir = init_db(output_db, DatabaseArguments::new(ClientVersion::default()))?;
 
     output_datadir.update(|tx| {
         tx.import_table_with_range::<tables::BlockBodyIndices, _>(
--- a/crates/cli/commands/src/stage/dump/mod.rs
+++ b/crates/cli/commands/src/stage/dump/mod.rs
@@ -3,7 +3,11 @@ use crate::common::{AccessRights, CliNodeComponents, CliNodeTypes, Environment, 
 use clap::Parser;
 use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_cli::chainspec::ChainSpecParser;
-use reth_db::{init_db, mdbx::DatabaseArguments, DatabaseEnv};
+use reth_db::{
+    init_db,
+    mdbx::{DatabaseArguments, SyncMode},
+    DatabaseEnv,
+};
 use reth_db_api::{
     cursor::DbCursorRO, database::Database, models::ClientVersion, table::TableImporter, tables,
     transaction::DbTx,
@@ -135,7 +139,8 @@ pub(crate) fn setup<N: NodeTypesWithDB>(
 
     info!(target: "reth::cli", ?output_db, "Creating separate db");
 
-    let output_datadir = init_db(output_db, DatabaseArguments::new(ClientVersion::default()))?;
+    let output_datadir =
+        init_db(output_db, DatabaseArguments::new(ClientVersion::default(), SyncMode::Durable))?;
 
     output_datadir.update(|tx| {
         tx.import_table_with_range::<tables::BlockBodyIndices, _>(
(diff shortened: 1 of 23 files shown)


fynnss (approved):

crates/node/core/src/args/database.rs
```diff
@@ -34,6 +37,13 @@ pub struct DatabaseArgs {
     /// Maximum number of readers allowed to access the database concurrently.
     #[arg(long = "db.max-readers")]
     pub max_readers: Option<u64>,
+    /// Controls how aggressively the database synchronizes data to disk.
+    #[arg(
+        long = "db.sync-mode",
+        value_parser = SyncModeValueParser::default(),
+        default_value = "durable"
+    )]
+    pub sync_mode: SyncMode,
```
mattsse: can we make this optional

crates/cli/commands/src/stage/dump/mod.rs
```diff
@@ -135,7 +139,8 @@ pub(crate) fn setup<N: NodeTypesWithDB>(
 
     info!(target: "reth::cli", ?output_db, "Creating separate db");
 
-    let output_datadir = init_db(output_db, DatabaseArguments::new(ClientVersion::default()))?;
+    let output_datadir =
+        init_db(output_db, DatabaseArguments::new(ClientVersion::default(), SyncMode::Durable))?;
```
mattsse: I'd like to avoid changing the new function for this

crates/storage/db/src/implementation/mdbx/mod.rs
```diff
@@ -329,7 +344,7 @@ impl DatabaseEnv {
             DatabaseEnvKind::RW => {
                 // enable writemap mode in RW mode
                 inner_env.write_map();
-                Mode::ReadWrite { sync_mode: SyncMode::Durable }
+                Mode::ReadWrite { sync_mode: args.sync_mode }
```
mattsse, line 358: this makes sense

crates/storage/db/src/implementation/mdbx/mod.rs
```diff
@@ -103,17 +103,31 @@ pub struct DatabaseArguments {
...
     fn default() -> Self {
-        Self::new(ClientVersion::default())
+        Self::new(ClientVersion::default(), SyncMode::Durable)
     }
 }
 
 impl DatabaseArguments {
     /// Create new database arguments with given client version.
-    pub fn new(client_version: ClientVersion) -> Self {
+    pub fn new(client_version: ClientVersion, sync_mode: SyncMode) -> Self {
```
mattsse: instead of adding this to new, can we init this always with durable

and add another `fn with_sync_mode` that sets this

crates/storage/db/src/implementation/mdbx/mod.rs
```diff
@@ -444,7 +459,7 @@ impl DatabaseEnv {
                     LogLevel::Extra => 7,
                 });
             } else {
-                return Err(DatabaseError::LogLevelUnavailable(log_level))
+                return Err(DatabaseError::LogLevelUnavailable(log_level));
```
mattsse: please undo additional ;

crates/storage/db/src/lib.rs
mattsse, line 1: why are those using different syncmodes?

mattsse (changes requested):
supportive but I'd like to change the setup for this a bit so that we don't always need to provide the syncmode

unclear to me why some changes are now using SafeNoSync

this change should not alter existing logic

eabz:
Hey @mattsse, thanks for the feedback!

I’m really enthusiastic about making my first contribution to the reth repository.
I reviewed your suggestions and reverted many of my changes to the `DatabaseArguments::new` function. I also added a new function, `with_sync_mode`, to ensure that the flag is only passed to the database when explicitly used by the user.

Additionally, I removed all the modifications I had made to the test functions.

I believe this approach aligns much better with the current codebase. I’m still getting familiar with how things work in `reth`, but I’m excited to keep learning and improving.

Let me know if there is something else I need to change.

crates/node/core/src/args/database.rs
```diff
@@ -175,6 +185,64 @@ fn parse_byte_size(s: &str) -> Result<usize, String> {
...
+        arg: Option<&Arg>,
+        value: &std::ffi::OsStr,
+    ) -> Result<Self::Value, Error> {
+        let raw =
+            value.to_str().ok_or_else(|| Error::raw(ErrorKind::InvalidUtf8, "Invalid UTF-8"))?;
+        let val = raw.trim().to_ascii_lowercase();
+
+        let parsed = match val.as_str() {
+            "durable" => SyncMode::Durable,
+            "safe-no-sync" | "safenosync" | "safe_no_sync" => SyncMode::SafeNoSync,
```
mattsse: can we make this a fromstr impl on SyncMode?

mattsse (dismissed):
last suggestion

other than that lgtm

pending @shekhirin

shekhirin (approved):
Nice, this is useful. LGTM!

crates/node/core/src/args/database.rs
```diff
@@ -34,6 +37,12 @@ pub struct DatabaseArgs {
     /// Maximum number of readers allowed to access the database concurrently.
     #[arg(long = "db.max-readers")]
     pub max_readers: Option<u64>,
+    /// Controls how aggressively the database synchronizes data to disk.
+    #[arg(
+        long = "db.sync-mode",
+        value_parser = value_parser!(SyncMode),
```
mattsse, line 43: unsure if we even need this now that syncmode is fromstr

mattsse (approved):

PR #18946, opened 2025-10-11
fix(examples): change method to launch with debug capabilities

TilakMaddy:
Without launching with debug capabilities, the example would be stuck waiting for the tx hash notification

```rust
    let head = notifications.next().await.unwrap();

    let tx = &head.tip().body().transactions().next().unwrap();
    assert_eq!(*tx.tx_hash(), hash);
    println!("mined transaction: {hash}");
```


mattsse (approved):
makes sense for the dev node example

PR #18947, opened 2025-10-11
fix(example): launch with debug capabilities

TilakMaddy:
Without auto mining, cast command to create tx will just be stuck forever.
So it's better to launch_with_debug_capabilities

```
cast send 0x0000000000000000000000000000000000000000 \
  --value 0.1ether \
  --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
  --priority-gas-price 2gwei \
  --gas-price 20gwei
  --nonce 5
```


mattsse (approved):
makes sense

PR #18952, opened 2025-10-13
perf: eliminate needless Vec allocation in tx_lookup stage unwind

sky-coderay:
Remove unnecessary `collect()` call in TransactionLookupStage::unwind that was creating a temporary Vec only to immediately iterate over it in reverse.

This optimization reduces memory complexity from O(n) to O(1) for the unwind operation by replacing the `range.collect::<Vec<_>>().rev()` pattern with direct arithmetic-based reverse iteration using `range_end - i as u64`.

Benchmark results show significant memory savings:
- Before: O(n) heap allocation for temporary Vec
- After: O(1) memory usage with stack-only operations
- Memory allocation eliminated: ~79ns saved per unwind operation
- Trade-off: ~8ns additional computation time (87ns vs 79ns)


mattsse (commented):
this isn't equivalent because this inst the same as enumarte

PR #18955, opened 2025-10-13
refactor: unify `Pipeline` creation codepaths

klkvr:
Changes node launching logic to reuse the main `Pipeline` instance for all of the consistency checks.

With this change, we now only instantiate the pipeline once allowing to simplify logic for custom stages


mattsse (approved):
makes sense

PR #18960, opened 2025-10-13
feat(engine): improve payload validator tracing spans

shekhirin:
Towards #18892 

This PR has two main changes:
- All `instrument` macro calls now have a `debug` level specified. This is done to prevent losing context for `DEBUG` level events  — previously if the `instrument` macro created a span with level `TRACE`, then `DEBUG` events won't have this span as a parent, unless the env filter is specified to include these `TRACE` spans. Just creating a span doesn't emit an event itself, so it doesn't make the stdout noisier.
- More spans with `engine::tree*` target is created, giving us a better overview of payload validation pipeline image


mediocregopher (approved):

mediocregopher (commented):
Does setting the level to trace mean that these fields won't get included in e.g. WARN logs? If so maybe it would be useful to also have these applied to all logs? Having more context in each individual log makes them easier to use imo

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -366,7 +374,9 @@ where
         let parent_hash = input.parent_hash();
         let block_num_hash = input.num_hash();
 
-        trace!(target: "engine::tree", block=?block_num_hash, parent=?parent_hash, "Fetching block state provider");
+        trace!(target: "engine::tree::payload_validator", "Fetching block state provider");
```
yongkangc, line 377: do we not want the block num hash here?

yongkangc (approved):
overall looks good to me 👍🏻 


one comment: 

- i think span would help within `build_account_multiproof`

crates/engine/tree/src/tree/payload_validator.rs
Rjected: That info will be added by the `instrument` call at the top of the fn

PR #18961, opened 2025-10-13
fix: SIGSEGV handler; use sa_handler to match handler signature

maradini77:
### Summary
- Switch `sigaction` configuration to `sa_handler` to match the handler’s `extern "C" fn(c_int)` signature.

### Rationale
- Avoids ambiguity between `sa_sigaction`/`SA_SIGINFO` and `sa_handler`.
- Improves portability and correctness across platforms.


mattsse (commented):
this doesnt fix anything because only changes docs

PR #18962, opened 2025-10-13
fix: use max B256 for upper bound in empty-storage check

maradini77:
### Summary
- Replace reversed zero with the real maximum B256 (all 0xFF) when sweeping trailing accounts for empty storage.

### Rationale
- Reversing `B256::ZERO` yields zero, not the maximum, truncating the verification range and potentially missing non-empty storages after the last processed account.


mediocregopher (approved):
Good catch, thanks!

PR #18963, opened 2025-10-13
fix: replace println! with tracing::debug! in print_stack

maradini77:
### Summary
- Switch `print_stack` from stdout prints to structured logging via `tracing::debug!`.


mattsse (commented):
we want print here

PR #18964, opened 2025-10-13
feat(cli): Reuse a single StaticFileProducer across file import chunks

GarmashAlex:
Previously, each file-import chunk created a new StaticFileProducer and passed it into a freshly built pipeline. StaticFileProducer is a lightweight Arc<Mutex<...>> wrapper with no chunk-specific state and is safe to clone and reuse. Creating it per chunk caused unnecessary allocations and event-channel setups without any functional benefit.
This change creates one StaticFileProducer before the import loop and passes static_file_producer.clone() into each pipeline build. It aligns the import path with the node’s engine lifecycle (where a single producer is reused), reduces overhead, and preserves behavior.

### DIFF
--- a/crates/optimism/cli/src/commands/import.rs
+++ b/crates/optimism/cli/src/commands/import.rs
@@ -71,6 +71,10 @@ impl<C: ChainSpecParser<ChainSpec = OpChainSpec>> ImportOpCommand<C> {
             .sealed_header(provider_factory.last_block_number()?)?
             .expect("should have genesis");
 
+        // Reuse a single StaticFileProducer across all chunks
+        let static_file_producer =
+            StaticFileProducer::new(provider_factory.clone(), PruneModes::default());
+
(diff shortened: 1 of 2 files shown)


crates/optimism/cli/src/commands/import.rs
```diff
@@ -71,6 +71,10 @@ impl<C: ChainSpecParser<ChainSpec = OpChainSpec>> ImportOpCommand<C> {
             .sealed_header(provider_factory.last_block_number()?)?
             .expect("should have genesis");
 
+        // Reuse a single StaticFileProducer across all chunks
```
mattsse: please dont add comments like pr descriptions

these are totally useless without the pr context

mattsse (changes requested):
doc nits

joshieDo (approved):

mattsse (approved):

PR #18967, opened 2025-10-13
refactor: replace magic numbers with named constants in execute tests

0xtutankhamun:
Replace hardcoded numeric literals with descriptive constants in EIP test cases. 

Constants include history buffer length, test beacon root byte, withdrawal request input length, and Gwei conversion value.


mattsse (commented):
these are just test so magic nums are fine here

PR #18968, opened 2025-10-13
fix: overflow in BlockNumber range conversion at u64::MAX

maradini77:
### Description
- **Summary**: Prevent overflow when converting an inclusive end bound into an exclusive one for BlockNumber ranges by treating u64::MAX as unbounded.
- **Change**: In `crates/storage/db-api/src/models/accounts.rs`, `Bound::Included(n)` now checks `n == u64::MAX` and returns `Bound::Unbounded`; otherwise uses `n + 1` as before. 
- **Why**: Avoids debug panics and release wrap-around; ensures correct and predictable range semantics at the upper boundary.


mattsse (commented):
block number can't reach u64::max so this isnt necessary

PR #18970, opened 2025-10-13
fix: overflow in inclusive range length calculation for static-file segments

maradini77:
### Description
- **Summary**: Use saturating arithmetic to compute inclusive range lengths without overflowing at u64::MAX.
- **Change**: In `crates/static-file/types/src/segment.rs`, replaced `(end + 1) - start` with `end.saturating_sub(start).saturating_add(1)` for both `tx_len` and `block_len`. 
- **Why**: Prevents debug panics and release wrap-around; ensures correct lengths at upper boundary.


mattsse (commented):
not needed because end > start

PR #18973, opened 2025-10-13
fix(trie): Fixes to pruning of MerkleChangeSets

mediocregopher:
Closes #18950 

This PR contains a few related fixes:

* Adds MerkleChangeSets pruning segment to the default set (it wasn't previously present)

* Prune the StoragesTrieChangeSets table, not StoragesChangeSets

* Allows for configuration of the MerkleChangeSets PruneMode. The PruneMode _must_ be set for this segment, as we never want to keep the full history of trie changesets. The default is `Distance(10064)`. If someone really wanted to they could manually set it to `Before(0)` to keep all history.

* Because PruneModes now has a field which can't be None it doesn't make sense to have `none` or `is_empty` methods. These have been removed.

### DIFF
--- a/crates/node/core/src/args/pruning.rs
+++ b/crates/node/core/src/args/pruning.rs
@@ -126,6 +126,7 @@ impl PruningArgs {
                     storage_history: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
                     // TODO: set default to pre-merge block if available
                     bodies_history: None,
+                    merkle_changesets: PruneMode::Full,
                     receipts_log_filter: Default::default(),
                 },
             }
--- a/crates/node/core/src/args/pruning.rs
+++ b/crates/node/core/src/args/pruning.rs
@@ -126,7 +126,7 @@ impl PruningArgs {
                     storage_history: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
                     // TODO: set default to pre-merge block if available
                     bodies_history: None,
-                    merkle_changesets: PruneMode::Full,
+                    merkle_changesets: PruneMode::Distance(MINIMUM_PRUNING_DISTANCE),
                     receipts_log_filter: Default::default(),
                 },
             }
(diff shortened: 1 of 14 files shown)


RomanHodulak (approved):

crates/node/core/src/args/pruning.rs
```diff
@@ -126,6 +126,7 @@ impl PruningArgs {
                     storage_history: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
                     // TODO: set default to pre-merge block if available
                     bodies_history: None,
+                    merkle_changesets: PruneMode::Distance(MINIMUM_PRUNING_DISTANCE),
```
shekhirin, line 129: given your tests over the weekend, this is about 20-25GB on Mainnet, right?

shekhirin (approved):
LGTM, did you test with an existing config.toml to ensure that nodes won't break when upgrading to a version with these prune config changes?

crates/node/core/src/args/pruning.rs
mediocregopher: Having applied this change on reth-3 already, on mainnet it comes out to about 1.5 GiB total. Which makes me think that the 25GiB must have covered more data than just the weekend, there might have been leftover data from when I was testing the pipeline sync. But anyway it's definitely a manageable amount of data now.

crates/prune/types/src/target.rs
```diff
@@ -36,8 +36,13 @@ pub enum HistoryType {
     StorageHistory,
 }
 
+/// Default pruning mode for merkle changesets - aggressively prune to finalized block
```
mediocregopher:
```suggestion
/// Default pruning mode for merkle changesets
```

crates/prune/types/src/segment.rs
```diff
@@ -3,6 +3,9 @@ use derive_more::Display;
 use thiserror::Error;
 
 /// Segment of the data that can be pruned.
+///
+/// NOTE new variants must be added to the end of this enum. The variant index is encoded directly
+/// when writing to the P`runeCheckpoint` table, so changing the order here will corrupt the table.
```
Rjected:
```suggestion
/// when writing to the `PruneCheckpoint` table, so changing the order here will corrupt the table.
```

Rjected (approved):

PR #18977, opened 2025-10-13
feat: support custom stages

klkvr:
Based on #18955 

Allows defining custom stages on `NodeAddOns`. Those stages are added to pipeline and advanced in `PersistenceService`.

Open questions:
- Is there a better place for the `custom_stages` getter? Maybe `Node` trait?
- How appropriate is it to advance stages directly in the `PersistenceService`? This should work fine for most of the cases, especially for custom indexes. Another approach would be allowing to define custom hooks invoked when persisting but it feels like this might just lead to redundancy in custom stage impls

### DIFF
--- a/crates/engine/tree/src/persistence.rs
+++ b/crates/engine/tree/src/persistence.rs
@@ -11,8 +11,10 @@ use reth_provider::{
     StageCheckpointWriter,
 };
 use reth_prune::{PrunerError, PrunerOutput, PrunerWithFactory};
-use reth_stages::{BoxedStage, ExecInput, ExecOutput, StageError, UnwindInput, UnwindOutput};
-use reth_stages_api::{MetricEvent, MetricEventsSender};
+use reth_stages_api::{
+    BoxedStage, ExecInput, ExecOutput, MetricEvent, MetricEventsSender, StageError, UnwindInput,
+    UnwindOutput,
+};
 use std::{
     sync::mpsc::{Receiver, SendError, Sender},
     time::Instant,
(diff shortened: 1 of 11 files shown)


crates/node/api/src/node.rs
```diff
@@ -203,6 +204,13 @@ pub trait NodeAddOns<N: FullNodeComponents>: Send {
         self,
         ctx: AddOnsContext<'_, N>,
     ) -> impl Future<Output = eyre::Result<Self::Handle>> + Send;
+
+    /// Returns additional stages to be added to the pipeline.
+    fn extra_stages(
+        &self,
+    ) -> Vec<BoxedStage<<N::Provider as DatabaseProviderFactory>::ProviderRW>> {
+        Vec::new()
+    }
```
mattsse, line 213: this makes sense here I believe

crates/node/builder/src/setup.rs
```diff
@@ -129,7 +132,8 @@ where
                 stage_config.execution.into(),
                 stage_config.execution_external_clean_threshold(),
                 exex_manager_handle,
-            )),
+            ))
+            .add_stages_before(extra_stages, StageId::Finish),
```
mattsse, line 136: I think we're fine if these are put before finish stage

crates/engine/tree/src/persistence.rs
```diff
@@ -149,10 +169,28 @@ where
             number: block.recovered_block().header().number(),
         });
 
-        if last_block_hash_num.is_some() {
+        if let Some(num_hash) = last_block_hash_num {
             let provider_rw = self.provider.database_provider_rw()?;
 
             provider_rw.save_blocks(blocks)?;
+
+            for stage in &mut self.custom_stages {
```
mattsse, line 177: can we document here that we need invoke those after we save the blocks, because we want them to run last

mattsse (changes requested):
makes sense

I think invoking the stages after saving blocks make sense, I guess we could also introduce another trait for this that extend stage so that we can directly provide the new blocks but for now this seems easier

PR #18978, opened 2025-10-13
chore: replace poll_next_unpin loop with poll_recv_many

### LINKED ISSUE #18972: replace poll_next_unpin loop with poll_recv_many
### Describe the feature

we currently do: https://github.com/paradigmxyz/reth/blob/0f919a949e77df7b2a9b18e4990f32c1a4891c46/crates/net/network/src/transactions/mod.rs#L1537-L1538

however there's a better way for this which is `Receiver::poll_recv_many`

for this we need to remove this stream wrapper:

https://github.com/paradigmxyz/reth/blob/0f919a949e77df7b2a9b18e4990f32c1a4891c46/crates/net/network/src/transactions/mod.rs#L425-L425

as limit we can use `SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE` 

## TODO
* remove receiverstream wrapper and update the poll loop to use poll_recv_many

### Additional context

_No response_

stevencartavia:
closes #18972

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -1533,7 +1533,9 @@ where
             &mut new_txs,
             SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE,
         ) {
-            Poll::Ready(count) => count > 0,
+            Poll::Ready(count) => {
+                count == SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE
+            }
             Poll::Pending => false,
         };
         if !new_txs.is_empty() {
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -28,8 +28,7 @@ use self::constants::{tx_manager::*, DEFAULT_SOFT_LIMIT_BYTE_SIZE_TRANSACTIONS_B
 use crate::{
     budget::{
         DEFAULT_BUDGET_TRY_DRAIN_NETWORK_TRANSACTION_EVENTS,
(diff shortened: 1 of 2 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -1529,14 +1528,14 @@ where
...
-            DEFAULT_BUDGET_TRY_DRAIN_POOL_IMPORTS,
-            this.pending_transactions.poll_next_unpin(cx),
-            |hash| new_txs.push(hash)
-        );
+        let maybe_more_pending_txns = match this.pending_transactions.poll_recv_many(
+            cx,
+            &mut new_txs,
+            SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE,
+        ) {
+            Poll::Ready(count) => count > 0,
```
mattsse: don't think this is accurate, because we only have maybe more if count is equal to SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE

mattsse (changes requested):
smol nit

crates/net/network/src/transactions/mod.rs
```diff
@@ -1529,14 +1528,16 @@ where
...
+        let maybe_more_pending_txns = match this.pending_transactions.poll_recv_many(
+            cx,
+            &mut new_txs,
+            SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE,
+        ) {
+            Poll::Ready(count) => {
+                count == SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE
+            }
+            Poll::Pending => false,
+        };
```
emhane, line 1540: I see you guys chose to revert to draining this sub stream cc @op-will

mattsse (approved):

PR #18985, opened 2025-10-14
feat(otlp-tracing): enable to export traces with grpc export with `tracing-otlp` and `tracing-otlp-protocol` arg

lean-apple:
Follow-up for https://github.com/paradigmxyz/reth/pull/18242.

Closes https://github.com/paradigmxyz/reth/issues/18982.

Add `tracing-otlp-protocol` flag to either enable `http` or `grpc` export, if not specificied it will fallback to `http`.

Examples of use : 
- For http :
```shell
cargo run --bin reth -- --tracing-otlp=http://localhost:4318 node --dev
# or
cargo run --bin reth -- --tracing-otlp=http://localhost:4318 --tracing-otlp-protocol http node --dev
```
- For grpc :
```
cargo run --bin reth -- --tracing-otlp=http://localhost:4317 --tracing-otlp-protocol grpc node --dev
```

I have tested it locally with these config/docker-compose file, this time it will redirect the trace to `Jaeger`, we can then visualize the traces @ `localhost:16686`.

```
name: test-otlp-reth

services:
  jaeger:
    image: jaegertracing/all-in-one:latest
    environment:
      - COLLECTOR_OTLP_ENABLED=true
    ports:
      - "4317:4317"
      - "4318:4318"
      - "16686:16686"

  otel-collector:
    image: otel/opentelemetry-collector:latest
    command: ["--config=/etc/otel-collector-config.yaml"]
    volumes:
      - ./test-otlp.yaml:/etc/otel-collector-config.yaml
    ports:
      - "4317:4317"
      - "4318:4318"
    profiles:
      - collector
 ```

and this `test-otlp.yaml` file : 

```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

exporters:
  debug:
    verbosity: detailed
  logging:
    loglevel: debug

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug, logging]
```

### CODE REVIEW

File: crates/optimism/cli/src/app.rs [hunk 1]
```diff
@@ -114,23 +117,51 @@ where
...
 
             self.guard = self.cli.logs.init_tracing_with_layers(layers)?;
             info!(target: "reth::cli", "Initialized tracing, debug log directory: {}", self.cli.logs.log_file_directory);
         }
         Ok(())
     }
+
+    /// Initialize OTLP tracing export based on protocol type.
+    #[cfg(feature = "otlp")]
+    fn init_otlp_export(
```
lean-apple, line 145: Not sure where to put it to avoid the duplicate with op-cli app code

crates/cli/runner/src/lib.rs
```diff
@@ -36,11 +36,15 @@ impl CliRunner {
...
-}
 
-// === impl CliRunner ===
+    /// Executes an async block on the tokio runtime and blocks until completion.
+    pub fn block_on<F, T>(&self, fut: F) -> T
+    where
+        F: Future<Output = T>,
+    {
+        self.tokio_runtime.block_on(fut)
+    }
```
lean-apple: I can add this in another pr or we can use https://github.com/paradigmxyz/reth/pull/18018.

crates/ethereum/cli/src/app.rs
```diff
@@ -96,7 +98,8 @@ where
                 self.cli.logs.log_file_directory.join(chain_spec.chain().to_string());
         }
 
-        self.init_tracing()?;
+        self.init_tracing(Some(&runner))?;
```
Rjected: do we ever call `self.init_tracing(None)`? if not then maybe we should require the runtime for `init_tracing`, even though it may be unused if grpc is not used

crates/optimism/cli/src/app.rs
```diff
@@ -114,23 +117,51 @@ where
...
+        let protocol = self.cli.traces.protocol;
+        let level_filter = self.cli.traces.otlp_filter.clone();
+
+        match protocol {
+            OtlpProtocol::Grpc => {
+                let runner =
+                    runner.ok_or_else(|| eyre!("Runner required for gRPC OTLP initialization"))?;
+                runner.block_on(async {
+                    layers.with_span_layer("reth".to_string(), endpoint, level_filter, protocol)
+                })?;
```
Rjected, line 159: if this isn't an async fn (ie, no await), then why do we need to use the runner and call with `block_on`? Does the `build` fn for tonic need to be called in the context of a runtime?

Rjected (commented):
this looks good to me, I just have a question about the runtime and why it's needed

crates/optimism/cli/src/app.rs
lean-apple: It's a bit tricky, it's linked to the tonic exporter (there is one per observability event type),  and its builder.

For example for the span one, when we activate `with_tonic()` on the span builder, it will call 
[`build_span_exporter`](https://github.com/open-telemetry/opentelemetry-rust/blob/d72103720e2d39f6e28f2f928c388e85c85cef4d/opentelemetry-otlp/src/span.rs#L72), 

and then in this function, it will call [`build_channel`](https://github.com/open-telemetry/opentelemetry-rust/blob/d72103720e2d39f6e28f2f928c388e85c85cef4d/opentelemetry-otlp/src/exporter/tonic/mod.rs#L345),  and finally among this [function](https://github.com/open-telemetry/opentelemetry-rust/blob/d72103720e2d39f6e28f2f928c388e85c85cef4d/opentelemetry-otlp/src/exporter/tonic/mod.rs#L170) , 

I would say this part [`Channel::from_shared()`](https://github.com/open-telemetry/opentelemetry-rust/blob/d72103720e2d39f6e28f2f928c388e85c85cef4d/opentelemetry-otlp/src/exporter/tonic/mod.rs#L239),  needs a runtime for the tokio executor : 
--> `Endpoint::new_uri` at this level set up.
 --> [SharedExec::tokio()](https://docs.rs/tonic/latest/src/tonic/transport/channel/endpoint.rs.html#98),  

There is also probably the [`connect_lazy`](https://github.com/open-telemetry/opentelemetry-rust/blob/d72103720e2d39f6e28f2f928c388e85c85cef4d/opentelemetry-otlp/src/exporter/tonic/mod.rs#L254) inside `build_channel` that requires a runtime.

crates/ethereum/cli/src/app.rs
lean-apple: https://github.com/paradigmxyz/reth/pull/18985/commits/311d7d41fde4a8dcde285a491e9e5b8daa621f1e

crates/optimism/cli/src/app.rs
lean-apple: Yet I mainly discovered it was needed because of this error `there is no reactor running, must be called from the context of a Tokio 1.x runtime` when testing it, with nearly the same error path mentioned just before among the opentelemetry and tonic deps

Rjected: makes sense, then yeah we can keep this, mind just adding a doc comment explaining this?

crates/optimism/cli/src/app.rs
Rjected: IMO we can figure out how to properly dedup tracing + otlp init for OP / eth in a followup

crates/node/core/src/args/trace.rs
```diff
@@ -28,6 +30,21 @@ pub struct TraceArgs {
...
+    /// OTLP transport protocol to use for exporting traces.
+    ///
+    /// - `http`: expects endpoint path to end with `/v1/traces`
+    /// - `grpc`: expects endpoint without a path
+    ///
+    /// Defaults to HTTP if not specified.
+    #[arg(
+        long = "tracing-otlp-protocol",
+        global = true,
+        value_name = "PROTOCOL",
```
Rjected, line 43: let's name this env `OTEL_EXPORTER_OTLP_PROTOCOL`

Rjected (changes requested):
just one nit about the env var name, otherwise this looks good to me

crates/node/core/src/args/trace.rs
lean-apple: Done https://github.com/paradigmxyz/reth/pull/18985/commits/be76ab5cce71cb837d63897d5c79805206a6ea3b.

crates/optimism/cli/src/app.rs
lean-apple: Added a comment in top of `init_otlp_export` https://github.com/paradigmxyz/reth/pull/18985/commits/8a2867819f715058fee6aca4261cc5ed8f58727a

Rjected (approved):
lgtm, thanks!

PR #18988, opened 2025-10-14
fix(discv5): get `fork_id` from `Enr` for all network stacks

### LINKED ISSUE #18924: Discv5 doesn't get `fork_id` from `Enr` when discovering peer on OPEL network stack
### Describe the bug

Image

After looking at the code , I've seen that only `ETH` network stack is handled, others are ignored (there is a todo comment):
https://github.com/paradigmxyz/reth/blob/6f96a328128eb79a3009669b0a32f232971a99c9/crates/net/discv5/src/lib.rs#L323-L326

Once modified to accept also `OPEL`:

Image
 

Are there any side effects that I might have missed with this modification? Is this relevant and would you like to have this modified ? Thanks 👍

### Steps to reproduce

```bash
RUST_LOG=info,net::discv5=trace ./target/release/op-reth node --dev\
  --disable-dns-discovery\
  --disable-discv4-discovery\
  --enable-discv5-discovery\
  --disable-nat\
   --no-persist-peers\
```

### Node logs

```text

```

### Platform(s)

Mac (Apple Silicon)

### Container Type

Not running in a container

master (6f96a328128eb79a3009669b0a32f232971a99c9)

2

dev

Archive (default)

_No response_

cargo build -r -p op-reth

### Code of Conduct

mablr:
Fix #18924

### DIFF
--- a/crates/net/discv5/src/error.rs
+++ b/crates/net/discv5/src/error.rs
@@ -13,7 +13,7 @@ pub enum Error {
     #[error("network stack identifier is not configured")]
     NetworkStackIdNotConfigured,
     /// Missing key used to identify rlpx network.
-    #[error("fork missing on enr, with local network id '{id}', and 'eth' always accepted as fallback", id = String::from_utf8_lossy(.0))]
+    #[error("fork missing on enr, key {0:?} and key 'eth' missing")]
     ForkMissing(&'static [u8]),
     /// Failed to decode [`ForkId`](reth_ethereum_forks::ForkId) rlp value.
     #[error("failed to decode fork id, 'eth': {0:?}")]
--- a/crates/net/discv5/src/error.rs
+++ b/crates/net/discv5/src/error.rs
@@ -13,7 +13,7 @@ pub enum Error {
     #[error("network stack identifier is not configured")]
     NetworkStackIdNotConfigured,
     /// Missing key used to identify rlpx network.
-    #[error("fork missing on enr, key missing")]
+    #[error("fork missing on enr, with local network id '{id}', and 'eth' always accepted as fallback", id = String::from_utf8_lossy(.0))]
     ForkMissing(&'static [u8]),
     /// Failed to decode [`ForkId`](reth_ethereum_forks::ForkId) rlp value.
     #[error("failed to decode fork id, 'eth': {0:?}")]
--- a/crates/net/discv5/src/lib.rs
+++ b/crates/net/discv5/src/lib.rs
@@ -924,7 +924,7 @@ mod test {
         let enr_with_opel = Enr::builder()
             .add_value_rlp(
                 NetworkStackId::OPEL,
-                alloy_rlp::encode(&EnrForkIdEntry::from(fork_id)).into(),
+                alloy_rlp::encode(EnrForkIdEntry::from(fork_id)).into(),
             )
             .build(&sk)
             .unwrap();
@@ -937,7 +937,7 @@ mod test {
         let enr_with_eth = Enr::builder()
             .add_value_rlp(
                 NetworkStackId::ETH,
-                alloy_rlp::encode(&EnrForkIdEntry::from(fork_id)).into(),
+                alloy_rlp::encode(EnrForkIdEntry::from(fork_id)).into(),
             )
             .build(&sk)
             .unwrap();
--- a/crates/net/discv5/src/lib.rs
+++ b/crates/net/discv5/src/lib.rs
@@ -388,11 +388,7 @@ impl Discv5 {
                 (key != NetworkStackId::ETH)
                     .then(|| {
                         // Fallback: trying to get fork id from Enr with 'eth' as network stack id
-                        trace!(
-                            target: "net::discv5",
-                            "fork id not found for '{key}' network stack id, trying 'eth'",
-                            key = String::from_utf8_lossy(key),
-                        );
+                        trace!(target: "net::discv5", "Fork id not found for key, trying 'eth'...");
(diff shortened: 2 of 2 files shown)


mattsse (approved):
lgtm

pending @emhane

emhane (approved):

emhane:
thanks @mablr ! ETH is EL and ETH2 is CL.

ok so based on op-geth behaviour, let's do this
here
https://github.com/paradigmxyz/reth/blob/856ad087766476f9fd20624c136951f654d5fdaa/crates/net/discv5/src/lib.rs#L387-L391
we want to do smthg like this
```rust
        let Some(key) = self.fork_key else { return Err(Error::NetworkStackIdNotConfigured) };
        let fork_id = enr
            .get_decodable::<EnrForkIdEntry>(key)
            .ok_or(enr
                .get_decodable::<EnrForkIdEntry>(NetworkStackId::ETH)
                .ok_or(Error::ForkMissing(key)?
            )?
            .map(Into::into)?;
```
the syntax in my example is a bit off but I think you get the point? would you mind implementing it? would be good if we had a trace log for the case when the configured key isn't found. we will also need to change the error message for `Error::ForkMissing(key)` to say that neither `{key}` nor 'eth' were found among the ENR kv-pairs. need to double check also that metrics are update correctly so they don't increment the count for e.g. 'opel' hits when actually it just found the 'eth' key.

crates/net/discv5/src/error.rs
```diff
@@ -13,7 +13,7 @@ pub enum Error {
     #[error("network stack identifier is not configured")]
     NetworkStackIdNotConfigured,
     /// Missing key used to identify rlpx network.
-    #[error("fork missing on enr, key missing")]
+    #[error("fork missing on enr, with local network id '{id}', and 'eth' always accepted as fallback", id = String::from_utf8_lossy(.0))]
```
emhane:
```suggestion
    #[error("fork missing on enr, key {0} and key 'eth' missing"}]
```
can be simplified to. don't need to decode it.

crates/net/discv5/src/lib.rs
```diff
@@ -384,7 +384,23 @@ impl Discv5 {
...
             .get_decodable::<EnrForkIdEntry>(key)
-            .ok_or(Error::ForkMissing(key))?
+            .or({
+                (key != NetworkStackId::ETH)
+                    .then(|| {
+                        // Fallback: trying to get fork id from Enr with 'eth' as network stack id
+                        trace!(
+                            target: "net::discv5",
+                            "fork id not found for '{key}' network stack id, trying 'eth'",
+                            key = String::from_utf8_lossy(key),
```
emhane:
```suggestion
                            key = String::from_utf8_lossy(key),
                            "Fork id not found for key, trying 'eth'...",
```
message usually goes last. since we have the kv pairs we don't substitute values into the message.

crates/net/discv5/src/lib.rs
```diff
@@ -384,7 +384,23 @@ impl Discv5 {
         let Some(key) = self.fork_key else { return Err(Error::NetworkStackIdNotConfigured) };
         let fork_id = enr
             .get_decodable::<EnrForkIdEntry>(key)
-            .ok_or(Error::ForkMissing(key))?
+            .or({
+                (key != NetworkStackId::ETH)
+                    .then(|| {
+                        // Fallback: trying to get fork id from Enr with 'eth' as network stack id
+                        trace!(
+                            target: "net::discv5",
```
emhane:
```suggestion
                        trace!(target: "net::discv5",
```
can save some whitespace with target on same line like that and then put the next values in column like you have done, nice.

emhane (approved):
nice! nitpicks remaining

crates/net/discv5/src/lib.rs
```diff
@@ -388,11 +388,7 @@ impl Discv5 {
                 (key != NetworkStackId::ETH)
                     .then(|| {
                         // Fallback: trying to get fork id from Enr with 'eth' as network stack id
-                        trace!(
-                            target: "net::discv5",
-                            "fork id not found for '{key}' network stack id, trying 'eth'",
-                            key = String::from_utf8_lossy(key),
-                        );
+                        trace!(target: "net::discv5", "Fork id not found for key, trying 'eth'...");
```
emhane:
```suggestion
                        trace!(target: "net::discv5",
                            key = String::from_utf8_lossy(key),
                            "Fork id not found for key, trying 'eth'..."
                        );
```
excuse me if wasn't clear, we want to keep the key kv-pair like it was!

mablr: no problem, I reverted it. Assuming decoding here is useful.

crates/net/discv5/src/lib.rs
```diff
@@ -388,7 +388,10 @@ impl Discv5 {
                 (key != NetworkStackId::ETH)
                     .then(|| {
                         // Fallback: trying to get fork id from Enr with 'eth' as network stack id
-                        trace!(target: "net::discv5", "Fork id not found for key, trying 'eth'...");
+                        trace!(target: "net::discv5",
+                            "Fork id not found for '{key}' network stack id, trying 'eth'...",
+                            key = String::from_utf8_lossy(key)
+                        );
```
emhane, line 394:
```suggestion
                        trace!(target: "net::discv5",
                            key = String::from_utf8_lossy(key),
                            "Fork id not found for key, trying 'eth'..."
                        );
```

emhane:
Screenshot 2025-10-16 at 11 19 44
check out how it looks with the message and then the kv-pairs

PR #18989, opened 2025-10-14
chore(optimism): remove unnecessary Debug bounds from header generics

MozirDmitriy:
- Removed Debug trait bounds from verify_withdrawals_root and verify_withdrawals_root_prehashed, and dropped the unused Debug import.
- Debug was not used in these functions (no formatting/logging of header), and its presence over-constrained the API.
- Aligns with canyon validation which only requires H: BlockHeader, improving consistency across consensus validation.


mattsse (approved):
fair

PR #18990, opened 2025-10-14
ci(hive): use Go 1.24.x in hive workflow

MamunC0der:
Update .github/workflows/hive.yml to install Go 1.24.x via actions/setup-go@v6.

- Replace "^1.13.1" with "1.24.x"
- Ensures compatibility with benchmarks using b.Loop()
- Improves consistency with the rest of CI jobs


mattsse (commented):
doesnt seem necessary

PR #18991, opened 2025-10-14
refactor(tokio-util): use public API instead of direct field access in rate limit test

anim001k:
### Summary
Replaced direct field access to `Rate::duration` with the public `duration()` method in rate limit test for better encapsulation.

### Changes
- Updated test to use `limit.rate.duration()` instead of `limit.rate.duration`
- Consistent with existing test patterns in the same file


mattsse (commented):
I prefer using the full path here

PR #18992, opened 2025-10-14
refactor(storage): fix ChainStateKey enum variant name

anim001k:
### Changes
- Renamed enum variant from `LastSafeBlockBlock` to `LastSafeBlock`
- Updated corresponding `Encode` and `Decode` implementations
- Updated all references in database provider

### DIFF
--- a/crates/storage/db-api/src/tables/mod.rs
+++ b/crates/storage/db-api/src/tables/mod.rs
@@ -531,7 +531,7 @@ pub enum ChainStateKey {
     /// Last finalized block key
     LastFinalizedBlock,
     /// Last safe block key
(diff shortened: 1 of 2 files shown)


mattsse (approved):
this looks okay, dont think the serde impl is used anywhere

but perhaps @shekhirin can double check this

shekhirin (approved):
LGTM, @mattsse it's used here https://github.com/paradigmxyz/reth/blob/5065890823eb6358845218f5e0f3219c6544e63a/crates/storage/provider/src/providers/blockchain_provider.rs#L105-L111 but I believe it's fine because we'll just fallback to finalized block and the correct safe block will be written on next persistence.

PR #18994, opened 2025-10-14
chore: remove unused Sealable import from blockchain provider

eeemmmmmm:
The file had an unused alloy_primitives::Sealable import hidden behind #![allow(unused)]. Dropping it cuts dead weight while leaving behaviour unchanged. cargo test -p reth-provider still passes, so nothing regressed.


mattsse (commented):
superseded by #18867

PR #18996, opened 2025-10-14
fix: start producer after subscription to avoid timing flakiness

anim001k:
### Summary
Start the producer thread after the subscription is established in `test_rpc_subscription` to remove timing-related flakiness.

### Changes
- Move `std::thread::spawn(move || produce_items(tx))` to run after `subscribe(...)` returns.


mattsse (commented):
we dont need this

PR #18998, opened 2025-10-14
refactor(storage/provider): remove redundant AsRef<Self> impl from DatabaseProvider

sashaodessa:
### Summary
Remove a no-op `AsRef<Self>` implementation for `DatabaseProvider`; borrowing `&DatabaseProvider` does not require a trait.


mattsse (commented):
we need this

PR #19012, opened 2025-10-15
perf: background init of workers

### LINKED ISSUE #18995: perf: intialise worker in bg instead
### Describe the feature

`ProofWorkerHandle::new()` blocks the calling thread during worker setup. we should returning the channels immediately and doing the setup in the background  instead.

### Additional context
https://github.com/paradigmxyz/reth/pull/18934#discussion_r2428551192

yongkangc:
- spawn workers in the background and return the channel handle first. part of cleanup and micro optimisations of worker pool. 

closes: https://github.com/paradigmxyz/reth/issues/18995

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -308,7 +308,7 @@ fn account_worker_loop<Tx>(
                 );
                 tracker.set_precomputed_storage_roots(storage_root_targets_len as u64);
 
-                let storage_proof_receivers = match queue_storage_proofs(
+                let storage_proof_receivers = match dispatch_storage_proofs(
                     &storage_work_tx,
                     &input.targets,
                     &mut storage_prefix_sets,
@@ -568,7 +568,7 @@ where
 /// computation. This enables interleaved parallelism for better performance.
 ///
 /// Propagates errors up if queuing fails. Receivers must be consumed by the caller.
-fn queue_storage_proofs(
+fn dispatch_storage_proofs(
     storage_work_tx: &CrossbeamSender<StorageWorkerJob>,
     targets: &MultiProofTargets,
     storage_prefix_sets: &mut B256Map<PrefixSet>,
@@ -963,7 +963,7 @@ impl ProofWorkerHandle {
     }
 
     /// Queue an account multiproof computation
-    pub fn queue_account_multiproof(
+    pub fn dispatch_account_multiproof(
         &self,
         input: AccountMultiproofInput,
     ) -> Result<Receiver<AccountMultiproofResult>, ProviderError> {
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -121,26 +121,9 @@ fn storage_worker_loop<Factory>(
     Factory: DatabaseProviderFactory<Provider: BlockReader>,
 {
     // Create db transaction before entering work loop
-    let proof_tx = match view.provider_ro() {
-        Ok(provider) => {
-            let tx = provider.into_tx();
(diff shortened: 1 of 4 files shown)


mattsse (approved):

crates/trie/parallel/src/proof_task.rs
```diff
@@ -112,34 +118,78 @@ enum StorageWorkerJob {
...
+    let mut trie_cursor_factory: Option<_> = None;
+    let mut hashed_cursor_factory: Option<_> = None;
+    let mut blinded_provider_factory: Option<_> = None;
 
     let mut storage_proofs_processed = 0u64;
     let mut storage_nodes_processed = 0u64;
 
     while let Ok(job) = work_rx.recv() {
+        // Perform lazy initialization on first job
+        if proof_tx.is_none() {
```
shekhirin: can we initialize it not lazily, but eagerly, to not spend time on it when the first job is received?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -258,35 +314,79 @@ fn storage_worker_loop<Tx>(
...
+                    tracing::error!(
+                        target: "trie::proof_task",
+                        worker_id,
+                        error = %e,
+                        "Account worker failed to initialize transaction, exiting"
+                    );
+                    return;
+                }
+            }
+        }
```
shekhirin: it's the same as for storage proofs, right? can be deduplicated?

crates/trie/parallel/src/proof_task.rs
yongkangc: yeah makes sense

crates/trie/parallel/src/proof_task.rs
```diff
@@ -258,15 +263,21 @@ fn storage_worker_loop<Tx>(
...
     storage_work_tx: CrossbeamSender<StorageWorkerJob>,
     worker_id: usize,
     #[cfg(feature = "metrics")] metrics: ProofTaskTrieMetrics,
 ) where
-    Tx: DbTx,
+    Factory: DatabaseProviderFactory<Provider: BlockReader>,
 {
+    // Create db transaction before entering work loop
+    let provider = view.provider_ro()
+        .expect("Account worker failed to initialize: database unavailable");
```
yongkangc, line 278: instead of having expect 

https://github.com/paradigmxyz/reth/pull/19012/files#diff-e3fb5658ad9c8b5024a6858cc898487c1961023c6361aeab5b4dcdd7ef781fb6L1231-L1233

its much simpler to expect here, and we dont have to propogate the error up

crates/trie/parallel/src/proof_task.rs
```diff
@@ -112,14 +111,20 @@ enum StorageWorkerJob {
...
     work_rx: CrossbeamReceiver<StorageWorkerJob>,
     worker_id: usize,
     #[cfg(feature = "metrics")] metrics: ProofTaskTrieMetrics,
 ) where
-    Tx: DbTx,
+    Factory: DatabaseProviderFactory<Provider: BlockReader>,
 {
+    // Create db transaction before entering work loop
+    let provider = view.provider_ro()
+        .expect("Storage worker failed to initialize: database unavailable");
```
yongkangc, line 125: Same for 

> instead of having expect 
> 
> https://github.com/paradigmxyz/reth/pull/19012/files#diff-e3fb5658ad9c8b5024a6858cc898487c1961023c6361aeab5b4dcdd7ef781fb6L1231-L1233
> 
> its much simpler to expect here, and we dont have to propogate the error up

mediocregopher (approved):

mattsse (approved):

PR #19022, opened 2025-10-15
feat: add pending sequence as pub

loocapro:
Makes the `FlashBlockPendingSequence` public.

### DIFF
--- a/crates/optimism/flashblocks/src/sequence.rs
+++ b/crates/optimism/flashblocks/src/sequence.rs
@@ -11,6 +11,15 @@ use tracing::{debug, trace, warn};
 /// The size of the broadcast channel for completed flashblock sequences.
 const FLASHBLOCK_SEQUENCE_CHANNEL_SIZE: usize = 128;
 
+impl<T> Default for FlashBlockPendingSequence<T>
+where
+    T: SignedTransaction,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -11,9 +11,18 @@ use tracing::{debug, trace, warn};
...
 const FLASHBLOCK_SEQUENCE_CHANNEL_SIZE: usize = 128;
 
+impl<T> Default for FlashBlockPendingSequence<T>
+where
+    T: SignedTransaction,
+{
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
mattsse: this impl is out of order, does not belong above type def

crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -70,15 +80,15 @@ where
...
+    pub fn insert(&mut self, flashblock: FlashBlock) -> eyre::Result<()> {
         if flashblock.index == 0 {
             trace!(number=%flashblock.block_number(), "Tracking new flashblock sequence");
 
             // Flash block at index zero resets the whole state.
             self.clear_and_broadcast_blocks();
 
             self.inner.insert(flashblock.index, PreparedFlashBlock::new(flashblock)?);
-            return Ok(())
+            return Ok(());
```
mattsse: please ser no more 😭

mattsse (changes requested):
nits and fmt issues

mattsse (approved):

PR #19028, opened 2025-10-15
fix: return ExitCode from main instead of calling process::exit

vtjl10:
Description
- Summary: Replace direct std::process::exit(1) with returning std::process::ExitCode in bin/reth/src/main.rs for graceful shutdown.
- Why: Ensures destructors run, background tasks can finish, and logs/metrics flush; improves testability and maintainability.


mattsse (commented):
we already exit with 1 on error

PR #19029, opened 2025-10-15
refactor(net): Deduplicate response handling in EthRequestHandler

vtjl10:
This PR refactors `EthRequestHandler` to eliminate duplicated code for handling p2p requests.

**Summary**

The logic for fetching data and enforcing response limits (`MAX_*_SERVE` and `SOFT_RESPONSE_LIMIT`) was repeated across multiple request handlers, particularly `on_bodies_request` and `get_receipts_response`.

This change introduces a single generic helper function, `get_response_items`, to encapsulate this shared logic. The existing handlers have been updated to use this new function, resulting in cleaner and more maintainable code.

**Changes**

- **Added** a new private generic function `get_response_items` in `eth_requests.rs`.
- **Refactored** `on_bodies_request` and `get_receipts_response` to delegate response fetching and limit checking to the new helper.

This improves maintainability by ensuring that any future changes to response limiting logic only need to be made in one place.

### DIFF
--- a/crates/net/network/src/eth_requests.rs
+++ b/crates/net/network/src/eth_requests.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use alloy_consensus::{BlockHeader, ReceiptWithBloom};
 use alloy_eips::BlockHashOrNumber;
+use alloy_primitives::B256;
 use alloy_rlp::Encodable;
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this doesnt remove code and removes flexibility in case we need to change something

PR #19039, opened 2025-10-15
fix(net): correct error messages for decrypt and header paths

GarmashAlex:
This change corrects two misleading ECIES error messages to match the actual failure sites and improve debugging clarity. TagCheckDecryptFailed was incorrectly worded as “tag check failure in read_header” despite only being raised in the ECIES decrypt path (EncryptedMessage::check_integrity during auth/ack payload decryption), and InvalidHeader was incorrectly worded as “invalid body data” although it is raised exclusively by read_header for malformed/undersized RLPx headers


mattsse (approved):
fair

PR #19045, opened 2025-10-15
chore(primitives-traits): relax SignerRecoverable bounds for Extended<B,T>

sashass1315:
Replace asymmetric and unnecessary bounds B: SignedTransaction + IsTyped2718, T: SignedTransaction with B: SignerRecoverable, T: SignerRecoverable in impl SignerRecoverable for Extended<B,T>.

Reason: The impl only delegates recovery methods and does not require typed-2718 or full SignedTransaction capabilities. This makes bounds minimal, symmetric, and consistent with actual usage, while keeping impl SignedTransaction for Extended<B,T> unchanged where B: IsTyped2718 is required to satisfy the Decodable2718 supertrait.

### DIFF
--- a/crates/primitives-traits/src/extended.rs
+++ b/crates/primitives-traits/src/extended.rs
@@ -142,8 +142,8 @@ where
 
 impl<B, T> SignerRecoverable for Extended<B, T>
 where
-    B: SignedTransaction + IsTyped2718,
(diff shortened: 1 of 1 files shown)


mattsse (approved):
looks like this is enough

PR #19047, opened 2025-10-15
chore: handle JSON parsing safely in custom_chain

0xsimka:
replaced `unwrap()` with proper error handling so invalid genesis JSON won’t crash the node. functionality stays the same.


mattsse (commented):
this is fine for the example

PR #19050, opened 2025-10-15
feat: use env filter for otlp, respect otel env var

prestwich:
- enhance the OTLP  args by allowing a full env filter
    - Note: this is backwards compatible, as all levels are also valid filter directives
- allow lazy url specification by setting the path if the path is unset
- respect [OTEL conventions](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) by allowing the url to be set via the standard env var for trace exporting
- improve documentation

### DIFF
--- a/crates/ethereum/cli/src/app.rs
+++ b/crates/ethereum/cli/src/app.rs
@@ -119,7 +119,7 @@ where
                 layers.with_span_layer(
                     "reth".to_string(),
                     output_type.clone(),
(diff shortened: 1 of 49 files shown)


mattsse (commented):
this does make sense but would appreciate another review from @Rjected or @shekhirin 

cc @lean-apple for vis

Rjected (approved):
ty, the env filter is much more flexible, and env var makes sense

PR #19051, opened 2025-10-15
fix(session): remove Clone derive from SessionCounter

radik878:
Remove Clone from SessionCounter to prevent accidental duplication of mutable state. SessionCounter tracks live counters within SessionManager and is not cloned anywhere; keeping Clone is a footgun that could lead to diverging metrics if copies are modified independently. For cloneable counters we use Arc-based wrappers (see DisconnectionsCounter); SessionCounter should remain non-clone to reflect its ownership and mutation model.


mattsse (approved):
I like this

PR #19052, opened 2025-10-15
refactor(txpool): reuse cached gas_limit value

gancerlory:
Eliminates a redundant `transaction.gas_limit()` call by reusing the already-cached `transaction_gas_limit` variable.


mattsse (approved):
makes sense

PR #19053, opened 2025-10-15
refactor(consensus): eliminate redundant header method calls

gancerlory:
Cache header field method results to avoid calling `gas_used()`, `gas_limit()`, `number()`, `timestamp()`, and `max_blob_gas_per_block()` multiple times within the same validation functions.


mattsse (commented):
@DaniPopes do you think this is worth doing?

DaniPopes:
No these are just simple getters

PR #19056, opened 2025-10-15
[wip] feat(osaka): convert blobs around transition

klkvr:


crates/transaction-pool/src/lib.rs
```diff
@@ -511,9 +510,40 @@ where
...
+                TransactionValidationOutcome::Valid {
+                    transaction:
+                        ValidTransaction::ValidWithSidecar {
+                            transaction,
+                            sidecar: BlobTransactionSidecarVariant::Eip4844(sidecar),
+                        },
+                    ..
+                } if self.pool.is_eip7594_activated() => {
+                    let this = self.clone();
+                    tokio::spawn(async move {
```
mattsse: do we need this spawn here even if the converter also spawns the converstion?

mattsse (changes requested):
in order for this to be reachable we also need to disable the check in validator itself because this currently enforce pre vs post sidecar versions

crates/transaction-pool/src/lib.rs
klkvr: yeah, otherwise we'd force insertion of entire batch to wait for single tx's semaphore

crates/transaction-pool/src/lib.rs
```diff
@@ -420,6 +424,84 @@ where
...
+                            sidecar:
+                                PoolTransactionSidecar::Eip4844 { sidecar, should_convert: true },
+                        },
+                    ..
+                } => {
+                    let this = self.clone();
+                    let hash = *transaction.hash();
+                    tokio::task::spawn(async move {
+                        let Some(converted) =
+                            this.pool.blob_sidecar_converter().convert(sidecar).await
```
mattsse, line 461: this feels a bit weird because we don't really ensure that we actually add the transaction if I understand this correctly

PR #19063, opened 2025-10-16
fix(net): Increment out_of_order_requests in BodiesDownloader on range reset

GarmashAlex:
This change increments the bodies downloader metric out_of_order_requests when the download range is reset and the new start block is below the last block returned by the stream. This matches the metric’s documented intent (“count out-of-order re-requests after unwind or external writes”) and aligns behavior with the headers downloader, which already increments its analogous metric on sync target changes. The change adds a simple check in BodiesDownloader::set_download_range before self.clear() to increment the counter when range.start() < latest_queued_block_number. No functional behavior is altered; this improves observability and monitoring accuracy for out-of-order body requests.


mattsse (approved):
this makes sense because we also use this when updating the target

PR #19069, opened 2025-10-16
feat: add subscription stream function for transaction receipts

### LINKED ISSUE #19055: Add subscription stream function for transaction receipts
### Describe the feature

prep for https://github.com/alloy-rs/alloy/pull/2974

we can already implement a function for this, similar to:

https://github.com/paradigmxyz/reth/blob/5c19ce75805d8477b6e839cc0afd259c8f99da77/crates/rpc/rpc/src/eth/pubsub.rs#L87-L90

we need to leverage the conversion trait for the tx to receipt conversion

https://github.com/paradigmxyz/reth/blob/5c19ce75805d8477b6e839cc0afd259c8f99da77/crates/rpc/rpc/src/eth/pubsub.rs#L65-L67

https://github.com/paradigmxyz/reth/blob/5c19ce75805d8477b6e839cc0afd259c8f99da77/crates/rpc/rpc-convert/src/transaction.rs#L179-L187

basically just like

https://github.com/paradigmxyz/reth/blob/5c19ce75805d8477b6e839cc0afd259c8f99da77/crates/rpc/rpc-eth-api/src/helpers/block.rs#L126-L162

which we could actually convert to a helper fn in rpconvert first maybe so that this can be reused 

maybe @futreall

### Additional context

_No response_

futreall:
Implements transaction_receipts_stream method in EthPubSub similar to new_headers_stream.
Adds build_convert_receipt_inputs helper function in rpc-convert for reusability.
Prep work for alloy-rs/alloy#2974.

Closes #19055


crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -201,7 +203,13 @@ where
                 Ok(())
             }
             _ => {
-                // TODO: implement once https://github.com/alloy-rs/alloy/pull/2974 is released
+                // For now, check if this is a TransactionReceipts subscription kind
+                // This is a temporary solution until alloy#2974 is merged
+                if format!("{:?}", kind).contains("TransactionReceipts") {
+                    // We need to handle this differently since transaction_receipts_stream
+                    // has different trait bounds than the current context
+                    return Err(invalid_params_rpc_err("TransactionReceipts subscription requires additional trait bounds - see alloy#2974"));
```
mattsse: I'd like to keep this unchanged for now

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -372,7 +380,46 @@ where
...
+where
+    Eth: RpcNodeCore<Provider: CanonStateSubscriptions<Primitives = N>>
+        + EthApiTypes<
+            RpcConvert: RpcConvert<Primitives = N, Network = Eth::NetworkTypes, Error = Eth::Error>,
+        >,
+{
+    /// Returns a stream that yields all transaction receipts from new blocks.
+    fn transaction_receipts_stream(
+        &self,
+    ) -> impl Stream<Item = reth_rpc_convert::RpcReceipt<Eth::NetworkTypes>> {
```
mattsse: actually, now that I'm reviewing this we can make this more efficient specifically for the upcoming subscription kind and already make this accept an optional list of hashes

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -372,7 +380,46 @@ where
...
+        let eth_api = self.eth_api.clone();
+        self.eth_api.provider().canonical_state_stream().flat_map(move |new_chain| {
+            let mut all_receipts = Vec::new();
+
+            // Process all blocks in the committed chain - exactly like new_headers_stream pattern
+            for (block, receipts) in new_chain.committed().blocks_and_receipts() {
+                // Build ConvertReceiptInput using our helper function
+                let inputs = build_convert_receipt_inputs(block, receipts);
```
mattsse: then we perhaps might not need this because we need to filter them receipts based on the hashes input

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -372,7 +380,46 @@ where
...
     }
+}
+
+impl<N: NodePrimitives<SignedTx: TxHashRef>, Eth> EthPubSubInner<Eth>
+where
+    Eth: RpcNodeCore<Provider: CanonStateSubscriptions<Primitives = N>>
+        + EthApiTypes<
+            RpcConvert: RpcConvert<Primitives = N, Network = Eth::NetworkTypes, Error = Eth::Error>,
+        >,
+{
```
mattsse: we need to wire this into the match stmt above so we can instead extend the bounds on the existing impl block already, then we dont need to unify them later

mattsse (changes requested):
cool, a few suggestions

crates/rpc/rpc/src/eth/pubsub.rs
futreall: removed it completely

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -200,14 +202,31 @@ where
 
                 Ok(())
             }
-            _ => {
-                // TODO: implement once https://github.com/alloy-rs/alloy/pull/2974 is released
-                Err(invalid_params_rpc_err("Unsupported subscription kind"))
```
mattsse, line 197: we can keep this unchanged

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -200,14 +202,31 @@ where
...
+            RpcConvert: RpcConvert<Primitives = N, Network = Eth::NetworkTypes, Error = Eth::Error>,
+        >,
+{
+    /// Returns a stream that yields all transaction receipts from new blocks.
+    pub fn transaction_receipts_stream(
+        &self,
+        hashes: Option<Vec<TxHash>>,
+    ) -> impl Stream<Item = reth_rpc_convert::RpcReceipt<Eth::NetworkTypes>> {
+        self.inner.transaction_receipts_stream(hashes)
+    }
```
mattsse: to answer your question we can also move this into the existing block and instead already update the bounds of the existing impl block because we'll need to do this anyway

mattsse (changes requested):
few more nits and q for @klkvr

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -1100,6 +1102,63 @@ impl TryFromTransactionResponse<alloy_network::Ethereum>
...
+///
+/// This is extracted from the pattern used in `eth_getBlockReceipts` to enable reuse in
+/// subscription streams.
+pub fn build_convert_receipt_inputs<'a, N: NodePrimitives>(
+    block: &'a reth_primitives_traits::RecoveredBlock<N::Block>,
+    receipts: &'a [N::Receipt],
+) -> Vec<ConvertReceiptInput<'a, N>>
+where
+    N::SignedTx: TxHashRef,
+{
```
mattsse, line 977: @klkvr thoughts on this

whould we perhaps make this a native fn of RecoveredBlock ?

klkvr: yep sgtm, though`ConvertReceiptInput::from_block_and_receipts` might be eaiser here because `ConvertReceiptInput` is RPC-only type

also I would prefer this to be re-used in https://github.com/paradigmxyz/reth/blob/447394f1f05d9878c02da8ed339180a4a949a433/crates/rpc/rpc-eth-api/src/helpers/block.rs#L126-L157

futreall: @mattsse separate PR for the refactor or include here?

mattsse: @futreall let's do a separate one for this

PR #19075, opened 2025-10-16
fix(rpc): correct highest_block value in eth_syncing response

Fallengirl:
Fix logical inconsistency in eth_syncing RPC endpoint where highest_block was incorrectly set to current_block during synchronization.
When is_syncing() returns true, highest_block should represent the network's target height, not the current local block. Setting highest_block = current_block is misleading and contradicts Ethereum RPC specification.
This ensures honest reporting when the actual network target is unavailable, preventing misleading sync progress information.


mattsse (commented):
we can keep this as current block

PR #19080, opened 2025-10-16
chore(docker): remove apt-get upgrade to ensure reproducible and faster builds

zeevick10:
### Description
- **Summary**: Remove `apt-get -y upgrade` from the build stage in `Dockerfile`.
- **Why**: Avoid non-deterministic builds, reduce cache busting, and speed up image builds.
- **Change**: Replace `RUN apt-get update && apt-get -y upgrade && apt-get install -y ...` with `RUN apt-get update && apt-get install -y ...`.


mattsse (commented):
this has been working fine but I do the the upgrade isnt necessary here?
@shekhirin ?

shekhirin (approved):
Yeah, not necessary.

PR #19081, opened 2025-10-16
ci: verify geth tarball checksum before extraction

zeevick10:
### Description
- **Summary**: Add SHA256 verification to `.github/assets/install_geth.sh` before extracting the tarball.
- **Change**: Download `*.sha256` and run `sha256sum -c` prior to `tar -xvf`.
- **Why**: Reduce supply-chain risk and prevent corrupted/altered artifacts from running in CI.


mattsse (commented):
we only run this in ci so we dont need this

PR #19082, opened 2025-10-16
ci(check_wasm): enable set -e -o pipefail for reliable pipeline failures

zeevick10:
### Description
- **Summary**: Replace global `set +e` with `set -e -o pipefail` at the start of `.github/assets/check_wasm.sh`.
- **Why**: Prevent hidden failures in the `cargo | jq | grep | sort` pipeline; ensure early, accurate error reporting.


mattsse (commented):
this is working fine

PR #19083, opened 2025-10-16
fix(net): remove redundant remove of evicted hash in fetcher

MozirDmitriy:
Remove a redundant call to remove an already-evicted key from hashes_pending_fetch after insert_and_get_evicted. The LruCache::insert_and_get_evicted API guarantees the evicted entry is already removed from that cache, so calling remove again is a no-op. This avoids an unnecessary operation without changing behavior.


mattsse (approved):
good find

PR #19084, opened 2025-10-16
feat: convert blobs at RPC

klkvr:


mattsse (approved):
lgtm!

as discussed one suggestion re osaka activation check

crates/rpc/rpc/src/eth/helpers/transaction.rs
```diff
@@ -34,7 +42,52 @@ where
...
+                )));
+            };
+
+            let sidecar = match sidecar {
+                BlobTransactionSidecarVariant::Eip4844(sidecar) => {
+                    let latest = self
+                        .provider()
+                        .latest_header()?
+                        .ok_or(EthApiError::HeaderNotFound(BlockId::latest()))?;
+                    if self.provider().chain_spec().is_osaka_active_at_timestamp(latest.timestamp())
```
mattsse: should we also use latest.timestamp() + 12 or 20 here?

klkvr: changed to start converting 12s before in f88ddd9 but I think it should anyway be fine because any legacy sidecars we accept will end up handled by https://github.com/paradigmxyz/reth/pull/19095

mattsse (approved):

PR #19086, opened 2025-10-16
fix: Remove deprecated SealedBlockFor alias from primitives

maradini77:
Remove the deprecated SealedBlockFor type alias which duplicated SealedBlock.
The alias adds noise to the public API and may confuse users. It’s already deprecated in favor of RecoveredBlock.


mattsse (commented):
we can keep this because we're removing this entire crte eventually

PR #19088, opened 2025-10-16
feat(cli): add method `CliRunner::block_on`

lean-apple:
Add method for `CliRunner::block_on`.

It is be useful for https://github.com/paradigmxyz/reth/pull/18985 to enable some asynchronous step.


mattsse (approved):
lgtm, reasonable

PR #19089, opened 2025-10-16
fix: support rlp hex in read_header_from_file

mattsse:
this would fail if the file content is hex rlp

this now supports both variants


RomanHodulak (approved):
LGTM

PR #19090, opened 2025-10-16
refactor: simplify GasCap conversion in EthApi

rnkrtt:
Replace `gas_cap.into().into()` with `gas_cap.into().0` for direct field access instead of redundant double conversion. Improves code readability while maintaining identical functionality.


mattsse (commented):
this is equivalent

PR #19092, opened 2025-10-16
refactor(net): remove unused secret_key from Peer and PeerConfig

Forostovec:
This refactor removes the redundant secret_key fields from Peer and PeerConfig in crates/net/network/src/test_utils/testnet.rs. The node’s secret key is already the single source of truth within NetworkConfig/NetworkManager/NetworkHandle and is used for identity/ENR. Keeping duplicate copies at the testnet layer served no functional purpose. All constructors/usages were updated accordingly. This change is non-breaking: the removed fields were private, and public method signatures remain unchanged.


mattsse (commented):
doesnt hurt to keep this

PR #19095, opened 2025-10-16
feat: convert pooled blobs transition

klkvr:
Based on #19084

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -504,6 +504,7 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
                     let spawner = task_spawner.clone();
                     let client = client.clone();
                     task_spawner.spawn(Box::pin(async move {
+                        let mut interval = tokio::time::interval(Duration::from_secs(1));
                         loop {
                             // Loop and replace blob transactions until we reach Osaka transition
                             // block after which no legacy blobs are going to be accepted.
@@ -519,7 +520,7 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
                                 if !tx.transaction.is_eip4844() {
                                     continue;
                                 }
-                                let tx_hash = *tx.transaction.tx_hash();
+                                let tx_hash = *tx.transaction.hash();
 
                                 // Fetch sidecar from the pool
                                 let Ok(Some(sidecar)) = pool.get_blob(tx_hash) else {
@@ -546,9 +547,7 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
                                 let pool = pool.clone();
                                 spawner.spawn(Box::pin(async move {
                                     // Convert sidecar to EIP-7594 format
-                                    let Some(sidecar) =
-                                        converter.convert(Arc::unwrap_or_clone(sidecar)).await
-                                    else {
+                                    let Some(sidecar) = converter.convert(sidecar).await else {
                                         return;
                                     };
 
@@ -567,16 +566,14 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
                             if last_iteration {
                                 break;
                             }
+
+                            interval.tick().await;
                         }
                     }));
                 }
             }
         }
     }
-}
-
-async fn convert_blob_sidecars<Pool, Tasks>(pool: P, spawner: Tasks) {
-    let interval = tokio::time::interval(Duration::from_secs(1));
 }
 
 struct FinalizedBlockTracker {
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -496,14 +496,20 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
                 // keep track of mined blob transactions
                 blob_store_tracker.add_new_chain_blocks(&blocks);
 
-                // At Osaka transition, we need to convert blobs to new format
+                // If Osaka activates in 2 slots we need to convert blobs to new format.
                 if !chain_spec.is_osaka_active_at_timestamp(tip.timestamp()) &&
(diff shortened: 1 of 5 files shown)


crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
                 blob_store_tracker.add_new_chain_blocks(&blocks);
+
+                // At Osaka transition, we need to convert blobs to new format
+                if !chain_spec.is_osaka_active_at_timestamp(tip.timestamp()) &&
+                    chain_spec.is_osaka_active_at_timestamp(tip.timestamp().saturating_add(12))
+                {
+                    let pool = pool.clone();
+                    let spawner = task_spawner.clone();
+                    let client = client.clone();
+                    task_spawner.spawn(Box::pin(async move {
```
mattsse, line 507: ~~could we perhaps make this a standalone fn, feel like this would be easier to delete later, but maybe doesnt really matter~~

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
 
                 // keep track of mined blob transactions
                 blob_store_tracker.add_new_chain_blocks(&blocks);
+
+                // At Osaka transition, we need to convert blobs to new format
+                if !chain_spec.is_osaka_active_at_timestamp(tip.timestamp()) &&
+                    chain_spec.is_osaka_active_at_timestamp(tip.timestamp().saturating_add(12))
```
mattsse: it's technically possible that we reach this twice if we go through a reorg but imo we can ignore this case

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
+                                let Ok(Some(sidecar)) = pool.get_blob(tx_hash) else {
+                                    continue;
+                                };
+                                // Ensure it is a legacy blob
+                                if !sidecar.is_eip4844() {
+                                    continue;
+                                }
+                                // Remove transaction and sidecar from the pool, both are in memory
+                                // now
+                                let Some(tx) = pool.remove_transactions(vec![tx_hash]).pop() else {
```
mattsse, line 542: let me add a new api fn for this

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
+                            let last_iteration =
+                                client.latest_header().ok().flatten().is_none_or(|header| {
+                                    client
+                                        .chain_spec()
+                                        .is_osaka_active_at_timestamp(header.timestamp())
+                                });
+
+                            let AllPoolTransactions { pending, queued } = pool.all_transactions();
+                            for tx in pending.into_iter().chain(queued) {
+                                if !tx.transaction.is_eip4844() {
```
mattsse: we can do a into_iter + filter here as well

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
+                        let mut interval = tokio::time::interval(Duration::from_secs(1));
+                        loop {
+                            // Loop and replace blob transactions until we reach Osaka transition
+                            // block after which no legacy blobs are going to be accepted.
+                            let last_iteration =
+                                client.latest_header().ok().flatten().is_none_or(|header| {
+                                    client
+                                        .chain_spec()
+                                        .is_osaka_active_at_timestamp(header.timestamp())
+                                });
```
mattsse, line 522: this exit condition assumes we're making it past the block, if that doesn't happen this loop will keep going, but since we have the timeout this is fine

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,82 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
+                                // Ensure it is a legacy blob
+                                if !sidecar.is_eip4844() {
+                                    continue;
+                                }
+                                // Remove transaction and sidecar from the pool, both are in memory
+                                // now
+                                let Some(tx) = pool.remove_transactions(vec![tx_hash]).pop() else {
+                                    continue;
+                                };
+                                pool.delete_blob(tx_hash);
```
mattsse, line 545: there's a potential race condition here if block building is still on going, that's why I think we should sleep for 2.5s before we start the loop by that time we can assume that there's already 1 block ready and we can be a bit more reckless here

mattsse (commented):
this is pretty straight forward actually

logic checks out only have one suggestion to sleep for a few seconds before we start doing this

crates/transaction-pool/src/maintain.rs
mattsse: I think we need to kick this of 2 slots (24s) in advance so that at `head == osaka - 12s` we have the new blobs available so that they can be included in the osaka activation block?

crates/transaction-pool/src/maintain.rs
```diff
@@ -494,6 +495,89 @@ pub async fn maintain_transaction_pool<N, Client, P, St, Tasks>(
...
+                    chain_spec.is_osaka_active_at_timestamp(tip.timestamp().saturating_add(24))
+                {
+                    let pool = pool.clone();
+                    let spawner = task_spawner.clone();
+                    let client = client.clone();
+                    task_spawner.spawn(Box::pin(async move {
+                        // Start converting not eaerlier than 4 seconds into current slot to ensure
+                        // that our pool only contains valid transactions for the next block (as
+                        // it's not Osaka yet).
+                        tokio::time::sleep(Duration::from_secs(4)).await;
```
mattsse, line 511: nice

mattsse (approved):
lgtm!

PR #19098, opened 2025-10-16
feat: add Pool::remove_transaction(hash)

### LINKED ISSUE #19097: Add Pool::remove_transaction(hash)
### Describe the feature

we have https://github.com/paradigmxyz/reth/blob/462de8e4c1686f4069db60533d29f566b24f1ce9/crates/transaction-pool/src/traits.rs#L432-L440

ideally we also have just remove_transaction(hash) -> Option

this can be impl as default like `remove_txs(vec![hash]).pop()`

@stevencartavia 

### Additional context

_No response_

stevencartavia:
closes #19097


mattsse (approved):
easy

cc @klkvr

PR #19099, opened 2025-10-16
fix: make functions const where possible (clippy::missing_const_for_fn)

avorylli:
while fixing clippys in another pr with my changes, found out some other clippy and fixed'em


mattsse (commented):
this changes the enum order which we should not do

PR #19102, opened 2025-10-17
refactor: naming fix for multiproof dispatch

yongkangc:
The methods now consistently use dispatch instead of queue, which better reflects that they're just sending messages to the workers rather than actually queuing work directly.


yongkangc (commented):
This PR should go in after https://github.com/paradigmxyz/reth/pull/18960

mediocregopher (approved):

PR #19103, opened 2025-10-17
fix(cli): Remove duplicit static file header and transaction append

RomanHodulak:
Fixes an import bug, in which this command fails.

```
reth init-state ./dump.json --datadir=test --chain ./chainspec.json --header=./header.rlp --header-hash=0x4f05e4392969fc82e41f6d6a8cea379323b0b2d3ddf7def1a33eec03883e3a33 --total-difficulty=0 --without-evm
```

By reporting

```
2025-10-16T15:42:50.687043Z  INFO Appending first valid block.
Error: trying to append data to Headers as block #1700 but expected block #1701
```

### DIFF
--- a/crates/cli/commands/Cargo.toml
+++ b/crates/cli/commands/Cargo.toml
@@ -99,6 +99,7 @@ proptest-arbitrary-interop = { workspace = true, optional = true }
 
 [dev-dependencies]
 reth-ethereum-cli.workspace = true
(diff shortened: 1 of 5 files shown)


mattsse (approved):
ah this makes sense

I believe this happened after we moved headers entire to static files

can we add an e2e test with those commands to ensure we keep this working?

mattsse (approved):

PR #19106, opened 2025-10-17
fix(cli): remove redundant EthChainSpec bound in run_with_components

sashass1315:
This change removes the unnecessary EthChainSpec constraint from CliApp::run_with_components. The EthChainSpec bound is already implied by N: CliNodeTypes via NodeTypes, so requiring it explicitly here was redundant and made the CliApp path stricter than the shared implementation and the public CLI API. Aligning the bounds with run_commands_with and Cli::run_with_components


mattsse (commented):
needs fmt

mattsse (approved):

PR #19109, opened 2025-10-17
fix: Simplify debug logging in TransactionFetcher by removing redundant format! calls

maximevtush:
Replace two format!("{:?}", new_announced_hashes) calls with direct {:?} arguments in tracing::debug! to avoid duplicated formatting and unnecessary allocations; leverage tracing’s native formatting.


mattsse (commented):
not worth it

PR #19112, opened 2025-10-17
fix: RethAccessList equality to enforce length

Galoretka:
This change fixes a bug in the test-only PartialEq<AccessList> implementation for RethAccessList where equality was determined via zipped iteration and could incorrectly return true for lists that share a common prefix but differ in length; now using Iterator::eq to ensure both content and length match, add a unit test that asserts inequality for prefix-only cases to guard against regressions


mattsse (commented):
this type is only used for tests

PR #19113, opened 2025-10-17
feat(rpc): improve eth_simulateV1 transaction validation

AJStonewee:
Implements transaction validation error codes (-38010 to -38014) for `eth_simulateV1` API to improve spec compliance.

Adds 5 new error variants to `EthSimulateError` as defined in the execution-apis spec.

### DIFF
--- a/crates/rpc/rpc-eth-types/src/simulate.rs
+++ b/crates/rpc/rpc-eth-types/src/simulate.rs
@@ -19,7 +19,7 @@ use reth_evm::{
     execute::{BlockBuilder, BlockBuilderOutcome, BlockExecutor},
     Evm,
 };
-use reth_primitives_traits::{BlockBody as _, BlockTy, NodePrimitives, Recovered, RecoveredBlock};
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-types/src/simulate.rs
```diff
@@ -33,6 +33,29 @@ use revm::{
...
+    FeeCapTooLow,
+    /// Not enough gas provided to pay for intrinsic gas.
+    #[error("intrinsic gas too low")]
+    GasTooLow,
+    /// Insufficient funds for gas fees and value.
+    #[error("insufficient funds for gas * price + value: address {address}")]
+    InsufficientFunds {
+        /// Address with insufficient funds.
+        address: Address,
+    },
```
mattsse, line 58: @klkvr I believe all of those are already covered via the EthApiError ?

klkvr: yep I believe so

Issue #18906, opened 2025-10-08
Remove consistentviewDB


jenpaff:
### Describe the feature

After #18460 and #18800 we can remove obsolete consistentviewDB

0xKarl98:
Can i take this ?

PR #19114, opened 2025-10-17
chore(trie): Add lifetime to cursors returned from Trie/HashedCursorFactorys

mediocregopher:
Towards #18906 

In the original implementation of OverlayStateProvider I assumed it would be possible to obtain a DBProvider with Clone by wrapping a DBProvider in an Arc. This turns out to not be true, because DBProvider is both `&mut self` and `self` methods.

Because of this it becomes necessary to refactor OverlayStateProvider such that the cursors it returns from its Trie/HashedCursorFactory are bound to its lifetime.

In order to accomplish this the factory traits themselves had to be modified such that the associated types now have lifetimes bound to `Self`. At a lower level our underlying cursor types are not bound to the txs which spawn them, but at an even lower level MDBX does not support using a cursor after its TX is committed/aborted, so this lifetime is actually correct anyway.

I've also added auto_impls for different references to the factorys/cursors that were missing them.

### DIFF
--- a/crates/trie/db/src/hashed_cursor.rs
+++ b/crates/trie/db/src/hashed_cursor.rs
@@ -20,18 +20,23 @@ impl<T> DatabaseHashedCursorFactory<T> {
 }
 
 impl<TX: DbTx> HashedCursorFactory for DatabaseHashedCursorFactory<&TX> {
-    type AccountCursor = DatabaseHashedAccountCursor<<TX as DbTx>::Cursor<tables::HashedAccounts>>;
-    type StorageCursor =
-        DatabaseHashedStorageCursor<<TX as DbTx>::DupCursor<tables::HashedStorages>>;
+    type AccountCursor<'a>
+        = DatabaseHashedAccountCursor<<TX as DbTx>::Cursor<tables::HashedAccounts>>
+    where
+        Self: 'a;
+    type StorageCursor<'a>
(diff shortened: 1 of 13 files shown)


Rjected (approved):
I see, is the idea here that rather than using the consistent view provider, we will be able to create multiple cursors, ie one per thread, which all have a ref to the same tx + in memory data?

crates/trie/db/src/hashed_cursor.rs
```diff
@@ -20,18 +20,23 @@ impl<T> DatabaseHashedCursorFactory<T> {
...
-    type AccountCursor = DatabaseHashedAccountCursor<<TX as DbTx>::Cursor<tables::HashedAccounts>>;
-    type StorageCursor =
-        DatabaseHashedStorageCursor<<TX as DbTx>::DupCursor<tables::HashedStorages>>;
-
-    fn hashed_account_cursor(&self) -> Result<Self::AccountCursor, DatabaseError> {
+    type AccountCursor<'a>
+        = DatabaseHashedAccountCursor<<TX as DbTx>::Cursor<tables::HashedAccounts>>
+    where
+        Self: 'a;
+    type StorageCursor<'a>
```
mattsse, line 27: I see

imo this makes sense

mattsse (approved):
makes sense

PR #19115, opened 2025-10-17
feat(stateless): make UncompressedPublicKey serializable

jsign:
After merging https://github.com/paradigmxyz/reth/pull/17841, it was noted that the new `UncompressedPublicKey`, which is an input to the stateless block validator function, wasn't bincode serializable.

zkVMs require the inputs to be serialized to be passed to the guest program (which in this case is the STF), same reason why `StatelessInput` is bincode serializable: https://github.com/paradigmxyz/reth/blob/a5618f57a8a8f80f13739cae640340dd3600b3c6/crates/stateless/src/lib.rs#L58-L70

This PR makes the new `UncompressedPublicKey` also bincode serializable, since arrays of length 65 aren't included in generic u8 array impl. (Open to other style if it sounds better!)

### DIFF
--- a/crates/stateless/src/recover_block.rs
+++ b/crates/stateless/src/recover_block.rs
@@ -2,15 +2,28 @@ use crate::validation::StatelessValidationError;
 use alloc::vec::Vec;
 use alloy_consensus::BlockHeader;
 use alloy_primitives::{Address, Signature, B256};
+use core::ops::Deref;
 use reth_chainspec::EthereumHardforks;
 use reth_ethereum_primitives::{Block, TransactionSigned};
 use reth_primitives_traits::{Block as _, RecoveredBlock};
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, Bytes};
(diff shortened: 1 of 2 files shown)


crates/stateless/src/recover_block.rs
```diff
@@ -2,15 +2,28 @@ use crate::validation::StatelessValidationError;
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UncompressedPublicKey(#[serde_as(as = "Bytes")] pub [u8; 65]);
+
+impl Deref for UncompressedPublicKey {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
jsign, line 26: Mostly to avoid `vk.0` in other places where it is used. Open for feedback if you prefer to avoid `Deref` impl.

mattsse (approved):
makes sense

failure unrelated

PR #19116, opened 2025-10-17
fix: Add support for init-state for op-reth chains that are not op-mainnet…

kahuang:
…, from a non-0 block.


mattsse (approved):
lgtm pending @RomanHodulak

RomanHodulak (approved):
I can see that the command now acts the same as for L1 with the exception of op-mainnet without OVM, in which case it imports the hardcoded bedrock genesis block.

The exception is consistent with the previous version, but the previous version would also skip `setup_without_evm` entirely for non op-mainnet, ignoring the header passed into the command.

PR #19117, opened 2025-10-18
feat: add capacity metrics for tries

Rjected:
This adds metrics for trie capacity, i.e., how many entries have we allocated for `nodes` and `values` maps respectively in trie data structures. This is so that we can understand how much space is being used by reused sparse tries in general.


yongkangc (approved):
Lgtm! 

One question would this just be looking at space allocated or what kind of insight around perf can we derive

mattsse (approved):
nice

PR #19119, opened 2025-10-18
fix: hold trace/debug semaphore permits correctly

reject-i:
The trace/debug RPC handlers tried to rate-limit themselves with a semaphore, but the permit was never actually held: we stored the Result from acquire_trace_permit() and immediately dropped it, so the OwnedSemaphorePermit was released right away and AcquireError was silently ignored. This change unwraps the Result, retains the permit for the lifetime of each RPC call, and converts AcquireError into a JSON-RPC internal error. With that, tracing endpoints finally respect the configured concurrency limit instead of running unbounded in parallel.


mattsse (commented):
we can ignore this case here, because the semaphore won't be closed

PR #19121, opened 2025-10-18
chore: remove redundant collect in debug trace

reallesee:
Remove the redundant `.into_iter().collect()` in `debug_trace_raw_block`; the first `collect::<Result<Vec<_>, _>>()?` already returns the required `Vec` of signer addresses, so looping again only re-allocates the same data. Dropping it keeps the tracing logic untouched while trimming unnecessary work.


mattsse (approved):
ty!

PR #19124, opened 2025-10-18
fix: accurate build features reporting in `reth --version`

### LINKED ISSUE #19123: --no-default-features build still reports jemalloc under “Build Features” (FreeBSD)
### Describe the bug

When building Reth on FreeBSD with --no-default-features, the resulting binary links only against libc.so.7 (the system jemalloc), but reth --version still shows Build Features: jemalloc.

This appears to be a false positive in the build metadata, not an actual linkage to the jemallocator crate.

### Steps to reproduce

git clone https://github.com/paradigmxyz/reth
cd reth
cargo clean
RUSTFLAGS="-C target-cpu=native -C link-arg=-lgcc -C link-arg=-static-libgcc" cargo build --release reth --no-default-features
ldd target/release/reth | grep malloc   # → nothing
ldd target/release/reth | grep libc     # → libc.so.7
./target/release/reth --version

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

1.8.2-dev
Commit SHA: a718752bf5703778993a9b660c0ea578da41a4bc

whatever comes with v1.8.2!

mainnet

Full via --full flag

none, default, full node

RUSTFLAGS="-C target-cpu=native -C link-arg=-lgcc -C link-arg=-static-libgcc" cargo build --release -p reth --no-default-features

### Code of Conduct

leniram159:
Replace `VERGEN_CARGO_FEATURES` with direct checks of `CARGO_FEATURE_*` environment variables. Accurate reporting of features actually enabled for the crate being built.

Fixes #19123

### DIFF
--- a/bin/reth/Cargo.toml
+++ b/bin/reth/Cargo.toml
@@ -28,7 +28,7 @@ workspace = true
 
 [dependencies]
 # reth
-reth-ethereum-cli = { workspace = true, default-features = false }
+reth-ethereum-cli.workspace = true
 reth-chainspec.workspace = true
 reth-primitives.workspace = true
 reth-db = { workspace = true, features = ["mdbx"] }
--- a/bin/reth/Cargo.toml
+++ b/bin/reth/Cargo.toml
@@ -28,7 +28,7 @@ workspace = true
 
 [dependencies]
 # reth
-reth-ethereum-cli.workspace = true
+reth-ethereum-cli = { workspace = true, default-features = false }
 reth-chainspec.workspace = true
 reth-primitives.workspace = true
 reth-db = { workspace = true, features = ["mdbx"] }
@@ -81,7 +81,9 @@ backon.workspace = true
 tempfile.workspace = true
 
 [features]
-default = ["jemalloc", "reth-revm/portable"]
+default = ["jemalloc", "otlp", "reth-revm/portable"]
+
+otlp = ["reth-ethereum-cli/otlp"]
 
 dev = ["reth-ethereum-cli/dev"]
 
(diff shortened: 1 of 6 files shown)


mattsse (changes requested):
instead of doing this

we need to remove these from the default features

https://github.com/paradigmxyz/reth/blob/a718752bf5703778993a9b660c0ea578da41a4bc/crates/ethereum/cli/Cargo.toml#L38-L38

and instead enable them properly in the binaries

crates/ethereum/cli/Cargo.toml
```diff
@@ -35,7 +35,7 @@ tracing.workspace = true
 tempfile.workspace = true
 
 [features]
-default = ["jemalloc", "otlp"]
```
mattsse, line 40: imo this makes sense 

because we make them configurable via the binary

mattsse (approved):
these lgtm

pending @Rjected or @shekhirin 

cc @fgimenez for vis

shekhirin:
@mattsse I don't think this works
```console
❯ cargo r --features min-error-logs --quiet -- --version
reth-ethereum-cli Version: 1.8.2
Commit SHA: eb9dfb41dd0c8afb8506bac968cb42d3e714d4d6
Build Timestamp: 2025-10-30T14:26:01.095541000Z
Build Features: jemalloc,otlp
Build Profile: debug
```

bin/reth/Cargo.toml
```diff
@@ -81,7 +81,9 @@ backon.workspace = true
 tempfile.workspace = true
 
 [features]
-default = ["jemalloc", "reth-revm/portable"]
+default = ["jemalloc", "otlp", "reth-revm/portable"]
+
+otlp = ["reth-ethereum-cli/otlp"]
```
mattsse, line 86: @shekhirin this is coming from here

shekhirin (approved):

PR #19125, opened 2025-10-18
fix(rpc): map existing transaction errors to eth_simulateV1 specific error codes

AJStonewee:
Currently `eth_simulateV1` returns generic -32602 for all transaction validation errors, but the spec requires specific codes like -38010 for nonce too low, -38014 for insufficient funds, etc. The error types already exist in `RpcInvalidTransactionError`, they just weren't mapped correctly in the simulateV1 context. 

Adds proper error code mapping by checking error messages and returning the right codes.

Inspired by #19113 - my bad for not realizing these errors were already in the codebase!


mattsse (commented):
we shouldnt do this like this

PR #19127, opened 2025-10-19
perf(storage): optimize wiped storage entry collection to avoid Vec allocation

### LINKED ISSUE #19077: perf(storage): optimize wiped storage entry collection to avoid unnecessary Vec allocation
### Problem
There's a [TODO comment at provider.rs:2279-2287](https://github.com/paradigmxyz/reth/blob/386eaa3ff68b5e05bfa64d76837a676baf3582cc/crates/storage/provider/src/providers/database/provider.rs#L2279-L2287) indicating that the collection of wiped storage entries could be rewritten more efficiently:

```rust
// TODO(mediocregopher): This could be rewritten in a way which doesn't require
// collecting wiped entries into a Vec like this, see
// `write_storage_trie_changesets`.
let mut wiped_storage = Vec::new();
if wiped {
    tracing::trace!(?address, "Wiping storage");
    if let Some((_, entry)) = storages_cursor.seek_exact(address)? {
        wiped_storage.push((entry.key, entry.value));
        while let Some(entry) = storages_cursor.next_dup_val()? {
            wiped_storage.push((entry.key, entry.value))
        }
    }
}
```

Currently, the code:
1. Allocates a Vec to collect all wiped storage entries
2. Iterates through the cursor and pushes each entry into the Vec
3. This intermediate collection adds allocation overhead and an extra iteration

### Proposed Optimization
Refactor to use the approach from `write_storage_trie_changesets` which likely:
- Processes entries directly from the cursor without intermediate collection
- Avoids unnecessary Vec allocations
- Reduces memory pressure during block persistence

### Context
This optimization is part of the broader [static file write optimization effort (#19057)](https://github.com/paradigmxyz/reth/issues/19057) to reduce block persistence latency and ensure write throughput matches block production rates.

WilliamNwoke:
Resolves #19077

Refactored the implementation to use a streaming iterator approach that processes wiped storage entries directly from the database cursor without intermediate collection. Key changes:

Added `StorageWipedEntriesIter`: A new iterator that yields wiped storage entries on-demand from the database cursor

Integrated with `StorageRevertsIter`: Combined with existing reverts processing for efficient merging

### DIFF
--- a/crates/storage/provider/src/changesets_utils/state_reverts.rs
+++ b/crates/storage/provider/src/changesets_utils/state_reverts.rs
@@ -1,4 +1,9 @@
-use alloy_primitives::{B256, U256};
+use alloy_primitives::{Address, B256, U256};
+use reth_db_api::{
+    cursor::{DbCursorRO, DbDupCursorRO},
+    tables,
+};
+use reth_storage_errors::provider::ProviderResult;
 use revm_database::states::RevertToSlot;
 use std::iter::Peekable;
 
@@ -24,8 +29,8 @@ where
     }
 
     /// Consume next revert and return it.
-    fn next_revert(&mut self) -> Option<(B256, U256)> {
-        self.reverts.next().map(|(key, revert)| (key, revert.to_previous_value()))
+    fn next_revert(&mut self) -> Option<(B256, RevertToSlot)> {
+        self.reverts.next()
     }
 
     /// Consume next wiped storage and return it.
@@ -39,7 +44,7 @@ where
     R: Iterator<Item = (B256, RevertToSlot)>,
     W: Iterator<Item = (B256, U256)>,
 {
-    type Item = (B256, U256);
+    type Item = (B256, RevertToSlot);
 
     /// Iterate over storage reverts and wiped entries and return items in the sorted order.
     /// NOTE: The implementation assumes that inner iterators are already sorted.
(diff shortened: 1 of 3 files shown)


WilliamNwoke:
Please review @yongkangc @mattsse

crates/storage/provider/src/changesets_utils/wiped_storage.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl<'a, C> WipedStorageIter<'a, C>
+where
+    C: DbCursorRO<PlainStorageState> + DbDupCursorRO<PlainStorageState>,
+{
+    /// Create a new iterator for wiped storage entries at the given address.
+    pub fn new(
+        cursor: &'a mut C,
+        address: alloy_primitives::Address,
```
yongkangc: could we do it in a simpler way where we don't need this abstraction? 

this abstraction seems to make it more complicated

yongkangc (changes requested):
Think we could simplify this without the abstraction


ref:
https://github.com/paradigmxyz/reth/blob/2f9281b6c10674525b3bc7820898b770258d2f67/crates/storage/provider/src/providers/database/provider.rs#L2383 cc @mediocregopher would be great to get ur thoughts as well :)

crates/storage/provider/src/changesets_utils/wiped_storage.rs
WilliamNwoke: Simplified and reverted!

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1689,22 +1732,34 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
                 // See [StorageWipe::Primary] for more details.
                 //
-                // TODO(mediocregopher): This could be rewritten in a way which doesn't require
-                // collecting wiped entries into a Vec like this, see
-                // `write_storage_trie_changesets`.
-                let mut wiped_storage = Vec::new();
-                if wiped {
+                // Note: This code performs an intermediate Vec allocation to collect storage slot
+                // changes before sorting.
+                let wiped_storage_iter: Box<dyn Iterator<Item = (B256, U256)>> = if wiped {
```
mediocregopher: I think this could be simplified even further to remove the Box/dyn. It should be possible to do something like:

```
let wiped_storage_iter = (if wiped { Some(storages_cursor.walk_dup(...)? } else { None }).into_iter()
```

We also want to be sure to not be dropping errors. This will likely mean modifying the `StorageRevertsIter` to accept an iterator over Results.

mediocregopher (changes requested):
He @WilliamNwoke , sorry it took so long to look at this. I think the code could be simplified a little further, left a comment about that 🙏

crates/storage/provider/src/providers/database/provider.rs
WilliamNwoke: Alright!

WilliamNwoke:
Please review @mediocregopher @yongkangc

crates/storage/provider/src/changesets_utils/state_reverts.rs
```diff
@@ -5,15 +5,19 @@ use std::iter::Peekable;
...
+{
     reverts: Peekable<R>,
     wiped: Peekable<W>,
 }
 
 impl<R, W> StorageRevertsIter<R, W>
 where
     R: Iterator<Item = (B256, RevertToSlot)>,
-    W: Iterator<Item = (B256, U256)>,
+    W: Iterator<Item = Result<(B256, U256), reth_storage_errors::provider::ProviderError>>,
```
mediocregopher: nit: let's import this at the top of the file. You can use ProviderResult as well for even more brevity.

crates/storage/provider/src/changesets_utils/state_reverts.rs
```diff
@@ -24,54 +28,61 @@ where
...
+                                Some(Ok((key, value)))
+                            }
+                            Err(e) => Some(Err(e)),
+                        }
                     }
                 }
             }
-            (Some(_revert), None) => self.next_revert(),
-            (None, Some(_wiped)) => self.next_wiped(),
+            (Some(_), Some(Err(_))) => self.next_wiped(),
```
mediocregopher: Let's not ignore the error, return it from here instead

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1689,22 +1732,35 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
                 //
-                // TODO(mediocregopher): This could be rewritten in a way which doesn't require
-                // collecting wiped entries into a Vec like this, see
-                // `write_storage_trie_changesets`.
-                let mut wiped_storage = Vec::new();
-                if wiped {
+                // Note: This code performs an intermediate Vec allocation to collect storage slot
+
+                // changes before sorting.
+                let wiped_storage_iter: Vec<
```
mediocregopher: So this is still allocating a Vec? The purpose of this change was to not do this. I would recommend reviewing the implementation of [write_storage_trie_changesets](https://github.com/paradigmxyz/reth/blob/dd25caec12e8401a72d683bcefc0307861dd90ef/crates/storage/provider/src/providers/database/provider.rs#L2345) to understand how this can be done

mediocregopher (changes requested):

PR #19128, opened 2025-10-19
refactor(cli): rename Entries::new_with_raw_values to Entries::new

sashaodessa:
## Description:

Renames `Entries::new_with_raw_values` to `Entries::new` in the TUI module to eliminate redundant naming and improve code clarity.

## Changes

- Renamed `Entries::new_with_raw_values` to `Entries::new`
- Updated constructor call in `DbListTUI::new`

## Rationale

The original method name was redundant since the boolean parameter `raw` already indicates the nature of the values being created. The simplified name makes the API more intuitive and reduces cognitive load.

### DIFF
--- a/crates/cli/commands/src/db/tui.rs
+++ b/crates/cli/commands/src/db/tui.rs
@@ -47,10 +47,10 @@ enum Entries<T: Table> {
 }
 
 impl<T: Table> Entries<T> {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we can keep just raw here

PR #19133, opened 2025-10-19
fix(engine): Eliminates spurious warning logs in prewarm task

### LINKED ISSUE #19101: Fix warning logs for prewarm
### Describe the feature

There's alot of these warning logs. 

ts=2025-10-16T06:42:31.387075846Z level=warn target=engine::tree::prewarm message="Failed to send transaction to worker" task_idx=10 error="sending on a closed channel"

My best guess for the warns is that some prewarm task is finishing after main execution has completed and everything is cleaned up

### Additional context

_No response_

0x00101010:
Closes https://github.com/paradigmxyz/reth/issues/19101

## Changes made

1. Fixes spurious warning logs issue
2. Reworked tx distribution code to improve readability

## Problem

The prewarm cache task was logging frequent warnings:
level=warn target=engine::tree::prewarm message="Failed to send transaction to worker"
task_idx=10 error="sending on a closed channel"

These warnings occur when the main block execution completes before the prewarm distribution thread finishes
sending all transactions to workers. This is expected behavior: workers check the `terminate_execution` flag and
exit early as an optimization when they're no longer needed.

## Root Cause

When main execution completes, it sends `TerminateTransactionExecution` to set the `terminate_execution` atomic
flag. Workers polling this flag exit their receive loops, dropping their channel receivers. The distribution
thread continues draining the `pending` channel and attempts to send to workers that have already exited,
resulting in `SendError`.

This race condition is by design - it allows workers to exit immediately when signaled rather than waiting for
all transactions to be distributed.

## Solution

**Two-part fix:**

1. **Add early termination check** in the distribution loop (line 172) - stops distributing transactions
immediately when `terminate_execution` is set

3. **Remove warning logs** for `SendError` (lines 190-194, 199-203) - replaced with comments explaining that
workers listen to `terminate_execution` and may exit early when signaled

ref #18892

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -167,59 +167,43 @@ where
 
             let mut tx_index = 0usize;
 
-            // Handle first transaction - special case for system transactions
-            if let Ok(first_tx) = pending.recv() {
-                // Move the transaction into the indexed wrapper to avoid an extra clone
-                let indexed_tx = IndexedTransaction { index: tx_index, tx: first_tx };
-                // Compute metadata from the moved value
-                let tx_ref = indexed_tx.tx.tx();
-                let is_system_tx = tx_ref.ty() > MAX_STANDARD_TX_TYPE;
-                let first_tx_hash = tx_ref.tx_hash();
+            // Distribute transactions to workers
+            for tx in core::iter::from_fn(|| pending.recv().ok()) {
+                // Stop distributing if termination was requested
+                if ctx.terminate_execution.load(Ordering::Relaxed) {
+                    trace!(
+                        target: "engine::tree::payload_processor::prewarm",
+                        "Termination requested, stopping transaction distribution"
+                    );
+                    break;
+                }
 
-                // Check if this is a system transaction (type > 4)
-                // System transactions in the first position typically set critical metadata
-                // that affects all subsequent transactions (e.g., L1 block info, fees on L2s).
-                if is_system_tx {
-                    // Broadcast system transaction to all workers to ensure they have the
-                    // critical state. This is particularly important for L2s like Optimism
-                    // where the first deposit transaction contains essential block metadata.
+                let indexed_tx = IndexedTransaction { index: tx_index, tx };
+                let is_system_tx = indexed_tx.tx.tx().ty() > MAX_STANDARD_TX_TYPE;
+
+                // System transactions (type > 4) in the first position set critical metadata
+                // that affects all subsequent transactions (e.g., L1 block info on L2s).
+                // Broadcast the first system transaction to all workers to ensure they have
+                // the critical state. This is particularly important for L2s like Optimism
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -165,61 +165,44 @@ where
...
             let mut tx_index = 0usize;
+            while let Ok(tx) = pending.recv() {
+                // Stop distributing if termination was requested
+                if ctx.terminate_execution.load(Ordering::Relaxed) {
+                    trace!(
+                        target: "engine::tree::payload_processor::prewarm",
+                        "Termination requested, stopping transaction distribution"
+                    );
+                    break;
+                }
```
mattsse, line 177: ah, I see

that could indeed result in warnings like

> message="Failed to send transaction to worker"

mattsse: this check is pretty cheap so we can keep this, perhaps this is reachable for blocks with 1 txs or not at all but doesnt really hurt

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -165,61 +165,44 @@ where
...
-                if is_system_tx {
-                    // Broadcast system transaction to all workers to ensure they have the
-                    // critical state. This is particularly important for L2s like Optimism
-                    // where the first deposit transaction contains essential block metadata.
+                // System transactions (type > 4) in the first position set critical metadata
+                // that affects all subsequent transactions (e.g., L1 block info on L2s).
+                // Broadcast the first system transaction to all workers to ensure they have
+                // the critical state. This is particularly important for L2s like Optimism
+                // where the first deposit transaction (type 126) contains essential block metadata.
+                if tx_index == 0 && is_system_tx {
```
mattsse, line 187: this is a lot cleaner than doing the loop twice and only costs us 2 checks that are probably predicted out

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -165,61 +165,44 @@ where
...
-            while let Ok(executable) = pending.recv() {
-                let indexed_tx = IndexedTransaction { index: tx_index, tx: executable };
-                let task_idx = executing % workers_needed;
-                if let Err(err) = handles[task_idx].send(indexed_tx) {
-                    warn!(
-                        target: "engine::tree::payload_processor::prewarm",
-                        task_idx,
-                        error = %err,
-                        "Failed to send transaction to worker"
-                    );
```
mattsse, line 221: @yongkangc this was potentially reachable if this was trailing the actual execution, which could happen
and all prewarming tasks already checked if it was terminated

mattsse (approved):
cool, this makes sense, tysm!

pending @yongkangc 


I also removed the duplicated counter https://github.com/paradigmxyz/reth/pull/19133/commits/fc8dc4611bdf35003c34581dbc7b87eeb24b24b5

yongkangc (approved):
lgtm here

PR #19135, opened 2025-10-20
refactor: improve error messages for Option unwrapping in critical paths

sky-coderay:
Fixed unhelpful panic messages by replacing .expect().unwrap() chains with descriptive .expect() calls.

Before:
```rust
let range = input.get_next_tx_num_range(&provider).expect("Expected range").unwrap();
```
Test result: `called Option::unwrap() on a None value`

After:
```rust
let range = input.get_next_tx_num_range(&provider).expect("Expected range").expect("Range should not be None");
```
Test result: `Range should not be None`

The difference is huge when debugging production issues - you immediately know WHICH Option was None instead of hunting through dozens of unwrap() calls.

 Instead of changing the project's unwrap() style entirely, this focuses on cases where we already have expect() but then immediately unwrap() anyway.

These particular locations in prune segments and flashblocks stream handling are critical for debugging when storage operations fail.

### DIFF
--- a/crates/optimism/flashblocks/src/ws/stream.rs
+++ b/crates/optimism/flashblocks/src/ws/stream.rs
@@ -496,8 +496,11 @@ mod tests {
         let mut stream = WsFlashBlockStream::with_connector(ws_url, connector);
 
         let expected_message = flashblock;
-        let actual_message =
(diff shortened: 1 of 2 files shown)


mattsse (commented):
this is just test code

PR #19136, opened 2025-10-20
fix: remove redundant RethRpcModule::variant_count() and use all_variants().len()

anim001k:
### Description
- Replace all call sites of variant_count() with all_variants().len() to eliminate duplicate API and a second source of truth for module count.
- Remove variant_count() from RethRpcModule; the single source is now STANDARD_VARIANTS via all_variants().

- Changes:
  - crates/rpc/rpc-server-types/src/module.rs: inline variant count via all_variants().len(), delete variant_count(), update related tests.


mattsse (commented):
this pub fn exists because then this doesn feature doesnt need a trait import

PR #19138, opened 2025-10-20
fix: Inline value match in SparseTrie::find_leaf to remove redundant wrapper

anim001k:
### Description
- Replace local helper function `check_value_match` with inlined comparison in `SparseTrie::find_leaf`.
- Avoids a redundant wrapper and defers cloning `Vec<u8>` until error construction.

- Rationale:
  - Reduces cognitive load and eliminates local duplication without changing behavior.
  - Keeps clones only on error path; hot path unaffected.


crates/trie/sparse/src/trie.rs
```diff
@@ -969,24 +969,6 @@ impl SparseTrieInterface for SerialSparseTrie {
         full_path: &Nibbles,
         expected_value: Option<&Vec<u8>>,
     ) -> Result<LeafLookup, LeafLookupError> {
-        // Helper function to check if a value matches the expected value
-        fn check_value_match(
```
shekhirin, line 977: can we just add an `#[inline]` attribute here? It doesn't clone the value unless the error is returned anyway, as @mediocregopher said.

shekhirin (changes requested):

crates/trie/sparse/src/trie.rs
```diff
@@ -964,29 +964,12 @@ impl SparseTrieInterface for SerialSparseTrie {
         self.rlp_buf.clear();
     }
 
+    #[inline]
```
shekhirin: sorry, I meant adding an `#[inline]` attribute on `check_value_match`, and that's basically the only change that's needed in this PR, because we already don't clone anything until we return the error

shekhirin (changes requested):

shekhirin (approved):

PR #19139, opened 2025-10-20
feat: update fork id in discv5[WIP]

cuiweixie:
remove todo


mattsse (changes requested):
this makes sense, but is this even correct here:

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/net/discv5/src/lib.rs#L109-L109

looks like this does double rlp encode

should

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/net/discv5/src/lib.rs#L109-L109

be `insert_raw_rlp` instead?

can we add a simple roundtrip test for this fn?

ref discv4

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/net/discv4/src/lib.rs#L1756-L1759

mattsse (approved):
actually, this is correct

PR #19141, opened 2025-10-20
fix(prune): Disable pruning limits

mediocregopher:
This PR disables the limits imposed in the Engine API on the Pruner. With the new Accounts/StoragesChangeSets tables the pruner was no longer able to complete all work within the normal limits, so we would have had to raise them. In reality these limits are not really necessary anymore, as pruning happens asynchronously to block processing and so should always be let run to completion.


mattsse (approved):
makes sense

jenpaff (approved):

PR #19144, opened 2025-10-20
chore(config): clean up gas limit code

0xMushow:
smol cleanup


mattsse (approved):
these will change again eventually but it isn't a big change to reintroduce separate arms again when that happens

PR #19146, opened 2025-10-20
feat(e2e): add builder API for configuring test node setups

fgimenez:
introduces a builder API that accepts closures to modify `TreeConfig` and `NodeConfig`, enabling configuration of settings like `persistence_threshold`. 

refactors `reth-e2e-test-utils` setup functions to use the builder internally, maintaining backward compatibility

### DIFF
--- a/crates/e2e-test-utils/src/setup_builder.rs
+++ b/crates/e2e-test-utils/src/setup_builder.rs
@@ -42,8 +42,6 @@ where
 {
     num_nodes: usize,
     chain_spec: Arc<N::ChainSpec>,
-    is_dev: bool,
-    tree_config: reth_node_api::TreeConfig,
     attributes_generator: F,
     connect_nodes: bool,
     tree_config_modifier: Option<TreeConfigModifier>,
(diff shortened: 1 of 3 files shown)


crates/e2e-test-utils/src/setup_builder.rs
```diff
@@ -0,0 +1,222 @@
...
+        PayloadAttributesBuilder<<N::Payload as PayloadTypes>::PayloadAttributes>,
+{
+    num_nodes: usize,
+    chain_spec: Arc<N::ChainSpec>,
+    is_dev: bool,
+    tree_config: reth_node_api::TreeConfig,
+    attributes_generator: F,
+    connect_nodes: bool,
+    tree_config_modifier: Option<TreeConfigModifier>,
+    node_config_modifier: Option<NodeConfigModifier<N::ChainSpec>>,
```
mattsse, line 48: I think this is a good step in the right direction, if needed we can iterate on this a bit more

crates/e2e-test-utils/src/setup_builder.rs
```diff
@@ -0,0 +1,222 @@
...
+        + 'static,
+    LocalPayloadAttributesBuilder<N::ChainSpec>:
+        PayloadAttributesBuilder<<N::Payload as PayloadTypes>::PayloadAttributes>,
+{
+    /// Creates a new builder with the required parameters.
+    pub fn new(
+        num_nodes: usize,
+        chain_spec: Arc<N::ChainSpec>,
+        is_dev: bool,
+        tree_config: reth_node_api::TreeConfig,
```
mattsse: these could now technically be replaced with the builder closure entirely

should we do this in this pr already?

mattsse (approved):
lgtm, one idea

pending @klkvr

klkvr (approved):

crates/e2e-test-utils/src/setup_builder.rs
fgimenez: sure good point, done here https://github.com/paradigmxyz/reth/pull/19146/commits/fd264038dc0bba0677fe9f29f22caf39fc21ca3c

PR #19149, opened 2025-10-20
feat(optimism): add interop revalidation attempt metrics and per-tx tracking

GarmashAlex:
Implement aggregated metrics for interop supervisor revalidation attempts and track attempts per transaction in-memory to fulfill the TODO without introducing high-cardinality labels. Each supervisor check increments a global counter, and when a transaction is removed due to invalidation or interop reclassification, the number of attempts is recorded in a histogram. The number of active tracked entries is exposed as a gauge. This provides visibility into revalidation pressure and behavior while aligning with project metric best practices and avoiding per-hash labels.


mattsse (commented):
dont think we need this really

PR #19150, opened 2025-10-20
fix(net): correct variable in test_peer_rotation assertion

reject-i:
Fix copy-paste bug in test_peer_rotation where second_peer assertion was incorrectly checking first_peer variable instead of second_peer. This test validates peer switching logic when one peer becomes unavailable, and the incorrect assertion could have masked real bugs in the peer selection algorithm.

https://github.com/paradigmxyz/reth/blob/main/crates/net/network/src/fetch/mod.rs#L518


mattsse (commented):
this is fine

PR #19151, opened 2025-10-20
chore: remove total difficulty from `HeaderProvider`

### LINKED ISSUE #16586: drop support for total difficulty table entirely
### Describe the feature

we stopped using total difficulty a while ago and require that all difficulty activated hardforks (merge) are known and completed.

we currently still maintain the total difficulty table, which is effectively unused now.
so we can drop support for:

https://github.com/paradigmxyz/reth/blob/6c05d595cffd91f97dc3e7c2819d7b874ac5ca85/crates/storage/db-api/src/tables/mod.rs#L309-L313

and all read/writes/unwinds etc.
we should entirely stop using this.

however there could be some edge cases when a user downgrades, so this changes must be done with this in mind

### Additional context

_No response_

joshieDo:
* removes total difficulty provider methods from `StorageApi`
* adds `era_utils::calculate_td_by_number` so we are able to generate era files with the same checksums. slightly expensive, since it has to iterate through all headers up to num.
* `static_file_provider.append_header` now always sets TTD column to zero. This has the side-effect that static files will have a different checksum after this PR. 

closes #16586

### DIFF
--- a/crates/node/core/src/node_config.rs
+++ b/crates/node/core/src/node_config.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 use alloy_consensus::BlockHeader;
 use alloy_eips::BlockHashOrNumber;
-use alloy_primitives::{BlockNumber, B256};
+use alloy_primitives::{BlockNumber, B256, U256};
 use eyre::eyre;
 use reth_chainspec::{ChainSpec, EthChainSpec, MAINNET};
 use reth_config::config::PruneConfig;
--- a/crates/node/core/src/node_config.rs
+++ b/crates/node/core/src/node_config.rs
@@ -330,12 +330,6 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
             .header_by_number(head)?
             .expect("the header for the latest block is missing, database is corrupt");
 
-        let total_difficulty = provider
-            .header_td_by_number(head)?
-            // total difficulty is effectively deprecated, but still required in some places, e.g.
-            // p2p
-            .unwrap_or_default();
-
         let hash = provider
             .block_hash(head)?
             .expect("the hash for the latest block is missing, database is corrupt");
@@ -344,7 +338,7 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
             number: head,
             hash,
             difficulty: header.difficulty(),
-            total_difficulty,
+            total_difficulty: U256::ZERO,
             timestamp: header.timestamp(),
         })
     }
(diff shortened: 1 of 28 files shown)


crates/node/core/src/node_config.rs
```diff
@@ -360,7 +338,7 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
             number: head,
             hash,
             difficulty: header.difficulty(),
-            total_difficulty,
+            total_difficulty: U256::ZERO,
```
joshieDo, line 341: im guessing so, but is this okay?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1312,6 +1311,26 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
     pub fn tx_index(&self) -> &RwLock<SegmentRanges> {
         &self.static_files_tx_index
     }
+
+    /// Fetches the total difficulty for a given block number from static files.
+    ///
+    /// Returns `None` if the block is not found or if the static file is missing.
+    pub fn header_td_by_number(&self, num: BlockNumber) -> ProviderResult<Option<U256>> {
```
joshieDo: if we want to be able to still generate era files that have the same checksum, then we do need to keep track total difficulty in some way. Same applies for our static files.

Since static files already have a column for total difficulty (and we can't remove it), i think it's okay to keep this method part of `StaticFileProvider`?

joshieDo: nvm, recalculating when necessary

crates/stages/stages/src/stages/headers.rs
```diff
@@ -134,11 +128,8 @@ where
             }
             last_header_number = header.number();
 
-            // Increase total difficulty
-            td += header.difficulty();
-
             // Append to Headers segment
-            writer.append_header(header, td, header_hash)?;
+            writer.append_header(header, U256::ZERO, header_hash)?;
```
Rjected: do we ever call this with something nonzero after this PR? if not then maybe we should just remove the arg?

crates/node/core/src/node_config.rs
Rjected: I think this should be fine, in p2p this is deprecated as of `eth/69` and the node builder only uses the `Head` to get the timestamp

Rjected (commented):
just a q about simplifying `append_header` args otherwise lgtm

crates/stages/stages/src/stages/headers.rs
mattsse: yeah I think we should remove this from the function arg entirely

mattsse (changes requested):
very nice, I only have the same pedantic suggestion as @Rjected 

imo it would be fine to always use zero for the export as well but we can keep this

mattsse (approved):
lgtm!

PR #19152, opened 2025-10-20
feat(storage): replace unreachable todo!() with explicit unreachable!() in compact derive

MozirDmitriy:
- The todo!() branch inside StructHandler::from for is_flag_type(ftype) && !is_compact is unreachable because is_compact is always set to true for flag-types in load_field_from_segments (is_flag_type(ftype) || #[maybe_zero]).
- codify the invariant by using unreachable!("flag-type fields are always compact in Compact derive").


mattsse (approved):
makes sense

pending @joshieDo

joshieDo (approved):

PR #19153, opened 2025-10-20
fix: clarify helper name for receipt offsets 

maradini77:
### Description
- **Summary**: Rename a misleading helper to accurately describe its behavior and reduce cognitive load.
- **Rationale**: The previous name suggested computing per-tx gas used, but the function returns cumulative gas and log index offsets before the target tx. Clear naming helps prevent misuse.
- **Changes**:
  - Rename `calculate_gas_used_and_next_log_index` to `compute_offsets_before_tx`.


mattsse (commented):
I dont want to rename this

PR #19155, opened 2025-10-20
feat(engine): improve payload validator tracing spans 2

Rjected:
This does a few main things over https://github.com/paradigmxyz/reth/pull/18960:

* First, the default level for otlp traces are set to `debug`. This is mainly to avoid a frequently called `trace` level span in `HashBuilder`: https://github.com/alloy-rs/trie/blob/3401e6e8a8a44594bca7694dccd44a0f2130f92a/src/hash_builder/mod.rs#L234

Screenshot 2025-10-20 at 2 07 16 PM

* `TrieNodeIter`and `TrieWalker` spans are downgraded to trace, these were too frequent and hurt performance

Screenshot 2025-10-20 at 12 32 30 PM

* Downgrades ecies `instrument` calls to `trace` (I just prefer these be opt-in even in otlp)

* Moves `ChainOrchestrator::poll` instrument call `name` to `target`.

* Make parallel sparse trie traces and spans consistent

* Add account and storage proof task spans linked to workers:

Screenshot 2025-10-20 at 4 27 11 PM
Screenshot 2025-10-20 at 4 27 27 PM

* Lower multiproof task traces from `debug` to `trace`: [`698227f` (#19155)](https://github.com/paradigmxyz/reth/pull/19155/commits/698227f5e14459b0b093ab1cd7d6288e90d796c4) - these were causing lots of log volume

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -329,7 +329,7 @@ where
         skip_all,
         fields(
             parent = ?input.parent_hash(),
-            block_num_hash = ?input.num_hash()
+            type_name = ?input.type_name(),
         )
     )]
     pub fn validate_block_with_state<T: PayloadTypes<BuiltPayload: BuiltPayload<Primitives = N>>>(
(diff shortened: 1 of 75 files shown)


shekhirin (approved):

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -324,6 +323,15 @@ where
     /// - Block execution
     /// - State root computation
     /// - Fork detection
+    #[instrument(
+        level = "debug",
+        target = "engine::tree::payload_validator",
+        skip_all,
+        fields(
+            parent = ?input.parent_hash(),
+            type_name = ?input.type_name(),
```
shekhirin, line 332: don't we want to have the block hash and number here too?

Rjected: removed because it was redundant, block hash num is also present in `on_new_payload`

PR #19156, opened 2025-10-20
chore: grammar tweak in comment

mdqst:
a grammatical mistake in the comment `/// Error gather proofs.` - it should be `/// Error gathering proofs.` to use the correct gerund form after “Error.”


mattsse (approved):
sure

PR #19157, opened 2025-10-20
chore: remove deprecated code

AJStonewee:
this part is not using in anywhere, I think we should remove this


mattsse (commented):
we can keep this for a bit longer

PR #19158, opened 2025-10-20
feat(net): mplement disconnect handling in ProtocolProxy via control channel

radik878:
Implemented CanDisconnect for ProtocolProxy by adding a dedicated disconnect control channel to signal the underlying P2PStream.
Hooked the disconnect signal into both the handshake loop and RlpxSatelliteStream::poll_next, invoking start_disconnect on the connection.
.


mattsse (commented):
I think not having this is fine because if the stream is terminated this would trigger a disconnect anyway

radik878:
> I think not having this is fine because if the stream is terminated this would trigger a disconnect anyway

could i remove todo then or leave it as is is fine?

PR #19159, opened 2025-10-21
fix(engine): shrink tries after clearing

Rjected:
This adds `shrink_nodes_to` and `shrink_values_to` methods to all sparse trie implementations. This allows us to manage memory after clearing the sparse trie, so that the sparse tries do not have unbounded memory growth.

This was previously possible because there were roughly 600 active storage tries, and 600 cleared storage tries. For example, we may fill a storage trie with many nodes because it corresponds to a large contract like XEN. We would then clear this trie, and keep it around for reuse.

It's unlikely that the allocated space would be effectively reused, because cleared tries that we reuse are picked indiscriminately. So this allocated space would stick around in a cleared or active trie, while we then allocate even more space in other storage tries, for other large contracts.

Now, we shrink allocated space in tries before saving the trie using these `shrink_to` methods.

## Memory Usage Impact

Previously there was a striking similarity between the capacity, and the allocated space, for example here is the memory usage after 3000 blocks:
Screenshot 2025-10-22 at 7 24 45 PM

And here is the node capacity metric:
Screenshot 2025-10-22 at 7 24 56 PM

Here is the memory usage now:
Screenshot 2025-10-22 at 7 04 36 PM

And the node capacity metric:
Screenshot 2025-10-22 at 7 05 07 PM

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -66,10 +66,27 @@ use configured_sparse_trie::ConfiguredSparseTrie;
 pub const PARALLEL_SPARSE_TRIE_PARALLELISM_THRESHOLDS: ParallelismThresholds =
     ParallelismThresholds { min_revealed_nodes: 100, min_updated_nodes: 100 };
 
-/// Default node capacity for shrinking the sparse trie.
+/// Default node capacity for shrinking the sparse trie. This is used to limit the number of trie
+/// nodes in allocated sparse tries.
+///
+/// Node maps have a key of `Nibbles` and value of `SparseNode`.
+/// The `size_of::<Nibbles>` is 40, and `size_of::<SparseNode>` is 80.
+///
+/// If we have 4 million entries of 120 bytes each, this conservative estimate comes out at around
+/// 480MB.
 pub const SPARSE_TRIE_MAX_NODES_SHRINK_CAPACITY: usize = 4_000_000;
 
-/// Default value capacity for shrinking the sparse trie.
+/// Default value capacity for shrinking the sparse trie. This is used to limit the number of values
+/// in allocated sparse tries.
+///
+/// There are storage and account values, the largest of the two being account values, which are
+/// essentially `TrieAccount`s.
+///
+/// Account value maps have a key of `Nibbles` and value of `TrieAccount`.
+/// The `size_of::<Nibbles>` is 40, and `size_of::<TrieAccount>` is 104.
+///
+/// If we have 1 million entries of 144 bytes each, this conservative estimate comes out at around
+/// 144MB.
 pub const SPARSE_TRIE_MAX_VALUE_SHRINK_CAPACITY: usize = 1_000_000;
 
 /// Entrypoint for executing the payload.
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -72,9 +72,9 @@ pub const PARALLEL_SPARSE_TRIE_PARALLELISM_THRESHOLDS: ParallelismThresholds =
 /// Node maps have a key of `Nibbles` and value of `SparseNode`.
 /// The `size_of::<Nibbles>` is 40, and `size_of::<SparseNode>` is 80.
 ///
-/// If we have 4 million entries of 120 bytes each, this conservative estimate comes out at around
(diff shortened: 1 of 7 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -426,9 +432,18 @@ where
...
+            // this one.
+            let mut cleared_trie = ClearedSparseStateTrie::from_state_trie(trie);
+
+            // Shrink the sparse trie so that we don't have ever increasing memory.
+            cleared_trie.shrink_to(
+                SPARSE_TRIE_MAX_NODES_SHRINK_CAPACITY,
+                SPARSE_TRIE_MAX_NODES_SHRINK_CAPACITY,
+            );
+
+            cleared_sparse_trie.lock().replace(cleared_trie);
```
mattsse, line 477: we're now acquiring the lock later, I think this is fine?

wdyt @mediocregopher

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -63,6 +63,12 @@ use configured_sparse_trie::ConfiguredSparseTrie;
 pub const PARALLEL_SPARSE_TRIE_PARALLELISM_THRESHOLDS: ParallelismThresholds =
     ParallelismThresholds { min_revealed_nodes: 100, min_updated_nodes: 100 };
 
+/// Default node capacity for shrinking the sparse trie.
+pub const SPARSE_TRIE_MAX_NODES_SHRINK_CAPACITY: usize = 4_000_000;
+
+/// Default value capacity for shrinking the sparse trie.
+pub const SPARSE_TRIE_MAX_VALUE_SHRINK_CAPACITY: usize = 1_000_000;
```
mattsse: can we document here what this targets?

this is capacity but not obvious what values or how much memory this targets

mattsse (commented):
shrinking makes sense here

have some q re target memory

crates/engine/tree/src/tree/payload_processor/mod.rs
Rjected: Yeah, these were somewhat arbitrarily picked but can document how much memory this should correspond to

crates/engine/tree/src/tree/payload_processor/mod.rs
mediocregopher: Yes this is fine, it means there will be more opportunity (but still unlikely) for the newPayload call to come in and see an empty `cleared_sparse_trie`, at which point it will initialize a fresh one and continue without issue.

crates/trie/sparse/src/state.rs
```diff
@@ -847,6 +897,35 @@ impl<S: SparseTrieInterface> StorageTries<S> {
...
+        }
+
+        // Shrink cleared storage tries
+        for trie in &mut self.cleared_tries {
+            trie.shrink_nodes_to(node_size_per_trie);
+            trie.shrink_values_to(value_size_per_trie);
+        }
+
+        // Shrink the default trie
+        self.default_trie.shrink_nodes_to(node_size_per_trie);
```
mediocregopher: nit: not sure this does anything, `default_trie` never has anything called on it besides `clone`

crates/trie/sparse/src/state.rs
```diff
@@ -818,6 +844,30 @@ where
     pub fn value_capacity(&self) -> usize {
         self.state.value_capacity() + self.storage.total_value_capacity()
     }
+
+    /// Shrink the sparse trie's capacity to the given size.
+    /// This helps reduce memory usage when the trie has excess capacity.
+    /// The capacity is distributed equally across the account trie and all storage tries.
+    pub fn shrink_to(&mut self, size: usize) {
```
mediocregopher: I don't think this one is ever called, and it's implementation seems a bit wrong anyway (no separation of node vs value capacities)

mediocregopher (changes requested):
Couple cleanup comments but overall looks good 👍

crates/trie/sparse/src/state.rs
Rjected: yeah an artifact of an older version, will remove

crates/trie/sparse/src/state.rs
Rjected: yeah will remove this

mediocregopher (approved):

mattsse (approved):
gg

PR #19160, opened 2025-10-21
chore(fs-util): remove unused ReadLink error variant

jenniferzelo:
Remove unused FsPathError::ReadLink and its ctor. It was a half-implemented variant with no read_link wrapper and zero call sites.


mattsse (commented):
we can keep this

PR #19170, opened 2025-10-21
perf: fix redundant Arc clone in file_client tests

reject-i:
Found some redundant Arc::clone(&Arc::new(client)) calls in the file_client tests. This is basically creating an Arc and immediately cloning it which is unnecessary work. Changed it to just Arc::new(client). Small performance improvement for test runs and cleaner code overall. Two instances fixed on lines 818 and 893.


mattsse (approved):
fair

PR #19171, opened 2025-10-21
refactor: decouple max proof task concurrency from inflight proof limits

### LINKED ISSUE #19100: Refactor: Decouple logic for max concurrency with worker and in flight request
Right now max concurrency as a variable is only used for in-flight requests. Previously it was used for both workers and in-flight requests. We should make it clear and remove any decoupling. 

**What we need to do** 
1. Rename max concurrency as a variable for inflight limit
2. Remove max concurrency as a cli arg
3. Bench different values of inflight limit

**Motivation:**
if max_concurrency cli arg removed, which reduced max_concurrency default from 256 to 32 and this reduction could cause perf degredation. Happened due to how we make use of inflight allowance with workers. 

Context:
https://github.com/paradigmxyz/reth/issues/18872

yongkangc:
Closes: https://github.com/paradigmxyz/reth/issues/19100

Problem:
- Now that we have moved our workers to a pool based system see https://github.com/paradigmxyz/reth/pull/18887 we have also configured the workers independently. 
- But it's confusing that we have `max proof task concurrency` variable, which is now only used for inflight limit, and the naming doesn't reflect the actual nature of the use case. 

Changes:
1. Remove `max proof task concurrency` as we don't need it anymore in the code. It only serves as inflight proof limit. 
2. Introduced inflight limit. This would be removed in https://github.com/paradigmxyz/reth/pull/19178 but this pr aims to just remove cli Arg for now to keep it simple. 

Benches:no regression while running


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -34,6 +34,10 @@ use std::{
 };
 use tracing::{debug, error, trace};
 
+/// Default upper bound for inflight multiproof calculations. These would be sitting in the queue
+/// waiting to
+pub const DEFAULT_MULTIPROOF_INFLIGHT_LIMIT: usize = 256;
```
yongkangc: initially wanted this as a cli arg, but dont think it makes sense as its too hard to configure

yongkangc:
initial bench with 256 as queue limit shows some perf improvements. will be exploring removal of pending queue now.


image
image

yongkangc:
Closing as we have https://github.com/paradigmxyz/reth/pull/19178 now

crates/engine/tree/src/tree/payload_processor/multiproof.rs
mattsse: reasonable

mattsse (approved):
makes sense, especially with #19183 coming up

PR #19173, opened 2025-10-21
chore: fix misleading log message for body size check

mdqst:
the validation checks `body_size`, but the log message incorrectly says *"Header exceeds..."*.
this could confuse anyone reading the logs later. fixed to correctly reference the body instead of the header.


mattsse (approved):
fair enough

mdqst:
[mattsse](https://github.com/mattsse), thanks! pls merge it.

PR #7938, opened 2024-04-28
feat: write pruning config if --full is present

lakshya-sky:
Closes #7923

* Save prune config to reth.toml.
* warn if prune config present but run with --full.

PR #19174, opened 2025-10-21
fix(cli): prune config saving to file

shekhirin:
## Problem

Since https://github.com/paradigmxyz/reth/pull/7938 was merged, we had a bunch of changes to pruning configuration that made the initial PR useless.
1. `PruningArgs::prune_config()` [never](https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/node/core/src/args/pruning.rs#L109-L166) returns `None`, and because of that we never update the prune config in the file. https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/node/builder/src/launch/common.rs#L181-L186
2. For the same reason, we never logged the following warning. https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/node/builder/src/launch/common.rs#L187-L189

## Solution

This PR fixes this behaviour in the following way:
1. `PruningArgs::prune_config()` returns `None` if it has the default configuration. https://github.com/paradigmxyz/reth/blob/7478e900169c949fedfc6f0af4ae4bbbeb494a13/crates/node/core/src/args/pruning.rs#L169
2. If prune config in the file doesn't match the config from non-empty CLI arguments, the file is updated with the config from CLI. https://github.com/paradigmxyz/reth/blob/3553bea320cc40699ccd6f5268847f56e97bef9b/crates/node/builder/src/launch/common.rs#L183-L187

This is a backward-compatible change that has the following effect:
1. If you have an existing archive node datadir and you start a node with any pruning CLI arguments, the config file will be updated.
2. If you have an existing pruned (full / custom) node datadir and you start a node with no pruning CLI arguments, the warning log will be emitted, and the prune config from the file will be used.
3. If you have an existing pruned node datadir and you start a node with different pruning CLI arguments than what's in the config file, the config file will be updated.

### DIFF
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -23,6 +23,7 @@ pub struct Config {
     // TODO(onbjerg): Can we make this easier to maintain when we add/remove stages?
     pub stages: StageConfig,
     /// Configuration for pruning.
+    #[cfg_attr(feature = "serde", serde(default))]
     pub prune: PruneConfig,
     /// Configuration for the discovery service.
     pub peers: PeersConfig,
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -23,8 +23,7 @@ pub struct Config {
     // TODO(onbjerg): Can we make this easier to maintain when we add/remove stages?
     pub stages: StageConfig,
     /// Configuration for pruning.
-    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
-    pub prune: Option<PruneConfig>,
+    pub prune: PruneConfig,
     /// Configuration for the discovery service.
     pub peers: PeersConfig,
     /// Configuration for peer sessions.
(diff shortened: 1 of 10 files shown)


crates/config/src/config.rs
```diff
@@ -23,8 +23,7 @@ pub struct Config {
     // TODO(onbjerg): Can we make this easier to maintain when we add/remove stages?
     pub stages: StageConfig,
     /// Configuration for pruning.
-    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
-    pub prune: Option<PruneConfig>,
+    pub prune: PruneConfig,
```
mattsse, line 27: this is better but should this be marked with serde(default) because this is now mandatory and previously it was not

mattsse (approved):
checks out

only one q re serde

PR #19175, opened 2025-10-21
fix: prevent accidental overwrite of existing era1 files

forkfury:
Add file existence check before creating era1 files to prevent data loss. 

When a file already exists, the export process now logs a warning and skips that file instead of silently overwriting it.


mattsse (commented):
I think we're fine with overwriting them

PR #19176, opened 2025-10-21
refactor: remove `FullNodePrimitives`

klkvr:
This trait is not very useful and we're only using it in provider code which sometimes leads to confusing errors


mattsse (approved):
makes sense

PR #19178, opened 2025-10-21
perf: rm pending queue from MultiproofManager

### LINKED ISSUE #19177: perf: remove pending queue for MultiproofManager
### Describe the feature

In previous discussions, @mattsse has mentioned that we might not even need Pending Queue as we already have crossbeam which serves as the queue. 

In this PR https://github.com/paradigmxyz/reth/pull/19171 , we realised that not having this queue might be beneficial for perf, as the `pending_queue` was a key bottleneck (increasing limit improved performance greatly - seen on nightly metrics as well)

Therefore we should explore remove pending queue and `is_full()` from `MultiProofManager`

We do have to keep in mind possible perf regression from https://github.com/paradigmxyz/reth/pull/18727/files which introduced chunking only when queue is full

### Additional context

_No response_

yongkangc:
closes: https://github.com/paradigmxyz/reth/issues/19177

This is a follow up from https://github.com/paradigmxyz/reth/pull/19171 where we removed the cli args. here, we remove the inflight limits completely. 

The core changes are instead of manually allocating task from a VecDeque, we make use of the unbounded crossbeam channel as queue to allocate tasks and perform backpressure.


yongkangc:
Initial Bench on mainnet shows improvement of 6%. Note that main and feature is swapped. so baseline is feature and vice versa. 

image


latency_comparison

mattsse (approved):
lgtm, we need this anyway for #19183

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -756,7 +723,7 @@ impl MultiProofTask {
             chunks += 1;
         };
 
-        if should_chunk && let Some(chunk_size) = self.chunk_size {
```
shekhirin, line 759: leaving a comment to not forget to need to fix this, because now we always chunk even if all workers are busy

shekhirin (changes requested):

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -364,6 +358,9 @@ pub struct MultiproofManager {
     /// a big account change into different chunks, which may repeatedly
     /// revisit missed leaves.
     missed_leaves_storage_roots: Arc<DashMap<B256, B256>>,
+    /// Counter tracking available workers. Workers decrement when starting work,
+    /// increment when finishing. Used to determine whether to chunk multiproofs.
+    available_workers: Arc<AtomicUsize>,
```
shekhirin: I think we want to track account and storage workers separately, because all account workers can be busy, but we have some availability in storage workers

crates/trie/parallel/src/proof_task.rs
```diff
@@ -893,6 +940,10 @@ impl ProofWorkerHandle {
         let (storage_work_tx, storage_work_rx) = unbounded::<StorageWorkerJob>();
         let (account_work_tx, account_work_rx) = unbounded::<AccountWorkerJob>();
 
+        // Initialize availability counters
+        let storage_available_workers = Arc::new(AtomicUsize::new(storage_worker_count));
```
mediocregopher: Rather than initializing the atomic with the worker count, if you instead increment prior to entering the while-loop inside the account/storage_worker_loop it will accomplish the same thing and better handle the case where the worker errors while initializing its provider.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -298,6 +321,12 @@ fn account_worker_loop<Factory>(
     let mut account_nodes_processed = 0u64;
 
     while let Ok(job) = work_rx.recv() {
+        // Reserve a worker slot with saturating decrement to prevent underflow.
+        // Only decrement if counter is above zero.
+        let decremented = available_workers
```
mediocregopher: if you were to do an initial increment prior to the while-loop then you could safely simplify this to just a `fetch_add` and assume it was correctly decremented.

mediocregopher (approved):

crates/trie/parallel/src/proof_task.rs
yongkangc: thats much better, addressed

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -643,6 +630,11 @@ pub(crate) struct MultiProofTaskMetrics {
     pub first_update_wait_time_histogram: Histogram,
     /// Total time spent waiting for the last proof result.
     pub last_proof_wait_time_histogram: Histogram,
+
+    /// Histogram of pending storage worker tasks in the queue.
+    pub pending_storage_tasks_histogram: Histogram,
+    /// Histogram of pending account worker tasks in the queue.
+    pub pending_account_tasks_histogram: Histogram,
```
shekhirin: can we move this up where we have inflight metric, and rename to `pending_account_multiproofs_histogram` and `pending_storage_multiproofs_histogram`?

yongkangc: thanks, thats clearer name, fixed in [2e87b73](https://github.com/paradigmxyz/reth/pull/19178/commits/2e87b731110b2d53fa0f56693fee84b86a3eafaa)

shekhirin (approved):

PR #19180, opened 2025-10-21
refactor(ipc): simplify RpcServiceCfg from enum to struct

davidjsonn:
Converts RpcServiceCfg from an enum with two variants to a simple struct. 

The OnlyCalls variant was never used in practice (had #[allow(dead_code)] since creation) and only CallsAndSubscriptions was ever instantiated. This change removes the unnecessary complexity of pattern matching and error handling for subscriptions that would never occur.

The IPC server always needs full RPC functionality including subscriptions, so having a "calls only" mode was theoretical overhead without practical benefit.


mattsse (approved):
fair, 

I think this is coming from ported jsonrpsee ws code, we can change this

PR #19184, opened 2025-10-21
refactor(prune): remove receipts log filter segment

### LINKED ISSUE #18889: Remove `receipts_log_filter` pruning and everything associated with it
### Describe the feature

With [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110), it's no longer required for an EL node to be able to serve deposit contract logs via `eth_getLogs` for a CL node to sync/attest/propose, if an EL node is not willing to serve them over p2p. I got a confirmation from CL client devs that it's safe to delete them.

Currently, we do special treatment for receipts when we're running in full node, pruning only receipts that aren't associated with the deposit contract. Now, we can just prune all receipts beyond some distance, same as we do with changesets.

This is a breaking change for users who depend on `prune.segments.receipts_log_filter` config or `--prune.receiptslogfilter` CLI argument, so we need to remove this functionality in two steps:
1. Remove the `ReceiptsLogFilter` pruning, but leave the config entry and CLI argument, making them a no-op that just logs a warning on startup.
2. In a couple releases, deprecate the config and CLI argument.

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/18889


crates/node/core/src/args/pruning.rs
```diff
@@ -59,12 +60,10 @@ pub struct PruningArgs {
...
-    // Receipts Log Filter
-    /// Configure receipts log filter. Format:
-    /// <`address`>:<`prune_mode`>... where <`prune_mode`> can be 'full', 'distance:<`blocks`>', or
-    /// 'before:<`block_number`>'
-    #[arg(long = "prune.receiptslogfilter", value_name = "FILTER_CONFIG", conflicts_with_all = &["receipts_full", "receipts_pre_merge", "receipts_distance",  "receipts_before"], value_parser = parse_receipts_log_filter)]
-    pub receipts_log_filter: Option<ReceiptsLogPruneConfig>,
+    /// Receipts Log Filter
+    #[arg(long = "prune.receiptslogfilter", value_name = "FILTER_CONFIG", hide = true)]
+    #[deprecated]
+    pub receipts_log_filter: Option<String>,
```
mattsse, line 66: should we just yeet this entirely?

crates/node/core/src/args/pruning.rs
```diff
@@ -157,13 +157,14 @@ impl PruningArgs {
...
-            // over the logs filter
-            config.segments.receipts.take();
+
+        // Log warning if receipts_log_filter is set (deprecated feature)
+        #[expect(deprecated)]
+        if self.receipts_log_filter.is_some() {
+            tracing::warn!(
+                target: "reth::cli",
+                "The --prune.receiptslogfilter flag is deprecated and has no effect. It will be removed in a future release."
+            );
```
mattsse, line 167: ah I see we keep it for this warn message, makes sense

mattsse (approved):
all of this lgtm

pending @RomanHodulak

RomanHodulak (approved):
The `ReceiptsByLogs` is now unused?

https://github.com/paradigmxyz/reth/blob/e810df943b962dae2587d1c8fb11b9acd4a59747/crates/prune/prune/src/segments/user/receipts_by_logs.rs#L28

PR #19185, opened 2025-10-21
feat: replace Payment todo!() with explicit unsupported error

Forostovec:
The previous todo!() for CustomTransaction::Payment in execute_transaction_without_commit and commit_transaction caused panics. Direct forwarding to the inner executor is not compatible due to trait bounds requiring RecoveredTx<OpTxEnvelope>. This change keeps Op handling as before and returns a clear BlockExecutionError for Payment instead of panicking. This avoids crashes while documenting that Payment execution is currently unsupported by OpBlockExecutor.


mattsse (commented):
I think the todo here is fine

PR #19186, opened 2025-10-21
fix: add missing excess_blob_gas validation in EIP-4844 header validation

avorylli:
Add validation for excess_blob_gas field presence and multiple-of-DATA_GAS_PER_BLOB check in validate_4844_header_standalone function. 

This ensures the function matches its documentation and provides complete EIP-4844 header field validation.


mattsse (commented):
we dont need this here because validated via validate_against_parent_4844

PR #19188, opened 2025-10-21
chore(engine): Remove ConsistentDbView

mediocregopher:
This PR replaces usage of ConsistenDbView with the OverlayStateProviderFactory. The OverlayStateProviderFactory will handle reverting both hashed and trie state to the correct target block whenever a new tx is made, as well as applying any necessary in-memory overlay on top of that.


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -290,7 +290,6 @@ impl From<MultiproofInput> for PendingMultiproofTask {
 /// Input parameters for spawning a dedicated storage multiproof calculation.
 #[derive(Debug)]
 struct StorageMultiproofInput {
-    config: MultiProofConfig,
```
mediocregopher: MultiProofConfig is how we previously conveyed reverts+in-memory overlay to the proof tasks. OverlayStateProvider now does this job, so it's not necessary to convey the config anymore.

The prefix_sets field of the config is no longer applicable either, as we are always generating proofs off a DB state where hashed state and trie tables are at the same block.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -802,52 +811,49 @@ where
...
+                let multiproof_provider_factory =
+                    OverlayStateProviderFactory::new(self.provider.clone())
+                        .with_block_number(Some(block_number))
+                        .with_trie_overlay(Some(multiproof_config.nodes_sorted))
+                        .with_hashed_state_overlay(Some(multiproof_config.state_sorted));
+
                 // Use state root task only if prefix sets are empty, otherwise proof generation is
                 // too expensive because it requires walking all paths in every proof.
                 let spawn_start = Instant::now();
-                let (handle, strategy) = if trie_input.prefix_sets.is_empty() {
```
mediocregopher, line 820: I'm not sure that prefix_sets ever wasn't empty, but it definitely is always empty now

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1030,44 +1037,17 @@ where
...
         }
 
-        // Convert the historical block to the block number.
+        // Convert the historical block to the block number
         let block_number = provider
             .convert_hash_or_number(historical)?
             .ok_or_else(|| ProviderError::BlockHashNotFound(historical.as_hash().unwrap()))?;
 
-        // Retrieve revert state for historical block.
-        let (revert_state, revert_trie) = if block_number == best_block_number {
```
mediocregopher, line 1064: Reverts are now handled by the OverlayStateProviderFactory, it's only necessary to collect the highest persisted ancestor block number here (so that the OverlayStateProviderFactory knows which block to revert to) and any in-memory overlay data

crates/storage/storage-api/src/database_provider.rs
```diff
@@ -160,6 +160,29 @@ pub trait DatabaseProviderFactory: Send + Sync {
 /// Helper type alias to get the associated transaction type from a [`DatabaseProviderFactory`].
 pub type FactoryTx<F> = <<F as DatabaseProviderFactory>::DB as Database>::TX;
 
+/// A trait which can be used to describe any factory-like type which returns a read-only provider.
+pub trait DatabaseProviderROFactory {
```
mediocregopher, line 164: This trait serves multiple purposes:

* Allow passing OverlayStateProviderFactory as a generic throughout the code, which will help keep things flexible in the future.
* By not constraining Provider to `DBProvider` here we don't force OverlayStateProvider to expose its inner TX, which will make it difficult to misuse.
* Allows us to constrain the Provider to be `TrieCursorFactory + HashedCursorFactory`. The alternative would be defining two new TrieCursorFactoryFactory + HashedCursorFactoryFactory traits, which seemed worse.

crates/trie/parallel/benches/root.rs
```diff
@@ -37,7 +38,7 @@ pub fn calculate_state_root(c: &mut Criterion) {
             provider_rw.commit().unwrap();
         }
 
-        let view = ConsistentDbView::new(provider_factory.clone(), None);
+        let view = OverlayStateProviderFactory::new(provider_factory.clone());
```
mediocregopher: If we could add a blanket implementation for TrieCursorFactory and HashedCursorFactory on DBProvider, then (combined with the DatabaseProviderROFactory blanket implementation in this PR) I don't think wrapping this factory would be necessary. But doing that created some import cycles so I left it alone for now.

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -200,50 +197,45 @@ where
...
         let (to_sparse_trie, sparse_trie_rx) = channel();
-        // spawn multiproof task, save the trie input
-        let (trie_input, state_root_config) = MultiProofConfig::from_input(trie_input);
-        self.trie_input = Some(trie_input);
+
+        // We rely on the cursor factory to provide whatever DB overlay is necessary to see a
+        // consistent view of the database, including the trie tables. Because of this there is no
+        // need for an overarching prefix set to invalidate any section of the trie tables, and so
+        // we use an empty prefix set.
+        let prefix_sets = Arc::new(TriePrefixSetsMut::default());
```
shekhirin, line 225: can `prefix_sets` be removed from `ProofTaskCtx` too then?

crates/storage/provider/src/providers/state/overlay.rs
```diff
@@ -104,13 +107,8 @@ where
...
 
-        // Use the higher of the two lower bounds (or error if neither is available)
-        let Some(lower_bound) = stage_lower_bound.max(prune_lower_bound) else {
-            return Err(ProviderError::InsufficientChangesets {
-                requested: requested_block,
-                available: 0..=upper_bound,
-            })
-        };
+        // Use the higher of the two lower bounds. If neither is available assume unbounded.
+        let lower_bound = stage_lower_bound.max(prune_lower_bound).unwrap_or(0);
```
mediocregopher, line 111: Without this change the e2e tests wouldn't complete, because they don't run a pipeline to set the PruneCheckpoint. I think it's fine generally, as we shouldn't ever have a real condition where the PruneCheckpoint wouldn't be set.

crates/engine/tree/src/tree/payload_processor/mod.rs
mediocregopher: I think so yes, I had thought it would be good to keep it around in case we wanted to/needed to use it for some future use-case, but at the moment removing it would be possible.

crates/trie/parallel/benches/root.rs
shekhirin: nit
```suggestion
        let factory = OverlayStateProviderFactory::new(provider_factory.clone());
```

crates/trie/parallel/src/proof.rs
```diff
@@ -28,14 +26,7 @@ use tracing::trace;
 /// that has proof targets.
 #[derive(Debug)]
 pub struct ParallelProof {
-    /// The sorted collection of cached in-memory intermediate trie nodes that
```
shekhirin, line 31: oh, this was already unused 👍

crates/engine/tree/src/tree/payload_processor/mod.rs
shekhirin: can do as a follow-up, just nicer to not pass defaults everywhere if it's unused anyway

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -407,8 +408,7 @@ where
         let env = ExecutionEnv { evm_env, hash: input.hash(), parent_hash: input.parent_hash() };
 
         // Plan the strategy used for state root computation.
-        let state_root_plan = self.plan_state_root_computation(&input, &ctx);
-        let persisting_kind = state_root_plan.persisting_kind;
```
shekhirin, line 411: `PersistingKind` can be removed completely in a follow-up PR

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -814,50 +818,49 @@ where
...
-                            (
-                                self.payload_processor.spawn_cache_exclusive(
-                                    env,
-                                    txs,
-                                    provider_builder,
-                                ),
-                                StateRootStrategy::Parallel,
-                            )
-                        }
+                let (handle, strategy) = match self.payload_processor.spawn(
```
shekhirin, line 837: unrelated to this PR: I don't think `self.payload_processor.spawn` is actually fallible, should be possible to simplify this even more without a fallback to `self.payload_processor.spawn_cache_exclusive`, can be done in a follow-up PR

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -915,48 +918,24 @@ where
...
-        // 1. No persistence is in progress
-        // 2. Config allows it
-        let strategy = if can_run_parallel {
-            if self.config.use_state_root_task() {
-                StateRootStrategy::StateRootTask
-            } else {
-                StateRootStrategy::Parallel
-            }
-        } else {
+    fn plan_state_root_computation(&self) -> StateRootPlan {
```
shekhirin, line 923: can be done in a follow-up PR: can just return an enum now, the function is much simpler

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -994,105 +974,40 @@ where
...
-                block_number,
-                best_block_number,
-                accounts = revert_state.accounts.len(),
-                storages = revert_state.storages.len(),
-                "Non-empty revert state"
-            );
-            (revert_state, revert_trie)
-        };
-
-        input.append_cached(revert_trie.into(), revert_state);
```
shekhirin, line 1088: checks out, this was needed only for reverting the trie state

shekhirin (approved):
LGTM, only nits and follow-ups for cleanup

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -678,24 +676,35 @@ where
...
+            .with_block_number(Some(block_number))
+            .with_trie_overlay(Some(multiproof_config.nodes_sorted))
+            .with_hashed_state_overlay(Some(multiproof_config.state_sorted));
+
+        // The `hashed_state` argument is already taken into account as part of the overlay, but we
+        // need to use the prefix sets which were generated from it to indicate to the
+        // ParallelStateRoot which parts of the trie need to be recomputed.
+        let prefix_sets = Arc::into_inner(multiproof_config.prefix_sets)
+            .expect("MultiProofConfig was never cloned")
+            .freeze();
```
Rjected, line 705: the use of `into_inner` here makes me think we should have a function that constructs a `MultiProofConfig` from trie input, that gives us back the original prefix sets.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -393,37 +353,28 @@ fn storage_worker_loop<Factory>(
...
 {
-    // Create db transaction before entering work loop
-    let provider =
-        view.provider_ro().expect("Account worker failed to initialize: database unavailable");
-    let proof_tx = ProofTaskTx::new(provider.into_tx(), task_ctx, worker_id);
+    // Create provider from factory
+    let provider = task_ctx
+        .factory
+        .database_provider_ro()
+        .expect("Account worker failed to initialize: unable to create provider");
```
Rjected, line 393: nit for followup - we should make this fn return a `Result<()>` so we can just propagate this error

Rjected (approved):
this makes sense to me, also just nits that can be done as a follow up

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -121,8 +121,6 @@ where
     >,
     /// Whether to disable the parallel sparse trie.
     disable_parallel_sparse_trie: bool,
-    /// A cleared trie input, kept around to be reused so allocations can be minimized.
-    trie_input: Option<TrieInput>,
```
mattsse, line 125: epic

crates/engine/tree/src/tree/payload_validator.rs
mattsse: nice

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -678,24 +676,35 @@ where
...
-        persisting_kind: PersistingKind,
         parent_hash: B256,
         hashed_state: &HashedPostState,
         state: &EngineApiTreeState<N>,
     ) -> Result<(B256, TrieUpdates), ParallelStateRootError> {
-        let consistent_view = ConsistentDbView::new_with_latest_tip(self.provider.clone())?;
+        let provider = self.provider.database_provider_ro()?;
+
+        let (mut input, block_number) =
+            self.compute_trie_input(provider, parent_hash, state, None)?;
```
mattsse, line 686: I think all of this is tracked in https://github.com/paradigmxyz/reth/issues/19250

crates/engine/tree/src/tree/payload_validator.rs
mattsse: can we track this @shekhirin

crates/storage/provider/src/providers/state/overlay.rs
mattsse: I think this is fine

mattsse (approved):
awesome

this now unlocks a ton of additional simplifications

PR #19195, opened 2025-10-21
[wip] feat: extend `ChainStorage` capabilities

klkvr:
Extends `ChainStorage` capabilities to allow it defining custom tables and adds more hooks for writing state to those tables

Introduces `custom-index` example for speeding up `eth_transactionBySenderAndNonce` with a custom index


crates/storage/provider/src/providers/database/chain.rs
```diff
@@ -1,26 +1,32 @@
...
 
     /// Provides access to the chain writer.
-    fn writer<TX, Types>(&self) -> impl ChainStorageWriter<DatabaseProvider<TX, Types>, Primitives>
+    fn writer<TX, Types>(&self) -> impl ChainStorageWriter<DatabaseProvider<TX, Types>, N>
     where
         TX: DbTxMut + DbTx + 'static,
-        Types: NodeTypesForProvider<Primitives = Primitives>;
+        Types: NodeTypesForProvider<Primitives = N>;
+
+    /// Optionally specifies extra database tables required for node operation.
```
mattsse, line 23: can we document here how this is used?

crates/storage/storage-api/src/chain.rs
```diff
@@ -37,67 +34,47 @@ pub trait BlockBodyWriter<Provider, Body: BlockBody> {
...
-    T: BlockBodyWriter<Provider, <Primitives::Block as Block>::Body>
-{
-}
+    /// Hook to write any custom indices or state that is not covered by other methods.
+    fn write_custom_state(
+        &self,
+        _provider: &Provider,
+        _state: &ExecutionOutcome<N::Receipt>,
+    ) -> ProviderResult<()> {
+        Ok(())
```
mattsse, line 44: ah noop like this is easier than adding some optional logic, so this makes sense

crates/storage/storage-api/src/chain.rs
```diff
@@ -37,67 +34,47 @@ pub trait BlockBodyWriter<Provider, Body: BlockBody> {
...
+    }
 
-/// Input for reading a block body. Contains a header of block being read and a list of pre-fetched
-/// transactions.
-pub type ReadBodyInput<'a, B> =
-    (&'a <B as Block>::Header, Vec<<<B as Block>::Body as BlockBody>::Transaction>);
+    /// Hook to revert any [`ChainStorageWriter::write_custom_state`] changes.
+    fn remove_custom_state_above(
```
mattsse, line 48: these two traits make a ton of sense and should unblock a lot of functionality

the core limitation then is that implementations must still handle the in memory sections, which I think is fine because we also do this for all of our impls

examples/custom-index/src/provider.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl<DB: Database + Clone + Unpin + DatabaseMetrics + 'static> SenderTxReader
+    for BlockchainProvider<NodeTypesWithDBAdapter<CustomNode, DB>>
+{
+    fn transaction_by_sender_and_idx(
+        &self,
+        sender: Address,
+        tx_index: TxNumber,
+    ) -> ProviderResult<Option<Self::Transaction>> {
+        let database = self.database_provider_ro()?;
```
mattsse, line 37: what this currently doesn't do is respect the in memory section

maybe not worth adding this for this example but should be useful for users that use this as a template?

emmajam:

Issue #19197, opened 2025-10-21
Add support for eth_fillTransaction


mattsse:
### Describe the feature

this should fill all missing fields and derive type based on the fields,
similar to how this works

https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/rpc/rpc-eth-types/src/simulate.rs#L116-L117

## TODO
* define new api fn with `RpcTxReq<T::NetworkTypes>` next to https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/rpc/rpc-eth-api/src/core.rs#L667-L668
* define return type (first in reth, then upstream to alloy): https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/internal/ethapi/api.go#L1695-L1695
* implement the endpoint, this can use 

same as https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/internal/ethapi/api.go#L1620-L1620

ref https://github.com/ethereum/go-ethereum/pull/19915
ref https://github.com/ethereum/execution-apis/issues/508

PR #19199, opened 2025-10-21
feat: eth_fillTransaction

kumaryash90:
Towards #19197 

_WIP_

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
@@ -13,13 +13,13 @@ use alloy_consensus::{
 };
 use alloy_dyn_abi::TypedData;
 use alloy_eips::{eip2718::Encodable2718, BlockId};
-use alloy_network::TransactionBuilder;
+use alloy_network::{TransactionBuilder, TransactionBuilder4844};
 use alloy_primitives::{Address, Bytes, TxHash, B256, U256};
 use alloy_rpc_types_eth::{BlockNumberOrTag, TransactionInfo};
 use futures::{Future, StreamExt};
 use reth_chain_state::CanonStateSubscriptions;
 use reth_node_api::BlockBody;
-use reth_primitives_traits::{RecoveredBlock, SignedTransaction};
+use reth_primitives_traits::{RecoveredBlock, SignedTransaction, TxTy};
 use reth_rpc_convert::{transaction::RpcConvert, RpcTxReq};
 use reth_rpc_eth_types::{
     utils::binary_search, EthApiError, EthApiError::TransactionConfirmationTimeout,
@@ -440,8 +440,7 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
     fn fill_transaction(
         &self,
         mut request: RpcTxReq<Self::NetworkTypes>,
-    ) -> impl Future<Output = Result<FillTransactionResult<RpcTxReq<Self::NetworkTypes>>, Self::Error>>
-           + Send
+    ) -> impl Future<Output = Result<FillTransactionResult<TxTy<Self::Primitives>>, Self::Error>> + Send
     where
         Self: EthApiSpec + LoadBlock + EstimateCall + LoadFee,
     {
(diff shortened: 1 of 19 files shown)


crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -436,6 +436,23 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
         }
     }
 
+    /// Fills the defaults on a given unsigned transaction.
+    fn fill_transaction(
+        &self,
+        request: RpcTxReq<Self::NetworkTypes>,
+    ) -> impl Future<Output = Result<FillTransactionRes<RpcTxReq<Self::NetworkTypes>>, Self::Error>> + Send
```
mattsse: @klkvr we want to introduce `type UnsignedTx;` to

https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/rpc/rpc-convert/src/rpc.rs#L14-L14

and then have a `build_unsinged` on

https://github.com/paradigmxyz/reth/blob/792b82d8956d91255c022abf7ea39a81781d0796/crates/rpc/rpc-convert/src/transaction.rs#L122-L122

I assume?

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -436,6 +436,23 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
+    where
+        Self: EthApiSpec,
+    {
+        async move {
+            // TODO: implement transaction filling logic
+            let _ = request;
+            Err(Self::Error::from_eth_err(EthApiError::Unsupported(
+                "eth_fillTransaction not yet implemented",
+            )))
```
mattsse: here's we'd need to derive the txtype and fill missing fields (mostly gas).

it looks like the reference impl doesnt do any fancy gaslimit estimation or use suggested fees, but I feel we should do this here?

crates/rpc/rpc-eth-types/src/transaction.rs
```diff
@@ -2,11 +2,21 @@
...
+use alloy_primitives::{Bytes, B256};
 use alloy_rpc_types_eth::TransactionInfo;
 use reth_ethereum_primitives::TransactionSigned;
 use reth_primitives_traits::{NodePrimitives, Recovered, SignedTransaction};
 use reth_rpc_convert::{RpcConvert, RpcTransaction};
+use serde::{Deserialize, Serialize};
+
+/// Response type for `eth_fillTransaction` RPC method.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FillTransactionRes<T> {
```
mattsse:
```suggestion
pub struct FillTransactionResult<T> {
```

mattsse (changes requested):
great start, defer to @klkvr on detailed todos

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
klkvr: geth does this in here https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/internal/ethapi/transaction_args.go#L103

we already do similar stuff here https://github.com/paradigmxyz/reth/blob/a9b1e0677dc9f9d59660929fcb099e95767d437d/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L406-L418

so what we could try is just extracting this into a helper `fill_request` method that will be used in both `fill_transaction` and `send_transaction`

kumaryash90: thanks @klkvr @mattsse. 

i've added draft implementation logic - tried following geth's setDefaults impl, also using send_transaction and resolve_transaction (in simulate.rs) for reference.

used LoadFee for gas suggestions. other logic is similar to geth and send_transaction.

let me know if this aligns with the requirements for eth_fillTransaction function.

also, i was somewhat unclear on the encoding of unsigned tx when returning. so kept it as placeholder for now, while i find the right way to do it.

mattsse (approved):
sweet

Issue #19182, opened 2025-10-21
refactor: clean up multiproof code for better abstractions


yongkangc:
### Describe the feature

Currently its hard to navigate. We should reorder the file in
```
consts
types
pub structs
fns
```
the main type should always be top of file. the main type should always be top of file
just trips me up if I navigate to a file proof_task and then the first types are some private internal types and I have to navigate up and down

### Additional comments to address:
https://github.com/paradigmxyz/reth/pull/19203/files#r2456286060

> we can maybe separate the handling of messages into separate methods so it's easier to navigate, but not a blocker
https://github.com/paradigmxyz/reth/pull/19203#discussion_r2455291026

> https://github.com/paradigmxyz/reth/pull/19203#discussion_r2459710820
yeah this feels a bit weird because we first clone stuff from self into the
> StorageMultiproofInput only to drop it here
> imo this entire type StorageMultiproofInput is redundant

> To do that, we convert it back to the storage proof only https://github.com/paradigmxyz/reth/blob/aeff51002648bd29253eff5960bd6f05ddc1db2d/crates/trie/parallel/src/proof.rs#L145-L150
> 
> feels like a massive footgun, but Im lacking context
> 
> Previously we did this [https://github.com/paradigmxyz/reth/blob/4a24cb3b499b997f93de9b5a01bfd0ecfe8339c1/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L[…]71](https://github.com/paradigmxyz/reth/blob/4a24cb3b499b997f93de9b5a01bfd0ecfe8339c1/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L451-L471), in order to send the result of storage proof calculation back to the multiproof manager. This is basically constructing a multiproof that has just the storage proof.
> When receiving the ProofCalculated in multiproof manager, we didn't distinguish between account or storage proofs, it's just a multiproof.
> Now, we do the same here https://github.com/paradigmxyz/reth/blob/aeff51002648bd29253eff5960bd6f05ddc1db2d/crates/trie/parallel/src/proof_task.rs#L251-L262, but this codepath is also used in ParallelProof::storage_proof, and it should return only the storage proof as-is, not the multiproof with the storage proof in it.
> To do that, we convert it back to the storage proof only https://github.com/paradigmxyz/reth/blob/aeff51002648bd29253eff5960bd6f05ddc1db2d/crates/trie/parallel/src/proof.rs#L145-L150
> 
[...]

PR #19203, opened 2025-10-22
perf: Eliminate spawn_blocking in multiproof manager

yongkangc:
Closes: https://github.com/paradigmxyz/reth/issues/19183

This PR refactors the multiproof pipeline to improve performance and simplify its design by eliminating a `spawn_blocking` hop for handling proof results.

Previously, the `MultiproofManager` would spawn a blocking task to await the result of each proof calculation, which introduced unnecessary overhead and complexity.

The new implementation streamlines this process:
1.  Proof workers now send their results (`ProofResultMessage`) directly to the `MultiProofTask` event loop via a dedicated `crossbeam` channel.
2.  The `MultiProofTask` uses `crossbeam::select!` to concurrently handle both incoming control messages (like state updates) and the proof results from workers.

This change removes the intermediate blocking task, resulting in a more direct, efficient, and easier-to-understand data flow.

The updated message flow is as follows:
```text
MultiProofTask -> MultiproofManager -> ProofWorkerHandle -> Storage/Account Worker
       ^                                          |
       |                                          v
ProofResultMessage <-------- ProofResultSender ---
```

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -21,12 +21,7 @@ use reth_trie_parallel::{
     proof::ParallelProof,
     proof_task::{AccountMultiproofInput, ProofResultMessage, ProofWorkerHandle},
 };
-use std::{
-    collections::BTreeMap,
-    ops::DerefMut,
-    sync::Arc,
-    time::Instant,
-};
+use std::{collections::{BTreeMap, VecDeque}, ops::DerefMut, sync::Arc, time::Instant};
 use tracing::{debug, error, instrument, trace};
 
 /// A trie update that can be applied to sparse trie alongside the proofs for touched parts of the
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -511,7 +511,7 @@ impl MultiproofManager {
                 account_targets,
                 storage_targets,
                 ?source,
-                "Starting multiproof calculation",
+            "Dispatching multiproof to workers"
             );
 
             let start = Instant::now();
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -1167,8 +1167,11 @@ impl MultiProofTask {
                             }
                         }
                         Err(_) => {
-                            error!(target: "engine::tree::payload_processor::multiproof", "Proof result channel closed unexpectedly");
-                            return
+                            // SAFETY: This is unreachable because `self.multiproof_manager` owns
+                            // `proof_result_tx` (the sender), which keeps the channel open for
+                            // the entire lifetime of this task. The channel cannot close while
+                            // `self` exists.
+                            unreachable!("proof result channel closed while multiproof manager still exists")
                         }
                     }
                 }
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -21,7 +21,6 @@ use reth_trie::{
 use reth_trie_parallel::{
     proof::ParallelProof,
     proof_task::{AccountMultiproofInput, ProofResultMessage, ProofWorkerHandle},
-    stats::ParallelTrieTracker,
 };
 use std::{
     collections::BTreeMap,
@@ -396,7 +395,7 @@ impl MultiproofManager {
     /// Spawns a single storage proof calculation task.
     fn spawn_storage_proof(&mut self, storage_multiproof_input: StorageMultiproofInput) {
         let StorageMultiproofInput {
-            config,
+            config: _,
             source,
             hashed_state_update,
             hashed_address,
@@ -406,57 +405,42 @@ impl MultiproofManager {
             ..
         } = storage_multiproof_input;
 
-        let storage_proof_worker_handle = self.proof_worker_handle.clone();
-        let missed_leaves_storage_roots = self.missed_leaves_storage_roots.clone();
-        let proof_result_tx = self.proof_result_tx.clone();
+        let storage_targets = proof_targets.len();
 
-        self.executor.spawn_blocking(move || {
-            let storage_targets = proof_targets.len();
+        trace!(
+            target: "engine::root",
+            proof_sequence_number,
+            ?proof_targets,
+            storage_targets,
+            ?source,
+            "Dispatching storage proof to workers"
+        );
 
-            trace!(
-                target: "engine::tree::payload_processor::multiproof",
-                proof_sequence_number,
-                ?proof_targets,
-                storage_targets,
-                "Starting dedicated storage proof calculation",
-            );
-            let start = Instant::now();
-            let proof_result = ParallelProof::new(
-                config.nodes_sorted,
-                config.state_sorted,
-                config.prefix_sets,
-                missed_leaves_storage_roots,
-                storage_proof_worker_handle,
-            )
-            .with_branch_node_masks(true)
-            .with_multi_added_removed_keys(Some(multi_added_removed_keys))
-            .storage_proof(hashed_address, proof_targets);
-            let elapsed = start.elapsed();
-            trace!(
-                target: "engine::tree::payload_processor::multiproof",
-                proof_sequence_number,
-                ?elapsed,
-                ?source,
-                storage_targets,
-                "Storage multiproofs calculated",
-            );
+        let start = Instant::now();
 
-            let result_msg = match proof_result {
-                Ok(proof) => {
-                    let multiproof = DecodedMultiProof::from_storage_proof(hashed_address, proof);
-                    let stats = ParallelTrieTracker::default().finish();
-                    Ok((multiproof, stats))
-                }
-                Err(e) => Err(e),
-            };
+        // Create prefix set from targets
+        let prefix_set = reth_trie::prefix_set::PrefixSetMut::from(
+            proof_targets.iter().map(reth_trie::Nibbles::unpack),
+        );
+        let prefix_set = prefix_set.freeze();
 
-            let _ = proof_result_tx.send(ProofResultMessage {
-                sequence_number: proof_sequence_number,
-                result: result_msg,
-                elapsed,
-                state: hashed_state_update,
-            });
-        });
+        // Build computation input (data only)
+        let input = reth_trie_parallel::proof_task::StorageProofInput::new(
+            hashed_address,
+            prefix_set,
+            proof_targets,
+            true, // with_branch_node_masks
+            Some(multi_added_removed_keys),
+        );
+
+        // Dispatch to storage worker - NO spawn_blocking!
+        if let Err(e) = self.proof_worker_handle.dispatch_storage_proof(
+            input,
+            (self.proof_result_tx.clone(), proof_sequence_number, hashed_state_update, start),
+        ) {
+            error!(target: "engine::root", ?e, "Failed to dispatch storage proof");
+            return;
+        }
 
         self.inflight += 1;
         self.metrics.inflight_multiproofs_histogram.record(self.inflight as f64);
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -21,6 +21,7 @@ use reth_trie::{
 use reth_trie_parallel::{
     proof::ParallelProof,
     proof_task::{AccountMultiproofInput, ProofResultMessage, ProofWorkerHandle},
+    root::ParallelStateRootError,
     stats::ParallelTrieTracker,
 };
 use std::{
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -460,8 +460,8 @@ impl MultiproofManager {
             hashed_state_update,
             proof_targets,
             proof_sequence_number,
+            state_root_message_sender: _,
             multi_added_removed_keys,
-            ..
         } = multiproof_input;
 
         let missed_leaves_storage_roots = self.missed_leaves_storage_roots.clone();
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -896,12 +896,11 @@ impl MultiProofTask {
     ///    [`MultiproofManager::spawn`].
     ///    * If the list of proof targets is empty, the [`MultiProofMessage::EmptyProof`] message is
     ///      sent back to this task along with the original state update.
-    ///    * Otherwise, the multiproof is calculated and the [`MultiProofMessage::ProofCalculated`]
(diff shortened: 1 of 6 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -109,28 +106,13 @@ pub(super) enum MultiProofMessage {
...
-#[derive(Debug)]
-pub(super) struct ProofCalculated {
-    /// The index of this proof in the sequence of state updates
-    sequence_number: u64,
-    /// Sparse trie update
-    update: SparseTrieUpdate,
-    /// The time taken to calculate the proof.
-    elapsed: Duration,
-}
-
```
yongkangc, line 133: We can remove ProofCalculated entirely because the task loop no longer
  forwards worker results through`ProofCalculated`.

Results now arrive as `ProofResultMessage` over `proof_result_rx`, and
  the loop converts them straight into SparseTrieUpdates inline—no
  intermediate wrapper needed.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -817,9 +922,7 @@ enum AccountWorkerJob {
     /// Account multiproof computation request
     AccountMultiproof {
         /// Account multiproof input parameters
-        input: AccountMultiproofInput,
-        /// Channel to send result back to original caller
-        result_sender: Sender<AccountMultiproofResult>,
+        input: Box<AccountMultiproofInput>,
```
yongkangc, line 997: using box here as `AccountMultiproofInput` is a larger input that we pass around

crates/trie/parallel/src/proof.rs
```diff
@@ -182,26 +198,36 @@ impl ParallelProof {
...
+        let (result_tx, result_rx) = crossbeam_unbounded();
+        let account_multiproof_start_time = Instant::now();
 
         let input = AccountMultiproofInput {
             targets,
             prefix_sets,
             collect_branch_node_masks: self.collect_branch_node_masks,
             multi_added_removed_keys: self.multi_added_removed_keys.clone(),
             missed_leaves_storage_roots: self.missed_leaves_storage_roots.clone(),
+            proof_result_sender: (
```
shekhirin: this is getting harder to read, let's introduce a separate struct?

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -997,147 +1017,187 @@ impl MultiProofTask {
...
-                        trace!(target: "engine::tree::payload_processor::multiproof", "processing MultiProofMessage::FinishedStateUpdates");
-                        updates_finished = true;
-                        updates_finished_time = Some(Instant::now());
-                        if self.is_done(
-                            proofs_processed,
-                            state_update_proofs_requested,
-                            prefetch_proofs_requested,
-                            updates_finished,
-                        ) {
+            crossbeam_channel::select! {
```
shekhirin, line 1020: I think it's a very nice change: we now have a channel that we receive the state updates on, and a separate channel that we receive the proof results on. Much cleaner imo, we can maybe separate the handling of messages into separate methods so it's easier to navigate, but not a blocker

shekhirin (commented):
Will review properly in a bit, but looks good directionally

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -404,72 +387,49 @@ impl MultiproofManager {
     /// Spawns a single storage proof calculation task.
     fn spawn_storage_proof(&mut self, storage_multiproof_input: StorageMultiproofInput) {
```
mattsse: I'd like to rename this now to just `send_` removing `spawn_` from the fn names because misleading

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -639,25 +566,119 @@ pub(crate) struct MultiProofTaskMetrics {
...
-/// It is responsible of  initializing a blinded sparse trie and subscribe to
-/// transaction state stream. As it receives transaction execution results, it
-/// fetches the proofs for relevant accounts from the database and reveal them
-/// to the tree.
-/// Then it updates relevant leaves according to the result of the transaction.
-/// This feeds updates to the sparse trie task.
+/// ## Architecture: Dual-Channel Multiproof System
+///
+/// This task orchestrates parallel proof computation using a dual-channel architecture that
+/// separates control messages from proof computation results:
```
mattsse, line 571: this is very helpful

crates/engine/tree/src/tree/payload_processor/multiproof.rs
mattsse, line 1020: all of this looks good to me

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -997,147 +1017,187 @@ impl MultiProofTask {
...
+                            ) {
+                                debug!(
+                                    target: "engine::tree::payload_processor::multiproof",
+                                    "State updates finished and all proofs processed, ending calculation"
+                                );
+                                break
+                            }
+                        }
+                        Err(_) => {
+                            error!(target: "engine::tree::payload_processor::multiproof", "Proof result channel closed unexpectedly");
```
mattsse, line 1173: this should be unreachable because always keep the channel open via the 

MultiproofManager.proof_result_tx

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -332,13 +305,20 @@ impl MultiproofInput {
...
+/// # Flow
+/// 1. `MultiProofTask` asks the manager to spawn either storage or account proof work.
+/// 2. The manager builds the request, clones `proof_result_tx`, and hands everything to
+///    [`ProofWorkerHandle`].
+/// 3. A worker finishes the proof and sends a [`ProofResultMessage`] through the channel included
+///    in the job.
+/// 4. `MultiProofTask` consumes the message from the same channel and sequences it with
+///    `ProofSequencer`.
 #[derive(Debug)]
 pub struct MultiproofManager {
```
mattsse, line 319: this type now simply acts as container to groups certain fields within the `MultiProofTask`

I personally find this introduces more complexity because now you also need to know what this entity does even if this now only does simple dispatch, similar situation with the ProofSequencer

all of this could be squashed into the 
MultiProofTask

but we can decide this later

mattsse (changes requested):
I like this a lot especially because this removes a totally redundant spawn

imo there's room for cleanup (not in this pr tho) see also #19182

crates/trie/parallel/src/proof.rs
yongkangc: good suggestion! addressed in [5aa0087](https://github.com/paradigmxyz/reth/pull/19203/commits/5aa00875e45f211ae77f40c3128e0bb36735b585)

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: i changed to the hierarchy and naming to dispatch (felt more fitting than send) :


dispatch()                    // Entry point - checks if empty, then sends
    ~~→ send_multiproof_task()    // removed shallow function~~
        → dispatch_storage_proof()   // dispatch to storage workers
        → dispatch_multiproof()      // dispatch to account workers


what do you think @mattsse ?

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: this makes sense, let me configure and see how to clean that up 

> we can maybe separate the handling of messages into separate methods so it's easier to navigate, but not a blocker

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: thanks for the note, noted this down as a cleanup in 
https://github.com/paradigmxyz/reth/issues/19182

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: good pt, changed to `unreachable!` [01f1db8](https://github.com/paradigmxyz/reth/pull/19203/commits/01f1db802636748c0dd60425d7d07f4ceae5435e)

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: @shekhirin just did a prototype, i think its easier to review in a separate pr :) noted down in https://github.com/paradigmxyz/reth/issues/19182

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -493,91 +452,58 @@ impl MultiproofManager {
...
+    fn dispatch_multiproof(&mut self, multiproof_input: MultiproofInput) {
         let MultiproofInput {
             config,
             source,
             hashed_state_update,
             proof_targets,
             proof_sequence_number,
-            state_root_message_sender,
             multi_added_removed_keys,
+            ..
```
shekhirin: nit, to be the same as in `dispatch_storage_proof`

```suggestion
            state_root_message_sender: _,
```

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -997,147 +1016,164 @@ impl MultiProofTask {
...
+                                updates_finished,
+                            ) {
+                                debug!(
+                                    target: "engine::tree::payload_processor::multiproof",
+                                    "State updates finished and all proofs processed, ending calculation"
+                                );
+                                    break
+                                }
+                            }
+                            MultiProofMessage::EmptyProof { sequence_number, state } => {
```
shekhirin, line 1089: something is going on with indentation here, rustfmt gave up I guess

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -310,15 +283,15 @@ impl StorageMultiproofInput {
...
+/// Input parameters for dispatching a multiproof calculation.
 #[derive(Debug)]
 struct MultiproofInput {
     config: MultiProofConfig,
     source: Option<StateChangeSource>,
     hashed_state_update: HashedPostState,
     proof_targets: MultiProofTargets,
     proof_sequence_number: u64,
-    state_root_message_sender: Sender<MultiProofMessage>,
+    state_root_message_sender: CrossbeamSender<MultiProofMessage>,
```
shekhirin, line 294: now it's only used for `EmptyProof` message, I wonder if we can remove it altogether and just send the empty proof to `proof_result_tx`? No need to address it in this PR, just a thought cc @mattsse

shekhirin (approved):
Very nice, LGTM

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: makes sense! resolved

crates/engine/tree/src/tree/payload_processor/multiproof.rs
mattsse: makes sense

crates/engine/tree/src/tree/payload_processor/multiproof.rs
shekhirin: @yongkangc I think we may need to fix this manually, idk why rustfmt doesn't do it for us

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -385,91 +368,67 @@ impl MultiproofManager {
...
         let StorageMultiproofInput {
-            config,
-            source,
             hashed_state_update,
             hashed_address,
             proof_targets,
             proof_sequence_number,
-            state_root_message_sender,
             multi_added_removed_keys,
+            state_root_message_sender: _,
```
mattsse, line 389: yeah this feels a bit weird because we first clone stuff from self into the 

StorageMultiproofInput only to drop it here

imo this entire type StorageMultiproofInput is redundant

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -997,147 +1016,164 @@ impl MultiProofTask {
...
+                                );
+                                break
+                            }
+                        }
+                        Err(_) => {
+                            // SAFETY: This is unreachable because `self.multiproof_manager` owns
+                            // `proof_result_tx` (the sender), which keeps the channel open for
+                            // the entire lifetime of this task. The channel cannot close while
+                            // `self` exists.
+                            unreachable!("proof result channel closed while multiproof manager still exists")
```
mattsse: this checks out because select does return `RecvError` and not TryRecvError so this is only reachable if the channel is disconnected.

tho maybe we should just return the an error here

crates/trie/parallel/src/proof_task.rs
```diff
@@ -522,15 +648,25 @@ where
...
+                        })?;
+
+                        // Extract storage proof from the multiproof wrapper
+                        let (mut multiproof, _stats) = proof_msg.result?;
+                        let proof =
+                            multiproof.storages.remove(&hashed_address).ok_or_else(|| {
+                                ParallelStateRootError::Other(format!(
+                                    "storage proof not found in multiproof for {hashed_address}"
+                                ))
+                            })?;
```
mattsse, line 667: why do we need to do this now?

this is added a few times and unclear what changed here?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -580,8 +716,13 @@ where
 
     // Consume remaining storage proof receivers for accounts not encountered during trie walk.
     for (hashed_address, receiver) in storage_proof_receivers {
-        if let Ok(Ok(proof)) = receiver.recv() {
-            collected_decoded_storages.insert(hashed_address, proof);
+        if let Ok(proof_msg) = receiver.recv() {
+            // Extract storage proof from the multiproof wrapper
+            if let Ok((mut multiproof, _stats)) = proof_msg.result &&
+                let Some(proof) = multiproof.storages.remove(&hashed_address)
```
mattsse, line 721: same here

mattsse (commented):
one last question otherwise lgtm

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: > receiving the ProofCalculated in multiproof manager
This is incorrect. we should get rid of them

mattsse (approved):

crates/trie/parallel/src/proof_task.rs
```diff
@@ -176,13 +248,35 @@ fn storage_worker_loop<Factory>(
...
+                            hashed_address,
+                            storage_proof,
+                        );
+                        let stats = crate::stats::ParallelTrieTracker::default().finish();
+                        Ok((multiproof, stats))
+                    }
+                    Err(e) => Err(e),
+                };
+
+                // Send result directly to MultiProofTask
```
shekhirin: let's remove this comment, because this codepath is also used from `ParallelProof::storage_proof`

yongkangc: @shekhirin removed comment

PR #19204, opened 2025-10-22
perf(net): convert Bytes to BytesMut to avoid reallocation

robinsdan:
If `Bytes` has only one reference to the data, `Into<BytesMut>` would avoid cloning the underlying buffer.


mattsse (approved):
I believe this is correct even if Bytes is unique

PR #19207, opened 2025-10-22
fix: OverlayStateProviderFactory: validating trie changeset range and revert target

mediocregopher:
When determining the available range of trie changeset data we previously only took into account the StageCheckpoint. This was wrong for two reasons:

* The StageCheckpoint can have its inner block range cleared, specifically on unwind or reorgs. So we can't rely on the starting bound being present.

* The StageCheckpoint's lower bound doesn't get updated by pruning, only the prune checkpoint does. So in general we'll want to rely on the prune checkpoint.

This also fixes an issue in OverlayStateProviderFactory related to which block_number was being reverted, and clarifies that behavior in the docs.

### DIFF
--- a/crates/storage/provider/src/providers/state/overlay.rs
+++ b/crates/storage/provider/src/providers/state/overlay.rs
@@ -16,6 +16,7 @@ use reth_trie_db::{
     DatabaseHashedCursorFactory, DatabaseHashedPostState, DatabaseTrieCursorFactory,
 };
 use std::sync::Arc;
+use tracing::debug;
 
 /// Factory for creating overlay state providers with optional reverts and overlays.
 ///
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/state/overlay.rs
```diff
@@ -64,25 +73,47 @@ where
...
+        let prune_lower_bound =
+            prune_checkpoint.and_then(|chk| chk.block_number.map(|block| block + 1));
+
+        // Use the higher of the two lower bounds (or error if neither is available)
+        let Some(lower_bound) = stage_lower_bound.max(prune_lower_bound) else {
+            return Err(ProviderError::InsufficientChangesets {
+                requested: requested_block,
+                available: 0..=upper_bound,
+            })
+        };
```
mattsse, line 113: I see

mattsse (approved):
> want to rely on the prune checkpoint

makes sense, lgtm!

PR #19208, opened 2025-10-22
refactor(engine): remove deprecated CLI arguments

davidjsonn:
Removes three deprecated fields from EngineArgs that have been sitting as dead code for months. The caching_and_prewarming_enabled field was deprecated back in June 2025, while parallel_sparse_trie_enabled and precompile_cache_enabled were both marked deprecated since August. These fields were completely ignored in the tree_config() method where the actual configuration happens. The functionality was moved to new fields with inverted logic like prewarming_disabled, parallel_sparse_trie_disabled, and precompile_cache_disabled. CLI backward compatibility is still maintained through aliases defined on the working fields, so existing scripts won't break.


mattsse (commented):
we need to keep them so that this doesnt break existing setups, can remove later

PR #19209, opened 2025-10-22
refactor(static-file): remove unused segments

shekhirin:
With https://github.com/paradigmxyz/reth/pull/18681, we don't need the headers and transactions static file producers anymore, as the data is written directly to static files.

This change is not breaking, as we've been writing to both database and static files for a very long time now (since Engine refactoring where we introduced the new background persistence mechanism), so the static file producer for headers and transactions hasn't actually been running for a while now.

For receipts segment, we _might_ still need it, because for full nodes receipts are currently written to database **only**, but since the [removal of receipts log filter](https://github.com/paradigmxyz/reth/pull/19184), full nodes can write receipts directly to static files too.


crates/static-file/static-file/src/static_file_producer.rs
```diff
@@ -314,68 +299,47 @@ mod tests {
...
+            HighestStaticFiles { headers: None, receipts: Some(3), transactions: None }
         );
 
         let targets = static_file_producer
             .get_static_file_targets(HighestStaticFiles {
-                headers: Some(4),
+                headers: None,
                 receipts: Some(4),
-                transactions: Some(4),
+                transactions: None,
```
joshieDo: could we just remove `headers` and `transactions` from `HighestStaticFiles`?

shekhirin: yeah, we can, removed

joshieDo (approved):

mattsse (approved):
yeeet

PR #19214, opened 2025-10-22
perf: check prewarm termination multiple times

mattsse:
we were only checking the atomicbool that indicates when we can stop prewarming (e.g. block fully executed) before we started execution, 
we should also do this after and stop sending messages once terminated

these relaxed checks are cheap so we can check this a few times

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -350,7 +350,7 @@ where
             }
         }
 
-        trace!(target: "engine::tree::prewarm", "Completed prewarm execution");
+        debug!(target: "engine::tree::prewarm", "Completed prewarm execution");
 
         // save caches and finish
         if let Some(Some(state)) = final_block_output {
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -233,8 +233,19 @@ where
         });
     }
 
+    /// Returns true if prewarming was terminated and no more transactions should be prewarmed.
+    fn is_execucation_terminated(&self) -> bool {
```
yongkangc:
```suggestion
    fn is_execution_terminated(&self) -> bool {
```

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -233,8 +233,19 @@ where
...
+        self.ctx.terminate_execution.load(Ordering::Relaxed)
+    }
+
     /// If configured and the tx returned proof targets, emit the targets the transaction produced
     fn send_multi_proof_targets(&self, targets: Option<MultiProofTargets>) {
+        if self.is_execucation_terminated() {
```
yongkangc:
```suggestion
        if self.is_execution_terminated() {
```

yongkangc (approved):
just one naming nit

PR #19215, opened 2025-10-22
fix(cli): prune CLI argument names

shekhirin:
1. `--block-interval` -> `--prune.block-interval`
2. Add missing hyphens (like `--prune.senderrecovery.full` -> `--prune.sender-recovery.full`)

Backward-compatible, I also added old names as aliases.


RomanHodulak (approved):

mattsse (approved):
lgtm, this is non breaking because we keep aliases

PR #19220, opened 2025-10-22
rpc/ipc: drop unused stop handle from ServiceData

dennsikl:
Brief housekeeping in the IPC server: keep the per-connection subscription limiter, but stop storing an unused StopHandle copy in ServiceData. The limiter still lives on ServiceData so we clone it per request, preserving the existing max_subscriptions_per_connection limits


mattsse (commented):
I believe we need to keep this one

PR #19221, opened 2025-10-22
chore: fix incorrect comment about `BlockReader` interface

mdqst:
noticed that the previous comment incorrectly described what `BlockReader` interacts with.
it actually deals with tables containing full blocks (not just headers), providing methods like `block`, `block_by_hash`, `sealed_block_with_senders`, etc.


mattsse (commented):
not worth a pr

PR #19222, opened 2025-10-22
feat: warning log when blocked on execution cache

mediocregopher:


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -596,8 +599,16 @@ impl ExecutionCache {
...
     /// - It exists and matches the requested parent hash
     /// - No other tasks are currently using it (checked via Arc reference count)
+    #[instrument(level = "debug", target = "engine::tree::payload_processor", skip(self))]
     pub(crate) fn get_cache_for(&self, parent_hash: B256) -> Option<SavedCache> {
+        let start = Instant::now();
         let cache = self.inner.read();
+
+        let elapsed = start.elapsed();
+        if elapsed.as_millis() > 5 {
+            warn!(blocked_for=?elapsed, "Blocked waiting for execution cache mutex");
```
shekhirin, line 609: needs a target, events don't inherit the span target afaik

yongkangc (approved):
makes sense to me

crates/engine/tree/src/tree/payload_processor/mod.rs
shekhirin: oh, maybe it's fine here, because it's going to be in stdout?

mattsse (approved):
reasonable until fixed properly

shekhirin (approved):

PR #19224, opened 2025-10-22
refactor(e2e): relax bound

klkvr:
Relaxes `PayloadBuilderAttributes: From<reth_payload_builder::EthPayloadBuilderAttributes>` on `NodeBuilderHelper`

We already impose `LocalPayloadAttributesBuilder` bound everywhere so we can just use it


mattsse (approved):
makes sense

PR #19226, opened 2025-10-22
fix(reth-bench): Lower block channel capacity and make it configurable

mediocregopher:
reth-bench spawns a background task to fetch blocks from RPC, pushing those blocks into a channel. The channel previously had a capacity of 1000, which did not cause sufficient backpressure to prevent us from hitting rate-limits on RPC providers. The capacity has been lowered to 20, but can be configured higher if necessary.

This also improves the error propagation from that task. Previously it would just panic when there was an error, but that panic was getting lost. Now it propagates the error to the main thread which properly shows it to the user.


mattsse (approved):
makes sense to make this configurable

PR #19228, opened 2025-10-22
fix: rustc-hash conditional compilation in reth-ethereum-forks

reject-i:
Running cargo test -p reth-primitives-traits --lib --no-default-features was failing with `unresolved import rustc_hash` error because rustc_hash::FxHashMap was imported with only `std` feature but rustc-hash dependency requires `rustc-hash` feature.

Fixed conditional compilation to require both `std` and `rustc-hash` features for FxHashMap import and added std::HashMap fallback when rustc-hash is disabled.

### DIFF
--- a/crates/ethereum/hardforks/src/hardforks/mod.rs
+++ b/crates/ethereum/hardforks/src/hardforks/mod.rs
@@ -2,7 +2,7 @@ mod dev;
 pub use dev::DEV_HARDFORKS;
 
 use crate::{ForkCondition, ForkFilter, ForkId, Hardfork, Head};
-#[cfg(feature = "std")]
+#[cfg(all(feature = "std", feature = "rustc-hash"))]
 use rustc_hash::FxHashMap;
 #[cfg(feature = "std")]
 use std::collections::hash_map::Entry;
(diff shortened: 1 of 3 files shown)


crates/ethereum/hardforks/src/hardforks/mod.rs
```diff
@@ -48,8 +48,10 @@ pub trait Hardforks: Clone {
 #[derive(Default, Clone, PartialEq, Eq)]
 pub struct ChainHardforks {
     forks: Vec<(Box<dyn Hardfork>, ForkCondition)>,
-    #[cfg(feature = "std")]
+    #[cfg(all(feature = "std", feature = "rustc-hash"))]
     map: FxHashMap<&'static str, ForkCondition>,
+    #[cfg(all(feature = "std", not(feature = "rustc-hash")))]
+    map: std::collections::HashMap<&'static str, ForkCondition>,
```
mattsse: looks like rustc-hash is no fully no-std,

can we unify this to always use fxhashmap ?

basically get rid of 

map: std::collections::HashMap<&'static str, ForkCondition>,

mattsse (changes requested):

crates/ethereum/hardforks/src/hardforks/mod.rs
reject-i: @mattsse thanks! I tried that route, but ran into the same compile error we started with. In our workspace `reth-ethereum-forks` is pulled in with `default-features = false`, so the `rustc-hash` feature never gets activated unless every downstream crate explicitly enables it. In that case `rustc_hash::FxHashMap` isn’t even compiled, and the build fails again. I’m keeping the `std::HashMap` fallback for now—if we’re okay enforcing `rustc-hash` everywhere, happy to flip it back

mattsse: yeah my suggestion would be to remove the rustc-hash feature and always use fxhashmap because it looks like the nostd issues got fixed in rustc-hash 2.0

mattsse (commented):
actually we cant do this because fxhashmap requires std

PR #19233, opened 2025-10-22
chore(trie): Use Vec<Option<...>> in HashedPostStateCursors

### LINKED ISSUE #18848: Use Vec<(B256, Option<...>)> in HashedPostStateCursors
### Describe the feature

The [HashedPostStateAccountCursor](https://github.com/paradigmxyz/reth/blob/fafe44d3864bda61f936ae37a4a9901d00922bff/crates/trie/trie/src/hashed_cursor/post_state.rs#L43) and [HashedPostStateStorageCursor](https://github.com/paradigmxyz/reth/blob/fafe44d3864bda61f936ae37a4a9901d00922bff/crates/trie/trie/src/hashed_cursor/post_state.rs#L177C12-L177C40) currently use a combination of iterator over updated accounts/slots and a separate HashSet of cleared accounts/slots to act as an overlay over the underlying Cursor.

This could be made more efficient by removing the HashSet on each, and making the iterator into an iterator over an `Option<Account>` (for accounts) or `U256` (for storages, U256::ZERO indicates deleted).

This change will require accompanying changes to [HashedPostStateSorted](https://github.com/paradigmxyz/reth/blob/fafe44d3864bda61f936ae37a4a9901d00922bff/crates/trie/common/src/hashed_state.rs#L462).

A previous PR was already merged, #18479 , which performed essentially this exact change but for TrieUpdates. This PR can be used as a guide; the logic and necessary changes should be effectively the same, just the types are different.

### Additional context

_No response_

0x00101010:
Fixes https://github.com/paradigmxyz/reth/issues/18848

## Summary

  This PR refactors and simplifies the `HashedPostStateCursor` implementation by removing redundant state tracking and unifying the cursor types. The changes reduce code complexity while maintaining the same functional behavior.

  ## Key Changes

  ### 1. Unified Cursor Implementation

  - **Merged cursor types**: Combined `HashedPostStateAccountCursor` and `HashedPostStateStorageCursor` into a single generic `HashedPostStateCursor<C, V>` type
  - **Removed redundant `wiped` field**: The `wiped` state had 1:1 correspondence with the cursor being `None`/`Some`, so the field was redundant
  - **Simplified cursor construction**: Changed from passing `Some(NoopHashedCursor::default())` to passing `None` directly when no database cursor is needed

  ### 2. Simplified `HashedPostStateSorted` Structure

  **Before:**
  ```rust
  pub struct HashedAccountsSorted {
      pub accounts: Vec<(B256, Account)>,
      pub destroyed_accounts: B256Set,
  }

  pub struct HashedStorageSorted {
      pub non_zero_valued_slots: Vec<(B256, U256)>,
      pub zero_valued_slots: B256Set,
      pub wiped: bool,
  }

  After:
  pub struct HashedPostStateSorted {
      pub accounts: Vec<(B256, Option<Account>)>,  // None = destroyed
      pub storages: B256Map<HashedStorageSorted>,
  }

  pub struct HashedStorageSorted {
      pub storage_slots: Vec<(B256, Option<U256>)>,  // None = zero-valued
      pub wiped: bool,
  }

  This change:
  - Uses Option<T> consistently to represent deletions/zero-values
  - Eliminates separate collections for tracking destroyed/zero-valued entries
  - Maintains sorted order in a single vector per type
  - Simplifies iteration logic by removing the need to merge and re-sort

  3. Improved ForwardInMemoryCursor

  - Changed from iterator-based to index-based traversal
  - Added has_any() method for checking if any entry satisfies a predicate
  - More efficient current() implementation that doesn't clone the iterator

  4. Simplified is_storage_empty() Logic

  Before:
  fn is_storage_empty(&mut self) -> Result<bool, DatabaseError> {
      let post_state_is_empty = /* complex logic checking if all are zero */;
      Ok(post_state_is_empty && self.cursor.as_mut().map_or(self.wiped == Some(true), |c| c.is_storage_empty()?))
  }

  After:
  fn is_storage_empty(&mut self) -> Result<bool, DatabaseError> {
      // Storage is not empty if it has non-zero slots
      if self.post_state_cursor.has_any(|(_, value)| value.is_some()) {
          return Ok(false);
      }

      // If no non-zero slots in post state, check the database
      // Returns true if cursor is None (wiped storage or empty DB)
      self.cursor.as_mut().map_or(Ok(true), |c| c.is_storage_empty())
  }

  5. Documentation Improvements

  - Added comprehensive documentation to HashedPostStateSorted explaining the Option<T> semantics
  - Clarified that Some(value) means update and None means deletion
  - Improved comments in critical sections

  6. Test Coverage

  - Added regression test all_storage_slots_deleted_not_wiped_exact_keys to cover edge case where all storage slots are deleted (zero-valued) but storage is not wiped
  - Test ensures repeated seek() operations correctly return None for all deleted slots
  - Added test case for non-wiped storage with all zero-value entries

  Testing

  All 177 trie-related tests pass:
  - reth-trie: 36 tests
  - reth-trie-db: 70 tests
  - reth-trie-sparse: 36 tests
  - reth-trie-sparse-parallel: 35 tests

  Benefits

  1. Less code: Reduced by ~30 net lines while adding comprehensive tests
  2. Simpler logic: Single generic cursor type instead of multiple specialized types
  3. Better performance: Index-based cursor traversal instead of iterator cloning
  4. Clearer semantics: Option<T> explicitly communicates update vs deletion
  5. Easier to maintain: Less duplicate code and clearer abstractions

### DIFF
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -332,17 +332,8 @@ impl HashedPostState {
     }
 
     /// Converts hashed post state into [`HashedPostStateSorted`].
-    pub fn into_sorted(self) -> HashedPostStateSorted {
-        let mut accounts: Vec<_> = self.accounts.into_iter().collect();
-        accounts.sort_unstable_by_key(|(address, _)| *address);
-
-        let storages = self
-            .storages
-            .into_iter()
-            .map(|(hashed_address, storage)| (hashed_address, storage.into_sorted()))
-            .collect();
-
-        HashedPostStateSorted { accounts, storages }
+    pub fn into_sorted(mut self) -> HashedPostStateSorted {
+        self.drain_into_sorted()
     }
 
     /// Converts hashed post state into [`HashedPostStateSorted`], but keeping the maps allocated by
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -425,10 +425,7 @@ impl HashedStorage {
 
     /// Converts hashed storage into [`HashedStorageSorted`].
     pub fn into_sorted(self) -> HashedStorageSorted {
-        let mut storage_slots: Vec<_> = self
-            .storage
-            .into_iter()
-            .collect();
+        let mut storage_slots: Vec<_> = self.storage.into_iter().collect();
         storage_slots.sort_unstable_by_key(|(key, _)| *key);
 
         HashedStorageSorted { storage_slots, wiped: self.wiped }
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -428,7 +428,6 @@ impl HashedStorage {
         let mut storage_slots: Vec<_> = self
             .storage
             .into_iter()
-            .map(|(hashed_slot, value)| (hashed_slot, (!value.is_zero()).then_some(value)))
             .collect();
         storage_slots.sort_unstable_by_key(|(key, _)| *key);
 
@@ -508,7 +507,7 @@ impl AsRef<Self> for HashedPostStateSorted {
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub struct HashedStorageSorted {
     /// Sorted collection of updated storage slots, None indicates zero valued.
-    pub storage_slots: Vec<(B256, Option<U256>)>,
+    pub storage_slots: Vec<(B256, U256)>,
     /// Flag indicating whether the storage was wiped or not.
     pub wiped: bool,
 }
@@ -520,7 +519,7 @@ impl HashedStorageSorted {
     }
 
     /// Returns reference to updated storage slots.
-    pub fn storage_slots_ref(&self) -> &[(B256, Option<U256>)] {
+    pub fn storage_slots_ref(&self) -> &[(B256, U256)] {
         &self.storage_slots
     }
 
@@ -1128,19 +1127,19 @@ mod tests {
         // Test normal extension
         let mut storage1 = HashedStorageSorted {
             storage_slots: vec![
-                (B256::from([1; 32]), Some(U256::from(10))),
-                (B256::from([3; 32]), Some(U256::from(30))),
-                (B256::from([5; 32]), None),
+                (B256::from([1; 32]), U256::from(10)),
+                (B256::from([3; 32]), U256::from(30)),
+                (B256::from([5; 32]), U256::ZERO),
             ],
             wiped: false,
         };
 
         let storage2 = HashedStorageSorted {
             storage_slots: vec![
-                (B256::from([2; 32]), Some(U256::from(20))),
-                (B256::from([3; 32]), Some(U256::from(300))), // Override
-                (B256::from([4; 32]), Some(U256::from(40))),
-                (B256::from([6; 32]), None),
+                (B256::from([2; 32]), U256::from(20)),
+                (B256::from([3; 32]), U256::from(300)), // Override
+                (B256::from([4; 32]), U256::from(40)),
+                (B256::from([6; 32]), U256::ZERO),
             ],
             wiped: false,
         };
@@ -1149,32 +1148,32 @@ mod tests {
 
         assert_eq!(storage1.storage_slots.len(), 6);
         assert_eq!(storage1.storage_slots[0].0, B256::from([1; 32]));
-        assert_eq!(storage1.storage_slots[0].1, Some(U256::from(10)));
+        assert_eq!(storage1.storage_slots[0].1, U256::from(10));
         assert_eq!(storage1.storage_slots[1].0, B256::from([2; 32]));
-        assert_eq!(storage1.storage_slots[1].1, Some(U256::from(20)));
+        assert_eq!(storage1.storage_slots[1].1, U256::from(20));
         assert_eq!(storage1.storage_slots[2].0, B256::from([3; 32]));
-        assert_eq!(storage1.storage_slots[2].1, Some(U256::from(300))); // Should have storage2's value
+        assert_eq!(storage1.storage_slots[2].1, U256::from(300)); // Should have storage2's value
         assert_eq!(storage1.storage_slots[3].0, B256::from([4; 32]));
-        assert_eq!(storage1.storage_slots[3].1, Some(U256::from(40)));
+        assert_eq!(storage1.storage_slots[3].1, U256::from(40));
         assert_eq!(storage1.storage_slots[4].0, B256::from([5; 32]));
-        assert_eq!(storage1.storage_slots[4].1, None);
+        assert_eq!(storage1.storage_slots[4].1, U256::ZERO);
         assert_eq!(storage1.storage_slots[5].0, B256::from([6; 32]));
-        assert_eq!(storage1.storage_slots[5].1, None);
+        assert_eq!(storage1.storage_slots[5].1, U256::ZERO);
         assert!(!storage1.wiped);
 
         // Test wiped storage
         let mut storage3 = HashedStorageSorted {
             storage_slots: vec![
-                (B256::from([1; 32]), Some(U256::from(10))),
-                (B256::from([2; 32]), None),
+                (B256::from([1; 32]), U256::from(10)),
+                (B256::from([2; 32]), U256::ZERO),
             ],
             wiped: false,
         };
 
         let storage4 = HashedStorageSorted {
             storage_slots: vec![
-                (B256::from([3; 32]), Some(U256::from(30))),
-                (B256::from([4; 32]), None),
+                (B256::from([3; 32]), U256::from(30)),
+                (B256::from([4; 32]), U256::ZERO),
             ],
             wiped: true,
         };
@@ -1185,9 +1184,9 @@ mod tests {
         // When wiped, should only have storage4's values
         assert_eq!(storage3.storage_slots.len(), 2);
         assert_eq!(storage3.storage_slots[0].0, B256::from([3; 32]));
-        assert_eq!(storage3.storage_slots[0].1, Some(U256::from(30)));
+        assert_eq!(storage3.storage_slots[0].1, U256::from(30));
         assert_eq!(storage3.storage_slots[1].0, B256::from([4; 32]));
-        assert_eq!(storage3.storage_slots[1].1, None);
+        assert_eq!(storage3.storage_slots[1].1, U256::ZERO);
     }
 
     #[test]
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -468,9 +468,7 @@ impl HashedPostStateSorted {
 
     /// Returns `true` if there are no account or storage updates.
     pub fn is_empty(&self) -> bool {
-        self.accounts.accounts.is_empty() &&
-            self.accounts.destroyed_accounts.is_empty() &&
-            self.storages.is_empty()
+        self.accounts.is_empty() && self.storages.is_empty()
     }
 
     /// Returns the total number of updates including all accounts and storage updates.
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -9,7 +9,7 @@ use crate::{
 use alloc::{borrow::Cow, vec::Vec};
 use alloy_primitives::{
     keccak256,
-    map::{hash_map, B256Map, B256Set, HashMap, HashSet},
+    map::{hash_map, B256Map, HashMap, HashSet},
     Address, B256, U256,
 };
 use itertools::Itertools;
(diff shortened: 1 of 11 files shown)


crates/trie/common/src/hashed_state.rs
```diff
@@ -515,47 +491,45 @@ impl AsRef<Self> for HashedPostStateSorted {
...
 
 /// Sorted hashed storage optimized for iterating during state trie calculation.
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub struct HashedStorageSorted {
-    /// Sorted hashed storage slots with non-zero value.
-    pub non_zero_valued_slots: Vec<(B256, U256)>,
-    /// Slots that have been zero valued.
-    pub zero_valued_slots: B256Set,
+    /// Sorted collection of updated storage slots, None indicates zero valued.
+    pub storage_slots: Vec<(B256, Option<U256>)>,
```
mediocregopher: I believe instead of `Option<U256>` you can use `U256::ZERO` to signal a value removal, unless you discovered some reason this wouldn't be the case.

crates/trie/common/src/hashed_state.rs
```diff
@@ -515,47 +491,45 @@ impl AsRef<Self> for HashedPostStateSorted {
...
-        self.accounts.retain(|(addr, _)| !other.destroyed_accounts.contains(addr));
-
-        // Extend the sorted accounts vector
-        extend_sorted_vec(&mut self.accounts, &other.accounts);
-
-        // Merge destroyed accounts sets
-        self.destroyed_accounts.extend(&other.destroyed_accounts);
-    }
-}
+// /// Sorted account state optimized for iterating during state trie calculation.
```
mediocregopher: Reminder to delete this before submitting

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,312 +25,184 @@ where
...
-            Some(account) => {
-                let entry = self.next_inner(account)?;
-                self.last_account = entry.as_ref().map(|entry| entry.0);
-                entry
-            }
-            // no previous entry was found
-            None => None,
+        // Only pass storage_slots if there are any non-zero (Some) values.
+        // This ensures is_storage_empty() works correctly by checking post_state_cursor is None.
+        let storage_slots_opt = if storage_slots.iter().any(|(_, v)| v.is_some()) {
```
mediocregopher: I don't think this is what you want. It's important to retain "none" values as part of the overlay (aka the post-state). If we are iterating over the DB and there exists a value at some key, but a "none" in the overlay for that key, we want to skip the DB value.

For supporting `is_storage_empty` we should do a check like this, but it should not affect the overlay which is used by the HashedPostStateCursor.

mediocregopher (commented):
Hey @0x00101010, great work on this so far, at a high level it looks correct, just a few hopefully smaller issues to work out.

crates/trie/common/src/hashed_state.rs
0x00101010: Generally just make it easier and consistent in how to handle none values, if I discard Option here, it's gonna be tricky to implement `HashedPostStateCursor` to handle both Account and Storage

crates/trie/common/src/hashed_state.rs
```diff
@@ -324,17 +324,8 @@ impl HashedPostState {
 
     /// Converts hashed post state into [`HashedPostStateSorted`].
     pub fn into_sorted(self) -> HashedPostStateSorted {
```
mediocregopher: You can replace the body of this method with just `self.drain_into_sorted()`, like was done with TrieUpdatesSorted, to reduce duplicate code

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,141 +25,184 @@ where
...
+    cursor_entry: Option<(B256, V)>,
+    /// Forward-only in-memory cursor over underlying V.
+    post_state_cursor: ForwardInMemoryCursor<'a, B256, Option<V>>,
+    /// The last hashed key that was returned by the cursor.
     /// De facto, this is a current cursor position.
-    last_account: Option<B256>,
+    last_key: Option<B256>,
+    /// Tracks whether `seek` has been called. Used to prevent re-seeking the DB cursor
+    /// when it has been exhausted by iteration.
+    seeked: bool,
```
mediocregopher, line 130: In the InMemoryTrieCursor this was included only for `cfg(debug_assertions)`, I would like to do that here as well.

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,141 +25,184 @@ where
...
+        }
     }
 
-    fn seek_inner(&mut self, key: B256) -> Result<Option<(B256, Account)>, DatabaseError> {
-        // Take the next account from the post state with the key greater than or equal to the
-        // sought key.
-        let post_state_entry = self.post_state_cursor.seek(&key);
+    /// Asserts that the next entry to be returned from the cursor is not previous to the last entry
+    /// returned.
+    fn set_last_key(&mut self, next_entry: &Option<(B256, V)>) {
```
mediocregopher, line 159: This is missing the debug_assert that InMemoryTrieOverlay has for this method

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,141 +25,184 @@ where
...
-        let post_state_entry = self.post_state_cursor.first_after(&last_account);
-
-        // If post state was given precedence or account was cleared, move the cursor forward.
-        let mut db_entry = self.cursor.seek(last_account)?;
-        while db_entry.as_ref().is_some_and(|(address, _)| {
-            address <= &last_account || self.is_account_cleared(address)
-        }) {
-            db_entry = self.cursor.next()?;
+    /// Seeks the `cursor_entry` field of the struct to the subsequent entry using the cursor.
+    fn cursor_next(&mut self) -> Result<(), DatabaseError> {
```
mediocregopher, line 181: missing debug_assert on `self.seeked`

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,141 +25,184 @@ where
...
     /// If the post state contains the exact match for the key, return it.
     /// Otherwise, retrieve the next entries that are greater than or equal to the key from the
     /// database and the post state. The two entries are compared and the lowest is returned.
     ///
-    /// The returned account key is memoized and the cursor remains positioned at that key until
-    /// [`HashedCursor::seek`] or [`HashedCursor::next`] are called.
+    /// The returned key is memoized and the cursor remains positioned at that key until
+    /// [`HashedCursor::next`] is called.
+    ///
+    /// Note: `seek` can be called multiple times by `TrieNodeIter`
```
mediocregopher: nit: I don't think we should be mentioning `TrieNodeIter` here, the previous version of the comment was fine

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -168,170 +211,52 @@ where
     /// If the cursor is positioned at the entry, return the entry with next greater key.
     /// Returns [None] if the previous memoized or the next greater entries are missing.
     ///
-    /// NOTE: This function will not return any entry unless [`HashedCursor::seek`] has been
-    /// called.
+    /// NOTE: This function will not return any entry unless [`HashedCursor::seek`] has been called.
```
mediocregopher, line 278: We should use the debug_assert on `self.seeked`, and this comment can be updated to indicate that it will panic

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -168,170 +211,52 @@ where
...
+        // Storage is not empty if it has non-zero slots.
+        if self.post_state_cursor.has_any(|(_, value)| value.is_some()) {
+            return Ok(false);
+        }
+
+        // If no non-zero slots in post state, check the database.
+        // Returns true if cursor is None (wiped storage or empty DB).
+        self.cursor.as_mut().map_or(Ok(true), |c| c.is_storage_empty())
     }
 }
```
mediocregopher, line 326: I'd like to see a proptest like was implemented for InMemoryTrieCursor 🙏

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -25,141 +25,184 @@ where
...
-    /// Reference to the collection of account keys that were destroyed.
-    destroyed_accounts: &'a B256Set,
-    /// The last hashed account that was returned by the cursor.
+pub struct HashedPostStateCursor<'a, C, V> {
+    /// The underlying `database_cursor`. If None then it is assumed there is no DB data.
+    cursor: Option<C>,
+    /// Entry that `database_cursor` is currently pointing to.
+    cursor_entry: Option<(B256, V)>,
+    /// Forward-only in-memory cursor over underlying V.
+    post_state_cursor: ForwardInMemoryCursor<'a, B256, Option<V>>,
```
mediocregopher: I would introduce a trait like:

```
trait HashedPostStateCursorValue<V> {
    fn as_option(&self) -> Option<&'_ V>
}
```

And implement that for both `Option<Account>` and `U256` here. Then that trait can be used in the ForwardInMemoryCursor instead of `Option<V>`. This way we can don't need to use `Option<U256>` on the HashedStorageSorted.

crates/trie/common/src/hashed_state.rs
mediocregopher: I think we can do it without duplicating significant code, see comment at : https://github.com/paradigmxyz/reth/pull/19233/files#r2481365676

crates/trie/trie/src/hashed_cursor/post_state.rs
0x00101010: There are relevant proptests there already:

https://github.com/paradigmxyz/reth/blob/780161a6472952704c5e1cc6d8f56fbda172d660/crates/trie/db/tests/post_state.rs#L178-L209

https://github.com/paradigmxyz/reth/blob/780161a6472952704c5e1cc6d8f56fbda172d660/crates/trie/db/tests/post_state.rs#L446-L490

crates/trie/trie/src/hashed_cursor/post_state.rs
0x00101010: This `seeked` field is needed here for this case (not only debug purposes):

1. cursor created
2. all in memory state are none or zero (deleted)
3. db has corresponding entries
4. multiple seeks are called 

Without `seeked`, we don't have a way to distinguish if db is exhausted (should not seek again) or there's no seek yet (should seek).

See `all_storage_slots_deleted_not_wiped_exact_keys()` in `tests/post_state.rs` for the specific test case. Without this, lots of integration tests will fail, for example `cargo test --test e2e_testsuite test_local_rpc_tests_compat --profile test --manifest-path /Users/francis/src/reth/main/crates/rpc/rpc-e2e-tests/Cargo.toml 2>&1`

Depending on if there should be multiple continuous `seek` calls, this potential bug exists on `InMemoryTrieCursor` as well.

Try test it with:
```
#[test]
fn test_all_storage_slots_deleted_not_wiped_exact_keys() {
    // This test reproduces an edge case where:
    // - cursor is not None (not wiped)
    // - All in-memory entries are deletions (None values)
    // - Database has corresponding entries
    // - Expected: NO leaves should be returned (all deleted)

    // Generate 42 trie node entries with keys distributed across the keyspace
    let db_nodes: Vec<(Nibbles, BranchNodeCompact)> = (0..42)
        .map(|i| {
            let key_bytes = vec![(i * 6) as u8, i as u8]; // Spread keys across keyspace
            let nibbles = Nibbles::from_nibbles_unchecked(key_bytes);
            (nibbles, BranchNodeCompact::new(i as u16, i as u16, 0, vec![], None))
        })
        .collect();

    // Create in-memory entries with same keys but all None values (deletions)
    let in_memory_nodes: Vec<(Nibbles, Option<BranchNodeCompact>)> =
        db_nodes.iter().map(|(key, _)| (*key, None)).collect();

    let db_nodes_map: BTreeMap<Nibbles, BranchNodeCompact> = db_nodes.into_iter().collect();
    let db_nodes_arc = Arc::new(db_nodes_map);
    let visited_keys = Arc::new(Mutex::new(Vec::new()));
    let mock_cursor = MockTrieCursor::new(db_nodes_arc, visited_keys);

    let mut cursor = InMemoryTrieCursor::new(Some(mock_cursor), &in_memory_nodes);

    // Seek to beginning should return None (all nodes are deleted)
    let result = cursor.seek(Nibbles::default()).unwrap();
    assert_eq!(
        result, None,
        "Expected no entries when all nodes are deleted, but got {:?}",
        result
    );

    // Test seek operations at various positions - all should return None
    let seek_keys = vec![
        Nibbles::from_nibbles([0x00]),
        Nibbles::from_nibbles([0x5d]),
        Nibbles::from_nibbles([0x5e]),
        Nibbles::from_nibbles([0x5f]),
        Nibbles::from_nibbles([0xc2]),
        Nibbles::from_nibbles([0xc5]),
        Nibbles::from_nibbles([0xc9]),
        Nibbles::from_nibbles([0xf0]),
    ];

    for seek_key in seek_keys {
        let result = cursor.seek(seek_key).unwrap();
        assert_eq!(
            result, None,
            "Expected None when seeking to {:?} but got {:?}",
            seek_key, result
        );
    }

    // next() should also always return None
    let result = cursor.next().unwrap();
    assert_eq!(result, None, "Expected None from next() but got {:?}", result);
}
```

crates/trie/trie/src/hashed_cursor/post_state.rs
0x00101010: See reason from the above comment thread

crates/trie/trie/src/hashed_cursor/post_state.rs
0x00101010: See above

crates/trie/trie/src/hashed_cursor/post_state.rs
0x00101010: done

crates/trie/trie/src/hashed_cursor/post_state.rs
mediocregopher: Thanks for the heads up here, I've submitted a PR for the InMemoryTrieCursor [here](https://github.com/paradigmxyz/reth/pull/19464).

I'm not 100% that it's a critical bug there, as it requires calling `seek` on the cursor even after the previous `seek` returned `None`, which doesn't really make sense to do. But it's worth handling at any rate. It's interesting that the same case for HashedPostStateCursor produces hive errors... I would guess there's some oddity in the TrieNodeIter that's causing it.

PR #19235, opened 2025-10-22
revert: "fix(engine): flatten storage cache (#18880)"

Rjected:
This reverts commit 169a1fb97b677731e2281b5e3ca5c372a1dcc616.

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -1,8 +1,5 @@
 //! Execution cache implementation for block processing.
-use alloy_primitives::{
-    map::{DefaultHashBuilder, HashSet},
-    Address, StorageKey, StorageValue, B256,
-};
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -302,70 +300,65 @@ pub(crate) struct ExecutionCache {
...
-        // NOTE: this must collect because the invalidate function should not be called while we
-        // hold an iter for it
-        let storage_entries = self
-            .storage_cache
-            .iter()
-            .filter_map(|entry| addresses.contains(&entry.key().0).then_some(*entry.key()))
-            .collect::<Vec<_>>();
-        for key in storage_entries {
-            self.storage_cache.invalidate(&key)
-        }
```
mattsse, line 368: note for future reference: we had to due this due to deadlock restrictions but this is problematic with lots of entries, especially on base

mattsse (approved):

PR #19237, opened 2025-10-22
chore: align env filter comment with configured directives

Fallengirl:
Updated the comment above DEFAULT_ENV_FILTER_DIRECTIVES in crates/tracing/src/layers.rs so that it now lists every dependency the filter array actually disables, keeping the documentation aligned with the code.


mattsse (approved):
sure why not

PR #19238, opened 2025-10-22
fix(node): remove unused ConsensusLayerHealthEvent variants

radik878:
Removed two enum variants that were never generated by the event stream:
  - HasNotBeenSeenForAWhile
  - NeverReceivedUpdates

  These variants were defined in the enum and had corresponding handlers
  in node.rs, but the poll_next implementation in cl.rs only generates
  NeverSeen and HaveNotReceivedUpdatesForAWhile events.

  The variants were likely remnants from an earlier implementation that
  tracked exchangeTransitionConfiguration (now deprecated post-merge) in
  addition to forkchoiceUpdated. Current implementation only tracks
  forkchoice updates via last_received_update_timestamp, making the
  additional variants unreachable dead code.

  Updated the outdated comment referencing "FCU and transition config
  exchange" to accurately reflect current behavior.

### DIFF
--- a/crates/node/events/src/cl.rs
+++ b/crates/node/events/src/cl.rs
@@ -61,7 +61,7 @@ impl<H: Send + Sync> Stream for ConsensusLayerHealthEvents<H> {
                 ))
             }
 
-            // We never had both FCU and transition config exchange.
+            // We never received any forkchoice updates.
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense

PR #19240, opened 2025-10-22
fix: hive tests consume test suite

jenpaff:
follow up from https://github.com/paradigmxyz/reth/pull/19230 

re-running hive test from branch before merging https://github.com/paradigmxyz/reth/actions/runs/18777666039

### DIFF
--- a/.github/assets/hive/build_simulators.sh
+++ b/.github/assets/hive/build_simulators.sh
@@ -11,7 +11,7 @@ go build .
 
 # Run each hive command in the background for each simulator and wait
 echo "Building images"
-./hive -client reth --sim "ethereum/eels" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz -sim.timelimit 1s || true &
+./hive -client reth --sim "ethereum/eels" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz --sim.buildarg branch=v5.3.0 -sim.timelimit 1s || true &
 ./hive -client reth --sim "ethereum/engine" -sim.timelimit 1s || true &
 ./hive -client reth --sim "devp2p" -sim.timelimit 1s || true &
 ./hive -client reth --sim "ethereum/rpc-compat" -sim.timelimit 1s || true &
--- a/.github/assets/hive/build_simulators.sh
+++ b/.github/assets/hive/build_simulators.sh
@@ -11,7 +11,7 @@ go build .
 
 # Run each hive command in the background for each simulator and wait
 echo "Building images"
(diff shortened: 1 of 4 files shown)


.github/assets/hive/build_simulators.sh
```diff
@@ -11,7 +11,7 @@ go build .
 
 # Run each hive command in the background for each simulator and wait
 echo "Building images"
-./hive -client reth --sim "ethereum/eest" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz --sim.buildarg branch=v5.3.0 -sim.timelimit 1s || true &
+./hive -client reth --sim "ethereum/eels" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz --sim.buildarg branch=v5.3.0 -sim.timelimit 1s || true &
```
shekhirin: They're migrating the repo, see https://github.com/ethereum/hive/pull/1358

So I think we should use a different branch here, maybe `forks/osaka`? Unsure what fixtures we should use.

jenpaff: removed

.github/assets/hive/build_simulators.sh
```diff
@@ -11,7 +11,7 @@ go build .
 
 # Run each hive command in the background for each simulator and wait
 echo "Building images"
-./hive -client reth --sim "ethereum/eest" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz --sim.buildarg branch=v5.3.0 -sim.timelimit 1s || true &
+./hive -client reth --sim "ethereum/eels" --sim.buildarg fixtures=https://github.com/ethereum/execution-spec-tests/releases/download/v5.3.0/fixtures_develop.tar.gz -sim.timelimit 1s || true &
```
shekhirin, line 15: we no longer pin to any branch, is this ok? @fgimenez

fgimenez: i believe it would be better to keep the branch pin, otherwise we would be open to failures due to unreleased changes landing on eels repo main branch

jenpaff: it was failing during the build step , but can take another look

fgimenez: yep, looks like tests are being migrated to https://github.com/ethereum/execution-specs and the simulator is built from this repo now in hive, but the tags are different from what we had in the old repo, and fixtures still come from the old one

just added a TODO comment to pin the image once we have proper tags and fixtures coming from the new repo

RomanHodulak: From the [readme](https://github.com/ethereum/execution-spec-tests/blob/main/README.md)

> Client Teams: Test fixture releases will continue from [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests) as usual. No changes to your consumption workflow.

fgimenez: thx! will update the TODO to reflect this

fgimenez (approved):

PR #19241, opened 2025-10-22
feat: add pruning of transactions from static-files

joshieDo:
* adds `Bodies` segment implementation
    * moves the pre-merge expiration logic from node launch here
    * generalizes `StaticFileProvider::delete_transactions_below` to `StaticFileProvider::delete_segment_below_block`
* adds `BodiesAdapter` wrapper over `Bodies` on `crates/prune/prune` while we don't decouple: https://github.com/paradigmxyz/reth/pull/19023
* `delete_jar` reports which `SegmentHeader` got deleted. This contains the block and transaction range.


crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -0,0 +1,44 @@
...
+use reth_prune_types::{PruneMode, PrunePurpose, PruneSegment, SegmentOutput};
+
+/// Segment adapter for pruning transactions in static files (bodies history).
+///
+/// This wraps the implementation from `reth_prune_static_files` and implements
+/// the `Segment` trait required by the pruner.
+#[derive(Debug)]
+pub struct BodiesAdapter {
+    inner: Bodies,
+}
```
joshieDo, line 17: temporary while we dont decouple

crates/node/core/src/args/pruning.rs
```diff
@@ -126,8 +126,10 @@ impl PruningArgs {
                     receipts: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
                     account_history: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
                     storage_history: Some(PruneMode::Distance(MINIMUM_PRUNING_DISTANCE)),
-                    // TODO: set default to pre-merge block if available
-                    bodies_history: None,
+                    bodies_history: chain_spec
+                        .ethereum_fork_activation(EthereumHardfork::Paris)
+                        .block_number()
+                        .map(PruneMode::Before),
```
mattsse, line 137: I'm not sure we should do this in this pr

I'd prefer a dedicated pr for this

crates/prune/static-files/src/bodies.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        if deleted_headers.is_empty() {
+            return Ok(SegmentOutput::done())
+        }
+
+        let tx_ranges = deleted_headers.iter().filter_map(|header| header.tx_range());
+
+        let pruned = tx_ranges
+            .clone()
+            .map(|range| range.end().saturating_sub(range.start()).saturating_add(1))
```
mattsse: I have a feeling that we need this again, can introduce a function for this, like `range.distance() + 1` or something?

crates/prune/static-files/src/bodies.rs
```diff
@@ -0,0 +1,75 @@
...
+            .sum::<u64>() as usize;
```
mattsse: I think this is just unwrap_or_default() ?

mattsse (changes requested):
tiny nit, other than that, checks out

crates/node/core/src/args/pruning.rs
shekhirin: I would rather have it in the same PR, because it removes the explicit `ctx.expire_pre_merge_transactions()` call, and we now rely on the pruner running on node startup with new bodies pruning.

shekhirin (approved):

crates/node/core/src/args/pruning.rs
mattsse: I see, makes sense

mattsse (approved):
lgtm!

PR #19242, opened 2025-10-22
chore: swap order for canon stream

mattsse:
we should create this first before submitting the tx so that we ensure we definitely receiving all blocks


klkvr (approved):
makes sense

RomanHodulak (approved):

PR #19243, opened 2025-10-22
feat(metrics): add push gateway support for Prometheus metrics

### LINKED ISSUE #19229: Support pushing Prometheus metrics to a remote URL
### Describe the feature

`prometheus` crate [supports](https://docs.rs/prometheus/latest/prometheus/#features) pushing metrics to a remote PushGateway. We need to add support for it via a new CLI argument.

### Additional context

_No response_

mablr:
Closes #19229

This spawns a task to push metrics to prometheus following configurable interval (default 5sec).
Added CLI flags for push gateway url and push interval.

### DIFF
--- a/crates/node/core/src/args/metric.rs
+++ b/crates/node/core/src/args/metric.rs
@@ -1,6 +1,6 @@
 use clap::Parser;
-use reth_cli_util::parse_socket_address;
-use std::net::SocketAddr;
+use reth_cli_util::{parse_duration_from_secs, parse_socket_address};
+use std::{net::SocketAddr, time::Duration};
 
 /// Metrics configuration.
 #[derive(Debug, Clone, Default, Parser)]
@@ -10,4 +10,22 @@ pub struct MetricArgs {
     /// The metrics will be served at the given interface and port.
     #[arg(long="metrics", alias = "metrics.prometheus", value_name = "PROMETHEUS", value_parser = parse_socket_address, help_heading = "Metrics")]
     pub prometheus: Option<SocketAddr>,
+
+    /// URL for pushing Prometheus metrics to a push gateway.
+    ///
+    /// If set, the node will periodically push metrics to the specified push gateway URL.
+    #[arg(long = "metrics.push.url", value_name = "PUSH_GATEWAY_URL", help_heading = "Metrics")]
+    pub push_gateway_url: Option<String>,
+
+    /// Interval in seconds for pushing metrics to push gateway.
+    ///
+    /// Default: 5 seconds
+    #[arg(
+        long = "metrics.push.interval",
+        default_value = "5",
+        value_parser = parse_duration_from_secs,
+        value_name = "SECONDS",
+        help_heading = "Metrics"
+    )]
+    pub push_gateway_interval: Duration,
 }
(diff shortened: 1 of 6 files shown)


mablr (commented):
I tested it locally like this :

```
docker run -d -p 9091:9091 prom/pushgateway

reth node --metrics=0.0.0.0:9001 --metrics.push.url=http://localhost:9091/metrics/job/test

curl http://localhost:9091/metrics
```

It seems to push correctly.

I'm wondering if it's relevant to add a test that spawns a HTTP server in order to inspect the PUT requests sent by the `MetricServer`. Wdyt ?

crates/node/metrics/src/server.rs
```diff
@@ -128,6 +162,59 @@ impl MetricServer {
...
+                loop {
+                    tokio::select! {
+                        _ = &mut signal => {
+                            tracing::info!("Shutting down task to push metrics to gateway");
+                            break;
+                        }
+                        _ = tokio::time::sleep(interval) => {
+                            hooks.iter().for_each(|hook| hook());
+                            let handle = install_prometheus_recorder();
+                            let metrics = handle.handle().render();
```
shekhirin: I think the `handle` initialization can be moved outside the loop body

crates/node/metrics/src/server.rs
```diff
@@ -128,6 +162,59 @@ impl MetricServer {
...
+                let client = reqwest::Client::builder()
+                    .build();
+
+                let client = match client {
+                    Ok(c) => c,
+                    Err(err) => {
+                        tracing::error!(%err, "Failed to create HTTP client to push metrics to gateway");
+                        return;
+                    }
+                };
```
shekhirin: let's initialize this outside the task, so that we can return an error from this method, and propagate it the same way as when calling `start_endpoint`

shekhirin (changes requested):
only two comments, this is basically what we need, yeah!

crates/node/core/src/args/metric.rs
```diff
@@ -10,4 +10,22 @@ pub struct MetricArgs {
...
+    ///
+    /// Default: 5 seconds
+    #[arg(
+        long = "metrics.push.interval",
+        default_value = "5",
+        value_parser = parse_duration_from_secs,
+        value_name = "SECONDS",
+        help_heading = "Metrics"
+    )]
+    pub push_gateway_interval: Duration,
```
shekhirin: one more thing: we may want to add pushing of OTLP traces in the future, so let's change the CLI argument names here
```suggestion
    #[arg(long = "metrics.prometheus.push.url", value_name = "PUSH_GATEWAY_URL", help_heading = "Metrics")]
    pub push_gateway_url: Option<String>,

    /// Interval in seconds for pushing metrics to push gateway.
    ///
    /// Default: 5 seconds
    #[arg(
        long = "metrics.prometheus.push.interval",
        default_value = "5",
        value_parser = parse_duration_from_secs,
        value_name = "SECONDS",
        help_heading = "Metrics"
    )]
    pub push_gateway_interval: Duration,
```

mablr: let's go !

shekhirin (approved):

mattsse (approved):
this lgtm

@Rjected likely has more context here

Rjected (approved):
This should work, although I think it would be good to figure out how to add config options to how we currently build our `PrometheusRecorder`, so we can use the builtin `push-gateway` support:
https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.with_push_gateway

If we were to implement auth then I would prefer using the builtin support as well

PR #19245, opened 2025-10-22
fix: return hashed peer key as id

klkvr:
Changes behavior to match Geth and return keccak of peer's public key (same as `PeerId`) in `admin_peers` endpoint

### DIFF
--- a/crates/rpc/rpc/src/admin.rs
+++ b/crates/rpc/rpc/src/admin.rs
@@ -14,6 +14,7 @@ use reth_network_types::PeerKind;
 use reth_rpc_api::AdminApiServer;
 use reth_rpc_server_types::ToRpcResult;
 use reth_transaction_pool::TransactionPool;
+use revm_primitives::keccak256;
 
 /// `admin` API implementation.
 ///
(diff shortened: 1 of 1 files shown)


shekhirin (approved):
I guess it's fine, but as far as I can tell, Nethermind also returns an address instead of a keccak of ID.

crates/rpc/rpc/src/admin.rs
```diff
@@ -74,34 +75,25 @@ where
...
-                        eth: Some(EthPeerInfo::Info(EthInfo {
-                            version: peer.status.version as u64,
-                        })),
-                        snap: None,
-                        other: Default::default(),
-                    },
-                })
-            }
+            infos.push(PeerInfo {
+                id: keccak256(peer.remote_id.as_slice()).to_string(),
```
mattsse, line 79: still unclear to me why we should return the hashed key here?

mattsse (approved):
if this makes it equivalent, I tink this is fine, tho unclear why

klkvr:
For context, relevant Geth code from RPC to the actual id calculation:
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/node/api.go#L307
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/server.go#L1056
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/peer.go#L570
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/peer.go#L152
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/enode/node.go#L165
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/enode/node.go#L58
https://github.com/ethereum/go-ethereum/blob/367b5fbe4211a01cfa15b3166b68b0d0dc5cecdc/p2p/enode/idscheme.go#L90

It does look like Nethermind returns just the public key

PR #19257, opened 2025-10-23
chore: remove unnecessary Clone derive from DummyInspector

SashaMalysehko:
removed unused Clone derive from DummyInspector. The inspector is constructed and passed by mutable reference into the EVM and later returned by value; no code path clones it, and the relevant EVM helper trait bounds for this flow do not require Clone. Keeping an unnecessary Clone derive increases surface area and can hide accidental copies in future refactors without benefit here.


mattsse (commented):
this is fine

PR #19259, opened 2025-10-23
feat(reth-bench): add file-based block loading for benchmarks (#19162)

### LINKED ISSUE #19162: Add Block Storage module with load functionality from file
### Describe the feature

Add Block Storage module with load functionality from file + Add CLI arguments for mode to load from file

We would wanna stream the blocks sequentially to avoid overloading the memory especially for 2000 blocks benches. We could probably do in batches of 50 - 100 blocks.
The latency differences is something that we would have to account for as well from the io of file. But its probably minimal.

### Additional context

_No response_

mackcee:
Resolves #19162 when complete

## Summary
Adds file-based block loading to reth-bench, allowing benchmarks to replay blocks from pre-generated files instead of fetching from an RPC.

## Changes
- Added `block-storage` command to generate block files from RPC
- Implemented `BlockFileReader` with configurable batch reading for memory efficiency
- Added `--from-file` flag to benchmark commands
- Refactored to use `BlockSource` enum for clean RPC/File separation
- Made auth provider optional (only required for actual benchmarking, not file generation)
- Added support for early optimism blocks (Engine API v1 not supported in Optimism)

## Usage
```bash
# Generate block file
reth-bench block-storage --rpc-url  --from X --to Y --output ./blocks

# Run benchmark from file
reth-bench new-payload-fcu --from-file ./blocks/blocks.bin --jwt-secret 
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7324,11 +7324,13 @@ dependencies = [
 name = "reth-bench"
 version = "1.8.2"
 dependencies = [
+ "alloy-consensus",
  "alloy-eips",
  "alloy-json-rpc",
  "alloy-primitives",
  "alloy-provider",
  "alloy-pubsub",
+ "alloy-rlp",
  "alloy-rpc-client",
  "alloy-rpc-types-engine",
  "alloy-transport",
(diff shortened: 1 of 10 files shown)


Charlie-Mack:
Closing as issue was closed and no longer required

PR #19260, opened 2025-10-23
chore: remove db pruning of header/txs segments

joshieDo:
Follow-up to https://github.com/paradigmxyz/reth/pull/19209. 

These pruner segments were related to moving data (headers/txs) from DB to static files. Since we now always write these structures to static files, pruning them from the database is no longer necessary.


crates/prune/prune/src/segments/user/transaction_lookup.rs
```diff
@@ -47,18 +52,26 @@ where
...
         // pre-merge history is dropped and then later tx lookup pruning is enabled) then we can
-        // only prune from the tx checkpoint and onwards.
-        if let Some(txs_checkpoint) = provider.get_prune_checkpoint(PruneSegment::Transactions)? &&
+        // only prune from the lowest static file.
+        if let Some(lowest_range) =
+            provider.static_file_provider().get_lowest_range(StaticFileSegment::Transactions) &&
             input
                 .previous_checkpoint
-                .is_none_or(|checkpoint| checkpoint.block_number < txs_checkpoint.block_number)
+                .is_none_or(|checkpoint| checkpoint.block_number < Some(lowest_range.start()))
```
joshieDo, line 60: Also, I'm pretty sure this also fixes a bug on `TransactionLookup` pruning. The code checked the `PruneSegment::Transactions` checkpoint and used it as a fallback starting point if the `TransactionLookup` checkpoint was behind.

However, the `Transactions` pruner (which we're removing) was responsible for moving transaction data from the database to static files - not for data expiration/deletion. This meant the `Transactions` checkpoint was always at or near the chain tip (since we don't actually store it in DB). Resulting in the `TransactionLookup` never being able to actually clean-up anything.

mattsse: I see, this makes sense

crates/prune/prune/src/segments/user/transaction_lookup.rs
```diff
@@ -47,18 +52,26 @@ where
...
-                transactions_checkpoint = ?input.previous_checkpoint,
-                "No TransactionLookup checkpoint found, using Transactions checkpoint as fallback"
-            );
+            let new_checkpoint = lowest_range.start().saturating_sub(1);
+            if let Some(body_indices) = provider.block_body_indices(new_checkpoint)? {
+                input.previous_checkpoint = Some(PruneCheckpoint {
+                    block_number: Some(new_checkpoint),
+                    tx_number: Some(body_indices.last_tx_num()),
+                    prune_mode: self.mode,
+                });
```
mattsse, line 68: makes sense

crates/cli/commands/src/stage/drop.rs
```diff
@@ -78,7 +78,6 @@ impl<C: ChainSpecParser> Command<C> {
             StageEnum::Bodies => {
                 tx.clear::<tables::BlockBodyIndices>()?;
                 tx.clear::<tables::Transactions<TxTy<N>>>()?;
-                reset_prune_checkpoint(tx, PruneSegment::Transactions)?;
```
mattsse, line 81: right, we no longer need this then,
are there any implications if this is removed?

mattsse (approved):
lgtm

we can also do the clear on launch separately, in case we decide to do that

crates/cli/commands/src/stage/drop.rs
joshieDo: i think txlookup was the only one

PR #19261, opened 2025-10-23
fix: infinite loop in eth_test_protocol_satellite test

sashaodessa:
**Problem:** The test contains an infinite loop that ignores stream termination, potentially causing CPU busy-looping when the stream ends.
**Solution:** Replace `loop { let _ = st.next().await; }` with `while st.next().await.is_some() {}` to properly exit when the stream terminates.
**Impact:** Prevents potential resource waste and improves test reliability by following proper `Stream` semantics.


mattsse (commented):
this is just for testing so we can keep this as is

PR #19262, opened 2025-10-23
docs: add safety comment for unsafe env::set_var usage

Lil-Duckling-22:
Adds a SAFETY comment explaining why the unsafe std::env::set_var call is safe in this context. The call happens in main() before any threads are spawned, making it safe despite the function being marked unsafe due to potential race conditions in multi-threaded environments.


mattsse (commented):
this isnt needed because convention after edition 2024

PR #19265, opened 2025-10-23
chore: rm `StaticFileReceipts` pruner

joshieDo:
Similar to https://github.com/paradigmxyz/reth/pull/19260, but for receipts.

This only removes the pruner segment that was being run after copying receipts from DB to static files.
```rust
//! - [`crate::segments::static_file::Receipts`] is responsible for pruning receipts on an archive
//!   node after static file producer has finished
```

The receipt pruner used for full node (eg. only keep unfinalized receipts) is still kept @ https://github.com/paradigmxyz/reth/blob/3883df3e6ca648a5d295b1dbb1daaa3d6be29030/crates/prune/prune/src/segments/user/receipts.rs


crates/prune/prune/src/segments/set.rs
```diff
@@ -58,7 +56,7 @@ where
     /// Creates a [`SegmentSet`] from an existing components, such as [`StaticFileProvider`] and
     /// [`PruneModes`].
     pub fn from_components(
-        static_file_provider: StaticFileProvider<Provider::Primitives>,
+        _static_file_provider: StaticFileProvider<Provider::Primitives>,
```
joshieDo, line 59: kept it for https://github.com/paradigmxyz/reth/pull/19241

mattsse (approved):
yeeeeeeeeeet

PR #19271, opened 2025-10-24
fix: no_std compatibility in reth-optimism-chainspec

strmfos:
Fix compilation failures when building without default features (no_std environment). The crate was using std::sync::Arc and std::str::FromStr imports which aren't available in no_std builds, causing tests to fail with "unresolved module" errors. Replace these with their core/alloc equivalents: use alloc::sync::Arc in test code, core::str::FromStr for parsing, and add missing ToString import.


mattsse (approved):
fair enough

Issue #3665, opened 2023-07-07
Tracking: discovery quality metrics


Rjected:
### Describe the feature

I was very impressed by https://github.com/ethereum/go-ethereum/pull/27621 and thought it would be a good idea for us to replicate some of these metrics. This tracks what we would need to add in reth, and what we would need to add in the existing grafana dashboard. Each of these tasks are likely small, so anyone who would like to take one should let me know, and I'll create an issue to track the individual task!

## Discovery

We would need an additional metric for nodes in kbuckets, and graph in grafana:
```[tasklist]
### Discovery Metrics
- [ ] `discover/bucket/{index}/count` - the number of nodes in bucket index
- [ ] Dashboard for the kbuckets metric
```

## `eth` handshake

We would need to create metrics for `eth` handshake errors that map to these:
```[tasklist]
### Eth protocol metrics
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/peer` - unexpected peer behavior
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/timeout` - handshake timeout
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/network` - wrong network id
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/version` - wrong protocol version
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/genesis` - wrong genesis
- [ ] `eth/protocols/eth/{ingress|egress}/handshake/error/forkid` - wrong forkid
```

## `p2p` dials

We already have some error metrics:
https://github.com/paradigmxyz/reth/blob/526f624e1cbfd659184873bffe12ec602678b57f/crates/net/network/src/metrics.rs#L64-L110

So we should make sure that the error metrics are in the grafana dashboard and named appropriately, and add the metrics which we don't already have:
```[tasklist]
### P2P dial metrics
- [ ] `p2p/dials/success` - the total number of dials that result in a peer being added to the peerset
- [ ] `p2p/dials/error/connection` - unable to initiate TCP connection to target
- [ ] `p2p/dials/error/saturated` - local client is already connected to its maximum number of peers
- [ ] `p2p/dials/error/known` - dialing an already connected peer
- [ ] `p2p/dials/error/self` - dialing the local node's id
- [ ] `p2p/dials/error/useless` - dialing a peer that doesn't share an capabilities with the local node
- [ ] `p2p/dials/error/id/unexpected` - dialed peer repsoned with different id than expected
[...]

PR #19274, opened 2025-10-24
feat(discv5): peers per kbucket metric

emhane:
Ref https://github.com/paradigmxyz/reth/issues/3665

- Adds metric for keeping track of how many peers are in each discv5 kbucket

todo: add intermediary struct for not holding read lock on kbuckets at same time as writing to gauges 
optional: keep track of lowest kbucket index with a peer to skip reading kbucket peer count for any buckets bellow it (lower indices are likely to be empty buckets)


mattsse (commented):
dont think we need this as a crate and proc macro

PR #19277, opened 2025-10-24
fix(trie): Rewrite InMemoryTrieOverlay (with proptests!)

mediocregopher:
The `InMemoryTrieOverlay::next` method was previously very broken, in that it did not ever descend into lower parts of the trie. For example a given current key of `0xA`, and subsequent keys of `0xA0` and `0xB`, `next` was returning `0xB`. This was due to incorrect usage of the `Nibbles::increment` method.

Fixing this issue involved essentially rewriting the entire type. The new implementation includes proptests to ensure that it functions correctly.

Some extra debug checks have also been added to ensure that the overlay is actually being used correctly.


crates/trie/trie/src/trie_cursor/in_memory.rs
```diff
@@ -132,50 +174,82 @@ impl<'a, C: TrieCursor> InMemoryTrieCursor<'a, C> {
...
+            self.in_memory_cursor.first_after(&last_key);
+        }
+
+        if let Some((key, _)) = &self.cursor_entry &&
+            key == &last_key
+        {
+            self.cursor_next()?;
+        }
+
+        let entry = self.next_inner()?;
```
Rjected: hmm, I wonder if there's a better name for `next_inner` because in this method, the cursors are mostly being advanced here by the `cursor_next` / `first_after` calls (except in the case of removed nodes).

Rjected (approved):
The methods seem equivalent to the old behavior, except now we always iterate in lexicographic order, the tests are also nice

crates/trie/trie/src/trie_cursor/in_memory.rs
```diff
@@ -83,47 +88,84 @@ impl<'a, C: TrieCursor> InMemoryTrieCursor<'a, C> {
...
+    }
+
+    /// Seeks the `cursor_entry` field of the struct using the cursor.
+    fn cursor_seek(&mut self, key: Nibbles) -> Result<(), DatabaseError> {
+        if let Some(entry) = self.cursor_entry.as_ref() &&
+            entry.0 >= key
+        {
+            // If already seeked to the given key then don't do anything. Also if we're seeked past
+            // the given key then don't anything, because `TrieCursor` is specifically a
+            // forward-only cursor.
```
shekhirin, line 121: I would guess seeking to a key less than the current key should not happen because it will return an incorrect result, can we add a debug assert for that here?

shekhirin (approved):

PR #19278, opened 2025-10-24
fix(txpool): correct propagate field name in Debug output

aso20455:
Fixed field name in ValidPoolTransaction Debug implementation. 
The field was displayed as "pragate" instead of the correct "propagate".


mattsse (approved):
fine

Issue #19107, opened 2025-10-17
feat(rpc): add configurable EVM memory limit


wojtek0x:
### Describe the feature

Currently, the only constraint on how much memory the EVM can allocate indirectly comes from the gas cap.
If the gas cap is set very high, it’s possible for a program to request more memory than the host system can provide. This can lead to Reth crashing with an error such as:
```
memory allocation of 30957566848569344 bytes failed
```

To prevent such out-of-memory crashes, Reth could support an explicit EVM memory limit configuration.

**Proposed Solution**
Introduce a command-line flag to cap the maximum memory the EVM can allocate per RPC request:
```
--rpc.evm-memory-limit <MB>
```

### Additional context

We run Reth with `--rpc.gas-cap` set to `uint64::MAX` to support complex EVM computations.
Occasionally Reth crashes due to buggy contracts that attempt invalid allocations (e.g., `MSTORE` with a huge offset). We haven't run into this problem with Geth, which appears to have additional memory limit check and returns `MemoryOOG` in such situation.

wojtek0x:
@mattsse @rakita I drafted two PRs for this, I'll appreciate a review:
* https://github.com/bluealloy/revm/pull/3125
* https://github.com/paradigmxyz/reth/pull/19279

PR #19279, opened 2025-10-24
feat: add --rpc.evm-memory-limit flag

wojtek0x:
This implements https://github.com/paradigmxyz/reth/issues/19107. It's a draft based on unmerged changes in `remv` - https://github.com/bluealloy/revm/pull/3125

Example usage:
```
reth node --rpc.gascap 18446744073709551615 --rpc.evm-memory-limit 10000000 --full --datadir /mnt/data/reth --authrpc.addr 127.0.0.1 --authrpc.port 8551 --http --http.addr 0.0.0.0 --http.port 8545 --http.api debug,eth,net,trace,txpool,web3,rpc,admin --chain mainnet --metrics 0.0.0.0:9001 --ws --ws.addr 0.0.0.0 --ws.port 3334 --ws.api debug,eth,net,trace,txpool,web3,rpc
```

```
curl http://127.0.0.1:8545/  -H 'Content-Type: application/json' -d '{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_call",
  "params": [
    {
      "from": "0x0000000000000000000000000000000000000001",
      "data": "0x600063009fffe05200",
      "value": "0x0"
    },
    "latest"
  ]
}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32003,"message":"out of memory: memory limit exceeded during memory expansion"}}
```


mattsse (commented):
awesome, clean, this lgtm

can merge after next revm bump

ref https://github.com/bluealloy/revm/pull/3129#pullrequestreview-3387936414

mattsse (approved):

crates/node/core/src/args/rpc_server.rs
```diff
@@ -188,6 +188,16 @@ pub struct RpcServerArgs {
     )]
     pub rpc_gas_cap: u64,
 
+    /// Maximum memory the EVM can allocate per RPC request.
+    #[arg(
+        long = "rpc.evm-memory-limit",
+        alias = "rpc-evm-memory-limit",
+        value_name = "MEMORY_LIMIT",
+        value_parser = MaxOr::new(RangedU64ValueParser::<u64>::new().range(1..)),
+        default_value_t = u64::MAX
```
mattsse: @rakita the revm default for this is (1 << 32) - 1

should we use that here?

rakita: u64::MAX would panic for big allocations, so it definetly needs to be changed.

4gb ( `1 <<32-1`) is an okay number, but it depends on machine they are running.

rakita: Very roughly, 10M gas can max the memory to around 10mb, 1 gigagas can max it around 1gb.

mattsse:
awesome, ty

I gave up earlier when github was choking (again)

mattsse (approved):

PR #19280, opened 2025-10-24
fix(trie): Fix trie_reverts not returning sorted nodes

mediocregopher:
As a more future-proof fix, I've also made the fields of the TrieNodesSorted struct private, and introduced a constructor which will do a debug_assert that the input is already sorted.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2077,7 +2077,7 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> TrieWriter for DatabaseProvider
         let mut account_trie_cursor = tx.cursor_write::<tables::AccountsTrie>()?;
 
         // Process sorted account nodes
-        for (key, updated_node) in &trie_updates.account_nodes {
+        for (key, updated_node) in trie_updates.account_nodes_ref() {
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2302,7 +2305,7 @@ impl<TX: DbTx + 'static, N: NodeTypes> TrieReader for DatabaseProvider<TX, N> {
                 .push((nibbles.0, node_value));
         }
 
-        Ok(trie_updates)
+        Ok(TrieUpdatesSorted::new(account_nodes, storage_tries))
```
mediocregopher, line 2309: This case does not require explicit sorting, as the vecs are constructed in sorted order inherently thanks to the DB

crates/storage/provider/src/providers/database/provider.rs
Rjected: maybe this comment should be updated to reflect that what we want to return is a list sorted by path, not sorted by block and path within each block

Rjected (approved):
makes sense to me, just have a doc nit

DaniPopes (approved):

yongkangc (approved):

mattsse (approved):
makes sense

PR #19284, opened 2025-10-24
chore(codecs): replace todo with unimplemented in Compact derive

reallesee:
Replaced `todo!()` with `unimplemented!()` for unit structs and unions in the Compact derive macro.

These aren't planned features - they're design limitations. Unit structs have no fields to serialize, and unions can't be safely handled by derive macros. Better semantics and clearer error messages for users who try to derive Compact on unsupported types.


mattsse (approved):
fine, this message is better and makes it easier to identify where the panic is coming from

PR #19286, opened 2025-10-24
fix(node): classify connect_async failures as WebSocket and use Url parse error

phrwlk:
The connect() function incorrectly mapped websocket handshake/transport failures from tokio-tungstenite connect_async into EthStatsError::InvalidUrl, which conflated network errors with URL format issues. This patch reclassifies those errors to EthStatsError::WebSocket and uses the existing Url(ParseError) variant for Url::parse failures. It also avoids unnecessary URL string conversion by passing url.as_str() to connect_async. This ensures accurate error semantics consistent with the error enum design and improves debuggability.


mattsse (approved):
this is cleaner, I like it

PR #19287, opened 2025-10-24
fix(fs): correct ReadLink error message and add missing read_link wra…

gap-editor:
My PR fixes a small inconsistency in the `fs` utilities module:

**Fixed**: the `FsPathError::ReadLink` message was identical to `Read` ("failed to read from …"), which made it hard to distinguish link read failures in logs.
  Now it clearly says *“failed to read link …”*.
**Added**: the missing `read_link()` wrapper for `std::fs::read_link`, providing consistent API coverage and error context like other helpers (`read()`, `write()`, `read_dir()`).

### DIFF
--- a/crates/fs-util/src/lib.rs
+++ b/crates/fs-util/src/lib.rs
@@ -39,7 +39,7 @@ pub enum FsPathError {
     },
 
     /// Error variant for failed read link operation with additional path context.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this seems appropriate even tho we don't use it, just in case we want to extract all of this at some point

PR #19288, opened 2025-10-24
docs: fix duplicate timer comment in pinger documentation

crStiv:
Corrected timeout_timer comment in pinger docs to distinguish it from ping_interval timer


mattsse (commented):
these docs are pretty much outdated not worth updating

PR #19293, opened 2025-10-25
chore(net): fix misleading comment for decompressed message size check

JacksonEi:
Comment said we check compressed length but the variable is `decompressed_len`, so fixed the comment to match.


mattsse (commented):
is fine but not worth fixing

PR #19294, opened 2025-10-25
chore: remove unused deprecated type aliases

sky-coderay:
This removes three deprecated type aliases that are no longer used anywhere in the codebase:
- RecoveredTx<T>
- TransactionSignedEcRecovered<T>  
- PooledTransactionsElementEcRecovered<T>

These types were deprecated in #13882 (Jan 20) as part of the migration to alloy::Recovered. The last actual usage was removed back in #13690 (Jan 6), which even removed a "TODO: remove this foreign type" comment from the code. They've just been sitting around as dead exports since then.

Since everything now uses alloy::Recovered directly, keeping these deprecated aliases just adds noise and might confuse new contributors who could think they should still be using them.

Also removed an unused import of `Recovered` in transaction/mod.rs that was only needed for these type aliases.

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -41,16 +41,10 @@ pub use alloy_consensus::{
     ReceiptWithBloom,
 };
 
-/// Recovered transaction
-#[deprecated(note = "use `Recovered` instead")]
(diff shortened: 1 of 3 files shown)


mattsse (commented):
we can keep this until we wipe this crate

Issue #18936, opened 2025-10-10
Avoid cloning the transaction in payload processor


shekhirin:
### Describe the feature

We can avoid this clone by wrapping the `tx` into `Arc` and sending it to both prewarm and execution channels https://github.com/paradigmxyz/reth/blob/397a30defbc7d577dd31dfbc269ad764a5e00449/crates/engine/tree/src/tree/payload_processor/mod.rs#L322

This should work, because none of the transaction traits functions expect an owned value.

0xKarl98:
Yeah sure , but ATM i have taken another issue .

If i can't allocate the time slot properly , will tell

0xKarl98:
@shekhirin I can take it , as the issue i said didn't assign to me and it is gone 
So now i can take this one :)

PR #19295, opened 2025-10-25
perf: wrap tx with Arc to avoid deep cloning

0xKarl98:
This is on issue #18936 , @shekhirin  @mattsse  @yongkangc  Please check if the requests are met


crates/evm/evm/src/execute.rs
```diff
@@ -457,7 +457,43 @@ where
     }
 
     fn into_recovered(self) -> Recovered<Executor::Transaction> {
-        self.tx.into_recovered()
+        // Try to unwrap the Arc, or clone if there are multiple references
+        match Arc::try_unwrap(self.tx) {
+            Ok(tx) => tx.into_recovered(),
+            Err(arc_tx) => (*arc_tx).clone().into_recovered(),
+        }
+    }
```
mattsse: I think we can use `Arc::unwrap_or_clone` here

crates/evm/evm/src/execute.rs
```diff
@@ -641,16 +677,16 @@ pub struct WithTxEnv<TxEnv, T> {
...
     fn tx(&self) -> &Tx {
-        self.tx.tx()
+        self.tx.as_ref().tx()
     }
 
     fn signer(&self) -> &Address {
-        self.tx.signer()
+        self.tx.as_ref().signer()
     }
 }
```
mattsse, line 654: @klkvr should we impl these traits for Arc ?

crates/evm/evm/src/execute.rs
```diff
@@ -457,7 +457,43 @@ where
...
+        match Self::try_unwrap(self) {
+            Ok(with_tx_env) => {
+                // Extract the Arc<T> and try to unwrap it
+                match Arc::try_unwrap(with_tx_env.tx) {
+                    Ok(tx) => tx.into_recovered(),
+                    Err(arc_tx) => {
+                        // If there are multiple references to the tx, clone it
+                        (*arc_tx).clone().into_recovered()
+                    }
+                }
```
mattsse: hmm, do we need these two arc checks even?

crates/evm/evm/src/engine.rs
```diff
@@ -23,14 +23,14 @@ pub trait ExecutableTxIterator<Evm: ConfigureEvm>:
     Iterator<Item = Result<Self::Tx, Self::Error>> + Send + 'static
 {
     /// The executable transaction type iterator yields.
-    type Tx: ExecutableTxFor<Evm> + Clone + Send + 'static;
+    type Tx: ExecutableTxFor<Evm> + Clone + Send + Sync + 'static;
```
mattsse, line 26: I see we need to add sync so that we can arc it

crates/evm/evm/src/execute.rs
```diff
@@ -457,7 +457,43 @@ where
...
+        // Try to unwrap the Arc, or clone if there are multiple references
+        match Arc::try_unwrap(self.tx) {
+            Ok(tx) => tx.into_recovered(),
+            Err(arc_tx) => (*arc_tx).clone().into_recovered(),
+        }
+    }
+}
+
+impl<T, Executor> ExecutorTx<Executor>
+    for Arc<WithTxEnv<<<Executor as BlockExecutor>::Evm as Evm>::Tx, T>>
```
mattsse: hmm, why do we need to impl this for `Arc<WithTxEnv` dont see where we're using WithTxEnv in an arc

but perhaps this is even better?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -326,7 +331,7 @@ where
         let (execute_tx, execute_rx) = mpsc::channel();
         self.executor.spawn_blocking(move || {
             for tx in transactions {
-                let tx = tx.map(|tx| WithTxEnv { tx_env: tx.to_tx_env(), tx });
+                let tx = tx.map(|tx| WithTxEnv { tx_env: tx.to_tx_env(), tx: Arc::new(tx) });
```
mattsse, line 329: since we're cloning `tx`, perhaps it actually makes more sense to wrap just `WithTxEnv` here?

mattsse (changes requested):
left some comments

wdyt @klkvr

crates/engine/tree/src/tree/payload_processor/mod.rs
klkvr: we'll still end up cloning and taking ownership of `tx_env` when using it so it would be the same perf I believe

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -197,45 +199,52 @@ where
...
-        P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
-        F: DatabaseProviderROFactory<Provider: TrieCursorFactory + HashedCursorFactory>
+        P: DatabaseProviderFactory<Provider: BlockReader>
+            + BlockReader
+            + StateProviderFactory
+            + StateReader
             + Clone
-            + Send
             + 'static,
+        I::Tx: Sync,
```
klkvr:
```suggestion
```

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -287,6 +296,7 @@ where
     ) -> PayloadHandle<WithTxEnv<TxEnvFor<Evm>, I::Tx>, I::Error>
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
+        I::Tx: Sync,
```
klkvr:
```suggestion
```

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -308,7 +318,10 @@ where
         mpsc::Receiver<WithTxEnv<TxEnvFor<Evm>, I::Tx>>,
         mpsc::Receiver<Result<WithTxEnv<TxEnvFor<Evm>, I::Tx>, I::Error>>,
         usize,
-    ) {
+    )
+    where
+        I::Tx: Sync,
+    {
```
klkvr:
```suggestion
    ) {
```

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -44,7 +43,10 @@ use tracing::{debug, debug_span, instrument, trace, warn};
 
 /// A wrapper for transactions that includes their index in the block.
 #[derive(Clone)]
-struct IndexedTransaction<Tx> {
+struct IndexedTransaction<Tx>
+where
+    Tx: Clone,
+{
```
klkvr:
```suggestion
struct IndexedTransaction<Tx> {
```

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -462,7 +464,7 @@ where
         sender: Sender<PrewarmTaskEvent>,
         done_tx: Sender<()>,
     ) where
-        Tx: ExecutableTxFor<Evm>,
+        Tx: ExecutableTxFor<Evm> + Clone,
```
klkvr: why do we need the `Clone` bound here?

crates/evm/evm/src/execute.rs
klkvr: yeah I don't think we should need this impl. `ExecutorTx` is only used during block building where we need ownership over both tx and tx env, so `Arc`'ing is not useful for that codepath

klkvr (commented):
directionally lgtm, just some comments on redundant bounds

crates/engine/tree/src/tree/payload_processor/mod.rs
0xKarl98: Done

crates/engine/tree/src/tree/payload_processor/mod.rs
0xKarl98: Done , thank you :)

crates/engine/tree/src/tree/payload_processor/prewarm.rs
0xKarl98: Because we need to implement Clone trait for Tx here  , will remove it once i have done remove the double Arc , as it looks weird to mattese

crates/evm/evm/src/engine.rs
0xKarl98: This already add Sync to it :)

shekhirin (approved):
LGTM, deferring to @mattsse or @klkvr here

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -318,10 +326,16 @@ where
...
+                let tx = tx.map(|tx| WithTxEnv { tx_env: tx.to_tx_env(), tx: Arc::new(tx) });
                 // only send Ok(_) variants to prewarming task
                 if let Ok(tx) = &tx {
-                    let _ = prewarm_tx.send(tx.clone());
+                    // Avoid deep cloning by creating a new WithTxEnv with Arc::clone for efficient
+                    // sharing
+                    let prewarm_tx_env = WithTxEnv {
+                        tx_env: tx.tx_env.clone(), // Clone tx_env only when necessary
+                        tx: Arc::clone(&tx.tx),    // Efficient Arc shallow clone
+                    };
```
mattsse, line 337: wait, I believe this is the same as 
```
                    let _ = prewarm_tx.send(tx.clone());
```
?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -184,7 +182,13 @@ where
...
                     // where the first deposit transaction contains essential block metadata.
                     for handle in &handles {
-                        if let Err(err) = handle.send(indexed_tx.clone()) {
+                        // Create a new IndexedTransaction with the same index and a cloned tx
+                        // Since Tx is WithTxEnv<TxEnvFor<Evm>, Arc<I::Tx>>, cloning is efficient
+                        let tx_for_worker = IndexedTransaction {
+                            index: indexed_tx.index,
+                            tx: indexed_tx.tx.clone(),
+                        };
+                        if let Err(err) = handle.send(tx_for_worker) {
```
mattsse, line 191: I think we can leave this unchanged?

crates/evm/evm/src/execute.rs
```diff
@@ -457,7 +457,11 @@ where
     }
 
     fn into_recovered(self) -> Recovered<Executor::Transaction> {
-        self.tx.into_recovered()
+        // Avoid deep cloning when Arc has only one reference
+        match Arc::try_unwrap(self.tx) {
+            Ok(tx) => tx.into_recovered(), // No cloning needed - direct ownership transfer
+            Err(arc_tx) => (*arc_tx).clone().into_recovered(), // Clone only when necessary
```
mattsse:
```suggestion
Arc::unwrap_or_clone(self.tx).into_recovered()
```

mattsse (changes requested):
cool, last nits

crates/engine/tree/src/tree/payload_processor/prewarm.rs
0xKarl98: Sure , i just suspects it would violates the rules that we avoid cloning when send transactions to channels  . Otherwise , we could keep it the same

crates/evm/evm/src/execute.rs
0xKarl98: Changed

PR #19296, opened 2025-10-25
fix(engine): module doc to reflect schnellru::LruMap backend

VolodymyrBg:
- Replaced an outdated reference to “moka” with the correct schnellru::LruMap in the module-level documentation of 
crates/engine/tree/src/tree/precompile_cache.rs
- The implementation imports and uses schnellru::LruMap (with ByLength limiter) and parking_lot::Mutex; other comments in the file already describe LruMap semantics.
- This change ensures the documentation accurately reflects the current implementation. No functional code changes.


mattsse (approved):
let this one slide, because actually a valid doc fix

PR #19299, opened 2025-10-25
chore(ethereum): remove redundant std::default::Default import

Galoretka:
Removed the unnecessary std::default::Default import. Default is available via the Rust prelude, so the explicit import was redundant


yongkangc (approved):

mattsse (approved):
fine

PR #19301, opened 2025-10-26
fix: highest_nonces update in PendingPool::remove_transaction

Forostovec:
Previously, PendingPool::remove_transaction unconditionally wrote ancestor(id) into highest_nonces after any removal, which incorrectly downgraded the recorded highest nonce when removing a non-maximum transaction and could also drop the highest entry incorrectly during cascade removals if the immediate ancestor no longer existed. The fix only updates highest_nonces when the removed transaction was the current highest for that sender and then recomputes the new highest from the remaining transactions of that sender in by_id, preserving correct invariants and ensuring remove_to_limit and related logic operate on true per-sender maxima.

### DIFF
--- a/crates/transaction-pool/src/pool/pending.rs
+++ b/crates/transaction-pool/src/pool/pending.rs
@@ -334,22 +334,23 @@ impl<T: TransactionOrdering> PendingPool<T> {
         let tx = self.by_id.remove(id)?;
         self.size_of -= tx.transaction.size();
 
-        let removed_was_highest = match self.highest_nonces.get(&id.sender) {
-            Some(highest) => highest.transaction.nonce() == id.nonce,
-            None => false,
-        };
-        if removed_was_highest {
-            if let Some((_, new_highest)) = self
-                .by_id
-                .range((id.sender.start_bound(), Unbounded))
-                .take_while(|(other, _)| other.sender == id.sender)
-                .last()
-            {
-                self.highest_nonces.insert(id.sender, new_highest.clone());
-            } else {
-                self.highest_nonces.remove(&id.sender);
+        if let Some(highest) = self.highest_nonces.get(&id.sender) {
+            if highest.transaction.nonce() == id.nonce {
+                if let Some((_, new_highest)) = self
+                    .by_id
+                    .range((
+                        id.sender.start_bound(),
+                        std::ops::Bound::Included(TransactionId::new(id.sender, u64::MAX)),
+                    ))
+                    .last()
+                {
+                    self.highest_nonces.insert(id.sender, new_highest.clone());
+                } else {
+                    self.highest_nonces.remove(&id.sender);
+                }
             }
         }
+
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/pending.rs
```diff
@@ -342,13 +334,21 @@ impl<T: TransactionOrdering> PendingPool<T> {
         let tx = self.by_id.remove(id)?;
         self.size_of -= tx.transaction.size();
 
-        if let Some(highest) = self.highest_nonces.get(&id.sender) {
-            if highest.transaction.nonce() == id.nonce {
```
mattsse, line 346: I'd like to keep if let logic here

crates/transaction-pool/src/pool/pending.rs
```diff
@@ -342,13 +334,21 @@ impl<T: TransactionOrdering> PendingPool<T> {
...
+        let removed_was_highest = match self.highest_nonces.get(&id.sender) {
+            Some(highest) => highest.transaction.nonce() == id.nonce,
+            None => false,
+        };
+        if removed_was_highest {
+            if let Some((_, new_highest)) = self
+                .by_id
+                .range((id.sender.start_bound(), Unbounded))
+                .take_while(|(other, _)| other.sender == id.sender)
+                .last()
```
mattsse: we can improve this slightly if we instead use the highest id as the upper threshold, then we don't then the take_while and can use the range's last() fn

mattsse (changes requested):
this makes a lot of sense, ty!

left two nits

crates/transaction-pool/src/pool/pending.rs
Forostovec: > we can improve this slightly if we instead use the highest id as the upper threshold, then we don't then the take_while and can use the range's last() fn

made corrections

crates/transaction-pool/src/pool/pending.rs
```diff
@@ -342,14 +334,23 @@ impl<T: TransactionOrdering> PendingPool<T> {
...
+                .range((
+                    id.sender.start_bound(),
+                    std::ops::Bound::Included(TransactionId::new(id.sender, u64::MAX)),
+                ))
+                .last()
+            {
+                self.highest_nonces.insert(id.sender, new_highest.clone());
+            } else {
                 self.highest_nonces.remove(&id.sender);
             }
```
mattsse: I don't think this is correct, because this would remove the highest entry if the removed tx isnt the highest?

mattsse (changes requested):

crates/transaction-pool/src/pool/pending.rs
mattsse: nvm, this is correct

mattsse (approved):

Rjected (approved):
makes sense to me

Issue #17177, opened 2025-07-02
feat: snap sync stage


rkrasiuk:
## Description

Create new `SnapSyncStage` that will be responsible for querying peers for ranges of trie data and inserting it into the database. Leave the healing algorithm and database inserts of trie nodes as placeholders for now. If enabled, this stage should _replace_ `SenderRecoveryStage`, `ExecutionStage` and `PruneSenderRecoveryStage`.

The stage should subscribe to the stream of head headers from the consensus engine to be able to update the target state root in snap peer requests.

The stage should adhere to the following algorithm:
1. Retrieve the latest header from the engine
2. Check if the hashed state in `tables::HashedAccounts` is empty:
  a. empty - start downloading ranges from `0x0000...` account hash
  b. not empty - start downloading ranges from the last entry in the database
3. Paginate over trie ranges using `GetAccountRange` request (leave storage ranges for the follow up)
  a. If no data is returned for current target state root, return to 1. 
4. Repeat 1. to 3. until the state is retrieved (final range until account `0xffff...` is fetched)

chimera-defi:
Hi there, I made this with some AI help, really as more of a learning exercise but would love to get a review/feedback/guidance in case it is helpful towards adding snapsync. 
https://github.com/paradigmxyz/reth/pull/18865

Feel free to close it if its way off.   
Would be nice to get snapsync since its a bit hard to test reth in prod currently with the slow and storage heavy sync that exists right now.

PR #19303, opened 2025-10-26
refactor: add more Snap response types 

guha-rahul:
tracks #17177 
- currently snapClient methods return `Self::Output`, which is hardcoded to `Future<Output = PeerRequestResult<AccountRangeMessage>>`
- added an enum which contains all the Response types for snap sync requests


mattsse (approved):
makes sense

PR #19304, opened 2025-10-26
feat(jovian/block-validation): fix block validation for jovian

theochap:
## Description

In Jovian, the `blob_gas_used` field has been repurposed to hold the DA footprint block limit value computed during block execution. We should change consensus rules to diverge from the L1 behavior from Jovian onwards.

Added tests to increment coverage on this behavior.

### DIFF
--- a/crates/optimism/consensus/src/lib.rs
+++ b/crates/optimism/consensus/src/lib.rs
@@ -65,7 +65,7 @@ where
         block: &RecoveredBlock<N::Block>,
         result: &BlockExecutionResult<N::Receipt>,
     ) -> Result<(), ConsensusError> {
-        validate_block_post_execution(block.header(), &self.chain_spec, &result.receipts)
+        validate_block_post_execution(block.header(), &self.chain_spec, result)
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/optimism/consensus/src/lib.rs
```diff
@@ -111,7 +111,13 @@ where
...
-        if self.chain_spec.is_ecotone_active_at_timestamp(block.timestamp()) {
+        // Blob gas used validation
+        // In Jovian, the blob gas used computation has changed. We are moving the blob base fee
+        // validation to post-execution since the DA footprint calculation is stateful.
+        // Pre-execution we only validate that the blob gas used is present in the header.
+        if self.chain_spec.is_jovian_active_at_timestamp(block.timestamp()) {
+            block.blob_gas_used().ok_or(ConsensusError::BlobGasUsedMissing)?;
+        } else if self.chain_spec.is_ecotone_active_at_timestamp(block.timestamp()) {
             validate_cancun_gas(block)?;
         }
```
mattsse, line 122: this else if looks correct to me

because this does

https://github.com/paradigmxyz/reth/blob/159ff01cd266172eec0a79d48aa2f85b69ffb1c4/crates/consensus/common/src/validation.rs#L80-L92

mattsse (approved):
this makes sense to me

pending @klkvr @emhane

emhane (approved):
lgtm. will open debt issue to move that blob gas used check into a new jovian model, not a prio now though in order to unblock.

PR #19305, opened 2025-10-26
feat(reth-optimism-node): Add OP E2E mineblock test with isthmus activated at genesis

### LINKED ISSUE #15447: Add OP E2E mineblock test with isthmus activated at genesis
### Describe the feature

we have this for OP:

https://github.com/paradigmxyz/reth/blob/0c9dd9a27ddeb53f13a3430007c136aff9596e14/crates/optimism/node/tests/e2e/testsuite.rs#L13-L21

but this suses a chainspec that doesn't enable isthmus at genesis

## TODO
* duplicate this test but modify the chainspec so that isthmus is also activated at genesis

we can use this as base:

https://github.com/paradigmxyz/reth/blob/0c9dd9a27ddeb53f13a3430007c136aff9596e14/crates/optimism/node/tests/e2e/testsuite.rs#L19-L21

but enable isthmus at genesis
or duplicate the genesis.json and enable isthmus there

### Additional context

_No response_

mablr:
Closes #15447

- duplicated `test_testsuite_op_assert_mine_block` but with isthmus activated at genesis
- workaround to handle both v2/v3 payloads (tell me if you'd like another design/approach)

### DIFF
--- a/crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
+++ b/crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
@@ -98,31 +98,66 @@ where
                 finalized_block_hash: parent_hash,
             };
 
-            let fcu_result = EngineApiClient::<Engine>::fork_choice_updated_v2(
+            // Try v2 first for backwards compatibility, fall back to v3 on error.
+            match EngineApiClient::<Engine>::fork_choice_updated_v2(
                 &engine_client,
                 fork_choice_state,
                 Some(self.payload_attributes.clone()),
             )
-            .await?;
-
-            debug!("FCU result: {:?}", fcu_result);
-
-            // check if we got a valid payload ID
-            match fcu_result.payload_status.status {
-                PayloadStatusEnum::Valid => {
-                    if let Some(payload_id) = fcu_result.payload_id {
-                        debug!("Got payload ID: {payload_id}");
-
-                        // get the payload that was built
-                        let _engine_payload =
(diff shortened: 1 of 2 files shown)


crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
```diff
@@ -98,31 +98,66 @@ where
...
-
-            // check if we got a valid payload ID
-            match fcu_result.payload_status.status {
-                PayloadStatusEnum::Valid => {
-                    if let Some(payload_id) = fcu_result.payload_id {
-                        debug!("Got payload ID: {payload_id}");
+            .await
+            {
+                Ok(fcu_result) => {
+                    debug!("FCU v2 result: {:?}", fcu_result);
```
emhane:
```suggestion
                    debug!(
                        ?fcu_result,
                        "FCU v2 result"
                    );
```
we use this kv-pair format for logs. see the `tracing` docs

emhane (changes requested):
nice! smol comment on log format

crates/e2e-test-utils/src/testsuite/actions/produce_blocks.rs
mablr: Damn, I made the same mistake as last time. I swear I'll keep it in mind this time.😅

emhane (approved):
lgtm. pending @mattsse

mattsse (approved):
ty frens

PR #19308, opened 2025-10-27
refactor: cleanup abstractions in multiproofs

### LINKED ISSUE #19182: refactor: clean up multiproof code for better abstractions

yongkangc:
Refactors multiproof computation code to close #19182

### Changes

1. Optimize Storage Proof Hot Path
- Introduced `ProofResult` enum to eliminate unnecessary allocations when returning storage proofs
- Storage proofs now returned directly instead of being wrapped in full `DecodedMultiProof` structure

2. Remove Dead Code
- `StorageMultiproofInput` (never constructed)
- `PendingMultiproofTask` (unnecessary wrapper)
- `AccountMultiproofResult` (unused alias)

3. Replace Unreachable Error Handling
- Changed defensive error returns to `unreachable!()` and `debug_assert_eq!()` for impossible conditions with storage proofs. 

4. Code Organization
- Reorganized `proof_task.rs` and `multiproof.rs` for logical flow. i.e main types at the start

### Future PR
- I wanted to limit the scope of this PR to just these 3 files. I have noticed another issue which is wrapping and unwrapping of `DecodedMultiProof`. This is a redudant abstraction that we should remove. tracking:
https://github.com/paradigmxyz/reth/issues/19309

### Testing
- Running on reth4 to test the refactor

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -72,87 +72,18 @@ impl MultiproofManager {
     }
 
     /// Dispatches a new multiproof calculation to worker pools.
-    fn dispatch(&mut self, input: PendingMultiproofTask) {
+    fn dispatch(&mut self, input: MultiproofInput) {
         // If there are no proof targets, we can just send an empty multiproof back immediately
-        if input.proof_targets_is_empty() {
+        if input.proof_targets.is_empty() {
             debug!(
-                sequence_number = input.proof_sequence_number(),
+                sequence_number = input.proof_sequence_number,
                 "No proof targets, sending empty multiproof back immediately"
             );
             input.send_empty_proof();
             return
         }
 
-        match input {
-            PendingMultiproofTask::Storage(storage_input) => {
-                self.dispatch_storage_proof(storage_input);
-            }
-            PendingMultiproofTask::Regular(multiproof_input) => {
-                self.dispatch_multiproof(multiproof_input);
-            }
-        }
-    }
-
-    /// Dispatches a single storage proof calculation to worker pool.
-    fn dispatch_storage_proof(&mut self, storage_multiproof_input: StorageMultiproofInput) {
-        let StorageMultiproofInput {
-            hashed_state_update,
-            hashed_address,
-            proof_targets,
-            proof_sequence_number,
-            multi_added_removed_keys,
-            state_root_message_sender: _,
-        } = storage_multiproof_input;
-
-        let storage_targets = proof_targets.len();
-
-        trace!(
-            target: "engine::tree::payload_processor::multiproof",
-            proof_sequence_number,
-            ?proof_targets,
-            storage_targets,
-            "Dispatching storage proof to workers"
-        );
-
-        let start = Instant::now();
-
-        // Create prefix set from targets
-        let prefix_set = reth_trie::prefix_set::PrefixSetMut::from(
-            proof_targets.iter().map(reth_trie::Nibbles::unpack),
-        );
-        let prefix_set = prefix_set.freeze();
-
-        // Build computation input (data only)
-        let input = reth_trie_parallel::proof_task::StorageProofInput::new(
-            hashed_address,
-            prefix_set,
-            proof_targets,
-            true, // with_branch_node_masks
-            Some(multi_added_removed_keys),
-        );
-
-        // Dispatch to storage worker
-        if let Err(e) = self.proof_worker_handle.dispatch_storage_proof(
-            input,
-            reth_trie_parallel::proof_task::ProofResultContext::new(
-                self.proof_result_tx.clone(),
-                proof_sequence_number,
-                hashed_state_update,
-                start,
-            ),
-        ) {
-            error!(target: "engine::tree::payload_processor::multiproof", ?e, "Failed to dispatch storage proof");
-            return;
-        }
-
-        self.inflight += 1;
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1,11 +1,7 @@
 //! Multiproof task related functionality.
```
yongkangc, line 1: The module in multiproof.rs is now organised as 
```
//! This module is organized as follows:
//! - Main Types: [MultiproofManager], [MultiProofTask]
//! - Metrics: [MultiProofTaskMetrics]
//! - Supporting Types: [SparseTrieUpdate], [MultiProofConfig], [MultiProofMessage],
//!   [StateHookSender]
//! - Internal Types: [ProofSequencer], [MultiProofTaskState], [MultiproofInput]
//! - Helper Functions: [evm_state_to_hashed_post_state], [get_proof_targets]
```

crates/trie/parallel/src/proof_task.rs
```diff
@@ -77,1255 +77,1290 @@ use std::{
...
 use tokio::runtime::Handle;
-use tracing::{debug, debug_span, error, trace};
+use tracing::{debug_span, error, trace};
 
 #[cfg(feature = "metrics")]
 use crate::proof_task_metrics::ProofTaskTrieMetrics;
 
 type StorageProofResult = Result<DecodedStorageMultiProof, ParallelStateRootError>;
 type TrieNodeProviderResult = Result<Option<RevealedNode>, SparseTrieError>;
-type AccountMultiproofResult =
```
yongkangc, line 87: removed `AccountMultiproofResult` as it is redundant. Both account and storage results now come as `ProofResult`, with the account case carried by `ProofResult::AccountMultiproof(multiproof, stats)`.

crates/trie/parallel/src/proof_task.rs
```diff
@@ -77,1255 +77,1300 @@ use std::{
...
+                                    )));
+                                }
+                                proof
+                            }
+                            ProofResult::AccountMultiproof(..) => {
+                                return Err(ParallelStateRootError::Other(
+                                    "expected storage proof, got account multiproof".to_string(),
+                                ));
+                            }
+                        };
```
yongkangc, line 1159: i think the errors here shouldnt be likely. However logging it here jic

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -24,287 +20,6 @@ use reth_trie_parallel::{
...
-}
-
-/// A pending multiproof task, either [`StorageMultiproofInput`] or [`MultiproofInput`].
-#[derive(Debug)]
-enum PendingMultiproofTask {
-    /// A storage multiproof task input.
-    Storage(StorageMultiproofInput),
-    /// A regular multiproof task input.
-    Regular(MultiproofInput),
-}
```
yongkangc, line 225: removed this 

After we removed the old storage/account enum and always pass a single MultiproofInput value into `MultiproofManager::dispatch`, the alias was redundant. 

Removing it lets thecall sites work directly with MultiproofInput, without  indirection

PR #19311, opened 2025-10-27
refactor(trie): Unify proof return types

yongkangc:
Part of #19182 cleanup effort - split into smaller PRs for easier review.

Introduces a new `ProofResult` enum that unifies the return types for proof calculations:

## Changes

- Introduced `ProofResult` enum with `AccountMultiproof` and `StorageProof` variants
- Added `into_multiproof()` helper method for conversion when needed
- Removed unused `AccountMultiproofResult` type alias

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -360,8 +360,7 @@ impl MultiproofManager {
     }
 
     /// Dispatches a new multiproof calculation to worker pools.
-    fn dispatch(&mut self, input: impl Into<PendingMultiproofTask>) {
-        let input = input.into();
+    fn dispatch(&mut self, input: PendingMultiproofTask) {
         // If there are no proof targets, we can just send an empty multiproof back immediately
         if input.proof_targets_is_empty() {
             debug!(
@@ -760,7 +759,7 @@ impl MultiProofTask {
                 self.multiproof_manager.proof_worker_handle.has_available_storage_workers();
 
         let mut dispatch = |proof_targets| {
-            self.multiproof_manager.dispatch(MultiproofInput {
+            self.multiproof_manager.dispatch(PendingMultiproofTask::Regular(MultiproofInput {
                 config: self.config.clone(),
                 source: None,
                 hashed_state_update: Default::default(),
@@ -768,7 +767,7 @@ impl MultiProofTask {
                 proof_sequence_number: self.proof_sequencer.next_sequence(),
                 state_root_message_sender: self.tx.clone(),
                 multi_added_removed_keys: Some(multi_added_removed_keys.clone()),
-            });
+            }));
             chunks += 1;
         };
 
@@ -905,7 +904,7 @@ impl MultiProofTask {
             );
             spawned_proof_targets.extend_ref(&proof_targets);
 
-            self.multiproof_manager.dispatch(MultiproofInput {
+            self.multiproof_manager.dispatch(PendingMultiproofTask::Regular(MultiproofInput {
                 config: self.config.clone(),
                 source: Some(source),
                 hashed_state_update,
@@ -913,7 +912,7 @@ impl MultiProofTask {
                 proof_sequence_number: self.proof_sequencer.next_sequence(),
                 state_root_message_sender: self.tx.clone(),
                 multi_added_removed_keys: Some(multi_added_removed_keys.clone()),
-            });
+            }));
 
             chunks += 1;
         };
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -759,15 +759,18 @@ impl MultiProofTask {
                 self.multiproof_manager.proof_worker_handle.has_available_storage_workers();
 
         let mut dispatch = |proof_targets| {
-            self.multiproof_manager.dispatch(PendingMultiproofTask::Regular(MultiproofInput {
-                config: self.config.clone(),
-                source: None,
-                hashed_state_update: Default::default(),
-                proof_targets,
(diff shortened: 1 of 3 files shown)


crates/trie/parallel/src/proof_task.rs
```diff
@@ -84,8 +84,35 @@ use crate::proof_task_metrics::ProofTaskTrieMetrics;
 
 type StorageProofResult = Result<DecodedStorageMultiProof, ParallelStateRootError>;
 type TrieNodeProviderResult = Result<Option<RevealedNode>, SparseTrieError>;
-type AccountMultiproofResult =
```
yongkangc, line 87: removed `AccountMultiproofResult` as it is redundant. Both account and storage results now come as ProofResult, with the account case carried by ProofResult::AccountMultiproof(multiproof, stats).

crates/trie/parallel/src/proof_task.rs
```diff
@@ -657,14 +676,20 @@ where
                             )
                         })?;
 
-                        // Extract storage proof from the multiproof wrapper
-                        let (mut multiproof, _stats) = proof_msg.result?;
-                        let proof =
-                            multiproof.storages.remove(&hashed_address).ok_or_else(|| {
```
yongkangc, line 663: This was the key issue pointed out as footgun. where we are doing potentially double conversion in the hot path.

mattsse (commented):
imo this makes sense but would like another review from @shekhirin

mediocregopher (approved):
LGTM

crates/trie/parallel/src/proof_task.rs
```diff
@@ -84,8 +84,35 @@ use crate::proof_task_metrics::ProofTaskTrieMetrics;
...
 type StorageProofResult = Result<DecodedStorageMultiProof, ParallelStateRootError>;
 type TrieNodeProviderResult = Result<Option<RevealedNode>, SparseTrieError>;
-type AccountMultiproofResult =
-    Result<(DecodedMultiProof, ParallelTrieStats), ParallelStateRootError>;
+
+/// Result of a proof calculation, which can be either an account multiproof or a storage proof.
+#[derive(Debug)]
+pub enum ProofResult {
+    /// Account multiproof with statistics
+    AccountMultiproof(DecodedMultiProof, ParallelTrieStats),
```
shekhirin: can we make these named fields, similar to `StorageProof`?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -716,9 +741,9 @@ where
     // Consume remaining storage proof receivers for accounts not encountered during trie walk.
     for (hashed_address, receiver) in storage_proof_receivers {
         if let Ok(proof_msg) = receiver.recv() {
-            // Extract storage proof from the multiproof wrapper
-            if let Ok((mut multiproof, _stats)) = proof_msg.result &&
-                let Some(proof) = multiproof.storages.remove(&hashed_address)
+            // Extract storage proof from the result
+            if let Ok(ProofResult::StorageProof { hashed_address: addr, proof }) = proof_msg.result &&
+                addr == hashed_address
```
shekhirin: we didn't have this check before, unsure if we need it now?

shekhirin (changes requested):
We need to preserve the functionality for dispatching storage-only proofs, even though it's not wired in yet. The idea is that we should be able to skip the account multiproof generation altogether, if we don't need the account proof (it has already been fetched), and only fetch the storage multiproof using [`StorageProof::storage_multiproof`](https://github.com/paradigmxyz/reth/blob/be73e4a246717ea51b5f0e8bef62ea73027cd17d/crates/trie/trie/src/proof/mod.rs#L297).

Also small nits.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -440,18 +456,6 @@ impl MultiproofManager {
             .record(self.proof_worker_handle.pending_account_tasks() as f64);
     }
 
-    /// Signals that a multiproof calculation has finished.
-    fn on_calculation_complete(&mut self) {
```
yongkangc, line 444: moved down

crates/trie/parallel/src/proof_task.rs
yongkangc: addressed this here: [34e790b](https://github.com/paradigmxyz/reth/pull/19311/commits/34e790b95a10bb43087f298421cd5de071a38390)

crates/trie/parallel/src/proof_task.rs
yongkangc: yeah i dont think we need it

removed here:

[1624209](https://github.com/paradigmxyz/reth/pull/19311/commits/1624209606a1fa288f03cad135cedcaf1fae1edc)

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -756,18 +759,15 @@ impl MultiProofTask {
...
-                    source: None,
-                    hashed_state_update: Default::default(),
-                    proof_targets,
-                    proof_sequence_number: self.proof_sequencer.next_sequence(),
-                    state_root_message_sender: self.tx.clone(),
-                    multi_added_removed_keys: Some(multi_added_removed_keys.clone()),
-                }
-                .into(),
-            );
+            self.multiproof_manager.dispatch(PendingMultiproofTask::Regular(MultiproofInput {
```
shekhirin: can we do `.into()` here like before?

yongkangc: initially wanted it to be more explicit. so we dont miss that in the future. 

but added back here
[379cadf](https://github.com/paradigmxyz/reth/pull/19311/commits/379cadfcc4f395592ccca5a655a723d46ec3e9bc)

shekhirin (approved):

PR #19316, opened 2025-10-27
chore: remove redundant PhantomData from NodeHooks

sashass1315:
- Removed _marker: PhantomData<Node> field and its initialization in NodeHooks::new() in crates/node/builder/src/hooks.rs

Rationale: 
Node and AddOns are already present in the trait object fields, so the marker was redundant and provided no additional dropck, variance, or auto-trait benefits. This simplifies the struct without functional changes.


mattsse (approved):
seems, okay, we dont need to guard this against deconstructions imo

PR #19319, opened 2025-10-27
docs(eth-wire): update docs to reflect eth-wire-types, alloy_rlp, version-aware decoding, and RLPx multiplexing

radik878:
- Types moved: Document that protocol message types live in crates/net/eth-wire-types/ and are re-exported by eth-wire.
- RLP crate: Replace internal rlp references with alloy_rlp and derive macros.
- Decoding API: Note ProtocolMessage decodes via decode_message(EthVersion, &mut bytes) instead of decode().
- Messages/IDs: Update EthMessage<N> to include NewPooledTransactionHashes66/68, Receipts69, BlockRangeUpdate (0x11).
- P2PStream: Switch to SharedCapabilities, correct ID mask/normalize using MAX_RESERVED_MESSAGE_ID + 1.
- Pinger: Fix comments and use timeout_timer.poll(cx) for WaitingForPong.
- EthStream: Describe version-aware decoding and prohibit Status after handshake (disconnect on breach); mention start_send_broadcast/start_send_raw.
- Handshakes: P2P uses HelloMessageWithProtocols; ETH uses UnifiedStatus + ForkFilter with timeout.
- Multiplexing: Add RlpxProtocolMultiplexer/RlpxSatelliteStream and role of SharedCapabilities.
- Links: Use relative repo paths to current files to prevent link rot.

Reason: Prior docs reflected older structure/APIs (types in eth-wire, internal rlp, non-versioned decoding, single capability offset).
This aligns documentation with the implemented behavior across Eth66–69 and current module layout.

### DIFF
--- a/docs/crates/eth-wire.md
+++ b/docs/crates/eth-wire.md
@@ -9,48 +9,70 @@ This crate can be thought of as having 2 components:
 2. Abstractions over Tokio Streams that operate on these types.
 
 (Note that ECIES is implemented in a separate `reth-ecies` crate.)
+Additionally, this crate focuses on stream implementations (P2P and Eth), handshakes, and multiplexing. The protocol
+message types and RLP encoding/decoding live in the separate `eth-wire-types` crate and are re-exported by `eth-wire`
+for convenience.
 ## Types
 The most basic Eth-wire type is a `ProtocolMessage`. It describes all messages that reth can send/receive.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
these docs aren't great but this is def an improvement because the existing ones are super outdated

PR #19320, opened 2025-10-27
Fix Vec Capacity in random_block_range

wedjob0X:
precompute the exact block count for RangeInclusive to avoid repeated reallocations
keep existing behaviour; cargo test -p reth-testing-utils passes


mattsse (commented):
this is only test code so this is fine

PR #19323, opened 2025-10-27
fix: improve variable naming clarity in full block downloader

viktorking7:
## Description

This PR improves code readability by renaming local variables in the full block downloader to better reflect their actual purpose and ordering.

### Changes
- Rename `headers_falling` → `headers_desc` 
- Rename `headers_rising` → `headers_asc` 

### Why
The previous names (`falling`/`rising`) were metaphorical and ambiguous. The new names explicitly indicate the sort order (descending/ascending), making the code's intent clearer and reducing cognitive load for future maintainers.


mattsse (commented):
falling is better here

PR #19325, opened 2025-10-27
chore: remove dead OpL1BlockInfo.number field and writes

phrwlk:
The OpL1BlockInfo.number (AtomicU64) field was write-only and had no
getter; it was not read anywhere in this crate or the repository.
Validation logic relies on timestamps for interop activation and L1 data
gas fee computation, so the block number is unnecessary.

This change removes the field and its stores in new() and
update_l1_block_info() while preserving timestamp behavior.


mattsse (approved):
makes sense because all op hardforks are activated by timestmap so we dont need this

PR #19326, opened 2025-10-27
perf(tree): only chunk multiproof targets if needed

shekhirin:
Doesn't have any visible improvement, but should be an overall positive change because we will not go through the chunking iterator anymore for state updates that don't need chunking.

Additionally, adds a `chunks` field to the spans, with `0` value meaning that no chunking has been applied.

```
Statistics:
Mean percent difference: 0.27%
Median percent difference: -0.29%
Standard deviation: 13.33%
Min: -35.05%
Max: 69.82%
Total blocks analyzed: 1000
```

image


mediocregopher (approved):

mattsse (approved):
makes sense

PR #19331, opened 2025-10-27
fix(op-reth): use latest for runtime image

ghost:
otw glibc linking made during builder wont be there, and binary wont launch

in general consider use tags for builder/runtime in base images

### DIFF
--- a/DockerfileOp
+++ b/DockerfileOp
@@ -31,7 +31,7 @@ RUN cargo build --profile $BUILD_PROFILE --features "$FEATURES" --bin op-reth --
 RUN ls -la /app/target/$BUILD_PROFILE/op-reth
 RUN cp /app/target/$BUILD_PROFILE/op-reth /app/op-reth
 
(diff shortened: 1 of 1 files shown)


DockerfileOp
```diff
@@ -31,7 +31,7 @@ RUN cargo build --profile $BUILD_PROFILE --features "$FEATURES" --bin op-reth --
 RUN ls -la /app/target/$BUILD_PROFILE/op-reth
 RUN cp /app/target/$BUILD_PROFILE/op-reth /app/op-reth
 
-FROM ubuntu:22.04 AS runtime
+FROM ubuntu AS runtime
```
mattsse, line 34: we have the same here https://github.com/paradigmxyz/reth/blob/ffeaa4772d01313062b0b127eb4dcc980550aec0/Dockerfile#L44-L44

mattsse (approved):

PR #19334, opened 2025-10-27
chore: remove dead scenario history types and field from MockTransactionSimulator

phrwlk:
Remove unused types Scenario, ExecutedScenario, ExecutedScenarios and the executed field from MockTransactionSimulator in crates/transaction-pool/src/test_utils/pool.rs
These entities are not referenced anywhere in the repository. The simulator does not record scenario history and ScenarioType::HigherNonce is unimplemented, indicating the history feature was never completed


mattsse (commented):
we can keep this, doesnt hurt

PR #19338, opened 2025-10-28
fix(op-reth/consensus): fixes header validation for jovian. decouple excess blob gas and blob gas used

theochap:
## Description

This PR follows #19304. It fixes header validation post-Jovian by decoupling the `excess_blob_gas` and the `blob_gas_used` header fields. 

I added unit tests to ensure this part of the code is well covered.

### DIFF
--- a/crates/optimism/consensus/src/lib.rs
+++ b/crates/optimism/consensus/src/lib.rs
@@ -18,9 +18,9 @@ use core::fmt::Debug;
 use reth_chainspec::EthChainSpec;
 use reth_consensus::{Consensus, ConsensusError, FullConsensus, HeaderValidator};
 use reth_consensus_common::validation::{
-    validate_against_parent_4844, validate_against_parent_eip1559_base_fee,
-    validate_against_parent_hash_number, validate_against_parent_timestamp, validate_cancun_gas,
-    validate_header_base_fee, validate_header_extra_data, validate_header_gas,
+    validate_against_parent_eip1559_base_fee, validate_against_parent_hash_number,
+    validate_against_parent_timestamp, validate_cancun_gas, validate_header_base_fee,
+    validate_header_extra_data, validate_header_gas,
 };
 use reth_execution_types::BlockExecutionResult;
 use reth_optimism_forks::OpHardforks;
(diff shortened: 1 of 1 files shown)


crates/optimism/consensus/src/lib.rs
```diff
@@ -188,9 +188,32 @@ where
             &self.chain_spec,
         )?;
 
-        // ensure that the blob gas fields for this block
-        if let Some(blob_params) = self.chain_spec.blob_params_at_timestamp(header.timestamp()) {
-            validate_against_parent_4844(header.header(), parent.header(), blob_params)?;
```
mattsse, line 193: I see we don't really need the parent validation

crates/optimism/consensus/src/lib.rs
```diff
@@ -188,9 +188,32 @@ where
...
+                blob_gas_used != 0
+            {
+                return Err(ConsensusError::BlobGasUsedDiff(GotExpected {
+                    got: blob_gas_used,
+                    expected: 0,
+                }));
+            }
+
+            let excess_blob_gas =
+                header.excess_blob_gas().ok_or(ConsensusError::ExcessBlobGasMissing)?;
```
mattsse, line 209: I see, excess_blob_gas must _always_ be 0 after ecotone

crates/optimism/consensus/src/lib.rs
```diff
@@ -188,9 +188,32 @@ where
...
+        // Ensure that the blob gas fields for this block are correctly set.
+        // In the op-stack, the excess blob gas is always 0 for all blocks after ecotone.
+        // The blob gas used and the excess blob gas should both be set after ecotone.
+        // After Jovian, the blob gas used contains the current DA footprint.
+        if self.chain_spec.is_ecotone_active_at_timestamp(header.timestamp()) {
+            let blob_gas_used = header.blob_gas_used().ok_or(ConsensusError::BlobGasUsedMissing)?;
+
+            // Before Jovian and after ecotone, the blob gas used should be 0.
+            if !self.chain_spec.is_jovian_active_at_timestamp(header.timestamp()) &&
+                blob_gas_used != 0
```
mattsse, line 200: I see, after jovian the blob gas used value isnt validated, right?

mattsse (approved):
lgtm

pending @emhane or @klkvr for sanity check

emhane (approved):
fine. though would be nice if we can move these checks into `ecotone` and `jovian` modules returning errors that are specific to OP protocol instead e.g. `MissingBlobGasPostEcotone`. can do this as follow up issue though assuming unit testing has enough coverage...could also be good to do the clean up before fork though and then not touch the logic after the fork as it's then considered tested code running in production on mainnet.

crates/optimism/consensus/src/lib.rs
theochap: Blob gas used value is validated in the `validate_post_execution`, here: https://github.com/theochap/reth/blob/f5886251786ce9bf80c824725a2f2fc9dd809d0a/crates/optimism/consensus/src/validation/mod.rs#L88-L106

PR #19340, opened 2025-10-28
perf(engine): return sorted data from compute_trie_input

### LINKED ISSUE #19249: Return sorted data from compute_trie_input
### Describe the feature

`compute_trie_input` is used to compute the in-memory overlay which will be passed to the MultiProofTask so that proof computation can take into account in-memory blocks which haven't been persisted to the database yet.

https://github.com/paradigmxyz/reth/blob/4548209e7b00351a993779820b78db86b8628c9b/crates/engine/tree/src/tree/payload_validator.rs#L999-L1006

`compute_trie_input` returns a TrieInput which contains HashedPostState and TrieUpdates, both unsorted data structures.

https://github.com/paradigmxyz/reth/blob/4548209e7b00351a993779820b78db86b8628c9b/crates/trie/common/src/input.rs#L5-L15

The `TrieInput` then gets converted into a MultiProofConfig, which is almost exactly the same but uses HashedPostStateSorted and TrieUpdatesSorted.

https://github.com/paradigmxyz/reth/blob/4548209e7b00351a993779820b78db86b8628c9b/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L66-L78

This conversion uses a few milliseconds per block:

Image

-----

We should introduce a TrieInputSorted which contains the sorted forms of the fields, and which can therefore be converted into a MultiProofConfig for free. This type should be returned from `compute_trie_input`.

Ideally we would want to eliminate all places in the hot path where we are converting to/from the sorted trie input data. This will mean modifying ExecutedBlock to hold HashedPostStateSorted and TrieUpdatesSorted as well, and possibly other unforeseen changes.

### Additional context

_No response_

yongkangc:
Closes #19249

Eliminates sorting overhead per block by returning `TrieInputSorted` instead of unsorted `TrieInput` from `compute_trie_input`, and have `ExecutedBlock` utilise the `TrieInputSorted`

Previously, `MultiProofConfig::from_input()` would call `drain_into_sorted()` on both nodes and state every block, performing expensive sorting operations:


crates/trie/common/src/input.rs
```diff
@@ -118,4 +122,75 @@ impl TrieInput {
...
+        }
+    }
+}
+
+/// Sorted variant of [`TrieInput`] for efficient proof generation.
+///
+/// This type holds sorted versions of trie data structures, which eliminates the need
+/// for expensive sorting operations during multiproof generation.
+#[derive(Default, Debug, Clone)]
+pub struct TrieInputSorted {
```
yongkangc, line 147: we should write some test for this

crates/trie/common/src/input.rs
```diff
@@ -118,4 +122,75 @@ impl TrieInput {
...
+    /// Converts trie input into [`TrieInputSorted`], draining the internal maps while keeping
+    /// their allocated capacity.
+    ///
+    /// This effectively clears all the fields in the [`TrieInput`] while preserving the
+    /// `HashMap` capacity for reuse. This allows us to reuse the allocated space for subsequent
+    /// block validations, avoiding repeated allocations and rehashing in hot paths.
+    ///
+    /// The sorted output allocates new `Vec` space, but the original `HashMap` capacity is
+    /// retained for the next cycle.
+    pub fn drain_into_sorted(&mut self) -> TrieInputSorted {
```
yongkangc: do we really need this?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -802,15 +802,15 @@ where
         match strategy {
             StateRootStrategy::StateRootTask => {
                 // get allocated trie input if it exists
-                let allocated_trie_input = self.trie_input.take();
+                let allocated = self.trie_input.take();
```
yongkangc: lets not change the name for this

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -982,9 +987,10 @@ where
         parent_hash: B256,
         state: &EngineApiTreeState<N>,
         allocated_trie_input: Option<TrieInput>,
-    ) -> ProviderResult<(TrieInput, BlockNumber)> {
-        // get allocated trie input or use a default trie input
+    ) -> ProviderResult<(TrieInputSorted, TrieInput, BlockNumber)> {
```
yongkangc: why do we still need `TrieInput`? cant we remove that?

crates/trie/common/src/input.rs
```diff
@@ -118,4 +122,75 @@ impl TrieInput {
...
+            nodes: input.nodes.into_sorted(),
+            state: input.state.into_sorted(),
+            prefix_sets: input.prefix_sets,
+        }
+    }
+
+    /// Append state to the input by reference and extend the prefix sets.
+    pub fn append_ref(&mut self, state: &HashedPostState) {
+        self.prefix_sets.extend(state.construct_prefix_sets());
+        let sorted_state = state.clone().into_sorted();
```
yongkangc: so basically we are applying sorting here, would that be a concern for performance regression?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -820,8 +820,13 @@ where
 
                 // Convert the TrieInput into a MultProofConfig, since everything uses the sorted
                 // forms of the state/trie fields.
-                let (trie_input, multiproof_config) = MultiProofConfig::from_input(trie_input);
-                self.trie_input.replace(trie_input);
+                let (mut cleared_sorted_input, multiproof_config) =
```
mediocregopher: The conversion to MultiProofConfig is no longer really necessary, you can use the `trie_input` fields directly when constructing the OverlaystateProviderFactory

crates/engine/tree/src/tree/payload_validator.rs
mediocregopher: Definitely, yes, `allocated_trie_input` should be a TrieInputSorted

crates/trie/common/src/input.rs
mediocregopher: imo it's not needed

crates/trie/common/src/input.rs
mediocregopher: if we're not using this method in this PR then I don't think we should add it. This logic made sense on TrieInput when we were not able to revert trie data, but now it's not necessary I don't think

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: yeah this makes alot of sense, thanks. yeeted

crates/trie/common/src/input.rs
yongkangc: makes sense, but currently we are using it here
https://github.com/paradigmxyz/reth/pull/19340/files#diff-6b5414154ff2b0a9b6cb854533d6b0a6024e0c1877742b0fa6d8b4d2e790cf82L646-L649

yongkangc: I feel like theres alot more code that i can get rid off, currently still looking through / understanding the flow to see how it can be simpler.

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: removed :)

crates/trie/common/src/input.rs
mediocregopher: I see, in this case I would have `append_ref` at least accept a `&HashedPostStateSorted`, and do the `into_sorted` outside, so it might be possible in the future to eliminate that too

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1759,8 +1759,8 @@ where
         Ok(Some(ExecutedBlock {
             recovered_block: Arc::new(RecoveredBlock::new_sealed(block, senders)),
             execution_output: Arc::new(execution_output),
-            hashed_state: Arc::new(hashed_state),
-            trie_updates: Arc::new(trie_updates.into()),
+            hashed_state: Arc::new(hashed_state.into_sorted()),
```
mediocregopher, line 1797: Even if we still have an `into_sorted` here, it's better that it's hashed state because that's usually an order of magnitude smaller than trie updates 👍

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -912,16 +897,19 @@ where
...
     /// It works as follows:
     /// 1. Collect in-memory blocks that are descendants of the provided parent hash using
     ///    [`crate::tree::TreeState::blocks_by_hash`]. This returns the highest persisted ancestor
     ///    hash (`block_hash`) and the list of in-memory descendant blocks.
-    /// 2. Extend the `TrieInput` with the contents of these in-memory blocks (from oldest to
+    /// 2. Extend the `TrieInputSorted` with the contents of these in-memory blocks (from oldest to
     ///    newest) to build the overlay state and trie updates that sit on top of the database view
     ///    anchored at `block_hash`.
+    ///
+    /// The in-memory blocks are already sorted, so we can directly append them without expensive
```
mediocregopher: nit: not sure this comment belongs here, or that it makes much sense anyway

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -946,10 +930,24 @@ where
...
-        );
+        // Extend with contents of parent in-memory blocks directly in sorted form.
+        let mut input = TrieInputSorted::default();
+        let mut blocks_iter = blocks.iter().rev();
+
+        if let Some(first) = blocks_iter.next() {
+            input.state = Arc::clone(&first.hashed_state);
+            input.nodes = Arc::clone(&first.trie_updates);
+
+            // Only clone and mutate if there are multiple in-memory blocks.
```
mediocregopher: Nice

crates/optimism/payload/src/builder.rs
```diff
@@ -389,8 +389,8 @@ impl<Txs> OpBuilder<'_, Txs> {
         let executed: ExecutedBlock<N> = ExecutedBlock {
             recovered_block: Arc::new(block),
             execution_output: Arc::new(execution_outcome),
-            hashed_state: Arc::new(hashed_state),
-            trie_updates: Arc::new(trie_updates),
+            hashed_state: Arc::new(hashed_state.into_sorted()),
```
mediocregopher: This is a complication that I hadn't thought about... that we might be making block building slower by forcing the builder to do the sorting itself, rather than the validator. There's a few of possibilities here:

* Use an Either for hashed_state/trie_updates fields on ExecutedBlock, so they can be sorted or not.
* Use a different ExecutedBlock struct for block building, which doesn't seem wrong but might be a big change
* Use an ExecutedBlockExt kind of struct in the engine tree, which might look like:
```
struct ExecutedBlockExt {
    executed_block: ExecutedBlock, // retains unsorted updates/state
    trie_updates_sorted: TrieUpdatesSorted,
    hashed_state_sorted: HashedPostStateSorted,
}
```
...and leave other places untouched

cc @mattsse

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -296,12 +296,10 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
             self.write_state(&execution_output, OriginalValuesKnown::No)?;
 
             // insert hashes and intermediate merkle nodes
-            self.write_hashed_state(&Arc::unwrap_or_clone(hashed_state).into_sorted())?;
+            self.write_hashed_state(&hashed_state)?;
```
mediocregopher, line 328: 🔥 🔥 🔥

crates/trie/common/src/input.rs
```diff
@@ -119,3 +138,92 @@ impl TrieInput {
...
+        Arc::make_mut(&mut self.state).extend_ref(&sorted_state);
+    }
+
+    /// Clears all data, reusing allocations if possible via `Arc::make_mut`.
+    pub fn clear(&mut self) {
```
mediocregopher: Do we use this method? I don't think it makes much sense to use it, given that we're holding onto Arcs

mediocregopher (commented):
Looking good! I think the biggest open question is what to do about ExecutedBlock, which seems to get used in a lot of places that maybe we don't want this change to touch.

crates/optimism/payload/src/builder.rs
yongkangc: hmm yeah, would like to get @mattsse thoughts on what might be the best way to deal with this re block building and `ExecutedBlock` abstraction

Personally i feel like `ExecutedBlockExt` kind of struct in the engine tree might make sense to isolate changes

crates/trie/common/src/input.rs
yongkangc: if we do this, it becaomes a thin layer of wrapper that i think we can get rid off

>I see, in this case I would have append_ref at least accept a &HashedPostStateSorted, and do the into_sorted outside, so it might be possible in the future to eliminate that too

yongkangc: i think having this part in the function is ok, since the function is small enough and not too much load cognitively
i.e 

```
        // Extend input with the block we are validating.
        input.prefix_sets.extend(hashed_state.construct_prefix_sets());
        let sorted_hashed_state = hashed_state.clone().into_sorted();
        Arc::make_mut(&mut input.state).extend_ref(&sorted_hashed_state);
```

yongkangc: addressed [769ee47](https://github.com/paradigmxyz/reth/pull/19340/commits/769ee4727568400b71cae21a7c0cd0d29e150f6d)

crates/trie/common/src/input.rs
yongkangc: yup good note! removed

yongkangc: [dc42373](https://github.com/paradigmxyz/reth/pull/19340/commits/dc423735d7bcf839b156f055275f10593e1efbb7)

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -641,8 +641,10 @@ where
     ) -> Result<(B256, TrieUpdates), ParallelStateRootError> {
         let (mut input, block_hash) = self.compute_trie_input(parent_hash, state)?;
 
-        // Extend with block we are validating root for.
-        input.append_ref(hashed_state);
+        // Extend state overlay with current block's sorted state.
+        input.prefix_sets.extend(hashed_state.construct_prefix_sets());
+        let sorted_hashed_state = hashed_state.clone().into_sorted();
```
mediocregopher, line 645: Is this clone needed?

crates/engine/tree/src/tree/payload_validator.rs
mattsse: ah this makes sense

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -921,10 +929,24 @@ where
             debug!(target: "engine::tree::payload_validator", historical = ?block_hash, blocks = blocks.len(), "Parent found in memory");
         }
 
-        // Extend with contents of parent in-memory blocks.
-        input.extend_with_blocks(
-            blocks.iter().rev().map(|block| (block.hashed_state(), block.trie_updates())),
-        );
+        // Extend with contents of parent in-memory blocks directly in sorted form.
+        let mut input = TrieInputSorted::default();
+        let mut blocks_iter = blocks.iter().rev();
```
mattsse: this could also be .peekable()

then the > 1 check can be peek().is_some

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -921,10 +929,24 @@ where
...
+
+        if let Some(first) = blocks_iter.next() {
+            input.state = Arc::clone(&first.hashed_state);
+            input.nodes = Arc::clone(&first.trie_updates);
+
+            // Only clone and mutate if there are multiple in-memory blocks.
+            if blocks.len() > 1 {
+                let state_mut = Arc::make_mut(&mut input.state);
+                let nodes_mut = Arc::make_mut(&mut input.nodes);
```
mattsse, line 923: a I see

crates/optimism/payload/src/builder.rs
mattsse: imo this is fine here

the only reason why op computes this here is so that this payload can be reused in derivation mode (where newPayload == locally built payload), which we can reuse in the tree, hence this is optional.

so we can keep this here

mattsse (commented):
all of this lgtm

although I haven't reviewed all the order+extend changes line by line

crates/optimism/payload/src/builder.rs
yongkangc: sounds good, thanks @mattsse

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: ah nice this is a good one 👍🏻 

addressed it here: [ee96745](https://github.com/paradigmxyz/reth/pull/19340/commits/ee967452569bddad6ea8f58704f7574ec0b0b0bd)

@mattsse

mediocregopher (commented):
One little question but overall LGTM, will need rebasing with #19430 merged

crates/trie/common/src/input.rs
```diff
@@ -171,6 +171,12 @@ impl TrieInputSorted {
             prefix_sets: input.prefix_sets,
         }
     }
+
+    /// Append sorted state to the input by reference and extend the prefix sets.
+    pub fn append_ref(&mut self, state: &HashedPostStateSorted) {
```
yongkangc: felt like there was alot going one, so abstracted instead

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: @mediocregopher yup we need the clone 

since [parallel computation](https://github.com/paradigmxyz/reth/blob/af643c3df18f0ff751eb7d17aef59ff2b23d1fbe/crates/engine/tree/src/tree/payload_validator.rs#L645) requires sorted state, but the [fallback path](https://github.com/paradigmxyz/reth/blob/af643c3df18f0ff751eb7d17aef59ff2b23d1fbe/crates/engine/tree/src/tree/payload_validator.rs#L495) needs the unsorted version and clones it.

to remove this clone, we would have to update `StateRootProvider` to accept `HashedPostStateSorted` [here](https://github.com/paradigmxyz/reth/blob/b48c72fad20c0964306392066f5c077a376e5fa9/crates/storage/storage-api/src/trie.rs#L29), but that's a larger API change prob best left for another ticket?

mediocregopher (approved):

PR #19347, opened 2025-10-28
fix: make get_exact error on missing blobs to match trait contract

sashass1315:
- Align InMemoryBlobStore::get_exact() with BlobStore contract: return error if any requested blob is missing and keep order.
- Add empty-input fast path for consistency with DiskFileBlobStore.
- This fix is necessary because the previous implementation used filter_map, silently dropping missing entries, which violated the trait docs and broke callers expecting all-or-error semantics (e.g., TransactionPoolExt::get_all_blobs_exact() and the beacon sidecar example).


mattsse (approved):
this is correct

PR #19353, opened 2025-10-28
fix(trie): use block hash in OverlayStateProviderFactory

mediocregopher:
By "pinning" the overlay to a block hash, rather than block number, we can ensure that unwinds which are occurring during a payload validation cause the state root task to fail if the unwind goes past the historical block number.

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -15,7 +15,7 @@ use crate::tree::{
 use alloy_consensus::transaction::Either;
 use alloy_eips::{eip1898::BlockWithParent, NumHash};
 use alloy_evm::Evm;
-use alloy_primitives::{BlockNumber, B256};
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -976,38 +969,30 @@ where
         skip_all,
         fields(parent_hash)
     )]
-    fn compute_trie_input<TP: DBProvider + BlockNumReader + TrieReader>(
+    fn compute_trie_input(
         &self,
-        provider: TP,
```
mediocregopher, line 981: This eliminates the need for a provider in `compute_trie_input`, so really this solves two problems

mattsse (approved):
makes sense

yongkangc (approved):
nice, this is great. ties to https://github.com/paradigmxyz/reth/issues/19249

PR #19354, opened 2025-10-28
perf(codecs): avoid String allocation in proc macro type checking

strmfos:
While looking at the derive macro code, I noticed we're converting syn::Ident to String just to compare it with string literals in should_use_alt_impl(). This creates unnecessary heap allocations during macro expansion.

Since syn::Ident already implements PartialEq<&str>, we can compare directly without the .to_string().as_str() chain. This matches the pattern already used on line 155 in the same file for the "maybe_zero" check.

The change is semantically identical but skips the temporary String allocation on every type check, which should help compile times for projects with lots of Compact derives.


mattsse (approved):
this makes sense

PR #19355, opened 2025-10-28
feat(exex): Provide TrieUpdates and HashedPostState in all ExEx notifications

mediocregopher:
This changes the TrieUpdates field of the Chain type (which is used in ExEx notifications) from an Option into a BTreeMap mapping the block number to the updates pulled from the ExecutedBlock. Similarly HashedPostState is added.

With this in place all ExEx subscribers have full access to all trie state changes produced by each block.


crates/evm/execution-types/src/chain.rs
```diff
@@ -73,9 +70,12 @@ impl<N: NodePrimitives> Chain<N> {
     pub fn from_block(
         block: RecoveredBlock<N::Block>,
         execution_outcome: ExecutionOutcome<N::Receipt>,
-        trie_updates: Option<TrieUpdates>,
+        trie_updates: Arc<TrieUpdates>,
     ) -> Self {
-        Self::new([block], execution_outcome, trie_updates)
+        let block_number = block.header().number();
+        let mut trie_updates_map = BTreeMap::new();
+        trie_updates_map.insert(block_number, trie_updates);
```
shekhirin: can be simplified with `::from([...])`

crates/evm/execution-types/src/chain.rs
```diff
@@ -35,10 +34,8 @@ pub struct Chain<N: NodePrimitives = reth_ethereum_primitives::EthPrimitives> {
     ///
     /// Additionally, it includes the individual state changes that led to the current state.
     execution_outcome: ExecutionOutcome<N::Receipt>,
-    /// State trie updates after block is added to the chain.
-    /// NOTE: Currently, trie updates are present only for
-    /// single-block chains that extend the canonical chain.
-    trie_updates: Option<TrieUpdates>,
+    /// State trie updates for each block in the chain, keyed by block number.
+    trie_updates: BTreeMap<BlockNumber, Arc<TrieUpdates>>,
```
mattsse, line 38: I see

mattsse (approved):
makes sense,

lgtm

shekhirin (approved):

mediocregopher:
> Hey @mediocregopher, just wanted to get an idea of how long it might take to get this PR merged.

Hi @dhyaniarun1993 , sorry this got dropped down my list of priorities. The main issue was with the serialization/deserialization of the WAL entries, such that they remained backwards compatible. It turned into a big headache and I put it aside. I'd be open to someone else taking a stab at this, or if not it will likely have to wait for the new year.

crates/evm/execution-types/src/chain.rs
```diff
@@ -452,7 +493,12 @@ pub(super) mod serde_bincode_compat {
     {
         blocks: RecoveredBlocks<'a, N::Block>,
         execution_outcome: serde_bincode_compat::ExecutionOutcome<'a, N::Receipt>,
-        trie_updates: Option<TrieUpdates<'a>>,
+        #[serde(default, skip_serializing, rename = "trie_updates_legacy")]
```
dhyaniarun1993, line 491:
```suggestion
        #[serde(default, rename = "trie_updates_legacy")]
```

dhyaniarun1993 (commented):
I tested the new format against WALs generated on the current `main` branch, and deserialization works correctly.

The failing tests are caused by the `skip_serializing` on `_trie_updates_legacy`, which removes that field from the binary layout and breaks bincode’s positional decoding.
With `skip_serializing` removed (but #[serde(default)] kept), the format remains backward compatible. Am I missing any other cases?

dhyaniarun1993:
> @dhyaniarun1993 sorry there was one other thing I had wanted to do here which I didn't realize I hadn't pushed yet. Basically we want the fields to be `TrieUpdatesSorted` and `HashedPostStateSorted`, in general we're moving towards using the sorted forms of those types. I had implemented the bincode serialization for `TrieUpdatesSorted` (now pushed), but it's not tested and we're missing serialization for `HashedPostStateSorted` still.

Got it. Thanks for the udpate. I can add the missing serialization for `HashedPostStateSorted` and will do some testing as well.

PR #19356, opened 2025-10-28
fix: remove PersistenceState from TreeCtx

rnkrtt:
Fixes https://github.com/paradigmxyz/reth/issues/19352

### DIFF
--- a/crates/engine/tree/src/tree/state.rs
+++ b/crates/engine/tree/src/tree/state.rs
@@ -1,8 +1,6 @@
 //! Functionality related to tree state.
 
 use crate::engine::EngineApiKind;
-#[cfg(test)]
-use alloy_eips::eip1898::BlockWithParent;
 use alloy_eips::BlockNumHash;
 use alloy_primitives::{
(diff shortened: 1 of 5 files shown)


crates/engine/tree/src/tree/state.rs
```diff
@@ -1,7 +1,9 @@
 //! Functionality related to tree state.
 
 use crate::engine::EngineApiKind;
-use alloy_eips::{eip1898::BlockWithParent, BlockNumHash};
+#[cfg(test)]
+use alloy_eips::eip1898::BlockWithParent;
```
mattsse: instead of feature gating this, could we use the full path where we currently use `BlockWithParent` then we can remove this import here

mattsse (changes requested):
cool, tiny nit

crates/engine/tree/src/tree/state.rs
rnkrtt: alright, thanks!!!

mattsse (approved):

PR #19357, opened 2025-10-28
fix(docker): fix docker image builds for ubuntu

theochap:
## Description

`make IMAGE_NAME=op-reth DOCKER_IMAGE_NAME=op-reth PROFILE=release op-docker-build-push-nightly` fails on Ubuntu 24.04.3 LTS. This should fix builds.

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,7 +1,7 @@
 [build]
 pre-build = [
     # Use HTTPS for package sources
-    "apt-get update && apt-get install --assume-yes --no-install-recommends ca-certificates",
+    "apt-get update && apt-get install --assume-yes --no-install-recommends ca-certificates apt-transport-https",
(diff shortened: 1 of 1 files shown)


shekhirin (changes requested):
Needs fixing

Cross.toml
```diff
@@ -12,7 +12,7 @@ pre-build = [
     # rust-bindgen dependencies: llvm-dev libclang-dev (>= 10) clang (>= 10)
     # See: https://github.com/cross-rs/cross/wiki/FAQ#using-clang--bindgen for
     # recommended clang versions for the given cross and bindgen version.
-    "apt-get update && apt-get install --assume-yes --no-install-recommends llvm-dev libclang-dev clang",
+    "apt-get update && apt-get install --assume-yes --no-install-recommends llvm-dev libclang-5.0-dev clang-5.0",
```
mattsse, line 15: from the ci output, I assume we should leave this unchanged

PR #19360, opened 2025-10-28
feat: Output the block execution outputs after validating (reth-stateless)

wraitii:
This changes the signature of `stateless_validation` and `stateless_validation_with_trie` to output the block execution output, which notably includes transaction receipts.
This makes it possible to check the status of a particular transaction or the logs of a particular transaction whilst also proving a block executed correctly.

Performance impact is likely negligible - with optimisations this _should_ just be RVO-d out. If this is a concern, we can potentially simply return receipts and/or requests instead.

Unfortunately this would be a breaking change - not sure if reth uses semver. Alternatively, new functions can be provided - I haven't done so because honestly I haven't found a good name. Some users of this crate I know are `zeth` and the ETH applied cryptography team (who I've quickly chatted about this PR with), but it's a pretty easy fix downstream.

Alternatives:
- There are several functions in `stateless_validation_with_trie` that are not public, making it impossible to construct custom versions of that functions from the bricks `reth-stateless` provides. I think it would be good to make those public regardless, but haven't done it here.
- We could also output the `current_block` instead of pre-hashing it, and leave it to the caller to hash or not, as needed.

### DIFF
--- a/crates/stateless/src/validation.rs
+++ b/crates/stateless/src/validation.rs
@@ -17,8 +17,11 @@ use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_consensus::{Consensus, HeaderValidator};
 use reth_errors::ConsensusError;
 use reth_ethereum_consensus::{validate_block_post_execution, EthBeaconConsensus};
-use reth_ethereum_primitives::{Block, EthPrimitives};
-use reth_evm::{execute::Executor, ConfigureEvm};
+use reth_ethereum_primitives::{Block, EthPrimitives, EthereumReceipt};
+use reth_evm::{
+    execute::{BlockExecutionOutput, Executor},
+    ConfigureEvm,
+};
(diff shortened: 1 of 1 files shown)


mattsse (approved):
seems okay

pending @kevaundray

kevaundray (approved):
This also LGTM -- we initially wanted the API for this function to be standardized eventually for the attester usecase, but I think we could add a specific function eventually that is standardized across all ELs. This function would call this more general function.

For making everything public, my only (general) reservation here would be that currently it's quite easy to analyze security because there are only a few entrypoints to this crate; we can for example make assumptions on private functions because we know exactly where they will be called. If everything were to be made public, it may become a bit more difficult to do this, so at least for me, it would be good to know what some of the usecases are here before making everything public.

Saying that, It may be within Reth's scope to do this as it encourages folks to use this crate in a modular way, so ultimately I would delegate to Matt on this.

PR #19364, opened 2025-10-28
fix(cli): restrict mouse click selection to list content area

MozirDmitriy:
Persist the key list widget Rect during rendering and use it in handle_event() to hit-test mouse clicks, preventing selection jumps when clicking outside the list. The row is mapped to a content index accounting for borders and combined with list_state offset to compute the selected item, and selection is applied only when the computed index is within the current entries. This change adds an optional Rect field to DbListTUI, initializes it to None, assigns it in ui() after layout split, and adjusts the mouse down handler accordingly while leaving scroll behavior unchanged.


mattsse (commented):
I think we can keep this as is

PR #19365, opened 2025-10-28
feat: insert at timestamp

mattsse:
helper fn to insert a timestamp based fork at the correct position.

this is primarily useful for new forks that are added on top and we can assume those are always timestamp based


klkvr (approved):

fgimenez (approved):
nice!

PR #19367, opened 2025-10-28
fix: deduplicate TxpoolExt RPC methods

vtjl10:
Deduplicate RPC method logic in `examples/node-custom-rpc/src/main.rs` by extracting shared behavior into private helpers. Added `transaction_count_inner()` and `clear_txpool_inner()` on `TxpoolExt` and updated both `TxpoolExtApiServer` impls (prod and test) to call the shared helpers.


mattsse (commented):
we can keep this as is

PR #19368, opened 2025-10-28
feat(jovian/rpc): update receipts to transmit over RPC with Jovian compatible fields

theochap:
## Description

Updates the tx receipts to transmit over RPC to be consistent with Jovian updates. Ie, to contain the updated correct value of the `blob_base_fee`.

Added unit tests.

### DIFF
--- a/crates/optimism/rpc/src/eth/receipt.rs
+++ b/crates/optimism/rpc/src/eth/receipt.rs
@@ -1,7 +1,7 @@
 //! Loads and formats OP receipt RPC response.
 
 use crate::{eth::RpcNodeCore, OpEthApi, OpEthApiError};
-use alloy_consensus::{transaction::TransactionMeta, BlockHeader, Receipt, TxReceipt};
+use alloy_consensus::{BlockHeader, Receipt, TxReceipt};
 use alloy_eips::eip2718::Encodable2718;
 use alloy_rpc_types_eth::{Log, TransactionReceipt};
 use op_alloy_consensus::{OpReceiptEnvelope, OpTransaction};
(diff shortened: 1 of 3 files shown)


crates/optimism/rpc/src/eth/receipt.rs
```diff
@@ -275,6 +277,71 @@ pub struct OpReceiptBuilder {
...
+            .is_jovian_active_at_timestamp(meta.timestamp)
+        {
+            // In jovian, we're using the blob gas used field to store the current da
+            // footprint's value.
+            let da_footprint_gas_scalar = l1_block_info.da_footprint_gas_scalar.unwrap_or_default();
+            (
+                Some(
+                    // Estimate the size of the transaction in bytes and multiply by the DA
+                    // footprint gas scalar.
+                    // TODO(@theochap): Instead of recomputing the encoding of the tx, we could add
```
theochap: I will open an issue for that if the reviewers think this is relevant. Note: this will also affect the l1 impl because we'll have to add a method to the `TxReceipt` trait.

crates/optimism/rpc/src/eth/receipt.rs
```diff
@@ -275,6 +277,71 @@ pub struct OpReceiptBuilder {
...
+    ) -> TransactionReceipt<E>
+    where
+        N: NodePrimitives,
+    {
+        let ConvertReceiptInput { tx, meta, receipt, gas_used, next_log_index } = input;
+        let from = tx.signer();
+
+        let (blob_gas_used, blob_gas_price) = if chain_spec
+            .is_jovian_active_at_timestamp(meta.timestamp)
+        {
```
mattsse: could we move this check to here:

https://github.com/paradigmxyz/reth/blob/715369b819308df96a9b329c30f4ce88ac36a168/crates/optimism/rpc/src/eth/receipt.rs#L314-L314

and simply modify the `core_receipt`'s fields

then we don't need this function and keep every as is and only update the blob fields if jovian is activate

mattsse (changes requested):

crates/optimism/rpc/src/eth/receipt.rs
theochap: Done in 
3c00687

mattsse (approved):
lgtm

I only moved this is jovian active check into the fn, then we dont need a new fn just for this

mattsse (approved):

PR #19369, opened 2025-10-28
feat(rpc): implement `debug_dbGet`

### LINKED ISSUE #19362: Unimplemented debug_dbGet
### Describe the bug

https://github.com/paradigmxyz/reth/blob/v1.8.2/crates/rpc/rpc/src/debug.rs#L1140-L1142

The `debug_dbGet` handler is not yet implemented, which leads to errors for clients that rely on it:
- https://github.com/ethereum-optimism/optimism/blob/v1.14.3/op-service/sources/debug_client.go#L30-L50
- https://github.com/op-rs/kona/blob/kona-node%2fv1.1.3/bin/host/src/single/handler.rs#L246-L262

FYI, this file defines how DB is stored in Geth: https://github.com/ethereum/go-ethereum/blob/v1.16.5/core/rawdb/schema.go#L120
- Particularly, the key of a bytecode is `0x63 ++ code_hash` (33 bytes in total).

I acknowledge that the specification of `debug_dbGet` is somewhat ambiguous and appears to be client-specific. However, since the function fn `debug_code_by_hash` is already defined, we can define `debug_db_get` as follows:

- If the key is 33 bytes long and its first byte is 0x63, return the corresponding code.
- Otherwise, return `EthApiError::InvalidParams`.

### Steps to reproduce

```
$ cast rpc debug_dbGet "0x636e49e66782037c0555897870e29fa5e552daf4719552131a0abce779daec0a5d" 
null
```

Instead, it should returns the bytecode of EIP-2935 (address `0x0000F90827F1C53a10cb7A02335B175320002935`):

```
0x3373fffffffffffffffffffffffffffffffffffffffe14604657602036036042575f35600143038111604257611fff81430311604257611fff9006545f5260205ff35b5f5ffd5b5f35611fff60014303065500
```

### Node logs

```text
N/A
```

### Platform(s)

_No response_

### Container Type

Docker

v1.8.1

N/A

N/A

Archive (default)

_No response_

_No response_

### Code of Conduct

mablr:
Closes #19362 

According to Geth impl, bytecode keys are: 0x63 (prefix) + 32 bytes (code_hash)
https://github.com/ethereum/go-ethereum/blob/v1.16.5/core/rawdb/schema.go#L120

### DIFF
--- a/crates/rpc/rpc-builder/tests/it/http.rs
+++ b/crates/rpc/rpc-builder/tests/it/http.rs
@@ -1694,3 +1694,52 @@ async fn test_eth_fee_history_raw() {
     )
     .await;
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_debug_db_get() {
+    reth_tracing::init_test_tracing();
+
+    let handle = launch_http(vec![RethRpcModule::Debug]).await;
+    let client = handle.http_client().unwrap();
+
+    // Valid test cases
+    let valid_test_cases = [
+        ("0x630000000000000000000000000000000000000000000000000000000000000000"),
+        ("630000000000000000000000000000000000000000000000000000000000000000"),
+    ];
+
+    for key in valid_test_cases {
+        DebugApiClient::<()>::debug_db_get(&client, key.into()).await.unwrap();
+    }
+
+    // Invalid test cases
+    let test_cases = [
+        ("0x0000", "Key must be 33 bytes, got 2"),
+        ("0000", "Key must be 33 bytes, got 2"),
+        (
+            "0x000000000000000000000000000000000000000000000000000000000000000000",
+            "Key prefix must be 0x63",
+        ),
+        (
+            "000000000000000000000000000000000000000000000000000000000000000000",
+            "Key prefix must be 0x63",
+        ),
+        ("0xc0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
+        ("c0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
+    ];
+
+    let match_error_msg = |err: jsonrpsee::core::client::Error, expected: String| -> bool {
+        match err {
+            jsonrpsee::core::client::Error::Call(error_obj) => {
+                error_obj.code() == ErrorCode::InvalidParams.code() &&
+                    error_obj.message() == expected
+            }
+            _ => false,
+        }
+    };
+
+    for (key, expected) in test_cases {
+        let err = DebugApiClient::<()>::debug_db_get(&client, key.into()).await.unwrap_err();
+        assert!(match_error_msg(err, expected.into()));
+    }
+}
--- a/crates/rpc/rpc-builder/tests/it/http.rs
+++ b/crates/rpc/rpc-builder/tests/it/http.rs
@@ -1702,10 +1702,9 @@ async fn test_debug_db_get() {
     let handle = launch_http(vec![RethRpcModule::Debug]).await;
     let client = handle.http_client().unwrap();
 
-    // Valid test cases
     let valid_test_cases = [
-        ("0x630000000000000000000000000000000000000000000000000000000000000000"),
-        ("630000000000000000000000000000000000000000000000000000000000000000"),
+        "0x630000000000000000000000000000000000000000000000000000000000000000",
+        "c00000000000000000000000000000000",
     ];
 
     for key in valid_test_cases {
@@ -1715,17 +1714,13 @@ async fn test_debug_db_get() {
     // Invalid test cases
     let test_cases = [
         ("0x0000", "Key must be 33 bytes, got 2"),
-        ("0000", "Key must be 33 bytes, got 2"),
+        ("00", "Key must be 33 bytes, got 2"),
         (
             "0x000000000000000000000000000000000000000000000000000000000000000000",
             "Key prefix must be 0x63",
         ),
-        (
-            "000000000000000000000000000000000000000000000000000000000000000000",
-            "Key prefix must be 0x63",
-        ),
+        ("000000000000000000000000000000000", "Key prefix must be 0x63"),
         ("0xc0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
-        ("c0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
     ];
 
     let match_error_msg = |err: jsonrpsee::core::client::Error, expected: String| -> bool {
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -1148,7 +1148,7 @@ where
     /// Currently supported:
     /// - Contract bytecode associated with a code hash.
     ///   The code hash must be provided as a hex string prefixed with 0x.
-    ///   See Geth impl: https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/core/rawdb/schema.go#L120
+    ///   See Geth impl: <https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/core/rawdb/schema.go#L120>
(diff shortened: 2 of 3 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -1143,8 +1143,29 @@ where
...
+            .into());
+        }
+        if key_bytes[0] != 0x63 {
+            return Err(EthApiError::InvalidParams(
+                "Key prefix must be 0x63 for bytecode".to_string(),
+            )
+            .into());
+        }
+
+        let code_hash = B256::from_slice(&key_bytes[1..33]);
```
mattsse: this endpoint supports more lookups that we perhaps want to add eventually

I think we can document this here, for now just support code by hash is fine

crates/rpc/rpc/src/debug.rs
```diff
@@ -1143,8 +1143,29 @@ where
         Ok(())
     }
 
-    async fn debug_db_get(&self, _key: String) -> RpcResult<()> {
-        Ok(())
+    async fn debug_db_get(&self, key: String) -> RpcResult<Option<Bytes>> {
+        // According to Geth impl, bytecode keys are: 0x63 (prefix) + 32 bytes (code_hash)
+        // https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/core/rawdb/schema.go#L120
+        let key_bytes =
+            decode(&key).map_err(|_| EthApiError::InvalidParams("Invalid hex key".to_string()))?;
```
mattsse: unsure if this really works here because:

https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/common/bytes.go#L98-L105

could we add a simple test case for this like `c0000000000000000000000000000000000000000000000000000000000000000`

should be an example input for codehash

mattsse (changes requested):
this lgtm, but I'm unsure about the hex decoding so would appreciate another check

crates/rpc/rpc/src/debug.rs
mablr: added some doc

crates/rpc/rpc/src/debug.rs
mablr: I added a test, wdyt ? It is in the right place ?

crates/rpc/rpc-builder/tests/it/http.rs
```diff
@@ -1694,3 +1694,52 @@ async fn test_eth_fee_history_raw() {
...
+        (
+            "0x000000000000000000000000000000000000000000000000000000000000000000",
+            "Key prefix must be 0x63",
+        ),
+        (
+            "000000000000000000000000000000000000000000000000000000000000000000",
+            "Key prefix must be 0x63",
+        ),
+        ("0xc0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
+        ("c0000000000000000000000000000000000000000000000000000000000000000", "Invalid hex key"),
```
mattsse: I believe this should pass

see https://github.com/op-rs/kona/blob/aa7e89596afa96f08eb1b2e14db269fabeb673ea/bin/host/src/single/handler.rs#L245

crates/rpc/rpc/src/debug.rs
```diff
@@ -1143,8 +1143,30 @@ where
         Ok(())
     }
 
-    async fn debug_db_get(&self, _key: String) -> RpcResult<()> {
-        Ok(())
+    /// `debug_db_get` - database key lookup
+    ///
+    /// Currently supported:
+    /// - Contract bytecode associated with a code hash.
+    ///   The code hash must be provided as a hex string prefixed with 0x.
```
mattsse: looking at the kona code and how this is parsed

https://github.com/op-rs/kona/blob/aa7e89596afa96f08eb1b2e14db269fabeb673ea/bin/host/src/single/handler.rs#L245

I'm not sure this is correct because if the prefix is missing the this will just use the content

https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/common/bytes.go#L101-L103

mattsse (changes requested):

crates/rpc/rpc-builder/tests/it/http.rs
mablr: These strings are actually invalid, because of their odd length.

Here is a snippet where we create a code key like in kona
```rust
// Code key like in Kona
let kona_code_key: Bytes = [&[b'c'], B256::ZERO.as_slice()].concat().into();

// Show that valid test cases are equal to the kona code key
assert_eq!(Bytes::from_str(valid_test_cases[0]).unwrap(), kona_code_key);
assert_eq!(Bytes::from_str(valid_test_cases[1]).unwrap(), kona_code_key);
// Fails due to odd length 
assert!(Bytes::from_str(test_cases[5].0).is_err());
```

On the other hand, I can also implement the fallback to also accept code hash without the Geth prefix scheme, like this: https://github.com/op-rs/kona/blob/1b136596d612de0ba5d17aee49df4fd52873a3fd/bin/host/src/single/handler.rs#L252-L262

crates/rpc/rpc/src/debug.rs
mablr: mb this is actually unclear, i'll fix it when the other review comment will be resolved (final behavior decided)

crates/rpc/rpc-builder/tests/it/http.rs
mattsse: hmm, does this mean `c0000000000000000000000000000000000000000000000000000000000000000` (which kona is sending) is invalid?

unclear why if geth does

https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/internal/ethapi/dbapi.go#L26-L26

https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/common/bytes.go#L101-L103

mablr: Kona does not send`c0000000000000000000000000000000000000000000000000000000000000000`, it concatenates the prefix byte `&[b'c']` (eg `&[0x63]`) and the `code_hash` a `B256` as slice of `u8`. Then it is converted to alloy's Bytes, then the serialization to string will give: `0x630000000000000000000000000000000000000000000000000000000000000000`.

mablr: Ok I see, Geth handles differently the key :
- with `0x` prefix it decode the string as hex ("0x6300" -> [0x63, 0x00])
- without `0x` it return the string as raw bytes ('c0' ->  -> [0x63, 0x00])

The current PR impl IS compatible with Kona as it uses alloy's Bytes (serialized into hex string starting with `0x`)

So the modification to apply is to not decode string, when it is not prefixed with `0x`. Then the check with already apply to the blob are OK (length == 33 && first byte == 0x63).

crates/rpc/rpc/src/debug.rs
mablr: done

crates/rpc/rpc-builder/tests/it/http.rs
mattsse: ah I see, this is indeed b"c"

thanks for walking me through this

crates/rpc/rpc/src/debug.rs
```diff
@@ -1143,8 +1143,38 @@ where
...
+            .into());
+        }
+        if key_bytes[0] != 0x63 {
+            return Err(EthApiError::InvalidParams("Key prefix must be 0x63".to_string()).into());
+        }
+
+        let code_hash = B256::from_slice(&key_bytes[1..33]);
+
+        // No block ID is provided, so it defaults to the latest block
+        self.debug_code_by_hash(code_hash, None).await.map_err(Into::into)
```
mattsse, line 1177: it's fine to just support codehash for now because this is the only thing used by kona

mattsse (approved):
tysm!

PR #19371, opened 2025-10-28
fix: remove duplicate debug log

kilavvy:
In `crates/trie/db/src/state.rs`, duplicate debug logging was removed to reduce log noise. Specifically, the `debug!` call was removed from the `incremental_root_with_updates()` method because it duplicated an identical message already logged by the `incremental_root()` method. This change eliminates redundant output that provided no unique context, while the debug log in `incremental_root()` has been retained for necessary observability.


mattsse (commented):
hmm, we also have debug! here

https://github.com/paradigmxyz/reth/blob/6651ae78521730740cc4af59b41c7c4d3cb1d1e7/crates/trie/db/src/state.rs#L157-L157

should we remove this there as well or keep both @shekhirin

mattsse (commented):
we can keep this

PR #19372, opened 2025-10-28
perf: optimize SyncHeight event handling to avoid recursive calls

DeVikingMark:
Replaces recursive event creation with direct batch updates when sync height changes. 

Eliminates 16 redundant StageCheckpoint events per SyncHeight event, reducing computational overhead during blockchain synchronization.

### DIFF
--- a/crates/stages/api/src/metrics/listener.rs
+++ b/crates/stages/api/src/metrics/listener.rs
@@ -52,9 +52,7 @@ impl MetricsListener {
         trace!(target: "sync::metrics", ?event, "Metric event received");
         match event {
             MetricEvent::SyncHeight { height } => {
-                // Efficiently update all stage checkpoints to the new sync height
-                // This avoids creating 16 separate StageCheckpoint events and recursive calls
-                self.sync_metrics.update_all_stages_height(height);
+                self.update_all_stages_height(height);
             }
             MetricEvent::StageCheckpoint { stage_id, checkpoint, max_block_number, elapsed } => {
                 let stage_metrics = self.sync_metrics.get_stage_metrics(stage_id);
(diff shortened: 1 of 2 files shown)


crates/stages/api/src/metrics/sync_metrics.rs
```diff
@@ -16,6 +16,17 @@ impl SyncMetrics {
...
+    /// Updates all stage checkpoints to the given height efficiently.
+    /// This avoids creating multiple individual events when sync height changes.
+    pub(crate) fn update_all_stages_height(&mut self, height: u64) {
+        for stage_id in StageId::ALL {
+            let stage_metrics = self.get_stage_metrics(stage_id);
+            stage_metrics.checkpoint.set(height as f64);
+            stage_metrics.entities_processed.set(height as f64);
+            stage_metrics.entities_total.set(height as f64);
+        }
+    }
```
mattsse: can we make this a fn of metriclistener instead?

crates/stages/api/src/metrics/listener.rs
```diff
@@ -52,17 +52,9 @@ impl MetricsListener {
...
-                        checkpoint: StageCheckpoint {
-                            block_number: height,
-                            stage_checkpoint: None,
-                        },
-                        max_block_number: Some(height),
-                        elapsed: Duration::default(),
-                    });
-                }
+                // Efficiently update all stage checkpoints to the new sync height
+                // This avoids creating 16 separate StageCheckpoint events and recursive calls
```
mattsse: this comment requires context from this pr and should therefor be removed

mattsse (changes requested):
makes sense,

one suggestion

mattsse (approved):
makes sense to not do this recursively

PR #19374, opened 2025-10-28
feat: add default warn log on stream closure in poll_nested_stream_with_budget!

Forostovec:
his change implements the pending TODO in poll_nested_stream_with_budget! by utilizing the provided $target and $label to emit a default ::tracing::warn! when the polled stream returns Poll::Ready(None) and no custom on_ready_none handler is supplied. The macro is split into two arms: one that preserves the existing behavior for call sites passing an explicit closure handler and another that provides a default warning log for call sites that do not, ensuring consistent and informative behavior across the codebase without altering existing semantics. The metered wrapper remains unchanged and forwards its optional handler as before.


mattsse (commented):
we can ignore this

PR #19375, opened 2025-10-29
feat: impl a function to create new instance of TransactionEvents

datnguyencse:
- The TransactionPool trait includes functions that return TransactionEvents, but currently, it’s not possible to initialize the TransactionEvents outside of Reth. This limitation makes it difficult to implement a custom transaction pool.

### DIFF
--- a/crates/transaction-pool/src/pool/listener.rs
+++ b/crates/transaction-pool/src/pool/listener.rs
@@ -30,7 +30,7 @@ pub struct TransactionEvents {
 
 impl TransactionEvents {
     /// Create a new instance of this stream.
-    pub fn new(hash: TxHash, events: UnboundedReceiver<TransactionEvent>) -> Self {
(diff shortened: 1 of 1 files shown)


yongkangc:
I dont see any issues with this. pending @mattsse

yongkangc (approved):

crates/transaction-pool/src/pool/listener.rs
```diff
@@ -29,6 +29,11 @@ pub struct TransactionEvents {
 }
 
 impl TransactionEvents {
+    /// Create a new instance of this stream.
+    pub fn new(hash: TxHash, events: UnboundedReceiver<TransactionEvent>) -> Self {
```
mattsse:
```suggestion
    pub const fn new(hash: TxHash, events: UnboundedReceiver<TransactionEvent>) -> Self {
```

mattsse (approved):

PR #19380, opened 2025-10-29
fix(engine): remove redundant parent_to_child cleanup in insert_executed

sky-coderay:
Drop unnecessary parent_to_child cleanup loop in TreeState::insert_executed. This method only inserts blocks, never removes them, so iterating through all parent-child relationships to clean up stale references is redundant. The proper cleanup already happens in remove_by_hash where it belongs.

Same pattern as #18648 but for TreeState instead of BlockBuffer.

### DIFF
--- a/crates/engine/tree/src/tree/state.rs
+++ b/crates/engine/tree/src/tree/state.rs
@@ -107,10 +107,6 @@ impl<N: NodePrimitives> TreeState<N> {
         self.blocks_by_number.entry(block_number).or_default().push(executed);
 
         self.parent_to_child.entry(parent_hash).or_default().insert(hash);
-
-        for children in self.parent_to_child.values_mut() {
-            children.retain(|child| self.blocks_by_hash.contains_key(child));
-        }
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/state.rs
```diff
@@ -107,10 +107,6 @@ impl<N: NodePrimitives> TreeState<N> {
         self.blocks_by_number.entry(block_number).or_default().push(executed);
 
         self.parent_to_child.entry(parent_hash).or_default().insert(hash);
-
-        for children in self.parent_to_child.values_mut() {
-            children.retain(|child| self.blocks_by_hash.contains_key(child));
-        }
```
mattsse, line 113: I see, this really doesnt belong here and is kinda pointless here

mattsse (approved):
I believe this is correct

but would appreciate another review from @Rjected or @fgimenez

fgimenez (approved):
nice, lgtm

PR #19381, opened 2025-10-29
feat(static-file): dynamic static file size

shekhirin:
Closes #19273

This PR doesn't make the static file size configurable yet, only introduces support for deriving the number of blocks per file from existing files instead of using the hardcoded value.
 
See this test for demonstration https://github.com/paradigmxyz/reth/blob/c200982d33ae4078df9e525fa47537f97dfc9209/crates/storage/provider/src/providers/static_file/mod.rs#L552-L635


crates/storage/db/src/static_file/mod.rs
```diff
@@ -17,12 +14,11 @@ pub use masks::*;
 use reth_static_file_types::{SegmentHeader, SegmentRangeInclusive, StaticFileSegment};
 
 /// Alias type for a map of [`StaticFileSegment`] and sorted lists of existing static file ranges.
-type SortedStaticFiles =
-    HashMap<StaticFileSegment, Vec<(SegmentRangeInclusive, Option<SegmentRangeInclusive>)>>;
+type SortedStaticFiles = HashMap<StaticFileSegment, Vec<(SegmentRangeInclusive, SegmentHeader)>>;
```
mattsse, line 17: is it useful if we introduce a helper type for this?

crates/static-file/types/src/lib.rs
```diff
@@ -79,11 +79,12 @@ impl StaticFileTargets {
 
 /// Each static file has a fixed number of blocks. This gives out the range where the requested
 /// block is positioned. Used for segment filename.
-pub const fn find_fixed_range(
+pub fn find_fixed_range(
+    start: Option<BlockNumber>,
```
mattsse: can we update docs here

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -250,7 +249,9 @@ pub struct StaticFileProviderInner<N> {
     earliest_history_height: AtomicU64,
     /// Max static file block for each segment
     static_files_max_block: RwLock<HashMap<StaticFileSegment, u64>>,
-    /// Available static file block ranges on disk indexed by max transactions.
+    /// Expected on disk static file block ranges indexed by max expected blocks.
+    static_files_expected_block_index: RwLock<SegmentRanges>,
+    /// Available on disk static file block ranges indexed by max transactions.
     static_files_tx_index: RwLock<SegmentRanges>,
```
mattsse, line 305: can we add additional docs how these two relate to each other?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -301,8 +303,48 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
     /// Each static file has a fixed number of blocks. This gives out the range where the requested
     /// block is positioned.
-    pub const fn find_fixed_range(&self, block: BlockNumber) -> SegmentRangeInclusive {
-        find_fixed_range(block, self.blocks_per_file)
+    ///
+    /// For already initialized static files, this function will use their existing ranges, instead
+    /// of the current fixed number of blocks.
+    ///
+    /// See `[Self::find_fixed_range]` documentation for more information.
+    pub fn find_fixed_range_with_block_index(
```
mattsse, line 364: i find some of this implementation not very intuitive, and would like to see a few additional docs here that explain what this actually returns.

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -301,8 +303,48 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
+                .iter()
+                .find(|(max_block, _)| block <= **max_block)
+                .filter(|(_, range)| range.contains(block))
+            {
+                return *range
+            } else if let Some((_, range)) = block_index.last_key_value() {
+                return find_fixed_range(Some(range.end() + 1), block, self.blocks_per_file)
+            }
+        }
+        find_fixed_range(None, block, self.blocks_per_file)
```
mattsse: what does this do?

mattsse (changes requested):
I'm being extra pedantic here re docs, because I had a very hard time figuring out what all of this is when I worked on the initial history expiry implementation and would like to really explain all of this as best as we can

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -301,9 +303,62 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
+            {
+                // Found matching range for an existing file using block index
+                return *range
+            } else if let Some((_, range)) = block_index.last_key_value() {
+                // Didn't find matching range for an existing file, derive a new range from the end
+                // of the last existing file range.
+                let blocks_after_last_range = block - range.end();
+                let segments_to_skip = (blocks_after_last_range - 1) / self.blocks_per_file;
+                let start = range.end() + 1 + segments_to_skip * self.blocks_per_file;
+                return SegmentRangeInclusive::new(start, start + self.blocks_per_file - 1)
```
joshieDo: can we add tests on  `crates/storage/provider/src/providers/static_file/mod.rs` that cover this

joshieDo (approved):

crates/storage/db/src/static_file/mod.rs
shekhirin: I think it's fine as is, I will simplify this type with this later https://github.com/paradigmxyz/reth/commit/5b66e371974a3a4dd932312a5a6fed21423d5779

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: added docs and tests

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: done

shekhirin:
Returning to draft for now, this method needs adjustment as well https://github.com/paradigmxyz/reth/blob/c200982d33ae4078df9e525fa47537f97dfc9209/crates/storage/provider/src/providers/static_file/manager.rs#L1180-L1201

crates/storage/provider/src/providers/static_file/manager.rs
mattsse, line 305: unclear to me what this means exactly, what's the difference here between the two sets?

can we add an exampl eto the docs

mattsse (changes requested):
this makes sense now, the only things that isnt immediately obvious are the two segmentranges rwlocks, can we add some additional docs what this is exactly

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: added docs, will unify these into one lock in the future, so that we have a single segment -> indices map

mattsse (approved):

crates/storage/db/src/static_file/mod.rs
```diff
@@ -39,25 +35,18 @@ pub fn iter_static_files(path: &Path) -> Result<SortedStaticFiles, NippyJarError
...
-                        entry.get_mut().push((block_range, tx_range));
-                    }
-                    Entry::Vacant(entry) => {
-                        entry.insert(vec![(block_range, tx_range)]);
-                    }
-                }
+            if let Some(block_range) = jar.user_header().block_range() {
+                static_files
+                    .entry(segment)
+                    .and_modify(|headers| headers.push((block_range, jar.user_header().clone())))
```
mattsse: should this be sorted?

crates/storage/db/src/static_file/mod.rs
mattsse, line 17: why are we now storing the entire header?

this isnt kept in sync with the jar file, right?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -344,15 +355,67 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
+                .iter()
+                .find(|(max_block, _)| block <= **max_block)
+                .filter(|(_, range)| range.contains(block))
+            {
+                // Found matching range for an existing file using block index
+                return *range
+            } else if let Some((_, range)) = block_index.last_key_value() {
+                // Didn't find matching range for an existing file, derive a new range from the end
+                // of the last existing file range.
+                let blocks_after_last_range = block - range.end();
```
mattsse, line 385: this assumes the block is always higher, is this always the case?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -344,15 +355,67 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
+                let segments_to_skip = (blocks_after_last_range - 1) / blocks_per_file;
+                let start = range.end() + 1 + segments_to_skip * blocks_per_file;
```
mattsse, line 387: okay, I see this checks the distance using the highest existing range

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -344,15 +355,67 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
...
         block: BlockNumber,
     ) -> SegmentRangeInclusive {
         let blocks_per_file =
             self.blocks_per_file.get(&segment).copied().unwrap_or(DEFAULT_BLOCKS_PER_STATIC_FILE);
+
+        if let Some(block_index) = block_index {
+            // Find first block range that contains the requested block
+            if let Some((_, range)) = block_index
+                .iter()
+                .find(|(max_block, _)| block <= **max_block)
```
mattsse: do we even need this check here?

because isnt max block always in the range?

mattsse (commented):
I think all of this is correct, but not super easy to follow, tests look good

but I have some additional questions

also would like another @joshieDo review

crates/storage/db/src/static_file/mod.rs
shekhirin: https://github.com/paradigmxyz/reth/blob/9156f444193d631494de34c36d37e2a1ca472368/crates/storage/db/src/static_file/mod.rs#L47-L51

crates/storage/db/src/static_file/mod.rs
shekhirin: because we now use more fields, didn't want to include a tuple https://github.com/paradigmxyz/reth/blob/9156f444193d631494de34c36d37e2a1ca472368/crates/storage/provider/src/providers/static_file/manager.rs#L864-L877

it's not kept in sync, same as it wasn't before. It's used only momentarily during index initialization, and discarded after.

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: No need for `contains` check: `max_block` is kept in sync with `range`.

mattsse (approved):
okiee, pending last sanity @joshieDo check

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: yep, added a comment

joshieDo (approved):

PR #19383, opened 2025-10-29
fix: Don't always clone in-memory overlays in OverlayStateProviderFactory

mediocregopher:
If there are no reverts then there's no need to call `extend_ref`, which clones the full overlay value; we can just use the overlays directly.


yongkangc (approved):
lgtm, leaving note

> If the reverts are empty, that means that dB doesnt change between tx creation, specifically it means the block is part of chain which is anchored to db tip. If the DB changes during validation then this would stop being the case and there would need to be reverts

shekhirin (approved):

PR #19384, opened 2025-10-29
feat: add `Metadata` table and `StorageSettings` to `ProviderFactory`

joshieDo:
towards https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097

* add generic `Metadata` table.
* add `StorageSettings` entry to above table.
* Initialize `StorageSettings` on `init_genesis`
* Cache storage settings on `ProviderFactory` and `DatabaseProvider` and after `init_genesis`

https://github.com/paradigmxyz/reth/pull/19384 <- https://github.com/paradigmxyz/reth/pull/19399 <- https://github.com/paradigmxyz/reth/pull/19401

### DIFF
--- a/crates/storage/db-api/src/models/metadata.rs
+++ b/crates/storage/db-api/src/models/metadata.rs
@@ -14,26 +14,26 @@ pub struct StorageSettings {
     /// Whether this node always writes receipts to static files.
     ///
     /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be written to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
-    pub receipts_on_static_files: bool,
+    pub receipts_in_static_files: bool,
 }
 
 impl StorageSettings {
     /// Creates a new `StorageSettings` with default values.
     pub const fn new() -> Self {
-        Self { receipts_on_static_files: false }
+        Self { receipts_in_static_files: false }
     }
 
     /// Creates `StorageSettings` for legacy nodes.
     ///
-    /// This explicitly sets `receipts_on_static_files` to `false`, ensuring older nodes
+    /// This explicitly sets `receipts_in_static_files` to `false`, ensuring older nodes
     /// continue writing receipts to the database when receipt pruning is enabled.
     pub const fn legacy() -> Self {
-        Self { receipts_on_static_files: false }
+        Self { receipts_in_static_files: false }
     }
 
     /// Sets the `receipts_static_files` flag to true.
-    pub const fn with_receipts_on_static_files(mut self) -> Self {
-        self.receipts_on_static_files = true;
+    pub const fn with_receipts_in_static_files(mut self) -> Self {
+        self.receipts_in_static_files = true;
         self
     }
 }
--- a/crates/storage/db-api/src/models/metadata.rs
+++ b/crates/storage/db-api/src/models/metadata.rs
@@ -23,6 +23,14 @@ impl StorageSettings {
         Self { receipts_on_static_files: false }
     }
 
+    /// Creates `StorageSettings` for legacy nodes.
+    ///
+    /// This explicitly sets `receipts_on_static_files` to `false`, ensuring older nodes
+    /// continue writing receipts to the database when receipt pruning is enabled.
+    pub const fn legacy() -> Self {
+        Self { receipts_on_static_files: false }
+    }
+
     /// Sets the `receipts_static_files` flag to true.
     pub const fn with_receipts_on_static_files(mut self) -> Self {
         self.receipts_on_static_files = true;
--- a/crates/storage/db-api/src/models/metadata.rs
+++ b/crates/storage/db-api/src/models/metadata.rs
@@ -14,18 +14,18 @@ pub struct StorageSettings {
     /// Whether this node always writes receipts to static files.
     ///
     /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be written to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
-    pub receipts_static_files: bool,
+    pub receipts_on_static_files: bool,
 }
 
 impl StorageSettings {
     /// Creates a new `StorageSettings` with default values.
     pub const fn new() -> Self {
-        Self { receipts_static_files: false }
+        Self { receipts_on_static_files: false }
     }
 
     /// Sets the `receipts_static_files` flag to true.
     pub const fn with_receipts_on_static_files(mut self) -> Self {
-        self.receipts_static_files = true;
+        self.receipts_on_static_files = true;
         self
     }
 }
--- a/crates/storage/db-api/src/models/metadata.rs
+++ b/crates/storage/db-api/src/models/metadata.rs
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 pub struct StorageSettings {
     /// Whether this node always writes receipts to static files.
     ///
-    /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be writtten to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
+    /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be written to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
     pub receipts_static_files: bool,
 }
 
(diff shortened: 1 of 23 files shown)


crates/storage/db-api/src/models/metadata.rs
```diff
@@ -0,0 +1,31 @@
...
+/// behaviour of new or old nodes respectively.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize, Compact)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[add_arbitrary_tests(compact)]
+pub struct StorageSettings {
+    /// Whether this node always writes receipts to static files.
+    ///
+    /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be written to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
+    pub receipts_on_static_files: bool,
+}
```
mattsse, line 18: can we make this so that we can easily extend this and store arbitrary info?

crates/storage/db-api/src/tables/mod.rs
```diff
@@ -540,6 +540,13 @@ tables! {
         type Key = ChainStateKey;
         type Value = BlockNumber;
     }
+
+    /// Stores generic node metadata as key-value pairs.
+    /// Can store feature flags, configuration markers, and other node-specific data.
+    table Metadata {
+        type Key = String;
+        type Value = Vec<u8>;
+    }
```
mattsse, line 549: oh I see this also works because then we can store arbitrary metadata entries like key value storage

crates/storage/storage-api/src/metadata.rs
```diff
@@ -0,0 +1,54 @@
...
+pub trait MetadataWriter: Send + Sync {
+    /// Write a metadata value
+    fn write_metadata(&self, key: &str, value: Vec<u8>) -> ProviderResult<()>;
+
+    /// Write storage settings for this node
+    ///
+    /// Be sure to update provider factory cache with
+    /// [`StorageSettingsCache::set_storage_settings_cache`].
+    fn write_storage_settings(&self, settings: StorageSettings) -> ProviderResult<()> {
+        use reth_codecs::Compact;
```
mattsse: should we store this as compact or can/should we even do serde?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -248,8 +250,9 @@ impl<TX: DbTxMut, N: NodeTypes> DatabaseProvider<TX, N> {
         static_file_provider: StaticFileProvider<N::Primitives>,
         prune_modes: PruneModes,
         storage: Arc<N::Storage>,
+        storage_settings: Arc<RwLock<StorageSettings>>,
```
mattsse, line 255: we might want to extend this in the future with things like snapsync node or smth, but for now i think this is okay

crates/storage/provider/src/providers/database/builder.rs
```diff
@@ -103,15 +102,15 @@ impl<N> ProviderFactoryBuilder<N> {
         config: impl Into<ReadOnlyConfig>,
     ) -> eyre::Result<ProviderFactory<NodeTypesWithDBAdapter<N, Arc<DatabaseEnv>>>>
     where
-        N: NodeTypes,
+        N: NodeTypesForProvider,
```
mattsse, line 105: why is this necessary?

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -64,31 +66,47 @@ pub struct ProviderFactory<N: NodeTypesWithDB> {
...
+        let storage_settings = DatabaseProvider::<_, N>::new(
+            db.tx()?,
+            chain_spec.clone(),
+            static_file_provider.clone(),
+            Default::default(),
+            Default::default(),
+            Arc::new(RwLock::new(StorageSettings::default())),
+        )
+        .storage_settings()?
+        .unwrap_or_default();
```
mattsse: this is correct here because this defaults to false, could we make this explicit with a constructor fn on StorageSettings

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -64,31 +66,47 @@ pub struct ProviderFactory<N: NodeTypesWithDB> {
...
+    ) -> ProviderResult<Self> {
+        let storage_settings = DatabaseProvider::<_, N>::new(
+            db.tx()?,
+            chain_spec.clone(),
+            static_file_provider.clone(),
+            Default::default(),
+            Default::default(),
+            Arc::new(RwLock::new(StorageSettings::default())),
+        )
+        .storage_settings()?
```
mattsse: this feels a bit strange, why does this need StorageSettings as input in order to return it?

mattsse (changes requested):

crates/storage/provider/src/providers/database/mod.rs
joshieDo: yeah it feels weird because its a bit cyclical, open to suggestions.

The overall idea is that both `ProviderFactory` and `DatabaseProvider` are instantiated with a cached storage settings object. But to obtain it in the first place, `ProviderFactory` needs to create a `DatabaseProvider` to fetch it from storage. That's what's happening here.

crates/storage/provider/src/providers/database/builder.rs
joshieDo: want to read and cache `StorageSettings` on `ProviderFactory::new`. For that, we need to create a `DatabaseProvider` which requires `NodeTypesForProvider` afaict

crates/storage/storage-api/src/metadata.rs
joshieDo: yeah, actually agree with that _a lot_

crates/storage/db-api/src/models/metadata.rs
Rjected: yeah I think we will also want something for snap sync here, maybe we could change `receipts_on_static_files` to `static_files_v2_enabled` since I think I'll just reuse this value to determine whether or not to write changesets in https://github.com/paradigmxyz/reth/pull/18882

maybe this should just be a map?

joshieDo: couldnt we just add new booleans here as we need them and make the sf roll out changes more gradually?

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -64,31 +66,48 @@ pub struct ProviderFactory<N: NodeTypesWithDB> {
...
-        Self {
+    ) -> ProviderResult<Self> {
+        let legacy_settings = StorageSettings::legacy();
+        let storage_settings = DatabaseProvider::<_, N>::new(
+            db.tx()?,
+            chain_spec.clone(),
+            static_file_provider.clone(),
+            Default::default(),
+            Default::default(),
+            Arc::new(RwLock::new(legacy_settings)),
```
mattsse, line 98: hmm, isn't this the same as creating Self first and then updating the storage_settings?

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -64,31 +66,48 @@ pub struct ProviderFactory<N: NodeTypesWithDB> {
...
     /// Create new database provider factory.
     pub fn new(
         db: N::DB,
         chain_spec: Arc<N::ChainSpec>,
         static_file_provider: StaticFileProvider<N::Primitives>,
-    ) -> Self {
-        Self {
+    ) -> ProviderResult<Self> {
+        let legacy_settings = StorageSettings::legacy();
+        let storage_settings = DatabaseProvider::<_, N>::new(
```
mattsse, line 92: please document what's going on here, why are we doing this

mattsse (changes requested):
smol style suggestion

crates/storage/db-api/src/models/metadata.rs
joshieDo: https://github.com/paradigmxyz/reth/blob/642925011d6ccfe5aff18896f7eb279fd0ffb76e/crates/storage/storage-api/src/metadata.rs#L35-L39

this is also being encoded as `serde_json` so extending this should be fine

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -0,0 +1,39 @@
...
+/// These should be set during `init_genesis` or `init_db` depending on whether we want dictate
+/// behaviour of new or old nodes respectively.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize, Compact)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[add_arbitrary_tests(compact)]
+pub struct StorageSettings {
+    /// Whether this node always writes receipts to static files.
+    ///
+    /// If this is set to FALSE AND receipt pruning IS ENABLED, all receipts should be written to DB. Otherwise, they should be written to static files. This ensures that older nodes do not need to migrate their current DB tables to static files. For more, read: <https://github.com/paradigmxyz/reth/issues/18890#issuecomment-3457760097>
+    pub receipts_on_static_files: bool,
```
shekhirin: pedantic nit

```suggestion
    pub receipts_in_static_files: bool,
```

mattsse (approved):

PR #19387, opened 2025-10-29
refactor: extract ExecutionWitness construction into helper function

vtjl10:
Extracts duplicate logic for constructing `ExecutionWitness` from `ExecutionWitnessRecord` into a reusable helper function `build_execution_witness`.

### Changes

- Added `build_execution_witness` helper function in `reth-stateless` crate
- Updated `debug_execution_witness_for_block` in `reth-rpc` to use the helper
- Updated blockchain test case to use the helper instead of inline construction

### Motivation

The same logic for creating `ExecutionWitness` from `ExecutionWitnessRecord`, state witness, and headers was duplicated in two places:
1. `debug_execution_witness_for_block` method in `reth-rpc`
2. Blockchain test case in `ef-tests`

This refactoring centralizes the logic, making it easier to maintain and ensuring consistency across both usages.

### DIFF
--- a/testing/ef-tests/src/cases/blockchain_test.rs
+++ b/testing/ef-tests/src/cases/blockchain_test.rs
@@ -21,8 +21,8 @@ use reth_provider::{
 };
 use reth_revm::{database::StateProviderDatabase, witness::ExecutionWitnessRecord, State};
 use reth_stateless::{
-    trie::StatelessSparseTrie, validation::stateless_validation_with_trie, ExecutionWitness,
-    UncompressedPublicKey,
+    trie::StatelessSparseTrie, validation::stateless_validation_with_trie, build_execution_witness,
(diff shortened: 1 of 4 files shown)


testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -281,11 +281,9 @@ fn run_case(
             .map_err(|err| Error::block_failed(block_number, program_inputs.clone(), err))?;
 
         // Generate the stateless witness
-        // TODO: Most of this code is copy-pasted from debug_executionWitness
```
mattsse, line 284: this is fine, because this can't really be simplified and it's only a few loc

mattsse (commented):
just extracting the constructor fn isnt very useful

Issue #19373, opened 2025-10-28
support `pending` block tag in `eth_getLogs` for flashblocks


emiliano-conduitxyz:
### Describe the feature

Reth already support the full spec of flashblocks + eip7966

Pending logs are today retrievable via getReceipts using pending tag, but that means the filter/aggregation has to be made client side (getlogs till latest + aggregate all the receipts of all the fblocks till now) which is not great

This to add support directly in getLogs to make filtering on the node end for great UX and better performance (less latency and data fetch)

0xKarl98:
Is this task assignable ?

mattsse:
assigned, ty @0xKarl98

mattsse:
marking this as closed via #19388

PR #19388, opened 2025-10-29
feat: support pending block tag in eth_getLogs for flashblocks

0xKarl98:
This closes issue #19373 , please check if it met the requests @yongkangc  @mattsse  @Rjected


crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -487,32 +487,108 @@ where
...
+                    .is_some_and(|b| matches!(b, BlockNumberOrTag::Pending)) ||
+                    to_block.is_some_and(|b| matches!(b, BlockNumberOrTag::Pending));
+
+                if has_pending {
+                    // Handle pending blocks separately
+                    let mut all_logs = Vec::new();
+
+                    // If pending block is requested, try to get it from provider
+                    if let Ok(Some((block, receipts))) =
+                        self.provider().pending_block_and_receipts()
```
mattsse: her we'd need to use this fn

https://github.com/paradigmxyz/reth/blob/8a795e7d40535d7bd7c8e7b3493936a599a30377/crates/rpc/rpc-eth-api/src/helpers/block.rs#L170-L170

which means we need to add the trait bound

crates/rpc/rpc/src/eth/filter.rs
```diff
@@ -487,32 +487,108 @@ where
...
+                    let non_pending_to = if matches!(to_block, Some(BlockNumberOrTag::Pending)) {
+                        None
+                    } else {
+                        to_block
+                    };
+
+                    if non_pending_from.is_some() || non_pending_to.is_some() {
+                        // Process non-pending range using the original logic
+                        let info = self.provider().chain_info()?;
+                        let start_block = info.best_number;
```
mattsse: this part and below is a bit hard to read,

for now let's keep the main version unchanged and only handle the case where from to are both pending or start is pending

we can then generalize this later

mattsse (changes requested):
cool

left some comments, we can first focus on just the pending tag logic, we can then separately handle the case where we have `from(num)..pending` seperately

0xKarl98:
@mattsse  Have applied your suggests , please check again :)

mattsse (approved):

klkvr (approved):

PR #19390, opened 2025-10-29
fix(engine): trigger live sync after backfill completes at finalized

fgimenez:
Context: Fixes hive sync tests

Hive run https://github.com/paradigmxyz/reth/actions/runs/18916415942

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1575,9 +1575,11 @@ where
         };
 
         // Check if there are more blocks to sync between current head and FCU target
-        if let Some(head_block) = self.state.buffer.block(&sync_target_state.head_block_hash) {
+        if let Some(lowest_buffered) =
+            self.state.buffer.lowest_ancestor(&sync_target_state.head_block_hash)
+        {
             let current_head_num = self.state.tree_state.current_canonical_head.number;
-            let target_head_num = head_block.number();
+            let target_head_num = lowest_buffered.number();
 
             if let Some(distance) = self.distance_from_local_tip(current_head_num, target_head_num)
             {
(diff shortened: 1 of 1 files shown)


RomanHodulak (commented):
Awesome find, looks good assuming that current_head_num < target_head_num or distance_from_local_tip() can handle that.

Pending @mattsse

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1574,6 +1574,30 @@ where
             return Ok(())
         };
 
+        // Check if there are more blocks to sync between current head and FCU target
+        if let Some(head_block) = self.state.buffer.block(&sync_target_state.head_block_hash) {
```
mattsse: I think we can use `buffer.lowest_ancestor(head_block_hash)` here?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1574,6 +1574,30 @@ where
...
+                debug!(
+                    target: "engine::tree",
+                    %current_head_num,
+                    %target_head_num,
+                    %distance,
+                    "Backfill complete, downloading remaining blocks to reach FCU target"
+                );
+
+                self.emit_event(EngineApiEvent::Download(DownloadRequest::BlockRange(
+                    sync_target_state.head_block_hash,
```
mattsse: and then `head_block.parent_hash` here because we already have that block?

crates/engine/tree/src/tree/mod.rs
fgimenez: great point, done

crates/engine/tree/src/tree/mod.rs
fgimenez: indeed, no need to redownload it, using `lowest_buffered.parent_hash()` here now

mattsse (approved):
lgtm

PR #19392, opened 2025-10-29
chore: removing is_fork method

Ayushdubey86:
Hello @mattsse ! Do check this pr out once you're free!


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2474,6 +2422,11 @@ where
         self.state.tree_state.insert_executed(executed.clone());
         self.metrics.engine.executed_blocks.set(self.state.tree_state.block_count() as f64);
 
+        let is_fork = {
+            let executed_number = executed.recovered_block().number();
+            let canonical_number = self.state.tree_state.current_canonical_head.number;
+            executed_number <= canonical_number
+        };
```
yongkangc, line 2429: my view is that we should still keep the function to do is_fork check so that we can 
1. reduce cognitive load, allows us to add clear docstring as well
2. allow us to write unit test
3. since its single responsibility

yongkangc (changes requested):
it seems like we are overhauling the isfork logic to make it simpler? might want another look from @shekhirin @mattsse to see if its right!

mattsse:
closing this as dup of #19416

PR #19393, opened 2025-10-29
chore(primitives-traits): gate test-only modules

Galoretka:
- Gate withdrawal and log modules with #[cfg(test)]
- These files contain only #[cfg(test)] test code and no public items, so compiling them in non-test builds is unnecessary.
- This change prevents empty private modules from being built outside tests and avoids internal namespace noise without impacting public API or downstream crates.


mattsse (approved):
this makes sense, these only exist for tests

PR #19396, opened 2025-10-29
chore: fix unused deps

mattsse:
fix unused warnings


mattsse:
not dealing with this feature snafu

PR #19398, opened 2025-10-29
chore: reuse gzip read buffer to avoid per-iteration allocation

Forostovec:
Optimize gzip chunk reading by allocating the 64 KiB read buffer once per callto read_gzip_chunk() and reusing it across loop iterations. Previously, a newVec<u8> was allocated on each iteration, causing unnecessary allocations for large chunks. This change preserves behavior while reducing allocation pressure and aligning with standard AsyncRead buffering practices.


mattsse (approved):
makes sense

PR #19399, opened 2025-10-29
feat: always write receipts to static files on new nodes

joshieDo:
On top of https://github.com/paradigmxyz/reth/pull/19384

On old nodes we are writing receipts to database if pruning is enabled, otherwise to static files.

Moving forward and using `StorageSettings` the new logic should be:
```
if `old_node` & `any_receipt_pruning` -> mdbx
else if `new_node` & `receipt_log_filter` -> mdbx
else -> SF
```

For now, only enabled through `--static-files.receipts` when launching the node


crates/storage/db-common/src/init.rs
```diff
@@ -164,7 +164,7 @@ where
     static_file_provider.latest_writer(StaticFileSegment::Transactions)?.increment_block(0)?;
 
     // Behaviour reserved only for new nodes should be set here.
-    let storage_settings = StorageSettings::new();
+    let storage_settings = StorageSettings::new().with_receipts_in_static_files();
```
joshieDo: ensures that only new nodes will do it

crates/stages/stages/src/stages/execution.rs
```diff
@@ -185,11 +185,17 @@ where
...
     {
-        // If there's any receipts pruning configured, receipts are written directly to database and
-        // inconsistencies are expected.
-        if provider.prune_modes_ref().has_receipts_pruning() {
+        // On old nodes, if there's any receipts pruning configured, receipts are written directly
+        // to database and inconsistencies are expected.
+        if provider.prune_modes_ref().has_receipts_pruning() &&
+            !provider.cached_storage_settings().receipts_in_static_files
+        {
             return Ok(())
```
joshieDo, line 197: one downside here, is that the resulting size of a Full node at the end of pipeline sync will be momentarily larger because of receipts (which will be pruned).

shekhirin (dismissed):
LGTM, cc @mattsse

mediocregopher (approved):

mattsse (approved):

crates/stages/stages/src/stages/execution.rs
joshieDo: not anymore: https://github.com/paradigmxyz/reth/pull/19399/commits/ffa9181fb90d4ecf6a70335e4f1ca41f1fb2f101

crates/storage/provider/src/either_writer.rs
```diff
@@ -81,3 +82,36 @@ where
...
+impl EitherWriter<'_, (), ()> {
+    /// Returns the destination for writing receipts.
+    ///
+    /// The rules are as follows:
+    /// - If the node should not always write receipts to static files, and any receipt pruning is
+    ///   enabled, write to the database.
+    /// - If the node should always write receipts to static files, but receipt log filter pruning
+    ///   is enabled, write to the database.
+    /// - Otherwise, write to static files.
+    pub fn receipts_destination<P: DBProvider + StorageSettingsCache>(
```
joshieDo, line 95: taken from https://github.com/paradigmxyz/reth/pull/19630

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1620,8 +1646,16 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
 
         // All receipts from the last 128 blocks are required for blockchain tree, even with
         // [`PruneSegment::ContractLogs`].
-        let prunable_receipts =
-            PruneMode::Distance(MINIMUM_PRUNING_DISTANCE).should_prune(first_block, tip);
+        //
+        // Receipts can only be skipped if we're dealing with legacy nodes that write them to
+        // Database, OR if receipts_in_static_files is enabled but no receipts exist in static
+        // files yet. Once receipts exist in static files, we must continue writing to maintain
+        // continuity and have no gaps.
```
mattsse, line 1652: I see

mattsse (approved):
pending @shekhirin

shekhirin (approved):

mattsse (approved):

crates/node/core/src/args/static_files.rs
```diff
@@ -34,4 +44,13 @@ impl StaticFilesArgs {
...
         }
     }
+
+    /// Converts the static files arguments into [`StorageSettings`].
+    pub fn to_settings(&self) -> StorageSettings {
+        if self.receipts {
+            StorageSettings::new().with_receipts_in_static_files()
+        } else {
+            StorageSettings::legacy()
+        }
```
shekhirin, line 54: nit: think it's nicer to be able to do
```suggestion
        StorageSettings::legacy().with_receipts_in_static_files(self.receipts)
```

shekhirin (approved):

PR #19401, opened 2025-10-29
feat: prune receipts from static files

joshieDo:
* on top of https://github.com/paradigmxyz/reth/pull/19399

ensures we can prune receipts from static files (similar to how we do for txs)


crates/prune/prune/src/segments/mod.rs
```diff
@@ -14,6 +20,39 @@ pub use user::{
...
+    let deleted_headers =
+        provider.static_file_provider().delete_segment_below_block(segment, input.to_block + 1)?;
+
+    if deleted_headers.is_empty() {
+        return Ok(SegmentOutput::done())
+    }
+
+    let tx_ranges = deleted_headers.iter().filter_map(|header| header.tx_range());
+
+    let pruned = tx_ranges.clone().map(|range| range.len()).sum::<u64>() as usize;
```
mattsse, line 44: sum is weird here, can we do unwrap_or_default because this is jus a number

crates/prune/prune/src/segments/receipts.rs
```diff
@@ -21,8 +26,16 @@ where
...
+        + StorageSettingsCache
+        + StaticFileProviderFactory
         + NodePrimitivesProvider<Primitives: NodePrimitives<Receipt: Value>>,
 {
+    // Check if receipts are stored on static files
+    if provider.cached_storage_settings().receipts_on_static_files {
+        return segments::prune_static_files(provider, input, StaticFileSegment::Receipts)
+    }
+
+    // Original database implementation for when receipts are not on static files (old nodes)
```
mattsse: shoudl we introduce some traces here?

mattsse (approved):
lgtm!

shekhirin (approved):

PR #19404, opened 2025-10-30
chore(net): avoid cloning GetBlockBodies request

sashass1315:
- Replace InflightBodiesRequest payload from Vec<B256> to () and stop cloning the bodies request vector when inserting into inflight_bodies_requests in StateFetcher.
- The inflight request contents were not used for validation or retries; only the response channel is needed. This removes an unnecessary allocation/copy of potentially large vectors.


mattsse (approved):
this makes sense, because we currently dont do anything with this and the caller is responsible for this

PR #19406, opened 2025-10-30
fix(cli): Metrics log when passed metrics port 0

emhane:
Upstreaming https://github.com/op-rs/op-reth/pull/334

Bug: When no value was passed to metrics flag, a random port was assigned, but port 0 was shown as metric port in logs
Fix: Log actual used metrics port


mattsse (approved):
this makes sense because for port 0 we'd choose a random port so we should only log this after launch

pending @Rjected @shekhirin

shekhirin (approved):

PR #19407, opened 2025-10-30
fix: Prune checkpoint fixes

### LINKED ISSUE #19389: fix: panic on replay after multiple unwinds due to invalid MerkleChangeSets checkpoint range
### Describe the bug
Account worker panics when replaying after performing multiple consecutive unwinds due to invalid `MerkleChangeSets` checkpoint range.

### Expected behavior
Replay should execute successfully without panicking

### Steps to reproduce

  1. Unwind to block N (e.g., 23676119)
  2. Unwind again to block N-1 (e.g., 23676118)
  3. Attempt to replay with `reth-bench ... --advance 1`

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

main (commit `5091482dec7ccd7df03b349030eb1f0382a8f1d5`)

2

 mainnet

Full via --full flag

_No response_

### Code of Conduct

mediocregopher:
Fixes #19389 

This PR contains three interrelated fixes:

1) Do not update prune checkpoint during unwind unless the previous checkpoint is higher than the unwind target. Since unwinding doesn't actually affect the lower bound of a segment's computed block range (ie the prune checkpoint) it doesn't make sense to update the prune checkpoint in most cases.

2) Use an explicit array of `PruneSegment` variants in `get_prune_checkpoints`. This prevents us from attempting read prune checkpoints which are unknown to the codebase, which can happen when testing an older version of reth against a newer datadir which had a new segment added to it. This would previously cause an `unreachable` panic.

3) Always use prune checkpoint to track the computed lower bound of MerkleChangeSets. Trying to use both the prune checkpoint and stage checkpoint's block range resulted in edge cases; it's simpler to just use prune checkpoint.


crates/prune/types/src/segment.rs
```diff
@@ -36,10 +36,19 @@ pub enum PruneSegment {
...
+/// Array of [`PruneSegment`]s actively in use.
+pub const PRUNE_SEGMENTS: [PruneSegment; 7] = [
+    PruneSegment::SenderRecovery,
+    PruneSegment::TransactionLookup,
+    PruneSegment::Receipts,
+    PruneSegment::ContractLogs,
+    PruneSegment::AccountHistory,
+    PruneSegment::StorageHistory,
+    PruneSegment::MerkleChangeSets,
+];
```
joshieDo, line 53: if you rebase/merge main there's a new segment

crates/stages/stages/src/stages/prune.rs
```diff
@@ -104,8 +104,11 @@ where
         // checkpoints.
         let prune_checkpoints = provider.get_prune_checkpoints()?;
         for (segment, mut checkpoint) in prune_checkpoints {
-            checkpoint.block_number = Some(input.unwind_to);
-            provider.save_prune_checkpoint(segment, checkpoint)?;
+            // Only update the checkpoint if unwind_to is lower than the existing checkpoint.
+            if checkpoint.block_number.is_none_or(|block| input.unwind_to < block) {
```
joshieDo: why would we replace it if it's none?

joshieDo: we might have to add a similar logic here for when checkpoint.tx_number > highest tx number as well

joshieDo (approved):

PR #19410, opened 2025-10-30
chore: Use move-based append in with_resolved_peers to avoid allocation and clones

Fibonacci747:
Replace extend(...clone()) with a move-based transfer in with_resolved_peers():

- Use std::mem::take on NodeConfig.network.trusted_peers and append the  elements into Config.peers.trusted_nodes.
- Eliminates both the temporary Vec allocation and per-element cloning of  TrustedPeer (which may include domain strings).
- Semantics are preserved because the engine launch path does not read  NodeConfig.network.trusted_peers after with_resolved_peers(), and the  networking stack consumes the TOML-backed PeersConfig.trusted_nodes.
- CLI code paths that only have &self continue to clone by necessity.


mattsse (commented):
we don't want to mutate this here

PR #19412, opened 2025-10-30
chore: OverlayStateProviderFactory: don't query for reverts unless necessary

mediocregopher:
With this PR the factory will only make two DB calls in the happy case (anchor == db tip)
- Convert block hash to block number - this is necessary to prevent races with unwinding, which we identified tuesday during the audit
- Check the MerkleChangeSets stage checkpoint - check that the anchor == db tip


mattsse (approved):
lgtm

pending @shekhirin

shekhirin (approved):

PR #19413, opened 2025-10-30
fix(codecs): return remaining slice in EIP-1559 from_compact

CreeptoGengar:
Return the remaining buffer from inner TxEip1559::from_compact to maintain cursor advancement and fix chained deserialization.


mattsse (commented):
@joshieDo 

I dont think we should touch this?

joshieDo (approved):

PR #19414, opened 2025-10-30
chore: mark OpBuilderConfig::constrained_da_config as deprecated

Snezhkko:
The method is unused across the repository and external code search.
It is redundant with OpDAConfig::max_da_tx_size() and max_da_block_size() which already return Option<u64> for “no limit.”
Builders and RPC paths directly use OpDAConfig and its per-limit getters.


mattsse (commented):
we can keep this

Issue #19411, opened 2025-10-30
Add tracing features to node-core crate


mattsse:
### Describe the feature

in node-core we don't have features for tracing like:

https://github.com/paradigmxyz/reth/blob/d87d0d1a1f592fe2f5cee025acf29d7f68ab4c86/crates/ethereum/cli/Cargo.toml#L66-L70

## todo

* replicate features in here https://github.com/paradigmxyz/reth/blob/d87d0d1a1f592fe2f5cee025acf29d7f68ab4c86/crates/node/core/Cargo.toml#L76-L76
* run `zepter` to update feature forwarding

PR #19415, opened 2025-10-30
chore: add tracing features to node-core crate

kumaryash90:
Towards #19411


mattsse (approved):
ty, turns out we also need to add these to zepter's feature forwarding set

PR #19416, opened 2025-10-30
chore: simplify fork detection in `EngineApiTreeHandler`

### LINKED ISSUE #19359: remove is_fork fn
### Describe the feature

this function is only used for reporting

https://github.com/paradigmxyz/reth/blob/5091482dec7ccd7df03b349030eb1f0382a8f1d5/crates/engine/tree/src/tree/mod.rs#L937-L937

here:

https://github.com/paradigmxyz/reth/blob/5091482dec7ccd7df03b349030eb1f0382a8f1d5/crates/engine/tree/src/tree/mod.rs#L2480-L2485

we can derive the same info during:

https://github.com/paradigmxyz/reth/blob/5091482dec7ccd7df03b349030eb1f0382a8f1d5/crates/engine/tree/src/tree/mod.rs#L2475-L2475

because treestate has the current canonical head:

https://github.com/paradigmxyz/reth/blob/5091482dec7ccd7df03b349030eb1f0382a8f1d5/crates/engine/tree/src/tree/state.rs#L37-L38

so we have a fork if the `executed` blockheight is <= current_canonical_head
and can use this as `is_fork`

## TODO
* remove is_fork fn
* replace logic if height check

### Additional context

_No response_

Peponks9:
closes #19359

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -932,41 +932,6 @@ where
     /// * walking back from the current head to verify that the target hash is not already part of
     ///   the canonical chain.
     ///
-    /// The header is required as an arg, because we might be checking that the header is a fork
-    /// block before it's in the tree state and before it's in the database.
-    fn is_fork(&self, target: BlockWithParent) -> ProviderResult<bool> {
-        let target_hash = target.block.hash;
-        // verify that the given hash is not part of an extension of the canon chain.
-        let canonical_head = self.state.tree_state.canonical_head();
-        let mut current_hash;
-        let mut current_block = target;
-        loop {
-            if current_block.block.hash == canonical_head.hash {
-                return Ok(false)
-            }
-            // We already passed the canonical head
-            if current_block.block.number <= canonical_head.number {
-                break
-            }
-            current_hash = current_block.parent;
-
-            let Some(next_block) = self.sealed_header_by_hash(current_hash)? else { break };
-            current_block = next_block.block_with_parent();
-        }
-
-        // verify that the given hash is not already part of canonical chain stored in memory
-        if self.canonical_in_memory_state.header_by_hash(target_hash).is_some() {
-            return Ok(false)
-        }
-
-        // verify that the given hash is not already part of persisted canonical chain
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2502,6 +2458,10 @@ where
 
         // emit insert event
         let elapsed = start.elapsed();
+        // A block is considered a fork if its height is less than or equal to the current canonical
+        // head height
+        let is_fork =
+            executed.recovered_block().number() <= self.state.tree_state.canonical_block_number();
```
mattsse: @mediocregopher @shekhirin  I believe  in theory this might not detect all fork blocks

for example if we already have an alternative block for the same canonical_block_number height buffered and this block is + 1

so perhaps we do need walk up the chain in the state buffer until canonical height for this until we reach the current_canonical_head height?

which is similar to to the removed is_fork function but should just be this loop

https://github.com/paradigmxyz/reth/blob/6fafff5f145f99bb8cccf3e1b88ea009955ab648/crates/engine/tree/src/tree/mod.rs#L943-L955

and use the buffered blocks, because we've executed the block we know it connects back to the canonical chain

mediocregopher: Agreed, just checking the block height isn't sufficient, the fork could have a height equal to the canonical, or even one less and this new block is the one which makes it equal.

lean-apple:
```suggestion
let is_fork = executed.recovered_block().parent_hash() !=
            self.state.tree_state.canonical_block_hash();
``` 
Doing some follow-up there, 
@mediocregopher, @mattsse 
would this condition cover the edge cases, since a block is a fork if its parent isn't the current canonical head.

Peponks9 (commented):
@mattsse @mediocregopher thanks for the feedback after reading it defo agree. so would it be ok to proceed with replacing the height check with hash-based check?

mattsse (commented):
very sorry, I totally forgot we had this...

closing per https://github.com/paradigmxyz/reth/pull/20441

PR #10627, opened 2024-08-30
chore: remove PerfectHasingFunction Filters

nkysg:
Follow up https://github.com/paradigmxyz/reth/pull/10259

PR #19420, opened 2025-10-31
chore: inline hardcoded filter name in segment filename

rnkrtt:
Removes unnecessary .to_string() call in filename_with_configuration.

After #10627 removed the Filters enum, filters_name became a hardcoded "none" literal but was still being converted to
String unnecessarily. This just inlines it directly into the format macro.

Also removes the now-unused ToString import.

### DIFF
--- a/crates/static-file/types/src/segment.rs
+++ b/crates/static-file/types/src/segment.rs
@@ -1,8 +1,5 @@
 use crate::{BlockNumber, Compression};
-use alloc::{
-    format,
-    string::{String, ToString},
-};
+use alloc::{format, string::String};
 use alloy_primitives::TxNumber;
 use core::{ops::RangeInclusive, str::FromStr};
 use derive_more::Display;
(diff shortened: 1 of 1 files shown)


crates/static-file/types/src/segment.rs
```diff
@@ -83,11 +80,9 @@ impl StaticFileSegment {
     ) -> String {
         let prefix = self.filename(block_range);
 
-        let filters_name = "none".to_string();
-
         // ATTENTION: if changing the name format, be sure to reflect those changes in
         // [`Self::parse_filename`.]
-        format!("{prefix}_{}_{}", filters_name, compression.as_ref())
```
mattsse, line 90: can we keep it like this but remove the .to_string call?

mattsse (changes requested):
one nit

mattsse (approved):

PR #19423, opened 2025-10-31
feat: optimize ECIES body decode to avoid extra allocation/copy

sashass1315:
- Replace creating a new BytesMut and copying plaintext with truncating the split buffer based on the slice length returned by read_body.
- This preserves in-place decryption semantics and eliminates unnecessary allocation and memcpy on every message body, reducing overhead under load.


mattsse (commented):
i dont want to change this

PR #19424, opened 2025-10-31
fix(beacon-api-sidecar): use correct block metadata for reorged blobs

avorylli:
Use metadata from the actual reorged block instead of new.tip() when creating ReorgedBlob events. 
Previously, all reorged transactions from different blocks received the same metadata from the new chain tip, losing information about which block each transaction originally belonged to.


avorylli:
also fixed clippy although it wasnt due to my changes,

blockheader hasn't been used so i deleted it

mattsse (approved):
this is indeed correct

PR #19425, opened 2025-10-31
feat(reth-bench): Default --wait-time to 250ms

mediocregopher:
When reth is not given some breathing room to persist blocks performance degrades significantly. Since this is not representative of real-world conditions we should default to waiting a short time between blocks.


shekhirin (approved):

joshieDo (approved):

yongkangc:
nice

PR #19427, opened 2025-10-31
perf: only chunk if more > 1 available

mattsse:
there's no point in chunking if we only have 1 idle worker

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -208,12 +208,12 @@ impl ProofWorkerHandle {
 
     /// Returns true if there are available storage workers to process tasks.
     pub fn has_available_storage_workers(&self) -> bool {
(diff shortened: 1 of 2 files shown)


crates/trie/parallel/src/proof_task.rs
```diff
@@ -208,12 +208,12 @@ impl ProofWorkerHandle {
 
     /// Returns true if there are available storage workers to process tasks.
     pub fn has_available_storage_workers(&self) -> bool {
-        self.storage_available_workers.load(Ordering::Relaxed) > 0
+        self.storage_available_workers.load(Ordering::Relaxed) > 1
```
shekhirin: let's return just the usize from this method and check on caller side, otherwise this method doesn't just show if there are available workers anymore

shekhirin (approved):
agreed

yongkangc (approved):
makes sense

PR #19430, opened 2025-10-31
chore: BuiltPayloadExecutedBlock

mediocregopher:
Introduce a copy of the ExecutedBlock type for use specifically in payload building. This will let us diverge the ExecutedBlock type which is used for validation without affecting performance of block building.


mattsse (commented):
this separation makes sense imo

we could/should support `either` representation here

crates/payload/primitives/src/traits.rs
```diff
@@ -1,16 +1,58 @@
...
+    /// Block's execution outcome.
+    pub execution_output: Arc<ExecutionOutcome<N::Receipt>>,
+    /// Block's hashed state.
+    pub hashed_state: Either<Arc<HashedPostState>, Arc<HashedPostStateSorted>>,
+    /// Trie updates that result from calculating the state root for the block.
+    pub trie_updates: Either<Arc<TrieUpdates>, Arc<TrieUpdatesSorted>>,
+}
+
+impl<N: NodePrimitives> From<BuiltPayloadExecutedBlock<N>> for reth_chain_state::ExecutedBlock<N> {
+    fn from(block: BuiltPayloadExecutedBlock<N>) -> Self {
```
mattsse: can we move the impl of this function to `BuiltPayloadExecutedBlock::into_executed_payload`

which the from impl then calls,

I find this a lot easier because then this shows up with intellisense and I find `.into_executed_payload()` instead of `.into()` easier to follow

crates/payload/primitives/src/traits.rs
```diff
@@ -1,16 +1,58 @@
...
+/// including the recovered block, execution outcome, hashed state, and trie updates.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BuiltPayloadExecutedBlock<N: NodePrimitives> {
+    /// Recovered Block
+    pub recovered_block: Arc<RecoveredBlock<N::Block>>,
+    /// Block's execution outcome.
+    pub execution_output: Arc<ExecutionOutcome<N::Receipt>>,
+    /// Block's hashed state.
+    pub hashed_state: Either<Arc<HashedPostState>, Arc<HashedPostStateSorted>>,
+    /// Trie updates that result from calculating the state root for the block.
```
mattsse, line 35: can we add an additional oneliner to each, why this supports both variants

mattsse (changes requested):
this makes sense

very pedantic nits

yongkangc (approved):
Lgtm, thanks for this

mattsse (approved):
lgtm

PR #19432, opened 2025-10-31
feat(banlist): add time-aware ban checks (`is_banned_ip_at`, `is_bann…

gap-editor:
Adds new helper methods to check whether a peer/IP is *currently* banned without requiring an
explicit `evict(now)`. This improves call-site ergonomics and reduces the risk of stale ban
entries influencing logic. No behavior changes to existing methods.


mattsse (commented):
we currently dont have a usecase for this

PR #19434, opened 2025-10-31
chore: avoid allocating empty AccountStorageCache when no storage entries

sashass1315:
Avoid allocating and inserting an empty per-account storage cache when there are no storage entries to add.

Previously, insert_storage_bulk() always created and inserted an Arc<AccountStorageCache> even if the iterator was empty. This led to unnecessary allocations and capacity consumption.

The new logic early-returns on empty iterators and only re-inserts into storage_cache when at least one slot is actually inserted. This preserves behavior (get_storage semantics are unchanged) while eliminating waste and keeping weigher/TTL refreshes only when the cache genuinely changes.


mattsse (commented):
dont think we need this

PR #19435, opened 2025-10-31
fix(payload): emit events for Freeze payload outcomes

avorylli:
Handle BuildOutcome::Freeze variant in BetterPayloadEmitter by sending frozen payloads to the broadcast channel, same as Better outcomes. 

This ensures subscribers receive all valid payload events, including final frozen payloads used in deterministic scenarios.


mattsse (approved):
makes sense

PR #19438, opened 2025-11-01
feat(tracing-otlp): make trace id ratio sample customizable with `--tracing-otlp.sample-ratio` arg

### LINKED ISSUE #19282: Allow to control the otlp traces sample ratio from 0.0 to 1.0
### Describe the feature

Right now, the `SpanExporter` use the sample ratio set up by [default](https://docs.rs/opentelemetry_sdk/0.31.0/src/opentelemetry_sdk/trace/provider.rs.html#93) with 
`sampler: Box::new(Sampler::ParentBased(Box::new(Sampler::AlwaysOn)))`, which samples every trace, from parent to child.

It is done through the builder initialization :
https://github.com/paradigmxyz/reth/blob/f29f4caf0e8ad19093967a446d80c023e3b0028b/crates/tracing-otlp/src/lib.rs#L44

Yet we might need to decrease this ratio.

We can make the sampler traces ratio configurable through the CLI args [`TraceArg`](https://github.com/paradigmxyz/reth/blob/dc781126c249f020d7abdac197fdd4bcb848ac28/crates/node/core/src/args/trace.rs#L11)

The behavior by default should be kept if no float provided through the CLI, and if specified it should be between 0 and 1.

See also https://docs.rs/opentelemetry_sdk/0.31.0/opentelemetry_sdk/trace/enum.Sampler.html#variant.TraceIdRatioBased for the configurable approach.

Discussed with @Rjected

### Additional context

Also this method https://docs.rs/opentelemetry_sdk/0.31.0/opentelemetry_sdk/trace/struct.TracerProviderBuilder.html#method.with_sampler may be useful to set up the configured sampler.

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/19282.

- Add a new arg `--tracing-otlp.sample-ratio` to enable to configure the ratio of traces exported

- Add `OtlpConfig` to avoid the long list of args for some functions


Rjected (approved):
LGTM!

lean-apple:
> @lean-apple do you mind rebasing and running `make update-book-cli`

Sure, it should be ok now

PR #19439, opened 2025-11-01
perf(era-utils): optimize accumulator hash extraction

CreeptoGengar:
Store only first 32 bytes of compressed header data instead of cloning the entire vector. 
Reduces memory allocations and improves performance when processing large block ranges during era1 export.


mattsse (commented):
we dont need to change this

PR #19440, opened 2025-11-01
refactor(provider, cli): simplify getting provider for index or range

shekhirin:
1. Introduces `StaticFileProvider::get_segment_provider` to deduplicate the logic of getting either block or transaction provider based on the segment.
2. Improves perf for `reth db get static-file` CLI by getting the necessary segment provider using index instead of probing all segments starting from the end.


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -365,14 +365,28 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+    pub fn get_segment_provider(
+        &self,
+        segment: StaticFileSegment,
+        start: u64,
+    ) -> ProviderResult<StaticFileJarProvider<'_, N>> {
+        if segment.is_block_based() {
+            self.get_segment_provider_for_block(segment, start, None)
+        } else {
+            self.get_segment_provider_for_transaction(segment, start, None)
+        }
```
joshieDo, line 411: cc @Rjected just in case

joshieDo (approved):

crates/storage/provider/src/providers/static_file/manager.rs
Rjected: makes sense, for changesets I think we will want to continue using `segment_provider_for_block` for getting segments

PR #19442, opened 2025-11-01
fix: refocus txpool example description on pool validator

crStiv:
Refocuses txpool example description on the transaction pool with custom validator instead of network component to match the Mempool section focus.


mattsse (commented):
the exisitng wording is fine here

PR #19444, opened 2025-11-02
fix: honor OpenDiskFileBlobStore and implement reindex

phrwlk:
- Wire OpenDiskFileBlobStore in DiskFileBlobStore::open() to support both Clear (delete+create) and ReIndex (create if missing and rebuild in-memory state).
- Implement DiskFileBlobStoreInner::reindex() to scan the blob directory, reconstruct size_tracker, and rebuild versioned_hashes_to_txhash from decoded sidecars.
- Add a unit test to verify reindex preserves visibility of on-disk blobs across restarts.
- This change fixes the previously ignored open option and ensures blobstore contents remain discoverable after process restarts, aligning implementation with the documented API.


mattsse (commented):
imo just wiping on restart is appropriate

PR #19447, opened 2025-11-02
fix: compare execution witness codes/keys as sets

DeVikingMark:
The comparison between re-executed and healthy node witnesses was failing due to order sensitivity in codes and keys vectors. ExecutionWitnessRecord may produce different ordering than our BTreeMap-based collection.

This change converts codes and keys to BTreeSets before comparison, making the comparison order-insensitive while preserving the same semantic meaning (since these fields represent unordered sets of bytecode and preimages).

### DIFF
--- a/crates/engine/invalid-block-hooks/src/witness.rs
+++ b/crates/engine/invalid-block-hooks/src/witness.rs
@@ -21,7 +21,7 @@ use revm_database::{
 };
 use serde::Serialize;
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::BTreeMap,
     fmt::Debug,
     fs::File,
     io::Write,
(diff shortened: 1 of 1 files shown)


crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -246,7 +252,23 @@ where
             let filename = format!("{}.witness.healthy.json", block_prefix);
             let healthy_path = self.save_file(filename, &healthy_node_witness)?;
 
-            if witness != &healthy_node_witness {
+            // Compare witness as sets to avoid order sensitivity in codes and keys vectors
+            let codes_match = {
+                let re_executed_codes: BTreeSet<_> = witness.codes.iter().collect();
```
mediocregopher: Rather than collecting fields into new BTreeSets, could we just sort the Vecs in place? I would be supportive of an `ExecutedWitness::normalize(&mut self)` method, which sorts all fields, and call that here for both witnesses.

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -246,7 +263,15 @@ where
...
+            let mut normalized_witness = witness.clone();
+            let mut normalized_healthy_witness = healthy_node_witness.clone();
+            normalized_witness.normalize();
+            normalized_healthy_witness.normalize();
+
+            let witness_matches = normalized_witness == normalized_healthy_witness;
+
+            if !witness_matches {
                 let filename = format!("{}.witness.diff", block_prefix);
                 let diff_path = self.save_diff(filename, witness, &healthy_node_witness)?;
```
mediocregopher: Do we want to use the normalized versions for the diff as well? The implementation looks to me like a line-by-line diff, so normalized would be better I'd think

crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -25,6 +25,23 @@ use std::{collections::BTreeMap, fmt::Debug, fs::File, io::Write, path::PathBuf}
...
+}
+
+impl ExecutionWitnessExt for ExecutionWitness {
+    fn normalize(&mut self) {
+        // Sort codes and keys to ensure order-insensitive comparison
+        // These fields represent unordered sets of bytecode and preimages
+        self.codes.sort();
+        self.keys.sort();
+    }
+}
```
Rjected: Do we need a trait for this? Can we not just call `codes.sort()` and `keys.sort()` in place

DeVikingMark:
Maybe we can do something like this?

mattsse (commented):
we can keep this as is

PR #19449, opened 2025-11-02
fix(engine): use actual pool size instead of notification counter

avorylli:
The accumulated counter was tracking notification events rather than actual
transactions in the pool, causing blocks to be mined at incorrect thresholds
when transactions were removed or discarded.

Replaced the notification-based counter with a direct check of the pool's
transaction count, ensuring blocks are mined only when the actual number of
transactions reaches the configured threshold.


mattsse (commented):
pending txs aren't executable, so we cant use the sum of here

PR #19451, opened 2025-11-02
fix: avoid unnecessary self.clone() in OpNetworkBuilder::network_config

sashass1315:
Replace self.clone() destructure with direct bool copies in OpNetworkBuilder::network_config. OpNetworkBuilder only holds bool fields, so cloning the builder to extract flags is redundant.
This change avoids an unnecessary allocation/copy of the struct, simplifies the code, and aligns with patterns used elsewhere in the codebase. No functional changes.


mattsse (approved):
makes sense

PR #19457, opened 2025-11-03
refactor(rpc): remove redundant comment in bundle execution

Fallengirl:
Remove redundant duplicate comment in bundle transaction processing. The comment "need to apply the state changes of this call before executing the next call" was duplicated both outside and inside the if block. Kept only the outer comment to improve code readability


mattsse (commented):
we can keep this

Issue #19433, opened 2025-10-31
Improve flashblock ethsendrawsync


mattsse:
### Describe the feature

we can still do a few things here

Syzygy106:
Hmmm? What the things?

PR #19459, opened 2025-11-03
feat: add broadcast channel for received flashblocks

mattsse:
more prep for https://github.com/paradigmxyz/reth/issues/19433

this introduces a new broadcast channel for received flashblocks

and changes the broadcast rx type to sender in the flashblockslistner, because otherwise this would always keep a subscription open and fully buffer the receiver

### DIFF
--- a/crates/optimism/flashblocks/src/lib.rs
+++ b/crates/optimism/flashblocks/src/lib.rs
@@ -50,7 +50,7 @@ pub type InProgressFlashBlockRx = tokio::sync::watch::Receiver<Option<FlashBlock
 
 /// Container for all flashblocks-related listeners.
 ///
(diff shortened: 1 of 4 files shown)


crates/optimism/flashblocks/src/lib.rs
```diff
@@ -47,31 +53,24 @@ pub type InProgressFlashBlockRx = tokio::sync::watch::Receiver<Option<FlashBlock
 /// Groups together the three receivers that provide flashblock-related updates.
```
fgimenez: we no longer have three receivers
```suggestion
/// Groups together the channels for flashblock-related updates.
```

fgimenez (commented):
smol doc nit, lgtm

PR #19460, opened 2025-11-03
docs: add error hierarchy documentation to trie error types

DeVikingMark:
Adds module-level documentation explaining the error type hierarchy and relationships between trie error types. Improves individual error type documentation to clarify when each type should be used and how they relate to each other.

This makes it easier for developers to understand which error type to use in different contexts and how errors propagate through the system.


mattsse (commented):
this will change shortly so we dont really need this

PR #19462, opened 2025-11-03
perf: improve ethsendrawsync for op with flashblock

### LINKED ISSUE #19433: Improve flashblock ethsendrawsync

mattsse:
closes #19433

the current impl is suboptimal for flashblocks, we can populate the receipt as soon as we receive the flashblock from the websocket subscription whcih includes the OpReceipt and the block data


klkvr (approved):
nice

PR #19463, opened 2025-11-03
refactor(node-builder): use explicit loop instead of map in invalid block hook creation

CreeptoGengar:
Replace functional `.map().collect()` pattern with an explicit `for` loop when creating invalid block hooks. 
This improves code readability while maintaining the same behavior. 
Each hook still receives its own cloned instances of provider and evm_config, which is necessary since hooks take ownership.


mattsse (commented):
this map collect is fine

PR #19464, opened 2025-11-03
fix(trie): InMemoryTrieCursor case where all DB nodes are deleted

mediocregopher:
This was pointed out by @0x00101010 in a [separate PR](https://github.com/paradigmxyz/reth/pull/19233#discussion_r2483813862). Basically if the in-memory overlay completely deleted all nodes in the trie table, then there was no way to distinguish the cases of the db cursor being exhausted or having never been used; the `cursor_entry` field would be `None` in both cases.


Rjected (approved):
This makes sense to me, I wonder if, to make it more explicit, we should have an enum for `cursor_entry`, which would be:
```rust
enum CursorState {
    /// Initial cursor state, when we haven't seeked, open to renames
    Initial
    /// The cursor is currently pointing to an entry.
    Entry {
        /// Path to the node
        path: Nibbles,
        /// The node
        node: BranchNodeCompact,
    },
    /// The DB cursor is exhausted
    Exhausted
}
```

PR #19465, opened 2025-11-03
refactor(prune): derive EnumIter instead of explicit array of segments

shekhirin:
Follow-up to https://github.com/paradigmxyz/reth/pull/19407

### DIFF
--- a/crates/prune/types/Cargo.toml
+++ b/crates/prune/types/Cargo.toml
@@ -17,6 +17,7 @@ reth-codecs = { workspace = true, optional = true }
 alloy-primitives.workspace = true
 derive_more.workspace = true
 strum.workspace = true
+strum_macros.workspace = true
 thiserror.workspace = true
 
 modular-bitfield = { workspace = true, optional = true }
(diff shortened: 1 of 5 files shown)


mediocregopher (approved):

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -83,6 +83,7 @@ use std::{
     ops::{Deref, DerefMut, Not, Range, RangeBounds, RangeFrom, RangeInclusive},
     sync::Arc,
 };
+use strum::IntoEnumIterator;
```
mattsse: one trick how to avoid this is

`fn prune_variants(&self) {Self::iter()}`

then we can also document properly why this is excluding a few variants

mattsse (approved):

crates/prune/types/Cargo.toml
```diff
@@ -16,6 +16,8 @@ reth-codecs = { workspace = true, optional = true }
 
 alloy-primitives.workspace = true
 derive_more.workspace = true
+strum.workspace = true
+strum_macros.workspace = true
```
DaniPopes: why is this imported separately?

crates/storage/provider/src/providers/database/provider.rs
shekhirin: nice, fixed

PR #19469, opened 2025-11-03
fix: update `min_block` on `StaticFileProvider::update_index`

joshieDo:
Ensures that `min_block` is updated when the node only has 1 static file of the segment.

Without this PR:
* If `reth node` is spawned completely from scratch, `get_lowest_static_file_block` would always report `None`. And we would never prune until a restart.
* If `reth node` is spawned after an initial `init_genesis` but without progressing the chain (eg. reth init-state), `min_block` would always report `[0,0]` and would report the wrong `block_height` below, leading to deleting the only static file prematurely https://github.com/paradigmxyz/reth/blob/846025545cb0d15f09a098d6a39f2f85d0559f40/crates/storage/provider/src/providers/static_file/manager.rs#L476-L483

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -615,6 +615,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         segment: StaticFileSegment,
         segment_max_block: Option<BlockNumber>,
     ) -> ProviderResult<()> {
+        let mut min_block = self.static_files_min_block.write();
         let mut max_block = self.static_files_max_block.write();
         let mut tx_index = self.static_files_tx_index.write();
 
@@ -628,6 +629,22 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                     &self.path.join(segment.filename(&fixed_range)),
                 )
                 .map_err(ProviderError::other)?;
+
+                // When the node is first started from scratch this min_block is set but needs to be
+                // updated as we progress the chain. Examples:
+                // - init: None -> [0..=100]
+                // - init: [0..=0], update with [0..=100] (eg. `op-reth node` after `reth
+                //   init-state`)
+                if let Some(current_block_range) = jar.user_header().block_range().copied() {
+                    min_block
+                        .entry(segment)
+                        .and_modify(|current_min| {
+                            if current_block_range.start() == current_min.start() {
+                                *current_min = current_block_range;
+                            }
+                        })
+                        .or_insert(current_block_range);
+                }
 
                 // Updates the tx index by first removing all entries which have a higher
                 // block_start than our current static file.
@@ -678,6 +695,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
             None => {
                 tx_index.remove(&segment);
                 max_block.remove(&segment);
+                min_block.remove(&segment);
             }
         };
 
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -627,10 +627,9 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                 max_block.insert(segment, segment_max_block);
                 let fixed_range = self.find_fixed_range(segment_max_block);
 
-                let jar = NippyJar::<SegmentHeader>::load(
-                    &self.path.join(segment.filename(&fixed_range)),
-                )
-                .map_err(ProviderError::other)?;
+                let path = self.path.join(segment.filename(&fixed_range));
+                let jar = NippyJar::<SegmentHeader>::load(&path)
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -629,6 +630,22 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+                // When the node is first started from scratch this min_block is set but needs to be
+                // updated as we progress the chain. Examples:
+                // - init: None -> [0..=100]
+                // - init: [0..=0], update with [0..=100] (eg. `op-reth node` after `reth
+                //   init-state`)
+                if let Some(current_block_range) = jar.user_header().block_range().copied() {
+                    min_block
+                        .entry(segment)
+                        .and_modify(|current_min| {
+                            if current_block_range.start() == current_min.start() {
```
shekhirin, line 656: why don't we compare them and update the `current_min` only if `current_block_range.start() < current_min.start()`?

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin:
```suggestion
```

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -629,6 +630,22 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                 )
                 .map_err(ProviderError::other)?;
 
+                // When the node is first started from scratch this min_block is set but needs to be
+                // updated as we progress the chain. Examples:
+                // - init: None -> [0..=100]
+                // - init: [0..=0], update with [0..=100] (eg. `op-reth node` after `reth
+                //   init-state`)
```
mattsse: not 100% instantly obvious what this really means exactly

mattsse (changes requested):

crates/storage/provider/src/providers/static_file/manager.rs
joshieDo: because start doesn't change, only the end range does. example:
* in the beginning of a node: `0..=0 -> 0..=1` after inserting block 1.

joshieDo: if we're inserting block 500k, `min_block` shouldnt be updated here and should remain `0..=499_999`, since thats the lowest/min range.

if we ever delete`0..=499_999`,  the whole index gets reinitialized with `Self::initialize_index()`, which would set it to 500k..=500k

mattsse (approved):

mattsse (approved):

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin, line 231: let's update this doc to mention that it also gets updated in `update_index`?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -624,10 +627,37 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                 max_block.insert(segment, segment_max_block);
                 let fixed_range = self.find_fixed_range(segment_max_block);
 
-                let jar = NippyJar::<SegmentHeader>::load(
-                    &self.path.join(segment.filename(&fixed_range)),
-                )
-                .map_err(ProviderError::other)?;
+                let path = self.path.join(segment.filename(&fixed_range));
+                let jar = NippyJar::<SegmentHeader>::load(&path)
+                    .map_err(|e| ProviderError::other(format!("Failed to load {path:?}: {e}")))?;
```
shekhirin: instead of this, we should use our fs wrapper here

https://github.com/paradigmxyz/reth/blob/ea69063aae501cd24f383515ec8bfd3869d3f5c3/crates/storage/nippy-jar/src/lib.rs#L202

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -624,10 +627,37 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+                // Without this update, min_block can remain at genesis (e.g. Some([0..=0]) or None)
+                // even after syncing to higher blocks (e.g. [0..=100]). A stale
+                // min_block causes get_lowest_static_file_block() to return the
+                // wrong end value, which breaks pruning logic that relies on it for
+                // safety checks.
+                //
+                // Example progression:
+                // 1. Node starts, initialize_index() sets min_block = [0..=0]
+                // 2. Sync to block 100, this update sets min_block = [0..=100]
+                // 3. Pruner calls get_lowest_static_file_block() -> returns 100 (correct) Without
```
shekhirin:
```suggestion
                // 3. Pruner calls get_lowest_static_file_block() -> returns 100 (correct). Without
```

shekhirin (approved):

PR #19476, opened 2025-11-03
chore: add queued reason to event

mattsse:
we can attach the reason why this tx was queued to the event


Rjected (approved):
seems useful

PR #19477, opened 2025-11-03
chore(engine): remove unused BlockReceived event variant

aharvey101:
## Summary
Remove the unused `BlockReceived` variant from `ConsensusEngineEvent` and all associated handling. This variant is no longer useful and was only used for reporting. The removal also eliminates the dependency on `alloy-eips`.

## Changes
- Removed `BlockReceived(BlockNumHash)` variant from `ConsensusEngineEvent` enum
- Removed Display impl handling for the variant
- Removed unused import `alloy_eips::BlockNumHash`
- Removed `alloy-eips` dependency from `crates/engine/primitives`
- Removed event emission in `crates/engine/tree/src/tree/mod.rs`
- Removed event handling in `crates/node/events/src/node.rs`
- Removed event handling in `crates/ress/provider/src/pending_state.rs`

## Testing
- All existing tests pass
- Formatting validated with `cargo +nightly fmt --all`
- Linting validated with clippy

Closes #19474

### DIFF
--- a/examples/exex-subscription/src/main.rs
+++ b/examples/exex-subscription/src/main.rs
@@ -6,8 +6,7 @@
 use alloy_primitives::{Address, U256};
 use futures::TryStreamExt;
 use jsonrpsee::{
-    core::SubscriptionResult, proc_macros::rpc, tracing, PendingSubscriptionSink,
-    SubscriptionMessage,
+    core::SubscriptionResult, proc_macros::rpc, PendingSubscriptionSink, SubscriptionMessage,
 };
 use reth_ethereum::{
     exex::{ExExContext, ExExEvent, ExExNotification},


crates/engine/tree/src/tree/mod.rs
```diff
@@ -538,9 +538,6 @@ where
         // This validation **MUST** be instantly run in all cases even during active sync process.
 
         let num_hash = payload.num_hash();
-        let engine_event = ConsensusEngineEvent::BlockReceived(num_hash);
-        self.emit_event(EngineApiEvent::BeaconConsensus(engine_event));
```
mattsse, line 541: oh I was actually wrong, this is used and we should keep it then

let's just only do the `LiveSyncProgress` variant then

mattsse (changes requested):

PR #19479, opened 2025-11-04
feat(flashblocks): Flashblocks engine sidecar

0x00101010:
Towards: https://github.com/paradigmxyz/reth/issues/18255, see bottom for follow up needed.

# feat(op-reth): integrate flashblocks with consensus engine

## Summary

This PR completes the integration of flashblocks with the consensus engine by:

- Adding structured conversion infrastructure from flashblock sequences to execution payloads
- Automatically spawning the consensus client when flashblocks are enabled
- Improving error handling and observability with detailed logging

### Integration

**`crates/optimism/rpc/src/eth/mod.rs`**
- Added automatic spawning of `FlashBlockConsensusClient` when flashblocks are enabled
- Wired up `beacon_engine_handle` to enable payload submission

**`crates/node/builder/src/rpc.rs`**
- Exposed `beacon_engine_handle` in `EthApiCtx` for RPC builders to access

**`examples/custom-node/src/engine.rs`**
- Added example implementation of `From<&FlashBlockCompleteSequence>` for custom execution data types
- Shows how to derive custom extensions from flashblock sequences

### Dependencies

**`crates/optimism/flashblocks/Cargo.toml`**
- Added: `op-alloy-rpc-types-engine` (required for `OpExecutionData`)
- Removed: `ringbuffer` (no longer needed with simplified FCU logic)

## Testing

- All existing tests pass
- New unit tests for conversion functions
- Clippy clean with all features enabled

## Follow up needed

- ~~handle optional state root calculations scenario~~ Done
- ~~handle Jovian upgrade situation where blob_gas_used is not 0, need to update `rollup-boost`, `op-rbuilder`, and this repo for consistent types.~~ Done
- ~~ideally we move all the flashblocks relevant data structures into `op-alloy-rpc-types-engine`~~ Done
- add functionality to work with different forks


crates/optimism/flashblocks/src/convert.rs
```diff
@@ -0,0 +1,182 @@
...
+//!
+//! This module provides the canonical implementations for converting flashblock data structures
+//! into the various versions of execution payloads required by the Engine API.
+//!
+//! TODO: ideally we move this to op-alloy-rpc-types-engine where `from_block_unchecked` is defined
+//! and define a similar called `from_flashblock_sequence_unchecked`.
+//! The goal is to have the conversion logic in one place to avoid missing steps.
+
+use crate::{
+    ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashBlockCompleteSequence,
```
0x00101010: @mattsse what's your thoughts on moving above data structures into `op-alloy-rpc-types-engine`?

Currently they're defined in rollup-boost and imported by op-rbuilder, etc.

If we're making reth to support flashblocks natively, it seems reasonable to move those into `op-alloy-rpc-types-engine` and update all other repos to refer to it.

This helps with
1. centralize commonly shared data structures
2. also we can move this convert.rs functionality directly into `op-alloy-rpc-types-engine` and sit alongside with OpExecutionData to share the same block / flashblock -> `ExecutionData` conversion logic

mattsse: > @mattsse what's your thoughts on moving above data structures into op-alloy-rpc-types-engine?

yeah that would be ideal, even if the format changes we can still make that work

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -1,80 +1,115 @@
...
+                        finalized_block_hash: B256::ZERO,
                     };
 
-                    // Send FCU
-                    let _ = self
+                    match self
                         .engine_handle
-                        .fork_choice_updated(state, None, EngineApiMessageVersion::V3)
-                        .await;
+                        .fork_choice_updated(fcu_state, None, EngineApiMessageVersion::V3)
```
mattsse: version shouldnt matter here, so this seems okay

crates/optimism/flashblocks/src/convert.rs
mattsse, line 1: yep let's move all of this to op-alloy

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -459,7 +464,12 @@ where
                                  + From<ExecutionPayloadBaseV1>
                                  + Unpin,
         >,
-        Types: NodeTypes<ChainSpec: Hardforks + EthereumHardforks>,
+        Types: NodeTypes<
+            ChainSpec: Hardforks + EthereumHardforks,
+            Payload: reth_node_api::PayloadTypes<
+                ExecutionData: for<'a> From<&'a FlashBlockCompleteSequence> + ExecutionPayload,
```
mattsse: I think I'd prefer a TryFrom requirement here, just so that you can still use all of these types if you cant satisfy this conversion

crates/node/builder/src/rpc.rs
```diff
@@ -1137,6 +1142,8 @@ pub struct EthApiCtx<'a, N: FullNodeTypes> {
     pub config: EthConfig,
     /// Cache for eth state
     pub cache: EthStateCache<PrimitivesTy<N::Types>>,
+    /// Handle to the beacon consensus engine
+    pub beacon_engine_handle: ConsensusEngineHandle<<N::Types as NodeTypes>::Payload>,
```
mattsse: let's name this engine_handle because I want to get rid of this beacon_ prefix where possible

mattsse: great, this is something I wanted to for a while

mattsse (changes requested):
actually pretty simple, very nice

imo we should first offload all things flashblock types to op-alloy and then merge this

crates/node/builder/src/rpc.rs
0x00101010: done

0x00101010:
> do you mind rebasing this on top of #19608 so that it's easier to review?

Done

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -1,86 +1,143 @@
...
-                    // Send FCU
-                    let _ = self
-                        .engine_handle
-                        .fork_choice_updated(state, None, EngineApiMessageVersion::V3)
-                        .await;
+            // FCU to update tip, only after successful engine_newPayload
+            let fcu_state = alloy_rpc_types_engine::ForkchoiceState {
+                head_block_hash: block_hash,
+                safe_block_hash: B256::ZERO,
+                finalized_block_hash: B256::ZERO,
```
klkvr, line 97: what's the motivation for setting those to zero? not sure how to properly set those if the only source of truth we have are flashblocks but maybe we could keep the `head-32`?

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -1,86 +1,143 @@
...
     }
 }
+
+impl From<&FlashBlockCompleteSequence> for OpExecutionData {
+    fn from(sequence: &FlashBlockCompleteSequence) -> Self {
+        let mut data = Self::from_flashblocks_unchecked(sequence);
+        // Replace payload's state_root with the calculated one. For flashblocks, there was an
+        // option to disable state root calculation for blocks, and in that case, the payload's
+        // state_root will be zero, and we'll need to locally calculate state_root before
+        // proceeding to call engine_newPayload.
```
klkvr, line 135: I'm wondering if we should somehow remove the need to compute state root here entirely?

engine will anyway end up computing it when verifying the payload, so this would just mean we do the work twice? cc @mattsse 

though I guess it's kind of tricky because engine relies on the `newPayload` to come with a computed block hash

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -506,9 +515,15 @@ where
             let flashblocks_sequence = service.block_sequence_broadcaster().clone();
             let received_flashblocks = service.flashblocks_broadcaster().clone();
             let in_progress_rx = service.subscribe_in_progress();
-
             ctx.components.task_executor().spawn(Box::pin(service.run(tx)));
 
+            info!(target: "reth::cli", "Launching FlashBlockConsensusClient");
```
klkvr, line 520: should this be optional? e.g if you want to run `op-node` and a flashblocks engine client

klkvr (commented):
this does make sense, left some questions

defer to @mattsse for final review

mattsse (commented):
I believe this was superseded by another one

PR #19480, opened 2025-11-04
fix(net): avoid unnecessary Sender clones in ActiveSessionHandle

Fibonacci747:
Remove .clone() before try_send and send in ActiveSessionHandle::{disconnect,try_disconnect} 
Cloning tokio::mpsc::Sender does not affect channel capacity and is not required since both methods take &self
Aligns with existing pattern in session/mod.rs::send_message() and reduces minor overhead


mattsse (commented):
we should clone here here because we definitely want this delivered

PR #19481, opened 2025-11-04
fix(net): remove capacity inflation from buffered blocks size calculation

JacksonEi:
Fixes `buffered_blocks_size_bytes` in `BodiesDownloader::buffer_bodies_response` to reflect only actual block data by removing `response.capacity() * mem::size_of::<BlockResponse<B>>()`.


mattsse (approved):
that makes sense

PR #19484, opened 2025-11-04
fix: extract helper for execution witness generation

anim001k:
Factor out the core “re-execute and collect ExecutionWitness” logic into a small internal helper in `invalid-block-hooks` and document intended reuse.
The hook duplicates logic that exists in the Debug API; extracting the core makes behavior clearer and reduces future drift, preparing a clean path to unify both call sites.


mattsse (commented):
we dont need this

PR #19487, opened 2025-11-04
chore(trie): Use Vec<Option<...>> in HashedPostStateCursors

### LINKED ISSUE #18848: Use Vec<(B256, Option<...>)> in HashedPostStateCursors

mediocregopher:
Fixes https://github.com/paradigmxyz/reth/issues/18848

This continues the work done https://github.com/paradigmxyz/reth/pull/19233. I've done a bit of cleanup and added more comprehensive proptests.

This PR modifies the in-memory representation of `HashedPostStateSorted` so that it only uses a single Vec for both updates and deletions, rather than a Vec for updates and HashSet for deletions. This allows the type to more closely map to how changesets are stored in the database, and will make it simpler to serialize/deserialize this data in general (e.g. for https://github.com/paradigmxyz/reth/pull/19430).

Changing the in-memory representation of `HashedPostStateSorted` requires changing how the in-memory overlay for the database works. This was done in a manner essentially identical how InMemoryTrieCursor is implemented now, with a small complication that a trait is required to handle the two different value types of the hashed state tables (Account and U256).


crates/trie/common/src/hashed_state.rs
```diff
@@ -362,17 +344,8 @@ impl HashedPostState {
...
-        for (hashed_address, info) in self.accounts.drain() {
-            if let Some(info) = info {
-                updated_accounts.push((hashed_address, info));
-            } else {
-                destroyed_accounts.insert(hashed_address);
-            }
-        }
-        updated_accounts.sort_unstable_by_key(|(address, _)| *address);
-        let accounts = HashedAccountsSorted { accounts: updated_accounts, destroyed_accounts };
+        let mut accounts: Vec<_> = self.accounts.drain().collect();
```
yongkangc, line 347: nice simplification - instead of filtering accounts into two separate collections, then sort the valid ones we just drain and sort

crates/trie/common/src/hashed_state.rs
```diff
@@ -452,41 +425,33 @@ impl HashedStorage {
...
 }
 
 /// Sorted hashed post state optimized for iterating during state trie calculation.
 #[derive(PartialEq, Eq, Clone, Default, Debug)]
 pub struct HashedPostStateSorted {
-    /// Updated state of accounts.
-    pub accounts: HashedAccountsSorted,
-    /// Map of hashed addresses to hashed storage.
+    /// Sorted collection of account updates. `None` indicates a destroyed account.
+    pub accounts: Vec<(B256, Option<Account>)>,
```
yongkangc, line 439: would a type alias be nicer / more intuitive here?

crates/trie/db/tests/post_state.rs
yongkangc, line 223: was thinking that it might be helpful to add a simple docstring on the cases that are being tested with this to make it easier to understand and follow what this test is testing?

since we have a few cases: 

```
    // empty from the get go
    // not empty
    // all zero values, but not wiped
    // wiped storage, but post state has zero-value entries
    // wiped storage, but post state has non-zero entries
```

maybe something like

```suggestion
/// Tests `is_storage_empty()` correctly distinguishes wiped storage from storage with zero values.
///
/// Key distinction:
/// - `wiped = true`: Storage cleared/deleted → empty
/// - `wiped = false` with zeros: Explicit zero values → not empty
///
/// Test cases:
/// 1. No entries → empty
/// 2. Non-zero values → not empty
/// 3. Zero values, not wiped → not empty
/// 4. Wiped + zero post-state → empty
/// 5. Wiped + non-zero post-state → not empty
#[test]
```

just a suggestion as i am not familiar with this part of the code

crates/trie/db/tests/post_state.rs
```diff
@@ -244,6 +244,23 @@ fn storage_is_empty() {
         assert!(!cursor.is_storage_empty().unwrap());
     }
 
+    // all zero values, but not wiped
```
yongkangc: does this test catch implementations that would incorrectly treat zero values as empty storage?

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -168,170 +254,241 @@ where
...
+                result.dedup_by(|a, b| a.0 == b.0);
+                result
+            })
+        }
+
+        proptest! {
+            #![proptest_config(ProptestConfig::with_cases(1000))]
+
+            #[test]
+            fn proptest_hashed_post_state_cursor(
```
yongkangc: this is the main prop test. just my own nit, but i feel like having a docstring here would be easier to follow because of the setup code. e.g

```suggestion
        /// Tests `HashedPostStateCursor` produces identical results to a pre-merged cursor
        /// across 1000 random scenarios.
        ///
        /// For random DB entries and post-state changes, creates two cursors:
        /// - Control: pre-merged data (expected behavior)  
        /// - Test: `HashedPostStateCursor` (lazy overlay)
        ///
        /// Executes random sequences of `next()` and `seek()` operations, asserting
        /// both cursors return identical results.
        #[test]
        fn proptest_hashed_post_state_cursor(
```

yongkangc (approved):
lgtm overall, left some questions and nits

crates/trie/common/src/hashed_state.rs
mediocregopher: Imo it's ok as-is... but open to it if others thing it would be easier to read. Which part were you thinking would be better as an alias?

crates/trie/db/tests/post_state.rs
mediocregopher: This is testing that given a _single_ zero value in the overlay (which indicates the slot was deleted), `is_storage_empty` still returns false, since the db has other non-zero slots.

mattsse (approved):

PR #19488, opened 2025-11-04
feat(reth-bench-compare): upstream from personal repo

shekhirin:
Upstreamed from https://github.com/shekhirin/reth-bench-compare

A lot of this code can be simplified, it was 100% Claude written.


mediocregopher (approved):

mattsse (approved):
lgtm

this is just tooling and doesnt affect the node itself

PR #19489, opened 2025-11-04
chore: Various cleanups after consistent DB view removal

mediocregopher:
Closes #19341

This implements various minor improvements which were pointed out during review of the ConsistentDbView removal PR.

* Remove prefix_sets from proof tasks. We no longer use prefix_sets at all during proof calculations.

* Spawning state root task can no longer fail, so we can simplify a bunch of logic around spawn_payload_validator, including no longer needing to fallback to a new strategy.

* Propagate errors from account/storage proof workers.


shekhirin (approved):

mattsse (approved):
lgtm

yongkangc (approved):
Good stuff

PR #19490, opened 2025-11-04
refactor(era): move to e2s module e2s types and file handling

lean-apple:
Part 1 for some refactor to make cleaner the files architecture of https://github.com/paradigmxyz/reth/tree/main/crates/era/src.

Right now, just the e2s types and file handling is moved to its own module.


mattsse (approved):
lgtm, ty!

PR #19491, opened 2025-11-04
fix: spawn block fetching blocking

mattsse:
we previously didnt spawn this as blocking

but this needs to fetch he blocks in the range

this introduces chunking over the block range and also performs the reads on a blocking task

this could still be improved so that blocks closer to the tip are fetched from the cache, but this can be done separately

recommend to hide whitespace for review

### DIFF
--- a/crates/rpc/rpc/src/trace.rs
+++ b/crates/rpc/rpc/src/trace.rs
@@ -429,6 +429,7 @@ where
                 block_traces.push(traces);
             }
 
+            #[allow(clippy::iter_with_drain)]
             let block_traces = futures::future::try_join_all(block_traces.drain(..)).await?;
             all_traces.extend(block_traces.into_iter().flatten().flat_map(|traces| {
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/trace.rs
```diff
@@ -389,80 +389,97 @@ where
...
-                        base_block_reward,
-                    )
-                    .into_iter()
-                    .filter(|trace| matcher.matches(&trace.trace)),
+        let mut all_traces = Vec::new();
+        for chunk_start in (start..end).step_by(self.inner.eth_config.max_tracing_requests) {
+            let chunk_end =
+                std::cmp::min(chunk_start + self.inner.eth_config.max_tracing_requests as u64, end);
+
+            let mut block_traces = Vec::with_capacity((chunk_end - chunk_start) as usize);
```
shekhirin: nit: allocation can be moved outside of the loop and the vector cleared at the end of chunk processing

shekhirin (approved):

PR #19493, opened 2025-11-04
fix: use cost when checking fee cap

klkvr:
re-use `cost` returned by `EthPoolTransaction` implementation instead of computing fee again here

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -396,15 +396,12 @@ where
             match self.tx_fee_cap {
                 Some(0) | None => {} // Skip if cap is 0 or None
                 Some(tx_fee_cap_wei) => {
-                    // max possible tx fee is (gas_price * gas_limit)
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -396,15 +396,12 @@ where
             match self.tx_fee_cap {
                 Some(0) | None => {} // Skip if cap is 0 or None
                 Some(tx_fee_cap_wei) => {
-                    // max possible tx fee is (gas_price * gas_limit)
-                    // (if EIP1559) max possible tx fee is (max_fee_per_gas * gas_limit)
-                    let gas_price = transaction.max_fee_per_gas();
-                    let max_tx_fee_wei = gas_price.saturating_mul(transaction_gas_limit as u128);
+                    let max_tx_fee_wei = transaction.cost().saturating_sub(transaction.value());
```
mattsse, line 399: this makes sense

mattsse (approved):

PR #19494, opened 2025-11-04
Simplify panic handling in TaskExecutor critical spawns

Doryu001:
Rewrite the critical task wrappers to await catch_unwind inside an async block, eliminating the extra map_err/map(drop) chain. Drop the unused TryFutureExt import while keeping FutureExt for catch_unwind. Preserve the tracing instrumentation and panic reporting logic while keeping the control flow clearer.


mattsse (commented):
I like what we have now

PR #19497, opened 2025-11-04
docs(banlist): document timeout update behavior on re-ban

avorylli:
Adds documentation clarifying that banning an already banned peer or IP updates the timeout to the new value. 

Also fixes a test compilation error by replacing PeerId::random() with PeerId::new([1; 64]) since the random() method doesn't exist on FixedBytes<64>.


mattsse (approved):
fine

PR #19498, opened 2025-11-05
fix: validate negative count values in E2S index parsing

dennsikl:
Added validation for negative count values during E2S index parsing. Previously, the code read count as i64 and immediately cast it to usize without checking. If a malformed file contained a negative number (e.g., -1), it would wrap to a huge value like 18 quintillion, causing the subsequent Vec::with_capacity() to panic with OOM.

Now we validate that count is non-negative before the cast and return a clear error if the file is corrupted.


mattsse (commented):
this isnt reachable

PR #19499, opened 2025-11-05
perf(trie): replace static Vec with const slice for empty updates

reallesee:
Replaced `static Vec::new()` with `const &[]` for EMPTY_UPDATES in InMemoryTrieCursorFactory::storage_trie_cursor.

The old version was creating a runtime-initialized static variable every time, which added unnecessary overhead. Since we just need an empty slice as a fallback when there are no storage trie updates, using a const reference to an empty array is more efficient and idiomatic.


reallesee:
@shekhirin fair, but static Vec still takes 24 bytes in memory vs const &[] taking zero. Small thing but why waste memory?

shekhirin (approved):
Makes sense!

PR #19500, opened 2025-11-05
feat(optimism): Simplify trait bounds in revalidate_interop_txs_stream

Fibonacci747:
Removed redundant bounds (MaybeInteropTransaction, Clone, Sync, 'static) from TItem in revalidate_interop_txs_stream() since the function body does not use them.
Cleaned up unused import of MaybeInteropTransaction
Kept only PoolTransaction, Transaction, and Send which are actually required by access_list(), hash(), and concurrent buffering.


mattsse (approved):
seems okay

PR #19501, opened 2025-11-05
fix(stages-types): resolve compilation errors in tests

davidjsonn:
Tests in reth-stages-types weren't compiling due to missing vec! macro import and some dev dependencies.

Fixed by:
- Adding #[cfg(test)] use alloc::vec for test builds
- Adding missing modular-bitfield and reth-trie-common[reth-codec] to dev-dependencies

cargo test -p reth-stages-types --lib now works properly


mediocregopher (approved):
Confirmed that tests were previously broken and this does fix them

shekhirin (approved):

PR #19508, opened 2025-11-05
feat(provider, static-file): transaction senders segment

### LINKED ISSUE #19030: TransactionSenders static files segment
### Describe the feature

https://github.com/paradigmxyz/reth/blob/8b18a1a282788cbd9acd53bd93ee31fe56644004/crates/storage/db-api/src/tables/mod.rs#L489-L496

Move to static files, indexed same as `Transactions` segment

### Additional context

_No response_

shekhirin:
Closes #19030

This PR adds new `TransactionSenders` static file segment and implements appending to/unwinding from it in pipeline and live sync using `EitherWriter`.

This functionality is gated behind storage settings and can be activated either on a new node via `reth init --static-files.transaction-senders`, or on an existing node via `reth db settings set transaction_senders_in_static_files true`.

Mainnet: 50GB -> 36GB (post-merge transactions), 28% space savings

image


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2930,7 +2948,21 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockWrite
...
+            let static_file_transaction_sender_num = self
+                .static_file_provider
+                .get_highest_static_file_tx(StaticFileSegment::TransactionSenders);
+
+            let to_delete = static_file_transaction_sender_num
+                .map(|static_num| (static_num + 1).saturating_sub(unwind_tx_from))
+                .unwrap_or_default();
+
+            self.static_file_provider
+                .latest_writer(StaticFileSegment::Receipts)?
```
```suggestion
                .latest_writer(StaticFileSegment::TransactionSenders)?
```

crates/stages/stages/src/stages/sender_recovery.rs
```diff
@@ -107,8 +116,28 @@ where
...
+                while let Some((block, index)) = blocks_with_indices.peek() {
+                    if index.contains_tx(tx) {
+                        block_numbers.push(*block);
+                        return block_numbers
+                    }
+                    blocks_with_indices.next();
+                }
+                block_numbers
+            });
+            let fold_elapsed = start.elapsed();
```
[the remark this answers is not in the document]
shekhirin: nah it's correct

crates/node/core/src/args/static_files.rs
```diff
@@ -28,6 +32,16 @@ pub struct StaticFilesArgs {
...
+
+    /// Store transaction senders in static files instead of the database.
+    ///
+    /// When enabled, transaction senders will be written to static files on disk instead of the
+    /// database.
+    ///
+    /// Note: This setting can only be configured at genesis initialization. Once
+    /// the node has been initialized, changing this flag requires re-syncing from scratch.
+    #[arg(long = "static-files.transaction-senders")]
+    pub transaction_senders: bool,
```
joshieDo, line 44: probably should add a --static-files.all (not for this PR)

crates/stages/stages/src/stages/sender_recovery.rs
```diff
@@ -84,6 +86,14 @@ where
             input.next_block_range_with_transaction_threshold(provider, self.commit_threshold)?
         else {
             info!(target: "sync::stages::sender_recovery", "No transaction senders to recover");
+            EitherWriter::new_senders(
+                provider,
+                provider
+                    .static_file_provider()
+                    .get_highest_static_file_block(StaticFileSegment::Transactions)
```
joshieDo:
```suggestion
                    .get_highest_static_file_block(StaticFileSegment::TransactionSenders)
```

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -7,33 +7,45 @@ use serde::{Deserialize, Serialize};
...
+    pub transaction_senders_in_static_files: bool,
 }
 
-impl StorageSettings {
-    /// Creates a new `StorageSettings` with default values.
-    pub const fn new() -> Self {
-        Self { receipts_in_static_files: false }
+impl Default for StorageSettings {
+    fn default() -> Self {
+        Self::legacy()
```
joshieDo: maybe we just remove Default altogether and just use legacy instead of default/new? ive grown to hate Default. and it becomes way more explicit

shekhirin: removed

crates/storage/db-models/src/blocks.rs
```diff
@@ -64,6 +64,11 @@ impl StoredBlockBodyIndices {
     pub const fn tx_count(&self) -> NumTransactions {
         self.tx_count
     }
+
+    /// Returns true if the block contains a transaction with the given number.
+    pub const fn contains_tx(&self, tx_num: TxNumber) -> bool {
+        tx_num >= self.first_tx_num && tx_num <= self.next_tx_num()
```
joshieDo:
```suggestion
        self.tx_count > 0 && tx_num >= self.first_tx_num && tx_num <= self.next_tx_num()
```

i think?

crates/storage/db/src/static_file/masks.rs
```diff
@@ -33,12 +33,18 @@ add_static_file_mask! {
...
 add_static_file_mask! {
-    #[doc = "Mask for selecting a single transaction from Transactions static file segment"]
+    #[doc = "Mask for selecting a single transaction from `Transactions` static file segment"]
     TransactionMask<T>, T, 0b1
 }
+
+// TRANSACTION SENDER MASKS
+add_static_file_mask! {
+    #[doc = "Mask for selecting a single transaction sender from `TransactionSenders` static file segment"]
+    TransactionSenderMask, alloy_primitives::Address, 0b1
```
joshieDo: super nit
```suggestion
    TransactionSenderMask, Address, 0b1
```

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -467,11 +467,19 @@ impl<N: ProviderNodeTypes> TransactionsProvider for ProviderFactory<N> {
         &self,
         range: impl RangeBounds<TxNumber>,
     ) -> ProviderResult<Vec<Address>> {
-        self.provider()?.senders_by_tx_range(range)
+        if EitherWriter::senders_destination(self).is_static_file() {
+            self.static_file_provider.senders_by_tx_range(range)
+        } else {
+            self.provider()?.senders_by_tx_range(range)
+        }
```
joshieDo: just a note, that if we use the db tool to change the destination, we lose access to the previous senders. just noting and its fine by me, its not something users should be using anyway

crates/storage/db-models/src/blocks.rs
joshieDo: or actually
```suggestion
        tx_num >= self.first_tx_num && tx_num < self.next_tx_num()
```

crates/stages/stages/src/stages/sender_recovery.rs
```diff
@@ -107,8 +116,28 @@ where
...
 
+        let start = Instant::now();
+        let block_body_indices =
+            provider.block_body_indices_range(range_output.block_range.clone())?;
+        let block_body_indices_elapsed = start.elapsed();
+        let mut blocks_with_indices = range_output.block_range.zip(block_body_indices).peekable();
+
         for range in batch {
-            recover_range(range, provider, tx_batch_sender.clone(), &mut senders_cursor)?;
+            // Pair each transaction number with its block number
```
joshieDo, line 126: what if we really paired them?

the first time i looked at the following below
```rust
    tx_range: Range<TxNumber>,
    block_numbers: Vec<BlockNumber>,
```

i was slightly confused. Maybe we could build and pass `recover_range(Vec<(TxNumber, BlockNumber)>,....`

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -679,31 +679,25 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> DatabaseProvider<TX, N> {
...
-
-                let mut senders = Vec::with_capacity(body.transactions().len());
-                for (tx_num, tx) in tx_range.zip(body.transactions()) {
-                    match known_senders.get(&tx_num) {
-                        None => {
+                let known_senders = self.senders_by_tx_range(tx_range)?;
+
+                body.transactions()
+                    .iter()
+                    .zip(known_senders.into_iter().map(Some).chain(std::iter::repeat(None)))
```
joshieDo: unsure what happens here when we add pruning

what if there is a chance that on a `tx_range` of `2..=12 `, `known_senders` is actually returning `5..=12`... we'd be matching the txs incorrectly right?

joshieDo: in general, we might want to change our range query responses to be returning` Vec<Option<V>>`  instead, the more historical pruning we add

crates/node/core/src/args/static_files.rs
```diff
@@ -41,16 +55,24 @@ impl StaticFilesArgs {
...
+
         if self.receipts {
-            StorageSettings::new().with_receipts_in_static_files()
-        } else {
-            StorageSettings::legacy()
+            settings = settings.with_receipts_in_static_files();
         }
+        if self.transaction_senders {
+            settings = settings.with_transaction_senders_in_static_files();
+        }
```
mattsse: we could make these setters that accept bools?

crates/primitives-traits/src/block/recovered.rs
```diff
@@ -83,8 +83,8 @@ impl<B: Block> RecoveredBlock<B> {
     }
 
     /// Returns an iterator over the recovered senders.
-    pub fn senders_iter(&self) -> impl Iterator<Item = &Address> {
-        self.senders.iter()
+    pub fn senders_iter(&self) -> impl Iterator<Item = Address> {
+        self.senders.iter().copied()
```
mattsse: this is breaking, can we undo this, the caller can do copied

crates/storage/provider/src/providers/static_file/jar.rs
```diff
@@ -274,15 +275,20 @@ impl<N: NodePrimitives<SignedTx: Decompress + SignedTransaction>> TransactionsPr
         &self,
         range: impl RangeBounds<TxNumber>,
     ) -> ProviderResult<Vec<Address>> {
-        let txs = self.transactions_by_tx_range(range)?;
-        Ok(reth_primitives_traits::transaction::recover::recover_signers(&txs)?)
+        let range = to_range(range);
+        let mut cursor = self.cursor()?;
+        let mut senders = Vec::with_capacity((range.end - range.start) as usize);
```
mattsse: this is too dangerous because this might allocate a ton of stuff
https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/storage/provider/src/lib.rs#L63-L63
,we should cap this to smth sane value

crates/stages/stages/src/stages/sender_recovery.rs
shekhirin: it's a bit about the way we use the range internally — we don't really iterate over it, we chunk it and send to workers. I would prefer to do it as a follow-up, because it's a bit invasive for `recover_range` function body.

crates/storage/provider/src/either_writer.rs
mattsse, line 1: is this actually very easy to keep track of where to write stuff to

mattsse (changes requested):
it was actually quite easy to follow this

some nits

crates/storage/provider/src/providers/static_file/jar.rs
shekhirin: oops very good catch, fixed in other methods too

crates/storage/provider/src/providers/database/provider.rs
shekhirin: yep this was incorrect, got it fixed. It took introducing an `EitherReader` to abstract away that functionality.

crates/cli/commands/src/db/settings.rs
```diff
@@ -83,8 +88,10 @@ impl Command {
             println!("No storage settings found, creating new settings.");
         }
 
-        let mut settings @ StorageSettings { receipts_in_static_files: _ } =
-            settings.unwrap_or_default();
+        let mut settings @ StorageSettings {
+            receipts_in_static_files: _,
+            transaction_senders_in_static_files: _,
+        } = settings.unwrap_or_else(StorageSettings::legacy);
```
mattsse, line 94: why is this pattern needed here?

crates/storage/provider/src/providers/static_file/jar.rs
```diff
@@ -104,12 +106,10 @@ impl<N: NodePrimitives<BlockHeader: Value>> HeaderProvider for StaticFileJarProv
         &self,
         range: impl RangeBounds<BlockNumber>,
     ) -> ProviderResult<Vec<Self::Header>> {
-        let range = to_range(range);
-
         let mut cursor = self.cursor()?;
-        let mut headers = Vec::with_capacity((range.end - range.start) as usize);
+        let mut headers = Vec::with_capacity(range_size_hint(&range).unwrap_or(1024));
```
mattsse, line 110: this can still be abused in theory at least I believe:`0..=u64::MAX` for example

but should be okay because we don't use it like this

mattsse (approved):
all of this lgtm!

one suggestion, if we dont do it already, can we also log the storage setting on launch?

crates/cli/commands/src/db/settings.rs
shekhirin: to not forget a new entry in `SetCommand` when we add a new field to `StorageSettings`

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1022,6 +1044,11 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                         continue;
                     }
                 }
+                StaticFileSegment::TransactionSenders => {
+                    if !provider.cached_storage_settings().transaction_senders_in_static_files {
```
joshieDo: nit: could we use EitherWriter::senders_destination.is_database? here as well

crates/storage/provider/src/providers/static_file/jar.rs
```diff
@@ -258,31 +256,34 @@ impl<N: NodePrimitives<SignedTx: Decompress + SignedTransaction>> TransactionsPr
         &self,
         range: impl RangeBounds<TxNumber>,
     ) -> ProviderResult<Vec<Self::Transaction>> {
-        let range = to_range(range);
         let mut cursor = self.cursor()?;
-        let mut txes = Vec::with_capacity((range.end - range.start) as usize);
+        let mut txs = Vec::with_capacity(range_size_hint(&range).unwrap_or(1024));
```
joshieDo, line 260: oof

joshieDo (approved):

PR #19510, opened 2025-11-05
chore(net): fix misleading comment about uncompressed message size check

JacksonEi:
Fixes misleading comment that claims to check "compressed length" when the code actually checks  `first_message_bytes.len()`, which is the uncompressed message size.


mattsse (approved):
this is actually correct

PR #19520, opened 2025-11-05
refactor(era): move era1 types and file handling to new module

lean-apple:
Move `era1` types to their own module folder `era1` for better crate files architecture.

From https://github.com/paradigmxyz/reth/tree/main/crates/era/src to https://github.com/lean-apple/reth/tree/chore-folder-refactor-2/crates/era/src.

Move also some common utilities for file and decoding to `common` module.

Follow-up of https://github.com/paradigmxyz/reth/pull/19490.

cc @mattsse @RomanHodulak


crates/era/src/lib.rs
lean-apple, line 1: Decoding traits were moved to common module

mattsse (approved):
this is better yes

PR #19522, opened 2025-11-05
feat(engine): record newPayload/forkchoiceUpdated metrics outside of RPC

shekhirin:
Nodes building on top of Reth SDK and not using Engine API RPC directly, but via handle, don't have newPayload/forkchoiceUpdated metrics, because we record them in RPC layer.

This PR moves these metrics to `engine` crate and records them there.

### DIFF
--- a/crates/rpc/rpc-engine-api/src/metrics.rs
+++ b/crates/rpc/rpc-engine-api/src/metrics.rs
@@ -1,8 +1,4 @@
-use std::time::Duration;
-
-use crate::EngineApiError;
-use alloy_rpc_types_engine::{ForkchoiceUpdated, PayloadStatus, PayloadStatusEnum};
-use metrics::{Counter, Gauge, Histogram};
+use metrics::{Counter, Histogram};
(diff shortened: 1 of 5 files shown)


crates/rpc/rpc-engine-api/src/metrics.rs
```diff
@@ -58,58 +50,6 @@ pub(crate) struct EngineApiLatencyMetrics {
     pub(crate) get_blobs_v2: Histogram,
 }
 
-/// Metrics for engine API forkchoiceUpdated responses.
-#[derive(Metrics)]
-#[metrics(scope = "engine.rpc")]
```
mattsse, line 63: this does change the scope,

do we need to update anything else for this, like dashboards?

mattsse (approved):
this makes sense, but doesnt this break the current dashboard?

pending @Rjected

shekhirin:
@mattsse idts, couldn't find mentions of these metrics in dashboard json

Rjected (approved):
lgtm

PR #19523, opened 2025-11-05
docs(chainspec): improve fork_id and satisfy documentation

CreeptoGengar:
Add algorithm overview and edge case notes to fork_id docs.
Clarify satisfy helper purpose and usage context.


mattsse (approved):
I let this one slide because linking the eip is useful and the docs are better

PR #19529, opened 2025-11-05
fix(transaction-pool): preallocate batch buffer and retain capacity via split_off

Fibonacci747:
Optimize BatchTxProcessor buffer management to avoid repeated reallocations:

- Preallocate the reusable receive buffer `buf` to `max_batch_size` (clamped to >= 1).
- Replace `mem::take` + `reserve(1)` with `split_off(0)` so `buf` retains its capacity
  between batches, avoiding growth during `poll_recv_many`.

Motivation:
The previous pattern reset `buf` to tiny capacity every cycle, causing Vec growth and reallocations on each batch under sustained load. This is contrary to the module’s high-throughput intent. The new approach keeps a hot buffer sized for `max_batch_size`,
reducing allocations, CPU overhead, and tail latency, without changing behavior.


mattsse (commented):
we can keep this as is

PR #19530, opened 2025-11-05
perf(era-utils): avoid unnecessary PathBuf clone in export

avorylli:
Move file_path into vector instead of cloning. Logging uses a reference before the move, eliminating unnecessary heap allocations when collecting exported file paths.

### DIFF
--- a/crates/era-utils/src/export.rs
+++ b/crates/era-utils/src/export.rs
@@ -219,12 +219,12 @@ where
             writer.write_accumulator(&accumulator)?;
             writer.write_block_index(&block_index)?;
             writer.flush()?;
-            created_files.push(file_path.clone());
 
             info!(
                 target: "era::history::export",
                 "Wrote ERA1 file: {file_path:?} with {blocks_written} blocks"
             );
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense

crates/era-utils/src/export.rs
```diff
@@ -219,12 +219,12 @@ where
             writer.write_accumulator(&accumulator)?;
             writer.write_block_index(&block_index)?;
             writer.flush()?;
-            created_files.push(file_path.clone());
 
             info!(
                 target: "era::history::export",
                 "Wrote ERA1 file: {file_path:?} with {blocks_written} blocks"
             );
+            created_files.push(file_path);
```
lean-apple, line 227: The issue is, now, the trace is in advance regarding what has been executed.
We can use `created_files.last().unwrap()` in the trace to avoid cloning, but cloning here doest not seem a big perf issue.

mattsse: hmm the trace is still after the flush, so logically nothing changed here

PR #19531, opened 2025-11-05
refactor: convert proof counters to MultiProofTask fields

hawkadrian:
Convert `prefetch_proofs_requested`, `state_update_proofs_requested`, and `proofs_processed` from local variables in `run()` to fields of `MultiProofTask` struct.

- Add three new fields to `MultiProofTask` struct
- Initialize fields in `new()` constructor
- Update all usages in `run()` method to use `self.` prefix
- Remove TODO comment and local variable declarations


mattsse (commented):
this has drifted

PR #19533, opened 2025-11-06
docs: explain default db size and error in CLI help

### LINKED ISSUE #19234: Missing documentation for default database size limit (4 TB) causes “environment map size limit reached” error
### Describe the change

While running a reth node, I encountered the following fatal error:

```
{"timestamp":"2025-10-22T15:01:27.533428Z","level":"ERROR","fields":{"message":"Persistence service failed","err":"ProviderError(Database(Write(DatabaseWriteError { info: DatabaseErrorInfo { message: \"environment map size limit reached\", code: -30792 }, operation: CursorInsert, table_name: \"AccountsHistory\", key: [77, 110, 34, 10, 217, 231, 13, 72, 176, 97, 15, 118, 50, 102, 6, 24, 10, 249, 97, 142, 255, 255, 255, 255, 255, 255, 255, 255] })))"}}
--
```

After some digging, I discovered that this happens when the database reaches 4 TB, which is the default maximum size configured internally.
This limit isn’t mentioned in the docs, so the only way to learn about it was by browsing the source:

```rs
impl DatabaseArguments {
    /// Create new database arguments with given client version.
    pub fn new(client_version: ClientVersion) -> Self {
        Self {
            client_version,
            geometry: Geometry {
                size: Some(0..(4 * TERABYTE)),
                growth_step: Some(4 * GIGABYTE as isize),
                shrink_threshold: Some(0),
                page_size: Some(PageSize::Set(default_page_size())),
            },
            log_level: None,
            max_read_transaction_duration: None,
            exclusive: None,
        }
    }

    /// Sets the upper size limit of the db environment, the maximum database size in bytes.
    pub const fn with_geometry_max_size(mut self, max_size: Option<usize>) -> Self {
        if let Some(max_size) = max_size {
            self.geometry.size = Some(0..max_size);
        }
        self
    }
}

// ...other code
```

Please document the default database size (4 TB). Thanks!

### Additional context

_No response_

Andrurachi:
The --db.max-size flag documentation was missing the default value (8TB) and the specific error ("environment map size limit reached") that users encounter.

This change adds the missing context to the CLI help text to prevent confusion.

Fixes #19234

### DIFF
--- a/crates/node/core/src/args/database.rs
+++ b/crates/node/core/src/args/database.rs
@@ -25,7 +25,12 @@ pub struct DatabaseArgs {
     /// NFS volume.
     #[arg(long = "db.exclusive")]
     pub exclusive: Option<bool>,
-    /// Maximum database size (e.g., 4TB, 8MB)
+    /// Maximum database size (e.g., 4TB, 8MB).
+    ///
+    /// This sets the "map size" of the database. If the database grows beyond this
+    /// limit, the node will stop with an "environment map size limit reached" error.
+    ///
+    /// The default value is 8TB.
     #[arg(long = "db.max-size", value_parser = parse_byte_size)]
     pub max_size: Option<usize>,
     /// Database growth step (e.g., 4GB, 4KB)
--- a/crates/node/core/src/args/database.rs
+++ b/crates/node/core/src/args/database.rs
@@ -25,7 +25,7 @@ pub struct DatabaseArgs {
     /// NFS volume.
     #[arg(long = "db.exclusive")]
     pub exclusive: Option<bool>,
-    /// Maximum database size (e.g., 4TB, 8MB).
+    /// Maximum database size (e.g., 4TB, 8TB).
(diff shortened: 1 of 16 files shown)


shekhirin (approved):

crates/node/core/src/args/database.rs
```diff
@@ -25,7 +25,12 @@ pub struct DatabaseArgs {
     /// NFS volume.
     #[arg(long = "db.exclusive")]
     pub exclusive: Option<bool>,
-    /// Maximum database size (e.g., 4TB, 8MB)
+    /// Maximum database size (e.g., 4TB, 8MB).
```
shekhirin: 😅

```suggestion
    /// Maximum database size (e.g., 4TB, 8TB).
```

shekhirin:
@Andrurachi feel free to reply in the issue so I can assign you to it. Thanks!

PR #19534, opened 2025-11-06
fix: remove redundant header insertion in extend_blocks and tests

phrwlk:
The mock provider’s extend_blocks() called add_header() and then add_block(), but add_block() already inserts the header. This resulted in redundant header insertions (double locking and duplicate HashMap inserts). In tests, persist_blocks() additionally called extend_headers(), effectively causing triple insertion.

Changes:
- test_utils/mock.rs: extend_blocks() now only calls add_block().
- engine/tree/tests.rs: removed extend_headers(...) in persist_blocks().

This keeps header insertion centralized in add_block(), aligns extend_* helper semantics with others, reduces unnecessary work, and preserves expected ordering.


shekhirin (approved):
Good catch

PR #19535, opened 2025-11-06
fix(prune): use saturating_sub in PruneLimiter::deleted_entries_limit_left

Snezhkko:
Replace limit - deleted with limit.saturating_sub(deleted) in PruneLimiter::deleted_entries_limit_left() to prevent underflow when the deleted count exceeds the current limit (which can happen after lowering limits mid-run).
Add tests to ensure the function returns Some(0) when deleted >= limit, including after lowering the limit via both set_deleted_entries_limit() and floor_deleted_entries_limit_to_multiple_of().
This change hardens the limiter against panic/wraparound and matches the intended semantics of “remaining deletions.”

### DIFF
--- a/crates/prune/prune/src/limiter.rs
+++ b/crates/prune/prune/src/limiter.rs
@@ -96,7 +96,7 @@ impl PruneLimiter {
 
     /// Returns the number of deleted entries left before the limit is reached.
     pub fn deleted_entries_limit_left(&self) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


shekhirin (changes requested):
The change itself LGTM, can we please combine these tests into one that just checks a scenario with no saturation and with saturation?

shekhirin (approved):

PR #19536, opened 2025-11-06
refactor: remove redundant .to_string().as_str() in core modules

davidjsonn:
Remove redundant .as_str() calls in assertions and array initialization. Rust's Deref trait handles these conversions automatically.


mattsse (commented):
I prefer the as_str call here, because all of these need &str

PR #19537, opened 2025-11-06
fix: update docs and arg comments

Vitaliyr888:
`crates/node/core/src/args/network.rs:` these options configure peer counts rather than individual requests; the comments now describe the actual meaning of the parameters.

`docs/vocs/docs/pages/sdk/custom-node/transactions.mdx:` type 0x2A is hexadecimal 42, matching the #[envelope(ty = 42)] attribute, so the doc comment now accurately reflects the code.

`testing/ef-tests/src/models.rs:` the fork sequence jumps from Homestead to the DAO (the DAO hard fork preceded Tangerine Whistle), so the label now matches the historical transition.


mattsse (approved):
I let this one slide because this actually fixes cli docs

PR #19540, opened 2025-11-06
chore: add new engine metrics to dashboard, fix multiproof charts

shekhirin:
1. Engine `forkchoiceUpdated` and `newPayload` charts now have only percentiles in the legend, without a breakdown per method version. image
2. Multiproof charts now display pending requests and active workers, according to new metrics image


mattsse (commented):
uTACK

mattsse (approved):

PR #19541, opened 2025-11-06
ci: check that Grafana dashboard doesn't contain DS_PROMETHEUS

shekhirin:
Our main Grafana dashboard uses the `datasource` variable, so we can't use the default `${DS_PROMETHEUS}` variable in panels.

Example of a failing workflow: https://github.com/paradigmxyz/reth/actions/runs/19133586708/job/54680080831


mattsse (approved):
makes sense

PR #19544, opened 2025-11-06
perf(execution): preallocate stage buffers

arreaaas:
We now size the execution loop’s vecs up front so the hot path stops reallocating on every block. When ExEx hooks are active we still keep the blocks buffered, but only then; otherwise we skip that allocation entirely. Behaviour stays identical, just cheaper.


mattsse (commented):
this isnt needed

PR #19545, opened 2025-11-06
fix: broken URL in download.mdx

kilavvy:
**Description:**  
This PR updates the outdated download link in `download.mdx` to the correct Merkle snapshot URL.  

- Replaced `https://downloads.merkle.io` with `https://www.merkle.io/snapshots`

### DIFF
--- a/docs/vocs/docs/pages/cli/reth/download.mdx
+++ b/docs/vocs/docs/pages/cli/reth/download.mdx
@@ -81,7 +81,7 @@ Database:
           - https://publicnode.com/snapshots (full nodes & testnets)
 
           If no URL is provided, the latest mainnet archive snapshot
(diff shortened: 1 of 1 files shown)


kilavvy:
> We set these docs dynamically from the default base URL, so I think it's fine to have it as-is
> 
> https://github.com/paradigmxyz/reth/blob/b305247b5e120b857c22b37116b311285986a4f7/crates/cli/commands/src/download.rs#L84-L87

@shekhirin Thanks! Understood about the dynamic URL. Maybe we could add a note in the docs to avoid confusion for others who might think the link is broken?
Снимок экрана 2025-11-06 в 3 17 26 PM

mattsse (commented):
this is generated

PR #19549, opened 2025-11-06
fix: Update filter last poll timestamp in filter logs

ercecan:
While working on implementing geth filter rpc endpoints on Citrea I noticed the last poll timestamp is not updated at each poll like in `eth_filterChanges` implementation. 

This may cause a problem for people using `eth_filterLogs` endpoint only, because if the poll timestamp is not updated the filter will be evicted from active filters no matter how many times it is used.

This pr aims to fix that by updating last poll timestamp.


mattsse (approved):
this is absolutely right, ty!

PR #19553, opened 2025-11-06
fix: ensure transaction lookup can prune

joshieDo:
Ensures Bodies pruning waits for TransactionLookup to finish before deleting static file jars.

## Problem
TransactionLookup needs to read transaction data to compute hashes before pruning its index. If Bodies deletes the jars first, tx_lookup fails.

## Solution
- Bodies now checks tx_lookup's checkpoint before pruning
- If tx_lookup hasn't caught up, Bodies returns `WaitingOnSegment` and limits pruning to what tx_lookup has processed
- Refactored `prune_static_files` to compute checkpoint from actual deleted range (more accurate)

## Segment ordering
```
TransactionLookup (runs first) → reads txs, prunes lookup table
Bodies (runs second) → deletes static file jars for already-pruned blocks
```


crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -40,12 +70,29 @@ where
...
 
         if deleted_headers.is_empty() {
-            return Ok(SegmentOutput::done())
+            return Ok(SegmentOutput {
+                progress: PruneProgress::Finished,
+                pruned: 0,
+                checkpoint: input
+                    .previous_checkpoint
+                    .map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            })
```
joshieDo: drive-by and unsure.

on one hand returning an empty checkpoint is wrong, on the other deleted_headers is empty because input.to_block falls in the middle of a static file (eg. block 10 in a range 0..=499_999k) 

opted for finished, but could also be not_done with a interrupt reason, thoughts?

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -2,30 +2,60 @@ use crate::{
...
+
+        let tx_lookup_checkpoint = provider
+            .get_prune_checkpoint(PruneSegment::TransactionLookup)?
+            .and_then(|cp| cp.block_number);
+
+        // Check if tx_lookup will eventually need to prune any blocks <= our target
+        Ok(tx_lookup_mode
+            .next_pruned_block(tx_lookup_checkpoint)
+            .is_some_and(|next_block| next_block <= target)
+            .not())
```
shekhirin: why is it not enough to just check `tx_lookup_checkpoint >= target` here?

joshieDo: you could have a `Before(100)` for txlookup and a `Before(200)` for bodies. (lets forget sf ranges for now)

our tx_lookup_checkpoint would always be 99 and bodies would never run

crates/prune/prune/src/segments/user/bodies.rs
mattsse: why are we checking deleted_headers here?

```
let deleted_headers = provider
            .static_file_provider()
            .delete_segment_below_block(StaticFileSegment::Transactions
```

mattsse: unsure what the impact of this is

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -40,12 +70,29 @@ where
...
     fn prune(&self, provider: &Provider, input: PruneInput) -> Result<SegmentOutput, PrunerError> {
+        if !self.is_tx_lookup_done(provider, input.to_block)? {
+            debug!(
+                to_block = input.to_block,
+                "Transaction lookup still has work to be done up to target block"
+            );
+            return Ok(SegmentOutput::not_done(
+                PruneInterruptReason::WaitingOnSegment(PruneSegment::TransactionLookup),
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
```
mattsse: I think this I understand

crates/prune/prune/src/segments/user/bodies.rs
joshieDo: `deleted_headers` refers to a list of `SegmentHeader { block_range, tx_range}` of static files that we _were able to_ delete.

if the prune target is a block in the middle of a static file, we would return empty here

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -40,23 +95,48 @@ where
...
+            })
         }
 
         let tx_ranges = deleted_headers.iter().filter_map(|header| header.tx_range());
 
         let pruned = tx_ranges.clone().map(|range| range.len()).sum::<u64>() as usize;
 
+        // The highest block number in the deleted files is the actual checkpoint. to_block might
+        // refer to a block in the middle of a undeleted file.
+        let checkpoint_block = deleted_headers
```
joshieDo: `if checkpoint_block != input.to_block` we can't return `PruneProgress::not_done`, otherwise the `PrunerStage` would get stuck trying over and over again

crates/prune/types/src/mode.rs
```diff
@@ -80,6 +80,38 @@ impl PruneMode {
...
+    /// This is independent of the current tip and indicates what block is next in the pruning
+    /// sequence according to this mode's configuration. Returns `None` if no more blocks will
+    /// be pruned (i.e., the mode has reached its target).
+    ///
+    /// # Examples
+    ///
+    /// - `Before(10)` with checkpoint at block 5 returns `Some(6)`
+    /// - `Before(10)` with checkpoint at block 9 returns `None` (done)
+    /// - `Distance(100)` with checkpoint at block 1000 returns `Some(1001)` (always has more)
+    /// - `Full` always returns the next block after checkpoint
```
mattsse, line 101: these are helpful

crates/prune/types/src/mode.rs
```diff
@@ -80,6 +80,38 @@ impl PruneMode {
...
+    pub const fn next_pruned_block(&self, checkpoint: Option<BlockNumber>) -> Option<BlockNumber> {
+        let next = match checkpoint {
+            Some(c) => c + 1,
+            None => 0,
```
mattsse, line 105: should we start at genesis here or 1?

crates/prune/prune/src/segments/set.rs
```diff
@@ -72,8 +72,10 @@ where
         } = prune_modes;
 
         Self::default()
-            // Bodies - run first since file deletion is fast
-            .segment_opt(bodies_history.map(Bodies::new))
+            // Transaction lookup (should be run before bodies)
+            .segment_opt(transaction_lookup.map(TransactionLookup::new))
+            // Bodies
+            .segment_opt(bodies_history.map(|mode| Bodies::new(mode, transaction_lookup)))
```
mattsse: can we add more note for dummies: because we still need bodies for txlookup

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -2,30 +2,85 @@ use crate::{
...
 
 /// Segment responsible for pruning transactions in static files.
 ///
 /// This segment is controlled by the `bodies_history` configuration.
 #[derive(Debug)]
 pub struct Bodies {
     mode: PruneMode,
+    /// Transaction lookup prune mode. Used to determine if we need to wait for tx lookup pruning
+    /// before deleting transaction bodies.
+    tx_lookup_mode: Option<PruneMode>,
```
mattsse, line 22: okay, now this makes sense to me

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -2,30 +2,85 @@ use crate::{
...
+        input: &PruneInput,
+    ) -> Result<Option<BlockNumber>, PrunerError>
+    where
+        Provider: PruneCheckpointReader,
+    {
+        let Some(tx_lookup_mode) = self.tx_lookup_mode else { return Ok(Some(input.to_block)) };
+
+        let tx_lookup_checkpoint = provider
+            .get_prune_checkpoint(PruneSegment::TransactionLookup)?
+            .and_then(|cp| cp.block_number);
```
mattsse, line 48: this should also have the uncommitted txlookip checkpoint, right?

even if not, this wouldnt be an issue because this would then just trail 1 prune run behind

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -2,30 +2,85 @@ use crate::{
...
+
+        // Determine the safe prune target, if any.
+        let to_block = match tx_lookup_mode.next_pruned_block(tx_lookup_checkpoint) {
+            None => Some(input.to_block), /* tx_lookup is done */
+            Some(tx_lookup_next) if tx_lookup_next > input.to_block => Some(input.to_block), /* tx_lookup is ahead */
```
mattsse: can we document what this means exactly, 

basically why can't we also prune up to `tx_lookup_next`?

crates/prune/prune/src/segments/user/bodies.rs
```diff
@@ -40,23 +95,48 @@ where
...
+                    .map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            })
         }
 
         let tx_ranges = deleted_headers.iter().filter_map(|header| header.tx_range());
 
         let pruned = tx_ranges.clone().map(|range| range.len()).sum::<u64>() as usize;
 
+        // The highest block number in the deleted files is the actual checkpoint. to_block might
+        // refer to a block in the middle of a undeleted file.
```
mattsse:
```suggestion
        // refer to a block in the middle of an undeleted file.
```

mattsse (commented):
one more question

yongkangc (approved):
lgtm

Rjected (approved):
makes sense

PR #19554, opened 2025-11-06
refactor(provider): introduce `EitherWriter`

shekhirin:
With https://github.com/paradigmxyz/reth/issues/18682, we're doing a lot of conditional provider logic for new static files segments: old nodes should read and write from DB, new nodes from static files.

For that reason, we introduce such constructions https://github.com/paradigmxyz/reth/blob/a6c0ac8cf4ade4eafd38ea2d09457bb02ba0ed68/crates/storage/provider/src/providers/database/provider.rs#L1614-L1628

This PR simplifies it by introducing an `EitherWriter` that gets initialized with either a database cursor, or a static file writer.

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -1,0 +1,43 @@
+//! Generic writer abstraction for writing to either database tables or static files.
+
+use crate::providers::StaticFileProviderRWRefMut;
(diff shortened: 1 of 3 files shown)


joshieDo (approved):
me likes

mattsse (approved):
this makes sense

this works because this is mutually exclusive

PR #19556, opened 2025-11-06
chore(provider): exhaustive segment match in static files check

shekhirin:
Makes the consistency check less error-prone when we add new static file segments, because the match will start failing and we'll need to explicitly handle new segments.


mattsse (approved):
this is equivalent

lgtm

PR #19558, opened 2025-11-06
fix(net): preserve ECIESError in connect_without_timeout

phrwlk:
Previously, ECIESStream::connect_without_timeout() remapped any error from ECIESCodec::new_client() to io::Error::other("invalid handshake"), which collapsed structured ECIESError variants (e.g., Secp256k1) into a generic IO error.

This broke downstream error classification in the networking layer, which matches on ECIESErrorImpl variants to decide on backoff and whether a failure is fatal.

Changes:
- Propagate the original ECIESError from ECIESCodec::new_client() to preserve semantics.
- Add a regression test that passes an invalid remote_id and asserts the error’s inner
  variant is ECIESErrorImpl::Secp256k1(_), not IO.

This aligns client error handling with server-side incoming() and ensures consistent, actionable error semantics for callers.

### DIFF
--- a/crates/net/ecies/src/stream.rs
+++ b/crates/net/ecies/src/stream.rs
@@ -67,8 +67,7 @@ where
         secret_key: SecretKey,
         remote_id: PeerId,
     ) -> Result<Self, ECIESError> {
-        let ecies = ECIESCodec::new_client(secret_key, remote_id)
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is better

PR #19560, opened 2025-11-06
docs(e2s): document empty file behavior in read_version

forkfury:
Added documentation clarifying that read_version() returns Ok(None) when the file is empty or contains no entries


mattsse (commented):
I think this is obvious enough

PR #19561, opened 2025-11-06
test(static-file): StaticFileSegment string and serde roundtrips

shekhirin:
1. Adds tests for `StaticFileSegment` string and serde roundtrips that are used in filenames and segment headers
2. Removes unneeded `strum(serialize = "...")` attributes

### DIFF
--- a/crates/static-file/types/src/segment.rs
+++ b/crates/static-file/types/src/segment.rs
@@ -40,7 +40,7 @@ pub enum StaticFileSegment {
 impl StaticFileSegment {
     /// Returns the segment as a string.
     pub const fn as_str(&self) -> &'static str {
-        // `strum` doesn't generate a comment for `into_str` when using `IntoStaticStr` derive
+        // `strum` doesn't generate a doc comment for `into_str` when using `IntoStaticStr` derive
         // macro, so we need to manually implement it.
         match self {
             Self::Headers => "headers",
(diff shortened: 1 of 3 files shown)


crates/static-file/types/src/segment.rs
```diff
@@ -21,28 +21,27 @@ use strum::{AsRefStr, EnumIs, EnumString};
...
-    #[strum(serialize = "receipts")]
     /// Static File segment responsible for the `Receipts` table.
     Receipts,
 }
 
 impl StaticFileSegment {
     /// Returns the segment as a string.
     pub const fn as_str(&self) -> &'static str {
+        // `strum` doesn't generate a comment for `into_str` when using `IntoStaticStr` derive
+        // macro, so we need to manually implement it.
```
mattsse: what does generate a comment mean?

mattsse (approved):
lgtm, 1 q re comment

crates/static-file/types/src/segment.rs
shekhirin: a doc comment, so the clippy is reporting a warning

afaik there's no way to silence missing_docs just for one method, and i don't want to silence it for the whole file

mattsse: ahhh

crates/static-file/types/src/segment.rs
```diff
@@ -21,28 +21,27 @@ use strum::{AsRefStr, EnumIs, EnumString};
...
     Transactions,
-    #[strum(serialize = "receipts")]
     /// Static File segment responsible for the `Receipts` table.
     Receipts,
 }
 
 impl StaticFileSegment {
     /// Returns the segment as a string.
     pub const fn as_str(&self) -> &'static str {
+        // `strum` doesn't generate a comment for `into_str` when using `IntoStaticStr` derive
```
shekhirin:
```suggestion
        // `strum` doesn't generate a doc comment for `into_str` when using `IntoStaticStr` derive
```

PR #19564, opened 2025-11-06
refactor: provide default implementation for send_raw_transaction

klkvr:
Changes `EthTransactions` trait:
- `send_transaction` is renamed to `send_transaction_request`
- new `send_transaction` is introduced which accepts `WithEncoded<Recovered<PooledTx>>` and is expected to have the same behavior as `send_raw_transaction` currently
- default implementation for `send_raw_transaction` is provided

the idea is to allow conditionally routing transactions without requiring to do decoding twice


crates/rpc/rpc/src/eth/helpers/transaction.rs
```diff
@@ -39,9 +39,11 @@ where
     /// Decodes and recovers the transaction and submits it to the pool.
     ///
     /// Returns the hash of the transaction.
```
mattsse: I think we can remove all the docs here because trait fn

mattsse (approved):
lgtm!

Issue #19256, opened 2025-10-23
Chunking Dynamically instead of using fixed limits


yongkangc:
### Describe the feature

Currently we are hardcoding the chunk size here
https://github.com/paradigmxyz/reth/blob/60e3eded5e0a19dd03b9efb16e08ca58ccb47eb9/crates/engine/primitives/src/config.rs#L32-L33

And we only want to chunk when there are available workers. This has a huge impact for PST, as seen from https://github.com/paradigmxyz/reth/pull/18727/files because it's just harmful to chunk when you can't process the multiproof immediately, because it will go into the queue and the latency will be higher.

Chunking rules: 
https://github.com/paradigmxyz/reth/blob/1dc50aaf17fd5cc0dfe55629186972dc7c72b0eb/crates/engine/primitives/src/config.rs#L32-L33
https://github.com/paradigmxyz/reth/blob/1dc50aaf17fd5cc0dfe55629186972dc7c72b0eb/crates/trie/common/src/proofs.rs#L94-L114

**The key question is how can we chunk smarter?** 

some ideas could be possible such as: 

1. Use the number of idle workers to calculate a dynamic chunk size instead of hardcoding one.
2. Having an adaptive chunking that allows us to chunk based on size of blocks, available workers
3. adapting chunking size based on queue depth
4. size based chunking (chunking only if > size)

And possibly moving decision to chunk upstream to Proof task where the Proof task should do the chunking to distribute the work among workers (or something similar)

0xKarl98:
Can i try this ?

PR #19568, opened 2025-11-07
feat: dynamic chunking instead of hardcoding 

0xKarl98:
This is on #19256 @yongkangc  @mattsse  Please check if this behaviour is correct 

Short summary : 

Dynamic multiproof chunking now adapts to actual worker capacity instead of blindly slicing every target set. 
I added a WorkerSnapshot + select_dynamic_chunk_size helper in tree::payload_processor::multiproof that reads idle/pending counts from ProofWorkerHandle, and both on_prefetch_proof and on_state_update use it to limit fan-out to what can be consumed immediately. 

When no spare workers exist (or work size is below the baseline), we dispatch the batch as a single proof to avoid inflating queues—addressing the PST latency regression called out in the issue.

### DIFF
--- a/crates/engine/tree/benches/multiproof_chunking.rs
+++ b/crates/engine/tree/benches/multiproof_chunking.rs
@@ -30,6 +30,7 @@ fn count_chunks(targets: &MultiProofTargets, chunk_size: usize) -> usize {
     targets.clone().chunks(chunk_size).count()
 }
 
+/// Benchmarks how many chunks the dynamic policy emits for different idle worker counts.
 fn bench_chunking(c: &mut Criterion) {
     let targets = make_targets(512, 4);
     let work_units = targets.chunking_length();
--- a/crates/engine/tree/benches/multiproof_chunking.rs
+++ b/crates/engine/tree/benches/multiproof_chunking.rs
@@ -20,7 +20,7 @@ fn simulate_processing(work_units: usize, chunk_size: usize, workers: usize) -> 
     while remaining > 0 {
         let current = min(remaining, chunk_size) as f64;
         remaining -= current as usize;
-        let duration = PER_CHUNK_OVERHEAD_NS + current * PER_TARGET_COST_NS;
+        let duration = current.mul_add(PER_TARGET_COST_NS, PER_CHUNK_OVERHEAD_NS);
         for _ in 0..PER_CHUNK_SPIN {
             spin = spin.wrapping_add(1);
         }
--- a/crates/engine/tree/benches/multiproof_chunking.rs
+++ b/crates/engine/tree/benches/multiproof_chunking.rs
@@ -1,0 +1,86 @@
+#![allow(missing_docs)]
+
+use alloy_primitives::{map::B256Set, B256};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use reth_engine_tree::tree::{select_dynamic_chunk_size, WorkerSnapshot};
+use reth_trie::MultiProofTargets;
+
+const DEFAULT_BASE_CHUNK: usize = 10;
+
+fn make_targets(accounts: usize, slots: usize) -> MultiProofTargets {
+    let mut targets = MultiProofTargets::default();
+    for account in 0..accounts {
+        let hashed = make_hash(account as u64);
+        let mut storage = B256Set::default();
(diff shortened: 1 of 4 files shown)


crates/engine/tree/benches/multiproof_chunking.rs
```diff
@@ -0,0 +1,87 @@
...
+    bytes[24..].copy_from_slice(&value.to_be_bytes());
+    B256::from(bytes)
+}
+
+fn count_chunks(targets: &MultiProofTargets, chunk_size: usize) -> usize {
+    targets.clone().chunks(chunk_size).count()
+}
+
+/// Benchmarks how many chunks the dynamic policy emits for different idle worker counts.
+fn bench_chunking(c: &mut Criterion) {
```
mediocregopher, line 41: I'm curious what these benchmarks are intended to do/show? How should we use the results from them?

mediocregopher (commented):
Overall code looks plausible to me, I'm going to run a benchmark to see if this has any noticeable effect on performance.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -530,6 +530,71 @@ impl MultiproofManager {
...
+    if work_units <= base_chunk || work_units <= 1 {
+        return None;
+    }
+
+    let idle_capacity = workers.idle_capacity();
+    if idle_capacity <= 1 {
+        return None;
+    }
+
+    let dynamic_size = base_chunk.saturating_mul(idle_capacity).min(work_units.saturating_sub(1));
```
mediocregopher, line 589: I would expect this to be something like `work_units / idle_capacity`, with maybe `base_chunk` not being required at all... can you explain the reasoning behind this algo which is being used instead?

crates/engine/tree/benches/multiproof_chunking.rs
0xKarl98: We fix a synthetic MultiProofTargets set (512 accounts × 4 slots) and time how long it takes to run targets.chunks(size) under three scenarios:

- Static – always split with the old constant (chunk_size = 10).

- Dynamic (idle=8) – feed select_dynamic_chunk_size a snapshot with 8 idle workers; the helper returns a much larger chunk size, so only ~26 chunks are produced.

- Dynamic (idle=2) – same workload but only 2 idle workers; chunk size is smaller, so ~103 chunks are emitted.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
0xKarl98: > instead

 I went with the base_chunk * idle_capacity scaling because the existing config already gives us a safe minimum chunk size that operators tweak today. 
 
 Multiplying by idle_capacity keeps that knob relevant and avoids the small‑chunk path that work_units / idle_capacity would produce when the workload is tiny but the worker pool is huge (e.g., 512 workers reading a 20-target set). The min(work_units - 1) clamp then ensures we still split at least once when there’s meaningful parallelism.

crates/engine/tree/benches/multiproof_chunking.rs
0xKarl98: If you run this command for benchmark : 
cargo bench --package reth-engine-tree --bench multiproof_chunking --features test-utils

this is the result from my end :
image

crates/engine/tree/src/tree/payload_processor/multiproof.rs
0xKarl98: Take the case where the amount of wokers are way larger than proof target , e.g. 512 workers doing 20 target set , and we only consider account proof here (that's why available_storage is 0 here)  , then we have : 

base_chunk = 10
available_accounts = 512, available_storage = 0, no pending work ⇒ idle_capacity = max(512, 0) - 0 = 512
work_units = proof_targets.chunking_length() = 20

The computes:
dynamic_size = min(base_chunk * idle_capacity, work_units - 1)
             = min(10 * 512, 20 - 1)
             = min(5120, 19)
             = 19
             
So we ask the iterator to split into chunks of at most 19 items. With 20 total work units that produces two dispatches (19 + 1), i.e. we “split once” instead of producing 20 single-item chunks or clamping to zero , which avoids over-segmentation .

PR #19570, opened 2025-11-07
chore: drop unused PoolTransaction import in txpool tracing submit

reallesee:
PoolTransaction was only there when the example manually called try_from_consensus. Once we added add_consensus_transaction(_and_subscribe) in #17431/#17442 the conversion moved inside the pool API, so the import has been dead weight ever since.


mattsse (commented):
not worth doing

PR #19571, opened 2025-11-07
chore: scope otlp deps to feature flag

davidjsonn:
The Ethereum CLI was always pulling in `reth-tracing-otlp` and `url`, even when the OTLP feature was disabled. That not only triggered unused-dependency warnings but also forced extra crates into every build. This patch gates the imports and moves both dependencies under the `otlp` feature, matching how the rest of the codebase handles OTLP support. Builds without OTLP are now clean, and enabling the feature keeps the same behaviour as before.


mattsse (commented):
I think we can do without

PR #19575, opened 2025-11-07
fix(transaction-pool): blob_tx_priority argument order to use (cap, current)

Snezhkko:
This change corrects the argument order passed to blob_tx_priority in BlobTransaction::new and BlobTransaction::update_priority to follow the documented fee_delta(max_tx_fee, current_fee) semantics. Previously, the arguments were inverted, effectively comparing current network fees against transaction caps, which could zero out priorities and break eviction ordering. The fix ensures deltas become negative when caps are below current fees, aligning with BlobOrd’s eviction logic and the filtering semantics used elsewhere in this module. Added unit tests assert exact priority values to prevent regressions in argument orientation.

### DIFF
--- a/crates/transaction-pool/src/pool/blob.rs
+++ b/crates/transaction-pool/src/pool/blob.rs
@@ -276,10 +276,10 @@ impl<T: PoolTransaction> BlobTransaction<T> {
         pending_fees: &PendingFees,
     ) -> Self {
         let priority = blob_tx_priority(
-            transaction.max_fee_per_blob_gas().unwrap_or_default(),
             pending_fees.blob_fee,
-            transaction.max_fee_per_gas(),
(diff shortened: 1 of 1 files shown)


mattsse (approved):
cool this is indeed the correct order

https://github.com/paradigmxyz/reth/blob/4394860df4024d6a81a7f06704139f244ebb0135/crates/transaction-pool/src/pool/blob.rs#L381-L387

mattsse (changes requested):
tests are failing because this now calcs the inverse

https://github.com/paradigmxyz/reth/blob/main/crates/transaction-pool/src/pool/blob.rs#L382-L389

so we need to swap the order here as well, 
the existing code is correct only the variable naming is incorrect

PR #19576, opened 2025-11-07
refactor: deduplicate file-saving logic in witness hook

kilavvy:
### Description
- Extracted a small helper `save_comparison_files` to save paired “original” and “re_executed” JSON artifacts.
- Replaced duplicated code in `validate_bundle_state` and `validate_state_root_and_trie` with the new helper.


mattsse (commented):
this doesnt really deduplicate and has more lines

PR #19577, opened 2025-11-07
feat(engine): metric for time diff between newPayload end and next forkchoiceUpdated start

shekhirin:
Follow-up to https://github.com/paradigmxyz/reth/pull/19522


yongkangc (approved):
Ah nice, moving moves these metrics to engine crate and recording them there instead

joshieDo (approved):

mattsse (approved):
makes sense

PR #19579, opened 2025-11-07
fix(provider): prevent underflow in transaction range query at block boundary

strmfos:
Fixes integer underflow in `get_in_memory_or_storage_by_tx_range` when querying transactions at block boundaries.

The code calculates `skip + (remaining.min(block_tx_count - skip) - 1)` to create a range. When `skip == block_tx_count`, this becomes `0 - 1` and underflows. This happens when a query starts exactly at a block boundary and that block has 0 transactions (rare but possible based on test code throughout the repo).

The existing check at line 482 only skips blocks where the range starts after the block ends, not at the boundary itself.
Fixed by using `saturating_sub` and checking `count > 0` before creating the range.


mattsse (commented):
this is fine

PR #19582, opened 2025-11-07
fix(storage): propagate IO errors in parse_accounts to avoid silent partial imports

sashass1315:
Previously parse_accounts used a while-let Ok loop around read_line which swallowed I/O errors by exiting the loop and returning Ok(collector). This could cause a silent partial import of the state dump and lead to confusing follow-up failures like state root mismatches. This change replaces the loop with one that uses the ? operator on read_line, ensuring any I/O error is propagated to init_from_state_dump and ultimately to the CLI. This aligns error handling with parse_state_root and makes failure modes explicit, without introducing extra allocations or changing the successful path behavior.


mattsse (approved):
I let this one slide, because this should actually error in this case

PR #19588, opened 2025-11-07
chore(trie): Allow reusing Hashed/TrieCursors

mediocregopher:
Towards #19512

In order to implement a reusable proof calculator we first need the underlying cursors used by the calculator to be reusable.

For account cursors this isn't so difficult, a new `reset` method is introduced which resets the ForwardInMemoryCursor and other state fields.

For storage cursors the situation is complicated because reusing the cursor might involve using it for a completely different hashed address.

To handle this the storage cursors are given a `set_hashed_address` method which effectively resets them, as well as pulls out the correct overlay for the chosen address. Implementing this requires that the cursors now hold onto the full
`&TrieUpdatesSorted`/`&HashedPostStateSorted`.

It also requires slightly different handling of the wiped case; before we were simply storing an underlying cursor in an Option, with None indicating wiped, but now we need to always have an underlying cursors (so we can reuse it for a future non-wiped storage). A new boolean is introduced instead.


crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -281,14 +320,24 @@ where
...
+
+    fn reset(&mut self) {
+        // Reset the cursors
+        self.cursor.reset();
+        self.post_state_cursor.reset();
+
+        // Reset cursor state
+        self.cursor_entry = None;
+        self.last_key = None;
+        self.seeked = false;
```
mattsse: we could use `let Self {all fields} = self;` here and then set the &mut fields.

I prefer this because then everytime we add a field we're reminded to properly update the reset fn as well (dont ask how I know -.-)

mattsse (commented):
the reset functionality makes a lot of sense to me

dont really feel confident with reviewing the trie logic changes, nothing raised any flags, but would appreciate another review from @Rjected

crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -113,31 +104,79 @@ where
...
+    fn get_storage_overlay(
+        post_state: &'a HashedPostStateSorted,
+        hashed_address: B256,
+    ) -> (ForwardInMemoryCursor<'a, B256, U256>, bool) {
+        const EMPTY_UPDATES: &[(B256, U256)] = &[];
+
+        let post_state_storage = post_state.storages.get(&hashed_address);
+        let cursor_wiped = post_state_storage.is_some_and(|u| u.is_wiped());
+        let storage_slots =
+            post_state_storage.map(|u| u.storage_slots_ref()).unwrap_or(EMPTY_UPDATES);
```
Rjected: can we do this
```suggestion
        let post_state_storage = post_state.storages.get(&hashed_address);
        let cursor_wiped = post_state_storage.is_some_and(|u| u.is_wiped());
        let storage_slots =
            post_state_storage.map(|u| u.storage_slots_ref()).unwrap_or(&[]);
```

Rjected (approved):
lgtm, just have a nit / question about the empty value const, otherwise I think this should work, when we integrate this with more things it would be cool to see some tests that use `reset` / `set_hashed_address`

crates/trie/trie/src/hashed_cursor/post_state.rs
mediocregopher: Cool! yes this works, updated the InMemoryTrieCursor as well

PR #19591, opened 2025-11-07
fix: correct execution stage clean threshold

dennsikl:
- switch execution_external_clean_threshold to use the Merkle rebuild threshold
- this matches the doc comment ("switching between incremental and full calculations") and what ExecutionStage::can_prune_changesets actually needs

With the old code we compared against merkle.incremental_threshold, which only limits how many blocks we can process incrementally. When that value is smaller than the real rebuild cutoff, ExecutionStage thinks Merkle will rebuild from scratch and allows pruning of change sets too early. That can leave the database in an inconsistent state if Merkle keeps running incrementally. Using rebuild_threshold fixes the logic and keeps pruning decisions in sync with the hashing pipeline.


joshieDo (approved):
makes sense, i think this hasn't really been an issue because of the `.max( `statements after & hashing stages having a default clean threshold of 500k

would still like @Rjected confirmation

mattsse (commented):
This uses the wrong threshold - should use incremental_threshold (7k), not rebuild_threshold (100k).

PR #19594, opened 2025-11-07
feat(db): introduce --db.page-size argument

### LINKED ISSUE #19546: Introduce `--db.page-size` argument
## Problem

To be able to use `--db.max-size` larger than 8TB, you need to bump the page size to 8KB. This limitation is coming from the fact that MDBX has a max page number of `2^31` https://github.com/paradigmxyz/reth/blob/65acaf330e23fe51f44ce3fdbbae3f216a291bed/crates/storage/libmdbx-rs/mdbx-sys/libmdbx/mdbx.c#L2241. With 4KB pages, it means a maximum database size of `2^31 * 4096 = 8TB`.

## Solution

To work around this limitation, you need to increase the page size to the next power of 2, which is 8KB. Currently, we don't have a way to set the page size via CLI arguments, we need to add it as `--db.page-size` argument.

Additionally, the documentation for the CLI argument should explain that it's not possible to change the page size without re-creating the database, effectively meaning a resync of the node.

### Additional context

_No response_

Andrurachi:
MDBX has a 31-bit limit for page numbers (2^31), which with the default 4KB page size, imposes a maximum database size of 8TB.

This commit introduces the --db.page-size CLI argument. This allows users to set a larger page size (e.g., 8KB) on database creation, enabling a larger maximum database size (e.g., 16TB with 8KB pages).

The CLI help text includes a warning that this value can only be set once when the database is first created and that changing it requires a full node resync.

Also adds parser tests for the new argument.

Fixes #19546

### DIFF
--- a/crates/node/core/src/args/database.rs
+++ b/crates/node/core/src/args/database.rs
@@ -33,6 +33,18 @@ pub struct DatabaseArgs {
     /// The default value is 8TB.
     #[arg(long = "db.max-size", value_parser = parse_byte_size)]
     pub max_size: Option<usize>,
+    /// Database page size (e.g., 4KB, 8KB, 16KB).
+    ///
+    /// Specifies the page size used by the MDBX database.
+    ///
+    /// The page size determines the maximum database size.
+    /// MDBX supports up to 2^31 pages, so with the default 4KB page size, the maximum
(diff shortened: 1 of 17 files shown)


mattsse (approved):
this lgtm

pending @shekhirin

shekhirin (approved):
LGTM, good docs!

PR #19595, opened 2025-11-07
fix: extract common file writing logic

kilavvy:
Eliminates code duplication by extracting common file writing logic from `save_file` and `save_diff` methods into a shared `write_file` helper method.

### DIFF
--- a/crates/engine/invalid-block-hooks/src/witness.rs
+++ b/crates/engine/invalid-block-hooks/src/witness.rs
@@ -4,6 +4,7 @@ use alloy_rpc_types_debug::ExecutionWitness;
 use pretty_assertions::Comparison;
 use reth_engine_primitives::InvalidBlockHook;
 use reth_evm::{execute::Executor, ConfigureEvm};
+use reth_fs_util::{create_file, FsPathError};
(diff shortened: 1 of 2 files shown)


crates/engine/invalid-block-hooks/src/witness.rs
```diff
@@ -364,26 +364,26 @@ where
         Ok(())
     }
 
-    /// Serializes and saves a value to a JSON file in the output directory
-    fn save_file<T: Serialize>(&self, filename: String, value: &T) -> eyre::Result<PathBuf> {
+    fn write_file(&self, filename: String, contents: &[u8]) -> eyre::Result<PathBuf> {
         let path = self.output_directory.join(filename);
-        File::create(&path)?.write_all(serde_json::to_string(value)?.as_bytes())?;
-
+        File::create(&path)?.write_all(contents)?;
```
mattsse: can we map the error here so that the error includes the filepath, like

https://github.com/paradigmxyz/reth/blob/32a80e8c495cb75571a00a1d62eb2f67c0fe7d87/crates/fs-util/src/lib.rs#L50-L57

mattsse (changes requested):
I'll allow this if we can map the io error on file::create

PR #19598, opened 2025-11-08
perf: disable storage branch masks for non-target accounts

Bashmunta:
Disable collecting storage branch-node masks in Proof::multiproof() when the current account leaf is not a proof target. We only need the storage root for account RLP encoding in this case, so collecting masks is unnecessary overhead.

This reduces minor allocations and bookkeeping without changing correctness or public API. Storage masks remain collected for target accounts to support blinded trie node retrieval and proof introspection.


mediocregopher (approved):
This looks fine to me, especially since `Proof::multiproof` isn't used in any critical paths at the moment anyway.

shekhirin (approved):

PR #19600, opened 2025-11-08
refactor(provider): introduce `EitherWriter::new_receipts`

shekhirin:
Hides the logic of choosing between database vs static files into `EitherWriter`. Will be especially useful for other segments where we need to do this conditional writer creation in many places, and it can all be abstracted away into one `EitherWriter::new_` function call.


joshieDo (approved):

mattsse (approved):
lgtm!

PR #19603, opened 2025-11-08
feat(engine): batch multiproof messages

### LINKED ISSUE #19168: perf: Drain and aggregate PrefetchProofs / StateUpdate messages
### Describe the feature
we should do similar thing here as we do in sparse trie updates

[[https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs#L[…]6](https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs#L%5B%E2%80%A6%5D6)](https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs#L81-L86)

basically drain the channel, aggregate everything, and then batch

### Additional context

In `MultiProofTask::run`, mirror the sparse trie pattern
[reth/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs](https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs#L80-L105)

by draining additional PrefetchProofs / StateUpdate messages with try_recv and merging their targets before handing them to the manager. Current handling without draining:

[reth/crates/engine/tree/src/tree/payload_processor/multiproof.rs](https://github.com/paradigmxyz/reth/blob/e9598ba5ac4e32600e48b93d197a25603b1c644b/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L1014-L1073)

0xSooki:
fixes #19168 

**Batching additions:**

* Added batching for `PrefetchProofs` messages
* Added batching for `StateUpdate` messages

**Testing enhancements:**

* Added `test_prefetch_proofs_batching` to verify that multiple `PrefetchProofs` messages are correctly batched and processed as a single unit. 
* Added `test_state_update_batching` to ensure that multiple `StateUpdate` messages are merged and handled together as intended.


0xSooki:
@mediocregopher done

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1109,15 +1109,33 @@ impl MultiProofTask {
...
+                                let mut num_batched = 1;
+
+                                loop {
+                                    match self.rx.try_recv() {
+                                        Ok(MultiProofMessage::PrefetchProofs(next_targets)) => {
+                                            merged_targets.extend(next_targets);
+                                            num_batched += 1;
+                                        }
+                                        Ok(other_msg) => {
+                                            let _ = self.tx.send(other_msg);
```
mediocregopher, line 1122: If this message is a `MultiProofMessage::StateUpdate`, then I believe this will result in state updates being consumed in the incorrect order. Currently they _must_ be consumed in the order they were produced, as that ultimately determines their sequence number and therefore the order they get sent to the state root task.

If there is a way to push to the front of a crossbeam channel then some kind of buffer local to function will need to be used.

mediocregopher (changes requested):
cc @yongkangc

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: agreed with @mediocregopher here. I think this could lead to a bug 

eg
```
Channel:  [Prefetch] [StateUpdate #1] [StateUpdate #2]
                    │
                    ▼
                recv() ───▶ Process Prefetch
                    │
                    ▼
                try_recv() ───▶ Gets StateUpdate #1
                    │
                    ▼
                "Wrong type, send back"
                    │
                    ▼
                send() ───▶ Goes to BACK of queue
                    │
                    ▼
      Channel:  [StateUpdate #2] [StateUpdate #1]
                                        ▲
                                        │
                                NOW OUT OF ORDER!
```

We would have to use a buffer so that we hold the message in your hand and process it immediately after current work, so nothing jumps ahead. something like 
`let mut pending: Option<MultiProofMessage> = None;`

or Vecdeque i.e `let mut pending: VecDeque<MultiProofMessage> = VecDeque::new();`

yongkangc:
@0xSooki was wondering if u have made any progress re https://github.com/paradigmxyz/reth/pull/19603#pullrequestreview-3499396774 or maybe you need additional guidance?

yongkangc (changes requested):
apologies for dropping this @0xSooki I was out due to commitments. 

I think we would still need to find a way to address this 
https://github.com/paradigmxyz/reth/pull/19603/files#r2579912182

PR #19604, opened 2025-11-08
Remove redundant range cloning in storage provider and storage history stage

Doryu001:
reuse the precomputed block range bounds in DatabaseProvider::remove_state_above and take_state_above instead of cloning the inclusive range multiple times pass the same cached bounds through the storage history indexing stage to avoid unnecessary RangeInclusive::clone keep the behaviour identical while trimming redundant allocations on hot paths


mattsse (commented):
range clone is fine because it's the same as copy here

PR #19606, opened 2025-11-08
fix(chainspec): use correct Jovian timestamp constant for Base Sepolia test

hawkadrian:
Fixes the base_sepolia_forkids test to use the correct BASE_SEPOLIA_JOVIAN_TIMESTAMP constant instead of OP_SEPOLIA_JOVIAN_TIMESTAMP, matching the actual chain spec configuration, and removes the obsolete TODO comment about updating the timestamp when Jovian is planned.


mattsse (approved):
this is okay, because this is the same superchain timestamp

but for consistency, I let this one through

PR #19611, opened 2025-11-09
fix(no_std): use `alloc::string::ToString` instead of `crate::alloc::...`

gap-editor:
**description**

* Replace `use crate::alloc::string::ToString;` with `use alloc::string::ToString;`.
* With `#![no_std]` and `extern crate alloc;`, the correct path to `ToString` is via the `alloc` crate root, not a local `crate::alloc` module.
* No functional changes.


mattsse (commented):
this is the same type

PR #19614, opened 2025-11-09
feat: add support for eip-7872 Max blob flag for local builders

### LINKED ISSUE #19613: Add support for eip-7872 Max blob flag for local builders
### Describe the feature

See eip for motivation https://eips.ethereum.org/EIPS/eip-7872

to support this we need a few things

## TODO
* add cli arg to PayloadBuilder cli args: https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/node/core/src/args/payload_builder.rs#L14-L21
* route this parameter to the actual ethereum builder impl which happens here: https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/ethereum/node/src/payload.rs#L55-L55, we can add this optional param to https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/node/core/src/cli/config.rs#L17-L17 and https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/ethereum/payload/src/config.rs#L6-L6
* then respect this param during building: https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/ethereum/payload/src/lib.rs#L201-L201 and cap it

### Additional context

_No response_

stevencartavia:
closes #19613

### DIFF
--- a/crates/ethereum/payload/src/lib.rs
+++ b/crates/ethereum/payload/src/lib.rs
@@ -198,8 +198,14 @@ where
     let mut block_transactions_rlp_length = 0;
 
     let blob_params = chain_spec.blob_params_at_timestamp(attributes.timestamp);
-    let max_blob_count =
+    let protocol_max_blob_count =
         blob_params.as_ref().map(|params| params.max_blob_count).unwrap_or_default();
+
+    // Apply user-configured blob limit (EIP-7872)
+    let max_blob_count = builder_config
+        .max_blobs_per_block
+        .map(|user_limit| std::cmp::min(user_limit as u64, protocol_max_blob_count).max(1))
+        .unwrap_or(protocol_max_blob_count);
 
     let is_osaka = chain_spec.is_osaka_active_at_timestamp(attributes.timestamp);
 
--- a/crates/ethereum/payload/src/lib.rs
+++ b/crates/ethereum/payload/src/lib.rs
@@ -199,12 +199,13 @@ where
 
     let blob_params = chain_spec.blob_params_at_timestamp(attributes.timestamp);
     let protocol_max_blob_count =
-        blob_params.as_ref().map(|params| params.max_blob_count).unwrap_or_default();
+        blob_params.as_ref().map(|params| params.max_blob_count).unwrap_or_else(Default::default);
 
     // Apply user-configured blob limit (EIP-7872)
+    // Per EIP-7872: if the minimum is zero, set it to one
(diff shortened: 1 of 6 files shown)


crates/ethereum/payload/src/lib.rs
```diff
@@ -198,9 +198,15 @@ where
     let mut block_transactions_rlp_length = 0;
 
     let blob_params = chain_spec.blob_params_at_timestamp(attributes.timestamp);
-    let max_blob_count =
+    let protocol_max_blob_count =
         blob_params.as_ref().map(|params| params.max_blob_count).unwrap_or_default();
```
mattsse: can we move this into unwrap_or_else

crates/node/core/src/args/payload_builder.rs
```diff
@@ -35,6 +35,10 @@ pub struct PayloadBuilderArgs {
     /// Maximum number of tasks to spawn for building a payload.
     #[arg(long = "builder.max-tasks", default_value = "3", value_parser = RangedU64ValueParser::<usize>::new().range(1..))]
     pub max_payload_tasks: usize,
+
+    /// Maximum number of blobs to include per block.
+    #[arg(long = "builder.max-blobs", value_name = "COUNT")]
+    pub max_blobs_per_block: Option<u8>,
```
mattsse: this will (hopefully) eventually exceed 256, so we can just make this a u64 instead, then we can avoid type casts

crates/ethereum/payload/src/lib.rs
```diff
@@ -198,9 +198,15 @@ where
...
 
+    // Apply user-configured blob limit (EIP-7872)
+    let max_blob_count = builder_config
+        .max_blobs_per_block
+        .map(|user_limit| std::cmp::min(user_limit as u64, protocol_max_blob_count).max(1))
```
mattsse: is max(1) part of the eip? that this can't be lower than that?
1 doesnt really make much sense really, so imo 0 is also a valid value here

mattsse (changes requested):
cool, smol nits

mattsse (approved):
lgtm!

PR #19616, opened 2025-11-09
refactor(rpc-convert): move rpc conversion traits/impls to alloy-evm

mablr:
Move rpc conversion traits/impls to alloy-evm.
https://github.com/alloy-rs/evm/pull/220


crates/rpc/rpc-convert/src/lib.rs
```diff
@@ -11,18 +11,16 @@
...
 
 pub use block::TryFromBlockResponse;
-pub use fees::{CallFees, CallFeesError};
 pub use receipt::TryFromReceiptResponse;
 pub use rpc::*;
 pub use transaction::{
-    EthTxEnvError, IntoRpcTx, RpcConvert, RpcConverter, TransactionConversionError,
-    TryFromTransactionResponse, TryIntoSimTx, TxInfoMapper,
+    RpcConvert, RpcConverter, TransactionConversionError, TryFromTransactionResponse, TryIntoSimTx,
+    TxInfoMapper,
```
mattsse, line 23: can we also re-export the types we moved to alloy-evm here?

then this becomes non-breaking

examples/custom-node/src/rpc.rs
```diff
@@ -3,14 +3,16 @@ use crate::{
     primitives::{CustomHeader, CustomTransaction},
 };
 use alloy_consensus::error::ValueError;
+use alloy_evm::{
+    rpc::{EthTxEnvError, TryIntoTxEnv},
+    EvmEnv,
+};
```
mattsse: and we don't need to update imports

mattsse (changes requested):
pedantic nit

examples/custom-node/src/rpc.rs
mablr: Hmm, that's weird `TryIntoTxEnv` was not re-exported before 🤨 while `EthTxEnvError` was.

Let's re-export everything we moved to alloy-evm.

examples/custom-node/src/rpc.rs
```diff
@@ -3,14 +3,15 @@ use crate::{
...
 };
 use alloy_consensus::error::ValueError;
+use alloy_evm::EvmEnv;
 use alloy_network::TxSigner;
 use op_alloy_consensus::OpTxEnvelope;
 use op_alloy_rpc_types::{OpTransactionReceipt, OpTransactionRequest};
 use reth_op::rpc::RpcTypes;
 use reth_rpc_api::eth::{
-    transaction::TryIntoTxEnv, EthTxEnvError, SignTxRequestError, SignableTxRequest, TryIntoSimTx,
+    EthTxEnvError, SignTxRequestError, SignableTxRequest, TryIntoSimTx, TryIntoTxEnv,
```
mablr, line 12: The only breaking thing: `TryIntoTxEnv` is now re-exported from `transaction` module.

Which is actually cleaner, I think.

examples/custom-node/src/rpc.rs
mattsse: I see

mattsse (approved):
lgtm

pending @klkvr

klkvr (approved):
nice!

Issue #16896, opened 2025-06-18
Support for ERA (not ERA1) hosts in the downloader


RomanHodulak:
### Describe the feature

Currently, the `reth_era_downloader` can deal with ERA1 hosts.

https://github.com/paradigmxyz/reth/blob/9ab57f70e3e5686b59a6f93fc41363d46707de73/crates/era-downloader/Cargo.toml#L2

The goal is to add support for ERA (not ERA1) hosts.

One notable difference is that compared to ERA1 hosts, these don't tend to contain checksums.txt, which is unfortunate as it makes file integrity checks impossible.

Another one is that, of course, the file format is different. However, the downloader is not dealing with the encoding at all, so this is not a concern.

PR #19617, opened 2025-11-10
refactor(era-downloader): support both era and era1 file types in downloader

lean-apple:
Towards https://github.com/paradigmxyz/reth/issues/16896.

Yet it would nice to have internal era files hosting for mainnet and hoodi files :)


crates/era-downloader/tests/it/download.rs
```diff
@@ -48,3 +48,19 @@ async fn test_getting_file_after_fetching_file_list(url: &str) {
     let actual_count = client.files_count().await;
     assert_eq!(actual_count, expected_count);
 }
+
+#[test_case("https://mainnet.era.nimbus.team/"; "nimbus")]
+#[tokio::test]
+async fn test_getting_era_file_url_after_fetching_file_list(url: &str) {
```
lean-apple, line 54: I don't know if it is worthy to also do era/era1 folders for test

crates/era-downloader/src/client.rs
```diff
@@ -156,68 +166,72 @@ impl<Http: HttpClient + Clone> EraClient<Http> {
         count
     }
 
-    /// Fetches the list of ERA1 files from `url` and stores it in a file located within `folder`.
+    /// Fetches the list of ERA1/ERA files from `url` and stores it in a file located within
+    /// `folder`.
+    /// For era files, checksum.txt file does not exist, so the checksum verification is
+    /// skipped.
     pub async fn fetch_file_list(&self) -> eyre::Result<()> {
```
lean-apple, line 173: @RomanHodulak I tried to make it slightly more readable but also to keep the parallel tasks.

crates/era-downloader/tests/it/download.rs
mattsse: I think this is sufficient

crates/era-downloader/src/client.rs
```diff
@@ -156,68 +166,72 @@ impl<Http: HttpClient + Clone> EraClient<Http> {
...
+        let ext = self.era_type.extension();
+        let ext_len = ext.len();
+
         while let Some(line) = lines.next_line().await? {
-            if let Some(j) = line.find(".era1") &&
+            if let Some(j) = line.find(ext) &&
                 let Some(i) = line[..j].rfind(|c: char| !c.is_alphanumeric() && c != '-')
             {
-                let era = &line[i + 1..j + 5];
+                let era = &line[i + 1..j + ext_len];
```
mattsse, line 212: makes sense

mattsse (approved):
very nice!

PR #19618, opened 2025-11-10
fix engine_getPayloadV4 compliance

CertiK-Geth:
Fix the issue on engine API: engine_getPayloadV4, to make it compliant with osaka engine API spec. 
According to issue listed in Fusaka contest: https://audits.sherlock.xyz/contests/1140/voting/156, the issue of API: engine_getPayloadV4 should also follow the spec restriction as the PR: https://github.com/paradigmxyz/reth/pull/18669 for API: getBlobsV1.

### DIFF
--- a/crates/rpc/rpc-engine-api/src/engine_api.rs
+++ b/crates/rpc/rpc-engine-api/src/engine_api.rs
@@ -464,6 +464,13 @@ where
         &self,
         payload_id: PayloadId,
     ) -> EngineApiResult<EngineT::ExecutionPayloadEnvelopeV4> {
+        let current_timestamp =
+            SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs();
+        if self.inner.chain_spec.is_osaka_active_at_timestamp(current_timestamp) {
+            return Err(EngineApiError::EngineObjectValidationError(
+                reth_payload_primitives::EngineObjectValidationError::UnsupportedFork,
+            ));
+        }
         self.get_payload_inner(payload_id, EngineApiMessageVersion::V4).await
     }
 


crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -464,6 +464,13 @@ where
         &self,
         payload_id: PayloadId,
     ) -> EngineApiResult<EngineT::ExecutionPayloadEnvelopeV4> {
+        let current_timestamp =
+            SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs();
+        if self.inner.chain_spec.is_osaka_active_at_timestamp(current_timestamp) {
+            return Err(EngineApiError::EngineObjectValidationError(
+                reth_payload_primitives::EngineObjectValidationError::UnsupportedFork,
+            ));
+        }
```
mattsse, line 473: we can't use a system time check because this might be called via the opstack engineapi as well, which can derive the chain via newpayload.

this this must use the timestamp of the payload itself in `get_payload_inner`

the tricky part is that now getpyload and newpayload versions diverged, so we need to handle this 

here

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc-engine-api/src/engine_api.rs#L365-L367

and cant add this here

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc-engine-api/src/engine_api.rs#L365-L367

mattsse (changes requested):
thanks, but relying on the systemtime here isnt appropriate

crates/rpc/rpc-engine-api/src/engine_api.rs
CertiK-Geth: Thanks, LGTM in this case.

mattsse (commented):
superseded by #19929

PR #19619, opened 2025-11-10
fix: log correct account worker count

dennsikl:
The span for account proof workers was still logging storage_worker_count because of a copy/paste slip. Swapped it to account_worker_count so the tracing output matches the pool we actually spawn.


mattsse (approved):
lgtm

pending @yongkangc

shekhirin (approved):

PR #19620, opened 2025-11-10
fix(net): fix update_root ENR/LINK handling

Forostovec:
update_root handled partial root changes inverted: when only link_root changed it set SyncState::Enr, and when only enr_root changed it set SyncState::Link. Additionally, the case where both roots changed was treated as “unchanged”, and the case where neither changed triggered a full resync (Pending). This contradicted EIP-1459 semantics (enr-root -> ENR subtree, link-root -> link subtree), caused incorrect query sequences, potential missed updates, and unnecessary DNS lookups, and led to “unexpected entry” logs during resolution.

Fixes:
- Renamed local flags to enr_unchanged/link_unchanged for clarity.

Corrected the state mapping:
- (true, true): early return (no resync).
- (false, true): clear unresolved_nodes, set SyncState::Enr (ENR changed).
- (true, false): clear unresolved_links, set SyncState::Link (LINK changed).
- (false, false): clear both, set SyncState::Pending (both changed).

Added unit tests covering all four scenarios and verifying emitted SyncAction ordering (Link then Enr when both changed).

### DIFF
--- a/crates/net/dns/src/lib.rs
+++ b/crates/net/dns/src/lib.rs
@@ -583,8 +583,14 @@ mod tests {
         // await recheck timeout
         tokio::time::sleep(config.recheck_interval).await;
 
+        let mut new_root = root.clone();
+        new_root.sequence_number = new_root.sequence_number.saturating_add(1);
+        new_root.enr_root = "NEW_ENR_ROOT".to_string();
+        new_root.sign(&secret_key).unwrap();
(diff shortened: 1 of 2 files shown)


mattsse (approved):
ah, nice find, this is indeed flipped

mattsse (changes requested):
but I think we now need to update the existing test_recheck_tree test as well

mattsse (approved):

PR #19622, opened 2025-11-10
feat: debug_storageRangeAt RPC Method

0xKarl98:
This closes issue #17609  , short summary : 

 - Implements debug_storageRangeAt with full historical accuracy
 - Adds a storage-range capable provider path, runs block replay with StorageDiffInspector, overlays per-transaction diffs via StorageRangeOverlay, and returns QuickNode-compatible results. 
 - Includes the necessary trait extensions, wiring through BlockchainProvider, and helper tests  
 
 Covers by the reference materials here :
 - https://github.com/paradigmxyz/reth/pull/18033
 - https://www.quicknode.com/docs/ethereum/debug_storageRangeAt
 
 cc @mattsse @shekhirin


crates/rpc/rpc-eth-api/src/helpers/storage_diff_inspector.rs
mattsse, line 1: once approved, we should move this one to the inspectors crate

crates/rpc/rpc/Cargo.toml
```diff
@@ -40,6 +40,7 @@ reth-consensus.workspace = true
 reth-consensus-common.workspace = true
 reth-node-api.workspace = true
 reth-trie-common.workspace = true
+secp256k1.workspace = true
```
mattsse, line 45: why is this needed?

crates/rpc/rpc/src/debug.rs
```diff
@@ -1014,13 +1157,15 @@ where
...
-        _max_result: u64,
-    ) -> RpcResult<()> {
-        Ok(())
+        block_hash: B256,
+        tx_idx: usize,
+        contract_address: Address,
+        key_start: B256,
+        max_result: u64,
+    ) -> RpcResult<RpcStorageRangeResult> {
+        self.storage_range_at(block_hash, tx_idx, contract_address, key_start, max_result)
```
mattsse: this should use `Self::` syntax

crates/rpc/rpc/src/debug.rs
```diff
@@ -208,6 +211,133 @@ where
...
+            return Err(EthApiError::InvalidParams(format!(
+                "transaction index {tx_idx} out of bounds for block with {tx_count} transactions"
+            ))
+            .into())
+        }
+
+        let storage_provider = self
+            .provider()
+            .storage_range_by_block_hash(block_hash)
+            .map_err(Eth::Error::from_eth_err)?;
```
mattsse: this does some IO, right?

so this should be spawned as blocking

crates/storage/storage-api/src/state.rs
```diff
@@ -92,6 +109,20 @@ pub trait StateProvider:
...
+    /// Returns storage slots for `account` starting at `start_key` (inclusive) and capped by
+    /// `max_slots`. When additional slots are available beyond `max_slots`, `next_key` in the
+    /// result indicates where the caller should resume.
+    fn storage_range(
+        &self,
+        account: Address,
+        start_key: StorageKey,
+        max_slots: usize,
+    ) -> ProviderResult<StorageRangeResult>;
+}
```
mattsse, line 122: maybe @Rjected or @shekhirin have an opinion on how to implement this feature on the existing provider stack

mattsse (changes requested):
a few suggestions

and would like some input from re the db api extensions

crates/rpc/rpc/Cargo.toml
0xKarl98: Because reth-rpc directly depends on secp256k1 (tests in crates/rpc/rpc/src/debug.rs use Secp256k1, SecretKey, Keypair, and public_key_to_address, and the crate has used secp256k1 as _; to satisfy unused-deps lint), it must declare it as a direct dependency.

crates/rpc/rpc/src/debug.rs
0xKarl98: Thought they are equivelant tho , but will change that

crates/rpc/rpc/src/debug.rs
0xKarl98: good idea , have changed that

wtdcode:
Any progress on this?

crates/storage/storage-api/src/state.rs
wtdcode: Also ping @Rjected and @shekhirin if you have an idea about the intended provider design.

@0xKarl98 I would assume this is the biggest blocker.

0xKarl98: I removed the Send + Sync bounds to avoid forcing thread‑safety on providers that are not designed to be shared across threads (e.g., types holding DbTx or &Provider references)

wdyt

wtdcode:
@0xKarl98 Hi, any reason closing this? Or accidentally removing your branch?

@mattsse Can I take over this?

PR #19624, opened 2025-11-10
perf(rpc): Improve trace_filter block range processing

### LINKED ISSUE #19528: Improve trace_filter block range processing
### Describe the feature

currently we do this sequentially:

https://github.com/paradigmxyz/reth/blob/ba8be3fb64d51d712f62ce4ed26d46842bd155e3/crates/rpc/rpc/src/trace.rs#L398-L415

this does

1. spawn a blocking task to fetch the block range
2. the spawn tracing future per block

this can be significantly improved by doing things concurrently and also leveraging the rpc cache, we already have a few optional helpers like:

https://github.com/paradigmxyz/reth/blob/ba8be3fb64d51d712f62ce4ed26d46842bd155e3/crates/rpc/rpc-eth-types/src/cache/mod.rs#L206-L207

that we could use here as well if the requested range is close to the head block (within ~1000 blocks)

## TODO
* convert this function into something that yields fetched blocks to something that can  process them concurrently 

```rust
  let mut tracing_futures_ordered =...;
  let mut block_stream_ordered =...;
  tokio::select! {
       block_stream_ordered.next() => tracing_futures_ordered.push(...),
       tracing_futures_ordered.next() => {process trace, maybe exit}
   } 

```

### Additional context

_No response_

frankudoags:
closes https://github.com/paradigmxyz/reth/issues/19528

### DIFF
--- a/crates/rpc/rpc/src/trace.rs
+++ b/crates/rpc/rpc/src/trace.rs
@@ -933,12 +933,12 @@ where
         // block_num as a u64
         let block_hash = FixedBytes::random();
 
-        // Try cache first using maybe_cached_block_and_receipts
-        let (maybe_block, _receipts) = self
+        // Try cache first using get_recovered_block
+        let maybe_block = self
             .trace_api
             .eth_api
             .cache()
-            .maybe_cached_block_and_receipts(block_hash)
+            .get_recovered_block(block_hash)
             .await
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this is unfortunately too complex for now

I think we need to delay this a bit until we have proper additional index support data to make this actually viable

PR #19626, opened 2025-11-10
feat(bench-compare): add OTLP tracing support for baseline/feature comparison

yongkangc:
Adds OpenTelemetry tracing support to `reth-bench-compare` to enable distributed trace comparison between baseline and feature benchmark runs.

### CLI Arguments
Added `TraceArgs` to support standard OTLP tracing configuration:
- `--tracing-otlp[=URL]` - OTLP endpoint (default: http://localhost:4318/v1/traces)

### Service Name Differentiation
Uses `OTEL_SERVICE_NAME` environment variable to create distinct services in Jaeger:
- **Baseline runs**: `reth-baseline`
- **Feature runs**: `reth-feature`

This allows easy filtering and side-by-side comparison in the Jaeger UI.

## Usage Example

1. Start Jaeger
```bash
docker run -d --name jaeger \
  -p 4318:4318 \
  -p 16686:16686 \
  jaegertracing/all-in-one:latest
```

2. Run Benchmark with Tracing
```bash
reth-bench-compare \
  --baseline-ref main \
  --feature-ref my-optimization \
  --tracing-otlp=http://localhost:4318/v1/traces \
  --blocks 100
```

4. Compare Traces in Jaeger
- Navigate to http://localhost:16686
- Select `reth-baseline` service → view baseline traces
- Select `reth-feature` service → view feature traces
- Use Jaeger's "Compare" feature for side-by-side analysis

image


bin/reth-bench-compare/src/node.rs
```diff
@@ -119,6 +121,7 @@ impl NodeManager {
         &self,
         binary_path_str: &str,
         additional_args: &[String],
+        ref_type: &str,
```
mediocregopher, line 124: nit: if this is only used for logging I say let's leave it out, we could use a span to add it to all logs from outside this call, or just infer the ref_type from context when reading the logs

bin/reth-bench-compare/src/node.rs
```diff
@@ -146,6 +149,13 @@ impl NodeManager {
             "--trusted-only".to_string(),
         ]);
 
+        // Add tracing arguments if OTLP endpoint is configured
+        if let Some(ref endpoint) = self.tracing_endpoint {
+            info!("Enabling OTLP tracing export to: {} (service: reth-{})", endpoint, ref_type);
+            // Endpoint requires equals per clap settings in reth
+            reth_args.push(format!("--tracing-otlp={}", endpoint));
```
shekhirin, line 156: should we just add `--tracing-otlp.service-name` argument here?

mediocregopher (approved):
minor nit but LGTM otherwise

bin/reth-bench-compare/src/node.rs
yongkangc: the ref_type functionality will allow us to separate service within reth-bench-compare, easier for filtering

bin/reth-bench-compare/src/node.rs
Rjected: Do we have this flag in reth yet?

yongkangc: not yet we dont @Rjected 

>Do we have this flag in reth yet?

yongkangc: i think having it as env works for now? since we would have to add this flag in reth as well

i dont see a downside for using env?  @shekhirin

shekhirin: makes sense

Rjected: oh nvm saw that this PR added it

shekhirin (approved):

PR #19629, opened 2025-11-10
chore: bump op-revm v12.0.2 patch

rakita:
Bump op-revm patch to fix op rpc panic


mattsse (approved):
can we get a release here as well to see what's included https://github.com/bluealloy/revm/releases

PR #19630, opened 2025-11-10
feat(provider, prune): return receipts log filter

### LINKED ISSUE #19621: `v1.9.1` crashes immediately on Hoodi
### Describe the bug

We tried to upgrade from `1.8.4` to `1.9.1`, unfortunately we are getting this log

We tried to enable trace logs (`-vvvvv`), but the output above is all we got. Downgrading to `1.8.4` fixes the issue

CL: lighthouse `8.0.0`

### Steps to reproduce

1. Update to `v1.9.1`
2. Start container

### Node logs

```text
reth_hoodi-1  | 2025-11-06T20:12:25.131782Z  INFO StaticFileProducer initialized
reth_hoodi-1  | 2025-11-06T20:12:25.135575Z ERROR shutting down due to error
reth_hoodi-1  | Error: trying to append row to Receipts at index #488 but expected index #239
reth_hoodi-1  | 
reth_hoodi-1  | Location:
reth_hoodi-1  |     /home/runner/work/reth/reth/crates/node/builder/src/launch/engine.rs:157:9
```

### Platform(s)

Linux (x86)

### Container Type

Docker

`v1.9.1`

```
Current database version: 2
Local database version: 2
```

Hoodi

Pruned with custom reth.toml config

```
[prune]
block_interval = 5

# Stakewise
[prune.segments.receipts_log_filter.0xa7d1ac9d6f32b404c75626874ba56f7654c1dc0f]
before = 94070

[prune.segments.receipts_log_filter.0x00000000219ab540356cBB839Cbe05303d7705Fa]
before = 0
```

_No response_

### Code of Conduct

shekhirin:
Fixes https://github.com/paradigmxyz/reth/issues/19621

This PR returns the previously removed receipts log filter pruning, but makes it work only with `Full` and `Before` prune modes. Most of the changes are reverted from https://github.com/paradigmxyz/reth/pull/19184.

We're returning it, because some stakers depend on having receipts from certain staking protocols, and they don't want to store all 300GB+ of receipts on Mainnet.

We don't have `Distance` prune mode for this segment anymore because it's not useful for the type of pruning stakers do, and for us to support it, we would need to add a prune segment back. `Full` and `Before` modes work without a separate pruning segment, writing only specific receipts at the time of persisting.

### DIFF
--- a/crates/prune/prune/src/segments/set.rs
+++ b/crates/prune/prune/src/segments/set.rs
@@ -59,7 +59,6 @@ where
         _static_file_provider: StaticFileProvider<Provider::Primitives>,
         prune_modes: PruneModes,
     ) -> Self {
-        #[expect(deprecated)]
         let PruneModes {
             sender_recovery,
             transaction_lookup,
@@ -68,7 +67,7 @@ where
             storage_history,
             bodies_history,
             merkle_changesets,
-            receipts_log_filter: (),
+            receipts_log_filter: _,
         } = prune_modes;
 
         Self::default()
--- a/crates/prune/types/src/lib.rs
+++ b/crates/prune/types/src/lib.rs
@@ -22,7 +22,7 @@ use std::collections::BTreeMap;
 
 use alloy_primitives::{Address, BlockNumber};
 pub use checkpoint::PruneCheckpoint;
-use derive_more::{Deref, From};
+use derive_more::{Deref, DerefMut, From};
 pub use event::PrunerEvent;
 pub use mode::PruneMode;
 pub use pruner::{
@@ -30,17 +30,17 @@ pub use pruner::{
     SegmentOutputCheckpoint,
 };
 pub use segment::{PrunePurpose, PruneSegment, PruneSegmentError};
-use serde::Deserialize;
+use serde::{de::Error, Deserialize, Deserializer};
 pub use target::{PruneModes, UnwindTargetPrunedError, MINIMUM_PRUNING_DISTANCE};
 
 /// Configuration for pruning receipts not associated with logs emitted by the specified contracts.
-#[derive(Debug, Clone, PartialEq, Eq, Default, Deref, From)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, Deref, DerefMut, From)]
 #[cfg_attr(any(test, feature = "serde"), derive(serde::Serialize, serde::Deserialize))]
 pub struct ReceiptsLogPruneConfig(pub BTreeMap<Address, PruneMode>);
 
 impl ReceiptsLogPruneConfig {
     /// Creates an empty config.
-    pub const fn empty() -> Self {
+    pub const fn new() -> Self {
         Self(BTreeMap::new())
     }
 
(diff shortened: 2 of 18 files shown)


crates/storage/provider/src/either_writer.rs
```diff
@@ -16,6 +18,32 @@ pub enum EitherWriter<'a, CURSOR, N> {
     StaticFile(StaticFileProviderRWRefMut<'a, N>),
 }
 
+#[derive(Debug, EnumIs)]
+#[allow(missing_docs)]
+pub enum EitherWriterDestination {
+    Database,
+    StaticFile,
+}
```
mattsse: this type is out of order

crates/storage/provider/src/either_writer.rs
```diff
@@ -16,6 +18,32 @@ pub enum EitherWriter<'a, CURSOR, N> {
...
+
+impl EitherWriter<'_, (), ()> {
+    /// Returns the destination for writing receipts.
```
mattsse, line 22: this needs docs about what the rules are here

crates/prune/types/src/segment.rs
```diff
@@ -115,6 +116,9 @@ pub enum PruneSegmentError {
     /// Invalid configuration of a prune segment.
     #[error("the configuration provided for {0} is invalid")]
     Configuration(PruneSegment),
+    /// Unsupported receipts log filter prune mode for address.
+    #[error("distance prune mode is not supported for receipts log filter, check address {0}")]
```
mattsse: what should be checked here with the address?

crates/prune/types/src/lib.rs
mattsse, line 1: is this what we had previously?

can we reference the previous pr in the pr description?

mattsse (changes requested):
a few nits

definitely need @joshieDo to review as well and need to take a look at the previous pr

crates/storage/provider/src/either_writer.rs
shekhirin: done

crates/storage/provider/src/either_writer.rs
shekhirin: done

crates/prune/types/src/segment.rs
shekhirin: fixed

crates/prune/types/src/target.rs
```diff
@@ -100,9 +100,11 @@ pub struct PruneModes {
     )]
     pub merkle_changesets: PruneMode,
     /// Receipts log filtering has been deprecated and will be removed in a future release.
```
joshieDo:
```suggestion
    /// Receipts pruning configuration by retaining only those receipts that contain logs emitted
    /// by the specified addresses, discarding others. This setting is overridden by `receipts`.
    ///
    /// The [`BlockNumber`](`crate::BlockNumber`) represents the starting block from which point
    /// onwards the receipts are preserved.
```

crates/prune/prune/src/segments/set.rs
joshieDo, line 84: do we no longer need the [following](https://github.com/paradigmxyz/reth/pull/19184/files#diff-180e595d6b89d64b1246bbd46f2553dd41f529f6a9db3aacdc377f226f1ff994)? 

```
           // Receipts by logs
            .segment_opt(
                (!receipts_log_filter.is_empty())
                    .then(|| ReceiptsByLogs::new(receipts_log_filter.clone())),
            )
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -264,7 +264,7 @@ impl<TX, N: NodeTypes> AsRef<Self> for DatabaseProvider<TX, N> {
     }
 }
 
-impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX, N> {
+impl<TX: DbTx + DbTxMut + 'static, N: NodeTypes + NodeTypesForProvider> DatabaseProvider<TX, N> {
```
joshieDo: `NodeTypes + ` is this necessary?

crates/prune/prune/src/segments/set.rs
shekhirin: nope, because we don't have a prune segment anymore! Pruning for receipts log filter is now handled at the time of writing.

crates/storage/provider/src/providers/database/provider.rs
shekhirin: no, removed

crates/prune/prune/src/segments/set.rs
joshieDo: dont we follow the `MINIMUM_PRUNING_DISTANCE` "rule", so we do need the pruner to remove the other receipts once its out of that range no?

emmajam:
@shekhirin do we still want this?

PR #19632, opened 2025-11-10
fix: require config path when not using --default

strmfos:
Without this change `reth config` silently tried to read from an empty path whenever neither --default nor --config was provided, which produced a confusing “config file does not exist:” error; now the command explicitly asks for --config <FILE> or --default, so users immediately know how to proceed.


mattsse (approved):
this makes sense

PR #19633, opened 2025-11-10
chore(trie): Add metrics to track all cursor operations in proof workers

mediocregopher:
This PR implements instrumentation wrappers for TrieCursor and HashedCursor. Each wrapper records data into cache structs which can then be extracted and either inspected or recorded via prometheus.

This functionality is then used to instrument the proof workers, such that:
* We include in traces the number of each kind of cursor method which was invoked for a particular proof, giving an idea of overall database operations.
* At the end of each proof worker's lifetime we record in prometheus the overall operations performed.


yongkangc (approved):
Lgtm! This would be useful

> This functionality is then used to instrument the proof workers, such that:
> * We include in traces the number of each kind of cursor method which was invoked for a particular proof, giving an idea of overall database operations.
> * At the end of each proof worker's lifetime we record in prometheus the overall operations performed.

mattsse (approved):
lgtm!

shekhirin (approved):

mediocregopher:
Confirmed no perf regression from these changes

image

PR #19634, opened 2025-11-10
chore: remove unused latest_update_kind from TxPool

phrwlk:
The latest_update_kind field in TxPool was write-only and never read by any component, tests, or metrics. PoolUpdateKind still flows through callers via CanonicalStateUpdate and into TxPool::on_canonical_state_change, but no consumer needs the last update kind persisted in TxPool. This removes the dead state, its initialization, and its only assignment, and renames the parameter to update_kind to avoid unused variable warnings. This keeps API behavior unchanged for callers and reduces unnecessary state.


mattsse (approved):
we can indeed remove this because not pub

PR #19637, opened 2025-11-10
fix(storage): validate blocks_per_file to prevent division by zero

prestoalvarez:
The `StaticFileProviderBuilder` methods `with_blocks_per_file()` and `with_blocks_per_file_for_segment()` accepted `0` as a valid value, which could lead to division by zero panics when calculating block ranges in `find_fixed_range_with_block_index()`.


mattsse (commented):
not reachable

PR #19640, opened 2025-11-10
refactor: capped `to` to `best_block`

### LINKED ISSUE #19636: cap --to target at existing block height
### Describe the feature

we currently don't cap the --to arg to min(best_block_number) which can cause panics

https://github.com/paradigmxyz/reth/blob/7b3300841f9fbce247db7702fd049fd755d985ab/crates/cli/commands/src/re_execute.rs#L68-L68

## TODO

* if to is some and > best_block, log a warn and use `max()` instead

### Additional context

_No response_

akronim26:
# Fixes #19636 
- Caps the requested `to` block at the current `best_block`.
- If the user requests a block higher than `best_block`, execution now emits a warning instead of silently proceeding.

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -66,19 +66,17 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
 
         let min_block = self.from;
         let best_block = provider.best_block_number()?;
-        let max_block = if let Some(to) = self.to {
+        let mut max_block = best_block;
+        if let Some(to) = self.to {
             if to > best_block {
                 warn!(
                     requested = to,
                     best_block,
                     "Requested --to is beyond available chain head; clamping to best block"
                 );
-                best_block
             } else {
-                to
+                max_block = to;
             }
-        } else {
-            best_block
         };
 
         let total_blocks = max_block - min_block;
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/re_execute.rs
```diff
@@ -65,7 +65,21 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
         let components = components(provider_factory.chain_spec());
 
         let min_block = self.from;
-        let max_block = self.to.unwrap_or(provider.best_block_number()?);
+        let best_block = provider.best_block_number()?;
```
mattsse, line 68: we can make this `let mut max_block`

crates/cli/commands/src/re_execute.rs
```diff
@@ -65,7 +65,21 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
+        let max_block = if let Some(to) = self.to {
+            if to > best_block {
+                warn!(
+                    requested = to,
+                    best_block,
+                    "Requested --to is beyond available chain head; clamping to best block"
+                );
+                best_block
```
mattsse: and then this just becomes max_block = to

mattsse (changes requested):
smol style suggestion

mattsse (approved):
cool, ty

PR #19642, opened 2025-11-11
chore: Remove unnecessary dead_code attributes from benchmark utils

JacksonEi:
Those constants and functions in are actively used across multiple benchmark files, so removed `#[allow(dead_code)]` attributes.


mattsse (commented):
we need to keep these due to how clippy works

PR #19645, opened 2025-11-11
fix: preserve parent beacon root in pending env

aso20455:
- stop zeroing `parent_beacon_block_root` when we build pending env attrs
- add a small regression test that fails if we reintroduce the bug


mattsse (commented):
what's the motivation for this?

iirc zero is the default for eth_simulatv1, so this seems also appropriate here

mattsse (approved):
yeah I dont see any harm in doing this

PR #19661, opened 2025-11-11
Remove unused SignerRecoverable import from txpool tracing example

Doryu001:
drop the stale SignerRecoverable import in examples/txpool-tracing/src/submit.rs keep the example free of dead code and avoid future unused-import warnings


mattsse (commented):
this import is required

Issue #19589, opened 2025-11-07
Make sure that we can just delete static files and it will not break provider access


shekhirin:
### Describe the feature

```console
❯ ls /mnt/nvme/reth/mainnet/static_files/static_file_* | first 6
╭───┬────────────────────────────────────────────────────────────────────────────────┬──────┬──────────┬────────────╮
│ # │                                      name                                      │ type │   size   │  modified  │
├───┼────────────────────────────────────────────────────────────────────────────────┼──────┼──────────┼────────────┤
│ 0 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_0_499999               │ file │ 158.2 MB │ a year ago │
│ 1 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_0_499999.conf          │ file │     75 B │ a year ago │
│ 2 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_0_499999.off           │ file │  12.0 MB │ a year ago │
│ 3 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_10000000_10499999      │ file │ 276.8 MB │ a year ago │
│ 4 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_10000000_10499999.conf │ file │     75 B │ a year ago │
│ 5 │ /mnt/nvme/reth/mainnet/static_files/static_file_headers_10000000_10499999.off  │ file │  12.0 MB │ a year ago │
╰───┴────────────────────────────────────────────────────────────────────────────────┴──────┴──────────┴────────────╯
```

Make sure that I can just delete `static_file_headers_0_499999`, `static_file_headers_0_499999.conf` and `static_file_headers_0_499999.off` files, and the provider will start returning `None` for all headers in that range.

shekhirin:
We can't delete all files for a segment right now, because if you delete all transaction static files, then the consistency check will get triggered with the unwind to block 0

```console
2025-11-13T19:16:16.036968Z  INFO Setting unwind target. checkpoint_block_number=23786102 unwind_target=0 segment=TransactionSenders

thread 'main' panicked at /home/shekhirin/reth/crates/cli/commands/src/common.rs:154:13:
assertion `left != right` failed: A static file <> database inconsistency was found that would trigger an unwind to block 0
```

[...]

PR #19662, opened 2025-11-11
feat(stages): get transaction range starting from first available block

shekhirin:
Towards https://github.com/paradigmxyz/reth/issues/19589

Ensures that we don't return a range of transactions that aren't available in static files anymore, in case the user deleted the static files manually.


crates/stages/api/src/stage.rs
```diff
@@ -70,15 +71,24 @@ impl ExecInput {
     /// Return the next block range determined the number of transactions within it.
     /// This function walks the block indices until either the end of the range is reached or
     /// the number of transactions exceeds the threshold.
```
mattsse, line 73: not obvious to me what this function returns

```
(Range<TxNumber>, RangeInclusive<BlockNumber>, bool)
```

can we either overexplain what this is or introduce a struct for this

crates/stages/api/src/stage.rs
```diff
@@ -70,15 +71,24 @@ impl ExecInput {
...
     ) -> Result<(Range<TxNumber>, RangeInclusive<BlockNumber>, bool), StageError>
     where
-        Provider: BlockReader,
+        Provider: StaticFileProviderFactory + BlockReader,
     {
-        let start_block = self.next_block();
+        let Some(lowest_transactions_block) = provider
+            .static_file_provider()
+            .get_lowest_range(StaticFileSegment::Transactions)
+            .map(|range| range.start())
```
mattsse: can we make this a helper, I'd prefer this than having map range .start

crates/stages/api/src/stage.rs
```diff
@@ -70,15 +71,24 @@ impl ExecInput {
...
+        else {
+            return Ok((0..0, 0..=0, true));
+        };
+
+        let start_block = self.next_block().max(lowest_transactions_block);
```
mattsse, line 96: why is this capped?

mattsse (changes requested):
very pedantic nits

crates/stages/api/src/stage.rs
shekhirin: I will introduce a struct in a follow-up PR, this is unrelated to current changes

crates/stages/api/src/stage.rs
shekhirin: explained in a comment

crates/stages/api/src/stage.rs
mattsse: reasonable

mattsse (approved):

PR #19663, opened 2025-11-11
chore(examples): trim unused imports in txpool tracing submit

0xLogicalx:
remove unused alloy and reth transaction pool imports from the txpool tracing submit helper keep the example focused on the types it actually consumes


mattsse (commented):
this example is fine as is

PR #19664, opened 2025-11-12
refactor(provider): explicit static file segment matches

shekhirin:
When adding new segments, it can be easy to miss these `if transactions { transaction } else { receipt }`, so I refactored them into an explicit matches over static file segment enum.

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -1196,20 +1196,21 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                     // TODO(joshie): is_block_meta
                     writer.prune_headers(highest_static_file_block - checkpoint_block_number)?;
                 }
-                StaticFileSegment::Transactions => {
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1191,17 +1191,27 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
-                    writer.prune_transactions(number, checkpoint_block_number)?;
+            match segment {
+                StaticFileSegment::Headers => {
+                    // TODO(joshie): is_block_meta
+                    writer.prune_headers(highest_static_file_block - checkpoint_block_number)?;
+                }
+                StaticFileSegment::Transactions | StaticFileSegment::Receipts => {
+                    if let Some(block) = provider.block_body_indices(checkpoint_block_number)? {
+                        // todo joshie: is querying block_body_indices a potential issue once bbi is
+                        // moved to sf as well
```
joshieDo:
```suggestion
```

drive-by: no longer applicable since block body indices will remain in DB

joshieDo (approved):

PR #19666, opened 2025-11-12
feat: limit handling of incoming txs to trusted peers

0xKarl98:
This closes issue #19652 , please check if it meest the requirement @mattsse @Rjected @klkvr 

closes #19652

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -1283,11 +1283,11 @@ where
     }
 
     /// Returns true if the ingress policy allows processing messages from the given peer.
-    fn accepts_incoming_from(&self, peer_id: PeerId) -> bool {
+    fn accepts_incoming_from(&self, peer_id: &PeerId) -> bool {
         if self.config.ingress_policy.allows_all() {
             return true;
         }
-        let Some(peer) = self.peers.get(&peer_id) else {
+        let Some(peer) = self.peers.get(peer_id) else {
             return false;
         };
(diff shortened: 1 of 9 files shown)


crates/node/core/src/args/network.rs
```diff
@@ -162,6 +162,12 @@ pub struct NetworkArgs {
     #[arg(long = "tx-propagation-policy", default_value_t = TransactionPropagationKind::All)]
     pub tx_propagation_policy: TransactionPropagationKind,
 
+    /// Transaction ingress policy
+    ///
+    /// Determines which peers' transactions are accepted over P2P.
+    #[arg(long = "tx-ingress-policy", default_value_t = TransactionIngressPolicy::All)]
+    pub tx_ingress_policy: TransactionIngressPolicy,
```
mattsse, line 169: this makes sense

crates/net/network/src/transactions/mod.rs
```diff
@@ -1282,10 +1282,25 @@ where
         }
     }
 
+    /// Returns true if the ingress policy allows processing messages from the given peer.
+    fn accepts_incoming_from(&self, peer_id: PeerId) -> bool {
+        if self.config.ingress_policy.allows_all() {
+            return true;
+        }
```
mattsse, line 1289: this is great, then this is basically 0 overhead for the default case

mattsse (approved):
this is great, clean


this has effectively 0 overhead for the default case

PR #19667, opened 2025-11-12
feat(flashblock): improve state root calculation condition

0x00101010:
Towards: https://github.com/paradigmxyz/reth/issues/18255

Improve flashblock state_root calculation condition


crates/optimism/flashblocks/src/sequence.rs
```diff
@@ -13,18 +13,24 @@ use tracing::{debug, trace, warn};
...
 const FLASHBLOCK_SEQUENCE_CHANNEL_SIZE: usize = 128;
 
+/// Outcome from executing a flashblock sequence.
+#[derive(Debug, Clone, Copy)]
+pub struct SequenceExecutionOutcome {
+    /// The block hash of the executed pending block
+    pub block_hash: B256,
+    /// Properly computed state root
+    pub state_root: B256,
+}
```
mattsse, line 23: this is better, yes

crates/optimism/flashblocks/src/service.rs
```diff
@@ -190,9 +185,38 @@ where
...
 
-        // Check if state root must be computed
-        let compute_state_root =
-            self.compute_state_root && self.blocks.index() >= Some(FB_STATE_ROOT_FROM_INDEX as u64);
+        // Auto-detect when to compute state root: only if the builder didn't provide it (sent
+        // B256::ZERO) and we're near the expected final flashblock index.
+        //
+        // Background: Each block period receives multiple flashblocks at regular intervals.
+        // The sequencer sends an initial "base" flashblock at index 0 when a new block starts,
+        // then subsequent flashblocks are produced every FLASHBLOCK_BLOCK_TIME intervals (200ms).
```
mattsse, line 202: this is very helpful

crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -506,9 +515,15 @@ where
...
             let in_progress_rx = service.subscribe_in_progress();
-
             ctx.components.task_executor().spawn(Box::pin(service.run(tx)));
 
+            info!(target: "reth::cli", "Launching FlashBlockConsensusClient");
+            let flashblock_client = FlashBlockConsensusClient::new(
+                ctx.engine_handle.clone(),
+                flashblocks_sequence.subscribe(),
+            )?;
+            ctx.components.task_executor().spawn(Box::pin(flashblock_client.run()));
```
mattsse: I think we need one more setting for this 

maybe we call this `--flashblock-consensus` or `--flashblock-engine` ?

the we wire from cli to Opethapibuilder


https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/optimism/rpc/src/eth/mod.rs#L390-L390

like we do with flashblocks url:

https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/optimism/node/src/node.rs#L196-L196

https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/optimism/node/src/node.rs#L838-L838

crates/optimism/flashblocks/src/consensus.rs
```diff
@@ -1,86 +1,143 @@
...
-                    // Send FCU
-                    let _ = self
-                        .engine_handle
-                        .fork_choice_updated(state, None, EngineApiMessageVersion::V3)
-                        .await;
+            // FCU to update tip, only after successful engine_newPayload
+            let fcu_state = alloy_rpc_types_engine::ForkchoiceState {
+                head_block_hash: block_hash,
+                safe_block_hash: B256::ZERO,
+                finalized_block_hash: B256::ZERO,
```
mattsse: this works once the node caught up, but if you want to sync a new node via flashblocks then we also need to choose a finalized block so we can kick off p2p backfill sync

I think we could simply use the parent block here for now?

ideally we track the some recent blocks based on 32slots*6secs blocks here and assume that new blocks are being built on the canonical chain

mattsse (changes requested):
this is great, only have two last nits

crates/optimism/flashblocks/src/consensus.rs
0x00101010: I see, so we do want to use this entirely without op-node. In that case, I'll add back the tracking feature

crates/optimism/rpc/src/eth/mod.rs
0x00101010: Done, also reworked `FlashblockConsensusClient` a little

crates/optimism/flashblocks/src/service.rs
```diff
@@ -190,9 +185,38 @@ where
...
+        //
+        // Note: Pathological cases may result in fewer flashblocks than expected (e.g., builder
+        // downtime, flashblock execution exceeding timing budget). When this occurs, we won't
+        // compute the state root, causing FlashblockConsensusClient to lack precomputed state for
+        // engine_newPayload. This is safe: we still have op-node as backstop to maintain
+        // chain progression.
+        let block_time_ms = (base.timestamp - latest.timestamp()) * 1000;
+        let expected_final_flashblock = block_time_ms / FLASHBLOCK_BLOCK_TIME;
+        let compute_state_root = last_flashblock.diff.state_root.is_zero() &&
+            self.blocks.index() >= Some(expected_final_flashblock.saturating_sub(1));
```
mattsse, line 229: can we put this behind `if self.compute_state_root` because only then we need to check this and derive whether to compute it from this

klkvr (approved):

crates/optimism/flashblocks/src/service.rs
```diff
@@ -60,8 +61,6 @@ pub struct FlashBlockService<
     in_progress_tx: watch::Sender<Option<FlashBlockBuildInfo>>,
     /// `FlashBlock` service's metrics
     metrics: FlashBlockServiceMetrics,
-    /// Enable state root calculation from flashblock with index [`FB_STATE_ROOT_FROM_INDEX`]
-    compute_state_root: bool,
```
mattsse, line 64: could we keep this perhaps and enable when we have flashblocks consensus?
because for regular rpc pending tag support we dont need to waste time computing it

mattsse (changes requested):
last nit

crates/optimism/flashblocks/src/service.rs
0x00101010: great idea, done

mattsse (approved):

PR #19668, opened 2025-11-12
fix(tree-engine): also update in-memory tree when there are attributes in an unwind forkchoice

hai-rise:
This revisits this concern: https://github.com/paradigmxyz/reth/pull/17938#discussion_r2288587824, which is correct!

If we don't unwind the in-memory tree when there are attributes, the newly built payload won't even be added to the tree, due to:
https://github.com/paradigmxyz/reth/blob/01f497bc494203356f8891b2e9dab05282210b28/crates/engine/tree/src/tree/mod.rs#L1360-L1363

For instance, if we are at block 100, and there is a forkchoice that reorgs back to block 80 with attributes to build block 81 immediately, then currently the in-memory tree isn't updated, leading it to reject the new payload 81 because 81 < 100. Then a new forkchoice for block 81 will also be rejected due to:
https://github.com/paradigmxyz/reth/blob/01f497bc494203356f8891b2e9dab05282210b28/crates/engine/tree/src/tree/mod.rs#L659

This would freeze the Sequencer until either:
- A random forkchoice update without attributes arrives, or
- A new payload is called to force-insert payload 81 (which defeats the purpose of the `InsertExecutedBlock` optimisation).

This PR makes it so that the new payload 81 will be accepted immediately, and that the Sequencer can progress with a tight forkchoice-with-attributes loop even when there's a reorg.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1125,6 +1125,15 @@ where
             if self.engine_kind.is_opstack() ||
                 self.config.always_process_payload_attributes_on_canonical_head()
             {
+                // We need to effectively unwind the _canonical_ chain to the FCU's head, which is
+                // part of the canonical chain. We need to update the latest block state to reflect
+                // the canonical ancestor. This ensures that state providers and the transaction
+                // pool operate with the correct chain state after forkchoice update processing, and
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah yes, this should happen before we handle the attributes

PR #19670, opened 2025-11-12
feat: add --p2p-secret-key-hex which is similiar nodekeyhex in geth

cuiweixie:


crates/node/core/src/args/network.rs
```diff
@@ -81,9 +82,16 @@ pub struct NetworkArgs {
...
-    #[arg(long, value_name = "PATH")]
+    #[arg(long, value_name = "PATH", conflicts_with = "p2p_secret_key_hex")]
     pub p2p_secret_key: Option<PathBuf>,
 
+    /// Hex encoded secret key to use for this node.
+    ///
+    /// This will also deterministically set the peer ID. Cannot be used together with
+    /// `--p2p-secret-key`.
+    #[arg(long, value_name = "HEX", conflicts_with = "p2p_secret_key")]
+    pub p2p_secret_key_hex: Option<String>,
```
mattsse: I believe we can make this a B256 because we need a 32byte slice for the secret key

mattsse (changes requested):
makes sense, smol suggestion re type

crates/node/core/src/args/network.rs
cuiweixie: > I believe we can make this a B256 because we need a 32byte slice for the secret key

Done

mattsse (approved):
makes sense

mattsse:
@cuiweixie mind running `cargo +nightly fmt --all` on this

cant make edits because opened from okx fork

cuiweixie:
> @cuiweixie mind running `cargo +nightly fmt --all` on this
> 
> cant make edits because opened from okx fork

Done, pls trigger the workflows.

mattsse (approved):

mattsse:
now we need to run `make update-book-cli`

if you could open future prs from a non org fork, that'd be appreciated 🙏

PR #19672, opened 2025-11-12
fix(ethstats): prevent shutdown on read errors

Fibonacci747:
On any WebSocket read error, the read task in EthStatsService::run() broke and sent a shutdown signal, causing the whole service to exit instead of reconnecting. This contradicted the docstring promise to auto-reconnect on connection loss.
The reconnection branch used an async guard in tokio::select! (if self.conn.read().await.is_none()), which is invalid and prevented the reconnect ticker from functioning.
## solution
In the read-loop, distinguish errors:
- ConnectionError::Serialization(_): log and continue (keep the connection).
- Transport/close errors: log, take and close the connection, set self.conn to None, and continue so the reconnect ticker can re-establish the connection.
- Move the connection-state check inside the reconnection branch after reconnect_interval.tick() (no .await in the guard).

### DIFF
--- a/crates/node/ethstats/src/ethstats.rs
+++ b/crates/node/ethstats/src/ethstats.rs
@@ -558,24 +558,32 @@ where
 
         // Start the read loop in a separate task
         let read_handle = {
-            let conn = self.conn.clone();
+            let conn_arc = self.conn.clone();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this makes sense

PR #19673, opened 2025-11-12
perf(engine, trie): chunk based on idle workers

aleexeyy:
Addresses #19256

Chunking logic is now based on the amount of available idle workers, instead of hardcoded chunk size. I kept the hardcoded chunk size as the minimum chunk size that we can have, to avoid unnecessary chunking.

Benchmarks:

Screenshot 2025-11-12 at 12 19 16

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/chunking.rs
+++ b/crates/engine/tree/src/tree/payload_processor/chunking.rs
@@ -1,0 +1,80 @@
+//! Generic chunk planning utilities for multiproof scheduling in the engine.
+
+/// A computed plan for splitting `total` items into `chunks` parts.
+///
+/// Sizes are distributed as evenly as possible:
+/// - `base` items per chunk
+/// - the first `increased` chunks get one extra item (size `base + 1`)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) struct ChunkPlan {
+    /// Number of chunks to produce.
+    pub chunks: usize,
+    /// Base size for each chunk.
+    pub base: usize,
+    /// Number of leading chunks that receive an extra item.
+    pub increased: usize,
+}
+
+impl ChunkPlan {
+    /// Returns the size for the `index`-th chunk (0-based).
+    #[cfg(test)]
+    pub(super) fn chunk_size(&self, index: usize) -> usize {
+        self.base + usize::from(index < self.increased)
+    }
+
+    /// Returns an iterator of chunk sizes with length `self.chunks`.
+    #[cfg(test)]
+    pub(super) fn sizes(&self) -> impl Iterator<Item = usize> + '_ {
+        (0..self.chunks).map(|i| self.chunk_size(i))
+    }
+}
+
+/// Computes a chunk plan given the total amount of items, the number of idle workers
+/// available to process chunks concurrently, and a minimum chunk size.
+pub(super) fn compute_chunk_plan(total: usize, idle: usize, min_chunk: usize) -> Option<ChunkPlan> {
+    if idle == 0 || total == 0 {
+        return None;
+    }
+
+    let max_chunks_amount = total / min_chunk;
+    let chunks = max_chunks_amount.min(idle);
+
+    (chunks >= 2).then(|| {
+        let base = total / chunks;
+        let increased = total % chunks;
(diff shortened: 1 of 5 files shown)


yongkangc:
hey @aleexeyy sorry for dropping the ball on this. this looks promising and we would like to run some benchmarks on our infra. do you mind rebasing this?

gakonst:
@yongkangc this is the kind of thing where it's quicker for you to just rebase it and benchmark it -- it's a two file rebase

yongkangc:
Got it @gakonst rebasing and benching

crates/engine/tree/src/tree/payload_processor/chunking.rs
```diff
@@ -0,0 +1,80 @@
+//! Generic chunk planning utilities for multiproof scheduling in the engine.
+
+/// A computed plan for splitting `total` items into `chunks` parts.
+///
+/// Sizes are distributed as evenly as possible:
+/// - `base` items per chunk
+/// - the first `increased` chunks get one extra item (size `base + 1`)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) struct ChunkPlan {
```
yongkangc, line 9: instead of having this as a separate file, which creates complexity and additional code overhead to manage, i think it would be better to inline it in areas where we would need it.

crates/trie/common/src/proofs.rs
```diff
@@ -139,15 +146,23 @@ impl ChunkedMultiProofTargets {
...
 }
 
 impl Iterator for ChunkedMultiProofTargets {
     type Item = MultiProofTargets;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let chunk = self.flattened_targets.by_ref().take(self.size).fold(
+        let extra = usize::from(self.increased > 0);
+
+        let chunk = self.flattened_targets.by_ref().take(self.size + extra).fold(
```
yongkangc, line 165: I am slightly confused by this, what is the main goal for this?

crates/trie/common/src/proofs.rs
```diff
@@ -161,6 +176,9 @@ impl Iterator for ChunkedMultiProofTargets {
         if chunk.is_empty() {
             None
         } else {
+            if extra == 1 {
```
yongkangc, line 179: also here, what is the main logic for this? 

also would be great if u could include in the pr the overall high level logic of chunking based on idle worker

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -766,22 +782,30 @@ impl MultiProofTask {
...
-                chunks += 1;
+        let dispatched_chunks = if let Some(min_chunk) = self.min_chunk_size {
+            let total = proof_targets.chunking_length();
+            let idle = self.multiproof_manager.proof_worker_handle.available_account_workers();
+
+            if let Some(plan) = compute_chunk_plan(total, idle, min_chunk) {
+                let mut c = 0usize;
+                for proof_targets_chunk in proof_targets.chunks_with(plan.base, plan.increased) {
+                    dispatch(proof_targets_chunk);
+                    c += 1;
```
yongkangc, line 795: ideally we should not use c as the variable name - its quite confusing to the reader

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -766,22 +782,30 @@ impl MultiProofTask {
...
-        {
-            let mut chunks = 0usize;
-            for proof_targets_chunk in proof_targets.chunks(chunk_size) {
-                dispatch(proof_targets_chunk);
-                chunks += 1;
+        let dispatched_chunks = if let Some(min_chunk) = self.min_chunk_size {
+            let total = proof_targets.chunking_length();
+            let idle = self.multiproof_manager.proof_worker_handle.available_account_workers();
+
+            if let Some(plan) = compute_chunk_plan(total, idle, min_chunk) {
```
yongkangc, line 791: instead of declaring a struct, maybe we can inline it in multiproof.rs?

yongkangc (changes requested):
@aleexeyy good start. Overall, left some comments on nits, var names

The overall logic is unclear, so would like some clarification on that either comments etc

PR #19676, opened 2025-11-12
fix: correct argument order in save_diff calls for invalid block hooks

strmfos:
The save_diff function in witness.rs had its arguments backwards in 3 places.
The function signature expects (original, new) but we were passing (new, original).

This meant when debugging invalid blocks, the diff files would show changes going the wrong way - from the re-executed value back to the original, instead of from original to re-executed. Pretty confusing when you're trying to figure out what actually changed.

See lines 272-275 & 383: files named `.original.json` and `.re_executed.json` match their variables, but save_diff calls had them swapped.


mattsse (commented):
@shekhirin is this right?

shekhirin (approved):
thank you!

PR #19011, opened 2025-10-15
revert: "feat: Add building and publishing of *.deb packages (#18615)"

shekhirin:
`.deb` changes break the release CI https://github.com/paradigmxyz/reth/actions/runs/18514837580/job/52763051740

I think we should revert it and submit a separate PR that adds the `.deb` changes again, and test it before merging.

ref #18615

PR #19678, opened 2025-11-12
feat: Reproducible builds and *.deb packages

bakhtin:
Improved version of https://github.com/paradigmxyz/reth/pull/18615 (that got [reverted](https://github.com/paradigmxyz/reth/pull/19011))

Differences compared to the original PR:
- Does not modify `release.yml` CI pipeline to prevent any accidental breakage. Instead, I refactored `release-reproducible` pipeline to publish reproducible binaries and DEB packages
- Uses Docker image for building reproducible binaries. I figured that different builder configurations fail reproducibility

### jemalloc reproducibility
The [original `jemalloc`](https://github.com/jemalloc/jemalloc) which is a [vendored dependency](https://github.com/tikv/jemallocator/tree/main/jemalloc-sys) of `tikv/jemallocator` is not reproducible by default and requires a [bit of hacking](https://github.com/NixOS/nixpkgs/pull/393724) to achieve that. 

Both Nix and Debian figured that for us. So instead of patching the build scripts of Reth just to make jemalloc reproducible I compile Reth with the statically linked jemalloc taken from Debian repository. See `JEMALLOC_OVERRIDE` environment variable and `Dockerfile.reproducible` for exact implementation.

The override does not add a runtime dependency and only implemented for reproducible builds, original Reth binary compilation left intact.

When `tikv/jemallocator` builds jemalloc as a vendored dependency it adds a prefix to all `*alloc` syscalls (`--with-jemalloc-prefix=_rjem_`). Debian's jemalloc is built without this prefix. It breaks the compilation when using `JEMALLOC_OVERRIDE`.

To fix that I added a feature to Cargo.toml to allow building with `unprefixed_malloc_on_supported_platforms` (https://github.com/tikv/jemallocator/tree/main/jemalloc-sys#cargo-features). The side effect is that all C/C++ dependencies of Reth will start using jemalloc. I was running Reth binary with overridden jemalloc for some time now and didn't find it being an issue. 

### release-reproducible pipeline
I refactored `release-reproducible` pipeline to trigger after the `release` pipeline has successfully finished. The pipeline builds signed release binaries, DEB packages, and a production-ready Docker image. It attaches the binaries to the just published release draft (without updating the release description). Example of the pipeline run is [here](https://github.com/flashbots/reth/actions/runs/19295820197). Example of attached binaries is [here](https://github.com/flashbots/reth/releases/tag/untagged-f9ee532fffdc20cc8723)

### reproducible-build pipeline
I refactored `reproducible-build` pipeline that runs on cron every two days. The original implementation is [currently failing](https://github.com/paradigmxyz/reth/actions/runs/19251991336/job/55038680961) because the builds are in fact not reproducible.

My new implementation of reproducible builds addresses this. To add even more randomness to the build process I updated the pipeline to build on ubuntu-24 and ubuntu-22. An example of successfully running pipeline is [here](https://github.com/flashbots/reth/actions/runs/19274929321).


mattsse:
let's try this again @shekhirin?  :)

shekhirin (approved):
This looks good to me, even if something doesn't work, it won't block our release CI because reproducible builds live in separate workflows.

shekhirin:
@bakhtin there's some conflicts with main, can you please resolve them? I don't have access to push to your branch.

mattsse (approved):

PR #19679, opened 2025-11-12
feat(op-rpc): support batch rpc calls in historical forwarding

### LINKED ISSUE #19677: support batch rpc calls in historical forwarding
### Describe the feature

as described in #19509

we need to handle batch requests, which we currently don't do

https://github.com/paradigmxyz/reth/blob/d9537a416a1ba153f0a6286672e7d099986781b5/crates/optimism/rpc/src/historical.rs#L148-L150

here we need to scan the individual requests in that batch and whether we need to forward them, partition them and then combine them again.

if no requests in that batch need to be forwarded, we can simply call

```
 self.inner.batch(req)
```

### Additional context

_No response_

reject-i:
Closes #19677 closes #19509

### DIFF
--- a/crates/optimism/rpc/src/historical.rs
+++ b/crates/optimism/rpc/src/historical.rs
@@ -5,11 +5,13 @@ use alloy_eips::BlockId;
 use alloy_json_rpc::{RpcRecv, RpcSend};
 use alloy_primitives::{BlockNumber, B256};
 use alloy_rpc_client::RpcClient;
+use futures::{stream::FuturesOrdered, StreamExt};
+use jsonrpsee::{types::ErrorCode, BatchResponseBuilder};
 use jsonrpsee_core::{
-    middleware::{Batch, Notification, RpcServiceT},
+    middleware::{Batch, BatchEntry, Notification, RpcServiceT},
     server::MethodResponse,
 };
-use jsonrpsee_types::{Params, Request};
+use jsonrpsee_types::{ErrorObject, Id, Params, Request};
 use reth_storage_api::{BlockReaderIdExt, TransactionsProvider};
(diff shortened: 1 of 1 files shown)


crates/optimism/rpc/src/historical.rs
```diff
@@ -171,6 +266,17 @@ impl<P> HistoricalRpcInner<P>
...
+    /// Checks if a request should be forwarded to the historical endpoint (synchronous check).
+    fn should_forward_request(&self, req: &Request<'_>) -> bool {
+        match req.method_name() {
+            "debug_traceTransaction" |
+            "eth_getTransactionByHash" |
+            "eth_getTransactionReceipt" |
+            "eth_getRawTransactionByHash" => self.should_forward_transaction(req),
+            method => self.should_forward_block_request(method, req),
+        }
+    }
```
mattsse, line 246: this is the same as maybe_forward_request

can we reuse the `should_forward` checks in maybe_forward_request

crates/optimism/rpc/src/historical.rs
```diff
@@ -145,8 +153,95 @@ where
...
+                    needs_forwarding = true;
+                    break;
+                }
+            }
+
+            if !needs_forwarding {
+                return inner_service.batch(req).await;
+            }
+
+            let entries: Vec<_> = req.into_iter().collect();
```
mattsse: maybe we can simplify this a bit and simply flatten all requests here and use self.call ?

mattsse (changes requested):
thanks, that was fast

this is a great start but I feel like we can simplify the batch handling a bit

mattsse (approved):
ty, turns out we can reuse and copy paste the default impl for batch handling here

PR #19680, opened 2025-11-12
feat(chains): add support for custom/ephemeral chains

### LINKED ISSUE #19635: Add `EthChainSpec::name` method
### Describe the feature

Right now when resolving datadir, we will only include chain name in the directory for chains that are added to alloy-chains enum, if it's not present, just chain id will be included

https://github.com/paradigmxyz/reth/blob/7997cd42832f4be41e0a20e5796f8dc194b7b137/crates/node/core/src/dirs.rs#L165-L169
https://github.com/paradigmxyz/reth/blob/7997cd42832f4be41e0a20e5796f8dc194b7b137/crates/node/core/src/dirs.rs#L12-L16

To support custom/ephemeral chains to customize this we can add something like this to the `EthChainSpec` trait
```rust
fn name(&self) -> String {
    self.chain().to_string()
}
```

And propagate it all the way to `platform_path_from_chain` from all callsites

We would want to change `resolve_datadir` to take `impl ChainSpec` and make use of the `name` method
https://github.com/paradigmxyz/reth/blob/7997cd42832f4be41e0a20e5796f8dc194b7b137/crates/node/core/src/args/datadir_args.rs#L33-L37

### Additional context

_No response_

aleexeyy:
### **Title**

Add spec-aware datadir naming via `EthChainSpec::name()`

### **Summary**

This PR introduces spec-aware datadir naming to support custom and ephemeral chains while preserving existing behavior for built-in chains.

* Adds `EthChainSpec::name()` with a default of `chain().to_string()` for human-friendly, stable directory naming.
* Updates `resolve_datadir` and related helpers to use `EthChainSpec` instead of plain `Chain`.
* Introduces `resolve_datadir_for_chain` for compatibility with existing chain-only call sites.
* Propagates spec-based naming across key modules and tests.

### **Why**

Allows custom or ephemeral networks not in `alloy-chains` to define meaningful datadir folder names instead of numeric chain IDs, improving clarity and usability.

### **Behavior**

* Built-in chains remain unchanged.
* Custom chains can override `name()` for stable folder naming.

### **Linked Issue**

Closes #19635

### DIFF
--- a/bin/reth-bench-compare/src/cli.rs
+++ b/bin/reth-bench-compare/src/cli.rs
@@ -231,8 +231,8 @@ impl Args {
                 PathBuf::from(expanded.as_ref())
             }
             None => {
-                // Use the same logic as reth: <datadir>/<chain>/jwt.hex
-                let chain_path = self.datadir.clone().resolve_datadir(self.chain);
+                // Use chain-id/name based path (bench tool only has `Chain`)
+                let chain_path = self.datadir.clone().resolve_datadir_for_chain(self.chain);
                 chain_path.jwt()
             }
         }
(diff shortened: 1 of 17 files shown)


mattsse (commented):
dup of [19692](https://github.com/paradigmxyz/reth/pull/19692)

PR #19686, opened 2025-11-12
feat: add --netrestrict to node cmd like in geth

cuiweixie:


crates/net/banlist/Cargo.toml
```diff
@@ -14,3 +14,6 @@ workspace = true
 [dependencies]
 # ethereum
 alloy-primitives.workspace = true
+
+# networking
+ipnet = "2.11"
```
mattsse: this should use from workspace
```suggestion
# networking
ipnet.workspace = true
```

crates/net/banlist/src/lib.rs
```diff
@@ -215,3 +215,163 @@ mod tests {
         assert!(!banlist.is_banned_ip(&ip));
     }
 }
+
+/// IP filter for restricting network communication to specific IP ranges using CIDR notation.
+///
+/// This is similar to geth's `--netrestrict` flag.
```
mattsse: I think we can omit these docs

crates/node/core/src/args/network.rs
```diff
@@ -205,6 +206,16 @@ pub struct NetworkArgs {
     /// Optional network ID to override the chain specification's network ID for P2P connections
     #[arg(long)]
     pub network_id: Option<u64>,
+
+    /// Restrict network communication to the given IP networks (CIDR masks).
+    ///
+    /// Comma separated list of CIDR network specifications.
+    /// Only peers with IP addresses within these ranges will be allowed to connect.
+    /// Similar to geth's --netrestrict flag.
```
mattsse:
```suggestion
```

crates/node/core/src/args/network.rs
```diff
@@ -381,6 +394,17 @@ impl NetworkArgs {
             get_secret_key(&secret_key_path)
         }
     }
+
+    /// Creates an IP filter from the netrestrict argument.
+    ///
+    /// Returns an error if the CIDR format is invalid.
+    pub fn ip_filter(&self) -> Result<IpFilter, ipnet::AddrParseError> {
+        if let Some(ref netrestrict) = self.netrestrict {
```
mattsse:
```suggestion
        if let Some(netrestrict) = &self.netrestrict {
```

mattsse (changes requested):
this looks very clean

thank you, this is indeed nice to have.

smol pedantic nits

mattsse (approved):
ref

https://github.com/ethereum/go-ethereum/blob/15ff378a8927eed211589bcf375aa5c528209b71/cmd/utils/flags.go#L871-L875

checks out, ty!

PR #19687, opened 2025-11-12
feat(trie): Implement skeleton of proof_v2

mediocregopher:
Towards #19512

This implements the skeleton of the new proof calculator rewrite. No actual logic is implemented yet, this only sets up most of the new types which will be involved.

* RevealedSparseNode is renamed to SparseTrieNode and moved to reth-trie-common. This is the type which will be returned from the calculator for proofs (letting avoid a translation step during sparse trie revealing). The rename helps to denote that this type is no longer just for revealing.

* ValueEncoder is defined. This is the primary mechanism by which we deal with the differences between storage and account tries, and which allows us to inject behavior in the future like dispatching storage root calculation to other threads for account proofs.

  * ValueEncoder::Value - Either Account or U256 for account and storage tries, respectively. This is the value returned from the DB.

  * ValueEncoder::Fut - A future-like type which will be called-upon later to encode the Value into its RLP form. For storage tries (U256) this is trivial. For account tries we need some mechanism to obtain the storage root of the account. A default `SyncAccountValueEncoder` is provided which synchronously computes the storage root when the future is invoked, but in later PRs we can to proof workers, add in caching, etc...

* ProofCalculator is where the actual logic of calculating proofs is going to live. For the moment it is un-implemented.


crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+    type Fut: ValueEncoderFut;
+
+    /// Returns a future-like value that will encode the value when called.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The key the value was stored at in the DB
+    /// * `value` - The value to encode
+    ///
+    /// The returned future will be called as late as possible in the algorithm to maximize
```
shekhirin: I find this a bit confusing, as the return type isn't really a future

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+
+                // Extract the hash - if it's a short node, it won't have a hash, so compute one
+                rlp_node.as_hash().unwrap_or_else(|| alloy_primitives::keccak256(&*buf))
+            })?;
+
+        // Combine account with storage root to create TrieAccount
+        let trie_account = self.account.into_trie_account(storage_root);
+
+        // Encode the trie account
+        buf.clear();
```
shekhirin: unsure if the `encode` implementation should be responsible for clearing the output buffer

also, I see that we clear it above in `map()`, is it correct?

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+use alloy_primitives::{B256, U256};
+use alloy_rlp::Encodable;
+use reth_execution_errors::trie::StateProofError;
+use reth_primitives_traits::Account;
+use reth_trie_common::RlpNode;
+use std::rc::Rc;
+
+/// A trait for deferred encoding of proof values.
+///
+/// This trait is implemented by types that can encode themselves into a buffer
```
yongkangc: what does encode themselves into a buffer on demand mean? this part slightly confused me

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/// A trait for encoding values for proof calculation.
+///
+/// This trait allows for lazy evaluation of encoding, enabling deferred
+/// computation of storage roots and other expensive operations.
+///
+/// The encoder takes a reference to itself and a value, returning a future-like
+/// type that will perform the encoding when needed.
+pub trait ValueEncoder {
```
yongkangc: non blocking, but took me a few looks to understand ValueEncoder. 

do you think `ProofValueEncoder` might be better?

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+                let root_node =
+                    nodes.first().expect("storage_proof always returns at least the root");
+                buf.clear();
+                root_node.node.encode(buf);
+
+                // Hash the encoded node (RlpNode::from_rlp handles hashing for nodes >= 32 bytes)
+                let rlp_node = RlpNode::from_rlp(buf);
+
+                // Extract the hash - if it's a short node, it won't have a hash, so compute one
+                rlp_node.as_hash().unwrap_or_else(|| alloy_primitives::keccak256(&*buf))
```
yongkangc: could we do something like `buf.as_slice()` for this?

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+    provider: Rc<P>,
+    hashed_address: B256,
+    account: Account,
+}
+
+impl<P> ValueEncoderFut for SyncAccountValueEncoderFut<P>
+where
+    P: TrieCursorFactory + HashedCursorFactory,
+{
+    fn encode(self, buf: &mut Vec<u8>) -> Result<(), StateProofError> {
```
yongkangc, line 108: does it makes sense to add docstring for what this impl does? 

i found it harder to understand from just the trait docstring

maybe something like 

```
// Computes the storage root for this account and RLP-encodes the resulting `TrieAccount` into `buf`
```

yongkangc (approved):
i left some comments. the only part that was confusing to me was the impl encode function and why we need to do that / what it needs to replace. 

i am guessing that it replaces https://github.com/paradigmxyz/reth/blob/3f1a7b37e91c0e6509a3145d9e4ac69302cf2d12/crates/trie/trie/src/proof/mod.rs#L118 where given an Account + hashed_address, we compute the right storage_root, build a TrieAccount, and write its RLP bytes into buf so the proof can include the correct account leaf.


other than these, lgtm and left some nits that are none blocking

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: It's not a literal `std::future::Future`, but it is a future in the more abstract sense that it's a handle on some asynchronous work which can be happening in the background. I had also considered the term "Promise" but that's more javascript-y. Open to other suggestions though...

yongkangc: i have the same thoughts as this, maybe i am missing something?

crates/trie/common/src/trie.rs
```diff
@@ -0,0 +1,45 @@
...
+impl TrieMasks {
+    /// Helper function, returns both fields `hash_mask` and `tree_mask` as [`None`]
+    pub const fn none() -> Self {
+        Self { hash_mask: None, tree_mask: None }
+    }
+}
+
+/// Carries all information needed by a sparse trie to reveal a particular node.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SparseTrieNode {
```
shekhirin: I find the rename to `SparseTrieNode` a bit confusing because it's returned from non sparse trie specific `proof_v2` module. Should it just be `TrieNode`?

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Returns a future-like value that will encode the value when called.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The key the value was stored at in the DB
+    /// * `value` - The value to encode
+    ///
+    /// The returned future will be called as late as possible in the algorithm to maximize
+    /// the time available for parallel computation (e.g., storage root calculation).
+    fn encoder_fut(&self, key: B256, value: Self::Value) -> Self::Fut;
```
mattsse: I think we can also do `-> impl ValueEncoderFut` unless we need named types for this

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+mod value;
+pub use value::*;
+
+/// A proof calculator that generates merkle proofs using only leaf data.
+///
+/// The calculator:
+/// - Accepts one or more B256 proof targets sorted lexicographically
+/// - Returns proof nodes sorted lexicographically by path
+/// - Returns only the root when given zero targets
```
shekhirin: would it make more sense semantically to return root if `Nibbles::default()` (`0x`) is provided, and nothing if no targets are provided?

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,158 @@
...
+/// computation of storage roots and other expensive operations.
+///
+/// The encoder takes a reference to itself and a value, returning a future-like
+/// type that will perform the encoding when needed.
+pub trait ValueEncoder {
+    /// The type of value being encoded (e.g., U256 for storage, Account for accounts).
+    type Value;
+
+    /// The type that will compute and encode the value when needed.
+    type Fut: ValueEncoderFut;
```
mattsse: this pattern makes a lot of sense, but I can see how `Fut` could be confusing here, maybe we can think about an alternative here, like `ValueEncoder -> Encodable` or deferred/lazy perhaps?

```
 pub trait DeferredValueEncoder {
      fn encode(self, buf: &mut Vec<u8>) -> Result<(), StateProofError>;
  }

  pub trait ValueEncoder {
      type Value;
      type DeferredEncoder: DeferredValueEncoder;
      fn deferred_encoder(&self, key: B256, value: Self::Value) -> Self::DeferredEncoder;
  }

```

mattsse (approved):
this was quite easy to follow, all makes sense.

only have some suggestions re `Fut`, but also something we could change later

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: Named types will be useful for future PRs, because we're going to be storing the "future" in an enum.

crates/trie/common/src/trie.rs
mediocregopher: I avoided `TrieNode` because that's already a type in alloy-trie that gets re-exported from reth-trie-common, and it's one of the fields on this struct. Wdyt about `ProofTrieNode`? We already have ProofTrieNodeProviderFactory elsewhere so I think it makes some sense

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: I removed one of the calls to `clear`, we can make sure in the ProofCalculator implementation to only pass cleared bufs into `encode`.

It is correct that we are `clear`ing after the encoding in the `map`. Within the `map` we are RLP-encoding the root node of the storage trie, which we then hash to get the storage root. We then re-use the buffer later to RLP encode the TrieAccount, which contains the storage root as a field.

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: Renamed according to matt's suggestion below

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: Removed this comment and expanded the comment on the `ValueEncoder` instead.

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: I'm open to renaming it, but tbh I don't think `ProofValueEncoder` adds any extra context that helps understand it. Maybe `LeafValueEncoder`?

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: Good catch, fixed this and generally cleaned up this area

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: Nice, implemented this :100:

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Yeah this makes sense, changed `targets` to be `IntoIterator<Item = Nibbles>` and updated docs

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -0,0 +1,161 @@
...
+/// initiate any asynchronous computation of the account's storage root we want to do. Later we call
+/// [`DeferredValueEncoder::encode`] to obtain the result of that computation and RLP-encode it.
+pub trait ValueEncoder {
+    /// The type of value being encoded (e.g., U256 for storage, Account for accounts).
+    type Value;
+
+    /// The type that will compute and encode the value when needed.
+    type DeferredEncoder: DeferredValueEncoder;
+
+    /// Returns a future-like value that will RLP-encode the value when called.
```
shekhirin: residue of future terminology

crates/trie/common/src/trie.rs
shekhirin: sgtm!

shekhirin (approved):

crates/trie/trie/src/proof_v2/value.rs
yongkangc: thats good `LeafValueEncoder`

yongkangc (approved):
lgtm

PR #19688, opened 2025-11-12
fix(docker): run container as non-root user

maximevtush:
Run the Reth node container as a non-root user to follow security best practices and reduce attack surface.


mattsse (commented):
i dont want to change this

PR #19691, opened 2025-11-12
perf(ress): avoid unnecessary clone in on_block_bodies_request

VolodymyrBg:
This change removes the redundant clone in on_block_bodies_request by passing the owned Vec<B256> directly to provider.block_bodies, which already accepts ownership. This eliminates an avoidable allocation and copy on potentially large vectors without altering behavior. Since the request vector is moved into the provider call, the trace logging was adjusted to log only the request length rather than the moved value. No public API or protocol logic is affected, and the change is consistent with the provider’s signature and usage patterns elsewhere.


mattsse (commented):
this is just a vec of hashes so this is fine to keep

PR #19692, opened 2025-11-12
feat: add `EthChainSpec::name` method

### LINKED ISSUE #19635: Add `EthChainSpec::name` method

Peponks9:
`resolve_datadir` in `datadir_args.rs` now takes `chainspec: impl EthChainSpec` instead of `chain: Chain`

closes #19635
cc @klkvr

### DIFF
--- a/bin/reth-bench-compare/src/cli.rs
+++ b/bin/reth-bench-compare/src/cli.rs
@@ -1,9 +1,10 @@
 //! CLI argument parsing and main command orchestration.
 
+use alloy_consensus::Header;
 use alloy_provider::{Provider, ProviderBuilder};
 use clap::Parser;
 use eyre::{eyre, Result, WrapErr};
-use reth_chainspec::Chain;
+use reth_chainspec::{Chain, EthChainSpec, HOLESKY, MAINNET, SEPOLIA};
 use reth_cli_runner::CliContext;
 use reth_node_core::args::{DatadirArgs, LogArgs, TraceArgs};
 use reth_tracing::FileWorkerGuard;
@@ -47,6 +48,18 @@ impl std::fmt::Display for DisableStartupSyncStateIdle {
             Self::Feature => write!(f, "feature"),
             Self::All => write!(f, "all"),
         }
+    }
+}
+
+fn get_chainspec(chain: Chain) -> &'static dyn EthChainSpec<Header = Header> {
+    if chain == Chain::mainnet() {
+        &*MAINNET
+    } else if chain == Chain::sepolia() {
+        &*SEPOLIA
+    } else if chain == Chain::holesky() {
+        &*HOLESKY
+    } else {
+        panic!("unsupported chain for bench-compare");
     }
 }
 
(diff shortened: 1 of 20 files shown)


bin/reth-bench-compare/src/cli.rs
```diff
@@ -50,6 +51,18 @@ impl std::fmt::Display for DisableStartupSyncStateIdle {
...
+    if chain == Chain::mainnet() {
+        &*MAINNET
+    } else if chain == Chain::sepolia() {
+        &*SEPOLIA
+    } else if chain == Chain::holesky() {
+        &*HOLESKY
+    } else {
+        panic!("unsupported chain for bench-compare");
+    }
+}
```
klkvr: hmm this is not ideal because we also need it to work for base chains

bin/reth-bench-compare/src/cli.rs
```diff
@@ -232,15 +245,15 @@ impl Args {
             }
             None => {
                 // Use the same logic as reth: <datadir>/<chain>/jwt.hex
-                let chain_path = self.datadir.clone().resolve_datadir(self.chain);
+                let chain_path = self.datadir.clone().resolve_datadir(get_chainspec(self.chain));
```
klkvr: maybe one way to work around this is to make `resolve_datadir`/`unwrap_or_chain_default` take `String` so that any `chain` arg value works here

klkvr (commented):
this lgtm overall but ideally we don't break reth-bench-compare

Peponks9 (commented):
thanks for the feeback @klkvr, will work based on this

bin/reth-bench-compare/src/main.rs
```diff
@@ -14,6 +14,9 @@
 #[global_allocator]
 static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
 
+// Suppress unused crate dependency warning
+use alloy_consensus as _;
+
```
klkvr: let's remove the dependency instead of suppresing the warning?

mattsse (commented):
since this a bit more invasive than expected and we can easily update chains, dont think this is that useful anymore

PR #19695, opened 2025-11-12
examples: remove unused imports from txpool tracing submit

Doryu001:
remove unused TransactionBuilder, PoolTransaction, and SignerRecoverable imports from examples/txpool-tracing/src/submit.rs silence compiler warnings and keep the example free of redundant dependencies


mattsse (commented):
this example is fine as is

PR #19696, opened 2025-11-12
feat: make PayloadBuilderArgs defaults customizable

### LINKED ISSUE #19609: Make PayloadBuilderArgs defaults customizable
### Describe the feature

similar to 

https://github.com/paradigmxyz/reth/pull/19437

but for 

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/node/core/src/args/payload_builder.rs#L14-L14

## TODO

* add a global static that holds the default values, that can be set
* update defaults in the clap args

### Additional context

_No response_

0xChaddB:
Make PayloadBuilderArgs defaults customizable, similar to #19437.

Closes #19609

### DIFF
--- a/crates/node/core/src/args/payload_builder.rs
+++ b/crates/node/core/src/args/payload_builder.rs
@@ -1,11 +1,13 @@
 use crate::{cli::config::PayloadBuilderConfig, version::default_extra_data};
 use alloy_consensus::constants::MAXIMUM_EXTRA_DATA_SIZE;
-use alloy_eips::merge::SLOT_DURATION;
 use clap::{
     builder::{RangedU64ValueParser, TypedValueParser},
     Arg, Args, Command,
 };
-use reth_cli_util::{parse_duration_from_secs, parse_duration_from_secs_or_ms};
+use reth_cli_util::{
+    parse_duration_from_secs, parse_duration_from_secs_or_ms,
+    parsers::format_duration_as_secs_or_ms,
+};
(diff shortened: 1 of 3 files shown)


mattsse (approved):
cool,

this lgtm

pending @klkvr for sanity check

crates/node/core/src/args/payload_builder.rs
```diff
@@ -6,14 +6,88 @@ use clap::{
...
+///
+/// Global defaults can be set via [`DefaultPayloadBuilderValues::try_init`].
+#[derive(Debug, Clone)]
+pub struct DefaultPayloadBuilderValues {
+    /// Default extra data for blocks
+    pub extra_data: String,
+    /// Default interval between payload builds in seconds
+    pub interval_secs: u64,
+    /// Default deadline for payload builds in seconds
+    pub deadline_secs: u64,
```
klkvr: I think we want more precision here than full seconds, can we have `Duration` here?

klkvr: also those need to be propagated as `default_value_t`

klkvr (changes requested):

crates/node/core/src/args/payload_builder.rs
mattsse: ah right, yeah we should change this to just duration

mattsse (commented):
this seems fine

mattsse (approved):

klkvr (approved):

PR #19697, opened 2025-11-12
fix(banlist): check ban expiration in is_banned methods

DeVikingMark:
Methods is_banned_ip and is_banned_peer now verify expiration time 
synchronously instead of relying solely on periodic evict calls.

This prevents false positives where expired bans are still considered
active between evict intervals.


mattsse (commented):
these should not check expiration because this is handled cheaper once for all in evict

PR #19699, opened 2025-11-12
chore: remove redundant clone in exex subscription RPC setup

0xLogicalx:
pass subscriptions_tx directly into StorageWatcherRpc::new instead of cloning behavior unchanged; avoids an extra UnboundedSender clone and keeps the example lean


mattsse (approved):
since this removes the allow lint entire, I allow this

PR #19700, opened 2025-11-13
perf: minimum of 32 proof workers for parallel state root computation

yongkangc:
## Summary

Changes the default worker count calculation to enforce a **minimum of 32 workers** instead of a maximum.

**Before**: `min(cpu_count * 2, 32)` - capped at 32 maximum
**After**: `max(cpu_count * 2, 32)` - enforced at 32 minimum

### Motivation

payload latency seems to have been caused by
- state root latency increase due to
- proof calculation increase ->
- total storage nodes increase (xen)
- pending multiproof request increase but active workers remaining the same
i.e pending multiproof request increase but active workers remaining the same -> more storage nodes but the proofs request are not being processed

That means that we are lacking in workers to process proofs and thats where the bottleneck is 
image

Previously testing on base, 32 was a point, and anything less could cause regression. So its a good sweet spot.

**Expected Impact**
This would reduce some back pressure when processing proofs especailly for xen. 

In the future we will dynamically increase / decrease workers based on available workers.

### DIFF
--- a/crates/engine/primitives/src/config.rs
+++ b/crates/engine/primitives/src/config.rs
@@ -10,14 +10,26 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
 pub const MIN_WORKER_COUNT: usize = 32;
 
 /// Returns the default number of storage worker threads based on available parallelism.
+///
+/// Uses the higher of:
+/// - 2x CPU thread count
+/// - 32 workers (minimum)
+///
+/// The minimum of 32 workers is based on performance testing showing that lower counts
+/// are insufficient for chains like Base, which require higher parallelism for proof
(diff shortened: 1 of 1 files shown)


crates/engine/primitives/src/config.rs
```diff
@@ -10,14 +10,26 @@ pub const DEFAULT_MEMORY_BLOCK_BUFFER_TARGET: u64 = 0;
 pub const MIN_WORKER_COUNT: usize = 32;
 
 /// Returns the default number of storage worker threads based on available parallelism.
```
theomart77, line 12: crates/engine/primitives/src/config.rs

gakonst:
what's the impact on e2e nP time?

mediocregopher (approved):

PR #19704, opened 2025-11-13
fix(db-api): resolve test compilation errors

davidjsonn:
Fixes compilation errors in reth-db-api tests by:

- Add alloy-primitives with "rand" feature to dev-dependencies to enable StorageKey::random() method in tests
- Implement manual Arbitrary trait for CompactClientVersion wrapper struct since ClientVersion has conditional Arbitrary implementation that canno be auto-derived through the wrapper macro

The errors were not caught by CI because unit tests in #[cfg(test)] modules are excluded from the test run configuration.


mattsse (approved):
we only need to fix the features here

PR #19708, opened 2025-11-13
chore: remove redundant eth/68 NewPooledTransactionHashes length validation

Fibonacci747:
This change removes a redundant and unreachable length validation for eth/68 NewPooledTransactionHashes in ActiveSession::on_incoming_message. Incoming messages are decoded via ProtocolMessage::decode_message which dispatches to NewPooledTransactionHashes68::decode; the decoder already enforces that hashes, types, and sizes vectors have equal lengths and returns an InvalidMessage error on mismatch. Because the stream yields either a decoded EthMessage or an EthStreamError, malformed payloads never reach on_incoming_message and are handled earlier by closing the connection. Removing the duplicate check simplifies the code path and avoids unnecessary work without changing behavior on invalid inputs.


mattsse (approved):
this makes sense, consumers are also checking this again

PR #19709, opened 2025-11-13
chore: bump revm-interpreter

klkvr:
we have 2 instances of the crate in depgraph rn


mattsse (approved):
maybe we should try to get rid of this direct import

PR #19710, opened 2025-11-13
fix: correct db access receipts example

zeroprooff:
Fixed the receipts_provider_example to use receipts_by_block instead of receipt. The previous code was incorrectly passing a block number to receipt(), which expects a transaction index, causing it to return None. Now it properly fetches all receipts for the block and iterates through their logs.


mattsse (commented):
this example is fine as is

PR #19713, opened 2025-11-13
Remove unused imports from txpool tracing example

0xLogicalx:
drop the module-level suppressor for unused items remove the dead `PoolTransaction` and `SignerRecoverable` imports


mattsse (commented):
we need to keep this

PR #19714, opened 2025-11-13
docs: fixed dead Sentry link

youyyytrok:
Hi! I fixed broken link in docs
image


mattsse (approved):
yeah, this one we should actually update

PR #19723, opened 2025-11-13
fix: ensure receipt consistency check is done over block number

joshieDo:
With https://github.com/paradigmxyz/reth/pull/19399, we might not write receipts on a static file during pipeline sync, even though we are incrementing the segment block number.

This lead to failing the second run of `ExecutionStage::ensure_consistency`, since the consistency check was being done over `TxNumber` instead of `BlockNumber`.

### DIFF
--- a/crates/stages/stages/src/stages/execution.rs
+++ b/crates/stages/stages/src/stages/execution.rs
@@ -233,11 +233,10 @@ where
             Ordering::Less => {
                 // If we are already in the process of unwind, this might be fine because we will
                 // fix the inconsistency right away.
-                if let Some(unwind_to) = unwind_to {
-                    if unwind_to <= static_file_block_num {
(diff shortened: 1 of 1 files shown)


crates/stages/stages/src/stages/execution.rs
```diff
@@ -209,36 +209,38 @@ where
             .map(|num| num + 1)
             .unwrap_or(0);
 
+        // Get highest block number in static files for receipts
+        let static_file_block_num = static_file_provider
+            .get_highest_static_file_block(StaticFileSegment::Receipts)
+            .unwrap_or(0);
```
mattsse, line 215: this makes to me

mattsse (approved):
lgtm

shekhirin (approved):

PR #19725, opened 2025-11-13
docs(revm): document lifetime requirements for CachedReads wrappers

forkfury:
Adds documentation explaining the lifetime parameter 'a in CachedReadsDbMut and CachedReadsDBRef. Clarifies that the lifetime ties these wrappers to the underlying CachedReads instance, ensuring memory safety. Also documents the use of RefCell for interior mutability in CachedReadsDBRef.


mattsse (approved):
these are useful, so I allow it

PR #19727, opened 2025-11-13
fix: ensure open db tx is dropped

mattsse:
this previously didnt drop the opened db tx

this also resets the executor every 2mins

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -64,7 +64,8 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
         let components = components(provider_factory.chain_spec());
 
         let min_block = self.from;
-        let best_block = DatabaseProviderFactory::database_provider_ro(&provider_factory)?.best_block_number()?;
+        let best_block = DatabaseProviderFactory::database_provider_ro(&provider_factory)?
+            .best_block_number()?;
         let mut max_block = best_block;
         if let Some(to) = self.to {
             if to > best_block {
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/re_execute.rs
```diff
@@ -166,9 +169,12 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
                     }
                     let _ = stats_tx.send(block.gas_used());
 
-                    // Reset DB once in a while to avoid OOM
-                    if executor.size_hint() > 1_000_000 {
+                    // Reset DB once in a while to avoid OOM or read tx timeouts
+                    if executor.size_hint() > 1_000_000 ||
```
klkvr, line 173: ah for some reason I was thinking this `1_000_000` would be enough back then

klkvr (approved):

crates/cli/commands/src/re_execute.rs
```diff
@@ -61,11 +61,11 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
     {
         let Environment { provider_factory, .. } = self.env.init::<N>(AccessRights::RO)?;
 
-        let provider = provider_factory.database_provider_ro()?;
         let components = components(provider_factory.chain_spec());
 
         let min_block = self.from;
-        let best_block = provider.best_block_number()?;
+        let best_block = DatabaseProviderFactory::database_provider_ro(&provider_factory)?
```
klkvr, line 67: oh this is the primary fix, makes sense

PR #19729, opened 2025-11-13
fix(stateless): export stateless_validation function

oooLowNeoNooo:
Export stateless_validation at crate level to match module documentation.

The documentation states that stateless_validation is the primary entry point, but it was only accessible via validation::stateless_validation. Now both stateless_validation and stateless_validation_with_trie are exported at the crate root for consistency.


mattsse (approved):
makes sense

PR #19731, opened 2025-11-13
docs: remove obsolete max_changesets documentation

avorylli:
Remove outdated reference to max_changesets field from ExecutionStageThresholds docs.

The struct only has four thresholds: max_blocks, max_changes, max_cumulative_gas, and max_duration.

### DIFF
--- a/crates/stages/types/src/execution.rs
+++ b/crates/stages/types/src/execution.rs
@@ -2,11 +2,8 @@ use core::time::Duration;
 
 /// The thresholds at which the execution stage writes state changes to the database.
 ///
-/// If either of the thresholds (`max_blocks` and `max_changes`) are hit, then the execution stage
-/// commits all pending changes to the database.
-///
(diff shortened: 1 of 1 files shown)


crates/stages/types/src/execution.rs
```diff
@@ -2,11 +2,8 @@ use core::time::Duration;
 
 /// The thresholds at which the execution stage writes state changes to the database.
 ///
-/// If either of the thresholds (`max_blocks` and `max_changes`) are hit, then the execution stage
-/// commits all pending changes to the database.
-///
-/// A third threshold, `max_changesets`, can be set to periodically write changesets to the
-/// current database transaction, which frees up memory.
```
mattsse, line 9: these are indeed outdated,

instead of shrinking the docs, can we update them so that the docs mention max_cumulative_gas and max_duration

mattsse (changes requested):

mattsse (approved):

PR #19734, opened 2025-11-13
perf: Optimize BatchTxProcessor buffer reuse to avoid repeated allocations

Snezhkko:
Initialize the batcher’s receive buffer with max_batch_size and preserve its capacity across batches. Instead of taking and resetting the buffer each iteration, drain its contents into a temporary batch vector and spawn processing. This keeps the hot poll_recv_many path from reallocating repeatedly and aligns with high-throughput batching goals. Removed the ineffective reserve(1) which discouraged capacity reuse.


mattsse (commented):
I don't think we should do this because this now requires an alloc right before we submit the batch

PR #19737, opened 2025-11-14
feat(bench-compare): add configurable OTLP trace queue size

yongkangc:
Motivation: trace loss issues observed where blocks were missing traces even for 10-30 block ranges.

**Summary**
- Fixes trace drops during benchmarks by setting `OTEL_BSP_MAX_QUEUE_SIZE` for spans as defaults (previously used `OTEL_BLRP_MAX_QUEUE_SIZE` which only affects logs).
- Adds `--otlp-max-queue-size` CLI flag with default 32768 to prevent span drops during intensive block processing.

### DIFF
--- a/bin/reth-bench-compare/src/node.rs
+++ b/bin/reth-bench-compare/src/node.rs
@@ -30,6 +30,7 @@ pub(crate) struct NodeManager {
     additional_reth_args: Vec<String>,
     comparison_dir: Option<PathBuf>,
     tracing_endpoint: Option<String>,
+    otlp_max_queue_size: usize,
 }
 
 impl NodeManager {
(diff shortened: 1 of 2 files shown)


mediocregopher (approved):

bin/reth-bench-compare/src/node.rs
```diff
@@ -267,8 +261,9 @@ impl NodeManager {
 
         // Set high queue size to prevent trace dropping during benchmarks
         if self.tracing_endpoint.is_some() {
-            cmd.env("OTEL_BLRP_MAX_QUEUE_SIZE", "10000");
-            // Set service name to differentiate baseline vs feature runs in Jaeger
+            cmd.env("OTEL_BSP_MAX_QUEUE_SIZE", self.otlp_max_queue_size.to_string()); // Traces
+            cmd.env("OTEL_BLRP_MAX_QUEUE_SIZE", "10000"); // Logs
+                                                          // Set service name to differentiate baseline vs feature runs in Jaeger
             cmd.env("OTEL_SERVICE_NAME", format!("reth-{}", ref_type));
```
shekhirin:
```suggestion
            // Set service name to differentiate baseline vs feature runs in Jaeger
            cmd.env("OTEL_SERVICE_NAME", format!("reth-{}", ref_type));
```

shekhirin (approved):

Issue #19705, opened 2025-11-13
Improve tx manager hash packing


mattsse:
### Describe the feature

this is currently enforced by a duration budget

https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/net/network/src/transactions/mod.rs#L1629-L1640

and always allocates some capacity

https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/net/network/src/transactions/fetcher.rs#L422-L429

all of this is a bit hard to read, and a bit suboptimal in general

https://github.com/paradigmxyz/reth/blob/7a599dc13018acc7f7d206581bf4cd4eb6368091/crates/net/network/src/transactions/fetcher.rs#L78-L84

## TODO
* create a more efficient implementation of txhash fetching
* all the packing logic should be testable in isolation

@0xKarl98

NikhilSharmaWe:
@mattsse is this still needed?

PR #19738, opened 2025-11-14
feat: Improve tx manager hash packing

0xKarl98:
This closes issue #19705 : 

- Introduced per-peer pending queues with optional event-driven scheduling (configurable via enable_peer_queue_scheduler). Announcements, request failures, and completions can trigger immediate fetch scheduling without relying solely on poll-time.

- Factored request packing into a reusable RequestBuilder that enforces hash-count and byte-size limits (mirroring geth’s scheduleFetches behavior), with boundary tests retained.

- Integrated capacity-aware search budgeting from main for idle-peer discovery and pending-hash intersection (search_breadth_budget_* and pool capacity closures) to avoid overloading the fetcher or pending pool.

- Added metrics for the peer-queue scheduler (queue size/hash count) to improve observability of the new flow.

- Request construction remains governed by both hash count and estimated payload size; scheduling can run event-driven while respecting configured capacity budgets.

Geth inspiration:

- State-machine layout with per-peer queues.

- Request construction governed by both hash count and estimated payload size.
Immediate, event-driven scheduling plus richer instrumentation.

cc @mattsse @Rjected  @klkvr @shekhirin


crates/net/network/src/transactions/fetcher.rs
```diff
@@ -496,6 +410,61 @@ impl<N: NetworkPrimitives> TransactionFetcher<N> {
         }
     }
 
+    fn try_request_via_peer_queue(
+        &mut self,
+        peers: &HashMap<PeerId, PeerMetadata<N>>,
+        search_durations: &mut TxFetcherSearchDurations,
+    ) -> bool {
```
mattsse, line 451: could we add some additional docs to this?

mattsse (commented):
cool, all of this looks very good

need a bit more time to review this in detail, but we can move forward with this

crates/net/network/src/transactions/fetcher.rs
0xKarl98: Have updated

PR #19739, opened 2025-11-14
perf(persistence): improve write batch for HashedPostState & TrieUpdatesSorted

duyquang6:
- Batch `HashedPostState` & `TrieUpdatesSorted`  in `save_blocks` to improve batch processing performance.

**Performance Results:**
- Tested on devnet with 60k TPS raw-transfer workload, with default config (engine.persistence-thresh-hold = 3):
  - Before: ~50ms per block → ~150ms total for 3 blocks
  - After: ~50ms total for 3 blocks (3x improvement)
Note: This assumes the number of accounts remains relatively stable across blocks, which aligns with typical production workloads.

Before: 
image
*Note: p0.9 is sum of total time for all persist blocks, so it will be multiple by number of block need to persist 

After:

image

Updated:

With --engine.persistence-threshold 6, so it need persist 7 blocks per interval:

For **TrieUpdatesSorted**:
Image

after:

Image

- reduce from 125ms -> ~100ms

Note: If we use default config: flush 3 blocks interval, the diff is not much different
But this still a win if in case our persistence very slow, and it need persist tons of block to catch-up with the tip

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -69,7 +69,8 @@ use reth_trie::{
         TrieCursorIter,
     },
     updates::{StorageTrieUpdatesSorted, TrieUpdatesSorted},
-    HashedPostState, HashedPostStateSorted, StoredNibbles, StoredNibblesSubKey, TrieChangeSetsEntry,
+    HashedPostState, HashedPostStateSorted, StoredNibbles, StoredNibblesSubKey,
+    TrieChangeSetsEntry,
 };
 use reth_trie_db::{
     DatabaseAccountTrieCursor, DatabaseStorageTrieCursor, DatabaseTrieCursorFactory,
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -69,7 +69,7 @@ use reth_trie::{
         TrieCursorIter,
     },
     updates::{StorageTrieUpdatesSorted, TrieUpdatesSorted},
-    HashedPostStateSorted, StoredNibbles, StoredNibblesSubKey, TrieChangeSetsEntry,
+    HashedPostState, HashedPostStateSorted, StoredNibbles, StoredNibblesSubKey, TrieChangeSetsEntry,
 };
 use reth_trie_db::{
     DatabaseAccountTrieCursor, DatabaseStorageTrieCursor, DatabaseTrieCursorFactory,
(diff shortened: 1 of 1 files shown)


shekhirin (approved):
LGTM, cc @mediocregopher

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -305,12 +306,14 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
+        let mut aggregated_hashed_state = HashedPostState::default();
+
         // TODO: Do performant / batched writes for each type of object
         // instead of a loop over all blocks,
         // meaning:
         //  * blocks
         //  * state
-        //  * hashed state
+        //  * hashed state (already done)
         //  * trie updates (cannot naively extend, need helper)
```
mediocregopher: This comment isn't accurate, we can collect a TrieUpdatesSorted and extend it with the trie updates from each block, and then only do a single `write_trie_updates_sorted` at the end, just like you're doing here with the hashed state. Might be worth trying out in this PR?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -324,15 +327,19 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
-            self.write_hashed_state(&Arc::unwrap_or_clone(hashed_state).into_sorted())?;
+            aggregated_hashed_state.extend(Arc::unwrap_or_clone(hashed_state));
 
             // sort trie updates and insert changesets
             let trie_updates_sorted = (*trie_updates).clone().into_sorted();
             self.write_trie_changesets(block_number, &trie_updates_sorted, None)?;
             self.write_trie_updates_sorted(&trie_updates_sorted)?;
         }
 
+        // batch insert hashes and intermediate merkle nodes
```
mediocregopher: nit: you're not inserting intermediate merkle nodes here

mediocregopher (approved):
One nitpick on a comment, and a suggestion for another easy win you could do here but it's optional

mediocregopher (approved):

crates/storage/provider/src/providers/database/provider.rs
duyquang6: hi @mediocregopher, about comment, that is obsolete I think

Let me try benchmark and update to you again

duyquang6: Thank for suggestion, here is bench result

I use `--engine.persistence-threshold 6`, so it need persist 7 blocks per interval:
before:

Image

after:

Image

reduce from 125ms -> ~100ms
Note: with default `--engine.persistence-threshold 2`: flush 3 blocks interval, the difference is not much

mediocregopher: Very nice, thanks for re-benching it :pray:

mediocregopher:
@duyquang6 this had to be reverted as it caused state root mismatches.

The issue is that the writing of changesets (`write_state` for state, and `write_trie_changesets` for trie nodes) must be called with the updates for the previous block having been written, otherwise incorrect changesets get written. By aggregating updates and writing them at the end these method calls are not called on top of correct DB state.

`write_trie_changesets` accepts an overlay argument which I think could be used with the updates aggregated so far to correctly write changesets even while aggregating. `write_state` does not accept an argument like this, one would need to be added to achieve the aggregation logic.

PR #19740, opened 2025-11-14
docs: add comprehensive documentation for NodeType enum

CreeptoGengar:
Adds detailed documentation to the NodeType enum explaining its role in the RESS protocol handshake and connection validation rules. 
The documentation now matches the style used for other enums in the module and provides clear guidance on when to use is_valid_connection.


mattsse (approved):
I allow it because I find this quite useful actually

PR #19743, opened 2025-11-14
feat: Inline TransactionConditional in OpPooledTransaction to avoid heap allocation

phrwlk:
Replace Option<Box<TransactionConditional>> with Option<TransactionConditional> in OpPooledTransaction and adjust setters/getters. Boxing was unnecessary because conditional is owned, accessed by &mut self on set, and read by reference for checks; no interior mutability or stable address is required. This removes a per-transaction heap allocation while preserving semantics and external API, contrasting with interop which intentionally uses Arc<AtomicU64> due to &self setter and shared state across clones.


mattsse (commented):
this type is none for almost all txs, so we can try to keep this type smaller

PR #19744, opened 2025-11-14
docs: resolve broken intra-doc link in transactions module

Fallengirl:
Remove impl-level doc comment referencing test_utils module to maintainconsistency with other impl blocks in the file, which document individual methods rather than the impl block itself.

The comment also referenced feature-gated code (test_utils::PeerHandle),making it potentially confusing for users without the test-utils feature.


mattsse (commented):
yeah this comment doesnt make much sense

mattsse (approved):

PR #19746, opened 2025-11-14
perf(rpc): reduce estimate gas trait bounds to EvmStateProvider

datnguyencse:
Replaces `StateProvider` with `EvmStateProvider` trait bound in `estimate_gas_with` since the EVM only needs the minimal database interface, not the full state provider API with block lookups, proofs, etc. This reduces trait overhead and makes dependencies explicit.

It also makes it easier to implement another state provider to feed `estimate_gas_with`, which we need.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/estimate.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/estimate.rs
@@ -6,21 +6,23 @@ use alloy_evm::overrides::apply_state_overrides;
 use alloy_network::TransactionBuilder;
 use alloy_primitives::{TxKind, U256};
 use alloy_rpc_types_eth::{state::StateOverride, BlockId};
+use core::fmt::Debug;
 use futures::Future;
 use reth_chainspec::MIN_TRANSACTION_GAS;
 use reth_errors::ProviderError;
-use reth_evm::{ConfigureEvm, Database, Evm, EvmEnvFor, EvmFor, TransactionEnv, TxEnvFor};
-use reth_revm::{database::StateProviderDatabase, db::State};
+use reth_evm::{ConfigureEvm, Evm, EvmEnvFor, EvmFor, TransactionEnv, TxEnvFor};
+use reth_revm::{database::{EvmStateProvider, StateProviderDatabase}, db::State};
 use reth_rpc_convert::{RpcConvert, RpcTxReq};
 use reth_rpc_eth_types::{
     error::{api::FromEvmHalt, FromEvmError},
     EthApiError, RevertError, RpcInvalidTransactionError,
 };
 use reth_rpc_server_types::constants::gas_oracle::{CALL_STIPEND_GAS, ESTIMATE_GAS_ERROR_RATIO};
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -92,10 +94,12 @@ pub trait EstimateCall: Call {
...
         // Check if this is a basic transfer (no input data to account with no code)
         let is_basic_transfer = if tx_env.input().is_empty() &&
-            let TxKind::Call(to) = tx_env.kind() &&
-            let Ok(code) = db.database.account_code(&to)
+            let TxKind::Call(to) = tx_env.kind()
         {
-            code.map(|code| code.is_empty()).unwrap_or(true)
+            match db.basic(to) {
+                Ok(Some(account)) => account.code_hash == KECCAK_EMPTY,
+                _ => true,
```
mattsse, line 104: why was this changed?

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
```diff
@@ -303,7 +307,7 @@ pub trait EstimateCall: Call {
         max_gas_limit: u64,
     ) -> Result<U256, Self::Error>
     where
-        DB: Database<Error = ProviderError>,
+        DB: Database<Error = ProviderError> + Debug,
```
mattsse: this  seems appropriate, but why is this necessary here?

can we undo?

mattsse (changes requested):
makes sense, but have questions re what appear to be unrelated changes

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
canximan: When switching from the StateProvider trait to EvmStateProvider, we can no longer use the account_code function and can only use the basic or basic_account functions from the Database or EvmStateProvider traits.

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
canximan: Nice catch, let me undo. Because I switched to using revm’s Database trait without Debug, I’ll switch back to reth_revm.

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
mattsse: ah I see, ty

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
datnguyencse: @mattsse In case you've missed this PR, it's ready again Sir.

crates/rpc/rpc-eth-api/src/helpers/estimate.rs
mattsse: then this is equivalent

mattsse (approved):
lgtm

PR #19750, opened 2025-11-14
feat: add --skip-invalid-blocks

mattsse:
instead of terminating, this now records any execution related errors

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -45,7 +45,7 @@ pub struct Command<C: ChainSpecParser> {
     #[arg(long, default_value = "10")]
     num_tasks: u64,
 
-    /// Continues with execution when an invalid block is encounterd and collects these blocks.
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/re_execute.rs
```diff
@@ -109,20 +115,33 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
+                'blocks: for block in start_block..end_block {
                     let block = provider_factory
                         .recovered_block(block.into(), TransactionVariant::NoHash)?
                         .unwrap();
-                    let result = executor.execute_one(&block)?;
+
+                    let result = match executor.execute_one(&block) {
+                        Ok(result) => result,
+                        Err(err) => {
+                            if skip_invalid_blocks {
```
klkvr, line 132: should this also re-instantiate executor at `block`?

klkvr (approved):

PR #19752, opened 2025-11-14
perf(trie): Cache overlays in the OverlayStateProviderFactory

mediocregopher:
When proof workers are initially spawned they all immediately create a tx and fetch from the database any reverts required to revert the trie back to the target anchor block. We've seen that querying these reverts can take a long time in high-load situations, upwards of 100+ ms.

This PR introduces a cache to OverlayStateProviderFactory which caches the overlays+reverts, with the cache key being the db tip. This way any txs creates with the same view of the DB (which should be most of them) will result in fetching the reverts only a single time.

## Benchmarks

Initial benchmarks are promising, though I am going to run a more comprehensive one on a high-load block range asap

```
Blocks: 38164759 to 38165758. Total: 1000
newPayload time:
• Baseline: 766,228ms
• Feature: 758,661ms
• Difference: 2.96% faster
Gas/sec:
• Baseline: 87.7 Mgas/s
• Feature: 88.6 Mgas/s
• Difference: 1.0% faster
```

image

### DIFF
--- a/crates/storage/provider/src/providers/state/overlay.rs
+++ b/crates/storage/provider/src/providers/state/overlay.rs
@@ -313,6 +313,18 @@ where
     /// cached value then this calculates the [`Overlay`] and populates the cache.
     #[instrument(level = "debug", target = "providers::state::overlay", skip_all)]
     fn get_overlay(&self, provider: &F::Provider) -> ProviderResult<Overlay> {
+        // If we have no anchor block configured then we will never need to get trie reverts, just
+        // return the in-memory overlay.
+        if self.block_hash.is_none() {
+            let trie_updates =
+                self.trie_overlay.clone().unwrap_or_else(|| Arc::new(TrieUpdatesSorted::default()));
+            let hashed_post_state = self
+                .hashed_state_overlay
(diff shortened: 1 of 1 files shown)


Rjected (approved):
nice, makes sense, we def dont need to be recalculating every time

crates/storage/provider/src/providers/state/overlay.rs
```diff
@@ -295,7 +306,89 @@ where
...
+
+        // If the overlay is not present then we need to calculate a new one. We grab a write lock,
+        // and then check the cache again in case some other thread populated the cache since we
+        // checked with the read-lock. If still not present we calculate and populate.
+        let mut cache_miss = false;
+        let overlay = match self
+            .overlay_cache
+            .as_ref()
+            .write()
+            .expect("poisoned mutex")
```
mattsse, line 345: we can replace this with a parkinglot rwlock which I think should already be a dep of this crate

mattsse (approved):
this makes sense to me

smol suggestion re mutex to get rid of the expect but can also be submitted separately

PR #19753, opened 2025-11-14
refactor(stages): empty transactions range

shekhirin:
Instead of relying on `tx_range.is_empty()`, return an explicit `None` that means that no tx range is available for the current input, and the checkpoint should be advanced all the way to the target block.

### DIFF
--- a/crates/stages/stages/src/stages/tx_lookup.rs
+++ b/crates/stages/stages/src/stages/tx_lookup.rs
@@ -150,39 +150,41 @@ where
             );
 
             if range_output.is_final_range {
+                let append_only =
+                    provider.count_entries::<tables::TransactionHashNumbers>()?.is_zero();
+                let mut txhash_cursor = provider
+                    .tx_ref()
+                    .cursor_write::<tables::RawTable<tables::TransactionHashNumbers>>()?;
+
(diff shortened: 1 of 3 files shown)


crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -143,42 +150,40 @@ where
...
+                break;
+            }
+        }
 
-                trace!(target: "sync::stages::transaction_lookup",
-                    total_hashes,
-                    "Transaction hashes inserted"
+        let append_only = provider.count_entries::<tables::TransactionHashNumbers>()?.is_zero();
+        let mut txhash_cursor =
+            provider.tx_ref().cursor_write::<tables::RawTable<tables::TransactionHashNumbers>>()?;
```
mattsse: dont really understand this change here

shekhirin: driveby, reverted and will do separately

mattsse (approved):
okay, now this was a lot easier to follow, ty

this lgtm, makes sense to map empty range to None

pending @joshieDo

joshieDo (approved):

PR #19754, opened 2025-11-14
refactor(cli): db subcommands

shekhirin:
1. Uses `db_exec` macro for all types of `AccessRights`
2. Uses `AccessRights::RoInconsistent` in `reth db stats` if `--inconsistent` flag is provided.

### DIFF
--- a/crates/cli/commands/src/db/mod.rs
+++ b/crates/cli/commands/src/db/mod.rs
@@ -92,7 +92,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
         match self.command {
             // TODO: We'll need to add this on the DB trait.
             Subcommands::Stats(command) => {
-                let access_rights = if command.inconsistent {
+                let access_rights = if command.skip_consistency_checks {
                     AccessRights::RoInconsistent
                 } else {
                     AccessRights::RO
--- a/crates/cli/commands/src/db/mod.rs
+++ b/crates/cli/commands/src/db/mod.rs
@@ -60,10 +60,11 @@ pub enum Subcommands {
     Path,
 }
 
-/// `db_ro_exec` opens a database in read-only mode, and then execute with the provided command
-macro_rules! db_ro_exec {
-    ($env:expr, $tool:ident, $N:ident, $command:block) => {
-        let Environment { provider_factory, .. } = $env.init::<$N>(AccessRights::RO)?;
+/// Initializes a provider factory with specified access rights, and then execute with the provided
+/// command
+macro_rules! db_exec {
+    ($env:expr, $tool:ident, $N:ident, $access_rights:expr, $command:block) => {
+        let Environment { provider_factory, .. } = $env.init::<$N>($access_rights)?;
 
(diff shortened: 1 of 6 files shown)


crates/cli/commands/src/db/stats.rs
```diff
@@ -18,6 +18,10 @@ use std::{sync::Arc, time::Duration};
 #[derive(Parser, Debug)]
 /// The arguments for the `reth db stats` command
 pub struct Command {
+    /// Skip consistency checks for static files.
+    #[arg(long, default_value_t = false)]
+    pub(crate) inconsistent: bool,
```
mattsse: can we name this skip-inconsistency-checks perhaps?

and an alias for --inconsistent 

then it's more obvious what this does

mattsse (changes requested):

crates/cli/commands/src/db/stats.rs
shekhirin: yep

crates/cli/commands/src/db/mod.rs
```diff
@@ -91,27 +92,32 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> Command<C>
         match self.command {
             // TODO: We'll need to add this on the DB trait.
             Subcommands::Stats(command) => {
-                db_ro_exec!(self.env, tool, N, {
+                let access_rights = if command.skip_consistency_checks {
+                    AccessRights::RoInconsistent
```
mattsse, line 96: hmm, unfamiliar with what this does exactly

crates/cli/commands/src/db/static_file_header.rs
```diff
@@ -34,21 +32,12 @@ enum Source {
...
-        // read-only without consistency checks.
-        let provider_factory = match env.init::<N>(AccessRights::RO) {
-            Ok(env) => env,
-            Err(err) => {
-                warn!(?err, "Failed to initialize environment");
-                env.init::<N>(AccessRights::RoInconsistent)?
-            }
+    pub fn execute<N: ProviderNodeTypes>(self, tool: &DbTool<N>) -> eyre::Result<()> {
+        let static_file_provider = tool.provider_factory.static_file_provider();
+        if let Err(err) = static_file_provider.check_consistency(&tool.provider_factory.provider()?)
```
mattsse, line 37: I thought this setting is about this?

crates/cli/commands/src/db/mod.rs
shekhirin: skips consistency checks for static files https://github.com/paradigmxyz/reth/blob/d150b0a39b622eb3be634f6a19e829e8aa10bc36/crates/cli/commands/src/common.rs#L142-L145

this is needed because the static files may be corrupted/in progress of writing, and we still want to try to read it

crates/cli/commands/src/db/static_file_header.rs
shekhirin: yes, but it returns an error with `AccessRights::RO`. Here, we just log a warning.

mattsse (approved):
ah I see, I get it now, ty

PR #19758, opened 2025-11-14
fix(prune): avoid extra iterator consumption

Forostovec:
# What was wrong:

Double consumption on early exit: the for key in &mut keys loop consumes the next item before the body runs. If the limiter triggers at loop start, we break, but the current key is already consumed. The subsequent keys.next().is_none() after the loop consumed yet another key.
Misleading done flag: done was derived by advancing the iterator after the loop, not by whether the loop actually exhausted it.

# Why this happens

`for` implicitly calls `next()`  before each iteration.

The extra `next()` to determine done advanced the iterator once more, skipping work and misreporting completion.

# What changed

1. Loop control:
- Replaced `for key in &mut keys` with an explicit loop that checks `limiter.is_limit_reached()` before calling keys.next().

2. Compute done within the loop:
- `done = true` only when keys.next() returns `None`.
- On limit reach, break without advancing and set `done = false`.

### DIFF
--- a/crates/prune/prune/src/db_ext.rs
+++ b/crates/prune/prune/src/db_ext.rs
@@ -23,7 +23,8 @@ pub(crate) trait DbTxPruneExt: DbTxMut {
 
         let mut deleted_entries = 0;
 
-        for key in &mut keys {
+        let done;
+        loop {
             if limiter.is_limit_reached() {
                 debug!(
                     target: "providers::db",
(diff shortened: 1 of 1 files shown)


crates/prune/prune/src/db_ext.rs
```diff
@@ -23,7 +23,8 @@ pub(crate) trait DbTxPruneExt: DbTxMut {
 
         let mut deleted_entries = 0;
 
-        for key in &mut keys {
+        let done;
+        loop {
```
mattsse: can we changes this so that we use `.peekable` on `keys.into_iter().peekable()`

and then 

```
while keys.peek().is_some() {
        ...
    key = keys.next()
}
```

mattsse (changes requested):
this looks correct

I have one style suggestion

crates/prune/prune/src/db_ext.rs
shekhirin: agreed, because I think with these changes we also incorrectly set `done = false;` if both the limit is reached and there's no more keys left to process.

Forostovec: > can we changes this so that we use `.peekable` on `keys.into_iter().peekable()`
> 
> and then
> 
> ```
> while keys.peek().is_some() {
>      if limiter.is_limit_reached() {
>         ...
>     key = keys.next()
> }
> ```

Like this? corrected

shekhirin: @mattsse maybe we can even use https://doc.rust-lang.org/std/iter/struct.Peekable.html#method.next_if here?

mattsse: yeah that's also an option @Forostovec

mattsse (approved):
this lgtm, ty!

PR #19761, opened 2025-11-14
fix(flashblocks): handle sink errors for ping/close responses

Bashmunta:
- Handle errors from poll_ready, start_send, and poll_flush when sending queued Ping/Close responses in WsFlashBlockStream::poll_next.
- On any sink error, reset state to Initial, clear sink/stream, and restart the connection loop.
- Prevents masking write failures and avoids lingering in Stream state with a broken sink.


mattsse (commented):
the start_send is fine here

PR #19762, opened 2025-11-14
Reduce cloning in random block generator

Doryu001:
avoid cloning block/request/withdrawal ranges in `random_block_range` reuse the existing range references and keep sampling behavior unchanged


mattsse (commented):
we can keep this as is,

are you interested in an actual useful pr or do we want to keep doing this?

PR #19763, opened 2025-11-14
Update node_iter.rs

wedjob0X:
collect the state once inside get_hash_builder_branch_nodes, removing the unnecessary Clone constraint and avoiding repeated iteration/cloning, fix clippy’s clone_on_copy lint by dereferencing Nibbles instead of cloning


mattsse (commented):
cloning the intoiterator has the same effect in the worst case, but might not be

PR #19768, opened 2025-11-14
fix(cli): always commit the unwind for stage run headers

### LINKED ISSUE #19766: `reth stage run headers` always fails with static file prune error
### Describe the bug

If you try `reth stage run headers`, for any range, this happens:
```
2025-11-14T21:15:58.521236Z  INFO Initialized tracing, debug log directory: /home/ubuntu/.cache/reth/logs/mainnet
2025-11-14T21:15:58.522207Z  INFO Opening storage db_path="/home/ubuntu/test-datadir/db" sf_path="/home/ubuntu/test-datadir/static_files"
2025-11-14T21:15:58.525536Z  INFO Verifying storage consistency.
2025-11-14T21:16:04.108527Z  INFO Executing stage stage=Headers
2025-11-14T21:16:05.317604Z  INFO Received headers total=10000 from_block=20000 to_block=10001
2025-11-14T21:16:06.319245Z  INFO Received headers total=10000 from_block=10000 to_block=1
2025-11-14T21:16:06.322853Z  INFO Writing headers total=20000
2025-11-14T21:16:06.328251Z ERROR shutting down due to error
Error: database integrity error occurred: Pruning should be committed before appending or pruning more data

Caused by:
    Pruning should be committed before appending or pruning more data

Location:
    /home/ubuntu/reth/crates/cli/commands/src/stage/run.rs:359:51
```

### Steps to reproduce

simplest repro: `reth stage run --datadir ~/test-datadir --from 0 --to 20000`

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

main

n/a

mainnet

Archive (default)

_No response_

_No response_

### Code of Conduct

Rjected:
fixes https://github.com/paradigmxyz/reth/issues/19766

Static files do not have transactions and we cannot change our local view of headers in the provider without committing. This caused problems with `stage run headers` which first unwinds the headers stage, queueing a prune but not committing it. Then, when the stage tried to write headers, it noticed the queued prune, which caused `Pruning should be committed before appending or pruning more data`

This causes `stage run` to require `--commit` for the headers, bodies, and execution stage.

### DIFF
--- a/crates/cli/commands/src/stage/run.rs
+++ b/crates/cli/commands/src/stage/run.rs
@@ -84,6 +84,9 @@ pub struct Command<C: ChainSpecParser> {
     /// Commits the changes in the database. WARNING: potentially destructive.
     ///
     /// Useful when you want to run diagnostics on the database.
+    ///
+    /// NOTE: This flag is currently required for the headers, bodies, and execution stages because
+    /// they use static files and must commit to properly unwind and run.
     // TODO: We should consider allowing to run hooks at the end of the stage run,
     // e.g. query the DB size, or any table data.
     #[arg(long, short)]
@@ -105,6 +108,14 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
         Comp: CliNodeComponents<N>,
         F: FnOnce(Arc<C::ChainSpec>) -> Comp,
     {
+        // Quit early if the stages requires a commit and `--commit` is not provided.
+        if self.requires_commit() && !self.commit {
+            return Err(eyre::eyre!(
+                "The stage {} requires overwriting existing static files and must commit, but `--commit` was not provided. Please pass `--commit` and try again.",
+                self.stage.to_string()
+            ));
+        }
+
         // Raise the fd limit of the process.
         // Does not do anything on windows.
         let _ = fdlimit::raise_fd_limit();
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/stage/run.rs
```diff
@@ -340,7 +340,13 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
                     provider_rw.save_stage_checkpoint(unwind_stage.id(), checkpoint)?;
                 }
 
-                if self.commit {
+                // NOTE: headers must be committed because headers cannot be written to if static
+                // file pruning is already queued.
+                //
+                // This also means that all `stage run header` commands commit for unwinding. This
+                // is because static files do not have transactions and we cannot change the view of
+                // headers without writing.
```
mattsse: oh, right, I believe this is correct

pending @joshieDo or @shekhirin

mattsse (approved):

crates/cli/commands/src/stage/run.rs
joshieDo: this might be confusing to whoever is using this - should we instead add a cli requirment that `run headers` should have `--commit` ?

joshieDo: and i guess this should also happen with ::Bodies and ::Execution (txs, receipts) ?

Rjected: actually yeah that makes sense, will forbid running without `--commit` for headers, bodies, and execution stage

Rjected: @joshieDo this now requires `--commit` for these stages and quits if not provided

joshieDo (approved):

PR #19769, opened 2025-11-14
refactor: rename state_root span to await_state_root

jenpaff:
Small nit: It would be nice to rename the span to make it clear whats happening. 

Screenshot 2025-11-14 at 08 11 32


Rjected (approved):
lgtm

PR #19771, opened 2025-11-14
feat(provider): change storage settings on existing nodes via CLI

shekhirin:
This PR makes it possible to change the receipts location (database/static files) on an existing node by modifying the storage settings stored in the DB using a new `reth db settings` subcommand.

```console
projects/oss/reth-copy on alexey/storage-settings-via-cli [!?] is 📦 v1.9.2 via 🦀 v1.91.1
❯ ./target/debug/reth db --chain hoodi settings get --quiet
Current storage settings:
  receipts_in_static_files = false

projects/oss/reth-copy on alexey/storage-settings-via-cli [!?] is 📦 v1.9.2 via 🦀 v1.91.1
❯ ./target/debug/reth db --chain hoodi settings set receipts_in_static_files false --quiet
receipts_in_static_files is already set to false

projects/oss/reth-copy on alexey/storage-settings-via-cli [!?] is 📦 v1.9.2 via 🦀 v1.91.1
❯ ./target/debug/reth db --chain hoodi settings set receipts_in_static_files true --quiet
Set receipts_in_static_files = true
Storage settings updated successfully.

projects/oss/reth-copy on alexey/storage-settings-via-cli [!?] is 📦 v1.9.2 via 🦀 v1.91.1
❯ ./target/debug/reth db --chain hoodi settings get --quiet
Current storage settings:
  receipts_in_static_files = true
```


crates/node/builder/src/launch/common.rs
```diff
@@ -482,6 +486,23 @@ where
...
+        let new_settings = old_settings.with_receipts_in_static_files_opt(receipts_in_static_files);
+
+        if new_settings == old_settings {
+            debug!(target: "reth::cli", settings = ?new_settings, "Storage settings are already up to date");
+        } else {
+            debug!(target: "reth::cli", ?old_settings, ?new_settings, "Updating storage settings");
+            provider_rw.write_storage_settings(new_settings)?;
+            provider_rw.set_storage_settings_cache(new_settings);
+            provider_rw.commit()?;
+        }
```
joshieDo: wondering if we can make this part of `ProviderFactory::new` in a non painful way? Either directly, or indirectly through `StaticFileProvider`

cause right now only `reth node` is really using the flag, even though it's present in way more commands

crates/cli/commands/src/db/settings.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    fn get<N: ProviderNodeTypes>(&self, tool: &DbTool<N>) -> eyre::Result<()> {
+        // Read storage settings
+        let provider = tool.provider_factory.provider()?;
+        let StorageSettings { receipts_in_static_files } =
+            provider.storage_settings()?.unwrap_or_default();
+
+        // Display settings
+        println!("Current storage settings:");
+        println!("  receipts_in_static_files = {receipts_in_static_files}");
```
Rjected: I think making this a `Display` or `Debug` would be good so that we don't have to update this every time we add fields to storage settings

crates/cli/commands/src/db/settings.rs
```diff
@@ -0,0 +1,98 @@
...
+            Subcommands::Get => self.get(tool),
+            Subcommands::Set(cmd) => self.set(cmd, tool),
+        }
+    }
+
+    fn get<N: ProviderNodeTypes>(&self, tool: &DbTool<N>) -> eyre::Result<()> {
+        // Read storage settings
+        let provider = tool.provider_factory.provider()?;
+        let StorageSettings { receipts_in_static_files } =
+            provider.storage_settings()?.unwrap_or_default();
```
Rjected: if `storage_settings` returns `None`, should we just mention that it doesn't exist

crates/cli/commands/src/db/settings.rs
```diff
@@ -0,0 +1,98 @@
...
+
+        Ok(())
+    }
+
+    fn set<N: ProviderNodeTypes>(&self, cmd: SetCommand, tool: &DbTool<N>) -> eyre::Result<()> {
+        // Read storage settings
+        let provider_rw = tool.provider_factory.database_provider_rw()?;
+        // Destruct settings struct to not miss adding support for new fields
+        let mut settings @ StorageSettings { receipts_in_static_files: _ } =
+            provider_rw.storage_settings()?.unwrap_or_default();
```
Rjected: would also be nice to log here that if `None`, we are writing to the table

crates/cli/commands/src/db/settings.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn get<N: ProviderNodeTypes>(&self, tool: &DbTool<N>) -> eyre::Result<()> {
+        // Read storage settings
+        let provider = tool.provider_factory.provider()?;
+        let storage_settings = provider.storage_settings()?;
+
+        // Display settings
+        match storage_settings {
+            Some(settings) => {
+                println!("Current storage settings:");
+                println!("{:#?}", settings);
```
Rjected:
```suggestion
                println!("{settings:#?}");
```

Rjected (approved):
LGTM

PR #19777, opened 2025-11-15
refactor(stages): deduplicate stage validation in replace method

sky-coderay:
Found a redundant check in StageSetBuilder::replace that was validating stage existence twice.

The removed stages.get() check is unnecessary because when stage IDs match, the subsequent set() call does the exact same check again. When IDs differ, index_of() validates existence through the order vector, which is kept in sync with stages by design.


mattsse (commented):
we can keep this as is

PR #19779, opened 2025-11-15
feat(rpc-provider): add `MeteredBatchRequests(Future)`

### LINKED ISSUE #19776: Add metrics for batch request handling
### Describe the feature

we currently don't add metrics for batches:

https://github.com/paradigmxyz/reth/blob/7fdd9c39e3af37dbc37a90688c20069617863a4a/crates/rpc/rpc-builder/src/metrics.rs#L129-L131

like we do for individual calls:

https://github.com/paradigmxyz/reth/blob/7fdd9c39e3af37dbc37a90688c20069617863a4a/crates/rpc/rpc-builder/src/metrics.rs#L116-L126

we want additional metrics for
* batch request started/finised/duration: https://github.com/paradigmxyz/reth/blob/7fdd9c39e3af37dbc37a90688c20069617863a4a/crates/rpc/rpc-builder/src/metrics.rs#L229-L230
* and also increase the RpcServerCallMetrics metrics but introduce a `batch_time` because this measures for single calls https://github.com/paradigmxyz/reth/blob/7fdd9c39e3af37dbc37a90688c20069617863a4a/crates/rpc/rpc-builder/src/metrics.rs#L247-L248

### Additional context

_No response_

lean-apple:
Closes https://github.com/paradigmxyz/reth/issues/19776.

### DIFF
--- a/crates/rpc/rpc-builder/src/metrics.rs
+++ b/crates/rpc/rpc-builder/src/metrics.rs
@@ -62,7 +62,7 @@ impl RpcRequestMetrics {
         Self::new(module, RpcTransport::WebSocket)
     }
 
-    /// Creates a new instance of the metrics layer for Ws.
+    /// Creates a new instance of the metrics layer for Ipc.
     pub(crate) fn ipc(module: &RpcModule<()>) -> Self {
         Self::new(module, RpcTransport::Ipc)
     }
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc-builder/src/metrics.rs
```diff
@@ -245,5 +300,5 @@ struct RpcServerCallMetrics {
     /// The number of failed calls
     failed_total: Counter,
     /// Response for a single call
-    time_seconds: Histogram,
```
mattsse, line 248: this might break existing existing metric setups, so I'd rather keep this unchanged

crates/rpc/rpc-builder/src/metrics.rs
```diff
@@ -127,7 +127,20 @@ where
...
 
     fn batch<'a>(&self, req: Batch<'a>) -> impl Future<Output = Self::BatchResponse> + Send + 'a {
-        self.inner.batch(req)
+        self.metrics.inner.connection_metrics.batches_started_total.increment(1);
+
+        for batch_entry in req.iter().flatten() {
+            let method_name = batch_entry.method_name();
+            if let Some(call_metrics) = self.metrics.inner.call_metrics.get(method_name) {
+                call_metrics.started_total.increment(1);
+            }
```
mattsse, line 136: unfortunately we can't add duration metrics for the individual calls, so just incrementing the calls is fine

mattsse (changes requested):
cool, smol nit

crates/rpc/rpc-builder/src/metrics.rs
lean-apple: Done, https://github.com/paradigmxyz/reth/pull/19779/commits/21e337010670468364ff21e25874ae91ffd4b4fd

mattsse (approved):

PR #19780, opened 2025-11-15
feat: add --http.rpcprefix and --ws.rpcprefix like in geth

cuiweixie:


cuiweixie:
> ## [CodSpeed Performance Report](https://codspeed.io/paradigmxyz/reth/branches/cuiweixie%3Arpcprefix?utm_source=github&utm_medium=comment&utm_content=header)
> ### Merging #19780 will **degrade performances by 31.33%**
> Comparing `cuiweixie:rpcprefix` ([b8af23f](https://github.com/paradigmxyz/reth/commit/b8af23f13058c57cd289d939896b7fe33df585e3)) with `main` ([fce0825](https://github.com/paradigmxyz/reth/commit/fce0825f81e029b8f260cd43a11c06cce5060e3a))
> 
> ### Summary
> `❌ 1` regression `✅ 80` untouched
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/paradigmxyz/reth/branches/cuiweixie%3Arpcprefix?utm_source=github&utm_medium=comment&utm_content=acknowledge)._
> 
> ### Benchmarks breakdown
> Benchmark	`BASE`	`HEAD`	Change
> ❌	[`remove_leaf[1000]`](https://codspeed.io/paradigmxyz/reth/branches/cuiweixie%3Arpcprefix?uri=crates%2Ftrie%2Fsparse%2Fbenches%2Fupdate.rs%3A%3Abenches%3A%3Aremove_leaf%3A%3Aremove_leaf%5B1000%5D&runnerMode=Instrumentation&utm_source=github&utm_medium=comment&utm_content=benchmark)	193.4 µs	281.6 µs	-31.33%

@mattsse It's weird. this pr should has no influence on this bench mark named remove_leaf.

crates/node/core/src/args/rpc_server.rs
```diff
@@ -67,6 +67,16 @@ pub struct RpcServerArgs {
...
 
+    /// HTTP path prefix for JSON-RPC endpoint
+    ///
+    /// Sets a custom path prefix for all HTTP RPC requests.
+    /// By default, the server accepts requests at the root path "/".
+    /// This option allows serving RPC requests under a custom path, for example "/rpc".
+    ///
+    /// Example: --http.rpcprefix /custom/path
+    #[arg(long = "http.rpcprefix", default_value = "/")]
+    pub http_rpcprefix: String,
```
mattsse: I'd prefer if this is an optional setting instead

crates/rpc/rpc-builder/src/path_prefix.rs
```diff
@@ -0,0 +1,178 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower::{Layer, Service};
+
+/// Layer that adds path prefix stripping functionality
+#[derive(Debug, Clone)]
+pub struct PathPrefixLayer {
+    /// The path prefix to strip from requests
+    prefix: String,
```
mattsse: can we make this an `Arc<str>`

crates/rpc/rpc-layer/src/path_prefix.rs
mattsse, line 1: can we move this into the rpc-layer crate instead?

crates/rpc/rpc-layer/src/path_prefix.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Normalizes a path prefix:
+    /// - Ensures it starts with `/`
+    /// - Removes trailing `/` unless it's the root path
+    fn normalize_prefix(prefix: &str) -> String {
+        let prefix = prefix.trim();
+
+        // Empty or root path
+        if prefix.is_empty() || prefix == "/" {
+            return "/".to_string();
+        }
```
mattsse, line 44: is this intended that this will pass through if empty?

mattsse (commented):
thanks!

I'm not sure we need actually need this because I think by default this will behave exactly the same as with this layer, e.g.

```
cast rpc web3_clientVersion -r http://localhost:8545/prefix/
"reth/v1.9.2-7fdd9c3/aarch64-apple-darwin"
```

cuiweixie:
> thanks!
> 
> I'm not sure we need actually need this because I think by default this will behave exactly the same as with this layer, e.g.
> 
> ```
> cast rpc web3_clientVersion -r http://localhost:8545/prefix/
> "reth/v1.9.2-7fdd9c3/aarch64-apple-darwin"
> ```

start with   
```
./target/debug/reth node --dev  --http --http.rpcprefix=/prefix  --http.api=web3,debug,eth,txpool,net --dev.block-time=1s
```

and then  send rpcc with out prefix. you will get 404. that's way expect to act like.
```
cast bn --rpc-url http://127.0.0.1:8545/
```
result in:  
```
Error: HTTP error 404 with empty body
```

crates/rpc/rpc-layer/src/path_prefix.rs
cuiweixie: > is this intended that this will pass through if empty?

yes

crates/rpc/rpc-builder/src/path_prefix.rs
cuiweixie: Done

crates/node/core/src/args/rpc_server.rs
cuiweixie: Done

crates/rpc/rpc-layer/src/path_prefix.rs
cuiweixie: Done

PR #19783, opened 2025-11-15
Tidy history reporting and trie test helpers

DotFunGuy:
reuse Cow<[u64]> in ethstats history reporting to skip extra Vecs and pass &[u64]

drop the Clone bound from trie test helper by collecting the input once before reuse


mattsse (commented):
this isnt needed

PR #19784, opened 2025-11-15
fix(net): use external ip for discv5 config

### LINKED ISSUE #19765: Wrong enode IP and discport used when resolving the peer identity (enode:// URL)
### Describe the bug

**Context**

We run a bunch of reth as well as geth instances in our environment,.

We currently provide both of our reth and geth instances with `--nat=extip:<a-private-ip>` setting. They're expected to be accessible at those private IPs.

Once those instances come up, we execute the `admin_nodeInfo` RPC to double-check their `enr` and `enode://` identity.

**Problem**

For reth, the `enode://` URL returned in the respponse does not contain the IP address we specified. Instead, it seems to contain a public IP (unknown to us) along with a totally different port as compared to the discovery port we specify `30303`.

On the other hand, for geth, the `enode://` URL returned in the response correctly ends up containing the exact private IP we provided, along with the discovery port we configured (happens to be "30303" everytime).

Here are some examples:

For reth we set up, `--nat extip:10.64.128.71 --discovery.v5.port 30303`

And when we called `admin_nodeInfo` RPC on it, we got 
```
   "enode": "enode://foobar@35.193.129.168:30304?discport=4152",
  "ip": "35.193.129.168",
  "ports": {
    "discovery": 4152,
    "listener": 30304
  },
```

The expectation was 

```
   "enode": "enode://foobar@10.64.128.71:30304?discport=30303",
  "ip": "10.64.128.71",
  "ports": {
    "discovery": 30303,
    "listener": 30304
  },
```

On the other hand, for geth, we setup `--nat 10.64.128.96` and we got (as expected)

```
  "enode": "enode://foobar@10.64.128.96:30304?discport=30303",
  "ip": "10.64.128.96",
  "ports": {
    "discovery": 30303,
    "listener": 30304
  },
```

**Consequence** 

All the reth instances end up advertising the `enode://` URL, containing the seemingly wrong IP, to which other reths can't connect and peer with. This causes no reth to be peered with any other reth.

**Interesting Note**

We were able to mitigate this by taking the wrong `enode://` URL provided by reth, and replacing the wrong IP with the right one by our hands. And then, providing the corrected `enode://` URL to other reth instances as a part of their `--bootnodes` arg.

This causes them to successfully peer with each other.

For eg: if the setting was `--nat extip:10.64.128.71 --discovery.v5.port 30303` and admin_nodeInfo RPC returned `"enode://foobar@35.193.129.168:30304?discport=4152",`, we manually took it and fixed it to `"enode://foobar@10.64.128.71:30304?discport=30303",`, and then fed that to the `--bootnodes` setting of other reth instances.

---

My perception is that, reth is trying to resolve the External/Public IP-port which seems to point to the private-ip we originally specified based on PAT (port address translation). Unfortunately, it doesn't seem to be doing so correctly.

### Steps to reproduce

1. Spin up reth with `--nat extip:10.64.128.71 --discovery.v5.port 30303`
2. Call admin_nodeInfo RPC on it.

### Node logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Kubernetes

v1.9.2

-

We are running op-reth

Full via --full flag

_No response_

_No response_

### Code of Conduct

mattsse:
closes #19765

for discv5 we must check if there's an external ip configured

this also adds some additional functions that were useful for the tests

### DIFF
--- a/crates/cli/commands/src/p2p/bootnode.rs
+++ b/crates/cli/commands/src/p2p/bootnode.rs
@@ -60,7 +60,7 @@ impl Command {
         if self.v5 {
             info!("Starting discv5");
             let config = Config::builder(self.addr).build();
-            let (_discv5, updates, _local_enr_discv5) = Discv5::start(&sk, config).await?;
+            let (_discv5, updates) = Discv5::start(&sk, config).await?;
             discv5_updates = Some(updates);
         };
 
--- a/crates/ethereum/node/tests/e2e/rpc.rs
+++ b/crates/ethereum/node/tests/e2e/rpc.rs
@@ -24,7 +24,7 @@ use reth_rpc_api::servers::AdminApiServer;
 use reth_tasks::TaskManager;
 use std::{
     sync::Arc,
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::{SystemTime, UNIX_EPOCH},
 };
 
 alloy_sol_types::sol! {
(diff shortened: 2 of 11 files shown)


Rjected (approved):
lgtm, tracking the local node record in the discv5 makes sense for this

yashvardhan-kukreja:
Hi @mattsse , thanks a lot for the quick fix. This worked well, I just confirmed in runtime.
Would you mind sharing the timeline for the next release?

PR #19787, opened 2025-11-16
chore(rpc): clean up engine_api test helper

MamunC0der:
Remove unused `chain_spec` field and redundant clone in `EngineApiTestHandle`.

Tests pass, clippy clean.


mattsse (commented):
we can keep this as is, we might need access to this in future tests and this would be easier that way

PR #19788, opened 2025-11-16
fix(tasks): prevent infinite recursion in TaskSpawner implementation

forkfury:
Fix infinite recursion in TaskSpawner trait methods for TaskExecutor.

When TaskSpawner::spawn or TaskSpawner::spawn_blocking were called via the trait, they would recursively call themselves instead of delegating to the inherent TaskExecutor methods. 
Changed to use Self::spawn and Self::spawn_blocking to explicitly call the inherent methods, matching the pattern already used in spawn_critical methods.


mattsse (approved):
this should indeed use Self here

PR #19789, opened 2025-11-16
docs: update rust version

VaibhavArora19:
The Minimum Supported Rust Version for RETH is now 1.88.0 as most of the crates do not compile below 1.88.0 as shown in the screenshot below:

Screenshot 2025-11-16 at 1 00 48 PM


mattsse (approved):
fair

Issue #19778, opened 2025-11-15
Add helpers for testing rpc requests with prestate


mattsse:
### Describe the feature

debugging tracing without an archive node is a bit cumbersome

we can simplify this a bit if we have the _correct_ prestate of a tx/block available (ref https://github.com/paradigmxyz/reth/issues/19703)

then we can inject this at genesis and should be able to trace the tx

## TODO
* add helpers for adding prestate to genesis and a sample test for the tx in https://github.com/paradigmxyz/reth/issues/19703

see also this e2e test that creates a chainspec from genesis and then launches the node

https://github.com/paradigmxyz/reth/blob/main/crates/ethereum/node/tests/e2e/eth.rs#L99-L112

for interacting with the trace endpoints an alloy provider can be used to then send the request to the server

### Additional context

@0xKarl98

Issue #19703, opened 2025-11-13
Prestate tracer (diffmode) inconsistent with other clients on selfdestruct


antoniolocascio:
### Describe the bug

The prestate tracer with diffmode enabled isn't reporting updates for an account that was selfdestructed in a transaction after being sent funds on a previous transaction in the same block.

My understanding is that, in this case, the "pre" field of that account should have some balance and the account should not be present in the "post". However, the account isn't present on either section. Geth response for the same request diverges from reth.

Concretely, on block 19484626 (Ethereum mainnet):
1. Transaction `0xbd80f521740efed4539f1de9508ec528fe21d88b264ad3ee96baaa14a005bac7` (index 14) performs a transfer to `0x5cf6756af36644fc13b2d5d8637a50ae7e6c5fef` (which previously had no balance).
2. Transaction `0x391f4b6a382d3bcc3120adc2ea8c62003e604e487d97281129156fd284a1a89d` (index 58) deploys `0x5cf6756af36644fc13b2d5d8637a50ae7e6c5fef` via CREATE2. This deployment ends with a SELFDESTRUCT, clearing its balance.

[reth_response.json](https://github.com/user-attachments/files/23518446/reth_response.json)

[geth_response.json](https://github.com/user-attachments/files/23518448/geth_response.json)

### Steps to reproduce

Steps to reproduce:

1. Do an RPC call to a reth archive node with the following request:
```
{"id":1,"jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["0x1294fd2",{"tracer":"prestateTracer","tracerConfig":{"diffMode":true}}]}
```
I used a reth node hosted by Alchemy.
2. Perform the same request to a geth node, compare differences for account `0x5cf6756af36644fc13b2d5d8637a50ae7e6c5fef`.

### Node logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

### What version/commit are you on?

v1.7.0-9d56da5

### What database version are you on?

Not sure how to check

### Which chain / network are you on?

Ethereum mainnet (1)

### What type of node are you running?

Archive (default)

### Code of Conduct

- [x] I agree to follow the Code of Conduct

mattsse:
@antoniolocascio mind trying with that endpoint and see if you can spot anything else that needs to get fixed?

https://reth-ethereum.ithaca.xyz/rpc

mattsse:
thanks @antoniolocascio 

[...]

PR #19790, opened 2025-11-16
feat: add helpers for testing rpc requests with prestate

0xKarl98:
This closes issue #19778 , which as : 

- We now load the geth prestateTracer dump for tx 0x391f…a1a89d, extend the genesis alloc with it, boot a temporary node, and invoke debug_traceCall (via an Alloy HTTP provider) with the prestate tracer settings
- The returned GethTrace::PreStateTracer frame is filtered for the self-destructing contract and related accounts, then compared against the original geth snapshot

- This end-to-end test proves that Reth’s tracer preserves the balance/code/storage for self-destructing contracts, matching geth’s behavior.

Related parent issue : #19703


crates/ethereum/node/tests/e2e/prestate.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let (mut nodes, _tasks, _) =
+        setup::<EthereumNode>(1, chain_spec, false, eth_payload_attributes).await?;
+    let node = nodes.pop().expect("node available");
+    let provider = ProviderBuilder::new().connect_http(node.rpc_url());
+
+    let trace: alloy_rpc_types_trace::geth::GethTrace = provider
+        .raw_request::<_, _>(
+            "debug_traceCall".into(),
+            (
```
mattsse: we can enable the debug-api feature for this in the alloy-provider crate for this, which provides some useful helpers exactly for this, see 

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/provider/src/ext/debug.rs#L17-L17

testing/testing-utils/src/prestate.rs
```diff
@@ -0,0 +1,239 @@
...
+    path::Path,
+    str::FromStr,
+};
+
+/// Map of accounts that should be inserted into the genesis alloc.
+pub type PrestateAlloc = BTreeMap<Address, GenesisAccount>;
+
+/// Errors that can occur when parsing or applying a prestate dump.
+#[derive(Debug)]
+pub enum PrestateError {
```
mattsse: don't think we need this, since this is just for test-utils, we can really just use eyre::Result for error handling

testing/testing-utils/src/prestate.rs
mattsse, line 1: don't think we need this really, we should be able to parse the json response into PrestateFrame and then simply convert the account objects into genesisaccount object

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/rpc-types-trace/src/geth/pre_state.rs#L129-L143

mattsse (changes requested):
nice,

some suggestions

crates/ethereum/node/tests/e2e/prestate.rs
0xKarl98: Yeah found it , debug_trace_call_prestate

mattsse (commented):
this was very useful, ty!

the last thing I need to figure out here is why exactly this doesn't match the expected prestate input, must be something stupid

crates/ethereum/node/tests/e2e/prestate.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Replays the selfdestruct transaction via `debug_traceCall` and ensures Reth's prestate matches
+/// Geth's captured snapshot.
+// <https://github.com/paradigmxyz/reth/issues/19703>
+#[tokio::test]
+async fn debug_trace_call_matches_geth_prestate_snapshot() -> Result<()> {
+    reth_tracing::init_test_tracing();
+
+    let mut genesis: Genesis = MAINNET.genesis().clone();
+    genesis.coinbase = address!("0x95222290dd7278aa3ddd389cc1e1d165cc4bafe5");
```
mattsse, line 30: ah nice, of course

mattsse (changes requested):
awesome,

I think we can undo the changes in the genesis file and solve that programatically, because then we can reuse that test more easily?

mattsse (approved):
sweet, this will be super useful for more tests in the future

PR #19791, opened 2025-11-16
chore(payload): remove redundant attr clone

sashass1315:
The BuildNewPayload path in PayloadBuilderService::poll() cloned the received attributes twice: once to pass ownership into 
new_payload_job and again to emit Events::Attributes. Since Events::Attributes requires an owned value and the original attr is not used afterwards, we can move attr into the event and keep a single clone for new_payload_job. This preserves semantics (attributes event still reflects what was received from the CL), avoids an unnecessary allocation/copy for potentially heavy attributes, and keeps logging (id, parent) intact as they are read before moving.


mattsse (approved):
letting those one pass because this should be as performant as possible

PR #19793, opened 2025-11-16
fix(era-downloader): align checksums with file index in fs::read_dir

ANtutov:
This change fixes a mismatch between locally scanned ERA1 files and checksums when starting from a non-zero block. Previously, 
read_dir() skipped a number of entries by count and then consumed checksums.txt from the beginning, which desynchronized files and checksums whenever the directory didn’t start at index 0 or had gaps. The function now computes the logical start file index from start_from, advances the checksum iterator by that index, and filters files by index using skip_while. This aligns file-to-checksum pairing with the global file index semantics used elsewhere (notably the remote path via expected_checksum(number)), preventing false checksum mismatches and early “less checksums than ERA files” errors.

### DIFF
--- a/crates/era-downloader/src/fs.rs
+++ b/crates/era-downloader/src/fs.rs
@@ -12,6 +12,8 @@ pub fn read_dir(
     start_from: BlockNumber,
 ) -> eyre::Result<impl Stream<Item = eyre::Result<EraLocalMeta>> + Send + Sync + 'static + Unpin> {
     let mut checksums = None;
+
+    // read all the files in the given dir and also read the checksums file
     let mut entries = fs::read_dir(dir)?
         .filter_map(|entry| {
             (|| {
@@ -29,6 +31,7 @@ pub fn read_dir(
                         return Ok(Some((number, path.into_boxed_path())));
                     }
                 }
+
                 if path.file_name() == Some("checksums.txt".as_ref()) {
                     let file = fs::open(path)?;
                     let reader = io::BufReader::new(file);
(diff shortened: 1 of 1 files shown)


crates/era-downloader/src/fs.rs
```diff
@@ -43,9 +43,14 @@ pub fn read_dir(
         .collect::<eyre::Result<Vec<_>>>()?;
     let mut checksums = checksums.ok_or_eyre("Missing file `checksums.txt` in the `dir`")?;
 
+    let start_index = start_from as usize / BLOCKS_PER_FILE;
+    for _ in 0..start_index {
+        checksums.next().transpose()?.ok_or_eyre("Got less checksums than ERA files")?;
+    }
```
mattsse, line 53: I see,
can we document this with a oneliner perhaps?

crates/era-downloader/src/fs.rs
```diff
@@ -43,9 +43,14 @@ pub fn read_dir(
...
+
     entries.sort_by(|(left, _), (right, _)| left.cmp(right));
 
-    Ok(stream::iter(entries.into_iter().skip(start_from as usize / BLOCKS_PER_FILE).map(
+    Ok(stream::iter(entries.into_iter().skip_while(move |(n, _)| *n < start_index).map(
```
mattsse, line 57: I see, this is equivalent because this now starts at the start_index + 1 effectively

mattsse (changes requested):
@lean-apple can't remember if you worked on this previously, if so perhaps worth checking.

@ANtutov one pedantic doc nit

crates/era-downloader/src/fs.rs
lean-apple: It seems indeed better but I think we should cover the case where we have non-continuous files also. 

Right now the files we used are hosted internally where it's continuous but in case of desync we will still have an issue, between the file index and the checksum one. It might be worthy as it is used in the [import-era cli](https://github.com/paradigmxyz/reth/blob/ab5c4452c1ff435ef96e52a57476b2f4f37f8275/crates/cli/commands/src/import_era.rs#L84).

mattsse (approved):
okay, this checks out

PR #19795, opened 2025-11-16
docs: clarify BetterPayloadEmitter emits both Better and Freeze outcomes

CreeptoGengar:
Update documentation to accurately reflect that BetterPayloadEmitter sends payloads for both Better and Freeze outcomes, not just Better.
Both represent valid payloads that need to be available to subscribers.


mattsse (approved):
letting this pass because this is an essential component and I think these docs are better

PR #19798, opened 2025-11-17
chore(provider): use parking_lot RwLock for overlay cache

### LINKED ISSUE #19770: Replace std rwlock with parkinglot
### Describe the feature

https://github.com/paradigmxyz/reth/blob/f88fae0ea1fb0143eb08563a29366b9f430e59fc/crates/storage/provider/src/providers/state/overlay.rs#L332-L332

uses std rwlock which has lock poisoning and requires expect, we already have parking_lot as dep and can relace the std rwlock with parking_lot::rwlock

### Additional context

_No response_

Andrurachi:
The `std::sync::RwLock` used for the `overlay_cache` in `OverlayStateProviderFactory` supports lock poisoning. This requires using `.expect("poisoned mutex")` on every lock acquisition, which will panic the thread if the lock is ever poisoned.

This commit replaces the `std::sync::RwLock` with `parking_lot::RwLock`. Since `parking_lot`'s locks are not poisonable, this change allows us to safely remove the `.expect()` calls.

Fixes #19770

### DIFF
--- a/crates/storage/provider/src/providers/state/overlay.rs
+++ b/crates/storage/provider/src/providers/state/overlay.rs
@@ -1,5 +1,6 @@
 use alloy_primitives::{BlockNumber, B256};
 use metrics::{Counter, Histogram};
+use parking_lot::RwLock;
 use reth_db_api::DatabaseError;
 use reth_errors::{ProviderError, ProviderResult};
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

PR #19799, opened 2025-11-17
perf(payload): remove string allocations and unused import

phrwlk:
Replace redundant String allocations when constructing EngineObjectValidationError::InvalidParams from string literals by converting &'static str directly into Box<dyn Error + Send + Sync> with .into(); &'static str implements core::error::Error and Box<dyn Error + Send + Sync> provides a From for error types, so avoiding to_string() removes unnecessary allocations without changing behavior or messages; this is no_std friendly with alloc, compatible with the project’s MSRV, and keeps tests and external API unaffected while reducing minor overhead in hot error paths.

### DIFF
--- a/crates/payload/primitives/src/lib.rs
+++ b/crates/payload/primitives/src/lib.rs
@@ -13,7 +13,6 @@
 
 extern crate alloc;
 
-use crate::alloc::string::ToString;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is fine, we can use the display impl of static str directly

PR #19800, opened 2025-11-17
perf(trie): optimize chunking configuration 

### LINKED ISSUE #19651: Add max chunking limit for storage proof fetching

yongkangc:
### Problem

XEN blocks contain transactions with 400-600 storage slots, causing state root calculations to become a performance bottleneck. XEN is responsible for ~90% of state root time, performed effectively sequentially.

### Solution

Optimize chunking configuration to balance parallelism and overhead:

1. Increase `DEFAULT_MULTIPROOF_TASK_CHUNK_SIZE` from 10 to 60
   - Controls the size of each chunk when chunking occurs. We don't want chunks to be too big, which means we can take advantage of other workers and all will be blocked by one long tasks. But too small means a lot of switching overhead  
2. Introduce `DEFAULT_MAX_TARGETS_FOR_CHUNKING` = 300
   - Forces chunking when proof targets exceed this threshold
   - Ensures large XEN proofs always get chunked regardless of worker availability

### Metrics
- Significantly lower pending storage request p90 size (due to better chunking, particularly for XEN)
image

-  Gas Throughput: +0.21% improvement overall
  - Median gas/s: 710.6M (baseline) → 712.1M (feature)

Latency Changes
- New Payload Average timing: +0.09% (overall average - negligible)
- Per-block latency:
  - Median: -0.07% (slightly faster)
  - Mean: +1.74% (slightly slower)


mediocregopher (approved):
This LGTM, just needs CI fixed

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -921,11 +930,16 @@ impl MultiProofTask {
...
+        let many_proof_targets =
+            not_fetched_state_update.chunking_length() > self.max_targets_for_chunking;
+
         // Only chunk if multiple account or storage workers are available to take advantage of
         // parallelism.
         let should_chunk = self.multiproof_manager.proof_worker_handle.available_account_workers() >
             1 ||
-            self.multiproof_manager.proof_worker_handle.available_storage_workers() > 1;
+            self.multiproof_manager.proof_worker_handle.available_storage_workers() > 1 ||
+            many_proof_targets;
```
mattsse: should we put this at first position in the OR chain here?

mattsse (approved):
lgtm, tiny nit+ CI

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: that looks nicer, thanks!

PR #19802, opened 2025-11-17
chore: add logs around `commit` and `update_index` on `StaticFileProvider`

joshieDo:


mattsse (approved):
we def want those

lgtm, pending @shekhirin

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -859,19 +851,23 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
                 // Update the cached provider.
+                debug!(target: "provider::static_file", ?segment, "Inserting updated jar into cache.");
                 self.map.insert((fixed_range.end(), segment), LoadedJar::new(jar)?);
 
                 // Delete any cached provider that no longer has an associated jar.
+                debug!(target: "provider::static_file", ?segment, "Cleaning up jar map.");
                 self.map.retain(|(end, seg), _| !(*seg == segment && *end > fixed_range.end()));
             }
             None => {
+                debug!(target: "provider::static_file", ?segment, "Removing segment from index.");
```
shekhirin: ocpd

```suggestion
                debug!(target: "provider::static_file", ?segment, "Inserting updated jar into cache");
                self.map.insert((fixed_range.end(), segment), LoadedJar::new(jar)?);

                // Delete any cached provider that no longer has an associated jar.
                debug!(target: "provider::static_file", ?segment, "Cleaning up jar map");
                self.map.retain(|(end, seg), _| !(*seg == segment && *end > fixed_range.end()));
            }
            None => {
                debug!(target: "provider::static_file", ?segment, "Removing segment from index");
```

shekhirin (approved):

PR #19803, opened 2025-11-17
refactor(provider): unify static file indexes into one struct

shekhirin:
This PR fixes a potential locking issue.

#### Thread A (Reader - getting transaction):
1. Acquires `static_files_tx_index.read()` https://github.com/paradigmxyz/reth/blob/cac2443dfc3bf9e728f9a89a69693e556c455609/crates/storage/provider/src/providers/static_file/manager.rs#L732
1. While holding this lock, calls `find_fixed_range()` https://github.com/paradigmxyz/reth/blob/cac2443dfc3bf9e728f9a89a69693e556c455609/crates/storage/provider/src/providers/static_file/manager.rs#L746
1. `find_fixed_range()` tries to acquire `static_files_expected_block_index.read()` https://github.com/paradigmxyz/reth/blob/cac2443dfc3bf9e728f9a89a69693e556c455609/crates/storage/provider/src/providers/static_file/manager.rs#L434

Lock order: `tx_index -> expected_block_index`

#### Thread B (Writer - committing transactions segment):
https://github.com/paradigmxyz/reth/blob/cac2443dfc3bf9e728f9a89a69693e556c455609/crates/storage/provider/src/providers/static_file/manager.rs#L772-L773

Lock order: `expected_block_index -> tx_index`

## Solution

Move locks into one struct behind an `RwLock`, and acquire locks on the whole struct. That way, it won't be possible to acquire locks on indexes in different order.


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -834,41 +829,35 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
                     // If the index is empty, just remove it.
-                    if tx_index.get(&segment).is_some_and(|index| index.is_empty()) {
-                        tx_index.remove(&segment);
-                    }
+                    index.available_block_range_by_max_tx.take_if(|index| index.is_empty());
                 }
 
                 // Update the cached provider.
-                self.map.insert((fixed_range.end(), segment), LoadedJar::new(jar)?);
+                self.jars.insert((fixed_range.end(), segment), LoadedJar::new(jar)?);
```
joshieDo: should we explicitly drop the indexes lock before entering here? probably so, but unsure if compiler is doing it

joshieDo: i don't think we have such scenario in the codebase, but i think there is an edge case if it's not dropped before self.jars methods: 

`threadA(???):` acquire `LoadedJarRef` 
`threadB(update_index)`: acquire `indexes` write lock
`threadA(???)`: while having the `LoadedJarRef`, block on `indexes` read lock
`threadB(update_index)`: block on `self.jars.insert`/`self.jars.retain`

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -877,57 +866,50 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
         for (segment, headers) in iter_static_files(&self.path).map_err(ProviderError::other)? {
             // Update first and last block for each segment
-            if let Some((block_range, _)) = headers.first() {
-                min_block.insert(segment, *block_range);
-            }
-            if let Some((block_range, _)) = headers.last() {
-                max_block.insert(segment, block_range.end());
-            }
+            let min_block_range = Some(headers.first().expect("headers are not empty").0);
+            let max_block = headers.last().expect("headers are not empty").0.end();
```
joshieDo, line 891: unsure we can call `expect` here. imagine a scenario where we create the writer (and commit?) but the node shuts down before we set the block range

shekhirin: I think we can, because entries in the return map can only be added here https://github.com/paradigmxyz/reth/blob/d313558241c14e90ed74837f5a383bc3ec1950db/crates/storage/db/src/static_file/mod.rs#L39-L42

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: I don't think so, because it will mean that the indices are no out of sync with `self.jars`. Unsure about the implications of this.

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -787,6 +768,28 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+                    })
+                    .or_insert_with(|| StaticFileSegmentIndex {
+                        min_block_range: None,
+                        max_block: segment_max_block,
+                        expected_block_ranges_by_max_block: BTreeMap::from([(
+                            fixed_range.end(),
+                            fixed_range,
+                        )]),
+                        available_block_ranges_by_max_tx: None,
+                    });
```
mattsse, line 796: this is a bit hard to follow without additional docs

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -729,12 +711,13 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         segment: StaticFileSegment,
         tx: u64,
     ) -> Option<SegmentRangeInclusive> {
-        let static_files = self.static_files_tx_index.read();
-        let segment_static_files = static_files.get(&segment)?;
+        let indexes = self.indexes.read();
```
mattsse, line 714: do we need to hold the read lock for the entire fn?

crates/storage/provider/src/providers/static_file/manager.rs
mattsse: we need a definitive answer to this

mattsse (commented):
afaict, all of these make sense and are equivalent, but this isn't easy to review

crates/storage/provider/src/providers/static_file/manager.rs
joshieDo: yeah makes sense, the iter function only adds if theres any block range

crates/storage/provider/src/providers/static_file/manager.rs
joshieDo: we should refactor the whole sf provider into a service

joshieDo: in the future

crates/storage/provider/src/providers/static_file/manager.rs
shekhirin: we do, because the ranges from this index https://github.com/paradigmxyz/reth/blob/90abe8dc9dd5159bb80a9db30381788541090c9e/crates/storage/provider/src/providers/static_file/manager.rs#L716-L720 should be in sync with this index https://github.com/paradigmxyz/reth/blob/90abe8dc9dd5159bb80a9db30381788541090c9e/crates/storage/provider/src/providers/static_file/manager.rs#L731

mattsse (approved):
lgtm

PR #19808, opened 2025-11-17
fix(rpc): clear_txpool returns actually removed transaction count

Snezhkko:
The Admin API docs require admin_clearTxpool to return the number of transactions that were removed from the pool. The previous implementation counted the number of hashes requested for removal, which could overstate the result in the presence of concurrent removals or races. This change uses the length of TransactionPool::remove_transactions(...) to report the number of transactions actually removed, aligning the behavior with the API contract and the pool’s semantics.


mattsse (approved):
right, this can potentially drift, but in this case this doesn't really matter

PR #19810, opened 2025-11-17
fix: keep latency plots aligned when filtering NaNs

aso20455:
I noticed our bench comparison script goes off the rails whenever a run reports zero total latency. The script filters out the NaN percent differences, but the rest of the arrays keep their original indices, so the plots and stats point at the wrong rows.

This change applies the same validity mask to every series (latencies, gas, block numbers, etc.). That keeps the data aligned and lets the script run cleanly even with zero-latency rows.

Tested locally with a CSV that reproduces the issue and confirmed the plots render correctly now.


mattsse (commented):
not needed

PR #19811, opened 2025-11-17
test(prune): fix bodies static files expected block range

shekhirin:
https://github.com/paradigmxyz/reth/blob/d5025392d03102c8c4bc549a7454abb914d78d87/crates/prune/prune/src/segments/user/bodies.rs#L260-L265

We can't use `initial_range` as first argument here, because it expects `expected_block_range` that should be calculated according to the fixed blocks per file configuration set on the provider level.


joshieDo (approved):

mattsse (approved):
makes sense

PR #19813, opened 2025-11-17
perf(ress): optimize witness cache cloning

DeVikingMark:
## Description

Optimizes witness cache cloning to avoid unnecessary double cloning when generating witnesses.

## Changes

- Replace `witness.as_ref().clone()` with `(*witness).clone()` for cache hits
- Eliminate double cloning on first witness generation by creating `Arc` once and cloning from it instead of cloning `Vec` before wrapping


mattsse (commented):
not needed here

PR #19815, opened 2025-11-17
chore: use unwrap_or_default instead of unwrap_or_else

wedjob0X:
This is a simple refactoring that improves code readability and addresses the clippy or_fun_call lint.


mattsse (commented):
not really necessary

Issue #19774, opened 2025-11-15
Improve consistency check UX


mattsse:
### Describe the feature

currently `consistency_check` at launch ensures that all the data in static files and db properly line up and eventually does some healing e.g. if the node was shut down between commits.

however this often returns hard to parse error messages with unclear instructions how to resolve.

ideally we can convert this into specific error variants that we can print instructions for.

this can happen with snapshots that mix and match data for example

PR #19816, opened 2025-11-17
chore: add logs to `StaticFileProvider::check_consistency`

joshieDo:
Adds logs to better observe consistency check behaviour on running nodes.

It's done on spawning the node, so I think it's okay if they're overly verbose.

towards https://github.com/paradigmxyz/reth/issues/19774


mattsse (approved):
lgtm

PR #19817, opened 2025-11-17
fix: stage drop command shouldnt delete jars

joshieDo:
Shouldnt drop the jars before we have actually committed the stage checkpoint to 0. We can only enqueue the pruning, and [let the DB provider worry about the commit order. ](https://github.com/paradigmxyz/reth/blob/23eb96c20925d3e849e2643c0e9b831777dee1eb/crates/storage/provider/src/providers/database/provider.rs#L3161-L3177)

```rust
// We can only enqueue the pruning because if the task is interrupted after we have deleted
// the files, BUT before we have committed the DB we'd lose essential data and would need to
// recover it manually (eg. copy paste static files from other node).
```

### DIFF
--- a/crates/cli/commands/src/stage/drop.rs
+++ b/crates/cli/commands/src/stage/drop.rs
@@ -48,9 +48,11 @@ impl<C: ChainSpecParser> Command<C> {
             _ => None,
         };
 
-        // We can only enqueue the pruning because if the task is interrupted after we have deleted
-        // the files, BUT before we have committed the DB we'd lose essential data and would need to
-        // recover it manually (eg. copy paste static files from other node).
+        // Calling `StaticFileProviderRW::prune_*` will instruct the writer to prune rows only
+        // when `StaticFileProviderRW::commit` is called. We need to do that instead of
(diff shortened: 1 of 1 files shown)


joshieDo:
blocked on https://github.com/paradigmxyz/reth/pull/19819

crates/cli/commands/src/stage/drop.rs
```diff
@@ -47,18 +48,35 @@ impl<C: ChainSpecParser> Command<C> {
             _ => None,
         };
 
-        // Delete static file segment data before inserting the genesis header below
+        // We can only enqueue the pruning because if the task is interrupted after we have deleted
+        // the files, BUT before we have committed the DB we'd lose essential data and would need to
+        // recover it manually (eg. copy paste static files from other node).
```
mattsse: hmm, not super obvious what this means exactly. unsure what `enqueue` refers to here.

or rather what operation this comment tries to explain

mattsse (commented):
pedantic doc nag

mattsse (approved):

shekhirin (approved):

PR #19818, opened 2025-11-17
docs(db): refresh db crate docs to current API and paths

radik878:
Updated the db crate documentation to align with the current codebase. Links were repointed to stable main paths instead of pinned commits and line ranges to avoid drift. The Table trait section now reflects the DUPSORT constant, and the Database trait snippet shows the Debug bound and the correct view closure signature taking a mutable transaction. The transactions API coverage was expanded by including DbTx::get_by_encoded_key and DbTxMut::append to document important performance and ergonomics improvements. Outdated examples were corrected: the stages Transaction example was replaced with a direct tx_mut usage, and HeaderProvider::header_by_number now demonstrates the actual static-file-only path. Static-file headers documentation now targets the current manager implementation with headers_range. The tables list was brought up to date by adding AccountsTrieChangeSets, StoragesTrieChangeSets, and Metadata, and by marking HeaderTerminalDifficulties as deprecated.

### DIFF
--- a/docs/crates/db.md
+++ b/docs/crates/db.md
@@ -8,12 +8,14 @@ The database is a central component to Reth, enabling persistent storage for dat
 
 Within Reth, the database is organized via "tables". A table is any struct that implements the `Table` trait.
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is better than what we have on main

PR #19822, opened 2025-11-18
feat(engine): Update execution cache on inserted executed blocks

### LINKED ISSUE #19747: Update execution cache on inserted executed blocks
### Describe the feature

we have a way to directly insert the executed block into the engine tree:

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/node/builder/src/launch/engine.rs#L281-L286

which listens to emitted payloads from the payload builder.
This is mainly useful for op-reth in derivation mode where the newpayload always becomes the next head

this then gets directly inserted into the state buffer:

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/engine/tree/src/tree/mod.rs#L1356-L1379

one limitation with this is that this bypasses and therefor invalidates the cross block execution cache which usually gets populated with the state after newpayload

this means for a sequencer that wants to insert locally built blocks directly the cache becomes redundant atm

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/engine/tree/src/tree/payload_processor/mod.rs#L98-L99

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L248-L248

what we can do is introduce a trait fn on

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/engine/tree/src/tree/payload_validator.rs#L948-L950

that is `fn on_inserted_exected_block` for example and would then update the cache via

https://github.com/paradigmxyz/reth/blob/2233f8661c6f0c35ea0419552e2af0869bbc4b25/crates/evm/execution-types/src/execution_outcome.rs#L46-L48

### Additional context

_No response_

0x00101010:
Closes https://github.com/paradigmxyz/reth/issues/19747.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -457,6 +457,41 @@ where
             cleared_sparse_trie.lock().replace(cleared_trie);
         });
     }
+
+    /// Updates the execution cache with the post-execution state from an inserted block.
+    ///
+    /// This is used when blocks are inserted directly (e.g., locally built blocks by sequencers)
+    /// to ensure the cache remains warm for subsequent block execution.
+    ///
+    /// The cache enables subsequent blocks to reuse account, storage, and bytecode data without
+    /// hitting the database, maintaining performance consistency.
+    pub(crate) fn on_inserted_executed_block(&self, block_hash: B256, bundle_state: &BundleState) {
+        debug!(target: "engine::caching", block_hash=?block_hash, "Updating execution cache for inserted block");
+
+        self.execution_cache.update_with_guard(|cached| {
+            // Take existing cache (if any) or create fresh caches
+            let (caches, cache_metrics) = if let Some(existing) = cached.take() {
+                existing.split()
+            } else {
+                (
+                    ExecutionCacheBuilder::default().build_caches(self.cross_block_cache_size),
+                    CachedStateMetrics::zeroed(),
+                )
+            };
+
+            // Insert the block's bundle state into cache
+            let new_cache = SavedCache::new(block_hash, caches, cache_metrics);
+            if new_cache.cache().insert_state(bundle_state).is_err() {
+                *cached = None;
+                debug!(target: "engine::caching", "cleared execution cache on update error");
+                return;
+            }
+            new_cache.update_metrics();
+
+            // Replace with the updated cache
+            *cached = Some(new_cache);
+        });
+    }
 }
 
 /// Handle to all the spawned tasks.
@@ -707,6 +742,7 @@ mod tests {
         test_utils::create_test_provider_factory_with_chain_spec,
         ChainSpecProvider, HashingWriter,
     };
+    use reth_revm::db::BundleState;
     use reth_testing_utils::generators;
     use reth_trie::{test_utils::state_root, HashedPostState};
     use revm_primitives::{Address, HashMap, B256, KECCAK_EMPTY, U256};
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -457,6 +457,41 @@ where
...
     }
+
+    /// Updates the execution cache with the post-execution state from an inserted block.
+    ///
+    /// This is used when blocks are inserted directly (e.g., locally built blocks by sequencers)
+    /// to ensure the cache remains warm for subsequent block execution.
+    ///
+    /// The cache enables subsequent blocks to reuse account, storage, and bytecode data without
+    /// hitting the database, maintaining performance consistency.
+    pub(crate) fn on_inserted_executed_block(&self, block_hash: B256, bundle_state: &BundleState) {
```
mattsse: I believe what we also need to enforce here is that the block's parent hash is the current anchor of the cache, oherwise we might encounter a situation where the cache doesn't track the block's parent hash

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1041,6 +1047,13 @@ where
     ) -> ValidationOutcome<N> {
         self.validate_block_with_state(BlockOrPayload::Block(block), ctx)
     }
+
+    fn on_inserted_executed_block(&self, block: ExecutedBlock<N>) {
+        self.payload_processor.on_inserted_executed_block(
+            block.recovered_block.hash(),
```
mattsse: there is which should be useful here

https://github.com/paradigmxyz/reth/blob/main/crates/payload/primitives/src/payload.rs#L34-L34

mattsse (changes requested):
this is great, left one suggestion

crates/engine/tree/src/tree/payload_validator.rs
0x00101010: great, updated

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -457,6 +458,58 @@ where
...
+                        // Restore the cache since we're not updating it
+                        *cached = Some(existing);
+                        return;
+                    }
+                    existing.split()
+                }
+                None => (
+                    ExecutionCacheBuilder::default().build_caches(self.cross_block_cache_size),
+                    CachedStateMetrics::zeroed(),
+                ),
```
mattsse, line 492: I see, this is the case we we have no cache yet and this was received before any regular payload.

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -457,6 +458,58 @@ where
...
+        bundle_state: &BundleState,
+    ) {
+        debug!(target: "engine::caching", ?block_with_parent, "Updating execution cache for inserted block");
+
+        self.execution_cache.update_with_guard(|cached| {
+            // Take existing cache (if any) or create fresh caches
+            let (caches, cache_metrics) = match cached.take() {
+                Some(existing) => {
+                    // Check if existing cache matches parent - if not, restore and skip
+                    if existing.executed_block_hash() != block_with_parent.parent {
```
mattsse: I wonder if we can this any simpler @yongkangc ?

mattsse (commented):
this lgtm

but because this is a bit sensitive, I'd like another review from @yongkangc @shekhirin and @Rjected

crates/engine/tree/src/tree/payload_processor/mod.rs
0x00101010: Or we can do something like this:

```rust
if self.execution_cache.get_cache_for(block_with_parent.parent).is_none() {
    warn!(
        target: "engine::caching",
        parent_hash = %block_with_parent.parent,
        "Cannot find cache for parent hash, skip updating cache with new state",
    );
    return;
}

self.execution_cache.update_with_guard(|cached| {
    // Take existing cache (if any) or create fresh caches
    let (caches, cache_metrics) = match cached.take() {
        Some(existing) => {
            existing.split()
        }
        None => (
            ExecutionCacheBuilder::default().build_caches(self.cross_block_cache_size),
            CachedStateMetrics::zeroed(),
        ),
    };

    // Insert the block's bundle state into cache
    let new_cache = SavedCache::new(block_with_parent.block.hash, caches, cache_metrics);
    if new_cache.cache().insert_state(bundle_state).is_err() {
        *cached = None;
        debug!(target: "engine::caching", "cleared execution cache on update error");
        return;
    }
    new_cache.update_metrics();

    // Replace with the updated cache
    *cached = Some(new_cache);
});
```

0x00101010: Actually we cannot do that:
1. this creates potential race conditions where another write happens between the read and the write operation in this function.
2. also get_cache_for does not properly return if cache is really none or just does not match the parent hash

Opting to move the comparison logic out of take()

yongkangc (approved):
lgtm, do not see any footguns here

one question is for the future: Should there be a metric for cache updates from inserted blocks?

shekhirin (approved):

mattsse (approved):

PR #19824, opened 2025-11-18
refactor: remove unused add_transactions_with_origins trait

datnguyencse:
This PR removes the unused `add_transactions_with_origins` trait from the
transaction pool.

The trait has no remaining call sites and is not part of the active TxPool API.
Cleaning it up reduces noise and makes the API surface easier to understand.

Changes:
- Remove the unused `add_transactions_with_origins` trait
- Drop related implementations and unused code paths

### DIFF
--- a/crates/transaction-pool/src/lib.rs
+++ b/crates/transaction-pool/src/lib.rs
@@ -384,23 +384,6 @@ where
         self.pool.validator().validate_transactions_with_origin(origin, transactions).await
     }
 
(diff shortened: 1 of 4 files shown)


mattsse (approved):
I see, this fn is more or less redundant I believe, so fine with removing it

any objections here @klkvr

klkvr (approved):

yongkangc (approved):

PR #19826, opened 2025-11-18
fix(stages-types): ensure no_std compatibility for tests

davidjsonn:
Running `cargo test -p reth-stages-types --lib --no-default-features` fails with 44 compilation errors, even though the crate declares `no_std` support. The main issues are that `arbitrary::Arbitrary` derive requires std through its dependencies, the `add_arbitrary_tests` macro uses `vec!` without proper imports, and `ToString` isn't in scope for no_std tests.

This fix adds `feature = "std"` guards to arbitrary-related attributes, which is consistent with how other Reth crates handle this. The Cargo.toml already declares `arbitrary = ["std", ...]`, so this just extends that contract to the derive attributes. Also added the missing `ToString` import from alloc in the test module.

After the fix, basic tests pass in no_std mode (5 tests), and all tests including proptests work with std enabled (15 tests). This matches the project architecture where arbitrary/proptest functionality requires std.

### DIFF
--- a/crates/stages/types/src/checkpoints.rs
+++ b/crates/stages/types/src/checkpoints.rs
@@ -220,9 +220,12 @@ impl reth_codecs::Compact for StorageRootMerkleCheckpoint {
 
 /// Saves the progress of `AccountHashing` stage.
 #[derive(Default, Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


mattsse (commented):
this isnt actually required

PR #19827, opened 2025-11-18
fix(rpc): metered getPayloadBodiesByHash timing to await before recording

Bashmunta:
Recorded the latency before awaiting the async operation, which only measured the time to create the future and severely under-reported the actual RPC duration. All other metered methods in engine_api.rs consistently await the operation and then record the elapsed time, and a repository-wide search shows no other instance of pre-await recording. This change applies the same pattern here by awaiting get_payload_bodies_by_hash_v1 before recording the histogram and then returning the already-resolved result, restoring correct latency semantics for the engine_getPayloadBodiesByHashV1 metric and aligning behavior with the rest of the module and the Optimism RPC callers.


mattsse (approved):
his explains the µs durations in the metrics :D

PR #19828, opened 2025-11-18
docs: align eth wire protocol description with supported versions

Fallengirl:
Updated the `reth-eth-wire` crate blurb so it accurately says we implement the eth wire protocol versions 66 through 69


mattsse (commented):
we can keep this as is

PR #19829, opened 2025-11-18
chore: document filter topic usage in db-access example

figtracer:
added better documentation for the usage of event_signature() and topic0 in db-access/receipts_provider_example with a simple erc20 transfer event


mattsse (approved):
this is better so I let this one through

PR #19832, opened 2025-11-18
fix: update AllTransactions metrics on removals and evictions

Galoretka:
AllTransactions metrics were updated on inserts but could remain stale after removals and evictions. Several removal paths invoked only TxPool::update_size_metrics, which updates sub-pool gauges but not AllTransactions gauges, and discard_worst did not update metrics at all. This change makes TxPool::update_size_metrics also refresh AllTransactions metrics, invokes update_size_metrics at the end of discard_worst, and updates AllTransactions metrics inside its removal methods. The result is consistent metrics for both sub-pools and the AllTransactions set across insertion, removal, and eviction flows without relying on incidental future inserts to refresh gauges. Tests cover remove-by-hash and eviction flows.

### DIFF
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -699,6 +699,7 @@ impl<T: TransactionOrdering> TxPool<T> {
         self.metrics.blob_pool_transactions.set(stats.blob as f64);
         self.metrics.blob_pool_size_bytes.set(stats.blob_size as f64);
         self.metrics.total_transactions.set(stats.total as f64);
+        self.all_transactions.update_size_metrics();
     }
 
     /// Updates transaction type metrics for the entire pool.
@@ -1229,7 +1230,7 @@ impl<T: TransactionOrdering> TxPool<T> {
                 queued_limit  => (queued_pool, queued_transactions_evicted),
             ]
         );
-
+        self.update_size_metrics();
         removed
     }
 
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -1229,7 +1230,7 @@ impl<T: TransactionOrdering> TxPool<T> {
                 queued_limit  => (queued_pool, queued_transactions_evicted),
             ]
         );
-
+        self.update_size_metrics();
```
mattsse: we shouldn't do this here because there's no need to update the metrics here

crates/transaction-pool/src/pool/txpool.rs
```diff
@@ -699,6 +699,7 @@ impl<T: TransactionOrdering> TxPool<T> {
         self.metrics.blob_pool_transactions.set(stats.blob as f64);
         self.metrics.blob_pool_size_bytes.set(stats.blob_size as f64);
         self.metrics.total_transactions.set(stats.total as f64);
+        self.all_transactions.update_size_metrics();
```
mattsse, line 702: if we add this here, why do we need to also call this in the other locations

mattsse (commented):
this makes sense in general, but I don't think we need all the additional calls then?

mattsse (commented):
we can leave the metrics as is here

PR #19833, opened 2025-11-18
docs(chainspec): document Optimism Mainnet hardfork block numbers

oooLowNeoNooo:
## Problem

The special case handling for Optimism Mainnet hardfork block numbers lacked documentation. The code used magic numbers (3950000 for Berlin, 105235063 for Bedrock) without explaining why Optimism Mainnet differs from other L2 chains, making it unclear for developers why these specific values are used.

## Solution

Added inline documentation explaining:
- Why Optimism Mainnet has a different hardfork history (Berlin activated at block 3_950_000, Bedrock migration at 105_235_063)
- Historical context for these block numbers
- Reference to `BEDROCK_HEADER` constant for additional context
- Improved function documentation with link to op-geth source

Also formatted numeric literals with underscores for better readability.


mattsse (commented):
I dont want to modify this

PR #19834, opened 2025-11-18
refactor(node): remove duplicate stage status logging logic

sashaodessa:
Removes duplicate `match` statement for stage progress/ETA logging that was present in both `PipelineEvent::Ran` handler and periodic status logger.


mattsse (commented):
this isnt actually equivalent

PR #19842, opened 2025-11-18
fix(net): poll inner stream before reserving and terminate on None

GarmashAlex:
The spawned header/body downloader tasks reserved a PollSender slot before polling the inner stream and returned Pending or even Pending on Ready(None), which kept an OwnedPermit reserved without sending. This reduced effective channel capacity and could leak the reservation indefinitely when the inner stream completed, also preventing proper task termination. Reordered the logic to first poll the inner stream and only reserve a slot when an item is available, and changed handling of Ready(None) to return Poll::Ready(()) to terminate the task. This aligns with tokio-util PollSender semantics and the patterns used elsewhere in the codebase, preventing capacity loss and hangs.

### DIFF
--- a/crates/net/downloaders/src/bodies/task.rs
+++ b/crates/net/downloaders/src/bodies/task.rs
@@ -146,21 +146,24 @@ impl<T: BodyDownloader> Future for SpawnedDownloader<T> {
                 }
             }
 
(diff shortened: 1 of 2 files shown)


mattsse (commented):
this is incorrect because this could lead to dropped messages

PR #19848, opened 2025-11-19
refactor(tree): remove unnecessary block clone

Vitaliyr888:
Remove unnecessary block cloning in the first loop of canonical chain traversal. The second loop (lines 722-724) already does this correctly by extracting parent_hash first, then moving the block without cloning.


mattsse (approved):
fair

PR #19851, opened 2025-11-19
chore(ethereum): avoid cloning requests in try_into_v4

ANtutov:
This change removes an unnecessary clone of the execution requests in EthBuiltPayload::try_into_v4 by switching to Option::take and passing ownership directly into ExecutionPayloadEnvelopeV4. The previous implementation cloned self.requests even though self is immediately consumed by try_into for the inner V3 envelope, leading to avoidable allocations and copies for potentially large alloy_eips::eip7685::Requests. The new approach mirrors try_into_v5 which already moves requests without cloning, keeps the conversion logic centralized by still delegating to try_into_v3 for envelope_inner, and preserves all semantics while improving performance.


mattsse (approved):
this makes sense

perhaps one more sanity check @Rjected

mattsse (approved):

PR #19852, opened 2025-11-19
fix(net): resolve external ip on startup

xzav3r:
## Background
+ Nodes behind NAT should advertise a correct external IP immediately to improve early connectivity.
+ Using interval(...) delayed the first resolution by one full interval.

## Changes
+ Discv4/NAT: switch from `interval(...)` to` interval_at(Instant::now(), …)` so the first tick happens at startup.
+ Tests: add unit test in discv4 config to assert immediate first tick and correct IP.

## Impact
+ No API changes: behavior applies only when `--rollup.discovery.v4` is set.
+ Faster external IP availability on boot: early peer discovery/inbound reachability.
+ Low risk: covered by unit test.

### DIFF
--- a/crates/net/discv4/src/config.rs
+++ b/crates/net/discv4/src/config.rs
@@ -346,7 +346,7 @@ mod tests {
 
         // Create a config with external IP resolver
         let mut builder = Discv4Config::builder();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is better yes, ty!

mattsse (approved):

PR #19853, opened 2025-11-19
feat(engine): new payload last gas/s metric

shekhirin:
Records gas/s throughput for the last new payload call.


yongkangc (approved):

theomart77 (commented):
Update commits

PR #19856, opened 2025-11-19
chore(engine, trie): cleanup traces

shekhirin:
1. Remove `prewarm task` span that had only the `prewarm` span
2. Rename `prewarm` span into` prewarm and caching`
3. Remove `plan_state_root_computation` span
4. Remove `subtrie` span from parallel sparse trie, as it's too noisy and not very useful
5. Start `state root task` span when we start waiting for the state root
6. Add spans for post-execution validation before we wait for the state root

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -481,7 +481,12 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
     /// # Panics
     ///
     /// If payload processing was started without background tasks.
-    #[instrument(level = "debug", target = "engine::tree::payload_processor", skip_all)]
+    #[instrument(
+        level = "debug",
+        target = "engine::tree::payload_processor",
+        name = "await_state_root",
+        skip_all
+    )]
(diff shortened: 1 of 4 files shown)


yongkangc (approved):
Nice, lgtm

mattsse (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -478,12 +481,7 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
     /// # Panics
     ///
     /// If payload processing was started without background tasks.
-    #[instrument(
-        level = "debug",
-        target = "engine::tree::payload_processor",
-        skip_all,
-        name = "await_state_root"
-    )]
+    #[instrument(level = "debug", target = "engine::tree::payload_processor", skip_all)]
```
Rjected: why rm the rename here?

Rjected (approved):
just have one comment about naming but I like the new spans and the spans that were removed due to noise

crates/engine/tree/src/tree/payload_processor/mod.rs
shekhirin: ah oops, missed that it was intentional in https://github.com/paradigmxyz/reth/pull/19769

reverted

PR #19858, opened 2025-11-19
feat: add eth_getAddressesInBlock RPC method

### LINKED ISSUE #19610: Add eth_getAddressesInBlock support
### Describe the feature

see also: https://github.com/paradigmxyz/reth/issues/4054#issuecomment-1669562345

and https://github.com/ethereum/execution-apis/pull/456/files

## TODO
* add ethapi trait fn and impl 

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc-eth-api/src/core.rs#L56-L56

we still need to figure out in which trait fn we want to put this, like
https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc-eth-api/src/helpers/trace.rs#L27-L27

or perhaps a new one that combines some of

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc-eth-api/src/helpers/mod.rs#L46-L46

but what we need as implementation is more or less similar to

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc/src/debug.rs#L88-L97

or 

https://github.com/paradigmxyz/reth/blob/43e5cc7989fb4ef8c303ddf639c7a082a51deab1/crates/rpc/rpc/src/debug.rs#L645-L657

where we need to
1. fetch the block
2. execute the entire block
3. check all accessed accounts, including logs and withdrawals in the block
4. return a unique set of addresses 

### Additional context

_No response_

0xChaddB:
Add a new method `eth_getAddressesInBlock` that returns all unique addresses that appear in a given block.

Implementation: 
   - New `EthAddresses` trait added to `FullEthApi`
  
Fixes #19610

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -449,3 +449,13 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> {
         Ok(())
     }
 }
+
+/// Collects all addresses that appear in blocks during execution.
+#[async_trait::async_trait]
+pub trait EthAddresses: LoadState<Error: FromEvmError<Self::Evm>> {
+    /// Returns the list of addresses that appeared in the given block.
+    async fn get_addresses_in_block(
+        &self,
+        block_id: BlockId,
+    ) -> Result<Vec<alloy_primitives::Address>, Self::Error>;
+}
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -1,8 +1,10 @@
 //! Loads a pending block from database. Helper trait for `eth_` call and trace RPC methods.
 
-use super::{Call, LoadBlock, LoadPendingBlock, LoadState, LoadTransaction};
+use super::{
+    Call, EthBlocks, LoadBlock, LoadPendingBlock, LoadReceipt, LoadState, LoadTransaction,
+};
 use crate::FromEvmError;
-use alloy_consensus::{transaction::TxHashRef, BlockHeader};
+use alloy_consensus::{transaction::TxHashRef, BlockHeader, TxReceipt};
 use alloy_primitives::B256;
 use alloy_rpc_types_eth::{BlockId, TransactionInfo};
 use futures::Future;
@@ -24,7 +26,7 @@ use revm_inspectors::tracing::{TracingInspector, TracingInspectorConfig};
 use std::sync::Arc;
 
 /// Executes CPU heavy tasks.
-pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> {
+pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + LoadBlock + Call + EthBlocks {
     /// Executes the [`TxEnvFor`] with [`reth_evm::EvmEnv`] against the given [Database] without
     /// committing state changes.
     fn inspect<DB, I>(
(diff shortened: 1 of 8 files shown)


crates/rpc/rpc/src/eth/core.rs
```diff
@@ -181,6 +183,150 @@ where
...
+        set.insert(block.header().beneficiary());
+        if let Some(withdrawals) = block.body().withdrawals() {
+            for withdrawal in withdrawals {
+                set.insert(withdrawal.address);
+            }
+        }
+
+        use reth_primitives_traits::SignerRecoverable;
+        for tx in block.body().transactions() {
+            if let Ok(signer) = tx.recover_signer() {
```
mattsse: this should already be recovered so we can iterate over

https://github.com/paradigmxyz/reth/blob/819330c5964572cb820cc28a79b0a2797c50b7c2/crates/primitives-traits/src/block/recovered.rs#L341-L345

crates/rpc/rpc/src/eth/core.rs
```diff
@@ -181,6 +183,150 @@ where
...
+
+        let block = match self
+            .cache()
+            .get_recovered_block(block_hash)
+            .await
+            .map_err(|e| reth_rpc_eth_types::EthApiError::Internal(e.into()))?
+        {
+            Some(block) => block,
+            None => return Ok(Vec::new()),
+        };
```
mattsse: I believe this should already be just

https://github.com/paradigmxyz/reth/blob/819330c5964572cb820cc28a79b0a2797c50b7c2/crates/rpc/rpc-eth-api/src/helpers/block.rs#L281-L281

crates/rpc/rpc/src/eth/core.rs
```diff
@@ -181,6 +183,150 @@ where
...
+        let this = self.clone();
+        let block_for_execution = block.clone();
+        let witness_record = self
+            .spawn_with_state_at_block(
+                block_for_execution.parent_hash().into(),
+                move |state_provider| {
+                    let db = StateProviderDatabase::new(&state_provider);
+                    let block_executor = this.evm_config().executor(db);
+
+                    let mut record = ExecutionWitnessRecord::default();
```
mattsse: I don't think we need this here and can instead take a look at the full state output and iterate over the accounts in there

crates/rpc/rpc/src/eth/core.rs
```diff
@@ -181,6 +183,150 @@ where
...
+                let hash_b256 = alloy_primitives::B256::from(hashed_addr);
+
+                // Validate this is actually an account by checking if its hash exists in
+                // hashed_state This removes false positives while not
+                // relying on private fields
+                if witness_record.hashed_state.accounts.contains_key(&hash_b256) {
+                    set.insert(address);
+                }
+            }
+        }
```
mattsse: then we don't need any of this

crates/rpc/rpc/src/eth/core.rs
```diff
@@ -181,6 +183,150 @@ where
...
+
+        let Some(receipts) = self
+            .provider()
+            .receipts_by_block_id(block_id)
+            .map_err(|e| reth_rpc_eth_types::EthApiError::Internal(e.into()))?
+        else {
+            return Ok(set.into_iter().collect());
+        };
```
mattsse: we can use

https://github.com/paradigmxyz/reth/blob/819330c5964572cb820cc28a79b0a2797c50b7c2/crates/rpc/rpc-eth-api/src/helpers/block.rs#L170-L170

to fetch both block and receipts

crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -449,3 +449,13 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> {
         Ok(())
     }
 }
+
+/// Collects all addresses that appear in blocks during execution.
+#[async_trait::async_trait]
+pub trait EthAddresses: LoadState<Error: FromEvmError<Self::Evm>> {
+    /// Returns the list of addresses that appeared in the given block.
+    async fn get_addresses_in_block(
```
mattsse: any ideas where we can best integrate this @klkvr 

perhaps simply in the Trace api?

https://github.com/paradigmxyz/reth/blob/819330c5964572cb820cc28a79b0a2797c50b7c2/crates/rpc/rpc-eth-api/src/helpers/trace.rs#L27-L27

mattsse (changes requested):
this looks pretty good already

left some nits

last q we need solve @klkvr is where to put this

PR #19859, opened 2025-11-19
chore(trie): add spans for how long we wait on proofs in recv

Rjected:
This adds spans for when we are waiting in `recv`, or in general doing storage proof computation rather than traversing the account trie. This also labels the spans with the hashed address


shekhirin (approved):

mattsse (approved):
makes sense to me but not an expert, so ptal @shekhirin

PR #19861, opened 2025-11-19
fix(primitives-traits): improve test by enforcing full equality

Bashmunta:
This change fixes a logical bug in the test-only PartialEq<AccessList> implementation for RethAccessList. The previous implementation used zip-and-all, which truncates to the shorter iterator and incorrectly accepts proper prefixes as equal, potentially masking discrepancies between encoded and decoded access lists in roundtrip tests. The new implementation uses 
Iterator::eq, which checks both element-wise equality and that both iterators are exhausted, ensuring length equality as well


mattsse (commented):
this isnt needed, we can phase this out shortly anyway

PR #19862, opened 2025-11-19
fix(era): fix roundtrip, decoding receipt for era mainnet test

lean-apple:
Fixes https://github.com/paradigmxyz/reth/pull/17646.

After era 152x, we need to use `ReceiptEnvelope` to decode receipts.

- Use `ReceiptEnvelope` to decode receipts in tests for era1 after 1520

How to check : 

```
cargo nextest run --release --package reth-era --test it -- --ignored
```

- improve file testing with test-case to make more obvious what files we are testing
- use `release` mode for nextest daily run in GHA
- add nextest configuration for slow `reth-era/tests/it`


crates/era/tests/it/roundtrip.rs
```diff
@@ -152,10 +152,9 @@ async fn test_file_roundtrip(
         );
 
         // Decode receipts
-        let original_receipts_decoded =
-            original_block.receipts.decode::<Vec<ReceiptWithBloom>>()?;
+        let original_receipts_decoded = original_block.receipts.decode::<Vec<ReceiptEnvelope>>()?;
```
mattsse, line 152: ah these are typed receipts then?

lean-apple: These are the ones to enable the rlp decoding after the era 152x

mattsse: -.-

mattsse (approved):
nice, ty!

PR #19863, opened 2025-11-19
feat(trie): Proof rewrite: implement stack-based algorithm for calculating trie nodes from leaves

mediocregopher:
Towards #19512 

This PR implements the next step of the proof calculation rewrite. The ProofCalculator can now take a series of leaf values, sorted by key, and calculate all trie nodes in the MPT which is built from those leaves.

Not all nodes are kept in memory, but rather a stack-based approach is used to only keep those nodes in-memory which are required to calculate subsequent nodes for future keys. This approach is essentially identical to that used by [HashBuilder](https://docs.rs/alloy-trie/latest/alloy_trie/hash_builder/struct.HashBuilder.html) but is (imo) much easier to follow.

At the moment no retention of proof nodes is done; passed in proof targets are ignored and the root node is always returned. Proof retention will be implemented in the next PR.

## Testing

Proptests are used to generate random account tries and compare the root node generated by the legacy `reth-trie::Proof`  implementation with that generated by this one.


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -73,18 +373,72 @@ where
...
+            node.into_trie_node(&mut self.rlp_encode_buf)?
+        } else {
+            TrieNode::EmptyRoot
+        };
+
+        proof_nodes.push(ProofTrieNode {
+            path: Nibbles::new(), // root path
+            node: root_node,
+            masks: TrieMasks::none(),
+        });
```
mediocregopher, line 427: This will be removed in the next PR when real proof retention based on `targets` is added, but it's useful for now to verify the implementation

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -0,0 +1,157 @@
...
+        short_key: Nibbles,
+        /// The [`DeferredValueEncoder`] which will encode the leaf's value.
+        value: RF,
+    },
+    /// An extension node whose child branch has not yet been converted to an [`RlpNode`]
+    Extension {
+        /// The short key of the leaf.
+        short_key: Nibbles,
+        /// The node of the child branch.
+        child: BranchNode,
```
mediocregopher, line 24: I'm still not 100% this has to be BranchNode, vs just being the `RlpNode` of the branch. Will know for sure after the next PR

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// not an extension but instead another branch.
+    pub(crate) ext_len: u8,
+    /// A mask tracking which child nibbles are set on the branch so far. There will be a single
+    /// child on the stack for each set bit.
+    pub(crate) state_mask: TrieMask,
+    /// A subset of `state_mask`. Each bit is set if the `state_mask` bit is set and:
+    /// - The child is a branch which is stored in the DB.
+    /// - The child is an extension whose child branch is stored in the DB.
+    #[expect(unused)]
+    pub(crate) tree_mask: TrieMask,
```
mediocregopher, line 141: For now I'm expecting that these masks will _only_ be populated using the nodes already in the DB..... we assume that if a node isn't in the DB then it has empty masks. Maybe this isn't quite true (e.g. for root nodes, which could have masks but we don't ever retain the root) but for now I don't think it will harm anything.

crates/trie/trie/src/hashed_cursor/mock.rs
mediocregopher, line 13: These changes are duplicated from #19864, which this PR is based on so I don't know why GH is showing them.

crates/trie/trie/src/trie_cursor/mock.rs
mediocregopher, line 14: These changes are duplicated from #19864, which this PR is based on so I don't know why GH is showing them.

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -0,0 +1,157 @@
...
+    #[expect(unused)]
+    pub(crate) hash_mask: TrieMask,
+}
+
+/// Trims the first `len` nibbles from the head of the given `Nibbles`.
+///
+/// # Panics
+///
+/// Panics if the given `len` is greater than the length of the `Nibbles`.
+pub(crate) fn trim_nibbles_prefix(n: &Nibbles, len: usize) -> Nibbles {
```
mattsse, line 152: should we also upstream this to nybbles?

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+            .unwrap_or_else(|| Vec::with_capacity(16))
+    }
+
+    /// Pushes a new branch onto the `branch_stack`, while also pushing the given leaf onto the
+    /// `child_stack`.
+    ///
+    /// This method expects that there already exists a child on the `child_stack`, and that that
+    /// child has a non-zero short key. The new branch is constructed based on the top child from
+    /// the `child_stack` and the given leaf.
+    fn push_new_branch(&mut self, leaf_key: Nibbles, leaf_val: VE::DeferredEncoder) {
```
mattsse, line 108: oh I fully get the encoder + deferred encoding now

mattsse (commented):
cool, even with my rather limited trie knowledge these comments made it easy to follow along.

all of this makes sense to me

would like @shekhirin and @yongkangc to also take a look here

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+
+    /// Pops the top branch off of the `branch_stack`, and hashes it (and its extension node as
+    /// well, if there is one). The `branch_path` field will be updated accordingly.
+    ///
+    /// If there remains another branch on the stack then that branch is updated. If not then a
+    /// single [`ProofTrieBranchChild::Hash`] is left on the `child_stack`.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if `branch_stack` is empty.
```
yongkangc, line 202: would there ever be a situation that branch_stack is empty?

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+        let branch_path_len = self.branch_path.len() +
+            common_prefix_len +
+            if self.branch_stack.len() == 1 { 0 } else { 1 };
+        self.branch_path = leaf_key.slice_unchecked(0, branch_path_len);
+
+        trace!(
+            target: TRACE_TARGET,
+            ?leaf_short_key,
+            ?common_prefix_len,
+            new_branch = ?self.branch_stack.last().unwrap(),
```
yongkangc: i am guessing this unwrap() would not happen unless branch stack is empty?

crates/trie/trie/src/proof_v2/mod.rs
yongkangc, line 108: I see, so this basically inserts an intermediate branch when the current branch already has a child on the same nibble as the incoming leaf node

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+    /// the `child_stack` and the given leaf.
+    fn push_new_branch(&mut self, leaf_key: Nibbles, leaf_val: VE::DeferredEncoder) {
+        // First determine the new leaf's shortkey relative to the current branch. If there is no
+        // current branch then the short key is the full key.
+        let leaf_short_key = if self.branch_stack.is_empty() {
+            leaf_key
+        } else {
+            // When there is a current branch then trim off its path as well as the nibble that it
+            // has set for this leaf.
+            leaf_key.slice_unchecked(self.branch_path.len() + 1, leaf_key.len())
```
yongkangc: i see because the branch path length is < key length, so len() + 1 is in-bounds.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+        // Similarly, trim off the common prefix, plus one nibble for the new branch, from the new
+        // leaf's short key.
+        let leaf_nibble = leaf_short_key.get_unchecked(common_prefix_len);
+        let leaf_short_key = trim_nibbles_prefix(&leaf_short_key, common_prefix_len + 1);
+
+        // Push the new leaf onto the child stack; it will be the second child of the new branch.
+        // The new branch's first child is the child already on the top of the stack, for which
+        // we've already adjusted its short key.
+        self.child_stack
+            .push(ProofTrieBranchChild::Leaf { short_key: leaf_short_key, value: leaf_val });
```
yongkangc, line 157: this is why deffered encoding is important right? so we can encode only when we want to do pop_branch

yongkangc (approved):
i dont get 100%, but nothing blocking stands out to me and was able to understand the branching logic


good docs

crates/trie/trie/src/proof_v2/node.rs
yongkangc, line 152: do we want test for this? less for functionality but more to show how trimming nibbles work

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -46,13 +85,269 @@ where
...
+            // directly, otherwise a new branch must be created in-between this branch and that
+            // existing child.
+            let nibble = key.get_unchecked(common_prefix_len);
+            if curr_branch.state_mask.is_bit_set(nibble) {
+                // This method will also push the new leaf onto the `child_stack`.
+                self.push_new_branch(key, val);
+            } else {
+                curr_branch.state_mask.set_bit(nibble);
+
+                // Push the leaf onto the stack.
```
yongkangc: was thinking maybe we can add why we push leaf onto stack

e.g
```suggestion
                // Add this leaf as a new child of the current branch (no intermediate branch needed).
```

crates/trie/trie/src/proof_v2/node.rs
mediocregopher: not sure, it's fairly trivial, was just nice to have here to clarify what's happening

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Yes there are a few:
* For the first two leaves processed the branch_stack will be empty, only after the second leaf is done processing will there be a branch under construction.
* If (for example) the third leaf is then a child of that branch, then that branch will be popped (leaving branch_stack empty) and a new one created with that previous branch+extension as a child and the third leaf as the other child (leaving branch_stack with a single branch again).

In this implementation `pop_branch` is only called in the case where the incoming leaf's shared prefix with the `branch_path` is shorter than the `branch_path`. For that to happen `branch_path` must itself be non-empty, which means there must be a branch on the stack.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Right, we're pushing to the `branch_stack` immediately before this so it will never panic. I'll make this an `expect` to clarify though.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Right, it's impossible that a branch path would have a length of 32, which will always be the key length, so this will always be in-bounds

crates/trie/trie/src/hashed_cursor/mock.rs
shekhirin: probably need to merge main into this branch

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -0,0 +1,157 @@
...
+        /// The node of the child branch.
+        child: BranchNode,
+    },
+    /// A branch node whose children have already been flattened into [`RlpNode`]s.
+    Branch(BranchNode),
+}
+
+impl<RF: DeferredValueEncoder> ProofTrieBranchChild<RF> {
+    /// Converts this child into its RLP node representation. This potentially also returns an
+    /// `RlpNode` buffer which can be re-used for other `ProofTrieBranchChild`s.
```
shekhirin:
```suggestion
    /// `RlpNode` buffer which can be re-used for other [`ProofTrieBranchChild`]s.
```

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Exactly, basically we have the time between pushing a leaf onto the stack and calling `pop_branch` for its branch for the leaf's value to be fully resolved. For the accounts trie this gives us time to potentially dispatch a task to storage workers for missed leaves, so we don't have to do those synchronously like we do currently.

crates/trie/trie/src/proof_v2/node.rs
mediocregopher: Added some basic tests

shekhirin (approved):
Could follow along, really appreciate the comments

PR #19864, opened 2025-11-19
feat(trie): Implement set_hashed_address on mock trie cursors

mediocregopher:
This properly implements the `set_hashed_address` method on the MockTrieCursor and MockHashedCursor. This method was previously left unimplemented, but will be necessary for an upcoming feature.

Implementing this method required changing the internals of the mock cursors a bit, such that they are able to hold onto the full set of storage tries in order to switch between them, similar to how the in-memory cursor overlays do.


mattsse (approved):
mocktastic

pending @yongkangc @shekhirin

yongkangc (approved):
lgtm,

one question - what would the main enablement of this be used for? 

imo seems like we can now simulate switching a storage cursor between accounts in a single test which means we can better test Overlay cursors? feels like i am missing the bigger picture here

PR #19869, opened 2025-11-20
fix: disable state root task when no changesets

Rjected:
It's not safe to run the state root task or parallel state root anyways if there are not merkle changesets, so we disable it when we have no merkle changesets

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -38,6 +38,7 @@ use reth_provider::{
     StateRootProvider, TrieReader,
 };
 use reth_revm::db::State;
+use reth_stages_api::StageId;
 use reth_trie::{updates::TrieUpdates, HashedPostState, TrieInputSorted};
 use reth_trie_parallel::root::{ParallelStateRoot, ParallelStateRootError};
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
Changes makes sense to me

Why is it not safe to run without change sets? Or what might be the case that it doesn't work

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -852,14 +854,21 @@ where
...
     /// Determines the state root computation strategy based on configuration.
-    const fn plan_state_root_computation(&self) -> StateRootStrategy {
-        if self.config.state_root_fallback() {
+    fn plan_state_root_computation(&self) -> ProviderResult<StateRootStrategy> {
+        // If there are no changesets then it's not safe to use the parallel state root or state
+        // root task.
+        let changeset_progress_exists =
+            self.provider.get_stage_checkpoint(StageId::MerkleChangeSets)?.is_some();
+
+        let strategy = if self.config.state_root_fallback() || changeset_progress_exists {
```
shekhirin: i think?
```suggestion
        let strategy = if self.config.state_root_fallback() || !changeset_progress_exists {
```

shekhirin (approved):
LGTM, @mediocregopher please take a look

PR #19870, opened 2025-11-20
feat(cli): add metrics for download command

AJStonewee:
We're downloading 2TB+ mainnet snapshots that take 48+ hours but had zero visibility into what's happening. 

### Before:

`Downloading and extracting... 42.31% (904.23 GB / 2.08 TB)`

That's it. No way to track via Prometheus, no failure stats, nothing.

### After:
```rust
// Now exposed as Prometheus metrics:
cli_download_downloads_started_total: 3
cli_download_downloads_success_total: 2  
cli_download_downloads_failed_total: 1
cli_download_download_speed_bytes_per_second: 125829120  // ~120MB/s
cli_download_download_progress_percent: 42.31
.....
```

Can now properly monitor downloads:
- Track success rates across retries
- Alert on slow speeds (`rate(cli_download_downloaded_bytes_total[1m]) < 10MB`)
- Predict completion time
- Identify if extraction or download is the bottleneck

### DIFF
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -30,14 +30,10 @@ const EXTENSION_TAR_ZSTD: &str = ".tar.zst";
 /// Global static download defaults
 static DOWNLOAD_DEFAULTS: OnceLock<DownloadDefaults> = OnceLock::new();
 
-static DOWNLOAD_METRICS: OnceLock<DownloadMetrics> = OnceLock::new();
-
 /// Metrics for the download command
 #[derive(Metrics, Clone)]
 #[metrics(scope = "cli.download")]
 struct DownloadMetrics {
-    /// Download attempts started
-    downloads_started_total: Counter,
     /// Successful downloads
     downloads_success_total: Counter,
     /// Failed downloads
@@ -54,12 +50,6 @@ struct DownloadMetrics {
     download_file_size_bytes: Gauge,
     /// Progress (0-100)
     download_progress_percent: Gauge,
-}
-
-impl DownloadMetrics {
-    fn global() -> &'static Self {
-        DOWNLOAD_METRICS.get_or_init(DownloadMetrics::default)
-    }
 }
 
 /// Download configuration defaults
@@ -186,8 +176,9 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> DownloadCo
             "Starting snapshot download and extraction"
         );
 
+        let metrics = DownloadMetrics::default();
         let start = Instant::now();
-        let result = stream_and_extract(&url, data_dir.data_dir()).await;
+        let result = stream_and_extract(&url, data_dir.data_dir(), &metrics).await;
 
         match result {
             Ok(_) => {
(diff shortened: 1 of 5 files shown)


crates/cli/commands/src/download.rs
```diff
@@ -26,6 +30,38 @@ const EXTENSION_TAR_ZSTD: &str = ".tar.zst";
 /// Global static download defaults
 static DOWNLOAD_DEFAULTS: OnceLock<DownloadDefaults> = OnceLock::new();
 
+static DOWNLOAD_METRICS: OnceLock<DownloadMetrics> = OnceLock::new();
```
Rjected: I don't think we need a global for this

crates/cli/commands/src/download.rs
```diff
@@ -170,12 +222,27 @@ struct DownloadProgress {
     downloaded: u64,
     total_size: u64,
     last_displayed: Instant,
+    _started_at: Instant,
+    last_speed_update: Instant,
+    last_speed_bytes: u64,
+    metrics: &'static DownloadMetrics,
```
Rjected: can also remove this `&'static` by removing the global

crates/cli/commands/src/download.rs
```diff
@@ -170,12 +222,27 @@ struct DownloadProgress {
     downloaded: u64,
     total_size: u64,
     last_displayed: Instant,
+    _started_at: Instant,
```
Rjected: If this is unused then we should remove it

crates/cli/commands/src/download.rs
```diff
@@ -256,33 +332,57 @@ impl CompressionFormat {
...
-    match format {
-        CompressionFormat::Lz4 => {
-            let decoder = Decoder::new(progress_reader)?;
-            Archive::new(decoder).unpack(target_dir)?;
+    let metrics = DownloadMetrics::global();
+    let overall_start = Instant::now();
+
+    metrics.downloads_started_total.increment(1);
+
+    let result = (|| -> Result<()> {
```
Rjected: I think we can keep it as-is, without the closure

crates/cli/commands/src/download.rs
```diff
@@ -256,33 +332,57 @@ impl CompressionFormat {
     }
 }
 
-/// Downloads and extracts a snapshot, blocking until finished.
```
Rjected, line 259: Let's bring back this comment

Rjected (changes requested):
I think adding metrics for this is reasonable, I have a few suggestions

crates/cli/commands/src/download.rs
mattsse: yeah if this only for one download run, we can just init this as DownloadMetrics::default() in the command itself

crates/cli/commands/src/download.rs
```diff
@@ -256,33 +332,57 @@ impl CompressionFormat {
...
-    let format = CompressionFormat::from_url(url)?;
-
-    match format {
-        CompressionFormat::Lz4 => {
-            let decoder = Decoder::new(progress_reader)?;
-            Archive::new(decoder).unpack(target_dir)?;
+    let metrics = DownloadMetrics::global();
+    let overall_start = Instant::now();
+
+    metrics.downloads_started_total.increment(1);
```
mattsse: this metric is only useful if we make use of this command a lot?

because this command impl only ever does one download

mattsse (changes requested):
unsure if we have have a prometheus supported for this command
so we should check this as well

crates/cli/commands/src/download.rs
AJStonewee: deleted

crates/cli/commands/src/download.rs
AJStonewee: yep removed both - global static is gone and dropped downloads_started_total since it doesn't make sense here

good point about prometheus though, this command doesn't actually expose metrics (no --metrics flag). so kinda pointless right now?

should i just remove them or add the endpoint like stage run has?

mattsse: > should i just remove them or add the endpoint like stage run has?

if the motivation for this pr is to make them observable via prometheus then we should add support for this via a metrics arg imo and launch the prometheus listener if set

AJStonewee: added it - there's now a --metrics flag that starts the listener

copied the pattern from stage run, should work the same way

let me know if this is what you had in mind

PR #19871, opened 2025-11-20
refactor: remove futures_util from exex test utils

Fallengirl:
This removes the `futures_util` dependency from the Execution Extension test helpers. The crate only used it for the `FutureExt::poll_unpin` method in the `poll_once` helper. By switching to `Pin::new(&mut *self).poll(cx)` from the standard library, we achieve the same behavior without the external dependency, reducing build times and eliminating an unnecessary crate from the dependency tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8702,7 +8702,6 @@ version = "1.9.3"
 dependencies = [
  "alloy-eips",
  "eyre",
- "futures-util",
  "reth-chainspec",
  "reth-config",
  "reth-consensus",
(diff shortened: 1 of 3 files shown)


shekhirin (changes requested):
Hmm, given that `reth-exex` crate already depends on `futures`, maybe we can just use it in `reth-exex-test-utils` instead of `futures-util`? Test utils are always used in conjunction with the main ExEx crate.

mattsse (commented):
yeah we can keep this as is

PR #19873, opened 2025-11-20
chore(bench-compare): Add latency distribution stats to reth-bench-compare

yongkangc:
Add percentile-based latency statistics (p50, p90, p99) to provide a better overview of latency. Rename "avg" to "mean" for statistical precision.

### DIFF
--- a/bin/reth-bench-compare/src/comparison.rs
+++ b/bin/reth-bench-compare/src/comparison.rs
@@ -53,12 +53,12 @@ pub(crate) struct TotalGasRow {
 
 /// Summary statistics for a benchmark run.
 ///
-/// Latencies are in milliseconds and describe `engine_newPayloadV3` calls:
-/// - `total_new_payload_latency_ms`: sum across all processed blocks.
-/// - `avg_new_payload_latency_ms`: mean latency per block.
-/// - `median_new_payload_latency_ms`: p50 latency per block.
-/// - `p90_new_payload_latency_ms` / `p99_new_payload_latency_ms`: tail latencies per block.
-/// - Throughput numbers are computed using total duration.
+/// Latencies are derived from per-block `engine_newPayloadV3` timings (converted from µs to ms):
+/// - `total_new_payload_latency_ms`: sum of all blocks' latencies.
+/// - `avg_new_payload_latency_ms`: arithmetic mean latency across blocks.
+/// - `median_new_payload_latency_ms`: p50 latency across blocks.
+/// - `p90_new_payload_latency_ms` / `p99_new_payload_latency_ms`: tail latencies across blocks.
+/// `gas_per_second` and `blocks_per_second` use the total run duration from the gas trace.
 #[derive(Debug, Clone, Serialize)]
 pub(crate) struct BenchmarkSummary {
     pub total_blocks: u64,
--- a/bin/reth-bench-compare/src/comparison.rs
+++ b/bin/reth-bench-compare/src/comparison.rs
@@ -54,19 +54,14 @@ pub(crate) struct TotalGasRow {
 /// Summary statistics for a benchmark run.
 ///
 /// Latencies are derived from per-block `engine_newPayload` timings (converted from µs to ms):
-/// - `total_new_payload_latency_ms`: sum of all blocks' latencies. Used for calculating mean and
-///   provides total execution time spent in `engine_newPayload`.
 /// - `mean_new_payload_latency_ms`: arithmetic mean latency across blocks.
 /// - `median_new_payload_latency_ms`: p50 latency across blocks.
 /// - `p90_new_payload_latency_ms` / `p99_new_payload_latency_ms`: tail latencies across blocks.
-///
-/// `gas_per_second` and `blocks_per_second` use the total run duration from the benchmark.
 #[derive(Debug, Clone, Serialize)]
 pub(crate) struct BenchmarkSummary {
     pub total_blocks: u64,
     pub total_gas_used: u64,
     pub total_duration_ms: u128,
-    pub total_new_payload_latency_ms: f64,
     pub mean_new_payload_latency_ms: f64,
     pub median_new_payload_latency_ms: f64,
     pub p90_new_payload_latency_ms: f64,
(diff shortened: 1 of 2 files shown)


bin/reth-bench-compare/src/comparison.rs
```diff
@@ -50,13 +51,25 @@ pub(crate) struct TotalGasRow {
     pub time: u128,
 }
 
-/// Summary statistics for a benchmark run
+/// Summary statistics for a benchmark run.
+///
+/// Latencies are derived from per-block `engine_newPayloadV3` timings (converted from µs to ms):
```
shekhirin:
```suggestion
/// Latencies are derived from per-block `engine_newPayload` timings (converted from µs to ms):
```

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -50,13 +51,25 @@ pub(crate) struct TotalGasRow {
...
+/// - `total_new_payload_latency_ms`: sum of all blocks' latencies.
+/// - `mean_new_payload_latency_ms`: arithmetic mean latency across blocks.
+/// - `median_new_payload_latency_ms`: p50 latency across blocks.
+/// - `p90_new_payload_latency_ms` / `p99_new_payload_latency_ms`: tail latencies across blocks.
+///
+/// `gas_per_second` and `blocks_per_second` use the total run duration from the gas trace.
```
shekhirin: unsure what does gas trace mean here?

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -82,10 +95,16 @@ pub(crate) struct RefInfo {
     pub end_timestamp: Option<DateTime<Utc>>,
 }
 
-/// Summary of the comparison between references
+/// Summary of the comparison between references.
+///
+/// Percent deltas are `(feature - baseline) / baseline * 100`, so positive is slower/higher and
+/// negative is faster/lower.
```
shekhirin: a bit confusing what does higher and lower mean here


```suggestion
/// Percent deltas are `(feature - baseline) / baseline * 100`, so positive is slower and
/// negative is faster.
```

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -450,15 +496,27 @@ impl ComparisonGenerator {
 
         println!("Performance Changes:");
         println!(
-            "  NewPayload Latency: {:+.2}% (total avg change)",
+            "  NewPayload Latency (mean per block): {:+.2}%",
```
shekhirin: "mean per block" sounds a bit confusing, I think it's more like "mean over blocks"? But generally, can even be just "mean" same as in other print 
```suggestion
            "  NewPayload Latency mean: {:+.2}%",
```

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -473,7 +531,15 @@ impl ComparisonGenerator {
             baseline.total_gas_used,
             baseline.total_duration_ms as f64 / 1000.0
         );
-        println!("  Avg NewPayload: {:.2}ms", baseline.avg_new_payload_latency_ms);
+        println!("  NewPayload latency (ms):");
+        println!(
+            "    total: {:.2}, mean/block: {:.2}, p50: {:.2}, p90: {:.2}, p99: {:.2}",
```
shekhirin:
```suggestion
            "    total: {:.2}, mean: {:.2}, p50: {:.2}, p90: {:.2}, p99: {:.2}",
```

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -495,7 +561,15 @@ impl ComparisonGenerator {
             feature.total_gas_used,
             feature.total_duration_ms as f64 / 1000.0
         );
-        println!("  Avg NewPayload: {:.2}ms", feature.avg_new_payload_latency_ms);
+        println!("  NewPayload latency (ms):");
+        println!(
+            "    total: {:.2}, mean/block: {:.2}, p50: {:.2}, p90: {:.2}, p99: {:.2}",
```
shekhirin:
```suggestion
            "    total: {:.2}, mean: {:.2}, p50: {:.2}, p90: {:.2}, p99: {:.2}",
```

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -50,13 +51,25 @@ pub(crate) struct TotalGasRow {
...
+/// - `p90_new_payload_latency_ms` / `p99_new_payload_latency_ms`: tail latencies across blocks.
+///
+/// `gas_per_second` and `blocks_per_second` use the total run duration from the gas trace.
 #[derive(Debug, Clone, Serialize)]
 pub(crate) struct BenchmarkSummary {
     pub total_blocks: u64,
     pub total_gas_used: u64,
     pub total_duration_ms: u128,
-    pub avg_new_payload_latency_ms: f64,
+    pub total_new_payload_latency_ms: f64,
```
shekhirin: what is it useful for?

yongkangc: removed

bin/reth-bench-compare/src/comparison.rs
yongkangc: clarified thanks!

bin/reth-bench-compare/src/comparison.rs
```diff
@@ -296,9 +315,16 @@ impl ComparisonGenerator {
 
         let total_duration_ms = total_gas_data.last().unwrap().time / 1000; // Convert microseconds to milliseconds
 
-        let avg_new_payload_latency_ms: f64 =
-            combined_data.iter().map(|r| r.new_payload_latency as f64 / 1000.0).sum::<f64>() /
-                total_blocks as f64;
+        let latencies_ms: Vec<f64> =
```
mediocregopher, line 330: Is this value different from `total_duration_ms`? I would expect that this is what `total_duration_ms` is

mediocregopher (approved):
One question but generally lgtm

bin/reth-bench-compare/src/comparison.rs
yongkangc: latencies_ms is summed/aggregated from per-block new_payload_latency values

but total_duration_ms comes from total_gas_data.last().time / 1000

so one measures per-block newPayload latency, the other is total benchmark runtime

mediocregopher: i see, so `total_duration_ms` is effectively the total runtime of the benchmark, including `--wait-time`? In that case I'd almost say let's remove it as a value in the report all-together, it's not useful and could only cause confusion

shekhirin (approved):

PR #19875, opened 2025-11-20
chore(deps): clean up unused imports and dependencies in reth-downloaders

Fallengirl:
Some imports in test_utils/mod.rs were unconditionally imported but only used when the "file-client" feature is enabled or during testing. This adds proper cfg gates to ensure they're only compiled when actually needed.

Changes:
1. Added cfg attributes for feature-gated imports (futures, tokio, std::io)
2. Added explicit tempfile crate usage to satisfy unused_crate_dependencies lint
3. Added alloy-rlp and itertools to dev-dependencies for test compilation

### DIFF
--- a/crates/net/downloaders/src/test_utils/mod.rs
+++ b/crates/net/downloaders/src/test_utils/mod.rs
@@ -5,17 +5,9 @@
 #[cfg(any(test, feature = "file-client"))]
 use crate::{bodies::test_utils::create_raw_bodies, file_codec::BlockFileCodec};
 use alloy_primitives::B256;
-#[cfg(any(test, feature = "file-client"))]
-use futures::SinkExt;
 use reth_ethereum_primitives::BlockBody;
 use reth_testing_utils::generators::{self, random_block_range, BlockRangeParams};
-#[cfg(any(test, feature = "file-client"))]
-use std::io::SeekFrom;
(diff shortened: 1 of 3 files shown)


crates/net/downloaders/src/test_utils/mod.rs
```diff
@@ -5,11 +5,16 @@
...
 use futures::SinkExt;
 use reth_ethereum_primitives::BlockBody;
 use reth_testing_utils::generators::{self, random_block_range, BlockRangeParams};
-use std::{collections::HashMap, io::SeekFrom, ops::RangeInclusive};
+#[cfg(any(test, feature = "file-client"))]
+use std::io::SeekFrom;
+use std::{collections::HashMap, ops::RangeInclusive};
+#[cfg(any(test, feature = "file-client"))]
 use tokio::{fs::File, io::AsyncSeekExt};
+#[cfg(any(test, feature = "file-client"))]
```
mattsse: can we move these feature gated imports into the functions that use them?

mattsse (changes requested):
will allow if we can move the feature gated imports into the functions that require them

Fallengirl:
@mattsse pinging

mattsse (approved):

Issue #19874, opened 2025-11-20
Customize genesis block number


JimmyShi22:
## Support for Custom Genesis Block Number

Currently, Reth only supports initializing the genesis block at block number 0. We would like to request support for custom genesis block numbers by reading the `number` field from the `genesis.json` file.

### Use Case

**Blockchain Data Migration Scenario**

When migrating blockchain data from one chain to another, the `genesis.json` file is typically dumped from an existing chain at a specific block height. In this case, we need the new chain to continue producing blocks from that exact block height, rather than starting from 0.

For example:
1. Export the state and genesis configuration from an existing chain at block height 1000
2. Initialize a new Reth node with this dumped `genesis.json`
3. The new chain should continue from block 1001, 1002, etc., maintaining continuity with the original chain's block numbering

Without this feature, the block numbers would reset to 0, breaking the continuity and causing issues with:
- Block explorers and indexers that rely on consistent block numbering
- Smart contracts or applications that depend on specific block numbers
- Historical data integrity and traceability

### Proposed Implementation

The `genesis.json` file already contains a `number` field. The proposal is to:

1. Read the `number` field from `genesis.json` during initialization
2. If `number` is non-zero, use that value as the genesis block height instead of defaulting to 0
3. Maintain backward compatibility by defaulting to 0 when the field is not specified or is 0

Example `genesis.json` dumped from an existing chain:
``` json
{
  "number": "0x3e8",
  "timestamp": "0x64a1f2b0",
  "extraData": "0x",
  "gasLimit": "0x1000000",
  "stateRoot": "0x...",
  ...
}
```
In this example, the genesis block would be initialized at block number 1000 (0x3e8), and the chain would continue producing blocks 1001, 1002, etc.

### Motivation for Support

We are currently working with a forked version of Reth for our project. To keep our fork as close to upstream as possible and make it easier to sync with the latest Reth updates, we hope this feature can be officially supported rather than maintaining it as a custom patch.

We believe this feature would be valuable not only for our use case but also for other teams working with Reth in similar scenarios. 

If it is possible, we are happy to submit a PR to this issue.

JimmyShi22:
Thanks for the reply! Yes, this is a feasible way, we have tried it locally and it does take up a certain amount of disk space. There may be a better way to start generating blocks directly from the current block height, and this is the solution we are considering: https://github.com/paradigmxyz/reth/pull/19877

PR #19877, opened 2025-11-20
feat: support non-zero genesis block numbers

Vui-Chee:
## Add support for non-zero genesis block numbers

Adds support for chains with genesis blocks that don't start at block 0, enabling compatibility with custom chain configurations.

Attempts to address issue described in #19874.

### Key Changes

- **Chainspec**: Genesis header now includes `number` and `parent_hash` from chain spec
- **Static File Provider**: Added genesis block number tracking and getter/setter methods
- **Provider Factory**: Added `with_genesis_block_number()` builder method with documentation
- **Initialization**: Updated genesis initialization to use actual genesis block number instead of hardcoded 0
- **Static File Writer**: Added `append_header_direct()` for non-zero genesis blocks and updated block range calculations
- **CLI/Node Builder**: Automatically configure genesis block number from chain spec

This ensures all database operations, static file operations, and stage checkpoints correctly handle chains with non-zero genesis blocks.


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -311,6 +311,8 @@ pub struct StaticFileProviderInner<N> {
     blocks_per_file: HashMap<StaticFileSegment, u64>,
     /// Write lock for when access is [`StaticFileAccess::RW`].
     _lock_file: Option<StorageLock>,
+    /// Genesis block number, default is 0;
+    genesis_block_number: u64,
```
mattsse, line 329: can you elaborate on why we need this here?

crates/chainspec/src/spec.rs
```diff
@@ -73,6 +73,8 @@ pub fn make_genesis_header(genesis: &Genesis, hardforks: &ChainHardforks) -> Hea
         .then_some(EMPTY_REQUESTS_HASH);
 
     Header {
+        number: genesis.number.unwrap_or_default(),
+        parent_hash: genesis.parent_hash.unwrap_or_default(),
```
mattsse, line 84: this makes sense

mattsse (commented):
supportive in general, but would like to hear more about where and how this is currently causing issues, because I'm unsure why we need this on the db level 

for this I'd also like to see an e2e that uses a modified genesis block

crates/storage/provider/src/providers/static_file/manager.rs
Vui-Chee: When static files are empty, we need this to reliably fallback to custom genesis number.

Vui-Chee:
> > different blockchain to reth
> 
> welcome frens, and thx for the prs thus far.
> 
> as mentioned, not opposed to this in general, but I'd like to avoid doing this if feasible so please get back to us on [#19874 (comment)](https://github.com/paradigmxyz/reth/issues/19874#issuecomment-3561108254)
> 
> because I'd prefer if we could avoid introducing additional corner cases in the static file logic.
> 
> I assume you'd need to do some form of regenesis with state snapshot anyway then?

I will continue this conversion in #19874, but will this PR opened till we come to an mutual understanding.

mattsse (commented):
mind taking a look at this @joshieDo @shekhirin 

adding genesis handling doesnt look too bad imo

joshieDo (approved):
hard to know if there is any assumption that we're missing, but since we're defaulting to 0, should be fine.

I wonder if we can use this also to initialize the op-mainnet without bedrock instead of creating the dummy chain.

would still like another sanity check from @shekhirin

mattsse (approved):
this isnt too terrible, and I have a feeling that we'll need smth like this eventually anyway

crates/storage/db-common/src/init.rs
```diff
@@ -179,15 +184,26 @@ where
     // compute state root to populate trie tables
     compute_state_root(&provider_rw, None)?;
 
-    // insert sync stage
+    // set stage checkpoint to genesis block number for all stages
+    let checkpoint = StageCheckpoint { block_number: genesis_block_number, ..Default::default() };
```
shekhirin:
```suggestion
    let checkpoint = StageCheckpoint::new(genesis_block_number);
```

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -409,6 +412,11 @@ impl<N: NodePrimitives> StaticFileProviderInner<N> {
             block,
         )
     }
+
+    /// Get genesis block number
+    pub const fn get_genesis_block_number(&self) -> u64 {
```
shekhirin:
```suggestion
    pub const fn genesis_block_number(&self) -> u64 {
```

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1645,6 +1653,15 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
             .map(|index| &index.expected_block_ranges_by_max_block)
             .cloned()
     }
+
+    /// Set genesis block number.
+    pub fn set_genesis_block_number(&mut self, genesis_block_number: u64) {
```
shekhirin: should we move this to `StaticFileProviderBuilder` level instead, so that it can only be changed when building a new provider?

shekhirin (commented):
2 nits and one question, overall looks correct

crates/storage/provider/src/providers/static_file/manager.rs
Vui-Chee: done.

crates/storage/provider/src/providers/static_file/manager.rs
Vui-Chee: Sounds good. Done in [822819a](https://github.com/paradigmxyz/reth/pull/19877/commits/822819af58815611bcd5b4cf3ca9c5be628edc38).

shekhirin (approved):

PR #19878, opened 2025-11-20
refactor: tx policy traits

klkvr:
This refactors transaction policy traits and makes them dyn-compatible which allowed to remove an extra generic from `TransactionsManager`

More importantly, this changes `SignedTransaction` trait to enforce `IsTyped2718` supertrait, and changes `AnnouncementFilteringPolicy` implementations so that txtype validity checks are done by checking `N::PooledTransaction::is_type` instead of a hardcoded Ethereum `TxType`

This allows to properly handle networking for chains with custom txtypes

### DIFF
--- a/crates/net/network/src/transactions/config.rs
+++ b/crates/net/network/src/transactions/config.rs
@@ -1,4 +1,5 @@
-use std::{fmt::Debug, marker::PhantomData, str::FromStr};
+use core::fmt;
+use std::{fmt::Debug, str::FromStr};
 
 use super::{
     PeerMetadata, DEFAULT_MAX_COUNT_TRANSACTIONS_SEEN_BY_PEER,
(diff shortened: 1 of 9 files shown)


crates/net/network/src/transactions/config.rs
```diff
@@ -127,17 +127,19 @@ impl Default for TransactionFetcherConfig {
 }
 
 /// A policy defining which peers pending transactions are gossiped to.
-pub trait TransactionPropagationPolicy: Send + Sync + Unpin + 'static {
+pub trait TransactionPropagationPolicy<N: NetworkPrimitives>:
```
mattsse, line 130: this is better

mattsse (approved):
right, we can fully lift the N: type to trait generic

PR #19881, opened 2025-11-20
chore(payload): remove redundant Debug bounds

phrwlk:
Remove explicit + Debug bounds from PayloadJob::PayloadAttributes and PayloadJob::BuiltPayload in crates/payload/builder/src/traits.rs because BuiltPayload and PayloadBuilderAttributes already require fmt::Debug per crates/payload/primitives/src/traits.rs. This duplication provides no additional guarantees, may trigger clippy redundancy lints, and slightly increases API noise.


mattsse (commented):
we can keep these

PR #19887, opened 2025-11-20
fix: preserve basic nodes from peers file by avoiding peer_config overwrite

Fibonacci747:
The builder in NetworkArgs::network_config() previously called peer_config twice, where the second assignment rebuilt PeersConfig from a raw clone and overwrote the config that had already included basic nodes from the peers file via peers_config_with_basic_nodes_from_file, resulting in persisted peers being dropped on startup. This change composes peers_config from the file-enriched configuration first, then applies inbound/outbound and IP filter overrides, and sets peer_config only once so persisted peers are preserved as documented. The unit test ensures a NodeRecord present in a temporary peers file is included in the final NetworkConfig’s peers_config.basic_nodes. The test also uses reth_network::EthNetworkPrimitives to avoid an unresolved crate, and a redundant clone warning in network_config was addressed.


crates/node/core/src/args/network.rs
```diff
@@ -288,17 +288,15 @@ impl NetworkArgs {
...
             .with_ip_filter(ip_filter);
 
         // Configure basic network stack
         NetworkConfigBuilder::<N>::new(secret_key)
-            .peer_config(config.peers_config_with_basic_nodes_from_file(
-                self.persistent_peers_file(peers_file).as_deref(),
-            ))
             .external_ip_resolver(self.nat)
             .sessions_config(
                 SessionsConfig::default().with_upscaled_event_buffer(peers_config.max_peers()),
```
mattsse, line 304: oh I see, we indeed overwrite this

mattsse (approved):
nice find, ty!

PR #19888, opened 2025-11-20
chore: dont treat invalid fork as fatal

mattsse:
see docs, 

we shouldn't treat this as fatal and only tmp backoff here

### DIFF
--- a/crates/net/network/src/error.rs
+++ b/crates/net/network/src/error.rs
@@ -114,8 +114,12 @@ impl SessionError for EthStreamError {
                 P2PHandshakeError::NonHelloMessageInHandshake,
             )) => true,
             Self::EthHandshakeError(err) => {
+                #[allow(clippy::match_same_arms)]
                 match err {
-                    EthHandshakeError::NoResponse => false,
+                    EthHandshakeError::NoResponse => {
+                        // this happens when the conn simply stalled
+                        false
+                    }
                     EthHandshakeError::InvalidFork(_) => {
                         // this can occur when the remote or our node is running an outdated client,
                         // we shouldn't treat this as fatal, because the node can come back only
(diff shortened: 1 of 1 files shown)


crates/net/network/src/error.rs
```diff
@@ -113,7 +113,18 @@ impl SessionError for EthStreamError {
                 P2PHandshakeError::HelloNotInHandshake |
                 P2PHandshakeError::NonHelloMessageInHandshake,
             )) => true,
-            Self::EthHandshakeError(err) => !matches!(err, EthHandshakeError::NoResponse),
+            Self::EthHandshakeError(err) => {
+                match err {
+                    EthHandshakeError::NoResponse => false,
+                    EthHandshakeError::InvalidFork(_) => {
+                        // this can occur when the remote or our node is running an outdated client,
+                        // we shouldn't treat this as fatal, because the node can come back only
```
Rjected:
```suggestion
                        // we shouldn't treat this as fatal, because the node can come back online
```

crates/net/network/src/error.rs
```diff
@@ -144,7 +155,18 @@ impl SessionError for EthStreamError {
                         P2PStreamError::MismatchedProtocolVersion { .. }
                 )
             }
-            Self::EthHandshakeError(err) => !matches!(err, EthHandshakeError::NoResponse),
+            Self::EthHandshakeError(err) => {
+                match err {
+                    EthHandshakeError::NoResponse => false,
+                    EthHandshakeError::InvalidFork(_) => {
+                        // this can occur when the remote or our node is running an outdated client,
+                        // we shouldn't treat this as fatal, because the node can come back only
```
Rjected:
```suggestion
                        // we shouldn't treat this as fatal, because the node can come back online
```

crates/net/network/src/error.rs
```diff
@@ -113,7 +113,22 @@ impl SessionError for EthStreamError {
...
+                match err {
+                    EthHandshakeError::NoResponse => {
+                        // this happens when the conn simply stalled
+                        false
+                    }
+                    EthHandshakeError::InvalidFork(_) => {
+                        // this can occur when the remote or our node is running an outdated client,
+                        // we shouldn't treat this as fatal, because the node can come back only
+                        // with an updated version any time
+                        false
```
Rjected, line 127: yeah forkid mismatch should probably not be a disc ban

Rjected (approved):
makes sense to me, just have some typo fixes

PR #19890, opened 2025-11-20
chore(rpc-eth-types): use `FillTransaction` from alloy

mablr:
Smol follow-up for #19199 and https://github.com/alloy-rs/alloy/pull/3210

Bumped alloy to 1.1.2


mattsse (approved):
lgtm!

Issue #19882, opened 2025-11-20
Add snapclient implementation for Fetchclient


mattsse:
### Describe the feature

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/p2p/src/snap/client.rs#L23-L23

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/fetch/client.rs#L35-L35

this can be a noop impl for now but just to keep the scope of this small

@0xKarl98

mattsse:
Implemented in #26049. FetchClient now implements SnapClient with account-range, storage-range, bytecode, and block-access-list requests.

Issue #19883, opened 2025-11-20
Integrate snap-sync protocol in p2p stack


mattsse:
we already have support for the protocol itself

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/eth-wire/src/eth_snap_stream.rs#L60-L60

we need to integrate this into the p2p stack:

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/session/active.rs#L93-L93

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/session/conn.rs#L25-L38

for this we need new variant for the ethsnap stream

since this stream then bubbles up new snap messages, we need to rout them accordingly

for example we need new request/response variants:

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network-api/src/events.rs#L184-L186

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/message.rs#L84-L84

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/message.rs#L58-L59

all of this is similar to how we route eth requests, 
incoming requests eventually end up at this type:

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/eth_requests.rs#L55-L55

via:
https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/eth_requests.rs#L306-L306

for snap we can repurpose this type so that it can also handle snaprequests, for now just noops (empty responses).

snap support must be opt-in for now, and should be configured via the config:
protocols are stored in this set:
https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/eth-wire/src/hello.rs#L33-L34

so we need another helper setter here

https://github.com/paradigmxyz/reth/blob/d550e4eb07e677733b31d5dcddb8a31de501026a/crates/net/network/src/config.rs#L331-L334

that is like `fn set_snap_protocols` or `with_snap(bool){self.set_snap_protocols}`

[...]

PR #19893, opened 2025-11-21
feat: add snapclient implementation for Fetchclient

0xKarl98:
This closes issue  #19882 #19883 
Changes as bellow : 
- Introduce a snap capability toggle: with_snap on HelloMessage/NetworkConfigBuilder to optionally announce snap/1 during handshake 
- Extend the protocol stack to snap: use EthSnapStream as the combined eth+snap stream; add snap request/response enums and route them via PeerMessage::SnapRequest 
- Added snap RawCapability encoding tests and snap toggle test 
- Snap requests currently return noop/empty responses to keep the pipeline stable while integration lands 

cc @mattsse


crates/net/network/src/message.rs
```diff
@@ -57,6 +63,8 @@ pub enum PeerMessage<N: NetworkPrimitives = EthNetworkPrimitives> {
     PooledTransactions(NewPooledTransactionHashes),
     /// All `eth` request variants.
     EthRequest(PeerRequest<N>),
+    /// All `snap` request variants (wrapped in `PeerRequest`).
```
mattsse: do we even need this additional variant if we added snap variants to peerrequest?

crates/net/network/src/config.rs
```diff
@@ -653,6 +663,10 @@ impl<N: NetworkPrimitives> NetworkConfigBuilder<N> {
 
         let mut hello_message =
             hello_message.unwrap_or_else(|| HelloMessage::builder(peer_id).build());
+
+        if snap_enabled {
+            hello_message = hello_message.with_snap();
+        }
```
mattsse, line 669: cool, this should do it

crates/net/network/src/session/mod.rs
```diff
@@ -1151,22 +1151,43 @@ async fn authenticate_stream<N: NetworkPrimitives>(
...
-        {
-            Ok((multiplex_stream, their_status)) => (multiplex_stream, their_status),
-            Err(err) => {
-                return PendingSessionEvent::Disconnected {
-                    remote_addr,
-                    session_id,
-                    direction,
-                    error: Some(PendingSessionHandshakeError::Eth(err)),
+        let supports_snap =
+            multiplex_stream.shared_capabilities().iter_caps().any(|cap| cap.name() == "snap");
```
mattsse: can we add a helper type for this on sharecap?

crates/net/network/src/session/conn.rs
```diff
@@ -109,59 +135,109 @@ impl<N: NetworkPrimitives> From<EthSatelliteConnection<N>> for EthRlpxConnection
...
+            }
+            (EthRlpxConnProj::SnapSatellite(mut conn), msg) => {
+                conn.start_send_unpin(msg).map_err(Into::into)
+            }
+            // Sending snap message over eth-only connection is unsupported
+            (
+                EthRlpxConnProj::EthOnly(_) | EthRlpxConnProj::Satellite(_),
+                EthSnapMessage::Snap(_),
+            ) => Err(EthStreamError::UnsupportedMessage {
+                message_id: SnapMessageId::GetAccountRange as u8,
```
mattsse: I believe this should fail regardless

crates/net/network/src/session/conn.rs
```diff
@@ -109,59 +135,109 @@ impl<N: NetworkPrimitives> From<EthSatelliteConnection<N>> for EthRlpxConnection
...
+        }
     }
 }
 
-impl<N: NetworkPrimitives> Sink<EthMessage<N>> for EthRlpxConnection<N> {
+impl<N: NetworkPrimitives> Sink<EthSnapMessage<N>> for EthRlpxConnection<N> {
     type Error = EthStreamError;
 
     fn poll_ready(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        delegate_call!(self.poll_ready(cx))
```
mattsse, line 135: why can we no longer use this macro?

crates/net/eth-wire/src/multiplex.rs
```diff
@@ -142,6 +147,7 @@ impl<St> RlpxProtocolMultiplexer<St> {
         mut self,
         cap: &Capability,
         handshake: F,
+        additional_caps: Vec<SharedCapability>,
```
mattsse: why is this needed now?

mattsse (changes requested):
this looks amazing already

I have a few more questions and smol requests

still need to look into this in detail

crates/net/network/src/message.rs
0xKarl98: You are right , agree that this is redundant 

The snap requests would flow the same if we just sent them via PeerMessage::EthRequest because the inner PeerRequest discriminates the snap cases.

crates/net/network/src/session/mod.rs
0xKarl98: good idea

crates/net/network/src/session/conn.rs
0xKarl98: I think dropping it is a right call , as delegate_call! is for single-field delegation , can’t express the current per-variant behavior  .

crates/net/eth-wire/src/multiplex.rs
0xKarl98: We need it so the primary stream can own more than one capability (e.g., eth + snap). additional_caps tells the multiplexer to route those extra capabilities to the primary; without it, only the main cap goes to primary and snap messages would be treated as “non-primary” and misrouted.

crates/net/network/src/session/active.rs
```diff
@@ -850,6 +1033,8 @@ enum RequestState<R> {
 pub(crate) enum OutgoingMessage<N: NetworkPrimitives> {
     /// A message that is owned.
     Eth(EthMessage<N>),
+    /// A snap protocol message (wrapped in `EthSnapMessage`).
+    Snap(EthSnapMessage<N>),
```
mattsse: could we maybe flatten this to just

SnapProtocolMessage, imo then this makes it a bit easier to reason about

crates/net/network/src/session/active.rs
```diff
@@ -860,7 +1045,7 @@ impl<N: NetworkPrimitives> OutgoingMessage<N> {
     /// Returns true if this is a response.
     const fn is_response(&self) -> bool {
         match self {
-            Self::Eth(msg) => msg.is_response(),
+            Self::Eth(msg) | Self::Snap(EthSnapMessage::Eth(msg)) => msg.is_response(),
```
mattsse: we also need `is_response` for

https://github.com/paradigmxyz/reth/blob/main/crates/net/eth-wire-types/src/snap.rs#L192-L192

crates/net/network/src/session/active.rs
```diff
@@ -872,6 +1057,12 @@ impl<N: NetworkPrimitives> From<EthMessage<N>> for OutgoingMessage<N> {
     }
 }
 
+impl<N: NetworkPrimitives> From<EthSnapMessage<N>> for OutgoingMessage<N> {
+    fn from(value: EthSnapMessage<N>) -> Self {
+        Self::Snap(value)
```
mattsse: here we can do a simple match and then only put the snap variant into OutgoingMessage::Snap

crates/net/network/src/session/active.rs
```diff
@@ -313,8 +462,35 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
         let request_id = self.next_id();
 
         trace!(?request, peer_id=?self.remote_peer_id, ?request_id, "sending request to peer");
-        let msg = request.create_request_message(request_id);
-        self.queued_outgoing.push_back(msg.into());
+        let outgoing = match &request {
```
mattsse: this should probably consume the type, so that we can avoid the clone?

mattsse (changes requested):
cool,

a few more suggestions, need to take a look at the satallite changes in greaer detail

crates/net/network/src/session/active.rs
0xKarl98: Yeah have looked at it again , we can definitely flatten it

crates/net/network/src/session/active.rs
0xKarl98: Have reduced the unnessary clone

crates/net/network/src/message.rs
```diff
@@ -208,6 +248,50 @@ impl<N: NetworkPrimitives> PeerResponseResult<N> {
...
                 to_message!(resp, Receipts69, id)
             }
+            Self::SnapAccountRange(resp) => match resp {
+                Ok(resp) => {
+                    let snap = SnapProtocolMessage::AccountRange(resp);
+                    let encoded = snap.encode();
+                    Ok(EthMessage::Other(RawCapabilityMessage::new(
+                        snap.message_id() as usize,
+                        encoded.slice(1..).to_vec().into(),
+                    )))
```
mattsse: this feels more like a hack than the right solution,

what if we add another function that returns `EthSnapMessage` instead?

crates/net/eth-wire/src/multiplex.rs
mattsse: I believe we can avoid this if we check the inverse here, which is if the protocol doesn't match any of the multiplexed ones:

https://github.com/paradigmxyz/reth/blob/main/crates/net/eth-wire/src/multiplex.rs#L296-L297

we also route it to the primary

then we don't need any additional handling in the sessionmanager I believe

mattsse (changes requested):
very nice,

I only two last suggestions

crates/net/network/src/message.rs
0xKarl98: Have added one , but is this one what you have in mind ?

crates/net/network-api/src/events.rs
mattsse: could we perhaps change this to ethsnapmessage? then we dont need to awkwardly use raw encoding for the snap variants?

crates/net/network/src/session/active.rs
```diff
@@ -216,113 +224,301 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
         }
 
         match msg {
-            message @ EthMessage::Status(_) => OnIncomingMessageOutcome::BadMessage {
-                error: EthStreamError::EthHandshakeError(EthHandshakeError::StatusNotInHandshake),
-                message,
+            EthSnapMessage::Eth(msg) => match msg {
```
mattsse: can we perhaps extract this to on_incoming_eth_message?

crates/net/network/src/session/active.rs
```diff
@@ -856,6 +1061,8 @@ enum RequestState<R> {
 pub(crate) enum OutgoingMessage<N: NetworkPrimitives> {
     /// A message that is owned.
     Eth(EthMessage<N>),
+    /// A snap protocol message.
+    SnapProtocolMessage(EthSnapMessage<N>),
```
mattsse: this is kinda duplicated, so either we change the eth variant to EthSnapMessage or change SnapProtocolMessage to snapmessage?

mattsse (changes requested):
a few more style suggestions, but I think this is almost ready!

0xKarl98:
Think we can re-open another PR to inherit the code behaviour of this one , so closed

PR #19894, opened 2025-11-21
perf(trie): compute and sort trie inputs async

### LINKED ISSUE #19250: Compute trie input asynchronously
### Describe the feature

`compute_trie_input` is used to compute the in-memory overlay which will be passed to the MultiProofTask so that proof computation can take into account in-memory blocks which haven't been persisted to the database yet.

https://github.com/paradigmxyz/reth/blob/4548209e7b00351a993779820b78db86b8628c9b/crates/engine/tree/src/tree/payload_validator.rs#L999-L1006

This step takes a few milliseconds on every block:

Image

Once #18906 is complete this method will no longer rely on the state of persistence to perform its computation, as it will no longer deal with collecting reverts.

At that point it will be possible to not compute the trie input at the start of a block's payload validation, but instead at the end of its parent, after the RPC response has been returned to the caller.

This will likely involve modifying ExecutedBlock to retain the full trie input. It will also require that when in-memory blocks are removed from the tree the trie input of each block is re-computed, as the trie input of one in-memory block is dependent on the trie input of its parent.

### Additional context

_No response_

yongkangc:
Addresses #19250 by moving both trie sorting and trie input construction out of the hot path; validation can return once the state root is checked.

### Changes:
- We stop building trie input on the validation hot path. `validate_block_with_state` now returns an `ExecutedBlock` with a pending `DeferredTrieData`, and offloads the sorting and trie input building to `spawn_blocking`.
- That background task then reconstructs the ancestor overlay (using in-memory blocks), sorts the state and trie updates, builds the trie input, and stores it in the handle. Callers only wait when they actually need the sorted trie data.

### Approach
- Keep a single shared result guarded by `Mutex<DeferredState>`; first caller to `wait_cloned()` while pending computes and caches the result.
- Store unsorted inputs in `DeferredTrieData::pending`: `Arc<HashedPostState>`, `Arc<TrieUpdates>`, `anchor_hash`, and ancestor `DeferredTrieData` handles. This enables synchronous fallback computation with identical logic to the async task.
- Background task calls `wait_cloned()` which sorts the current block's state/updates, merges ancestor trie data (by calling their `wait_cloned()`), builds `ComputedTrieData` with `AnchoredTrieInput`, caches it, and returns.
- Fallback path: if a consumer calls `wait_cloned()` before the background task finishes, it acquires the lock, finds `Pending` state, computes synchronously from stored inputs, caches into `Ready`, and returns. Deadlock-free because ancestors form a DAG.
- Ancestors are referenced via `DeferredTrieData` handles so the hot path only clones lightweight handles; actual trie data is retrieved during compute via recursive `wait_cloned()` calls.

```
                        DeferredTrieData
                  ┌─────────────────────────┐
                  │  Mutex<DeferredState>   │
                  │  - Pending(inputs)      │
                  │  - Ready(bundle)        │
                  └───────────┬─────────────┘
                              │
        ┌─────────────────────┼─────────────────────┐
        │                     │                     │
   Validation             Background            Consumer
   (hot path)               Task              (DB/proofs)
   ──────────             ──────────          ──────────
   create pending         spawn_blocking      trie_data()
   return immediately          │                   │
        │                      ▼                   ▼
        │               wait_cloned()        wait_cloned()
        │                      │                   │
        │                      └─────────┬─────────┘
        │                                │
        │                         ┌──────▼──────┐
        │                         │ lock mutex  │
        │                         │             │
        │                         │ Ready?      │
        │                         │  → clone    │
        │                         │             │
        │                         │ Pending?    │
        │                         │  → compute  │
        │                         │  → cache    │
        │                         │  → return   │
        │                         └─────────────┘
        │
        └── First caller while Pending computes & caches;
            subsequent callers get cached result.
            Mutex ensures single computation.
```

### Flow
- Hot path: execute block → verify state root → gather unsorted state/updates + anchor hash + ancestor handles → create `DeferredTrieData::pending(...)` → spawn background task → return `ExecutedBlock` immediately.
- Background task: call `wait_cloned()` which sorts state/updates → merges ancestor trie data (via their `wait_cloned()`) → builds overlay → caches `ComputedTrieData` → returns.
- Synchronous fallback: if a consumer calls `wait_cloned()` before background completes, the same computation runs inline. The mutex ensures only one computation occurs; subsequent callers get the cached result.
- Consumers: call `trie_data()` which internally calls `wait_cloned()`; they either get the cached result or trigger fallback computation.

### Impact
- seeing a 3 - 5% improvement on base
    - 1000 block benches on base for a set of problematic blocks: https://github.com/paradigmxyz/reth/pull/19894#issuecomment-3569724316
- Latest benches: https://github.com/paradigmxyz/reth/pull/19894#issuecomment-3585906638

Distinct drop in block validation overhead, especially in p50 conditions. This is a set of problematic xen blocks:
image

---
### Other Approaches Considered?

Why not oneshot? 
- We need "set once, read many, cache for late readers" and sometimes block in sync contexts - e.g., DB writes or state providers) that need the trie data but don't run on a Tokio runtime. A oneshot is single-producer/single-consumer, gets consumed on first recv, and is async-only. It would still require a cache and sync wait path, adding complexity here.

Why not Mutex + Condvar?
- Using `Mutex` + `Condvar` requires some scaffolding code to handle errors and conditions. With Onelock, we could simplify the handling with blocking wait.

Why not Oncelock?
- OnceLock::wait() ties up the workers.
    - i.e [parking_lot::Condvar::wait](https://docs.rs/parking_lot/latest/parking_lot/struct.Condvar.html#method.wait) releases the mutex and that yields worker so other tasks like spawn blocking can run. [source](https://docs.rs/parking_lot/latest/src/parking_lot/condvar.rs.html#198-234). But with [OnceLock::wait()](https://doc.rust-lang.org/src/std/sync/once_lock.rs.html#200-204), it delegates to Once::wait which enqueues the waiter and calls thread::park() without releasing anything. [source](https://github.com/rust-lang/rust/blob/1ebbd87a62ce96a72b22da61b7c2c43893534842/library/std/src/sys/sync/once/futex.rs#L65-L108)


crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,221 @@
+use alloy_primitives::B256;
+use parking_lot::{Condvar, Mutex};
+use reth_trie::{updates::TrieUpdatesSorted, HashedPostStateSorted, TrieInputSorted};
+use std::{error::Error, fmt, sync::Arc};
+
+/// Sorted trie data computed for an executed block.
+/// These represent the complete set of sorted trie data required to persist
+/// block state and proof generation for a block.
```
mediocregopher:
```suggestion
/// block state for, and generate proofs on top of, a block.
```

crates/chain-state/src/in_memory.rs
```diff
@@ -719,29 +719,37 @@ impl<N: NodePrimitives> BlockState<N> {
...
             recovered_block: Default::default(),
             execution_output: Default::default(),
-            hashed_state: Default::default(),
-            trie_updates: Default::default(),
+            trie_data: DeferredTrieData::ready(ComputedTrieData::default()),
         }
     }
 }
 
+impl<N: NodePrimitives> PartialEq for ExecutedBlock<N> {
```
mediocregopher, line 751: In what situations is PartialEq/Eq for ExecutedBlock used? Could we just not support it?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1791,11 +1793,28 @@ where
...
+        let trie_input = Arc::new(TrieInputSorted::new(
+            Arc::clone(&sorted_trie_updates),
+            Arc::clone(&sorted_hashed_state),
+            prefix_sets,
+        ));
+        let trie_data = ComputedTrieData {
+            hashed_state: sorted_hashed_state,
+            trie_updates: sorted_trie_updates,
+            anchor_hash: block.parent_hash(),
+            trie_input,
```
mediocregopher: I think given this use-case it might make sense to stick anchor_hash and trie_input inside a shared Option, so we can simply opt to not generate them for cases like this where it doesn't make sense to.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +526,100 @@ where
...
+                            nodes_mut.extend_ref(data.trie_updates.as_ref());
+                        }
+                    }
+                }
+
+                let prefix_sets = hashed_state_for_trie.construct_prefix_sets();
+                let sorted_hashed_state = Arc::new(hashed_state_for_trie.into_sorted());
+                let sorted_trie_updates = Arc::new(trie_output_for_trie.into_sorted());
+
+                parent_trie_input.prefix_sets.extend(prefix_sets);
```
mediocregopher: The TrieInput for an ExecutedBlock shouldn't have a prefix set, it's never used, so no need to calculate it.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +526,100 @@ where
...
+                    if blocks_iter.peek().is_some() {
+                        let state_mut = Arc::make_mut(&mut parent_trie_input.state);
+                        let nodes_mut = Arc::make_mut(&mut parent_trie_input.nodes);
+                        for block in blocks_iter {
+                            let data = block.trie_data();
+                            state_mut.extend_ref(data.hashed_state.as_ref());
+                            nodes_mut.extend_ref(data.trie_updates.as_ref());
+                        }
+                    }
+                }
```
mediocregopher: Since this is largely copy-pasted from `compute_trie_input`, maybe we could break it out into a function which can be shared?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +526,100 @@ where
...
+        let deferred_compute_duration =
+            self.metrics.block_validation.deferred_trie_compute_duration.clone();
+
+        // Defer trie sorting and compute trie input to a background task so the validation hot path
+        // can return once the state root is checked. Consumers (DB writes, overlay
+        // providers, proofs) block on the handle only when they actually need the sorted
+        // trie data.
+        let task = move || {
+            let compute_start = Instant::now();
+            let result = std::panic::catch_unwind(AssertUnwindSafe(|| {
```
mediocregopher: What happens if the task panics and we don't catch it like this? If the whole process panics then I think that's acceptable, nothing in here should be panicking and there's a ton of extra code added here to handle it, imo it'd be better to have simpler code.

Put another way, if something in here is panicking then the logic is so broken it's probably better the process doesn't continue.

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +526,100 @@ where
...
+                        %block_hash,
+                        "deferred trie task panicked while sorting/assembling trie data"
+                    );
+                    deferred_handle_task.set_error(DeferredTrieDataError::panicked());
+                    deferred_compute_duration.record(compute_start.elapsed().as_secs_f64());
+                }
+            }
+        };
+
+        if let Ok(handle) = Handle::try_current() {
```
mediocregopher: Inside `BasicEngineValidator::new` we initialize the PayloadProcessor using a default WorkloadExecutor. I think we could keep a clone of that WorkloadExecutor on the BasicEngineValidator and use it here, rather than doing this.

crates/engine/tree/src/tree/payload_validator.rs
mediocregopher, line 938: What this method should be doing is checking the `blocks.last()` block to see if its `anchor_hash` matches the `block_hash` here, and if it does it should reuse the TrieInputSorted completely, so we don't have to calculate it here at all.

If it doesn't match then we continue on to the logic below which re-calculates the TrieInputSorted.

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub hashed_state: Arc<HashedPostStateSorted>,
+    /// Sorted trie updates produced by state root computation.
+    pub trie_updates: Arc<TrieUpdatesSorted>,
+    /// The persisted ancestor hash this trie input is anchored to.
+    pub anchor_hash: B256,
+    /// Trie input constructed from in-memory overlays.
+    pub trie_input: Arc<TrieInputSorted>,
+}
+
+impl PartialEq for ComputedTrieData {
```
mediocregopher: Same question as for ExecutedBlock regarding PartialEq

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl Error for DeferredTrieDataError {}
+
+#[derive(Debug, Default)]
+enum DeferredState {
+    #[default]
+    Pending,
+    Ready(ComputedTrieData),
+    Failed(DeferredTrieDataError),
```
mediocregopher: I think we could use a [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html), replacing the mutex+condvar in DeferredTrieData, and eliminate the need for this enum as well... assuming my point [here](https://github.com/paradigmxyz/reth/pull/19894/files#r2549365333) saying we can eliminate the error case

crates/chain-state/src/in_memory.rs
mattsse: yeah I think if users need Eq, they should use the block hash

crates/chain-state/src/deferred_trie.rs
mattsse: that's a good idea yes

crates/optimism/flashblocks/src/worker.rs
```diff
@@ -125,8 +128,12 @@ where
             ExecutedBlock {
```
mattsse: I think it would be useful to add some additional constructors to ExecutedBlock, because this is currently initialized by builders as well, and having an easy way to migrate this breaking change would be ideal

crates/engine/tree/src/tree/payload_validator.rs
mattsse: yep, we should definitely spawn this on the same rt

mattsse (changes requested):
seconding all of @mediocregopher comments

but functionally this approach lgtm

crates/chain-state/src/deferred_trie.rs
yongkangc: good idea, initially thought we couldnt use once lock as it doesn’t provide a blocking `wait()` for an externally set value, so we end up adding a Condvar or spin/backoff 

but good idea, tried

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: makes sense, good point here

crates/optimism/flashblocks/src/worker.rs
yongkangc: good point, @mattsse just to clarify do you mean constructors like 

```
    pub const fn new(
        recovered_block: Arc<RecoveredBlock<N::Block>>,
        execution_output: Arc<ExecutionOutcome<N::Receipt>>,
        trie_data: DeferredTrieData,
    ) -> Self {
        Self { recovered_block, execution_output, trie_data }
    }

    pub fn with_trie_data(
        recovered_block: Arc<RecoveredBlock<N::Block>>,
        execution_output: Arc<ExecutionOutcome<N::Receipt>>,
        trie_data: ComputedTrieData,
    ) -> Self {
        Self { recovered_block, execution_output, trie_data: DeferredTrieData::ready(trie_data) }
    }
```

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: i see where you are coming from, the tradeoff is that if something crashes the process, any panic within the blocking tasks leaves the underlying data unset which would cause the consumers of trie data to block forever.


but i think thats okay because it likely means that something has been broken before that? @mediocregopher

crates/chain-state/src/deferred_trie.rs
yongkangc: this would be used for test - but i think we can remove that? 

i.e
```
  // Without PartialEq:
  let result = deferred.wait_cloned();
  assert_eq!(result.hashed_state, expected.hashed_state);
  assert_eq!(result.trie_updates, expected.trie_updates);
  assert_eq!(result.anchor_hash, expected.anchor_hash);

  // With PartialEq:
  assert_eq!(deferred.wait_cloned(), expected);
```

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: thanks for the note, missed this logic

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: good point, moved over to `merge_overlay_trie_input`

crates/chain-state/src/in_memory.rs
yongkangc: right now the main breaking changes and why we needed it is for tests compatibility

else would have to use Arc::ptr_eq/is_none, state assertions compare hashes/numbers/receipts, and persistence/tree tests compare saved block hashes instead of whole structs

e.g this has to be changed
```
        assert_eq!(state.block(), block);
```

to something like this
```
        let retrieved_block = state.block();
        assert_eq!(retrieved_block.recovered_block().hash(), block.recovered_block().hash());
        assert_eq!(
            retrieved_block.execution_outcome().receipts,
            block.execution_outcome().receipts
        );
```

crates/optimism/flashblocks/src/worker.rs
yongkangc: added this after clarifications

crates/chain-state/src/in_memory.rs
yongkangc: removed Eq as its not needed for all tests, but retaining PartialEq

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: added

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +527,70 @@ where
...
+        let overlay_blocks_for_trie = overlay_blocks;
+        let deferred_compute_duration =
+            self.metrics.block_validation.deferred_trie_compute_duration.clone();
+
+        // Defer trie sorting and compute trie input to a background task so the validation hot path
+        // can return once the state root is checked. Consumers (DB writes, overlay
+        // providers, proofs) block on the handle only when they actually need the sorted
+        // trie data.
+        let task = move || {
+            let result = panic::catch_unwind(AssertUnwindSafe(|| {
```
yongkangc: having a panic here - the concern is that if something crashes the process, any panic within the blocking tasks leaves the underlying data unset which would cause the consumers of trie data to block forever.

crates/engine/tree/src/tree/payload_validator.rs
mediocregopher: > but i think thats okay because it likely means that something has been broken before that?

Right, if something panics in here then things have gone very wrong, whatever recovery we do outside of this probably won't actually fix the process.

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,295 @@
...
+/// These represent the complete set of sorted trie data required to persist
+/// block state for, and generate proofs on top of a block.
+#[derive(Clone, Debug, Default)]
+pub struct ComputedTrieData {
+    /// Sorted hashed post-state produced by execution.
+    pub hashed_state: Arc<HashedPostStateSorted>,
+    /// Sorted trie updates produced by state root computation.
+    pub trie_updates: Arc<TrieUpdatesSorted>,
+    /// The persisted ancestor hash this trie input is anchored to.
+    pub anchor_hash: Option<B256>,
```
mediocregopher: I'd like to put these in the same Option, the TrieInputSorted is not valid without knowing what its anchor hash is, so we don't want to allow the possibility that it won't be present.

yongkangc: i see that makes alot of sense, thanks for the note 👍🏻

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// Multiple threads can wait concurrently. All waiters wake when the value is set,
+    /// and each receives a cloned `ComputedTrieData` (Arc clones are cheap).
+    pub fn wait_cloned(&self) -> ComputedTrieData {
+        self.0.wait().clone()
+    }
+}
+
+impl fmt::Debug for DeferredTrieData {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let state = if self.0.get().is_some() { "ready" } else { "pending" };
```
mediocregopher: Looking at the [Debug impl for OnceLock](https://doc.rust-lang.org/src/std/sync/once_lock.rs.html#598) I think it already does basically what we want, we can just derive I think?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1791,11 +1792,17 @@ where
...
+        // Skip building trie input and anchor for DB-loaded blocks.
+        let trie_data =
+            ComputedTrieData::without_trie_input(sorted_hashed_state, sorted_trie_updates);
+
         Ok(Some(ExecutedBlock {
             recovered_block: Arc::new(RecoveredBlock::new_sealed(block, senders)),
             execution_output: Arc::new(execution_output),
-            hashed_state: Arc::new(hashed_state.into_sorted()),
-            trie_updates: Arc::new(trie_updates),
+            // Already computed from DB; expose via the deferred handle type.
```
mediocregopher: nit: not sure this comment is useful
```suggestion
```

crates/chain-state/src/deferred_trie.rs
yongkangc: @mediocregopher resolved this in [8ecbb53](https://github.com/paradigmxyz/reth/pull/19894/commits/8ecbb539cbb4a908a8a6107fe273761cb3c97b61)

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: also side note - allows us to see what fails in e2e

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +527,70 @@ where
...
+            .tree_state
+            .blocks_by_hash(block.parent_hash())
+            .unwrap_or_else(|| (block.parent_hash(), Vec::new()));
+
+        // Create a deferred handle to store the sorted trie data.
+        let deferred_trie_data = DeferredTrieData::pending();
+        let deferred_handle_task = deferred_trie_data.clone();
+        let hashed_state_for_trie = hashed_state;
+        let trie_output_for_trie = trie_output;
+        let overlay_blocks_for_trie = overlay_blocks;
```
mediocregopher: nit: I don't think these three renamings are necessary, none of these are used after the task is spawned.

crates/payload/primitives/src/traits.rs
```diff
@@ -62,8 +63,12 @@ impl<N: NodePrimitives> BuiltPayloadExecutedBlock<N> {
         reth_chain_state::ExecutedBlock {
             recovered_block: self.recovered_block,
             execution_output: self.execution_output,
-            hashed_state,
-            trie_updates,
+            trie_data: DeferredTrieData::ready(ComputedTrieData::with_trie_input(
```
mediocregopher: nit: use `without_trie_input`

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
```diff
@@ -372,8 +373,12 @@ pub trait LoadPendingBlock:
         Ok(ExecutedBlock {
             recovered_block: block.into(),
             execution_output: Arc::new(execution_outcome),
-            hashed_state: Arc::new(hashed_state.into_sorted()),
-            trie_updates: Arc::new(trie_updates.into_sorted()),
+            trie_data: DeferredTrieData::ready(ComputedTrieData::with_trie_input(
```
mediocregopher: nit: use `without_trie_input`

mediocregopher (changes requested):
mostly nits, looking good!

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
yongkangc: thanks for this, resolved all [b05110f](https://github.com/paradigmxyz/reth/pull/19894/commits/b05110f9f1049fe310b43b88b439a08adc23a503)

crates/chain-state/src/deferred_trie.rs
yongkangc: thanks, good note

crates/optimism/flashblocks/src/worker.rs
```diff
@@ -125,8 +128,12 @@ where
...
                 recovered_block: block.into(),
                 execution_output: Arc::new(execution_outcome),
-                hashed_state: Arc::new(hashed_state.into_sorted()),
-                trie_updates: Arc::default(),
+                trie_data: DeferredTrieData::ready(ComputedTrieData::with_trie_input(
```
mediocregopher: Missed this one, `without_trie_input`

mediocregopher (approved):
One final nit, LGTM 🚀

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /// Create a handle that is already populated with the given [`ComputedTrieData`].
+    ///
+    /// Useful when trie data is available immediately; [`Self::wait_cloned`] will return without
+    /// blocking.
+    pub fn ready(bundle: ComputedTrieData) -> Self {
+        let data = OnceLock::new();
+        data.set(bundle).unwrap(); // Safe: newly created OnceLock
+        Self(Arc::new(data))
```
mattsse: I think this is just `Self(Arc::new(bundle.into()))`

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// all waiters will block forever. This is intentional - trie computation failures are
+    /// considered unrecoverable and should crash the node.
+    ///
+    /// # Concurrency
+    ///
+    /// Multiple threads can wait concurrently. All waiters wake when the value is set,
+    /// and each receives a cloned `ComputedTrieData` (Arc clones are cheap).
+    pub fn wait_cloned(&self) -> ComputedTrieData {
+        self.0.wait().clone()
+    }
```
mattsse: hmm, this assumes that there's always something that's already computing the triedata,

imo this could be a potential footgun in case this isnt the case?

crates/chain-state/src/in_memory.rs
```diff
@@ -761,16 +796,42 @@ impl<N: NodePrimitives> ExecutedBlock<N> {
...
     }
 
-    /// Returns a reference to the hashed state result of the execution outcome
+    /// Returns the deferred trie data, blocking until it is available.
+    ///
+    /// This blocks the calling thread until the background trie computation completes.
+    #[inline]
+    pub fn trie_data(&self) -> ComputedTrieData {
+        self.trie_data.wait_cloned()
+    }
```
mattsse: this assumes that there is something that is doing this, otherwise this will deadlock, right?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -522,11 +529,68 @@ where
...
+            .state()
+            .tree_state
+            .blocks_by_hash(block.parent_hash())
+            .unwrap_or_else(|| (block.parent_hash(), Vec::new()));
+
+        // Create a deferred handle to store the sorted trie data.
+        let deferred_trie_data = DeferredTrieData::pending();
+        let deferred_handle_task = deferred_trie_data.clone();
+        let deferred_compute_duration =
+            self.metrics.block_validation.deferred_trie_compute_duration.clone();
```
mattsse: ca we make all of this a dedicated function, this existing fn is already quite large

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// If the background task computing trie data panics before calling [`Self::set_ready`],
+    /// all waiters will block forever. This is intentional - trie computation failures are
+    /// considered unrecoverable and should crash the node.
+    ///
+    /// # Concurrency
+    ///
+    /// Multiple threads can wait concurrently. All waiters wake when the value is set,
+    /// and each receives a cloned `ComputedTrieData` (Arc clones are cheap).
+    pub fn wait_cloned(&self) -> ComputedTrieData {
+        self.0.wait().clone()
```
mattsse: we don't have access to the internal oncestate and .get() does not block any ongoing initialization.

we could solve this with our own state tracking but this feels a bit redundant.

imo we can take the Hit here and always do `get_or_init` 

imo that'd be better than relying on something else to init it

mattsse (dismissed):

crates/chain-state/src/in_memory.rs
yongkangc: yes this is right, to counter that 

1.  we have `ExecutedBlock::new(ComputedTrieData)` is the safe default that wraps in  `DeferredTrieData::ready()` and never blocks
2. warning docs

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,549 @@
...
+    }
+
+    /// Returns trie data, computing synchronously if the async task hasn't completed.
+    ///
+    /// Uses a try-lock approach:
+    /// - If the async task has completed (`Ready`), returns the cached result.
+    /// - If pending, computes synchronously from stored inputs.
+    ///
+    /// This design eliminates deadlock risk: we never block waiting for another task.
+    /// All code paths are guaranteed to return (either cached or computed result).
```
yongkangc: The main fallback logic happens here

yongkangc: Overall flow:
- [`pending`](https://github.com/paradigmxyz/reth/blob/main/crates/chain-state/src/deferred_trie.rs#L62-L95) stores unsorted hashed state/trie updates, anchor hash, and ancestor handles while a background task sorts, merges, and calls `set_ready`.
- [`wait_cloned`](https://github.com/paradigmxyz/reth/blob/main/crates/chain-state/src/deferred_trie.rs#L96-L151) first tries a lock; if ready it returns cached data. If pending it computes synchronously, caches, and returns. When the try-lock is contended it blocks, rechecks, and does the same.
- [`compute_from_inputs`](https://github.com/paradigmxyz/reth/blob/main/crates/chain-state/src/deferred_trie.rs#L153-L195) sorts the current block’s unsorted hashed state/trie updates, merges ancestor trie data, and returns `ComputedTrieData`.

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,548 @@
...
+            // The async task is still pending, compute the trie data synchronously from the stored
+            // inputs.
+            DeferredState::Pending(inputs) => {
+                DEFERRED_TRIE_METRICS.deferred_trie_sync_fallback.increment(1);
+                let computed = Self::compute_from_inputs(inputs);
+                *state = DeferredState::Ready(computed.clone());
+                computed
+            }
+        }
+    }
```
mediocregopher, line 227:
```
        // Lock is contended - another thread/task holds the mutex (either the async
        // task calling set_ready(), or another waiter computing). We block until
        // available.
```
I thought we wanted to never block? I think you'll need to keep the inputs that are currently in Pending as fields on DeferredTrieData, then you can do something like:

```
        // If the deferred computation has completed then return the results immediately.
        if let Some(DeferredState::Ready(bundle)) = self.state.try_lock() {
            DEFERRED_TRIE_METRICS.deferred_trie_async_ready.increment(1);
            return bundle.clone();
        };

        // Otherwise compute the ComputedTrieData synchronously, so we never block.
        DEFERRED_TRIE_METRICS.deferred_trie_sync_fallback.increment(1);
        let computed = Self::compute_from_inputs(inputs);

        // If we can set the cached state then do so, otherwise just return.
        if let Some(mut DeferredState::Pending(_) @ state) = self.state.try_lock() {
            *state = DeferredState::Ready(computed.clone());
        };

        return computed
```

Basically just a OnceLock, but with metrics (which we want 👍 )

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -898,35 +910,189 @@ where
...
+
+        // Spawn background task to compute trie data. The task performs the same computation
+        // as the fallback path but runs asynchronously to avoid blocking the hot path.
+        let compute_trie_input_task = move || {
+            let result = panic::catch_unwind(AssertUnwindSafe(|| {
+                let compute_start = Instant::now();
+
```
mediocregopher: I think you could just call `deferred_handle_task.wait_cloned()` in here, and it would automatically do the synchronous computation and set the cached result without having to duplicate all of this code.

crates/chain-state/src/deferred_trie.rs
yongkangc: adding fallback logic

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: thats much simpler logic yup. added this in, thanks

yongkangc:
Made some changes based on @mediocregopher comments:

Current Logic now does: 
1. validation spawns a background spawn_blocking that calls wait_cloned() on the handle seeded with unsorted data + ancestors. Do computation and sorting of trie inputs off the hot path with Deferred Trie Data
2. This should be done in a background task, with the background calling wait_clone()
3. When wait_clone() is called from the background task, it does the computation of trie inputs and sorting asynchronously in the spawn blocking. 
4. The fallback if background task does not finish before another caller needs it is that wait_clone() is called again, at which it computes synchronously. This prevents a deadllock

@mediocregopher @mattsse should be good to go now

crates/chain-state/src/deferred_trie.rs
yongkangc: It prevents duplicate work under contention (at the cost of blocking), and avoids deadlock because whoever owns the lock must  eventually either set Ready or compute and set it.

I think thats okay here as it does not block indefinetely? and it prevents duplicate computation

If thread A holds lock on block X and waits for block Y, while thread B holds lock on block Y and waits for block X, you get deadlock. But this shouldn't happen with in context of blockchain since ancestors form a DAG (no cycles).


The alternative is that we spin off to compute instead , but this would cause heavy contention of resources

 @mediocregopher

yongkangc: think the issue is that without blocking, many parallel operations call wait_clone() at once, and this hurts the performance exponentially. so maybe the solution is oncelock::get_or_init here

yongkangc: for a deadlock to happen, we would need a cycle: Thread 1 holds Lock A, waits for Lock B; Thread 2 holds Lock B, waits for Lock A.

but i think that is unlikely to happen because:
1. If X waits on Y → Y is an parent of X
2. If Y waits on X → X is an paremt of Y
3. Both true → X is ancestor of Y AND Y is ancestor of X → cycle in blockchain itself

yongkangc: i.e 

```
Thread A (computing X)          Thread B (computing Y)
    holds X.state.lock()            holds Y.state.lock()
            │                               │
            │ compute_from_inputs()         │ compute_from_inputs()
            │ iterates X.ancestors          │ iterates Y.ancestors
            ▼                               ▼
    if Y ∈ X.ancestors:             if X ∈ Y.ancestors:
        Y.wait_cloned()                 X.wait_cloned()
            │                               │
            ▼                               ▼
        waits for Y.lock ──────────► waits for X.lock
                         DEADLOCK
```

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: makes sense, thanks for the note!

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,576 @@
...
+    ///
+    /// Sorts and builds trie input from the provided state and updates, then marks the handle
+    /// as ready so that [`Self::wait_cloned`] returns immediately without fallback computation.
+    ///
+    /// # Arguments
+    /// * `hashed_state` - Unsorted hashed post-state from execution
+    /// * `trie_updates` - Unsorted trie updates from state root computation
+    /// * `anchor_hash` - The persisted ancestor hash this trie input is anchored to
+    /// * `ancestors` - Deferred trie data from ancestor blocks for merging
+    pub fn compute_set_ready(
```
yongkangc: having this function here imo is to reduce cognitive load on payload_validator.rs where we might want to set as ready.

mediocregopher: There's two problems I see with this:
* The lock isn't being held while this is computing the trie input, so when we call this from validate_block_with_state it's possible that it doesn't finish before some other task calls wait_cloned, and that other task will be duplicating the computation.
* It doesn't really make sense to me to pass in all these values when the DeferredTrieData was initialized to them, seems confusing.

I'd prefer to get rid of this function and just call wait_cloned from validate_block_with_state like before.

yongkangc: >The lock isn't being held while this is computing the trie input, so when we call this from validate_block_with_state it's possible that it doesn't finish before some other task calls wait_cloned, and that other task will be duplicating the computation.

thats true, its a good gurantee to make sure that 👍🏻 

was initially thinking it would be unlikely that this could happen, and we wanna avoid a lock() call. i.e didnt want to add additional lock, but the guarantee is important

mediocregopher (approved):

crates/chain-state/src/in_memory.rs
```diff
@@ -719,30 +725,63 @@ impl<N: NodePrimitives> BlockState<N> {
...
+        recovered_block: Arc<RecoveredBlock<N::Block>>,
+        execution_output: Arc<ExecutionOutcome<N::Receipt>>,
+        trie_data: ComputedTrieData,
+    ) -> Self {
+        Self { recovered_block, execution_output, trie_data: DeferredTrieData::ready(trie_data) }
+    }
+
+    /// Create a new [`ExecutedBlock`] with deferred trie data.
+    ///
+    /// If the background task hasn't completed when `trie_data()` is called,
```
mattsse: this isn't immediately obvious what background task refers to here.


I'd like to rephrase the docs a bit because they currently require some context from the payload validation internals, we can use this as an example, like

"this is useful if the triedata is populated somewhere else, e.g. asynchronously after the block was validated"

crates/chain-state/src/in_memory.rs
```diff
@@ -719,30 +725,63 @@ impl<N: NodePrimitives> BlockState<N> {
...
     /// Recovered Block
     pub recovered_block: Arc<RecoveredBlock<N::Block>>,
     /// Block's execution outcome.
     pub execution_output: Arc<ExecutionOutcome<N::Receipt>>,
-    /// Block's sorted hashed state.
-    pub hashed_state: Arc<HashedPostStateSorted>,
-    /// Sorted trie updates that result from calculating the state root for the block.
-    pub trie_updates: Arc<TrieUpdatesSorted>,
+    /// Deferred trie data produced by execution.
+    /// These are computed in a background task to unblock the validation hot path
```
mattsse: we should expand on the motivation for this a bit more:

"This allows deferring the computation of the triedata which can be expensive ..."

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,437 @@
...
+
+/// Metrics for deferred trie computation.
+#[derive(Metrics)]
+#[metrics(scope = "sync.block_validation")]
+struct DeferredTrieMetrics {
+    /// Number of times deferred trie data was ready (async task completed first).
+    deferred_trie_async_ready: Counter,
+    /// Number of times deferred trie data required synchronous computation (fallback path).
+    deferred_trie_sync_fallback: Counter,
+}
```
mattsse: can we put this below the main type of the file?

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,437 @@
...
+/// If the deferred task hasn't completed, computes trie data synchronously
+/// from stored unsorted inputs rather than blocking.
+#[derive(Clone)]
+pub struct DeferredTrieData {
+    /// Shared deferred state holding either raw inputs (pending) or computed result (ready).
+    state: Arc<Mutex<DeferredState>>,
+}
+
+static DEFERRED_TRIE_METRICS: LazyLock<DeferredTrieMetrics> =
+    LazyLock::new(DeferredTrieMetrics::default);
```
mattsse, line 60: can we put this top of file

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -898,35 +908,147 @@ where
...
+        let deferred_handle_task = deferred_trie_data.clone();
+        let deferred_compute_duration =
+            self.metrics.block_validation.deferred_trie_compute_duration.clone();
+
+        // Spawn background task to compute trie data. Calling `wait_cloned` will compute from
+        // the stored inputs and cache the result, so subsequent calls return immediately.
+        let compute_trie_input_task = move || {
+            let result = panic::catch_unwind(AssertUnwindSafe(|| {
+                let compute_start = Instant::now();
+                let _ = deferred_handle_task.wait_cloned();
```
mattsse, line 1032: this should do it yeah

crates/chain-state/src/deferred_trie.rs
```diff
@@ -0,0 +1,437 @@
...
+    ) -> Self {
+        Self {
+            hashed_state,
+            trie_updates,
+            anchored_trie_input: Some(AnchoredTrieInput { anchor_hash, trie_input }),
+        }
+    }
+
+    /// Construct a bundle without trie input or anchor information.
+    pub const fn without_trie_input(
```
mattsse, line 253: this will likely be used by block builders, so we should elaborate on how this differs from `with_trie_input` and what the trie input anchor is exactly

mattsse (dismissed):
okay, this now actually quite straight forward.

I left a few (pedantic) doc nits because this is used quite a bit by externals and is breaking API

crates/chain-state/src/deferred_trie.rs
yongkangc: done, yup

crates/chain-state/src/in_memory.rs
yongkangc: thanks for the note matt

PR #19898, opened 2025-11-21
refactor(cli): deduplicate ethereum init_tracing implementations

shekhirin:
Projects building on SDK and using `ethereum::Cli` didn't have OTLP tracing initialized, because `Cli::init_tracing` method wasn't modified to do it.

To solve this, we additionally deduplicate the `init_tracing` method and have it only in `ethereum::Cli`, calling into it from `ethereum::CliApp`

### DIFF
--- a/crates/ethereum/cli/src/app.rs
+++ b/crates/ethereum/cli/src/app.rs
@@ -14,7 +14,7 @@ use reth_node_core::args::OtlpInitStatus;
 use reth_node_ethereum::{consensus::EthBeaconConsensus, EthEvmConfig, EthereumNode};
 use reth_node_metrics::recorder::install_prometheus_recorder;
 use reth_rpc_server_types::RpcModuleValidator;
-use reth_tracing::{FileWorkerGuard, Layers};
+use reth_tracing::{tracing_appender::non_blocking::WorkerGuard, FileWorkerGuard, Layers};
 use std::{fmt, sync::Arc};
 use tracing::{info, warn};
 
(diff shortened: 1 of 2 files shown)


crates/ethereum/cli/src/app.rs
```diff
@@ -97,38 +95,15 @@ where
                 self.cli.logs.log_file_directory.join(chain_spec.chain().to_string());
         }
 
-        self.init_tracing(&runner)?;
+        if self.guard.is_none() {
+            self.guard = self.cli.init_tracing(&runner, self.layers.take().unwrap_or_default())?;
+        }
```
mattsse: oh, I see this moves the init one level down?

crates/ethereum/cli/src/app.rs
```diff
@@ -97,38 +95,15 @@ where
...
         let _ = install_prometheus_recorder();
 
         run_commands_with::<C, Ext, Rpc, N>(self.cli, runner, components, launcher)
     }
-
-    /// Initializes tracing with the configured options.
-    ///
-    /// If file logging is enabled, this function stores guard to the struct.
-    /// For gRPC OTLP, it requires tokio runtime context.
-    pub fn init_tracing(&mut self, runner: &CliRunner) -> Result<()> {
```
mattsse, line 112: could we keep this fn, because some might rely on this

mattsse (approved):
I think this checks out, smol suggestion to keep the init_fn if possible, idk if it is

PR #19899, opened 2025-11-21
fix(storage): buffer advancement in Compact::from_compact for alloy wrappers

Snezhkko:
This change corrects the remainder handling in Compact::from_compact implementations for AlloyWithdrawal, AlloyHeader, and AlloyGenesisAccount by returning the advanced buffer slice produced by the inner helper’s from_compact rather than the original input slice. The Compact trait contract in storage/codecs/src/lib.rs explicitly requires from_compact to return the object and the buffer advanced past the consumed bytes; the previous code violated this by discarding the advanced slice. While vector decoding masks this behavior in some contexts by ignoring the inner remainder and manually advancing, these wrappers still broke the contract and could lead to subtle parsing errors in composed decoders that rely on the returned remainder. The fix is minimal and consistent with other alloy implementations (e.g., authorization_list and all transaction wrappers) that already propagate the advanced buffer correctly, ensuring correctness and future-proofing without changing encoding layouts or introducing allocations.

### DIFF
--- a/crates/storage/codecs/src/alloy/genesis_account.rs
+++ b/crates/storage/codecs/src/alloy/genesis_account.rs
@@ -92,7 +92,7 @@ impl Compact for AlloyGenesisAccount {
     }
 
     fn from_compact(buf: &[u8], len: usize) -> (Self, &[u8]) {
-        let (account, _) = GenesisAccount::from_compact(buf, len);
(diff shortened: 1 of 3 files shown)


mattsse (commented):
@joshieDo does this matter?

PR #19900, opened 2025-11-21
fix: correctly poll tx fetcher

klkvr:
Both `on_network_tx_event` and `on_fetch_hashes_pending_fetch` might result in new futures being pushed to `transaction_fetcher.inflight_requests`

However, right now, `transaction_fetcher` is polled before both of those, meaning that waker is never registered with new futures.

This PR fixes this and also adds a tx propagation test that spins up a network of 10 nodes and asserts that txs are propagated within 10ms.

Also did a drive-by change with parallelizing the e2e nodes setup

### DIFF
--- a/crates/ethereum/node/tests/e2e/p2p.rs
+++ b/crates/ethereum/node/tests/e2e/p2p.rs
@@ -1,10 +1,18 @@
 use crate::utils::{advance_with_random_transactions, eth_payload_attributes};
+use alloy_consensus::{SignableTransaction, TxEip1559, TxEnvelope};
+use alloy_eips::Encodable2718;
+use alloy_network::TxSignerSync;
 use alloy_provider::{Provider, ProviderBuilder};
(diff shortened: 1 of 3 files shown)


crates/ethereum/node/tests/e2e/p2p.rs
```diff
@@ -195,3 +203,94 @@ async fn test_reorg_through_backfill() -> eyre::Result<()> {
...
+
+    // Assert that all nodes have the transaction
+    for (i, node) in nodes.iter().enumerate() {
+        assert!(
+            node.rpc.inner.eth_api().transaction_by_hash(tx_hash).await?.is_some(),
+            "Node {i} should have the transaction"
+        );
+    }
+
+    // Build and send one more transaction to a random node, assert t
```
klkvr:
```suggestion
    // Build and send one more transaction to a random node
```

mattsse (approved):
nice, this checks out

PR #19904, opened 2025-11-21
feat: allow customizing dev block timestamp

klkvr:
Right now `LocalMiner` always enforces that block timestamp is always increasing https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/engine/local/src/miner.rs#L196-L202

However, in cases when mining is instant or a low block time configured, this causes block time to drift into the future.

This PR moves the timestamp choosing logic into `PayloadAttributesBuilder`, allowing to customize this behavior. This required passing `Header` generic into `PayloadAttributesBuilder`

This is sdk breaking but only for chains with custom headers

### DIFF
--- a/crates/engine/local/src/miner.rs
+++ b/crates/engine/local/src/miner.rs
@@ -106,7 +106,7 @@ pub struct LocalMiner<T: PayloadTypes, B, Pool: TransactionPool + Unpin> {
     mode: MiningMode<Pool>,
     /// The payload builder for the engine
     payload_builder: PayloadBuilderHandle<T>,
-    /// Timestamp for the next block.
+    /// Latest block in the chain so far.
     last_header: SealedHeaderFor<<T::BuiltPayload as BuiltPayload>::Primitives>,
     /// Stores latest mined blocks.
     last_block_hashes: VecDeque<B256>,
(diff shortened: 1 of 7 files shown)


crates/engine/local/src/miner.rs
```diff
@@ -107,35 +107,38 @@ pub struct LocalMiner<T: PayloadTypes, B, Pool: TransactionPool + Unpin> {
     /// The payload builder for the engine
     payload_builder: PayloadBuilderHandle<T>,
     /// Timestamp for the next block.
```
mattsse: need docs update

crates/engine/local/src/payload.rs
```diff
@@ -13,21 +15,36 @@ use std::sync::Arc;
...
 where
-    ChainSpec: Send + Sync + EthereumHardforks + 'static,
+    ChainSpec: EthChainSpec + EthereumHardforks + 'static,
 {
-    fn build(&self, timestamp: u64) -> EthPayloadAttributes {
+    fn build(&self, parent: &SealedHeader<ChainSpec::Header>) -> EthPayloadAttributes {
+        let mut timestamp =
+            std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs();
+
+        if self.enforce_increasing_timestamp {
```
mattsse, line 44: ah I see, this makes sense

mattsse (approved):
lgtm, pending smol doc nit

PR #19907, opened 2025-11-21
fix(exex): update lowest/highest committed block heights on insert

VolodymyrBg:
Update BlockCache::insert_notification_blocks_with_file_id to compute and persist both lowest_committed_block_height and highest_committed_block_height during committed chain insertion by aggregating the block numbers over the inserted chain and applying min/max with existing values. Previously, lowest_committed_block_height was left unset until finalize, causing the corresponding gauge to remain empty and misrepresenting the WAL state in Grafana, while highest_committed_block_height was overwritten with the current tip which could transiently decrease even when higher committed blocks still existed in the cache. The new logic performs a single pass over committed blocks, tracking chain-local min/max, then merges them with existing metrics, ensuring consistent, monotonic metrics immediately after commit and during cache reconstruction.

### DIFF
--- a/crates/exex/exex/src/wal/cache.rs
+++ b/crates/exex/exex/src/wal/cache.rs
@@ -108,15 +108,32 @@ impl BlockCache {
         }
 
         if let Some(committed_chain) = &committed_chain {
+            let mut chain_min: Option<BlockNumber> = None;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
I dont think we need to change this

PR #19908, opened 2025-11-22
fix(txpool): prevent removed transactions from being included in blocks

wo-o:
## Problem

A race condition exists where `BestTransactions` iterator creates an independent snapshot via `BTreeMap::clone()`. When the maintenance job removes transactions from the pool, existing iterators remain unaware, causing removed transactions to be included in blocks.

### Production Impact

During load testing with `--txpool.lifetime 600`, this caused sequencer failure:

```
2025-11-15T13:30:01.790071Z WARN The database read transaction has been open for too long.
    open_duration=190.664088693s self.txn_id=7039115

2025-11-15T13:30:01.947723Z WARN Attempt to calculate state root for an old block
    might result in OOM target=2846869
```

The sequencer stopped producing blocks because:
1. Maintenance removed expired transactions from pool
2. Block builder's snapshot still contained the removed transactions
3. Execution attempted to process non-existent transactions
4. Database read stuck
5. State root calculation failed

This is critical for transactions with time-based expiration.

### Root Cause

`crates/transaction-pool/src/pool/pending.rs:110` - The iterator snapshot is created via `BTreeMap::clone()`, which creates an independent copy. The pool has notification support for new transactions but lacks removal notifications.

## Solution

Add removal notifications following the existing pattern for new transactions:

1. `PendingPool` broadcasts transaction removals via new `removed_transaction_notifier` channel
2. `BestTransactions` subscribes to removal notifications on creation
3. Iterator processes removals before yielding next transaction

### DIFF
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -1,7 +1,7 @@
 use crate::{
     error::{Eip4844PoolTransactionError, InvalidPoolTransactionError},
     identifier::{SenderId, TransactionId},
-    pool::pending::PendingTransaction,
+    pool::pending::{PendingPoolEvent, PendingTransaction},
     PoolTransaction, Priority, TransactionOrdering, ValidPoolTransaction,
 };
 use alloy_consensus::Transaction;
@@ -93,17 +93,11 @@ pub struct BestTransactions<T: TransactionOrdering> {
     pub(crate) independent: BTreeSet<PendingTransaction<T>>,
     /// There might be the case where a yielded transactions is invalid, this will track it.
     pub(crate) invalid: HashSet<SenderId>,
-    /// Used to receive any new pending transactions that have been added to the pool after this
-    /// iterator was static filtered
+    /// Used to receive pool events (additions and removals) after this iterator was created.
     ///
-    /// These new pending transactions are inserted into this iterator's pool before yielding the
-    /// next value
-    pub(crate) new_transaction_receiver: Option<Receiver<PendingTransaction<T>>>,
-    /// Used to receive transaction removals from the pool after this iterator was created.
-    ///
-    /// Removed transactions are deleted from this iterator's snapshot before yielding the next
-    /// value, preventing inclusion of transactions that were removed by maintenance jobs.
-    pub(crate) removed_transaction_receiver: Option<Receiver<TransactionId>>,
+    /// New transactions are inserted and removed transactions are deleted from this iterator's
+    /// snapshot before yielding the next value.
+    pub(crate) event_receiver: Option<Receiver<PendingPoolEvent<T>>>,
     /// The priority value of most recently yielded transaction.
     ///
     /// This is required if new pending transactions are fed in while it yields new values.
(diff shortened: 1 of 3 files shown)


mattsse:
could you elaborate on why

> causing removed transactions to be included in blocks.

would be problematic, because atm I believe this would only happen if the pending pool is at capacity, but I assume you need this to satisfy some protocol level rules?


> Block builder's snapshot still contained the removed transactions
> Execution attempted to process non-existent transactions
> Database read stuck

not following the sequence of events here that's causing db/State root issues, maybe this has something to do with some protocol specific rules?
if the blockbuilder adds this tx into the block, how would this cause execution errors?

crates/transaction-pool/src/pool/best.rs
```diff
@@ -99,6 +99,11 @@ pub struct BestTransactions<T: TransactionOrdering> {
     /// These new pending transactions are inserted into this iterator's pool before yielding the
     /// next value
     pub(crate) new_transaction_receiver: Option<Receiver<PendingTransaction<T>>>,
+    /// Used to receive transaction removals from the pool after this iterator was created.
+    ///
+    /// Removed transactions are deleted from this iterator's snapshot before yielding the next
+    /// value, preventing inclusion of transactions that were removed by maintenance jobs.
+    pub(crate) removed_transaction_receiver: Option<Receiver<TransactionId>>,
```
mattsse: not opposed to this, but since this is an internal channel we can unify this by introducing 

```
enum PendingPoolEvent<T> {Added(Tx),Removed(tx)}
```

mattsse (changes requested):
lacking some context on why excluding removed txs is critical

but maybe this is similar to OP interop inclusion rules

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/optimism/txpool/src/transaction.rs#L104-L115

which is checked during block building for example

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/optimism/payload/src/builder.rs#L696-L696

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/optimism/payload/src/builder.rs#L730-L737

crates/transaction-pool/src/pool/best.rs
wo-o: @mattsse Thanks for the feedback. I think my original description was misleading.

The issue is not that removed transactions get included in blocks. The actual problem is that when a transaction is removed from the pool (e.g., due to --txpool.lifetime expiration), the BestTransactions iterator still holds references to it in its snapshot. During block building, this mismatch causes the iterator to get stuck trying to access transactions that no longer exist in the pool's state, leading to long-running DB read transactions and eventually blocking block production.

I've also unified the two notification channels into a single PendingPoolEvent enum as you suggested:
https://github.com/paradigmxyz/reth/pull/19908/commits/0f2d0afeee60e29cf67838a4c32af9f55923498c

mattsse: > leading to long-running DB read transactions and eventually blocking block production.

can you expand on this a bit more, because I don't quite understand how this would cause issues during building, is there something that just endlessly loops then? and how exactly

PR #19909, opened 2025-11-22
feat: update docs link in cli/SUMMARY.mdx

cuiweixie:
Use relative URLs so links can be clicked on GitHub as well. This ensures compatibility with older versions. We may need to access the documentation directly on GitHub — at least, I need this.


cuiweixie:
https://github.com/EmbarkStudios/cargo-deny/issues/804

mattsse (commented):
these are intended for https://reth.rs/cli/cli

but I believe these links should work as well, let's see

mattsse (approved):

PR #19911, opened 2025-11-22
refactor(engine): store hashes in blocks_by_number instead of ExecutedBlock

phrwlk:
This change replaces the per-height index blocks_by_number from a Vec<ExecutedBlock<N>> to a Vec<B256>, removing duplicated storage of full blocks. All code paths using this index only require block hashes for pruning and membership operations; the authoritative block data remains in blocks_by_hash. This reduces memory overhead and avoids unnecessary cloning in insert_executed, while preserving existing behavior in canonical removal and finalized sidechain pruning. Tests are updated accordingly to initialize the index with hashes.


mattsse (commented):
ExecutedBlock is just arc data so we can keep this as is

PR #19912, opened 2025-11-22
Fix Clippy regressions across trie, storage, and ethereum e2e tests

0xLogicalx:
Quote the engine_api IPC doc string and drop a redundant MDBX cast to appease Clippy. Switch trie prefix extension to iter().copied() and mark the sigsegv handler stub as const fn. Scope unix-only e2e imports and pin RNG outputs to u8/B256 to resolve type inference failures


mattsse (commented):
we can keep this as is

PR #19914, opened 2025-11-22
feat(op): pull chain config from superchain-registry github

0xPenryn:
Updates the `reth-optimism-chainspec` crate, adding functionality to pull the latest chain configuration directly from the superchain registry. If the superchain registry is unreachable, it will fall back to the local chain configurations and follow the existing behavior.

This means that op-reth doesn't need to cut a new release to reflect updates to the superchain registry; rather, the node operator simply needs to restart op-reth and it will pull the latest config.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9364,6 +9364,7 @@ dependencies = [
  "op-alloy-consensus",
  "op-alloy-rpc-types",
  "paste",
+ "reqwest",
  "reth-chainspec",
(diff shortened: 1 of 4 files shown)


mattsse:
Closing this PR as op-reth has been moved to its own repository. See https://github.com/paradigmxyz/reth/pull/21532 and the updated README for the new location. If this is still relevant, please re-open in the new op-reth repository. 🙏

PR #19915, opened 2025-11-22
fix: Extract common header locking logic in GasPriceOracle

vtjl10:
### Description

Removes duplicate code in `GasPriceOracle` by extracting the common header fetching and cache check logic from `suggest_tip_cap` and `op_suggest_tip_cap` into helper methods.

**Changes:**
- Added `lock_head_and_inner()` to fetch the latest header and acquire the inner mutex
- Added `cached_price_for_head()` to check if a cached price exists for the current head


mattsse (commented):
we can keep this as is

PR #19916, opened 2025-11-22
feat(engine): remove chunks shadowing and unify span recording

SashaMalysehko:
Removed the inner shadowed chunks counters in on_prefetch_proof and on_state_update and records the tracing span’s chunks field using the outer counter that is already used for metrics and return values. The previous pattern introduced two independent counters with identical intended semantics, which is unnecessary and risky for maintenance if future edits diverge their behavior. The new approach keeps a single source of truth, ensuring tracing, metrics, and logic remain aligned without changing runtime behavior.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -773,10 +773,8 @@ impl MultiProofTask {
             let Some(chunk_size) = self.chunk_size &&
             proof_targets.chunking_length() > chunk_size
         {
-            let mut chunks = 0usize;
             for proof_targets_chunk in proof_targets.chunks(chunk_size) {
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -773,10 +773,8 @@ impl MultiProofTask {
             let Some(chunk_size) = self.chunk_size &&
             proof_targets.chunking_length() > chunk_size
         {
-            let mut chunks = 0usize;
             for proof_targets_chunk in proof_targets.chunks(chunk_size) {
                 dispatch(proof_targets_chunk);
-                chunks += 1;
```
mattsse, line 779: this isn't correct because this removes the chunks itself

mattsse (commented):
these are different span infos

PR #19920, opened 2025-11-23
refactor: simplify rpc state provider traits

klkvr:
This changes RPC code so that we can operate on `&mut State<DB>` directly.


crates/rpc/rpc-eth-types/src/cache/db.rs
```diff
@@ -171,66 +170,11 @@ impl StateProvider for StateProviderTraitObjWrapper<'_> {
...
         &self,
         code_hash: &B256,
     ) -> reth_errors::ProviderResult<Option<reth_primitives_traits::Bytecode>> {
         self.0.bytecode_by_hash(code_hash)
     }
 }
-
-/// Hack to get around 'higher-ranked lifetime error', see
-/// <https://github.com/rust-lang/rust/issues/100013>
-pub struct StateCacheDbRefMutWrapper<'a, 'b>(pub &'b mut StateCacheDb<'a>);
```
mattsse, line 185: rip bozo

mattsse (approved):

PR #19921, opened 2025-11-23
refactor: better spawn_with_state_at_block API

klkvr:
Based on https://github.com/paradigmxyz/reth/pull/19920

Improves `spawn_with_state_at_block` API so that callers don't need to build database and clone `self` manually


crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -550,16 +544,20 @@ pub trait Call:
     /// Executes the closure with the state that corresponds to the given [`BlockId`] on a new task
     fn spawn_with_state_at_block<F, R>(
         &self,
-        at: BlockId,
+        at: impl Into<BlockId>,
```
mattsse, line 547: yeah I guess this fn is duplicated anyway so we can just accept inot here

mattsse (approved):
makes sense

PR #19924, opened 2025-11-23
chore: remove unused lifetime from map_internal_err impl

Bashmunta:
The impl generated by impl_to_rpc_result! for map_internal_err declared an extra lifetime parameter 'a that is not present in the ToRpcResult trait method and is not used by the function signature or bounds. This was a copy-paste artifact from the methods that legitimately accept borrowed data and could trigger the unused_lifetimes lint. The change removes the redundant lifetime so the impl matches the trait and avoids misleading generics without altering behavior.


mattsse (approved):
fine

PR #19925, opened 2025-11-23
refactor(tracing): introduce `DebugInspector`

klkvr:
This PR refactors `DebugApi` and encapsulates the tracing logic in a `DebugInspector` enum.

This gives more control to the tracing callsite as it now knows has control over inspector type, allowing to rewrite those APIs logic to use `BlockExecutor`s and potentially reuse more existing helpers

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -1077,7 +1077,7 @@ enum DebugInspector {
     FlatCallTracer(TracingInspector),
     Default(TracingInspector, GethDefaultTracingOptions),
     #[cfg(feature = "js-tracer")]
-    Js(Option<revm_inspectors::tracing::js::JsInspector>, String, serde_json::Value),
+    Js(Option<Box<revm_inspectors::tracing::js::JsInspector>>, String, serde_json::Value),
 }
 
 impl DebugInspector {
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -1428,3 +1070,268 @@ struct DebugApiInner<Eth> {
...
+    FourByte(FourByteInspector),
+    CallTracer(TracingInspector, CallConfig),
+    PreStateTracer(TracingInspector, PreStateConfig),
+    Noop(NoOpInspector),
+    Mux(MuxInspector, MuxConfig),
+    FlatCallTracer(TracingInspector),
+    Default(TracingInspector, GethDefaultTracingOptions),
+    #[cfg(feature = "js-tracer")]
+    Js(Box<revm_inspectors::tracing::js::JsInspector>, String, serde_json::Value),
+}
```
mattsse, line 1092: this makes sense, this is kinda similar to the mux tracer itself

we should upstream this to revm-inspectors once merged

crates/rpc/rpc/src/debug.rs
```diff
@@ -530,18 +359,14 @@ where
...
-                    call_tx_env,
-                    &mut db,
-                    None, // No transaction context for call tracing
-                    &mut None,
-                )?;
+                let mut inspector = DebugInspector::new(tracing_options)?;
+                let res =
+                    eth_api.inspect(&mut db, evm_env.clone(), tx_env.clone(), &mut inspector)?;
+                let trace =
+                    inspector.get_result(None, &tx_env, &evm_env.block_env, &res, &mut db)?;
```
mattsse, line 369: this makes me think that we could probably get rid of this clone by obtaining the required values before, which are usually only gaslimit,intput,fees but then, cloning txenv is pretty cheap in most cases

mattsse (approved):
cool, this is a lot better

PR #19931, opened 2025-11-24
feat(storage): log storage settings on startup

shekhirin:
```console
2025-11-24T09:59:31.281131Z  INFO Storage settings legacy=StorageSettings { receipts_in_static_files: false, transaction_senders_in_static_files: false } stored=Some(StorageSettings { receipts_in_static_files: false, transaction_senders_in_static_files: false }) current=StorageSettings { receipts_in_static_files: false, transaction_senders_in_static_files: false }
```


joshieDo (approved):

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -93,24 +93,32 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
...
             chain_spec.clone(),
             static_file_provider.clone(),
             Default::default(),
             Default::default(),
             Arc::new(RwLock::new(legacy_settings)),
         )
-        .storage_settings()?
-        .unwrap_or(legacy_settings);
+        .storage_settings()?;
+        let storage_settings = Arc::new(RwLock::new(stored_settings.unwrap_or(legacy_settings)));
```
mattsse: this change is a bit weird, why move the  arc rwlock to here?

mattsse (changes requested):
what if we log this here instead?

https://github.com/paradigmxyz/reth/blob/e03c9da85cf8ec7b9864c46d0bdcafff6689ef04/crates/node/builder/src/launch/engine.rs#L99-L102

mattsse (approved):

PR #19933, opened 2025-11-24
chore(test): Increase timeout for ethereum e2e tests

yongkangc:
## Summary
- raise the nextest slow-timeout for reth-node-ethereum e2e binary to ~5 minutes so long-running p2p/blob e2e tests stop hitting the 120s cap
- motivated by repeated timeouts in the integration job: https://github.com/paradigmxyz/reth/actions/runs/19631439252/job/56211995794?pr=19894


mattsse (approved):
let's see

PR #19937, opened 2025-11-24
fix: npm run build failed in docs/vocs directory

cuiweixie:


mattsse (approved):
okay, let's try that first

PR #19940, opened 2025-11-24
fix(txpool): ensure skipped high-priority transactions are tracked in BestTransactions

keanji-x:
When a high-priority transaction arrives late, we skip yielding it to enforce priority ordering. Previously, we also failed to track it in the pool.

This caused a critical bug: if a descendant of that skipped transaction arrived later, `ancestor()` check would fail (return None). The descendant was then incorrectly promoted to the `independent` set and yielded out of order (before its parent).

This change stashes skipped transactions in the `all` map so descendants correctly detect their dependencies.

### DIFF
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -126,7 +126,7 @@ impl<T: TransactionOrdering> BestTransactions<T> {
     }
 
     /// Non-blocking read on the new pending transactions subscription channel
-    fn try_recv(&mut self) -> Option<PendingTransaction<T>> {
+    fn try_recv(&mut self) -> Option<IncomingTransaction<T>> {
         loop {
             match self.new_transaction_receiver.as_mut()?.try_recv() {
                 Ok(tx) => {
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/best.rs
```diff
@@ -74,6 +74,11 @@ impl<T: TransactionOrdering> Iterator for BestTransactionsWithFees<T> {
     }
 }
 
+enum IncomingTransaction<T: TransactionOrdering> {
+    Process(PendingTransaction<T>),
+    Stash(PendingTransaction<T>),
+}
```
mattsse: unclear what these variants mean

crates/transaction-pool/src/pool/best.rs
```diff
@@ -170,11 +175,20 @@ impl<T: TransactionOrdering> BestTransactions<T> {
...
+                    IncomingTransaction::Process(tx) => {
+                        let tx_id = *tx.transaction.id();
+                        if self.ancestor(&tx_id).is_none() {
+                            self.independent.insert(tx.clone());
+                        }
+                        self.all.insert(tx_id, tx);
+                    }
+                    IncomingTransaction::Stash(tx) => {
+                        let tx_id = *tx.transaction.id();
+                        self.all.insert(tx_id, tx);
```
mattsse, line 184: I see, this prevents broken nonce chain

crates/transaction-pool/src/pool/best.rs
mattsse: can we add some docs here and move to below the impl block

mattsse (changes requested):
makes sense, although this would only result in a nonce error on execution but we can avoid this with this change

some style nits

crates/transaction-pool/src/pool/best.rs
keanji-x: Good idea, fixed

mattsse (approved):
nice find, ty!

Issue #19513, opened 2025-11-05
Proof Calc Rewrite: Cached Branch Nodes


mediocregopher:
### Describe the feature

Modify the proof calculator to use branch nodes cached in Accounts/StoragesTrie tables. When cached branch nodes are present we can skip descending into any of their children which:

* Have a pre-calculated hash in the cached branch node.
* Are not a child of the current proof target.

The proptest should be modified to initially use StateRoot to generate TrieUpdates for the test dataset, and include those TrieUpdates in the mock cursor factory.

PR #19941, opened 2025-11-24
feat(trie): Proof V2: retain proof nodes which match targets

mediocregopher:
Closes #19512

This PR gives the new `proof_v2::ProofCalculator` the ability to retain the `ProofTrieNode`s for all nodes in the trie which match the given proof targets. The legacy proof calculator retains proofs by checking each node against all targets, whereas V2 checks each node against only a single target. See the [`should_retain`](https://github.com/paradigmxyz/reth/pull/19941/files#diff-0f5e4d1b41c9f7f212a9fb1ec319846f1489e692499d00bbd356c3a2022b1131R154-R159) method for details on how this works.

To support proof retention the idea of "committing" a child is introduced. A child is "committed" to when there can be no further changes to it, such as an extension or a leaf being split to insert a new branch. When a child is committed it is converted to an RlpNode, checked to see if its proof should be retained, and pushed back onto the child stack as a `ProofTrieBranchChild::RlpNode`. This is done all as one step in order to best reuse buffers during RLP-encoding.

### DIFF
--- a/crates/trie/trie/src/proof_v2/mod.rs
+++ b/crates/trie/trie/src/proof_v2/mod.rs
@@ -216,6 +216,14 @@ where
             self.retained_proofs.last().map(|n| n.path),
         );
 
+        trace!(
+            target: TRACE_TARGET,
+            ?path,
+            last_retained_path = ?self.retained_proofs.last().map(|n| n.path),
+            target = ?targets.peek(),
+            "should_retain: called",
+        );
+
         // If the node in question is a prefix of the previously retained proof then we retain.
         //
         // This is required for cases where the target iterator is moved forward such that the
@@ -239,6 +247,11 @@ where
             target < path
         {
             targets.next();
+            trace!(
+                target: TRACE_TARGET,
+                target = ?targets.peek(),
+                "target < path, next target",
+            );
         }
 
         // If the node in question is a prefix of the target then we retain
@@ -332,13 +345,9 @@ where
         // Create an iterator over the paths of each child in the branch.
         let child_paths = {
             let branch_path = self.branch_path;
-            debug_assert!(
-                branch_path.len() < 64,
-                "branch_path {branch_path:?} is too long to extend"
-            );
             TrieMaskIter(branch.state_mask).map(move |nibble| {
                 let mut child_path = branch_path;
-                child_path.extend_from_slice_unchecked(&[nibble]);
+                child_path.push_unchecked(nibble);
                 child_path
             })
         };
@@ -493,8 +502,8 @@ where
 
         // If there are no targets then nothing could be returned, return early.
         if targets.peek().is_none() {
-            // TODO uncomment
-            //return Ok(Vec::new())
+            trace!(target: TRACE_TARGET, "Empty targets, returning");
+            return Ok(Vec::new())
         }
 
         // Ensure initial state is cleared. By the end of the method call these should be empty once
@@ -631,7 +640,7 @@ where
     }
 }
 
-/// A helper type for iterating over the indexes of the non-zero bits of a TrieMask.
+/// A helper type for iterating over the indexes of the non-zero bits of a [`TrieMask`].
 struct TrieMaskIter(TrieMask);
 
 impl Iterator for TrieMaskIter {
@@ -653,10 +662,11 @@ mod tests {
         proof::Proof,
         trie_cursor::{mock::MockTrieCursorFactory, TrieCursorFactory},
     };
-    use alloy_primitives::map::B256Map;
+    use alloy_primitives::map::{B256Map, B256Set};
     use alloy_rlp::Decodable;
+    use assert_matches::assert_matches;
     use itertools::Itertools;
-    use reth_trie_common::{HashedPostState, MultiProofTargets};
+    use reth_trie_common::{HashedPostState, MultiProofTargets, TrieNode};
     use std::collections::BTreeMap;
 
     /// Target to use with the `tracing` crate.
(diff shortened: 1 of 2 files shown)


crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -58,41 +61,53 @@ impl<RF: DeferredValueEncoder> ProofTrieBranchChild<RF> {
...
             }
             Self::Extension { short_key, child } => {
-                child.encode(buf);
-                let child_rlp_node = RlpNode::from_rlp(buf);
-                Ok(TrieNode::Extension(ExtensionNode { key: short_key, child: child_rlp_node }))
+                (TrieNode::Extension(ExtensionNode { key: short_key, child }), TrieMasks::none())
             }
-            Self::Branch(branch_node) => Ok(TrieNode::Branch(branch_node)),
-        }
+            // TODO store trie masks on branch
```
mediocregopher, line 93: Will be done as part of #19513

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -53,7 +55,17 @@ pub struct ProofCalculator<TC, HC, VE: LeafValueEncoder> {
     /// The children for the bottom branch in `branch_stack` are found at the bottom of this stack,
     /// and so on. When a branch is removed from `branch_stack` its children are removed from this
     /// one, and the branch is pushed onto this stack in their place (see [`Self::pop_branch`].
+    ///
+    /// Children on the `child_stack` are converted to [`ProofTrieBranchChild::RlpNode`]s via the
+    /// [`Self::commit_child`] method. Committing a child indicates that no further changes are
+    /// expected to happen to it (e.g. splitting its short key when inserting a new branch). Given
+    /// that keys are consumed in lexicographical order, only the most last child on the stack can
+    /// ever be modified, and therefore all children besides the last are expected to be
+    /// [`ProofTrieBranchChild::RlpNode`]s.
```
mediocregopher, line 64: Given this property, it might make more sense to split this into something like:
```
child_stack: Vec<RlpNode>
uncommitted_child: Option<ProofTrieBranchChild<VE::DeferredEncoder>>
```
I'm not sure that this would give us any performance gains or anything, but it might make the code more understandable and easier to validate as correct. Would be better done as a follow-up PR though.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -191,66 +406,53 @@ where
...
         );
 
-        // Update the branch_path. If this branch is the only branch then only its extension needs
-        // to be trimmed, otherwise we also need to remove its nibble from its parent.
-        let new_path_len = self.branch_path.len() -
-            branch.ext_len as usize -
-            if self.branch_stack.is_empty() { 0 } else { 1 };
-
-        debug_assert!(self.branch_path.len() >= new_path_len);
-        self.branch_path = self.branch_path.slice_unchecked(0, new_path_len);
```
mediocregopher, line 239: This logic didn't change, it only got re-ordered in the method a bit. We cannot modify the `branch_path` until _after_ `commit_child` has potentially been called on the popped branch, in the case that the branch has an extension node parent.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -99,13 +115,219 @@ where
             .unwrap_or_else(|| Vec::with_capacity(16))
     }
 
+    /// Returns true if the proof of a node at the given path should be retained. This may move the
```
yongkangc: maybe we can add a tldr on _why_ the proof should be retained so its easy to follow _how_ it works

e.g
```suggestion
    /// Returns true if the proof of a node at the given path should be retained. 
    /// A node is retained if its path is a prefix of any target.
    /// This may move the
```

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -16,15 +17,17 @@ pub(crate) enum ProofTrieBranchChild<RF> {
...
         /// The short key of the leaf.
         short_key: Nibbles,
-        /// The node of the child branch.
-        child: BranchNode,
+        /// The [`RlpNode`] of the child branch.
+        child: RlpNode,
     },
     /// A branch node whose children have already been flattened into [`RlpNode`]s.
     Branch(BranchNode),
+    // A node whose type is not known, as it has already been converted to an [`RlpNode`].
```
yongkangc:
```suggestion
    /// A node whose type is not known, as it has already been converted to an [`RlpNode`].
```

smol nit for consistency

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -53,7 +55,17 @@ pub struct ProofCalculator<TC, HC, VE: LeafValueEncoder> {
     /// The children for the bottom branch in `branch_stack` are found at the bottom of this stack,
     /// and so on. When a branch is removed from `branch_stack` its children are removed from this
     /// one, and the branch is pushed onto this stack in their place (see [`Self::pop_branch`].
+    ///
+    /// Children on the `child_stack` are converted to [`ProofTrieBranchChild::RlpNode`]s via the
+    /// [`Self::commit_child`] method. Committing a child indicates that no further changes are
+    /// expected to happen to it (e.g. splitting its short key when inserting a new branch). Given
+    /// that keys are consumed in lexicographical order, only the most last child on the stack can
```
yongkangc:
```suggestion
    /// that keys are consumed in lexicographical order, only the last child on the stack can
```

smol nit

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -99,13 +115,219 @@ where
...
+    ///
+    /// * targets.next() is called.
+    ///
+    /// * targets.peek() now returns (0x045, Some(0x678)). This does contain 0x04.
+    ///
+    /// * 0x04 is a prefix of 0x045, and so is retained.
+    /// ```
+    ///
+    /// Because paths in the trie are visited in depth-first order, it's imperative that targets are
+    /// given in depth-first order as well. If the targets where generated off of B256s, which is
```
yongkangc:
```suggestion
    /// given in depth-first order as well. If the targets were generated off of B256s, which is
```

do you mean this?

yongkangc (approved):
`should_retain` and `commit_child` was easy enough to understand! dont have any blocking comments 🚀

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -99,13 +115,219 @@ where
...
+        child_path: Nibbles,
+        child: ProofTrieBranchChild<VE::DeferredEncoder>,
+    ) -> Result<RlpNode, StateProofError> {
+        // If the child is already an `RlpNode` then there is nothing to do.
+        if let ProofTrieBranchChild::RlpNode(rlp_node) = child {
+            return Ok(rlp_node)
+        }
+
+        // If we should retain the child then do so.
+        if self.should_retain(targets, &child_path) {
```
yongkangc, line 203: do you think it would help if we can have a skip here if all targets are consumed? 

i.e
```
 // let targets_exhausted = targets.peek().is_none();
  if !targets_exhausted && self.should_retain(targets, &child_path) {
      // retain logic...
  }
```

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -507,18 +754,55 @@ where
...
+    /// iterator is exhausted then `Some(prev, None)` is returned on the subsequent call to
+    /// `WindowIter::next`, and `None` from the call after that.
+    type Item = (I::Item, Option<I::Item>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match (self.prev, self.iter.next()) {
+            (None, None) => None,
+            (None, Some(v)) => {
+                self.prev = Some(v);
+                self.next()
```
mattsse: this is recursive, but shouldnt be problematic because this should always yield in the next next call, but this can also be converted into a loop easily

mattsse (approved):
cool, this was actually very easy to follow with your comments

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: It's not actually possible for `targets` to ever be exhausted at this point. `should_retain` will only call `next` if the "next" target is Some, meaning it will never consume the final item in the iterator. If `targets` starts off exhausted then there is a fast-path in `proof_inner` to immediately return.

PR #19943, opened 2025-11-24
fix(transaction-pool): Spawn ValidationTask to keep channel open

radik878:
In crates/transaction-pool/src/validate/task.rs, the constructor TransactionValidationTaskExecutor::new() previously created the channel ValidationTask::new () channel and immediately lost the receiver without running ValidationTask::run(), which caused the channel to close and all send() calls to end with ValidationServiceUnreachable; now we save the receiver and immediately start the background runner via tokio::task::spawn(async move { task.run().await; }), so the validator actually consumes and executes tasks, and validation calls start working, and we added regression tests at the end of task.rs, which confirmed correctness after the fix.


crates/transaction-pool/src/validate/task.rs
```diff
@@ -181,7 +181,10 @@ impl<V> TransactionValidationTaskExecutor<V> {
     /// Initializes the executor with the provided validator and sets up communication for
     /// validation tasks.
     pub fn new(validator: V) -> Self {
```
mattsse: can we instead return the task itself here?

mattsse (changes requested):

crates/transaction-pool/src/validate/task.rs
radik878: > can we instead return the task itself here?

Yes

mattsse (approved):

PR #19947, opened 2025-11-24
stage-set: add #[track_caller] to index_of() and set() for clearer pa…

gap-editor:
#  **Description**

This PR improves the debugging experience of `StageSetBuilder` by adding `#[track_caller]` to two functions that may panic:

* `index_of`
* `set`

Both functions panic when a caller references a stage that does not exist in the set. Previously, these panics pointed to internal builder code, making it harder to identify the root cause.

### What this change does

Adds `#[track_caller]` to surface the original callsite in panic messages. Leaves all logic, comments, ordering, and formatting untouched. No behavioral or API changes.

### Why it matters

* Better panic diagnostics during pipeline/stage configuration.
* Consistent with other builder methods (`disable`, `disable_if`, etc.) that already use `#[track_caller]`.
* Helps developers quickly locate misconfigured stage definitions.

### DIFF
--- a/crates/stages/api/src/pipeline/set.rs
+++ b/crates/stages/api/src/pipeline/set.rs
@@ -65,6 +65,7 @@ impl<Provider> Debug for StageSetBuilder<Provider> {
 }
 
 impl<Provider> StageSetBuilder<Provider> {
+    #[track_caller]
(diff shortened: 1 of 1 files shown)


mattsse (commented):
not needed

PR #19953, opened 2025-11-25
feat: introduce PayloadValidator::payload_to_block

### LINKED ISSUE #19949: Introduce PayloadValidator::payload_to_block
### Describe the feature

we have:

https://github.com/paradigmxyz/reth/blob/118fd3b3720ca2c244d18102cc0c1632096b860f/crates/engine/primitives/src/lib.rs#L142-L145

which also does recovery:

https://github.com/paradigmxyz/reth/blob/118fd3b3720ca2c244d18102cc0c1632096b860f/crates/ethereum/node/src/engine.rs#L50-L51

this can be wasteful because during payload processing we've already done this, however we still want to support cases where the payload itself is the RecoveredBlock already

so we really need

```rust
    fn convert_payload_to_block(
        &self,
        payload: ExecutionData,
    ) -> Result<SealedBlock<Self::Block>, NewPayloadError> {
```

## TODO
* add function
* update all current ensure_well_formed_payload which we can make a default impl by matching over `convert_payload_to_block`

### Additional context

_No response_

stevencartavia:
closes #19949

### DIFF
--- a/crates/engine/primitives/src/lib.rs
+++ b/crates/engine/primitives/src/lib.rs
@@ -17,7 +17,7 @@ use reth_payload_primitives::{
     EngineApiMessageVersion, EngineObjectValidationError, InvalidPayloadAttributesError,
     NewPayloadError, PayloadAttributes, PayloadOrAttributes, PayloadTypes,
 };
-use reth_primitives_traits::{Block, RecoveredBlock};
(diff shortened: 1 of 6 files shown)


mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):

PR #19954, opened 2025-11-25
feat(flashblock): Enable eth_getTransactionByHash support for flashblock

0x00101010:
Enable eth_getTransactionByHash support for flashblock, bring it at parity with existing support

### DIFF
--- a/crates/optimism/rpc/src/eth/transaction.rs
+++ b/crates/optimism/rpc/src/eth/transaction.rs
@@ -222,7 +222,7 @@ where
         }
 
         // 3. check local pool
-        if let Some(tx) = self.pool().get(&hash).map(|tx| tx.transaction.clone().into_consensus()) {
+        if let Some(tx) = self.pool().get(&hash).map(|tx| tx.transaction.clone_into_consensus()) {
             return Ok(Some(TransactionSource::Pool(tx)));
         }
 
--- a/crates/optimism/rpc/src/eth/transaction.rs
+++ b/crates/optimism/rpc/src/eth/transaction.rs
@@ -197,8 +197,7 @@ where
         {
             let transaction = tx
                 .try_into_recovered_unchecked()
-                .map_err(|_| EthApiError::InvalidTransactionSignature)
-                .map_err(Self::Error::from_eth_err)?;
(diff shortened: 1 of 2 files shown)


crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -179,6 +182,60 @@ where
     OpEthApiError: FromEvmError<N::Evm>,
     Rpc: RpcConvert<Primitives = N::Primitives, Error = OpEthApiError>,
 {
+    async fn transaction_by_hash(
```
0x00101010, line 184: Note: Steps 1 and 3 duplicate logic from the default trait implementation rather than extracting to a helper method. This keeps the override self-contained and avoids adding a helper method to the trait API that would only exist to support this one override.

Also we check flashblocks first before checking pool (in case people are using nodes with txpool gossip enabled to serve rpc requests)

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -609,45 +609,36 @@ pub trait LoadTransaction: SpawnBlocking + FullEthApiTypes + RpcNodeCoreExt {
     > + Send {
         async move {
             // Try to find the transaction on disk
-            let mut resp = self
+            if let Some((tx, meta)) = self
```
0x00101010, line 612: Reworked the function a bit to avoid unnecessary nested loops and improve readability

crates/optimism/rpc/src/eth/transaction.rs
```diff
@@ -179,6 +182,60 @@ where
...
+            let Some(indexed_tx) = pending_block.block().find_indexed(hash)
+        {
+            let recovered_tx = pending_block
+                .block()
+                .recovered_transaction(indexed_tx.index())
+                .expect("transaction must exist at index returned by find_indexed")
+                .cloned();
+
+            let meta = indexed_tx.meta();
+            return Ok(Some(TransactionSource::Block {
```
0x00101010, line 215: Did not choose to add a new FlashBlock source type, current Block type seem to represent the situation well

mattsse (approved):
lgtm, ty

PR #19957, opened 2025-11-25
docs(cli): enhance documentation for memory block buffer target configuration

### LINKED ISSUE #19946: Add a better CLI doc description of `--engine.memory-block-buffer-target`
### Describe the feature

The comment for this flag:
https://github.com/paradigmxyz/reth/blob/d55d641225798442f774249158d43a796e7e87da/crates/node/core/src/args/engine.rs#L24

does not completely explain what this flag actually does internally, we should document this, specifically how it relates to persistence and `--engine.persistence-threshold`

### Additional context

_No response_

0xSooki:
closes #19946 

Updated the documentation for the `--memory-block-buffer-target` flag.

### DIFF
--- a/crates/node/core/src/args/engine.rs
+++ b/crates/node/core/src/args/engine.rs
@@ -22,6 +22,11 @@ pub struct EngineArgs {
     pub persistence_threshold: u64,
 
     /// Configure the target number of blocks to keep in memory.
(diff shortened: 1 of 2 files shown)


mattsse (commented):
not needed

PR #19960, opened 2025-11-25
docs(consensus): fix misleading docstring in validate_4844_header_standalone

eeemmmmmm:
The docstring claimed things that the function doesn't actually do (checking excess_blob_gas) and said it doesn't check blob_gas_used when it clearly does. Fixed to match actual behavior.


mattsse (approved):
this is indeed accurate

PR #19961, opened 2025-11-25
fix: docs link in https://reth.rs/cli/cli

cuiweixie:
use vocs vite plugin to transfer link like ./reth.mdx to /cli/reth.


mattsse (approved):
lets try this

PR #16099, opened 2025-05-07
chore: eth69 status message support

shane-moore:
## Changes
- resolves https://github.com/paradigmxyz/reth/issues/13186?issue=paradigmxyz%7Creth%7C16073
- additionally, `SessionManager`, `SessionEvent`, `PeerInfo` etc have their nested prop `status: Status` updated to `status: StatusMessage`

where
```
pub enum StatusMessage {
    /// The legacy status (`eth/66` through `eth/68`) with `total_difficulty`.
    Legacy(Status),
    /// The new `eth/69` status with no `total_difficulty`.
    Eth69(StatusEth69),
}
```

Now, `eth_handshake` no longer has the dependency of legacy status as a param or return value. These have been updated to `StatusMessage` as well. Should be simpler now since no longer need to convert a potential eth69 status back to a legacy status in `eth_handshake`.  It's also nice that `SessionManager` no longer only supports legacy status

## PR Status
- code compiles

## Todo:
- fix tests
- resolve comments below

@mattsse, lmk if we like this approach, and I'll finish it up

PR #19963, opened 2025-11-25
fix(net): correct Debug impl for StatusEth69

strmfos:
The Debug impl for StatusEth69 was not updated when earliest/latest fields were added in #16099. Display was fixed but Debug was missed.

Fixed by updating the type name and adding the missing fields to match the Display implementation.


mattsse (approved):
fine

PR #19967, opened 2025-11-25
feat(trie): Benchmarks comparing proof_v2 to "legacy" proof implementation

mediocregopher:
## Summary

This PR adds criterion benchmarks comparing the "legacy" proof calculator to the new V2 implementation. It also includes two small perf-related fixes in the V2 implementation which came out of benching and profiling.

The results from the benchmarks show that the new implementation is faster across all inputs, and is especially faster for large target sets. This is mostly due to two factors:

* Not iterating over all targets for each calculated trie node.
* Being more mindful about memory allocations.

## Benchmarks

Benchmarks can be run by doing:

```
cargo bench --profile=profiling -p reth-trie --bench proof_v2 --features test-utils -- --verbose --measurement-time=20
```

Results will be printed to stdout, but an HTML rendering with graphs and details will also be generated at `target/criterion/report/index.html`.

The benchmarks compare both implementations across different combinations of dataset and target set sizes. For small datasets/target sets the V2 implementation does ~10% better, and with much less variance:

shot-1764079666

As expected, larger target sets produce more dramatic differences, as the legacy implementation checks all trie nodes against all targets, while the V2 implementation only checks a single target against each trie node. For a small data set and a larger target set we see a 17% improvement:

shot-1764079788

Results get better as data sets and target sizes go up. At a dataset size of 1024 keys and 512 targets we see a 30% improvement:

shot-1764079847

Similarly about 30% at 10240 keys and 2048 targets:

shot-1764080080


crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -95,10 +103,6 @@ impl<RF: DeferredValueEncoder> ProofTrieBranchChild<RF> {
             Self::RlpNode(_) => panic!("Cannot call `into_proof_trie_node` on RlpNode"),
         };
 
-        // Encode the `TrieNode` to the buffer, so we can return the `RlpNode` for it at the end.
```
mediocregopher, line 98: This `encode` was leftover from a previous iteration, it should have been deleted but snuck through because it doesn't actually cause incorrect results.

yongkangc (approved):
nice one! lgtm 🚀  🔥 🔥 

also very useful way to bench as well

crates/trie/trie/benches/proof_v2.rs
```diff
@@ -0,0 +1,162 @@
...
+                            .expect("Proof generation failed");
+                    },
+                    BatchSize::SmallInput,
+                );
+            });
+        }
+    }
+}
+
+criterion_group!(proof_comparison, bench_proof_algos);
```
yongkangc, line 161: nice use of criterion

shekhirin (approved):

PR #19969, opened 2025-11-25
chore: avoid redundant factory construction in witness/proof overlay init

Bashmunta:
Refactor TrieWitness::overlay_witness and DatabaseStorageProof overlay methods to initialize directly with InMemoryTrieCursorFactory and HashedPostStateCursorFactory instead of creating base factories and immediately replacing them, eliminating redundant instantiations without changing behavior; this mirrors the established pattern used in Proof overlay paths, maintains lifetime and API correctness, and slightly reduces overhead while improving consistency across trie db integration code


mediocregopher (approved):
I changed the title to `chore` because this won't have any meaningful impact on perf. Worth merging though as those `Self::from_tx` calls are weird imo

PR #19977, opened 2025-11-25
docs(trace): fix trace_call method invocation table

ANtutov:
The trace_call section showed the wrong RPC method and parameter shape in its “Method invocation” table, incorrectly using trace_callMany with [trace[], block]. This contradicted both the prose in the same section and the example that follows, which use trace_call with a transaction object as the first parameter, a list of trace types as the second parameter, and an optional block selector as the third. Updating the table to {"method": "trace_call", "params": [tx, ["trace"|"vmTrace"|"stateDiff"], block]} makes the documentation consistent and accurate, preventing user confusion and aligning the quick reference with the actual API behavior.


mattsse (approved):
this is fine, ty!

mattsse (approved):

PR #19982, opened 2025-11-26
feat: add next_tx_and_priority for Besttransactions

mattsse:
this way we can also bubble up the prio value of the transaction

### DIFF
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -191,6 +191,7 @@ impl<T: TransactionOrdering> BestTransactions<T> {
     }
 
     /// Returns the next best transaction and its priority value.
+    #[allow(clippy::type_complexity)]
     pub fn next_tx_and_priority(
         &mut self,
     ) -> Option<(Arc<ValidPoolTransaction<T::Transaction>>, Priority<T::PriorityValue>)> {
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -220,7 +220,7 @@ impl<T: TransactionOrdering> BestTransactions<T> {
                 // transactions are returned
                 self.mark_invalid(
                     &best.transaction,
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/pool/best.rs
```diff
@@ -189,6 +189,50 @@ impl<T: TransactionOrdering> BestTransactions<T> {
...
+                    &best.transaction,
+                    InvalidPoolTransactionError::Eip4844(
+                        Eip4844PoolTransactionError::NoEip4844Blobs,
+                    ),
+                )
+            } else {
+                if self.new_transaction_receiver.is_some() {
+                    self.last_priority = Some(best.priority.clone())
+                }
+                return Some((best.transaction, best.priority))
```
yongkangc, line 232: I see we bubble up priority over here

yongkangc (approved):

yongkangc:
Actually curious what might be the cases we want to bubble up priority? Is it for error handling

yongkangc:
Actually curious what might be the cases we want to bubble up priority? Is it for error handling e.g https://github.com/paradigmxyz/reth/pull/19981

crates/transaction-pool/src/pool/best.rs
```diff
@@ -189,6 +189,50 @@ impl<T: TransactionOrdering> BestTransactions<T> {
             }
         }
     }
+
+    /// Returns the next best transaction and its priority value.
+    #[allow(clippy::type_complexity)]
+    pub fn next_tx_and_priority(
+        &mut self,
+    ) -> Option<(Arc<ValidPoolTransaction<T::Transaction>>, Priority<T::PriorityValue>)> {
```
klkvr, line 197: I guess this could also just emit `PendingTransaction`?

klkvr (approved):

crates/transaction-pool/src/pool/best.rs
mattsse: this would force the pending tx internally and we really just need the prio and tx itself

PR #19984, opened 2025-11-26
docs(optimism): fix incorrect data layout comments in L1 block info parsing

eeemmmmmm:
The comments in parse_l1_info_tx_isthmus and parse_l1_info_tx_jovian were copy-pasted from Ecotone and never updated:

- Said "data layout assumed for Ecotone" in Isthmus/Jovian functions
- Listed offsets that included the 4-byte selector, but the code operates
  on data *after* the selector is stripped

For example, the comment claimed operator_fee_scalar was at offset 164, but the code reads data[160..164]. This made the comments misleading when debugging or modifying the parsing logic.

Fixed by rewriting comments to show offsets relative to the start of data (after selector removal), which now match the actual slice indices in code.


mattsse (commented):
dont want to change this

PR #19988, opened 2025-11-26
feat(network): add support for bootnodes_v4(enode) and bootnodes_v5(enr)

cuiweixie:
add support for bootnodes_v4(enode) and bootnodes_v5(enr) in config file similiar in geth

### DIFF
--- a/crates/node/core/src/args/network.rs
+++ b/crates/node/core/src/args/network.rs
@@ -283,11 +283,11 @@ impl NetworkArgs {
         default_peers_file: PathBuf,
     ) -> NetworkConfigBuilder<N> {
         let addr = self.resolved_addr();
-        
+
         // Resolve bootnodes with priority: CLI > Config file > Chain spec > Default
         let cli_bootnodes = self.resolved_bootnodes();
         let chain_bootnodes_default = chain_spec.bootnodes().unwrap_or_else(mainnet_nodes);
-        
+
         // For discv4: CLI > Config file > Chain spec > Default
         let discv4_bootnodes = if let Some(cli_nodes) = &cli_bootnodes {
             cli_nodes.clone()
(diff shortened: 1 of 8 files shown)


crates/node/core/src/args/network.rs
```diff
@@ -280,9 +283,61 @@ impl NetworkArgs {
...
+        } else if !config.peers.bootnodes_v4.is_empty() {
+            config
+                .peers
+                .bootnodes_v4
+                .iter()
+                .filter_map(|peer| peer.resolve_blocking().ok())
+                .collect()
+        } else {
+            chain_bootnodes_default
+        };
```
mattsse: I find this a bit hard to follow

I'd like to extract boot node detection logic to a dedicated function instead

crates/node/core/src/args/network.rs
```diff
@@ -318,7 +373,13 @@ impl NetworkArgs {
             // apply discovery settings
             .apply(|builder| {
                 let rlpx_socket = (addr, self.port).into();
-                self.discovery.apply_to_builder(builder, rlpx_socket, chain_bootnodes)
+                self.discovery.apply_to_builder_with_discv5_bootnodes(
+                    builder,
+                    rlpx_socket,
+                    chain_bootnodes,
+                    discv5_bootnodes_enode,
+                    discv5_bootnodes_enr,
```
mattsse: very confusing why need both here

mattsse (changes requested):
not opposed to this setting but this needs to be easier to follow, I admit all these settings are a bit complex especially discv5

so if we can make this easier to parse than I'm okay with this

could you perhaps elaborate on the use case here?

crates/node/core/src/args/network.rs
cuiweixie: change to using enr only.

crates/node/core/src/args/network.rs
cuiweixie: extact a function to process this.

PR #19990, opened 2025-11-26
feat(persistence): batch write hashed_state

### LINKED ISSUE #20609: Flatten HashedPostState before persisting
### Describe the feature

When persisting hashed state we do so block-by-block:
https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/storage/provider/src/providers/database/provider.rs#L355-L370

We could instead accumulate state into a single HashedPostState and persist it all at once.

### Additional context

_No response_

duyquang6:
As discussed https://github.com/paradigmxyz/reth/pull/19739#issuecomment-3563008832
Batch write of `hashed_state` is safe, so I created this PR to cherry-pick old reverted commit

### Changes
- batch write hashed_state

*Note: The after result here is after we improve extend_sorted_vec, which is upcoming PR

#### erc20 transfers spam: ~100ms

image

#### native transfers spam: ~50ms

image

#### erc20 transfers spam: ~80ms

image

#### native transfers spam: ~30ms

image

### DIFF
--- a/crates/trie/common/src/utils.rs
+++ b/crates/trie/common/src/utils.rs
@@ -89,3 +89,111 @@ mod tests {
         assert_eq!(target, vec![(1, "a"), (2, "b"), (3, "c_new")]);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_extend_sorted_vec_empty_other() {
+        let mut target = vec![(1, "a"), (2, "b"), (3, "c")];
+        let other: Vec<(i32, &str)> = vec![];
+        extend_sorted_vec(&mut target, &other);
+        assert_eq!(target, vec![(1, "a"), (2, "b"), (3, "c")]);
+    }
+
+    #[test]
+    fn test_extend_sorted_vec_empty_target() {
+        let mut target: Vec<(i32, &str)> = vec![];
+        let other = vec![(1, "a"), (2, "b"), (3, "c")];
+        extend_sorted_vec(&mut target, &other);
+        assert_eq!(target, vec![(1, "a"), (2, "b"), (3, "c")]);
+    }
+
+    #[test]
+    fn test_extend_sorted_vec_no_overlap() {
+        let mut target = vec![(1, "a"), (3, "c"), (5, "e")];
+        let other = vec![(2, "b"), (4, "d"), (6, "f")];
+        extend_sorted_vec(&mut target, &other);
+        assert_eq!(target, vec![(1, "a"), (2, "b"), (3, "c"), (4, "d"), (5, "e"), (6, "f")]);
+    }
+
+    #[test]
+    fn test_extend_sorted_vec_with_duplicates_other_precedence() {
+        let mut target = vec![(1, "a"), (2, "target"), (3, "c")];
+        let other = vec![(2, "other"), (4, "d")];
+        extend_sorted_vec(&mut target, &other);
+        // other's value should take precedence for key 2
+        assert_eq!(target, vec![(1, "a"), (2, "other"), (3, "c"), (4, "d")]);
+    }
+
+    #[test]
+    fn test_extend_sorted_vec_all_other_before_target() {
+        let mut target = vec![(5, "e"), (6, "f")];
+        let other = vec![(1, "a"), (2, "b"), (3, "c")];
+        extend_sorted_vec(&mut target, &other);
+        assert_eq!(target, vec![(1, "a"), (2, "b"), (3, "c"), (5, "e"), (6, "f")]);
+    }
+
+    #[test]
+    fn test_extend_sorted_vec_all_other_after_target() {
+        let mut target = vec![(1, "a"), (2, "b")];
+        let other = vec![(5, "e"), (6, "f"), (7, "g")];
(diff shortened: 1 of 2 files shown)


crates/trie/common/src/utils.rs
```diff
@@ -4,11 +4,10 @@ use alloc::vec::Vec;
 /// Values from `other` take precedence for duplicate keys.
 ///
 /// This function efficiently merges two sorted vectors by:
```
mattsse, line 7: can we highlight this better that both inputs __must__ be sorted already for this to work

crates/trie/common/src/utils.rs
```diff
@@ -18,36 +17,152 @@ where
...
-                    // Same key, update target with other's value
-                    target_item.1 = other_iter.next().unwrap().1.clone();
-                    break;
-                }
-                Ordering::Greater => {
-                    // Other item comes after current target item, keep target unchanged
-                    break;
+
+    loop {
+        match (target_iter.peek(), other_iter.peek()) {
```
mattsse: isn't this just

https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge

now?

mattsse (changes requested):
pedantic doc nit

crates/trie/common/src/utils.rs
duyquang6: yes, thank, that make sense. There is small caveat that:
```Values from `other` take precedence for duplicate keys.```
So will use `merge_join_by` instead

File: crates/trie/common/src/utils.rs (hunk 1 above)
Comment (duyquang6) [reply crates/trie/common/src/utils.rs:7]: Yeah it already at the top: `Helper function to extend a sorted vector with another sorted vector.`
but I'll add an explicit note emphasizing that both inputs must be sorted.

File: crates/trie/common/src/utils.rs [hunk 3]
```diff
@@ -18,36 +19,16 @@ where
...
-
-    // Append collected new items, as well as any remaining from `other` which are necessarily also
-    // new, and sort if needed
-    if !to_insert.is_empty() || other_iter.peek().is_some() {
-        target.extend(to_insert);
-        target.extend(other_iter.cloned());
-        target.sort_unstable_by(|a, b| a.0.cmp(&b.0));
-    }
+        })
+        .collect();
```
mediocregopher: The previous implementation was specifically designed to avoid having to do a big collect like this; the resulting memory allocation from this collect dwarfs any ostensible speedup you get from not having to sort. I just did a bench comparing your implementation to the previous and this new one is about 2x slower for synthetic datasets:

Image

You can see the bench [here](https://github.com/paradigmxyz/reth/blob/86d6a363dddea6dc7024cb702041b01cf45d61c9/crates/trie/common/benches/extend_sorted_vec.rs) if you're curious

duyquang6: very useful bench @mediocregopher 
when I first use old version with aggregated hashed state to bench, the result is not good. This is why I think there something wrong with  `extend_ref` or `extend_sorted_vec` 

when I use your bench compare with custom own merge version (not used merge_join_by), it only shine at other target size smaller than other size, but overall case, old version still win. That give me some hint to use this function better, is keep target size and other size similar or larger so might benefit old version 

**shine case (new better)**
image

image

**but overall (size similar or target size > other) old still better**
image

image

duyquang6: reuse old extend_sorted version + use trick to clone on first item, and extend_ref on remain items 
```
            if aggregated_hashed_post_state_sorted.is_empty() {
                aggregated_hashed_post_state_sorted = (*hashed_state).clone();
            } else {
                aggregated_hashed_post_state_sorted.extend_ref(&hashed_state);
            }
```

I can reach better perf - better than sequential insert version, but still not good as new version

**native transfer 35ms**

image

duyquang6: I dump the raw data of `HashedPostStateSorted` when bench with native-transfer

here is bench result of `extend_ref`, new version of both is better than in this testcase

image

can double check the bench [here](https://github.com/duyquang6/reth/blob/bench-sorted-extend/crates/trie/common/benches/extend_ref.rs#L286) - already attach hashed state raw data
Could be raw data might have properties that benchmark doesn’t fully cover :thinking: ?

mediocregopher: @duyquang6 these are interesting results, your benches for `extend_sorted_vec_comparison/t10_o1000` conflict with what I originally saw in mine, but now I'm able to replicate, so there's some inconsistency there that I still need to figure out.

If yours is faster for t10_o1000 I expect it's because it's doing the full allocation up-front, whereas mine is likely doing two larger allocations at the end with the `extend` calls.

What do you think about trying out something like:

```
    // Where "50" is a made up number that needs to be tuned
    if other.len() > target.len() * 50 {
        return extend_sorted_vec_custom(target, other);
    }

    extend_sorted_vec(target, other)
```

This way maybe we better cover all cases.

duyquang6:
> may i know what tool u used to do erc20/native transfers spam?

Hi, we wrote a custom Rust script for benchmarking transaction throughput

PR #19991, opened 2025-11-26
fix(mev): correct refund calculation in sim_bundle

zeroprooff:
Fixed a bug where multiple transactions with refund_percent in a bundle were getting incorrect refund amounts.

The problem was that each transaction calculated its refund from the already-reduced refundable_value instead of the original amount. This caused later transactions to receive progressively smaller payouts.

For example, with two txs requesting 50% refund each from 1000 wei:
- First tx: got 500 wei (correct)
- Second tx: got 250 wei (50% of remaining 500, wrong - should be 500)

The fix stores the original refundable_value before the loop and uses that for all calculations. This matches the validation logic that ensures the sum of all refund percentages doesn't exceed 100%.

### DIFF
--- a/crates/rpc/rpc/src/eth/sim_bundle.rs
+++ b/crates/rpc/rpc/src/eth/sim_bundle.rs
@@ -340,6 +340,8 @@ where
                 }
 
                 // After processing all transactions, process refunds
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah I see, this would always decrease the paylout value

PR #19995, opened 2025-11-26
perf(payload): move parent_header in PayloadBuilderStack::build_empty_payload

Snezhkko:
This change removes two unnecessary Arc clones of the parent header in PayloadBuilderStack::build_empty_payload by destructuring the owned PayloadConfig and moving parent_header directly into the branch-specific PayloadConfig. Since PayloadConfig::parent_header is an Arc<SealedHeader<...>> and the function takes config by value, cloning the Arc only increments and later decrements the refcount without any semantic benefit. The refactoring preserves behavior and type safety, as the impl requires both sides to share the same Primitives, ensuring the header type matches between branches. The result is slightly reduced refcount churn with no API changes or logic alterations.


mattsse (approved):
i allow it

PR #19996, opened 2025-11-26
docs: document missing config sections and fields

Galoretka:
This change updates the configuration reference to align with the current configuration schema in code. It adds missing sections for the stages.era and stages.prune configurations, includes the optional dir field for stages.etl, extends the sessions section with pending_session_timeout and the limits subtable, and documents additional peers options including max_backoff_count, trusted_nodes_resolution_interval, incoming_ip_throttle_duration, and connection_info.max_concurrent_outbound_dials, as well as reputation_weights.bad_announcement. It also introduces the static_files section for per-segment block partitioning. These updates are grounded in the definitions in 
crates/config/src/config.rs
crates/net/network-types/src/peers/config.rs
crates/net/network-types/src/session/config.rs
crates/net/network-types/src/peers/reputation.rs
ensuring flag names and defaults match the implementation. The ip_filter field in PeersConfig is intentionally omitted because it is not deserializable from TOML (#[serde(skip)]) and therefore cannot be configured via reth.toml.

### DIFF
--- a/docs/vocs/docs/pages/run/configuration.mdx
+++ b/docs/vocs/docs/pages/run/configuration.mdx
@@ -15,6 +15,7 @@ The default data directory is platform dependent:
 The configuration file contains the following sections:
 
 -   [`[stages]`](#the-stages-section) -- Configuration of the individual sync stages
+    -   [`era`](#era)
     -   [`headers`](#headers)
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is actually useful

PR #19997, opened 2025-11-26
Refine precompile cache generics and clean up clippy regressions

0xLogicalx:
Require cached spec ids to be Copy, letting CachedPrecompile avoid needless cloning on hot paths. Quiet clippy across the workspace by const-qualifying the no-op handler, tightening doc comments, and dropping an unnecessary cast. Gate E2E test imports by target and qualify wallet usage, plus pin RNG calls to concrete types so tests compile everywhere.


mattsse (commented):
maybe next attempt

PR #19998, opened 2025-11-26
perf(payload): move fields instead of cloning in PayloadBuilderStack::try_build

figtracer:
following #19995 i was wondering if we could do something like this to avoid a couple more clones


mattsse (approved):
yeah this is even better actually

PR #20012, opened 2025-11-27
test: avoid redundant factory init in trie fuzz tests

JacksonEi:
Initialize StateRoot and StorageRoot directly with final factories instead of creating base factories via `from_tx`/`from_tx_hashed` and immediately replacing them with `with_trie_cursor_factory`.


mattsse (commented):
this is just for fuzzing, we can keep this as is

PR #20013, opened 2025-11-27
fix: propagate-only pending listener dropping promoted txs

sashass1315:
Remove the batch-level early-return in on_new_pending_transaction that suppressed propagate-only notifications when the inserted transaction itself was non-propagatable; the underlying iter already enforces per-item filtering, so the pre-check incorrectly prevented promotions from being announced to network listeners; added an integration test that creates a nonce-gap scenario with a local non-propagatable tx unblocking an external tx and verify that All listeners receive both pending hashes while PropagateOnly listeners receive only the promotable external hash, ensuring promotions are visible to the network as intended.

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -604,16 +604,8 @@ where
     /// [`TransactionPool`](crate::TransactionPool) trait for a custom pool implementation
     /// [`TransactionPool::pending_transactions_listener_for`](crate::TransactionPool).
     pub fn on_new_pending_transaction(&self, pending: &AddedPendingTransaction<T::Transaction>) {
-        let propagate_allowed = pending.is_propagate_allowed();
-
         let mut transaction_listeners = self.pending_transaction_listener.lock();
         transaction_listeners.retain_mut(|listener| {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -604,16 +604,8 @@ where
...
         let mut transaction_listeners = self.pending_transaction_listener.lock();
         transaction_listeners.retain_mut(|listener| {
-            if listener.kind.is_propagate_only() && !propagate_allowed {
-                // only emit this hash to listeners that are only allowed to receive propagate only
-                // transactions, such as network
-                return !listener.sender.is_closed()
-            }
-
             // broadcast all pending transactions to the listener
             listener.send_all(pending.pending_transactions(listener.kind))
```
mattsse, line 610: right, this already respects the propagate property

https://github.com/paradigmxyz/reth/blob/36c62807dc83993c851e997bf0cfb6643f8f2afb/crates/transaction-pool/src/pool/mod.rs#L1149-L1152

mattsse (approved):
right, I think we should do this otherwise we might end up not announcing descendant transactions that were promoted but can be propagated

PR #20016, opened 2025-11-27
Fix: Ensure Isthmus withdrawals root validation for non-canonical parent state in Optimism engine

petrokamin10:
Previously, when `state_by_block_hash` failed, the validator returned `Ok(())`, effectively skipping withdrawals root verification for blocks whose parent was not yet in the canonical chain.  
This created a window where blocks could bypass a critical consensus check during reorgs or while processing non-canonical branches.

  - Fix `OpEngineValidator::validate_block_post_execution_with_hashed_state` to always validate the Isthmus withdrawals root, even when the parent block state is not yet canonical.  
  - Resolve state lookups by first trying historical state and then pending state; if neither is available, return a consensus error instead of silently accepting the block.


crates/optimism/node/src/engine.rs
```diff
@@ -127,11 +127,23 @@ where
         block: &RecoveredBlock<Self::Block>,
     ) -> Result<(), ConsensusError> {
         if self.chain_spec().is_isthmus_active_at_timestamp(block.timestamp()) {
-            let Ok(state) = self.provider.state_by_block_hash(block.parent_hash()) else {
-                // FIXME: we don't necessarily have access to the parent block here because the
-                // parent block isn't necessarily part of the canonical chain yet. Instead this
-                // function should receive the list of in memory blocks as input
```
mattsse, line 133: this is a bit more tricky than that because technically, the block's parent isn't necessarily part of the canonical chain yet
in which case we'd fail to resolve it via the provider which only gives access to the canonical chain + pending block

so this change doesnt fully cover all scenarios, but in almost all cases we ca n assume that this is the case

mattsse (commented):
this is a bit more tricky, and the current design for this is a bit of a footgun

I guess we'd need to extend this with like `parent_state: BlockState` to alway provide access to any parent state

mattsse:
Closing this PR as op-reth has been moved to its own repository. See https://github.com/paradigmxyz/reth/pull/21532 and the updated README for the new location. If this is still relevant, please re-open in the new op-reth repository. 🙏

PR #20017, opened 2025-11-27
fix(cli): always attempt graceful shutdown in run_command_until_exit 

radik878:
Previously, `CliRunner::run_command_until_exit()` only invoked `task_manager.graceful_shutdown_with_timeout(5s)` on successful command completion, while the error path skipped it and relied on TaskManager drop to propagate shutdown. This caused graceful tasks to be signaled but not awaited in error scenarios, leading to abrupt termination under load and risking incomplete cleanup. This change moves the graceful shutdown call out of the success branch so it runs unconditionally, ensuring graceful tasks get a bounded window to finish regardless of whether the command exited cleanly or with an error. The behavior remains consistent with the success path, preserves the existing 5s cap, and aligns with the TaskManager’s semantics without altering public APIs.


mattsse (commented):
if this terminates then we completed

PR #20018, opened 2025-11-27
fix: keep peer status unchanged on new discovered updates

mattsse:
retains peer status for existing peers on discovery level updates.

admin level operations are allowed to modify peer status


klkvr (approved):
lgtm

there is also this https://github.com/paradigmxyz/reth/blob/2a6c33d4eafd8c694722bfba6a9089cea8971ad9/crates/net/network/src/peers.rs#L855-L880

but probably doesn't matter as only invoked on demand

PR #20020, opened 2025-11-27
fix(net): back off slightly after graceful connection termination

mattsse:
see docs for reasoning

this should avoid any races if the connection was closed locally but might still exists on the remote (pending cleanup or ip throttle)


klkvr (approved):
nice, makes sense

PR #20022, opened 2025-11-27
fix: remove dead resolved_links state from SyncTree and its usage

Fibonacci747:
The resolved_links map in SyncTree was a dead state: it was only written to in DnsDiscoveryService::on_resolved_entry() and never read or cleared. This caused unnecessary allocations and clones of LinkEntry while providing no functional value. Deduplication of trees is already ensured by DnsDiscoveryService’s trees HashMap keyed by LinkEntry, and DNS entry results are cached by dns_record_cache. Removing resolved_links eliminates the redundant memory growth and avoids needless cloning without changing behavior. The change removes the field, its accessor, and the single write site, and cleans up an unused import.

### DIFF
--- a/crates/net/dns/src/lib.rs
+++ b/crates/net/dns/src/lib.rs
@@ -268,9 +268,6 @@ impl<R: Resolver> DnsDiscoveryService<R> {
                     }
                     DnsEntry::Link(link_entry) => {
                         if kind.is_link() {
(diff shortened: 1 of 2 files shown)


mattsse (approved):
oh I see, this would just collect and never clean up and in fact we dont need this

PR #20024, opened 2025-11-28
feat: docs for op-reth

cuiweixie:
op-geth has some difference form reth. I think we might need to add docs for op-reth, at least the team i served do need it.


mattsse (commented):
hmm, this duplicates a lot of docs, but since op-reth is technically its own binary with some additional args, I'm thinking this is appropriate

wdyt @shekhirin

.github/workflows/lint.yml
```diff
@@ -187,9 +187,10 @@ jobs:
         with:
           cache-on-failure: true
       - run: cargo build --bin reth --workspace --features ethereum
+      - run: cargo build --bin op-reth -p op-reth --workspace --features ethereum
```
shekhirin: this should be equivalent
```suggestion
      - run: cargo build --bin reth --workspace
      - run: cargo build --bin op-reth --workspace
```

shekhirin (approved):
Makes sense, just a small nit

PR #20026, opened 2025-11-28
fix(rpc): unify EthFilterConfig TTL default with constant

ANtutov:
Replaced the hardcoded 5-minute TTL in EthFilterConfig::default() with DEFAULT_STALE_FILTER_TTL in crates/rpc/rpc-eth-types/src/builder/config.rs to eliminate duplicate sources of truth and prevent future drift. EthConfig::default() already uses this constant, and EthFilterConfig::default() is used directly in some call sites (e.g., EthFilter::new examples), so aligning both ensures consistent behavior across construction paths without changing the effective default


mattsse (approved):
this is indeed the same

PR #20031, opened 2025-11-28
chore: drop jar early

mattsse:
at the cost of a single path alloc we can drop the jar early before doing some io


shekhirin (approved):
makes sense

PR #20034, opened 2025-11-28
chore(node): gate otlp-only code and silence unused warnings

conomist:
move OtlpConfig behind #[cfg(feature = "otlp")] so it only compiles when the feature is enabled
silence the unused layers parameter for builds without otlp by explicitly binding it under the same cfg guard
keeps the tracing logic untouched when the feature is on while removing noise for minimal builds

### DIFF
--- a/crates/node/core/src/args/trace.rs
+++ b/crates/node/core/src/args/trace.rs
@@ -4,6 +4,8 @@ use clap::Parser;
 use eyre::WrapErr;
 use reth_tracing::{tracing_subscriber::EnvFilter, Layers};
 use reth_tracing_otlp::OtlpProtocol;
+#[cfg(feature = "otlp")]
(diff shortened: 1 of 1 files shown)


crates/node/core/src/args/trace.rs
```diff
@@ -4,6 +4,8 @@ use clap::Parser;
 use eyre::WrapErr;
 use reth_tracing::{tracing_subscriber::EnvFilter, Layers};
 use reth_tracing_otlp::OtlpProtocol;
+#[cfg(feature = "otlp")]
+use reth_tracing_otlp::OtlpConfig;
```
mattsse: this should be replaced with full path imports

mattsse (changes requested):
can this also feature gate the deps entirely?

conomist (commented):
work has done according your comment

PR #20040, opened 2025-11-29
feat: generate part of sidebar

cuiweixie:
base on https://github.com/paradigmxyz/reth/pull/20032.
I think automately update sidebar is really need to be added here


mattsse (commented):
hmm, could you perhaps give a screenshot example what this changes?

mattsse (approved):
seems fine

PR #20042, opened 2025-11-29
fix(ipc): make IpcServer Debug impl generic

Bashmunta:
Made this change to provide Debug for all IpcServer generic instantiations because it aligns with upstream patterns, satisfies workspace lint for missing debug, and avoids restricting Debug to default middleware without introducing extra bounds.


mattsse (approved):
fine

PR #20044, opened 2025-11-30
feat: add [TransactionPoolExt::filter_pooled_txs] for efficient transaction filtering

### LINKED ISSUE #19919: Add TransactionPoolExt::filter_pooled_txs(Fn(&) -> bool)
### Describe the feature

this is intended as a cheaper variant for

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/transaction-pool/src/traits.rs#L310-L321

that uses a predicate to also filter txs

we should add this to this trait:

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/transaction-pool/src/traits.rs#L645-L645

for example we can improve this:

https://github.com/paradigmxyz/reth/blob/9f3949cd35626f2685073103ccde4a759df72832/crates/optimism/txpool/src/maintain.rs#L118-L118

## TODO
* add and impl fn
* update in maintain_transaction_pool_conditional

### Additional context

_No response_

AnInsaneJimJam:
### Summary
Adds filter_pooled_txs method to efficiently filter pooled transactions using a predicate, avoiding the need to collect all transactions before filtering.

Changes
- Added  TransactionPoolExtExt trait with filter_pooled_txsmethod
- Implemented efficient filtering in Pool that applies predicate during iteration
- Updated maintain_transaction_pool_conditional
 to use the new method
 
 This closes #19919

### DIFF
--- a/crates/transaction-pool/src/traits.rs
+++ b/crates/transaction-pool/src/traits.rs
@@ -678,6 +678,38 @@ pub trait TransactionPoolExt: TransactionPool {
     fn cleanup_blobs(&self);
 }
 
+/// Extension methods for [`TransactionPoolExt`].
+///
+/// This trait provides additional utility methods that can be implemented more efficiently
+/// by specific pool implementations.
+pub trait TransactionPoolExtExt: TransactionPoolExt {
+    /// Returns the _full_ transaction objects for all transactions in the pool that are allowed to
+    /// be propagated and match the given predicate.
+    ///
+    /// This is a cheaper variant of [`TransactionPool::pooled_transactions`] that allows filtering
+    /// transactions using a predicate without collecting all transactions first.
+    ///
+    /// The predicate is applied to each transaction, and only transactions for which the predicate
+    /// returns `true` are included in the result.
+    ///
+    /// Note: This returns a `Vec` but should guarantee that all transactions are unique.
+    ///
+    /// Caution: In case of blob transactions, this does not include the sidecar.
+    ///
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/traits.rs
```diff
@@ -668,6 +678,38 @@ pub trait TransactionPoolExt: TransactionPool {
...
+    ///
+    /// Caution: In case of blob transactions, this does not include the sidecar.
+    ///
+    /// Consumer: Maintenance tasks, P2P
+    fn filter_pooled_txs<F>(
+        &self,
+        mut predicate: F,
+    ) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>>
+    where
+        F: FnMut(&Arc<ValidPoolTransaction<Self::Transaction>>) -> bool,
```
mattsse, line 705: can we move this to the existing trait instead, so that this can be specialized, because this is effectively how it was before

mattsse (changes requested):
the function lgtm, but we should move this into the existing TransactionPoolExt so that this can be specialized by implementers

crates/transaction-pool/src/traits.rs
AnInsaneJimJam: @mattsse auto_impl was not compiling due to methos with generic parameters. How Should I proceed regarding this?

mattsse (commented):
the optimism impls was removed so we ccan close this here, sorry about this

PR #20045, opened 2025-11-30
feat(metrics): add auth support for Prometheus push gateway mode

mablr:
https://github.com/paradigmxyz/reth/pull/19243#pullrequestreview-3384565326 follow-up.

Add basic auth support for Prometheus push gateway mode.

It's unfortunately not practicable to use directly the built-in [`PrometheusBuilder::with_push_gateway`](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.with_push_gateway), see https://github.com/metrics-rs/metrics/issues/634, so that's why I think PR may be relevant.


crates/node/core/src/args/metric.rs
```diff
@@ -28,8 +25,33 @@ pub struct MetricArgs {
...
+    /// Must be used together with `--metrics.prometheus.push.password`.
+    /// Only applies when a push gateway URL is configured.
+    #[arg(
+        long = "metrics.prometheus.push.username",
+        value_name = "USERNAME",
+        requires = "push_gateway_url",
+        requires = "push_gateway_password",
+        env = "RETH_METRICS_PUSH_USERNAME"
+    )]
+    pub push_gateway_username: Option<String>,
```
mattsse, line 41: should we perhaps just combine these two into one arg?

crates/node/core/src/args/metric.rs
```diff
@@ -28,8 +25,33 @@ pub struct MetricArgs {
         long = "metrics.prometheus.push.interval",
         default_value = "5",
         value_parser = parse_duration_from_secs,
-        value_name = "SECONDS",
-        help_heading = "Metrics"
```
mattsse, line 32: should we keep the heading here?

crates/node/builder/src/launch/common.rs
```diff
@@ -614,7 +614,12 @@ where
                         }
                     })
                     .build(),
-            ).with_push_gateway(self.node_config().metrics.push_gateway_url.clone(), self.node_config().metrics.push_gateway_interval);
+            ).with_push_gateway(
+                self.node_config().metrics.push_gateway_url.clone(),
+                self.node_config().metrics.push_gateway_interval,
+                self.node_config().metrics.push_gateway_username.clone(),
+                self.node_config().metrics.push_gateway_password.clone(),
```
mattsse, line 648: if we're only using these together perhaps we can make this a single Option instead?

mattsse (changes requested):

crates/node/core/src/args/metric.rs
mablr: As all fields of `MetricArgs` had this, I replaced them by adding to the struct `#[command(next_help_heading = "Metrics")]`.

I can revert this btw, and set `help_heading` for each arg.

crates/node/core/src/args/metric.rs
mablr: Do you mean :
- a single arg "username:password" that will be parsed ?
- or an inner flatten optional Clap arg struct containing username and password fields ?

mablr: I realize it should be possible to set username and leave password empty. (I'll fix that)

So the "username:password" argument to parse is definitely too complex compared to the benefit of having a single auth arg.

PR #20047, opened 2025-12-01
perf(trie): add HashedPostStateSorted::from_reverts

yongkangc:
## Summary

Adds `HashedPostStateSorted::from_reverts` that returns sorted hashed state directly from database changesets, eliminating the intermediate unsorted `HashedPostState` and the separate `.into_sorted()` call.

Closes #19382

## Benchmark (6% Faster)

Command: `cargo bench -p reth-trie-db from_reverts` on a synthetic dataset seeded into a temp DB: 256 unique accounts, each with 4 storage slots, all written in a single block of change sets.

- `unsorted_then_sort` (`HashedPostState::from_reverts().into_sorted()`): ~964µs (949–985µs)
- `direct_sorted` (`HashedPostStateSorted::from_reverts()`): ~906µs (902–910µs)

**~6% faster** for this dataset by skipping the intermediate unsorted post-state allocation and sort step while preserving semantics.

Benchmark violin plot
- bench code found: [298504a](https://github.com/paradigmxyz/reth/pull/20047/commits/298504ae734ab1854b51ba99fce10b98b779bf3c)

Other Bench for from reverts: https://github.com/paradigmxyz/reth/pull/20047#issuecomment-3599964129

## Impact

- Overlay state uses the sorted-from-reverts path directly, removing the redundant conversion and the linked TODO.
- Provides a public API for consumers that need trie-ready sorted state without materializing the unsorted map first.

### Future Works

I have another issue in the works, where we make use of `HashedPostStateSorted`. However that might be more breaking and I want to limit to just this scope. 
https://github.com/paradigmxyz/reth/issues/20048


crates/trie/common/src/hashed_state.rs
```diff
@@ -22,7 +22,9 @@ use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 
 use revm_database::{AccountStatus, BundleAccount};
 
-/// Representation of in-memory hashed state.
+/// In-memory hashed state that stores account and storage changes with keccak256-hashed keys in
+/// hash maps For state root computation, convert to [`HashedPostStateSorted`] which enables
+/// efficient ordered iteration required by the Merkle Patricia Trie.
```
mediocregopher: Not sure this extra information in the comment is helpful
```suggestion
/// In-memory hashed state that stores account and storage changes with keccak256-hashed keys in
/// hash maps.
```

crates/trie/db/src/state.rs
```diff
@@ -273,16 +302,62 @@ impl<TX: DbTx> DatabaseHashedPostState<TX> for HashedPostState {
...
+    ///   (`HashedPostStateSorted`) for cursors or trie iteration (e.g. overlay state roots,
+    ///   parallel state root, proof generation). This skips the intermediate `HashMap` allocation
+    ///   and sort done by `into_sorted`.
+    /// - Keep using the unsorted variant if you need to mutate/merge in map form first.
+    #[instrument(target = "trie::db", skip(tx), fields(range))]
+    fn from_reverts<KH: KeyHasher>(
+        tx: &TX,
+        range: impl RangeBounds<BlockNumber>,
+    ) -> Result<Self, DatabaseError> {
+        let (accounts, storages) = read_reverts_from_changesets(tx, range)?;
```
mediocregopher: One of the main objectives of this change is to eliminate unnecessary allocations; by reading reverts into this intermediate dataset you're kind of negating that. Data in MDBX is already sorted, so it should be sufficient to read it directly into Vecs and build the HashedPostStateSorted directly from those Vecs. You can check [trie_reverts](https://github.com/paradigmxyz/reth/blob/4f5b842543197dd28faea9f9424a6e46432341e3/crates/storage/provider/src/providers/database/provider.rs#L2235) implementation, this should be basically identical to it.

crates/trie/db/src/state.rs
```diff
@@ -135,6 +136,17 @@ pub trait DatabaseHashedPostState<TX>: Sized {
     ) -> Result<Self, DatabaseError>;
 }
 
+/// Extends [`HashedPostStateSorted`] with operations specific for working with a database
+/// transaction.
+pub trait DatabaseHashedPostStateSorted<TX>: Sized {
```
mediocregopher: Rather than introducing this new trait, could we make the existing one return the sorted data? If it means we have to `into::<HashedPostState>()` in some call sites that'd be preferable imo, we can address those in the other PR. Basically I would like to move away from using the unsorted data completely, as it's essentially always going to be converted to the sorted form in the end anyway.

mediocregopher (dismissed):

crates/trie/db/src/state.rs
yongkangc: nice! second moving away from unsorted as well :)

crates/trie/db/src/state.rs
```diff
@@ -112,26 +112,13 @@ pub trait DatabaseStateRoot<'a, TX>: Sized {
...
-        input: TrieInput,
-    ) -> Result<(B256, TrieUpdates), StateRootError>;
-
     /// Calculates the state root for provided [`HashedPostStateSorted`] using cached intermediate
     /// nodes.
-    fn overlay_root_from_nodes_sorted(
-        tx: &'a TX,
-        input: TrieInputSorted,
-    ) -> Result<B256, StateRootError>;
+    fn overlay_root_from_nodes(tx: &'a TX, input: TrieInputSorted) -> Result<B256, StateRootError>;
```
yongkangc, line 117: we are changing to make use of `TrieInputSorted` so that 

we can do this
https://github.com/paradigmxyz/reth/pull/20050

crates/trie/db/src/state.rs
```diff
@@ -99,40 +96,43 @@ pub trait DatabaseStateRoot<'a, TX>: Sized {
...
+    fn overlay_root(tx: &'a TX, post_state: &HashedPostStateSorted)
+        -> Result<B256, StateRootError>;
 
-    /// Calculates the state root for this [`HashedPostState`] and returns it alongside trie
+    /// Calculates the state root for this [`HashedPostStateSorted`] and returns it alongside trie
     /// updates. See [`Self::overlay_root`] for more info.
     fn overlay_root_with_updates(
         tx: &'a TX,
-        post_state: HashedPostState,
+        post_state: &HashedPostStateSorted,
```
yongkangc, line 112: This is the tldr of the change where we make use of `HashedPostStateSorted` instead of HashedPostState for overlay root

crates/trie/common/src/hashed_state.rs
```diff
@@ -507,6 +508,46 @@ impl HashedPostStateSorted {
         self.accounts.len() + self.storages.values().map(|s| s.len()).sum::<usize>()
     }
 
+    /// Construct [`TriePrefixSetsMut`] from hashed post state.
+    ///
+    /// The prefix sets contain the hashed account and storage keys that have been changed in the
+    /// post state.
+    pub fn construct_prefix_sets(&self) -> TriePrefixSetsMut {
```
yongkangc, line 539: behaviour is same as unsorted HashPostState - to build trie prefix

crates/chain-state/src/memory_overlay.rs
```diff
@@ -124,26 +124,34 @@ impl<N: NodePrimitives> AccountReader for MemoryOverlayStateProviderRef<'_, N> {
...
+    fn state_root_from_nodes(&self, mut input: TrieInputSorted) -> ProviderResult<B256> {
+        // Sort internal state and merge with sorted input
+        let internal_sorted = TrieInputSorted::from_unsorted(self.trie_input().clone());
+        input.prepend_self(internal_sorted);
         self.historical.state_root_from_nodes(input)
     }
 
     fn state_root_with_updates(
         &self,
         state: HashedPostState,
```
yongkangc, line 137: In the future, we should change this to accept sorted, maybe in the next pr since its q breaking

yongkangc:
after this experience, i really feel like we should change all unsorted to sorted - might be quite breaking but i think wanna do some cleanup there

for instance MemoryOverlayStateProviderRef should be just using sorted data

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -287,14 +289,15 @@ impl<Provider: DBProvider + BlockNumReader> StateRootProvider
...
+        let hashed_state_sorted = hashed_state.into_sorted();
+        revert_state.extend_ref(&hashed_state_sorted);
+        StateRoot::overlay_root(self.tx(), &revert_state)
             .map_err(|err| ProviderError::Database(err.into()))
     }
 
     fn state_root_from_nodes(&self, mut input: TrieInput) -> ProviderResult<B256> {
-        input.prepend(self.revert_state()?);
-        StateRoot::overlay_root_from_nodes(self.tx(), input)
+        input.prepend(self.revert_state()?.into());
```
yongkangc, line 299: this should be another PR where would we work with TrieInputSorted directly and merge sorted data. But that requires some additional function and can be breaking. 

should be another PR to cleanup in the future

and since this is not on the hot path of new payload

crates/trie/db/src/state.rs
```diff
@@ -174,115 +174,166 @@ impl<'a, TX: DbTx> DatabaseStateRoot<'a, TX>
...
+impl<TX: DbTx> DatabaseHashedPostState<TX> for HashedPostStateSorted {
+    /// Builds a sorted hashed post-state from reverts.
+    ///
+    /// Reads MDBX data directly into Vecs, using `HashSet`s only to track seen keys.
+    /// This avoids intermediate `HashMap` allocations since MDBX data is already sorted.
+    ///
+    /// - Reads the first occurrence of each changed account/storage slot in the range.
+    /// - Hashes keys and returns them already ordered for trie iteration.
     #[instrument(target = "trie::db", skip(tx), fields(range))]
     fn from_reverts<KH: KeyHasher>(
```
yongkangc, line 246: the core logic

crates/trie/db/src/state.rs
```diff
@@ -174,115 +174,166 @@ impl<'a, TX: DbTx> DatabaseStateRoot<'a, TX>
         Self::incremental_root_calculator(tx, range)?.root_with_progress()
     }
 
-    fn overlay_root(tx: &'a TX, post_state: HashedPostState) -> Result<B256, StateRootError> {
+    fn overlay_root(
+        tx: &'a TX,
+        post_state: &HashedPostStateSorted,
+    ) -> Result<B256, StateRootError> {
         let prefix_sets = post_state.construct_prefix_sets().freeze();
-        let state_sorted = post_state.into_sorted();
```
mediocregopher, line 179: These `into_sorted` calls have been moved outside this trait impl into the caller, in this case the historical state provider 👍

crates/trie/db/src/state.rs
```diff
@@ -174,115 +174,166 @@ impl<'a, TX: DbTx> DatabaseStateRoot<'a, TX>
...
         let mut account_changesets_cursor = tx.cursor_read::<tables::AccountChangeSets>()?;
+
         for entry in account_changesets_cursor.walk_range(account_range)? {
             let (_, AccountBeforeTx { address, info }) = entry?;
-            accounts.entry(address).or_insert(info);
+            if seen_accounts.insert(address) {
+                accounts.push((KH::hash_key(address), info));
+            }
         }
+        accounts.sort_unstable_by_key(|(hash, _)| *hash);
```
mediocregopher, line 263: This is necessary because we are not storing the hashed addresses/slots in mdbx, so we are not able to take advantage of its sorting.

mediocregopher (approved):
LGTM 👍 cc @mattsse for a second opinion

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -310,9 +310,11 @@ fn run_case(
         // Compute and check the post state root
         let hashed_state =
             HashedPostState::from_bundle_state::<KeccakKeyHasher>(output.state.state());
-        let (computed_state_root, _) =
-            StateRoot::overlay_root_with_updates(provider.tx_ref(), hashed_state.clone())
-                .map_err(|err| Error::block_failed(block_number, program_inputs.clone(), err))?;
+        let (computed_state_root, _) = StateRoot::overlay_root_with_updates(
+            provider.tx_ref(),
+            &hashed_state.clone().into_sorted(),
```
mattsse: I wonder if it's more efficient to add `clone_into_sorted` that uses `iter().cloned()` then we don't clone the map metadata

https://github.com/paradigmxyz/reth/blob/82a186104311c660e77982883542754542f1f206/crates/trie/common/src/hashed_state.rs#L365-L377

mattsse (approved):
makes sense, one idea for clone + into but also not sure if iter +cloned is any different

mediocregopher (approved):

testing/ef-tests/src/cases/blockchain_test.rs
yongkangc: @mattsse thanks 👍🏻 this is a good idea to avoid HashMap Metadata clone, and skip the alloc from hashmap

i.e skips the intermediate HashMap allocation entirely and clones key-value pairs directly into the final Vec.

yongkangc: Ran benchmarks comparing `clone().into_sorted()` vs `clone_into_sorted`:

**HashedPostState** (accounts x slots per account):
| Case | clone().into_sorted() | clone_into_sorted | Improvement |
|------|----------------------|-------------------|-------------|
| 64x0 | 2.12 us | 1.98 us | **6.6% faster** |
| 64x8 | 22.4 us | 14.4 us | **35.8% faster** |
| 64x64 | 182 us | 171 us | **6.0% faster** |
| 256x0 | 10.2 us | 9.2 us | **9.5% faster** |
| 256x8 | 90.2 us | 59.9 us | **33.6% faster** |
| 256x64 | 1.39 ms | 0.81 ms | **41.6% faster** |
| 1024x0 | 49.9 us | 46.4 us | **7.0% faster** |
| 1024x8 | 520 us | 378 us | **27.4% faster** |
| 1024x64 | 6.52 ms | 4.94 ms | **24.2% faster** |

**HashedStorage** (single account storage):
| Slots | Type | clone().into_sorted() | clone_into_sorted | Improvement |
|-------|------|----------------------|-------------------|-------------|
| 16 | dirty | 346 ns | 316 ns | **8.6% faster** |
| 16 | wiped | 339 ns | 293 ns | **13.4% faster** |
| 128 | dirty | 4.75 us | 4.81 us | ~flat |
| 128 | wiped | 4.78 us | 4.65 us | **2.9% faster** |
| 1024 | dirty | 53.4 us | 53.5 us | ~flat |
| 1024 | wiped | 53.5 us | 52.0 us | **2.8% faster** |

> **Note**: `wiped` = storage was cleared (e.g., contract SELFDESTRUCT), `dirty` = individual slot updates without a full wipe. The number indicates the count of storage slots.

**Summary**: `clone_into_sorted` is consistently faster, with **significant wins (24-42%) on larger states with storage**. The optimization avoids cloning HashMap metadata and directly iterates into sorted Vecs.

[Violin plots (gist)](https://gist.github.com/yongkangc/7f47ad27bc6b6d420e2b68771254657b)

Code: https://gist.github.com/yongkangc/7f47ad27bc6b6d420e2b68771254657b?permalink_comment_id=5889717#gistcomment-5889717

Issue #20048, opened 2025-12-01
make use of HashedPostState Sorted


yongkangc:
### Describe the feature

`DatabaseStateRoot::overlay_root` takes an unsorted `HashedPostState` and sorts it internally:

https://github.com/paradigmxyz/reth/blob/main/crates/trie/db/src/state.rs#L188-L197

For callers that already have `HashedPostStateSorted` (e.g. from `HashedPostStateSorted::from_reverts`), this forces an unnecessary conversion path.

Add `overlay_root_sorted` and `overlay_root_with_updates_sorted` variants that accept `HashedPostStateSorted` directly.

This would then get used e.g. in HistoricalStateProviderRef which currently calls `from_reverts` and then passes through `overlay_root`:

https://github.com/paradigmxyz/reth/blob/main/crates/storage/provider/src/providers/state/historical.rs#L288-L292

### Additional context

**Other areas to change:**
- [MerkleChangeSets stage](https://github.com/paradigmxyz/reth/blob/main/crates/stages/stages/src/stages/merkle_changesets.rs): it builds per-block reverts with HashedPostState::from_reverts to assemble cumulative reverts and trie inputs. That code ends up calling state-root routines that sort later; switching to HashedPostStateSorted::from_reverts plus the sorted overlay APIs would avoid the unsorted allocations in this loop-heavy stage.

PR #20050, opened 2025-12-01
perf(trie): add overlay root calculations for HashedPostStateSorted

yongkangc:
Closes: https://github.com/paradigmxyz/reth/issues/20048

## Summary

Adds sorted variants of overlay root methods that accept pre-sorted data directly, avoiding redundant conversions when callers already have sorted data. Migrates historical provider to use these paths.

For historical state root queries (e.g., `eth_getProof` at past blocks):

- **Skips HashMap allocation for state representation** - sorted path uses `Vec` instead of `B256Map` for accounts/storages
- **Skips Vec allocation + O(n log n) sort** - no `into_sorted()` conversion needed
- **Better cache locality** - sequential Vec iteration vs random HashMap iteration in prefix set construction

## Changes

- `HashedPostStateSorted::construct_prefix_sets_sorted()` - builds prefix sets by iterating sorted Vecs directly
- `overlay_root_sorted{_with_updates}` - accepts `&HashedPostStateSorted` directly
- `overlay_root_from_nodes_sorted{_with_updates}` - accepts `TrieInputSorted`
- Historical provider: `revert_state()` now returns `HashedPostStateSorted`, uses sorted APIs throughout

### Benches: 
https://github.com/paradigmxyz/reth/pull/20050#issuecomment-3594894230

### Ref
Builds on https://github.com/paradigmxyz/reth/pull/20047


yongkangc:
Bench: overlay_root bench (reth-trie-db)

Dataset: 1 block, 256 accounts × 4 slots seeded into Account/StorageChangeSets. Keccak hashing.

Results (Criterion, 100 samples, flat sampling):
- `overlay_root` unsorted (`HashedPostState`) → **1.826 ms** median
- `overlay_root_sorted` (`HashedPostStateSorted`) → **1.699 ms** median

Impact: Sorted fast-path is ~7.5% faster for this setup by skipping the HashMap→Vec conversion and using the sorted prefix-set builder. Hashing still dominates; larger/denser workloads may show bigger gaps.

image



Code: [61ed453](https://github.com/paradigmxyz/reth/pull/20050/commits/61ed453e28d0e4a6127f8c49ec5d26258c19e99f)

Command: `CARGO_TARGET_DIR=./target cargo bench -p reth-trie-db overlay_root`

crates/trie/common/src/hashed_state.rs
```diff
@@ -509,6 +509,53 @@ impl HashedPostStateSorted {
...
+    /// This mirrors [`HashedPostState::construct_prefix_sets`] but assumes the state is already
+    /// sorted, so it can avoid rebuilding hash maps or sorting keys a second time. This is the
+    /// fast-path for callers that already hold `HashedPostStateSorted` (e.g. `from_reverts`
+    /// sorted, sorted trie inputs).
+    ///
+    /// Behavior is identical to the unsorted path:
+    /// - destroyed accounts are tracked in `destroyed_accounts`
+    /// - storage wipes yield `PrefixSetMut::all()`
+    /// - storage overlays also register the account prefix to ensure trie walks cover it
+    pub fn construct_prefix_sets_sorted(&self) -> TriePrefixSetsMut {
```
mediocregopher, line 523: nit: including "sorted" in the name here makes it sound like the prefix set is sorted. I think it's fine for this to be the same name as on the unsorted HashedPostState
```suggestion
    pub fn construct_prefix_sets(&self) -> TriePrefixSetsMut {
```

crates/trie/db/src/state.rs
```diff
@@ -108,13 +109,27 @@ pub trait DatabaseStateRoot<'a, TX>: Sized {
     /// The state root for this [`HashedPostState`].
     fn overlay_root(tx: &'a TX, post_state: HashedPostState) -> Result<B256, StateRootError>;
 
+    /// Calculate the state root for this [`HashedPostStateSorted`], skipping the unsorted map
+    /// construction and sort.
+    fn overlay_root_sorted(
```
mediocregopher, line 114: Rather than add new methods, I would prefer to modify the existing ones to accept the sorted forms of HashedPostState and TrieInput. Each of this is already essentially calling `into_sorted` immediately inside the implementation, e.g. [here](https://github.com/paradigmxyz/reth/blob/4f5b842543197dd28faea9f9424a6e46432341e3/crates/trie/db/src/state.rs#L179).

mediocregopher (changes requested):

PR #20053, opened 2025-12-01
fix(rpc): check correct variable for division by zero in blob gas ratio

strmfos:
The `checked_blob_gas_used_ratio` function was checking the wrong variable.
It checked `blob_gas_used == 0` but should check `max_blob_gas_per_block == 0`to prevent division by zero.

The doc comment even says "accounting for the case where max_blob_gas_per_block is zero" but the code was checking blob_gas_used instead. This caused the function to return `inf` when called with non-zero blob_gas_used and zero max_blob_gas_per_block.


mattsse (approved):
originally this was because this only checked if there's even blob gas used, but if we use the maxblock gas check here then this is equivalent and even more correct

PR #20056, opened 2025-12-01
fix(provider): correct capacity calculation for inclusive block range

Fallengirl:
Capacity for RangeInclusive should be `end - start + 1`.
Matches the pattern already used in consistent.rs and bodies.rs.


mattsse (commented):
we can keep this as is

PR #20057, opened 2025-12-01
fix(rpc): improve/preserve error messages for TransactionConversionError

### LINKED ISSUE #20052: Improve error message for TransactionConversionError
### Describe the feature

we're currently ignoring the message for
https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/mod.rs#L166-L168

when doing:

https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/mod.rs#L339-L343

## TODO
* convert this into an enum: https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-convert/src/transaction.rs#L486-L489 with the string as one variant
* introduce other variant for cases like https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L436-L436

### Additional context

_No response_

figtracer:
fixes #20052

### DIFF
--- a/crates/rpc/rpc-convert/src/transaction.rs
+++ b/crates/rpc/rpc-convert/src/transaction.rs
@@ -486,28 +486,12 @@ where
 #[derive(Debug, thiserror::Error)]
 pub enum TransactionConversionError {
     /// Required fields are missing from the transaction request.
-    #[error("required fields missing from transaction request")]
-    MissingRequiredFields,
+    #[error("Failed to convert transaction into RPC response: {0}")]
+    FromTxReq(String),
 
-    /// Conversion from `Consensus` failed.
-    #[error(transparent)]
-    FromConsensus(#[from] FromConsensusError),
-}
-
-/// Error when converting from consensus transaction type.
-#[derive(Debug, thiserror::Error)]
-pub enum FromConsensusError {
-    /// Deposit transactions are special system transactions that are directly
-    /// included in blocks by the sequencer/validator and never enter the mempool.
-    #[error("deposit transactions cannot be pooled")]
-    DepositTransaction,
-    /// Blob transactions without sidecars cannot be pooled. After being included
-    /// in a block, the sidecar data is pruned, making the transaction unpoolable.
-    #[error("blob transaction missing sidecar")]
-    BlobSidecarMissing,
-    /// Transaction type is not supported for pool submission.
-    #[error("transaction type not supported for pool")]
-    UnsupportedForPool,
+    /// Other conversion errors.
+    #[error("{0}")]
+    Other(String),
 }
 /// Generic RPC response object converter for `Evm` and network `Network`.
 ///
@@ -847,7 +831,7 @@ where
         Ok(self
             .sim_tx_converter
             .convert_sim_tx(request)
-            .map_err(|_| TransactionConversionError::MissingRequiredFields)?)
+            .map_err(|e| TransactionConversionError::FromTxReq(e.to_string()))?)
     }
 
     fn tx_env(
--- a/crates/rpc/rpc-convert/src/transaction.rs
+++ b/crates/rpc/rpc-convert/src/transaction.rs
@@ -485,8 +485,15 @@ where
 
 /// Conversion into transaction RPC response failed.
 #[derive(Debug, Clone, Error)]
-#[error("Failed to convert transaction into RPC response: {0}")]
-pub struct TransactionConversionError(String);
+pub enum TransactionConversionError {
+    /// Conversion from `TxReq` to `SimTx` failed
+    #[error("Failed to convert RPC transaction request: {0}")]
+    FromTxReq(String),
+
+    /// Conversion from `Consensus` type failed
+    #[error("Failed to convert transaction from consensus: {0}")]
+    FromConsensus(String),
+}
 
 /// Generic RPC response object converter for `Evm` and network `Network`.
 ///
@@ -826,7 +833,7 @@ where
         Ok(self
             .sim_tx_converter
             .convert_sim_tx(request)
-            .map_err(|e| TransactionConversionError(e.to_string()))?)
+            .map_err(|e| TransactionConversionError::FromTxReq(e.to_string()))?)
     }
 
     fn tx_env(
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -484,9 +483,16 @@ where
...
-pub struct TransactionConversionError(String);
+#[derive(Debug, thiserror::Error)]
+pub enum TransactionConversionError {
+    /// Conversion from `TxReq` to `SimTx` failed
+    #[error("Failed to convert RPC transaction request: {0}")]
+    FromTxReq(String),
+
+    /// Conversion from `Consensus` type failed
+    #[error("Failed to convert transaction from consensus: {0}")]
+    FromConsensus(String),
```
emhane: matching on string isn't type safe. better make new error enums `FromTxReqError` and `FromConsensusError` if you need, with variants for each possible error.

emhane (changes requested):

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -485,15 +485,38 @@ where
...
+    FromConsensus(#[from] FromConsensusError),
 }
 
+/// Error when converting from RPC transaction request to simulated transaction.
+#[derive(Debug, thiserror::Error)]
+pub enum FromTxReqError {
+    /// Required fields are missing from the transaction request.
+    #[error("required fields missing from transaction request")]
+    MissingRequiredFields,
+}
```
figtracer: what variants are important here? @emhane

emhane: since you are able to have 
```rust
pub enum TransactionConversionError {
    /// Conversion from `TxReq` to `SimTx` failed.
    #[error(transparent)]
    FromTxReq(#[from] FromTxReqError),
    ..
}
```
instead of
```rust
pub enum TransactionConversionError {
    /// Conversion from `TxReq` to `SimTx` failed.
    #[error(transparent)]
    FromTxReq(String),
    ..
}
```
already, that means you have covered all the necessary variants. I agree that it doesn't make sense to have an enum with one variant only, so let's collapse this by removing this error and simply having
```rust
pub enum TransactionConversionError {
    /// Required fields are missing from the transaction request.
    #[error("required fields missing from transaction request")]
    MissingRequiredFields,

    /// Conversion from `Consensus` failed.
    #[error(transparent)]
    FromConsensus(#[from] FromConsensusError),
}
```
ok?

emhane (changes requested):

crates/rpc/rpc-convert/src/transaction.rs
figtracer: on it

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -826,7 +847,7 @@ where
         Ok(self
             .sim_tx_converter
             .convert_sim_tx(request)
-            .map_err(|e| TransactionConversionError(e.to_string()))?)
+            .map_err(|_| TransactionConversionError::MissingRequiredFields)?)
```
emhane: we don't want to loose the error returned by `convert_sim_tx` as it may have valuable info, what error type is that? we should have the `MissingRequiredFields` variant wrap that error type i.e. `MissingRequiredFields(TheErrorType)`

emhane (dismissed):

crates/rpc/rpc-convert/src/transaction.rs
figtracer: this is `convert_sim_tx`:
```rust
pub trait SimTxConverter<TxReq, SimTx>: Clone + Debug + Unpin + Send + Sync + 'static {
    type Err: Error;

    fn convert_sim_tx(&self, tx_req: TxReq) -> Result<SimTx, Self::Err>;
}

impl<TxReq, SimTx> SimTxConverter<TxReq, SimTx> for ()
where
    TxReq: TryIntoSimTx<SimTx> + Debug,
{
    type Err = ValueError<TxReq>;

    fn convert_sim_tx(&self, tx_req: TxReq) -> Result<SimTx, Self::Err> {
        tx_req.try_into_sim_tx()
    }
}

impl<TxReq, SimTx, F, E> SimTxConverter<TxReq, SimTx> for F
where
    TxReq: Debug,
    E: Error,
    F: Fn(TxReq) -> Result<SimTx, E> + Clone + Debug + Unpin + Send + Sync + 'static,
{
    type Err = E;

    fn convert_sim_tx(&self, tx_req: TxReq) -> Result<SimTx, Self::Err> {
        self(tx_req)
    }
}
```
since `TxReq` varies per network, i can't see any good way of solving this besides matching on `String` type which is not desirable, or:
```rust
    /// Required fields are missing from the transaction request.
    #[error("required fields missing from transaction request: {0}")]
    MissingRequiredFields(Box<dyn Error>),
```
but isn't that effectively doing the same thing as matching on `String`?

crates/rpc/rpc-convert/src/transaction.rs
```diff
@@ -484,10 +483,32 @@ where
...
+    #[error("deposit transactions cannot be pooled")]
+    DepositTransaction,
+    /// Blob transactions without sidecars cannot be pooled. After being included
+    /// in a block, the sidecar data is pruned, making the transaction unpoolable.
+    #[error("blob transaction missing sidecar")]
+    BlobSidecarMissing,
+    /// Transaction type is not supported for pool submission.
+    #[error("transaction type not supported for pool")]
+    UnsupportedForPool,
+}
```
mattsse: I don't think we need to limit this to a specific set

we can make one variant for

```
#[error("Failed to convert transaction into RPC response: {0}")]
```

and then have an `Other(String)` variant that only holds the error message

mattsse (changes requested):
cool, only one suggestion re the TransactionConversionError variants, fine if these just use (String) messages

crates/rpc/rpc-convert/src/transaction.rs
figtracer: i think we should just collapse everything into the initial approach since we can use String here; just commited the changes

mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):

PR #20060, opened 2025-12-01
refactor: remove dead state and make entries() always 0

Forostovec:
Simplifieв the mock DB to match its documented no-op behavior. Removes unused internal state from TxMock and makes entries() explicitly return 0. Deprecates DatabaseMock.data as an unused placeholder to avoid breaking external consumers while clarifying that the mock does not persist data.


mattsse (commented):
we can keep this as is

PR #20061, opened 2025-12-01
feat(optimism): Flashblock Receipts Stream

mablr:
Close #19681

Streams receipts from flashblocks as they arrive, before canonical block finalization.

  - Subscribes to received flashblocks and extracts receipts
  - Returns a Stream of `Log` items matched with a Filter via `matching_block_logs_with_tx_hashes`, similarly to [`‎EthPubSubInner::log_stream‎`](https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/rpc/rpc/src/eth/pubsub.rs#L365)
  - Tracks block metadata (number, timestamp) from base flashblock (index 0) for use across the flashblocks sequence
  - This uses fancy `scan` iterator method to maintain state without requiring a separate handler struct


crates/optimism/rpc/src/eth/mod.rs
```diff
@@ -125,6 +132,48 @@ impl<N: RpcNodeCore, Rpc: RpcConvert> OpEthApi<N, Rpc> {
...
+                        let receipts = fb
+                            .metadata
+                            .receipts
+                            .iter()
+                            .map(|(tx, receipt)| (*tx, receipt.clone()))
+                            .collect::<Vec<_>>();
+
+                        let all_logs = matching_block_logs_with_tx_hashes(
+                            &filter,
+                            BlockNumHash::new(state.0, fb.diff.block_hash),
```
mablr: Not sure if it's relevant to build a `BlockNumHash` from the flashblock hash (`diff.block_hash`).

Another matching method could be designed here, but Idk what are the actual filtering needs...

mattsse (approved):
this should do it, I think ideally we could also annotate the sequence info in the event type if we have it

PR #20064, opened 2025-12-02
test(era): complete int tests with roundtrip mainnet era files 

lean-apple:
- Add mainnet era files integration tests
 - Remove decompressing/decoding era1/era tests - they are redundant with roundtrip tests

How to check :
```
cargo nextest run --release --package reth-era --test it -- --ignored
```


crates/era/tests/it/era/dd.rs
lean-apple, line 1: redundant with era roundtrip test

crates/era/tests/it/era1/mod.rs
lean-apple, line 1: redundant with era1 roundtrip test

mattsse (approved):

PR #20066, opened 2025-12-02
perf(engine): batch multiproof messages 

### LINKED ISSUE #19168: perf: Drain and aggregate PrefetchProofs / StateUpdate messages

yongkangc:
fixes https://github.com/paradigmxyz/reth/issues/19168

### Problem

When state updates arrive faster than proof computation, each message triggers a separate proof task, causing redundant trie traversals for the same accounts.

### Solution

Batch consecutive same-type messages before dispatching to workers. Batching Logic:

1. **Receive first message** — Start accumulating
2. **Drain channel non-blocking** — Collect more same-type messages up to target/message limits
3. **Dispatch as single batch** — Merge targets and send to workers

Different-type messages encountered during drain are stored in `pending_msg` and processed in the next iteration to preserve ordering.

| Type | Target Limit | Message Limit |
|------|--------------|---------------|
| PrefetchProofs | 512 | 16 |
| StateUpdate | 64 | — |

### Impact

Benches: https://github.com/paradigmxyz/reth/pull/20066#issuecomment-3625777154

Pending Queue also saw a drop
image

---

[Detailed flow diagrams](https://gist.github.com/yongkangc/fb7cae0744cef46fbb1e833107676eb1)


yongkangc:
## Flow Diagrams

### Before: No Batching

```
Channel: [StateUpdate1, StateUpdate2, StateUpdate3, PrefetchProofs1]

Loop:
  recv StateUpdate1   -->  dispatch proof #1
  recv StateUpdate2   -->  dispatch proof #2
  recv StateUpdate3   -->  dispatch proof #3
  recv PrefetchProofs1  -->  dispatch proof #4

Result: 4 messages, 4 proof tasks, 4 trie traversals
```

### After: With Batching

```
Channel: [StateUpdate1, StateUpdate2, StateUpdate3, PrefetchProofs1]
State:   pending_msg = None

Loop iteration 1:
  pending_msg = None, so select recv StateUpdate1
  |
  +-- process_multiproof_message(StateUpdate1)
        |
        +-- try_recv() StateUpdate2, same type, MERGE
        +-- try_recv() StateUpdate3, same type, MERGE  
        +-- try_recv() PrefetchProofs1, different type
        |         |
        |         +-- pending_msg = Some(PrefetchProofs1)
        |
        +-- dispatch ONE merged batch as proof #1

Loop iteration 2:
  pending_msg = Some(PrefetchProofs1), skip select
  |
  +-- process_multiproof_message(PrefetchProofs1)
        |
        +-- dispatch proof #2

Result: 4 messages, 2 proof tasks, 2 trie traversals
```

### Message Type Handling

| Message Type | Batching |
|--------------|----------|
| `StateUpdate` | Merge: combine `state` fields |
| `PrefetchProofs` | Merge: extend `targets` |
| `FinishedStateUpdates` | No batch (signal) |
| `EmptyProof` | No batch (forward as-is) |

### Why pending_msg vs Channel Send-Back?

```
WRONG: self.tx.send(other_msg)
   [A, B, C, D]  -->  drain A,B  -->  send C back  -->  [D, C]
   C now AFTER D, order broken

CORRECT: pending_msg = Some(other_msg)
   [A, B, C, D]  -->  drain A,B  -->  pending=C  -->  next: process C, then D
   Order preserved
```

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1133,15 +1212,96 @@ impl MultiProofTask {
...
-                                state_update_proofs_requested += self.on_state_update(source, update);
+                                let mut merged_update = update;
+                                let mut num_batched = 1;
+                                let mut pending_msg: Option<MultiProofMessage> = None;
+
+                                loop {
+                                    match self.rx.try_recv() {
+                                        Ok(MultiProofMessage::StateUpdate(_new_source, next_update)) => {
+                                            merged_update.extend(next_update);
+                                            num_batched += 1;
```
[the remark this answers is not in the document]
yongkangc: Addressed: Added a comment explaining that `source` is only used for logging/observability, so using the first sources attribution is acceptable. The comment on line 1059-1061 now explains this:

```rust
// Batch consecutive StateUpdate messages into one.
// When batching, we use the first messages source for logging (observability only).
```

StateChangeSource does not affect correctness - it is purely for tracing/debugging purposes.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1109,17 +1109,96 @@ impl MultiProofTask {
...
+                                                );
+                                                break
+                                            }
+                                        }
+                                        MultiProofMessage::PrefetchProofs(_) => {
+                                            // This shouldn't happen as we're batching PrefetchProofs
+                                            unreachable!("PrefetchProofs should have been batched")
+                                        }
+                                    }
+                                }
```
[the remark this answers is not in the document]
yongkangc: Resolved: The code has been refactored to use a single `process_multiproof_message()` helper function (lines 982-1137) that handles all message types. The duplication is now eliminated.

The main loop simply calls:
```rust
if self.process_multiproof_message(msg, &mut pending_msg, ...) {
    break;
}
```

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1739,4 +1899,126 @@ mod tests {
...
+                assert!(merged_targets.contains_key(&addr2));
+                assert!(merged_targets.contains_key(&addr3));
+
+                task.on_prefetch_proof(merged_targets)
+            } else {
+                panic!("Expected PrefetchProofs message");
+            };
+
+        assert_eq!(proofs_requested, 1);
+    }
```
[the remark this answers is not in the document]
yongkangc: Addressed: Added `test_batching_preserves_ordering_with_different_message_type()` test that verifies exactly this scenario:

```
Queue: [PrefetchProofs1, PrefetchProofs2, StateUpdate1, StateUpdate2, PrefetchProofs3]
```

The test verifies:
1. First two PrefetchProofs are batched (not 3)
2. StateUpdate1 is stored in pending_msg (not lost)
3. StateUpdate2 is next in channel
4. PrefetchProofs3 is preserved at end

See lines 1971-2096.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1739,4 +1899,126 @@ mod tests {
...
+                assert_eq!(merged_update.len(), 2);
+                assert!(merged_update.contains_key(&addr1));
+                assert!(merged_update.contains_key(&addr2));
+
+                task.on_state_update(source, merged_update)
+            } else {
+                panic!("Expected StateUpdate message");
+            };
+        assert_eq!(proofs_requested, 1);
+    }
```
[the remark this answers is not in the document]
yongkangc: Addressed: The `test_batching_preserves_ordering_with_different_message_type()` test now includes this scenario. 

Regarding different sources: As explained in the reply to comment #1, `StateChangeSource` is only used for logging/observability - it does not affect correctness. The state data itself is what matters and gets correctly merged via `extend()`.

Additionally, added `test_pending_messages_get_full_batching_treatment()` which verifies that pending messages also get batching treatment in subsequent iterations.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1095,103 +1522,26 @@ impl MultiProofTask {
                     }
                 },
                 recv(self.rx) -> message => {
-                    match message {
```
yongkangc, line 1098: this is extracted to `process_multiproof_message`

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -745,47 +844,36 @@ impl MultiProofTask {
...
-            dispatch(proof_targets);
-        }
-
-        self.metrics.prefetch_proof_chunks_histogram.record(chunks as f64);
+        let chunking_len = proof_targets.chunking_length();
+        let available_account_workers =
+            self.multiproof_manager.proof_worker_handle.available_account_workers();
+        let available_storage_workers =
+            self.multiproof_manager.proof_worker_handle.available_storage_workers();
+        let outcome = dispatch_with_chunking(
```
mediocregopher: nit: it looks like `chunk` got renamed to `outcome`, if we're still returning number of chunks here then `chunks` or `num_chunks` would be a clearer name I think

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -895,69 +984,52 @@ impl MultiProofTask {
...
-            tracing::Span::current().record("chunks", chunks);
-        } else {
-            dispatch(not_fetched_state_update);
-        }
-
+        let available_account_workers =
+            self.multiproof_manager.proof_worker_handle.available_account_workers();
+        let available_storage_workers =
+            self.multiproof_manager.proof_worker_handle.available_storage_workers();
+        let outcome = dispatch_with_chunking(
```
mediocregopher: same nit here as with the prewarm case

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1051,272 @@ impl MultiProofTask {
...
+
+                if first_update_time.is_none() {
+                    self.metrics
+                        .first_update_wait_time_histogram
+                        .record(start.elapsed().as_secs_f64());
+                    *first_update_time = Some(Instant::now());
+                    debug!(target: "engine::tree::payload_processor::multiproof", "Started state root calculation");
+                }
+
+                // Dispatch first message immediately
```
mediocregopher: Did you do some testing that confirmed that doing this is impactful? My intuition says that it wouldn't matter a whole lot, since it's just the results of one tx, more likely than not it would only occupy a single worker while the rest are waiting for further batched messages.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1051,272 @@ impl MultiProofTask {
...
+    ///
+    /// If a different message type is encountered, it's stored in `pending_msg` for the next
+    /// iteration (preserving order without sending back to channel).
+    ///
+    /// Returns `true` if done, `false` to continue.
+    #[allow(clippy::too_many_arguments)]
+    fn process_multiproof_message(
+        &mut self,
+        msg: MultiProofMessage,
+        pending_msg: &mut Option<MultiProofMessage>,
```
mediocregopher: It'd be nice to pull out a lot (or all) of these `&mut`s into a separate struct, like `MultiproofBatchCtx` or something. It seems like a lot are just for metrics, if so maybe two structs, `MultiproofBatchCtx` and `MultiproofBatchMetrics`.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -534,6 +516,11 @@ pub(crate) struct MultiProofTaskMetrics {
     /// Histogram of the number of state update proof target chunks.
     pub state_update_proof_chunks_histogram: Histogram,
 
+    /// Histogram of prefetch proof batch sizes (number of messages merged).
+    pub prefetch_batch_size_histogram: Histogram,
```
yongkangc, line 555: we might want to look at targets merged as well, but maybe for a start messages works since targets is more or less fixed?

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -684,8 +706,50 @@ pub(super) struct MultiProofTask {
...
+    /// Processed first in next iteration to preserve message ordering.
+    pending_msg: Option<MultiProofMessage>,
+    /// Timestamp when the first state update or prefetch was received.
+    first_update_time: Option<Instant>,
+    /// Timestamp before the first state update or prefetch was received.
+    start: Instant,
+    /// Whether all state updates have been received.
+    updates_finished: bool,
+    /// Timestamp when state updates finished.
+    updates_finished_time: Option<Instant>,
```
mediocregopher: nit: these two fields are probably redundant, `updates_finished` could be inferred by `updates_finished_time.is_some()`

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1264,6 +1509,92 @@ fn get_proof_targets(
...
+            batch_update == next_update
+        }
+        _ => true,
+    }
+}
+
+/// Checks whether two state change sources refer to the same origin.
+fn same_state_change_source(lhs: StateChangeSource, rhs: StateChangeSource) -> bool {
+    match (lhs, rhs) {
+        (StateChangeSource::Transaction(a), StateChangeSource::Transaction(b)) => a == b,
```
mediocregopher, line 1516: My initial thinking here was that we could be more aggressively merging state updates together, since they appear sequentially from the main execution thread. But what I realized is that the proof sequencer is essentially already doing that:

https://github.com/paradigmxyz/reth/blob/28d6996fc4cf71aa030cef27b8ef1e248bc8372e/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L964-L980

Thinking about this more, I feel like the proof sequencer isn't necessary at all. The only thing that matters with respect to ordering is that the HashedPostState from state updates arrives at the sparse trie in tx execution order. Prewarm multiproof targets don't include a HashedPostState, and state update multiproofs from the main execution thread are naturally produced in order anyway.

If the proof workers sent the proof results directly to the sparse trie task (rather than returning them to MultiProofTask via the `proof_result_rx`) I think it would "just work". We would instead want to do batching of state updates in the MultiProofTask, the same way prewarm targets are batched.

cc @yongkangc @mattsse would be worth following up on, one less channel hop in the pipeline would be worthwhile.

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                        .record(ctx.start.elapsed().as_secs_f64());
+                    ctx.first_update_time = Some(Instant::now());
+                    debug!(target: "engine::tree::payload_processor::multiproof", "Started state root calculation");
+                }
+
+                // Accumulate messages including the first one
+                let mut accumulated_count = targets.chunking_length();
+                // Preallocate up to the message cap; avoids repeated reallocations when bursts
+                // arrive.
+                let mut accumulated_targets: Vec<MultiProofTargets> =
```
mediocregopher: nit: maybe a premature optimization, but you could move this `accumulated_targets`, as well as the one for the state updates, onto the MultiproofBatchCtx and re-use them across calls. If there's lots of messages buffered on the channel then this function could get called in a tight loop, so it might be impactful (though probably not much)

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                // Only attempt batching if below limits; otherwise skip straight to dispatch.
+                if accumulated_count < PREFETCH_MAX_BATCH_TARGETS &&
+                    accumulated_targets.len() < PREFETCH_MAX_BATCH_MESSAGES
+                {
+                    loop {
+                        if accumulated_count >= PREFETCH_MAX_BATCH_TARGETS ||
+                            accumulated_targets.len() >= PREFETCH_MAX_BATCH_MESSAGES
+                        {
+                            break;
+                        }
```
mediocregopher: nit: this could be a `while` I think

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                // Preallocate modestly; state updates are heavier per message, but we can see small
+                // bursts.
+                let mut accumulated_updates: Vec<(StateChangeSource, EvmState)> =
+                    Vec::with_capacity(STATE_UPDATE_BATCH_PREALLOC);
+                accumulated_updates.push((source, update));
+
+                loop {
+                    if accumulated_targets >= STATE_UPDATE_MAX_BATCH_TARGETS {
+                        break;
+                    }
```
mediocregopher: nit: another `while`

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                                ctx.pending_msg =
+                                    Some(MultiProofMessage::StateUpdate(next_source, next_update));
+                                break;
+                            }
+                            if accumulated_targets + next_estimate > STATE_UPDATE_MAX_BATCH_TARGETS
+                            {
+                                ctx.pending_msg =
+                                    Some(MultiProofMessage::StateUpdate(next_source, next_update));
+                                break;
+                            }
```
mediocregopher: These two checks are redundant, if the second is true then the first must have been true.
```suggestion
                            if accumulated_targets + next_estimate > STATE_UPDATE_MAX_BATCH_TARGETS
                            {
                                ctx.pending_msg =
                                    Some(MultiProofMessage::StateUpdate(next_source, next_update));
                                break;
                            }
```

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                    }
+                }
+
+                // Process all accumulated messages in a single batch
+                let num_batched = accumulated_updates.len();
+                self.metrics.state_update_batch_size_histogram.record(num_batched as f64);
+
+                let batch_source = accumulated_updates[0].0;
+                {
+                    let batch_update = &accumulated_updates[0].1;
```
mediocregopher:
```suggestion
                #[cfg(debug_assertions)]
                {
                    let batch_source = accumulated_updates[0].0;
                    let batch_update = &accumulated_updates[0].1;
```

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -979,6 +1016,246 @@ impl MultiProofTask {
...
+                        can_batch_state_update(batch_source, batch_update, *source, update)
+                    }));
+                }
+
+                // Merge all accumulated updates into a single EvmState payload.
+                let mut accumulated_iter = accumulated_updates.into_iter();
+                let (batch_source, mut merged_update) = accumulated_iter
+                    .next()
+                    .expect("state update batch always has at least one entry");
+                for (_, next_update) in accumulated_iter {
```
mediocregopher: `batch_source` only gets used for logging after this point, but imo it'd be clearer to keep the most recent `batch_source`

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1010,37 +1287,90 @@ impl MultiProofTask {
...
+        // messages; labeled so inner loops can `break 'main` once all work is complete.
+        'main: loop {
+            trace!(target: "engine::tree::payload_processor::multiproof", "entering main channel receiving loop");
 
-        // Timestamp before the first state update or prefetch was received
-        let start = Instant::now();
+            // Always drain proof results first to maintain priority over pending messages.
+            // This prevents priority inversion where pending_msg would bypass select_biased!
+            // and starve proof result processing.
+            while let Ok(proof_result) = self.proof_result_rx.try_recv() {
```
mediocregopher: I understand why `if let Some(msg) = ctx.pending_msg.take() {` was added prior to the crossbeam_select, but why do we do this loop above that? Processing the pending msg is non-blocking, so I would expect that doing that and then continuing on to the crossbeam_select would be sufficient.

mediocregopher (changes requested):

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: resolved!

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: good note, thanks - this makes sense

yongkangc: resolved this in e7164c1c94f9f037e5d0db0bd7315f5ee322f32f

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: resolve [9866615](https://github.com/paradigmxyz/reth/pull/20066/commits/986661596977e92cd3aebe012470956070731b41)

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: i think thats possible but we would have to take into consideration the parallel processing by workers which could finish out of order
i think if we want to do that, we would need a proof sequence index - but that might be the same logic as [add_proof](https://github.com/paradigmxyz/reth/blob/986661596977e92cd3aebe012470956070731b41/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L124)?

there might be some complexity in the ordering impl, but would simplify things - i think we could try

yongkangc: "Ok so for state update multiproofs we still need the sequencer, but for prefetch we don't, we can just forward them directly to the the sparse trie task without sequencing"

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: good point, thanks - missed that

yongkangc: adressed [98f3406](https://github.com/paradigmxyz/reth/pull/20066/commits/98f3406949877db64003ef8e07c9f51b99c851e9)

mediocregopher (approved):

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -684,8 +706,59 @@ pub(super) struct MultiProofTask {
     max_targets_for_chunking: usize,
 }
 
+/// Context for multiproof message batching loop.
+///
+/// Contains processing state that persists across loop iterations.
+struct MultiproofBatchCtx {
```
mattsse: I believe these new types are defined out of order and should be moved below the `impl MultiProofTask {` block

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1010,37 +1277,40 @@ impl MultiProofTask {
...
-        let mut updates_finished_time = None;
-
-        loop {
+        // Main event loop; select_biased! prioritizes proof results over control messages.
+        // Labeled so inner match arms can `break 'main` once all work is complete.
+        'main: loop {
             trace!(target: "engine::tree::payload_processor::multiproof", "entering main channel receiving loop");
 
+            if let Some(msg) = ctx.pending_msg.take() {
+                if self.process_multiproof_message(msg, &mut ctx, &mut batch_metrics) {
```
mattsse, line 1251: oh I see, this way we can batch any additional new incoming `PrefetchProofs` messages, if not we buffer the other message variant, makes sense

mattsse (approved):
pedantic style nit,

the only reason why the diff is so large is because we extracted the message handler for MultiProofMessage message

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: yup 100%

yongkangc: also updated docstring to be clearer about this point

crates/engine/tree/src/tree/payload_processor/multiproof.rs
yongkangc: done!

PR #20067, opened 2025-12-02
docs(engine): fix canonical_block_by_hash comments

JacksonEi:
Correct the `canonical_block_by_hash` documentation to state that it fetche finalized blocks from the database instead of returning `None`. The previou documentation contradicted the implementation, which queries the provider a only fails if the historical data has been pruned.


mattsse (approved):
I think this is more accurate

PR #20071, opened 2025-12-02
feat(persistence): implement RocksDB provider

### LINKED ISSUE #20005: Add reth-rocksdb crate and rocksdb provider impl
We should create a new crate called `reth-rocksdb` and add a provider `RocksDBProvider` which can manage a rocksdb instance and manage basic DB operations

duyquang6:
Resolve #20005 
Changes:
- Apply some recommend tuning config from [RocksDB docs](https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning), verified enabled by RocksDB log
- Expose api: Get, Put, Delete, WriteBatch
- Compatible with db-api non-dupsort [Table](https://github.com/duyquang6/reth/blob/71c124798c8a15e6aeec1b4a4e6e9fc24073d6ac/crates/storage/db-api/src/table.rs#L87)
- Add RocksDB metrics: count, latency
- Add tests & verify

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -98,7 +98,7 @@ jobs:
       matrix:
         partition: [1, 2]
         total_partitions: [2]
-    timeout-minutes: 30
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v5
       - uses: rui314/setup-mold@v1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -646,6 +646,9 @@ c-kzg = "2.1.5"
 # config
 toml = "0.8"
 
+# database
+rocksdb = { version = "0.24", default-features = false, features = ["bindgen-runtime", "lz4", "zstd", "jemalloc"] }
+
 # otlp obs
 opentelemetry_sdk = "0.31"
 opentelemetry = "0.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -646,8 +646,8 @@ c-kzg = "2.1.5"
 # config
 toml = "0.8"
 
-# database
-rocksdb = { version = "0.24", default-features = false, features = ["bindgen-runtime", "lz4", "zstd", "jemalloc"] }
+# rocksdb
+rocksdb = { version = "0.24" }
 
 # otlp obs
 opentelemetry_sdk = "0.31"
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -1,0 +1,513 @@
+use reth_db_api::{
+    table::{Compress, Decompress, Encode, Table},
+    DatabaseError,
+};
+use reth_storage_errors::{
+    db::{DatabaseErrorInfo, DatabaseWriteError, DatabaseWriteOperation, LogLevel},
+    provider::{ProviderError, ProviderResult},
+};
+use rocksdb::{Cache, ColumnFamilyDescriptor, Options, WriteBatch, DB};
+use std::{fmt, path::Path, sync::Arc};
+
+use super::metrics::{RocksDBMetrics, RocksDBOperation};
+
+/// Converts Reth's [`LogLevel`] to RocksDB's [`rocksdb::LogLevel`].
+fn convert_log_level(level: LogLevel) -> rocksdb::LogLevel {
+    match level {
+        LogLevel::Fatal => rocksdb::LogLevel::Fatal,
+        LogLevel::Error => rocksdb::LogLevel::Error,
+        LogLevel::Warn => rocksdb::LogLevel::Warn,
+        LogLevel::Notice => rocksdb::LogLevel::Info,
+        LogLevel::Verbose => rocksdb::LogLevel::Info,
+        LogLevel::Debug => rocksdb::LogLevel::Debug,
+        LogLevel::Trace => rocksdb::LogLevel::Debug,
+        LogLevel::Extra => rocksdb::LogLevel::Debug,
+    }
+}
+
+/// Builder for [`RocksDBProvider`].
+pub struct RocksDBBuilder {
+    path: std::path::PathBuf,
+    column_families: Vec<String>,
+    enable_metrics: bool,
+    enable_statistics: bool,
+    log_level: rocksdb::LogLevel,
+    block_cache: Cache,
+}
+
+impl fmt::Debug for RocksDBBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RocksDBBuilder")
+            .field("path", &self.path)
+            .field("column_families", &self.column_families)
+            .field("enable_metrics", &self.enable_metrics)
+            .finish()
+    }
+}
+
+impl RocksDBBuilder {
+    /// Creates a new builder with optimized default options.
+    pub fn new(path: impl AsRef<Path>) -> Self {
+        let cache = Cache::new_lru_cache(128 << 20);
+        Self {
+            path: path.as_ref().to_path_buf(),
+            column_families: Vec::new(),
+            enable_metrics: false,
+            enable_statistics: false,
+            log_level: rocksdb::LogLevel::Info,
+            block_cache: cache,
+        }
+    }
+
+    /// Creates optimized RocksDB options per RocksDB wiki recommendations.
+    fn default_options(
+        log_level: rocksdb::LogLevel,
+        cache: &Cache,
+        enable_statistics: bool,
+    ) -> Options {
+        // Follow recommend tuning guide from RocksDB wiki, see https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning
+        let mut table_options = rocksdb::BlockBasedOptions::default();
+        table_options.set_block_size(16 * 1024);
+        table_options.set_cache_index_and_filter_blocks(true);
+        table_options.set_pin_l0_filter_and_index_blocks_in_cache(true);
+        // Shared block cache for all column families.
+        table_options.set_block_cache(cache);
+        // Bloom filter: 10 bits/key = ~1% false positive rate, full filter for better read
+        // performance. this setting is good trade off a little bit of memory for better
+        // point lookup performance. see https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#configuration-basics
+        table_options.set_bloom_filter(10.0, false);
+        table_options.set_optimize_filters_for_memory(true);
+
+        let mut options = Options::default();
+        options.set_block_based_table_factory(&table_options);
+        options.create_if_missing(true);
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -652,6 +646,9 @@ c-kzg = "2.1.5"
 # config
 toml = "0.8"
 
+# database
+rocksdb = { version = "0.24", default-features = false, features = ["bindgen-runtime", "lz4", "zstd", "jemalloc"] }
```
mattsse: I think enabling these by default is okay

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+            let value = result.and_then(|value| T::Value::decompress(&value).ok());
+
+            Ok(value)
+        })
+    }
+
+    /// Puts a value into the specified table.
+    pub fn put<T: Table>(&self, key: T::Key, value: T::Value) -> ProviderResult<()> {
+        self.execute_metered(RocksDBOperation::Put, T::NAME, || {
+            let encoded_key = key.encode();
```
mattsse: we likely also want put_encoded I think

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+                )?;
+
+            let value = result.and_then(|value| T::Value::decompress(&value).ok());
+
+            Ok(value)
+        })
+    }
+
+    /// Puts a value into the specified table.
+    pub fn put<T: Table>(&self, key: T::Key, value: T::Value) -> ProviderResult<()> {
```
mattsse: looks like these can be accepted by ref?

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+
+        Ok(RocksDBProvider(Arc::new(RocksDBProviderInner { db, metrics })))
+    }
+}
+
+/// Inner state for `RocksDB` provider.
+#[derive(Debug)]
+struct RocksDBProviderInner {
+    db: DB,
+    metrics: Option<Arc<RocksDBMetrics>>,
```
mattsse: I assume we don't actually need the arc wrapper here?

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+/// Converts Reth's [`LogLevel`] to `RocksDB`'s [`rocksdb::LogLevel`].
+const fn convert_log_level(level: LogLevel) -> rocksdb::LogLevel {
+    match level {
+        LogLevel::Fatal => rocksdb::LogLevel::Fatal,
+        LogLevel::Error => rocksdb::LogLevel::Error,
+        LogLevel::Warn => rocksdb::LogLevel::Warn,
+        LogLevel::Notice | LogLevel::Verbose => rocksdb::LogLevel::Info,
+        LogLevel::Debug | LogLevel::Trace | LogLevel::Extra => rocksdb::LogLevel::Debug,
+    }
+}
```
mattsse: can we move this to end of file

mattsse (commented):
a few (early and pedantic) comments

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+            .field("column_families", &self.column_families)
+            .field("enable_metrics", &self.enable_metrics)
+            .finish()
+    }
+}
+
+impl RocksDBBuilder {
+    /// Creates a new builder with optimized default options.
+    pub fn new(path: impl AsRef<Path>) -> Self {
+        let cache = Cache::new_lru_cache(128 << 20);
```
Rjected: I would create a default cache size const and use it here

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+        options.set_block_based_table_factory(&table_options);
+        options.create_if_missing(true);
+        options.create_missing_column_families(true);
+        options.set_max_background_jobs(6);
+        options.set_bytes_per_sync(1_048_576);
+
+        options.set_bottommost_compression_type(rocksdb::DBCompressionType::Zstd);
+        options.set_bottommost_zstd_max_train_bytes(0, true);
+        options.set_compression_type(rocksdb::DBCompressionType::Lz4);
+        options.set_compaction_pri(rocksdb::CompactionPri::MinOverlappingRatio);
```
Rjected: nit: I'd import `DBCompressionType`, `CompactionPri`, `BlockBasedOptions` at the top of the file

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+        table_options.set_block_size(16 * 1024);
+        table_options.set_cache_index_and_filter_blocks(true);
+        table_options.set_pin_l0_filter_and_index_blocks_in_cache(true);
+        table_options.set_block_cache(cache);
+        // Bloom filter: 10 bits/key = ~1% false positive rate, full filter for better read
+        // performance. this setting is good trade off a little bit of memory for better
+        // point lookup performance. see https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#configuration-basics
+        table_options.set_bloom_filter(10.0, false);
+        // Optimize filters for memory.
+        table_options.set_optimize_filters_for_memory(true);
```
Rjected: maybe we could have a `fn default_table_options(cache: &Cache) -> BlockBasedOptions`? So that we don't have to repeat this in both `default_column_family_options` and `default_options`.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,509 @@
...
+}
+
+impl<'a> fmt::Debug for RocksDBBatch<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RocksDBBatch")
+            .field("provider", &self.provider)
+            .field("batch", &"<WriteBatch>")
+            .field("length", &self.inner.len())
+            .field("size_in_bytes", &self.inner.size_in_bytes())
+            .finish()
```
Rjected, line 366: can we add some comments for what `length` and `size_in_bytes` represent w.r.t. the `WriteBatch`

Rjected (commented):
makes sense to me but left some nits

Rjected (commented):
I think we do need to add windows conditional compilation in this PR in order for CI to pass

.github/workflows/lint.yml
```diff
@@ -98,9 +98,9 @@ jobs:
       matrix:
         partition: [1, 2]
         total_partitions: [2]
-    timeout-minutes: 30
+    timeout-minutes: 60
```
duyquang6, line 101: issue: https://github.com/paradigmxyz/reth/actions/runs/19922208464/job/57115796165?pr=20071
I think because we add more dependency so this check is take longer

duyquang6:
The conditional compilation has been updated, and the changes are ready for review.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,558 @@
...
+
+/// Default bytes per sync for `RocksDB` WAL writes (1 MB).
+const DEFAULT_BYTES_PER_SYNC: u64 = 1_048_576;
+
+/// Default bloom filter bits per key (~1% false positive rate).
+const DEFAULT_BLOOM_FILTER_BITS: f64 = 10.0;
+
+/// Builder for [`RocksDBProvider`].
+pub struct RocksDBBuilder {
+    path: std::path::PathBuf,
```
Rjected:
```suggestion
use std::{fmt, path::{Path, PathBuf}, sync::Arc};

use super::metrics::{RocksDBMetrics, RocksDBOperation};

/// Default cache size for `RocksDB` block cache (128 MB).
const DEFAULT_CACHE_SIZE: usize = 128 << 20;

/// Default block size for `RocksDB` tables (16 KB).
const DEFAULT_BLOCK_SIZE: usize = 16 * 1024;

/// Default max background jobs for `RocksDB` compaction and flushing.
const DEFAULT_MAX_BACKGROUND_JOBS: i32 = 6;

/// Default bytes per sync for `RocksDB` WAL writes (1 MB).
const DEFAULT_BYTES_PER_SYNC: u64 = 1_048_576;

/// Default bloom filter bits per key (~1% false positive rate).
const DEFAULT_BLOOM_FILTER_BITS: f64 = 10.0;

/// Builder for [`RocksDBProvider`].
pub struct RocksDBBuilder {
    path: PathBuf,
```

Rjected (commented):
just have one nit otherwise this looks good to me, think we can start implementing providers pretty easily after this

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,561 @@
...
+
+        let metrics = self.enable_metrics.then(RocksDBMetrics::default);
+
+        Ok(RocksDBProvider(Arc::new(RocksDBProviderInner { db, metrics })))
+    }
+}
+
+/// Some types don't support compression (eg. B256), and we don't want to be copying them to the
+/// allocated buffer when we can just use their reference.
+macro_rules! compress_to_buf_or_ref {
```
mattsse, line 195: this makes sense

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,561 @@
...
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+
+impl RocksDBProvider {
+    /// Creates a new `RocksDB` provider.
+    pub fn new(path: impl AsRef<Path>) -> ProviderResult<Self> {
+        RocksDBBuilder::new(path).build()
+    }
```
mattsse, line 230: would be nice to also have a `fn builder` convenience fn here, then you don't need to import two types

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// Executes a function and records metrics with the given operation and table name.
+    fn execute_with_operation_metric<T>(
+        &self,
+        operation: RocksDBOperation,
+        table: &'static str,
+        f: impl FnOnce(&Self) -> T,
+    ) -> T {
+        let start = std::time::Instant::now();
```
mattsse: this start is redundant if no metrics, so we can make this an optional as well and then simply do a match over both options for recording

mattsse (approved):
this lgtm, with one pedantic suggestion

pending @Rjected

Rjected (approved):
LGTM

PR #20075, opened 2025-12-02
feat(trie): Proof Rewrite: Use cached branch nodes

### LINKED ISSUE #19513: Proof Calc Rewrite: Cached Branch Nodes

mediocregopher:
Closes #19513

This PR introduces the usage of the AccountsTrie and StoragesTrie tables into the proof_v2 implementation. These tables contain cached information about intermediate nodes of the tries, notably precomputed hashes of some children, allowing us to skip over large sub-tries when doing any kind of trie calculation.

Incorporating this data into the proof_v2 calculator is a bit mind-bending, but it's done without affecting too much of the existing logic for calculating trie nodes based solely on leaf data. The new cache-aware code uses the cache data to drive the logic forward, deciding which ranges of leaves need to be iterated over to calculate portions of the trie for which no cached information is available.

I've left comments in both the PR and code to explain the logic as best as possible.

# Benches

As before the new implementation is faster across all dimensions, with speedups of up to 30% for larger target set sizes.

shot-1764781929
shot-1764781901


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -156,34 +224,47 @@ where
...
 
-        // If the path isn't in the current range then iterate forward until it is (or until there
-        // is no upper bound, indicating unbounded).
-        while upper.is_some_and(|upper| depth_first::cmp(path, &upper) != Ordering::Less) {
-            targets.next();
-            trace!(target: TRACE_TARGET, target = ?targets.peek(), "upper target <= path, next target");
-            let &(l, u) = targets.peek().expect("targets is never exhausted");
-            (lower, upper) = (l, u);
-        }
+        loop {
```
mediocregopher, line 203: This loop got reworked a bit to account for usage of the trie cursor.

Previously we expected to always visit calculated trie nodes in depth-first order, however we actually visit cached branch nodes from the trie-cursor in lexicographic order. Specifically we visit the children of cached branch nodes which we have a hash for, because if they should be retained then we need to descend into them anyway, ignoring the hash.

To account for this we always check `lower.starts_with(path)` on every loop, not just once after the loop. This ends up accounting for the two different ways of comparing against the target sequence, but it's a bit mind-bending.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -281,6 +372,9 @@ where
         }
 
         let child_path = self.last_child_path();
+        // TODO theoretically `commit_child` only needs to convert to an `RlpNode` if it's going to
```
mediocregopher, line 327: Something to think about for a future PR

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -320,37 +413,25 @@ where
...
-        leaf_key: Nibbles,
-        leaf_val: VE::DeferredEncoder,
-    ) -> Result<(), StateProofError> {
-        // First determine the new leaf's shortkey relative to the current branch. If there is no
-        // current branch then the short key is the full key.
-        let leaf_short_key = if self.branch_stack.is_empty() {
-            leaf_key
+    /// Returns the nibble of the branch's `state_mask` which should be set for the new child, and
+    /// short key that the next child should use.
+    fn push_new_branch(&mut self, new_child_path: Nibbles) -> (u8, Nibbles) {
```
mediocregopher, line 374: Previously this method was only used when pushing leafs onto the `child_stack`, now it is also used when pushing cached branches onto the `child_stack`. The core logic is the same, the only difference is that variables and documentation is made generic and `push_new_leaf` isn't called from here anymore; instead this method returns the nibble and short-key of the node which caused this new branch to be pushed, and the caller is expected to construct that node and push it onto the `child_stack` or `branch_stack` themselves.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +653,456 @@ where
...
+    /// bound is None then it is considered unbounded.
+    ///
+    /// It is expected that this method is "driven" by `next_uncached_key_range`, which decides
+    /// which ranges of keys need to be calculated based on what cached trie data is available.
+    #[instrument(
+        target = TRACE_TARGET,
+        level = "trace",
+        skip(self, value_encoder, targets, hashed_cursor_current),
+    )]
+    fn calculate_key_range(
```
mediocregopher, line 619: This method now contains the logic which was previously in `proof_inner` to construct the trie based on a sequence of keys (leafs). It's only significant change is that it is now bounded.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,454 @@ where
...
+        }
+    }
+
+    /// Pushes a new branch onto the `branch_stack` which is based on a cached branch obtained via
+    /// the trie cursor.
+    ///
+    /// If there is already a child at the top branch of `branch_stack` occupying this new branch's
+    /// nibble then that child will have its short-key split with another new branch, and this
+    /// cached branch will be a child of that splitting branch.
+    fn push_cached_branch(
```
mediocregopher, line 681: This method is the sibling of `push_leaf`. Where `push_leaf` was used to handle all changes required to push a single leaf onto the `child_stack`, this method handles all changes required to push a single cached branch onto the `branch_stack`.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,454 @@ where
...
+    /// # Returns
+    ///
+    /// - `None`: No more data to process, finish computation
+    ///
+    /// - `Some(lower, None)`: Indicates to call `push_leaf` on all keys starting at `lower`, with
+    ///   no upper bound. This method won't be called again after this.
+    ///
+    /// - `Some(lower, Some(upper))`: Indicates to call `push_leaf` on all keys starting at `lower`,
+    ///   up to but excluding `upper`, and then call this method once done.
+    #[instrument(target = TRACE_TARGET, level = "trace", skip_all)]
```
mediocregopher, line 853: This method is the core of this PR. It becomes the main "driver" of all the logic of the proof calculator, in that it simultaneously iterates over the trie cursor, pushing cached branches as it can, and decides when sections of the trie need to be calculated via their leaves using `calculate_key_range`.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,454 @@ where
...
+        }
+    }
+
     /// Internal implementation of proof calculation. Assumes both cursors have already been reset.
     /// See docs on [`Self::proof`] for expected behavior.
     fn proof_inner(
         &mut self,
         value_encoder: &VE,
-        targets: impl IntoIterator<Item = Nibbles>,
+        targets: impl IntoIterator<Item = B256>,
```
mediocregopher, line 1076: The changes to `should_retain` necessitated that targets are passed in as B256, so that they are in both lexicographic and depth-first order. This should be fine, I can't think of any use-cases where we'd want to pass in paths which aren't full keys anyway.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+                // - `commit_last_child` will check if the last child needs to be retained. We need
+                //   to check that before the subsequent `should_retain` call here to prevent
+                //   `targets` from being moved beyond the last child before it is checked.
+                // - If we do end up using the cached hash value, then we will need to commit the
+                //   last child before pushing a new one onto the stack anyway.
+                self.commit_last_child(targets)?;
+
+                if !self.should_retain(targets, &child_path) {
+                    // Pull this child's hash out of the cached branch node. To get the hash's index
+                    // we first need to calculate the mask of which cached hash's have already been
```
yongkangc:
```suggestion
                    // we first need to calculate the mask of which cached hashes have already been
```

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -303,8 +390,7 @@ where
         leaf_val: VE::DeferredEncoder,
     ) -> Result<(), StateProofError> {
         // Before pushing the new leaf onto the `child_stack` we need to commit the previous last
-        // child (ie the first child of this new branch), so that only `child_stack`'s final child
-        // is a non-RlpNode.
+        // child, so that only `child_stack`'s final child is a non-RlpNode.
         self.commit_last_child(targets)?;
 
         // Once the first child is committed we set the new child's bit on the top branch's
```
yongkangc: should this be first child or is it last child?


i,e
`self.commit_last_child(targets)?;`

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+    #[instrument(target = TRACE_TARGET, level = "trace", skip_all)]
+    fn next_uncached_key_range(
+        &mut self,
+        targets: &mut TargetsIter<impl Iterator<Item = Nibbles>>,
+        trie_cursor_state: &mut TrieCursorState,
+        hashed_key_current_path: Option<Nibbles>,
+    ) -> Result<Option<(Nibbles, Option<Nibbles>)>, StateProofError> {
+        // All trie data prior to the current cached branch, if any, has been computed. Any branches
+        // which were under-construction previously, and which are not on the same path as this
+        // cached branch, can be assumed to be completed; they will not have any further keys added
```
yongkangc, line 863:
```suggestion
        // Pop any under-construction branches that are now complete.
        // All trie data prior to the current cached branch, if any, has been computed. Any branches
        // which were under-construction previously, and which are not on the same path as this
        // cached branch, can be assumed to be completed; they will not have any further keys added.
```

Found that it was slightly hard to follow this. Would help i think if we mention what the purpose is and followed by the why.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+        if let Some(cached_path) = self.cached_branch_stack.last().map(|kv| kv.0) {
+            while !cached_path.starts_with(&self.branch_path) {
+                self.pop_branch(targets)?;
+            }
+        }
+
+        // `lower_bound` will be used to track the lower bound of the range which is returned from
+        // this method. If this is None then there are no further keys which need to be processed.
+        //
+        // This starts off being based off of the hashed cursor's current position, which is the
```
yongkangc:
```suggestion
        // This starts off being based on the hashed cursor's current position, which is the
```
small nit, the offs got to me 😆

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+            //
+            // NOTE we pop off the `cached_branch_stack` because cloning the `BranchNodeCompact`
+            // means cloning an Arc, which incurs synchronization overhead. We have to be sure to
+            // push the cached branch back onto the stack once done.
+            let (cached_path, cached_branch) = match (
+                self.cached_branch_stack.pop(),
+                &trie_cursor_state,
+                lower_bound.as_ref(),
+            ) {
+                (Some(cached), _, _) => cached,
```
yongkangc: i find these match condition slightly hard to follow as it requires some sort of context understanding on the possible states. 

e.g 
```
// Use Queued Cache, cached_branch_stack has something
(Some(cached), _, _) => cached

// No More Cache Data
(None, TrieCursorState::Exhausted, _) | (_, _, None) 

```

had to generate an ascii digram to understand the flow better
https://gist.github.com/yongkangc/1e571093db035be6e09942b8eb93cbfe

would a helper method with enum be better here?

yongkangc: for instance, something like 

```rust
  /// Result of trying to get the next cached branch to process.
  enum NextCachedBranch {
      /// Use this cached branch
      Ready(Nibbles, BranchNodeCompact),
      /// No more cached data, return this range to calculate
      NoCacheRemaining(Option<Nibbles>),
      /// Cursor is behind, seek to this position and retry
      SeekCursorTo(Nibbles),
      /// Need to calculate leaves before the cached branch (catch-up range)
      CatchUpFirst { lower: Nibbles, upper: Nibbles },
  }
```

then main loop simplifies to 
```rust
  loop {
	// helper
      let (cached_path, cached_branch) = match self.get_next_cached_branch(
          &mut trie_cursor_state,
          lower_bound,
          hashed_key_current_path,
      )? {
          NextCachedBranch::Ready(path, branch) => (path, branch),
          NextCachedBranch::NoCacheRemaining(lower) => {
              return Ok(lower.map(|l| (l, None)));
          }
          NextCachedBranch::SeekCursorTo(pos) => {
              *trie_cursor_state = TrieCursorState::new(self.trie_cursor.seek(pos)?);
              continue
          }
          NextCachedBranch::CatchUpFirst { lower, upper } => {
              return Ok(Some((lower, Some(upper))));
          }
      };

      // ... rest of the loop uses (cached_path, cached_branch)
  }
```

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+            // cache.
+            let curr_branch =
+                self.branch_stack.last().expect("top of branch_stack corresponds to cached branch");
+
+            let cached_state_mask = cached_branch.state_mask.get();
+            let curr_state_mask = curr_branch.state_mask.get();
+
+            // Determine all child nibbles which are set in the cached branch but not the
+            // under-construction branch.
+            let next_child_nibbles = curr_state_mask ^ cached_state_mask;
```
yongkangc, line 936: isee, we use xor to determine the remaining nibble that is cached but yet to processed

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +646,450 @@ where
...
+            if cached_branch.hash_mask.is_bit_set(child_nibble) {
+                // Commit the last child. We do this here for two reasons:
+                // - `commit_last_child` will check if the last child needs to be retained. We need
+                //   to check that before the subsequent `should_retain` call here to prevent
+                //   `targets` from being moved beyond the last child before it is checked.
+                // - If we do end up using the cached hash value, then we will need to commit the
+                //   last child before pushing a new one onto the stack anyway.
+                self.commit_last_child(targets)?;
+
+                if !self.should_retain(targets, &child_path) {
```
yongkangc, line 986: i see this is the main cache hash optimisation where we make use of the cache

yongkangc (approved):
left some comments, i got the logic of `next_uncached_key_range` and it makes sense to me where we want to 

1. Get next cached branch from the stack or cursor
2. Pop branch, find the next unprocessed child via xor mask
3. If the child has cached hash and not needed for proof -> use it
4. else we wanna either push nested cache or return to calculate hash from leaves
5. loop until done

also thanks for the comments :) 

would prob be useful for @shekhirin or @Rjected to take a look

crates/trie/trie/src/proof_v2/mod.rs
yongkangc: is there a need to keep depth first in the future?

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -25,7 +25,12 @@ pub(crate) enum ProofTrieBranchChild<RF> {
         child: RlpNode,
     },
     /// A branch node whose children have already been flattened into [`RlpNode`]s.
-    Branch(BranchNode),
+    Branch {
+        /// The node itself, for use during RLP encoding.
+        node: BranchNode,
+        /// Bitmasks carried over from cached `BranchNodeCompact` values, if any.
+        masks: TrieMasks,
```
mattsse, line 32: I see, this is effectively the big unlock here

crates/trie/trie/src/proof_v2/node.rs
```diff
@@ -98,8 +103,7 @@ impl<RF: DeferredValueEncoder> ProofTrieBranchChild<RF> {
             Self::Extension { short_key, child } => {
                 (TrieNode::Extension(ExtensionNode { key: short_key, child }), TrieMasks::none())
             }
-            // TODO store trie masks on branch
```
mattsse, line 101: 👍

crates/trie/trie/src/proof_v2/mod.rs
mattsse: I guess we'd need to track the state then somehow so we know which to convert?

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: I'd be open to this, could definitely be easier to read than this giant match which I agree is a beast

mattsse (commented):
I was more or less able to follow this thanks to the great docs, couldn't spot any obvious issues but I'm also not an expert when it comes to reviewing trie

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Yes we still need to do depth-first for the cases of iterating over hashed keys to build the trie from leaves. This implementation is still _technically_ depth-first, it just happens to handle lexicographic as well, so kind of a hybrid.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: We can tell the state of the child by which variant of the enum it is. The difficulty here is more in the `commit_child` logic, which would become `maybe_commit_child` given this change, so a bit confusing, didn't want to introduce that here.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Good catch, this comment was wrong

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Done :+1:

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -795,20 +1300,101 @@ impl<I: Iterator<Item: Copy>> Iterator for WindowIter<I> {
...
+    /// Need to calculate leaves from this range (exclusive upper) before the cached branch
+    /// (catch-up range). If None then
+    CalculateLeaves((Nibbles, Option<Nibbles>)),
+}
+
+/// Increments the nibbles and strips any trailing zeros.
+///
+/// This function wraps `Nibbles::increment` and when it returns a value with trailing zeros,
+/// it strips those zeros using bit manipulation on the underlying U256.
+fn increment_and_strip_trailing_zeros(nibbles: &Nibbles) -> Option<Nibbles> {
```
mediocregopher, line 1359: @mattsse @yongkangc I had to add this to address an issue found by the proptests. Basically once done processing a child of a branch we use `increment` to update the `uncalculated_lower_bound`, which is used to seek forward the trie cursor. Previously if the child was at `0x1F`, `increment` would return `0x20`, but if there were a trie node at `0x2` this would get skipped. This now strips the zeros after increment so we don't skip that node.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -599,6 +1092,12 @@ where
         #[cfg(not(debug_assertions))]
         let targets = targets.into_iter();
 
+        // Convert B256 targets into Nibbles.
+        let targets = targets.into_iter().map(|key| {
+            // SAFETY: key is a B256 and so is exactly 32-bytes.
+            unsafe { Nibbles::unpack_unchecked(key.as_slice()) }
```
shekhirin, line 1098: I feel like we should add to `nybbles` a safe function for unpacking `[u8; 32]`, basically a full path

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +605,485 @@ where
...
+    fn next_uncached_key_range(
+        &mut self,
+        targets: &mut TargetsIter<impl Iterator<Item = Nibbles>>,
+        trie_cursor_state: &mut TrieCursorState,
+        hashed_key_current_path: Option<Nibbles>,
+    ) -> Result<Option<(Nibbles, Option<Nibbles>)>, StateProofError> {
+        // Pop any under-construction branches that are now complete.
+        // All trie data prior to the current cached branch, if any, has been computed. Any branches
+        // which were under-construction previously, and which are not on the same path as this
+        // cached branch, can be assumed to be completed; they will not have any further keys added.
```
shekhirin:
```suggestion
        // cached branch, can be assumed to be completed; they will not have any further keys added
```

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +605,485 @@ where
...
+        // to them.
+        if let Some(cached_path) = self.cached_branch_stack.last().map(|kv| kv.0) {
+            while !cached_path.starts_with(&self.branch_path) {
```
shekhirin, line 866: does this check ensure that `self.pop_branch` will not panic?

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +605,485 @@ where
...
+        trace!(target: TRACE_TARGET, cached=?cached, "Pushed next trie node onto cached_branch_stack");
+
+        // If the calculated range is not caught up to the next cached branch it means there
+        // are portions of the trie prior to that branch which may need to be calculated;
+        // return the uncalculated range up to that branch to make that happen.
+        //
+        // If the next cached branch's path is all zeros then we can skip this catch-up step,
+        // because there cannot be any keys prior to that range.
+        let cached_path = &cached.0;
+        if uncalculated_lower_bound < cached_path && !PATH_ALL_ZEROS.starts_with(cached_path) {
```
shekhirin, line 824: can probably optimize this in `nybbles` by introducing a `is_zeroes()` method that checks `self.nibbles == U256::ZERO`

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -795,20 +1300,101 @@ impl<I: Iterator<Item: Copy>> Iterator for WindowIter<I> {
...
+
+    // If result is empty, just return it
+    if result.is_empty() {
+        return Some(result);
+    }
+
+    // Get access to the underlying U256 to detect trailing zeros
+    let uint_val = *result.as_mut_uint_unchecked();
+    let non_zero_prefix_len = 64 - (uint_val.trailing_zeros() / 4);
+    result.truncate(non_zero_prefix_len);
```
shekhirin, line 1370: can be upstreamed

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -570,27 +605,485 @@ where
...
+            // cache.
+            let curr_branch =
+                self.branch_stack.last().expect("top of branch_stack corresponds to cached branch");
+
+            let cached_state_mask = cached_branch.state_mask.get();
+            let curr_state_mask = curr_branch.state_mask.get();
+
+            // Determine all child nibbles which are set in the cached branch but not the
+            // under-construction branch.
+            let next_child_nibbles = curr_state_mask ^ cached_state_mask;
```
shekhirin, line 936: may be worth implementing bitwise operations for `TrieMask`, so that we can have good `Debug` formatting for these variables instead of u16 numbers

shekhirin (approved):
All of this makes sense, have only nits mostly about nibbles DX

Think we should ask amp to write docs with diagrams for this new crate when the work is done, will help to take a look at big picture again

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: That is part of it, for `cached_path.starts_with(&self.branch_path)` to be false, `self.branch_path` _must_ be non-zero, as a zero `branch_path` would always be a matching prefix. If `branch_path` is non-zero then there must be a corresponding branch node on the `branch_stack`.

However this serves more to close out any under-construction branches which are prior to the cached one. `branch_path`/`branch_stack` essentially track the currently under-construction branches, each of which must be descended from the previous. At this point in the logic the cached branch is leftover from the previous call to `next_uncached_key_range`, which must have already driven the hashed key forward to catch up to `cached_path`. So any branches prior to `cached_path` must be completed, and can be popped.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: https://github.com/alloy-rs/nybbles/issues/40

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: https://github.com/alloy-rs/trie/issues/114

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: https://github.com/alloy-rs/nybbles/issues/41

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: https://github.com/alloy-rs/nybbles/issues/42

PR #20083, opened 2025-12-02
fix(cli): correct BlockBodyIndices import range in setup()

Fibonacci747:
import_table_with_range expects an inclusive end bound. Using from - 1 and to + 1 could underflow/overflow for edge inputs (from == 0 or to == u64::MAX) and was inconsistent with other stage imports. Switch to Some(from), to to precisely copy [from..=to] and avoid numeric wraparounds.


mattsse (commented):
unsure here @joshieDo @shekhirin

PR #20087, opened 2025-12-03
feat(tx-pool): make metrics, listener structs, and fields public

datnguyencse:
This PR exposes the transaction pool listener and metrics APIs publicly by updating their visibility from pub(crate) to pub. This change allows external crates (and projects) to access and utilize these components to implement there own transaction pool.

Could you please review this change? I'd appreciate any feedback or suggestions.

Thanks.


mattsse (approved):
reasonable

PR #20089, opened 2025-12-03
feat(reth-bench-compare): add reth command to summary output

### LINKED ISSUE #20077: Add `reth` run command to `reth-bench-compare` summary
### Describe the feature

Right now we output some metrics like gas per second, metrics on newPayload and FCU latency at the end of the benchmark, but we should also output the exact `reth` commands / flags in the summary

### Additional context

_No response_

Vitaliyr888:
Closes https://github.com/paradigmxyz/reth/issues/20077

### DIFF
--- a/bin/reth-bench-compare/src/comparison.rs
+++ b/bin/reth-bench-compare/src/comparison.rs
@@ -21,9 +21,7 @@ pub(crate) struct ComparisonGenerator {
     feature_ref_name: String,
     baseline_results: Option<BenchmarkResults>,
     feature_results: Option<BenchmarkResults>,
-    /// Reth command used for baseline run (stored independently of results)
     baseline_command: Option<String>,
-    /// Reth command used for feature run (stored independently of results)
     feature_command: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


bin/reth-bench-compare/src/comparison.rs
```diff
@@ -206,6 +208,29 @@ impl ComparisonGenerator {
         Ok(())
     }
 
+    /// Set the reth command for a reference
+    pub(crate) fn set_ref_command(&mut self, ref_type: &str, command: String) -> Result<()> {
+        match ref_type {
+            "baseline" => {
+                if let Some(ref mut results) = self.baseline_results {
```
mediocregopher: Why does it matter if there are results yet or not?

yongkangc: I think we would wanna capture the command regardless of whether it succeeded or failed @Vitaliyr888

yongkangc (changes requested):
Most of it looks fine, just one question on success / failure that @mediocregopher pointed out

bin/reth-bench-compare/src/comparison.rs
Vitaliyr888: Makes sense! Moved commands to separate fields

mediocregopher (approved):

yongkangc (approved):

PR #20092, opened 2025-12-03
feat(cli): New subcommand: reth db search-changesets

mediocregopher:
This should not be merged because it will have to be rewritten once static file changesets are finished. It's useful to have around in the meantime though.


emmajam:
@mediocregopher do we still want this?

Issue #20069, opened 2025-12-02
Add support for testing_ rpc namespace and testing_buildBlockV1


mattsse:
### Describe the feature

see https://github.com/marcindsobczak/execution-apis/blob/main/src/testing/testing_buildBlockV1.md

for this we need 

a new rpc server trait and impl for the rpc function similar to

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/rpc/rpc/src/validation.rs#L47-L52

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/rpc/rpc/src/validation.rs#L478-L479

an implementation (can be eth specific for now)
a definition of the request type, similar to 

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/payload/primitives/src/traits.rs#L167-L167

but with the transactions (this is actually similar to the op payload attributes:

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/payload/primitives/src/traits.rs#L182-L182

then we need to tune 

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/ethereum/payload/src/lib.rs#L138-L151

so that we can build the block with the transactions

the actually integration of this if `--http.api testing` is provided we could do later, here for example:

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/ethereum/node/src/node.rs#L293-L293

PR #20094, opened 2025-12-03
feat: add support for testing_ rpc namespace and testing_buildBlockV1

0xKarl98:
This closes issue #20069  , summerized as below :
 
 - Implemented the testing_buildBlockV1 RPC namespace (testing-only, default disabled) per spec: 
    - Added TestingApi and TestingBuildBlockRequest (camelCase, payloadAttributes+transactions+extraData) in rpc-ap
    
    - Implemented server-side handler and builder: TestingBlockBuilder trait and EthTestingBlockBuilder execute only the provided raw signed transactions (order preserved), reject blob tx without sidecars, apply payloadAttributes/extraData, and return ExecutionPayloadEnvelopeV4 without touching the canonical chain.
  
    - Wired conditional registration: hidden CLI flags --http.testing / --ws.testing add testing to module selection, and in Ethereum node we construct EthTestingBlockBuilder and merge TestingApi only when the testing module is explicitly requested. Security note added to API docs
    
    - Expose a hidden --testing.max-concurrent flag (default 1), and wire testing RPC instantiation to use a semaphore + spawn_blocking so heavy block builds don’t block the async runtime and have bounded concurrency
   
References : 
[testing_buildBlockV1.md](https://github.com/marcindsobczak/execution-apis/blob/main/src/testing/testing_buildBlockV1.md) https://github.com/ethereum/execution-apis/pull/710

cc @mattsse


crates/node/core/src/args/rpc_server.rs
mattsse, line 1: these args we can exclude, we can make it so that --http.api also accepts testing

crates/rpc/rpc-api/src/testing.rs
```diff
@@ -0,0 +1,45 @@
...
+use jsonrpsee::proc_macros::rpc;
+use serde::{Deserialize, Serialize};
+
+/// Capability string for `testing_buildBlockV1`.
+pub const TESTING_BUILD_BLOCK_V1: &str = "testing_buildBlockV1";
+
+/// Request payload for `testing_buildBlockV1`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TestingBuildBlockRequest {
```
mattsse: let's name this V1
```suggestion
pub struct TestingBuildBlockRequestV1 {
```

crates/rpc/rpc-builder/src/config.rs
mattsse, line 1: we can exclude this from here, and instead instantiate and install here

https://github.com/paradigmxyz/reth/blob/a62b46a96070df0c6925238439b3c4ed16df5eaf/crates/ethereum/node/src/node.rs#L311-L314

crates/rpc/rpc-builder/src/lib.rs
mattsse, line 1: these changes we can all undo because we can instantiate this on demand specifically for eth

similar to flashbots api:

https://github.com/paradigmxyz/reth/blob/a62b46a96070df0c6925238439b3c4ed16df5eaf/crates/ethereum/node/src/node.rs#L307-L310

crates/rpc/rpc/src/testing.rs
```diff
@@ -0,0 +1,243 @@
...
+        + Send
+        + Sync
+        + 'static,
+    EthEvmConfig<ChainSpec>:
+        ConfigureEvm<Primitives = EthPrimitives, NextBlockEnvCtx = NextBlockEnvAttributes>,
+{
+    async fn build_block(
+        &self,
+        request: TestingBuildBlockRequest,
+    ) -> RpcResult<ExecutionPayloadEnvelopeV4> {
```
mattsse: this entire fn logic we need to spawn as blocking, ideally also add a semaphor for this so we can limit how many requests we allow at the same time

mattsse (changes requested):
this looks pretty good already

left some suggestions

crates/node/core/src/args/rpc_server.rs
0xKarl98: Yeah we can exclude that and make it more simple

crates/rpc/rpc/src/testing.rs
0xKarl98: In terms of the semaphor , which number do you think is appropriate by default ?

I just set the concurrent number to at least 1 and can be overriten if value provided

crates/ethereum/node/src/node.rs
```diff
@@ -313,6 +332,29 @@ where
                     .modules
                     .merge_if_module_configured(RethRpcModule::Eth, eth_config.into_rpc())?;
 
+                // testing_buildBlockV1: only wire when the hidden testing module is explicitly
+                // requested on any transport. Default stays disabled to honor security guidance.
+                let testing_requested = {
+                    let cfg = container.modules.module_config();
+                    cfg.contains_http(&RethRpcModule::Other("testing".to_string())) ||
+                        cfg.contains_ws(&RethRpcModule::Other("testing".to_string())) ||
+                        cfg.contains_ipc(&RethRpcModule::Other("testing".to_string()))
```
mattsse: I believe this is just merge_if_configured

crates/ethereum/node/src/node.rs
```diff
@@ -313,6 +332,29 @@ where
...
+                };
+                if testing_requested {
+                    let evm_config: EthEvmConfig<ChainSpec> =
+                        EthEvmConfig::new(testing_chain.clone());
+                    let builder = EthTestingBlockBuilder::new(
+                        testing_provider,
+                        evm_config,
+                        EthereumBuilderConfig::new(),
+                    );
+                    let testing_api = TestingApi::new(builder, testing_max_concurrent).into_rpc();
```
mattsse: imo we can always instantiate this and just call merge_if_configured

crates/node/core/src/args/rpc_server.rs
```diff
@@ -71,6 +71,10 @@ pub struct RpcServerArgs {
     #[arg(long)]
     pub ws: bool,
 
+    /// Max concurrent `testing_buildBlockV1` requests (hidden, defaults to 1).
+    #[arg(long = "testing.max-concurrent", default_value_t = 1usize, hide = true)]
+    pub testing_max_concurrent: usize,
```
mattsse: we can make some default assumptions here and dont really need this as a setting imo

crates/rpc/rpc/src/testing.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/// Trait for types that can build a block for `testing_buildBlockV1`.
+#[async_trait]
+pub trait TestingBlockBuilder: Send + Sync + std::fmt::Debug + 'static {
+    /// Build a block according to the testing API request.
+    async fn build_block(
+        &self,
+        request: TestingBuildBlockRequestV1,
+    ) -> RpcResult<ExecutionPayloadEnvelopeV4>;
+}
```
mattsse: I think this api could potentially be useful for other networks as well, so we can make an effort here to make this generic over the types, like N: NodePrimitives perhaps, or at least the actual TestingApi

crates/ethereum/node/src/node.rs
```diff
@@ -270,12 +275,22 @@ impl<N, EthB, PVB, EB, EVB, RpcMiddleware> NodeAddOns<N>
...
+    N::Provider: BlockReaderIdExt<
+            Block = EthBlock,
+            Header = <EthBlock as BlockTrait>::Header,
+            Transaction = TransactionSigned,
+            Receipt = Receipt,
+        > + StateProviderFactory
+        + ChainSpecProvider<ChainSpec = ChainSpec>
+        + Send
+        + Sync
+        + 'static,
```
mattsse: I'd like to avoid these changes if possible, I think the entire block building logic is already sufficiently generic enough and we can make a few changes to the TestingApi to route the types accordingly

crates/ethereum/node/src/node.rs
```diff
@@ -313,6 +332,29 @@ where
...
+                // requested on any transport. Default stays disabled to honor security guidance.
+                let testing_requested = {
+                    let cfg = container.modules.module_config();
+                    cfg.contains_http(&RethRpcModule::Other("testing".to_string())) ||
+                        cfg.contains_ws(&RethRpcModule::Other("testing".to_string())) ||
+                        cfg.contains_ipc(&RethRpcModule::Other("testing".to_string()))
+                };
+                if testing_requested {
+                    let evm_config: EthEvmConfig<ChainSpec> =
+                        EthEvmConfig::new(testing_chain.clone());
```
mattsse: all of these should be obtainable from the `ctx` type I believe

crates/rpc/rpc/src/testing.rs
```diff
@@ -0,0 +1,267 @@
...
+            .semaphore
+            .clone()
+            .acquire_owned()
+            .await
+            .map_err(|_| internal_rpc_err("testing_buildBlockV1 concurrency limiter closed"))?;
+
+        let builder = self.builder.clone();
+        let handle = Handle::current();
+        let join_handle =
+            tokio::task::spawn_blocking(move || handle.block_on(builder.build_block(request)));
```
mattsse: we can give this type a `Box<dyn TaskExecutor>` which can also obtain from the `ctx` where this type is created

crates/rpc/rpc/src/testing.rs
mattsse: I think let's go with just 1 by default for now

mattsse (changes requested):
cool, a few more nits

would be great to get a simple e2e test for this that calls this, see ethereum/node/tests/it for reference

paging @klkvr for suggestions re generics

crates/ethereum/node/src/node.rs
0xKarl98: yeah should use merge_if_module_configured instead

crates/ethereum/node/src/node.rs
0xKarl98: Yeah true

crates/node/core/src/args/rpc_server.rs
0xKarl98: Ahh alright let's go by default

crates/rpc/rpc/src/testing.rs
0xKarl98: done

crates/ethereum/node/src/node.rs
0xKarl98: Yeah have reduced the unnessary part

mattsse (approved):

PR #20095, opened 2025-12-03
fix(txpool): derive accurate queued reason for SubPool::Blob

Bashmunta:
Blob subpool aggregates all non-pending blob transactions, not just those missing the blob fee. The previous implementation always reported InsufficientBlobFee, misclassifying cases such as nonce gaps, insufficient balance, too much gas, parked ancestors, or insufficient base fee. This change derives the queued reason for blob transactions from TxState flags in a consistent order (structural first, then fees), and adds unit tests to cover the major combinations. Sub-pool routing is unchanged; only the reported reason is corrected.

### DIFF
--- a/crates/transaction-pool/src/pool/state.rs
+++ b/crates/transaction-pool/src/pool/state.rs
@@ -330,7 +330,7 @@ mod tests {
     #[test]
     fn test_blob_reason_insufficient_base_fee() {
         // Blob tx with all structural bits set and blob fee sufficient, but base fee insufficient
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I see, this doesnt look super pretty, but best what we can do and not that horrible

Bashmunta:
> pending clippy lint fixes

Did it mister

mattsse (approved):

yongkangc (approved):
this would be useful

PR #20096, opened 2025-12-03
chore(storage): Add trait object error variant to `DatabaseError`

emhane:
This PR contains one of the bindings which is needed to use `DatabaseError` in a type safe way in l2 code, for example in the OP historical proofs workstream. This allows downcast to OP error enum as opposed to adding additional wrapper cursor traits to be able to propagate OP errors in a type safe way (matching on error strings out ruled).

- Adds variant wrapping trait object error to `DatabaseError`
- Removes redundant `PartialEq` impl for `DatabaseError`, this was almost exclusively used in tests
- Updates tests to match on errors or compare the error variant data instead or relying on `PartialEq`

Ref https://github.com/op-rs/op-reth/pull/388

### DIFF
--- a/crates/storage/errors/src/db.rs
+++ b/crates/storage/errors/src/db.rs
@@ -2,15 +2,17 @@ use alloc::{
     boxed::Box,
     format,
     string::{String, ToString},
+    sync::Arc,
     vec::Vec,
 };
 use core::{
+    error::Error,
     fmt::{Debug, Display},
     str::FromStr,
 };
 
 /// Database error type.
-#[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
+#[derive(Clone, Debug, thiserror::Error)]
 pub enum DatabaseError {
     /// Failed to open the database.
     #[error("failed to open the database: {_0}")]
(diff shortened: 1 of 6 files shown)


mattsse (commented):
I think this would be quite useful, now that we rollout another db for index data 

wdyt @Rjected

crates/storage/errors/src/db.rs
```diff
@@ -48,6 +50,9 @@ pub enum DatabaseError {
     /// Other unspecified error.
     #[error("{_0}")]
     Other(String),
+    /// Other unspecified error.
+    #[error(transparent)]
+    Custom(#[from] Arc<dyn Error + Send + Sync>),
```
mattsse, line 55: I believe this should be okay, might also be useful for different db storages

mattsse (approved):
seems okay,
diff here is only test code

any objections @Rjected @joshieDo

crates/storage/errors/src/db.rs
emhane: perhaps we remove the variant `DatabaseError::Other(String)` variant eventually @mattsse ? just didn't want to introduce that breaking change in this pr

Issue #20054, opened 2025-12-01
handle more dedicated ethsimulate rpc errors 


mattsse:
### Describe the feature

currently simulatev1 returns:

https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-api/src/helpers/call.rs#L57-L57

which is https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/mod.rs#L59-L59

however ethsimulatev1 uses specific error codes that deviate from ethapi 

https://github.com/ethereum/execution-apis/blob/f910189261255eb0ebc49ee0e6f10f48562a41c3/src/eth/execute.yaml#L129

see also:

https://github.com/ethereum/go-ethereum/blob/6f2cbb7a27ba7e62b0bdb2090755ef0d271714be/internal/ethapi/errors.go#L102-L118

we can improve this by extending

https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/simulate.rs#L30-L40

ethsimulatev1 with the specified errors as variants and then enforce conversions from ethapierror, because the overlap with 

https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/mod.rs#L108-L110

and https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/mod.rs#L632-L634

is quite huge

so we could do add additional helpers for these to https://github.com/paradigmxyz/reth/blob/b3c00ed602f2a8805974f25be714a3bea26901e9/crates/rpc/rpc-eth-types/src/error/api.rs#L53-L53
or require that the error can be converted into an etsimulateerror

or we do 

```rust
enum EthSimulateErr<Err> {
 ...,
 Other(Err) 
}
```

and return that 

### Additional context

maybe @figtracer as well?

PR #20099, opened 2025-12-03
feat(rpc): handle dedicated eth_simulate errors

figtracer:
solves #20054 by:
- extending `EthSimulateError` following [execution-apis](https://github.com/ethereum/execution-apis/blob/f910189261255eb0ebc49ee0e6f10f48562a41c3/src/eth/execute.yaml#L129) and [geth/ethapi/errors](https://github.com/ethereum/go-ethereum/blob/6f2cbb7a27ba7e62b0bdb2090755ef0d271714be/internal/ethapi/errors.go#L102-L118)
- extending the trait `AsEthApiError` with as_simulate_error() to match RpcInvalidTransactionError counterparts
- small refactor at `simulate_v1` using this new method

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -29,7 +29,7 @@ use reth_revm::{database::StateProviderDatabase, db::State};
 use reth_rpc_convert::{RpcConvert, RpcTxReq};
 use reth_rpc_eth_types::{
     cache::db::StateProviderTraitObjWrapper,
-    error::FromEthApiError,
+    error::{AsEthApiError, FromEthApiError},
     simulate::{self, EthSimulateError},
     EthApiError, StateCacheDb,
 };
@@ -159,6 +159,13 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
                         .context_for_next_block(&parent, this.next_env_attributes(&parent)?)
                         .map_err(RethError::other)
                         .map_err(Self::Error::from_eth_err)?;
+                    let map_err = |e: EthApiError| -> Self::Error {
+                        match e.as_simulate_error() {
+                            Some(sim_err) => Self::Error::from_eth_err(EthApiError::other(sim_err)),
+                            None => Self::Error::from_eth_err(e),
+                        }
+                    };
+
                     let (result, results) = if trace_transfers {
                         // prepare inspector to capture transfer inside the evm so they are recorded
                         // and included in logs
@@ -173,7 +180,8 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
                             default_gas_limit,
                             chain_id,
                             this.converter(),
-                        )?
+                        )
+                        .map_err(map_err)?
                     } else {
                         let evm = this.evm_config().evm_with_env(&mut db, evm_env);
                         let builder = this.evm_config().create_block_builder(evm, &parent, ctx);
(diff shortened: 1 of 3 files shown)


crates/rpc/rpc-eth-types/src/simulate.rs
```diff
@@ -203,25 +263,31 @@ where
         let call = match result {
             ExecutionResult::Halt { reason, gas_used } => {
                 let error = Err::from_evm_halt(reason, tx.gas_limit());
+                let message = error.to_string();
+                // Use simulate-specific error code if available, otherwise fall back to standard
+                let code = error
+                    .as_simulate_error()
+                    .map(|sim_err| sim_err.error_code())
+                    .unwrap_or_else(|| error.into().code());
```
figtracer: this is not needed at all i'll just remove this

crates/rpc/rpc-eth-types/src/error/api.rs
```diff
@@ -74,6 +74,32 @@ pub trait AsEthApiError {
 
         false
     }
+
+    /// Returns [`EthSimulateError`] if this error maps to a simulate-specific error code.
+    fn as_simulate_error(&self) -> Option<EthSimulateError> {
```
mattsse, line 79: okay, this should actually work

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -159,6 +159,13 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
                         .context_for_next_block(&parent, this.next_env_attributes(&parent)?)
                         .map_err(RethError::other)
                         .map_err(Self::Error::from_eth_err)?;
+                    let map_err = |e: EthApiError| -> Self::Error {
+                        match e.as_simulate_error() {
+                            Some(sim_err) => Self::Error::from_eth_err(EthApiError::other(sim_err)),
+                            None => Self::Error::from_eth_err(e),
+                        }
+                    };
```
mattsse, line 167: this isn't too bad either

mattsse (approved):
I think this should do it

wdyt @klkvr

crates/rpc/rpc-eth-types/src/simulate.rs
```diff
@@ -36,12 +36,67 @@ pub enum EthSimulateError {
...
+            Self::BaseFeePerGasTooLow => -38012,
+            Self::IntrinsicGasTooLow => -38013,
+            Self::InsufficientFunds { .. } => -38014,
             Self::BlockGasLimitExceeded => -38015,
+            Self::BlockNumberInvalid => -38020,
+            Self::BlockTimestampInvalid => -38021,
+            Self::PrecompileSelfReference => -38022,
+            Self::PrecompileDuplicateAddress => -38023,
+            Self::SenderNotEOA => -38024,
+            Self::MaxInitCodeSizeExceeded => -38025,
```
klkvr, line 99: would be also great to check if any of those codes should apply here https://github.com/paradigmxyz/reth/blob/7e6a59b6acf0ab1390da416293ca0b3e5d11fca8/crates/rpc/rpc-eth-types/src/error/mod.rs#L695-L709

klkvr (approved):
this makes sense, though I think it's likely that there are some error code fixes that not only apply to simulateV1

crates/rpc/rpc-eth-types/src/simulate.rs
figtracer: i agree, let me take a look at it

figtracer: actually i think we can leave it like this, there's nothing else to be added here
```rust
     pub const fn error_code(&self) -> i32 { 
         match self { 
             Self::InvalidChainId | 
             Self::GasTooLow | 
             Self::GasTooHigh | 
             Self::GasRequiredExceedsAllowance { .. } | 
             ...
        } 
     }
```

the generic json-rpc codes stay at `RpcInvalidTransactionError` and the conversion for  (if needed) is handled by `as_simulate_error()`. the ones that don't match are specific

PR #20101, opened 2025-12-03
fix(chainspec): add ChainConfig to StatelessInput and add ChainConfig creator helpers

### LINKED ISSUE #17805: Evaluate working solution to use ChainConfig as zkVM input for the stateless crate
### Describe the feature

This is a feature request to allow [ChainConfig](https://github.com/alloy-rs/alloy/blob/14ad31d96ef1b168294703ba9ffe016ace5d96bc/crates/genesis/src/lib.rs#L299) to be the primary data structure used for zkVMs to receive the corresponding chain configuration for stateless block validation.

### Additional context

Today, reth has the [stateless crate](https://github.com/paradigmxyz/reth/tree/main/crates/stateless), which provides the [stateless_validation](https://github.com/paradigmxyz/reth/blob/main/crates/stateless/src/validation.rs#L132-L137) function to validate a block statelessly. To use this in zkVMs, the corresponding guest program must, roughly, receive those parameters via zkVM-inputs and call that function. The zkVM will generate a proof that computation was done correctly.

The way this works is that those inputs (i.e.,: block, witness, and chain configuration) are passed as bytes to the zkVM, usually Bincode-encoded to have an efficient deserialization (since this is also part of the proving).

Serializing blocks and witnesses works quite naturally, using native structs present in reth or alloy and simply bincode them (sometimes leveraging some existing features in the creates, since native structs aren’t Bincode-friendly). So that’s okay. 

Now, the other important part and the one related to this issue, is providing the input to create the correct `ChainSpec` of `stateless_validation(...)`. The `ChainSpec` indicates many required configurations for block validation and EVM execution such as: hardfork configurations, deposit contract address, and other stuff. 

`ChainSpec` is a complex structure that is not ideal to serialize. In a reth stateless validation guest program we’re using for benchmarking, we ended up using [ChainConfig](https://github.com/alloy-rs/alloy/blob/14ad31d96ef1b168294703ba9ffe016ace5d96bc/crates/genesis/src/lib.rs#L299) which is an idea from @kevaundray from a while ago — recently [he added Bincode-support](https://github.com/alloy-rs/alloy/pull/2630).

The idea is that we could use `ChainConfig` as the input for the zkVM, and leverage this [From<Genesis> for ChainSpec](https://github.com/paradigmxyz/reth/blob/main/crates/chainspec/src/spec.rs#L638) trait implementation. i.e., we create a Genesis with a set ChainConfig, and then do ` let _ : ChainSpec = genesis.into()`.

The advantage of this idea is that it could work quite generically, not only supporting mainnet blocks but also EEST tests/benchmark fixtures. We attempted and succeeded in doing this, but there were some blockers that we had to fix:

- The [MAINNET](https://github.com/paradigmxyz/reth/blob/main/crates/chainspec/src/spec.rs#L89) constructed ChainSpec builds an incomplete Genesis, since it is [deserialized from a JSON file](https://github.com/paradigmxyz/reth/blob/main/crates/chainspec/src/spec.rs#L90). This [JSON file does not contain the `chain_config` field](https://github.com/paradigmxyz/reth/blob/main/crates/chainspec/res/genesis/mainnet.json), which means that we end up having the [default value](https://github.com/alloy-rs/alloy/blob/main/crates/genesis/src/lib.rs#L27-L28). This means that using `MAINNET.genesis.config` to construct `ChainSpec` leads to a wrong result. Note that JSON files for other networks do have the `chain_config` field, so that they would work correctly.
- The `ChainSpecBuilder` used for EEST tests [constructs from MAINNET](https://github.com/paradigmxyz/reth/blob/bcbd2d64cedbde0c8f4f044a74da1bfb9324e2ce/testing/ef-tests/src/models.rs#L317), and thus also have similar problems.
- Some other nits reg full hardfork support in ChainSpecBuilder (more details on this in a PR below)

Hopefully, the above can give a good enough context on the intention and the blockers for doing this.

Now, we fixed all this in a reth fork, which contains all details on the approach:

- [PR with the changes to fix](https://github.com/kevaundray/reth/pull/20) all the mentioned problems.
- If you’re curious how this is [used in the guest program for a zkVM](https://github.com/eth-act/zkevm-benchmark-workload/blob/49fb87f694d573cd41d35d0a0bd41f39e168b77c/ere-guests/stateless-validator/sp1/src/main.rs#L19-L36).

We aren’t entirely sure if the changes in the PR would be the right approach/style to be upstreamed — the goal of this issue is to provide all the context and working solution to get more feedback on how this could be incorporated into main.

Thanks!

jsign:
Fixes #17805

A bit of a TL;DR of the issue and the goal of the PR:
- Add `ChainConfig` into `StatelessInput`. This formalizes that the guest program inputs must also receive the right chain configuration, to be able to setup the evm config and other related things correctly.
- Add a bunch of helpers to create the `ChainConfig` for mainnet or arbitrary chain configurations.

The last point is better explained in the issue. Mainly, the `.json` files from where `MAINNET` is created have an empty `chain_config`, thus `MAINNET.genesis.chain_config` is empty.

The PR basically provides helpers to easily create it for mainnet, which will be the 99.9% of use-cases. And also it exposes the more generic method for other chains -- this is useful for running EEST tests statelessly since each EEST test has a custom chain configuration.

### DIFF
--- a/crates/chainspec/src/spec.rs
+++ b/crates/chainspec/src/spec.rs
@@ -23,7 +23,7 @@ use alloy_eips::{
     eip1559::INITIAL_BASE_FEE, eip7685::EMPTY_REQUESTS_HASH, eip7840::BlobParams,
     eip7892::BlobScheduleBlobParams,
 };
-use alloy_genesis::Genesis;
+use alloy_genesis::{ChainConfig, Genesis};
 use alloy_primitives::{address, b256, Address, BlockNumber, B256, U256};
 use alloy_trie::root::state_root_ref_unhashed;
 use core::fmt::Debug;
@@ -37,6 +37,7 @@ use reth_network_peers::{
     NodeRecord,
 };
 use reth_primitives_traits::{sync::LazyLock, BlockHeader, SealedHeader};
+use std::collections::BTreeMap;
 
 /// Helper method building a [`Header`] given [`Genesis`] and [`ChainHardforks`].
 pub fn make_genesis_header(genesis: &Genesis, hardforks: &ChainHardforks) -> Header {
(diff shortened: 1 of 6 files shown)


crates/chainspec/src/spec.rs
```diff
@@ -936,9 +1061,16 @@ impl ChainSpecBuilder {
         self
     }
```
jsign, line 1050: There were some forks missing here, mainly DAO and the difficulty-bomb related ones. 
Probably not a big deal, but I included them for completeness.

crates/chainspec/src/spec.rs
```diff
@@ -240,6 +247,124 @@ pub static DEV: LazyLock<Arc<ChainSpec>> = LazyLock::new(|| {
...
+        (EthereumHardfork::Bpo4, "bpo4"),
+        (EthereumHardfork::Bpo5, "bpo5"),
+    ];
+
+    for (timestamp, blob_params) in &params.scheduled {
+        for (fork, name) in &bpo_forks {
+            if let ForkCondition::Timestamp(fork_ts) = hardforks.fork(*fork) &&
+                fork_ts == *timestamp
+            {
+                schedule.insert((*name).to_string(), *blob_params);
```
mattsse: I believe this is just fork.name().to_lowercase()

also added and released this https://github.com/alloy-rs/hardforks/pull/84

crates/chainspec/src/spec.rs
```diff
@@ -240,6 +247,124 @@ pub static DEV: LazyLock<Arc<ChainSpec>> = LazyLock::new(|| {
...
+    deposit_contract_address: Option<Address>,
+    blob_schedule: BTreeMap<String, BlobParams>,
+) -> ChainConfig {
+    // Helper to extract block number from a hardfork condition
+    let block_num = |fork: EthereumHardfork| -> Option<u64> {
+        match hardforks.fork(fork) {
+            ForkCondition::Block(n) => Some(n),
+            ForkCondition::TTD { activation_block_number, .. } => Some(activation_block_number),
+            _ => None,
+        }
```
mattsse: I think this is just `.block_number()`

mattsse (changes requested):
supportive, only a few nits

mattsse (approved):
reasonable, ty

PR #20105, opened 2025-12-03
chore: allow empty blobparams in ethconfig

mattsse:
this makes the endpoint also work on other networks that dont have blobs and also matches what geth is doing where the blobschedule field is optional in the first place and can be "null"

we can do this manually on alloy if these are set to 0 for example

https://github.com/ethereum/go-ethereum/blob/6f2cbb7a27ba7e62b0bdb2090755ef0d271714be/params/config.go#L1173-L1193

https://github.com/ethereum/go-ethereum/blob/6f2cbb7a27ba7e62b0bdb2090755ef0d271714be/internal/ethapi/api.go#L1177-L1177

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/config.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/config.rs
@@ -1,7 +1,10 @@
 //! Loads chain configuration.
 
 use alloy_consensus::{BlockHeader, Header};
-use alloy_eips::eip7910::{EthConfig, EthForkConfig, SystemContract};
+use alloy_eips::{
+    eip7840::BlobParams,
+    eip7910::{EthConfig, EthForkConfig, SystemContract},
+};
 use alloy_evm::precompiles::Precompile;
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/config.rs
```diff
@@ -72,14 +74,24 @@ where
...
+                .blob_params_at_timestamp(timestamp)
+                // no blob support, so we set this to zero
+                .unwrap_or(BlobParams {
+                    target_blob_count: 0,
+                    max_blob_count: 0,
+                    update_fraction: 0,
+                    min_blob_fee: 0,
+                    max_blobs_per_tx: 0,
+                    blob_base_cost: 0,
+                }),
```
klkvr: should this be `BlobParams::cancun()` instead? I guess ideally we just make it optional but not worth breaking alloy for this

klkvr (approved):
lgtm

PR #20107, opened 2025-12-03
refactor(trie): move revealed paths tracking into SerialSparseTrie

### LINKED ISSUE #14378: Keep revealed sparse trie paths on trie level
should we keep these on trie level instead?

_Originally posted by @rkrasiuk in https://github.com/paradigmxyz/reth/pull/14370#discussion_r1949548483_

Himess:
## Summary

  - Moves `revealed` paths tracking from separate collections in `SparseStateTrie` into the `SerialSparseTrie` struct itself
  - Adds revealed path methods to `SparseTrieInterface` trait for unified access
  - Simplifies code by removing redundant tracking in `StorageTries`

  Closes #14378


mediocregopher:
I'm sorry to say I don't think this is an impactful change at this point; we don't use SerialSparseTrie by default anymore, and in any case we will be reworking heavily how proofs/revealed nodes are tracked in an upcoming refactor. In the future please make sure to double check on the issue and get it assigned, especially for an older one like this :bow:

PR #20108, opened 2025-12-03
fix: set minimum pruning distance to 64 blocks for trie changesets

joshieDo:
Right now 10k is too much, leading to big tables. And the pipeline stage trims them to 64 blocks anyway, resulting in a potential big freelist.


mattsse (approved):
makes sense

mediocregopher (approved):

PR #20112, opened 2025-12-04
chore(rpc): protect gas_used_ratio from div by zero

JacksonEi:
Adds a division-by-zero check for `gas_used_ratio` calculation in `FeeHistoryEntry::new`, this prevents `NaN` or `Infinity` results if `gas_limit` is zero.


mattsse (commented):
we can ignore this because if the header gaslimit is 0 the entire chain is kinda borked

PR #20114, opened 2025-12-04
chore: update naming convention

### LINKED ISSUE #20080: Update mdbx naming convention for tables
### Describe the feature

In mdbx implementation, the **_create_tables** method calls **create_db** on the **Transaction** struct as seen in the screenshot below, this might be confusing for some people. The method name should be updated to **create_table** to follow the same naming convention as the current version of libmdbx also implements **create_table** and not **create_db**

Image

### Additional context

_No response_

VaibhavArora19:
Closes #20080


mattsse (commented):
this will cause a lot of conflicts with work that's wip

i dont think we should do this

VaibhavArora19:
Yeah, I didn't expect these many changes as well.

PR #20117, opened 2025-12-04
perf(sparse): move-encode account RLP into trie to avoid clone

ANtutov:
Encode account RLP into a fresh Vec and move it into the trie instead of cloning a reusable buffer. This removes one allocation and memcpy per account update. The persistent buffer in SparseStateTrie was removed since the trie must own the bytes; keeping it yielded no benefit and forced a clone. Mirrored the same change in the stateless path for consistency. Only allocation and copy reductions.


mattsse (commented):
this isn't right, this will now end up over allocating max size per acc iteration

and we need to give ownership so the clone is required

PR #20118, opened 2025-12-04
perf(engine): only recover senders once

### LINKED ISSUE #20115: Avoid full recovered block conversion during validation
### Describe the feature

https://github.com/paradigmxyz/reth/blob/a62b46a96070df0c6925238439b3c4ed16df5eaf/crates/engine/tree/src/tree/payload_validator.rs#L176-L187

which we do here:

https://github.com/paradigmxyz/reth/blob/a62b46a96070df0c6925238439b3c4ed16df5eaf/crates/engine/tree/src/tree/payload_validator.rs#L420-L420

most of this time is technically free, but we can reuse all the senders that we already recovered and only do `convert_payload_to_block` and then convert to recoveredblock

### Additional context

_No response_

klkvr:
Closes https://github.com/paradigmxyz/reth/issues/20115

Right now we recover senders in 3 places across the engine — on every downloaded block, when converting a payload to block, and when advancing the tx iterator on a separate thread.

Only the latter is necessary and the rest can simply reuse its outputs instead of doing extra work.

This PR changes engine to operate on `SealedBlock` for un-executed blocks and only construct `RecoveredBlock` once all transactions were executed

### DIFF
--- a/crates/engine/tree/src/tree/block_buffer.rs
+++ b/crates/engine/tree/src/tree/block_buffer.rs
@@ -1,7 +1,7 @@
 use crate::tree::metrics::BlockBufferMetrics;
 use alloy_consensus::BlockHeader;
 use alloy_primitives::{BlockHash, BlockNumber};
-use reth_primitives_traits::{Block, RecoveredBlock};
+use reth_primitives_traits::{Block, SealedBlock};
 use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 
 /// Contains the tree of pending blocks that cannot be executed due to missing parent.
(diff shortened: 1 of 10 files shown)


crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -18,7 +18,7 @@ use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 #[derive(Debug)]
 pub struct BlockBuffer<B: Block> {
     /// All blocks in the buffer stored by their block hash.
-    pub(crate) blocks: HashMap<BlockHash, RecoveredBlock<B>>,
+    pub(crate) blocks: HashMap<BlockHash, SealedBlock<B>>,
```
mattsse, line 21: I think these changes are okay, this should also safe redundant recovery on downloaded blocks because when we execute we can use the recovered txs from the execution just like for payloads

mattsse (approved):
lgtm,

a nice side-effect of this for downloaded blocks is that we no longer spent a lot of resources recovering senders if we download chunks of blocks, and delay this until execution

fyi @yongkangc

yongkangc (approved):
nice

Issue #6100, opened 2024-01-17
Maximum Peer Flags and Connected Peer Logging


SeaMonkey82:
### Describe the feature

Other execution clients have a generalized maximum peer flag rather than individual flags for outbound and inbound peers.  I prefer the ability to cap the overall peer count.  However, if the current implementation is preferable to others, and a generalized maximum peer flag isn't likely to be added, I suggest at least breaking down the connected peers in the client logging to inbound and outbound to give a better idea of which value needs a more restrictive limitation in order to reach an overall target number of peers.

jenpaff:
if i get this right, the issue suggests to just add a `--max-peers` flag, which could be a good first-timer issue

mattsse:
I think we can do that, if we're okay to apply the default ration `out/in` to this flag and derive the slots like this, but definitely +1 on improve log output here

gr8h:
Can i work on this?

PR #20119, opened 2025-12-04
feat: add inbound & outbound peer counts to network metrics & logs

gr8h:
This PR adds inbound & outbound peer counts to network metrics & logs. Logs now display: `connected_peers=X inbound=Y outbound=Z .....`

Partially cover: [6100](https://github.com/paradigmxyz/reth/issues/6100)

### DIFF
--- a/crates/node/events/src/node.rs
+++ b/crates/node/events/src/node.rs
@@ -65,6 +65,14 @@ impl NodeState {
         self.peers_info.as_ref().map(|info| info.num_connected_peers()).unwrap_or_default()
     }
 
+    fn num_inbound_peers(&self) -> usize {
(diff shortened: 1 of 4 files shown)


crates/node/events/src/node.rs
```diff
@@ -425,6 +435,8 @@ where
                         info!(
                             target: "reth::cli",
                             connected_peers = this.state.num_connected_peers(),
+                            inbound = this.state.num_inbound_peers(),
+                            outbound = this.state.num_outbound_peers(),
```
mattsse, line 439: this makes sense, but I'm wondering if we can make this a bit less verbose than inbound/outbound

any ideas @shekhirin @Rjected

gr8h: would `in`/`out` make more sense here?

PR #20121, opened 2025-12-04
feat(cli): repair-trie metrics

mediocregopher:
This adds an optional metrics endpoint to the `reth db repair-trie` sub-command, which can be used to track in prometheus when inconsistencies are found, which makes alerting much more reliable.

This also adds a small log message to include the current db tip, which is convenient to know when debugging issues.


crates/cli/commands/src/db/repair_trie.rs
```diff
@@ -1,36 +1,101 @@
...
 use tracing::{info, warn};
 
 const PROGRESS_PERIOD: Duration = Duration::from_secs(5);
 
+/// Metrics for tracking trie repair inconsistencies
+#[derive(Debug)]
+struct RepairTrieMetrics {
+    account_inconsistencies: Counter,
+    storage_inconsistencies: Counter,
+}
```
mattsse: can we move this below the command struct def and impl, because private type

crates/cli/commands/src/db/repair_trie.rs
```diff
@@ -1,36 +1,101 @@
...
     /// Execute `db repair-trie` command
-    pub fn execute<N: ProviderNodeTypes>(self, tool: &DbTool<N>) -> eyre::Result<()> {
+    pub async fn execute<N: ProviderNodeTypes>(
+        self,
+        task_executor: TaskExecutor,
+        tool: &DbTool<N>,
+    ) -> eyre::Result<()> {
+        // Set up metrics server if requested
+        if let Some(listen_addr) = self.metrics {
+            let config = MetricServerConfig::new(
```
mattsse: right, we need to spawn a server here

crates/ethereum/cli/src/app.rs
```diff
@@ -154,7 +154,7 @@ where
         Commands::ImportEra(command) => runner.run_blocking_until_ctrl_c(command.execute::<N>()),
         Commands::ExportEra(command) => runner.run_blocking_until_ctrl_c(command.execute::<N>()),
         Commands::DumpGenesis(command) => runner.run_blocking_until_ctrl_c(command.execute()),
-        Commands::Db(command) => runner.run_blocking_until_ctrl_c(command.execute::<N>()),
+        Commands::Db(command) => runner.run_command_until_exit(|ctx| command.execute::<N>(ctx)),
```
mattsse: what's the reason for this?

afaict, the command is still blocking?

mattsse (changes requested):
smol pedantic style and 1 q

yongkangc (approved):
makes sense 🚀

crates/ethereum/cli/src/app.rs
mediocregopher: The prometheus server needs a tokio executor. I can try to just create a new runtime inside the blocking command though, not sure it's necessary to affect all `db` sub-commands like this.

mediocregopher: Done :+1:

mattsse (approved):
ah, I see due to the graceful shutdown functions the server task relies on

I believe we can make this a bit nicer if we add support for this run_blocking_until_ctrl_c or make the command itself spawn the blocking task

but this will do as well

PR #20126, opened 2025-12-04
perf(engine): set thread priority to max

shekhirin:
Engine thread is the one doing EVM execution, and it should have the highest priority out of all other threads (RPC / prewarming / state root task)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -409,7 +410,15 @@ where
             evm_config,
         );
         let incoming = task.incoming_tx.clone();
-        std::thread::Builder::new().name("Engine Task".to_string()).spawn(|| task.run()).unwrap();
+        std::thread::Builder::new()
+            .name("Engine Task".to_string())
+            .spawn_with_priority(ThreadPriority::Max, |result| {
```
mattsse: can we document what this actually does?

mediocregopher (approved):

yongkangc (approved):

Issue #20110, opened 2025-12-03
OOM on 1.9.3 (27a8c0f) Post Fusaka


keccakk:
### Describe the bug

Had an OOM post Fusaka, memory seemed to spike up suddenly after a bad block. Reth restarted after about 4 minutes later.

### Steps to reproduce

```
[Service]
User=reth
Group=reth
Type=simple
Restart=always
RestartSec=5
ExecStart=reth node \
  --datadir /xxxxx/mainnet/reth \
  --authrpc.port 8551 \
  --authrpc.jwtsecret /xxxxx/ \
  --log.file.directory /xxxxx/mainnet/reth \
  --discovery.port 30303 \
  --port 30303 \
  --metrics 127.0.0.1:9001 \
  --http \
  --http.addr 0.0.0.0 \
  --http.port 8545 \
  --http.api "eth,net,trace,txpool,web3,rpc" \
  --http.corsdomain "XXXXX,XXXXX"
```

### Node logs

```text
Dec 03 17:22:14 arch reth[2585726]: 2025-12-03T22:22:14.750100Z  INFO Block added to canonical chain number=23935838 hash=0x6b5bfe7e65085295b7a1b74ae61cda2fa2d27646ea7519df81bf038106e3bdbb peers=131 txs=233 gas_used=30.47Mgas gas_throughput=58.64Mgas/second gas_limit=60.00Mgas full=50.8% base_fee=0.02Gwei blobs=7 excess_blobs=250 elapsed=519.643546ms
Dec 03 17:22:14 arch reth[2585726]: 2025-12-03T22:22:14.919121Z  INFO Canonical chain committed number=23935838 hash=0x6b5bfe7e65085295b7a1b74ae61cda2fa2d27646ea7519df81bf038106e3bdbb elapsed=806.2µs
Dec 03 17:22:23 arch reth[2585726]: 2025-12-03T22:22:23.740790Z  INFO Status connected_peers=131 latest_block=23935838
Dec 03 17:22:25 arch reth[2585726]: 2025-12-03T22:22:25.457464Z  INFO Received block from consensus engine number=23935839 hash=0x3b28163fd808f8febb46b220b7bf588e0ac1039f22ac81a7c3c017f622bd15b8
[...]

PR #20127, opened 2025-12-04
feat(engine): run sync state root if not enough parallelism

shekhirin:
Do not run state root task or parallel state root if the host has less than 5 cores.

ref #20110


mattsse (approved):
makes sense

PR #20128, opened 2025-12-04
fix(engine/tree): gate persistence on max(threshold, buffer-target)

SashaMalysehko:
Persistence gating only considered persistence_threshold, while the selection of blocks to persist was capped by memory_block_buffer_target. When the buffer target exceeded the threshold, the engine could repeatedly attempt persistence with an empty set. This change updates should_persist() to use the maximum of the two, ensuring persistence starts only when there are blocks eligible to persist and matching the documented expectation that the buffer target should not exceed the threshold.


mattsse (commented):
if the set is empty, then this is a noop anyway

PR #20131, opened 2025-12-05
fix(rpc): remove redundant blob_gas_used calculation in eth_callBundle

Fallengirl:
blob_gas_used was already computed on line 108 but the same sum was recalculated on line 116 instead of reusing the variable. This was introduced during the alloy 0.15.5 bump refactor.


mattsse (approved):
reasonable, ty

PR #20009, opened 2025-11-27
refactor: use DebugInspector from revm-inspectors

reject-i:
Closes #19932

PR #20132, opened 2025-12-05
chore: bump revm inspectors

mattsse:
bump dep and replace debuginspector with the one from inspectors


klkvr (approved):

reject-i:
@mattsse dup ? https://github.com/paradigmxyz/reth/pull/20009

Issue #19805, opened 2025-11-17
Add support for eth/70 eip-7542


mattsse:
https://eips.ethereum.org/EIPS/eip-7542 

this requires introducing a new eth70 variant and the protocol changes, new message variants,status

PR #20133, opened 2025-12-05
feat: add support for eth/70 eip-7542

0xKarl98:
This targets on issue #19805 , which i did below :

- Added eth/70 wire support: new StatusEth70 with block range, BlockRange type, and RequestBlockRange/SendBlockRange messages (IDs 0x0b/0x0c);

- Extended capability/version helpers to include eth/70 alongside eth/69; kept eth/69 compatibility tests and added eth/70 defaults.

- Handshake/status conversion now carries block range, validates eth/70 ranges and nonzero blockhash, and maps unified status to the proper message variant.

- eth/70 peers send SendBlockRange instead of BlockRangeUpdate, handle incoming range requests/responses, update remote range info, and throttle periodic range requests when info is stale

- Add eth/69-like tests 

cc @mattsse

### DIFF
--- a/crates/net/eth-wire-types/src/status.rs
+++ b/crates/net/eth-wire-types/src/status.rs
@@ -1,4 +1,4 @@
-use crate::EthVersion;
+use crate::{BlockRange, EthVersion};
 use alloy_chains::{Chain, NamedChain};
 use alloy_hardforks::{EthereumHardfork, ForkId, Head};
 use alloy_primitives::{hex, B256, U256};
@@ -9,11 +9,11 @@ use reth_codecs_derive::add_arbitrary_tests;
 
 /// `UnifiedStatus` is an internal superset of all ETH status fields for all `eth/` versions.
 ///
-/// This type can be converted into [`Status`] or [`StatusEth69`] depending on the version and
-/// unsupported fields are stripped out.
+/// This type can be converted into [`Status`], [`StatusEth69`], or [`StatusEth70`] depending on
+/// the negotiated version and unsupported fields are stripped out.
 #[derive(Clone, Debug, PartialEq, Eq, Copy)]
 pub struct UnifiedStatus {
-    /// The eth protocol version (e.g. eth/66 to eth/69).
+    /// The eth protocol version (e.g. eth/66 to eth/70).
     pub version: EthVersion,
     /// The chain ID identifying the peer’s network.
     pub chain: Chain,
@@ -29,6 +29,8 @@ pub struct UnifiedStatus {
     pub earliest_block: Option<u64>,
     /// The latest block number this node has (eth/69 only).
     pub latest_block: Option<u64>,
+    /// The block range this node can serve (eth/70).
+    pub block_range: Option<BlockRange>,
 }
 
 impl Default for UnifiedStatus {
@@ -45,6 +47,7 @@ impl Default for UnifiedStatus {
             total_difficulty: Some(U256::from(17_179_869_184u64)),
             earliest_block: Some(0),
             latest_block: Some(0),
+            block_range: None,
         }
     }
 }
@@ -68,6 +71,7 @@ impl UnifiedStatus {
             .total_difficulty(Some(head.total_difficulty))
             .earliest_block(Some(0))
             .latest_block(Some(head.number))
+            .block_range(Some(BlockRange { start_block: 0, end_block: head.number }))
             .build()
     }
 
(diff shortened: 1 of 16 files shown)


crates/net/eth-wire-types/src/status.rs
```diff
@@ -29,6 +29,8 @@ pub struct UnifiedStatus {
     pub earliest_block: Option<u64>,
     /// The latest block number this node has (eth/69 only).
     pub latest_block: Option<u64>,
+    /// The block range this node can serve (eth/70).
+    pub block_range: Option<BlockRange>,
```
mattsse, line 33: maybe we dont technically need this, but I think we can be exhaustive here and have all versioned fields, so let's keep this

crates/net/eth-wire-types/src/version.rs
```diff
@@ -27,14 +27,17 @@ pub enum EthVersion {
...
 
 impl EthVersion {
     /// The latest known eth version
-    pub const LATEST: Self = Self::Eth69;
+    pub const LATEST: Self = Self::Eth70;
 
     /// All known eth versions
-    pub const ALL_VERSIONS: &'static [Self] = &[Self::Eth69, Self::Eth68, Self::Eth67, Self::Eth66];
+    pub const ALL_VERSIONS: &'static [Self] =
+        &[Self::Eth70, Self::Eth69, Self::Eth68, Self::Eth67, Self::Eth66];
```
mattsse: I'd like to keep these unchanged for now, we can enable this once we're ready to enable full support for this

crates/net/eth-wire/src/handshake.rs
```diff
@@ -218,6 +218,19 @@ where
...
+                        return Err(EthHandshakeError::EarliestBlockGreaterThanLatestBlock {
+                            got: s.block_range.start_block,
+                            latest: s.block_range.end_block,
+                        }
+                        .into());
+                    }
+
+                    if s.blockhash.is_zero() {
+                        return Err(EthHandshakeError::BlockhashZero.into());
+                    }
```
mattsse, line 232: great

crates/net/network/src/session/active.rs
```diff
@@ -165,6 +173,7 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
     ///
     /// Returns an error if the message is considered to be in violation of the protocol.
     fn on_incoming_message(&mut self, msg: EthMessage<N>) -> OnIncomingMessageOutcome<N> {
+        let now = Instant::now();
```
mattsse: we can avoid this here and instead only instantiate this when we need to record this based on the message variant

crates/net/network/src/session/active.rs
```diff
@@ -746,14 +787,47 @@ impl<N: NetworkPrimitives> Future for ActiveSession<N> {
...
+                    };
+                    this.queued_outgoing.push_back(msg.into());
                     this.last_sent_latest_block = Some(current_latest);
                 }
             }
         }
 
+        if this.conn.version() >= EthVersion::Eth70 {
+            // If our view of the peer's range is stale, request an update (throttled).
+            let now = Instant::now();
```
mattsse: I think this is a bit too expensive,

can we combine this with this interval here:

https://github.com/paradigmxyz/reth/blob/190297083a3d82221c6ecb557d5a14026f719651/crates/net/network/src/session/active.rs#L737-L739

crates/net/network/src/session/mod.rs
```diff
@@ -538,13 +538,19 @@ impl<N: NetworkPrimitives> SessionManager<N> {
...
+                    });
+
                 // Configure the interval at which the range information is updated, starting with
                 // ETH69
                 let range_update_interval = (conn.version() >= EthVersion::Eth69).then(|| {
                     let mut interval = tokio::time::interval(RANGE_UPDATE_INTERVAL);
                     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
                     interval
                 });
+                let now = Instant::now();
```
mattsse: this can be moved into the closure, then we dont always call this

mattsse (changes requested):
this is awesome, only have a few requests

mattsse (changes requested):
I'm very sorry https://eips.ethereum.org/EIPS/eip-7542 was withdrawn and instead https://eips.ethereum.org/EIPS/eip-7975 will become eth70...

for this pr this means we need to remove the additional message variants and instead introduce a Receipts70 variant, similar to how receipts69 works I believe

PR #20134, opened 2025-12-05
chore: use drop guard for transact task

mattsse:
towards #20090

very basic and almost free check to skip (expensive) transact calls if the callsite (request) dropped/was terminated

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -25,7 +25,7 @@ use reth_evm::{
 };
 use reth_node_api::BlockBody;
 use reth_primitives_traits::Recovered;
-use reth_revm::{database::StateProviderDatabase, db::State};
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -535,10 +540,23 @@ pub trait Call:
...
+            let res = self
+                .spawn_with_call_at(request, at, overrides, move |db, evm_env, tx_env| {
+                    if cancel.is_cancelled() {
+                        // callsite dropped the guard
+                        return Err(EthApiError::InternalEthError.into())
+                    }
+                    this.transact(db, evm_env, tx_env)
+                })
+                .await;
+            drop(guard);
```
klkvr, line 557: will this not get dropped on its own once the future is cancelled?

klkvr (approved):

crates/rpc/rpc-eth-api/src/helpers/call.rs
mattsse: yeah just want to highlight this explicitly

PR #20136, opened 2025-12-05
feat(node): tx pool args defaults

### LINKED ISSUE #20098: Make TxPoolArgs defaults customizable
### Describe the feature

Similar to 

https://github.com/paradigmxyz/reth/issues/19609

but for 

https://github.com/paradigmxyz/reth/blob/8f5b4ae324d5c75cd866379dc083a0eb716ff617/crates/node/core/src/args/txpool.rs#L23

### Additional context

_No response_

shekhirin:
Closes https://github.com/paradigmxyz/reth/issues/20098

### DIFF
--- a/crates/node/core/src/args/txpool.rs
+++ b/crates/node/core/src/args/txpool.rs
@@ -34,14 +34,11 @@ pub struct DefaultTxPoolValues {
     queued_max_size: usize,
     blobpool_max_count: usize,
     blobpool_max_size: usize,
-    blob_cache_size: Option<u32>,
     disable_blobs_support: bool,
     max_account_slots: usize,
     price_bump: u128,
     minimal_protocol_basefee: u64,
-    minimum_priority_fee: Option<u128>,
     enforced_gas_limit: u64,
-    max_tx_gas_limit: Option<u64>,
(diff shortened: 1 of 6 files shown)


joshieDo (approved):

crates/node/core/src/args/txpool.rs
```diff
@@ -162,44 +403,73 @@ impl TxPoolArgs {
...
+            disable_transactions_backup,
+            max_batch_size,
         }
     }
 }
 
 impl RethTransactionPoolConfig for TxPoolArgs {
     /// Returns transaction pool configuration.
     fn pool_config(&self) -> PoolConfig {
+        let default_config = PoolConfig::default();
```
mattsse, line 472: why is this needed now?

mattsse (approved):
all of these lgtm, one q re what seems like an unrelated change

crates/node/core/src/args/txpool.rs
shekhirin: drive-by to explicitly set all fields of `PoolConfig` to not forget to adjust tx pool args when we add new fields

PR #20138, opened 2025-12-05
feat(txpool): handle more simulated scenarios in test_utils/pool.rs

figtracer:
- implements `ScenarioType::HigherNonce` with unit test
- implements `ScenarioType::BelowBaseFee` with unit test
- previously `MockPool::default()` created a pool with `pending_basefee = MIN_PROTOCOL_BASE_FEE` and it wasn't matching the simulation's `base_fee` so also added `create_pool()` to align simulation's `base_fee` with pool's `pending_basefee`. 
- implements `ScenarioType::FillNonceGap` with unit test
- previously we weren't using the generated `sender`, now we use `sender` in `with_sender`

### DIFF
--- a/crates/transaction-pool/src/test_utils/pool.rs
+++ b/crates/transaction-pool/src/test_utils/pool.rs
@@ -408,44 +408,6 @@ mod tests {
 
     #[test]
     fn test_many_random_scenarios() {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
cool, yep, this lgtm

mattsse (approved):
cool

PR #20139, opened 2025-12-05
feat: add `--max-peers` CLI flag

gr8h:
This PR adds `--max-peers` CLI flag for configuring total peer limits splitting into inbound and outbound connections.

## Changes

- Added `--max-peers` flag that splits total peers into ~2:1 inbound:outbound
- Mutually exclusive with `--max-outbound-peers` and `--max-inbound-peers`

Issue: [6100](https://github.com/paradigmxyz/reth/issues/6100)


mattsse (approved):
this does indeed seem useful

th 2:1 ratio makes sense, we could make the ratio configurable as well but I thinkt his should do

gr8h:
@mattsse feel free to ping me on other issues

PR #20142, opened 2025-12-05
feat(node): tx pool optional args defaults

shekhirin:
Follow-up to https://github.com/paradigmxyz/reth/pull/20136 for optional arguments

### DIFF
--- a/crates/node/core/src/args/txpool.rs
+++ b/crates/node/core/src/args/txpool.rs
@@ -552,6 +552,7 @@ impl RethTransactionPoolConfig for TxPoolArgs {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloy_primitives::address;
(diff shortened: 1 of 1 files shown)


crates/node/core/src/args/txpool.rs
```diff
@@ -3,7 +3,7 @@
 use crate::cli::config::RethTransactionPoolConfig;
 use alloy_eips::eip1559::{ETHEREUM_BLOCK_GAS_LIMIT_30M, MIN_PROTOCOL_BASE_FEE};
 use alloy_primitives::Address;
-use clap::Args;
+use clap::{builder::Resettable, Args};
```
mattsse, line 6: TIL

mattsse (approved):
lgtm,

pending some parse command tests

PR #20143, opened 2025-12-05
feat(engine): cli argument to disable state cache

shekhirin:
Adds an ability to disable cached state provider by passing `--engine.disable-state-cache` CLI argument.


crates/engine/primitives/src/config.rs
```diff
@@ -363,6 +373,12 @@ impl TreeConfig {
         self
     }
 
+    /// Setter for whether to disable state cache.
+    pub const fn without_state_cache(mut self, disable_state_cache: bool) -> Self {
```
joshieDo, line 379: _without_ is weird to me but i see its used by others as well

joshieDo (approved):

mattsse (commented):
doesnt this still handle cache updates after execution?

Rjected (approved):
lgtm

yongkangc (approved):
Nice

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -255,31 +255,35 @@ where
...
-        });
+                // Replace the shared cache with the new one; the previous cache (if any) is
+                // dropped.
+                *cached = Some(new_cache);
+            });
 
-        let elapsed = start.elapsed();
-        debug!(target: "engine::caching", parent_hash=?hash, elapsed=?elapsed, "Updated execution cache");
+            let elapsed = start.elapsed();
+            debug!(target: "engine::caching", parent_hash=?hash, elapsed=?elapsed, "Updated execution cache");
```
mattsse, line 283: these are now misleading if there's no cache

mattsse (commented):
this feels a bit messy

how does prewarm behave if caching is disabled?

should we even do prewarm if we don't make use of the shared cache or only for the proof prep?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
shekhirin: it's gated behind `if let Some(saved_cache) = saved_cache` which is triggered only if we use state cache

mattsse (approved):

PR #20144, opened 2025-12-05
feat: add support for debug_getBadblock

0xKarl98:
This targets on issue #20123 , as below :

- Add a shared BadBlockStore (bounded, deduped, newest-first) and wire it into DebugApi
-  Implement debug_getBadBlocks to expose recent invalid blocks with full transactions, hash, and RLP
- Listen for ConsensusEngineEvent::InvalidBlock, recover the sealed block with senders, and cache it for the debug API

cc @mattsse  @klkvr @Rjected

### DIFF
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -10,6 +10,7 @@ use crate::{
 };
 use alloy_rpc_types::engine::ClientVersionV1;
 use alloy_rpc_types_engine::ExecutionData;
+use futures::StreamExt;
 use jsonrpsee::{core::middleware::layer::Either, RpcModule};
 use reth_chain_state::CanonStateSubscriptions;
 use reth_chainspec::{ChainSpecProvider, EthChainSpec, EthereumHardforks, Hardforks};
@@ -23,6 +24,7 @@ use reth_node_core::{
     version::{version_metadata, CLIENT_CODE},
 };
 use reth_payload_builder::{PayloadBuilderHandle, PayloadStore};
+use reth_primitives_traits::RecoveredBlock;
 use reth_rpc::{
     eth::{core::EthRpcConverterFor, DevSigner, EthApiTypes, FullEthApiServer},
     AdminApi,
@@ -1017,6 +1019,23 @@ where
             let signers = DevSigner::from_mnemonic(config.dev.dev_mnemonic.as_str(), 20);
             registry.eth_api().signers().write().extend(signers);
         }
+
+        // keep track of invalid blocks for `debug_getBadBlocks`
+        let bad_block_store = registry.bad_block_store().clone();
+        let mut engine_events_stream = engine_events.new_listener();
+        node.task_executor().spawn_critical(
+            "collect bad blocks",
+            Box::pin(async move {
+                while let Some(event) = engine_events_stream.next().await {
+                    if let ConsensusEngineEvent::InvalidBlock(block) = event
+                        && let Ok(recovered) =
+                            RecoveredBlock::try_recover_sealed(block.as_ref().clone())
+                    {
+                        bad_block_store.insert(recovered);
+                    }
+                }
+            }),
+        );
 
         let mut registry = RpcRegistry { registry };
         let ctx = RpcContext {
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -1027,8 +1027,8 @@ where
             "collect bad blocks",
             Box::pin(async move {
                 while let Some(event) = engine_events_stream.next().await {
-                    if let ConsensusEngineEvent::InvalidBlock(block) = event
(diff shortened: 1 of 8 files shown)


crates/rpc/rpc-api/src/debug.rs
```diff
@@ -38,7 +38,7 @@ pub trait DebugApi<TxReq: RpcObject> {
 
     /// Returns an array of recent bad blocks that the client has seen on the network.
     #[method(name = "getBadBlocks")]
-    async fn bad_blocks(&self) -> RpcResult<Vec<Block>>;
+    async fn bad_blocks(&self) -> RpcResult<Vec<BadBlock>>;
```
mattsse, line 41: I think we can actually make this just serde_json::Value, seems easier or we use the actual rpc type see also 

https://github.com/paradigmxyz/reth/blob/190297083a3d82221c6ecb557d5a14026f719651/crates/rpc/rpc-eth-api/src/core.rs#L90-L91

crates/rpc/rpc/src/debug.rs
```diff
@@ -1045,21 +1131,22 @@ where
...
     }
 }
 
-struct DebugApiInner<Eth> {
+struct DebugApiInner<Eth: RpcNodeCore> {
     /// The implementation of `eth` API
     eth_api: Eth,
     // restrict the number of concurrent calls to blocking calls
     blocking_task_guard: BlockingTaskGuard,
+    bad_block_store: BadBlockStore<ProviderBlock<Eth::Provider>>,
```
mattsse, line 1151: I see we need the types for this

I guess an alternative here would be to just use `EthApiTypes` here

crates/rpc/rpc/src/debug.rs
```diff
@@ -37,36 +38,86 @@ use reth_tasks::pool::BlockingTaskGuard;
...
+
+    /// Inserts a recovered block, keeping only the most recent `limit` entries and deduplicating
+    /// by block hash.
+    pub fn insert(&self, block: RecoveredBlock<B>) {
+        let hash = block.hash();
+        let mut guard = self.inner.write();
+
+        // drop existing entry with same hash to keep most recent order
+        guard.retain(|b| b.hash() != hash);
+        guard.push_back(Arc::new(block));
```
mattsse, line 73: instead of removing, I think we should just check if we already have it

because it's common that we see bad blocks a few times, and this would then mess up the ordering

crates/rpc/rpc/src/debug.rs
```diff
@@ -37,36 +38,86 @@ use reth_tasks::pool::BlockingTaskGuard;
...
 /// This type provides the functionality for handling `debug` related requests.
-pub struct DebugApi<Eth> {
+pub struct DebugApi<Eth: RpcNodeCore> {
     inner: Arc<DebugApiInner<Eth>>,
 }
 
-// === impl DebugApi ===
+/// A bounded, deduplicating store of recently observed bad blocks.
+#[derive(Clone, Debug)]
+pub struct BadBlockStore<B: BlockTrait> {
```
mattsse, line 54: this type def is out of order, we should move this to after all the impl blocks

crates/node/builder/src/rpc.rs
```diff
@@ -1018,6 +1020,23 @@ where
             registry.eth_api().signers().write().extend(signers);
         }
 
+        // keep track of invalid blocks for `debug_getBadBlocks`
+        let bad_block_store = registry.bad_block_store().clone();
+        let mut engine_events_stream = engine_events.new_listener();
+        node.task_executor().spawn_critical(
```
mattsse, line 1026: this is kinda redundant if we dont have the debug_ endpoint installed because then the blockstore wont be useable, but this doesnt really hurt anyway

mattsse, line 1026: we can spawn this as regular task imo

mattsse (changes requested):
cool, a few suggestions

paging @klkvr re type generics

crates/rpc/rpc-api/src/debug.rs
0xKarl98: Yeah think we need to unify the block def in rpc here

crates/rpc/rpc/src/debug.rs
0xKarl98: Yeah def should do that , skip the duplicated bad block hash

crates/rpc/rpc/src/debug.rs
0xKarl98: Yeah will look more ordered

crates/node/builder/src/rpc.rs
0xKarl98: Have set a flag to enable blockstore  only when debug_ endpoint installed

0xKarl98:
Just closes this and open another with more proper name debug_getBadBlocks

PR #20146, opened 2025-12-05
refactor(net): remove unused enable_packet_filter and ban_duration from config

Bashmunta:
Removed enable_packet_filter and ban_duration from Discv4Config and their builder methods. These fields were not used by the discv4 service: inbound packet filtering is always applied in receive_loop, and ban duration is managed at the peers manager layer, not in discv4. Updated defaults and tests accordingly.


mattsse (approved):
reasonable, we dont need banning on this level really

PR #20148, opened 2025-12-05
perf: parallelize HashedPostStateSorted::from_reverts hashing/sorting

### LINKED ISSUE #20049: perf: parallelize HashedPostStateSorted::from_reverts hashing/sorting
## Summary
Explore a perf experiment for `HashedPostStateSorted::from_reverts` to parallelize the hashing/sorting step (not the DB cursor walks) so we can speed up large revert ranges without extra allocations.

## Motivation
- PR #20047 added the sorted-from-reverts path; benchmarks show ~6% win vs unsorted+sort on a small synthetic dataset (256 accounts x 4 slots). For larger reorg ranges, hashing/sorting could dominate.
- Tracker: #17920 (perf experiments).

## Proposal
- Keep DB cursor iteration sequential (required), but parallelize the post-collection hashing/sorting with rayon behind an opt-in feature flag (e.g. `parallel-from-reverts`).
- Use thresholds to avoid rayon overhead on small inputs (e.g. accounts > 1k and per-account slots > 32 use `par_sort_unstable_by_key`, else stay sequential).
- Hash keys during collection to avoid an extra map pass if it helps CPU profiles.

## Acceptance Criteria
- Implement gated parallel path with sensible thresholds and benchmarks comparing sequential vs parallel on large synthetic revert sets (e.g. 10k+ accounts, skewed slot distributions).
- Measure CPU and wall-time deltas; document when parallel path wins/loses.
- Keep default behavior unchanged (feature flag off) to avoid footprint changes.

## Risks / Notes
- rayon dependency and build footprint (hence feature flag).
- Parallel sorting only helps when key counts are large; overhead may regress small cases.

Andrurachi:
## Summary
Introduces a parallelized path for `HashedPostStateSorted::from_reverts` using `rayon`, gated behind a new `parallel-from-reverts` feature flag.

This optimization targets CPU bottlenecks caused by deep reorgs or blocks with skewed storage distributions. It implements an Account Count Threshold (2,500) to ensure no regressions on small or standard blocks.

## Motivation
Closes #20049

Processing large state reverts involves two distinct phases:
1.  **DB Cursor Walk (I/O Bound):** Sequential reading of account/storage changesets.
2.  **Sorting & Hashing (CPU Bound):** Organizing the raw data into a sorted state.

While the DB walk must remain sequential, the sorting phase becomes a bottleneck when thousands of accounts (or accounts with massive storage slots) need processing. This PR parallelizes the CPU-bound sorting phase, reducing wall-time for heavy blocks.

## Implementation Details

1.  **Feature Flag:** Added `parallel-from-reverts` (opt-in).
2.  **Strategy:** Parallelize the **outer loop** (iterating over accounts) rather than the inner loop (sorting slots). Benchmarks proved that inner-loop parallelization (`par_sort_unstable`) introduced too much overhead for typical slot counts (< 1,000).
3.  **Threshold:** A threshold of **2,500 Accounts** was established via micro-benchmarking.
    * `< 2,500`: Always Sequential (avoids Rayon overhead).
    * `>= 2,500`: Parallel (distributes load, handles skewed accounts).
4.  **Logic Preservation:** Preserves the existing eager hashing logic (hashing keys during collection) to ensure cache locality and consistency with the sequential path.

### 1. Micro-Benchmarks (`sorting_par_exp`)
Measured purely the sorting/hashing overhead (excluding DB reads).

| Scenario | Accounts | Slots/Acc | Seq Time | Par Time | Speedup |
| :--- | :--- | :--- | :--- | :--- | :--- |
| `Acc_Low` | 100 | 4 | 8 µs | 124 µs | -15x |
| `Acc_Med` | 1,000 | 4 | 73 µs | 122 µs | -0.6x |
| `Acc_High` | 10,000 | 4 | 790 µs | 612 µs | **1.3x** |
| `Skewed_10k`| 10k | Mixed | 64 ms | 14 ms | **4.6x** |

Skewed Distribution: 95% accounts have 4 slots, 5% have 2,000 slots.*

### 2. Integration Benchmarks (`integration_bench`)
Measured full lifecycle: DB Read -> Hashing -> Sorting -> Allocation.

| Scenario | Description | Seq Time | Par Time | Delta |
| :--- | :--- | :--- | :--- | :--- |
| `Small` | 1k Acc (Below Threshold) | 61.2 ms | 61.0 ms | **0%** (No Regression) |
| `Large_Uniform` | 10k Acc (20 slots) | 270 ms | 267 ms | **~1.2%** |
| `Large_Skewed` | 10k Acc (Whale Dist.) | 1,439 ms | 1,363 ms | **~5.3%** |

Since the total runtime is dominated by DB I/O, this actually represents a solid optimization of the available CPU-bound work. The threshold ensures no regression for small blocks.

## Checklist
- [x] Added `rayon` dependency (optional).
- [x] Implemented `parallel-from-reverts` feature flag.
- [x] Added benchmarks covering Uniform, Skewed, and Threshold scenarios.
- [x] Verified no regressions for inputs below the 2,500 account threshold.


crates/trie/db/src/state.rs
```diff
@@ -280,14 +283,37 @@ impl<TX: DbTx> DatabaseHashedPostState<TX> for HashedPostStateSorted {
...
-            .collect();
+        // Threshold based on benchmark
+        const PARALLEL_THRESHOLD: usize = 2_500;
+
+        // Check Feature Flag AND Threshold
+        let use_parallel =
+            cfg!(feature = "parallel-from-reverts") && storages.len() >= PARALLEL_THRESHOLD;
+
+        let hashed_storages = if use_parallel {
+            #[cfg(feature = "parallel-from-reverts")]
```
mediocregopher: In other places we feature-flag rayon usage behind `std`, imo we could do that here and not introduce a new feature flag.

https://github.com/paradigmxyz/reth/blob/245cca7ce2cbc4844bcc6610174425cac1dca455/crates/trie/sparse/src/state.rs#L264-L276

mediocregopher:
@yongkangc let's rbc this as well to confirm there's no regression :pray:

Andrurachi:
@yongkangc I've rebased on main to bring the breaking changes in MerkleChangeSets. The PR is compiling and benchmarks are passing. I'm using `cargo bench -p reth-trie-db --bench integration_bench` and `cargo bench -p reth-trie-db --bench integration_bench --features parallel-from-reverts` to run benchmarks.

Regarding moving the feature flag to std: I tried replacing parallel-from-reverts with the standard default = ["std"] pattern, but I hit a wall of Zepter errors regarding feature propagation.

To avoid pushing broken config or making you wait more than needed, I've kept parallel-from-reverts for this push. If required, could you please provide some advise on the correct Cargo.toml setup to satisfy Zepter for this specific crate?

yongkangc:
Hey @Andrurachi thanks for clarifying, I think that they should be removed from the merge so we don't have to maintain them

The results from this pr is good enough!

yongkangc:
Hey @Andrurachi thanks for clarifying, I think that they should be removed from the merge so we don't have to maintain them

The results from this pr is good enough!

Andrurachi:
Got it, I’ve removed the benchmarks and moved the parallelized path behind `std`.  
Thanks for the feedback!

Also, I really enjoyed working on this perf issue. If any similar performance-related tasks come up in the future, feel free to ping me. Happy to help :)

mediocregopher (approved):
LGTM 👍

yongkangc:
@mediocregopher currently running benchmarks on this for base (r0) and mainnet (r4)

yongkangc:
@Andrurachi current thoughts right now is that the pr probably has scheduling overhead from parallelization. The tradeoff in complexity right now might not be worth the result.
However there could be more investigation to find the bottleneck to speed it up better. That requires more work in profiling and tracing to find out the bottleneck 

thoughts? @mediocregopher @mattsse

Andrurachi:
@yongkangc Thanks for running these!

Interpreting the Mainnet data, the ~5% mean regression confirms the overhead on standard blocks, but the -32% p99 latency drop suggests the parallel path is a massive win for heavy blocks. It seems I'm just catching too many "light" blocks in the net.

Do you think it's worth bumping the threshold significantly (e.g., 10k or 15k)? That should fix the average while preserving the p99 speedup. Or would a threshold that high be unrealistic to hit in practice and that's why the complexity might not be worth the result?

Happy to tune and push if that logic tracks!

yongkangc:
Hey @Andrurachi 

Imo we would have to play with different threshold and do some testing. Bumping it to a higher threshold could work for improving a mean. 

Unsure of what might be a good number right jow but would wanna try 5, 10, 15k to see how it changes on mainnet.

Andrurachi:
Awesome. I've updated the PR with a threshold of 10,000 as a starting point.

Since I can't run the mainnet benches myself, I'll leave it there for you to test. If you want to see 5k or 15k next, just let me know and I can bump the number and push again.

Ready when you are!

PR #20155, opened 2025-12-06
fix(node): Added missing trait bound on RethRpcAddOns for EthereumAddOns

mackcee:
### Summary
  - add the missing `RpcMiddleware: RethRpcMiddleware` bound on `impl RethRpcAddOns for EthereumAddOns` so `with_rpc_middleware` works with custom middleware types
  - no behavior changes - fixes a compile-time regression when using non-Identity middleware
  
 ### Future Work
  - plan to add example code of launching a node with custom rpc middleware

### DIFF
--- a/crates/ethereum/node/src/node.rs
+++ b/crates/ethereum/node/src/node.rs
@@ -319,7 +319,8 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah, ty!

PR #20158, opened 2025-12-06
feat: add merge_if_module_configured_with that accepts a closure

### LINKED ISSUE #20156: Add merge_if_module_configured_with that accepts a closure
### Describe the feature

this function might not even need the given methods so it'd be great to have another fn that accepts a closure so we can avoid always instantiating the value

https://github.com/paradigmxyz/reth/blob/190297083a3d82221c6ecb557d5a14026f719651/crates/rpc/rpc-builder/src/lib.rs#L1649-L1653

### Additional context

_No response_

stevencartavia:
closes #20156

### DIFF
--- a/crates/rpc/rpc-builder/src/lib.rs
+++ b/crates/rpc/rpc-builder/src/lib.rs
@@ -1665,6 +1665,42 @@ impl TransportRpcModules {
         Ok(())
     }
 
+    /// Merge the given [`Methods`] in all configured transport modules if the given
+    /// [`RethRpcModule`] is configured for the transport, using a closure to lazily
+    /// create the methods only when needed.
+    ///
+    /// The closure is only called if at least one transport has the module configured.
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-builder/src/lib.rs
```diff
@@ -1665,6 +1665,42 @@ impl TransportRpcModules {
...
+
+        // Apply to configured transports
+        if self.module_config().contains_http(&module) {
+            self.merge_http(methods.clone())?;
+        }
+        if self.module_config().contains_ws(&module) {
+            self.merge_ws(methods.clone())?;
+        }
+        if self.module_config().contains_ipc(&module) {
+            self.merge_ipc(methods)?;
```
mattsse: this is basically just merge_configured_if_module now, so we can call this

mattsse (changes requested):

mattsse (approved):

PR #20159, opened 2025-12-06
feat: add example for launching a node with custom rpc middleware

mackcee:
depends on #20155

### Summary
- adds example for launching a node with custom rpc middleware that modifies the returned error message if the rpc call results in an error.


mattsse (approved):
this is useful indeed

PR #14308, opened 2025-02-07
feat(l2-withdrawals): consensus rules

emhane:
Checks out consensus rules from https://github.com/paradigmxyz/reth/pull/14124

PR #14209, opened 2025-02-04
feat(isthmus): withdrawals root in block building

emhane:
Closes https://github.com/paradigmxyz/reth/issues/14176

Ref https://github.com/ethereum-optimism/op-reth/pull/60

PR #20160, opened 2025-12-06
chore(optimism): deduplicate ADDRESS_L2_TO_L1_MESSAGE_PASSER

zeroprooff:
Found that `ADDRESS_L2_TO_L1_MESSAGE_PASSER` is defined in both `reth-optimism-primitives` and `reth-optimism-consensus`.

Looking at git history, the constant was first added to `primitives/predeploys.rs` in #14209, then accidentally copied to `consensus/validation/isthmus.rs` in #14308 

Since `reth-optimism-consensus` already depends on `reth-optimism-primitives`, there's no reason to duplicate it. Replaced the local definition with an import.

### DIFF
--- a/crates/optimism/consensus/src/validation/isthmus.rs
+++ b/crates/optimism/consensus/src/validation/isthmus.rs
@@ -2,16 +2,13 @@
 
 use crate::OpConsensusError;
 use alloy_consensus::BlockHeader;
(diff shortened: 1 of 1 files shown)


zeroprooff:
btw noticed there's a todo to move this to op-alloy - happy to take that on if useful!

mattsse (approved):
makes sense

PR #20162, opened 2025-12-06
docs: add architecture diagram for EVM node component

sashass1315:
Fill the previously empty Architecture section of the EVM node component docs with a mermaid diagram.

The diagram documents the main EVM building blocks (configuration, environment, block executor, block builder, block assembler and state database) and how they interact, matching the style used by other node-components docs (network, pool). This makes the EVM documentation more complete and easier to understand


mattsse (approved):
seems useful

PR #20164, opened 2025-12-06
fix: make inserted blocks part of fcu canonical

mattsse:
this will mitigate suboptimal behaviour when we're trying to close the gap after transitioning from backfill to live sync.

previously this would only attempt to make the chain canonical if we have `[canonicalhead +1,..fcu.head]` executed, which can be over 2 epochs of blocks and be an issue performance wise.

this will now attempt to also make the finalized block, safe block canonical, because we can assume those are definitely part of the canonical chain.

this should restrict it to 32 blocks at most


yongkangc (approved):
makes sense i see, this allows us to save sooner as we know that they are canon and we can save it.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2348,11 +2361,15 @@ where
         // try to append the block
         match self.insert_block(block) {
             Ok(InsertPayloadOk::Inserted(BlockStatus::Valid)) => {
-                if self.is_sync_target_head(block_num_hash.hash) {
-                    trace!(target: "engine::tree", "appended downloaded sync target block");
+                // check if we just inserted a block that's part of sync targets,
+                // head,safe,finalized
```
shekhirin:
```suggestion
                // check if we just inserted a block that's part of sync targets,
                // i.e. head, safe, or finalized
```

shekhirin (approved):

PR #20167, opened 2025-12-06
feat: add --datadir.min-free-disk  flag

cuiweixie:
add flag

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9191,6 +9191,7 @@ dependencies = [
  "serde",
  "shellexpand",
  "strum 0.27.2",
+ "sysinfo",
  "thiserror 2.0.17",
  "tokio",
  "toml",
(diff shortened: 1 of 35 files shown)


cuiweixie:
@mattsse please help me to solve the error.
in https://github.com/paradigmxyz/reth/actions/runs/19991030410/job/57331770979?pr=20167
```
  warning: librocksdb-sys@0.17.3+10.4.2: ar: /home/runner/work/reth/reth/target/debug/build/librocksdb-sys-a88a624b72dde647/out/librocksdb.a: error reading /home/runner/work/reth/reth/target/debug/build/librocksdb-sys-a88a624b72dde647/out/d11b32dac0d420e2-memory_util.o: No space left on device
  error: failed to run custom build command for `librocksdb-sys v0.17.3+10.4.2`
```
I think It's not caused by this pr.

mattsse (commented):
unsure if we need/want this

thoughts @shekhirin

mattsse (commented):
after thinking about this, i don't think this is something we need to build into the client itself, this can also be monitored separately

PR #20168, opened 2025-12-06
feat: add --fdlimit flag

cuiweixie:
add flag

### DIFF
--- a/crates/cli/commands/src/node.rs
+++ b/crates/cli/commands/src/node.rs
@@ -66,6 +66,13 @@ pub struct NodeCommand<C: ChainSpecParser, Ext: clap::Args + fmt::Debug = NoArgs
     #[arg(long, conflicts_with = "instance", global = true)]
     pub with_unused_ports: bool,
 
+    /// Raise the open file descriptor resource limit.
+    ///
+    /// Default: 0 (system default). When set to 0, uses the system's default file descriptor
+    /// limit. When set to a positive value, attempts to raise the limit to that value.
+    #[arg(long, value_name = "LIMIT", default_value_t = 0, global = true)]
+    pub fdlimit: u64,
+
     /// All datadir related arguments
     #[command(flatten)]
     pub datadir: DatadirArgs,
@@ -158,6 +165,7 @@ where
             metrics,
             instance,
             with_unused_ports,
+            fdlimit,
             network,
             rpc,
             txpool,
@@ -179,6 +187,7 @@ where
             chain,
             metrics,
             instance,
+            fdlimit,
             network,
             rpc,
             txpool,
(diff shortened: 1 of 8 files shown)


crates/node/builder/src/launch/common.rs
```diff
@@ -212,15 +212,51 @@ impl LaunchContext {
...
-    pub fn configure_globals(&self, reserved_cpu_cores: usize) {
+    pub fn configure_globals(&self, reserved_cpu_cores: usize, fdlimit: u64) {
         // Raise the fd limit of the process.
         // Does not do anything on windows.
-        match fdlimit::raise_fd_limit() {
-            Ok(fdlimit::Outcome::LimitRaised { from, to }) => {
-                debug!(from, to, "Raised file descriptor limit");
+        if fdlimit == 0 {
+            // Use system default (raise to maximum)
+            match fdlimit::raise_fd_limit() {
```
mattsse, line 220: this will alter current behaviour and will likely be super annoying

crates/cli/commands/src/node.rs
```diff
@@ -66,6 +66,13 @@ pub struct NodeCommand<C: ChainSpecParser, Ext: clap::Args + fmt::Debug = NoArgs
     #[arg(long, conflicts_with = "instance", global = true)]
     pub with_unused_ports: bool,
 
+    /// Raise the open file descriptor resource limit.
+    ///
+    /// Default: 0 (system default). When set to 0, uses the system's default file descriptor
+    /// limit. When set to a positive value, attempts to raise the limit to that value.
+    #[arg(long, value_name = "LIMIT", default_value_t = 0, global = true)]
+    pub fdlimit: u64,
```
mattsse, line 74: not sure we really need this

imo always raising is appropriate

mattsse (commented):
hmm, undecided about this

crates/node/builder/src/launch/common.rs
cuiweixie: 0 for old behavior may be okay.

crates/cli/commands/src/node.rs
cuiweixie: Even if the system allows setting a higher file descriptor limit, the actual availability of resources such as memory and CPU can impact the stability and performance of the program. Setting the file descriptor limit too high can lead to resource exhaustion, which may cause the system or application to become unstable. This is because every open file descriptor consumes system resources, and if the number of file descriptors is increased significantly, it can strain the system's capacity to manage them effectively. Therefore, it's important to consider the overall resource usage and system capabilities when adjusting file descriptor limits.

mattsse (commented):
this is kinda hard to get right, so we can rather expect the user to do this

PR #20169, opened 2025-12-06
feat: parallelize recovery

klkvr:


crates/evm/evm/src/engine.rs
```diff
@@ -18,22 +18,48 @@ pub trait ConfigureEngineEvm<ExecutionData>: ConfigureEvm {
...
-/// Iterator over executable transactions.
-pub trait ExecutableTxIterator<Evm: ConfigureEvm>:
-    Iterator<Item = Result<Self::Tx, Self::Error>> + Send + 'static
-{
+/// A helper trait representing a pair of a "raw" transactions iterator and a closure that can be
+/// used to convert them to an executable transaction. This tuple is used in the engine to
+/// parallelize heavy work like decoding or recovery.
+pub trait ExecutableTxTuple: Into<(Self::Iter, Self::Convert)> + Send + 'static {
+    /// Raw transaction that can be converted to an [`ExecutableTxIterator::Tx`]
+    type RawTx: Send + Sync + 'static;
```
mattsse: I think we should highlight here that this can be anything as long as the the converter can map this into the Tx type

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -312,21 +314,49 @@ where
...
+                    next_for_execution += 1;
+
+                    while queue.first_key_value().is_some_and(|(idx, _)| *idx == next_for_execution)
+                    {
+                        let _ = execute_tx.send(queue.pop_first().unwrap().1);
+                        next_for_execution += 1;
+                    }
+                } else {
+                    queue.insert(idx, tx);
+                }
```
mattsse: I'd like to replace this with while over `first_entry`

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -213,16 +213,36 @@ where
...
+
+                Ok((
+                    iter,
+                    Box::new(convert)
+                        as Box<
+                            dyn Fn(Either<_, _>) -> Result<Either<_, _>, _> + Send + Sync + 'static,
+                        >,
+                ))
+            }
             BlockOrPayload::Block(block) => {
```
mattsse, line 232: since this step looks slightly confusing, I think some additional docs would be helpful here

mattsse (changes requested):
this makes a lot of sense and is way easier than what I previously attempted

smol suggestion and a doc nit

we might also want to take the avail cores into account here but unsure because rayon already handles the bridging based on available threads

mattsse (approved):
lgtm!

shekhirin (approved):

DaniPopes (approved):

yongkangc (approved):

PR #20171, opened 2025-12-06
refactor(rpc): remove dead got_notif flag from RpcService batch handler

radik878:
The got_notif flag in RpcService::batch was written but never read, so it effectively had no impact on the batch response. Batch notification semantics for IPC are already correctly handled in ipc::process_batch_request, which does not rely on this flag.

This change removes the unused variable and simplifies the Notification match armandling semantics intact.


mattsse (approved):
reasonable

PR #20172, opened 2025-12-07
feat: add --http.vhosts and --ws.vhosts flags

cuiweixie:


crates/node/core/src/args/rpc_server.rs
```diff
@@ -67,6 +67,11 @@ pub struct RpcServerArgs {
     #[arg(long = "http.corsdomain")]
     pub http_corsdomain: Option<String>,
 
+    /// Comma separated list of virtual hostnames from which to accept HTTP requests (server
+    /// enforced). Accepts '*' wildcard.
+    #[arg(long = "http.vhosts", default_value = "localhost")]
+    pub http_vhosts: String,
```
mattsse: can we make these optional?

crates/rpc/rpc-layer/src/vhost_layer.rs
```diff
@@ -0,0 +1,518 @@
...
+};
+use tower::{Layer, Service};
+
+/// A trait for validating virtual hostnames from the Host header.
+pub trait VHostValidator {
+    /// Validates the Host header value against allowed virtual hostnames.
+    /// Returns `Ok(())` if the host is allowed, or an error response if not.
+    #[expect(clippy::result_large_err)]
+    fn validate(&self, host: Option<&str>) -> Result<(), HttpResponse>;
+}
```
mattsse: I think we can do without the trait for now?

crates/rpc/rpc-layer/src/vhost_layer.rs
```diff
@@ -0,0 +1,518 @@
...
+/// A validator that checks Host headers against a list of allowed virtual hostnames.
+///
+/// Supports wildcard matching with '*' character. For example:
+/// - `"*"` matches all hosts
+/// - `"*.example.com"` matches any subdomain of example.com
+/// - `"localhost"` matches exactly "localhost"
+#[derive(Clone, Debug)]
+pub struct AllowedVHosts {
+    /// List of allowed virtual hostname patterns
+    patterns: Vec<String>,
```
mattsse: this one should be Arc so it's cheaper to clone

crates/rpc/rpc-layer/src/vhost_layer.rs
```diff
@@ -0,0 +1,518 @@
...
+
+impl<S, V> Service<HttpRequest> for VHostService<S, V>
+where
+    S: Service<HttpRequest, Response = HttpResponse>,
+    V: VHostValidator,
+    Self: Clone,
+{
+    type Response = HttpResponse;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future>;
```
mattsse: I think for this we can use a boxed future and dont need a dedicated type?

crates/rpc/rpc-layer/src/vhost_layer.rs
mattsse: should this be a hashset perhaps?

mattsse (changes requested):
not opposed to this, but have some suggestions

crates/node/core/src/args/rpc_server.rs
cuiweixie: Done

crates/rpc/rpc-layer/src/vhost_layer.rs
cuiweixie: Done

crates/rpc/rpc-layer/src/vhost_layer.rs
cuiweixie: I have change to a new version, but not sure whether it is as your expect.

crates/rpc/rpc-layer/src/vhost_layer.rs
cuiweixie: Done.

mattsse (commented):
I dont think this does anything/is useful so dont want to bloat the feature set

PR #20173, opened 2025-12-07
refactor: to use map_or/is_none_or/is_some_and instead of map().unwrap_or()

cuiweixie:


mattsse (commented):
I personally find map_or harder to read than unwrap_or

PR #20175, opened 2025-12-07
refactor: map_or(Some()...) to Some(map_or())

cuiweixie:


mattsse (commented):
since this is the same, I don't think we need to change this

PR #20176, opened 2025-12-07
refactor: simplify code

cuiweixie:


mattsse (commented):
in this case I prefer the existing version because it's easier to parse

PR #20177, opened 2025-12-07
feat: add support for debug_getBadBlock

0xKarl98:
This closes issue #20123  , as below :

Add a shared BadBlockStore (bounded, deduped, newest-first) and wire it into DebugApi
Implement debug_getBadBlocks to expose recent invalid blocks with full transactions, hash, and RLP
Listen for ConsensusEngineEvent::InvalidBlock, recover the sealed block with senders, and cache it for the debug API
cc @mattsse @klkvr @Rjected

### DIFF
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -26,7 +26,7 @@ use reth_node_core::{
 use reth_payload_builder::{PayloadBuilderHandle, PayloadStore};
 use reth_primitives_traits::RecoveredBlock;
 use reth_rpc::{
-    eth::{core::EthRpcConverterFor, DevSigner, EthApiTypes, FullEthApiServer},
+    eth::{core::EthRpcConverterFor, DevSigner, EthApiTypes, FullEthApiServer, RpcNodeCore},
     AdminApi,
 };
 use reth_rpc_api::{eth::helpers::EthTransactions, IntoEngineApiRpcModule};
@@ -1208,6 +1208,7 @@ impl<'a, N: FullNodeComponents<Types: NodeTypes<ChainSpec: Hardforks + EthereumH
 pub trait EthApiBuilder<N: FullNodeComponents>: Default + Send + 'static {
     /// The Ethapi implementation this builder will build.
     type EthApi: EthApiTypes
+        + RpcNodeCore<Primitives = PrimitivesTy<N::Types>>
         + FullEthApiServer<Provider = N::Provider, Pool = N::Pool>
         + Unpin
         + 'static;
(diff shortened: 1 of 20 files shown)


crates/node/builder/src/rpc.rs
```diff
@@ -1018,6 +1026,22 @@ where
...
+                while let Some(event) = engine_events_stream.next().await {
+                    if let ConsensusEngineEvent::InvalidBlock(block) = event &&
+                        let Ok(recovered) =
+                            RecoveredBlock::try_recover_sealed(block.as_ref().clone())
+                    {
+                        bad_block_store.insert(recovered);
+                    }
+                }
+            }));
+        }
```
klkvr: hmm can we not have this being handled by `DebugApi` itself?

crates/rpc/rpc-eth-api/src/types.rs
```diff
@@ -28,6 +29,8 @@ pub trait EthApiTypes: Send + Sync + Clone {
     type NetworkTypes: RpcTypes;
     /// Conversion methods for transaction RPC type.
     type RpcConvert: RpcConvert<Network = Self::NetworkTypes>;
+    /// Provider block type used by RPC.
+    type ProviderBlock: BlockTrait;
```
klkvr: would prefer us to avoid changing this trait

examples/exex-hello-world/src/main.rs
```diff
@@ -66,7 +66,11 @@ async fn ethapi_exex<Node, EthApi>(
 ) -> eyre::Result<()>
 where
     Node: FullNodeComponents<Types: NodeTypes<ChainSpec: EthereumHardforks>>,
-    EthApi: FullEthApi,
+    EthApi: FullEthApi
+        + reth_ethereum::rpc::eth::RpcNodeCore<
+            Provider = Node::Provider,
+            Primitives = <Node::Types as NodeTypes>::Primitives,
+        >,
```
klkvr: hmm unclear why this is needed. I know that those bounds are sometimes problematic but we need to figure out solutions for this that don't break such user-facing APIs

klkvr (commented):
would prefer us to avoid doing breaking changes for this

crates/node/builder/src/rpc.rs
0xKarl98: Have moved this part of logic to inside DebugApi

crates/rpc/rpc-eth-api/src/types.rs
0xKarl98: Have revert changes to this trait , it remains the same as before

klkvr (approved):
pending @mattsse

mattsse (approved):

PR #15925, opened 2025-04-25
chore: Adding warning log for empty MPT root as withdrawals_root post-Isthmus

Ayushdubey86:
Hello @mattsse , this should close #15916 

Also it seems I am on fire are there any issue I can look into!

PR #20178, opened 2025-12-07
chore: add missing EMPTY_ROOT_HASH warning for prehashed validation

Fallengirl:
Follow-up to #15925

#15925 added a warning log when storage_root == EMPTY_ROOT_HASH in verify_withdrawals_root(). Should the same check be in verify_withdrawals_root_prehashed()? Both functions do the same validation, and the prehashed version is used in production (engine.rs)


mattsse (commented):
we can do without

PR #20179, opened 2025-12-07
docs: remove hardcoded values from prune segment comments

CreeptoGengar:
Replace hardcoded configuration values (3500, 5, 8750) in account_history and storage_history prune segment comments with generic formulas. 
This prevents the comments from becoming outdated when default values change. 
Also fixes incorrect  constant reference in storage_history.rs comment.


mattsse (commented):
we can keep this as is

PR #20183, opened 2025-12-07
fix: fail fast on config file read/parse errors

anim001k:
Replaced `unwrap_or_default()` with proper error propagation using `?` operator. This ensures that configuration errors are surfaced immediately, preventing silent failures.


mattsse (commented):
we can keep this as is

PR #20184, opened 2025-12-07
Avoid recomputing block RLP length in validation

0xcharry:
cache the block RLP length before the Osaka size check, reuse the cached value when constructing the oversized-block error


mattsse (commented):
we can keep this as is because this is only the error case

PR #20185, opened 2025-12-07
docs: add admin_peers and admin_clearTxpool sections

Galoretka:
Updated the admin namespace JSON-RPC documentation by adding two missing sections: admin_peers and admin_clearTxpool. Each section follows the established style with a method invocation table and a concise example. The sections are placed just before the admin_peerEvents documentation.


mattsse (approved):
sure why not

PR #20187, opened 2025-12-07
feat(primitives-traits): add recover_transactions_ref to avoid cloning

forkfury:
### Fix
Adds recover_transactions_ref() to BlockBody to avoid cloning transactions when owned values aren't needed. recover_transactions() clones all transactions, which is expensive for large blocks.
### Changes
- Added recover_transactions_ref() returning Result<impl Iterator<Item = Recovered<&Transaction>>, RecoveryError>
- Uses Recovered::new_unchecked() with references instead of cloning
- Matches the pattern used in RecoveredBlock::transactions_recovered()
- Keeps recover_transactions() unchanged for backward compatibility


mattsse (approved):
makes sense

PR #20190, opened 2025-12-08
refactor(storage): handle missing static file headers gracefully

eeemmmmmm:
Replaced `.expect()` call with proper Option handling in `remove_blocks_above()`. 

The previous code would panic if static files had no headers, but this is a valid state during initial sync or after genesis. Now we check if headers exist before attempting to prune them, matching the same pattern used for transactions in `remove_bodies_above()`.

Added debug logs for both branches to make it easier to track what's happening during block removal.


mattsse (commented):
we cna keep this as is

PR #20191, opened 2025-12-08
fix(network): preserve FIFO among High priority requests when Normal absent

sashass1315:
FIFO violation for high-priority requests:
In poll, when processing Priority::High, the insertion position is calculated as position(|req| req.is_normal_priority()).unwrap_or(0) and then insert(pos, request).

If the queue consists only of High requests, the new High request will be inserted at index 0, thereby placing it before the existing High requests and violating the required “add to the end of the High queue” behavior. The comment explicitly says “add this request to the back of the high-priority queue,” but the implementation in this case did the opposite.


mattsse (commented):
this isnt right

PR #20198, opened 2025-12-08
fix(rpc): validate fee history reward percentiles

SashaMalysehko:
Ensure eth_feeHistory reward_percentiles are fully validated against the [0, 100] range and checked for monotonicity for all entries, including single-element vectors and the last percentile. This prevents invalid user-supplied percentiles from slipping through RPC input and keeps the implementation aligned with the API expectations.


mattsse (approved):
I see, I believe this is actually a correct fix, need to check a few things tho

mattsse (approved):
checks out

PR #20202, opened 2025-12-08
fix(rpc): return error if toBlock exceeds current head

futreall:
Closes #20197

Returns `BlockRangeExceedsHead` error when `toBlock` is greater than current chain head in `eth_getLogs`.


mattsse (approved):
sweet

PR #20206, opened 2025-12-08
feat: add helper method to eth validator

klkvr:
adds a `validate_one_with_state_provider` method to validate transaction with an exisiting state provider impl

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -247,6 +247,20 @@ where
         }
     }
 
+    /// Validates a single transaction with the provided state provider.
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -247,6 +247,20 @@ where
         }
     }
 
+    /// Validates a single transaction with the provided state provider.
+    pub fn validate_one_with_state_provider(
+        &self,
+        origin: TransactionOrigin,
+        transaction: Tx,
+        state: impl AccountInfoReader,
```
mattsse, line 255: this works because auto impled for &

mattsse (approved):

PR #20207, opened 2025-12-08
perf: batch key/value length reads in EtlFile

gap-editor:
**Summary**
This PR optimizes the `read_next` method in the ETL data collector. Previously, the method performed two separate 8-byte reads to fetch the key and value lengths. This change consolidates them into a single 16-byte read.

**Motivation**

  * **Reduce Syscalls:** Halves the number of read calls required just to determine entry sizes, improving performance when iterating over large datasets.
  * **Cleanliness:** Uses slice indexing and `try_into()` for standard byte conversion.

### DIFF
--- a/crates/etl/src/lib.rs
+++ b/crates/etl/src/lib.rs
@@ -259,14 +259,13 @@ impl EtlFile {
             return Ok(None)
         }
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we can keep this as is

PR #20216, opened 2025-12-09
feat(transaction-pool): add `new_blob_pool_transactions_listener`

AJStonewee:
Noticed we have listener helpers for _pending/basefee/queued_ pools but blob pool was missing one

Added `new_blob_pool_transactions_listener()` in `TransactionPool` trait - same pattern as the other three.


mattsse (approved):
cool, this makes sense

mattsse (approved):
cool, this makes sense

PR #20217, opened 2025-12-09
docs: document state and block overrides for trace_call

Forostovec:
`trace_call` accepts optional `stateOverrides` and `blockOverrides` parameters, but the JSON-RPC docs only showed three arguments. This updates the `trace_call` section to describe both override objects and reflects them in the example method invocation so that the documentation matches the actual RPC interface.


mattsse (approved):
fine

PR #20222, opened 2025-12-09
ci: use depot runners

### LINKED ISSUE #20220: Switch CI to depot runners
### Describe the feature

Same as we have in Foundry, see https://github.com/foundry-rs/foundry/blob/master/.github/workflows/docker-publish.yml#L85-L101 for Docker example, and https://github.com/foundry-rs/foundry/blob/9dc83c98223df59d102dde47405f0650c97b256a/.github/workflows/ci.yml#L64 for a regular workflow that uses custom runner.

### Additional context

_No response_

shekhirin:
Closes #20220

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - depot-ubuntu-latest
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,3 +1,7 @@
 self-hosted-runner:
   labels:
     - depot-ubuntu-latest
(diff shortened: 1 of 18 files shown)


DaniPopes (commented):
as a general rule I use, keep using GHA ubuntu-latest for really simple jobs like small scripts and the "x-success" since depot usually has higher startup cost

and then you can tweak the core count (default/unspecified is 2, depot-ubuntu-latest-X for X power of two cores) for the slower ones. you can also drop partitions for tests if it becomes fast enough with e.g. 16 cores

DaniPopes (approved):

yongkangc (approved):
nice stuff

PR #20227, opened 2025-12-09
perf(chain-state): `executed_block_receipts_ref`

shekhirin:
This clone gets very expensive for blocks with many receipts.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -18,7 +18,7 @@ use reth_primitives_traits::{
 };
 use reth_storage_api::StateProviderBox;
 use reth_trie::{updates::TrieUpdatesSorted, HashedPostStateSorted, TrieInputSorted};
-use std::{collections::BTreeMap, sync::Arc, time::Instant};
+use std::{collections::BTreeMap, ops::Deref, sync::Arc, time::Instant};
 use tokio::sync::{broadcast, watch};
 
 /// Size of the broadcast channel used to notify canonical state events.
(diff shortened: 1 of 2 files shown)


crates/chain-state/src/in_memory.rs
```diff
@@ -646,6 +648,24 @@ impl<N: NodePrimitives> BlockState<N> {
...
+    pub fn executed_block_receipts_ref(&self) -> Option<&[N::Receipt]> {
+        let receipts = self.receipts();
+
+        debug_assert!(
+            receipts.len() <= 1,
+            "Expected at most one block's worth of receipts, found {}",
+            receipts.len()
+        );
+
+        receipts.first().map(|receipts| receipts.as_slice())
```
DaniPopes: nit u can still .as_deref().unwrap_or_default

DaniPopes (approved):

joshieDo (approved):

crates/storage/provider/src/providers/consistent.rs
```diff
@@ -1055,7 +1058,7 @@ impl<N: ProviderNodeTypes> ReceiptProvider for ConsistentProvider<N> {
         for block_state in self.head_block.iter().flat_map(|b| b.chain()) {
             let executed_block = block_state.block_ref();
             let block = executed_block.recovered_block();
-            let receipts = block_state.executed_block_receipts();
+            let Some(receipts) = block_state.executed_block_receipts_ref() else { continue };
```
mattsse: I believe skipping here makes sense

mattsse (approved):

crates/chain-state/src/in_memory.rs
shekhirin: hmm
```rust
        receipts.first().as_deref().unwrap_or_default()
```
doesn't work for some reason, but
```rust
        receipts.first().map(|receipts| receipts.deref()).unwrap_or_default()
```
does

crates/storage/provider/src/providers/consistent.rs
shekhirin: simplified, less LOC changes now

crates/chain-state/src/in_memory.rs
DaniPopes: right `.first()` is `Option<&Vec<_>>` so it would be `map(Deref::deref)` or `map(Vec::as_slice)` which is the original

PR #20228, opened 2025-12-09
feat(net): pool transactions import duration metric

shekhirin:


crates/net/network/src/transactions/mod.rs
```diff
@@ -1327,6 +1327,8 @@ where
         transactions: PooledTransactions<N::PooledTransaction>,
         source: TransactionSource,
     ) {
+        let start = Instant::now();
+
```
mattsse: this can be moved below

crates/net/network/src/metrics.rs
```diff
@@ -131,6 +131,8 @@ pub struct TransactionsManagerMetrics {
     /// capacity. Note, this is not a limit to the number of inflight requests, but a health
     /// measure.
     pub(crate) capacity_pending_pool_imports: Counter,
+    /// The time it took to import transactions.
+    pub(crate) pool_imports_duration: Histogram,
```
mattsse: what does this exactly represent?

the naming indicates that this is the time it took to import txs into the pool, however this isn't what is being measured by this metric, which is more like prepare or rather just recovery

mattsse (changes requested):

mattsse (approved):

PR #20230, opened 2025-12-09
perf(prune): use delete_current_duplicates for MerkleChangeSets tables

mediocregopher:
When pruning DUPSORT tables, rather than using `delete_current` which would only delete the single entry within the duplicate key for each table, we now use `delete_current_duplicates` which deletes the whole duplicate key at once. This results in much faster pruning for those tables, and less IO overall.

We cannot use this new method for other DUPSORT tables like AccountChangeSets and StorageChangeSets, as that logic relies on actually iterating over all entries in order to later update the AccountHistory/StorageHistory index tables.

### DIFF
--- a/crates/prune/prune/src/db_ext.rs
+++ b/crates/prune/prune/src/db_ext.rs
@@ -2,13 +2,13 @@ use crate::PruneLimiter;
 use reth_db_api::{
     cursor::{DbCursorRO, DbCursorRW, RangeWalker},
     table::{DupSort, Table, TableRow},
-    transaction::DbTxMut,
+    transaction::{DbTx, DbTxMut},
     DatabaseError,
 };
 use std::{fmt::Debug, ops::RangeBounds};
 use tracing::debug;
 
-pub(crate) trait DbTxPruneExt: DbTxMut {
+pub(crate) trait DbTxPruneExt: DbTxMut + DbTx {
     /// Prune the table for the specified pre-sorted key iterator.
     ///
     /// Returns number of rows pruned.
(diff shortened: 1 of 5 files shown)


crates/prune/prune/src/db_ext.rs
```diff
@@ -123,6 +123,52 @@ pub(crate) trait DbTxPruneExt: DbTxMut {
...
     }
+
+    /// Prune a DUPSORT table for the specified key range.
+    ///
+    /// Returns number of rows pruned.
+    fn prune_dupsort_table_with_range<T: DupSort>(
+        &self,
+        keys: impl RangeBounds<T::Key> + Clone + Debug,
+        limiter: &mut PruneLimiter,
+        mut delete_callback: impl FnMut(TableRow<T>),
```
mattsse, line 134: ah makes sense

mattsse (approved):
lgtm, pending @joshieDo

crates/prune/prune/src/db_ext.rs
```diff
@@ -123,6 +123,52 @@ pub(crate) trait DbTxPruneExt: DbTxMut {
 
         Ok(false)
     }
+
+    /// Prune a DUPSORT table for the specified key range.
+    ///
+    /// Returns number of rows pruned.
```
joshieDo, line 129: might be a bit misleading depending on how you look at it. `deleted_entries` is being incremented per key, which might have many rows.

joshieDo (approved):

crates/prune/prune/src/db_ext.rs
mediocregopher: True, i'll just get rid of that cause it's not used anyway

mediocregopher: Actually I made it return the real count by calling `entries` on the table, since that number is being used by the checkpoint

joshieDo (approved):

PR #20233, opened 2025-12-09
feat(cli): Allow walking a range of an MDBX table using `db mdbx get`

mediocregopher:
This change adds the ability to walk a range of an MDBX table using the `db mdbx get` sub-command, whereas previously we could only retrieve a single key.

This mode is activated when:

* Two arguments (start and end) are given for a non-DUPSORT table
* Three or four arguments (start, subkey-start, end, [subkey-end]) are given for a DUPSORT table

The end parameters are always exclusive. Existing behavior is unchanged.

To support this change a new trait, ValueWithSubKey, is defined and implemented for all dupsort table values, so that they can return their subkey in a generic way. This is a generally useful trait that could be used for other conveniences in the future, for example a range walker over dupsort tables.

### DIFF
--- a/crates/cli/commands/src/db/get.rs
+++ b/crates/cli/commands/src/db/get.rs
@@ -8,12 +8,17 @@ use reth_db::{
     RawDupSort,
 };
 use reth_db_api::{
-    table::{Decompress, DupSort, Table},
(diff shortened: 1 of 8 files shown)


joshieDo (approved):
lgtm

mattsse (approved):
nice!

PR #20234, opened 2025-12-09
docs: remove incorrect total_difficulty mention from process_iter

forkfury:
### Fix
Removed incorrect documentation claim that process_iter updates total_difficulty. 
The function only processes headers and bodies; total_difficulty is not part of its implementation.
### Changes
Removed "Adds on to total_difficulty and" from process_iter documentation in crates/era-utils/src/history.rs


mattsse (approved):
fine

PR #20235, opened 2025-12-09
docs(config): clarify PruneConfig::merge semantics

kurahin:
Clarified the exact merge semantics for PruneConfig based on current behavior and tests. Optional segment modes are taken from other only if self is None; block_interval is taken only when self uses the default value; merkle_changesets is always taken from other; receipts_log_filter is taken only when self is empty and other is non-empty. This matches the behavior asserted by test_prune_config_merge and avoids misleading expectations.


mattsse (approved):
I think these are actually useful

PR #20236, opened 2025-12-09
docs: add missing documentation for serde_bincode_compat::ExExNotification

hawkadrian:
## What

Adds missing documentation for `ExExNotification` enum in `serde_bincode_compat` module and removes `#[expect(missing_docs)]` attribute.

## Why

The enum is part of the public API (exported via `pub mod serde_bincode_compat`) but lacked documentation, which is inconsistent with other `serde_bincode_compat` types in the codebase. This improves API discoverability and follows the project's documentation standards.


mattsse (approved):
fine I allow it

PR #20241, opened 2025-12-10
docs(txpool): fix PoolSize total field comment to include blob pool

JacksonEi:
The comment incorrectly stated `total = pending + basefee + queued`, missing the blob pool component that was added in EIP-4844 support.


mattsse (approved):
fair

PR #20242, opened 2025-12-10
fix: add missing std feature propagation

gancerlory:
Fixes std feature propagation across the workspace.

When crate A depends on crate B and both have std features, A needs to activate B's std. Without this, we get silent feature mismatches that can cause compile errors or unexpected behavior in no_std environments.


mattsse (commented):
this isnt needed because these crates dont support no-std

PR #20245, opened 2025-12-10
fix(consensus): swap got/expected in validate_cancun_gas error

reallesee:
The GotExpected fields in validate_cancun_gas were swapped, making error messages confusing.

When validating blob gas, we compute the total from transactions and compare it against the header value. The error should report what we computed (got) vs what the header claims (expected), not the other way around.

This aligns with how other body-vs-header validations work in the same file (tx_root, ommers_hash, withdrawals_root).


mattsse (commented):
this is confusing either way, we settled on this order

PR #20246, opened 2025-12-10
docs: fix timestamp validation comment

forkfury:
## Fix

Corrects misleading doc comment for `validate_against_parent_timestamp` function.

## Changes

- Updated function doc comment to accurately state that it validates the block timestamp is greater than the parent timestamp, not that it's in the past
- The previous comment contradicted the actual validation logic which requires `header.timestamp() > parent.timestamp()`


mattsse (approved):
reasonable

PR #20247, opened 2025-12-10
docs: clarify network mode, tx gossip and NAT

sashass1315:
Documented NetworkMode semantics and how PoS vs PoW affect block broadcasting.
Added a dedicated note that block announcements are disabled on PoS networks and blocks are instead obtained from the consensus layer and requested over devp2p.
Also describe the tx_gossip_disabled flag and when it is useful, and briefly mention NAT resolver support for advertising the correct external address.


mattsse (approved):
I allow it

PR #20251, opened 2025-12-10
fix(db): add metrics for clear operation

CreeptoGengar:
Add Operation::Clear variant and wrap clear() method with operation metrics to match the pattern used by other write operations (put, delete). 
This ensures consistent metrics tracking for table clearing operations.


mattsse (commented):
clear is only for specific commands for which we dont need this because we dont commonly invoke this

PR #20253, opened 2025-12-10
feat(storage): add rocksdb provider into database provider

duyquang6:
Resolve #20152

Changes:
- `rocksdb_stub.rs` (stub for non-Unix, no feature), `rocksdb_provider.rs` (for unix, enabled rocksdb feature)
- Added `RocksDBProvider` to `ProviderFactory` with builder pattern
- Added `/rocksdb` directory alongside `/db` and `/static_files`
- Integration: Updated provider factory builder, node launch, CLI commands, and test utilities

### DIFF
--- a/crates/cli/commands/src/common.rs
+++ b/crates/cli/commands/src/common.rs
@@ -23,7 +23,7 @@ use reth_node_core::{
     dirs::{ChainPath, DataDirPath},
 };
 use reth_provider::{
-    providers::{BlockchainProvider, NodeTypesForProvider, StaticFileProvider},
+    providers::{BlockchainProvider, NodeTypesForProvider, RocksDBProvider, StaticFileProvider},
     ProviderFactory, StaticFileProviderFactory,
 };
 use reth_stages::{sets::DefaultStages, Pipeline, PipelineTarget};
@@ -98,18 +98,25 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
         }
 
         info!(target: "reth::cli", ?db_path, ?sf_path, "Opening storage");
-        let (db, sfp) = match access {
+        let (db, sfp, rocksdb_provider) = match access {
             AccessRights::RW => (
                 Arc::new(init_db(db_path, self.db.database_args())?),
                 StaticFileProvider::read_write(sf_path)?,
+                RocksDBProvider::builder(data_dir.rocksdb())
+                    .with_database_log_level(self.db.log_level)
+                    .build()?,
             ),
             AccessRights::RO | AccessRights::RoInconsistent => (
                 Arc::new(open_db_read_only(&db_path, self.db.database_args())?),
                 StaticFileProvider::read_only(sf_path, false)?,
+                RocksDBProvider::builder(data_dir.rocksdb())
+                    .with_database_log_level(self.db.log_level)
+                    .build()?,
             ),
         };
 
-        let provider_factory = self.create_provider_factory(&config, db, sfp, access)?;
+        let provider_factory =
+            self.create_provider_factory(&config, db, sfp, rocksdb_provider, access)?;
         if access.is_read_write() {
             debug!(target: "reth::cli", chain=%self.chain.chain(), genesis=?self.chain.genesis_hash(), "Initializing genesis");
             init_genesis_with_settings(&provider_factory, self.static_files.to_settings())?;
@@ -128,6 +135,7 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
         config: &Config,
         db: Arc<DatabaseEnv>,
         static_file_provider: StaticFileProvider<N::Primitives>,
+        rocksdb_provider: RocksDBProvider,
         access: AccessRights,
     ) -> eyre::Result<ProviderFactory<NodeTypesWithDBAdapter<N, Arc<DatabaseEnv>>>>
     where
@@ -138,6 +146,7 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
             db,
             self.chain.clone(),
             static_file_provider,
+            rocksdb_provider,
         )?
         .with_prune_modes(prune_modes.clone());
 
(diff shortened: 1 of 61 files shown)


crates/storage/provider/src/providers/rocksdb_stub.rs
duyquang6, line 1: this is no-op version of rocksdb provider which help simplify integrate into DatabaseProvider

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -144,6 +149,12 @@ impl<N: NodeTypesWithDB> StorageSettingsCache for ProviderFactory<N> {
     }
 }
 
+impl<N: NodeTypesWithDB> RocksDBProviderFactory for ProviderFactory<N> {
```
fgimenez, line 152: we should have an impl like this but for `BlockchainProvider` in crates/storage/provider/src/providers/blockchain_provider.rs:
```rust
  impl<N: ProviderNodeTypes> RocksDBProviderFactory for BlockchainProvider<N> {
      fn rocksdb_provider(&self) -> &RocksDBProvider {
          self.database.rocksdb_provider()
      }
  }
```
this follows the same delegation pattern used by `StaticFileProviderFactory`

crates/storage/provider/src/traits/rocksdb_provider.rs
```diff
@@ -0,0 +1,9 @@
+use crate::providers::RocksDBProvider;
+
+/// `RocksDB` provider factory.
+///
+/// This trait provides access to the `RocksDB` provider
+pub trait RocksDBProviderFactory {
```
fgimenez, line 6: `FullProvider` trait in crates/storage/provider/src/traits/full.rs includes `StaticFileProviderFactory` but not `RocksDBProviderFactory`, for consistency we should add it too (both in the definition and in the blanket impl)

crates/storage/provider/src/providers/rocksdb_stub.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn get<T>(&self, _key: T::Key) -> ProviderResult<Option<T::Value>>
+    where
+        T: Table,
+        T::Value: Decompress,
+    {
+        Err(UnsupportedProvider)
+    }
+
+    /// Put a value into `RocksDB` (stub implementation).
+    pub fn put<T>(&self, _key: T::Key, _value: T::Value) -> ProviderResult<()>
```
fgimenez: changed to match the signature of the real impl
```suggestion
    pub fn put<T>(&self, _key: T::Key, _value: &T::Value) -> ProviderResult<()>
```

crates/storage/provider/src/providers/rocksdb_stub.rs
```diff
@@ -0,0 +1,128 @@
...
+pub struct RocksDBBatch;
+
+impl RocksDBBatch {
+    /// Put a value into the batch (stub implementation).
+    pub fn put<T>(&self, _key: T::Key, _value: T::Value) -> ProviderResult<()>
+    where
+        T: Table,
+    {
+        Err(UnsupportedProvider)
+    }
```
fgimenez, line 104: we should add the missing `delete` method present in the real impl
```suggestion
    }
    /// Delete a value from the batch (stub implementation).
    pub fn delete<T>(&self, _key: T::Key) -> ProviderResult<()>
    where
        T: Table,
    {
        Err(UnsupportedProvider)
    }
```

fgimenez (commented):
looks good, some initial suggestions

duyquang6:
thank sir @fgimenez, I’ve pushed the fixes for the comments — it’s ready for review

fgimenez (approved):
lgtm, pending @Rjected @joshieDo @shekhirin @yongkangc

joshieDo (approved):

Rjected (approved):
This makes sense to me, let's merge https://github.com/paradigmxyz/reth/pull/20288 first and rebase this one so it can use the tx utils

crates/cli/commands/src/common.rs
```diff
@@ -98,18 +98,25 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
...
             AccessRights::RW => (
                 Arc::new(init_db(db_path, self.db.database_args())?),
                 StaticFileProvider::read_write(sf_path)?,
+                rocksdb_provider,
             ),
             AccessRights::RO | AccessRights::RoInconsistent => (
                 Arc::new(open_db_read_only(&db_path, self.db.database_args())?),
                 StaticFileProvider::read_only(sf_path, false)?,
+                // Note: TransactionDB only support read-write mode
+                rocksdb_provider,
```
duyquang6: I'm not sure if this is what we want @fgimenez, as https://github.com/paradigmxyz/reth/pull/20288#pullrequestreview-3571451057 suggest should remove readonly mode

crates/cli/commands/src/common.rs
```diff
@@ -108,8 +108,13 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
                 StaticFileProvider::read_only(sf_path, false)?,
             ),
         };
+        // TransactionDB only support read-write mode
+        let rocksdb_provider = RocksDBProvider::builder(data_dir.rocksdb())
```
duyquang6, line 114: context: there is AccessRights::RW and AccessRights::RO, in case of rocksdb, there is only support read write mode, so, in AccessRights::RO, behind rocksdb will use read write mode

I'm not sure if this is what we want sir @fgimenez, as https://github.com/paradigmxyz/reth/pull/20288#pullrequestreview-3571451057 suggest should remove readonly mode, so this code here adapt to it

fgimenez: yes, this approach is correct, `TransactionDB` does not support RO and we always create `RocksDB` in RW no matter the value of `AccessRights`. `AccessRights` enum controls `MDBX` and `StaticFiles` access, not `RocksDB`.

One remaining thing, we can now remove the unused `ReadOnlyRocsDBAccess` error variant from `crates/storage/errors/src/provider.rs`, not used anywhere

duyquang6: thanks, I fixed that :100:

PR #20254, opened 2025-12-10
fix(rpc): avoid signing Optimism deposit transactions

radik878:
SignableTxRequest for OpTransactionRequest was calling the signer before checking is_deposit(), even though Optimism deposit transactions are not supposed to be user-signed. This caused unnecessary signer calls (including potential hardware wallet prompts) for requests that are ultimately rejected as invalid. Reorder the logic to build the typed transaction, immediately reject deposits, and only then sign non-deposit transactions, aligning the RPC signing path with the Optimism deposit semantics.


mattsse (approved):
makes sense

PR #20255, opened 2025-12-10
feat: add support for eth/70 eip-7975

0xKarl98:
This closes issue #19805  , since eip-7542 has been withdrawn , this PR is implementing eip-7975 . 

Scope of this PR: 

- focus on server‑side eth/70 support and basic compatibility only; full client‑side eth/70 pagination and the EIP‑7975 safety checks for partial receipt downloads (using lastBlockIncomplete and transaction/gas‑limit constraints) are explicitly left for follow‑up PRs.

cc @mattsse


crates/net/eth-wire-types/src/receipts.rs
```diff
@@ -17,6 +17,23 @@ pub struct GetReceipts(
...
+///
+/// This follows EIP-7975 and represents the full wire message
+/// `[request-id, firstBlockReceiptIndex, [blockhash₁, ...]]`.
+#[derive(Clone, Debug, PartialEq, Eq, RlpEncodable, RlpDecodable)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[add_arbitrary_tests(rlp)]
+pub struct GetReceipts70 {
+    /// The request id used to correlate the response.
+    pub request_id: u64,
```
mattsse: I think we don't need this here because the request id is tracked in

https://github.com/paradigmxyz/reth/blob/2e6238746953a996649a34ffd9cd9e0f758e4355/crates/net/eth-wire-types/src/message.rs#L581-L581

crates/net/eth-wire-types/src/block_range.rs
mattsse, line 1: I think we can exclude all of this for this pr?

crates/net/eth-wire-types/src/message.rs
```diff
@@ -111,13 +125,24 @@ impl<N: NetworkPrimitives> ProtocolMessage<N> {
                 }
                 EthMessage::NodeData(RequestPair::decode(buf)?)
             }
-            EthMessageID::GetReceipts => EthMessage::GetReceipts(RequestPair::decode(buf)?),
+            EthMessageID::GetReceipts => {
+                if version >= EthVersion::Eth70 {
+                    EthMessage::GetReceipts70(GetReceipts70::decode(buf)?)
```
mattsse: this should also use RequestPair

crates/net/eth-wire-types/src/receipts.rs
```diff
@@ -17,6 +17,23 @@ pub struct GetReceipts(
...
+    /// Index into the receipts of the first requested block hash.
+    pub first_block_receipt_index: u64,
+    /// The block hashes to request receipts for.
+    pub block_hashes: Vec<B256>,
```
mattsse, line 34: then we need to adjust rlp encoding of this type so that this becomes:

```
firstBlockReceiptIndex: P, [blockhash₁: B_32, blockhash₂: B_32, ...]
```

and not 
```
[firstBlockReceiptIndex: P, [blockhash₁: B_32, blockhash₂: B_32, ...]]]
```

like it would with the macro

crates/net/eth-wire-types/src/status.rs
mattsse, line 1: these changes we can also undo

mattsse (changes requested):

crates/net/eth-wire-types/src/status.rs
```diff
@@ -378,15 +401,79 @@ impl Debug for StatusEth69 {
     }
 }
 
-/// `StatusMessage` can store either the Legacy version (with TD) or the
-/// eth/69 version (omits TD).
+/// Status message for `eth/70` including the block range as per EIP-7975.
```
mattsse: we dont need this anymore, right? so we can undo all status changes I believe

crates/net/eth-wire-types/src/message.rs
```diff
@@ -271,6 +296,15 @@ pub enum EthMessage<N: NetworkPrimitives = EthNetworkPrimitives> {
...
     Receipts69(RequestPair<Receipts69<N::Receipt>>),
+    /// Represents a Receipts request-response pair for eth/70.
+    #[cfg_attr(
+        feature = "serde",
+        serde(bound = "N::Receipt: serde::Serialize + serde::de::DeserializeOwned")
+    )]
+    ///
+    /// Note: The eth/70 encoding for `Receipts` in EIP-7975 inlines the
+    /// request id instead of using a generic [`RequestPair`].
+    Receipts70(Receipts70<N::Receipt>),
```
mattsse: this should also use RequestPair and we can implement encode manually for Receipts70

crates/net/network/src/session/active.rs
```diff
@@ -135,6 +137,8 @@ pub(crate) struct ActiveSession<N: NetworkPrimitives> {
     /// The last latest block number we sent in a range update
     /// Used to avoid sending unnecessary updates when block height hasn't changed significantly
     pub(crate) last_sent_latest_block: Option<u64>,
+    /// The last time we updated the remote block range information.
+    pub(crate) last_range_update: Option<Instant>,
```
mattsse: this isnt needed for this pr

crates/net/network/src/session/active.rs
```diff
@@ -176,6 +180,7 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
                     request_id,
                     rx: PeerResponse::$resp_item { response },
                     received: Instant::now(),
+                    metadata: ReceivedRequestMetadata::None,
```
mattsse: we can assume that we handle the request properly internally so we dont need to track this here

crates/net/network/src/session/active.rs
```diff
@@ -365,9 +463,25 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
...
+            ReceivedRequestMetadata::Eth70Receipts { first_block_receipt_index } => {
+                // For eth/70 GetReceipts requests we construct a `Receipts70`
+                // message that respects `firstBlockReceiptIndex` but still
+                // reuses the existing receipts lookup path.
+                self.build_eth70_receipts_response(id, first_block_receipt_index, resp)
+            }
+            ReceivedRequestMetadata::None => resp.try_into_message(id),
+        };
+
+        match msg_result {
```
mattsse: this we dont need to change, we can instead introduce another variant for this so we can handle the v70 request internally properly

crates/net/network/src/session/active.rs
```diff
@@ -376,6 +490,83 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
...
+    /// This applies the `firstBlockReceiptIndex` offset to the first block's
+    /// receipts and enforces the soft response size limit (`SOFT_RESPONSE_LIMIT`)
+    /// by potentially truncating the last block's receipts and setting
+    /// `lastBlockIncomplete = true` as defined in EIP-7975.
+    fn build_eth70_receipts_response(
+        &self,
+        id: u64,
+        first_block_receipt_index: u64,
+        resp: PeerResponseResult<N>,
+    ) -> RequestResult<EthMessage<N>> {
```
mattsse: this we can undo, this should be handled where we process the request

mattsse (changes requested):
cool, 

we can simplify a lot of things here, basically the session shouldnt have to deal with any response manipulation

this has some additional changes for the withdrawn eip that we can fully exclude from this pr

crates/net/network/src/session/active.rs
0xKarl98: Suspect so , but thanks for the confirm

crates/net/eth-wire-types/src/status.rs
0xKarl98: Yeah forgot to change the comment here

crates/net/network/src/session/active.rs
0xKarl98: Have removed them

crates/net/eth-wire-types/src/message.rs
0xKarl98: I think we have used RequestPair here , but the comment isn't clear , i.e. :
pub type Receipts70<T = Receipt> = crate::message::RequestPair<Receipts70Payload<T>>;

Will make others misunderstood that we don't use RequestPair here  but i wil simplify it

crates/net/network/src/session/active.rs
0xKarl98: makes sense

crates/net/network/src/session/mod.rs
```diff
@@ -538,9 +538,8 @@ impl<N: NetworkPrimitives> SessionManager<N> {
                 // negotiated version
                 let version = conn.version();
 
-                // Configure the interval at which the range information is updated, starting with
-                // ETH69
-                let range_update_interval = (conn.version() >= EthVersion::Eth69).then(|| {
+                // Configure the interval at which the range information is updated (eth/69 only)
+                let range_update_interval = (conn.version() == EthVersion::Eth69).then(|| {
```
mattsse: this is also the case for anything above, so we can keep this unchanged

crates/net/network/src/session/active.rs
```diff
@@ -746,9 +841,8 @@ impl<N: NetworkPrimitives> Future for ActiveSession<N> {
                 };
 
                 if should_send {
-                    this.queued_outgoing.push_back(
-                        EthMessage::BlockRangeUpdate(this.local_range_info.to_message()).into(),
-                    );
+                    let msg = EthMessage::BlockRangeUpdate(this.local_range_info.to_message());
+                    this.queued_outgoing.push_back(msg.into());
```
mattsse: please keep unchanged

crates/net/eth-wire/src/protocol.rs
```diff
@@ -84,5 +84,6 @@ mod tests {
         assert_eq!(Protocol::eth(EthVersion::Eth67).messages(), 17);
         assert_eq!(Protocol::eth(EthVersion::Eth68).messages(), 17);
         assert_eq!(Protocol::eth(EthVersion::Eth69).messages(), 18);
+        assert_eq!(Protocol::eth(EthVersion::Eth70).messages(), 17);
```
mattsse: this shouldnt be the case, this should exactly like eth69

crates/net/eth-wire-types/src/status.rs
```diff
@@ -378,15 +383,20 @@ impl Debug for StatusEth69 {
...
+/// (omits TD, includes block range).
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum StatusMessage {
     /// The legacy status (`eth/66` through `eth/68`) with `total_difficulty`.
     Legacy(Status),
     /// The new `eth/69` status with no `total_difficulty`.
     Eth69(StatusEth69),
+    /// The `eth/70` status, identical to `eth/69` .
+    Eth70(StatusEth70),
```
mattsse: we dont need to change this because the status message is unchanged for eth70 so we dont need to change anything here

crates/net/eth-wire-types/src/receipts.rs
```diff
@@ -17,6 +17,44 @@ pub struct GetReceipts(
...
+}
+
+/// Helper type for the full eth/70 request carrying the request id alongside
+/// the payload.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct GetReceipts70(pub crate::message::RequestPair<GetReceipts70Payload>);
```
mattsse: this deviates from how we handle requestpairs and the actual message, this should be the request type only, like:

https://github.com/paradigmxyz/reth/blob/a9c21a395dfc71520d6440c0f968cfb7c3d9bb4f/crates/net/eth-wire-types/src/receipts.rs#L15-L18

crates/net/eth-wire-types/src/message.rs
mattsse: what I meant is that we should use 
```
RequestPair<Receipts70<N::Receipt>>
```

here, because if we start mixing this this becomes a bit confusing,

we can fix the encoding issue by implementing rlp directly for `Receipts70` so that this matches what the request/response message expects

crates/net/eth-wire-types/src/message.rs
```diff
@@ -259,6 +275,12 @@ pub enum EthMessage<N: NetworkPrimitives = EthNetworkPrimitives> {
     NodeData(RequestPair<NodeData>),
     /// Represents a `GetReceipts` request-response pair.
     GetReceipts(RequestPair<GetReceipts>),
+    /// Represents a `GetReceipts` request for eth/70.
+    ///
+    /// Note: Unlike earlier protocol versions, the eth/70 encoding for
+    /// `GetReceipts` in EIP-7975 inlines the request id. The type still wraps
+    /// a [`RequestPair`], but with a custom inline encoding.
+    GetReceipts70(GetReceipts70),
```
mattsse: same here

crates/net/eth-wire-types/src/message.rs
```diff
@@ -66,10 +66,12 @@ impl<N: NetworkPrimitives> ProtocolMessage<N> {
...
         let message = match message_type {
-            EthMessageID::Status => EthMessage::Status(if version < EthVersion::Eth69 {
-                StatusMessage::Legacy(Status::decode(buf)?)
-            } else {
+            EthMessageID::Status => EthMessage::Status(if version >= EthVersion::Eth70 {
+                StatusMessage::Eth70(StatusEth70::decode(buf)?)
+            } else if version >= EthVersion::Eth69 {
                 StatusMessage::Eth69(StatusEth69::decode(buf)?)
+            } else {
+                StatusMessage::Legacy(Status::decode(buf)?)
```
mattsse: this we can also keep unchanged

.dprint-cache/plugin-cache-manifest.json
mattsse, line 1: can remove

.dprint-cache/locks/.4365269138195617005.lock.poll
mattsse, line 1: same

.dprint-cache/locks/.4365269138195617005.lock
mattsse, line 1: same

crates/net/eth-wire-types/src/message.rs
```diff
@@ -597,6 +635,106 @@ impl<T> RequestPair<T> {
...
+            request_id,
+            message: crate::receipts::GetReceipts70Payload {
+                first_block_receipt_index,
+                block_hashes,
+            },
+        })
+    }
+}
+
+impl<T> RequestPair<crate::receipts::Receipts70Payload<T>> {
```
mattsse: we can get the same thing if we implement rlp manually for `Receipts70` so that this encodes/decodes as `lastBlockIncomplete, [receipts...]`

this way we can keep using requestpair and everything is more consistent

mattsse (changes requested):
almost there 🙏

a few more suggestions

crates/net/eth-wire-types/src/status.rs
0xKarl98: Ahh okay i know te status is unchanged from 69 to 70 , but just want a identifier here . Will undo that

crates/net/eth-wire-types/src/message.rs
0xKarl98: Yeah have removed statusEth70 related stuff , just use 69 (even though i thought the same while coding )

crates/net/eth-wire-types/src/message.rs
0xKarl98: makes sense , should have eth/70 payload deal with manual rlp instead so that the downstream more consistent

0xKarl98:
pending @mattsse  @klkvr

mattsse (approved):
cool, tysm!

PR #20258, opened 2025-12-10
docs(jsonrpc): add missing debug namespace RPC methods

eeemmmmmm:
Found 4 undocumented but fully implemented debug methods while reviewing the codebase:

- `debug_getRawTransactions` - returns encoded transactions for a block
- `debug_traceCallMany` - batch trace multiple calls (useful for bundle simulation)
- `debug_executionWitness` - generate execution witness by block number
- `debug_executionWitnessByBlockHash` - same but by hash


mattsse (approved):
fine

PR #20260, opened 2025-12-10
docs: use canonical --rollup.sequencer and note aliases

phrwlk:
Source in code: in crates/optimism/node/src/args.rs, the flag is declared as canonical --rollup.sequencer with aliases --rollup.sequencer-http and --rollup.sequencer-ws:

```rust
/// Endpoint for the sequencer mempool (can be both HTTP and WS)
#[arg(long = "rollup.sequencer", visible_aliases = ["rollup.sequencer-http", "rollup.sequencer-ws"])]
pub sequencer: Option<String>,
```


mattsse (approved):
reasonable

PR #20266, opened 2025-12-10
feat: add recover_transactions_unchecked_ref to BlockBody

JacksonEi:
Adds unchecked variant of recover_transactions_ref for BlockBody trait, avoiding transaction clones during signature recovery without s-value validation.


mattsse (approved):
fine

PR #20275, opened 2025-12-10
docs(run): add ethereum testnet examples to ethereum guide

VolodymyrBg:
The ethereum run guide claimed to cover both mainnet and testnets but only showed commands for mainnet:

```
---
description: How to run Reth on Ethereum mainnet and testnets.
---
```
That's the beginning of the file but there wasn't a guide how to run reth on testnets


mattsse (commented):
dont think we need this

PR #20283, opened 2025-12-11
docs: fix stages order and add missing EraStage

eeemmmmmm:
The stages list in docs/crates/stages.md was outdated:
- Missing EraStage (ERA1 import)
- MerkleChangeSets was in wrong position (after history indexing instead of after MerkleStage execute)

Fixed the order to match the actual pipeline in sets.rs and added EraStage documentation.


mattsse (approved):
fine

PR #20282, opened 2025-12-11
feat: add `account_history_in_rocksdb` field to `StorageSettings`

yongkangc:
Closes #20150

Adds configuration flag to control whether `AccountsHistory` table is stored in RocksDB or current MDBX

PR #20288, opened 2025-12-11
feat: add RocksDB variant to EitherReader and EitherWriter

### LINKED ISSUE #20278: Add `RocksDB` variant to `EitherReader`
### Describe the feature

Since the `EitherReader` is how we control reading to our storage types based on `StorageSettings`, we should create a variant in `EitherReader` which encompasses reading from RocksDB. Now that we have `TransactionHashNumbers` and `StoragesHistory` settings in `StorageSettings`, we can also create constructors for those tables.

https://github.com/paradigmxyz/reth/blob/179e1644ec72b9afb05a46367f2b55baf7de8cce/crates/storage/provider/src/either_writer.rs#L295-L302

### Additional context

_No response_

yongkangc:
Closes #20278, #20279, #20280

Add RocksDB variants to EitherReader, EitherWriter, and EitherWriterDestination enums.

This is done by switching from `DB` to `TransactionDB` and add `RocksTx` wrapper providing MDBX-like transaction semantics.

## Changes
- Replace `DB` with `TransactionDB` in `RocksDBProvider`
- Add `RocksTx<'db>` wrapper with `get`, `put`, `delete`, `iter`, `commit`, `rollback`
- Update `EitherWriter`/`EitherReader` to use `RocksTx` instead of `RocksDBProvider`
- Add tests for read-your-writes, rollback, and iterator behavior


crates/storage/provider/src/either_writer.rs
```diff
@@ -109,6 +123,40 @@ impl<'a> EitherWriter<'a, (), ()> {
...
+        } else {
+            Ok(EitherWriter::Database(
+                provider.tx_ref().cursor_write::<tables::TransactionHashNumbers>()?,
+            ))
+        }
+    }
+
+    /// Creates a new [`EitherWriter`] for storages history based on storage settings.
+    #[cfg(all(unix, feature = "rocksdb"))]
+    pub fn new_storages_history<P>(
```
yongkangc: is this what you mean by constructor? @Rjected

crates/storage/provider/src/either_writer.rs
```diff
@@ -238,6 +294,40 @@ impl EitherReader<(), ()> {
...
+    ) -> ProviderResult<EitherReaderTy<P, tables::StoragesHistory>>
+    where
+        P: DBProvider + NodePrimitivesProvider + StorageSettingsCache + RocksDBProviderFactory,
+        P::Tx: DbTx,
+    {
+        if provider.cached_storage_settings().storages_history_in_rocksdb {
+            Ok(EitherReader::RocksDB(provider.rocksdb_provider()))
+        } else {
+            Ok(EitherReader::Database(provider.tx_ref().cursor_read::<tables::StoragesHistory>()?))
+        }
```
yongkangc, line 354: will have to add in constructor for account history after 
https://github.com/paradigmxyz/reth/pull/20282

is merged in

crates/storage/provider/src/either_writer.rs
```diff
@@ -109,6 +114,42 @@ impl<'a> EitherWriter<'a, (), ()> {
...
+        if provider.cached_storage_settings().storages_history_in_rocksdb {
+            Ok(EitherWriter::RocksDB(rocksdb))
+        } else {
+            Ok(EitherWriter::Database(provider.tx_ref().cursor_write::<tables::StoragesHistory>()?))
+        }
+    }
+
+    /// Creates a new [`EitherWriter`] for transaction hash numbers based on storage settings.
+    #[cfg(all(unix, feature = "rocksdb"))]
+    pub fn new_transaction_hash_numbers<P>(
```
yongkangc, line 147: is this what you mean by constructor? @Rjected

crates/storage/provider/src/either_writer.rs
```diff
@@ -34,13 +36,16 @@ type EitherWriterTy<'a, P, T> = EitherWriter<
...
+/// Represents a destination for writing data, either to database, static files, or `RocksDB`.
 #[derive(Debug, Display)]
 pub enum EitherWriter<'a, CURSOR, N> {
     /// Write to database table via cursor
     Database(CURSOR),
     /// Write to static file
     StaticFile(StaticFileProviderRWRefMut<'a, N>),
+    /// Write to `RocksDB`
+    #[cfg(all(unix, feature = "rocksdb"))]
+    RocksDB(RocksDBProvider),
```
joshieDo: we probably will want to use a rocksdb tx instead of just the provider

yongkangc: I see, could u expand on why we would wanna go with tx instead? @joshieDo

yongkangc: ideally we should explicitly control with DatabaseProvider::commit when we actually commit to disk and in what order. with an txn we can roll back and commit but not with a provider. this breaks API compatibility in terms of rollback which might be necessary for some parts

Rjected: This PR is just for EitherReader / EitherWriter, so IMO we def want a rocksdb TX type in here. And we want to manage the tx lifecycle separately (at the provider level). I think it's fine if we use a rocksdb TX in the enums, even if this means the constructors like `new_transaction_hash_numbers` have to take a rocksdb TX argument that would be removed later

crates/storage/provider/src/either_writer.rs
Rjected: yep. I think we need to make sure this fn is not gated by cfg though

crates/storage/provider/src/either_writer.rs
Rjected: merged

crates/storage/provider/src/either_writer.rs
yongkangc: @joshieDo @Rjected made changes from `DB` to `TransactionDB` and added `RocksTx` wrapper.

This gives us the similar tx semantics as MDBX (read-your-writes, explicit commit/rollback, iterators seeing uncommitted writes). Let me know if that aligns with the design we should aim for.

crates/storage/provider/src/either_writer.rs
yongkangc: Done\! Introduced type aliases (`RocksTxArg`, `RocksTxRefArg`) that resolve to `RocksTx<'a>` or `&'a RocksTx<'a>` when rocksdb is enabled, and `()` when disabled. The constructors are now always available - internal cfg blocks handle the routing.

fgimenez (approved):
lgtm, once this one is merged we will need to update #20253 to:
- Remove `RocksDBAccess` enum: `TransactionDB` doesn't support read-only mode
- Remove `RocksDBBuilder::read_only()` method: cannot be implemented with `TransactionDB`
- Remove read-only access checks in `put()`, `delete()`, `write_batch()` that return `ProviderError::ReadOnlyRocksDBAccess`
- Remove `access` field from `RocksDBProviderInner` struct

Rjected (approved):
LGTM, let's explore `TransactionDB` (or `OptimisticTransactionDB`) vs `WriteBatch` for read vs write in a follow up, if it makes sense

PR #20289, opened 2025-12-11
feat: add semaphore for blocking IO requests

mattsse:
Adds settings for explicitly limiting concurrent tokio blocking tasks intended for IO (for now enabled for ethcall)

Note: this isnt yet enforced for all spawned tasks and should be applied selectively, gonna use this to investigate https://github.com/paradigmxyz/reth/issues/20059

closes #20090

see also https://github.com/paradigmxyz/reth/issues/20059


crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -96,7 +96,7 @@ pub trait EthState: LoadState + SpawnBlocking {
     {
         Ok(async move {
             let _permit = self
-                .acquire_owned()
+                .acquire_weighted_blocking_io(5)
```
shekhirin: this feels like a magic number, can you elaborate why it's 5?

mattsse: this basically enforces most 5 concurrent requests for this endpoint, while still leaving permits available

shekhirin: can we make this configurable? I expect some users have more than 5 concurrent eth_getProof and it's expected behaviour that they're ready for in terms of RAM

mattsse: I can restore the previous behaviour by allocating the other semaphore

klkvr (approved):
makes sense

shekhirin (approved):

PR #20291, opened 2025-12-11
feat(flashblocks): support eth_getBlockTransactionCount for flashblocks

sieniven:
## Summary 

- Enable support for eth_getBlockTransactionCountByNumber and eth_getBlockTransactionCountByHash apis for flashblock
- Note that the old logic on pending tags is maintained for backwards compatibility since we cant assume the concrete implementations of local_pending_block

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/block.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/block.rs
@@ -74,13 +74,9 @@ pub trait EthBlocks: LoadBlock<RpcConvert: RpcConvert<Primitives = Self::Primiti
         block_id: BlockId,
     ) -> impl Future<Output = Result<Option<usize>, Self::Error>> + Send {
         async move {
-            if block_id.is_pending() {
-                // Pending block can be fetched directly without need for caching
-                return Ok(self
-                    .provider()
-                    .pending_block()
-                    .map_err(Self::Error::from_eth_err)?
-                    .map(|block| block.body().transaction_count()));
+            // If no pending block from provider, build the pending block locally.
+            if let Some(pending) = self.local_pending_block().await? {
+                return Ok(Some(pending.block.body().transaction_count()));
             }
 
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
that makes sense but should we perhaps change

https://github.com/paradigmxyz/reth/blob/1752d6fb99d020bca8b4f5de89b2847f61984009/crates/rpc/rpc-eth-api/src/helpers/block.rs#L77-L83

to just 

https://github.com/paradigmxyz/reth/blob/1752d6fb99d020bca8b4f5de89b2847f61984009/crates/rpc/rpc-eth-api/src/helpers/block.rs#L189-L192

instead?

mattsse (approved):

crates/rpc/rpc-eth-api/src/helpers/block.rs
```diff
@@ -74,13 +74,9 @@ pub trait EthBlocks: LoadBlock<RpcConvert: RpcConvert<Primitives = Self::Primiti
         block_id: BlockId,
     ) -> impl Future<Output = Result<Option<usize>, Self::Error>> + Send {
         async move {
-            if block_id.is_pending() {
```
mattsse, line 77: ah we need to keep this check actually

mattsse (changes requested):

crates/rpc/rpc-eth-api/src/helpers/block.rs
sieniven: @mattsse hey, thanks for the spot, apologies. We have amended the code on our end here: [148c6cd](https://github.com/paradigmxyz/reth/pull/20291/commits/148c6cd4a82beb07b868cf1eae81685d1b820567), note that the old logic on pending tags is maintained for backwards compatibility since we cant assume the concrete implementations of local_pending_block. Let us know if there any further issues 👍

mattsse (approved):

PR #20292, opened 2025-12-11
refactor(era): make era count in era file name optional

lean-apple:
To follow better this change, https://github.com/eth-clients/e2store-format-specs/pull/14 and the name in general

We only want the era count in the file if we export less than the max blocks/slots per file, so it's better to make it optional.

A part of the code was also refactored to deduplicate some era id logic.


mattsse (approved):
cool, makes sense

PR #20293, opened 2025-12-11
fix(rpc): avoid duplicate precompile computation for last fork

Bashmunta:
When the next configured fork is also the last configured fork, eth_config now reuses the already built next fork configuration for last instead of recomputing precompiles for the same timestamp. This keeps the EIP-7910 semantics (next and last identical in this case) while avoiding an extra evm_for_block call and redundant precompile map construction.


mattsse (commented):
we can keep this as is

PR #20297, opened 2025-12-11
docs: fix incorrect API example in node-components.mdx

aso20455:
The example code in SDK node-components documentation showed a non-existent API that wouldn't compile. Fixed to match the actual builder pattern used in examples/custom-node-components.

Changes:
- Replaced fake `ctx.components_builder()` with real `EthereumNode::components()`
- Fixed `.evm()` to `.executor()` (correct method name)
- Removed closure-based builder syntax that doesn't exist
- Added `.with_add_ons()` which was missing
- Changed `.build().await` to `.launch().await`
- Added links to working examples in the repo

### DIFF
--- a/docs/vocs/docs/pages/sdk/node-components.mdx
+++ b/docs/vocs/docs/pages/sdk/node-components.mdx
@@ -53,43 +53,55 @@ Provides external API access to the node:
 
 ## Component Customization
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
fine, this is an improvement overall

PR #20299, opened 2025-12-11
fix(net): correct config builder doc comments

radik878:
The doc comments for Discv4ConfigBuilder::enable_eip868 and ::enforce_expiration_timestamps were swapped and described the wrong flags.


mattsse (approved):
fine

PR #20300, opened 2025-12-11
docs: fix misleading links

sashass1315:
The link to Database trait → leads to a CLI file, not to an abstraction.
The link to Transaction → leads to a file with database errors, not to transaction implementation.

corrected links: 
https://github.com/paradigmxyz/reth/blob/main/crates/storage/db-api/src/database.rs
https://github.com/paradigmxyz/reth/blob/main/crates/storage/db-api/src/transaction.rs

### DIFF
--- a/docs/design/database.md
+++ b/docs/design/database.md
@@ -2,8 +2,8 @@
 
 ## Abstractions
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
fine

PR #20308, opened 2025-12-11
fix(chain-state): correct balance deduction in test block builder

PivasDesant:
Fixed a bug in generate_random_block where balance was being deducted incorrectly. The total cost of all transactions was being subtracted for each transaction in the loop, so with 3 transactions the balance would drop by 9x instead of 3x.

Changed it to subtract single_tx_cost() per iteration, which matches how get_execution_outcome does it.


mattsse (approved):
this is technically a fix yes

PR #20309, opened 2025-12-11
feat(rpc): add transaction index to EVM error messages

reject-i:
Closes #20301


crates/rpc/rpc-eth-types/src/error/mod.rs
```diff
@@ -195,6 +195,14 @@ pub enum EthApiError {
...
         error: jsonrpsee_types::ErrorObject<'static>,
     },
+    /// Error that occurred during multi-transaction execution with transaction index context.
+    #[error("transaction execution failed (index {tx_index}): {error}")]
+    IndexedEvmError {
+        /// Transaction index where the error occurred
+        tx_index: usize,
+        /// The underlying error
+        error: Box<Self>,
+    },
```
mattsse: I believe the error message for this would be a bit messy, because this would chain error messages

mattsse (changes requested):
what we'd need is a helper wrapper type for EvmError that wraps it like

```
EthApiEvmErr<..> {
  inner: EvmError,
  index: usize
}
```

so that we can format the error nicely

crates/rpc/rpc-eth-types/src/error/api.rs
```diff
@@ -105,9 +108,17 @@ pub trait FromEvmError<Evm: ConfigureEvm>:
...
+        + FromEvmHalt<HaltReasonFor<Evm>>
+        + FromRevert
+        + From<EthApiError>,
     Evm: ConfigureEvm,
+    EthApiError: From<EthApiEvmErr<EvmErrorFor<Evm, ProviderError>>>,
 {
+    fn from_evm_err_at_index(err: EvmErrorFor<Evm, ProviderError>, tx_index: usize) -> Self {
+        let indexed = EthApiEvmErr::new(err, tx_index);
+        EthApiError::from(indexed).into()
+    }
```
klkvr: if it's only used once imo we should just do it in-place instead of changing this trait with extra non-evm related bound

PR #20310, opened 2025-12-11
docs: add test documentation for GenesisAccount conversion tests

PivasDesant:
Added doc comments to the test functions. Makes it clearer what each test is checking when converting GenesisAccount to TrieAccount.


mattsse (commented):
not worth documenting test code

PR #20313, opened 2025-12-11
refactor(rpc): remove unnecessary Debug bound

Fibonacci747:
The generic impl for MultiConsumerLruCache required K: Debug even though none of the methods or the Debug impl format the key. This bound provided no functional value but restricted the cache from being used with non-Debug keys. Dropping the Debug constraint loosens the API without affecting existing usages (which already use B256 as the key) and keeps the implementation consistent with the struct and Debug impl bounds (K: Hash + Eq).


mattsse (commented):
we cna keep this because might be useful

PR #20318, opened 2025-12-11
refactor(stages): cache hashed address in storage hashing loop

AJStonewee:
Noticed while reviewing the storage hashing code that compute `keccak256(address)` for every storage entry, even though `PlainStorageState` is a DupSort table where entries from the same address are consecutive.

Caches the last hashed address to avoid redundant keccak256 calls within each worker chunk (100 entries).

### DIFF
--- a/crates/stages/stages/src/stages/hashing_storage.rs
+++ b/crates/stages/stages/src/stages/hashing_storage.rs
@@ -102,7 +102,7 @@ where
                 // Spawn the hashing task onto the global rayon pool
                 rayon::spawn(move || {
                     // Cache hashed address since PlainStorageState is sorted by address
-                    let (mut last_addr, mut hashed_addr) = (Address::ZERO, B256::ZERO);
+                    let (mut last_addr, mut hashed_addr) = (Address::ZERO, keccak256(Address::Zero));
                     for (address, slot) in chunk {
                         if address != last_addr {
                             last_addr = address;
--- a/crates/stages/stages/src/stages/hashing_storage.rs
+++ b/crates/stages/stages/src/stages/hashing_storage.rs
@@ -102,7 +102,8 @@ where
                 // Spawn the hashing task onto the global rayon pool
                 rayon::spawn(move || {
                     // Cache hashed address since PlainStorageState is sorted by address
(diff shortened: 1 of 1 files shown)


crates/stages/stages/src/stages/hashing_storage.rs
```diff
@@ -101,9 +101,15 @@ where
                 let chunk = chunk.collect::<Result<Vec<_>, _>>()?;
                 // Spawn the hashing task onto the global rayon pool
                 rayon::spawn(move || {
+                    // Cache hashed address since PlainStorageState is sorted by address
+                    let (mut last_addr, mut hashed_addr) = (Address::ZERO, B256::ZERO);
```
mediocregopher: Even though it will basically never matter, might as well make this use the correct hash for zero address:

```suggestion
                    let (mut last_addr, mut hashed_addr) = (Address::ZERO, keccak256(Address::Zero));
```

mediocregopher (changes requested):
Good change, just one nitpick

crates/stages/stages/src/stages/hashing_storage.rs
AJStonewee: Yeah, thats right, thanks!

AJStonewee:
@mediocregopher @mattsse  Hi this has been hold for a whale, mind having a check?

mediocregopher (approved):

crates/stages/stages/src/stages/hashing_storage.rs
```diff
@@ -101,9 +101,16 @@ where
                 let chunk = chunk.collect::<Result<Vec<_>, _>>()?;
                 // Spawn the hashing task onto the global rayon pool
                 rayon::spawn(move || {
+                    // Cache hashed address since PlainStorageState is sorted by address
+                    let (mut last_addr, mut hashed_addr) =
+                        (Address::ZERO, keccak256(Address::ZERO));
```
mattsse, line 106: I think we could even make this a const, but doing this once isnt a big deal imo

mattsse (approved):
makes sense

PR #20319, opened 2025-12-11
docs: refresh repo layout crate lists

sashass1315:
The repo layout documentation had drifted from the actual crates structure.

Updated the crate lists in the Storage, Networking, Consensus, Sync, Payloads and Misc sections to match the current contents of the crates/directory. The goal is to make the high-level project layout accurate and discoverable again for contributors, without changing any runtime behavior.


mattsse (approved):
makes sense, ty!

PR #20321, opened 2025-12-12
docs: add missing RPC namespaces to JSON-RPC intro

eeemmmmmm:
The intro.mdx listed only 8 RPC namespaces but reth actually has 14. Added the missing ones: reth, ots, flashbots, miner, mev, testing. Also marked miner and testing as sensitive since they can modify node behavior or include arbitrary transactions.

Fixed the "all" option description to show the complete list.


mattsse (approved):
seems alright

PR #20322, opened 2025-12-12
fix(docs): document discv5 discovery port 9200

dennsikl:
The ports.mdx page lists all network ports used by reth but was missing the discv5 discovery port (9200/UDP).

This port is configurable via --discovery.v5.port and used when --enable-discv5-discovery is enabled.


mattsse (approved):
reasonable

PR #20323, opened 2025-12-12
chore(chain-state): remove EthPrimitives re-export

zeroprooff:
Closes the TODO in `reth-chain-state` - removes the temporary `EthPrimitives` re-export.

All consumers now import directly from `reth-ethereum-primitives`.


mattsse (commented):
we can keep this re-export

zeroprooff:
should I remove the todo comment then? if you reopen, I'll update

PR #20326, opened 2025-12-12
chore(storage): fix reference in comment

ANtutov:
The TxnManager documentation was referencing a non-existent method with_max_read_transaction_duration, which is misleading for users trying to enable read transaction timeouts. Updated the comment to point to the actual constructor new_with_max_read_transaction_duration so that the docs accurately reflect the public API and reduce confusion when configuring read-tx-timeouts.


mattsse (commented):
we can keep this as is

PR #20329, opened 2025-12-12
docs: clarify receipts vs log filter pruning semantics

Bashmunta:
The previous comments and docs described receipts and receipts_log_filter as if they were mutually exclusive, using “overrides” wording. In reality the implementation allows them to be combined: receipts defines the global receipts pruning policy, while receipts_log_filter can be layered on top to additionally retain receipts for specific contracts, as used in the full node configuration.

This change updates the PruneModes documentation comments and the pruning section in configuration.mdx to describe the actual combined behavior. No pruning logic or CLI flags are changed, only the semantics in comments and user-facing docs are clarified.

### DIFF
--- a/crates/prune/types/src/target.rs
+++ b/crates/prune/types/src/target.rs
@@ -56,8 +56,8 @@ pub struct PruneModes {
     /// Transaction Lookup pruning configuration.
     #[cfg_attr(any(test, feature = "serde"), serde(skip_serializing_if = "Option::is_none"))]
     pub transaction_lookup: Option<PruneMode>,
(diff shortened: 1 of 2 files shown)


joshieDo:
unless im missing something, this is incorrect.

receipts pruner always runs first and overrides the log one

https://github.com/paradigmxyz/reth/blob/134164954b9a7773c9590b91ba8bb78b982e8997/crates/prune/prune/src/segments/set.rs#L83-L89

https://github.com/paradigmxyz/reth/blob/134164954b9a7773c9590b91ba8bb78b982e8997/crates/prune/prune/src/segments/receipts.rs#L82-L93

PR #20333, opened 2025-12-12
feat: Add `TrieUpdatesSorted` and `HashedPostStateSorted` in all ExEx notifications

dhyaniarun1993:
~~Based on~~ includes changes from https://github.com/paradigmxyz/reth/pull/19355


emhane:
seems like needs rebase. probably easier if you simply cherry-pick the commits that are new to this pr (the ones that don't have a pr link in brackets at end), onto the base branch, and then force push @dhyaniarun1993

crates/exex/exex/src/wal/storage.rs
```diff
@@ -229,6 +273,58 @@ mod tests {
...
+        use std::io::Write;
+
+        let mut rng = generators::rng();
+
+        // Create a block with a transaction
+        let block = random_block(&mut rng, 0, Default::default()).try_recover()?;
+        let block_number = block.header().number();
+
+        // Create some non-empty trie updates and hashed state to ensure the 4-field format is used
+        let trie_updates = TrieUpdatesSorted::default();
```
mediocregopher: Can we include some nominal trie updates/state changes to make this fully end-to-end? If someone accidentally breaks either of those serialization formats we would want this new `.wal` file test to break.

dhyaniarun1993: Sure. I add some cases here.

crates/trie/common/src/updates.rs
```diff
@@ -1097,11 +1097,134 @@ pub mod serde_bincode_compat {
...
+    /// struct Data {
+    ///     #[serde_as(as = "serde_bincode_compat::updates::TrieUpdatesSorted")]
+    ///     trie_updates: TrieUpdatesSorted,
+    /// }
+    /// ```
+    #[derive(Debug, Serialize, Deserialize)]
+    pub struct TrieUpdatesSorted<'a> {
+        account_nodes: Cow<'a, [(Nibbles, Option<BranchNodeCompact>)]>,
+        storage_tries: B256Map<StorageTrieUpdatesSorted<'a>>,
+    }
```
emhane, line 1119: I think we want to create a new module (can be part of this file) called serde_bincode_compat, and feature gate this code behind the existing feature for this crate `serde-bincode-compat`. same for the hashed post state .

emhane (changes requested):
we want to

crates/trie/common/src/updates.rs
dhyaniarun1993: Yep, we are already doing that https://github.com/op-rs/op-reth/blob/8605dc3b378f5bbce808e16a0dc7c0f0d7650433/crates/trie/common/src/lib.rs#L72-L78

emhane (approved):

crates/exex/exex/src/wal/storage.rs
```diff
@@ -202,6 +211,39 @@ mod tests {
...
+            ExExNotification::ChainCommitted { new } => {
+                // Check blocks
+                assert_eq!(new.blocks().len(), 1);
+                let block_number = new.tip().header().number();
+
+                // The key thing is that this WAL was created with the new 4-field format
+                // (blocks, execution_outcome, trie_updates, hashed_state)
+                // Even if the maps are empty, they should be present in the serialized data
+                assert!(
+                    new.trie_updates().contains_key(&block_number),
```
mediocregopher: @dhyaniarun1993 I see that your encoding actual data into the WAL below, but we need to fully assert those contents here, not just assert that there is _some_ trie updates/hashed post state data for the block.

emhane: @dhyaniarun1993 is ooo until 5th jan @mediocregopher

dhyaniarun1993: Yep, thanks for pointing it out. I will push the fixes

dhyaniarun1993: I have pushed the changes.

mediocregopher (approved):

crates/exex/exex/src/wal/storage.rs
mediocregopher: Thanks @dhyaniarun1993 , looks good!

crates/evm/execution-types/src/chain.rs
```diff
@@ -452,7 +491,19 @@ pub(super) mod serde_bincode_compat {
...
+        #[serde(default)]
+        trie_updates: BTreeMap<
+            BlockNumber,
+            reth_trie_common::serde_bincode_compat::updates::TrieUpdatesSorted<'a>,
+        >,
+        #[serde(default)]
+        hashed_state: BTreeMap<
+            BlockNumber,
+            reth_trie_common::serde_bincode_compat::hashed_state::HashedPostStateSorted<'a>,
+        >,
```
mattsse, line 506: cool, this should do it

crates/exex/exex/src/wal/storage.rs
```diff
@@ -178,12 +178,22 @@ where
...
     use reth_testing_utils::generators::{self, random_block};
-    use std::{fs::File, sync::Arc};
+    use reth_trie_common::{
+        updates::{StorageTrieUpdates, TrieUpdates},
+        BranchNodeCompact, HashedPostState, HashedStorage, Nibbles,
+    };
+    use std::{collections::BTreeMap, fs::File, sync::Arc};
 
-    // wal with 1 block and tx
+    // wal with 1 block and tx (old 3-field format)
```
mattsse, line 196: nice

mattsse (approved):
makes sense, I think this breaking change won't be difficult to deal with for users

PR #20336, opened 2025-12-12
feat(trie): Proof Rewrite: Support partial proofs

### LINKED ISSUE #19516: Proof Calc Rewrite: Partial Proofs
### Describe the feature

The proof target should be modified to include a minimum path length. Any nodes in a proof which are less than that path length should not be calculated or returned.

The proptest should be modified such that the control calculation, which uses the legacy proof calculator, simply filters all nodes whose path is shorter than the minimum.

### Additional context

_No response_

mediocregopher:
The key change here is the addition of the `Target` type, which wraps a B256 key to hold also a `min_len`. When a node's path length is less than the `min_len` then we do not retain it for that target.

Given a set of Targets we then intelligently determine which sub-tries of the overall trie actually need to be computed, so we don't waste time computing portions of the trie which no target could possibly match anyway. The method for determining sub-tries is a bit complex, but is heavily documented in the code with examples; more docs/examples can be added if necessary.

The actual algorithms for building the trie from leaves+cached branches has not significantly changed, except to add boundary checks for the sub-tries.

Closes #19516


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -1068,63 +1110,61 @@ where
...
+    /// Calculates trie nodes and retains proofs for targetted nodes within a sub-trie. The
+    /// sub-trie's bounds are denoted by the `lower_bound` and `upper_bound` arguments,
+    /// `upper_bound` is exclusive, None indicates unbounded.
+    #[instrument(
+        target = TRACE_TARGET,
+        level = "trace",
+        skip_all,
+        fields(prefix=?sub_trie_targets.prefix),
+    )]
+    fn proof_subtrie<'a>(
```
mediocregopher, line 1117: `proof_subtrie` has essentially replaced the previous `proof_inner` implementation; `proof_inner` previously drove the calculation of the trie for the entire trie, now `proof_subtrie` drives the calculation of a single sub-trie using roughly the same logic. `proof_inner` instead now loops over each sub-trie and calls on `proof_subtrie` on each.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -198,21 +198,47 @@ where
...
+            // - key: 0xabc0, min_len: 2
+            // - key: 0xabc1, min_len: 1
+            // - key: 0xabc2, min_len: 4 <-- current
+            // - key: 0xabc3, min_len: 3
+            //
+            // When the branch node at 0xabc is visited it will be after the targets has iterated
+            // forward to 0xabc2 (because all children will have been visited already). At this
+            // point the target for 0xabc2 will not match the branch due to its prefix, but any of
+            // the other targets would, so we need to check those as well.
+            if lower.key.starts_with(path) {
```
yongkangc, line 224: i see, with `min_len` this basically allows `should_retain` to filter which nodes is sent for proof output?

i.e which nodes are already cached and what nodes are needed

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -1313,19 +1423,38 @@ enum TrieCursorState {
 }
 
 impl TrieCursorState {
+    /// Creates a [`Self::Unseeked`] based on an entry returned from the cursor itself.
```
yongkangc, line 1422: what does entry returned from the cursor itself mean here? since it does not look like we are not returning any entry from the cursor

> /// Creates a [`Self::Unseeked`] based on an entry returned from the cursor itself.

yongkangc (commented):
x

crates/trie/trie/src/proof_v2/target.rs
```diff
@@ -0,0 +1,332 @@
...
+        // Helper to create nibbles from hex string (each character is a nibble)
+        let nibbles = |hex: &str| -> Nibbles {
+            if hex.is_empty() {
+                return Nibbles::new();
+            }
+            format!("0x{}", hex).parse().expect("valid nibbles hex string")
+        };
+
+        // Test cases: (input_targets, expected_output)
+        // Expected output format: Vec<(lower_bound_hex, upper_bound_hex_opt, Vec<key_hex>)>
```
yongkangc:
```suggestion
        // Expected output format: Vec<(prefix_hex, Vec<key_hex>)>
```

nit: looking at line 299, and the data, is the comment outdated?

yongkangc (approved):
LGTM! logic wise it makes sense to me and its quite minimal changes compared to what i would have expected

crates/trie/trie/src/proof_v2/target.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// # Panics
+    ///
+    /// This method panics if `min_len` is greater than 64.
+    pub fn with_min_len(mut self, min_len: u8) -> Self {
+        debug_assert!(min_len <= 64);
+        self.min_len = min_len;
+        self
+    }
+
+    // A helper function for getting the largest prefix of the sub-trie which contains a particular
```
mediocregopher, line 33: This comment is out of date

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: Right, we can use this to not retain nodes which are already known. In addition we are using `min_len` to not bother constructing portions of the trie which are already known (or not needed) in the first place.

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: This comment is out of date, will fix

crates/trie/trie/src/proof_v2/target.rs
mediocregopher: Yup, good catch :pray:

crates/trie/trie/src/proof_v2/mod.rs
yongkangc: ahh i see, its pretty neat that this is the core and its simple 💯

mattsse (approved):
cool, all of this makes sense, even for me

PR #20339, opened 2025-12-12
perf(net): don't fetch or import recently mined transactions

klkvr:
Closes #20328

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -63,8 +63,8 @@ use reth_primitives_traits::SignedTransaction;
 use reth_tokio_util::EventStream;
 use reth_transaction_pool::{
     error::{PoolError, PoolResult},
-    AddedTransactionOutcome, GetPooledTransactionLimit, PoolTransaction, PropagateKind,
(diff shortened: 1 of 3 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -1650,6 +1681,14 @@ where
             |cmd| this.on_command(cmd)
         );
 
+        // Drain pool events to handle mined transactions.
+        // We only care about `Mined` events for cleanup purposes.
+        while let Poll::Ready(Some(event)) = this.pool_events.poll_next_unpin(cx) {
+            if let FullTransactionEvent::Mined { tx_hash, .. } = event {
+                this.on_mined_transaction(tx_hash);
+            }
+        }
```
mattsse, line 1690: this will be quite a few updates and def more expensive on the pool side because we dont have any low level listeners for this by default. 

imo it would be better to have a command for this on the handle that sends the newly mined txs in bulk

basically similar to the maintain task then sends the set of mined txs per new block once per block

mattsse (changes requested):

emmajam:

PR #20340, opened 2025-12-12
feat(storage): add method to check invariants on RocksDB tables

### LINKED ISSUE #20153: Add method to check invariants on RocksDB tables
### Describe the feature

Right now we make sure that static files have been properly committed, and in the right order, by using `check_consistency` and `ensure_invariants`:

https://github.com/paradigmxyz/reth/blob/179e1644ec72b9afb05a46367f2b55baf7de8cce/crates/storage/provider/src/providers/static_file/manager.rs#L991-L1025
https://github.com/paradigmxyz/reth/blob/179e1644ec72b9afb05a46367f2b55baf7de8cce/crates/storage/provider/src/providers/static_file/manager.rs#L1254

We should add a version of this for tables that we've moved to RocksDB

### Additional context

_No response_

fgimenez:
Closes #20153

Adds invariant checking for RocksDB tables, similar to the existing `check_consistency` and `ensure_invariants` methods for static files.

- Add `first()` and `last()` methods to `RocksDBProvider` for efficient table bounds checking
- Add `check_consistency()` method that validates RocksDB tables against MDBX stage checkpoints
- For `TransactionHashNumbers`: maps max tx_number -> block_number via `TransactionBlocks` to detect if RocksDB is behind the `TransactionLookup` checkpoint
- For `StoragesHistory`: checks if table is empty when `IndexStorageHistory` checkpoint indicates data should exist
- Returns an unwind target block number when inconsistencies are detected

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -58,6 +58,7 @@ impl RocksDBProvider {
     /// Checks invariants for the `TransactionHashNumbers` table.
     ///
     /// Returns a block number to unwind to if `RocksDB` is behind the checkpoint.
+    /// If `RocksDB` is ahead of the checkpoint, excess entries are pruned (healed).
     fn check_transaction_hash_numbers<Provider>(
         &self,
         provider: &Provider,
@@ -81,8 +82,9 @@ impl RocksDBProvider {
             Some((_hash, tx_number)) => {
                 // If checkpoint is 0 but we have data, that's inconsistent
                 if checkpoint == 0 && tx_number > 0 {
-                    // RocksDB has data but stage hasn't run - need to clear
-                    return Ok(Some(0));
+                    // RocksDB has data but stage hasn't run - need to clear all
+                    self.prune_transaction_hash_numbers_above(0)?;
+                    return Ok(None);
                 }
 
                 // Map tx_number to block_number using TransactionBlocks table
@@ -92,9 +94,10 @@ impl RocksDBProvider {
                 // Seek to the first entry where key >= tx_number
                 if let Some((_, rocks_block)) = cursor.seek(tx_number)? {
                     // rocks_block is the block containing this transaction
-                    // If checkpoint block > rocks_block, RocksDB is behind
                     if checkpoint > rocks_block {
-                        tracing::info!(
+                        // RocksDB is behind checkpoint - this indicates something went wrong
+                        // since MDBX checkpoint is committed last when appending.
+                        tracing::warn!(
                             target: "reth::providers::rocksdb",
                             rocks_max_tx = tx_number,
                             rocks_block,
@@ -102,6 +105,39 @@ impl RocksDBProvider {
                             "TransactionHashNumbers behind checkpoint, unwind needed"
                         );
                         return Ok(Some(rocks_block));
+                    } else if checkpoint < rocks_block {
+                        // RocksDB is ahead of checkpoint - this is the common case during
+                        // recovery from a crash during unwinding. MDBX checkpoint was
+                        // committed first, so we need to prune excess RocksDB data.
+                        //
+                        // Find the max tx_number for the checkpoint block by iterating
+                        // TransactionBlocks to find the entry where block_number == checkpoint
+                        let mut max_tx_for_checkpoint = 0u64;
+                        let mut cursor_walk =
+                            provider.tx_ref().cursor_read::<tables::TransactionBlocks>()?;
+                        let mut walker = cursor_walk.walk(Some(0))?;
+                        while let Some((tx_num, block_num)) = walker.next().transpose()? {
+                            if block_num == checkpoint {
+                                max_tx_for_checkpoint = tx_num;
+                                break;
+                            } else if block_num > checkpoint {
+                                // We've passed the checkpoint block, use the previous tx as max
+                                break;
+                            }
+                            max_tx_for_checkpoint = tx_num;
+                        }
+
+                        if max_tx_for_checkpoint > 0 {
+                            tracing::info!(
+                                target: "reth::providers::rocksdb",
+                                rocks_max_tx = tx_number,
+                                rocks_block,
+                                checkpoint,
+                                max_tx_for_checkpoint,
+                                "TransactionHashNumbers ahead of checkpoint, pruning excess data"
+                            );
+                            self.prune_transaction_hash_numbers_above(max_tx_for_checkpoint)?;
+                        }
                     }
                 }
 
@@ -119,9 +155,46 @@ impl RocksDBProvider {
         }
     }
 
+    /// Prunes `TransactionHashNumbers` entries where `tx_number` > `max_tx_number`.
+    ///
+    /// Since `TransactionHashNumbers` is keyed by `TxHash`, we must iterate all entries
+    /// and delete those whose value exceeds the threshold.
+    fn prune_transaction_hash_numbers_above(&self, max_tx_number: u64) -> ProviderResult<()> {
+        let mut to_delete = Vec::new();
+
+        // Iterate all entries and collect those to delete
+        for result in self.iter::<tables::TransactionHashNumbers>()? {
+            let (key, tx_number) = result?;
+
+            if tx_number > max_tx_number {
+                to_delete.push(key);
+            }
+        }
+
+        let deleted_count = to_delete.len();
+        if deleted_count > 0 {
+            tracing::info!(
+                target: "reth::providers::rocksdb",
+                deleted_count,
+                max_tx_number,
+                "Pruning TransactionHashNumbers entries"
+            );
+
+            self.write_batch(|batch| {
+                for key in to_delete {
+                    batch.delete::<tables::TransactionHashNumbers>(key)?;
+                }
+                Ok(())
+            })?;
+        }
+
+        Ok(())
+    }
+
     /// Checks invariants for the `StoragesHistory` table.
     ///
     /// Returns a block number to unwind to if `RocksDB` is behind the checkpoint.
+    /// If `RocksDB` is ahead of the checkpoint, excess entries are pruned (healed).
     fn check_storages_history<Provider>(
         &self,
         provider: &Provider,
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,413 @@
...
+                    if checkpoint > rocks_block {
+                        tracing::info!(
+                            target: "reth::providers::rocksdb",
+                            rocks_max_tx = tx_number,
+                            rocks_block,
+                            checkpoint,
+                            "TransactionHashNumbers behind checkpoint, unwind needed"
+                        );
+                        return Ok(Some(rocks_block));
+                    }
```
joshieDo: either here or somewhere else, we should cover the other way around which is the most common occurrence
`else if checkpoint < rocks_block { self.delete.... } ` 

when appending data, mdbx stage checkpoint will be the last thing to be committed to disk :
https://github.com/paradigmxyz/reth/blob/64909d33e6b7ab60774e37f5508fb5ad17f41897/crates/storage/provider/src/providers/database/provider.rs#L3176-L3178

when unwinding storage (eg pipeline unwind/engine deep reorg), mdbx stage checkpoint will be the first thing to be commited to disk:
https://github.com/paradigmxyz/reth/blob/64909d33e6b7ab60774e37f5508fb5ad17f41897/crates/storage/provider/src/providers/database/provider.rs#L3169-L3175

joshieDo: > checkpoint > rocks_block

for this to happen, something has gone terrible wrong if we follow the above rules

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,413 @@
...
+
+        // Get the last entry from RocksDB TransactionHashNumbers
+        let rocks_last = self.last::<tables::TransactionHashNumbers>()?;
+
+        match rocks_last {
+            Some((_hash, tx_number)) => {
+                // If checkpoint is 0 but we have data, that's inconsistent
+                if checkpoint == 0 && tx_number > 0 {
+                    // RocksDB has data but stage hasn't run - need to clear
+                    return Ok(Some(0));
```
joshieDo: instead of issuing an unwind we could just prune it here

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: good point, added changes to remove excess rocksdb entries when it is ahead of mdbx, ptal

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: yep, done

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// If `RocksDB` is ahead of the checkpoint, excess entries are pruned (healed).
+    fn check_transaction_hash_numbers<Provider>(
+        &self,
+        provider: &Provider,
+    ) -> ProviderResult<Option<BlockNumber>>
+    where
+        Provider: DBProvider + StageCheckpointReader,
+    {
+        use reth_db::cursor::DbCursorRO;
+        use reth_db_api::transaction::DbTx;
```
joshieDo: nit, move them up

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,688 @@
...
+
+    /// Prunes `TransactionHashNumbers` entries where `tx_number` > `max_tx_number`.
+    ///
+    /// Since `TransactionHashNumbers` is keyed by `TxHash`, we must iterate all entries
+    /// and delete those whose value exceeds the threshold.
+    fn prune_transaction_hash_numbers_above(&self, max_tx_number: u64) -> ProviderResult<()> {
+        let mut to_delete = Vec::new();
+
+        // Iterate all entries and collect those to delete
+        for result in self.iter::<tables::TransactionHashNumbers>()? {
```
joshieDo: i dont think it's feasible to iterate over all rows. we need to fetch all possible transactions above the checkpoint, hash them and query the entries by key. We have a `TransactionLookup` pruner that does something of the kind.

kinda points to static files needing to be ahead of rocksdb

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,688 @@
...
+    ///
+    /// For `StoragesHistory`, the key contains `highest_block_number`, so we can iterate
+    /// and delete entries where `key.sharded_key.highest_block_number > max_block`.
+    fn prune_storages_history_above(&self, max_block: BlockNumber) -> ProviderResult<()> {
+        use reth_db_api::models::storage_sharded_key::StorageShardedKey;
+
+        let mut to_delete: Vec<StorageShardedKey> = Vec::new();
+
+        // Iterate all entries and collect those to delete
+        for result in self.iter::<tables::StoragesHistory>()? {
```
joshieDo, line 304: similarly we can't iterate like this. `crates/prune/prune/src/segments/user/storage_history.rs` we should try to emulate what pruner does and/or aggregate the pruning logic

joshieDo: but, we actually need storage changesets in static files first for this.

the way entries are selected are by looking at storage changesets for the block range. however, since StorageChangeSets are still in DB, and it is behind alongside the checkpoint, we cant really know which keys we need to compose to prune here (apart from iterating like this)

context:
https://github.com/paradigmxyz/reth/blob/948966781425b6ee6d9fec7e2a1078d58eacec04/crates/prune/prune/src/segments/user/storage_history.rs#L82-L113

joshieDo: id say lets wait for `storagechangesets` on static files before we proceed with `StoragesHistory`, OR
focus on the `accounthistory` since we almost have `accountchangests` on static files OR
go along with this but dont enable it until we have `storagechangesets` + create issue to fix this

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: sure done

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: makes sense, introduced `prune_transaction_hash_numbers_in_range` now instead of iterating all rows in TransactionHashNumbers to find entries to delete:
  - fetches transactions from the provider by tx range (transactions_by_tx_range)
  - computes their hashes in parallel using rayon
  - deletes the entries from RocksDB by key

ptal

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: ok, created issue here https://github.com/paradigmxyz/reth/issues/20417 and added a TODO refering to it

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,870 @@
...
+    /// and deletes the corresponding entries from `RocksDB` by key. This approach is more
+    /// scalable than iterating all rows because it only processes the transactions that
+    /// need to be pruned.
+    ///
+    /// # Requirements
+    ///
+    /// The provider must be able to supply transaction data (typically from static files)
+    /// so that transaction hashes can be computed. This implies that static files should
+    /// be ahead of or in sync with `RocksDB`.
+    pub fn prune_transaction_hash_numbers_in_range<Provider>(
```
joshieDo: wondering if we could have this on the EitherWriter instead? that way it could be used here and in the pruner

joshieDo: same for the prune storage history

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,870 @@
...
+                        return Ok(Some(rocks_block));
+                    } else if checkpoint < rocks_block {
+                        // RocksDB is ahead of checkpoint - this is the common case during
+                        // recovery from a crash during unwinding. MDBX checkpoint was
+                        // committed first, so we need to prune excess RocksDB data.
+                        //
+                        // Find the max tx_number for the checkpoint block by iterating
+                        // TransactionBlocks to find the entry where block_number == checkpoint
+                        let mut max_tx_for_checkpoint = 0u64;
+                        let mut cursor_walk =
```
joshieDo: Given the order of storage commits, static files is at the very least at the same height as rocksdb. So we need to use that to find the range of transactions we need to pull.


* find out the last tx num in MDBX (can query for the last on `TransactionBlocks` or `BlockBodyIndices`)
* find out the highest tx num in static files (eg. `static_file_provider()
            .get_highest_static_file_tx(StaticFileSegment::Transactions))

`mdbx_last_tx_num+1..=highest_tx`

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -0,0 +1,870 @@
...
+        // Get the TransactionLookup stage checkpoint
+        let checkpoint = provider
+            .get_stage_checkpoint(StageId::TransactionLookup)?
+            .map(|cp| cp.block_number)
+            .unwrap_or(0);
+
+        // Find the max tx_number in RocksDB by iterating all entries.
+        // Note: We can't use `last()` because that returns the entry with the largest
+        // hash key, not the entry with the largest tx_number value.
+        let max_tx = self
```
joshieDo: cant do this, we need to use static files: https://github.com/paradigmxyz/reth/pull/20340/files#r2625106003

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: very good point, done ptal

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: makes sense, done

crates/storage/provider/src/providers/rocksdb/invariants.rs
fgimenez: ok, using static files now

crates/storage/provider/src/either_writer.rs
```diff
@@ -418,6 +418,158 @@ where
...
+where
+    CURSOR: DbCursorRW<tables::TransactionHashNumbers> + DbCursorRO<tables::TransactionHashNumbers>,
+{
+    /// Prunes `TransactionHashNumbers` entries for the given transaction hashes.
+    ///
+    /// This deletes entries by seeking to each hash key and deleting it.
+    /// Returns the number of entries deleted.
+    pub fn prune_transaction_hash_numbers(
+        &mut self,
+        hashes: impl IntoIterator<Item = alloy_primitives::B256>,
```
joshieDo: nice

crates/storage/provider/src/either_writer.rs
```diff
@@ -418,6 +418,158 @@ where
...
+    /// Prunes `StoragesHistory` entries where `highest_block_number > max_block` (`RocksDB` only).
+    ///
+    /// For entries with `highest_block_number == u64::MAX` (sentinel), they are only
+    /// deleted if `max_block == 0` (clearing all).
+    ///
+    /// This variant is for use when the writer is known to be `RocksDB` at compile time.
+    /// Returns the number of entries deleted.
+    pub fn prune_storages_history_above_rocksdb(
+        &mut self,
+        max_block: BlockNumber,
```
joshieDo: can we leave a TODO/issue (maybe it already exists) that we need to revisit this since this will iterate the whole table and we shouldnt

crates/storage/provider/src/either_writer.rs
```diff
@@ -418,6 +418,158 @@ where
...
 
+impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N>
+where
+    CURSOR: DbCursorRW<tables::TransactionHashNumbers> + DbCursorRO<tables::TransactionHashNumbers>,
+{
+    /// Prunes `TransactionHashNumbers` entries for the given transaction hashes.
+    ///
+    /// This deletes entries by seeking to each hash key and deleting it.
+    /// Returns the number of entries deleted.
+    pub fn prune_transaction_hash_numbers(
```
joshieDo: i dont think we need the `_rocksdb` variant of this method?

crates/storage/provider/src/either_writer.rs
joshieDo: similarly, we shouldnt need the `_rocksdb` variant

fgimenez: sure, added TODO pointing to https://github.com/paradigmxyz/reth/issues/20417 and removed the whole `impl EitherWriter<'_, (), ()>` that defined `prune_transaction_hash_numbers_rocksdb` and `prune_storages_history_above_rocksdb`

crates/storage/provider/src/either_writer.rs
fgimenez: yep done

joshieDo (approved):

PR #20341, opened 2025-12-12
fix: only collect already tracked accounts

mattsse:
this would create a new entry for every changed account.

however we only need to track accounts for which we already have an entry because every unique account in the pool has an entry.
if it's missing then there's nothing to update


joshieDo (approved):

klkvr (approved):

yongkangc (approved):
Ah nice

PR #20344, opened 2025-12-12
chore: remove unused CustomContext alias from custom-node example

lisenokdonbassenok:
The CustomContext type alias in the custom-node example was unused and referenced OpTransaction<PaymentTxEnv>, a combination that is not used anywhere else in the codebase. The actual EVM context for this example is provided via OpContext<DB> through the EvmFactory implementation and consumed via EvmContextFor.


mattsse (commented):
we can keep this as a template

PR #20348, opened 2025-12-13
feat: Check CL/Reth capability compatibility

tonisives:
## Problem

As discussed on Telegram on the 4th of December:

During Fusaka fork, at least 2 operators updated Reth but forgot to update the CL. Reth and CL continued to work fine.

After the fork went live, Reth stopped producing blocks. This is because CL wasn't updated together with Reth.

## Goal

Warn operators when

- CL doesn't support new methods in Reth
- Reth doesn't support new methods in CL

## Approach

Trigger a console warning when CL calls engine_exchangeCapabilities - at startup or after Reth restart.

## Testing

I tested by simulating `engine_exchangeCapabilities` request from CL with 4 variants:

1. CL is missing capabilities compared to Reth

```
2025-12-13T05:06:35.391183Z  WARN Reth supports Engine API methods that CL doesn't. Consider upgrading your consensus client. missing=["engine_getBlobsV1", "engine_getBlobsV2", "engine_getClientVersionV1", "engine_getPayloadBodiesByHashV1", "engine_getPayloadBodiesByRangeV1", "engine_getPayloadV4", "engine_getPayloadV5", "engine_newPayloadV4"]
```

2. CL has more capabilities than Reth

```
2025-12-13T05:03:12.867856Z  WARN CL supports Engine API methods that Reth doesn't. Consider upgrading Reth. missing=["engine_getPayloadV6", "engine_newPayloadV5"]
```

3. CL and Reth have matching capabilities

```
no warning
```

4. With current versions (1.9.3 and Lighhouse 8.0.1) 
```
no warnings and syncing starts
```

I only tested with curl, because we cannot sync old CL versions with reduced capabilities.

This is the query for variant 3

```sh
curl -X POST http://127.0.0.1:8551 \
  -H "Authorization: Bearer $TOKEN" \
  -d '{
    "jsonrpc": "2.0",
    "method": "engine_exchangeCapabilities",
    "params": [[
      "engine_forkchoiceUpdatedV1",
      "engine_forkchoiceUpdatedV2",
      "engine_forkchoiceUpdatedV3",
      "engine_getClientVersionV1",
      "engine_getPayloadV1",
      "engine_getPayloadV2",
      "engine_getPayloadV3",
      "engine_getPayloadV4",
      "engine_getPayloadV5",
      "engine_newPayloadV1",
      "engine_newPayloadV2",
      "engine_newPayloadV3",
      "engine_newPayloadV4",
      "engine_getPayloadBodiesByHashV1",
      "engine_getPayloadBodiesByRangeV1",
      "engine_getBlobsV1",
      "engine_getBlobsV2"
    ]],
    "id": 1
  }'
```

## Precedence

There is precedence for adding this in other clients:
Nethermind: https://github.com/NethermindEth/nethermind/blob/c2f1fa47a9beec195800a12fdf1f11d56115cdd0/src/Nethermind/Nethermind.Merge.Plugin/Handlers/ExchangeCapabilitiesHandler.cs#L33
Erigon: https://github.com/erigontech/erigon/blob/e75dac829f8f54da34f8df10029d1b79dda068bc/execution/engineapi/engine_api_methods.go#L184
Missing in geth: https://github.com/ethereum/go-ethereum/blob/e63e37be5e48a7869abf3f17e2a8cfff1fbfbcbf/eth/catalyst/api.go#L981

## References

- Engine api spec: https://github.com/ethereum/execution-apis/tree/main/src/engine
- CL fork schedule method: http://localhost:5052/eth/v1/config/fork_schedule
  since Engine API is unidirectional, then we shouldn't use it.


crates/rpc/rpc-engine-api/src/capabilities.rs
```diff
@@ -21,43 +26,62 @@ pub const CAPABILITIES: &[&str] = &[
...
-    pub fn new(capabilities: impl IntoIterator<Item: Into<String>>) -> Self {
+    /// Creates from an iterator of capability strings.
+    pub fn new(capabilities: impl IntoIterator<Item = impl Into<String>>) -> Self {
         Self { inner: capabilities.into_iter().map(Into::into).collect() }
     }
 
-    /// Returns the list of all supported Engine capabilities for Prague spec.
-    fn prague() -> Self {
-        Self { inner: CAPABILITIES.iter().copied().map(str::to_owned).collect() }
-    }
```
mattsse, line 42: removing this makes sense

crates/rpc/rpc-engine-api/src/capabilities.rs
```diff
@@ -21,43 +26,62 @@ pub const CAPABILITIES: &[&str] = &[
...
+    }
+}
+
+/// Logs warnings if CL and EL capabilities don't match.
+///
+/// Called during `engine_exchangeCapabilities` to warn operators about
+/// version mismatches between the consensus layer and execution layer.
+pub fn log_capability_mismatches(cl_capabilities: &[String], el_capabilities: &EngineCapabilities) {
+    let cl_set: HashSet<&str> = cl_capabilities.iter().map(|s| s.as_str()).collect();
+    let el_set: HashSet<&str> = el_capabilities.inner.iter().map(|s| s.as_str()).collect();
```
mattsse: I think we don't strictly need this map+collect because this is already a set

and can do

```
    // CL has methods EL doesn't support
      let mut el_missing: Vec<_> = cl_capabilities
          .iter()
          .filter(|cap| !el_set.contains(cap.as_str()))
          .map(String::as_str)
          .collect();
```

mattsse (changes requested):
ty, this is great, I have one tiny nit.

I believe this is okay and should also work for opstack, but will double check

crates/rpc/rpc-engine-api/src/capabilities.rs
tonisives: Hi, thanks for the review. I made these changes that remove the hash allocation. The second one is still necessary.
Since the code is only ran once on startup, then effects are negligible. If it were up to me, then I think the initial  approach with `.difference()` looks cleaner.

I tested the new approach and it works the same and the first one. So both are ok with me

mattsse (approved):

PR #20352, opened 2025-12-13
docs: fix --color auto option description

eeemmmmmm:
Fix incorrect description for `--color auto` CLI option. The documentation incorrectly stated "Colors on" instead of "Auto-detect" in several CLI reference pages.


mattsse (approved):
fine

Issue #20327, opened 2025-12-12
Make blockstream in debugclient configurable


mattsse:
### Describe the feature

for debug purposes we use

https://github.com/paradigmxyz/reth/blob/ed3a8a03d59bd6db62972f4c87b608bc217b0b29/crates/consensus/debug-client/src/providers/rpc.rs#L14-L18

that subscribes to blocks from a node, this is currently used by all `DebugNode` via:

https://github.com/paradigmxyz/reth/blob/ed3a8a03d59bd6db62972f4c87b608bc217b0b29/crates/node/builder/src/launch/debug.rs#L57-L66

ideally this is can be configured, so we can change the debug trait so that we replace the `rpc_to_primitive_block` with something that returns 

https://github.com/paradigmxyz/reth/blob/ed3a8a03d59bd6db62972f4c87b608bc217b0b29/crates/consensus/debug-client/src/client.rs#L16-L16

instead, then the default impl would just return something like `-> impl BlockProvider { RpcBlockProvider }` in the trait.

@0xKarl98

PR #20354, opened 2025-12-13
feat: make blockstream in debugclient configurable

0xKarl98:
This closes issue #20327

what i did : 

- Made the debug blockstream configurable via DebugNode::block_provider, adding DebugBlockProviderContext/DebugBlockProviderFuture and a default rpc_block_from_response fast‑path (type match → no serde round‑trip, with JSON fallback).

-  Standardized dev payload builder selection with a single DevBuilderConfig enum (factory / builder / default+map), removing multiple option fields.

- Centralized consensus‑client spawning in spawn_consensus_client and unified factory vs default paths to avoid duplicated setup.

- Switched to a runtime‑configurable provider handle (Arc<dyn BlockProvider>) and a tiny adapter so we can swap RPC/Etherscan/custom providers without extra branching.

cc @mattsse  @klkvr


0xKarl98:
@mattsse  @klkvr  pending

crates/node/builder/src/launch/debug.rs
```diff
@@ -111,19 +289,21 @@ impl<L> DebugNodeLauncher<L> {
...
     inner: L,
     target: Target,
     local_payload_attributes_builder:
         Option<Box<dyn PayloadAttributesBuilder<PayloadAttrTy<N::Types>, HeaderTy<N::Types>>>>,
-    map_attributes:
-        Option<Box<dyn Fn(PayloadAttrTy<N::Types>) -> PayloadAttrTy<N::Types> + Send + Sync>>,
+    map_attributes: Option<PayloadAttrMapper<N>>,
+    rpc_consensus_convert: Option<RpcConsensusConvert<N>>,
+    debug_block_provider_factory: Option<DebugBlockProviderFactory<N, AddOns>>,
 }
```
klkvr: I think we can just extend this with 
```rust
debug_consensus_client: Option<Box<dyn BlockProvider<Block = BlockTy<N::Types>>>>
```

that can be configured on the future directly so that you can do

```rust
.launch_with_debug_capabilities()
.with_debug_consensus(client)
```

for making `BlockProvider` dyn-compatible we can just change all methods to return `Pin<Box<dyn Future>>`

the configured block provider should take precedence over CLI-configured values

klkvr (changes requested):
I think we should be able to do this in a less invasive way

crates/node/builder/src/launch/debug.rs
0xKarl98: To make those methods , we have been using async_traits which is equal to Pin<Box<dyn Future>> 

Besides that , have addressed your suggestions

mattsse (commented):
slightly simplified in #21345

PR #20355, opened 2025-12-13
docs: fix beacon port comments

Lorsmirq0cB:
Changed `port to listen on` to `port to connect to` in beacon API examples, these examples act as HTTP clients connecting to beacon nodes.


mattsse (commented):
this is just an example and we can keep as is

PR #20357, opened 2025-12-13
feat(events): add hardfork activation event printout

futreall:
Closes https://github.com/paradigmxyz/reth/issues/20350


crates/node/events/src/node.rs
```diff
@@ -43,12 +73,17 @@ struct NodeState {
     finalized_block_hash: Option<B256>,
     /// The time when we last logged a status message
     last_status_log_time: Option<u64>,
+    /// Information about hardforks for tracking activation.
+    hardforks: Arc<[HardforkInfo]>,
```
mattsse: I think we can do a simple Vec for this because this isnt shared

crates/node/events/src/node.rs
```diff
@@ -58,13 +93,35 @@ impl NodeState {
...
     fn num_connected_peers(&self) -> usize {
         self.peers_info.as_ref().map(|info| info.num_connected_peers()).unwrap_or_default()
     }
 
+    /// Checks if any hardfork has become active and logs it.
+    fn check_hardfork_activation(&mut self, block_number: u64, timestamp: u64) {
+        for fork_info in self.hardforks.iter() {
+            if self.activated_forks.contains(fork_info.name) {
+                continue;
+            }
```
mattsse: I see this finds the first not activated fork and the checks if it is active at the current block?

can we document this as well?

mattsse (changes requested):
this looks good, some nits

crates/node/events/src/node.rs
futreall: yep, will do

crates/node/events/src/node.rs
```diff
@@ -58,13 +92,36 @@ impl NodeState {
...
+
+            if fork_info.is_active(block_number, timestamp) {
+                info!(
+                    target: "reth::cli",
+                    hardfork = %fork_info.name,
+                    block_number,
+                    timestamp,
+                    "Hardfork activated"
+                );
+                self.activated_forks.insert(fork_info.name);
```
mattsse: this will likely print multiple times on launch.

I think we can make this a bit smarter here by just tracking the timestamp of the current active hardfork and the next one. we can also use the current system time to check activation

I'm okay with jus tracking timestamp based hardforks

mattsse (changes requested):

crates/node/events/src/node.rs
futreall: done, now only tracking pending timestamp-based hardforks and checking activation against system time. already active forks are filtered out at startup

PR #20358, opened 2025-12-13
chore(rpc): use `InvalidTxError` trait to match gas limit errors

mablr:
Replace `EthApiError` helpers `is_gas_too_high`, `is_gas_too_high`, with alloy-evm's `InvalidTxError` trait fns.


crates/rpc/rpc-eth-types/src/error/api.rs
```diff
@@ -56,26 +56,6 @@ pub trait AsEthApiError {
...
-
-    /// Returns `true` if error is
-    /// [`RpcInvalidTransactionError::GasTooLow`].
-    fn is_gas_too_low(&self) -> bool {
-        if let Some(err) = self.as_err() {
-            return err.is_gas_too_low()
-        }
-
-        false
-    }
```
mattsse, line 77: these are breaking, and it doesnt hurt to keep them, s I'd like to keep them

mattsse (changes requested):

crates/rpc/rpc-eth-types/src/error/api.rs
mablr: okay

PR #20361, opened 2025-12-14
fix(node): log forkchoice updates when only head_block_hash changes

reallesee:
The handler was only checking safe_block_hash and finalized_block_hash for logging, but head_block_hash was being updated without being checked. 

Added a check for head_block_hash changes when safe and finalized haven't changed.
This preserves the original behavior (log when both safe and finalized change) while also catching head-only updates, avoiding log spam when all three change together.

### DIFF
--- a/crates/node/events/src/node.rs
+++ b/crates/node/events/src/node.rs
@@ -215,8 +215,13 @@ impl NodeState {
             ConsensusEngineEvent::ForkchoiceUpdated(state, status) => {
                 let ForkchoiceState { head_block_hash, safe_block_hash, finalized_block_hash } =
                     state;
-                if self.safe_block_hash != Some(safe_block_hash) &&
-                    self.finalized_block_hash != Some(finalized_block_hash)
+                // Log when both safe and finalized change (significant event), or when only head
+                // changes.
+                if (self.safe_block_hash != Some(safe_block_hash) &&
+                    self.finalized_block_hash != Some(finalized_block_hash)) ||
+                    (self.head_block_hash != Some(head_block_hash) &&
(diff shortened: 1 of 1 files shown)


crates/node/events/src/node.rs
```diff
@@ -215,8 +215,13 @@ impl NodeState {
             ConsensusEngineEvent::ForkchoiceUpdated(state, status) => {
                 let ForkchoiceState { head_block_hash, safe_block_hash, finalized_block_hash } =
                     state;
-                if self.safe_block_hash != Some(safe_block_hash) &&
-                    self.finalized_block_hash != Some(finalized_block_hash)
+                // Log when both safe and finalized change (significant event), or when only head
+                // changes.
+                if (self.safe_block_hash != Some(safe_block_hash) &&
+                    self.finalized_block_hash != Some(finalized_block_hash)) ||
+                    (self.head_block_hash != Some(head_block_hash) &&
```
mattsse: I think this kinda makes sense

crates/node/events/src/node.rs
```diff
@@ -215,8 +215,13 @@ impl NodeState {
...
+                        self.safe_block_hash == Some(safe_block_hash) &&
+                        self.finalized_block_hash == Some(finalized_block_hash))
```
mattsse: but we dont need these checks really

mattsse (changes requested):

mattsse (approved):
lgtm pending @shekhirin @Rjected

mattsse (commented):
we can keep this as is actually because we log on new head anyway

PR #20367, opened 2025-12-15
chore(rpc): fix misleading link and comment

Snezhkko:
the generic get_payload_v4 helper referenced the engine_newPayloadV4 section instead of the engine_getPayloadV4 section
and
handler for fork_choice_updated_v3 was still labeled as handling engine_forkchoiceUpdatedV2, which is misleading when navigating the code against the execution-apis spec


mattsse (approved):
fine

PR #20372, opened 2025-12-15
feat(cli): limit number of blocks to re-execute

shekhirin:
Allows to limit the number of blocks to re-execute via `--total-blocks` argument and the number of blocks in a single consecutive batch via `--batch-size` argument. This will split the whole set of `from..=to` into chunks of equal size, which allows to cover the provided range via selecting only a subset of blocks to re-execute.

---

Recommend to review with whitespaces hidden: https://github.com/paradigmxyz/reth/pull/20372/changes?w=1

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -15,10 +15,13 @@ use reth_evm::{execute::Executor, ConfigureEvm};
 use reth_primitives_traits::{format_gas_throughput, BlockBody, GotExpected};
 use reth_provider::{
     BlockNumReader, BlockReader, ChainSpecProvider, DatabaseProviderFactory, ReceiptProvider,
-    TransactionVariant,
+    StaticFileProviderFactory, TransactionVariant,
 };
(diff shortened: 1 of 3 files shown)


crates/cli/commands/src/re_execute.rs
```diff
@@ -84,12 +95,63 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
         };
 
-        let total_blocks = max_block - min_block;
-        let total_gas = calculate_gas_used_from_headers(
-            &provider_factory.static_file_provider(),
-            min_block..=max_block,
-        )?;
-        let blocks_per_task = total_blocks / self.num_tasks;
+        let full_range_size = max_block - min_block;
+        let (blocks_to_execute, chunk_size) = if let Some(total_blocks) = self.total_blocks {
```
mattsse, line 99: this is very hard to read, unclear what this does really or if useful

shekhirin:
Not needed

PR #20377, opened 2025-12-15
feat(storage): use RocksDBBatch in EitherWriter and related modules

### LINKED ISSUE #20346: Rocksdb: Switch to Write batch for writes
we have 3 options. writebatch / optimistictx and tx. Currently we are going with TransactionDB for writes. 

We want to have WriteBatch for writes, and TransactionDB for reads, assuming we don't run into any situations where we actually do use the read part of the write tx

However we need two variants, similar to mdbx :

read-write
WriteBatch for writes. we dont need isolation here since there's no other writer that can mess with our view. We might need to use OptimisticTransaction IF we want to read through our writes (WriteBatch does not support reads). But we probably don't ? Also, if we ever move to concurrent writes then yes we need [optimistic]tx. Should be easily swappable hopefully.

read-only
we need to either have a transaction or create a snapshot (im not entirely sure on the differences)

looks like the snapshot is explained here, it seems to ensure that a tx where there are write conflicts fails at commit rather than insert time:
https://github.com/facebook/rocksdb/wiki/Transactions#setting-a-snapshot

I think we want to create a transaction & not worry about snapshots for now

yongkangc:
Closes #20346 by using RocksDB history writes to batch-only instead of TransactionDB

  **Changes**
  - Use batch-only path for history writers (no RYW needed)
  - Add batch tests

  **Performance — Batch writes are faster**
  - WAL on: batch ~10–18% faster at 1k/10k writes.
  - WAL off: batch ~3–13% faster or parity.
  - Artifacts:  https://gist.github.com/yongkangc/cd8486f54457fc7ad5ce4f1b8798e8ca.

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -208,24 +208,6 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
             Self::RocksDB(_) => Err(ProviderError::UnsupportedProvider),
         }
     }
-
-    /// Commits `RocksDB` writes if this is a `RocksDB` writer.
-    ///
-    /// For [`Self::Database`] and [`Self::StaticFile`], this is a no-op as they use
-    /// different commit patterns (MDBX transaction commit, static file sync).
-    ///
-    /// # Commit Order
-    ///
-    /// Call this AFTER the outer MDBX transaction commits successfully. This ensures
-    /// that if `RocksDB` commit fails, the primary data (MDBX) is still intact and
-    /// the `RocksDB` data (which is derived) can be rebuilt.
-    #[cfg(all(unix, feature = "rocksdb"))]
-    pub fn commit(self) -> ProviderResult<()> {
-        match self {
-            Self::Database(_) | Self::StaticFile(_) => Ok(()),
-            Self::RocksDB(mode) => mode.commit(),
-        }
-    }
 }
 
 impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N>
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -208,6 +208,24 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
             Self::RocksDB(_) => Err(ProviderError::UnsupportedProvider),
         }
     }
+
+    /// Commits `RocksDB` writes if this is a `RocksDB` writer.
+    ///
+    /// For [`Self::Database`] and [`Self::StaticFile`], this is a no-op as they use
+    /// different commit patterns (MDBX transaction commit, static file sync).
+    ///
+    /// # Commit Order
+    ///
+    /// Call this AFTER the outer MDBX transaction commits successfully. This ensures
+    /// that if `RocksDB` commit fails, the primary data (MDBX) is still intact and
+    /// the `RocksDB` data (which is derived) can be rebuilt.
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -422,6 +431,28 @@ impl<'a> RocksDBBatch<'a> {
...
+    pub fn commit(self) -> ProviderResult<()> {
+        self.provider.0.db.write_opt(self.inner, &WriteOptions::default()).map_err(|e| {
+            ProviderError::Database(DatabaseError::Commit(DatabaseErrorInfo {
+                message: e.to_string().into(),
+                code: -1,
+            }))
+        })
+    }
+
+    /// Returns the number of operations in this batch.
```
fgimenez:
```suggestion
    /// Returns the number of write operations (puts + deletes) queued in this batch.
```

fgimenez (approved):
lgtm, minor doc nit

crates/storage/provider/src/either_writer.rs
```diff
@@ -208,15 +209,21 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
...
+    ///
+    /// Call this AFTER the outer MDBX transaction commits successfully. This ensures
+    /// that if `RocksDB` commit fails, the primary data (MDBX) is still intact and
+    /// the `RocksDB` data (which is derived) can be rebuilt.
     #[cfg(all(unix, feature = "rocksdb"))]
     pub fn commit(self) -> ProviderResult<()> {
         match self {
             Self::Database(_) | Self::StaticFile(_) => Ok(()),
-            Self::RocksDB(tx) => tx.commit(),
+            Self::RocksDB(mode) => mode.commit(),
```
joshieDo: unsure if we actually want/need `EitherWriter::commit`

crates/storage/provider/src/either_writer.rs
```diff
@@ -208,15 +209,21 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
...
+    /// Commits `RocksDB` writes if this is a `RocksDB` writer.
     ///
     /// For [`Self::Database`] and [`Self::StaticFile`], this is a no-op as they use
     /// different commit patterns (MDBX transaction commit, static file sync).
+    ///
+    /// # Commit Order
+    ///
+    /// Call this AFTER the outer MDBX transaction commits successfully. This ensures
+    /// that if `RocksDB` commit fails, the primary data (MDBX) is still intact and
+    /// the `RocksDB` data (which is derived) can be rebuilt.
```
joshieDo: any commit logic should ideally be in one place only. right now it's at:

https://github.com/paradigmxyz/reth/blob/948966781425b6ee6d9fec7e2a1078d58eacec04/crates/storage/provider/src/providers/database/provider.rs#L3181-L3197

joshieDo (approved):
i'd remove the `commit`, but apart from that lgtm

crates/storage/provider/src/either_writer.rs
yongkangc: yeah makes sense here, to me seems like a shallow abstraction

yongkangc: will remove it first YAGNI, we can always add that back if we really need to

crates/storage/provider/src/either_writer.rs
yongkangc: i see, thanks for the note @joshieDo - removing that YAGNI

initially was thinking that either writer is a unified abstraction, but its clearer to just call rocksdb.commit() than eitherwrite.commit etc

PR #20382, opened 2025-12-15
fix: sync BodiesDownloader docs with current implementation

SashaMalysehko:
The network crate docs were still describing the old BodiesDownloader API (retries/batch_size/concurrency and a fetch_bodies helper). Updated the section to show the current generic BodiesDownloader shape and the internal request flow via BodiesRequestFuture::submit_request, including the actual get_block_bodies_with_priority call, so the documentation matches the existing code.

### DIFF
--- a/docs/crates/network.md
+++ b/docs/crates/network.md
@@ -382,36 +382,45 @@ fn get_or_init_fut(&mut self) -> HeadersRequestFuture {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
these docs are pretty much useless

but this is at least more accurate

mattsse (approved):

PR #20383, opened 2025-12-15
feat(engine): Use BAL in state root validation

### LINKED ISSUE #20272: BAL: Payload Validator: State Root Integration
### Describe the feature

The following changes should be made to optimize state root calculation when BAL is present.
 
* A new variant to [MultiProofMessage](https://github.com/paradigmxyz/reth/blob/b7d881510422ac03340a63e8c08239950e8226b0/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L81) should be added, `MultiProofMessage::BAL(BlockAccessList)`. When received the `MultiProofTask` should generate a single state update using the BAL, and process it using `on_state_update`. The MultiProofTask should not expect further messages when it receves a `MultiProofMessage::BAL`.

* When a BAL is present in the block being validated, the payload validator should spawn the prewarm tasks such that they do not send proof targets to the MultiProofTask.

* When a BAL is present in the block then the payload validator should immediately send a `MultiProofMessage::BAL` to the MultiProofTask once it has been spawned.

### Additional context

_No response_

mediocregopher:
Closes #20272 

This PR is the first step to integrate the BAL into the payload validator. When BAL is present, it will be sent to the MultiProofTask as-is. The MultiProofTask will convert it to a HashedPostState, and treat that as a single giant state updates to generate proof targets with, chunking the targets as it normally would.

In this case the prewarm tasks are configured to not send proof targets to the MultiProofTask, as that's no longer necessary.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/bal.rs
+++ b/crates/engine/tree/src/tree/payload_processor/bal.rs
@@ -1,0 +1,318 @@
+//! BAL (Block Access List, EIP-7928) related functionality.
+
+use alloy_consensus::constants::KECCAK_EMPTY;
+use alloy_eip7928::BlockAccessList;
+use alloy_primitives::{keccak256, U256};
+use reth_primitives_traits::Account;
+use reth_provider::{AccountReader, ProviderError};
+use reth_trie::{HashedPostState, HashedStorage};
+
+/// Converts a Block Access List into a [`HashedPostState`] by extracting the final state
+/// of modified accounts and storage slots.
+pub fn bal_to_hashed_post_state<P>(
+    bal: &BlockAccessList,
+    provider: &P,
+) -> Result<HashedPostState, ProviderError>
+where
+    P: AccountReader,
+{
+    let mut hashed_state = HashedPostState::with_capacity(bal.len());
+
+    for account_changes in bal {
+        let address = account_changes.address;
+        let hashed_address = keccak256(address);
+
+        // Get the latest balance (last balance change if any)
+        let balance = account_changes.balance_changes.last().map(|change| change.post_balance);
+
+        // Get the latest nonce (last nonce change if any)
+        let nonce = account_changes.nonce_changes.last().map(|change| change.new_nonce);
+
+        // Get the latest code (last code change if any)
+        let code_hash = if let Some(code_change) = account_changes.code_changes.last() {
+            if code_change.new_code.is_empty() {
+                Some(Some(KECCAK_EMPTY))
+            } else {
+                Some(Some(keccak256(&code_change.new_code)))
+            }
+        } else {
+            None
+        };
+
+        // Only fetch account from provider if we're missing any field
+        let existing_account = if balance.is_none() || nonce.is_none() || code_hash.is_none() {
+            provider.basic_account(&address)?
+        } else {
+            None
+        };
+
+        // Build the final account state
+        let account = Account {
+            balance: balance.unwrap_or_else(|| {
+                existing_account.as_ref().map(|acc| acc.balance).unwrap_or(U256::ZERO)
+            }),
+            nonce: nonce
+                .unwrap_or_else(|| existing_account.as_ref().map(|acc| acc.nonce).unwrap_or(0)),
+            bytecode_hash: code_hash.unwrap_or_else(|| {
+                existing_account.as_ref().and_then(|acc| acc.bytecode_hash).or(Some(KECCAK_EMPTY))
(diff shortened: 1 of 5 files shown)


crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -0,0 +1,318 @@
...
+            if code_change.new_code.is_empty() {
+                Some(Some(KECCAK_EMPTY))
+            } else {
+                Some(Some(keccak256(&code_change.new_code)))
+            }
+        } else {
+            None
+        };
+
+        // Only fetch account from provider if we're missing any field
```
mediocregopher, line 42: @rakita I wanted to double check with you that I'm not missing something here. The BAL only contains individual changes to the account tuple fields, so if any field of an account is not changed we need to fetch that field from the db in order to calculate SR.

crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -0,0 +1,318 @@
...
+        // Get the latest balance (last balance change if any)
+        let balance = account_changes.balance_changes.last().map(|change| change.post_balance);
+
+        // Get the latest nonce (last nonce change if any)
+        let nonce = account_changes.nonce_changes.last().map(|change| change.new_nonce);
+
+        // Get the latest code (last code change if any)
+        let code_hash = if let Some(code_change) = account_changes.code_changes.last() {
+            if code_change.new_code.is_empty() {
+                Some(Some(KECCAK_EMPTY))
```
mediocregopher, line 34: @mattsse should this just be `Some(None)`? I think they're equivalent but wasn't 100% sure

crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -0,0 +1,318 @@
...
+        let account = Account {
+            balance: balance.unwrap_or_else(|| {
+                existing_account.as_ref().map(|acc| acc.balance).unwrap_or(U256::ZERO)
+            }),
+            nonce: nonce
+                .unwrap_or_else(|| existing_account.as_ref().map(|acc| acc.nonce).unwrap_or(0)),
+            bytecode_hash: code_hash.unwrap_or_else(|| {
+                existing_account.as_ref().and_then(|acc| acc.bytecode_hash).or(Some(KECCAK_EMPTY))
+            }),
+        };
```
shekhirin, line 59: are these `unwrap_or` possible here? It means an account wasn't in the BAL, and it also doesn't exist in our database

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1146,6 +1168,58 @@ impl MultiProofTask {
...
+                        debug!(
+                            target: "engine::tree::payload_processor::multiproof",
+                            accounts = hashed_state.accounts.len(),
+                            storages = hashed_state.storages.len(),
+                            "Processing BAL state update"
+                        );
+
+                        // Use Transaction(0) as source for BAL-derived state updates
+                        batch_metrics.state_update_proofs_requested += self.on_hashed_state_update(
+                            StateChangeSource::Transaction(0),
```
shekhirin: would be nice to identify BAL state updates separately, as IIUC this `source` is only used for logging https://github.com/paradigmxyz/reth/blob/3c4740dab648b7f86ca814cce02a36bd5bfbeade/crates/engine/tree/src/tree/payload_processor/multiproof.rs#L472-L480

shekhirin (approved):

crates/engine/tree/src/tree/payload_processor/bal.rs
mediocregopher: If it's a brand new account then the changes would appear in the BAL, and it wouldn't be in the DB, so yes I think these are possible

crates/engine/tree/src/tree/payload_processor/bal.rs
rakita: True
and if the account is present and there are no changes this means that account is only read (This does not mean that there are not storage writes)

yongkangc (approved):
lgtm, clean

crates/engine/tree/src/tree/payload_processor/multiproof.rs
mediocregopher: I added a new `Source` enum to differentiate a `StateChangeSource` (from alloy_evm) from a BAL update.

PR #20385, opened 2025-12-15
perf: recover transactions in parallel during network import

klkvr:


crates/net/network/src/transactions/mod.rs
```diff
@@ -1368,51 +1369,49 @@ where
...
+                return false;
+            }
+            true
+        });
+
+        let txs_len = transactions.len();
+
+        let new_txs = transactions
+            .into_par_iter()
+            .filter_map(|tx| match tx.try_into_recovered() {
```
mattsse, line 1395: makes sense, we also do this when recovering txs in to recovered blocks, so this seems appropriate

mattsse (approved):

DaniPopes (approved):

PR #20394, opened 2025-12-15
fix: tx_count debug log to report actual broadcast size

zeevick10:
Log the real number of pooled transaction hashes (`tx_count`) instead of a boolean emptiness flag. Keeps broadcast debug output accurate for troubleshooting network transaction propagation.


mattsse (commented):
we can keep this as is

PR #20395, opened 2025-12-15
feat(cli): add rocksdb table flags

### LINKED ISSUE #20393: Add CLI flags to enable RocksDB storage
Add CLI flags to enable storing `TransactionHashNumbers` and `StoragesHistory` tables in RocksDB instead of MDBX.

- In `crates/node/core/src/args/static_files.rs`:
  - Add `tx_hash_in_rocksdb: bool` field with `#[arg(long = "storage.tx-hash-in-rocksdb")]`
  - Add `storages_history_in_rocksdb: bool` field with `#[arg(long = "storage.storages-history-in-rocksdb")]`
  - Update `to_settings()` to include the new flags in `StorageSettings`

- In `crates/storage/provider/src/lib.rs` (or `StorageSettings`):
  - Add `transaction_hash_numbers_in_rocksdb: bool` field
  - Add `storages_history_in_rocksdb: bool` field
  - Add corresponding `with_*` builder methods

CLI to enable the feature flag. For storage setting, we save them to mdbx during genesis. 

Basically replicate what is done for `transaction_senders_in_static_files`
crates/cli/commands/src/db/settings.rs

Rjected:
Fixes https://github.com/paradigmxyz/reth/issues/20393

### DIFF
--- a/crates/node/core/src/args/static_files.rs
+++ b/crates/node/core/src/args/static_files.rs
@@ -42,6 +42,27 @@ pub struct StaticFilesArgs {
     /// the node has been initialized, changing this flag requires re-syncing from scratch.
     #[arg(long = "static-files.transaction-senders")]
     pub transaction_senders: bool,
+
+    /// Store `TransactionHashNumbers` table in `RocksDB` instead of MDBX.
+    ///
+    /// Note: This setting can only be configured at genesis initialization. Once
+    /// the node has been initialized, changing this flag requires re-syncing from scratch.
+    #[arg(long = "storage.tx-hash-numbers-in-rocksdb")]
+    pub tx_hash_numbers_in_rocksdb: bool,
+
+    /// Store `StoragesHistory` table in `RocksDB` instead of MDBX.
+    ///
+    /// Note: This setting can only be configured at genesis initialization. Once
+    /// the node has been initialized, changing this flag requires re-syncing from scratch.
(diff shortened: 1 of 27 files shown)


yongkangc (approved):
lgtm!

crates/node/core/src/args/static_files.rs
```diff
@@ -42,6 +42,27 @@ pub struct StaticFilesArgs {
     /// the node has been initialized, changing this flag requires re-syncing from scratch.
     #[arg(long = "static-files.transaction-senders")]
     pub transaction_senders: bool,
+
+    /// Store `TransactionHashNumbers` table in `RocksDB` instead of MDBX.
+    ///
+    /// Note: This setting can only be configured at genesis initialization. Once
+    /// the node has been initialized, changing this flag requires re-syncing from scratch.
+    #[arg(long = "storage.tx-hash-numbers-in-rocksdb")]
```
joshieDo, line 50: feels a bit odd to be here, should we introduce `RocksArgs` instead so we can have the following:
--rocksdb.tx-hash-numbers

fgimenez (approved):
changes lgtm, but maybe we should wait to merge until the rocksdb backend is fully functional? otherwise if someone uses the cli flags things won't work as expected aiui

crates/node/core/src/args/static_files.rs
Rjected: actually yeah

PR #20397, opened 2025-12-15
docs: fix typo and improve FullBlockBody documentation

hawkadrian:
## What

Fixed typo in `FullBlockBody` docstring and expanded documentation with explanation of associated type bounds syntax and usage example.

## Why

The original documentation had a typo ("transaction" instead of "block body") and lacked explanation of the `BlockBody<Transaction: FullSignedTx>` syntax, making it unclear for developers how to use the trait. The improved docs clarify the trait's constraints and provide a concrete usage example from `NodePrimitives`.


mattsse (commented):
we can keep this as is

PR #20398, opened 2025-12-15
perf: use RwLock for transaction pool listeners

mattsse:
Transaction notifications are on the hot path (called for every pending transaction) but previously used `Mutex<Vec<Listener>>` which blocked all concurrent notifications. This changes `pending_transaction_listener` and `transaction_listener` to use `RwLock` instead.

The hot path now uses shared read locks allowing multiple threads to notify concurrently. Write locks are only acquired when dead listeners are detected (channel closed) for cleanup. Cleanup also happens when adding new listeners.

With typically 1-2 listeners (network + optional RPC), this eliminates lock contention on transaction notifications with zero change to semantics.

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -145,9 +145,9 @@ where
     /// Manages listeners for transaction state change events.
     event_listener: RwLock<PoolEventBroadcast<T::Transaction>>,
     /// Listeners for new _full_ pending transactions.
-    pending_transaction_listener: Mutex<Vec<PendingTransactionHashListener>>,
+    pending_transaction_listener: RwLock<Vec<PendingTransactionHashListener>>,
     /// Listeners for new transactions added to the pool.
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -654,16 +676,26 @@ where
...
-                // only emit this hash to listeners that are only allowed to receive propagate only
-                // transactions, such as network
-                return !listener.sender.is_closed()
+        let mut needs_cleanup = false;
+
+        {
+            let listeners = self.transaction_listener.read();
+            for listener in listeners.iter() {
+                if listener.kind.is_propagate_only() && !event.transaction.propagate {
+                    // Skip non-propagate transactions for propagate-only listeners
```
klkvr, line 688:
```suggestion
                    if listener.sender.is_closed() {
                        needs_cleanup = true;
                    }
                    // Skip non-propagate transactions for propagate-only listeners
```

klkvr (approved):

PR #20399, opened 2025-12-15
docs: add comment explaining unused function warning suppression

CreeptoGengar:
Adds a comment to clarify why the conditional compilation block references verify_and_compute_sender_unchecked_k256 without calling it.
This prevents compiler warnings when both k256 and secp256k1 features are enabled, as the k256 function is compiled but not used when secp256k1 has priority.


mattsse (commented):
we can keep as is

PR #20404, opened 2025-12-15
chore: remove redundant .as_str() calls after to_string()

0xcharry:
Remove redundant `.as_str()` conversions after `.to_string()`


mattsse (approved):
fine, even tho these are just tests

PR #20405, opened 2025-12-15
perf: defer transaction pool notifications until after lock release

mattsse:
Transaction pool was holding the write lock during all notification and blob storage operations, blocking critical state updates and other pool operations.

## Changes

`add_transaction` now collects metadata instead of performing notifications immediately. All post-insertion work (blob storage, notifications) happens in `on_added_transaction` after releasing the pool lock.

**Before:** Lock held during insertions + blob ops + notifications (~100ms-1s for 100 tx batch)
**After:** Lock held only during insertions (~1-10ms for 100 tx batch)

This still has one heap alloc for AddedTransactionMeta for the single tx case which we could specialize as well


klkvr (approved):
nice

PR #20406, opened 2025-12-15
perf: remove unnecessary channels from parallel trie operations

mattsse:
The channel-based result gathering in `reveal_nodes` and `update_subtrie_hashes` was unnecessary since rayon's `.collect()` achieves the same result more simply. Replaced `.for_each_init()` with channel pattern with direct `.collect()` in both methods, removing the `sync::mpsc` import.

unsure if the intention was to process them as fast as possible here, but this for_each with channel basically just acts as an expensive buffer here


shekhirin (approved):

yongkangc (approved):

yongkangc (commented):
on a side note, this looks much simpler

PR #20411, opened 2025-12-16
feat(net/p2p): support fixed external addresses with DNS resolution

theochap:
## Description

This PR adds support to specify external p2p discv4/5 addresses via DNS. Those addresses are resolved to fixed IP addresses to broadcast to peers in the p2p discovery network. This is especially useful in docker bridge networks where nodes are usually queried by DNS rather than IP addresses.


mattsse (approved):
this seems fine, ty!

PR #20412, opened 2025-12-16
feat(storage): Use EitherReader/EitherWriter in DatabaseProvider and HistoricalStateProvider

### LINKED ISSUE #20388: Use EitherReader/EitherWriter in DatabaseProvider and HistoricalStateProvider

yongkangc:
Closes #20388

Integrate RocksDB support into `DatabaseProvider` and `HistoricalStateProvider` using the `EitherReader`/`EitherWriter` abstraction.

**Changes**

**EitherReader/EitherWriter:**
- Added `read_account_history_shards()`, `read_storage_history_shards()` for history lookups
- `EitherWriter::RocksDB` now borrows batch from provider (single batch per provider model)

**DatabaseProvider:**
- Added RocksDB-specific `append_*_history_index` and `unwind_*_history_indices` methods with "last write wins" deduplication
- Updated `transaction_id()`, `insert_block()`, `remove_blocks_above()` to use EitherReader/EitherWriter
- Centralized RocksDB batch commits in `DBProvider::commit()` with correct ordering:
  - Forward: `static_files → rocks → mdbx`
  - Unwind: `mdbx → rocks → static_files`

**HistoricalStateProvider:**
- Added `history_info_from_shards()`, `account_history_lookup_rocksdb()`, `storage_history_lookup_rocksdb()`
- Conditional `AccountReader`/`StateProvider` impls for RocksDB builds


crates/storage/provider/src/either_writer.rs
```diff
@@ -213,6 +213,18 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
             Self::RocksDB(_) => Err(ProviderError::UnsupportedProvider),
         }
     }
+
+    /// Consumes the writer and returns the underlying `RocksDB` batch if present.
+    ///
+    /// Returns `Some(batch)` for [`Self::RocksDB`], `None` for other variants.
+    /// The caller is responsible for committing the batch.
+    #[cfg(all(unix, feature = "rocksdb"))]
+    pub fn into_rocksdb_batch(self) -> Option<RocksDBBatch<'a>> {
```
yongkangc: the reason why i have this is so that we can centralise the commit in provider 
https://github.com/paradigmxyz/reth/blob/98b787fe4d5a586ab5eb7149d93298a23a909c12/crates/storage/provider/src/providers/database/provider.rs#L3472

dont feel 100% about this, but not sure if theres a better way if we want to centrallise commit

fgimenez (approved):
lgtm

Rjected (commented):
I think it would be nice if we could use `EitherReader` for the historical state provider impls somehow, because I would rather not end up with duplicate gated impls for all of the historical state provider methods.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -524,21 +499,72 @@ impl LowestAvailableBlocks {
...
+
+    // Check if we're before the first-ever change
+    if rank == 0 && found_block != Some(block_number) && !has_previous_shard {
+        if let (Some(_), Some(bn)) = (lowest_available, found_block) {
+            return HistoryInfo::InChangeset(bn);
+        }
+        return HistoryInfo::NotYetWritten;
+    }
+
+    found_block.map(HistoryInfo::InChangeset).unwrap_or(HistoryInfo::InPlainState)
```
Rjected: Can we add back the comments that we had in `history_info`? so that we can be sure that it's actually equivalent?

crates/storage/provider/src/either_writer.rs
```diff
@@ -577,6 +619,131 @@ where
...
+                        &chunk,
+                        block_number,
+                        has_prev,
+                        lowest_available_block_number,
+                    ))
+                } else if lowest_available_block_number.is_some() {
+                    Ok(HistoryInfo::MaybeInPlainState)
+                } else {
+                    Ok(HistoryInfo::NotYetWritten)
+                }
```
Rjected, line 758: would also like to move the comments from the old fn here as well

Rjected (commented):
some comments, still going through the append / unwind logic

crates/storage/provider/src/either_writer.rs
yongkangc: added back!

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -634,7 +634,11 @@ impl<Tx, Err, R: Send + Sync + 'static> PayloadHandle<Tx, Err, R> {
 
         move |source: StateChangeSource, state: &EvmState| {
             if let Some(sender) = &to_multi_proof {
-                let _ = sender.send(MultiProofMessage::StateUpdate(source.into(), state.clone()));
+                // Convert EvmState to HashedPostState BEFORE sending to avoid issues with
+                // EvmState::extend overwriting is_changed flags during batching.
+                // This ensures each transaction's state changes are captured correctly.
+                let hashed_state = evm_state_to_hashed_post_state(state.clone());
+                let _ = sender.send(MultiProofMessage::StateUpdate(source.into(), hashed_state));
```
Rjected: not relevant for this PR right?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -167,10 +167,30 @@ pub struct DatabaseProvider<TX, N: NodeTypes> {
     storage_settings: Arc<RwLock<StorageSettings>>,
     /// `RocksDB` provider
     rocksdb_provider: RocksDBProvider,
+    /// Single `RocksDB` batch for this provider's lifetime.
+    /// All `RocksDB` writes accumulate here and commit together at provider commit time.
+    #[cfg(all(unix, feature = "rocksdb"))]
+    pending_rocks_batch: std::sync::Mutex<crate::providers::rocksdb::RocksDBBatch>,
```
Rjected: we already have this here https://github.com/paradigmxyz/reth/blob/30162c535e0c3ef9997c0745c90888e1905fc1c3/crates/storage/provider/src/providers/database/provider.rs#L169-L171

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -167,10 +167,30 @@ pub struct DatabaseProvider<TX, N: NodeTypes> {
...
+    /// In-memory overlay for history index last-shards written into the shared `RocksDB` batch.
+    ///
+    /// `RocksDBBatch` is write-only (no read-your-writes). When history indices are appended
+    /// multiple times before `commit`, subsequent appends must not rebuild shards from the
+    /// persisted `RocksDB` state alone, or earlier batch updates would be lost ("last write
+    /// wins").
+    #[cfg(all(unix, feature = "rocksdb"))]
+    pending_history_index_last_shards: std::sync::Mutex<PendingHistoryIndexLastShards>,
```
Rjected: the comment doesn't really explain why this is required, for history indexing specifically, are we reading whenever we append history indices? ie, the previous method for appending history indices was reading using the write tx?

Rjected (commented):
the impls look the same, have some more comments mostly about undersandability

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: sorry about this, this kinda got merged with another pr while resolving merge conflicts

crates/storage/provider/src/providers/database/provider.rs
yongkangc: ah i see, that just got merged in not too long ago. thanks for the note @Rjected , will rebase

crates/storage/provider/src/providers/database/provider.rs
yongkangc: clarified the comment much better here!

yongkangc: [5f06662](https://github.com/paradigmxyz/reth/pull/20412/commits/5f06662bf9304d5a8850260501f23cfaca719b70)

crates/storage/provider/src/providers/database/provider.rs
yongkangc: merged main in! addressed

PR #20414, opened 2025-12-16
refactor(db): Replace raw transaction usage in TransactionLookup with provider trait methods

### LINKED ISSUE #20281: Replace raw transaction usage in `TransactionLookup` with provider trait methods
### Describe the feature

Right now we use raw transactions in the `TransactionLookup` stage, for the `TransactionHashNumbers` table

https://github.com/paradigmxyz/reth/blob/179e1644ec72b9afb05a46367f2b55baf7de8cce/crates/stages/stages/src/stages/tx_lookup.rs#L153-L157

We should come up with provider trait methods that make sense for what we are doing with the raw txs in the stage. For example, we would want one for counting entries in the `TransactionHashNumbers` table:
https://github.com/paradigmxyz/reth/blob/179e1644ec72b9afb05a46367f2b55baf7de8cce/crates/stages/stages/src/stages/tx_lookup.rs#L154

and another for the write cursor usage - that one might be more involved because the access pattern is:
```rust
let cursor = tx.cursor_write();
for elem in list {
    if maybe {
        cursor.append(elem);
    else {
        cursor.insert(elem);
    }
}
```

### Additional context

_No response_

yongkangc:
Closes #20281

Introduce a new `TransactionHashNumbersWriter` trait that abstracts bulk writes to the `TransactionHashNumbers` table. The trait method accepts raw bytes from the ETL collector and handles the append-only optimization internally. 

Stage logic is now storage-agnostic. This allows us to handles RocksDB batch writes without stage changes

**Changes**

- Add `TransactionHashNumbersWriter` trait to `storage-api` with `insert_transaction_hash_numbers_raw` method
- Implement trait for `DatabaseProvider` with both MDBX (append/insert optimization) and RocksDB (batch write) paths
- Implement no-op version for `NoopProvider`
- Refactor `TransactionLookup` stage to use the new trait instead of direct cursor access

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8247,6 +8247,7 @@ dependencies = [
  "metrics",
  "metrics-util",
  "mini-moka",
+ "moka",
  "parking_lot",
  "proptest",
  "rand 0.8.5",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -587,6 +587,7 @@ url = { version = "2.3", default-features = false }
 zstd = "0.13"
 byteorder = "1"
 mini-moka = "0.10"
+moka = "0.12"
 tar-no-std = { version = "0.3.2", default-features = false }
 miniz_oxide = { version = "0.8.4", default-features = false }
 chrono = "0.4.41"
--- a/crates/engine/tree/Cargo.toml
+++ b/crates/engine/tree/Cargo.toml
@@ -53,6 +53,7 @@ futures.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "sync", "macros"] }
 mini-moka = { workspace = true, features = ["sync"] }
+moka = { workspace = true, features = ["sync"] }
 smallvec.workspace = true
 
 # metrics
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -393,7 +393,7 @@ where
             metrics,
             terminate_execution,
             precompile_cache_disabled,
-            mut precompile_cache_map,
+            precompile_cache_map,
         } = self;
 
         let mut state_provider = match provider.build() {
--- a/crates/engine/tree/src/tree/precompile_cache.rs
+++ b/crates/engine/tree/src/tree/precompile_cache.rs
@@ -1,50 +1,56 @@
 //! Contains a precompile cache backed by `schnellru::LruMap` (LRU by length).
 
 use alloy_primitives::Bytes;
-use parking_lot::Mutex;
+use dashmap::DashMap;
 use reth_evm::precompiles::{DynPrecompile, Precompile, PrecompileInput};
 use revm::precompile::{PrecompileId, PrecompileOutput, PrecompileResult};
 use revm_primitives::Address;
-use schnellru::LruMap;
-use std::{
-    collections::HashMap,
-    hash::{Hash, Hasher},
-    sync::Arc,
-};
+use std::{hash::Hash, sync::Arc};
 
 /// Default max cache size for [`PrecompileCache`]
 const MAX_CACHE_SIZE: u32 = 10_000;
 
 /// Stores caches for each precompile.
 #[derive(Debug, Clone, Default)]
-pub struct PrecompileCacheMap<S>(HashMap<Address, PrecompileCache<S>>)
+pub struct PrecompileCacheMap<S>(Arc<DashMap<Address, PrecompileCache<S>>>)
 where
-    S: Eq + Hash + std::fmt::Debug + Send + Sync + Clone;
+    S: Eq + Hash + std::fmt::Debug + Send + Sync + Clone + 'static;
 
 impl<S> PrecompileCacheMap<S>
 where
     S: Eq + Hash + std::fmt::Debug + Send + Sync + Clone + 'static,
 {
-    pub(crate) fn cache_for_address(&mut self, address: Address) -> PrecompileCache<S> {
+    pub(crate) fn cache_for_address(&self, address: Address) -> PrecompileCache<S> {
+        // Try just using `.get` first to avoid acquiring a write lock.
+        if let Some(cache) = self.0.get(&address) {
+            return cache.clone();
+        }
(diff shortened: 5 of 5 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3121,6 +3121,56 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> StatsReader for DatabaseProvid
...
+    for DatabaseProvider<TX, N>
+{
+    fn insert_transaction_hash_numbers_raw<I>(&self, hash_to_number_iter: I) -> ProviderResult<bool>
+    where
+        I: Iterator<Item = std::io::Result<(Vec<u8>, Vec<u8>)>>,
+    {
+        // Check if RocksDB is configured for this table
+        #[cfg(all(unix, feature = "rocksdb"))]
+        if self.cached_storage_settings().transaction_hash_numbers_in_rocksdb {
+            use reth_db_api::table::Decompress;
```
joshieDo: nit can we move it top of the file

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3121,6 +3121,56 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> StatsReader for DatabaseProvid
...
+            // RocksDB path: use batch writes
+            // Note: RocksDB doesn't have an append-only optimization like MDBX,
+            // so we always return false for append_only mode
+            self.rocksdb_provider().write_batch(|batch| {
+                for entry in hash_to_number_iter {
+                    let (hash_bytes, number_bytes) = entry.map_err(ProviderError::other)?;
+
+                    // Convert raw bytes to typed values
+                    let hash = TxHash::try_from(hash_bytes.as_slice())
+                        .map_err(|_| ProviderError::InvalidStorageOutput)?;
```
Rjected: nit: can we instead use `ProviderError::other`? the `InvalidStorageOutput` error isn't very descriptive

crates/storage/storage-api/src/transactions.rs
```diff
@@ -109,3 +109,39 @@ pub trait TransactionsProviderExt: BlockReader {
...
+/// This trait abstracts the bulk write operation for the `TransactionHashNumbers` table,
+/// handling the append-only optimization internally.
+#[auto_impl::auto_impl(&, Arc, Box)]
+pub trait TransactionHashNumbersWriter: Send + Sync {
+    /// Insert transaction hash to number mappings into the database.
+    ///
+    /// This method handles the append-only optimization internally:
+    /// - If the `TransactionHashNumbers` table is empty, uses `append` (O(1))
+    /// - Otherwise, uses `insert` (O(log n))
+    ///
```
Rjected: I think we can remove this comment because it's only relevant to the mdbx part of the implementation

crates/storage/storage-api/src/transactions.rs
```diff
@@ -109,3 +109,39 @@ pub trait TransactionsProviderExt: BlockReader {
         tx_range: Range<TxNumber>,
     ) -> ProviderResult<Vec<(TxHash, TxNumber)>>;
 }
+
+/// Writer trait for transaction hash to number mappings.
+///
+/// This trait abstracts the bulk write operation for the `TransactionHashNumbers` table,
+/// handling the append-only optimization internally.
```
Rjected: would also remove this mention of append-only optimization, I think it's an implementation detail that doesn't need to be in trait docs

crates/storage/provider/src/providers/database/provider.rs
yongkangc: makes more sense, good note

crates/storage/storage-api/src/transactions.rs
yongkangc: good point, addressed [79b4a00](https://github.com/paradigmxyz/reth/pull/20414/commits/79b4a007f63800b7b836eca0d915751ec61dce81)

crates/storage/storage-api/src/transactions.rs
yongkangc: makes sense! addressed [79b4a00](https://github.com/paradigmxyz/reth/pull/20414/commits/79b4a007f63800b7b836eca0d915751ec61dce81)

crates/storage/provider/src/providers/database/provider.rs
yongkangc: makes sense! addressed [79b4a00](https://github.com/paradigmxyz/reth/pull/20414/commits/79b4a007f63800b7b836eca0d915751ec61dce81)

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3121,6 +3121,51 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> StatsReader for DatabaseProvid
...
+                    let hash =
+                        TxHash::try_from(hash_bytes.as_slice()).map_err(ProviderError::other)?;
+                    let number: TxNumber = Decompress::decompress(&number_bytes)?;
+
+                    batch.put::<tables::TransactionHashNumbers>(hash, &number)?;
+                }
+                Ok(())
+            })?;
+
+            return Ok(false);
```
fgimenez: we are always returning `Ok(false)` here, but in the trait fn doc we say `Returns `Ok(true)` if the table was empty before insertion, `Ok(false)` otherwise.` https://github.com/paradigmxyz/reth/pull/20414/files#diff-92d898a32e1356e566dd6266a1c516f269575be896a5e4d1b22e0106ca30f5cdR132 

this return value seems to be used only for logging, maybe we should update the doc?

fgimenez (approved):
lgtm, smol nit

crates/storage/provider/src/providers/database/provider.rs
yongkangc: ahh good catch, thanks!

yongkangc: addressed in [d170fe8](https://github.com/paradigmxyz/reth/pull/20414/commits/d170fe845f1b98d27de977aa164a7c3990e75cec)

Rjected (approved):
lgtm

PR #20420, opened 2025-12-16
chore: add engine terminate

mattsse:


crates/engine/tree/src/tree/mod.rs
```diff
@@ -1302,22 +1314,7 @@ where
...
-                        // empty vec of blocks
-                        warn!(target: "engine::tree", "Persistence task completed but did not persist any blocks");
-                        return Ok(())
-                    };
-
-                    debug!(target: "engine::tree", ?last_persisted_block_hash, ?last_persisted_block_number, elapsed=?start_time.elapsed(), "Finished persisting, calling finish");
-                    self.persistence_state
-                        .finish(last_persisted_block_hash, last_persisted_block_number);
-                    self.on_new_persisted_block()?;
+                    self.on_persistence_complete(last_persisted_hash_num, start_time)?;
```
mattsse, line 1310: can we do this extraction first?

would make it easier to review

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1315,14 +1327,49 @@ where
...
     }
 
+    /// Finishes termination by persisting all remaining blocks and signaling completion.
+    ///
+    /// This blocks until all persistence is complete. Always signals completion,
+    /// even if an error occurs.
+    fn finish_termination(
+        &mut self,
+        pending_termination: oneshot::Sender<()>,
+    ) -> Result<(), AdvancePersistenceError> {
```
mattsse, line 1339: can we also add a trace here?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -277,6 +277,8 @@ where
     engine_kind: EngineApiKind,
     /// The EVM configuration.
     evm_config: C,
+    /// Termination signal sender, set when shutdown is requested.
+    pending_termination: Option<oneshot::Sender<()>>,
```
mattsse: hmm, why do we need to tmp store this?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1361,6 +1408,10 @@ where
                 FromOrchestrator::BackfillSyncFinished(ctrl) => {
                     self.on_backfill_sync_finished(ctrl)?;
                 }
+                FromOrchestrator::Terminate { tx } => {
+                    debug!(target: "engine::tree", "received terminate request");
+                    self.pending_termination = Some(tx);
```
mattsse: can we not drive the termination to completion here? this seems easier than yielding control back to the main loop for this?

we can make this function the return a bool if that helps to terminate the task

crates/node/builder/src/launch/engine.rs
```diff
@@ -295,6 +304,14 @@ impl EngineNodeLauncher {
             // advance the chain and await payloads built locally to add into the engine api tree handler to prevent re-execution if that block is received as payload from the CL
             loop {
                 tokio::select! {
+                    shutdown_req = &mut shutdown_rx => {
+                        if let Ok(req) = shutdown_req {
+                            debug!(target: "reth::cli", "received engine shutdown request");
+                            engine_service.orchestrator_mut().handler_mut().handler_mut().on_event(
+                                FromOrchestrator::Terminate { tx: req.done_tx }.into()
```
mattsse, line 311: ah I see this simply forwards the channel

crates/node/builder/src/rpc.rs
```diff
@@ -319,6 +321,51 @@ where
...
+    }
+
+    /// Requests a graceful engine shutdown.
+    ///
+    /// All remaining in-memory blocks will be persisted before the engine terminates.
+    ///
+    /// Returns a future that resolves when the shutdown is complete, or `None`
+    /// if shutdown was already requested.
+    pub async fn shutdown(&self) -> Option<oneshot::Receiver<()>> {
+        let mut guard = self.tx.lock().await;
```
mattsse: I think we can make this a regular mutex instead

crates/node/builder/src/rpc.rs
```diff
@@ -319,6 +321,51 @@ where
...
 
+/// Handle to trigger graceful engine shutdown.
+///
+/// This handle can be used to request a graceful shutdown of the engine,
+/// which will persist all remaining in-memory blocks before terminating.
+#[derive(Clone, Debug)]
+pub struct EngineShutdown {
+    /// Channel to send shutdown signal.
+    tx: Arc<Mutex<Option<oneshot::Sender<EngineShutdownRequest>>>>,
+}
```
mattsse: this type feels a bit random here

can me maybe move this to end of file?

crates/node/builder/src/rpc.rs
```diff
@@ -319,6 +321,51 @@ where
...
+        let (tx, rx) = oneshot::channel();
+        (Self { tx: Arc::new(Mutex::new(Some(tx))) }, rx)
+    }
+
+    /// Requests a graceful engine shutdown.
+    ///
+    /// All remaining in-memory blocks will be persisted before the engine terminates.
+    ///
+    /// Returns a future that resolves when the shutdown is complete, or `None`
+    /// if shutdown was already requested.
```
mattsse: should mention that this returns a none if a shutdown was already triggered

mattsse (commented):
some suggestions

mattsse (commented):
lgtm

yongkangc (approved):

joshieDo (approved):

crates/node/builder/src/launch/engine.rs
```diff
@@ -260,8 +261,16 @@ impl EngineNodeLauncher {
...
 
-        let RpcHandle { rpc_server_handles, rpc_registry, engine_events, beacon_engine_handle } =
-            add_ons.launch_add_ons(add_ons_ctx).await?;
+        let RpcHandle {
+            rpc_server_handles,
+            rpc_registry,
+            engine_events,
+            beacon_engine_handle,
+            engine_shutdown: _,
+        } = add_ons.launch_add_ons(add_ons_ctx).await?;
```
DaniPopes, line 270: what is the point of this?

PR #20423, opened 2025-12-16
chore: era decompression bounds

NaCl-Ezpz:
Bounds Snappy decompression output size for ERA consensus payloads to prevent unbounded memory allocation on malformed inputs. No behavior change for valid ERA files.


mattsse (commented):
I think we can keep this as is

mattsse (approved):
these bounds are pretty relaxed, but imo fine to have some checks for this in place at least

PR #20425, opened 2025-12-16
fix(optimism): delegate OpPooledTransaction::with_conditional to trait default

lisenokdonbassenok:
The problem was in duplication of with_conditional logic: The MaybeConditionalTransaction tray already has a default implementation of with_conditional, and OpPooledTransaction defines its own method of the same name with the same semantics. 
Replaced the inherent OpPooledTransaction::with_conditional implementation with a call to MaybeConditionalTransaction::with_conditional to avoid duplicated logic.


mattsse (commented):
we can keep this as is

PR #20428, opened 2025-12-16
feat(stages): use EitherWriter for TransactionLookupStage RocksDB writes

### LINKED ISSUE #20389: Modify TransactionLookupStage to use `EitherWriter` for RocksDB writes
Modify `TransactionLookupStage` to write transaction hash mappings to RocksDB when the feature is enabled.

In `crates/stages/stages/src/stages/tx_lookup.rs`:

1. [`execute()`](https://github.com/paradigmxyz/reth/blob/main/crates/stages/stages/src/stages/tx_lookup.rs#L76):
   - Currently writes to `tables::RawTable<tables::TransactionHashNumbers>` via cursor 
   - Change to: Use `EitherWriter::new_transaction_hash_numbers()` 
   - Call `put_transaction_hash_number()` for each hash/number pair from the ETL collector

2. [`unwind()`](https://github.com/paradigmxyz/reth/blob/main/crates/stages/stages/src/stages/tx_lookup.rs#L197):
   - Currently deletes from `TransactionHashNumbers` via cursor
   - Change to: Use `EitherWriter::new_transaction_hash_numbers()`
   - Call `delete_transaction_hash_number()` for each hash to remove

fgimenez:
Closes #20389 

Changes in `TransactionLookupStage` to use `EitherWriter` abstraction for writing transaction hash mappings, enabling seamless routing to either MDBX or RocksDB based on storage settings.

  - Add `EitherWriter::commit()` for RocksDB batch commits
  - Update execute() and unwind() to use `EitherWriter::new_transaction_hash_numbers()`
  - Update test factory to register RocksDB column families with `with_default_tables()`

### DIFF
--- a/crates/stages/stages/src/stages/tx_lookup.rs
+++ b/crates/stages/stages/src/stages/tx_lookup.rs
@@ -1,5 +1,6 @@
 use alloy_eips::eip2718::Encodable2718;
 use alloy_primitives::{TxHash, TxNumber};
+use num_traits::Zero;
 use reth_config::config::{EtlConfig, TransactionLookupConfig};
 use reth_db_api::{
     table::{Decode, Decompress, Value},
@@ -153,6 +154,10 @@ where
                 let total_hashes = hash_collector.len();
                 let interval = (total_hashes / 10).max(1);
 
+                // Use append mode when table is empty (first sync) - significantly faster
+                let append_only =
+                    provider.count_entries::<tables::TransactionHashNumbers>()?.is_zero();
+
                 // Create RocksDB batch if feature is enabled
                 #[cfg(all(unix, feature = "rocksdb"))]
                 let rocksdb = provider.rocksdb_provider();
@@ -170,6 +175,7 @@ where
                     if index > 0 && index.is_multiple_of(interval) {
                         info!(
                             target: "sync::stages::transaction_lookup",
+                            ?append_only,
                             progress = %format!("{:.2}%", (index as f64 / total_hashes as f64) * 100.0),
                             "Inserting hashes"
                         );
@@ -178,7 +184,7 @@ where
                     // Decode from raw ETL bytes
                     let hash = TxHash::decode(&hash_bytes)?;
                     let tx_num = TxNumber::decompress(&number_bytes)?;
-                    writer.put_transaction_hash_number(hash, tx_num)?;
+                    writer.put_transaction_hash_number(hash, tx_num, append_only)?;
                 }
 
                 // Extract and register RocksDB batch for commit at provider level
--- a/crates/stages/stages/src/stages/tx_lookup.rs
+++ b/crates/stages/stages/src/stages/tx_lookup.rs
@@ -1,19 +1,17 @@
 use alloy_eips::eip2718::Encodable2718;
 use alloy_primitives::{TxHash, TxNumber};
-use num_traits::Zero;
 use reth_config::config::{EtlConfig, TransactionLookupConfig};
 use reth_db_api::{
-    cursor::{DbCursorRO, DbCursorRW},
-    table::Value,
+    table::{Decode, Decompress, Value},
     tables,
     transaction::DbTxMut,
-    RawKey, RawValue,
 };
 use reth_etl::Collector;
 use reth_primitives_traits::{NodePrimitives, SignedTransaction};
 use reth_provider::{
-    BlockReader, DBProvider, PruneCheckpointReader, PruneCheckpointWriter,
-    StaticFileProviderFactory, StatsReader, TransactionsProvider, TransactionsProviderExt,
+    BlockReader, DBProvider, EitherWriter, PruneCheckpointReader, PruneCheckpointWriter,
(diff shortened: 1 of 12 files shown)


crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -150,33 +150,40 @@ where
...
-                        txhash_cursor.insert(key, &RawValue::<TxNumber>::from_vec(number))?
-                    }
+                    // Decode from raw ETL bytes
+                    let hash = TxHash::decode(&hash_bytes)?;
+                    let tx_num = TxNumber::decompress(&number_bytes)?;
+                    writer.put_transaction_hash_number(hash, tx_num)?;
                 }
 
+                // Commit the writer (no-op for MDBX, commits batch for RocksDB)
+                writer.commit()?;
```
Rjected: is there any way we can put the rocksdb commits where we do the mdbx commits:
https://github.com/paradigmxyz/reth/blob/dc830b30c68ff8d4419e0a2da6f78e0c1ae6edb7/crates/stages/api/src/pipeline/mod.rs#L485

this way we only have once place where we commit and can better reason about the ordering of commits across different storages?

Rjected (commented):
one comment about commit ordering

crates/storage/provider/src/either_writer.rs
```diff
@@ -187,6 +187,20 @@ impl<'a> EitherWriter<'a, (), ()> {
 }
 
 impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N> {
+    /// Commits the writer.
+    ///
+    /// - For [`Self::Database`]: No-op, as MDBX commits happen at the provider level.
+    /// - For [`Self::StaticFile`]: No-op, as static files have their own commit mechanism.
+    /// - For [`Self::RocksDB`]: Commits the underlying batch.
+    pub fn commit(self) -> ProviderResult<()> {
```
joshieDo: yeah like dan mentioned, we should avoid having a fn commit on any eitherwriter

crates/stages/stages/src/stages/tx_lookup.rs
fgimenez: makes sense, pushed changes to move batch commits to provider.commit level, ptal

crates/storage/provider/src/either_writer.rs
fgimenez: yep done in https://github.com/paradigmxyz/reth/pull/20428/commits/8af89567f44328d7e01fd7897faa7dad8d72bd9a ptal

Rjected (approved):
lgtm

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3178,8 +3205,18 @@ impl<TX: DbTx + 'static, N: NodeTypes + 'static> DBProvider for DatabaseProvider
...
+    /// Commit database transaction, static files, and pending `RocksDB` batches.
     fn commit(self) -> ProviderResult<bool> {
+        // Commit pending RocksDB batches first.
+        // RocksDB uses WAL so data is durable once the batch commit returns.
+        #[cfg(all(unix, feature = "rocksdb"))]
+        {
+            let batches = std::mem::take(&mut *self.pending_rocksdb_batches.lock());
+            for batch in batches {
+                self.rocksdb_provider.commit_batch(batch)?;
+            }
```
joshieDo: rocksdb shouldn't be the first, but rather always the middle one

joshieDo (changes requested):
needs the commit order fix

yongkangc (approved):
nice lgtm, pending joshie's comment

crates/storage/provider/src/providers/database/provider.rs
fgimenez: sure done here https://github.com/paradigmxyz/reth/pull/20428/commits/f438161220c67cacc142c3cf37bf59e9cd84ca5c ptal

crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -150,31 +150,41 @@ where
...
-                    let key = RawKey::<TxHash>::from_vec(hash);
-                    if append_only {
-                        txhash_cursor.append(key, &RawValue::<TxNumber>::from_vec(number))?
-                    } else {
-                        txhash_cursor.insert(key, &RawValue::<TxNumber>::from_vec(number))?
-                    }
+                    // Decode from raw ETL bytes
+                    let hash = TxHash::decode(&hash_bytes)?;
+                    let tx_num = TxNumber::decompress(&number_bytes)?;
+                    writer.put_transaction_hash_number(hash, tx_num)?;
```
joshieDo: can we move this `append_only` logic inside `put_transaction_hash_number` somwhow, at least until we release rocksdb? otherwise performance on first sync will be greatly impacted for mdbx

fgimenez: sure done here https://github.com/paradigmxyz/reth/pull/20428/commits/f76a515fdf7d5e3c6797c667912509d13f526586 ptal

joshieDo (approved):

PR #20429, opened 2025-12-16
perf: batch txpool validations when possible

klkvr:


crates/transaction-pool/src/validate/task.rs
```diff
@@ -205,75 +201,72 @@ where
...
+        // Create oneshot channels for all transactions
+        let mut receivers = Vec::with_capacity(len);
+        let mut hashes = Vec::with_capacity(len);
+
+        for (origin, transaction) in transactions {
+            let hash = *transaction.hash();
+            hashes.push(hash);
+
+            let (tx, rx) = oneshot::channel();
+            receivers.push((hash, rx));
```
mattsse, line 240: this feels a bit redundant, because now this is more or less the same as `join_all(txs.into_iter().map(self.validate)` ?

mattsse (changes requested):
I believe conceptionally this is a lot better, I have one q re the batch validation send, which now looks like just a join_all

we could perhaps also allow sending/recv Vec but I guess due to the recv_many all the txs we send at once should get batched I believe

should we add a histogram for batch_size?

emmajam:
@klkvr do you still want this?

emmajam:

PR #20441, opened 2025-12-16
refactor(engine): simplify fork detection in insert_block

### LINKED ISSUE #20438: Remove fork detection logic in insert_block
The `is_fork` check in `insert_block` is only used for emitting an event, but the current implementation does a full fork detection check via `self.is_fork(block_id)`.

**Location:** https://github.com/paradigmxyz/reth/blob/c5e26546394e8b2b63db868ebe340fec47118a2d/crates/engine/tree/src/tree/mod.rs#L2511-L2519

```rust
// determine whether we are on a fork chain
// TODO: get rid of this as we only use it for the event, can make it simpler
let is_fork = match self.is_fork(block_id) {
```

**Current behavior:**
- Calls `is_fork()` which may do unnecessary work
- Only used to set a flag in the resulting event

**Suggested improvement:**
- Either inline a simpler check that is sufficient for the event
- Or defer the fork detection to when the event is actually consumed
- Or remove if the event consumer does not actually need this information

leniram159:
Replaces the expensive `is_fork()` call in `insert_block` with a simple parent hash comparison. The fork flag is only used for event emission, so a full fork detection is unnecessary.

Closes #20438

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -926,48 +926,6 @@ where
         Ok(())
     }
 
-    /// Determines if the given block is part of a fork by checking that these
-    /// conditions are true:
-    /// * walking back from the target hash to verify that the target hash is not part of an
-    ///   extension of the canonical chain.
-    /// * walking back from the current head to verify that the target hash is not already part of
-    ///   the canonical chain.
-    ///
-    /// The header is required as an arg, because we might be checking that the header is a fork
-    /// block before it's in the tree state and before it's in the database.
-    fn is_fork(&self, target: BlockWithParent) -> ProviderResult<bool> {
-        let target_hash = target.block.hash;
-        // verify that the given hash is not part of an extension of the canon chain.
-        let canonical_head = self.state.tree_state.canonical_head();
-        let mut current_hash;
-        let mut current_block = target;
-        loop {
-            if current_block.block.hash == canonical_head.hash {
-                return Ok(false)
-            }
-            // We already passed the canonical head
-            if current_block.block.number <= canonical_head.number {
-                break
-            }
-            current_hash = current_block.parent;
-
-            let Some(next_block) = self.sealed_header_by_hash(current_hash)? else { break };
-            current_block = next_block.block_with_parent();
-        }
-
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2507,14 +2465,9 @@ where
...
-        let is_fork = match self.is_fork(block_id) {
-            Err(err) => {
-                let block = convert_to_block(self, input)?;
-                return Err(InsertBlockError::new(block, err.into()).into());
-            }
-            Ok(is_fork) => is_fork,
-        };
+        // determine whether we are on a fork chain by checking if the parent is the canonical
+        // head. This is a simple check that is sufficient for the event emission below.
+        let is_fork = block_id.parent != self.state.tree_state.canonical_block_hash();
```
mattsse: this isn't sufficient because the block can also be an extension of a non canonical block

since this is just for reporting we can  go by height, we have a fork if we receive a block which number is `<= canonical.number()`

mattsse (changes requested):

crates/engine/tree/src/tree/mod.rs
leniram159: Yeah make sense, fixed now

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2507,14 +2465,11 @@ where
...
-            Err(err) => {
-                let block = convert_to_block(self, input)?;
-                return Err(InsertBlockError::new(block, err.into()).into());
-            }
-            Ok(is_fork) => is_fork,
-        };
+        // determine whether we are on a fork chain by comparing the block number with the
+        // canonical head. This is a simple check that is sufficient for the event emission below.
+        // A block is considered a fork if its number is less than or equal to the canonical head,
+        // as this indicates there's already a canonical block at that height.
```
mattsse, line 2471: imo this is sufficient for reporting purposes, technically we could also have another block on a sidechain, but imo we can ignore this case

mattsse (approved):
lgtm pending @shekhirin @klkvr

klkvr (approved):

PR #20442, opened 2025-12-16
perf: skip already executed transactions in prewarming

### LINKED ISSUE #20433: Share current transaction index between execution and prewarming
Currently, the prewarming task processes transactions independently without knowing which transactions have already been executed by the main execution path. This leads to wasted work when prewarming runs transactions that execution has already completed.

https://github.com/paradigmxyz/reth/blob/c5e26546394e8b2b63db868ebe340fec47118a2d/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L525-L528

## Solution

Share the current transaction index from the execution task to the prewarming task, allowing prewarming to skip transactions that have already been executed. This could be implemented via an `AtomicUsize` counter shared between execution and prewarming

gancerlory:
Closes #20433

Prewarming was processing transactions without knowing which ones already ran in the main execution path. Now shares an `AtomicUsize` counter so prewarming can skip the ones that are done.

Pass original tx index through prewarm channel instead of using receive order,main execution updates shared counter after each tx, prewarming skips txs with index < current executed index

BTW its strange that the todo comment that in the issue is missing....

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -34,7 +34,7 @@ use reth_revm::{database::StateProviderDatabase, db::BundleState, state::EvmStat
 use reth_trie::MultiProofTargets;
 use std::{
     sync::{
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         mpsc::{self, channel, Receiver, Sender},
         Arc,
     },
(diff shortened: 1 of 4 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -168,8 +174,8 @@ where
             }
 
             // Distribute transactions to workers
-            let mut tx_index = 0usize;
-            while let Ok(tx) = pending.recv() {
+            let mut tx_count = 0usize;
+            while let Ok((original_index, tx)) = pending.recv() {
```
klkvr: nit: could just be `index`

emmajam:

PR #20445, opened 2025-12-16
perf: only populate cache during prewarm

mattsse:
this adds a prewarm flag to the concurrent flag, see docs why we don't need to populate the cache during regular exec


klkvr (approved):

DaniPopes (approved):

Rjected (approved):
lgtm

shekhirin (approved):

PR #20448, opened 2025-12-17
perf(engine): share Arc<ExecutionOutcome> to avoid cloning BundleState

### LINKED ISSUE #20437: Arc execution output state to avoid cloning in caching
In `payload_validator.rs`, when terminating the caching task with the execution output, we pass a reference to `output.state` which likely gets cloned downstream:

https://github.com/paradigmxyz/reth/blob/c5e26546394e8b2b63db868ebe340fec47118a2d/crates/engine/tree/src/tree/payload_validator.rs#L562

We should wrap the output state in an `Arc` before passing it, so that the caching code can share ownership without a deep clone. This would reduce memory allocations and improve performance during payload validation.

yongkangc:
Closes #20437

In `payload_validator.rs`, when terminating the caching task with the execution output, `output.state` (a `BundleState`) gets cloned downstream - an expensive deep copy.

So we wrap the `ExecutionOutcome` in `Arc` at the caching boundary and share it between:
1. The prewarm cache task (`terminate_caching`)
2. The deferred trie task (`spawn_deferred_trie_task`)

This avoids cloning the expensive `BundleState` entirely.

Changes
- Make `PrewarmTaskEvent<R>` generic over receipt type, holding `Arc<ExecutionOutcome<R>>`
- Make `CacheTaskHandle<R>` and `PayloadHandle<Tx, Err, R>` generic
- Create `Arc<ExecutionOutcome>` early in `payload_validator.rs`
- Share with both `terminate_caching` and `spawn_deferred_trie_task`


crates/evm/execution-types/src/execution_outcome.rs
```diff
@@ -64,7 +64,7 @@ pub struct ExecutionOutcome<T = reth_ethereum_primitives::Receipt> {
 impl<T> Default for ExecutionOutcome<T> {
     fn default() -> Self {
         Self {
-            bundle: Default::default(),
+            bundle: Arc::new(BundleState::default()),
```
DaniPopes: Default::default should still work

crates/evm/execution-types/src/execution_outcome.rs
```diff
@@ -340,14 +345,15 @@ impl<T> ExecutionOutcome<T> {
     /// Reverts  and receipts are not updated.
     pub fn prepend_state(&mut self, mut other: BundleState) {
         let other_len = other.reverts.len();
-        // take this bundle
-        let this_bundle = core::mem::take(&mut self.bundle);
+        // take this bundle, unwrapping from Arc
+        let this_bundle =
+            Arc::try_unwrap(core::mem::take(&mut self.bundle)).unwrap_or_else(|arc| (*arc).clone());
```
DaniPopes: Arc::unwrap_or_clone

crates/evm/execution-types/src/execution_outcome.rs
```diff
@@ -78,7 +78,7 @@ impl<T> ExecutionOutcome<T> {
     /// This constructor initializes a new `ExecutionOutcome` instance with the provided
     /// bundle state, receipts, first block number, and EIP-7685 requests.
     pub const fn new(
-        bundle: BundleState,
+        bundle: Arc<BundleState>,
```
DaniPopes: does it make sense here if we need mutable access in many places?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -649,7 +649,7 @@ impl<Tx, Err> PayloadHandle<Tx, Err> {
     /// Terminates the entire caching task.
     ///
     /// If the [`BundleState`] is provided it will update the shared cache.
-    pub(super) fn terminate_caching(&mut self, block_output: Option<&BundleState>) {
+    pub(super) fn terminate_caching(&mut self, block_output: Option<Arc<BundleState>>) {
```
mattsse: since this is private api, wouldnt it be easier to give this the arc'ed execution output?, then we dont need to change all the other fields?

mattsse (changes requested):

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: Ahh good note, that would be much cleaner 👍

crates/ethereum/evm/src/test_utils.rs
```diff
@@ -128,7 +128,8 @@ impl<'a, DB: Database, I: Inspector<EthEvmContext<&'a mut State<DB>>>> BlockExec
             blob_gas_used: 0,
         };
 
-        evm.db_mut().bundle_state = bundle;
+        // Unwrap Arc or clone if shared
+        evm.db_mut().bundle_state = Arc::unwrap_or_clone(bundle);
```
DaniPopes: the clone is moved here, i would recommend constructing ExecutionOutcome beforehand, arcing it in `validate_block_with_state`, and passing that to both `terminate_caching` and `spawn_deferred_trie_task`

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -685,10 +685,9 @@ impl CacheTaskHandle {
     /// Terminates the entire pre-warming task.
     ///
     /// If the [`BundleState`] is provided it will update the shared cache.
-    pub(super) fn terminate_caching(&mut self, block_output: Option<&BundleState>) {
+    pub(super) fn terminate_caching(&mut self, block_output: Option<Arc<BundleState>>) {
         if let Some(tx) = self.to_prewarm_task.take() {
-            // Only clone when we have an active task and a state to send
-            let event = PrewarmTaskEvent::Terminate { block_output: block_output.cloned() };
```
mediocregopher, line 691: Getting rid of this clone seems like the big win here

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -549,12 +549,16 @@ where
             .into())
         }
 
-        // terminate prewarming task with good state output
-        handle.terminate_caching(Some(&output.state));
+        // Take state from output and wrap in Arc for sharing with cache task
+        let state_arc = Arc::new(std::mem::take(&mut output.state));
```
mediocregopher: nit/question: can we just do `Arc::new(output.state)` here, and let rust's borrow checker assert that `state` won't be used again for the rest of the method? Would be clearer and less error-prone going forward, rather than leaving `output` in this half-constructed state

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1021,10 +1025,12 @@ where
     /// The validation hot path can return immediately after state root verification,
     /// while consumers (DB writes, overlay providers, proofs) get trie data either
     /// from the completed task or via fallback computation.
+    #[expect(clippy::too_many_arguments)]
     fn spawn_deferred_trie_task(
         &self,
         block: RecoveredBlock<N::Block>,
-        output: BlockExecutionOutput<N::Receipt>,
+        result: reth_execution_types::BlockExecutionResult<N::Receipt>,
```
mediocregopher: super nit: let's `use` this type at the top

crates/ethereum/evm/src/test_utils.rs
yongkangc: good point, that is pretty nice 👍🏻

crates/engine/tree/src/tree/payload_validator.rs
yongkangc: Addressed in f62386a7de 👍

DaniPopes (approved):

mattsse (approved):
the additional generic leakage isn't super pretty, but this is kinda necessary

lgtm

PR #20450, opened 2025-12-17
perf(engine): deduplicate state_provider_builder calls

### LINKED ISSUE #20435: Deduplicate state_provider_builder calls in payload validator
## Summary

There are two implementations of `state_provider_builder` that could be deduplicated.

https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/mod.rs#L2483

https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/payload_validator.rs#L365

## Solution

Instead of discarding the `state_provider_builder` here https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/mod.rs#L2483 pass it as the argument to `execute` closure inside the `ctx` https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/mod.rs#L2523 to not create it again here https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/payload_validator.rs#L364-L365

yongkangc:
Closes #20435

**Problem**
Two redundant `state_provider_builder` calls during block validation - one in `EngineApiTreeHandler::insert_block_inner` and another in `BasicEngineValidator::validate_block_with_state`. Each call traverses `blocks_by_hash` to build the overlay Vec.

**Solution**
Pass both the prebuilt `StateProviderBox` and the `StateProviderBuilder` through `TreeCtx` to avoid redundant lookups.

**Impact**
- Eliminates 1× `blocks_by_hash` HashMap traversal per block validation
- Eliminates 1× `Vec<ExecutedBlock>` allocation per block validation  

**Bench** ([code](https://gist.github.com/yongkangc/8159d84558a84f12550708140f237005))

| Overlay Size | single_lookup | double_lookup | Speedup |
|--------------|---------------|---------------|---------|
| 1 block | 10.18 ns | 20.15 ns | **2.0x** |
| 10 blocks | 102.82 ns | 203.91 ns | **2.0x** |
| 50 blocks | 510.36 ns | 1.04 µs | **2.0x** |
| 100 blocks | 1.24 µs | 2.59 µs | **2.1x** |

TL;DR: **~2x faster** for the builder step by avoiding the second lookup/build.

state-provider-builder-lines
state-provider-builder-violin


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2525,7 +2527,8 @@ where
             Ok(is_fork) => is_fork,
         };
 
-        let ctx = TreeCtx::new(&mut self.state, &self.canonical_in_memory_state);
+        let mut ctx = TreeCtx::new(&mut self.state, &self.canonical_in_memory_state);
+        ctx.set_precomputed_state(state_provider, provider_builder);
```
mediocregopher: stylistic nit: would prefer a `TreeCtx::with_state_provider(self, provider, builder) -> Self` style method

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -60,13 +60,38 @@ pub struct TreeCtx<'a, N: NodePrimitives> {
...
+    /// This is set by [`crate::tree::EngineApiTreeHandler`] after validating parent state exists.
+    /// The builder is type-erased via `dyn Any` because adding a generic `P` to `TreeCtx` would
+    /// require cascading changes to the public `EngineValidator` trait signature.
+    precomputed_state: Option<PrecomputedState>,
+}
+
+/// Precomputed state provider and builder for block validation.
+///
+/// Stored as type-erased builder to avoid adding generics to `TreeCtx`.
+pub struct PrecomputedState {
```
mediocregopher: another nit: calling this "state" is a bit confusing, it's not carrying state but instead the StateProvider and Builder. Maybe keep it obvious and go with "StateProviderAndBuilder" ?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -358,20 +406,52 @@ where
...
+        // Use precomputed state from TreeCtx (set by EngineApiTreeHandler) to avoid
+        // redundant state lookups. Fall back to computing if not available (legacy callers).
+        let (mut state_provider, provider_builder) = if let Some((provider, any_builder)) =
+            ctx.take_precomputed_state()
+        {
+            // Downcast the type-erased builder to the expected type
+            match any_builder.downcast::<StateProviderBuilder<N, P>>() {
+                Ok(builder) => (provider, *builder),
+                Err(_) => {
+                    // Type mismatch should never happen - indicates a programming error
```
mediocregopher: If it's only possible via programmer error then let's just panic and reduce the lines of code

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -60,13 +60,38 @@ pub struct TreeCtx<'a, N: NodePrimitives> {
     state: &'a mut EngineApiTreeState<N>,
     /// Reference to the canonical in-memory state
     canonical_in_memory_state: &'a CanonicalInMemoryState<N>,
+    /// Optional precomputed state provider and builder to avoid redundant lookups.
+    /// This is set by [`crate::tree::EngineApiTreeHandler`] after validating parent state exists.
+    /// The builder is type-erased via `dyn Any` because adding a generic `P` to `TreeCtx` would
```
mediocregopher: I'll defer to @mattsse and @shekhirin , but imo I'd prefer to keep the type safety and leak this type onto the EngineValidator, even if it's very ugly

yongkangc: hmm yeah good point, was debating whether to make it ugly or down cast but downcast is iffy

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1129,6 +1181,9 @@ pub trait EngineValidator<
     N: NodePrimitives = <<Types as PayloadTypes>::BuiltPayload as BuiltPayload>::Primitives,
 >: Send + Sync + 'static
 {
+    /// The provider type used for state lookups.
+    type Provider: Send + Sync;
```
mediocregopher: Provider cannot be Sync, see #20516 . If it's possible to skip the Send as well we should do so, better to keep things as unconstrained as possible

yongkangc: i see, thanks for the note!

yongkangc: @mediocregopher just removed the trait bound for send and sync

mediocregopher (approved):

mattsse (commented):
these are quite a few changes just to remove a duplicate function

Issue #20415, opened 2025-12-16
Add engine_getBlobsV3 support


mattsse:
### Describe the feature

ref https://github.com/ethereum/execution-apis/pull/719

this is a mix of

https://github.com/paradigmxyz/reth/blob/bbd51862d4e95e81b713cd61e97cfd18f02b9107/crates/transaction-pool/src/blobstore/disk.rs#L204-L207

and behaviour from

https://github.com/paradigmxyz/reth/blob/bbd51862d4e95e81b713cd61e97cfd18f02b9107/crates/transaction-pool/src/blobstore/disk.rs#L145-L148

basically dont short circuit if we dont have all entris, like we need to do for v2

https://github.com/paradigmxyz/reth/blob/bbd51862d4e95e81b713cd61e97cfd18f02b9107/crates/transaction-pool/src/blobstore/disk.rs#L261-L266

so basically, all of this already exists in the v2 impl so we can extract this and only keep this check in the v2

https://github.com/paradigmxyz/reth/blob/bbd51862d4e95e81b713cd61e97cfd18f02b9107/crates/transaction-pool/src/blobstore/disk.rs#L261-L266

but not v3

we also need to bubble up the api from

https://github.com/paradigmxyz/reth/blob/bbd51862d4e95e81b713cd61e97cfd18f02b9107/crates/rpc/rpc-api/src/engine.rs#L234-L242

### Additional context

@0xKarl98

0xKarl98:
Can assign me

figtracer:
this can be closed

PR #20451, opened 2025-12-17
feat: add engine_getBlobsV3 method

0xKarl98:
This closes issue #20415
cc @mattsse  @klkvr

### DIFF
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -1383,6 +1383,7 @@ where
         };
         let blobs_disabled =
             ctx.config.txpool.disable_blobs_support || ctx.config.txpool.blobpool_max_count == 0;
+        
         Ok(EngineApi::new(
             ctx.node.provider().clone(),
             ctx.config.chain.clone(),
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -1381,6 +1381,8 @@ where
             version: version_metadata().cargo_pkg_version.to_string(),
             commit: version_metadata().vergen_git_sha.to_string(),
         };
+        let blobs_disabled =
+            ctx.config.txpool.disable_blobs_support || ctx.config.txpool.blobpool_max_count == 0;
         Ok(EngineApi::new(
             ctx.node.provider().clone(),
             ctx.config.chain.clone(),
@@ -1392,6 +1394,8 @@ where
             EngineCapabilities::default(),
             engine_validator,
             ctx.config.engine.accept_execution_requests_hash,
+            ctx.node.network().clone(),
+            !blobs_disabled,
         ))
     }
 }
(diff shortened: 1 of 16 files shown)


crates/rpc/rpc-builder/tests/it/utils.rs
```diff
@@ -54,6 +54,8 @@ pub async fn launch_auth(secret: JwtSecret) -> AuthServerHandle {
         EngineCapabilities::default(),
         EthereumEngineValidator::new(MAINNET.clone()),
         false,
+        NoopNetwork::default(),
+        true,
```
mattsse: why do we need this change

crates/node/builder/src/rpc.rs
```diff
@@ -1381,6 +1381,9 @@ where
             version: version_metadata().cargo_pkg_version.to_string(),
             commit: version_metadata().vergen_git_sha.to_string(),
         };
+        let blobs_disabled =
+            ctx.config.txpool.disable_blobs_support || ctx.config.txpool.blobpool_max_count == 0;
```
mattsse: don't think this is related to this pr?

crates/transaction-pool/src/blobstore/disk.rs
```diff
@@ -59,6 +59,65 @@ impl DiskFileBlobStore {
     fn clear_cache(&self) {
         self.inner.blob_cache.lock().clear()
     }
+
+    fn get_by_versioned_hashes_eip7594(
```
mattsse, line 71: let's add some docs here about how this behaves

mattsse (changes requested):
this looks great already, only have a question re some additional code changes

crates/node/builder/src/rpc.rs
0xKarl98: Because in the spec , it states as this :
Client software **MUST** return `null` if syncing or otherwise unable to generally serve blob pool data.

So i use this to identify if we are currently syncing or any other reasons that causes blob pool can't serve

0xKarl98: After think about it again , we can remove this blobs_disabled , just keep the judge of if syncing

crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -792,6 +796,35 @@ where
...
+        }
+
+        if versioned_hashes.len() > MAX_BLOB_LIMIT {
+            return Err(EngineApiError::BlobRequestTooLarge { len: versioned_hashes.len() })
+        }
+
+        // Spec requires returning `null` if syncing.
+        if (*self.inner.is_syncing)() {
+            return Ok(None)
+        }
```
mattsse, line 819: ah I see, now this makes sense

mattsse (approved):

PR #20452, opened 2025-12-17
fix(payload): clarify PayloadTransactions mark_invalid semantics

phrwlk:
Clarified the contract of PayloadTransactions::mark_invalid so it matches how block builders and the transaction pool actually use it. Instead of promising to exclude only nonce descendants, the docs now state that implementations must avoid returning transactions that depend on the invalid one, and explicitly allow filtering by sender. The comment on BestPayloadTransactions was also updated to document that it drops all further transactions from the given sender in the current iteration, mirroring 
BestTransactions::mark_invalid.


mattsse (approved):
fine

PR #20453, opened 2025-12-17
chore(downloaders): improve panic messages in ReverseHeadersDownloader

SophiaR-baty:
Replace vague expect messages with specific field and condition descriptions, improving debuggability when invariants are violated.


mattsse (commented):
these are unreachable we can keep as is

PR #20455, opened 2025-12-17
docs(exex): document ChainRevert flow in how-it-works

sashass1315:
Updated the ExEx architecture mermaid sequence in how-it-works.mdx to include a dedicated “Revert Flow” using ChainRevert notifications. This makes the docs consistent with the actual ExExNotification variants and clarifies that ExExes are expected to handle reverts in addition to commits and reorgs.


mattsse (approved):
seems alright

PR #20456, opened 2025-12-17
chore: fix blob count in validation benchmark

dizer-ti:
The truncate logic was inside the while loop, so it never executed when the initial arbitrary-generated sidecar already had more blobs than num_blobs. This caused the benchmark to test with wrong blob counts (e.g. testing "1 blob" with 3+ blobs).

Moved truncate outside the loop to ensure exact blob count regardless of initial sidecar size.


mattsse (approved):
fine

PR #20459, opened 2025-12-17
perf: remove redundant contains_key check in ProofSequencer::add_proof

### LINKED ISSUE #20439: Measure and maybe cache precompile cache initialization in payload execution
Currently in payload validation, we call `map_pure_precompiles` to wrap each precompile with caching on every block execution:

https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/payload_validator.rs#L630-L645

This happens for every `execute_block_and_verify_preimages` call. The initialization involves:
- Iterating over all pure precompiles
- HashMap lookups/insertions for metrics
- Creating `CachedPrecompile` wrappers

We should measure how long this takes and investigate if the executor with cached precompiles can be initialized once and reused across blocks.

yongkangc:
Closes: https://github.com/paradigmxyz/reth/issues/20439
Removes a redundant `contains_key()` check inside the `while` loop in `ProofSequencer::add_proof`.

- The loop uses `while let Some(pending) = self.pending_proofs.remove(&current_sequence)` which naturally exits when the key doesn't exist. The inner `contains_key()` check on line 175 is unnecessary since `remove()` will return `None` on the next iteration anyway.

- So we can remove to save one `BTreeMap::contains_key()` lookup (O(log n)) per iteration while maintaining identical behavior.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -170,11 +170,6 @@ impl ProofSequencer {
         while let Some(pending) = self.pending_proofs.remove(&current_sequence) {
             consecutive_proofs.push(pending);
             current_sequence += 1;
-
-            // if we don't have the next number, stop collecting
-            if !self.pending_proofs.contains_key(&current_sequence) {
-                break;
-            }
         }
 
         self.next_to_deliver += consecutive_proofs.len() as u64;


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -170,11 +170,6 @@ impl ProofSequencer {
         while let Some(pending) = self.pending_proofs.remove(&current_sequence) {
```
yongkangc, line 170: this stops when the next sequence is missing, so we have the same behaviour

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -170,11 +170,6 @@ impl ProofSequencer {
...
             consecutive_proofs.push(pending);
             current_sequence += 1;
-
```
yongkangc, line 173: The while let with remove already stops when the next sequence is missing, so the behaviour should be the same.

onbjerg (approved):

PR #20465, opened 2025-12-17
chore(primitives-traits): correct SealedBlock::senders return description

radik878:
SealedBlock::senders returns `Result<Vec<Address>, RecoveryError>`, but the doc comment still claimed it returns `None` when sender recovery fails.


mattsse (approved):
fine

PR #20466, opened 2025-12-17
chore(trie): Rm redundant clone of propagated error

emhane:
Removes redundant clone of error that is propagated directly anyway

### DIFF
--- a/crates/trie/trie/src/node_iter.rs
+++ b/crates/trie/trie/src/node_iter.rs
@@ -160,15 +160,15 @@ where
     ///
     /// If `metrics` feature is enabled, it also updates the metrics.
     fn next_hashed_entry(&mut self) -> Result<Option<(B256, H::Value)>, DatabaseError> {
-        let result = self.hashed_cursor.next();
+        let next = self.hashed_cursor.next()?;
 
-        self.last_next_result = result.clone()?;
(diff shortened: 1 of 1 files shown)


crates/trie/trie/src/node_iter.rs
```diff
@@ -160,15 +160,15 @@ where
     ///
     /// If `metrics` feature is enabled, it also updates the metrics.
     fn next_hashed_entry(&mut self) -> Result<Option<(B256, H::Value)>, DatabaseError> {
-        let result = self.hashed_cursor.next();
+        let next = self.hashed_cursor.next()?;
 
-        self.last_next_result = result.clone()?;
+        self.last_next_result = next.clone();
```
mattsse: isn't this the same as just `.cloned()?`

mattsse (changes requested):

crates/trie/trie/src/node_iter.rs
emhane: does the compiler optimise away the clone when the result is an error that will be propagated?

mattsse: if this is an error then `.cloned()` isnt invoked

mattsse (approved):

PR #20468, opened 2025-12-17
feat(engine): Prefetch storage and accounts when BAL is provided

mediocregopher:
The primary change of this PR is to add a new code-path to the prewarm tasks. In this code-path the tasks will prefetch storage slots from the BAL. The main prewarm task divides the BAL into ranges of slots, and each task iterates over its range. Because the BAL is just Vecs of Vecs of Vecs we can partition the work this way without requiring any coordination or work scheduling. Each prewarm worker then iterates over its range of slots and fetches each using its CachedStateProvider, warming the cache.

In addition the MultiProofTask's BAL code-path has been updated to handle account prewarming. Since it needs to fetch accounts in order to create the HashedPostState for the block anyway, we pass a CachedStateProvider into it in order to warm the accounts in the cache at the same time.

This also contains a small fix the the BAL->HashedPostState code, so that it doesn't include accounts with no changes in the HashedPostState.


crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -1,17 +1,124 @@
...
 use reth_primitives_traits::Account;
 use reth_provider::{AccountReader, ProviderError};
 use reth_trie::{HashedPostState, HashedStorage};
+use std::ops::Range;
+
+/// Returns the total number of storage slots (both changed and read-only) across all accounts in
+/// the BAL.
+pub fn total_slots(bal: &BlockAccessList) -> usize {
+    bal.iter().map(|account| account.storage_changes.len() + account.storage_reads.len()).sum()
+}
```
mattsse, line 16: @Rimeeeeee @Soubhik-10 that's why we should proceed with https://github.com/alloy-rs/eips/pull/47

crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -1,17 +1,124 @@
...
+
+/// Iterator over storage slots in a [`BlockAccessList`], with range-based filtering.
+///
+/// Iterates over all `(Address, StorageKey)` pairs representing both changed and read-only
+/// storage slots across all accounts in the BAL. For each account, changed slots are iterated
+/// first, followed by read-only slots. The iterator intelligently skips accounts and slots
+/// outside the specified range for efficient traversal.
+#[derive(Debug)]
+pub(crate) struct BALSlotIter<'a> {
```
mattsse, line 25: can even upstream this entire type I think

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -43,6 +47,14 @@ use std::{
 };
 use tracing::{debug, debug_span, instrument, trace, warn, Span};
 
+/// Determines the prewarming mode: transaction-based or BAL-based.
+pub(super) enum PrewarmMode<Tx> {
+    /// Prewarm by executing transactions from a stream.
+    Transactions(Receiver<Tx>),
+    /// Prewarm by prefetching slots from a Block Access List.
+    BlockAccessList(Arc<BlockAccessList>),
+}
```
mattsse, line 56: I see, we definitely need these two modes until BAL is the norm

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -286,6 +298,86 @@ where
...
+
+        // Calculate number of workers needed (at most max_concurrency)
+        let workers_needed = total_slots.min(self.max_concurrency);
+
+        // Calculate slots per worker
+        let slots_per_worker = total_slots / workers_needed;
+        let remainder = total_slots % workers_needed;
+
+        // Spawn workers with their assigned ranges
+        for i in 0..workers_needed {
```
mattsse, line 348: we just did https://github.com/paradigmxyz/reth/pull/20575 , so I assume we can do the same here as well?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -297,13 +389,22 @@ where
...
+        Tx: ExecutableTxFor<Evm> + Clone + Send + 'static,
+    {
+        // Spawn execution tasks based on mode
+        match mode {
+            PrewarmMode::Transactions(pending) => {
+                self.spawn_all(pending, actions_tx);
+            }
+            PrewarmMode::BlockAccessList(bal) => {
+                self.run_bal_prewarm(bal, actions_tx);
+            }
```
mattsse, line 406: this makes a lot of sense to me

mattsse (approved):
all of this makes sense,

this is now an entirely separate branch and doesn't change how prewarming works until we receive a BAL (which is currently a noop) so this is de-facto unreachable for now

one q re fetch spawning @klkvr

crates/engine/tree/src/tree/payload_processor/bal.rs
```diff
@@ -75,9 +188,7 @@ where
                 }
             }
 
-            if !storage_map.storage.is_empty() {
-                hashed_state.storages.insert(hashed_address, storage_map);
-            }
+            hashed_state.storages.insert(hashed_address, storage_map);
```
shekhirin, line 191: why do we not check empty storage now?

shekhirin (approved):
LGTM, one question

crates/engine/tree/src/tree/payload_processor/bal.rs
mediocregopher: `storage_map` could only be empty if any `slot_changes` is empty, which wouldn't be included in the BAL in the first place. If an empty `storage_map` gets inserted here there shouldn't be any problems downstream anyway, so I figured it was a pointless check.

klkvr (approved):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
mediocregopher: Spawning the tx prewarm workers in a separate task was helpful because it allowed us to immediately begin pushing txs onto the channel, so the workers could immediately start working. In this case there is no channel for the workers to read from, the work is pre-partitioned using the Range, so I don't think spawning them in a separate task would actually change anything.

PR #20470, opened 2025-12-17
fix: implement Debug trait for EngineService

PivasDesant:
Implement Debug for EngineService

Same pattern as EngineApi - just using finish_non_exhaustive() since we can't derive it.


mattsse (commented):
this isnt needed

PR #20472, opened 2025-12-17
Stop overcounting tx offsets when iterating in-memory blocks

conomist:
Ensure block body indices calculation breaks once the target in-memory block is reached, so first_tx_num isn’t advanced past the requested block while still setting the correct tx_count.


mattsse (commented):
unsure here @joshieDo

PR #20473, opened 2025-12-17
Remove redundant PartialEq bound from NodePrimitives

splinter012:
Drop the explicit PartialEq bound from NodePrimitives; Eq already implies PartialEq, so this tightens the trait contract without changing behavior. Keeps API and implementations untouched; only removes duplicate constraint noise.


mattsse (commented):
can keep as is

PR #20475, opened 2025-12-18
perf(multiproof): batch consecutive state updates

### LINKED ISSUE #20223: perf: batch consecutive transaction state updates in multiproof task

yongkangc:
Closes #20223 by batching consecutive transaction state updates

 Batch `EvmState` updates using a custom `merge_evm_state` function that preserves `original_value` from the base state while taking `present_value` from the overlay. This ensures `is_changed()` returns correct results after batching.

**Previous SR Root Mismatch Attempt**

The bug was caused by batching at the `EvmState` level. When `EvmState::extend` merges states, storage slot `is_changed` flags from later transactions overwrite those from earlier ones. The `evm_state_to_hashed_post_state` function then filters by `is_changed()`, causing changed slots to be dropped.

e.g
- TX1 writes slot A (is_changed=true)
- TX2 reads slot A (is_changed=false)
- After `EvmState::extend`, slot A has `is_changed=false`
- Slot A is filtered out by `evm_state_to_hashed_post_state`
- Sparse trie never sees the change → **state root mismatch**

[Mermaid digram to explain](https://mermaid.live/edit#pako:eNqtVU9vmzAc_Sqee8kkGkFCILXUHdq00qQOqrXTqi0TcuBHQAWMbJO1q3rcbbcddum-XD_JbJxkJFGnTqoPlnnxe79_j3CHY5YAJjgt2Nc4o1yiy6NphdQSzWzOaZ2h8GzyeYrVTtDJoryQVAIhcCOhSlBvxtk1VK-n-Ith6ZXkHGKZs2qtpVfoKJXLK4cgUTCJrhDj-TyvaHFoW6jmIKCSh46NchGpRKo5JIeSN4AeH35uqIeDVmdAEAeaiB0150m5lBYCNrWGSmtGBfRNPT22AF7Q2816QlfdegdcaSyjkWeHU-n_2hQbKTFYlJHQjYwkizIqMkiimglpQJTmhQQu0Oy2o9fbSspTOo-_f6ALU__kfXh-fjJB7XwQZwosc1FSGWevNojdiaD9_TeqCd3m7kJDA7ndhhho1C3LQJ6BVC_NYctLwclHlbbaCSp1Q6N1J1AvZlz75jleCl7IS8ELeinQXtou6hqgVnPUDluJRgtaqFQkvQaBln7rL4OY3zZl_2W-_6z2hawXtNZ7-L6y3tvg-OzDRHnv2EwQibUHn7JeYKwXDLuz2IWM9QK32w4DjbplGahrPXOMCyrEBFIEnDOuayvI3vjIVssSUv9zkb3T0_YxZgXjZC9N0y2qaOIYhFiSbdtzO2Tb1vSnyTVnHbJDHTqANdmlLvXoDrkjoV4nk3oXC7xVThs3h1boWuHICoZW4FqqJ8vY2MJznieYaEtY2qAl1Y_4TtOnWGZQKscRdUwgpU0hp3ha3StaTatPjJUrJmfNPMOkNb6FmzpRM57kVL3Xf6-o3gM_Zk0lMTloFTC5wzeYDMZ-fzy2bce31e47tm_hW0zcg7439l3fde3hwHMPnHsLf2tj2v2xP7IwJLlk_J35RrWfqvs_HWcMQg)

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -5,14 +5,15 @@ use alloy_eip7928::BlockAccessList;
 use alloy_evm::block::StateChangeSource;
 use alloy_primitives::{
     keccak256,
-    map::{B256Set, HashSet},
-    B256,
+    map::{B256Set, HashMap, HashSet},
+    Address, B256, U256,
 };
 use crossbeam_channel::{unbounded, Receiver as CrossbeamReceiver, Sender as CrossbeamSender};
 use dashmap::DashMap;
 use derive_more::derive::Deref;
 use metrics::{Gauge, Histogram};
 use reth_metrics::Metrics;
+use reth_primitives_traits::Account;
 use reth_provider::AccountReader;
 use reth_revm::state::EvmState;
 use reth_trie::{
@@ -110,10 +111,11 @@ pub(super) enum MultiProofMessage {
     PrefetchProofs(MultiProofTargets),
     /// New state update from transaction execution with its source.
     ///
-    /// The state is already converted to a `HashedPostState` by the producer (state hook).
-    /// This avoids issues with `EvmState::extend` overwriting `is_changed` flags, which
-    /// could cause state updates to be lost during batching.
-    StateUpdate(Source, HashedPostState),
+    /// The state is "sealed" - the `is_changed` filter has been applied at the producer,
+    /// but keys are not yet hashed. This allows:
+    /// 1. Safe batching (no `is_changed` flags to corrupt)
+    /// 2. Deferred hashing (hash once after batching, not per-TX)
+    StateUpdate(Source, SealedStateUpdate),
     /// State update that can be applied to the sparse trie without any new proofs.
     ///
     /// It can be the case when all accounts and storage slots from the state update were already
@@ -211,7 +213,7 @@ impl Drop for StateHookSender {
 pub(crate) fn evm_state_to_hashed_post_state(update: &EvmState) -> HashedPostState {
     let mut hashed_state = HashedPostState::with_capacity(update.len());
 
-    for (address, account) in update.iter() {
+    for (address, account) in update {
         if account.is_touched() {
             let hashed_address = keccak256(*address);
             trace!(target: "engine::tree::payload_processor::multiproof", ?address, ?hashed_address, "Adding account to state update");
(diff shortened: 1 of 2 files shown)


gakonst:
What’s the perf impact?

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -216,19 +216,25 @@ pub(crate) fn evm_state_to_hashed_post_state(update: EvmState) -> HashedPostStat
...
-                .peekable();
-
             if destroyed {
                 hashed_state.storages.insert(hashed_address, HashedStorage::new(true));
-            } else if changed_storage_iter.peek().is_some() {
-                hashed_state
-                    .storages
-                    .insert(hashed_address, HashedStorage::from_iter(false, changed_storage_iter));
+            } else {
+                let changed_storage: Vec<_> = account
```
yongkangc, line 222: Collect changed storage slots directly to avoid Peekable allocation

PR #20477, opened 2025-12-18
docs: fix pruning segments list and config field name

aso20455:
1. Added missing `Bodies History` segment to the pruning segments list in `pruning.mdx`. This segment exists in the codebase (`bodies_history` field in `PruneModes` struct) and is documented in `configuration.mdx`, but was missing from the segments overview.

2. Updated `connect_trusted_nodes_only` to `trusted_nodes_only` in `configuration.mdx`. The actual field name in `PeersConfig` is `trusted_nodes_only`, while `connect_trusted_nodes_only` is just a serde alias kept for backwards compatibility.

3. Added `Bodies History` column to all RPC compatibility tables in the Pruned Node section. Methods that require historical transaction data (like `eth_getBlockByHash`, `debug_traceBlock`, `trace_block`, etc.) are marked as incompatible.


mattsse (commented):

Issue #20392, opened 2025-12-15
Initialize RocksDB provider and add consistency check on startup


fgimenez:
Requires #20340

Wire up RocksDB initialization during node launch and add a consistency check to verify data integrity.

1. Node launch initialization: ✅ 
   - In [`crates/node/builder/src/launch/common.rs`](https://github.com/paradigmxyz/reth/blob/main/crates/node/builder/src/launch/common.rs):
   - Create `RocksDBProvider` using `RocksDBBuilder::new(path).with_default_tables().build()`
   - Store in node components for access by stages and providers

2. Pass RocksDB to stages: (moved to a new issue [20593](https://github.com/paradigmxyz/reth/issues/20593))
   - `TransactionLookupStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - `IndexStorageHistoryStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - `IndexAccountHistoryStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - Options:
     - Add `RocksDBProvider` to stage structs
     - Pass via provider trait
     - Use a shared Arc reference

3. Consistency check on startup, call existing `check_consistency()` from #20340 ✅ 

4. Configuration validation: . moved to https://github.com/paradigmxyz/reth/issues/20482
   - Warn or error if user tries to change RocksDB flags after initial sync
   - The flags should only be set at genesis initialization

PR #20478, opened 2025-12-18
feat(storage): add RocksDB consistency check on startup

yongkangc:
Addresses [#20392](https://github.com/paradigmxyz/reth/issues/20392) - Initialize RocksDB provider and add consistency check on startup.

## Changes

- Add RocksDB `check_consistency()` call in node launch with correct ordering:
  1. `StaticFileProvider::check_file_consistency()` - heals interrupted writes
  2. `RocksDB::check_consistency()` - uses static file data for tx hash pruning  
  3. `StaticFileProvider::check_consistency()` - compares with MDBX checkpoints

Note for future:

- We have to follow up with passing RocksDB to the stages once implemented in https://github.com/paradigmxyz/reth/issues/20389
- And also we can spin off another pr for Configuration validation


crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,50 @@ where
...
+        if let Some(unwind_block) = unwind_target {
             // Highly unlikely to happen, and given its destructive nature, it's better to panic
             // instead.
+            let inconsistency_source = match (static_file_unwind, rocksdb_unwind) {
+                (Some(_), Some(_)) => "static file <> database and RocksDB <> database",
+                (Some(_), None) => "static file <> database",
+                (None, Some(_)) => "RocksDB <> database",
+                (None, None) => unreachable!(),
+            };
             assert_ne!(
```
fgimenez, line 551: this will panic when rocksdb returns `Some(0)` as unwind target, which can happen for instance when an existing node enables rocksdb flags. is this what we want? this will be a common case for nodes migrating to using rocksdb, maybe the message in the panic is not clear enough, or we want to let them unwind to 0 given that they will need to resync in any case, correct?

also, i see the comment "Highly unlikely to happen, and given its destructive nature, it's better to panic" given that it is more likely to happen now, if we want this behavior we should update the comment, wdyt?

yongkangc: Makes sense, addressed this by sending warning instead! Thanks for note

crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,63 @@ where
...
+            .check_consistency(&factory.provider()?)?
+            .map(|target| match target {
+                PipelineTarget::Unwind(block) => block,
+                PipelineTarget::Sync(_) => unreachable!("check_consistency returns Unwind"),
+            });
+
+        // RocksDB consistency check - runs after static files since it may use static file data
+        // for pruning transaction hashes.
+        let rocksdb_unwind =
+            factory.rocksdb_provider().check_consistency(&factory.database_provider_ro()?)?;
```
joshieDo: hm, this will be slightly trickier i think.

these two need to happen the other way around: eg. imagine a scenario where rocksdb and sf are at block N, ahead of the checkpoint.

if we prune from static files first, rocksdb wouldnt be able to delete the excess of transaction hashes, because static file would no longer have the transactions to hash

--

There's another matter at play here: `StaticFileProvider::check_consistency` does 2 types of consistency checks:

https://github.com/paradigmxyz/reth/blob/b3a792ad1eb468876a6ee1fe95a0ddc4a8dcca5a/crates/storage/provider/src/providers/static_file/manager.rs#L995-L998

`1) When a static file fails to commit but the underlying data was changed.`

Specifically this one should be done first:
https://github.com/paradigmxyz/reth/blob/b3a792ad1eb468876a6ee1fe95a0ddc4a8dcca5a/crates/storage/provider/src/providers/static_file/manager.rs#L1090-L1105

and then the order would become:

* `static_file_provider.check_file_consistency()` (internal to static files)
* `rocksdb_provider.check_consistency()` ( vs mdbx checkpoints where data comes from sf  (eg. tx from a txhash))
* `static_file_provider.check_storage_consistency()` (vs mdbx checkpoints)

can you refactor `check_file_consistency` and `check_storage_consistency` in the staticfileprovider in another PR first?  maybe with better names than these ha

crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,63 @@ where
...
+                (Some(_), Some(_)) => "static file <> database and RocksDB <> database",
+                (Some(_), None) => "static file <> database",
+                (None, Some(_)) => "RocksDB <> database",
+                (None, None) => unreachable!(),
+            };
+
+            // Unwind to block 0 from static file inconsistency is destructive and likely
+            // indicates corruption - panic to prevent data loss.
+            // However, RocksDB returning unwind to 0 is expected when migrating an existing
+            // node to use RocksDB flags - allow this case to proceed with a full resync.
```
joshieDo: no, do not - we're not planning to support that because would leave MDBX with a huge free list size

joshieDo: this unwind target is for the whole pipeline and all its stages, not just one stage

in the past, cases to unwind to 0 were some misconfig (eg. forget to copy over something) which would have been disastrous  if let through

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -120,7 +127,12 @@ impl RocksDBProvider {
                     highest_static_tx = highest_tx,
                     "Static files ahead of MDBX, pruning TransactionHashNumbers excess data"
                 );
-                self.prune_transaction_hash_numbers_in_range(provider, (mdbx_tx + 1)..=highest_tx)?;
+                // Use checked_add to prevent overflow when mdbx_tx == u64::MAX
+                if let Some(start) = mdbx_tx.checked_add(1) {
```
joshieDo: yeah i dont think we need to worry about that ?

crates/node/builder/src/launch/common.rs
yongkangc: thanks for the note @joshieDo 

split into another issue here https://github.com/paradigmxyz/reth/issues/20506

crates/node/builder/src/launch/common.rs
yongkangc: ahh okay thats terrible. thanks for the note

addressed in 2af821d

crates/storage/provider/src/providers/rocksdb/invariants.rs
yongkangc: i guess we wont ever reach there haha

addressed here [2af821d](https://github.com/paradigmxyz/reth/pull/20478/commits/2af821d5fb30fd5736b0367d433e6b46aabadab0)

crates/node/builder/src/launch/common.rs
yongkangc: addressed this here 
https://github.com/paradigmxyz/reth/blob/8eb4ba2f65f4b8e399aff45232da172526389d4b/crates/node/builder/src/launch/common.rs#L505-L530

crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,61 @@ where
...
-                unwind_target,
-                PipelineTarget::Unwind(0),
-                "A static file <> database inconsistency was found that would trigger an unwind to block 0"
+                unwind_block,
+                0,
+                "A {inconsistency_source} inconsistency was found that would trigger an unwind to block 0"
             );
 
-            info!(target: "reth::cli", unwind_target = %unwind_target, "Executing an unwind after a failed storage consistency check.");
+            let unwind_target = PipelineTarget::Unwind(unwind_block);
```
yongkangc, line 557: we need this now as we are combining 3 different targets

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -69,6 +69,13 @@ impl RocksDBProvider {
             unwind_target = Some(unwind_target.map_or(target, |t| t.min(target)));
         }
 
+        // Check AccountsHistory if stored in RocksDB
```
yongkangc, line 72: this was missing from 

https://github.com/paradigmxyz/reth/pull/20340/changes

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1184,6 +1184,44 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+    /// (indicating a pruning interruption that needs a database unwind).
+    pub fn check_file_consistency(&self) -> ProviderResult<Option<BlockNumber>> {
+        info!(target: "reth::cli", "Healing static file inconsistencies.");
+
+        let mut unwind_target: Option<BlockNumber> = None;
+
+        for segment in StaticFileSegment::iter() {
+            let (initial_highest_block, highest_block) = self.maybe_heal_segment(segment)?;
+
+            if initial_highest_block != highest_block {
```
yongkangc, line 1208: logic based on `check_consistency` 

https://github.com/paradigmxyz/reth/blob/8eb4ba2f65f4b8e399aff45232da172526389d4b/crates/storage/provider/src/providers/static_file/manager.rs#L1015

PR #20484, opened 2025-12-18
fix: session config should be read from config file

cuiweixie:


mattsse (approved):
oh, yeah this should be the configured one...

nice find, ty

PR #20485, opened 2025-12-18
refactor: remove dead storage multiproof path

Snezhkko:
Removed the unused StorageMultiproofInput / PendingMultiproofTask indirection in the multiproof task. MultiproofManager::dispatch now takes MultiproofInput directly and only routes account multiproof jobs.
The removed storage multiproof path was never constructed anywhere in the codebase, so the StorageMultiproofInput branch was effectively dead code.


mattsse (commented):
this actually seems alright

ptal @yongkangc

mediocregopher (approved):

mattsse (approved):

PR #20487, opened 2025-12-18
perf(engine): prevent duplicate block insertion in BlockBuffer

reallesee:
Add early return in `BlockBuffer::insert_block` when block already exists. Without this check, inserting the same block twice would add duplicate entries to `block_queue` (VecDeque), breaking FIFO eviction logic while other collections (HashSet-based) silently ignored duplicates. Follows the same pattern used in `TreeState::insert_executed`.


crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -66,6 +66,10 @@ impl<B: Block> BlockBuffer<B> {
     pub fn insert_block(&mut self, block: SealedBlock<B>) {
         let hash = block.hash();
 
+        if self.blocks.contains_key(&hash) {
+            return;
+        }
```
mattsse: seems fine to do but we should use the entry api here because we do `self.blocks.insert(hash, block);` if its not in there

mattsse (changes requested):

reallesee:
@mattsse pinging

crates/engine/tree/src/tree/block_buffer.rs
reallesee: > seems fine to do but we should use the entry api here because we do `self.blocks.insert(hash, block);` if its not in there

thanks, fixed

crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -66,16 +66,20 @@ impl<B: Block> BlockBuffer<B> {
...
-            // Evict oldest block if limit is hit
-            if let Some(evicted_hash) = self.block_queue.pop_front() {
-                self.remove_block(&evicted_hash);
-            }
+        // Evict oldest block if capacity is reached
+        if self.block_queue.len() >= self.max_blocks &&
+            let Some(evicted_hash) = self.block_queue.pop_front()
+        {
+            self.remove_block(&evicted_hash);
         }
```
mattsse: please undo this

crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -66,16 +66,20 @@ impl<B: Block> BlockBuffer<B> {
     pub fn insert_block(&mut self, block: SealedBlock<B>) {
         let hash = block.hash();
 
+        let entry = match self.blocks.entry(hash) {
+            std::collections::hash_map::Entry::Occupied(_) => return,
+            std::collections::hash_map::Entry::Vacant(entry) => entry,
```
mattsse: we can insert right here

mattsse (changes requested):

crates/engine/tree/src/tree/block_buffer.rs
reallesee: @mattsse undoing this triggers clippy::collapsible_if because of the nested ifs - collapse or add #[allow]?

mattsse: it only does this because you also modified the comments, please keep L78 and below unchanged

mattsse (changes requested):

mattsse (approved):

PR #20488, opened 2025-12-18
perf(db): eliminate redundant allocations in serialization benchmarks

0xxFloki:
Refactored measure_table_serialization to remove unnecessary per-iteration cloning of benchmark input data, improving benchmark accuracy and reducing memory overhead.


mattsse (commented):
we can keep this as is

PR #20489, opened 2025-12-18
perf(trie): remove more unnecessary channels

figtracer:
same reasoning as #20406

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -166,8 +166,7 @@ where
 
     // Update storage slots with new values and calculate storage roots.
     let span = tracing::Span::current();
-    let (tx, rx) = mpsc::channel();
-    state
+    let results: Vec<_> = state
         .storages
         .into_iter()
         .map(|(address, storage)| (address, storage, trie.take_storage_trie(&address)))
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -166,8 +166,7 @@ where
 
     // Update storage slots with new values and calculate storage roots.
     let span = tracing::Span::current();
-    let (tx, rx) = mpsc::channel();
-    state
+    let results: Vec<_> = state
```
mattsse, line 169: I believe this makes sense @mediocregopher 

should we even do `trie.take_storage_trie(&address).collect` first so that we can replace .par_bridge() with just into_par_iter?

mattsse (approved):
lgtm, pending @mediocregopher 

cc @klkvr

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
mediocregopher: When I originally wrote this my goal was to reduce allocations as much as possible. I recognize that a channel also can have allocs attached, but for large amounts of storage tries my bet was that it would work out less.

If we want to get rid of this channel now that's fine, probably it doesn't matter, if it turns out it does we can add it back, but imo let's not go allocating more than we have to.

mediocregopher (approved):

PR #20490, opened 2025-12-18
fix(flashblocks): add chainstate check on consensus update and add flashblock websocket connection retry delay

sieniven:


sieniven:
Closing this PR because flashblock consensus is intended to only run as the only source to advance the chainstate.

PR #20493, opened 2025-12-18
perf(db): remove unnecessary vector clone in benchmark setup

0xlupin:
The `set_up_db` benchmark helper was accepting `&Vec<...>` and cloning  the entire vector before iteration, causing unnecessary allocations on  the hot path.


mattsse (commented):
we can keep as is

PR #20498, opened 2025-12-18
chore(cli): correct p2p body error message

sashass1315:
The p2p Body subcommand now reports "Invalid number of bodies received" instead of "headers" when validating the response from get_block_bodies. This aligns the error message with the actual type being fetched and makes debugging failed body downloads less confusing.


mattsse (approved):
fine

PR #20502, opened 2025-12-18
perf: properly share precompile cache + use `moka`

klkvr:
Changes `PrecompileCacheMap` to use `DashMap` internally and switches individual precompile caches to use `moka`

### DIFF
--- a/crates/engine/tree/src/tree/precompile_cache.rs
+++ b/crates/engine/tree/src/tree/precompile_cache.rs
@@ -21,9 +21,14 @@ where
     S: Eq + Hash + std::fmt::Debug + Send + Sync + Clone + 'static,
 {
     pub(crate) fn cache_for_address(&self, address: Address) -> PrecompileCache<S> {
+        // Try just using `.get` first to avoid acquiring a write lock.
         if let Some(cache) = self.0.get(&address) {
             return cache.clone();
         }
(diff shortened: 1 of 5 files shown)


crates/engine/tree/src/tree/precompile_cache.rs
```diff
@@ -1,114 +1,83 @@
...
 
 impl<S> PrecompileCacheMap<S>
 where
     S: Eq + Hash + std::fmt::Debug + Send + Sync + Clone + 'static,
 {
-    pub(crate) fn cache_for_address(&mut self, address: Address) -> PrecompileCache<S> {
+    pub(crate) fn cache_for_address(&self, address: Address) -> PrecompileCache<S> {
+        if let Some(cache) = self.0.get(&address) {
+            return cache.clone();
+        }
```
shekhirin, line 27: why do we need to do this instead of just an entry call?

klkvr: I believe `.entry` would aquire a shard lock

shekhirin: right, they will both acquire, but `read` acquires only a read lock, while entry does an rw-lock

It's fine here because we will have very few misses, only for precompiles addresses and not inputs themselves

can we just add a comment pls?

shekhirin (approved):

Issue #20503, opened 2025-12-18
Make write_block_bodies operate on &Body


mattsse:
### Describe the feature

this requires taking ownership on write:

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/storage-api/src/chain.rs#L31-L31

but we dont need this and can clone here instead:

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/storage-api/src/chain.rs#L114-L124

and ideally we can have these operate on & as well:

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/db-api/src/models/blocks.rs#L14-L14

but this can be done separately

## TODO
* change write_block_bodies to &Body

figtracer:
gmgm

mattsse:
same as:

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/provider/src/providers/database/provider.rs#L2903-L2906

@figtracer 

because

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/provider/src/providers/database/provider.rs#L2919-L2919

figtracer:
this can be closed

PR #20504, opened 2025-12-18
feat(storage): make insert_block() operate with references

figtracer:
solves #20503

- adds `body()` to BlockResponse
- passes ref downstream through `insert_block()` instead of owned


crates/storage/storage-api/src/chain.rs
```diff
@@ -108,19 +108,21 @@ where
         let mut withdrawals_cursor =
             provider.tx_ref().cursor_write::<tables::BlockWithdrawals>()?;
 
-        for (block_number, body) in bodies {
+        for (block_number, body) in &bodies {
```
joshieDo: we are still owning `bodies` in this function. The goal is to have `write_block_bodies` only requiring a reference to `BlockBody`:

We should change the trait and impl:
https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/storage-api/src/chain.rs#L26-L32

figtracer: fixed

crates/storage/storage-api/src/chain.rs
```diff
@@ -28,7 +28,7 @@ pub trait BlockBodyWriter<Provider, Body: BlockBody> {
     fn write_block_bodies(
         &self,
         provider: &Provider,
-        bodies: Vec<(BlockNumber, Option<Body>)>,
+        bodies: &[(BlockNumber, Option<Body>)],
```
mattsse: this isn't quite what we want, because in order to get a slice like this, we still need owned Body,

we want here instead
 
```suggestion
        bodies: &[(BlockNumber, Option<&Body>)],
```

crates/storage/storage-api/src/chain.rs
```diff
@@ -113,14 +113,16 @@ where
...
             }
 
             // Write withdrawals if any
-            if let Some(withdrawals) = body.withdrawals &&
+            if let Some(withdrawals) = body.withdrawals.clone() &&
                 !withdrawals.is_empty()
             {
-                withdrawals_cursor.append(block_number, &StoredBlockWithdrawals { withdrawals })?;
+                withdrawals_cursor
+                    .append(*block_number, &StoredBlockWithdrawals { withdrawals })?;
```
mattsse: I think these clones are fine for now,

we can move further up the call graph and then change this ideally to ref as well

https://github.com/paradigmxyz/reth/blob/4dfaf238c931faa6388298840c606d8b15fe49a7/crates/storage/provider/src/providers/database/provider.rs#L334-L334

mattsse (changes requested):

crates/storage/storage-api/src/chain.rs
figtracer: do we keep the slice or revert to vec and only reference Body? ig since we don't need the vec after, i'll revert to vec but ref Body instead

crates/storage/storage-api/src/chain.rs
figtracer: on it

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3454,15 +3452,15 @@ mod tests {
...
+        provider_rw.insert_block(&data.genesis.clone().try_recover().unwrap()).unwrap();
         provider_rw
             .write_state(
                 &ExecutionOutcome { first_block: 0, receipts: vec![vec![]], ..Default::default() },
                 crate::OriginalValuesKnown::No,
             )
             .unwrap();
         for i in 0..3 {
-            provider_rw.insert_block(data.blocks[i].0.clone()).unwrap();
+            provider_rw.insert_block(&data.blocks[i].0.clone()).unwrap();
```
joshieDo: no need to clone i think

crates/storage/storage-api/src/block_writer.rs
```diff
@@ -53,7 +53,7 @@ pub trait BlockWriter: Send + Sync {
     /// and transition in the block.
     fn insert_block(
         &self,
-        block: RecoveredBlock<Self::Block>,
+        block: &RecoveredBlock<Self::Block>,
```
joshieDo, line 56: nice

crates/storage/provider/src/providers/database/provider.rs
joshieDo: and in all the other usages

figtracer: yep, missed it

crates/storage/storage-api/src/chain.rs
```diff
@@ -113,11 +113,12 @@ where
...
             // Write ommers if any
             if !body.ommers.is_empty() {
-                ommers_cursor.append(block_number, &StoredBlockOmmers { ommers: body.ommers })?;
+                ommers_cursor
+                    .append(block_number, &StoredBlockOmmers { ommers: body.ommers.clone() })?;
             }
 
             // Write withdrawals if any
-            if let Some(withdrawals) = body.withdrawals &&
+            if let Some(withdrawals) = body.withdrawals.clone() &&
```
figtracer, line 121: should get rid of these `clone()` but maybe in another PR if we open one

crates/storage/provider/src/providers/database/provider.rs
joshieDo: `make fix-lint` should be able to fix most of them

figtracer: > `make fix-lint` should be able to fix most of them

i usually don't run it because it takes too much time in my machine

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -356,7 +356,7 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
             let trie_data = block.trie_data();
             let ExecutedBlock { recovered_block, execution_output, .. } = block;
             let block_number = recovered_block.number();
-            self.insert_block(Arc::unwrap_or_clone(recovered_block))?;
+            self.insert_block(&recovered_block)?;
```
mattsse, line 359: nice!

crates/storage/storage-api/src/chain.rs
```diff
@@ -113,11 +113,12 @@ where
 
             // Write ommers if any
             if !body.ommers.is_empty() {
-                ommers_cursor.append(block_number, &StoredBlockOmmers { ommers: body.ommers })?;
+                ommers_cursor
+                    .append(block_number, &StoredBlockOmmers { ommers: body.ommers.clone() })?;
```
mattsse, line 117: this clone is basically free

crates/storage/storage-api/src/chain.rs
mattsse, line 121: and this clone is negligbile but we can get rid of this in theory

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -249,7 +249,7 @@ fn run_case(
 
         // Insert the block into the database
         provider
-            .insert_block(block.clone())
+            .insert_block(&block.clone())
```
mattsse:
```suggestion
            .insert_block(&block))
```

mattsse (approved):
nice

lgtm, pending @joshieDo @klkvr

testing/ef-tests/src/cases/blockchain_test.rs
```diff
@@ -249,7 +249,7 @@ fn run_case(
 
         // Insert the block into the database
         provider
-            .insert_block(block.clone())
+            .insert_block(&block))
```
figtracer:
```suggestion
            .insert_block(&block)
```

joshieDo (approved):

PR #20508, opened 2025-12-19
refactor(provider): extract heal_segment for NippyJar consistency

### LINKED ISSUE #20506: Refactor StaticFileProvider::check_consistency into separate file and storage checks
**Description**

As noted in [PR #20478 review](https://github.com/paradigmxyz/reth/pull/20478#discussion_r2631483537), the current `StaticFileProvider::check_consistency` method performs two types of consistency checks that need to be separated for correct ordering with RocksDB checks.

**Current Behavior**

`check_consistency` does 2 types of checks:
1. **File consistency** - When a static file fails to commit but the underlying data was changed (internal to static files)
2. **Storage consistency** - Comparing static files vs MDBX checkpoints

**Problem**

RocksDB consistency checks need static file data to prune transaction hashes. The correct order should be:

1. `static_file_provider.check_file_consistency()` - internal static file checks (run first)
2. `rocksdb_provider.check_consistency()` - vs MDBX checkpoints, uses SF data for tx hashes
3. `static_file_provider.check_storage_consistency()` - vs MDBX checkpoints (run last)

**Required Changes**

Refactor `StaticFileProvider::check_consistency` into two separate methods:
- `check_file_consistency()` - handles internal file integrity
- `check_storage_consistency()` - handles comparison with MDBX checkpoints

Reference: https://github.com/paradigmxyz/reth/blob/main/crates/storage/provider/src/providers/static_file/manager.rs#L995-L1105

**Blocked**

- PR #20478 (RocksDB consistency check) depends on this refactoring for correct ordering

yongkangc:
Extract file-level healing logic from `check_consistency()` into a dedicated `heal_segment()` method.

PR #20478 can add a public wrapper to use this:

```rust
pub fn check_file_consistency(&self) -> ProviderResult<()> {
    for segment in StaticFileSegment::iter() {
        self.heal_segment(segment)?;
    }
    Ok(())
}
```

Closes #20506

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -1054,7 +1054,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                 .collect();
 
         // Heal any file-level (NippyJar) inconsistencies
-        self.check_file_consistency()?;
+        self.check_file_consistency(provider)?;
 
         let mut unwind_target: Option<BlockNumber> = None;
         let mut update_unwind_target = |new_target: BlockNumber| {
@@ -1096,6 +1096,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
 
             // Get the pre-healing highest block (captured before check_file_consistency)
             let initial_highest_block = initial_highest_blocks.get(&segment).copied().flatten();
+            debug!(target: "reth::providers::static_file", ?segment, ?initial_highest_block, "Initial highest block for segment");
 
             // Get the post-healing highest block
             // The highest_block may have decreased if we healed from a pruning interruption.
@@ -1214,10 +1215,13 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         Ok(())
     }
 
-    /// Heals file-level (`NippyJar`) inconsistencies for all static file segments.
+    /// Heals file-level (`NippyJar`) inconsistencies for static file segments.
     ///
     /// This method ONLY handles recovery from interrupted file operations (partial writes,
     /// corrupted jars). It does NOT compare against database checkpoints or prune data.
+    ///
+    /// Segments configured to stay in the database (e.g., receipts or transaction senders on
+    /// certain node configurations) are skipped to avoid creating unwanted static files.
     ///
     /// Call this before any operations that need static file data to be in a consistent state,
     /// such as `RocksDB` consistency checks that need transaction data from static files.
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -74,6 +74,53 @@ impl StaticFileAccess {
...
 }
 
+/// Result of file-level consistency checks for static files.
+///
+/// This is an opaque token returned by [`StaticFileProvider::check_file_consistency`]
+/// that should be passed to [`StaticFileProvider::check_storage_consistency`].
+///
+/// Contains the healed state for each segment after file-level recovery,
+/// and any unwind target detected during file healing.
+#[derive(Debug, Default)]
```
fgimenez: is the `Default` derive required? let's remove if not, it can be a footgun to allow external consumers to create an empty `FileConsistencyResult` (`segment_state` would be empty and every segment would be skipped, no storage consistency would be run)

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1084,31 +1084,13 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
-            //   we are expected to still have. We need to check the Database and unwind everything
-            //   accordingly.
-            if self.access.is_read_only() {
-                debug!(target: "reth::providers::static_file", ?segment, "Checking segment consistency (read-only)");
-                self.check_segment_consistency(segment)?;
-            } else {
-                debug!(target: "reth::providers::static_file", ?segment, "Fetching latest writer which might heal any potential inconsistency");
-                // Fetching the writer will attempt to heal any file level inconsistency.
-                self.latest_writer(segment)?;
-            }
```
yongkangc, line 1105: this is extracted to check heal_segment

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1223,6 +1205,44 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+    /// if healing from a pruning interruption decreased the highest block.
+    ///
+    /// File consistency is broken if:
+    ///
+    /// * appending data was interrupted before a config commit, then data file will be truncated
+    ///   according to the config.
+    ///
+    /// * pruning data was interrupted before a config commit, then we have deleted data that we are
+    ///   expected to still have. We need to check the Database and unwind everything accordingly.
+    fn heal_segment(
```
joshieDo:
```suggestion
    fn maybe_heal_segment(
```

can we just add a note that we will return error if we are at read-only and a consistency issue was found

joshieDo (approved):

PR #20509, opened 2025-12-19
docs: fix incorrect documentation terminology

Lorsmirq0cB:
Corrects 'compressed' to 'decompressed' in p2pstream comment where snap::raw::decompress_len returns the decompressed size, fixes `trace.rs` default value documentation from TRACE to debug to match actual code.


mattsse (commented):
not needed

PR #20510, opened 2025-12-19
fix(flashblocks): Add flashblock ws connection retry period

sieniven:
## Summary

This PR adds websocket retry period to prevent unbounded retry loop on websocket connection failures on the flashblock service.

### DIFF
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -22,7 +22,7 @@ use tokio::{
 };
 use tracing::*;
 
-const CONNECTION_RETRY_PERIOD: Duration = Duration::from_secs(5);
+const CONNECTION_BACKOUT_PERIOD: Duration = Duration::from_secs(5);
 
 /// The `FlashBlockService` maintains an in-memory [`PendingFlashBlock`] built out of a sequence of
 /// [`FlashBlock`]s.
@@ -178,10 +178,10 @@ where
                             warn!(
                                 target: "flashblocks",
                                 %err,
-                                retry_period = CONNECTION_RETRY_PERIOD.as_secs(),
+                                retry_period = CONNECTION_BACKOUT_PERIOD.as_secs(),
                                 "Error receiving flashblock"
                             );
-                            sleep(CONNECTION_RETRY_PERIOD).await;
+                            sleep(CONNECTION_BACKOUT_PERIOD).await;
                         }
                         None => {
                             warn!(target: "flashblocks", "Flashblock stream ended");
--- a/crates/optimism/flashblocks/src/service.rs
+++ b/crates/optimism/flashblocks/src/service.rs
@@ -22,7 +22,7 @@ use tokio::{
 };
 use tracing::*;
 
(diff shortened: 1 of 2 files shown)


sieniven:
@0x00101010 would appreciate a review for this too 👍

crates/optimism/flashblocks/src/service.rs
```diff
@@ -12,10 +12,18 @@ use reth_primitives_traits::{AlloyBlockHeader, BlockTy, HeaderTy, NodePrimitives
...
+    sync::Arc,
+    time::{Duration, Instant},
+};
+use tokio::{
+    sync::{oneshot, watch},
+    time::sleep,
+};
 use tracing::*;
 
+const CONNECTION_RETRY_PERIOD: Duration = Duration::from_secs(5);
```
mattsse: can we name backoff perhaps?

crates/optimism/flashblocks/src/service.rs
```diff
@@ -167,7 +175,13 @@ where
...
                         }
                         Some(Err(err)) => {
-                            warn!(target: "flashblocks", %err, "Error receiving flashblock");
+                            warn!(
+                                target: "flashblocks",
+                                %err,
+                                retry_period = CONNECTION_RETRY_PERIOD.as_secs(),
+                                "Error receiving flashblock"
+                            );
+                            sleep(CONNECTION_RETRY_PERIOD).await;
```
mattsse: can you elaborate on why not immediately retrying is better here?

mattsse (changes requested):
naming suggestion and one q re motivation

crates/optimism/flashblocks/src/service.rs
sieniven: Resolved in a9857eea97

crates/optimism/flashblocks/src/service.rs
sieniven: Screenshot 2025-12-23 at 10 59 40 AM
On the case where the websocket connection breaks, the current flashblocks streamer immediately retries to reconnect. However on cases where perhaps the flashblock publisher is down, this spams the RPC node on retry attempts, as the socket creation and connection attempt immediately fails and the error is propagated and handled immediately on re-poll.

The backoff logic should give sufficient time to allow another retry without wasting unnecessary cpu cycles

crates/optimism/flashblocks/src/service.rs
```diff
@@ -12,10 +12,18 @@ use reth_primitives_traits::{AlloyBlockHeader, BlockTy, HeaderTy, NodePrimitives
...
+    sync::Arc,
+    time::{Duration, Instant},
+};
+use tokio::{
+    sync::{oneshot, watch},
+    time::sleep,
+};
 use tracing::*;
 
+const CONNECTION_BACKOUT_PERIOD: Duration = Duration::from_secs(5);
```
mattsse, line 25: lets rename this to CONNECTION_RETRY_BACKOFF

mattsse (approved):

PR #20512, opened 2025-12-19
fix(tree): correct block buffer eviction policy comment

hazeezs:
The comment said eviction is done by "last recently used" (LRU), but the code actually uses FIFO - it just pops from the front of a VecDeque without any access tracking. Fixed the comment to match the actual behavior.


mattsse (approved):
this is fine because actually accurate now

PR #20513, opened 2025-12-19
chore(tree): fix confuse comment on is_fork fn

duyquang6:
comment is not reflect actual code, just update small doc for match code intent


mattsse (approved):
ty, but this is about to be removed #20441

PR #20514, opened 2025-12-19
docs: correct load_toml_config comment

ANtutov:
The load_toml_config helper is synchronous and no longer resolves trusted peers. Its doc comment still claimed it was async, which is misleading for readers and for anyone searching for where trusted peers are handled.

Updated the comment to describe the actual side effects (loading the config, persisting pruning settings and merging static file options) and drop the stale async note.


mattsse (commented):
can keep as is

PR #20516, opened 2025-12-19
chore(db): Remove Sync from DbTx

mediocregopher:
There are two motivations for this change.

1) MDBX does not support using the same tx from different threads. Even with this change it is still technically possible to create a cursor with a tx in one thread, send the tx to another thread, create a new cursor there, and subsequently use both cursors from each thread in an unsafe way. In a followup PR I want to add a lifetime tying the cursors to the tx which should prevent this.

2) By removing Sync from DbTx we can add Cells to it, which can be used for a generic cursor freelist allowing transparent cursor reuse.

By far the biggest effect on the overall codebase from this change was on the StateProviders, which really wanted to be Sync. Specifically the InMemoryOverlay state provider needed some reorganization to make the compiler happy. Overall there should be no actual behavioral code changes.


crates/storage/storage-api/src/macros.rs
```diff
@@ -19,45 +20,46 @@ macro_rules! delegate_impls_to_as_ref {
...
-pub(crate) use delegate_impls_to_as_ref;
+pub use delegate_impls_to_as_ref;
 
 /// Delegates the provider trait implementations to the `as_ref` function of the type:
 ///
 /// [`AccountReader`](crate::AccountReader)
 /// [`BlockHashReader`](crate::BlockHashReader)
 /// [`StateProvider`](crate::StateProvider)
+#[macro_export]
 macro_rules! delegate_provider_impls {
```
mediocregopher, line 31: I moved this to `reth-storage-api` so it could be accessed by both chain-state crate (for in-mem state provider) and provider crate (where it was previously). The implementation is not changed, only the import paths for the errors.

crates/storage/storage-api/src/state.rs
```diff
@@ -27,10 +27,10 @@ pub trait StateReader: Send + Sync {
 }
 
 /// Type alias of boxed [`StateProvider`].
-pub type StateProviderBox = Box<dyn StateProvider>;
+pub type StateProviderBox = Box<dyn StateProvider + Send + 'static>;
```
mediocregopher, line 30: When we box a state provider it must be Send, but the state provider impls themselves don't have to be so we have a bit more wiggle room in cases where it's not necessary to box

examples/bsc-p2p/src/block_import/service.rs
```diff
@@ -65,7 +65,7 @@ where
 
 impl<Provider, T> ImportService<Provider, T>
 where
-    Provider: BlockNumReader + Clone + 'static,
+    Provider: BlockNumReader + Sync + Clone + 'static,
```
mediocregopher, line 68: afaict nothing uses this type, so I added Sync just for it and left it alone.

crates/storage/storage-api/src/macros.rs
```diff
@@ -19,45 +20,46 @@ macro_rules! delegate_impls_to_as_ref {
...
-                fn state_root_with_updates(&self, state: reth_trie::HashedPostState) -> reth_storage_errors::provider::ProviderResult<(alloy_primitives::B256, reth_trie::updates::TrieUpdates)>;
-                fn state_root_from_nodes_with_updates(&self, input: reth_trie::TrieInput) -> reth_storage_errors::provider::ProviderResult<(alloy_primitives::B256, reth_trie::updates::TrieUpdates)>;
+                fn state_root(&self, state: reth_trie::HashedPostState) -> reth_storage_api::errors::provider::ProviderResult<alloy_primitives::B256>;
+                fn state_root_from_nodes(&self, input: reth_trie::TrieInput) -> reth_storage_api::errors::provider::ProviderResult<alloy_primitives::B256>;
+                fn state_root_with_updates(&self, state: reth_trie::HashedPostState) -> reth_storage_api::errors::provider::ProviderResult<(alloy_primitives::B256, reth_trie::updates::TrieUpdates)>;
+                fn state_root_from_nodes_with_updates(&self, input: reth_trie::TrieInput) -> reth_storage_api::errors::provider::ProviderResult<(alloy_primitives::B256, reth_trie::updates::TrieUpdates)>;
```
shekhirin, line 52: why these changes?

shekhirin (approved):
LGTM, that was actually not painful at all to do, I'm surprised

crates/storage/storage-api/src/macros.rs
mediocregopher: This macro got moved to a different crate, so the import paths of the errors needed to be updated. The actual content of the macro hasn't changed.

mattsse (approved):
I think all of these make sense

PR #20517, opened 2025-12-19
refactor(consensus): simplify verify_receipts return

Vitaliyr888:
Remove unnecessary `?` + `Ok(())` pattern in `verify_receipts` function. Since `compare_receipts_root_and_logs_bloom` already returns `Result<(), ConsensusError>`, we can just return its result directly instead of unwrapping and rewrapping it.


mattsse (approved):
fine

PR #20519, opened 2025-12-19
perf(db): Reuse cursors in the Latest/HistoricalStateProviders

### LINKED ISSUE #20492: Re-use cursors in Latest/HistoricalStateProvider
### Describe the feature

Every time we read a storage slot, we create a new dup cursor only to read it one storage slot. This behaviour comes from the fact that duplicate keys in dupsort tables can't be read with a simple mdbx_get.

In Latest/HistoricalStateProvider we should cache cursors for the history tables and plain state tables which are created on the provider itself, so they can be re-used in subsequent calls to the same provider.

### Additional context

_No response_

mediocregopher:
This PR introduces a `ReusableCursor` type which automatically returns a cursor to a Cell on Drop. By holding a Cell for each table type we are able to reuse cursors in the state providers.

I wanted to make this even more generic and automatic, so that we automatically reuse dropped cursors on every Tx, but doing so is a bigger lift. This should at least get us something useful.

Closes #20492

Update:

This introduces a owned/borrowed enum so that we can propagate borrowed cursors and instantiate owned resuable cursors, which we currently do in here:

https://github.com/paradigmxyz/reth/blob/646fec9b4b36707b48d8271715c69d9c38cfddf8/crates/storage/provider/src/providers/database/provider.rs#L203-L203

we could further add an instance to:

https://github.com/paradigmxyz/reth/blob/646fec9b4b36707b48d8271715c69d9c38cfddf8/crates/storage/provider/src/providers/database/provider.rs#L154-L154

so that we can borrow there as well.

this is currently used in the execution stage:

https://github.com/paradigmxyz/reth/blob/646fec9b4b36707b48d8271715c69d9c38cfddf8/crates/stages/stages/src/stages/execution.rs#L298-L298

which is used for multi block execution so I believe the owned variant would be beneficial here as well

Closes RETH-179


mattsse (approved):
this makes sense,

but I believe all provider trait fn go through

https://github.com/paradigmxyz/reth/blob/baa97b38686a6d63d3d16b2620bbfcb73075420d/crates/storage/storage-api/src/macros.rs#L1-L21

which rn would create a new instance of the reusablecursors per as_ref call
so this only gives a benefit if a function of

https://github.com/paradigmxyz/reth/blob/baa97b38686a6d63d3d16b2620bbfcb73075420d/crates/storage/storage-api/src/macros.rs#L27-L31

would create two cursors for the same table

mattsse (approved):

crates/storage/provider/src/providers/state/cursor_reuse.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// A Cow-like type for cursor caches that can either own or borrow the cache.
+///
+/// This allows state providers to either:
+/// - Own their cursor cache (standalone `*Ref` construction)
+/// - Borrow from an owned wrapper type (via `as_ref()` delegation)
+///
+/// Uses `Box` for the owned variant to keep the enum pointer-sized, optimizing
+/// for the common `Borrowed` case used in macro delegation.
+pub(crate) enum CursorCache<'a, TX: DbTx> {
```
DaniPopes, line 66: this is just Cow no ?

crates/storage/provider/src/providers/state/cursor_reuse.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Container for reusable database cursors.
+///
+/// Holds optional cached cursors for frequently accessed tables. When a cursor is requested,
+/// it returns a cached cursor if available, otherwise creates a new one. The cursor is
+/// automatically returned to the cache when dropped via the `ReusableCursor` wrapper.
+///
+/// This reduces cursor allocation overhead for state providers that perform many sequential
+/// database operations.
+pub(crate) struct ReusableStateCursors<TX: DbTx> {
```
DaniPopes, line 17: whys this plural

crates/storage/provider/src/providers/state/latest.rs
```diff
@@ -157,7 +172,7 @@ impl<Provider: DBProvider + BlockHashReader> StateProvider
         account: Address,
         storage_key: StorageKey,
     ) -> ProviderResult<Option<StorageValue>> {
-        let mut cursor = self.tx().cursor_dup_read::<tables::PlainStorageState>()?;
+        let mut cursor = self.cursors.plain_storage_state(self.tx())?;
```
DaniPopes, line 175: this only works for a single table

PR #20521, opened 2025-12-19
fix(stages): Fix two bugs related to stage checkpoints and pipeline syncs

mediocregopher:
Bug 1: When unwinding more than 128 blocks we would end up completely clearing out the trie changeset tables. While not a bug on its own, when running a reth-bench on a recently unwound node we would frequently hit `insufficient changesets` errors because, given a smaller wait-time between blocks, we would be more likely to run persistence at the same time as a payload validation.

The fix for this bug is to actually clear the MerkleChangesets stage checkpoint if unwinding results in fewer than the retention threshold of blocks being present. This will induce a pipeline sync on the next startup which will regenerate trie changesets.

Bug 2: A recent change caused the logic which determines if a pipeline sync is required to be broken. This fixes it.

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -944,7 +944,7 @@ where
             StageId::ALL.into_iter().filter(|id| era_enabled || id != &StageId::Era);
 
         // Get the expected first stage based on config.
-        let first_stage = all_stages.next().expect("there must be at least once stage");
+        let first_stage = all_stages.next().expect("there must be at least one stage");
 
         // If no target was provided, check if the stages are congruent - check if the
         // checkpoint of the last stage matches the checkpoint of the first.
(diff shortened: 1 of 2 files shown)


crates/node/builder/src/launch/common.rs
```diff
@@ -950,20 +954,28 @@ where
             .unwrap_or_default()
             .block_number;
 
-        // Skip the first stage as we've already retrieved it and comparing all other checkpoints
-        // against it.
-        for stage_id in StageId::ALL.iter().skip(1) {
```
mediocregopher, line 955: The issue here was that we would `skip(1)` when iterating over all stages, but we've already skipped the first stage above when getting `first_stage`. So this would always end up comparing Headers stage to itself which would.... fail? I couldn't figure out why, even the logs don't really make sense:

```
2025-12-19T14:31:53.194224Z DEBUG consensus::engine: Checking stage against first stage first_stage_checkpoint=829728 stage_id=Headers stage_checkpoint=828598
2025-12-19T14:31:53.194263Z DEBUG consensus::engine: Pipeline sync progress is inconsistent first_stage_checkpoint=829728 inconsistent_stage_id=Headers inconsistent_stage_checkpoint=828598
```
Note that it's saying the stages are inconsistent even though the checkpoints are the same....

Anyway, I fixed it to not be comparing against itself and now it works as expected.

klkvr (approved):

PR #20523, opened 2025-12-19
chore: prevent false-positive log when peer not found in transaction propagation

vtjl10:
The trace log was emitted before checking if the peer exists, causing false-positive log entries when `propagate_full_transactions_to_peer` returns early due to a missing peer. This misleads debugging efforts and makes it appear that transactions were propagated when they were actually skipped.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -860,7 +860,6 @@ where
         peer_id: PeerId,
         propagation_mode: PropagationMode,
     ) -> Option<PropagatedTransactions> {
-        // Only log after we know the peer exists; otherwise we would emit a false-positive.
         let peer = self.peers.get_mut(&peer_id)?;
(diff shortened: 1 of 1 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -860,9 +860,9 @@ where
         peer_id: PeerId,
         propagation_mode: PropagationMode,
     ) -> Option<PropagatedTransactions> {
-        trace!(target: "net::tx", ?peer_id, "Propagating transactions to peer");
-
+        // Only log after we know the peer exists; otherwise we would emit a false-positive.
```
mattsse: you guys need to stop adding comments that are totally redundant and only relevant in the context of this change

please remove

mattsse (changes requested):

mattsse (approved):

PR #20525, opened 2025-12-19
chore(net/dns): remove redundant PartialEq bound on LinkEntry Eq impl

0xxFloki:
Removes redundant `PartialEq` trait bound from `LinkEntry<K>`'s `Eq` implementation.


mattsse (commented):
we can keep this as

PR #20526, opened 2025-12-19
perf(storage): eliminate redundant range clones in get_state method

0xlupin:
Removed three unnecessary RangeInclusive clones in the ConsistentProvider::get_state method by reusing already extracted boundary values.


mattsse (commented):
this clone is pretty much fake because it will be the same as copying two u64 so we can keep this as is

PR #20528, opened 2025-12-20
fix: remove stale blobs

mattsse:
if a tx becomes stale we remove it, but we should also remove stale blobs here

drive by for faster instant cmp:

```
 pub fn elapsed(&self) -> Duration {
        Instant::now() - *self
    }
```


yongkangc (approved):
Makes sense to remove stale blobs

PR #20529, opened 2025-12-20
fix: remove code as mention in todo

cuiweixie:


mattsse (commented):
I think we want to keep this for a bit longer

mattsse (commented):
desnt really hurt to keep this

PR #20536, opened 2025-12-20
fix: remove outdated TODO comment about rand migration

vtjl10:
The comment `// TODO(rand): update ::random calls after rand_09 migration` is no longer relevant as the codebase already uses rand 0.9 APIs (`rng.random()`, `Address::random()`, `B256::random()`). This misleading TODO suggests unfinished work when the migration is complete.


mattsse (commented):
this is still todo

PR #20538, opened 2025-12-20
chore(flashblocks): remove dead cached_reads wipe after finalize

MozirDmitriy:
FlashBlockPendingSequence::finalize already clears cached_reads, and this behavior is enforced by tests. The extra 
take_cached_reads call in SequenceManager::insert_flashblock always operated on None and had no effect, while its comment suggested that it was responsible for wiping the cache. Removing this no-op call and misleading comment makes the cache lifecycle clearer and keeps the implementation aligned with the tested invariants.


mattsse (commented):
I want to keep this just to make sure we properly clean up

PR #20539, opened 2025-12-20
fix(consensus): correct got/expected order in BlobGasUsedDiff

hawkadrian:
## What

Fix inconsistent field order in `BlobGasUsedDiff` error. Changed `got` to contain computed value from block body and `expected` to contain value from header.

## Why

Maintains consistency with other body vs header validation checks (ommers hash, transactions root, withdrawals root) where `got` is the computed value and `expected` is the header value. This improves error message clarity and follows the established pattern across the codebase.


mattsse (commented):
we can keep this as is

PR #20542, opened 2025-12-21
refactor(storage): extract shared find_changeset_block_from_index algorithm [1/3]

### LINKED ISSUE #20388: Use EitherReader/EitherWriter in DatabaseProvider and HistoricalStateProvider

yongkangc:
**Summary**

Extract the history lookup algorithm into a reusable function that can be shared between MDBX and RocksDB backends. 

**This is PR 1/3** of a stacked PR series that splits [#20412](https://github.com/paradigmxyz/reth/pull/20412) for easier review. Closes https://github.com/paradigmxyz/reth/issues/20388

**Review focus**

- Verify `find_changeset_block_from_index` correctly captures the original algorithm

**Stack**
  - PR #20542: refactor: extract shared find_changeset_block_from_index algorithm
  - PR #20543: feat: add RocksDB history lookup methods and owned batch type
  - PR #20544: feat: wire RocksDB into history lookups and HistoryWriter


crates/storage/provider/src/changesets_utils/trie.rs
```diff
@@ -43,20 +43,20 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         let Some(curr_path) = self.paths.next() else {
             // If there are no more paths then there is no further possible output.
-            return None
+            return None;
```
joshieDo: nooo `;`

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -166,48 +176,19 @@ impl<'b, Provider: DBProvider + BlockNumReader> HistoricalStateProviderRef<'b, P
...
-                // The chunk contains an entry for a write after our block, return it.
-                Ok(HistoryInfo::InChangeset(block_number))
-            } else {
-                // The chunk does not contain an entry for a write after our block. This can only
-                // happen if this is the last chunk and so we need to look in the plain state.
-                Ok(HistoryInfo::InPlainState)
-            }
+        if let Some(chunk) = cursor.seek(key)?.filter(|(key, _)| key_filter(key)).map(|x| x.1) {
+            // Check if there's a previous shard for the same key
+            let has_previous_shard = cursor.prev()?.is_some_and(|(key, _)| key_filter(&key));
```
joshieDo: >// This check is worth it, the `cursor.prev()` check is rarely triggered (the if will
            // short-circuit) and when it passes we save a full seek into the changeset/plain state
            // table.

before, this line was rarely being triggered, but now it will always. 

maybe we should make this a closure instead and pass that if we don't want to be passing the cursor?

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -243,6 +224,66 @@ impl<Provider: DBProvider + BlockNumReader> HistoricalStateProviderRef<'_, Provi
...
+/// * `chunk` - The block number list from the history index shard
+/// * `block_number` - Target block to look up
+/// * `has_previous_shard` - Whether there's a shard before this one for the same key
+/// * `lowest_available` - Lowest block where history is available (pruning boundary)
+///
+/// # Returns
+/// - `InChangeset(n)` - Look up the value in the changeset at block `n`
+/// - `InPlainState` - Look up the value in plain state (no changes after target block)
+/// - `NotYetWritten` - The key didn't exist at the target block
+/// - `MaybeInPlainState` - Pruning may have removed history, check plain state
```
joshieDo:
```suggestion
```

crates/storage/provider/src/providers/state/historical.rs
yongkangc: ahh before only when the target block is before the first write in the shard that it triggers with `&&`


good point, there might be a more elegant way to refactor this

> before, this line was rarely being triggered, but now it will always.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -166,48 +176,36 @@ impl<'b, Provider: DBProvider + BlockNumReader> HistoricalStateProviderRef<'b, P
...
-                }
-            } else if let Some(block_number) = block_number {
-                // The chunk contains an entry for a write after our block, return it.
-                Ok(HistoryInfo::InChangeset(block_number))
-            } else {
-                // The chunk does not contain an entry for a write after our block. This can only
-                // happen if this is the last chunk and so we need to look in the plain state.
-                Ok(HistoryInfo::InPlainState)
-            }
+            let is_before_first_write =
```
yongkangc, line 238: @joshieDo instead of going with closure, i extracted to a pure method - thought it was simpler

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -166,48 +176,36 @@ impl<'b, Provider: DBProvider + BlockNumReader> HistoricalStateProviderRef<'b, P
...
-                    Ok(HistoryInfo::NotYetWritten)
-                }
-            } else if let Some(block_number) = block_number {
-                // The chunk contains an entry for a write after our block, return it.
-                Ok(HistoryInfo::InChangeset(block_number))
-            } else {
-                // The chunk does not contain an entry for a write after our block. This can only
-                // happen if this is the last chunk and so we need to look in the plain state.
-                Ok(HistoryInfo::InPlainState)
-            }
```
yongkangc, line 210: this becomes part of `find_changeset_block_from_index`

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -166,48 +176,36 @@ impl<'b, Provider: DBProvider + BlockNumReader> HistoricalStateProviderRef<'b, P
...
+            let found_block = chunk.select(rank);
 
             // If our block is before the first entry in the index chunk and this first entry
             // doesn't equal to our block, it might be before the first write ever. To check, we
             // look at the previous entry and check if the key is the same.
             // This check is worth it, the `cursor.prev()` check is rarely triggered (the if will
             // short-circuit) and when it passes we save a full seek into the changeset/plain state
             // table.
-            if rank == 0 &&
-                block_number != Some(self.block_number) &&
```
yongkangc, line 191: these 2 logic goes into `needs_prev_shard_check` imo its easier to read and understand 

the part i didnt like about the previous code was that the code has too many conditionals and was hard to hold in my head

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -524,6 +522,49 @@ impl LowestAvailableBlocks {
...
+///
+/// This is a pure function shared by both MDBX and `RocksDB` backends.
+///
+/// # Arguments
+/// * `found_block` - The block number from the shard lookup
+/// * `is_before_first_write` - True if the target block is before the first write to this key. This
+///   should be computed as: `rank == 0 && found_block != Some(block_number) && !has_previous_shard`
+///   where `has_previous_shard` comes from a lazy `cursor.prev()` check.
+/// * `lowest_available` - Lowest block where history is available (pruning boundary)
+pub const fn find_changeset_block_from_index(
```
joshieDo: maybe as `impl HistoryInfo`?

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -32,6 +32,24 @@ use reth_trie_db::{
 
 use std::fmt::Debug;
 
+/// Result of a history lookup for an account or storage slot.
+///
+/// Indicates where to find the historical value for a given key at a specific block.
+#[derive(Debug, Eq, PartialEq)]
+pub enum HistoryInfo {
+    /// The key is written to, but only after our block (not yet written at the target block).
```
joshieDo:
```suggestion
    /// The key is written to, but only after our block (not yet written at the target block). Or it has never been written.
```

i think

crates/storage/provider/src/providers/state/historical.rs
joshieDo: `HistoryInfo::from_lookup`

joshieDo (dismissed):

crates/storage/provider/src/providers/state/historical.rs
yongkangc: Ah yes that's much nicer

yongkangc: Ah yes that's much nicer

yongkangc: Ah yes that's much nicer

crates/storage/provider/src/providers/state/historical.rs
yongkangc: Thanks for the note

yongkangc: Thanks for the note

crates/storage/provider/src/providers/state/historical.rs
yongkangc: addressed this @joshieDo

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -524,6 +560,14 @@ impl LowestAvailableBlocks {
...
+///
+/// Returns `true` when `rank == 0` (first entry in shard) and the found block doesn't match
+/// the target block number. In this case, we need to check if there's a previous shard.
+fn needs_prev_shard_check(rank: u64, found_block: Option<u64>, block_number: BlockNumber) -> bool {
+    rank == 0 && found_block != Some(block_number)
+}
+
 #[cfg(test)]
 mod tests {
     use crate::{
```
fgimenez, line 574: would be nice to document the expected behavior with a test wdyt? like:
```rust
    #[test]
    fn test_history_info_from_lookup() {
        // Before first write, no pruning
        assert_eq!(HistoryInfo::from_lookup(Some(10), true, None), HistoryInfo::NotYetWritten);

        // Before first write, no found block, no pruning
        assert_eq!(HistoryInfo::from_lookup(None, true, None), HistoryInfo::NotYetWritten);

        // Not before first write, found block
        assert_eq!(HistoryInfo::from_lookup(Some(10), false, None), HistoryInfo::InChangeset(10));

..................
    }
```

fgimenez (approved):

crates/storage/provider/src/providers/state/historical.rs
yongkangc: That's a good note! Thanks 🙏

yongkangc: That's a good note! Thanks 🙏

yongkangc:
@joshieDo would like a review from u before any merge :)

crates/storage/provider/src/providers/state/historical.rs
yongkangc: added https://github.com/paradigmxyz/reth/pull/20542/changes/a0808f457b1e9f1de24cb699c52220a73e625c83

Rjected (approved):
lgtm, I like that we can actually test `from_lookup` and `needs_prev_shard_check`

PR #20543, opened 2025-12-21
feat(storage): add RocksDB history lookup methods and owned batch type [2/3]

### LINKED ISSUE #20388: Use EitherReader/EitherWriter in DatabaseProvider and HistoricalStateProvider

yongkangc:
Add RocksDB infrastructure for history lookups without changing any behavior.

This is PR 2/3 of a stacked PR series that splits [#20412](https://github.com/paradigmxyz/reth/pull/20412) for easier review. Closes https://github.com/paradigmxyz/reth/issues/20388

Depends on: #20542

**Review focus:** 
- Correctness of `RocksTx` history lookup methods
- `RocksDBBatch` ownership change is safe

**Changes**

`rocksdb/provider.rs`
- Change `RocksDBBatch` to own `Arc<RocksDBProviderInner>` instead of borrowing
  - Allows storing `RocksDBBatch` in `DatabaseProvider` without lifetime issues
- Add `RocksTx::account_history_info()` method

**Stack**
  - PR #20542: 
  - PR #20543: 
  - PR #20544: 

Did some testing and benchmark here: https://gist.github.com/yongkangc/6dfedca18b842cd41ea2864f248c1a07
MDBX wins by 2x but we are still using Rocksdb for the optmised writes.

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -1,7 +1,10 @@
 use super::metrics::{RocksDBMetrics, RocksDBOperation};
+use crate::providers::{needs_prev_shard_check, HistoryInfo};
+use alloy_primitives::{Address, BlockNumber, B256};
 use reth_db_api::{
-    table::{Compress, Decompress, Encode, Table},
-    tables, DatabaseError,
+    models::{storage_sharded_key::StorageShardedKey, ShardedKey},
+    table::{Compress, Decode, Decompress, Encode, Table},
+    tables, BlockNumberList, DatabaseError,
 };
 use reth_storage_errors::{
     db::{DatabaseErrorInfo, DatabaseWriteError, DatabaseWriteOperation, LogLevel},
@@ -9,8 +12,8 @@ use reth_storage_errors::{
 };
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamilyDescriptor, CompactionPri, DBCompressionType,
-    IteratorMode, Options, Transaction, TransactionDB, TransactionDBOptions, TransactionOptions,
-    WriteBatchWithTransaction, WriteOptions,
+    DBRawIteratorWithThreadMode, IteratorMode, Options, Transaction, TransactionDB,
+    TransactionDBOptions, TransactionOptions, WriteBatchWithTransaction, WriteOptions,
 };
 use std::{
     fmt,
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -471,17 +480,22 @@ impl RocksDBProvider {
...
+/// Note: `WriteBatch` operations are applied in order. If the same key is updated multiple times,
+/// the last update wins. Ref: <https://github.com/facebook/rocksdb/wiki/Basic-Operations#atomic-updates>
+///
+/// This type owns an `Arc<RocksDBProviderInner>` to allow storing it in `DatabaseProvider`
+/// without lifetime issues.
 #[must_use = "batch must be committed"]
-pub struct RocksDBBatch<'a> {
-    provider: &'a RocksDBProvider,
+pub struct RocksDBBatch {
+    provider: Arc<RocksDBProviderInner>,
```
yongkangc: If `RocksDBBatch<'a>` borrows `&'a RocksDBProvider`, then `DatabaseProvider` would need to:                                                                                                             
  1. Store both the `RocksDBProvider` and the `RocksDBBatch<'a>` that borrows from it                      
  2. This creates a self-referential struct

joshieDo: isnt this just `RocksDBProvider`?  RocksDBProvider is already an arc over an inner.

I think we might have to revisit this struct in another PR, because it feels odd

Above we specify
> Uses `WriteBatchWithTransaction` for atomic writes without full transaction overhead.

but in fact `RocksDbProviderInner` holds a `Transaction`.

And if we're implementing read, maybe the best option would be to use the optimistic TX directly? unsure if we need to read through our writes here, but seems like a footgun if we do now or in the future

yongkangc: > but in fact RocksDbProviderInner holds a Transaction.

RocksDB naming is confusing - `RocksDbProviderInner` hold `TransactionDB` which is the database engine (supports transactions), not an active transaction.

Would update the comment so that this is clearer!

yongkangc: >And if we're implementing read, maybe the best option would be to use the optimistic TX directly? unsure if we need to read through our writes here, but seems like a footgun if we do now or in the future 

hmm good point @joshieDo , looking at mdbx pattern it seems like mdbx does have read our writes here.

However do we need read your writes for history tables? my impression was that it was append‑only write sink for history tables and its more performant if we do that in batches

joshieDo: > TransactionDB which is the database engine (supports transactions), not an active transaction.

oh i definitely got confused by the naming, mb

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -467,23 +467,27 @@ impl RocksDBProvider {
...
+/// Note: `WriteBatch` operations are applied in order. If the same key is updated multiple times,
+/// the last update wins. Ref: <https://github.com/facebook/rocksdb/wiki/Basic-Operations#atomic-updates>
+///
+/// This type owns a `RocksDBProvider` to allow storing it in `DatabaseProvider`
+/// without lifetime issues.
 #[must_use = "batch must be committed"]
-pub struct RocksDBBatch<'a> {
-    provider: &'a RocksDBProvider,
+pub struct RocksDBBatch {
+    provider: RocksDBProvider,
```
yongkangc: the provider: `RocksDBProvider` field in RocksDBBatch is effectively an Arc clone. This change is motivateed by: 


 - If RocksDBBatch borrows a provider, the batch can’t live longer than the borrow.
  - But we need to store the batch and commit it later, so it must outlive that borrow.
  - If it owned the provider (the Arc), the batch can live as long  as needed.
  - Otherwise we’d have to put lifetimes on DatabaseProvider and a bunch of APIs just to keep the borrow valid.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -666,6 +669,158 @@ impl<'db> RocksTx<'db> {
...
+        let mut rank = chunk.rank(block_number);
+        if rank.checked_sub(1).and_then(|r| chunk.select(r)) == Some(block_number) {
+            rank -= 1;
+        }
+        let found_block = chunk.select(rank);
+
+        // Lazy check for previous shard - only called when needed.
+        // If we can step to a previous shard for this same key, history already exists,
+        // so the target block is not before the first write.
+        let is_before_first_write = if rank == 0 && found_block != Some(block_number) {
```
fgimenez: nit, this check `rank == 0 && found_block != Some(block_number)` is done also in mdbx's `needs_prev_shard_check` here https://github.com/paradigmxyz/reth/blob/yk/pr1-history-info-abstraction/crates/storage/provider/src/providers/state/historical.rs#L567 maybe we couldmove that fn to a place where it could be reused, wdyt?

fgimenez (approved):
lgtm, smol nit

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc: good suggestion! thanks fede :)

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1094,4 +1249,179 @@ mod tests {
         let last = provider.last::<TestTable>().unwrap();
         assert_eq!(last, Some((20, b"value_20".to_vec())));
     }
+
+    #[test]
+    fn test_account_history_info_single_shard() {
+        use crate::providers::HistoryInfo;
```
joshieDo: nit: test mod top level (this and others)

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1094,4 +1249,179 @@ mod tests {
         let last = provider.last::<TestTable>().unwrap();
         assert_eq!(last, Some((20, b"value_20".to_vec())));
     }
+
+    #[test]
+    fn test_account_history_info_single_shard() {
```
joshieDo, line 1268: i wonder if we could move these tests out of rocksdb crate ? this is behaviour we want to test on both mdbx/rocksdb . and if they both pass then we're sure they're equivalent?

joshieDo (approved):

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc: yeah good idea, will create another pr to move this, thanks @joshieDo

PR #20548, opened 2025-12-21
fix(exex): update lowest_committed_block_height in WAL cache on insert

strmfos:
`lowest_committed_block_height` was never set when committing blocks to WAL cache, only `highest_committed_block_height` was updated. This caused the metric to always be `None` until the first WAL finalization.

Now both metrics are updated consistently on insert, matching the existing logic in `remove_before()`.


mattsse (approved):
I see, this makes sense and this is only used for reporting

Issue #20434, opened 2025-12-16
Pre-calculate keccak hashes for transaction addresses at newPayload start


shekhirin:
## Description

Even though we have a keccak cache now, we could benefit from pre-calculating keccak hashes for transaction addresses (senders, receivers) at the beginning of `newPayload` processing. This would ensure the cache is warm before parallel tasks (execution, prewarming, multiproof generation) start competing to compute the same hashes.

## Context

Currently, `multiproof_targets_from_state\ calculates keccaks for addresses during prewarm outcome processing:

https://github.com/paradigmxyz/reth/blob/4231f4b68879d673ebab41b8a7e434537e6c6f0c/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L585

## Solution

At the start of payload execution, iterate through transaction senders and recipients and pre-populate the keccak cache. This front-loads the computation before parallel tasks begin.

PR #20550, opened 2025-12-21
perf(engine): pre-populate keccak cache before parallel tasks begin

i-m-aditya:
Ref #20434


i-m-aditya:
@mediocregopher addressed txs decoding/recovering duplication—thoughts?

mediocregopher (approved):
This LGTM, would like @mattsse or @klkvr to double-check though. Going to benchmark as well to determine if this helps anything.

mattsse (commented):
dont think we need this

PR #20553, opened 2025-12-21
refactor(rpc): remove dead code and unnecessary allocation in sim_bundle

Aleksandr1732:
This PR removes dead code and eliminates an unnecessary allocation in the MEV sim bundle implementation.

### DIFF
--- a/crates/rpc/rpc/src/eth/sim_bundle.rs
+++ b/crates/rpc/rpc/src/eth/sim_bundle.rs
@@ -344,18 +344,20 @@ where
                 let original_refundable_value = refundable_value;
                 for item in &flattened_bundle {
                     if let Some(refund_percent) = item.refund_percent {
-                        // Get refund configurations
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/eth/sim_bundle.rs
```diff
@@ -344,18 +344,20 @@ where
                 let original_refundable_value = refundable_value;
                 for item in &flattened_bundle {
                     if let Some(refund_percent) = item.refund_percent {
-                        // Get refund configurations
-                        let refund_configs = item.refund_configs.clone().unwrap_or_else(|| {
-                            vec![RefundConfig { address: item.tx.signer(), percent: 100 }]
-                        });
+                        // Only need the number of refund configs; default to one without allocating.
```
mattsse, line 347: please stop adding comments like this, these are totally useless without the pr context

mattsse (commented):
closing this because of the non sensical comment,

would still accept that change otherwise

PR #20554, opened 2025-12-21
docs(libmdbx): add missing # Panics documentation for unwrap calls

letmehateu:
## Problem

Several functions in the transaction module use `.unwrap()` on channel receives and type 
conversions without documenting the panic conditions.

## Summary

Adds missing `# Panics` documentation to 4 public functions and 1 Drop implementation 
in `libmdbx-rs/transaction.rs` that call `.unwrap()` but don't document when they panic.


mattsse (commented):
can keep as is

PR #20555, opened 2025-12-21
refactor(engine): derive Copy for DefaultEngineValues

marukai67:
Derive Copy for DefaultEngineValues and replace .clone() with copy-by-dereference in EngineArgs::default().


mattsse (commented):
we can keep clone here in case we need to add a non copy field

PR #20557, opened 2025-12-21
fix(era-downloader): improve retry logic and error handling in download_to_file

forkfury:
### Fix
Fixes retry logic and error handling in download_to_file. 

The previous range-based retry (1..3) was unclear, and errors weren't handled for non-Era1 files. 

Partially downloaded files weren't cleaned up on failure.
### Changes
- Replace range-based retry with explicit attempt counter (MAX_DOWNLOAD_ATTEMPTS)
- Clean up partially downloaded files before retries
- Propagate errors correctly for both Era1 and Era file types
- Remove failed downloads and invalid checksum files automatically

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -372,7 +372,26 @@ where
             )
             .into())
         };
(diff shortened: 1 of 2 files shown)


mattsse (commented):
we can keep this as is, especially because this touches some other code

PR #20558, opened 2025-12-21
refactor(primitives-traits): remove redundant PartialEq bounds from SignedTransaction

0xlupin:
Remove redundant PartialEq trait bounds from SignedTransaction trait and its implementations.


mattsse (commented):
doesnt hurt to keep

PR #20563, opened 2025-12-22
docs: describe EVM payload building flow

lisenokdonbassenok:
Explained how the EVM component participates in payload building when driven by a consensus client over the Engine API. Documented the role of next-block attributes (timestamp, fee recipient, randomness, gas limit, withdrawals, beacon block root) and clarified that they are used to derive the next block environment before executing transactions and assembling a block candidate. This makes the EVM component docs consistent with the ConfigureEvm and NextBlockEnvAttributes APIs.


mattsse (commented):
not needed

PR #20565, opened 2025-12-22
fix(rpc): preserve nested bundle structure in mev_simBundle logs

meetrick:
Fixes log collection in `mev_simBundle` so nested bundles preserve their hierarchical structure.

The implementation now records logs in execution order and rebuilds the `SimBundleLogs` tree iteratively from the original bundle shape. This avoids misattributing logs when the same signed transaction appears multiple times in a bundle and adds regression tests for duplicate transactions and mismatched log reconstruction.


crates/rpc/rpc/src/eth/sim_bundle.rs
```diff
@@ -350,18 +396,18 @@ where
...
                         // Add gas used for payout transactions
                         total_gas_used += SBUNDLE_PAYOUT_MAX_COST * refund_configs.len() as u64;
 
                         // Calculate allocated refundable value (payout value) based on ORIGINAL
                         // refundable value This ensures all refund_percent
                         // values are calculated from the same base
-                        let payout_value = original_refundable_value * U256::from(refund_percent) /
-                            U256::from(100);
+                        let payout_value = original_refundable_value * U256::from(refund_percent)
+                            / U256::from(100);
```
mattsse: please don't change formatting, this makes it harder to review

crates/rpc/rpc/src/eth/sim_bundle.rs
```diff
@@ -83,6 +86,41 @@ impl<Eth> EthSimBundle<Eth>
...
+                BundleItem::Tx { tx, .. } => {
+                    // Compute the transaction hash from raw transaction bytes
+                    // The hash of an RLP-encoded signed transaction is its transaction hash
+                    let tx_hash = keccak256(tx);
+                    let tx_logs = tx_logs_map.get(&tx_hash).cloned();
+                    logs.push(SimBundleLogs { tx_logs, bundle_logs: None });
+                }
+                BundleItem::Bundle { bundle: nested_bundle } => {
+                    // Recursively build logs for nested bundle
+                    let nested_logs = Self::build_bundle_logs(nested_bundle, tx_logs_map);
```
mattsse: this is recursive, I'd like to avoid having recursive calls, even with a max depth of 5

crates/rpc/rpc/src/eth/sim_bundle.rs
```diff
@@ -330,15 +374,17 @@ where
                                 full_log
                             })
                             .collect();
-                        let sim_bundle_logs =
-                            SimBundleLogs { tx_logs: Some(tx_logs), bundle_logs: None };
-                        body_logs.push(sim_bundle_logs);
+                        tx_logs_map.insert(tx_hash, tx_logs);
```
mattsse: maybe a better approach here would be to include the bundle depth in the `FlattenedBundleItem` as an identifier then we can populate the `body_logs` properly

mattsse (changes requested):
thanks, all of this makes sense

but I believe we can also populate the logs if use smth like a stack of logs instead because I'd like to avoid having recursive calls here

crates/rpc/rpc/src/eth/sim_bundle.rs
meetrick: I'll keep the format as is for the next revision.

crates/rpc/rpc/src/eth/sim_bundle.rs
meetrick: got it! thanks

meetrick:
@mattsse Could you review again?

crates/rpc/rpc/src/eth/sim_bundle.rs
```diff
@@ -59,6 +62,9 @@ pub struct FlattenedBundleItem<T> {
     pub refund_percent: Option<u64>,
     /// Optional refund configs for the bundle item
     pub refund_configs: Option<Vec<RefundConfig>>,
+    /// Path from root bundle to this item, represented as indices at each nesting level.
+    /// For example, `[1, 0]` means `bundle.bundle_body[1].bundle.bundle_body[0]`.
+    pub bundle_path: Vec<usize>,
```
mattsse: is this even used?

meetrick: It looks like some code from the previous approach (bundle_path) is still there. This is not used..

meetrick: @mattsse  I removed bundle_path. You commented because it was unused and should be cleaned up, right?

mattsse (approved):

Issue #20491, opened 2025-12-18
Flashblocks: issue when handling chainstate updates on flashblocks consensus sync vs default sync


sieniven:
## Issue

Referencing to the really nice PR added by @0x00101010 https://github.com/paradigmxyz/reth/pull/19786, it seems like on further testing, there are 3 potentially problematic issues that could surface from this design.

1. Currently FCU sent on the flashblocks consensus sets unsafe, safe and finalized as the flashblocks sequence hash. This is incorrect and violates optimism finality states, and is resolved in the WIP PR: https://github.com/paradigmxyz/reth/pull/20490
2. The current flashblocks consensus layer doesnt take into account the possibility that the default CL sync could be ahead of flashblocks. This can be resolved pretty easily by adding a chainstate check on the flashblocks consensus client to check with the chain provider the current latest tip, and skip if the FCU call is behind the curren tip. A WIP PR https://github.com/paradigmxyz/reth/pull/20490 should resolve this.
3. The other issue is flashblocks consensus layer is ahead of the CL sync, then this situation may happen which could be undesirable.   - flashblocks advances chainstate to x -> sends FCU to x
   - CL sync is still at x-1 -> sends FCU to x-1
   - chainstate reorgs to x-1
   - Next flashblock sequence advances chainstate to x+1

I am thinking of what are the possible solutions to resolve this - perhaps allowing flashblocks RPC to advance the EL's chainstate might not be such a good idea afterall? However the performance improvements, if the chainstate is able to sync with flashblocks, is massive. However, resolving issue 2 doesnt seem too trivial as if the resolution is on the EL layer, it might touch core protocol which is dangerous if an actual re-org needs to happen.

## Solution 1

The easiest and least complicated solution to resolve this is for flashblocks consensus layer to not be able to advance chainstate, but pre-warm the payload and insert into the locally built blocks cache. Subsequently on CL sync, engine_newPayload will skip the payload validation logic all together which will help greatly in RPC syncs.

However, this solution could limit the flashblocks RPC potential since driving the chainstate and reducing chain latencies is dependent on the default CL sync.

## Solution 2

Perhaps an alternative solution will be to design a separate accumulated flashblock sequences chainstate that is used in the exposed eth APIs. We maintain this flashblocks state layer separate from the default sync (which updates the canonical chainstate). This flashblocks state layer is flushed once the default sync's block is received, and we similarly we implement solution 1 by default to speed up default RPC sync.

### Steps to reproduce

-

[...]

PR #20566, opened 2025-12-22
feat(flashblocks): Add async state root calculation and pre-warming flashblocks sequence execution to local execution cache

sieniven:
## Summary

This PR references the flashblocks RPC solution 1 mentioned in: https://github.com/paradigmxyz/reth/issues/20491. It optimizes the sync for flashblocks RPC (CL sync and flashblocks sync) by inserting the executed flashblocks sequence to the engine state tree, to prevent re-execution and re-calculating of state root of the same payload on engine_newPayload engine api. This should optimize the RPC sync speed and improve latencies on the RPC node.

1. Flashblocks RPC sync logic is now split into 2 portions. New incoming flashblocks sequence are validated and executed without state root calculation quickly, which updates the pending state on the eth api handler as quickly as possible
2. State root calculation and subsequent sync logic (flashblock or CL driven syncs) are shifted to async
3. If flashblocks RPC node chainstate is driven by the CL sync (`--flashblocks.consensus` turned off), which this PR optimizes the sync speed of the CL sync as unsafe payload commits on the CL will skip payload re-validation on cache hits, which the flashblocks RPC has already executed and validated the sequence as the current pending block
2. With `--flashblocks.consensus` enabled, this optimizes the RPC node sync with the flashblocks builder source, as we cache hit locally executed blocks prevent re-execution of the same payload on the flashblocks RPC node


crates/engine/tree/src/tree/mod.rs
```diff
@@ -1474,6 +1474,7 @@ where
                                 // handle the event if any
                                 self.on_maybe_tree_event(maybe_event)?;
                             }
+                            _ => {}
```
mattsse: can we do this fully exhaustive

crates/engine/primitives/src/message.rs
```diff
@@ -164,6 +165,11 @@ pub enum BeaconEngineMessage<Payload: PayloadTypes> {
         /// The sender for returning forkchoice updated result.
         tx: oneshot::Sender<RethResult<OnForkChoiceUpdated>>,
     },
+    /// Message to insert a locally built executed block into the engine state tree.
+    InsertExecutedBlock {
+        /// The executed block to insert.
+        block: ExecutedBlock<<Payload::BuiltPayload as BuiltPayload>::Primitives>,
+    },
```
mattsse: hmm, a bit unsure about exposing this because this assumes the block is fully valid

crates/optimism/flashblocks/src/cache.rs
```diff
@@ -181,8 +187,7 @@ impl<T: SignedTransaction> SequenceManager<T> {
         // chain progression.
         let block_time_ms = (base.timestamp - local_tip_timestamp) * 1000;
         let expected_final_flashblock = block_time_ms / FLASHBLOCK_BLOCK_TIME;
-        let compute_state_root = self.compute_state_root &&
-            last_flashblock.diff.state_root.is_zero() &&
+        let compute_state_root = last_flashblock.diff.state_root.is_zero() &&
```
mattsse: this will now always compute the state root?

dont think this is what we want here?

mattsse (commented):
I'm a bit sceptical that this is always useful, because this can delay the latest flashblock for SR computation making it not available for rpc as fast as possible

at the very least I'd like to see some timings for base for this

also curious what @0x00101010 thinks here

crates/engine/tree/src/tree/mod.rs
sieniven: Fixed in d35f59bc24

crates/engine/primitives/src/message.rs
sieniven: I see where youre coming from, I am unsure of an alternative safer approach to this - I have instead renamed this better to FlashblocksSequence instead so that we explicitly know this message is handled entirely only for flashblocks rpc layer (d35f59bc24).

crates/optimism/flashblocks/src/cache.rs
sieniven: The state root calculation has since been switched to async in this PR, which allows us to quickly update the pending state after flashblock payload execution.

sieniven:
@mattsse hey, would appreciate another review for this as this has been pending for awhile.  As flashblocks are still pre-confirmed state changes and we cannot assume the state confirmations of the chain until the default optimism block is received, this PR helps to allow flashblocks to optimize the RPC sync, while still allow support for CL/EL driven syncs (on top of flashblock consensus sync, which is unsafe in nature). Separately, we have also launched this feature on X Layer mainnet and testnet (chainstate is driven by CL), and it works fine and stable.

We are in the midst of getting your required metrics for Base mainnet - will send the metrics here asap once we get the datapoints.

PR #20567, opened 2025-12-22
fix(engine): sync invalid header cache count gauge on hit eviction

VolodymyrBg:
The invalid header cache now updates the count gauge after evicting an entry when its hit counter reaches `INVALID_HEADER_HIT_EVICTION_THRESHOLD`. Previously we only incremented `hit_evictions` and removed the entry from headers without adjusting count, so the metric could remain higher than the actual cache size after repeated hit-based evictions. This change keeps the count gauge consistent with `headers.len()` without altering cache behavior or eviction semantics.


mattsse (approved):
makes sense

PR #20569, opened 2025-12-22
refactor(stages): consolidate prune checkpoint pattern in 3 stages

letmehateu:
Refactors duplicated prune checkpoint save logic across 3 stage implementations (AccountHistory, StorageHistory, and TransactionLookup) into a single reusable helper function maybe_save_prune_checkpoint().


mattsse (commented):
we can keep this as is

PR #20570, opened 2025-12-22
fix(fs-util): append .tmp extension instead of replacing original

dizer-ti:
Fixed atomic_write_file where set_extension("tmp") was replacing the original file extension instead of appending .tmp to the filename.

Before: config.json -> config.tmp
After: config.json -> config.json.tmp

This caused a race condition when atomically writing files with the same name but different extensions (e.g. data.json and data.xml would both use data.tmp as their temp file).


mattsse (commented):
we can keep as is

PR #20572, opened 2025-12-22
perf: spawn prewarming before multiproof

klkvr:
Reorders spawning logic so that prewarming is spawned before `MultiProofTask`

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -51,7 +51,7 @@ use std::{
     },
     time::Instant,
 };
-use tracing::{debug, debug_span, error, instrument, warn, Span};
+use tracing::{debug, debug_span, instrument, warn, Span};
 
 pub mod bal;
 mod configured_sparse_trie;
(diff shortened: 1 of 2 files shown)


devnet.json
mattsse, line 1: empty?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -236,44 +236,15 @@ where
...
-
-        let multi_proof_task = MultiProofTask::new(
-            proof_handle.clone(),
-            to_sparse_trie,
-            config.multiproof_chunking_enabled().then_some(config.multiproof_chunk_size()),
-        );
-
-        // wire the multiproof task to the prewarm task
-        let to_multi_proof = Some(multi_proof_task.state_root_message_sender());
+        let (to_multi_proof, from_multi_proof) = crossbeam_channel::unbounded();
```
mattsse, line 239: I see, we create the channel outside now

mattsse (approved):
lgtm

needs rm redundant file

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -290,10 +261,37 @@ where
...
-                to_multi_proof.clone(),
+                Some(to_multi_proof.clone()),
             )
         };
 
+        // We rely on the cursor factory to provide whatever DB overlay is necessary to see a
+        // consistent view of the database, including the trie tables. Because of this there is no
+        // need for an overarching prefix set to invalidate any section of the trie tables, and so
+        // we use an empty prefix set.
+
```
mediocregopher:
```suggestion
```

We don't need this comment anymore imo

mediocregopher (approved):

PR #20574, opened 2025-12-22
fix(era-downloader): ensure download errors are returned for all file types

hawkadrian:
### What:
Ensures download_to_file returns download errors for all file types, not just Era1.
### Why:
Previously, when all retry attempts failed for non-Era1 files, the function returned Ok(path) even though the download failed. 

This could cause the caller to treat a failed download as successful, leading to incorrect behavior downstream.


mattsse (commented):
we only need this for era1

PR #20575, opened 2025-12-22
perf: spawn prewarm workers in parallel

klkvr:
Closes #20573

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -164,12 +164,7 @@ where
             };
 
             // Initialize worker handles container
-            let mut handles = Vec::with_capacity(workers_needed);
-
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -536,27 +531,43 @@ where
...
+        for _ in 0..workers_needed {
+            let (tx, rx) = mpsc::channel();
+            handles.push(tx);
+            receivers.push(rx);
+        }
+
+        // Spawn a separate task spawning workers in parallel.
+        let executor = task_executor.clone();
+        let span = Span::current();
+        task_executor.spawn_blocking(move || {
```
mattsse, line 556: oh I see, this offloads spawning

mattsse (approved):
makes sense

PR #20580, opened 2025-12-23
fix: fuse shutdown

mattsse:
the sender can be dropped before this future terminates so this can end up polling the terminated channel multiple times

we can simply fuse here, because if the shutdown succeeds then the `engine_service` will also terminate

ref #20420

### DIFF
--- a/crates/node/builder/src/launch/engine.rs
+++ b/crates/node/builder/src/launch/engine.rs
@@ -9,7 +9,7 @@ use crate::{
     NodeBuilderWithComponents, NodeComponents, NodeComponentsBuilder, NodeHandle, NodeTypesAdapter,
 };
 use alloy_consensus::BlockHeader;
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

yongkangc (approved):
ahh i see! polling a completed future again can panic so we need fuse

crates/node/builder/src/launch/engine.rs
DaniPopes, line 313: what about breaking here?

mattsse: not needed because this will terminate once actually shut down

https://github.com/paradigmxyz/reth/blob/c822c96d3528a33b65a7e150b7e9e230a4ee89f0/crates/node/builder/src/launch/engine.rs#L322-L323

PR #20584, opened 2025-12-23
feat: integrate newPayload into ethstats

### LINKED ISSUE #20578: Integrate newPayload into ethstats
### Describe the feature

ref https://github.com/ethereum/go-ethereum/pull/33395/changes

for this we need a new event type in

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/node/ethstats/src/events.rs#L1-L1

and then we need a way to subscribe to new blocks which we can do via 

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/engine/primitives/src/event.rs#L25-L30

which already includes the duration (how long it took)

like

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/node/ethstats/src/ethstats.rs#L309-L309

but `reportNewPayload`

here we then need to give access to the event channel

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/node/builder/src/launch/common.rs#L1072-L1072

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/node/builder/src/launch/engine.rs#L392-L392

https://github.com/paradigmxyz/reth/blob/eec76a3fafd671e73b5d5af74fd21a9d1e85f869/crates/node/builder/src/launch/engine.rs#L188-L188

@stevencartavia 

### Additional context

_No response_

stevencartavia:
closes #20578

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -1060,13 +1060,10 @@ where
     }
 
     /// Spawns the [`EthStatsService`] service if configured.
-    pub async fn spawn_ethstats<E>(&self, engine_events: Option<E>) -> eyre::Result<()>
+    pub async fn spawn_ethstats<St>(&self, mut engine_events: St) -> eyre::Result<()>
     where
-        E: Stream<
-                Item = reth_engine_primitives::ConsensusEngineEvent<
-                    <T::Types as reth_node_api::NodeTypes>::Primitives,
-                >,
-            > + Send
+        St: Stream<Item = reth_engine_primitives::ConsensusEngineEvent<PrimitivesTy<T::Types>>>
+            + Send
             + Unpin
             + 'static,
     {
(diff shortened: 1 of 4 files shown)


crates/node/ethstats/src/ethstats.rs
```diff
@@ -50,7 +50,7 @@ const READ_TIMEOUT: Duration = Duration::from_secs(30);
 /// authentication, stats reporting, block notifications, and connection management.
 /// It maintains a persistent `WebSocket` connection and automatically reconnects
 /// when the connection is lost.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
mattsse, line 53: why do we need this to be clone?

crates/node/builder/src/launch/common.rs
```diff
@@ -1070,6 +1090,36 @@ where
         info!(target: "reth::cli", "Starting EthStats service at {}", url);
 
         let ethstats = EthStatsService::new(url, network, provider, pool).await?;
+
+        // If engine events are provided, spawn listener for new payload reporting
+        if let Some(engine_events) = engine_events {
+            let ethstats_clone = ethstats.clone();
+            let task_executor = self.task_executor().clone();
+            task_executor.spawn_critical(
+                "ethstats engine events listener",
```
mattsse: I think we can spawn a regular task here

crates/node/builder/src/launch/common.rs
```diff
@@ -1061,6 +1061,26 @@ where
...
+        self.spawn_ethstats_with_optional_engine_events(
+            None::<futures::stream::Empty<reth_engine_primitives::ConsensusEngineEvent<_>>>,
+        )
+        .await
+    }
+
+    /// Spawns the [`EthStatsService`] service if configured, with optional engine events.
+    pub async fn spawn_ethstats_with_optional_engine_events<E>(
+        &self,
+        engine_events: Option<E>,
```
mattsse: we can remove the option here

crates/node/builder/src/launch/common.rs
```diff
@@ -1061,6 +1061,26 @@ where
...
     /// Spawns the [`EthStatsService`] service if configured.
     pub async fn spawn_ethstats(&self) -> eyre::Result<()> {
+        self.spawn_ethstats_with_optional_engine_events(
+            None::<futures::stream::Empty<reth_engine_primitives::ConsensusEngineEvent<_>>>,
+        )
+        .await
+    }
+
+    /// Spawns the [`EthStatsService`] service if configured, with optional engine events.
+    pub async fn spawn_ethstats_with_optional_engine_events<E>(
```
mattsse: let's just give spawn_ethstats the stream argument

crates/node/builder/src/launch/common.rs
```diff
@@ -1070,6 +1090,36 @@ where
...
+
+        // If engine events are provided, spawn listener for new payload reporting
+        if let Some(engine_events) = engine_events {
+            let ethstats_clone = ethstats.clone();
+            let task_executor = self.task_executor().clone();
+            task_executor.spawn_critical(
+                "ethstats engine events listener",
+                Box::pin(async move {
+                    let mut events_stream = engine_events;
+                    while let Some(event) = StreamExt::next(&mut events_stream).await {
```
mattsse: can we use .next() here?

crates/node/builder/src/launch/common.rs
```diff
@@ -1070,6 +1090,36 @@ where
...
+                        use reth_engine_primitives::ConsensusEngineEvent;
+                        if let ConsensusEngineEvent::ForkBlockAdded(executed, duration) = event {
+                            let block_hash = executed.recovered_block.num_hash().hash;
+                            let block_number = executed.recovered_block.num_hash().number;
+                            if let Err(e) = ethstats_clone
```
mattsse: we need a match here instead for both variants

https://github.com/paradigmxyz/reth/blob/66db0839a078b5b8f78c8262085442040926cab6/crates/engine/primitives/src/event.rs#L26-L26

https://github.com/paradigmxyz/reth/blob/66db0839a078b5b8f78c8262085442040926cab6/crates/engine/primitives/src/event.rs#L30-L30

mattsse (changes requested):
cool, only a few suggestions

crates/node/ethstats/src/ethstats.rs
stevencartavia: needed here because the main ethstats service (ethstats.run()) and engine events listener run in separate tasks, I think, with independent ownership. I tried to avoid it using Arc but couldn't since run() consumes self by value

mattsse (approved):

PR #20585, opened 2025-12-23
perf: spawn storage root tasks in parallel

Lorsmirq0cB:
Spawns storage root calculation tasks in parallel instead of serially, reduces task initialization overhead.

### DIFF
--- a/crates/trie/parallel/src/root.rs
+++ b/crates/trie/parallel/src/root.rs
@@ -100,33 +100,46 @@ where
         // Get runtime handle once outside the loop
         let handle = get_runtime_handle();
 
+        // Collect all storage root targets with their channels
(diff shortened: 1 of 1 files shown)


crates/trie/parallel/src/root.rs
```diff
@@ -100,31 +100,50 @@ where
...
+                        )
+                        .calculate(retain_updates)?)
+                    })();
+                    let _ = tx.send(result);
+                });
+            }
+        });
+
+        // Insert all receivers into the storage_roots map
+        for (hashed_address, rx) in receivers {
```
mediocregopher: Is this intermediate `receives` Vec necessary? Could we instead insert directly into `storage_roots` in the for-loop above?

mediocregopher (approved):

mediocregopher:
cc @yongkangc

PR #20586, opened 2025-12-23
feat: switch samply feature for CLI flags

DaniPopes:
The dependency is very lightweight and we want the CLI flag anyway for more fine-grained control.


yongkangc (approved):

mattsse (approved):
seems okay to not feature gate this

PR #20587, opened 2025-12-23
perf: cache hashed address in unwind_storage_hashing

Lorsmirq0cB:
Caches address hash for consecutive storage entries in unwind_storage_hashing, avoiding redundant keccak256 calculations.


mattsse (commented):
we can keep this as is because not critical perf wise

PR #20590, opened 2025-12-23
perf(witness): avoid unnecessary HashMap clone when converting to BTreeMap

doocho:
Replace `BTreeMap::from_iter(map.clone())` pattern with direct iteration and collection.
This avoids allocating an intermediate cloned HashMap when converting to BTreeMap for sorted comparison. Keys (U256, B256) are Copy types, so only values need to be cloned.


yongkangc (approved):
makes sense, before we have to allocate new hashmap to copy the keys 

but now we iterate on original hashmap, clone and build the btree

i.e reduce one full hashmap allocation and dellocation

mattsse (approved):
this is fine but kinda negligible

PR #20592, opened 2025-12-23
docs(exex): correct comparison order in backfill docs

hazeezs:
The doc comment had the comparison operands swapped - it said `node_head.number < exex_head.number` when describing ExEx being behind, but the actual code compares `exex_head.number.cmp(&node_head.number)`. Fixed to match the code logic.


mattsse (approved):
allowed, because this is actually a valid fix

PR #20594, opened 2025-12-23
feat(storage): add AccountsHistory RocksDB consistency check

### LINKED ISSUE #20595: Add invariance check for account history
### Describe the feature

Followup from https://github.com/paradigmxyz/reth/issues/20153

We now need to add the same invariance check, but for account history instead.

### Additional context

_No response_

yongkangc:
closes https://github.com/paradigmxyz/reth/issues/20595

RocksDB consistency checks only cover TransactionHashNumbers and StoragesHistory, but AccountsHistory was missing.

Add `check_accounts_history()` invariant check following the same pattern as
StoragesHistory.

**Related**
- Sibling PR: #20596 (consistency check ordering fix)

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -282,6 +282,18 @@ impl RocksDBProvider {
                         "StoragesHistory ahead of checkpoint, pruning excess data"
                     );
                     self.prune_storages_history_above(checkpoint)?;
+                    return Ok(None);
+                }
+
+                // If RocksDB is behind the checkpoint, request an unwind to rebuild.
+                if max_highest_block < checkpoint {
+                    tracing::warn!(
+                        target: "reth::providers::rocksdb",
+                        rocks_highest = max_highest_block,
+                        checkpoint,
+                        "StoragesHistory behind checkpoint, unwind needed"
+                    );
+                    return Ok(Some(max_highest_block));
                 }
 
                 Ok(None)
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -327,6 +334,125 @@ impl RocksDBProvider {
...
+
+    /// Prunes `AccountsHistory` entries where `highest_block_number` > `max_block`.
+    ///
+    /// For `AccountsHistory`, the key is `ShardedKey<Address>` which contains
+    /// `highest_block_number`, so we can iterate and delete entries where
+    /// `key.highest_block_number > max_block`.
+    ///
+    /// TODO(<https://github.com/paradigmxyz/reth/issues/20417>): this iterates the whole table,
+    /// which is inefficient. Use changeset-based pruning instead.
+    fn prune_accounts_history_above(&self, max_block: BlockNumber) -> ProviderResult<()> {
```
mattsse, line 437: I see, so rn this requires scanning the entire table?

yongkangc: yup, will have to do some refactoring with `prune_accounts_history_above` and `prune_storages_history_above` once https://github.com/paradigmxyz/reth/issues/20417 is in @mattsse

joshieDo: yeah we need changesets in SF to redo this, kinda as a placeholder here + testing asap other components ig

joshieDo (approved):
feel like we could dedup storage/account history here, but we can do it another time

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -327,6 +346,125 @@ impl RocksDBProvider {
...
+                        target: "reth::providers::rocksdb",
+                        "AccountsHistory has data but checkpoint is 0, clearing all"
+                    );
+                    self.prune_accounts_history_above(0)?;
+                    return Ok(None);
+                }
+
+                // Find the max highest_block_number (excluding u64::MAX sentinel) across all
+                // entries
+                let mut max_highest_block = 0u64;
```
joshieDo, line 384: i feel like there was already some kind of rocksdb checkpoint block somewhere?

yongkangc: @joshieDo 

  After searching the codebase, i dont think is an existing RocksDB checkpoint mechanism:

  1. No metadata table - No RocksDB*Progress or checkpoint tracking table exists
  2. `last()` won't help - Keys are `ShardedKey<Address>` sorted by `(address, highest_block_number)`, so last() returns the largest address, not the largest block number

PR #20596, opened 2025-12-23
fix(storage): rocksdb consistency check on startup

yongkangc:
Addresses [#20392](https://github.com/paradigmxyz/reth/issues/20392) - Initialize RocksDB provider and add consistency check on startup.

## Changes

- Add RocksDB `check_consistency()` call in node launch with correct ordering:
  1. `StaticFileProvider::check_file_consistency()` - heals interrupted writes
  2. `RocksDB::check_consistency()` - uses static file data for tx hash pruning  
  3. `StaticFileProvider::check_consistency()` - compares with MDBX checkpoints

Note for future:

- We have to follow up with passing RocksDB to the stages once implemented in https://github.com/paradigmxyz/reth/issues/20389
- And also we can spin off another pr for Configuration validation

**Related**
- Sibling PR: #20594 (AccountsHistory support)

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -498,18 +498,17 @@ where
         )?
         .with_prune_modes(self.prune_modes());
 
-        // Check for consistency between MDBX, static files, and RocksDB. If any
-        // inconsistencies are found, unwind to the first block that's consistent across all
-        // storage layers.
+        // Keep MDBX, static files, and RocksDB aligned. If any check fails, unwind to the
+        // earliest consistent block.
         //
-        // The ordering is critical:
-        // 1. File healing - heals NippyJar inconsistencies without pruning data
-        // 2. RocksDB check - needs static file tx data for hash lookups
-        // 3. Static file checkpoint check - compares with MDBX, may prune data
+        // Order matters:
+        // 1) heal static files (no pruning)
+        // 2) check RocksDB (needs static-file tx data)
+        // 3) check static-file checkpoints vs MDBX (may prune)
         //
-        // We compute a combined unwind target from all checks and run a single unwind pass.
+        // Compute one unwind target and run a single unwind.
 
-        // Step 1: Heal file-level inconsistencies (no pruning)
+        // Step 1: heal file-level inconsistencies (no pruning)
         factory.static_file_provider().check_file_consistency()?;
 
         let provider_ro = factory.database_provider_ro()?;
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -66,8 +66,9 @@ use reth_node_metrics::{
 };
 use reth_provider::{
     providers::{NodeTypesForProvider, ProviderNodeTypes, RocksDBProvider, StaticFileProvider},
-    BlockHashReader, BlockNumReader, ProviderError, ProviderFactory, ProviderResult,
-    StageCheckpointReader, StaticFileProviderBuilder, StaticFileProviderFactory,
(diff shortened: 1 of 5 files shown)


crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,61 @@ where
...
+        let file_unwind = factory.static_file_provider().check_file_consistency()?;
+
+        // Step 2: RocksDB consistency check (needs static files tx data)
+        let rocksdb_unwind =
+            factory.rocksdb_provider().check_consistency(&factory.database_provider_ro()?)?;
+
+        // Step 3: Static file checkpoint consistency (may prune)
+        let static_file_unwind = factory
+            .static_file_provider()
+            .check_consistency(&factory.provider()?)?
```
joshieDo: nit, can use the same provider_ro i think

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1184,6 +1184,45 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+            if initial_highest_block != highest_block {
+                info!(
+                    target: "reth::providers::static_file",
+                    ?initial_highest_block,
+                    unwind_target = highest_block,
+                    ?segment,
+                    "Setting unwind target."
+                );
+                let target = highest_block.unwrap_or_default();
+                unwind_target = Some(unwind_target.map_or(target, |t| t.min(target)));
```
joshieDo: i understand this is the current behaviour, but im unsure if we need to return an `unwind_target` at all. target should always be >= stage checkpoint , which then gets healed on `check_consistency`

crates/node/builder/src/launch/common.rs
```diff
@@ -501,20 +502,61 @@ where
...
+                (None, None, Some(_)) => "static file <> MDBX",
+                (None, None, None) => unreachable!(),
+            };
             assert_ne!(
-                unwind_target,
-                PipelineTarget::Unwind(0),
-                "A static file <> database inconsistency was found that would trigger an unwind to block 0"
+                unwind_block,
+                0,
+                "A {inconsistency_source} inconsistency was found that would trigger an unwind to block 0"
```
joshieDo: can we move this to an helper, and maybe just increment the err str or smth, because this 3-way-match is ugly

yongkangc: Good note, initially thought that it might be a shallow function

crates/node/builder/src/launch/common.rs
yongkangc: thanks for note, addrssed

crates/storage/provider/src/providers/static_file/manager.rs
yongkangc: good point, addressed this 👍🏻

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1184,6 +1158,74 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
+    pub fn check_file_consistency<Provider>(&self, provider: &Provider) -> ProviderResult<()>
+    where
+        Provider: DBProvider + ChainSpecProvider + StorageSettingsCache,
+    {
+        info!(target: "reth::cli", "Healing static file inconsistencies.");
+
+        for segment in self.segments_to_check(provider) {
+            let (initial_highest_block, highest_block) = self.maybe_heal_segment(segment)?;
+
+            let _ = (initial_highest_block, highest_block, segment);
```
fgimenez: why do we need this binding?
```suggestion
            let _ = self.maybe_heal_segment(segment)?;
```

fgimenez (approved):
lgtm, minor nit

PR #20597, opened 2025-12-23
refactor(metrics): deduplicate MeteredReceiver implementations

aso20455:
Extract common receiver methods (recv, try_recv, close, poll_recv, Stream impl) into impl_metered_receiver! macro to reduce code duplication between MeteredReceiver and UnboundedMeteredReceiver.


mattsse (commented):
we can keep this as is

PR #20598, opened 2025-12-23
fix(cli): constrain download execute generic

kurahin:
Add a CliNodeTypes<ChainSpec = C::ChainSpec> bound to DownloadCommand::execute and import CliNodeTypes from crate::common. This removes the unused type parameter and aligns the download command's execute signature with other CLI commands and the run_commands_with dispatcher.


mattsse (commented):
these additional bounds aren't necessary

PR #20603, opened 2025-12-23
docs: clarify state helpers header comment

yyhrnk:
The state helpers module had a copy-pasted header mentioning loading a pending block, which belongs to the pending_block helpers instead. This updates the top-level comment to describe state-related RPC helpers (accounts, storage, proofs) so that navigation and code search reflect the actual responsibility of the module, without changing any behavior.


mattsse (commented):
not needed

PR #20604, opened 2025-12-23
fix(stages): only iterate over active stages

joshieDo:
During engine sync, we iterate over all stages in `StageId::ALL` and bump their checkpoints via `update_pipeline_stages`. During the pipeline consistency check, we iterate all checkpoints and compare them to the header stage.

However, the pipeline itself conditionally adds `PruneSenderRecoveryStage` only if sender pruning is enabled:

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/stages/stages/src/sets.rs#L337-L340

This can result in the following weird scenario if sender pruning is not enabled:

1. `PruneSenderRecovery` checkpoint is set to block 100 by engine sync (via `update_pipeline_stages`)
2. Run unwind command to block 95. Since `PruneSenderRecoveryStage` is not in the pipeline, its checkpoint stays at 100 (ahead of other stages, but not an issue yet)
3. Run pipeline forward to block 105. Once again, `PruneSenderRecoveryStage` is not in the pipeline and won't run, so its checkpoint stays at 100
4. Now `PruneSenderRecovery` (100) < `Headers` (105), which triggers the inconsistency check:

or

1. Engine sync to block 50 → update_pipeline_stages creates PruneSenderRecovery at 50
2. Pipeline sync continues to block 100 → Headers at 100, but PruneSenderRecovery stuck at 50
3. CLI unwind to block 60 → Headers at 60, PruneSenderRecovery still at 50
4. reth node → consistency check: 50 < 60

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/node/builder/src/launch/common.rs#L975-L985

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -1563,9 +1563,11 @@ impl<TX: DbTxMut, N: NodeTypes> StageCheckpointWriter for DatabaseProvider<TX, N
         block_number: BlockNumber,
         drop_stage_checkpoint: bool,
     ) -> ProviderResult<()> {
-        // iterate over all existing stages in the table and update its progress.
+        // iterate over active stages and update their progress.
         let mut cursor = self.tx.cursor_write::<tables::StageCheckpoints>()?;
-        for stage_id in StageId::ALL {
+        // TODO: era_enabled should come from node config
+        let era_enabled = true;
+        for stage_id in StageId::active(era_enabled, self.prune_modes_ref()) {
             let (_, checkpoint) = cursor.seek_exact(stage_id.to_string())?.unwrap_or_default();
             cursor.upsert(
                 stage_id.to_string(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10734,6 +10734,7 @@ dependencies = [
  "proptest-arbitrary-interop",
  "rand 0.9.2",
  "reth-codecs",
+ "reth-prune-types",
  "reth-trie-common",
  "serde",
 ]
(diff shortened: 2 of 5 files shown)


mediocregopher (approved):
Makes sense to me 👍

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1563,9 +1563,11 @@ impl<TX: DbTxMut, N: NodeTypes> StageCheckpointWriter for DatabaseProvider<TX, N
         block_number: BlockNumber,
         drop_stage_checkpoint: bool,
     ) -> ProviderResult<()> {
-        // iterate over all existing stages in the table and update its progress.
+        // iterate over active stages and update their progress.
         let mut cursor = self.tx.cursor_write::<tables::StageCheckpoints>()?;
-        for stage_id in StageId::ALL {
+        // TODO: era_enabled should come from node config
+        let era_enabled = true;
```
joshieDo, line 1569: was not sure how to bring `era_enabled` here

hardcoding it to `true` here, results in the same behaviour as before. And should be fine since we filter it out anyway in the pipeline consistency check.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1563,9 +1563,11 @@ impl<TX: DbTxMut, N: NodeTypes> StageCheckpointWriter for DatabaseProvider<TX, N
         block_number: BlockNumber,
         drop_stage_checkpoint: bool,
     ) -> ProviderResult<()> {
-        // iterate over all existing stages in the table and update its progress.
+        // iterate over active stages and update their progress.
         let mut cursor = self.tx.cursor_write::<tables::StageCheckpoints>()?;
-        for stage_id in StageId::ALL {
+        // TODO: era_enabled should come from node config
```
mattsse, line 1568: what is the actually todo here?

if this is always true here, do we even need the bool arg now?

this isnt enabled by default so setting this to true here, feels wrong

PR #20615, opened 2025-12-23
fix(ipc): trim leading whitespace in StreamCodec decode

meetrick:
## Summary

Trim leading whitespace from decoded JSON-RPC messages in `StreamCodec`, resolving a TODO and improving IPC message handling.

## Problem

When `StreamCodec::decode()` parses JSON-RPC messages over IPC, it includes leading whitespace (newlines, tabs, spaces) that appears between messages. This is unnecessary since:

1. Leading whitespace is not part of the JSON message
2. It wastes memory storing meaningless bytes
3. It clutters debug logs with extra whitespace
4. The `start_idx` variable already tracks where JSON actually begins, but was not being used to discard prior bytes

## Before

```
Input:  "{ msg: 1 }\n\n\n\n{ msg: 2 }"
Output: "\n\n\n\n{ msg: 2 }"  // includes leading whitespace
```

## After

```
Input:  "{ msg: 1 }\n\n\n\n{ msg: 2 }"
Output: "{ msg: 2 }"  // clean JSON only
```

## User Benefits

- **Cleaner logs**: Debug output shows clean JSON without leading garbage
- **Memory efficiency**: No allocation for unnecessary whitespace bytes
- **Predictable output**: Messages always start with `{` or `[`

## Changes

- `stream_codec.rs`: Discard bytes before `start_idx` in `decode()`
- Updated `whitespace` test expectations
- Updated `huge` test to start with `{` directly
- Removed TODO comment

### DIFF
--- a/crates/rpc/ipc/src/stream_codec.rs
+++ b/crates/rpc/ipc/src/stream_codec.rs
@@ -113,7 +113,11 @@ impl tokio_util::codec::Decoder for StreamCodec {
                 is_escaped = byte == b'\\' && !is_escaped && in_str;
 
                 if depth == 0 && idx != start_idx && idx - start_idx + 1 > whitespaces {
-                    let bts = buf.split_to(idx + 1);
+                    // Discard any leading whitespace before the JSON start
+                    if start_idx > 0 {
+                        let _ = buf.split_to(start_idx);
+                    }
+                    let bts = buf.split_to(idx + 1 - start_idx);
                     return match String::from_utf8(bts.into()) {
                         Ok(val) => Ok(Some(val)),
                         Err(_) => Ok(None),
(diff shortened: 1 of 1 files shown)


yongkangc (approved):

crates/rpc/ipc/src/stream_codec.rs
```diff
@@ -113,7 +113,10 @@ impl tokio_util::codec::Decoder for StreamCodec {
                 is_escaped = byte == b'\\' && !is_escaped && in_str;
 
                 if depth == 0 && idx != start_idx && idx - start_idx + 1 > whitespaces {
-                    let bts = buf.split_to(idx + 1);
+                    if start_idx > 0 {
+                        let _ = buf.split_to(start_idx);
```
mattsse: I think this should use .advance?
see docs for split_to

mattsse (changes requested):

mattsse (approved):

Issue #7836, opened 2024-04-24
Block hash data inconsistency when reading data from a different process that leads to incorrect EVM execution.


dvush:
### Describe the bug

I believe that there are 2 issues 

1. When database is opened from a different process, trying to get the last 256 block hashes sometimes fails.
2. This will cause evm execution for txs that use `BLOCKHASH` opcode have incorrect execution without explicit fail. 

### Steps to reproduce

1. Have reth beta running on the machine
2. In the separate process open the database and static files storage (`ProviderFactory`)
3. For each new block try reading last 256 (except the current one) block hashes 

example code to detect that:
```rust
    for i in 1u64..=256 {
        let hash = provider_factory.block_hash(current_block_number - i)?;
        if hash.is_none() {
            eyre::bail!(
                "Missing historical block hash for block {}, current block: {}",
                current_block_number - i,
                current_block_number
            );
        }
```

### Platform(s)

Linux (x86)

### What version/commit are you on?

reth Version: 0.2.0-beta.6
Commit SHA: ac29b4b73

(external process and the node use the same version)

### What database version are you on?

Current database version: 2
Local database version: 2

### Which chain / network are you on?

mainnet

### What type of node are you running?

Archive (default)

### Code of Conduct

- [X] I agree to follow the Code of Conduct

0x416e746f6e:
🌷🌷🌷🌷🌷

(trying to see if some flowers could help bring some attention to this issue)

jenpaff:
@0x416e746f6e could you confirm that the issue is fixed? It seems like with Pectra this will become a non-issue since the last 256 block hashes will be served from state with (EIP-2935)[https://eips.ethereum.org/EIPS/eip-2935]

0x416e746f6e:
> [@0x416e746f6e](https://github.com/0x416e746f6e) could you confirm that the issue is fixed? It seems like with Pectra this will become a non-issue since the last 256 block hashes will be served from state with [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935), feel free to re-open if you're still running into issues here

@dvush do you know if this is fixed?  I am completely out of context

dvush:
As of today its not fixed but lets check after Pectra upgrade.

[...]

PR #20616, opened 2025-12-23
feat(rpc): add `reth_subscribeLatestPersistedBlock` subscription

cakevm:
When using direct database access `(examples/db-access)` and subscribing to new header events, it is currently necessary to add a delay to wait for persistence. This happens because the new header event over WebSockets is triggered before persistence. Even when using `--engine.persistence-threshold 0`, the event is emitted before persistence is triggered (and this is totally fine).

This PR adds a WebSocket subscription that emits notifications when blocks are persisted to disk, allowing external services to know when database reads reflect the latest canonical state. This can also be used when `--engine.persistence-threshold 0` is not provided, to determine when the state has been written.

Usage: `{"jsonrpc":"2.0","method":"reth_subscribeLatestPersistedBlock","params":[],"id":1}`

PS: I am currently investigating whether https://github.com/paradigmxyz/reth/issues/7836 is still an issue. You can find a test repo here: https://github.com/cakevm/reth-stale-direct-db. This also allows you to test this feature. I tested it with and without `--engine.persistence-threshold 0`.

I can downgrade this to `debug` but thought it may be useful:
```
ConsensusEngineEvent::LatestPersistedBlock(num_hash) => {
   info!(number=num_hash.number, hash=?num_hash.hash, "Latest persisted block");
}
```
Looks like:
```
2025-12-23T16:26:02.298841Z  INFO Received block from consensus engine number=24076495 hash=0xba428297deae3da18ba9b1731e141283c189daccce1050bb25213f8e185a08a1
2025-12-23T16:26:02.333813Z  INFO State root task finished state_root=0x80fd2171cd0728134f98ac7e2d83d3e28817f517168d00c52a1aa565f6c756fe elapsed=7.41032ms
2025-12-23T16:26:02.333914Z  INFO Block added to canonical chain number=24076495 hash=0xba428297deae3da18ba9b1731e141283c189daccce1050bb25213f8e185a08a1 peers=4 txs=307 gas_used=21.73Mgas gas_throughput=707.14Mgas/second gas_limit=60.00Mgas full=36.2% base_fee=0.10Gwei blobs=8 excess_blobs=998 elapsed=30.722264ms
2025-12-23T16:26:02.447327Z  INFO Canonical chain committed number=24076495 hash=0xba428297deae3da18ba9b1731e141283c189daccce1050bb25213f8e185a08a1 elapsed=354.086µs
2025-12-23T16:26:02.448215Z  INFO New payload job created id=0x028ee883bdbafbdc parent=0xba428297deae3da18ba9b1731e141283c189daccce1050bb25213f8e185a08a1
2025-12-23T16:26:02.949877Z  INFO Latest persisted block number=24076495 hash=0xba428297deae3da18ba9b1731e141283c189daccce1050bb25213f8e185a08a1 <-- NEW
```

### DIFF
--- a/crates/engine/primitives/src/event.rs
+++ b/crates/engine/primitives/src/event.rs
@@ -32,6 +32,8 @@ pub enum ConsensusEngineEvent<N: NodePrimitives = EthPrimitives> {
     CanonicalChainCommitted(Box<SealedHeader<N::BlockHeader>>, Duration),
     /// The consensus engine processed an invalid block.
     InvalidBlock(Box<SealedBlock<N::Block>>),
+    /// Latest persisted block with its number and hash.
+    LatestPersistedBlock(BlockNumHash),
 }
 
 impl<N: NodePrimitives> ConsensusEngineEvent<N> {
@@ -72,6 +74,9 @@ where
             }
             Self::BlockReceived(num_hash) => {
                 write!(f, "BlockReceived({num_hash:?})")
+            }
+            Self::LatestPersistedBlock(block) => {
+                write!(f, "LatestPersistedBlock({block:?})")
             }
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10307,6 +10307,7 @@ dependencies = [
  "reth-storage-api",
  "reth-tasks",
  "reth-testing-utils",
+ "reth-tokio-util",
  "reth-transaction-pool",
  "reth-trie-common",
  "revm",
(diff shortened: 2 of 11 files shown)


crates/engine/primitives/src/event.rs
```diff
@@ -32,6 +32,8 @@ pub enum ConsensusEngineEvent<N: NodePrimitives = EthPrimitives> {
     CanonicalChainCommitted(Box<SealedHeader<N::BlockHeader>>, Duration),
     /// The consensus engine processed an invalid block.
     InvalidBlock(Box<SealedBlock<N::Block>>),
+    /// Latest persisted block with its number and hash.
+    LatestPersistedBlock(BlockNumHash),
```
mattsse, line 36: we can emit this in various ways, after looking at #20626 I'm a bit undecided what the best solution for this is.

I'm currently favoring a watch channel in the canonical tracker, which I think would be slightly more convenient over all

mattsse:
this makes a lot of sense imo, but I think a watch channel with header/block would be nicer from an sdk pov

mattsse:
superseded by #20877

Issue #19514, opened 2025-11-05
Proof Calc Rewrite: Integration


mediocregopher:
### Describe the feature

Integrate the new proof calculator into the proof tasks used by the state root task (sparse trie) during payload validation.

Integration will require:

* Using storage proof workers to calculate account storage roots, rather than doing so synchronously within account proofs.

* Caching account storage roots across account proof workers.

Initially the new implementation should only be enabled via a CLI flag, which will later be made into a default.

PR #20617, opened 2025-12-23
feat(trie): Add flag to enable proof v2 for storage proof workers

mediocregopher:
Towards #19514

* A new flag `--engine.enable-proof-v2` is added.
* The `StorageProofInput` type is made into an enum with a Legacy and V2 variant, so the storage proof workers can serve either kind of request.
* When submitting proof requests the flag is used to decide which variant of `StorageProofInput` to use.
* When the flag is enabled the storage workers will create a `proof_v2::StorageProofCalculator` during their startup, and have it available for calculating `StorageProofInput::V2`s.
* A new enum `StorageProofResult` is introduced, which functions similarly to `StorageProofInput`. The account workers will convert either variant of `StorageProofResult` to a `DecodedMultiProof`. Once account workers also support V2 proofs this step won't be necessary.


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -621,11 +625,6 @@ impl MultiProofTask {
         }
     }
```
mediocregopher, line 620: clippy complained this isn't used anymore, even though it was used in the tests. I just inlined the `tx.clone()` in the tests instead.

shekhirin (approved):
lfg

mattsse (approved):
lgtm, this is an entirely different code branch

yongkangc (approved):
nice, LGTM

PR #20620, opened 2025-12-23
docs: align admin_nodeInfo example with actual RPC response

MoNyAvA:
Update the admin_nodeInfo JSON example in the admin namespace docs to better match the actual RPC response shape.

- Show protocols.eth.difficulty as null instead of a PoW-style numeric value
- Add the protocols.eth.config field to reflect the chain configuration payload


mattsse (commented):
we can keep as is

PR #20622, opened 2025-12-23
docs: clarify parentBeaconBlockRoot validation rules

phrwlk:
The validation logic for parentBeaconBlockRoot already matches the Cancun/Prague Engine API, but the doc comment on validate_parent_beacon_block_root_presence incorrectly implied timestamp-based UnsupportedFork behavior for V1/V2 and pre-Cancun V3. This change updates the documentation to describe the actual version-dependent requirements (V1–V2: field unsupported, V3+: field required) and makes it explicit that fork-specific timestamp checks and UnsupportedFork errors are handled by validate_payload_timestamp.


mattsse (commented):
we can keep this as is

PR #20623, opened 2025-12-23
fix: resolve TODO in provider.rs

letmehateu:
Resolved TODO comment and replaced `.expect()` with `.unwrap_or_default()` for handling missing static file headers.


mattsse (commented):
we can keep this, using default might be problematic here

PR #20624, opened 2025-12-24
chore: correct misleading comment in find_node failure handling

arreaaas:
Fixed incorrect comment in evict_failed_find_nodes. The old comment claimed the node responded with fewer entries than expected, but the condition `!answered` triggers only when there was no response at all. The answered flag gets set on any response, so this branch handles complete timeouts, not partial responses.


mattsse (commented):
we can keep this as is

Issue #20621, opened 2025-12-23
Add metrics for `ForkchoiceStateTracker` state


Rjected:
### Describe the feature

Right now we have the `ForkchoiceStateTracker` which keeps track of the latest FCUs received by the engine API. We should track these in metrics, so that we can see what the latest valid state is without logs

https://github.com/paradigmxyz/reth/blob/29438631be7b6492ac5e5c3cb5202b78e8792059/crates/engine/primitives/src/forkchoice.rs#L4-L15

reject-i:
i’ll handle this

PR #20628, opened 2025-12-24
feat(engine/metrics): add latest valid fcu

figtracer:
#20621

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1424,10 +1424,31 @@ where
                                     if let Some(last_valid) =
                                         self.state.forkchoice_state_tracker.last_valid_state()
                                     {
+                                        let head = self
+                                            .sealed_header_by_hash(last_valid.head_block_hash)
+                                            .ok()
+                                            .flatten()
+                                            .map(|header| header.number())
+                                            .unwrap_or(0);
+
+                                        let safe = self
+                                            .sealed_header_by_hash(last_valid.safe_block_hash)
+                                            .ok()
+                                            .flatten()
+                                            .map(|header| header.number())
+                                            .unwrap_or(0);
+
+                                        let finalized = self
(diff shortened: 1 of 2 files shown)


Rjected (changes requested):
Should have specified this better in the issue, but I don't think it's useful to have the hash bits as a gauge. We should instead have a number for the latest valid fcu head / safe / finalized

figtracer:
this will run if fcu is invalid as well, but ig it's not that relevant

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1413,15 +1413,36 @@ where
...
                                     self.state
                                         .forkchoice_state_tracker
-                                        .set_latest(state, res.outcome.forkchoice_status());
+                                        .set_latest(state, forkchoice_status);
+
+                                    // update forkchoice state tracker metrics with last valid state
+                                    if let Some(last_valid) =
+                                        self.state.forkchoice_state_tracker.last_valid_state()
+                                    {
+                                        let block_num = |hash| {
```
mattsse, line 1427: his now most likely always goes to disk for the finalized/safe block, before we send the respond

we need to avoid this

we already track the finalized/safe/head block, unsure what the use case for these metrics are here

is this just for tracking what we receive as fcu or what we actually have as finalized/safe/head @Rjected ?

because we already have:

https://github.com/paradigmxyz/reth/blob/dea27a55a860ecc85edbea276696ddb2bab8b3b8/crates/chain-state/src/chain_info.rs#L100-L131

mattsse (commented):
not perfectly clear to me what the use case for this is

> We should track these in metrics, so that we can see what the latest valid state is without logs

the last valid tags would track:

https://github.com/paradigmxyz/reth/blob/dea27a55a860ecc85edbea276696ddb2bab8b3b8/crates/chain-state/src/chain_info.rs#L154-L161

so perhaps we add the metrics there instead?

PR #20629, opened 2025-12-24
docs: remove outdated gas metrics TODO in backfill job

forkfury:
## Fix

Removes outdated TODO comment about gas metrics reporting in backfill job execution.

## Changes

- Replace TODO comment with explanatory note that gas metrics are already tracked via `cumulative_gas` and logged at batch level using `format_gas_throughput`
- Individual block-level gas metrics are not needed for backfill operations


mattsse (commented):
yeah we can remove this todo, but not with a comment that explains why we can remove it lmao

PR #20630, opened 2025-12-24
fix(exex): distinguish revert from reorg in notification conversion

DeVikingMark:
## Description

Fixes incorrect conversion of `CanonStateNotification::Reorg` to `ExExNotification`. When the `new` chain is empty (indicating a revert), the conversion now correctly creates `ChainReverted` instead of `ChainReorged`.

## Changes

- Added empty chain check in `From<CanonStateNotification>` implementation
- `Reorg` with empty `new` chain now converts to `ChainReverted`
- `Reorg` with non-empty `new` chain continues to convert to `ChainReorged`


mattsse (commented):
we can keep this as reorg because a reorg requires a new chain segment

PR #20635, opened 2025-12-24
perf(engine): paralellize evm_state_to_hashed_post_state()

figtracer:
according to benchmarks, this increasingly improves perf if accounts accessed >20, below this value a sequential approach is more beneficial due to overhead

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -203,8 +203,7 @@ impl Drop for StateHookSender {
 }
 
 pub(crate) fn evm_state_to_hashed_post_state(update: EvmState) -> HashedPostState {
-    let hashed: Vec<_> = update
-        .into_par_iter()
+    update.into_par_iter()
         .filter_map(|(address, account)| {
             if !account.is_touched() {
                 return None;
@@ -235,17 +234,7 @@ pub(crate) fn evm_state_to_hashed_post_state(update: EvmState) -> HashedPostStat
 
             Some((hashed_address, info, hashed_storage))
         })
-        .collect();
-
-    let mut hashed_state = HashedPostState::with_capacity(hashed.len());
-    for (hashed_address, info, hashed_storage) in hashed {
-        hashed_state.accounts.insert(hashed_address, info);
-        if let Some(storage) = hashed_storage {
-            hashed_state.storages.insert(hashed_address, storage);
-        }
-    }
-
-    hashed_state
+        .collect()
 }
 
 /// Input parameters for dispatching a multiproof calculation.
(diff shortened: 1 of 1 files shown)


gakonst:
Nice - what's the impact of this overall?

figtracer:
the function is called by the payload processor ```MultiProofTask::on_state_update()``` when we have state changes

i got these results when benchmarking on my machine
| Accounts Accessed| ~Before (µs) | ~After (µs) | ~Speedup    |
|------------------|-------------|------------|------------|
| 10               | 31          | 38         | 0.82x      |
| 20               | 61          | 51         | 1.2x  |
| 50               | 289         | 125        | 2.3x  |
| 100              | 571         | 214        | 2.7x   |
| 200              | 1,200       | 387        | 3.1x   |
| 300              | 1,810       | 560        | 3.2x  |

i think there's another PR working on this as well #20475

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -202,31 +203,45 @@ impl Drop for StateHookSender {
...
+
+            Some((hashed_address, info, hashed_storage))
+        })
+        .collect();
+
+    let mut hashed_state = HashedPostState::with_capacity(hashed.len());
+    for (hashed_address, info, hashed_storage) in hashed {
+        hashed_state.accounts.insert(hashed_address, info);
+        if let Some(storage) = hashed_storage {
+            hashed_state.storages.insert(hashed_address, storage);
```
mattsse: this we can simplify with 

https://github.com/paradigmxyz/reth/pull/20653

after

mattsse (changes requested):
imo this makes a lot of sense

we can even simplify the collect impl with #20653

crates/engine/tree/src/tree/payload_processor/multiproof.rs
figtracer: on it

figtracer:
there's potentially another optimization, similar to this one, here:
https://github.com/paradigmxyz/reth/blob/05b3a8668cae5c06fad3051853be5884b6c1299f/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L768-L779

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -202,35 +203,38 @@ impl Drop for StateHookSender {
...
+            } else {
+                let storage: Vec<_> = account
+                    .storage
+                    .into_iter()
+                    .filter(|(_slot, value)| value.is_changed())
+                    .map(|(slot, value)| (keccak256(B256::from(slot)), value.present_value))
+                    .collect();
+
+                if storage.is_empty() {
+                    None
```
mattsse, line 229: this is equivalent

mattsse (approved):

PR #6908, opened 2024-03-01
feat(node-core, storage): static files segment metrics

shekhirin:
Holesky
```console
➜  reth (alexey/static-files-segment-metrics) curl -s localhost:9001 | rg '^reth_static_files_segment' | sort
reth_static_files_segment_entries{segment="headers"} 1049527
reth_static_files_segment_entries{segment="receipts"} 32422175
reth_static_files_segment_entries{segment="transactions"} 32422175
reth_static_files_segment_files{segment="headers"} 3
reth_static_files_segment_files{segment="receipts"} 3
reth_static_files_segment_files{segment="transactions"} 3
reth_static_files_segment_size{segment="headers"} 456090364
reth_static_files_segment_size{segment="receipts"} 2274578626
reth_static_files_segment_size{segment="transactions"} 20599193088
```

To my surprise, no overhead 🤔 
```console
➜  reth (alexey/static-files-segment-metrics) hyperfine 'curl -s localhost:9001' -n 'alexey/static-files-segment-metrics'
Benchmark 1: alexey/static-files-segment-metrics
  Time (mean ± σ):      27.4 ms ±   1.8 ms    [User: 1.8 ms, System: 2.7 ms]
  Range (min … max):    25.6 ms …  37.7 ms    73 runs

➜  reth (main) hyperfine 'curl -s localhost:9001' -n 'main'
Benchmark 1: main
  Time (mean ± σ):      28.6 ms ±   1.1 ms    [User: 1.9 ms, System: 2.9 ms]
  Range (min … max):    27.5 ms …  36.7 ms    92 runs
```

---

image

PR #20636, opened 2025-12-25
docs: remove outdated static file metrics TODOs

SophiaR-baty:
Removes outdated metrics TODOs present since Feb 2024, author implemented static file metrics in PR #6908 without these duration metrics, elapsed time already tracked via events.


mattsse (commented):
still todo

PR #20637, opened 2025-12-25
chore(primitives-traits): correct set_timestamp parameter name and type

aso20455:
Fixed copy-paste error in HeaderMut trait. The set_timestamp method had parameter named `number` with type `BlockNumber` instead of `timestamp: u64`. This was clearly copied from set_block_number and not updated. The code worked because both are u64, but the naming was misleading.


mattsse (approved):
yeah we shouldnt use the type alias here

PR #20638, opened 2025-12-25
perf: allocate signer vec exact size

mattsse:
we know the exact tx count and can allocate accordingly.

currently we'd end up with spare capacity


yongkangc (approved):
nice

PR #20639, opened 2025-12-25
chore: shrink range result vec to fit

mattsse:
it's possible that we over allocate here and these vecs frozen, so we should shrink to fit them here


yongkangc (approved):
nice!

PR #20640, opened 2025-12-25
chore(e2e): remove unused next_payload_id from NodeState

lisenokdonbassenok:
The next_payload_id field in NodeState was only ever written once in GenerateNextPayload and never read anywhere in the e2e test suite or upstream sources. This field provided no additional information beyond payload_id_history and latest_payload_* and effectively acted as dead state. This change removes the field, its default initialization and Debug output, and drops the lone write in GenerateNextPayload, keeping NodeState focused on actually used engine state without changing runtime behavior.


mattsse (commented):
we can keep this

PR #13362, opened 2024-12-12
feat(ethereum-forks): remove total difficulty for hardfork check

aroralanuk:
- fixes https://github.com/paradigmxyz/reth/issues/13174

PR #20641, opened 2025-12-25
fix(rpc): use EthereumHardforks trait for Paris activation check

davidjsonn:
This fixes a regression introduced in #13362 where the Paris hardfork check was hardcoded for mainnet and sepolia only. For any other chain (custom PoW networks, testnets, etc.), `is_paris_activated` was always set to `true`, which meant block rewards were never calculated - even on chains that haven't merged yet.

The original code before #13362 used `chain_spec.is_paris_active_at_block()` which works correctly for all chains. The refactor to remove total difficulty checks accidentally replaced this with a broken hardcoded approach.

This change restores the proper behavior by using the `EthereumHardforks` trait method, which checks the chainspec configuration for any network.

### DIFF
--- a/crates/rpc/rpc/src/trace.rs
+++ b/crates/rpc/rpc/src/trace.rs
@@ -17,7 +17,7 @@ use alloy_rpc_types_trace::{
 };
 use async_trait::async_trait;
 use jsonrpsee::core::RpcResult;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is better actually, ty

PR #20643, opened 2025-12-26
fix(chainspec): use lazy error formatting in chain spec macro

oooLowNeoNooo:
## Problem

`alloc::format!` in the macro runs on every `LazyLock` init, even when there's no error. This allocates memory unnecessarily.

## Solution

Use `unwrap_or_else` so the error message is only formatted when an error occurs. This avoids allocation on the happy path and matches the pattern used elsewhere in the codebase.


mattsse (approved):
fine

PR #20645, opened 2025-12-26
fix(trie): verify empty storage tries when no storage updates exist

ANtutov:
The trie verifier previously assumed that if state root updates contained no storage branch nodes then there was nothing to validate for storage tries. In that scenario we never instantiated a storage SingleVerifier nor invoked verify_empty_storages, so stray storage nodes in the trie tables were silently ignored when hashed storage was empty for all accounts.

This change makes try_next call verify_empty_storages over the full account range when no BranchNode::Storage was ever observed. This ensures that any remaining storage nodes in the trie tables are reported as StorageExtra, keeping the trie tables consistent with the hashed state. A regression test was added to cover the case where hashed storage is empty but the storage trie still contains nodes


mattsse (commented):
dont think we need this

PR #20646, opened 2025-12-26
fix(db-api): set is_done=true in RangeWalker when cursor returns None

dizer-ti:
Set `is_done` to `true` in `RangeWalker::next()` when cursor returns `None`, regardless of the range bound type. Previously, `is_done` was only set for `Unbounded` ranges, leading to unnecessary `cursor.next()` calls on subsequent 
 iterations for bounded ranges.

### DIFF
--- a/crates/storage/db-api/src/cursor.rs
+++ b/crates/storage/db-api/src/cursor.rs
@@ -282,7 +282,7 @@ impl<T: Table, CURSOR: DbCursorRO<T>> Iterator for RangeWalker<'_, T, CURSOR> {
             },
             Some(res @ Err(_)) => Some(res),
             None => {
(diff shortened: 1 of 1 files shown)


crates/storage/db-api/src/cursor.rs
```diff
@@ -282,7 +282,7 @@ impl<T: Table, CURSOR: DbCursorRO<T>> Iterator for RangeWalker<'_, T, CURSOR> {
             },
             Some(res @ Err(_)) => Some(res),
             None => {
-                self.is_done = matches!(self.end_key, Bound::Unbounded);
```
mattsse: I believe this doesn't really matter because this iterator returns None here anyway and should terminate 
so we can keep this as is @joshieDo ?

PR #20647, opened 2025-12-26
chore: remove duplicate BlockBodyIndices table import

strmfos:
BlockBodyIndices table was imported twice during execution stage dump - first in setup() with range from-1..to+1, then again in import_tables_with_range() with from..to. The second import is redundant since the data already exists from the first call. Removed the unnecessary duplicate.


mattsse (commented):
we can keep this

PR #20648, opened 2025-12-26
s3 downloader: handle EmptyContentLength error explicitly

Olexandr88:
Explicitly handle missing Content-Length in S3 downloader.


mattsse (commented):
dont think we need this

PR #20655, opened 2025-12-28
refactor: unify witness generation with DebugApi using ExecutionWitnessRecord

forkfury:
## Fix

Unifies witness generation with `DebugApi::debug_execution_witness` by replacing custom implementation with `ExecutionWitnessRecord::from_executed_state()`.

## Changes

- Replace `collect_execution_data()` and `generate()` with `ExecutionWitnessRecord::from_executed_state()`
- Remove duplicate `state_by_block_hash()` call in `re_execute_block()`
- Add `witness` feature to `reth-revm` dependency in `Cargo.toml`
- Update tests to use `ExecutionWitnessRecord` instead of custom functions
- Remove TODO comment about unification


mattsse (commented):
we can keep this as is

PR #20657, opened 2025-12-28
fix(banlist): check ban expiration in is_banned methods

mattvaux:
Methods is_banned_ip and is_banned_peer now verify 
that temporary bans haven't expired before returning true. 

Previously, expired bans were only removed by periodic evict calls, 
causing peers to be incorrectly blocked between eviction cycles.


mattsse (commented):
this isn't needed because the expiry is supposed to be maintained separately

PR #20660, opened 2025-12-29
chore(rpc): shrink active filters HashMap after clearing stale entries

meetrick:
## Summary

Add `shrink_to_fit()` call after removing stale filters in `EthFilter::clear_stale_filters()`.

## Problem

When filters are removed via `HashMap::retain()`, the HashMap keeps its allocated capacity. This means if many filters are created and then expire, the memory stays allocated until the node restarts.

**Example scenario:**
1. 10,000 filters are created → HashMap capacity grows to ~10,000
2. All filters expire and get removed by `clear_stale_filters()`
3. HashMap now has 0 elements but still holds capacity for ~10,000 entries
4. This memory is never released

This is known as the "high water mark" problem.

## Solution

Call `shrink_to_fit()` after `retain()` to release unused capacity.

This follows the same pattern used in `EthStateCache::shrink_queues()` (added in #14105).

## Before vs After

| Scenario | Before | After |
|----------|--------|-------|
| 10,000 filters created then expired | Capacity: ~10,000 (memory retained) | Capacity: 0 (memory released) |
| Memory released when? | Only on node restart | Every 5 minutes (stale_filter_ttl) |

## Testing

- `cargo check -p reth-rpc --lib` ✅
- `cargo test -p reth-rpc --lib -- filter` ✅ (10 tests passed)

### DIFF
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -195,8 +195,6 @@ where
 
             is_valid
         });
-        // Release excess capacity after removing stale filters to prevent
-        // the HashMap from retaining memory indefinitely (high water mark problem).
         filters.shrink_to_fit();
     }
 }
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is reasonable because we can simply re-alloc for new filters

PR #20661, opened 2025-12-29
feat(txpool): add configurable batch timeout to transaction pool insertions

keanji-x:
Adds configurable transaction pool insertion batching with `--txpool.batch-timeout`.

The batcher now has two explicit modes. Immediate mode (no timeout, the default) processes requests as soon as they arrive and spawns insertion tasks without a concurrency bound, identical to the previous behavior. Batch-and-timeout mode coalesces requests until either `max-batch-size` is reached or the timeout expires after the first buffered transaction, and bounds in-flight insertion batches via `max_concurrent_batches` (derived from `txpool.additional-validation-tasks + 3`) so submission bursts cannot create unbounded concurrent batches while keeping validation workers saturated.

### DIFF
--- a/crates/node/core/src/args/txpool.rs
+++ b/crates/node/core/src/args/txpool.rs
@@ -55,6 +55,8 @@ pub struct DefaultTxPoolValues {
     transactions_backup_path: Option<PathBuf>,
     disable_transactions_backup: bool,
     max_batch_size: usize,
+    /// Batch timeout for collecting transactions before flushing
+    batch_timeout: Duration,
 }
 
 impl DefaultTxPoolValues {
@@ -247,6 +249,12 @@ impl DefaultTxPoolValues {
         self.max_batch_size = v;
         self
     }
+
+    /// Set the default batch timeout
+    pub const fn with_batch_timeout(mut self, v: Duration) -> Self {
+        self.batch_timeout = v;
+        self
+    }
 }
 
 impl Default for DefaultTxPoolValues {
@@ -282,6 +290,7 @@ impl Default for DefaultTxPoolValues {
             transactions_backup_path: None,
             disable_transactions_backup: false,
             max_batch_size: 1,
+            batch_timeout: Duration::ZERO,
         }
     }
 }
@@ -411,6 +420,12 @@ pub struct TxPoolArgs {
     /// Max batch size for transaction pool insertions
     #[arg(long = "txpool.max-batch-size", default_value_t = DefaultTxPoolValues::get_global().max_batch_size)]
     pub max_batch_size: usize,
+
+    /// Batch timeout for transaction pool insertions (e.g. "50ms" or "0" for immediate).
+    /// When non-zero, transactions are batched until either max-batch-size is reached OR timeout
+    /// expires. Use "0" for immediate processing (zero-cost path, original behavior).
+    #[arg(long = "txpool.batch-timeout", value_parser = parse_duration_from_secs_or_ms, value_name = "DURATION", default_value = format_duration_as_secs_or_ms(DefaultTxPoolValues::get_global().batch_timeout))]
+    pub batch_timeout: Duration,
 }
 
 impl TxPoolArgs {
@@ -464,6 +479,7 @@ impl Default for TxPoolArgs {
             transactions_backup_path,
             disable_transactions_backup,
             max_batch_size,
+            batch_timeout,
         } = DefaultTxPoolValues::get_global().clone();
         Self {
             pending_max_count,
@@ -496,6 +512,7 @@ impl Default for TxPoolArgs {
             transactions_backup_path,
             disable_transactions_backup,
             max_batch_size,
+            batch_timeout,
         }
     }
 }
@@ -546,6 +563,11 @@ impl RethTransactionPoolConfig for TxPoolArgs {
     /// Returns max batch size for transaction batch insertion.
     fn max_batch_size(&self) -> usize {
         self.max_batch_size
+    }
+
+    /// Returns batch timeout for transaction batch insertion.
+    fn batch_timeout(&self) -> Duration {
+        self.batch_timeout
     }
 }
 
@@ -625,6 +647,7 @@ mod tests {
             transactions_backup_path: Some(PathBuf::from("/tmp/txpool-backup")),
             disable_transactions_backup: false,
             max_batch_size: 10,
+            batch_timeout: Duration::from_millis(50),
         };
 
         let parsed_args = CommandParser::<TxPoolArgs>::parse_from([
@@ -685,6 +708,8 @@ mod tests {
             "/tmp/txpool-backup",
             "--txpool.max-batch-size",
             "10",
+            "--txpool.batch-timeout",
+            "50ms",
         ])
         .args;
 
--- a/crates/node/core/src/args/txpool.rs
+++ b/crates/node/core/src/args/txpool.rs
@@ -10,10 +10,11 @@ use reth_transaction_pool::{
     maintain::MAX_QUEUED_TRANSACTION_LIFETIME,
     pool::{NEW_TX_LISTENER_BUFFER_SIZE, PENDING_TX_LISTENER_BUFFER_SIZE},
     validate::DEFAULT_MAX_TX_INPUT_BYTES,
-    LocalTransactionConfig, PoolConfig, PriceBumpConfig, SubPoolLimit, DEFAULT_PRICE_BUMP,
-    DEFAULT_TXPOOL_ADDITIONAL_VALIDATION_TASKS, MAX_NEW_PENDING_TXS_NOTIFICATIONS,
-    REPLACE_BLOB_PRICE_BUMP, TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER,
-    TXPOOL_SUBPOOL_MAX_SIZE_MB_DEFAULT, TXPOOL_SUBPOOL_MAX_TXS_DEFAULT,
+    BatchConfig, LocalTransactionConfig, PoolConfig, PriceBumpConfig, SubPoolLimit,
+    DEFAULT_PRICE_BUMP, DEFAULT_TXPOOL_ADDITIONAL_VALIDATION_TASKS,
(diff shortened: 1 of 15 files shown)


crates/rpc/rpc/src/eth/builder.rs
```diff
@@ -42,6 +42,7 @@ pub struct EthApiBuilder<N: RpcNodeCore, Rpc, NextEnv = ()> {
     task_spawner: Box<dyn TaskSpawner + 'static>,
     next_env: NextEnv,
     max_batch_size: usize,
+    batch_timeout: Duration,
```
mattsse: nit: could we introude a dedicated Batchconfig type?

crates/transaction-pool/src/batcher.rs
```diff
@@ -39,28 +47,80 @@ where
...
         pool: Pool,
         max_batch_size: usize,
+        batch_timeout: Duration,
     ) -> (Self, mpsc::UnboundedSender<BatchTxRequest<Pool::Transaction>>) {
         let (request_tx, request_rx) = mpsc::unbounded_channel();
 
-        let processor = Self { pool, max_batch_size, buf: Vec::with_capacity(1), request_rx };
+        // Only create interval if timeout is non-zero (zero-cost abstraction)
+        let interval = if batch_timeout.is_zero() {
+            None
```
mattsse: can we make this an option then instead?

crates/transaction-pool/src/batcher.rs
```diff
@@ -98,23 +170,51 @@ where
...
+                match this.request_rx.as_mut().poll_recv_many(cx, this.buf, *this.max_batch_size) {
+                    Poll::Ready(0) => return Poll::Ready(()), // Channel closed
+                    Poll::Ready(_) => {
+                        Self::spawn_batch(this.pool, this.buf);
+                        this.buf.reserve(*this.max_batch_size);
+                        // continue to check for more requests
+                    }
+                    Poll::Pending => return Poll::Pending,
+                }
+            },
```
mattsse: all of this looks a bit too complex

how should this behave exactly?

can we not use just the interval to yield pending if it's not ready yet?

mattsse (changes requested):
this makes sense and I#ve thought about this as well because most of the time this ends up just batching 1 single request, but even waiting a bit should be very beneficial here.

left some suggestions and a q re the poll logic

crates/transaction-pool/src/batcher.rs
keanji-x: Fixed, thanks for review

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,46 @@ where
...
-                let pool = this.pool.clone();
-                tokio::spawn(async move {
-                    Self::process_batch(&pool, batch).await;
-                });
-                this.buf.reserve(1);
-
-                continue;
+            // 1. Collect available requests (non-blocking)
+            while this.buf.len() < *this.max_batch_size {
+                match this.request_rx.as_mut().poll_recv(cx) {
```
mattsse: I think this should keep using poll_recv_many

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,46 @@ where
...
+                !this.buf.is_empty() && interval.as_mut().poll_tick(cx).is_ready()
+            } else {
+                // Immediate mode: always flush if we have items
+                // This corresponds to "zero cost" greedy processing
+                !this.buf.is_empty()
+            };
+
+            if batch_full || timeout_ready {
+                Self::spawn_batch(this.pool, this.buf);
+                this.buf.reserve(*this.max_batch_size);
```
mattsse: this is wasteful, because this will most likely over allocate

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,46 @@ where
...
+                        // Channel closed, flush remaining and exit
+                        Self::spawn_batch(this.pool, this.buf);
+                        return Poll::Ready(())
+                    }
+                    Poll::Pending => break,
+                }
+            }
+
+            // 2. Check flush conditions
+            let batch_full = this.buf.len() >= *this.max_batch_size;
```
mattsse: if the batch is full we can spawn it immediately and then only need to reset the interval and can skip polling it

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,46 @@ where
...
+            let timeout_ready = if let Some(mut interval) = this.interval.as_mut().as_pin_mut() {
+                // If we have items and interval ticked, we should flush
+                !this.buf.is_empty() && interval.as_mut().poll_tick(cx).is_ready()
+            } else {
+                // Immediate mode: always flush if we have items
+                // This corresponds to "zero cost" greedy processing
+                !this.buf.is_empty()
+            };
+
+            if batch_full || timeout_ready {
```
mattsse: I find timeout_ready a bit confusing here because if no timeout is set, then this represents something else

mattsse (changes requested):
I'd like to restructure the poll logic a bit so that this behaves like:

1. poll_recv many
2. if batch empty -> return early
3. if batch full -> spawn + continue
4. if batch not full and interval is set: poll interval


I'd also like to see some manual poll tests for this

crates/transaction-pool/src/batcher.rs
keanji-x: fixed

keanji-x:
Hi @mattsse 👋 Sorry for the delay on this — I've addressed all the feedback from both review rounds and rebased onto latest main. Here's a summary of what's been updated:

1. Introduced a dedicated BatchConfig struct (as suggested)
2. Made `interval` an `Option<Interval>` for zero-cost immediate mode
3. Restructured the poll logic to follow your suggested flow: poll_recv_many → empty check → batch full spawn (skip interval) → interval poll for partial batches
4. Removed the wasteful buf.reserve (using std::mem::take now)
5. Renamed timeout_ready → should_flush for clarity
6. Added manual poll tests for: full batch, partial batch with interval, immediate mode, and channel close flush

Would appreciate another look when you get a chance. Thanks! 🙏

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,62 @@ where
...
+                if let Some(mut interval) = this.interval.as_mut().as_pin_mut() {
+                    interval.as_mut().reset();
+                }
+                continue
+            }
+
+            // 4. Batch not full - check if we should flush based on mode
+            let should_flush = if let Some(mut interval) = this.interval.as_mut().as_pin_mut() {
+                // Batch-and-timeout mode: flush only if interval ticked
+                interval.as_mut().poll_tick(cx).is_ready()
```
mattsse: I think this can use the ready! macro here?

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,62 @@ where
...
+                // Immediate mode: always flush if we have items (greedy processing)
+                true
+            };
+
+            if should_flush {
+                Self::spawn_batch(this.pool, this.buf);
+                // Reset interval if present
+                if let Some(mut interval) = this.interval.as_mut().as_pin_mut() {
+                    interval.as_mut().reset();
+                }
```
mattsse: this seems uneccessary if we can only enter this once we have a new tick?

crates/transaction-pool/src/batcher.rs
```diff
@@ -99,22 +171,62 @@ where
...
+                    Poll::Pending => {
+                        // No items immediately available
+                    }
+                }
+            }
+
+            // 2. Early return if buffer is empty
+            if this.buf.is_empty() {
+                return Poll::Pending
+            }
```
mattsse: this makes sense here, can we clarify that we can ignore the interval interval here because there's nothing buffered

crates/transaction-pool/src/batcher.rs
mattsse: isnt this always ready after we have

1. flushed all
2. return pending
3. ...wait for new tx to arrive (e.g. 1s)
4. then this is immediately ready?

crates/node/core/src/args/txpool.rs
```diff
@@ -411,6 +421,12 @@ pub struct TxPoolArgs {
     /// Max batch size for transaction pool insertions
     #[arg(long = "txpool.max-batch-size", default_value_t = DefaultTxPoolValues::get_global().max_batch_size)]
     pub max_batch_size: usize,
+
+    /// Batch timeout for transaction pool insertions (e.g. "50ms" or "0" for immediate).
```
mattsse, line 425: does this also support sub ms?

keanji-x: No, `parse_duration_from_secs_or_ms`only supports ms and s suffixes. I can extend `parse_duration_from_secs_or_ms` to also support us in a follow-up PR if needed.

crates/transaction-pool/src/batcher.rs
keanji-x: Yes, in low-frequency scenarios (tx arrival interval > batch_timeout), poll_tick will be immediately ready, effectively degrading to immediate mode. This is intentional — the timeout is designed for high-throughput batching; when transactions arrive infrequently, there's no benefit in waiting for more items, and flushing immediately reduces latency.

crates/transaction-pool/src/batcher.rs
keanji-x: Added a clarifying comment: "no need to poll the interval since there is nothing buffered to flush

crates/transaction-pool/src/batcher.rs
keanji-x: You're right, removed the reset here.

crates/transaction-pool/src/batcher.rs
keanji-x: 👍 Good call, refactored to use ready! — removed the should_flush variable entirely.

emmajam:
hey @keanji-x do we still want this PR?

PR #20662, opened 2025-12-29
perf: preallocate block body indices vec

SophiaR-baty:
Preallocates exact capacity for block_body_indices vector in receipts_by_block_range, reduces memory reallocations during block range iteration.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -1413,7 +1413,12 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> ReceiptProvider for DatabasePr
         }
 
         // collect block body indices for each block in the range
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1453,7 +1453,8 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> ReceiptProvider for DatabasePr
         }
 
         // collect block body indices for each block in the range
-        let mut block_body_indices = Vec::new();
+        let range_len = (block_range.end() - block_range.start() + 1) as usize;
```
mattsse: this could be malformed, so we should check that the range isnt empty here

mattsse (changes requested):

mattsse (commented):
this is just unchecked with extra steps

PR #20663, opened 2025-12-29
refactor(db-api): remove dead data field from DatabaseMock

kurahin:
The public DatabaseMock::data field was never read or written anywhere in the codebase and was not connected to TxMock or cursor implementations. At the same time its documentation claimed it was internal storage for the mock database, which does not match the actual stateless, no-op behavior of DatabaseMock and TxMock. This change removes the dead data field and its misleading comment, keeping DatabaseMock as a simple empty mock type that reflects the current semantics without altering any runtime behavior or tests.


mattsse (commented):
we can keep this

PR #20664, opened 2025-12-29
perf(trie): flatten sparse trie branch node masks to reduce overhead

yongkangc:
Follow up from #20659
Consolidate into a single `HashMap<Nibbles, BranchNodeMasks>`, reusing the `BranchNodeMasks` struct and `BranchNodeMasksMap` type alias introduced in #20659.

**Problem**
`SerialSparseTrie` and `ParallelSparseTrie` use two separate `HashMap<Nibbles, TrieMask>` fields (`branch_node_tree_masks` and `branch_node_hash_masks`) to store branch node mask information. This results in:
- Double HashMap overhead per path
- Two lookups instead of one when accessing both masks for the same path
- Increased memory footprint

**Impact**

Benchmark:

E2E: https://github.com/paradigmxyz/reth/pull/20664#issuecomment-3698212688

Micro: https://gist.github.com/yongkangc/a7ebeb047362fa1690fbd4a1c264ed43

| Operation | Elements | Old (µs) | New (µs) | Speedup |
|-----------|----------|----------|----------|---------|
| Insert | 1000 | 43.4 | 28.0 | **1.55x** |
| Lookup | 1000 | 13.8 | 5.6 | **2.47x** |
| Clone | 1000 | 3.3 | 1.7 | **1.94x** |
| Clear | 1000 | 2.4 | 0.7 | **3.42x** |
| Conditional Access | 1000 | 8.5 | 4.3 | **1.98x** |

Memory:
- Old: 96 bytes per struct (two HashMaps × 48 bytes each)
- New: 48 bytes per struct (single HashMap)
- Savings: 48 bytes (50% reduction)

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -9,8 +9,8 @@ use alloy_trie::{BranchNodeCompact, TrieMask, EMPTY_ROOT_HASH};
 use reth_execution_errors::{SparseTrieErrorKind, SparseTrieResult};
 use reth_trie_common::{
     prefix_set::{PrefixSet, PrefixSetMut},
-    BranchNodeRef, ExtensionNodeRef, LeafNodeRef, Nibbles, ProofTrieNode, RlpNode, TrieMasks,
-    TrieNode, CHILD_INDEX_RANGE,
+    BranchNodeMasks, BranchNodeMasksMap, BranchNodeRef, ExtensionNodeRef, LeafNodeRef, Nibbles,
+    ProofTrieNode, RlpNode, TrieMasks, TrieNode, CHILD_INDEX_RANGE,
 };
 use reth_trie_sparse::{
     provider::{RevealedNode, TrieNodeProvider},
@@ -112,10 +112,12 @@ pub struct ParallelSparseTrie {
     prefix_set: PrefixSetMut,
     /// Optional tracking of trie updates for later use.
     updates: Option<SparseTrieUpdates>,
-    /// When a bit is set, the corresponding child subtree is stored in the database.
-    branch_node_tree_masks: HashMap<Nibbles, TrieMask>,
-    /// When a bit is set, the corresponding child is stored as a hash in the database.
-    branch_node_hash_masks: HashMap<Nibbles, TrieMask>,
+    /// Branch node masks containing tree_mask and hash_mask for each path.
+    /// - `tree_mask`: When a bit is set, the corresponding child subtree is stored in the
+    ///   database.
+    /// - `hash_mask`: When a bit is set, the corresponding child is stored as a hash in the
+    ///   database.
+    branch_node_masks: BranchNodeMasksMap,
     /// Reusable buffer pool used for collecting [`SparseTrieUpdatesAction`]s during hash
     /// computations.
     update_actions_buffers: Vec<Vec<SparseTrieUpdatesAction>>,
@@ -136,8 +138,7 @@ impl Default for ParallelSparseTrie {
             lower_subtries: [const { LowerSparseSubtrie::Blind(None) }; NUM_LOWER_SUBTRIES],
             prefix_set: PrefixSetMut::default(),
             updates: None,
-            branch_node_tree_masks: HashMap::default(),
-            branch_node_hash_masks: HashMap::default(),
+            branch_node_masks: BranchNodeMasksMap::default(),
             update_actions_buffers: Vec::default(),
             parallelism_thresholds: Default::default(),
             #[cfg(feature = "metrics")]
@@ -187,11 +188,14 @@ impl SparseTrieInterface for ParallelSparseTrie {
 
         // Update the top-level branch node masks. This is simple and can't be done in parallel.
         for ProofTrieNode { path, masks, .. } in &nodes {
-            if let Some(tree_mask) = masks.tree_mask {
-                self.branch_node_tree_masks.insert(*path, tree_mask);
-            }
-            if let Some(hash_mask) = masks.hash_mask {
-                self.branch_node_hash_masks.insert(*path, hash_mask);
+            if masks.tree_mask.is_some() || masks.hash_mask.is_some() {
+                self.branch_node_masks.insert(
+                    *path,
+                    BranchNodeMasks {
+                        tree_mask: masks.tree_mask.unwrap_or_default(),
+                        hash_mask: masks.hash_mask.unwrap_or_default(),
+                    },
+                );
             }
         }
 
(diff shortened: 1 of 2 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2316,6 +2303,9 @@ impl SparseSubtrieInner {
                             // SAFETY: it's a child, so it's never empty
                             let last_child_nibble = child_path.last().unwrap();
 
+                            // Get the branch node masks for this path once (single lookup)
+                            let path_masks = branch_node_masks.get(&path);
```
DaniPopes:
```suggestion
                            let mut path_masks_storage = None;
                            let path_masks = || path_masks_storage.get_or_insert_with(|| branch_node_masks.get(&path));
```

same below

yongkangc: thanks for the note!

yongkangc: basically a lazy pattern so we dont need to call for conditional lookups

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -187,11 +188,14 @@ impl SparseTrieInterface for ParallelSparseTrie {
 
         // Update the top-level branch node masks. This is simple and can't be done in parallel.
         for ProofTrieNode { path, masks, .. } in &nodes {
-            if let Some(tree_mask) = masks.tree_mask {
-                self.branch_node_tree_masks.insert(*path, tree_mask);
-            }
-            if let Some(hash_mask) = masks.hash_mask {
-                self.branch_node_hash_masks.insert(*path, hash_mask);
+            if masks.tree_mask.is_some() || masks.hash_mask.is_some() {
```
yongkangc, line 191: This is okay because Branch nodes in the database are stored as `BranchNodeCompact`, which contains both tree_mask and hash_mask together. When a proof node is constructed, the masks come from this compact representation.

i.e previously 

```
 pub struct TrieMasks {
      pub hash_mask: Option<TrieMask>,
      pub tree_mask: Option<TrieMask>,
  }
```

  This type allows 4 states:
  1. (None, None)
  2. (Some, None)
  3. (None, Some) - cant happen
  4. (Some, Some) - cant happen

However that cant happen because [BranchNodeCompact](https://github.com/alloy-rs/trie/blob/main/src/nodes/branch.rs#L195-L210) (from database) has both masks as non-optional fields:

```
// From alloy_trie - stored in database
struct BranchNodeCompact {
    tree_mask: TrieMask,  // NOT optional
    hash_mask: TrieMask,  // NOT optional
    // ...
}

```

So When constructing TrieMasks:

Branch node: Both masks come from BranchNodeCompact → [always (Some, Some)](https://github.com/alloy-rs/trie/blob/main/src/nodes/branch.rs#L252C1-L286C1)
Non-branch: Uses [TrieMasks::none()](https://github.com/paradigmxyz/reth/blob/main/crates/trie/common/src/trie.rs#L48-L50) → always (None, None)

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2409,19 +2402,17 @@ impl SparseSubtrieInner {
...
-                        // need to remove the node update and add the node itself to the list of
-                        // removed nodes.
-                        update_actions.push(SparseTrieUpdatesAction::InsertRemoved(path));
-                    } else if branch_node_tree_masks.get(&path).is_none_or(|mask| mask.is_empty()) &&
-                        branch_node_hash_masks.get(&path).is_none_or(|mask| mask.is_empty())
-                    {
-                        // If new tree and hash masks are empty, and they were previously empty
-                        // as well, we need to remove the node update.
-                        update_actions.push(SparseTrieUpdatesAction::RemoveUpdated(path));
+                    } else {
```
yongkangc, line 2405: the logic here is equivalent where is bascially just   "Did this path previously have non-empty masks?"

```
  if (previously had non-empty masks) {
      InsertRemoved    // was in DB, now gone → mark as removed
  } else {
      RemoveUpdated    // wasn't in DB anyway → just remove the update
  }

```

crates/trie/sparse-parallel/src/trie.rs
yongkangc: @DaniPopes just addressed this, let me know if there are still concerns u might have!

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -112,10 +112,12 @@ pub struct ParallelSparseTrie {
...
-    /// When a bit is set, the corresponding child subtree is stored in the database.
-    branch_node_tree_masks: HashMap<Nibbles, TrieMask>,
-    /// When a bit is set, the corresponding child is stored as a hash in the database.
-    branch_node_hash_masks: HashMap<Nibbles, TrieMask>,
+    /// Branch node masks containing `tree_mask` and `hash_mask` for each path.
+    /// - `tree_mask`: When a bit is set, the corresponding child subtree is stored in the
+    ///   database.
+    /// - `hash_mask`: When a bit is set, the corresponding child is stored as a hash in the
+    ///   database.
+    branch_node_masks: BranchNodeMasksMap,
```
mattsse, line 120: can't really comment on whether combining those into one is sound

mattsse (commented):
all the logic changes lgtm

would like one more review from someone that can comment on whether this is totally fine to combine

mattsse (approved):
nvm, I just overlooked this type:

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/trie/common/src/trie.rs#L12-L18

so this seems fine

crates/trie/sparse-parallel/src/trie.rs
Rjected: hmm, makes me think whether we should just use `Option<BranchNodeMasks>` instead of `TrieMasks` in `ProofTrieNode` or something similar

crates/trie/sparse-parallel/src/trie.rs
Rjected: we need both masks for every branch node

Rjected (approved):
logic looks the same to me

yongkangc:
Bench compare

 Chain: mainnet
Comparing:
• Baseline: origin/main
• Feature: origin/yk/sparse-trie-masks-consolidation
Blocks: 24117211 to 24119210 (2000 blocks)

Overall Metrics:
• Total newPayload time: 1,171,826ms → 1,164,312ms (4.5% faster)
• Gas/sec: 51.9 → 52.3 Mgas/s (0.65% faster)

Latency Metrics:
• Median (p50): 72.83ms → 68.41ms (6.06% faster)
• p90: 144.63ms → 140.49ms (2.86% faster )
• p99: 220.68ms → 211.66ms (4.08% faster )

Per-Block Change Analysis:
• Mean of change per block: 1.56% faster
• Median of change per block: 3.53% faster

crates/trie/sparse-parallel/src/trie.rs
yongkangc: As discussed, think this is a good idea w little downsides

PR #20669, opened 2025-12-29
perf(engine): parellelize multiproof_targets_from_state

figtracer:
same reasoning as this #20635

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -770,13 +770,19 @@ fn multiproof_targets_from_state(state: EvmState) -> (MultiProofTargets, usize) 
     let results: Vec<_> = state
         .into_par_iter()
         .filter_map(|(address, account)| {
+            // if the account was not touched, or if the account was selfdestructed, do not
+            // fetch proofs for it
+            //
+            // Since selfdestruct can only happen in the same transaction, we can skip
+            // prefetching proofs for selfdestructed accounts
+            //
+            // See: https://eips.ethereum.org/EIPS/eip-6780
             if !account.is_touched() || account.is_selfdestructed() {
                 return None;
             }
 
             let hashed_address = keccak256(address);
 
-            // Pre-allocate storage set
             let mut storage_set =
                 B256Set::with_capacity_and_hasher(account.storage.len(), Default::default());
             storage_set.extend(
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -771,6 +771,13 @@ fn multiproof_targets_from_state(state: EvmState) -> MultiProofTargets {
     state
         .into_par_iter()
         .filter_map(|(address, account)| {
+            // if the account was not touched, or if the account was selfdestructed, do not
+            // fetch proofs for it
+            //
+            // Since selfdestruct can only happen in the same transaction, we can skip
+            // prefetching proofs for selfdestructed accounts
+            //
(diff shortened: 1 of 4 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -766,33 +767,42 @@ where
...
+                    .filter(|(_, slot)| slot.is_changed())
+                    .map(|(key, _)| keccak256(B256::new(key.to_be_bytes()))),
+            );
+
+            Some((hashed_address, storage_set))
+        })
+        .collect();
+
+    let mut targets = MultiProofTargets::default();
+    let mut storage_targets = 0usize;
```
mattsse: isn't this equivalent to smth like `MultiProofTargets::storage_targets_count`?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -766,33 +767,42 @@ where
...
-            continue
-        }
-
-        let mut storage_set =
-            B256Set::with_capacity_and_hasher(account.storage.len(), Default::default());
-        for (key, slot) in account.storage {
-            // do nothing if unchanged
-            if !slot.is_changed() {
-                continue
+    let results: Vec<_> = state
```
mattsse: I think we could collect this directly into `B256Map`? then we don't need to collect into vec then insert into map?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -766,33 +767,42 @@ where
 /// Returns a set of [`MultiProofTargets`] and the total amount of storage targets, based on the
 /// given state.
 fn multiproof_targets_from_state(state: EvmState) -> (MultiProofTargets, usize) {
```
mattsse: I think we can change this to just return MultiProofTargets and then add a helper fn for counting the targets?

mattsse (changes requested):
makes sense, smol suggestions

crates/engine/tree/src/tree/payload_processor/prewarm.rs
figtracer: i think we don't have that available but we can add it

crates/trie/common/src/proofs.rs
```diff
@@ -28,6 +29,15 @@ impl FromIterator<(B256, B256Set)> for MultiProofTargets {
...
 }
 
+impl FromParallelIterator<(B256, B256Set)> for MultiProofTargets {
+    fn from_par_iter<I>(par_iter: I) -> Self
+    where
+        I: IntoParallelIterator<Item = (B256, B256Set)>,
+    {
+        Self(par_iter.into_par_iter().collect())
+    }
+}
```
mattsse, line 41: great

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -765,37 +767,26 @@ where
...
-    let mut targets = MultiProofTargets::with_capacity(state.len());
-    let mut storage_targets = 0;
-    for (addr, account) in state {
-        // if the account was not touched, or if the account was selfdestructed, do not
-        // fetch proofs for it
-        //
-        // Since selfdestruct can only happen in the same transaction, we can skip
-        // prefetching proofs for selfdestructed accounts
-        //
-        // See: https://eips.ethereum.org/EIPS/eip-6780
```
mattsse, line 778: can we keep these docs?

mattsse (changes requested):
smol nit

mattsse (approved):

PR #20670, opened 2025-12-29
chore: use with capacity

mattsse:
also ingored .ck folder for [ck-search](https://github.com/BeaconBay/ck) which was used to find this


Rjected (approved):
nice

PR #20671, opened 2025-12-29
chore(stages): use with_capacity() at populate_range()

figtracer:
uses the provided range of blocks to allocate the vector


mattsse (approved):
this is safe and we can easily do this, ty!

PR #20672, opened 2025-12-29
chore: shrink outgoing broadcast messages

mattsse:
the number of items per peer specific message is dynamic and is buffered internally and also immutable, we can freeze the vecs before emitting it to avoid having over allocated capacity for the time this message is buffered


DaniPopes (approved):

yongkangc (approved):
Ah nice find

PR #20674, opened 2025-12-30
perf(provider): optimize RocksDB history pruning with changeset-based approach

meetrick:
## Summary
Optimizes `StoragesHistory` and `AccountsHistory` pruning in RocksDB by using
MDBX changesets instead of iterating the entire history tables.
- Closes #20417

## Problem
During consistency checks (e.g. after crash recovery), RocksDB history pruning
currently scans the entire `StoragesHistory` / `AccountsHistory` tables to
delete entries above a given block.

On mainnet-scale databases, this results in unnecessary full table scans and
slow startup times.

## Solution
Use MDBX `StorageChangeSets` and `AccountChangeSets` to identify only the
(address, key) pairs that changed in the excess block range, and prune
corresponding history entries using `iter_from` seeks.

If changeset data is unavailable or incomplete, the logic safely falls back
to a full table scan to preserve correctness.

### Key Changes
- Add `iter_from` to `RocksDBProvider` for efficient seek-based iteration
- Prune history entries based on MDBX changesets instead of full scans
- Retain a full-scan fallback for safety
- Add a defensive check to catch missed entries in edge cases

## Testing
- All existing RocksDB invariant tests pass
- Added tests to verify the defensive fallback when changesets are incomplete:
  - `test_storages_history_defensive_check_catches_missed_entries`
  - `test_accounts_history_defensive_check_catches_missed_entries`


gakonst:
@rjected shouldn't this work w static file changesets too?

Rjected:
@meetrick I'm referring to this PR https://github.com/paradigmxyz/reth/pull/18882 which we will merge soon, and we will be rolling it out with the rocksdb storage code. Since it's not merged yet I don't expect this PR to change, but I would expect the direct DB calls to be replaced with methods that use providers once it is merged

Issue #20667, opened 2025-12-29
Avoid event mutex locking


mattsse:
### Describe the feature

we have this event listener:

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/transaction-pool/src/pool/mod.rs#L288-L290

that is only intended for special use cases (e.g. subscribe to events by a specific tx)
but we always obtain a write lock in some paths:

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/transaction-pool/src/pool/mod.rs#L615-L615

we do have is_empty to short circuit, but this still requires obtaining a lock

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/transaction-pool/src/pool/listener.rs#L122-L126

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/transaction-pool/src/pool/mod.rs#L792-L795

since we don't make use of these event channels in any vanilla code, we can further optimize this keep track of whether we have previously added a subscription (e.g. in an atomicbool) then we can check this first before we try to acquire the lock.
the bool is then set the first time a subscription is installed

https://github.com/paradigmxyz/reth/blob/05ed753e58fef9780ef9f74c4e9337c5bb1e2635/crates/transaction-pool/src/pool/mod.rs#L281-L290

@0xKarl98

PR #20678, opened 2025-12-30
feat: avoid mutex locking

0xKarl98:
This closes issue #20667
### ChangeSet

-  Implemented an event-listener fast path in the txpool to avoid acquiring the event_listener write lock when no listeners are present 

- Added an atomic listener-activity flag that is set on first subscription and can be reset when all listeners are gone

- Centralized the repeated lock/check/broadcast/update pattern into a single helper (with_event_listener) to reduce duplication and keep the event emission logic cohesive.

cc @mattsse

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -309,18 +309,18 @@ where
 
     #[inline]
     fn has_event_listeners(&self) -> bool {
-        self.has_event_listeners.load(Ordering::Acquire)
+        self.has_event_listeners.load(Ordering::Relaxed)
     }
 
     #[inline]
     fn mark_event_listener_installed(&self) {
-        self.has_event_listeners.store(true, Ordering::Release);
+        self.has_event_listeners.store(true, Ordering::Relaxed);
     }
 
     #[inline]
     fn update_event_listener_state(&self, listener: &PoolEventBroadcast<T::Transaction>) {
         if listener.is_empty() {
-            self.has_event_listeners.store(false, Ordering::Release);
+            self.has_event_listeners.store(false, Ordering::Relaxed);
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/transaction-pool/src/pool/mod.rs
```diff
@@ -279,14 +290,53 @@ where
...
-        self.event_listener.write().subscribe_all()
+        let mut listener = self.event_listener.write();
+        let events = listener.subscribe_all();
+        self.mark_event_listener_installed();
+        events
+    }
+
+    #[inline]
+    fn has_event_listeners(&self) -> bool {
+        self.has_event_listeners.load(Ordering::Acquire)
```
mattsse: we can use relaxed here, we dont need to strictly enforce any ordering here

crates/transaction-pool/src/pool/mod.rs
```diff
@@ -279,14 +290,53 @@ where
...
+
+    #[inline]
+    fn mark_event_listener_installed(&self) {
+        self.has_event_listeners.store(true, Ordering::Release);
+    }
+
+    #[inline]
+    fn update_event_listener_state(&self, listener: &PoolEventBroadcast<T::Transaction>) {
+        if listener.is_empty() {
+            self.has_event_listeners.store(false, Ordering::Release);
```
mattsse: same here

crates/transaction-pool/src/pool/mod.rs
```diff
@@ -815,7 +850,7 @@ where
             for tx in &discarded {
                 listener.discarded(tx.hash());
             }
-        }
+        });
```
mattsse:
```suggestion
        })
```

mattsse (changes requested):
this is great, only a few tiny suggestions

mattsse (approved):

PR #20679, opened 2025-12-30
perf: pre-alloc removed vec

mattsse:
this will always fill at least `max(1, count - limit)` transactions, so we can pre-alloc this


yongkangc (approved):
Nice's makes sense

PR #20683, opened 2025-12-30
fix(config): call validate() when loading static file config

ethfanWilliam:
validate() was already implemented but never called in from_path(), so invalid blocks_per_file values (like 0) would only blow up at runtime.

now we validate right after parsing.


mattsse (commented):
we dont need to do this here

PR #20684, opened 2025-12-30
chore(engine): make reinsert_reorged_blocks() operate with ref

figtracer:
changed ```reinsert_reorged_blocks()``` to accept a ref instead of owning to prevent ```clone()```2 times inside on_canonical_chain_update()


figtracer:
there's barely any positive change with this actually

PR #20687, opened 2025-12-30
chore: use chain_id() method instead of direct field access in prometheus setup

livehardd:
use the existing `chain_id()` helper instead of directly accessing `self.left().config.chain.chain()`. Matches the pattern used elsewhere in the file and reduces coupling to internal structure.


mattsse (approved):
this is indeed euqivalent

PR #20691, opened 2025-12-31
chore: optimize evm_env if header is available

### LINKED ISSUE #20690: Optimize evm_env if header is available
### Describe the feature

we currently always go through:

https://github.com/paradigmxyz/reth/blob/dc931f56693821b7db37b8b081c94601c2315bcb/crates/rpc/rpc-eth-api/src/helpers/state.rs#L259-L271

for things like:

https://github.com/paradigmxyz/reth/blob/dc931f56693821b7db37b8b081c94601c2315bcb/crates/rpc/rpc-eth-api/src/helpers/trace.rs#L165-L171

but this will always do a redundant header lookup:

https://github.com/paradigmxyz/reth/blob/dc931f56693821b7db37b8b081c94601c2315bcb/crates/rpc/rpc-eth-api/src/helpers/state.rs#L277-L288

what we can do here is introduce another helper fn that is `evm_env_for_header(&SealedHeader`

which basically just forwards to:

https://github.com/paradigmxyz/reth/blob/dc931f56693821b7db37b8b081c94601c2315bcb/crates/rpc/rpc-eth-api/src/helpers/state.rs#L284-L288

@0xKarl98 

### Additional context

_No response_

0xKarl98:
This closes #20690
cc @mattsse

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/state.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/state.rs
@@ -293,11 +293,7 @@ pub trait LoadState:
                     .sealed_header_by_id(at)
                     .map_err(Self::Error::from_eth_err)?
                     .ok_or_else(|| EthApiError::HeaderNotFound(at))?;
-                let evm_env = self
-                    .evm_config()
-                    .evm_env(&header)
-                    .map_err(RethError::other)
-                    .map_err(Self::Error::from_eth_err)?;
+                let evm_env = self.evm_env_for_header(&header)?;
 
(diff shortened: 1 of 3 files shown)


crates/rpc/rpc-eth-api/src/helpers/state.rs
mattsse: let's call the new fn with the header here

mattsse (changes requested):
tiny nit

mattsse (approved):
nice

PR #20694, opened 2026-01-01
refactor: extract runtime shutdown logic into separate function

forkfury:
### Fix
Removes code duplication in runtime shutdown logic.
### Changes
- Extracts shared shutdown_runtime_on_separate_thread function
- Replaces 26 lines of duplicated code with function calls
- Maintains existing behavior


mattsse (commented):
we can keep this as is, not worth unifying

PR #20696, opened 2026-01-01
refactor(rpc): remove unused BlockTransactionsResponseSender

GarmashAlex:
The cache for recovered blocks never enqueued the Right variant of Either, so BlockTransactionsResponseSender and the corresponding match arms were effectively dead code. Simplify BlockLruCache to use a single sender type, remove the unreachable Either::Right branches and the unused type alias, and keep the cache semantics unchanged while reducing complexity


mattsse (approved):
this makes sense because this is indeed unused

PR #20697, opened 2026-01-02
fix(stages): ensure all blocks get changesets in AccountHashingStage::seed

meetrick:
## Summary

Fix a bug in `AccountHashingStage::seed` where changesets were silently truncated when `accounts < blocks`.

## Problem

The `seed` function used `zip` to iterate over blocks and accounts:

```rust
for (t, (addr, acc)) in opts.blocks.zip(&accounts) {
```

Since `zip` stops at the shorter iterator, when `SeedOpts` was configured with fewer accounts than blocks (e.g., `accounts: 10, blocks: 0..=500`), only 10 changesets were generated instead of 501. This caused the incremental hashing mode to query empty changeset ranges, meaning it was never properly tested.

The original `TODO: Rewrite this` comment and the complex documentation explaining how to configure `SeedOpts` to work around this limitation indicated that this was a known issue.

## Solution

Changed `zip(&accounts)` to `zip(accounts.iter().cycle())` so that accounts are reused cyclically to cover all blocks. This ensures that every block in the specified range gets a corresponding changeset entry, regardless of the number of accounts.

## Result

After this fix, tests now generate changesets for all blocks as originally intended. The incremental hashing code path is now properly exercised during testing. The workaround documentation in `SeedOpts` is no longer needed and has been simplified.

### DIFF
--- a/crates/stages/stages/src/stages/hashing_account.rs
+++ b/crates/stages/stages/src/stages/hashing_account.rs
@@ -106,7 +106,8 @@ impl AccountHashingStage {
 
             let mut acc_changeset_cursor =
                 provider.tx_ref().cursor_write::<tables::AccountChangeSets>()?;
-            for (t, (addr, acc)) in opts.blocks.zip(&accounts) {
+            // Use cycle() to ensure all blocks get changesets, cycling through accounts if needed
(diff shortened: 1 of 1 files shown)


mattsse (commented):
since this is just test code, we can keep this as is

PR #20698, opened 2026-01-02
refactor(stages): extract IncrementalSettings struct from MerkleStage

meetrick:
refactor(stages): extract IncrementalSettings struct from MerkleStage

## Problem

The `MerkleStage` enum duplicated the `rebuild_threshold` and `incremental_threshold` fields across both the `Execution` and `Both` variants. This required making the same changes in multiple places when modifying threshold-related behavior, increasing maintenance overhead and the risk of inconsistencies.

## Analysis

Both variants use these thresholds in exactly the same way inside `execute()`: they determine when to rebuild the trie versus apply incremental updates, and they control the chunk size in incremental mode. The `unwind()` path does not reference these values at all. The only behavioral difference between the variants is that `Execution` skips unwind, while `Both` supports both execution and unwind. No behavioral changes are introduced by this refactor, making it safe to consolidate the duplicated fields.

## Solution

Introduce a new `IncrementalSettings` struct that encapsulates both threshold fields and provides a `Default` implementation. The relevant `MerkleStage` variants now store this struct instead of duplicating the fields. This refactor is purely structural and does not change stage semantics or public behavior.

## Changes

- `crates/stages/stages/src/stages/merkle.rs`: added `IncrementalSettings`, updated enum variants, and adjusted match patterns
- `crates/cli/commands/src/stage/dump/merkle.rs`: updated `MerkleStage::Execution` construction
- `crates/stages/stages/benches/criterion.rs`: updated `MerkleStage::Both` construction

Code paths using `new_execution()` or `default_unwind()` required no changes.

## Verification

- All 104 tests in `reth-stages` passed
- Full workspace build succeeded
- Clippy passed with no warnings

### DIFF
--- a/crates/cli/commands/src/stage/dump/merkle.rs
+++ b/crates/cli/commands/src/stage/dump/merkle.rs
@@ -160,11 +160,11 @@ where
     info!(target: "reth::cli", "Executing stage.");
     let provider = output_provider_factory.database_provider_rw()?;
 
-    let mut stage = MerkleStage::Execution {
+    let mut stage = MerkleStage::Execution(reth_stages::stages::IncrementalSettings {
         // Forces updating the root instead of calculating from scratch
         rebuild_threshold: u64::MAX,
         incremental_threshold: u64::MAX,
-    };
+    });
 
     loop {
         let input = reth_stages::ExecInput {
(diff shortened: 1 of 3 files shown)


mattsse (commented):
unsure if we really need this @Rjected

PR #20699, opened 2026-01-02
feat: add sload/sstore metrics

cuiweixie:
remove todo


gakonst (changes requested):
This will almost certainly be a regression perf wide because of non trivial inspector overhead in revm's step - so we can't merge it as is, thx for taking the time and we'll circle back

PR #20701, opened 2026-01-02
perf(era-utils): replace Box<dyn Fn> with function pointer

doocho:
The `decode` function is a pure function that doesn't capture any state, so using `Box<dyn Fn>` was unnecessary overhead.

This change:
- Removes heap allocation from `Box::new(decode)`
- Eliminates dynamic dispatch (vtable lookup) on each iterator call
- Simplifies the code from 5 lines to 1 line

The function pointer `fn(...)` is sufficient since `decode` doesn't need to capture any environment.


mattsse (approved):
seems fine

PR #20703, opened 2026-01-02
feat: make metrics layer configurable

### LINKED ISSUE #20232: Make metrics layer configurable
### Describe the feature

I want to configure buckets, and right now to do so i need to fork reth or uproot all metrics and replace it with different crate.
There is easy workaround for this:
When running `install_prometheus_recorder` pass `PrometheusBuilder`, that should be used in `PrometheusRecorder::install()`.
With this i could run `install_prometheus_recorder` with custom builder, that would have my buckets set up before calling OpNode::run(). 

### Additional context

_No response_

0xKarl98:
closes #20232

### ChangeSet:
 - install_prometheus_recorder() is now a zero‑arg default install backed by OnceLock::get_or_init, calling PrometheusRecorder::install().expect(...).
- Added init_prometheus_recorder(PrometheusRecorder) to explicitly set the global recorder before any installs.
Exposed PrometheusRecorder::install_with_builder for custom builder configuration prior to global injection.
- Removed the one‑shot init hook (set_metrics_init*) and the Option<PrometheusBuilder> path; all call sites now use the default install.
- Kept the LazyLock → OnceLock switch so installation is explicit and customizable without hidden initialization.
cc @mattsse

### DIFF
--- a/crates/node/metrics/src/recorder.rs
+++ b/crates/node/metrics/src/recorder.rs
@@ -13,6 +13,7 @@ use std::sync::{atomic::AtomicBool, Mutex, OnceLock};
 /// Callers must run [`PrometheusRecorder::spawn_upkeep`] manually.
 ///
 /// If a builder is provided, it is registered as the one-time initialization path.
+/// The default install path returns an error if the recorder cannot be installed.
 pub fn install_prometheus_recorder(
     builder: Option<PrometheusBuilder>,
 ) -> eyre::Result<&'static PrometheusRecorder> {
@@ -32,7 +33,7 @@ pub fn install_prometheus_recorder(
         ));
     }
 
-    Ok(PROMETHEUS_RECORDER_HANDLE.get_or_init(|| PrometheusRecorder::install(None).unwrap()))
+    PROMETHEUS_RECORDER_HANDLE.get_or_try_init(|| PrometheusRecorder::install(None))
 }
 
 /// The default Prometheus recorder handle. We use a global static to ensure that it is only
--- a/crates/node/metrics/src/recorder.rs
+++ b/crates/node/metrics/src/recorder.rs
@@ -3,7 +3,7 @@
 use eyre::WrapErr;
 use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
 use metrics_util::layers::{PrefixLayer, Stack};
-use std::sync::{atomic::AtomicBool, Mutex, OnceLock};
+use std::sync::{atomic::AtomicBool, OnceLock};
 
 /// Installs the Prometheus recorder as the global recorder.
 ///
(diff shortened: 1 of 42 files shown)


crates/node/metrics/src/recorder.rs
mattsse, line 1: hmm, 

I wonder if we can make this a bit less invasive, if we keep the logic of

```
pub fn install_prometheus_recorder() -> &'static PrometheusRecorder
```

but obtained from the static oncelock


but this then installs the global recorder like it did before, but we also introduce an `init_promethereus_recorder(PrometheusRecorder)` that sets the global, so that users can manually override this

I think that'd be simpler

crates/node/metrics/src/recorder.rs
```diff
@@ -3,22 +3,119 @@
...
+
+    // No init ran and no recorder is installed, so fall back to the default install path.
+    install_default_recorder()
 }
 
 /// The default Prometheus recorder handle. We use a global static to ensure that it is only
 /// installed once.
-static PROMETHEUS_RECORDER_HANDLE: LazyLock<PrometheusRecorder> =
-    LazyLock::new(|| PrometheusRecorder::install().unwrap());
+static PROMETHEUS_RECORDER_HANDLE: OnceLock<PrometheusRecorder> = OnceLock::new();
```
mattsse: this change we can make, but I'd like to simplify the rest

one global default (install_prometheus_recorder), that does get or init with PrometheusRecorder::install().unwrap()

and another fn that accepts a PrometheusRecorder that sets the global

mattsse (changes requested):
cool, I think we can simplify this a bit more on the recorder level

and then also get rid of the optional arg, if the expectation is that the user will manually set the global recorder

crates/node/metrics/src/recorder.rs
```diff
@@ -11,27 +11,31 @@ use std::sync::{atomic::AtomicBool, OnceLock};
...
-/// Returns an error if a recorder has already been installed.
-pub fn try_install_prometheus_recorder_with_builder(
-    builder: PrometheusBuilder,
-) -> eyre::Result<&'static PrometheusRecorder> {
-    let recorder = PrometheusRecorder::install_with_builder(builder)?;
-    PROMETHEUS_RECORDER_HANDLE
-        .set(recorder)
-        .map_err(|_| eyre::eyre!("Prometheus recorder already installed"))?;
-    Ok(PROMETHEUS_RECORDER_HANDLE.get().expect("recorder is set"))
-}
```
mattsse, line 33: can we keep this fn as well, I think wouldn't hurt to keep this even tho this can now be done with the new init_prometheus_recorder as well

mattsse (changes requested):

crates/node/metrics/src/recorder.rs
0xKarl98: Yeah i have thought about the same , can definitately keep it

Was trying to reduce redundancy before

mattsse (approved):

mattsse:
I believe this should do it

PR #20705, opened 2026-01-02
feat(cli): validate storage settings at startup

yongkangc:
Add validation to error when CLI storage flags differ from persisted settings, instead of silently ignoring user input.

Closes #20482

Users changing storage flags (e.g., `--static-files.receipts`) after genesis get no feedback - the node silently uses stored settings.

**Solution**

Compare CLI settings against stored settings at startup. On mismatch, return actionable error:

```
Storage settings mismatch!

Stored in DB: { receipts_in_static_files: false, ... }
From CLI:     { receipts_in_static_files: true, ... }

Storage flags cannot be changed after genesis.
Either remove the conflicting CLI flags, or delete the database and re-sync.
```

- Validation in CLI layer, not storage layer (keeps `ProviderFactory` simple)
- Only compares CLI-configurable fields via `static_file_settings_eq()` - RocksDB flags are ignored since they're not exposed in CLI

### DIFF
--- a/IMPLEMENTATION_PLAN.md
+++ b/IMPLEMENTATION_PLAN.md
@@ -1,0 +1,144 @@
+# Issue #20482: Add Configuration Validation for RocksDB Flags at Startup
+
+## Problem
+
+Reth uses `StorageSettings` to control where data is stored (MDBX vs RocksDB vs static files). These flags are:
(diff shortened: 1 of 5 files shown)


yongkangc:
blocked by #20393

PR #20707, opened 2026-01-02
refactor(trie): replace TrieMasks with Option<BranchNodeMasks>

yongkangc:
Follow up to  https://github.com/paradigmxyz/reth/pull/20659

**Problem**
The `TrieMasks` struct allows invalid states where one mask exists without the other.

**Solution**
Replace `TrieMasks` with `Option<BranchNodeMasks>` throughout the codebase.

This prevents invalid states and simplifies mask handling. also as a bonus some memory savings: 

Memory savings:
- TrieMasks: 2 × Option<TrieMask> = 2 × 4 bytes = 8 bytes
- Option<BranchNodeMasks>: 1 byte discriminant + 2 × 2 bytes = ~6 bytes
So ~2 bytes saved per mask instance.

invariant: BranchNodeCompact always has both tree_mask and hash_mask together - they are never used independently. This justifies the refactor.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/configured_sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/configured_sparse_trie.rs
@@ -1,7 +1,7 @@
 //! Configured sparse trie enum for switching between serial and parallel implementations.
 
 use alloy_primitives::B256;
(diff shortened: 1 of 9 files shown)


mediocregopher (approved):
Nice, LGTM

PR #20710, opened 2026-01-02
feat(cli): make Cli extensible with custom subcommands

Zygimantass:
Add a generic `SubCmd` parameter to `Cli` and `Commands` that allows consumers to add their own subcommands to the reth CLI.

- Add `SubCmd` generic parameter with `NoSubCmd` as default
- Add `ExtendedCommand` trait for executing custom subcommands
- Add `Commands::Ext` variant using `#[command(flatten)]`
- Export `ExtendedCommand` and `NoSubCmd` from lib.rs
- Add test demonstrating extensible subcommand usage

Example consumer PR: https://github.com/tempoxyz/tempo/pull/1839

Amp-Thread-ID: https://ampcode.com/threads/T-019b7ef3-7b08-736f-b149-b93fd7b7ee1d

### DIFF
--- a/crates/ethereum/cli/src/app.rs
+++ b/crates/ethereum/cli/src/app.rs
@@ -1,4 +1,8 @@
-use crate::{interface::Commands, Cli};
+use crate::{
+    interface::{Commands, NoSubCmd},
+    Cli,
+};
+use clap::Subcommand;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
makes sense

PR #20711, opened 2026-01-02
refactor: optimize check whether all blobs ready

cuiweixie:


mattsse (approved):
makes sense

crates/transaction-pool/src/blobstore/disk.rs
```diff
@@ -75,20 +75,24 @@ impl DiskFileBlobStore {
...
-
+        let mut missing_count = result.len();
         // first scan all cached full sidecars
         for (_tx_hash, blob_sidecar) in self.inner.blob_cache.lock().iter() {
             if let Some(blob_sidecar) = blob_sidecar.as_eip7594() {
                 for (hash_idx, match_result) in
                     blob_sidecar.match_versioned_hashes(versioned_hashes)
                 {
                     result[hash_idx] = Some(match_result);
+                    missing_count -= 1;
```
mattsse, line 86: this will work, even if the user passed in duplicated hashes

mattsse (approved):

PR #20712, opened 2026-01-02
feat: simplify find logic

cuiweixie:
avoid vector alloc that using filter collect.


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1467,19 +1467,22 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> ReceiptProvider for DatabasePr
             return Ok(Vec::new());
         }
 
-        // find blocks with transactions to determine transaction range
-        let non_empty_blocks: Vec<_> =
-            block_body_indices.iter().filter(|indices| indices.tx_count > 0).collect();
```
mattsse, line 1472: I don't mind this additional collect, and I find this easier to parse than the change, since this isn't a hot fn and the index type is rather small, I'd rather keep it as is

but wdyt @joshieDo

joshieDo:
agree with @mattsse

PR #20714, opened 2026-01-02
refactor(trie): consolidate StorageTries parallel maps into one

figtracer:
- created new struct `StorageTrieEntry` for better cache locality
- updated all calling sites to use this new struct

this makes storage related lookups in the sparse trie 2x faster and we also save 48 bytes per `StorageTrie` instance

benchmark results:

| Op        | Size | Old (two maps) | New (single map) | Speedup   |
| ----------------- | ---- | -------------- | ---------------- | --------- |
| **Lookup**        | 100  | 612 ns         | 280 ns           | **2.19x** |
| **Lookup**        | 1000 | 7.35 µs        | 3.31 µs          | **2.22x** |
| **Lookup**        | 5000 | 35.6 µs        | 15.6 µs          | **2.28x** |
| **Clear**         | 500  | 14.2 µs        | 12.5 µs          | **1.14x** |
| **Insert**        | 5000 | 1.54 ms        | 1.51 ms          | **1.02x** |

### DIFF
--- a/crates/trie/sparse/src/state.rs
+++ b/crates/trie/sparse/src/state.rs
@@ -814,19 +814,46 @@ where
     }
 }
 
+/// Entry containing both the sparse trie and its revealed paths for a storage account.
+#[derive(Debug)]
+struct StorageTrieEntry<S> {
+    /// The sparse trie for this storage account.
+    trie: SparseTrie<S>,
+    /// Collection of revealed trie node paths.
+    revealed_paths: HashSet<Nibbles>,
+}
+
+impl<S: Default> Default for StorageTrieEntry<S> {
+    fn default() -> Self {
+        Self { trie: SparseTrie::default(), revealed_paths: HashSet::default() }
+    }
+}
+
+impl<S: SparseTrieInterface> StorageTrieEntry<S> {
+    /// Clears the entry, returning it for reuse.
+    fn clear(mut self) -> Self {
+        self.trie = self.trie.clear();
+        self.revealed_paths.clear();
+        self
+    }
+}
+
+impl<S: SparseTrieInterface + Clone> StorageTrieEntry<S> {
+    /// Creates a new entry from a default trie template.
+    fn new_from_default(default_trie: &SparseTrie<S>) -> Self {
+        Self { trie: default_trie.clone(), revealed_paths: HashSet::default() }
+    }
+}
+
 /// The fields of [`SparseStateTrie`] related to storage tries. This is kept separate from the rest
 /// of [`SparseStateTrie`] both to help enforce allocation re-use and to allow us to implement
 /// methods like `get_trie_and_revealed_paths` which return multiple mutable borrows.
 #[derive(Debug, Default)]
 struct StorageTries<S = SerialSparseTrie> {
-    /// Sparse storage tries.
-    tries: B256Map<SparseTrie<S>>,
-    /// Cleared storage tries, kept for re-use.
-    cleared_tries: Vec<SparseTrie<S>>,
-    /// Collection of revealed storage trie paths, per account.
-    revealed_paths: B256Map<HashSet<Nibbles>>,
-    /// Cleared revealed storage trie path collections, kept for re-use.
-    cleared_revealed_paths: Vec<HashSet<Nibbles>>,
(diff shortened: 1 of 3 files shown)


crates/trie/sparse/src/state.rs
```diff
@@ -205,22 +211,22 @@ where
...
     /// Returns mutable reference to storage sparse trie if it was revealed.
     pub fn storage_trie_mut(&mut self, address: &B256) -> Option<&mut S> {
-        self.storage.tries.get_mut(address).and_then(|e| e.as_revealed_mut())
+        self.storage.entries.get_mut(address).and_then(|e| e.trie.as_revealed_mut())
     }
 
     /// Takes the storage trie for the provided address.
     pub fn take_storage_trie(&mut self, address: &B256) -> Option<SparseTrie<S>> {
-        self.storage.tries.remove(address)
+        self.storage.entries.remove(address).map(|e| e.trie)
```
yongkangc: Difference:
- Old: Only removes trie, revealed_paths for this account persists
- New: Removes both trie and revealed_paths
This is a behavior change. If something called take_storage_trie and expected revealed_paths to still be there, it would break.

not 100% sure about this, defer to @mediocregopher or @shekhirin

crates/trie/sparse/src/state.rs
```diff
@@ -835,83 +866,67 @@ impl<S: SparseTrieInterface> StorageTries<S> {
...
-            self.cleared_tries.pop().unwrap_or_else(|| self.default_trie.clone())
-        });
-
-        let revealed_paths = self
-            .revealed_paths
-            .entry(account)
-            .or_insert_with(|| self.cleared_revealed_paths.pop().unwrap_or_default());
-
-        (trie, revealed_paths)
+        let entry = self.get_entry_mut(account);
```
yongkangc, line 914: Difference in creation logic:
- Old: For a new account, tries to pop from cleared_tries, else clones default_trie. Separately, tries to pop from `cleared_revealed_paths`, else creates empty HashSet.
- New: Tries to pop from cleared_entries (which has BOTH trie and paths), else creates new entry with cloned `default_trie` and empty HashSet.

Potential issue: In the old code, cleared_tries and cleared_revealed_paths were independent pools. we could have 10 cleared tries and 5 cleared paths. In the new code, they're always paired in `cleared_entries`.

yongkangc (commented):
this is promising, however regarding the trie logic would defer to @mediocregopher or @shekhirin to take a closer look at it

mattsse (commented):
@shekhirin is this similar to your arena wip?

crates/trie/sparse/src/state.rs
```diff
@@ -835,83 +866,67 @@ impl<S: SparseTrieInterface> StorageTries<S> {
...
+            self.cleared_entries
+                .pop()
+                .unwrap_or_else(|| StorageTrieEntry::new_from_default(&self.default_trie))
+        })
     }
 
     /// Returns the `SparseTrie` and the set of already revealed trie node paths for an account's
     /// storage, creating them if they didn't previously exist.
+    ///
+    /// Uses a single `HashMap` lookup instead of two separate lookups.
```
mediocregopher: Unnecessary comment
```suggestion
```

crates/trie/sparse/src/state.rs
mediocregopher: Imo this should be fine, I don't believe there was a situation where the length of revealed paths and tries was different in the old code.

crates/trie/sparse/src/state.rs
mediocregopher: Good catch, this will indeed break other code. This method is used in two places, and they both do the same thing:

* Call `take_storage_trie` on an account
* Modify the storage trie
* Put the account's storage trie back using `insert_storage_trie`

Given this change the first step will discard the revealed_paths, such that on the third step the trie will appear to have no revealed paths.

I think the best thing to do would be to change these methods to `take_storage_trie_entry` and `insert_storage_trie_entry`. It will result in a larger PR but I think it's the cleanest solution.

mediocregopher (changes requested):

mediocregopher (approved):

PR #20715, opened 2026-01-02
fix: history index stages incorrectly detect first sync with non-zero…

livehardd:
The history index stages were hardcoding block 0 to check for first sync, which fails when genesis starts at a non-zero block. This prevented the first-sync optimization (clearing tables and rebuilding from scratch) from running.

Changed to check if the history table is empty instead, and use the actual checkpoint block number as the starting point rather than assuming 0.

### DIFF
--- a/crates/stages/stages/src/stages/index_account_history.rs
+++ b/crates/stages/stages/src/stages/index_account_history.rs
@@ -2,7 +2,11 @@ use super::{collect_history_indices, load_history_indices};
 use alloy_primitives::Address;
 use reth_config::config::{EtlConfig, IndexHistoryConfig};
 use reth_db_api::{
-    cursor::DbCursorRO, models::ShardedKey, table::Decode, tables, transaction::{DbTx, DbTxMut},
+    cursor::DbCursorRO,
+    models::ShardedKey,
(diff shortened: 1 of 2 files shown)


crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -91,13 +97,16 @@ where
         }
 
         let mut range = input.next_block_range();
-        let first_sync = input.checkpoint().block_number == 0;
+        // Check if the history table is empty to determine if this is the first sync.
+        // We cannot use checkpoint == 0 because genesis block number might not be 0.
+        let first_sync =
+            provider.tx_ref().cursor_read::<tables::AccountsHistory>()?.last()?.is_none();
```
mattsse, line 103: we shouldnt use raw cursors here because this will complicate things with static files I think.

also I dont think there's a valid reason for this, even if the node is initialized with block > 0 dont think we insert history

mattsse (commented):
wdyt @Rjected @joshieDo

livehardd:
@mattsse hi  mattsse, the bot closed this would you mind re-opening this?

PR #20717, opened 2026-01-02
perf: avoid one vec alloc

cuiweixie:


crates/chain-state/src/memory_overlay.rs
```diff
@@ -49,11 +49,14 @@ impl<'a, N: NodePrimitives> MemoryOverlayStateProviderRef<'a, N> {
...
-            )
+            let mut input = TrieInput::default();
+            // Iterate from oldest to newest
+            for block in self.in_memory.iter().rev() {
+                let data = block.trie_data();
+                input.nodes.extend_from_sorted(&data.trie_updates);
+                input.state.extend_from_sorted(&data.hashed_state);
+            }
+            input
         })
```
mattsse, line 60: I believe this is the same:

https://github.com/paradigmxyz/reth/blob/e8cc91ebc2fdfb4befdbc468b0a6af74cacdf123/crates/trie/common/src/input.rs#L49-L61

because already sorted

we can't use from_blocks_sorted here because that only accepts a & type, maybe we can add another helper for this

mattsse (approved):
lgtm, pending @yongkangc

yongkangc (approved):

PR #20719, opened 2026-01-03
perf: remove unnecessary code

cuiweixie:
1. num_real_targets >= len of vec always true, so  take is not need.  
2. sort is call below, so sorted is not needed.


mattsse (commented):
I believe this checks out, but deferring to @mediocregopher because this touches existing test code

mediocregopher (approved):
Good catch 🙏

PR #20723, opened 2026-01-03
perf: using kth algorithm to get suggest gas price

cuiweixie:


mattsse (commented):
I'd like to keep this as is because this is easier to understand

PR #20725, opened 2026-01-03
docs: update eth-wire TransactionSigned struct definition

andrewshab3:
Update the documentation to reflect the current implementation.


mattsse (commented):
this type is besically deprecated

PR #20728, opened 2026-01-03
docs: fix doc comment errors

crypomen9:
Clarifies broadcast channel return value descriptions in service, improves forkchoice update return value description in consensus.


mattsse (approved):
fine

PR #20729, opened 2026-01-03
fix: inclusive range off-by-one

cuiweixie:


crates/rpc/rpc/src/trace.rs
mattsse, line 387: ah, right this is a bug because we'd process overlapping ranges

mattsse (approved):

PR #20731, opened 2026-01-03
fix: progress should not be finished

cuiweixie:


mattsse (approved):
I believe this is correct

pending @joshieDo

crates/prune/prune/src/pruner.rs
```diff
@@ -191,6 +191,8 @@ where
 
         for segment in &self.segments {
             if limiter.is_limit_reached() {
+                output.progress =
+                    PruneProgress::HasMoreData(PruneInterruptReason::DeletedEntriesLimitReached);
```
joshieDo, line 195: i think this might be unnecessary/wrong since it should be covered by eacher prune segment and assigned here

https://github.com/paradigmxyz/reth/blob/790a73cd2af02849bba85bd48b1e5fe90b2292fe/crates/prune/prune/src/pruner.rs#L250-L261

PR #20733, opened 2026-01-03
docs(evm): add documentation for ExecutableTxIterator trait

endorina:
Adds comprehensive documentation for the ExecutableTxIterator trait,
explaining its role as a marker trait that combines ExecutableTxTuple
with EVM-specific transaction constraints. Documents its usage in
parallel transaction processing and provides cross-references to
related types.


mattsse (commented):
we can keep this as is

PR #20739, opened 2026-01-05
perf(trie): use mem::take instead of clone for account RLP buffer

yongkangc:
**Summary**
Replace `self.account_rlp_buf.clone()` with `core::mem::take()` to avoid allocation and memcpy when updating account leaves.

**Performance Impact**
Eliminates one allocation + memcpy per account update. The buffer is re-used on the next encode operation (starts empty, grows as needed).

Trade-offs
this pattern loses the pre-allocated capacity after each `mem::take()`, so subsequent encodes may need to grow the buffer. However, CPU savings from avoiding the copy typically outweigh the small reallocation cost for typical account RLP sizes.

### DIFF
--- a/crates/trie/sparse/src/state.rs
+++ b/crates/trie/sparse/src/state.rs
@@ -728,7 +728,8 @@ where
         let nibbles = Nibbles::unpack(address);
         self.account_rlp_buf.clear();
         account.into_trie_account(storage_root).encode(&mut self.account_rlp_buf);
-        self.update_account_leaf(nibbles, self.account_rlp_buf.clone(), provider_factory)?;
+        let rlp_encoded_account = core::mem::take(&mut self.account_rlp_buf);
+        self.update_account_leaf(nibbles, rlp_encoded_account, provider_factory)?;
 
         Ok(true)
     }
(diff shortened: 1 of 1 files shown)


Rjected (commented):
I am not convinced this saves any allocation, I think it just moves it to encoding-time? Conceptually, we need to allocate these anyways:
* each `value` in the `values` map, and
* some capacity for encoding

And right now we rarely re-allocate during encoding

crates/trie/sparse/src/state.rs
```diff
@@ -728,7 +728,8 @@ where
         let nibbles = Nibbles::unpack(address);
         self.account_rlp_buf.clear();
         account.into_trie_account(storage_root).encode(&mut self.account_rlp_buf);
-        self.update_account_leaf(nibbles, self.account_rlp_buf.clone(), provider_factory)?;
+        let rlp_encoded_account = core::mem::take(&mut self.account_rlp_buf);
```
mattsse, line 731: > subsequent encodes may need to grow the buffer.

I think this is now equivalent to simply  creating a new Vec?

do we then even need the buf at all?

PR #20743, opened 2026-01-05
docs: fix discv5 multiaddr peer id conversion comment

kurahin:
The documentation for discv4_id_to_multiaddr_id incorrectly stated that the function converts a PeerId to reth_network_peers::PeerId. In reality, it converts our PeerId into discv5::libp2p_identity::PeerId. This updates the doc comment to reference the correct target type so that the documentation matches the actual behavior and does not mislead callers.


mattsse (approved):
fine

PR #20753, opened 2026-01-05
perf: preallocate capacity in block body processing

livehardd:
Adds capacity hint to avoid reallocation during iteration. small change but this runs on every block.

### DIFF
--- a/crates/primitives-traits/src/block/body.rs
+++ b/crates/primitives-traits/src/block/body.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 use alloc::{fmt, vec::Vec};
 use alloy_consensus::{
-    transaction::{Recovered, TxHashRef},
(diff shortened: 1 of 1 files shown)


crates/primitives-traits/src/block/body.rs
```diff
@@ -190,13 +190,12 @@ pub trait BlockBody:
...
-        self.recover_signers().map(|signers| {
-            self.transactions()
-                .iter()
-                .zip(signers)
-                .map(|(tx, signer)| tx.clone().with_signer(signer))
-                .collect()
-        })
+        let mut recovered = Vec::with_capacity(self.transactions().len());
+        for tx in self.transactions() {
+            let signer = tx.recover_signer()?;
```
mattsse, line 195: this change isn't needed and even regresses perf because this now performs recovery sequentially and also both zip inputs are exactsizeiter so collect only allocs self.transactions().len()

PR #20755, opened 2026-01-05
fix(engine): do not create another cache for multiproof task

shekhirin:
### Problem

Every time `cache_for` is called, we check that it's going to be the only cache in use: https://github.com/paradigmxyz/reth/blob/790a73cd2af02849bba85bd48b1e5fe90b2292fe/crates/engine/tree/src/tree/payload_processor/mod.rs#L786-L788

First `cache_for` call is happening here https://github.com/paradigmxyz/reth/blob/fa05d19f1b225259500b49f46468f314a1070855/crates/engine/tree/src/tree/payload_processor/mod.rs#L252

Second is here https://github.com/paradigmxyz/reth/blob/fa05d19f1b225259500b49f46468f314a1070855/crates/engine/tree/src/tree/payload_processor/mod.rs#L297

Which leads to always creating a new cache on the second call

### Solution

Get cache once with `cache_for` and pass to all consumers.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -3,8 +3,8 @@
 use super::precompile_cache::PrecompileCacheMap;
 use crate::tree::{
     cached_state::{
-        CachedStateMetrics, ExecutionCache as StateExecutionCache, ExecutionCacheBuilder,
-        SavedCache,
+        CachedStateMetrics, CachedStateProvider, ExecutionCache as StateExecutionCache,
+        ExecutionCacheBuilder, SavedCache,
     },
     payload_processor::{
         prewarm::{PrewarmCacheTask, PrewarmContext, PrewarmMode, PrewarmTaskEvent},
@@ -30,7 +30,9 @@ use reth_evm::{
 };
 use reth_execution_types::ExecutionOutcome;
 use reth_primitives_traits::NodePrimitives;
-use reth_provider::{BlockReader, DatabaseProviderROFactory, StateProviderFactory, StateReader};
+use reth_provider::{
+    BlockReader, DatabaseProviderROFactory, StateProvider, StateProviderFactory, StateReader,
+};
 use reth_revm::{db::BundleState, state::EvmState};
 use reth_trie::{hashed_cursor::HashedCursorFactory, trie_cursor::TrieCursorFactory};
 use reth_trie_parallel::{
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -651,12 +649,12 @@ impl<Tx, Err, R: Send + Sync + 'static> PayloadHandle<Tx, Err, R> {
...
     /// Returns a clone of the caches used by prewarming
     pub(super) fn caches(&self) -> Option<StateExecutionCache> {
-        self.prewarm_handle.cache.clone()
+        self.prewarm_handle.saved_cache.clone().map(|cache| cache.cache().clone())
     }
 
     /// Returns a clone of the cache metrics used by prewarming
     pub(super) fn cache_metrics(&self) -> Option<CachedStateMetrics> {
-        self.prewarm_handle.cache_metrics.clone()
+        self.prewarm_handle.saved_cache.clone().map(|cache| cache.metrics().clone())
```
mattsse: any reason why this does option clone instead of as_ref + clone?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -651,12 +649,12 @@ impl<Tx, Err, R: Send + Sync + 'static> PayloadHandle<Tx, Err, R> {
 
     /// Returns a clone of the caches used by prewarming
     pub(super) fn caches(&self) -> Option<StateExecutionCache> {
-        self.prewarm_handle.cache.clone()
+        self.prewarm_handle.saved_cache.clone().map(|cache| cache.cache().clone())
```
mattsse: same q

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -294,14 +295,18 @@ where
         // spawn multi-proof task
         let parent_span = span.clone();
         self.executor.spawn_blocking({
-            let saved_cache = self.cache_for(parent_hash);
-            let cache = saved_cache.cache().clone();
-            let cache_metrics = saved_cache.metrics().clone();
+            let saved_cache = prewarm_handle.saved_cache.clone();
```
mattsse: I kinda dislike this pattern of using a scope that returns a closure, especially because we already clone the parent_span one line above, so we have the scope for just the saved_cache clone

can we also move this outside so that we  only have `move` 

I find this a lot easier to read

mattsse (changes requested):

mattsse (approved):

PR #20757, opened 2026-01-05
fix(cli): rely on Config::from_path for missing config files

GarmashAlex:
reth config now relies on the documented behavior of Config::from_path to create a configuration file with default values when the specified path does not exist. This removes the manual path.exists() check that prevented the NotFound branch in from_path from running and made reth config --config <FILE> inconsistent with other commands such as node and p2p, which already trust from_path to handle missing files.


mattsse (commented):
we can keep this here because config command should show you the correct config and not alter the disk

PR #20758, opened 2026-01-05
refactor(ethereum): cache RLP lengths in ethereum payload builder

VolodymyrBg:
default_ethereum_payload was calling tx.inner().length() twice per transaction and attributes.withdrawals().length() on every iteration of the builder loop. Both values are pure and do not change during block construction, so the repeated calls only add overhead on a hot path. This change caches the RLP length of withdrawals once before the loop and the RLP length of each transaction once per iteration, reusing these values for the Osaka block-size estimation and the running transaction RLP length accumulator, without changing behavior.


mattsse (approved):
makes sense

PR #20759, opened 2026-01-05
docs: fix outdated references in static-file README

andrewshab3:
The static-file README.md contains several outdated references to source code:

  1. ``StaticFileProvider`` link points to line 44, but the struct is defined at line 86
  2. ``StaticFileJarProvider`` link points to line 42, but the struct is defined at line 29
  3. ``Query Diagrams`` section references non-existent `Provider` at line 41, should reference `ProviderFactory` at line 62


mattsse (commented):
we can keep this as is because this would become outdated soon anyway

PR #20761, opened 2026-01-05
docs: update ReverseHeadersDownloader structure

letmehateu:
Updated the documentation to reflect the current ReverseHeadersDownloader struct


mattsse (commented):
not worth updating

PR #20762, opened 2026-01-05
docs: update BodiesDownloader structure

letmehateu:
Updated documentation to reflect the current BodiesDownloader structure


mattsse (commented):
not worth updating

PR #20763, opened 2026-01-05
docs: update poll_action structure

letmehateu:
Updated documentation to reflect the current `poll_action` structure


mattsse (commented):
not worth updating this anymore, these docs arent that useful

PR #20768, opened 2026-01-06
perf: handle RPC errors instead of panicking

ethfanWilliam:
These were in the benchmark initialization path and would panic on network issues or missing blocks. Since the function returns Result anyway, now it returns clear error messages instead.


mattsse (approved):
this is better

needs fmt

mattsse (approved):

PR #20770, opened 2026-01-06
chore: remove outdated network TODO

crypomen9:
Removes outdated TODO comment about spawning txpool in NetworkConfig::start_network(), the txpool field is unit type and intentionally not spawned in this simplified setup.


mattsse (commented):
we can keep this to highlight that this should be spawned as well

PR #20774, opened 2026-01-06
perf(trie): reuse overlay in deferred trie overlay computation

yongkangc:
Eliminate O(N²) complexity in `sort_and_build_trie_input()` by reusing the parent block's already-merged trie overlay instead of re-merging all ancestors for every block.

**Problem**

Previously, each block iterated through ALL ancestors calling `wait_cloned()` on each to merge their state into the overlay. For block N, this meant N-1 `wait_cloned()` calls, leading to O(N²) total calls for N blocks.

This became a severe bottleneck under load when blocks arrive faster than the async task can complete (sync fallback path).

**Solution**

Fast path (O(1)): When parent's `anchor_hash` matches, clone its overlay which already contains all ancestors merged. Since `TrieInputSorted` uses `Arc` internally, cloning is O(1).

Slow path (O(N)): When anchor changes (after persist/reorg), rebuild from all ancestors. This only happens at persist boundaries, not per block.

Benchmark: https://github.com/paradigmxyz/reth/pull/20774#issuecomment-3738494912


crates/chain-state/src/deferred_trie.rs
```diff
@@ -190,6 +236,31 @@ impl DeferredTrieData {
...
 
+    /// Merge all ancestors into a single overlay.
+    ///
+    /// This is a rare fallback path, only used when no ancestor has a cached
+    /// `anchored_trie_input` (e.g., blocks created via alternative constructors).
+    /// In normal operation, the parent always has a cached overlay and this
+    /// function is never called.
+    ///
+    /// Iterates ancestors oldest -> newest so later state takes precedence.
+    fn merge_ancestors_into_overlay(ancestors: &[Self]) -> TrieInputSorted {
```
yongkangc: previous implementation, case 2

crates/chain-state/src/deferred_trie.rs
```diff
@@ -164,9 +172,71 @@ impl DeferredTrieData {
...
+                    if !sorted_hashed_state.is_empty() {
+                        Arc::make_mut(&mut overlay.state).extend_ref(&sorted_hashed_state);
+                    }
+                    if !sorted_trie_updates.is_empty() {
+                        Arc::make_mut(&mut overlay.nodes).extend_ref(&sorted_trie_updates);
+                    }
+                    overlay
+                }
+                // Case 2: Parent exists but anchor mismatch or no cached overlay.
+                // We must rebuild from the ancestors list (which only contains unpersisted blocks).
```
yongkangc, line 203: This is the current approach

mediocregopher (approved):

yongkangc:
### 500 MS
image


```
=== BENCHMARK COMPARISON SUMMARY ===
Timestamp: 2026-01-12 UTC
Baseline: origin/main
Feature:  origin/fix/deferred-trie-on2-complexity

Performance Changes:
  NewPayload Latency per-block mean change:   +0.32%
  NewPayload Latency per-block median change: -3.66%
  NewPayload Latency per-block std dev:       33.57%
  Total newPayload time change:               +0.02%
  NewPayload Latency p50:                     -1.14%
  NewPayload Latency p90:                     -1.90%
  NewPayload Latency p99:                     +3.13%
```


### 0 ms

PR #20777, opened 2026-01-06
feat: support additional eth_subscribe handlers

0xKarl98:
This closes #20750
### Changes 
- Parse the subscription `kind` and raw params at the callsite: if the `kind` maps to a known `SubscriptionKind` (newHeads/logs/newPendingTransactions/syncing), we validate and normalize params upfront and then invoke the standard handlers. 
- If the `kind` is unknown, we try a per‑kind custom handler (registered via `register_handler`) first, then fall back to a global `fallback_handler` if configured; otherwise we return `invalid_params`.
   - For known kinds, invalid params are rejected immediately with `invalid_params` to preserve strict RPC semantics and avoid masking client errors. Custom subscriptions are only accepted when explicitly registered (per‑kind handler or global fallback), and they must provide their own subscription stream; otherwise the request is rejected.

-  Invalid paths are explicitly:
    - Known kind + invalid params (e.g., `logs` expects `Filter`/none; `newPendingTransactions` expects `bool`/none; deserialization/type mismatch) → `invalid_params`.
   - Unknown kind + no registered handler/fallback → `invalid_params`.
   - Handler/fallback returns `Err(ErrorObject)` → forwarded as‑is (e.g., `invalid_params` or other error).

cc @mattsse


crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -289,6 +386,8 @@ struct EthPubSubInner<EthApi> {
     eth_api: EthApi,
     /// The type that's used to spawn subscription tasks.
     subscription_task_spawner: Box<dyn TaskSpawner>,
+    /// Fallback handler for unknown subscription kinds or params.
+    fallback_handler: Option<FallbackHandler>,
```
mattsse, line 436: we need a way to inject additional handlers in here, so I think this could be like an `Arc<Rwlock<HashMap<String, Handler>` for example

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -90,27 +126,59 @@ where
     pub async fn handle_accepted(
         &self,
         accepted_sink: SubscriptionSink,
-        kind: SubscriptionKind,
-        params: Option<Params>,
+        kind: String,
+        raw_params: Option<Box<JsonRawValue>>,
```
mattsse, line 140: I'd like to keep this fn as is

and instead handle parsing this on the callsite

like if kind deserializes into SubscriptionKind then parse into `Option<Params>` and invoke this function, otherwise check the fallback

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -50,7 +58,35 @@ impl<Eth> EthPubSub<Eth> {
...
+    /// params.
+    pub fn with_spawner_and_fallback_handler<F>(
+        eth_api: Eth,
+        subscription_task_spawner: Box<dyn TaskSpawner>,
+        fallback_handler: F,
+    ) -> Self
+    where
+        F: Fn(String, Option<Box<JsonRawValue>>) -> Result<PubSubStream, ErrorObject<'static>>
+            + Send
+            + Sync
```
mattsse, line 68: we can keep all of this as is, and instead add smth like `register_handler()` that inserts a new handler

this is similar to the `Methods` type

```
pub struct Methods {
	callbacks: Arc<FxHashMap<&'static str, MethodCallback>>,
```

https://github.com/paradigmxyz/reth/blob/485eb2e8d544080e61e33d2b3fa1ee41245ec6ad/crates/rpc/rpc-builder/src/lib.rs#L1755-L1760

basically, we want a way to inject additional handlers after the fact, like:

https://github.com/paradigmxyz/reth/blob/485eb2e8d544080e61e33d2b3fa1ee41245ec6ad/examples/exex-subscription/src/main.rs#L172-L172

mattsse (changes requested):
great start

crates/rpc/rpc/src/eth/pubsub.rs
0xKarl98: makes sense

crates/rpc/rpc/src/eth/pubsub.rs
0xKarl98: Yeah should do it

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -38,6 +40,36 @@ pub struct EthPubSub<Eth> {
...
+    params: Option<Params>,
+}
+
+#[derive(Debug)]
+enum ParseSubscriptionError {
+    UnsupportedKind,
+    InvalidParams(&'static str),
+}
+
+impl ParseSubscriptionError {
```
mattsse: can we move these private types below the ethpubinner impl block so that the order is

pub 
private

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -50,7 +82,51 @@ impl<Eth> EthPubSub<Eth> {
...
+
+    /// Creates a new, shareable instance with a fallback handler for unknown subscription kinds or
+    /// params.
+    pub fn with_spawner_and_fallback_handler<F>(
+        eth_api: Eth,
+        subscription_task_spawner: Box<dyn TaskSpawner>,
+        fallback_handler: F,
+    ) -> Self
+    where
+        F: Fn(String, Option<Box<JsonRawValue>>) -> Result<PubSubStream, ErrorObject<'static>>
```
mattsse, line 66: I think we can do without this fn

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -90,59 +166,80 @@ where
...
+            }
+        }
+    }
+
+    /// Dispatches validated subscription kinds to their concrete stream implementations.
+    async fn handle_parsed(
+        &self,
+        accepted_sink: SubscriptionSink,
+        parsed: ParsedSubscription,
     ) -> Result<(), ErrorObject<'static>> {
```
mattsse, line 177: this should remain unchanged entirely, easier to review that way if we dont alter existing code too much, so this should remain

```
 pub async fn handle_accepted(
        &self,
        accepted_sink: SubscriptionSink,
        kind: SubscriptionKind,
        params: Option<Params>,
```

with the same body

crates/rpc/rpc/src/eth/pubsub.rs
```diff
@@ -289,6 +460,10 @@ struct EthPubSubInner<EthApi> {
     eth_api: EthApi,
     /// The type that's used to spawn subscription tasks.
     subscription_task_spawner: Box<dyn TaskSpawner>,
+    /// Additional handlers keyed by subscription kind.
+    additional_handlers: Arc<RwLock<HashMap<String, SubscriptionHandler>>>,
+    /// Fallback handler for unknown subscription kinds or params.
+    fallback_handler: Option<FallbackHandler>,
```
mattsse, line 436: this we don't need because if no additional_handlers match then we return an error

mattsse (changes requested):
cool, a few more style suggestions

PR #20779, opened 2026-01-06
fix: shrink default storage trie capacity

Bashmunta:
Update StorageTries::shrink_to to also shrink the default_trie so the implementation matches the documented behavior. Previously the capacity distribution logic counted the default trie when computing per-trie limits but never applied the shrink, which made the comment misleading and could leave the default trie over-allocated compared to active and cleared tries.


mattsse (commented):
this isnt necessary bcause this type is effectively immutable

PR #20780, opened 2026-01-06
perf: use parallel signature recovery in debug_trace_raw_block

livehardd:
uses BlockBody's recover_signers instead of manual iteration to enable parallel recovery when rayon is enabled.

before:
`block.body().transactions().iter().map(|tx| tx.recover_signer()...).collect()`

after:
`block.body().recover_signers()`

Make ECDSA recovery run in parallel for blocks with many transactions, should improve debug tracing performance.


mattsse (approved):
this makes sense ty!

mattsse:
@livehardd just an unrelated action timeout

PR #20781, opened 2026-01-06
fix(transaction-pool): finalized block number should never decrease

aso20455:
FinalizedBlockTracker was using replace() which overwrites the stored block number regardless of whether it's higher or lower. Since finalized blocks are irreversible by definition, the tracker should ignore any value that's smaller than what we already have.

Updated the logic to only accept strictly increasing block numbers and fixed the corresponding test assertion.


mattsse (approved):
this makes sense, because this now doesn't alter the tracker if this would in theory be a lower block

is doesnt really matter because this is defacto only used for finalizing blobs

PR #20782, opened 2026-01-06
docs: align net JSON-RPC docs with implementation

yyhrnk:
Update net_listening, net_peerCount and net_version documentation to reflect actual return types and formats in Reth. Clarify that net_listening always returns true, document hex-encoded quantity for net_peerCount and string network ID for net_version so client implementers are not misled by outdated examples.


mattsse (approved):
fair

PR #20783, opened 2026-01-06
perf: preallocate capacity for filter chunk results

livehardd:
These headers are already bloom-filtered so most will have receipts, meaning final size is close to initial capacity. reduces allocation overhead when processing large log queries.


mattsse (approved):
this makes sense

PR #20785, opened 2026-01-06
perf(storage): flatten TrieUpdates before persist

cwkang1998:
Accumulates TriesUpdateSorted and batch write at at the end. Prior benchmark done in #19739 indicates this should improve batch processing performance.

Both #19739 and its succeesor #19991,, contains benchmarks that measures the performance improvements. I am not really familiar on how to do that, so if that's something that's required, would appreciate some guidance.

Closes #20611


mediocregopher (commented):
This looks correct to me, just needs merge conflict fixed. I'll also want to test this very thoroughly on real nodes before merging, as this area caused us problems in the past.

PR #20786, opened 2026-01-06
docs(flashblocks): fix outdated API examples in test_utils module documentation

andrewshab3:
Fixes outdated code examples in the `flashblocks/test_utils.rs` module documentation that reference old API that no longer exists.


mattsse (commented):
I believe 2s is the default block time

PR #20787, opened 2026-01-06
chore: add debug log for peer lookup miss in tx fetcher

livehardd:
The comment says "peer should always exist" but there's a defensive else branch with silent failure. added debug logging to track if/when this happens.


mattsse (commented):
this is basically unreachable and we dont need any logs here

PR #20788, opened 2026-01-06
chore: relax = requirement on cc dependency

SuperFluffy:
Need this because `aws-lc-sys` has `cc = "1.26.26`, which prevents cargo from finding a version of cc that fulfills both the `reth` and the `aws-ls-sys` requirement.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -694,7 +694,7 @@ ahash = "0.8"
 anyhow = "1.0"
 bindgen = { version = "0.71", default-features = false }
 block-padding = "0.3.2"
-cc = "=1.2.15"
+cc = "1.2.15"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -694,7 +694,7 @@ ahash = "0.8"
 anyhow = "1.0"
 bindgen = { version = "0.71", default-features = false }
 block-padding = "0.3.2"
-cc = "=1.2.15"
+cc = "1.2.15"
```
SuperFluffy, line 697: Alternatively, set this to `=1.2.26` to force an update to `Cargo.lock`?

mattsse (approved):

PR #20789, opened 2026-01-06
fix: eliminate TOCTOU race condition in remove_persisted_blocks

letmehateu:
Fixes a Time-of-Check-Time-of-Use (TOCTOU) race condition in `CanonicalInMemoryState::remove_persisted_blocks()` that could lead to inconsistent state between the `numbers` and `blocks` maps during concurrent persistence and reorg operations.


mattsse (commented):
this race condition is only hypothetical, because there's only one single entity that modifies these internals, so never concurrent writes

PR #20790, opened 2026-01-06
docs: clarify TreeRootEntry::content unsigned format

Snezhkko:
The documentation for TreeRootEntry::content did not match the actual string being signed. It mentioned a sig field and omitted the enrtree-root:v1 prefix, which is misleading given EIP-1459 specifies that the signature is computed over the unsigned content without the sig= part. This change updates the comment to document the real unsigned payload, including the root prefix and excluding the signature field.


mattsse (approved):
sure

PR #20796, opened 2026-01-07
docs: update permalink to source code

bobtajson:
Updated the link to the legacy source code to match the example


mattsse (commented):
we can keep this as is, the new permalink will also become stale

PR #20767, opened 2026-01-06
perf(trie): add cursor locality optimization for storage cursors

yongkangc:
**Problem**
Storage trie cursors use `seek_by_key_subkey` (O(log N)) for every lookup, even during forward-sequential traversal where keys are accessed in sorted order.

**Solution**
Track `last_key` in `DatabaseHashedStorageCursor` and `DatabaseStorageTrieCursor`. When seeking forward past the current position, use `next_dup`/`next_dup_val` (O(1)) instead of re-seeking.

**Bench**
HashedStorages Cursor

| Entries | Optimized (forward_sequential_seek) | Baseline (always_seek) | Improvement |
|---------|-------------------------------------|------------------------|-------------|
| 100     | 12.4 µs                            | 15.8 µs                | **27%**     |
| 1,000   | 123 µs                             | 178 µs                 | **31%**     |
| 10,000  | 1.24 ms                            | 2.15 ms                | **42%**     |

StoragesTrie Cursor

| Entries | Optimized (forward_sequential_seek) | Baseline (always_seek) | Improvement |
|---------|-------------------------------------|------------------------|-------------|
| 100     | 14.1 µs                            | 18.2 µs                | **29%**     |
| 1,000   | 141 µs                             | 215 µs                 | **34%**     |
| 10,000  | 1.41 ms                            | 2.58 ms                | **45%**     |

Benchmark results and code: https://gist.github.com/yongkangc/0b5d9fde2758cf87b2ab4c4a43c7e173

https://github.com/paradigmxyz/reth/issues/20766#issue-3783674047

Closes: https://github.com/paradigmxyz/reth/issues/20766

PR #20798, opened 2026-01-07
perf(trie): add cursor locality for account cursors (experiment)

figtracer:
this a follow up to #20767 but applied on account cursors

consider if can this have integration regressions when combined with the other pr


figtracer:
i cant really benchmark this atm, will search for something less invasive

PR #20799, opened 2026-01-07
perf(rpc): reuse EVM instance in debug_traceBlock

t-nakatani:
This closes #17045

Noticed this issue was sitting idle, so I took a stab at it.

## Summary

Reuse a single EVM instance across all transactions when tracing a block in the `debug_` API, matching the optimization already present in the `trace_` API.

## Motivation

Previously, `debug_traceBlock` (and related methods) called `eth_api.inspect()` for each transaction in a loop, which creates a new EVM instance every time:

## Changes

This PR implements EVM reuse manually in `debug.rs`:

1. Create the EVM once before the loop using `evm_with_env_and_inspector()`
2. Call `evm.transact()` for each transaction (reusing the same EVM)
3. Access inspector and database through `evm.components_mut()` for `get_result()`, `fuse()`, and `commit()`

> we recently added support for reusing the evm for block tracing:
which we can mirror for the debug_ api as well.

The `trace_` API already solved this problem using `create_tracer().try_trace_many()` and `trace_block_until_with_inspector()`. However, these APIs require `Inspector: Clone`.

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -121,7 +121,6 @@ where
                 let mut transactions = block.transactions_recovered().enumerate().peekable();
                 let inspector = DebugInspector::new(opts).map_err(Eth::Error::from_eth_err)?;
 
-                // Create EVM once and reuse it for all transactions in the block
                 let mut evm = eth_api.evm_config().evm_with_env_and_inspector(
                     &mut db,
                     evm_env.clone(),
@@ -133,7 +132,6 @@ where
                     let tx_env = eth_api.evm_config().tx_env(tx);
                     let res = evm.transact(tx_env.clone()).map_err(Eth::Error::from_evm_err)?;
 
-                    // Access inspector and db through EVM components
                     let (db_ref, inspector_ref, _) = evm.components_mut();
                     let result = inspector_ref
                         .get_result(
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -116,6 +116,7 @@ where
             .spawn_with_state_at_block(block.parent_hash(), move |eth_api, mut db| {
                 let mut results = Vec::with_capacity(block.body().transactions().len());
 
+                let block_env = evm_env.block_env.clone();
                 eth_api.apply_pre_execution_changes(&block, &mut db, &evm_env)?;
 
                 let mut transactions = block.transactions_recovered().enumerate().peekable();
@@ -123,7 +124,7 @@ where
 
                 let mut evm = eth_api.evm_config().evm_with_env_and_inspector(
                     &mut db,
-                    evm_env.clone(),
+                    evm_env,
                     inspector,
                 );
 
@@ -141,7 +142,7 @@ where
                                 tx_index: Some(index),
                             }),
                             &tx_env,
-                            &evm_env.block_env,
+                            &block_env,
                             &res,
                             db,
                         )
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -119,18 +119,23 @@ where
                 eth_api.apply_pre_execution_changes(&block, &mut db, &evm_env)?;
 
                 let mut transactions = block.transactions_recovered().enumerate().peekable();
-                let mut inspector = DebugInspector::new(opts).map_err(Eth::Error::from_eth_err)?;
+                let inspector = DebugInspector::new(opts).map_err(Eth::Error::from_eth_err)?;
+
+                // Create EVM once and reuse it for all transactions in the block
+                let mut evm = eth_api.evm_config().evm_with_env_and_inspector(
+                    &mut db,
+                    evm_env.clone(),
```
mattsse: hmm, ideally we can also avoid the clone here but unsure if possible because we need to borrow the db as mut

crates/rpc/rpc/src/debug.rs
```diff
@@ -119,18 +119,23 @@ where
...
 
-                    let res = eth_api.inspect(
-                        &mut db,
-                        evm_env.clone(),
-                        tx_env.clone(),
-                        &mut inspector,
-                    )?;
-                    let result = inspector
+                    // Access inspector and db through EVM components
+                    let (db_ref, inspector_ref, _) = evm.components_mut();
```
mattsse: let's drop the _ref suffix here

crates/rpc/rpc/src/debug.rs
```diff
@@ -119,18 +119,23 @@ where
                 eth_api.apply_pre_execution_changes(&block, &mut db, &evm_env)?;
 
                 let mut transactions = block.transactions_recovered().enumerate().peekable();
-                let mut inspector = DebugInspector::new(opts).map_err(Eth::Error::from_eth_err)?;
+                let inspector = DebugInspector::new(opts).map_err(Eth::Error::from_eth_err)?;
+
+                // Create EVM once and reuse it for all transactions in the block
```
mattsse: I think we can drop this comment here

crates/rpc/rpc/src/debug.rs
```diff
@@ -119,18 +119,23 @@ where
...
+                    let res = evm.transact(tx_env.clone()).map_err(Eth::Error::from_evm_err)?;
 
-                    let res = eth_api.inspect(
-                        &mut db,
-                        evm_env.clone(),
-                        tx_env.clone(),
-                        &mut inspector,
-                    )?;
-                    let result = inspector
+                    // Access inspector and db through EVM components
```
mattsse: this comment isnt very useful

mattsse (changes requested):
cool, this makes sense

I have a few pedantic style nits

crates/rpc/rpc/src/debug.rs
t-nakatani: @mattsse Thanks for review! 
fixed in b774afb9c5b58ea4742cab7490f5bde9b3baef54

inspector.get_result() needs both &mut db (from components_mut) and &block_env, and evm_env has already been moved into the EVM, the clone seems necessary here.

BlockEnv is much smaller than the full EvmEnv (just block metadata vs cfg + block + other fields), so the cost should be acceptable. 

Please let me know if you have other ideas!

crates/rpc/rpc/src/debug.rs
t-nakatani: fixed in e2f37cf2c92ba2f1712d80863ec39093be69eeef

crates/rpc/rpc/src/debug.rs
t-nakatani: fixed in dec631dda48663e35a7ee7c2e88955936bf21aee

crates/rpc/rpc/src/debug.rs
t-nakatani: fixed in dec631dda48663e35a7ee7c2e88955936bf21aee

PR #20800, opened 2026-01-07
docs: clarify pending pending_block build_block docs

phrwlk:
The previous documentation for build_block described behavior for the ActualPending origin, but this function is only used for locally derived pending blocks from the latest canonical header. This could mislead readers into thinking it is responsible for handling execution-layer pending blocks received from the CL.


mattsse (approved):
seems alright

PR #20805, opened 2026-01-07
chore(reth-bench-compare): skip last unwind

DaniPopes:
We can skip the last unwind since there's nothing more to benchmark.


mattsse (approved):
I believe this makes sense @shekhirin ?

PR #20808, opened 2026-01-07
feat(reth-bench-compare): Do unwind first

mediocregopher:
reth-bench-compare currently performs its unwind _after_ each of its workflows (warmup, baseline, and feature). This means that:

* When synced to the tip, you must first manually unwind by however many blocks you want to bench against, otherwise they won't be available.
* When RBC completes the node will be behind by however many blocks were benched against.

This PR changes RBC to perform the unwind step prior to each workflow. This requires initially starting the node to get the current block number, but this is fine.

### DIFF
--- a/bin/reth-bench-compare/src/cli.rs
+++ b/bin/reth-bench-compare/src/cli.rs
@@ -521,8 +521,12 @@ async fn run_warmup_phase(
     args: &Args,
     is_optimism: bool,
     baseline_commit: &str,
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
this is better, we always unwind first anyway in the nushell script

mattsse (approved):

shekhirin (approved):

PR #20809, opened 2026-01-07
refactor(config): delegate PruneConfig::has_receipts_pruning

kurahin:
`PruneConfig::has_receipts_pruning` duplicated the same condition already implemented on `PruneModes`. This change makes the config-layer helper method delegate directly to `self.segments.has_receipts_pruning()`, so there is a single source of truth for the receipts pruning check. Public API and behavior remain unchanged, but future changes to `PruneModes` logic will be consistently reflected everywhere.


mattsse (approved):
this is indeed equivalent

PR #20626, opened 2025-12-24
feat(engine): add persistence completion wake channel

cakevm:
Based on https://github.com/paradigmxyz/reth/pull/20616 I noticed that there is kind of a fixed delay until a block(s) is finally persisted ~500ms. The idea in the PR is to add channel to notify the listener when the persistence is finished. This allows to wake up before the timeout is triggered, since it may that there is no new engine event that will trigger the wakeup. The timeout is still there as safety net. The latency is still higher than I expected, but not fixed and better as before.

Overall the next message is proceed when:
- An incoming engine message
- A persistence completion signal <-- NEW
- A fallback timeout (500ms)

https://github.com/paradigmxyz/reth/blob/ad37490e7dd151d5ac3767099bb4f485e86f603c/crates/engine/tree/src/tree/mod.rs#L1203-L1218

---

I used https://github.com/cakevm/reth-stale-direct-db to test the latency:
Before (usually it is around 500ms, I picked some worst case). `latency_ms=501` is the latency between newHead and the persisted event from https://github.com/paradigmxyz/reth/pull/20616 is received.
```
2025-12-24T09:06:50.079711Z  INFO reth_stale_direct_db::monitor: New head received block_number=24081475 hash=0x7bc412b263752e1e213f333d075e11ca417753e19ded09baaf7486e09198eb66 pending=0
2025-12-24T09:06:50.581421Z  INFO reth_stale_direct_db::monitor: New latest persisted block received block_number=24081475 hash=0x7bc412b263752e1e213f333d075e11ca417753e19ded09baaf7486e09198eb66 blocks_to_flush=1 max_latency_ms=501
2025-12-24T09:06:50.581441Z  INFO reth_stale_direct_db::monitor: Verifying block block_number=24081475 rpc_block_hash=0x7bc412b263752e1e213f333d075e11ca417753e19ded09baaf7486e09198eb66 latency_ms=501
2025-12-24T09:06:50.581674Z  INFO reth_stale_direct_db::monitor: Consistency check passed (256 block hashes accessible) block_number=24081475 db_last_block=24081475
2025-12-24T09:06:50.581682Z  INFO reth_stale_direct_db::monitor: Block hash matches block_number=24081475 db_block_hash=0x7bc412b263752e1e213f333d075e11ca417753e19ded09baaf7486e09198eb66
[...]

PR #20813, opened 2026-01-07
perf(engine): use crossbeam select for persistence events

mattsse:
Follow-up to #20626 which added a signal channel for persistence completion.

This PR takes it further by replacing the 500ms `recv_timeout` polling approach with proper event-driven waiting using `crossbeam_channel::select_biased!`. The engine now waits on both the incoming message channel and persistence response channel simultaneously, waking immediately when either is ready.

Previously, if a persistence task completed but no new engine message arrived, the engine would wait up to 500ms before processing the completion. Now it wakes immediately.

Changes:
- Use `crossbeam_channel` for persistence responses instead of `tokio::sync::oneshot`
- Use `select_biased!` to prioritize persistence completion over new messages, ensuring disk writes are processed as soon as they complete and in memory state is updated and ready to perform the next regular engine message
- Remove the `recv_timeout` polling pattern entirely

### DIFF
--- a/crates/engine/tree/src/engine.rs
+++ b/crates/engine/tree/src/engine.rs
@@ -6,6 +6,7 @@ use crate::{
     download::{BlockDownloader, DownloadAction, DownloadOutcome},
 };
 use alloy_primitives::B256;
+use crossbeam_channel::Sender;
 use futures::{Stream, StreamExt};
 use reth_chain_state::ExecutedBlock;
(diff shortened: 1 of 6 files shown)


Rjected (approved):
converting this all to proper nonblocking select in place of the 500ms wait makes sense, lgtm

yongkangc (approved):
LGTM, makes sense esp on high load

PR #20820, opened 2026-01-07
perf: remove redundant contains_key

livehardd:
The contains_key check is redundant since the while let loop immediately tries to remove the same key anyway.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -153,16 +153,6 @@ impl ProofSequencer {
             self.pending_proofs.insert(sequence, update);
         }
 
-        // BTreeMap::first_key_value is more efficient than contains_key for checking
-        // if the next expected proof exists, since we only need to check the first key
-        let Some(first_key) = self.pending_proofs.first_key_value().map(|(k, _)| *k) else {
-            return Vec::new()
-        };
-
-        if first_key != self.next_to_deliver {
-            return Vec::new()
-        }
-
         let mut consecutive_proofs = Vec::with_capacity(self.pending_proofs.len());
         let mut current_sequence = self.next_to_deliver;
 
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -153,6 +153,11 @@ impl ProofSequencer {
             self.pending_proofs.insert(sequence, update);
         }
 
+        // return early if we don't have the next expected proof to avoid unnecessary allocation
+        if self.pending_proofs.is_empty() {
+            return Vec::new()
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -153,11 +153,6 @@ impl ProofSequencer {
             self.pending_proofs.insert(sequence, update);
         }
 
-        // return early if we don't have the next expected proof
```
mattsse, line 156: this is correct, but I'm unsure if this is always the case when pending_proofs is empty, because if we skip this then we always alloc

mattsse (commented):
wdyt @mediocregopher @yongkangc

crates/engine/tree/src/tree/payload_processor/multiproof.rs
livehardd: Nice catch! I didn't realized this before. how about updating to use `is_empty()` check instead—avoids both the redundant hash lookup and unnecessary allocations when pending_proofs is empty. 
PTAL

mattsse: ah cool, yeah this makes sense then

mattsse: ah actually this doesn't avoid the case where the vec isn't empty but doesnt contain the next_to_deliver

livehardd: 😂 all right, lets try `first_key_value()` to check if the smallest key matches next_to_deliver.

mediocregopher (approved):
This seems fine to me, I think it could be done even without the `first_key_value` check and rely only on if any calls to `remove` return anything, but probably not super impactful

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -153,8 +153,13 @@ impl ProofSequencer {
...
-        if !self.pending_proofs.contains_key(&self.next_to_deliver) {
+        // BTreeMap::first_key_value is more efficient than contains_key for checking
+        // if the next expected proof exists, since we only need to check the first key
+        let Some(first_key) = self.pending_proofs.first_key_value().map(|(k, _)| *k) else {
+            return Vec::new()
+        };
+
+        if first_key != self.next_to_deliver {
```
mattsse: this seems now a bit convoluted just for the fast pass

let's remove

mattsse (changes requested):

crates/engine/tree/src/tree/payload_processor/multiproof.rs
livehardd: Sure.

mattsse (approved):

PR #20821, opened 2026-01-07
docs: clarify bodies downloader set_download_range semantics

ANtutov:
The previous documentation for BodiesDownloader::set_download_range described an exclusive range and implied draining and re-requesting queued bodies, which does not match the actual implementation or tests. This change updates the comment to document the inclusive range and the three real code paths (suffix no-op, consecutive range, and full reset) so that callers and maintainers can rely on accurate behavior without reading the implementation.


mattsse (approved):
these are better

PR #20822, opened 2026-01-07
fix: invalid header cache state corrupted by insert checks

livehardd:
`get()` increments `hit_count` and can evict entries, so using it for existence checks causes unintended side effects. Switched to `peek()` for `insert()` and `insert_with_invalid_ancestor()`.

### DIFF
--- a/crates/engine/tree/src/tree/invalid_headers.rs
+++ b/crates/engine/tree/src/tree/invalid_headers.rs
@@ -58,7 +58,7 @@ impl InvalidHeaderCache {
         header_hash: B256,
         invalid_ancestor: BlockWithParent,
     ) {
-        if self.get(&header_hash).is_none() {
+        if self.headers.peek(&header_hash).is_none() {
             warn!(target: "consensus::engine", hash=?header_hash, ?invalid_ancestor, "Bad block with existing invalid ancestor");
             self.insert_entry(header_hash, invalid_ancestor);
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this should go through `get` because we need to track how often these were provided

livehardd:
@mattsse 

After double check by test I think using `get()` here is a bug. When the same invalid block is received from multiple peers, each insert attempt increments hit_count even though we're just checking if it exists.

After 128 duplicate inserts (easy to hit during network splits or attacks), the entry gets evicted and we have to revalidate a known-bad block.

I added a test that shows this: `test_duplicate_insert_does_not_corrupt_hit_count`

peek() is the standard pattern for existence checks elsewhere in the codebase (e.g., `crates/net/network/src/cache.rs:49`). If you want to track duplicate insert attempts, we should add a separate metric instead of corrupting hit_count.

should we reopen this or just leave it?

mattsse:
> After 128 duplicate inserts (easy to hit during network splits or attacks), the entry gets evicted and we have to revalidate a known-bad block.

this is intentional, if the CL really wants to submit this block, then we revalidate once more after exceeding the Hit count

livehardd:
> > After 128 duplicate inserts (easy to hit during network splits or attacks), the entry gets evicted and we have to revalidate a known-bad block.
> 
> this is intentional, if the CL really wants to submit this block, then we revalidate once more after exceeding the Hit count

Thanks for the info!

PR #20823, opened 2026-01-07
perf: Add cancellation support to trace API methods

ethfanWilliam:
Adds CancelOnDrop mechanism to trace_call, trace_raw_transaction, trace_call_many, trace_filter, trace_block, and replay_block_transactions. 

These methods can be long-running (especially trace_filter which can scan up to 100 blocks), so they should handle request cancellation the same way transact_call_at does. Without this, canceled requests keep consuming CPU/memory until completion which is wasteful.


mattsse (commented):
these apis are behind stricter ratelimiting so we don't really need this here

PR #20824, opened 2026-01-07
feat(rpc): add debug_getBlockAccessList endpoint for EIP-7928

### LINKED ISSUE #20269: Add support for debug_getBlockAccessList
### Describe the feature

new debug level endpoint that re-executes the block and returns the BAL

@Soubhik-10 @Rimeeeeee 

### Additional context

_No response_

Sabnock01:
## Summary

Implements the `debug_getBlockAccessList` RPC method that re-executes a block and returns the Block Access List (BAL) as defined in [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928).

- Uses `StorageInspector` to track SLOAD operations during transaction execution
- Collects state changes (balance, nonce, code, storage) per transaction
- Returns addresses sorted lexicographically per EIP-7928 specification
- Distinguishes between storage reads (accessed but not written) and storage writes

Closes #20269

Update: Just stub impl for now since `revm` updates will include BAL-building


mattsse (changes requested):
ty ser!

with the upcoming revm update bal building will be available out of the box, but we can already stub in this endpoint so I'd merge this already with a return unimplemented error

because the impl will be drastically simpler

mattsse (approved):
ty!
this will help keep the diff small after next revm upgrade

PR #20825, opened 2026-01-07
perf(sidecar): skip processing for blocks without blob txs

gap-editor:
This PR introduces a performance optimization to the MinedSidecarStream.

Changes:

Added an early return check in process_block to verify if the block contains any EIP-4844 transactions.

Reasoning:
Previously, the stream would proceed to query the transaction pool via get_all_blobs_exact and potentially set up Beacon API request logic even for blocks that had no blob transactions. Since most blocks currently may not contain blobs, this avoids unnecessary overhead and logic execution for the majority of canonical state notifications.


mattsse (commented):
this is just example code, we can keep as is

PR #20828, opened 2026-01-08
chore: correct `StorageHistory` prune map size constant name

Fibonacci747:
StorageHistory pruning comment referenced ACCOUNT_HISTORY_TABLES_TO_PRUNE even though the code uses STORAGE_HISTORY_TABLES_TO_PRUNE. This was a copy-paste leftover from AccountHistory and could confuse readers reviewing memory usage estimates. The change only updates the comment to match the actual constant name and does not affect runtime behavior.


mattsse (approved):
fair

PR #20718, opened 2026-01-03
feat(cli): add `reth report` command for debug report generation

meetrick:
## Motivation
When users encounter state root mismatch or EVM execution errors during sync, they need to manually collect multiple pieces of information (checksums, logs, block info, version history, etc.) for bug reports. This process is error-prone, time-consuming, and results in inconsistent reports.

This PR addresses issue #18569 and the TODO in the codebase:
```rust
// TODO: automate the process outlined below so the user can just send in a debugging package
```

## Solution
Add a new top-level `reth report` command that automates the collection of all necessary debugging information into a single `.tar.gz` archive.

### Usage
```bash
reth report --block <BLOCK_NUMBER> [--checksum] [-y] [--output report.tar.gz]
```

### What it collects
| Information | Description |
|-------------|-------------|
| **metadata.json** | Timestamp, reth version, chain name, data directory |
| **block_info.json** | Block number, hash, parent hash, state root, transactions/receipts root |
| **db_stats.json** | Latest block number, table statistics (entries, size), optional checksums |
| **version_history.json** | Full VersionHistory table dump |
| **logs/** | All `reth.log*` files from the log directory |
| **invalid_block_hooks/** | All files from the invalid block hooks directory |

### Features
- **User consent prompt**: Privacy notice with confirmation before collecting data (`-y` to skip)
- **GitHub issue link**: Generates a pre-filled issue creation URL
- **Checksum option**: `--checksum` flag to include database table checksums (takes a long time)
- **Flexible output**: Custom output path via `--output` flag

### Error guidance
Updated error messages to guide users to run `reth report`:
- State root mismatch errors (`INVALID_STATE_ROOT_ERROR_MESSAGE`)
- EVM execution errors (`DEBUG_REPORT_SUGGESTION`)

## Example Output
```
======================================================================
RETH DEBUG REPORT GENERATOR
======================================================================

This tool will collect the following information:
  • Database table statistics
  • VersionHistory table (client version history)
  • Log files from: /path/to/datadir/logs
  • Invalid block hook outputs from: /path/to/datadir/invalid_block_hooks
  • Block header information for block #12345000

⚠️  PRIVACY NOTICE:
  The generated archive may contain sensitive information about your node.
  Please review the contents before sharing publicly.

Do you want to proceed? [y/N]: y

======================================================================
DEBUG REPORT GENERATED SUCCESSFULLY
======================================================================

Archive saved to: reth_report_20260108_123456.tar.gz

📋 To create a GitHub issue, visit:
[...]

PR #20832, opened 2026-01-08
feat(cli): add reth report command for debug report generation

meetrick:
## Motivation
When users encounter state root mismatch or EVM execution errors during sync, they need to manually collect multiple pieces of information (checksums, logs, version history, etc.) for bug reports. This process is error-prone, time-consuming, and results in inconsistent reports.

This PR addresses issue #18569 and the TODO in the codebase:
```rust
// TODO: automate the process outlined below so the user can just send in a debugging package
```

## Solution
Add a new top-level `reth report` command that automates the collection of all necessary debugging information into a single `.tar.gz` archive.

### Usage
```bash
reth report [--checksum] [-y] [--output report.tar.gz]
```

### What it collects
| Information | Description |
|-------------|-------------|
| **metadata.json** | Timestamp (UTC), reth version, chain name, data directory |
| **db_stats.json** | Latest block number, table statistics (entries, size), optional checksums |
| **stage_checkpoints.json** | Pipeline stage progress (sync status) |
| **prune_checkpoints.json** | Pruning progress for each segment |
| **version_history.json** | Full VersionHistory table dump |
| **logs/** | All `reth.log*` files from the cache directory |
| **invalid_block_hooks/** | All files from the invalid block hooks directory |

### Features
- **User consent prompt**: Privacy notice with confirmation before collecting data (`-y` to skip)
- **GitHub issue link**: Generates a pre-filled issue creation URL
- **Checksum option**: `--checksum` flag to include database table checksums (takes a long time)
- **Flexible output**: Custom output path via `--output` flag

### Error guidance
Updated error messages to guide users to run `reth report`:
- State root mismatch errors (`INVALID_STATE_ROOT_ERROR_MESSAGE`)
- EVM execution errors (`DEBUG_REPORT_SUGGESTION`)

## Example Output
```
======================================================================
RETH DEBUG REPORT GENERATOR
======================================================================

This tool will collect the following information:
  • Database table statistics
  • Stage checkpoints (sync progress)
  • Prune checkpoints (pruning progress)
  • VersionHistory table (client version history)
  • Log files from: /Users/user/Library/Caches/reth/logs
  • Invalid block hook outputs from: /path/to/datadir/invalid_block_hooks

⚠️  PRIVACY NOTICE:
  The generated archive may contain sensitive information about your node.
  Please review the contents before sharing publicly.

Do you want to proceed? [y/N]: y

======================================================================
DEBUG REPORT GENERATED SUCCESSFULLY
======================================================================

Archive saved to: reth_report_20260108_123456Z.tar.gz

📋 To create a GitHub issue, visit:
https://github.com/paradigmxyz/reth/issues/new?title=Debug%20Report&labels=C-bug

⚠️  IMPORTANT:
   • You must be logged into GitHub in your browser
   • The issue will be created under YOUR GitHub account
   • Please attach the generated archive to your issue
   • Review the archive contents before sharing
```

## Documentation
Added CLI documentation for both `reth` and `op-reth`:

| File | Change |
|------|--------|
| `docs/vocs/docs/pages/cli/reth/report.mdx` | New command documentation |
| `docs/vocs/docs/pages/cli/op-reth/report.mdx` | New command documentation |
| `docs/vocs/sidebar-cli-reth.ts` | Added sidebar entry |
| `docs/vocs/sidebar-cli-op-reth.ts` | Added sidebar entry |

## Related
Closes #18569 (MVP implementation without S3 upload - can be added in follow-up PR)


meetrick:
@mediocregopher When I reset the branch in PR #20718, the PR was automatically closed, and I didn’t have permission to reopen it. So I recreated the PR using the same branch with the updated code.
I have now incorporated the feedback from Issue #18569, could you please take a look and let me know what you think?

crates/cli/commands/src/db/mod.rs
```diff
@@ -10,7 +10,7 @@ use std::{
     sync::Arc,
 };
 mod account_storage;
-mod checksum;
+pub mod checksum;
```
mediocregopher, line 13: Is this necessary?

crates/stages/api/src/pipeline/mod.rs
```diff
@@ -159,14 +159,14 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
                     PipelineTarget::Sync(tip) => self.set_tip(tip),
                     PipelineTarget::Unwind(target) => {
                         if let Err(err) = self.move_to_static_files() {
-                            return (self, Err(err.into()))
+                            return (self, Err(err.into()));
```
mediocregopher: Let's leave out these unrelated fmt changes

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+    transactions_root: String,
+    /// Receipts root
+    receipts_root: String,
+    /// Timestamp
+    timestamp: u64,
+}
+
+/// Database statistics
+#[derive(Debug, Serialize)]
+struct DbStats {
```
mediocregopher, line 70: Something that would be helpful to include would be also the stage checkpoints and prune checkpoints

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        info!(target: "reth::cli", block = ?self.block, "Generating debug report archive");
+
+        // Determine output path
+        let output_path = self.output.clone().unwrap_or_else(|| {
+            let timestamp = humantime::format_rfc3339_seconds(SystemTime::now())
+                .to_string()
+                .replace([':', '-'], "")
+                .replace('T', "_")
+                .split('Z')
```
mediocregopher: I think we're losing timezone information here. Can we either keep the timezone or ensure that the time is always in UTC? Probably the former would be less confusing for the user

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        // Create the archive
+        let file = File::create(&output_path)?;
+        let encoder = GzEncoder::new(file, Compression::default());
+        let mut archive = Builder::new(encoder);
+
+        // Collect and add metadata
+        let chain_spec = provider_factory.chain_spec();
+        let metadata = ReportMetadata {
+            generated_at: humantime::format_rfc3339(SystemTime::now()).to_string(),
```
mediocregopher: nitpick: let's use a single "now" value across the whole `execute`

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+            reth_version: env!("CARGO_PKG_VERSION").to_string(),
+            chain: chain_spec.chain().to_string(),
+            target_block: self.block,
+            data_dir: data_dir.data_dir().display().to_string(),
+        };
+        add_json_to_archive(&mut archive, "metadata.json", &metadata)?;
+
+        // Collect block info if specified
+        if let Some(block) = self.block {
```
mediocregopher: I'm not sure this info is worth collecting, in general the block header isn't the issue, it's going to match what we see in an explorer. I think it'd be better to give the user one less step and remove this.

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+                    build_timestamp: version.build_timestamp,
+                });
+            }
+            Ok::<(), eyre::Report>(())
+        })??;
+        add_json_to_archive(&mut archive, "version_history.json", &version_history)?;
+        info!(target: "reth::cli", entries = version_history.len(), "Version history collected");
+
+        // Collect log files
+        let logs_collected = collect_logs(&data_dir.data_dir().join("logs"), &mut archive)?;
```
mediocregopher: Did you test this part? I think logs are generally found in the cache directory, not the data dir

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+        println!("   • You must be logged into GitHub in your browser");
+        println!("   • The issue will be created under YOUR GitHub account");
+        println!("   • Please attach the generated archive to your issue");
+        println!("   • Review the archive contents before sharing");
+
+        Ok(())
+    }
+
+    /// Returns the chain spec if available
+    pub fn chain_spec(&self) -> Option<&Arc<C::ChainSpec>> {
```
mediocregopher: This method seems redundant

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+                    checksum: format!("ERROR: {e}"),
+                });
+            }
+        }
+    }
+
+    Ok(checksums)
+}
+
+fn collect_logs<W: Write>(log_dir: &Path, archive: &mut Builder<W>) -> eyre::Result<usize> {
```
mediocregopher: I would just do `collect_directory_recursive` instead of a custom log file finder

crates/cli/commands/src/report.rs
```diff
@@ -0,0 +1,481 @@
...
+) -> eyre::Result<()> {
+    let json = serde_json::to_string_pretty(data)?;
+    let bytes = json.as_bytes();
+
+    let mut header = tar::Header::new_gnu();
+    header.set_path(name)?;
+    header.set_size(bytes.len() as u64);
+    header.set_mode(0o644);
+    header.set_mtime(
+        SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap_or_default().as_secs(),
```
mediocregopher: Use the same "now" value as rest of the command

mediocregopher (changes requested):
Thanks @meetrick , I left some comments and I'll get some others on the team to review it too. Sorry it took so long to get to it.

crates/cli/commands/src/db/mod.rs
meetrick: Regarding the pub mod checksum change, I wanted to clarify your concern.

1. Are you questioning whether the checksum feature itself is needed in the report command? (In that case, I can remove the --checksum option entirely, and users can run reth db checksum separately if needed)

2. Or are you concerned about making the module public? (If so, I could reimplement the checksum logic locally in report.rs to avoid exposing the module)

Could you let me know which approach you would prefer?

crates/stages/api/src/pipeline/mod.rs
meetrick: While fixing a lint error, I accidentally modified some unrelated parts as well. I’ll revert the unnecessary changes.

crates/cli/commands/src/report.rs
meetrick: Agree

crates/cli/commands/src/report.rs
meetrick: I agreed, I will fix it

crates/cli/commands/src/report.rs
meetrick: I agreed, I will fix it

crates/cli/commands/src/report.rs
meetrick: I agreed, I will fix it

crates/cli/commands/src/report.rs
meetrick: I thought including it in the report would let users get all the information at once without checking an external explorer.
That said, if we want to keep things lightweight, it’s probably better to remove this.

crates/cli/commands/src/report.rs
meetrick: I missed somthing. Thanks review for details

crates/cli/commands/src/report.rs
meetrick: got it!

crates/cli/commands/src/report.rs
meetrick: I agreed

crates/stages/api/src/pipeline/mod.rs
meetrick: fix it

crates/cli/commands/src/report.rs
meetrick: Updated

crates/cli/commands/src/report.rs
meetrick: Fix it

crates/cli/commands/src/report.rs
meetrick: Fix it

crates/cli/commands/src/report.rs
meetrick: fix it

crates/cli/commands/src/report.rs
meetrick: Fix it

crates/cli/commands/src/report.rs
meetrick: Fix it

crates/cli/commands/src/report.rs
meetrick: Remove it

crates/cli/commands/src/report.rs
meetrick: Fix it

PR #20833, opened 2026-01-08
perf(trie): avoid cloning Arc contents in deferred trie fallback

### LINKED ISSUE #20802: Try unwrap pending inputs
### Describe the feature

these inputs are arced but they might actually be unique
https://github.com/paradigmxyz/reth/blob/5fa1b99bb65be3a574f10d7afc8e0a748238df54/crates/chain-state/src/deferred_trie.rs#L215-L222

so instead of borrowing, we can try unwrapping them and operate on Cow here:

https://github.com/paradigmxyz/reth/blob/5fa1b99bb65be3a574f10d7afc8e0a748238df54/crates/chain-state/src/deferred_trie.rs#L152-L153

for borrowed we keep using:

https://github.com/paradigmxyz/reth/blob/5fa1b99bb65be3a574f10d7afc8e0a748238df54/crates/chain-state/src/deferred_trie.rs#L158-L159

and for owned we directly convert into

we like need to change this to an Option or something euqivalent so that we can take owenership
https://github.com/paradigmxyz/reth/blob/5fa1b99bb65be3a574f10d7afc8e0a748238df54/crates/chain-state/src/deferred_trie.rs#L215-L215

### Additional context

_No response_

yongkangc:
Closes #20802. Use `Arc::try_unwrap` to move data when the Arc is unique instead of always cloning via `clone_into_sorted()`.

**Problem**
In `wait_cloned()`, when falling back to synchronous computation, the `PendingInputs` are discarded after use. The `Arc<HashedPostState>` and `Arc<TrieUpdates>` often have refcount=1, yet we were always cloning.

**Solution**
- Use `mem::replace` to take ownership of `PendingInputs`
- Call `Arc::try_unwrap` to check if we have the only reference
- If unique: use `into_sorted()` (moves data, no allocation)
- If shared: fall back to `clone_into_sorted()` (same as before)

### DIFF
--- a/crates/chain-state/src/deferred_trie.rs
+++ b/crates/chain-state/src/deferred_trie.rs
@@ -441,4 +441,22 @@ mod tests {
         let (_, account) = &overlay_state[0];
         assert_eq!(account.unwrap().nonce, 2);
     }
+
+    /// Verifies `Arc::try_unwrap` fallback: when Arcs are shared, external references stay valid.
+    #[test]
+    fn shared_arc_fallback_preserves_external_refs() {
+        let hashed_arc = Arc::new(HashedPostState::default());
+        let trie_arc = Arc::new(TrieUpdates::default());
+
+        // Hold external references (refcount = 2, so try_unwrap must fail)
+        let external_hashed = Arc::clone(&hashed_arc);
+        let external_trie = Arc::clone(&trie_arc);
+
+        let deferred = DeferredTrieData::pending(hashed_arc, trie_arc, B256::ZERO, Vec::new());
+        let _ = deferred.wait_cloned();
+
+        // External refs must still be valid (proves clone path was used, not move)
+        assert_eq!(Arc::strong_count(&external_hashed), 1);
+        assert_eq!(Arc::strong_count(&external_trie), 1);
+    }
 }


crates/chain-state/src/deferred_trie.rs
```diff
@@ -131,56 +131,47 @@ impl DeferredTrieData {
...
+    fn sort_and_build_trie_input(
+        hashed_state: Arc<HashedPostState>,
+        trie_updates: Arc<TrieUpdates>,
         anchor_hash: B256,
         ancestors: &[Self],
     ) -> ComputedTrieData {
-        // Sort the current block's hashed state and trie updates
-        let sorted_hashed_state = Arc::new(hashed_state.clone_into_sorted());
-        let sorted_trie_updates = Arc::new(trie_updates.clone_into_sorted());
+        let sorted_hashed_state = Arc::new(match Arc::try_unwrap(hashed_state) {
```
yongkangc: now we avoid the clone here

 - Before: always clone → allocate + copy + extra writes.
  - After: do `Arc::try_unwrap` (refcount check + branch).
      - If unique: no clone, no extra alloc, no copy → fewer instructions and less memory traffic.
      - If shared: pay a small refcount/branch (strong count + `try_unwrap` branch prediction) cost and then do the same clone as before.

PR #20834, opened 2026-01-08
feat(metrics): configurable jeprof pprof dumps directory

shekhirin:
When jemalloc collects a pprof heap dump, it first writes it to a temp file, then reads it, and removes the file. That file is written to the `/tmp` directory that may not be accessible on certain setups, and this will result in such errors
```console
❯ curl localhost:6060/debug/pprof/heap
Failed to dump pprof: Read-only file system (os error 30) at path "/tmp/.tmpQpafXK"
```

This PR allows to configure the pprof dumps directory, and sets the default to `<DIR>/<CHAIN_ID>/pprof`.

The files are removed immediately once read, so we don't expect any disk usage growth.


crates/cli/commands/src/db/repair_trie.rs
```diff
@@ -53,11 +56,13 @@ impl Command {
         self,
         tool: &DbTool<N>,
         task_executor: TaskExecutor,
+        data_dir: &ChainPath<DataDirPath>,
     ) -> eyre::Result<()> {
         // Set up metrics server if requested
         let _metrics_handle = if let Some(listen_addr) = self.metrics {
             let chain_name = tool.provider_factory.chain_spec().chain().to_string();
             let executor = task_executor.clone();
+            let pprof_dump_dir = data_dir.data_dir().to_path_buf();
```
mediocregopher: Is `to_path_buf` right here? shouldn't it be `pprofs_dumps`?

crates/cli/commands/src/stage/run.rs
```diff
@@ -153,6 +153,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
                         }
                     })
                     .build(),
+                data_dir.data_dir().to_path_buf(),
```
mediocregopher: same here

mediocregopher (approved):

DaniPopes (approved):

yongkangc (approved):

PR #20837, opened 2026-01-08
docs: fix incorrect API reference and mermaid include in Pipeline docs

maximevtush:
Fixes documentation issues in `Pipeline` rustdoc comments:

 **Incorrect method reference**: Changed `Pipeline::set_max_block` → `PipelineBuilder::with_max_block`
   - The referenced method doesn't exist on `Pipeline`
   - The correct API is available via `PipelineBuilder`


mattsse (commented):
we can keep this as set_max_block

Issue #19515, opened 2025-11-05
Proof Calc Rewrite: Metrics and Traces


mediocregopher:
### Describe the feature

Introduces any useful metrics and traces/span which would be helpful for monitoring and debugging proof calculations. Audit any existing metrics and traces which were added in previous PRs to ensure they make sense in the context of subsequent changes.

PR #20838, opened 2026-01-08
feat(trie): Keep cached storage roots on proof workers

mediocregopher:
Towards #19515

Renames `missed_leaves_storage_roots`  to `storage_root_cache`, and populate it also for non-missed leaf cases via the storage workers.

This also changes how the cache is propagated; it is now held on the worker structs themselves rather than being propagated via the proof inputs. This simplifies the code a bit.

Related to RETH-166


mattsse (approved):
makes sense

yongkangc (approved):

PR #20842, opened 2026-01-08
chore(engine): remove state update batching in multiproof

mattsse:
Removes state update batching logic that was effectively unreachable and incorrect.

The batching was problematic for two reasons:

1. `can_batch_state_update` required matching sources and identical payloads for pre/post-block updates, making batching nearly impossible in practice.

2. The batching used `HashMap::extend` for merging state, which is incorrect because proper state merging requires checking existing entries and only updating changed storage slots, not blindly overwriting. See revm's `commit_transition` for the correct approach: https://github.com/bluealloy/revm/blob/2befb622978d29a70a5b73f8902a484d294f33de/crates/database/src/in_memory_db.rs#L193-L194

State updates are now processed directly via `on_state_update` without batching. Prefetch batching remains unchanged since it operates on proof targets (simple HashMap merge), not state.


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1433,66 +1354,6 @@ where
     1
 }
 
-/// Checks whether two state updates can be merged in a batch.
-///
-/// Transaction updates with the same transaction ID (`StateChangeSource::Transaction(id)`)
-/// are safe to merge because they originate from the same logical execution and can be
-/// coalesced to amortize proof work.
-fn can_batch_state_update(
```
mattsse, line 1441: this fortunately always returns false, so basically all of the state batching is a noop

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -959,80 +949,15 @@ impl MultiProofTask {
...
-
-                // Merge all accumulated updates into a single EvmState payload.
-                // Use drain to preserve the buffer allocation.
-                let mut accumulated_iter = ctx.accumulated_state_updates.drain(..);
-                let (mut batch_source, mut merged_update) = accumulated_iter
-                    .next()
-                    .expect("state update batch always has at least one entry");
-                for (next_source, next_update) in accumulated_iter {
-                    batch_source = next_source;
-                    merged_update.extend(next_update);
```
mattsse, line 1023: this was incorrect because this is using Hashmap::extend which might alter the aggregated map incorrectly

shekhirin (approved):

yongkangc (approved):
Thanks Matt, good find

PR #20843, opened 2026-01-08
chore(consensus): Remove associated type `Consensus::Error`

emhane:
Upstreams https://github.com/op-rs/op-reth/pull/543

Removes redundant `Error` associated type on `Consensus` trait


mattsse (approved):
imo this makes sense

consensus error has

https://github.com/paradigmxyz/reth/blob/b38d37a1e10bdd461a07e4e054859d315c69c01c/crates/consensus/consensus/src/lib.rs#L410-L412

which should be sufficient but perhaps we also add a dyn error variant here

wdyt @klkvr

PR #20844, opened 2026-01-08
refactor(engine): buffer state updates and process in parallel

figtracer:
this buffers state updates and processes all in parallel

less granularity on parallelization as compared to #20635 and following #20842 

avg transaction count per block seems to be around 300, so the lower bound of the benchmarks doesn't look concerning
 
benchmarks:
https://gist.github.com/figtracer/e4e146d1a9d5817139a7c25aebb0c9bd


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1010,10 +1061,39 @@ impl MultiProofTask {
...
+                        num_deferred,
+                        "Processing deferred state updates"
+                    );
+
+                    // Merge all EVM states into one, properly combining storage slots
+                    let merged_state = merge_evm_states(deferred_states);
+
+                    // Hash everything in parallel and process
+                    batch_metrics.state_update_proofs_requested += self.on_state_update(
+                        Source::Evm(StateChangeSource::Transaction(0)),
```
figtracer: this is not good maybe add another Source

PR #20845, opened 2026-01-08
perf(engine): avoid unnecessary B256 copy in get_proof_targets

mattsse:
Use reference iteration instead of destructuring pattern to avoid copying B256 (32 bytes) on each iteration. The copy only happens when actually needed for insertion into the targets map.

```rust
// Before: copies B256 on every iteration
for &hashed_address in state_update.accounts.keys() {
    if !fetched_proof_targets.contains_key(&hashed_address) {

// After: only copies when inserting
for hashed_address in state_update.accounts.keys() {
    if !fetched_proof_targets.contains_key(hashed_address) {
        targets.insert(*hashed_address, ...);
```


yongkangc (approved):
nice find

PR #20841, opened 2026-01-08
feat: add `edge` feature flag to reth

joshieDo:
Allows enabling all unstable features with the feature flag `edge` for tests/benchmarks/etc.

Right now, it enables all new storage features set by `StorageSettings::edge()`

blocked until the following fixes https://github.com/paradigmxyz/reth/pull/20846 https://github.com/paradigmxyz/reth/pull/20850

PR #20846, opened 2026-01-08
fix: initialize `transaction-senders` sf during genesis

joshieDo:
fixes [CI failure](https://github.com/paradigmxyz/reth/actions/runs/20820256303/job/59806550216) on https://github.com/paradigmxyz/reth/pull/20841 when running with all storage settings enabled

This has not been an issue because:
* tx senders in static files is always disabled
* if enabled we usually run pipeline stage first which calls `ensure_at_block`. This will initialize it, if its not. So, if pipeline is not started (eg. test, new chain), this fails when trying to append senders.

https://github.com/paradigmxyz/reth/blob/8eecad3d1d433ed509373713c21c31504290d17d/crates/stages/stages/src/stages/sender_recovery.rs#L89-L96 

https://github.com/paradigmxyz/reth/blob/8eecad3d1d433ed509373713c21c31504290d17d/crates/storage/provider/src/providers/static_file/writer.rs#L427-L433

### DIFF
--- a/crates/cli/commands/src/init_state/without_evm.rs
+++ b/crates/cli/commands/src/init_state/without_evm.rs
@@ -99,6 +99,7 @@ where
 /// * Headers: It will push an empty block.
 /// * Transactions: It will not push any tx, only increments the end block range.
 /// * Receipts: It will not push any receipt, only increments the end block range.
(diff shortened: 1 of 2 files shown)


Rjected (approved):
makes sense

PR #20847, opened 2026-01-08
perf: fix duplicated hash

livehardd:
Caching address hashes during storage unwind. no reason to hash the same address multiple times when processing bulk storage changes.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2675,19 +2675,18 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> HashingWriter for DatabaseProvi
         // Use HashMap to cache hashed addresses and avoid redundant keccak256 computation.
         let mut address_hash_cache = HashMap::new();
         let mut hashed_storages = Vec::new();
-        
+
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2672,12 +2672,22 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> HashingWriter for DatabaseProvi
...
+        for (BlockNumberAddress((_, address)), storage_entry) in changesets {
+            let hashed_address = *address_hash_cache
+                .entry(address)
+                .or_insert_with(|| keccak256(address));
+            
+            hashed_storages.push((
+                hashed_address,
+                keccak256(storage_entry.key),
+                storage_entry.value,
+            ));
```
mattsse: this is correct but unclear how beneficial this would be

mattsse (commented):
this is potentially beneficial cross blocks since we expect duplicate blocks here

since this is unwind, this isnt too critical, so we could either do this or not

wdyt @Rjected @joshieDo

livehardd:
@mattsse hi  mattsse, the bot closed this would you mind re-opening this?

PR #20849, opened 2026-01-08
docs(reth): fix outdated comments and document missing features

meetrick:
Comments and documentation in `bin/reth/src/lib.rs` were not updated when the source code changed over time:

- The `rpc::result` module comment still referenced `reth_rpc::rpc` (typo since 2023), but the actual code imports from `reth_rpc_server_types::result`
- Several default features (`otlp`, `js-tracer`, `keccak-cache-global`) were added in late 2025, but never documented
- Other features (`dev`, `jemalloc-symbols`, `tracy-allocator`, `snmalloc`, etc.) were also missing from documentation

This PR updates the comments to match the current codebase and improves readability by organizing feature flags into categories.


mattsse (approved):
makes sense

PR #20850, opened 2026-01-08
fix: pre-calculate transitions on `append_blocks_with_state` 

joshieDo:
fixes [CI failure](https://github.com/paradigmxyz/reth/actions/runs/20824783921/job/59823053445) on https://github.com/paradigmxyz/reth/pull/20841 when running with all storage settings enabled

`append_blocks_with_state` assumes that it can read `account changesets` after writing them, which is not the case with account changesets in static files (and storage in the future). This PR pre-calculates the transitions before writing to state, which in any case makes more sense than just checking what we just write.

Afaict, this method is only used in tests, so we should probably hide it behind `test-utils` ?


Rjected (approved):
yeah looks like it's only used in test-utils

PR #20852, opened 2026-01-08
feat(metrics): add block_size_bytes histogram to engine tree

shemnon:
Track in-memory size of validated blocks via new block_size_bytes metric, enabling visibility into block size distribution during Engine API calls.


mattsse (commented):
this seems alright, but I think I'd like to change a few things here first, because this has a tiny bit of overhead, that we can reduce by moving a few things around

PR #20853, opened 2026-01-08
feat: allow `TransactionHashNumbers` to be written to `rocksdb` during live sync

joshieDo:
We were only using `EitherWriter/EitherReader` during the stage. This PR:
* `EitherWriter/EitherReader` adds to other places like tests and live sync so it can be used with [reth-edge](https://github.com/paradigmxyz/reth/pull/20841)
* some helpers to make the whole `cfg` less ugly

Still missing the pruning bit - follow up

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -17,10 +17,10 @@ use crate::{
     DBProvider, EitherReader, EitherWriter, EitherWriterDestination, HashingWriter, HeaderProvider,
     HeaderSyncGapProvider, HistoricalStateProvider, HistoricalStateProviderRef, HistoryWriter,
     LatestStateProvider, LatestStateProviderRef, OriginalValuesKnown, ProviderError,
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -595,6 +620,25 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> DatabaseProvider<TX, N> {
...
+        let rocksdb = self.rocksdb_provider();
+        #[cfg(all(unix, feature = "rocksdb"))]
+        let rocksdb_tx = rocksdb.tx();
+        #[cfg(all(unix, feature = "rocksdb"))]
+        let rocksdb_tx_ref = &rocksdb_tx;
+        #[cfg(not(all(unix, feature = "rocksdb")))]
+        let rocksdb_tx_ref = ();
+
+        f(rocksdb_tx_ref)
+    }
```
Rjected: good fn

Rjected (approved):
lgtm

PR #20854, opened 2026-01-08
refactor(ecies): avoid duplicate keccak digest in MAC::update_body

radik878:
Removed redundant call to digest() inside MAC::update_body and reuses the already computed digest instead. The previous implementation cloned and finalized the Keccak256 state twice in a row without any intermediate updates, which produced the same value but added extra work on every frame. The new code keeps the exact same RLPx MAC semantics by using the single digest result both as AES input and for the XOR step, matching the specification while reducing unnecessary hashing overhead.

### DIFF
--- a/crates/net/ecies/src/mac.rs
+++ b/crates/net/ecies/src/mac.rs
@@ -57,7 +57,7 @@ impl MAC {
         self.hasher.update(data);
         let prev = self.digest();
         let aes = Aes256Enc::new_from_slice(self.secret.as_ref()).unwrap();
-        let mut encrypted = self.digest().0;
+        let mut encrypted = prev.0;
 
         aes.encrypt_padded::<NoPadding>(&mut encrypted, B128::len_bytes()).unwrap();
         for i in 0..16 {


crates/net/ecies/src/mac.rs
```diff
@@ -57,7 +57,7 @@ impl MAC {
         self.hasher.update(data);
         let prev = self.digest();
         let aes = Aes256Enc::new_from_slice(self.secret.as_ref()).unwrap();
-        let mut encrypted = self.digest().0;
+        let mut encrypted = prev.0;
```
mattsse, line 60: this is actually the same hash because unchanged

https://github.com/paradigmxyz/reth/blob/a06644944f7becb4a5e3e7c0a7fc37e3bd6d6361/crates/net/ecies/src/mac.rs#L69-L73

mattsse (approved):
lgtm

pending sanity check @Rjected

Rjected (approved):
yeah this looks the same to me

PR #20856, opened 2026-01-08
chore(exex): log lagging exex information

danyalprout:
### Description
We run multiple ExEx's on our node and occasionally get the WAL warning log. Updated the log to print out the ExEx which is responsible for the lag.


danyalprout:
Going to close this for now, it's the finalized head that's causing the lag not the exex, so this log line is technically incorrect.

PR #20857, opened 2026-01-08
refactor(net): remove incorrect `dead_code` lint suppression from `TrustedPeersResolver::set_interval`

0xxFloki:
Removes the incorrect #[allow(dead_code)] attribute from the TrustedPeersResolver::set_interval method, as this method is actively used in the codebase.


mattsse (approved):

mattsse:
actually this is required to make clippy happy

PR #20859, opened 2026-01-08
perf(engine): eliminate redundant allocation in block transaction iterator

0xlupin:
Removes an unnecessary intermediate Vec allocation when creating a parallel iterator over block transactions in the payload validator.


mattsse (approved):
this makes sense actually and is the recommended rayon way for slices

PR #20860, opened 2026-01-08
Simplify core trait bounds by dropping redundant PartialEq

0xcharry:
Removes duplicated PartialEq from NodePrimitives and SignedTransaction bounds (already implied by Eq), no logic changes; purely trait constraint cleanup, verified with cargo check -p reth-primitives-traits


mattsse (commented):
this doesnt hurt, we can keep this as is

PR #20861, opened 2026-01-08
feat(network): add customizable announcement filtering policy to APIs

shemnon:
Enable SDK chains to configure announcement filtering by exposing the AnnouncementFilteringPolicy as a parameter in the network builder API instead of hardcoding to strict filtering.

- Add transactions_with_policies() accepting both propagation and announcement policies
- Add start_network_with_policies() to BuilderContext for custom announcement filters
- Existing transactions_with_policy() defaults to StrictEthAnnouncementFilter

### DIFF
--- a/crates/net/network/src/builder.rs
+++ b/crates/net/network/src/builder.rs
@@ -3,7 +3,9 @@
 use crate::{
     eth_requests::EthRequestHandler,
     transactions::{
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense, doesnt hurt

PR #20863, opened 2026-01-08
perf(engine): destructure leaf to avoid clone()

figtracer:
eliminates one `Vec<u8>` clone per revealed leaf node

### DIFF
--- a/crates/trie/sparse/src/trie.rs
+++ b/crates/trie/sparse/src/trie.rs
@@ -546,40 +546,43 @@ impl SparseTrieInterface for SerialSparseTrie {
                     self.reveal_node_or_hash(child_path, &ext.child)?;
                 }
             },
-            TrieNode::Leaf(leaf) => match self.nodes.entry(path) {
-                Entry::Occupied(mut entry) => match entry.get() {
-                    // Replace a hash node with a revealed leaf node and store leaf node value.
-                    SparseNode::Hash(hash) => {
-                        let mut full = *entry.key();
-                        full.extend(&leaf.key);
-                        self.values.insert(full, leaf.value.clone());
-                        entry.insert(SparseNode::Leaf {
-                            key: leaf.key,
-                            // Memoize the hash of a previously blinded node in a new leaf
-                            // node.
-                            hash: Some(*hash),
-                        });
-                    }
-                    // Left node already exists.
-                    SparseNode::Leaf { .. } => {}
-                    // All other node types can't be handled.
-                    node @ (SparseNode::Empty |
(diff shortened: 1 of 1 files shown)


crates/trie/sparse/src/trie.rs
```diff
@@ -546,40 +546,43 @@ impl SparseTrieInterface for SerialSparseTrie {
...
+                                node: Box::new(node.clone()),
+                            }
+                            .into())
+                        }
+                    },
+                    Entry::Vacant(entry) => {
+                        let mut full = *entry.key();
+                        full.extend(&leaf_key);
+                        entry.insert(SparseNode::new_leaf(leaf_key));
+                        self.values.insert(full, leaf_value);
```
mattsse, line 582: ah cool, no need to clone the leaf value here

crates/trie/sparse/src/trie.rs
```diff
@@ -546,40 +546,43 @@ impl SparseTrieInterface for SerialSparseTrie {
...
-                    SparseNode::Leaf { .. } => {}
-                    // All other node types can't be handled.
-                    node @ (SparseNode::Empty |
-                    SparseNode::Extension { .. } |
-                    SparseNode::Branch { .. }) => {
-                        return Err(SparseTrieErrorKind::Reveal {
-                            path: *entry.key(),
-                            node: Box::new(node.clone()),
+            TrieNode::Leaf(leaf) => {
+                let (leaf_key, leaf_value) = (leaf.key, leaf.value);
```
mattsse, line 550: ah nice, that makes sense

crates/trie/sparse/src/trie.rs
```diff
@@ -546,40 +546,43 @@ impl SparseTrieInterface for SerialSparseTrie {
...
+                match self.nodes.entry(path) {
+                    Entry::Occupied(mut entry) => match entry.get() {
+                        // Replace a hash node with a revealed leaf node and store leaf node value.
+                        SparseNode::Hash(hash) => {
+                            let mut full = *entry.key();
+                            full.extend(&leaf_key);
```
mattsse, line 556: this is the unlock

mattsse (approved):
cool, lgtm
pending sanity review @mediocregopher (recommend hiding whitespace)


we have a similar pattern here:


https://github.com/paradigmxyz/reth/blob/a06644944f7becb4a5e3e7c0a7fc37e3bd6d6361/crates/trie/sparse-parallel/src/trie.rs#L1941-L1941

but this operates on &

mediocregopher (approved):

PR #20864, opened 2026-01-08
perf(engine): simplify get_prefetch_proof_targets

mattsse:
Simplifies `get_prefetch_proof_targets` by using the existing `MultiProofTargets::retain_difference` method instead of a two-phase approach with duplicate lookups. The logic is equivalent afaict.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -603,30 +603,31 @@ impl MultiProofTask {
         skip_all,
         fields(accounts = targets.len(), chunks = 0)
     )]
-    fn on_prefetch_proof(&mut self, targets: MultiProofTargets) -> u64 {
-        let proof_targets = self.get_prefetch_proof_targets(targets);
-        self.fetched_proof_targets.extend_ref(&proof_targets);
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
seems equivalent to me

crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -672,48 +672,15 @@ impl MultiProofTask {
...
-            // keep only the storage slots that have not been fetched yet
-            //
-            // we already removed subsets, so this should only remove duplicates
-            target_storage.retain(|slot| !fetched_storage.contains(slot));
-
-            duplicates += prev_target_storage_len - target_storage.len();
-        }
+        let before = targets.chunking_length();
+        targets.retain_difference(&self.fetched_proof_targets);
+        let duplicates = before - targets.chunking_length();
```
mediocregopher: I would be open to not computing `duplicated` and getting rid of the trace, but up to you

mediocregopher (approved):

crates/engine/tree/src/tree/payload_processor/multiproof.rs
mattsse: yeah, good point, lets get rid of this entirely then

PR #20865, opened 2026-01-08
perf(engine): add new hashes_buf to RlpNodeBuffers

figtracer:
we were allocating a new vec for every branch node processed. to fix this, we only reuse a buffer

### DIFF
--- a/crates/trie/sparse/src/trie.rs
+++ b/crates/trie/sparse/src/trie.rs
@@ -1613,7 +1613,7 @@ impl SerialSparseTrie {
 
                     let mut tree_mask = TrieMask::default();
                     let mut hash_mask = TrieMask::default();
-                    let mut hashes = Vec::new();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I believe this avoids over-allocations in the new vec, and uses a fixed alloc for drain+collect instead, so we keep the buf excess capacity

this makes sense to me, but unsure how impacful if at all really

pending @mediocregopher

mediocregopher (commented):
I'm not convinced this is more impactful than a `Vec::with_capacity` instead of the `Vec::new`, since the `drain(..).collect()` is going to allocate a Vec anyway. @figtracer am I missing something?

Going forward I would suggest not bothering with SerialSparseTrie, it's not enabled by default and it's likely we'll decide to stop supporting it sooner than later. Please focus on ParallelSparseTrie for optimization work.

PR #20866, opened 2026-01-08
perf(trie): optimize ChunkedHashedPostState sorting

mattsse:
Uses unstable sort instead of stable sort for `ChunkedHashedPostState` by encoding the ordering constraint in a dedicated `FlattenedStateOrder` enum. The enum's derived `Ord` ensures the required ordering (wipe < update < account), eliminating the need for stable sort to preserve insertion order.

by using an additional order key `FlattenedStateOrder`, we get unique keys for unstable sorting

### DIFF
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -741,11 +741,30 @@ pub struct ChunkedHashedPostState {
     size: usize,
 }
 
+/// Order discriminant for sorting flattened state items.
+/// Ordering: `StorageWipe` < `StorageUpdate` (by slot) < `Account`
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+enum FlattenedStateOrder {
+    StorageWipe,
+    StorageUpdate(B256),
+    Account,
+}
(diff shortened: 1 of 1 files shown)


crates/trie/common/src/hashed_state.rs
```diff
@@ -741,42 +741,60 @@ pub struct ChunkedHashedPostState {
...
                 (address, FlattenedHashedPostStateItem::Account(account))
             }))
-            // We need stable sort here to preserve the order for each address:
-            // 1. Storage wipes
-            // 2. Storage updates
-            // 3. Account update
-            .sorted_by_key(|(address, _)| *address);
+            // Sort by address, then by item order to ensure correct application sequence:
+            // 1. Storage wipes (must come first to clear storage)
+            // 2. Storage updates (apply new values after wipe)
```
mattsse: ~~@shekhirin this might end up reorder storage updates so I'm not entirely sure this is okay, if not we can add the slot to the order key~~

Fixed this by using the slot in the order key

shekhirin (approved):
nice, makes sense

PR #20867, opened 2026-01-09
feat(exex): Make WAL Block Threshold Configurable

refcell:
### Problem

The `WAL_BLOCKS_WARNING` threshold is hardcoded to `128` blocks, which is appropriate for Ethereum mainnet with ~12 second block times. However, L2 chains have significantly faster block times (often 2 seconds or less), causing the WAL to accumulate blocks much faster than on mainnet. This results in the following [ExExManager warning](https://github.com/paradigmxyz/reth/blob/1866db4d5/crates/exex/exex/src/manager.rs#L393-L399) appearing frequently even during normal operation:

```
WAL contains too many blocks and is not getting cleared. That will lead to increased disk space usage. Check that you emit the FinishedHeight event from your ExExes. blocks=...
```

This creates noise in logs and can mask real issues where ExExes are genuinely failing to emit `FinishedHeight` events.

### Solution

Make the WAL blocks warning threshold configurable via a builder pattern, allowing node operators to set an appropriate value based on their chain's block time.

### Usage

```rust
// Configure a higher threshold for L2 chains with faster block times
ctx.exex_launcher(installed_exexes)
    .with_wal_blocks_warning(768)  // 6x default for ~2 second blocks
    .launch()
    .await?;
```

Or when constructing `ExExManager` directly:

```rust
ExExManager::new(provider, handles, capacity, wal, finalized_stream).with_wal_blocks_warning(1024)
```

### Changeset

- Renamed `WAL_BLOCKS_WARNING` to `DEFAULT_WAL_BLOCKS_WARNING` and kept the default at 128
- Added `wal_blocks_warning` field to `ExExManager` with a `with_wal_blocks_warning()` builder method
- Added `wal_blocks_warning` field to `ExExLauncher` with a `with_wal_blocks_warning()` builder method
- Added `exex_launcher()` method that returns the launcher before calling `.launch()`, enabling custom configuration
- Updated the warning log to include the configured threshold for better observability

### DIFF
--- a/crates/exex/exex/src/manager.rs
+++ b/crates/exex/exex/src/manager.rs
@@ -38,11 +38,14 @@ use tokio_util::sync::{PollSendError, PollSender, ReusableBoxFuture};
 /// or 17 minutes of 1-second blocks.
 pub const DEFAULT_EXEX_MANAGER_CAPACITY: usize = 1024;
 
-/// The maximum number of blocks allowed in the WAL before emitting a warning.
+/// Default maximum number of blocks allowed in the WAL before emitting a warning.
 ///
-/// This constant defines the threshold for the Write-Ahead Log (WAL) size. If the number of blocks
-/// in the WAL exceeds this limit, a warning is logged to indicate potential issues.
-pub const WAL_BLOCKS_WARNING: usize = 128;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
hello ser!

nice, this lgtm!

PR #20868, opened 2026-01-09
feat(bench): add persistence-based flow optimization for rbc

yongkangc:
**Problem**
The current reth-bench implementation uses static wait times between block submissions, resulting in ~long rbc time for 1000 blocks (at 500ms wait). 

**Solution**
Implement an event-driven pipelined approach using WebSocket-based persistence notifications. The benchmark now tracks the gap between sent and persisted blocks, waiting only when the gap reaches a configurable threshold (default N=2).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7704,6 +7704,7 @@ dependencies = [
  "csv",
  "eyre",
  "futures",
+ "http",
  "humantime",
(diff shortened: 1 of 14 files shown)


yongkangc:
Recreating PR from main repo branch

PR #20869, opened 2026-01-09
feat(reth-bench-compare): add persistence-based flow optimization for reth-bench

yongkangc:
**Problem**
reth-bench-compare uses static wait times between block submissions, which is slow.

**Solution**
Wait for persistence to finish before sending the next payload. By default, persistence-based flow waits for every Nth block to be persisted (where N is the persistence threshold, default 2). Uses a new `reth_subscribeLatestPersistedBlock` WebSocket subscription to track persistence progress.

If `--wait-time` is provided with a non-zero value, uses the traditional fixed sleep intervals instead of persistence-based flow.

For block counts not evenly divisible by N, a final sync waits for the last block to persist before completing.

**Changes**
- Add `reth_subscribeLatestPersistedBlock` RPC subscription
- Emit `LatestPersistedBlock` event when blocks are persisted
- Default to persistence-based flow (wait every N blocks for persistence to catch up)
- Support `--wait-time` flag for traditional fixed delay mode
- Handle remainder blocks with final sync + timeout

**Expected Impact**
Faster benchmarks when using persistence-based flow (default), with option to fall back to traditional wait-time mode if needed.


bin/reth-bench/src/bench/output.rs
```diff
@@ -15,6 +15,37 @@ pub(crate) const COMBINED_OUTPUT_SUFFIX: &str = "combined_latency.csv";
 /// This is the suffix for new payload output csv files.
 pub(crate) const NEW_PAYLOAD_OUTPUT_SUFFIX: &str = "new_payload_latency.csv";
 
+/// This is the suffix for persistence stats csv files.
+pub(crate) const PERSISTENCE_STATS_SUFFIX: &str = "persistence_stats.csv";
+
+/// Statistics about persistence waits during the benchmark.
+#[derive(Debug)]
+pub(crate) struct PersistenceStats {
```
yongkangc: just for testing

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -1,26 +1,140 @@
...
 use std::time::{Duration, Instant};
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
+use url::Url;
+
+const PERSISTENCE_THRESHOLD: u64 = DEFAULT_PERSISTENCE_THRESHOLD;
+const PERSISTENCE_CHECKPOINT_TIMEOUT: Duration = Duration::from_secs(60);
+const PERSISTENCE_TIMEOUT: Duration = Duration::from_secs(300);
+
+struct PersistenceTracker {
```
yongkangc: this just for testing to remove

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -175,24 +314,51 @@ impl Command {
             return Err(error);
         }
 
+        // Final sync: wait for remaining blocks (persistence mode only)
```
shekhirin: do we need this? I think we can stop the bench right after last newPayload response

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -154,17 +275,35 @@ impl Command {
...
+                    wait_for_persistence(
+                        sub.stream_mut(),
+                        block_number,
+                        &mut tracker.last_persisted,
+                        PERSISTENCE_CHECKPOINT_TIMEOUT,
+                    )
+                    .await?;
+                    debug!(target: "reth-bench", persisted = tracker.last_persisted, "Persistence caught up");
+                }
+            }
```
shekhirin: should this be below `info!(&combined_result)` same as it was before?

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -213,4 +379,107 @@ impl Command {
...
+        let jwt_str = std::fs::read_to_string(&auth_jwt)?;
+        let jwt = JwtSecret::from_hex(&jwt_str)?;
+        let ws_url = self.derive_local_rpc_url()?;
+
+        let claims = alloy_rpc_types_engine::Claims::default();
+        let token = jwt.encode(&claims)?;
+
+        info!("Connecting to WebSocket at {} for persistence subscription", ws_url);
+
+        let ws_connect = WsConnect::new(ws_url.to_string()).with_auth(Authorization::Bearer(token));
```
shekhirin: @mattsse is there any way to do it easier?

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -11,16 +17,107 @@ use crate::{
...
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
+use url::Url;
+
+/// Wrapper that keeps both the subscription stream and the underlying provider alive.
+/// The provider must be kept alive for the subscription to continue receiving events.
+struct PersistenceSubscription {
+    _provider: RootProvider<Ethereum>,
+    stream: SubscriptionStream<BlockNumHash>,
+}
```
mattsse: this type does not belong here, this belongs at the end of the file

bin/reth-bench/src/bench/new_payload_fcu.rs
mattsse: we dont need auth for the subscription because is exposed on the regular rpc server, so we shouldnt need this

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -213,4 +379,107 @@ impl Command {
...
+                .parse()
+                .wrap_err_with(|| format!("Failed to parse WebSocket RPC URL: {ws_url}"))?;
+            info!(target: "reth-bench", ws_url = %parsed, "Using provided WebSocket RPC URL");
+            Ok(parsed)
+        } else {
+            let derived = engine_url_to_ws_url(&self.benchmark.engine_rpc_url)?;
+            debug!(target: "reth-bench", engine_url = %self.benchmark.engine_rpc_url, %derived, "Derived WebSocket URL");
+            Ok(derived)
+        }
+    }
```
mattsse: I understand why we need this but this is a bit hard to follow and could be simplified imo + properly documented why we need this.

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -11,16 +17,107 @@ use crate::{
...
+        Self { blocks_sent: 0, last_persisted: 0, last_block_number: 0 }
+    }
+
+    fn record_block(&mut self, block_number: u64) {
+        self.blocks_sent += 1;
+        self.last_block_number = block_number;
+    }
+
+    fn should_wait(&self) -> bool {
+        self.blocks_sent.is_multiple_of(PERSISTENCE_THRESHOLD + 1)
```
mattsse: this should be a config field, for when we want to experiment with different values

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -154,17 +275,35 @@ impl Command {
...
             };
 
-            // current duration since the start of the benchmark minus the time
-            // waiting for blocks
-            let current_duration = total_benchmark_duration.elapsed() - total_wait_time;
+            // Handle waiting based on mode
+            if let Some(wait_duration) = wait_time_mode {
+                tokio::time::sleep(wait_duration).await;
+            } else if let Some(ref mut sub) = persistence_sub {
+                tracker.record_block(block_number);
```
mattsse: I have a feeling that we might need this pattern again, so perhaps useful to make the persistencetracker (or waiter) include this as either duration or sub instead and we only need to invoke this as `waiter.on_block(block_number).await?;` here

I feel this would be cleaner

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -175,24 +314,51 @@ impl Command {
             return Err(error);
         }
 
+        // Final sync: wait for remaining blocks (persistence mode only)
+        if let Some(ref mut sub) = persistence_sub {
+            let remaining_gap = tracker.last_block_number.saturating_sub(tracker.last_persisted);
+
+            if remaining_gap > PERSISTENCE_THRESHOLD {
```
mattsse: same here

mattsse (changes requested):
some arch suggestions but feature wise this should be okay

bin/reth-bench/src/bench/new_payload_fcu.rs
yongkangc: makes sense. i think we can have that as a default for starters

bin/reth-bench/src/bench/new_payload_fcu.rs
yongkangc: removed auth

yongkangc: removed auth

bin/reth-bench/src/bench/new_payload_fcu.rs
yongkangc: moved to end

bin/reth-bench/src/bench/new_payload_fcu.rs
yongkangc: yup that does feel much cleaner, thanks for note

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -213,4 +276,278 @@ impl Command {
...
+            .map_err(|_| eyre::eyre!("Failed to set WSS scheme for URL: {url}"))?,
+        "ws" | "wss" => {}
+        scheme => {
+            return Err(eyre::eyre!(
+            "Unsupported URL scheme '{scheme}' for URL: {url}. Expected http, https, ws, or wss."
+        ))
+        }
+    }
+
+    ws_url.set_port(Some(8546)).map_err(|_| eyre::eyre!("Failed to set port for URL: {url}"))?;
```
mattsse, line 364: this should work, dont think we need this configurable

mattsse (approved):

shekhirin (approved):

PR #20871, opened 2026-01-09
test(storage): add parametrized MDBX/RocksDB history lookup equivalence tests

yongkangc:
**Problem**
#20543 added RocksDB history lookup tests, but they only tested the RocksDB backend. There was no verification that MDBX and RocksDB produce identical results for the same inputs.

**Solution**
Add parametrized tests that run identical scenarios against both backends and assert they return the same `HistoryInfo`, so its behaviorally equivalent.

**Changes**
- Add `run_account_history_scenario()` and `run_storage_history_scenario()` test helpers
- Add `test_account_history_info_both_backends` covering 4 scenarios (single shard, multiple shards, no history, pruning boundary)
- Add `test_storage_history_info_both_backends` covering 2 scenarios

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -10,7 +10,9 @@ use std::{
 #[cfg(all(unix, feature = "rocksdb"))]
 use crate::providers::rocksdb::RocksDBBatch;
 use crate::{
-    providers::{StaticFileProvider, StaticFileProviderRWRefMut},
+    providers::{
+        needs_prev_shard_check, HistoryInfo, StaticFileProvider, StaticFileProviderRWRefMut,
+    },
     StaticFileProviderFactory,
 };
 use alloy_primitives::{map::HashMap, Address, BlockNumber, TxHash, TxNumber};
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/either_writer.rs
```diff
@@ -1125,6 +1129,362 @@ mod rocksdb_tests {
...
+        lowest_available: Option<BlockNumber>,
+        expected: HistoryInfo,
+    }
+
+    /// Runs the same account history queries against both MDBX and `RocksDB` backends,
+    /// asserting they produce identical results.
+    fn run_account_history_scenario(
+        scenario_name: &str,
+        address: Address,
+        shards: &[(BlockNumber, Vec<BlockNumber>)], // (shard_highest_block, blocks_in_shard)
```
Rjected: if this is unclear then the tuple should be a struct

crates/storage/provider/src/either_writer.rs
```diff
@@ -1125,6 +1129,362 @@ mod rocksdb_tests {
...
+                i,
+                query.block_number,
+                query.lowest_available,
+                mdbx_result,
+                query.expected
+            );
+        }
+
+        rocks_tx.rollback().unwrap();
+        drop(temp_dir);
```
Rjected: does this need to be explicitly dropped?

crates/storage/provider/src/either_writer.rs
```diff
@@ -1125,6 +1129,362 @@ mod rocksdb_tests {
...
+        // Setup MDBX
+        let factory = create_test_provider_factory();
+        let mdbx_provider = factory.database_provider_rw().unwrap();
+        {
+            let tx = mdbx_provider.tx_ref();
+            let mut cursor = tx.cursor_write::<tables::AccountsHistory>().unwrap();
+            for (highest_block, blocks) in shards {
+                let key = ShardedKey::new(address, *highest_block);
+                let value = IntegerList::new(blocks.clone()).unwrap();
+                cursor.upsert(key, &value).unwrap();
```
joshieDo: lets use `eitherwriter`

crates/storage/provider/src/either_writer.rs
```diff
@@ -1125,6 +1129,362 @@ mod rocksdb_tests {
...
+            }
+        }
+        mdbx_provider.commit().unwrap();
+
+        // Setup RocksDB
+        let (temp_dir, rocks_provider) = create_rocksdb_provider();
+        for (highest_block, blocks) in shards {
+            let key = ShardedKey::new(address, *highest_block);
+            let value = IntegerList::new(blocks.clone()).unwrap();
+            rocks_provider.put::<tables::AccountsHistory>(key, &value).unwrap();
```
joshieDo: lets use `eitherwriter` (maybe only loop once with the above)

crates/storage/provider/src/either_writer.rs
```diff
@@ -1125,6 +1129,362 @@ mod rocksdb_tests {
...
+            };
+            let mdbx_state = HistoricalStateProviderRef::new_with_lowest_available_blocks(
+                &mdbx_ro,
+                query.block_number,
+                lowest_blocks,
+            );
+            let mdbx_result = mdbx_state.storage_history_lookup(address, storage_key).unwrap();
+
+            // RocksDB query
+            let rocks_result = rocks_tx
```
joshieDo: eitherreader

yongkangc:
@joshieDo just addressed your comments, thanks for review

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -204,7 +204,7 @@ impl<'b, Provider: DBProvider + ChangeSetReader + BlockNumReader>
         Ok(HashedStorage::from_reverts(self.tx(), address, self.block_number)?)
     }
 
-    fn history_info<T, K>(
```
yongkangc, line 207: wrapper to keep backwards compatiblilty

joshieDo (approved):

PR #20873, opened 2026-01-09
perf(trie): defer child RLP conversion in proof_v2 for async encoder support

meetrick:
## Summary

This PR implements the TODO comment in `proof_v2/mod.rs` that suggested deferring RLP conversion of non-retained children to optimize for async encoders.

**Before:**
```rust
// TODO: If this is not being retained as part of the proof, and is a Leaf or
// Extension node, then we don't need to commit to an RlpNode yet -- it's fine to
// push it back on the child stack as-is.
```

**After:** The TODO has been implemented and removed.

## What & Why
The `proof_v2` module uses `DeferredValueEncoder` trait to enable lazy RLP encoding of leaf values. The design intention was to allow async encoders to perform background work (e.g., storage root calculation) while the trie continues processing other nodes.

However, the previous implementation called `encode()` immediately in `commit_last_child()` whenever a new leaf was added. This meant the deferred encoder had virtually no time to complete any async work before being forced to block and produce the encoded value.

## How
The optimization defers RLP conversion for children that won't be retained as proof nodes:

1. **`commit_last_child()`**: Now checks `should_retain()` before converting. If the child doesn't need to be part of the proof, it's pushed back onto the stack as-is (Leaf, Extension, or Branch variant) instead of being immediately converted to `RlpNode`.

2. **`pop_branch()`**: When collecting children to build a branch node, any uncommitted children (non-RlpNode variants) are converted at this point using `into_rlp()`.

3. **Edge case handling**: Added a safety check for cached branches where `state_mask` is empty, which would otherwise cause a panic in `last_child_path()`.

## Impact

**Current reth:** This optimization has **no immediate performance impact** because all existing `DeferredValueEncoder` implementations are synchronous:
- `StorageDeferredValueEncoder`
- `SyncAccountDeferredValueEncoder`

**Future benefit:** When async encoders are introduced (e.g., for parallel storage root calculation), this optimization will provide measurable performance improvements without requiring additional code changes.

## Simulation Test
To validate the potential benefit of this optimization, I conducted a local simulation test:

I created a mock encoder that simulates async work by tracking the time between `deferred_encoder()` creation and `encode()` invocation. The encoder assumes a fixed 100µs async work duration per leaf. If `encode()` is called before the async work would complete, it counts the remaining time as "blocking time" that would have been spent waiting.

The test compared two scenarios:
- **Before (immediate)**: `encode()` called immediately after `deferred_encoder()` creation, simulating the old behavior
- **After (deferred)**: `encode()` called through `ProofCalculator` with the optimization, where conversion happens during `pop_branch()`

### Results
| Metric | Before | After |
|--------|--------|-------|
| Simulated async work | 100µs per encoder |
| Number of operations | 50 encode calls |
| Total blocking time | 4.95ms | 3.33ms |
| **Improvement** | — | **32.6% reduction** |

The deferred approach allows approximately 50µs of additional trie processing to occur between encoder creation and the `encode()` call, reducing effective blocking time when async work takes around 100µs to complete.

## Test Plan
- All existing `proof_v2` tests pass (9 tests including proptest with 8000 cases)
- All `reth-trie` tests pass (47 tests total)
- No compilation warnings
- Benchmark shows no performance regression for synchronous encoders


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -349,14 +349,26 @@ where
             return Ok(())
         }
 
+        // If state_mask is empty (e.g., cached branch), we can't compute child_path.
```
mediocregopher: Is this a case which you actually ran into as a result of the main changes? If not I would like to remove it, replacing with a `debug_assert_ne!(state_mask, 0)` if anything

mediocregopher (commented):
This looks good, basically what I was imagining the change would look like. Left one question about the extra check added to commit_last_child.

crates/trie/trie/src/proof_v2/mod.rs
meetrick: @mediocregopher While running the full test suite, I encountered cases where state_mask is empty during proptest_proof_with_targets (property-based test). This check handles that case.

mediocregopher: @meetrick I did some investigating and I would like to cherry-pick https://github.com/paradigmxyz/reth/commit/b52bb527dd4833ac0a9e61f89d8306c7a20dd684 onto this branch as a solution to this which is a bit cleaner I think, if you could do that we can merge :pray:

meetrick: @mediocregopher I went ahead and cherry-picked b52bb52 and added it to the PR.
As you mentioned, it does look cleaner now. I’ll use this as a reference for this kind of code style going forward!

meetrick: Done

mediocregopher (approved):

PR #20874, opened 2026-01-09
refactor: extract common TypedValueParser validation logic

ethfanWilliam:
The same UTF-8 validation and error formatting pattern repeated across 5 different TypedValueParser implementations. Pulled out the common bits into utility functions to make future clap API updates easier to handle.

1.Added value_parser_utils.rs for reth-node-core parsers
2.Added utils.rs for reth-cli parsers  
3.Refactored LogLevelValueParser, InvalidBlockSelectionValueParser, RpcModuleSelectionValueParser, ExtraDataValueParser, and chainspec Parser

### DIFF
--- a/crates/cli/cli/src/chainspec.rs
+++ b/crates/cli/cli/src/chainspec.rs
@@ -1,5 +1,6 @@
 use std::{fs, path::PathBuf, sync::Arc};
 
+use crate::utils;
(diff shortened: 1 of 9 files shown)


mattsse (commented):
we can keep this as is

PR #20876, opened 2026-01-09
feat(chain-state): add persisted block tracking

mattsse:
Adds tracking for the last persisted block in `ChainInfoTracker` and `CanonicalInMemoryState`. This enables subscribing to persistence events and querying the last persisted block.

Introduces `PersistedBlockSubscriptions` trait for subscribing to persisted block events, integrated into `FullProvider` with implementations for `BlockchainProvider` and `NoopProvider`.

Extracted from #20616.

cc @cakevm

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -317,6 +317,7 @@ impl<N: NodePrimitives> CanonicalInMemoryState<N> {
     /// This will update the links between blocks and remove all blocks that are [..
     /// `persisted_height`].
     pub fn remove_persisted_blocks(&self, persisted_num_hash: BlockNumHash) {
+        self.set_persisted(persisted_num_hash);
         // if the persisted hash is not in the canonical in memory state, do nothing, because it
         // means canonical blocks were not actually persisted.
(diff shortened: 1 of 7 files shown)


crates/chain-state/src/in_memory.rs
```diff
@@ -317,6 +317,7 @@ impl<N: NodePrimitives> CanonicalInMemoryState<N> {
     /// This will update the links between blocks and remove all blocks that are [..
     /// `persisted_height`].
     pub fn remove_persisted_blocks(&self, persisted_num_hash: BlockNumHash) {
+        self.set_persisted(persisted_num_hash);
```
shekhirin, line 320: should we do it in the end of this method, so that notification is sent when we finished updating the in-memory tree?

mattsse: don't think so, because this is independent to what is stored in mem, for the intended feature in rbc this also doesnt matter

shekhirin (approved):

yongkangc (approved):
makes sense where we add the block tracking and broadcast the updates and simpler

PR #20878, opened 2026-01-09
feat(metrics): Add metrics for save_block steps and computed trie input sizes

mediocregopher:
This PR adds metrics for the following:

* Each step of save_blocks, so we can see which steps are taking the most time.

* Total sizes of HashedPostStateSorted/TrieInputsSorted per-block.

* Total sizes of the anchored (flattened) overlay computed for each block, which is used as trie input for its child block.

The second two are computed and recorded async as part of the deferred trie task, so they won't affect nP latency.


yongkangc (approved):
lgmt, useful metrics

mattsse (approved):

PR #20880, opened 2026-01-09
fix(optimism): clarify pending sequence insert contract

Galoretka:
The previous comment on `FlashBlockPendingSequence::insert` claimed that a flashblock with index 0 “resets the set”, but the implementation does not clear any prior state. Actual reset and caching logic now lives in`SequenceManager::insert_flashblock`, which finalizes and clears the previous sequence before inserting a new index-0 block.

This change updates the documentation to reflect the real contract: index 0 is treated as the first element of a new sequence, and callers are responsible for finalizing or clearing any previous sequence before inserting it.


mattsse (commented):
imo the existing docs are clear enough

PR #20882, opened 2026-01-09
docs: fix MerkleStage naming in stages documentation

maximevtush:
The documentation incorrectly referenced `MerkleUnwindStage` and `MerkleExecuteStage` as separate stages, but these don't exist in the codebase. The actual implementation uses a single `MerkleStage` with two execution modes (unwind and execute), which return different `StageId` values (`MerkleUnwind` and `MerkleExecute`).


mattsse (commented):
these are actually called like this so we should keep this nomenclature

PR #20883, opened 2026-01-09
docs(engine): update outdated EthBuiltPayload comment

letmehateu:
The comment stated "the empty-block here is always available and full-block will be set/updated afterward", which implied that the struct stores both an empty block and a full block that can be updated in place. This was accurate when the struct contained a `_initial_empty_block: SealedBlock` field.


mattsse (approved):
this is indeed better

PR #20886, opened 2026-01-09
docs: clarify BatchTxProcessor local origin

MozirDmitriy:
The batcher is only used from the RPC path and always inserts transactions into the pool with TransactionOrigin::Local. The previous comment on  process_batch mentioned grouping by origin, which does not reflect the actual API or implementation and can mislead readers about supported use cases. This change updates the doc comment to state explicitly that the function processes a batch of local transaction requests, keeping the behavior unchanged but aligning the documentation with reality.


mattsse (commented):

PR #20887, opened 2026-01-09
chore: remove deprecated and unused ExecuteOutput struct

ethfanWilliam:
Removed the ExecuteOutput struct that was already marked as deprecated. The deprecation note explicitly recommended using BlockExecutionResult or BlockExecutionOutput instead. Since there are no references to this type, it's safe to remove and prevents accidental usage of the old API.

_By the way, got a question here, why don’t we remove this deprecated code directly and instead leave it commented out here?Is it because the process requires deprecated code to be commented out first?_


mattsse (approved):
this has been deprecated for long enough so we can remove


yes because some users that use this as dependency might rely on the types

PR #20893, opened 2026-01-09
refactor(prune): remove unnecessary Clone + Debug bounds from range parameters

Aleksandr1732:
Removes redundant trait bounds (Clone + Debug) from RangeBounds parameters in pruning functions. These bounds were not utilized in the implementation and unnecessarily restricted the API surface.


mattsse (commented):
we can keep this because this quite helpful for debugging

PR #20896, opened 2026-01-09
feat: add StaticFileSegment::StorageChangeSets

Rjected:
Adds storage changeset static files, which re-use the changeset offsets introduced in https://github.com/paradigmxyz/reth/pull/18882

The entries are sorted by address and key rather than just address.

Adds support with the `--static-files.storage-change-sets` flag

Closes RETH-174


gakonst (commented):
### Suggestion
Consider adding an equivalence test: same query set against MDBX-only vs static-file providers should produce identical results for:
- `storage_changesets_range(from..=to)`
- `get_storage_before_block(block, addr, key)`

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -2012,6 +2131,124 @@ impl<N: NodePrimitives> ChangeSetReader for StaticFileProvider<N> {
...
+        &self,
+        range: RangeInclusive<BlockNumber>,
+    ) -> ProviderResult<Vec<(BlockNumberAddress, StorageEntry)>> {
+        self.walk_storage_changeset_range(range).collect()
+    }
+
+    fn storage_changeset_count(&self) -> ProviderResult<usize> {
+        let mut count = 0;
+
+        let static_files = iter_static_files(&self.path).map_err(ProviderError::other)?;
```
joshieDo, line 2482: there should be a better way rather than read from disk again

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1922,17 +1997,14 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
         let mut storages_cursor = self.tx_ref().cursor_dup_write::<tables::PlainStorageState>()?;
-        let mut storage_changeset_cursor =
-            self.tx_ref().cursor_dup_write::<tables::StorageChangeSets>()?;
         for (block_index, mut storage_changes) in reverts.storage.into_iter().enumerate() {
             let block_number = first_block + block_index as BlockNumber;
 
             tracing::trace!(block_number, "Writing block change");
             // sort changes by address.
             storage_changes.par_sort_unstable_by_key(|a| a.address);
+            let mut changeset = Vec::new();
```
joshieDo: could we call with_capacity here

joshieDo (approved):
would like to review again once conflicts are in - but it seems to me that we might have some perf impact, because instead of just iterating over the _range, we're collecting it 

unsure if there is a better way to handle it tho, or if it's even a significant impact

crates/storage/provider/src/providers/database/provider.rs
Rjected: yep now uses with_capacity

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1270,13 +1271,58 @@ impl<TX: DbTx, N: NodeTypes> StorageChangeSetReader for DatabaseProvider<TX, N>
...
+                .transpose()
+        }
+    }
+
+    fn storage_changesets_range(
+        &self,
+        range: RangeInclusive<BlockNumber>,
+    ) -> ProviderResult<Vec<(BlockNumberAddress, StorageEntry)>> {
+        if self.cached_storage_settings().storage_changesets_in_static_files {
+            self.static_file_provider.storage_changesets_range(range)
```
joshieDo, line 1312: i think it makes sense,

but we do it differently on account_changesets_range : https://github.com/paradigmxyz/reth/blob/d22056d42f1bd659fea37ad2c9e72564684e88b7/crates/storage/provider/src/providers/database/provider.rs#L1368-L1398

unsure

Rjected: I prefer this way of doing it, I think we should rely on storage settings as the source of truth for where we should be checking, and not have the other checks for highest static files

joshieDo: agree

joshieDo (approved):

yongkangc (approved):

PR #20907, opened 2026-01-10
docs: fix broken Next Chapter link in reth-db crate docs

crypomen9:
Pointed a broken \"Next Chapter\" link to `eth-wire.md`, following the crate list order in `README.md`.


mattsse (commented):
we can keep as is

PR #20908, opened 2026-01-10
fix: typos in error messages and comments

sashaodessa:
Corrected `initiate` to `initialize` 
Corrected `Returns true if the this type` to `Returns true if this type`


mattsse (commented):
these are fine

Issue #20431, opened 2025-12-16
Improve transaction handling in payload processing


shekhirin:
There are several issues with how transactions are handled during payload processing that could be improved for correctness and efficiency.

#### 1. Incorrect transaction index in prewarm task
- [`crates/engine/tree/src/tree/payload_processor/prewarm.rs:183`](https://github.com/paradigmxyz/reth/blob/main/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L183)

The indexes used are not actual order indexes. The `pending` channel should send `(idx, tx)` tuples to preserve correct ordering.

#### 2. Missing `is_system_tx` method on transaction type
- [`crates/engine/tree/src/tree/payload_processor/prewarm.rs:185`](https://github.com/paradigmxyz/reth/blob/main/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L185)

The system transaction check (`tx.ty() > MAX_STANDARD_TX_TYPE`) is not accounting for transaction types introduced by Reth SDK chains. This should be exposed as an `is_system_tx()` method on the transaction type itself.

#### 3. Redundant transaction decoding
- [`crates/ethereum/payload/src/validator.rs:79`](https://github.com/paradigmxyz/reth/blob/main/crates/ethereum/payload/src/validator.rs#L79)

Transactions are already decoded by `tx_iterator` but not passed to `try_into_block_with_sidecar`, causing redundant decoding work.

ahsenkamal:
umm no, first two parts are done but I'm kind of stuck for the third part and would need some guidance to continue

gr8h:
Hey everyone, I'm new to Reth and wanted to try this out. Please check my solution and let me know if it makes sense.
@mattsse @Vui-Chee

PR #20911, opened 2026-01-10
fix(payload): send correct transaction index in prewarming

ahsenkamal:
Fixes first part of #20431 

- Fix incorrect transaction index in prewarm task
    - Included the original index of transactions from the iterator while passing them over channel by sending `(idx, tx)`

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -52,7 +52,7 @@ use tracing::{debug, debug_span, instrument, trace, warn, Span};
 /// Determines the prewarming mode: transaction-based or BAL-based.
 pub(super) enum PrewarmMode<Tx> {
     /// Prewarm by executing transactions from a stream.
-    Transactions(Receiver<Tx>),
+    Transactions(Receiver<(usize, Tx)>),
     /// Prewarm by prefetching slots from a Block Access List.
     BlockAccessList(Arc<BlockAccessList>),
 }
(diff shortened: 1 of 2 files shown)


crates/primitives-traits/src/transaction/signed.rs
mattsse, line 1: could you submit this separately first

crates/primitives-traits/src/transaction/signed.rs
```diff
@@ -48,6 +62,11 @@ pub trait SignedTransaction:
     + TxHashRef
     + IsTyped2718
 {
+    /// Returns whether this transaction is a system transaction
+    fn is_system_tx(&self) -> bool {
+        self.ty() > MAX_STANDARD_TX_TYPE
```
mattsse: this should by default return false

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -181,8 +167,8 @@ where
             let handles = ctx.clone().spawn_workers(workers_needed, &executor, actions_tx.clone(), done_tx.clone());
 
             // Distribute transactions to workers
-            let mut tx_index = 0usize;
-            while let Ok(tx) = pending.recv() {
+            let mut total_tx_executed = 0usize;
+            while let Ok((tx_index, tx)) = pending.recv() {
```
mattsse, line 185: I dont think we need the index change here, so let's keep this as is

mattsse (changes requested):

crates/primitives-traits/src/transaction/signed.rs
ahsenkamal: opened #21407

crates/primitives-traits/src/transaction/signed.rs
ahsenkamal: done

crates/engine/tree/src/tree/payload_processor/prewarm.rs
ahsenkamal: It was mentioned in the issue, am I missing something?

PR #20917, opened 2026-01-10
fix: dead documentation link

viktorking7:
* Updated the `snmalloc` repository link to the correct **microsoft/snmalloc** URL in `bin/reth/src/lib.rs`.


mattsse (commented):
we can keep this as is

PR #20918, opened 2026-01-10
chore: remove unused field

ethfanWilliam:
Cleaned up unused field in test helper struct. The chain_spec field was marked with #[allow(dead_code)] and was never accessed in any tests, only the provider and from_api fields are actually used.


mattsse (commented):
we can keep this as is, it's possible we'll need this for future test and the arc clone doesnt hurt here

PR #20919, opened 2026-01-10
chore(trie): save clock reads per block

figtracer:
same as #20916, time is measured more accurately too but not that relevant

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -1133,12 +1133,16 @@ impl MultiProofTask {
             }
         }
 
+        let now = Instant::now();
+        let total_time = ctx.first_update_time.map(|t| now.duration_since(t));
+        let time_since_updates_finished = ctx.updates_finished_time.map(|t| now.duration_since(t));
+
         debug!(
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -1133,12 +1133,16 @@ impl MultiProofTask {
             }
         }
 
+        let now = Instant::now();
+        let total_time = ctx.first_update_time.map(|t| now.duration_since(t));
+        let time_since_updates_finished = ctx.updates_finished_time.map(|t| now.duration_since(t));
```
mattsse, line 1138: this is now invoked outside the debug! macro, which I think could be redundant depending on the enabled tracing features, so I'm not exactly sure this is beneficial in _all_ cases

we could wrap this in a

if tracing::enabled!(tracing::Level::DEBUG) {

but perhaps at this point not worth doing

figtracer: i think it has a bit less overhead for most cases since these metrics are always recorded

maybe time is also measured more accurately since we're only using `now` as a ref; i could be wrong though

PR #20921, opened 2026-01-10
fix(libmdbx): remove duplicate assertion

letmehateu:
Removes a duplicate assertion in `test_iter_empty_dup_database` test that was checking the same iterator method twice.


mattsse (commented):
we can keep this

PR #20922, opened 2026-01-10
docs: add documentation for decompression size limit constants

crStiv:
Adds detailed documentation to MAX_DECOMPRESSED_SIGNED_BEACON_BLOCK_BYTES and 
MAX_DECOMPRESSED_BEACON_STATE_BYTES constants explaining their purpose as DoS 
protection mechanisms and the rationale behind the chosen values.


mattsse (commented):
dont think these add additional value

PR #20925, opened 2026-01-11
Makefile: POSIX compliant IO redirect

ahsenkamal:
Solves #20912 

Replaced `&>` operator with POSIX compliant `> ___ 2>&1` that runs properly on sh/dash


mattsse (commented):
these has been working well in ci

PR #20926, opened 2026-01-11
fix(rpc): correct parameter name in MinerApi::set_gas_limit

davidjsonn:
Fix copy-paste error in `set_gas_limit` implementation where the parameter was incorrectly named `_gas_price` instead of `_gas_limit`. This aligns the implementation with the trait definition in `rpc-api`.


mattsse (commented):
we can keep this

PR #20927, opened 2026-01-11
chore: replace io::copy with write_all for Bytes writing

DeVikingMark:
### Description
Replaces io::copy with write_all for writing Bytes chunks. write_all is more appropriate for byte slices.


mattsse (approved):
this makes sense here

PR #20928, opened 2026-01-11
refactor(prune): return actual deleted count from prune_table_with_range

davidjsonn:
Changed `prune_table_with_range` to return the count of actually deleted entries instead of processed entries. Previously, the counter was incremented even when entries were skipped by the filter.

Inlined `prune_table_with_range_step` logic and removed the now-unused function.


mattsse (commented):

PR #20930, opened 2026-01-11
perf(payload): remove unnecessary parent_header clone

davidjsonn:
Reorder operations to avoid cloning parent_header before wrapping it in Arc. The hash is now extracted first, then the header is moved into Arc instead of being cloned. Saves one full header allocation per payload job.


mattsse (approved):
makes sense

PR #20931, opened 2026-01-11
fix(optimism): add missing Holocene hardfork to DEV_HARDFORKS

davidjsonn:
DEV_HARDFORKS was missing OpHardfork::Holocene between Granite and Prague. All other hardfork configurations (OP_MAINNET, OP_SEPOLIA, BASE_MAINNET, BASE_SEPOLIA) include Holocene in the correct position. This caused dev chains to skip Holocene activation, breaking any Holocene-specific logic like the base fee calculation.


mattsse (approved):
this makes sense

PR #20932, opened 2026-01-11
fix(net): use saturating_sub consistently in transaction fetcher

dennsikl:
Use `saturating_sub` for budget decrement in `fill_request_from_hashes_pending_fetch` (line 740) to match the same pattern already used in `find_any_idle_fallback_peer_for_any_pending_hash` (line 223).


mattsse (commented):
this isnt necessary because we always check for 0

PR #20933, opened 2026-01-11
fix(metrics): correct section comment for UnboundedMeteredReceiver

avorylli:
The section divider comment incorrectly referenced MeteredReceiver instead of UnboundedMeteredReceiver. 

Updated to match the actual  impl block that follows.


mattsse (commented):
we can keep this as is

PR #20935, opened 2026-01-11
fix: remove unused random bytes generation in test

forkfury:
### Fix
Removed unused random bytes generation in test_chain_bincode_roundtrip.
### Changes
- Removed unused bytes array and rand::rng().fill() call
- Removed unused use rand::Rng; import
The test creates ExecutionOutcome with default values, so the random data generation was unnecessary leftover code.


mattsse (commented):

PR #20937, opened 2026-01-11
fix(trie): Update branch masks when revealing blinded nodes

mediocregopher:
The method `SerialSparseTrie::reveal_node` automatically adds the revealed node's masks to the `branch_node_masks` map. However in ParallelSparseTrie this cannot be done, because `reveal_node` is implemented on the SparseSubtrie type and the `branch_node_masks` is held on the higher-level ParallelSparseTrie.

The regular `ParallelSparseTrie::reveal_node` method handles this correctly, however when revealing blinded nodes the PST was incorrectly assuming that `SparseSubtrie::reveal_node` would update the branch masks. This results in missing tree mask bits in certain cases.

The fix here is to manually add masks to `branch_node_masks` when revealing blinded nodes in PST.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -6922,7 +6922,7 @@ mod tests {
             },
         ];
 
-        // Create a sparse trie and reveal nodes with the specific structure from prompt.md
+        // Create a sparse trie and reveal nodes
         let mut trie = ParallelSparseTrie::default()
             .with_root(
                 TrieNode::Extension(ExtensionNode {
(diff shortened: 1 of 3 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -436,7 +448,11 @@ impl SparseTrieInterface for ParallelSparseTrie {
 
             // If we didn't update the target leaf, we need to call update_leaf on the subtrie
             // to ensure that the leaf is updated correctly.
-            subtrie.update_leaf(full_path, value, provider, retain_updates)?;
+            if let Some((revealed_path, revealed_masks)) =
```
mattsse, line 451: oh I see, that makes sense to me

mattsse (approved):
cool, lgtm pending @shekhirin

shekhirin (approved):

PR #20938, opened 2026-01-11
Remove dead test-only helpers from storage state providers

0xcharry:
Drop unused test assertions in historical.rs that were silencing dead_code, remove empty test module in latest.rs to avoid needless dead_code suppression


mattsse (commented):
we can keep this

PR #20939, opened 2026-01-11
docs(rpc): fix incorrect transport in with_ipc comment

letmehateu:
The comment incorrectly stated that the method configures the http transport, but it sets the module for the ipc transport.


mattsse (approved):
sure

PR #20941, opened 2026-01-11
perf: remove iteration in pooled transaction getters

ethfanWilliam:
`pooled_transactions_hashes()` and `pooled_transactions_hashes_max()` were both iterating the pool independently. Same with the non-hash versions. Wasn't making much sense to duplicate all that filter+map logic when we can just call the _max variant with usize::MAX.

Updated all four methods to use the _max implementations.

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -346,16 +346,12 @@ where
 
     /// Returns transactions in the pool that can be propagated
     pub fn pooled_transactions(&self) -> Vec<Arc<ValidPoolTransaction<T::Transaction>>> {
-        let mut out = Vec::new();
-        self.append_pooled_transactions(&mut out);
-        out
+        self.pooled_transactions_max(usize::MAX)
     }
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
can you clarify which iteration this removes?

crates/transaction-pool/src/pool/mod.rs
```diff
@@ -346,16 +346,12 @@ where
 
     /// Returns transactions in the pool that can be propagated
     pub fn pooled_transactions(&self) -> Vec<Arc<ValidPoolTransaction<T::Transaction>>> {
-        let mut out = Vec::new();
-        self.append_pooled_transactions(&mut out);
-        out
+        self.pooled_transactions_max(usize::MAX)
```
mattsse, line 349: I see now, the reason why they appear duplicated is that this is slightly more efficient because we don't need to use the `.take()` combinator which here would then do redundant count checks.

so we can keep this as is

ethfanWilliam: Oh my bad

PR #20945, opened 2026-01-12
feat: integrate TrieDB for optimized state root calculation

cliff0412:
## Summary

  This PR integrates [Base's TrieDB](https://github.com/base/triedb) into reth for optimized state root calculation during block production. TrieDB is a memory-mapped trie database with page design optimized for MPT (Merkle Patricia Trie) structure, unlike MDBX which uses a binary tree page structure (which is for general kv store).

  ### Key Changes

  - Add `triedb` feature flag to enable TrieDB support
  - Integrate TrieDB into `ProviderFactory` architecture via `TrieDBProvider`
  - Implement state persistence to TrieDB after block production
  - Add support for `init_genesis` with TrieDB
  - Add microbenchmarks and metrics for state root calculation

  ### Performance Improvements

  | Dataset | TPS Improvement | State Root Speedup |
  |---------|-----------------|-------------------|
  | Small (1GB genesis) | **+20.58%** | **3.18x** |
  | Medium (3GB genesis) | **+23.86%** | **6.27x** |

  **Microbenchmarks** show up to **131x speedup** for isolated state root calculations (1,000 account overlay: 416ms → 3.17ms).

  ### Why TrieDB is Faster

  CPU profiling reveals that without TrieDB, the `TrieUpdates::extend_from_sorted` functions become a bottleneck when `PERSISTENCE_THRESHOLD` increases (6x overhead increase from threshold=2 to threshold=32).

  TrieDB **eliminates this bottleneck entirely** by:
  - Merging plain state (addresses, storage keys) instead of pre-computed trie nodes, eliminating `extend_from_sorted`
  - Reducing overall CPU usage by 20-22% compared to MDBX

  For detailed benchmarks and profiling analysis, see [docs/performance/perf_triedb.md](https://github.com/okx/reth/blob/feature/cliff/triedb/docs/performance/perf_triedb.md).

  ## Test Plan

  - [x] Run microbenchmarks: `cargo bench -p reth-provider --features "test-utils,triedb" --bench bench_state_root_overlay`
  - [x] Run integration tests with `triedb` feature enabled

  ## Related

  - TrieDB repository: https://github.com/base/triedb
  - TrieDB genesis performance issue: https://github.com/base/triedb/issues/179


gakonst:
We should explore replacing what remains of MDBX's Hashed tables and Trie tables when replaced with this? That should be compatible with our current roadmap around sparse trie and proofs v2, no? @mediocregopher

PR #20946, opened 2026-01-12
docs: fix incorrect panic documentation in Transaction::from_compact

bobtajson:
Fixes misleading panic documentation in `Transaction::from_compact` that incorrectly referenced Optimism-specific functionality.


mattsse (commented):
we can keep this as is

PR #20950, opened 2026-01-12
docs: align replace_.. docs in TransportRpcModules with real behavior

lisenokdonbassenok:
Updated the documentation of TransportRpcModules::replace_http, replace_ws, replace_ipc and replace_configured to match their actual semantics. The comments previously claimed that these helpers fail when the method being replaced is missing and that the boolean return value indicates whether any replacement occurred, which was not true. The docs now describe that the old methods are always removed if present, new methods are merged, errors only come from the merge step, and replace_configured always returns Ok(true) on success.


mattsse (commented):

PR #20951, opened 2026-01-12
docs: update run command

letmehateu:
Switch the run command to the actual workspace package example-full-contract-state (the example isn’t built via --example).


mattsse (commented):
we can keep as is

PR #20952, opened 2026-01-12
perf(engine): save clock reads and improve consistency in multiproof metrics

BearSoltokien:
Reduces clock reads from four to one when recording multiproof metrics, ensures all timing measurements use the same end timestamp.


figtracer:
this is a dup of #20919

mattsse (commented):

PR #20953, opened 2026-01-12
chore: export OpBuilderConfig types from optimism payload root

ethfanWilliam:
exports OpBuilderConfig, OpDAConfig, and OpGasLimitConfig from optimism/payload rootbefore:
`use reth_optimism_payload::config::{OpBuilderConfig, OpDAConfig, OpGasLimitConfig};`

after:
`use reth_optimism_payload::{OpBuilderConfig, OpDAConfig, OpGasLimitConfig};`

Matches how ethereum/payload already works with pub use config::*. makes the API cleaner and consistent across modules.

### DIFF
--- a/crates/optimism/payload/src/lib.rs
+++ b/crates/optimism/payload/src/lib.rs
@@ -28,6 +28,7 @@ pub mod validator;
 pub use validator::OpExecutionPayloadValidator;
 
 pub mod config;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this code will be moved out so we can keep as is

PR #20954, opened 2026-01-12
perf(eth-consensus): drop useless receipt alloc

rejected-l:
Removed an unnecessary O(N) allocation and RLP encoding of receipts that happened during receipt root validation failures. The code was eagerly serializing all receipts into a vector just for a debug log, which creates useless overhead and is a potential DoS vector.


mattsse (commented):
we want this for error reporting

PR #20955, opened 2026-01-12
fix(engine): preserve payload transactions when building reorg block

andrewshab3:
When creating a reorg block in `create_reorg_head`, the function was incorrectly overwriting `candidate_transactions` with transactions from the reorg target block during the ancestor walk loop.


mattsse (commented):
we can keep this as is

PR #20959, opened 2026-01-12
docs(db-common): clarify init-from-state-dump expectations and checkpoints

kilavvy:
Clarifies `init_from_state_dump` documentation in `reth-db-common` and adds a small test to lock in the expected state-dump format.

- Document that the **first line must be** the `{ "root": ... }` JSON object (not “can be”)
- Clarify that computing the state root writes trie tables but **does not** advance stage checkpoints implicitly (checkpoints are set explicitly afterwards)
- Add a unit test to ensure `parse_state_root` errors if the first line is not a root line

The previous comments did not match the actual behavior, which can mislead users/integrators when debugging `init-state` workflows.

### DIFF
--- a/crates/storage/db-common/src/init.rs
+++ b/crates/storage/db-common/src/init.rs
@@ -493,7 +493,7 @@ where
 
     let expected_state_root = header.state_root();
 
-    // first line can be state root
+    // First line must be the expected state root.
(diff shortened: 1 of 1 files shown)


mattsse (commented):

PR #20964, opened 2026-01-12
feat(db): implement extra dup methods

meyer9:
Implements `last_dup` and `prev_dup` which help with scanning dup items in reverse. We use this in the Proofs ExEx for finding the last entry before a certain block number.


dhyaniarun1993 (approved):

mattsse (approved):
makes sense

PR #20971, opened 2026-01-12
feat(reth-bench): create gas ramp-up bench command

Rjected:
This is a reth-bench command that sends empty blocks to the node, so that we can bump the gaslimit before sending benchmark data

### DIFF
--- a/bin/reth-bench/src/bench/gas_limit_ramp.rs
+++ b/bin/reth-bench/src/bench/gas_limit_ramp.rs
@@ -1,0 +1,214 @@
+//! Benchmarks empty block processing by ramping the block gas limit.
+
+use crate::{
+    authenticated_transport::AuthenticatedTransportConnect,
+    bench::{
+        helpers::{build_payload, prepare_payload_request, rpc_block_to_header},
+        output::{
+            write_benchmark_results, CombinedResult, NewPayloadResult, TotalGasOutput, TotalGasRow,
+        },
+    },
+    valid_payload::{call_forkchoice_updated, call_new_payload, payload_to_new_payload},
+};
+use alloy_eips::BlockNumberOrTag;
+use alloy_provider::{network::AnyNetwork, Provider, RootProvider};
+use alloy_rpc_client::ClientBuilder;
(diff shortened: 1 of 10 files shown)


bin/reth-bench/src/bench/gas_limit_ramp.rs
```diff
@@ -0,0 +1,267 @@
...
+        let is_optimism = !provider
+            .get_code_at(address!("0x420000000000000000000000000000000000000F"))
+            .await?
+            .is_empty();
+
+        if is_optimism {
+            return Err(eyre::eyre!(
+                "gas-limit-ramp is only supported for Ethereum chains (non-Optimism)"
+            ));
+        }
```
mattsse: I think we can ignore optimism support for this for now

bin/reth-bench/src/bench/gas_limit_ramp.rs
```diff
@@ -0,0 +1,267 @@
...
+        // Fetch the current head block as parent
+        let parent_block = provider
+            .get_block_by_number(BlockNumberOrTag::Latest)
+            .full()
+            .await?
+            .ok_or_else(|| eyre::eyre!("Failed to fetch latest block"))?;
+
+        let (mut parent_header, mut parent_hash) = rpc_block_to_header(parent_block)?;
+
+        validate_post_merge_header(&parent_header)?;
```
mattsse: this should be part of the node internally already so we can ignore?

bin/reth-bench/src/bench/helpers.rs
```diff
@@ -0,0 +1,259 @@
...
+};
+use eyre::OptionExt;
+use reth_chainspec::{
+    ChainSpec, EthereumHardforks, NamedChain, DEV, HOLESKY, HOODI, MAINNET, SEPOLIA,
+};
+use reth_node_api::EngineApiMessageVersion;
+use std::sync::Arc;
+
+/// Map a chain ID to a known chain spec, if available.
+pub(crate) fn chain_spec_from_id(chain_id: u64) -> Option<Arc<ChainSpec>> {
```
mattsse: we can make this a function for the Chainspec type itself

PR #20972, opened 2026-01-12
fix(stages): use static files for unwind in SenderRecovery stage

yongkangc:
**Summary**

The SenderRecovery stage unwind now correctly handles static files when senders are stored there.

**Problem**

Previously, the `SenderRecovery` stage unwind only removed entries from the database `TransactionSenders` table using `unwind_table_by_num`, but did not handle static files when senders are configured to be stored there. This could leave stale data in static files after an unwind.

Use `EitherWriter::prune_senders` which handles both database and static file storage destinations, consistent with how unwind is handled elsewhere in the codebase (e.g., `DatabaseProvider::remove_block_data_above`).

### DIFF
--- a/crates/stages/stages/src/stages/sender_recovery.rs
+++ b/crates/stages/stages/src/stages/sender_recovery.rs
@@ -161,7 +161,6 @@ where
             .ok_or(ProviderError::BlockBodyIndicesNotFound(unwind_to))?
             .next_tx_num();
 
-        // Prune senders from both database and static files
         EitherWriter::new_senders(provider, unwind_to)?.prune_senders(unwind_tx_from, unwind_to)?;
 
         Ok(UnwindOutput {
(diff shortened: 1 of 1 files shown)


crates/stages/stages/src/stages/sender_recovery.rs
```diff
@@ -155,12 +155,14 @@ where
...
-        let latest_tx_id = provider
+        // Lookup the next tx id after unwind_to block (first tx to remove)
+        let unwind_tx_from = provider
             .block_body_indices(unwind_to)?
             .ok_or(ProviderError::BlockBodyIndicesNotFound(unwind_to))?
-            .last_tx_num();
-        provider.tx_ref().unwind_table_by_num::<tables::TransactionSenders>(latest_tx_id)?;
+            .next_tx_num();
+
+        // Prune senders from both database and static files
```
joshieDo:
```suggestion
```

joshieDo (approved):

PR #20974, opened 2026-01-12
perf(db): throttle metrics reporting

DaniPopes:
db and static files metrics iterate over a bunch of entries, which is pretty expensive. throttle only these metrics to improve overall performance.

fixes #20966

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -1264,12 +1264,12 @@ pub fn metrics_hooks<N: NodeTypesWithDB>(provider_factory: &ProviderFactory<N>) 
     Hooks::builder()
         .with_hook({
             let db = provider_factory.db_ref().clone();
-            move || throttle!(Duration::from_mins(5), || db.report_metrics())
+            move || throttle!(Duration::from_secs(5 * 60), || db.report_metrics())
         })
         .with_hook({
             let sfp = provider_factory.static_file_provider();
             move || {
-                throttle!(Duration::from_mins(5), || {
+                throttle!(Duration::from_secs(5 * 60), || {
                     if let Err(error) = sfp.report_metrics() {
                         error!(%error, "Failed to report metrics from static file provider");
                     }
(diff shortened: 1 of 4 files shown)


crates/tracing/src/throttle.rs
```diff
@@ -0,0 +1,78 @@
...
+macro_rules! throttle {
+    ($duration:expr, || $expr:expr) => {{
+        static START: ::std::sync::LazyLock<::std::time::Instant> =
+            ::std::sync::LazyLock::new(::std::time::Instant::now);
+        static LAST: ::core::sync::atomic::AtomicU64 =
+            ::core::sync::atomic::AtomicU64::new($crate::__private::THROTTLE_NOT_YET_RUN);
+
+        if $crate::__private::should_throttle(&START, &LAST, $duration.as_millis() as u64) {
+            $expr
+        }
```
mattsse: this makes sense

crates/node/builder/src/launch/common.rs
```diff
@@ -1266,6 +1259,26 @@ where
     head: Head,
 }
 
+/// Returns the metrics hooks for the node.
+pub fn metrics_hooks<N: NodeTypesWithDB>(provider_factory: &ProviderFactory<N>) -> Hooks {
+    Hooks::builder()
+        .with_hook({
+            let db = provider_factory.db_ref().clone();
+            move || throttle!(Duration::from_mins(5), || db.report_metrics())
+        })
```
mattsse: cool

crates/node/builder/src/launch/common.rs
```diff
@@ -1266,6 +1259,26 @@ where
     head: Head,
 }
 
+/// Returns the metrics hooks for the node.
+pub fn metrics_hooks<N: NodeTypesWithDB>(provider_factory: &ProviderFactory<N>) -> Hooks {
+    Hooks::builder()
+        .with_hook({
+            let db = provider_factory.db_ref().clone();
+            move || throttle!(Duration::from_secs(5 * 60), || db.report_metrics())
```
mattsse, line 1267: every 5mins seems more than appropriate

mattsse (approved):

PR #20976, opened 2026-01-12
chore(db): remove unused `set_up_db` helper from bench utils

marukai67:
Removes dead code from database benchmark utilities.


mattsse (commented):

PR #20977, opened 2026-01-13
fix: correct comment references and remove duplicate in codecs derive

gap-editor:
Fixed documentation comments and a code typo in the codecs derive macro.

### Changes
- **generator.rs**: Swapped incorrect doc comments between `generate_from_compact` and `generate_to_compact` functions
- **mod.rs**: Removed duplicate "Address" entry in the `should_use_alt_impl` type array


mattsse (commented):
we can keep this as is

PR #20978, opened 2026-01-13
fix: add missing AccountsHistory and StoragesHistory to RocksDB metrics

HuberyJulianay:
The metrics setup was only tracking TransactionHashNumbers but we actually have 3 tables in RocksDB (AccountsHistory and StoragesHistory too). This would panic on the expect() call when trying to record metrics for the other two tables.

Just added them to the array.


mattsse (approved):
I believe this is correct @yongkangc ?

PR #20980, opened 2026-01-13
perf(static-file): save clock reads in writer commit methods

BearSoltokien:
Reduces clock reads from two to one in both commit methods, ensures metrics and debug logs use the same timestamp.


mattsse (commented):
we can keep this for debug

PR #20982, opened 2026-01-13
fix: fix RocksDB initialization in rpc-db example

HuberyJulianay:
The rpc-db example wasn't registering column families when creating RocksDBProvider. This would cause runtime errors if the code tried to access TransactionHashNumbers, AccountsHistory, or StoragesHistory tables.


mattsse (commented):
we can keep this as is

PR #20983, opened 2026-01-13
feat(debug): add `debug_setFailpoint` for runtime fault injection testing

AshinGau:
## Summary

This PR introduces runtime fault injection support via the `fail` crate, enabling crash recovery testing for reth's multi-storage architecture. Since reth now uses multiple storage backends (MDBX, RocksDB, and Static Files) without cross-system atomic transactions, this feature allows testing and validating the consistency recovery mechanisms when crashes occur between storage system commits.

## Motivation

Reth's storage layer has evolved to use three different storage systems:
- **MDBX**: Primary database for state data (transactional)
- **RocksDB**: Used for history indices (https://github.com/paradigmxyz/reth/issues/19942)
- **Static Files**: Immutable historical data

While MDBX provides block-level transactionality internally, the overall commit process involves multiple independent storage systems that cannot be atomically committed together. This breaks the traditional single-database transactional guarantees and requires careful recovery logic to handle partial commits.

Inspired by [Aptos's failpoint implementation](https://github.com/aptos-labs/aptos-core), this PR adds failpoint support to test these critical consistency boundaries.

### New Feature: `failpoints`

A new optional feature `failpoints` that can be enabled at build time:

```bash
cargo build -p reth --features failpoints

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -1092,6 +1092,23 @@ where
     async fn debug_write_mutex_profile(&self, _file: String) -> RpcResult<()> {
         Ok(())
     }
+
+    /// Handler for `debug_setFailpoint`
+    ///
+    /// Sets a failpoint with the given name and action for runtime fault injection.
+    async fn debug_set_failpoint(&self, name: String, actions: String) -> RpcResult<()> {
+        #[cfg(feature = "failpoints")]
+        {
+            fail::cfg(name, &actions).map_err(internal_rpc_err)
+        }
+        #[cfg(not(feature = "failpoints"))]
+        {
+            let _ = (name, actions);
+            Err(internal_rpc_err(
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: crates/rpc/rpc/src/debug.rs [hunk 1]
```diff
@@ -1092,6 +1092,23 @@ where
...
         Ok(())
     }
+
+    /// Handler for `debug_setFailpoint`
+    ///
+    /// Sets a failpoint with the given name and action for runtime fault injection.
+    async fn debug_set_failpoint(&self, name: String, actions: String) -> RpcResult<()> {
+        #[cfg(feature = "failpoints")]
+        {
+            fail::cfg(name, &actions).map_err(internal_rpc_err)
```
mattsse, line 1102: what is the expected behaviour of this, because this will panic in the rpc call context and wont affect anything else

PR #20984, opened 2026-01-13
feat(storage): split static file commit into sync_all and finalize

joshieDo:
Currently, committing static files performs three sequential operations (per segment):
1. fsync data file
2. write pending offsets to disk + fsync offset file
3. write pending configuration + fsync config file + fsync dir

--

This PR splits sf writers `commit()` into:
- `sync_all()` - performs (1) and (2): syncs data and offsets to disk
- `finalize()` - performs (3): writes configuration and marks writer clean

When parallelizing writes in the future, only `finalize()` needs to be sequential across all storage types. `sync_all()` can be run in parallel. 

This only applies to appending data, and not unwinding (eg. we don't want to erase data, before we know that `db.commit` has succeeded)

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -1851,6 +1851,8 @@ pub trait StaticFileWriter {
     fn has_unwind_queued(&self) -> bool;
 
     /// Finalizes all static file writers by committing their configuration to disk.
+    ///
+    /// Returns an error if prune is queued (use [`Self::commit`] instead).
     fn finalize(&self) -> ProviderResult<()>;
 }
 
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1849,6 +1849,9 @@ pub trait StaticFileWriter {
 
     /// Returns `true` if the static file provider has unwind queued.
     fn has_unwind_queued(&self) -> bool;
+
+    /// Finalizes all static file writers by committing their configuration to disk.
```
pepyakin, line 1853: leave a note about the err condition?

pepyakin (approved):

yongkangc (approved):

mattsse (approved):

PR #20986, opened 2026-01-13
fix(trie): account for flags in sorted is_empty()

aso20455:
Follow-up to #20801


mattsse (commented):

PR #20987, opened 2026-01-13
perf: use in-memory length for static files metrics

DaniPopes:


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -491,29 +494,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
-                    .unwrap_or_default();
-                let offsets_size = reth_fs_util::metadata(offsets_path)
-                    .map(|metadata| metadata.len())
-                    .unwrap_or_default();
-                let config_size = reth_fs_util::metadata(config_path)
-                    .map(|metadata| metadata.len())
-                    .unwrap_or_default();
-
-                size += data_size + index_size + offsets_size + config_size;
+                size += jar_provider.size() as u64;
```
DaniPopes, line 497: config size is not in-memory but it's negligible (<4kb per jar)

crates/storage/nippy-jar/src/lib.rs
```diff
@@ -201,11 +201,11 @@ impl<H: NippyJarHeader> NippyJar<H> {
         let config_path = path.with_extension(CONFIG_FILE_EXTENSION);
         let config_file = File::open(&config_path)
             .inspect_err(|e| {
-                warn!( ?path, %e, "Failed to load static file jar");
+                warn!(?path, %e, "Failed to load static file jar");
             })
             .map_err(|err| reth_fs_util::FsPathError::open(err, config_path))?;
 
-        let mut obj = Self::load_from_reader(config_file)?;
+        let mut obj = Self::load_from_reader(io::BufReader::new(config_file))?;
```
DaniPopes, line 208: drive-by perfington to avoid multiple syscalls for reading the config file

mattsse (approved):

shekhirin (approved):
good code

PR #20993, opened 2026-01-13
feat: parallelize `save_blocks` 

joshieDo:
* parallelize `save_blocks`
* `insert_block` now calls `save_blocks`

---

## PR Stack

```
#21045 (metrics)    #21003 (rocksdb)
         \           /
          \         /
           #21012 (account changesets)
              ↓
           #20993 (parallelize save_blocks) ◀
              ↓
            main
```

- https://github.com/paradigmxyz/reth/pull/21045
- https://github.com/paradigmxyz/reth/pull/21003
- https://github.com/paradigmxyz/reth/pull/21012
- https://github.com/paradigmxyz/reth/pull/20993 ◀

### DIFF
--- a/crates/optimism/cli/src/commands/import_receipts.rs
+++ b/crates/optimism/cli/src/commands/import_receipts.rs
@@ -18,7 +18,7 @@ use reth_optimism_primitives::{bedrock::is_dup_tx, OpPrimitives, OpReceipt};
 use reth_primitives_traits::NodePrimitives;
 use reth_provider::{
     providers::ProviderNodeTypes, DBProvider, DatabaseProviderFactory, OriginalValuesKnown,
-    ProviderFactory, StageCheckpointReader, StageCheckpointWriter, StateWriter,
+    ProviderFactory, StageCheckpointReader, StageCheckpointWriter, StateWriteConfig, StateWriter,
     StaticFileProviderFactory, StatsReader,
 };
 use reth_stages::{StageCheckpoint, StageId};
@@ -228,7 +228,11 @@ where
             ExecutionOutcome::new(Default::default(), receipts, first_block, Default::default());
 
         // finally, write the receipts
-        provider.write_state(&execution_outcome, OriginalValuesKnown::Yes, true)?;
+        provider.write_state(
+            &execution_outcome,
+            OriginalValuesKnown::Yes,
+            StateWriteConfig::default(),
+        )?;
     }
 
     // Only commit if we have imported as many receipts as the number of transactions.
--- a/crates/optimism/cli/src/commands/import_receipts.rs
+++ b/crates/optimism/cli/src/commands/import_receipts.rs
@@ -228,7 +228,7 @@ where
             ExecutionOutcome::new(Default::default(), receipts, first_block, Default::default());
 
         // finally, write the receipts
-        provider.write_state(&execution_outcome, OriginalValuesKnown::Yes)?;
+        provider.write_state(&execution_outcome, OriginalValuesKnown::Yes, true)?;
     }
 
     // Only commit if we have imported as many receipts as the number of transactions.
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2,7 +2,6 @@ use crate::{
     changesets_utils::{
         storage_trie_wiped_changeset_iter, StorageRevertsIter, StorageTrieCurrentValuesIter,
     },
-    either_writer::EitherWriter,
     providers::{
         database::{chain::ChainStorage, metrics},
         rocksdb::RocksDBProvider,
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -35,7 +35,7 @@ use alloy_primitives::{
 use itertools::Itertools;
 use parking_lot::RwLock;
 use rayon::slice::ParallelSliceMut;
-use reth_chain_state::ExecutedBlock;
+use reth_chain_state::{ComputedTrieData, ExecutedBlock};
 use reth_chainspec::{ChainInfo, ChainSpecProvider, EthChainSpec};
 use reth_db_api::{
     cursor::{DbCursorRO, DbCursorRW, DbDupCursorRO, DbDupCursorRW},
@@ -85,6 +85,7 @@ use std::{
     fmt::Debug,
     ops::{Deref, DerefMut, Range, RangeBounds, RangeFrom, RangeInclusive},
     sync::Arc,
+    thread,
     time::{Duration, Instant},
 };
 use tracing::{debug, trace};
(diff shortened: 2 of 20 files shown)


crates/optimism/cli/src/commands/import_receipts.rs
```diff
@@ -228,7 +228,7 @@ where
             ExecutionOutcome::new(Default::default(), receipts, first_block, Default::default());
 
         // finally, write the receipts
-        provider.write_state(&execution_outcome, OriginalValuesKnown::Yes)?;
+        provider.write_state(&execution_outcome, OriginalValuesKnown::Yes, true)?;
```
mattsse: can we make this a config struct because i assume well add more settings eventually?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3030,89 +3194,40 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockExecu
...
-    /// If ommers are not empty, this will modify [`BlockOmmers`](tables::BlockOmmers).
-    /// If withdrawals are not empty, this will modify
-    /// [`BlockWithdrawals`](tables::BlockWithdrawals).
-    ///
-    /// If the provider has __not__ configured full sender pruning, this will modify either:
-    /// * [`StaticFileSegment::TransactionSenders`] if senders are written to static files
-    /// * [`tables::TransactionSenders`] if senders are written to the database
-    ///
-    /// If the provider has __not__ configured full transaction lookup pruning, this will modify
-    /// [`TransactionHashNumbers`](tables::TransactionHashNumbers).
```
mattsse, line 3059: is this no longer accurate?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -357,97 +377,235 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
+            for (i, block) in blocks.iter().enumerate() {
+                let recovered_block = block.recovered_block();
+
+                let start = Instant::now();
+                self.insert_block_mdbx_only(recovered_block, tx_nums[i])?;
+                total_insert_block += start.elapsed();
+
+                // Full saving mode: write state, hashed state, and trie data
+                if save_mode.is_full() {
+                    let trie_data = block.trie_data();
```
mattsse: this can be called later?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3030,89 +3194,40 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockExecu
...
-    ///
-    /// If the provider has __not__ configured full sender pruning, this will modify either:
-    /// * [`StaticFileSegment::TransactionSenders`] if senders are written to static files
-    /// * [`tables::TransactionSenders`] if senders are written to the database
-    ///
-    /// If the provider has __not__ configured full transaction lookup pruning, this will modify
-    /// [`TransactionHashNumbers`](tables::TransactionHashNumbers).
+    /// This is a convenience method primarily used in tests. For production use,
+    /// prefer [`Self::save_blocks`] which handles execution output and trie data.
     fn insert_block(
```
mattsse, line 3239: when do we even call this?

separately

https://github.com/paradigmxyz/reth/blob/f101e09d049fec06f19ebe657cc50799aea231d3/crates/storage/provider/src/providers/database/provider.rs#L3369-L3369

appears to be unused?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3030,89 +3194,40 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider + 'static> BlockExecu
...
-    /// If withdrawals are not empty, this will modify
-    /// [`BlockWithdrawals`](tables::BlockWithdrawals).
-    ///
-    /// If the provider has __not__ configured full sender pruning, this will modify either:
-    /// * [`StaticFileSegment::TransactionSenders`] if senders are written to static files
-    /// * [`tables::TransactionSenders`] if senders are written to the database
-    ///
-    /// If the provider has __not__ configured full transaction lookup pruning, this will modify
-    /// [`TransactionHashNumbers`](tables::TransactionHashNumbers).
+    /// This is a convenience method primarily used in tests. For production use,
```
mattsse, line 3237: ah nvm, I think I understand now

mattsse (changes requested):
some suggestions, but afaict the flow makes sense, in parallelizes as much as possible

crates/optimism/cli/src/commands/import_receipts.rs
joshieDo: https://github.com/paradigmxyz/reth/pull/21012 added here

pepyakin (approved):
I've went through the code and it looks reasonable to me.

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -356,98 +376,251 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
...
+
+        thread::scope(|s| {
+            // SF writes
+            let sf_handle = s.spawn(|| sf_provider.write_blocks_data(&blocks, &tx_nums, sf_ctx));
+
+            // MDBX writes
+            for (i, block) in blocks.iter().enumerate() {
+                let recovered_block = block.recovered_block();
+
+                let start = Instant::now();
```
yongkangc, line 502: should we have the instant::now() outside the loop? right now in this scope we have a couple of this ~5

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1821,20 +1994,26 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
-
-        let tip = self.last_block_number()?.max(last_block);
 
         let (plain_state, reverts) =
             execution_outcome.bundle.to_plain_state_and_reverts(is_value_known);
 
         self.write_state_reverts(reverts, first_block)?;
         self.write_state_changes(plain_state)?;
 
+        if !write_receipts {
```
yongkangc:
```suggestion
        // // SF thread handles receipts when write_receipts=false
        if !write_receipts {
```

I guess this might be slightly confusing right now because i am unsure what the bool does. 

maybe instead of boolean we could use a config / enum?

yongkangc: write_receipts flag = "should this method write receipts to MDBX?"

The confusion is the flag name is ambiguous to me because:

- write_receipts = true → write to MDBX
- write_receipts = false → skip (SF thread writes to static files instead)

yongkangc (approved):
@joshieDo left some comments, overall flow lgtm

crates/storage/provider/src/providers/database/provider.rs
joshieDo: handled already in another PR: https://github.com/paradigmxyz/reth/pull/21012

crates/storage/provider/src/providers/database/provider.rs
mattsse: this seems fine, I think we want this per block

mattsse (approved):
checks out

PR #20995, opened 2026-01-13
refactor(net): avoid per-poll shrink_to_fit in ActiveSession

Snezhkko:
ActiveSession::poll used to call shrink_to_fit on its internal buffers on every Poll::Pending. This adds unnecessary reallocations and extra allocator work in a hot path, while the queues are already bounded and small. Removing the per-poll shrink keeps protocol behavior unchanged but reduces useless work under bursty load and long-lived sessions.

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -769,9 +769,6 @@ impl<N: NetworkPrimitives> Future for ActiveSession<N> {
                 this.pending_message_to_session = Some(msg);
             }
         }
-
-        this.shrink_to_fit();
-
         Poll::Pending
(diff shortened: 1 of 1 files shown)


crates/net/network/src/session/active.rs
```diff
@@ -769,9 +769,6 @@ impl<N: NetworkPrimitives> Future for ActiveSession<N> {
                 this.pending_message_to_session = Some(msg);
             }
         }
-
-        this.shrink_to_fit();
```
mattsse, line 773: where do we do this else because we might want to shrink in case we buffered a bunch of stuff

PR #20997, opened 2026-01-13
feat(trie): in-memory trie changesets

mediocregopher:
Introduces an in-memory cache for trie changesets to eliminate database writes during live sync, reducing I/O overhead while maintaining reorg support.

Key capabilities:
- Cache stores trie changesets for up to 64 blocks (2 epochs)
- Automatic eviction of older blocks using sliding window policy
- Thread-safe concurrent access with RwLock protection
- On-demand computation for cache misses via get_or_compute()
- Comprehensive metrics tracking (hits, misses, evictions, memory)

Usage:
- Changesets computed asynchronously in deferred trie task after block validation, integrated into existing background task without blocking validation path
- Centralized OverlayStateProviderFactory creation in validate_block_with_state eliminates duplicate trie input computations across validation paths
- Cache automatically populated during normal block processing
- Reorg operations transparently access cached or on-the-fly computed changesets via get_or_compute() API
- MerkleChangeSets pipeline stage marked deprecated, with database tables and write paths to be removed in follow-up PR

Architecture:
- ChangesetCache (crates/engine/tree/src/cache/): Core cache implementation with eviction, metrics, and thread-safety
- compute_trie_changesets (crates/trie/trie/src/changesets.rs): Changeset computation logic using TrieCursorFactory
- Integration points: EngineApiTreeHandler holds cache instance, PayloadValidator receives cache and factory, deferred trie task performs computation and caching

This change reduces write amplification during live sync, simplifies the pipeline architecture, and provides a foundation for removing legacy changeset persistence in a future cleanup phase.


mediocregopher (commented):
## Review Summary

Good architectural direction—caching changesets in-memory will significantly reduce I/O during live sync. A few issues to address:

### Critical Issues

1. **Non-idempotent cache insertion** — Duplicate `block_hash` inserts accumulate entries in `block_numbers[block_number]`, causing unbounded growth. If the same block gets inserted twice (background task + `get_or_compute()` race), the vector grows without bound even if `entries` stays bounded. Fix by checking if hash exists before pushing, or switch to `HashSet<B256>`.

2. **Lock poisoning panics** — `self.inner.read().unwrap()` / `write().unwrap()` will permanently break on any panic. Use `unwrap_or_else(|e| e.into_inner())` or switch to `parking_lot::RwLock`.

3. **Empty range handling** — `get_or_compute_range(provider, (block_number + 1)..=db_tip_block)` when `block_number == db_tip_block` creates an inverted range. Add early return:
   ```rust
   if start_block > end_block {
       return Ok(TrieUpdatesSorted::default());
   }
   ```

4. **Wiped storage cursor positioning** — `TrieCursorIter::new(cursor)` assumes cursor is at start, but it may have been seeked previously. Explicitly rewind before iterating.

### Minor Issues

- TOCTOU race in `get_or_compute()` can cause duplicate expensive computations (acceptable for now)
- Side-chain explosion: `block_numbers[height]` can grow unbounded with many competing payloads

### Missing Tests

- Cache idempotency (double insert same hash)
- Empty/inverted range behavior  
- Wiped storage cursor positioning regression
- Reorg scenario end-to-end

crates/trie/db/src/changesets.rs
```diff
@@ -0,0 +1,877 @@
...
+
+    Ok(TrieUpdatesSorted::new(account_nodes, storage_tries))
+}
+
+/// Thread-safe handle to the changeset cache.
+///
+/// This type wraps a shared, mutable reference to the changeset cache.
+/// The `RwLock` enables concurrent reads while ensuring exclusive access for writes.
+#[derive(Debug, Clone)]
+pub struct ChangesetCacheHandle {
```
shekhirin: I find the `Handle` naming a bit confusion here because we usually have an `fn handle()` method that returns such type

I think would be better to have a `ChangesetCacheInner` type and call this one `ChangesetCache`

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2814,22 +2703,11 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> StorageTrieWriter for DatabaseP
 
         let mut changeset_cursor =
             self.tx_ref().cursor_dup_write::<tables::StoragesTrieChangeSets>()?;
+        let curr_values_cursor = self.tx_ref().cursor_dup_read::<tables::StoragesTrie>()?;
```
mediocregopher, line 2733: This code got modified to not require the StorageTrieCurrentValuesIter type anymore, but it's dead code and can be deleted in a later PR anyway

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1365,6 +1375,21 @@ where
 
         debug!(target: "engine::tree", ?last_persisted_block_hash, ?last_persisted_block_number, elapsed=?start_time.elapsed(), "Finished persisting, calling finish");
         self.persistence_state.finish(last_persisted_block_hash, last_persisted_block_number);
+
+        // Evict trie changesets for blocks below the finalized block, but keep at least 64 blocks
+        if let Some(finalized) = self.canonical_in_memory_state.get_finalized_num_hash() {
+            let min_threshold = last_persisted_block_number.saturating_sub(64);
+            let eviction_threshold = finalized.number.min(min_threshold);
```
mattsse, line 1382: this makes sense, we perhaps could use a setting for this but 2 epochs makes sense

crates/trie/db/src/changesets.rs
```diff
@@ -0,0 +1,840 @@
...
+    Ok(TrieUpdatesSorted::new(account_nodes, storage_tries))
+}
+
+/// Thread-safe changeset cache.
+///
+/// This type wraps a shared, mutable reference to the cache inner.
+/// The `RwLock` enables concurrent reads while ensuring exclusive access for writes.
+#[derive(Debug, Clone)]
+pub struct ChangesetCache {
+    inner: Arc<RwLock<ChangesetCacheInner>>,
```
yongkangc, line 252: Do you think dashmap might be better? due to multiple writers hitting the cache

It is getting populated during the deferred trie task 

Non blocking

crates/node/builder/src/launch/common.rs
```diff
@@ -470,7 +471,10 @@ where
     /// Returns the [`ProviderFactory`] for the attached storage after executing a consistent check
     /// between the database and static files. **It may execute a pipeline unwind if it fails this
     /// check.**
-    pub async fn create_provider_factory<N, Evm>(&self) -> eyre::Result<ProviderFactory<N>>
+    pub async fn create_provider_factory<N, Evm>(
+        &self,
+        changeset_cache: ChangesetCache,
```
mattsse, line 476: this seems fine

crates/node/core/src/args/stage.rs
```diff
@@ -38,11 +38,6 @@ pub enum StageEnum {
     ///
     /// Handles Merkle tree-related computations and data processing.
     Merkle,
-    /// The merkle changesets stage within the pipeline.
-    ///
-    /// Handles Merkle trie changesets for storage and accounts.
-    #[value(name = "merkle-changesets")]
-    MerkleChangeSets,
```
mattsse, line 45: rip bozo

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -552,10 +551,6 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
                     self.write_hashed_state(&trie_data.hashed_state)?;
                     timings.write_hashed_state += start.elapsed();
 
-                    let start = Instant::now();
-                    self.write_trie_changesets(block_number, &trie_data.trie_updates, None)?;
-                    timings.write_trie_changesets += start.elapsed();
```
mattsse, line 557: hell yeah

docs/vocs/docs/pages/cli/reth/stage/drop.mdx
```diff
@@ -131,18 +131,17 @@ Static Files:
...
           Possible values:
-          - headers:           The headers stage within the pipeline
-          - bodies:            The bodies stage within the pipeline
-          - senders:           The senders stage within the pipeline
-          - execution:         The execution stage within the pipeline
-          - account-hashing:   The account hashing stage within the pipeline
-          - storage-hashing:   The storage hashing stage within the pipeline
-          - hashing:           The account and storage hashing stages within the pipeline
-          - merkle:            The merkle stage within the pipeline
-          - merkle-changesets: The merkle changesets stage within the pipeline
```
yongkangc, line 142: nice that we removed this whole satge

crates/trie/db/src/changesets.rs
```diff
@@ -0,0 +1,840 @@
...
+use reth_trie::{
+    changesets::compute_trie_changesets,
+    trie_cursor::{InMemoryTrieCursorFactory, TrieCursor, TrieCursorFactory},
+    HashedPostStateSorted, KeccakKeyHasher, StateRoot, TrieInputSorted,
+};
+use reth_trie_common::updates::{StorageTrieUpdatesSorted, TrieUpdatesSorted};
+use std::{
+    collections::{BTreeMap, HashMap},
+    ops::RangeInclusive,
+    sync::{Arc, RwLock},
```
mattsse: should we use parkinglot rwlock here?

mattsse (commented):
removing stages 👍

left one suggestion re std rwlock, afaict all of this checks out

crates/trie/db/src/changesets.rs
mattsse: we dont expect too many entries and the read+clone pattern dominates here

mattsse (approved):
yeet

PR #20999, opened 2026-01-13
refactor(custom-node): remove redundant Sized bound in FromTxCompact impl

0xxFloki:
Removes redundant `where Self: Sized` bound from the `FromTxCompact` implementation for `CustomTransaction`.


mattsse (commented):
we can keep this

PR #21000, opened 2026-01-13
refactor(fs-util): use create_file wrapper in atomic_write_file

kilavvy:
Removes code duplication in `atomic_write_file` by using the existing `create_file` wrapper instead of manually creating a file with error handling.
The `create_file` wrapper already provides the same functionality (creating a file with proper error context), so duplicating this logic inline violates DRY principles and creates a maintenance risk. Using the wrapper ensures consistency and reduces the chance of divergence if error handling logic changes in the future.


mattsse (commented):
we can keep as is

PR #21001, opened 2026-01-13
perf: use fixed-map for StaticFileSegment maps

DaniPopes:
minor change, change hashmap for `[Option<X>; N]` for static files map; O(1) lookup and super fast, deterministic iteration

### DIFF
--- a/crates/static-file/types/Cargo.toml
+++ b/crates/static-file/types/Cargo.toml
@@ -33,5 +33,6 @@ std = [
     "serde/std",
     "strum/std",
     "serde_json/std",
+    "fixed-map/std",
 ]
 clap = ["dep:clap"]
--- a/crates/static-file/types/Cargo.toml
+++ b/crates/static-file/types/Cargo.toml
@@ -15,6 +15,7 @@ workspace = true
 alloy-primitives.workspace = true
 
 clap = { workspace = true, features = ["derive"], optional = true }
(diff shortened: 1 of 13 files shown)


crates/static-file/types/Cargo.toml
```diff
@@ -32,5 +33,6 @@ std = [
     "serde/std",
     "strum/std",
     "serde_json/std",
+    "fixed-map/std",
```
pepyakin, line 36: redundant given the root cargo.toml.

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -138,11 +138,11 @@ impl<P: AsRef<Path>> StaticFileProviderBuilder<P> {
     /// If it is changed for an existing node, existing static files will not be affected and will
     /// be finished with the old blocks per file setting, but new static files will use the new
     /// setting.
-    pub fn with_blocks_per_file_for_segments(
-        mut self,
-        segments: HashMap<StaticFileSegment, u64>,
-    ) -> Self {
-        self.blocks_per_file.extend(segments);
+    #[allow(clippy::boxed_local)]
```
pepyakin: Would be great to add justification.

pepyakin (approved):

mattsse (approved):
checks out

PR #21005, opened 2026-01-13
fix: use POSIX-compliant IO redirection in Makefile

asmitadhungana:
## Summary
- Replace `&>` with `> /dev/null 2>&1` in `ensure-typos` and `ensure-dprint` targets
- The `&>` operator is bash-specific and fails in sh/dash shells that make typically uses
- This caused command detection to incorrectly report tools as not installed even when available in PATH

Closes #20912

## Test plan
- Verified `make ensure-typos` works correctly when typos is installed
- Verified `make ensure-dprint` works correctly when dprint is installed
- Confirmed the POSIX-compliant syntax works with /bin/sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -453,7 +453,7 @@ lint-typos: ensure-typos
 	typos
 
 ensure-typos:
-	@if ! command -v typos &> /dev/null; then \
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this is working

PR #21007, opened 2026-01-13
perf: reduce lock contention in on_inserted_executed_block

yongkangc:
**Summary**
Move expensive `insert_state` operation outside the write lock to eliminate 10-25ms blocking on `get_cache_for` reads.
`Blocked waiting for execution cache mutex blocked_for=25ms`

**Changes**
- Restructure `on_inserted_executed_block` to use two brief lock acquisitions instead of one long hold
- Add `update_with_guard_return` helper to extract data from lock closure
- Add race-safety guard to avoid clobbering newer cache entries

**Before**
```
Lock acquired → check parent → take cache → insert_state (10-25ms) → swap → release
```

**After**
```
Lock acquired → check parent → take cache → release (fast)
insert_state (10-25ms, no lock held)
Lock acquired → swap → release (fast)
```

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -556,40 +556,54 @@ where
         block_with_parent: BlockWithParent,
         bundle_state: &BundleState,
     ) {
-        self.execution_cache.update_with_guard(|cached| {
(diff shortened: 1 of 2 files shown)


Rjected (commented):
i think we actually want this inside the write lock because we need to make sure the updates are atomic? unless I'm misunderstanding this

PR #21009, opened 2026-01-13
perf: avoid cloning RangeInclusive in execute_range

CreeptoGengar:
Replaced `self.range.clone()` with direct iteration using a while loop and local `block_number` variable. This eliminates unnecessary cloning of the range for large block ranges while maintaining the same functionality.
The change preserves the original behavior: range boundaries are saved at the start for calculations, and the range is properly updated after batch execution completes.


mattsse (commented):
we can keep this as is

PR #21013, opened 2026-01-13
fix(pipeline): only check history limits when history-dependent stages need unwinding

yongkangc:
Fixes cases where a static file inconsistency (e.g., corrupted TransactionSenders) triggers an unwind that fails due to pruned history, even though the affected stage doesn't need history to unwind.

The `ensure_unwind_target_unpruned` check was applied universally, but only three stages actually need account/storage changesets to unwind:
- `Execution`: reverts plain state using changesets
- `AccountHashing`: reads `AccountChangeSets` via `unwind_account_hashing_range()`
- `StorageHashing`: reads `StorageChangeSets` via `unwind_storage_hashing_range()`

Other stages like Headers, Bodies, and SenderRecovery can unwind without history data.

### DIFF
--- a/crates/cli/commands/src/common.rs
+++ b/crates/cli/commands/src/common.rs
@@ -27,13 +27,14 @@ use reth_provider::{
         BlockchainProvider, NodeTypesForProvider, RocksDBProvider, StaticFileProvider,
         StaticFileProviderBuilder,
     },
-    ProviderFactory, StaticFileProviderFactory,
+    BlockNumReader, ProviderFactory, PruneCheckpointReader, StageCheckpointReader,
+    StaticFileProviderFactory,
 };
-use reth_stages::{sets::DefaultStages, Pipeline, PipelineTarget};
+use reth_stages::{sets::DefaultStages, Pipeline, StageId};
 use reth_static_file::StaticFileProducer;
 use std::{path::PathBuf, sync::Arc};
 use tokio::sync::watch;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 /// Struct to hold config and datadir paths
 #[derive(Debug, Parser)]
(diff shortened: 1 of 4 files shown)


crates/cli/commands/src/common.rs
```diff
@@ -174,12 +174,33 @@ impl<C: ChainSpecParser> EnvironmentArgs<C> {
...
-                PipelineTarget::Unwind(0),
+                unwind_block, 0,
                 "A static file <> database inconsistency was found that would trigger an unwind to block 0"
             );
 
+            // Pre-check history availability before attempting unwind
+            let provider = factory.provider()?;
+            if let Err(err) = check_unwind_history_available(&provider, &prune_modes, unwind_block)
+            {
+                error!(
```
mediocregopher, line 187: nit: let's combine these errors into a single one

crates/node/builder/src/launch/common.rs
```diff
@@ -547,6 +547,28 @@ where
                 inconsistency_source
             );
 
+            // Pre-check history availability before attempting unwind
+            if let Err(err) =
+                check_unwind_history_available(&provider_ro, &self.prune_modes(), unwind_block)
+            {
+                error!(
```
mediocregopher, line 554: nit: let's combine these errors into one

crates/stages/api/src/pipeline/mod.rs
```diff
@@ -34,6 +34,60 @@ use progress::*;
 use reth_errors::RethResult;
 pub use set::*;
 
+/// Stages that require account/storage history changesets to unwind.
+///
+/// - `Execution`: needs changesets to revert plain state
+/// - `AccountHashing`: calls `unwind_account_hashing_range` which reads `AccountChangeSets`
+/// - `StorageHashing`: calls `unwind_storage_hashing_range` which reads `StorageChangeSets`
+const STAGES_REQUIRING_HISTORY_FOR_UNWIND: [StageId; 3] =
```
mediocregopher: I think Merkle might as well? It's dependent on Account/StorageHashing, so maybe it's automatically covered, but just to be complete.

emmajam:
@yongkangc do we need this PR?

PR #21014, opened 2026-01-13
feat(cli): parse URL path and display ETA in `reth download`

shekhirin:
Allows to pass a path that has query arguments in it but the file itself still has a valid extension. Additionally, displays an ETA.

### DIFF
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -171,12 +171,14 @@ struct DownloadProgress {
     downloaded: u64,
     total_size: u64,
     last_displayed: Instant,
+    started_at: Instant,
 }
 
 impl DownloadProgress {
     /// Creates new progress tracker with given total size
     fn new(total_size: u64) -> Self {
-        Self { downloaded: 0, total_size, last_displayed: Instant::now() }
+        let now = Instant::now();
+        Self { downloaded: 0, total_size, last_displayed: now, started_at: now }
     }
 
     /// Converts bytes to human readable format (B, KB, MB, GB)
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):

crates/cli/commands/src/download.rs
```diff
@@ -201,8 +216,24 @@ impl DownloadProgress {
             let formatted_total = Self::format_size(self.total_size);
             let progress = (self.downloaded as f64 / self.total_size as f64) * 100.0;
 
+            // Calculate ETA based on current speed
+            let elapsed = self.started_at.elapsed();
+            let eta = if self.downloaded > 0 {
+                let remaining = self.total_size.saturating_sub(self.downloaded);
+                let speed = self.downloaded as f64 / elapsed.as_secs_f64();
```
pepyakin, line 223: is it guaranteed that `as_secs` will always be greater than 0.0?

crates/cli/commands/src/download.rs
```diff
@@ -201,8 +216,24 @@ impl DownloadProgress {
...
+                if speed > 0.0 {
+                    Duration::from_secs_f64(remaining as f64 / speed)
+                } else {
+                    Duration::ZERO
+                }
+            } else {
+                Duration::ZERO
```
pepyakin, line 230: do I understand it right that if it does not really progress then the duration is going to be displayed as 0sec?

pepyakin (approved):
fine by me

crates/cli/commands/src/download.rs
shekhirin: ye, should be ok

crates/cli/commands/src/download.rs
shekhirin: yep, we can also do ∞ but I think doesn't matter

yongkangc (approved):

PR #21015, opened 2026-01-13
fix: fix the nightly docker builds 

livehardd:
fixes the nightly docker builds that were failing when PROFILE=profiling is set.

The `docker_build_push` and `op_docker_build_push` functions only pass the `FEATURES` parameter to the child `make` call, but the subsequent `cp` command uses the `PROFILE` and `CARGO_TARGET_DIR` parameters from the parent `make`. This results in a "file not found" error because the binary is being built in the wrong directory.


mattsse (commented):
we can keep this as is

PR #21016, opened 2026-01-13
refactor: change PayloadTypes::block_to_payload to accept &SealedBlock

livehardd:
changes block_to_payload to accept &SealedBlock instead of owned value to avoid unnecessary cloning.

removes forced .clone() calls at miner.rs and e2e-test-utils where blocks are only read to construct payloads.


mattsse (commented):
this fn is primarily used for testing and conversion, this change would make it more expensive for instances where the payload is also the block

PR #21017, opened 2026-01-13
refactor(stages): reuse history index cache buffers in `collect_history_indices`

MozirDmitriy:
Refactor `collect_history_indices` to use cache.drain() and BlockNumberList::new_pre_sorted(indices.into_iter()) when flushing the cache. This avoids an extra pass over the map plus a separate clear() and aligns the implementation with collect_account_history_indices, while keeping the same semantics and progress reporting. No behavioral changes are expected; this is a small performance and clarity improvement in the history indexing path.


mattsse (approved):
this is indeed equivalent and we can reuse here

PR #21020, opened 2026-01-14
chore: remove redundant root calculation in stateless trie

BearSoltokien:
Remove redundant storage root calculation in stateless trie, eliminating unnecessary hash computation.


mattsse (commented):
this modifies the trie, so we need

PR #21023, opened 2026-01-14
test: add error scenario coverage for testing_buildBlockV1

livehardd:
adds error path tests for testing_buildBlockV1. current coverage only tests the happy path.

new tests:
invalid parent hash
timestamp earlier than parent block  
tx with insufficient gas (1000 vs 21000 needed)
tx with wrong nonce


mattsse (commented):
can keep this as is

PR #21039, opened 2026-01-14
feat: Support for sending logs through OTLP

kamsz:
This PR adds ability to specify OTLP endpoint to send logs to store them inside external systems (like Loki/VictoriaLogs/any OTLP compatible software).

### DIFF
--- a/crates/node/core/src/args/trace.rs
+++ b/crates/node/core/src/args/trace.rs
@@ -1,4 +1,4 @@
-//! Opentelemetry tracing configuration through CLI args.
+//! Opentelemetry tracing and logging configuration through CLI args.
 
 use clap::Parser;
 use eyre::WrapErr;
@@ -6,7 +6,7 @@ use reth_tracing::{tracing_subscriber::EnvFilter, Layers};
 use reth_tracing_otlp::OtlpProtocol;
 use url::Url;
 
-/// CLI arguments for configuring `Opentelemetry` trace and span export.
+/// CLI arguments for configuring `Opentelemetry` trace and logs export.
 #[derive(Debug, Clone, Parser)]
 pub struct TraceArgs {
     /// Enable `Opentelemetry` tracing export to an OTLP endpoint.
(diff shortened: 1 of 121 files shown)


mattsse (approved):

crates/node/core/src/args/trace.rs
```diff
@@ -30,9 +30,29 @@ pub struct TraceArgs {
...
+        env = "OTEL_EXPORTER_OTLP_LOGS_ENDPOINT",
+        global = true,
+        value_name = "URL",
+        num_args = 0..=1,
+        default_missing_value = "http://localhost:4318/v1/logs",
+        require_equals = true,
+        value_parser = parse_otlp_endpoint,
+        help_heading = "Logging"
+    )]
+    pub logs_otlp: Option<Url>,
```
lean-apple, line 51: It could have been nice to separate the logs and trace args and have a proper logs args module as we have already trace/metric args, it would be more in the "opentelemetry" protocol mindset :)

crates/tracing-otlp/src/lib.rs
```diff
@@ -115,6 +154,43 @@ impl OtlpConfig {
...
+        endpoint: Url,
+        protocol: OtlpProtocol,
+    ) -> eyre::Result<Self> {
+        Ok(Self { service_name: service_name.into(), endpoint, protocol })
+    }
+
+    /// Returns the service name.
+    pub fn service_name(&self) -> &str {
+        &self.service_name
+    }
```
lean-apple, line 181: Is there a reason for not reusing the otlp traces config ?

crates/node/core/src/args/trace.rs
kamsz: they're separate now, what exactly do you mean?

there's `--logs-otlp` for logs, `--tracing-otlp` for traces, and for metrics you can use existing push gateway functionality to push them through otel collector (`--metrics.prometheus.push.url`)

crates/tracing-otlp/src/lib.rs
kamsz: yes as you might be scraping logs through different means already (e.g. within k8s clusters via existing fluent-bit daemonset) and not necessarily want to ship everything to otlp endpoint; this way you can be selective what you want to send through otlp and avoid data duplication

PR #21042, opened 2026-01-14
perf: don't clone entire keys set

klkvr:
Instead of cloning the entire set we can only clone accounts relevant for the given multiproof

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/multiproof.rs
+++ b/crates/engine/tree/src/tree/payload_processor/multiproof.rs
@@ -20,6 +20,7 @@ use reth_trie_parallel::{
         AccountMultiproofInput, ProofResultContext, ProofResultMessage, ProofWorkerHandle,
     },
 };
+use revm_primitives::map::{hash_map, B256Map};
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -705,7 +718,30 @@ impl MultiProofTask {
         }
 
         // Clone+Arc MultiAddedRemovedKeys for sharing with the dispatched multiproof tasks
-        let multi_added_removed_keys = Arc::new(self.multi_added_removed_keys.clone());
+        let multi_added_removed_keys = Arc::new(MultiAddedRemovedKeys {
+            account: self.multi_added_removed_keys.account.clone(),
+            storages: {
+                let mut storages = B256Map::default();
```
shekhirin: can probably do `with_capacity(not_fetched_state_update.storage.len())`

mattsse (commented):
does that make sense @mediocregopher

DaniPopes (approved):

mediocregopher (approved):

PR #21044, opened 2026-01-14
docs(banlist): fix doc comments referencing ip instead of peer

gap-editor:
Fix copy-paste errors in doc comments where peer methods incorrectly reference "ip":
- `is_banned_peer`: "ip" → "peer"  
- `unban_peer`: "ip address" → "peer"


mattsse (commented):
we can keep this

PR #21050, opened 2026-01-14
fix: improve valueError to preserve transaction context

ethfanWilliam:
Changed TransactionConversionError to use Debug format instead of Display when wrapping ValueError. Keeps the complete transaction request data in error messages instead of just a string.

Before: "Failed to convert transaction into RPC response: Required fields missing"
After: Shows the actual TransactionRequest with all fields so you can see exactly what's missing

(The ValueError<TxReq> is generic so can't store it directly, but using {:?} gives us basically the same debugging info.)


mattsse (commented):
we can keep this as is

PR #21051, opened 2026-01-14
fix: propagate keccak-cache-global feature to reth-optimism-cli

ethfanWilliam:
fixes the feature propagation issue mentioned in #20524. the keccak-cache-global feature now properly flows through op-reth -> reth-optimism-cli -> reth-node-core, matching how asm-keccak is already set up. this should make the feature show up correctly in op-reth --version output.


mattsse (approved):
this is weird, this should have been forwarded via zepter

ethfanWilliam:
> this is weird, this should have been forwarded via zepter

Not sure why didn't catch by CI, BTW would you please re-touch https://github.com/paradigmxyz/reth/pull/20887?

PR #21072, opened 2026-01-14
feat(storage): add tracing spans

pepyakin:
Add tracing instrumentation to the block persistence code paths for better observability and performance debugging. You can see them in action in the persistence service visible here https://share.firefox.dev/3ZelM78.


mattsse (approved):
very nice, this is a lot cleaner and will definitely help 

pending @joshieDo

joshieDo (approved):

DaniPopes (approved):

PR #21073, opened 2026-01-14
feat(mdbx): more tracing observability

figtracer:
this PR aims at migrating https://github.com/figtracer/reth-mdbx-profiler into a reth exex
- the codebase would be much smaller 
- data would also be 100% accurate, since we wouldn't need to make assumptions based on page fault-ops timing

for that i wanted to add a couple of extra tracing spans which may be useful as well for other experiments

edit: this would probably cause a lot of overhead as well so idk if leaving it in tracing without any feature gate would be a good idea

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/cursor.rs
+++ b/crates/storage/db/src/implementation/mdbx/cursor.rs
@@ -2,7 +2,7 @@
 
 use super::utils::*;
 use crate::{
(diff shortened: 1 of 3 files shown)


mattsse (commented):
these are quite hot unsure what the impact of these additional spans would be @joshieDo @shekhirin

PR #21075, opened 2026-01-14
refactor: avoid redundant allocation in create_reorg_head

Fibonacci747:
`create_reorg_head` always cloned transactions from next_block into candidate_transactions, even when depth > 0. In that case the vector was immediately overwritten in the loop with into_transactions() from the selected reorg target child and the initial allocation was never used.

This change makes the initialization of candidate_transactions conditional on depth == 0, preserving the existing reorg semantics while avoiding an unnecessary allocation and copy for deeper reorg simulations.


mattsse (commented):
we can keep this

PR #21077, opened 2026-01-14
fix(db): change commit return type from Result<bool> to Result<()>

pepyakin:
The boolean return value from mdbx_txn_commit was never checked by callers. When MDBX returns MDBX_RESULT_TRUE, it means the transaction was aborted due to previous errors - this is now properly surfaced as Error::BotchedTransaction instead of being silently ignored.

### DIFF
--- a/crates/storage/db-api/src/mock.rs
+++ b/crates/storage/db-api/src/mock.rs
@@ -92,10 +92,10 @@ impl DbTx for TxMock {
 
     /// Commits the transaction.
     ///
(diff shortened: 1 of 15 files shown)


mattsse (approved):
lgtm

ref https://libmdbx.dqdkfa.ru/group__c__transactions.html#ga3146a36cf1d0603cc6e3d52e97ec38a9

pending @joshieDo

joshieDo (approved):

shekhirin (approved):

PR #21082, opened 2026-01-14
feat(reth-bench): add generate-big-block command

Rjected:
Adds a new reth-bench subcommand that generates large blocks by:

* Fetching transactions and receipts from existing blocks via RPC
* Packing them until target gas (based on actual gas_used) is reached
* Calling `testing_buildBlockV1` to build the block
* Executing via `newPayload` + `forkchoiceUpdated`

This requires ramping up the gas limit with the included gas-limit-ramp benchmark command beforehand, and --from should be set to the block before ramping up the gas limit.

Also adds:

gas-limit-ramp: Generates empty blocks with incrementally increasing gas limits, saving payloads for later replay
replay-payloads: Replays payloads from either gas-limit-ramp or generate-big-block output directories

Note: When count == 1, the payload is only generated and saved, not executed. When count > 1, each payload is executed before building the next

Usage:

First ramp up the gas limit with empty blocks:

```
reth-bench gas-limit-ramp \
  --blocks 1000 \
  --engine-rpc-url http://localhost:8551 \
  --jwt-secret ~/.local/share/reth/mainnet/jwt.hex \
  --output ./gas-ramp-payloads
```

Then generate big blocks using ramped gas limit:

```
reth-bench generate-big-block \
  --rpc-url https://eth-mainnet.g.alchemy.com/v2/<key> \
  --engine-rpc-url http://localhost:8551 \
  --testing-rpc-url http://localhost:8545 \
  --jwt-secret ~/.local/share/reth/mainnet/jwt.hex \
  --from 22000000 \
  --target-gas 100000000 \
  --count 50 \
  --output-dir ./big-block-payloads
```

Then replay payloads (gas ramp first, then big blocks):

```
reth-bench replay-payloads \
  --engine-rpc-url http://localhost:8551 \
  --jwt-secret ~/.local/share/reth/mainnet/jwt.hex \
  --gas-ramp-dir ./gas-ramp-payloads \
  --payload-dir ./big-block-payloads
```

Or replay just gas ramp payloads:

```
reth-bench replay-payloads \
  --engine-rpc-url http://localhost:8551 \
  --jwt-secret ~/.local/share/reth/mainnet/jwt.hex \
  --gas-ramp-dir ./gas-ramp-payloads \
  --payload-dir ./gas-ramp-payloads
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,9 +106,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.24"
+version = "0.2.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b163ff4acf0eac29af05a911397cc418a76e153467b859398adc26cb9335a611"
+checksum = "25db5bcdd086f0b1b9610140a12c59b757397be90bd130d8d836fc8da0815a34"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
@@ -448,7 +448,7 @@ dependencies = [
  "foldhash 0.2.0",
  "getrandom 0.3.4",
  "hashbrown 0.16.1",
- "indexmap 2.12.1",
+ "indexmap 2.13.0",
  "itoa",
  "k256",
  "keccak-asm",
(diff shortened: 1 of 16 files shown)


mattsse (approved):
lgtm, I think this is fine and we can fix as we go

PR #21085, opened 2026-01-14
refactor(db): return Result<()> from DbTx::commit

DaniPopes:
MDBX commit returns a bool indicating whether the transaction was aborted due to a previous error. Instead of propagating this bool up the stack where it was universally ignored, convert it to an error when true. This makes the API cleaner - commit either succeeds with () or fails with a DatabaseError.

HT @prestwich


mattsse (commented):
i believe this is a dup of #21077

shekhirin (approved):

PR #21087, opened 2026-01-15
refactor: decouple bodies stage from transaction index

BearSoltokien:
Switch bodies stage consistency checks to use BlockBodyIndices, removing unnecessary dependency on transaction index tables.


mattsse (commented):
we need to use transactionblocks here because we dont necesarrily have bodyindices

PR #21088, opened 2026-01-15
feat: configurable EVM execution limits

rezzmah:
closes #21018

i.e. allows for configurable `MAX_CODE_SIZE`, `MAX_INITCODE_SIZE`, and `tx_gas_limit_cap` on existing chain specs.

This PR doesn't implement genesis parsing into the chainspec as that requires an opinionated approach on the genesis file shape. 

i realised a valid alternative is to have a custom `ChainSpecParser` which also meets most requirements

Closes RETH-186


crates/optimism/evm/src/lib.rs
```diff
@@ -179,7 +189,16 @@ where
             self.chain_spec().next_block_base_fee(parent, attributes.timestamp).unwrap_or_default(),
             self.chain_spec(),
             self.chain_spec().chain().id(),
-        ))
+        );
```
mattsse: I'd be nicer if we can make this like `.with_limits()` then we dont need to duplicate all of the setters, so I'd like to move the EvmLimits type to the alloy repo instead then we can use it in evmenv and elsewhere

mattsse (changes requested):
I think directionally this is appropriate, but I'd like to avoid the setter code duplications

wdyt @klkvr

crates/optimism/evm/src/lib.rs
rezzmah: Agreed that's cleaner. I'll start on it

crates/ethereum/evm/src/lib.rs
```diff
@@ -245,10 +247,6 @@ where
             cfg_env.set_max_blobs_per_tx(blob_params.max_blobs_per_tx);
         }
 
-        if self.chain_spec().is_osaka_active_at_timestamp(timestamp) {
```
rezzmah, line 251: I'm kind of torn about this. 

In most places we're relying on the `None` value of `cfg_env.tx_gas_limit_cap` so that the `tx_gas_limit_cap` is derived in revm based on the Osaka fork. 

https://github.com/bluealloy/revm/blob/7f3083b9ac5d72f31519a853c3e51ede1ff8641f/crates/context/src/cfg.rs#L377-L385

The alternative is we be consistent everywhere in reth explicitly define the bound at forks and don't rely on revm defaults.



The easier approach is to leave as `None` and rely on revm

crates/optimism/evm/src/lib.rs
rezzmah: https://github.com/alloy-rs/evm/pull/253

rezzmah:
TODO: figure out how best to update:

https://github.com/paradigmxyz/reth/blob/bf272c94327f066a442ef10a2ca320d419ecc3f1/crates/consensus/common/src/validation.rs#L157-L169

&&

https://github.com/paradigmxyz/reth/blob/1ec6ed9cee4a2ed6e8e55b99eaa0fbdccd04f1a7/crates/transaction-pool/src/validate/eth.rs#L513-L521

crates/ethereum/evm/src/lib.rs
rezzmah: Changed it so that `None` behaviour semantics now mean that `None` means no limit

crates/chainspec/src/spec.rs
```diff
@@ -544,6 +585,28 @@ impl<H: BlockHeader> ChainSpec<H> {
...
+                .map(|(_, params)| *params)
+                .unwrap_or_default(),
+        };
+
+        // EIP-7825: resolve tx gas limit cap for Osaka
+        if params.tx_gas_limit_cap.is_none() && self.is_osaka_active_at_timestamp(timestamp) {
+            EvmLimitParams { tx_gas_limit_cap: Some(MAX_TX_GAS_LIMIT_OSAKA), ..params }
+        } else {
+            params
+        }
```
rezzmah: This seems like the best way to unify the logic so all calling contexts don't have to do a special `is_osaka` check which results in a disjoint source of truth.

basically the `None` case mimicks the default revm behaviour, and other cases correctly override this across all contexts.

This does also mean that when we call
https://github.com/paradigmxyz/reth/blob/4354bd1899caca09e32c2f861b78e2566ba42b74/crates/ethereum/evm/src/lib.rs#L164

We're always setting the value post-Osaka and not relying on revm's default behaviour anymore, except in pre-Osaka case

crates/chainspec/src/spec.rs
```diff
@@ -388,6 +394,37 @@ impl From<ForkBaseFeeParams> for BaseFeeParamsKind {
...
+    Constant(EvmLimitParams),
+    /// Variable EVM limits based on hardfork activation
+    Variable(ForkEvmLimitParams),
+}
+
+impl Default for EvmLimitParamsKind {
+    fn default() -> Self {
+        Self::Constant(EvmLimitParams::default())
+    }
+}
```
rezzmah: I was initially concerned that using this default could cause cause pre-shanghai blocks to fail execution if they had a larger init code size than `MAX_INITCODE_SIZE` since we're explicitly setting the limit in a supposedly non-fork aware way. 

However, this value is propagated to revm and only really queried here which only retrieves the value if the specId is shanghai, making this safe.

https://github.com/bluealloy/revm/blob/a31fa98efe2208af326cf70f59d374165e0df363/crates/handler/src/validation.rs#L216-L222

crates/transaction-pool/src/validate/eth.rs
```diff
@@ -361,10 +360,14 @@ where
         }
 
         // Check whether the init code size has been exceeded.
-        if self.fork_tracker.is_shanghai_activated() &&
-            let Err(err) = transaction.ensure_max_init_code_size(MAX_INIT_CODE_BYTE_SIZE)
-        {
-            return Err(TransactionValidationOutcome::Invalid(transaction, err))
+        if self.fork_tracker.is_shanghai_activated() {
+            let max_initcode_size = self
```
rezzmah: an alternative way to do this could be to just remove the `is_shanghai_activated` check and make 
`max_initcode_size` an `Option` that behaves similarly to `tx_gas_limit_cap`, i.e.

```rs
pub struct EvmLimitParams {
    pub max_code_size: usize,
    /// Make this Optional as well
    pub max_initcode_size: Option<usize>,
    pub tx_gas_limit_cap: Option<u64>,
}
```
and

```rs
    pub fn evm_limit_params_at_timestamp(&self, timestamp: u64) -> EvmLimitParams {
        let mut params = match &self.evm_limit_params {
            EvmLimitParamsKind::Constant(params) => *params,
            EvmLimitParamsKind::Variable(ForkEvmLimitParams(fork_params)) => fork_params
                .iter()
                .rev()
                .find(|(fork, _)| {
                    self.hardforks.is_fork_active_at_timestamp(fork.clone(), timestamp)
                })
                .map(|(_, params)| *params)
                .unwrap_or_default(),
        };

        // EIP-7825: resolve tx gas limit cap for Osaka
        if params.tx_gas_limit_cap.is_none() && self.is_osaka_active_at_timestamp(timestamp) {
            params.tx_gas_limit_cap = Some(MAX_TX_GAS_LIMIT_OSAKA);
        }
        if params.max_initcode_size.is_none() && self.is_shanghai_activated(timestamp) {
            params.tx_gas_limit_cap = Some(MAX_INIT_CODE_BYTE_SIZE)
        }
        params
    }
```

However, the current approach works with the caveat of edge cases for custom networks don't enable shanghai but want to tweak the evm limits

crates/optimism/chainspec/src/lib.rs
```diff
@@ -1345,4 +1350,40 @@ mod tests {
             assert!(!content.contains(eth_hf.name()));
         }
     }
+
+    #[test]
+    fn test_variable_evm_limits_with_custom_hardfork() {
```
rezzmah: this test showcases the main unlock from this PR

crates/chainspec/src/spec.rs
```diff
@@ -388,6 +394,37 @@ impl From<ForkBaseFeeParams> for BaseFeeParamsKind {
...
 pub struct ForkBaseFeeParams(Vec<(Box<dyn Hardfork>, BaseFeeParams)>);
 
+/// Enum for constant or fork-dependent EVM limit parameters.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum EvmLimitParamsKind {
+    /// Constant EVM limits for all blocks
+    Constant(EvmLimitParams),
+    /// Variable EVM limits based on hardfork activation
+    Variable(ForkEvmLimitParams),
+}
```
mattsse: this makes sense but only saves 1 lookup so maybe not wrth it but we can keep

crates/chainspec/src/spec.rs
mattsse: I dont understand why this is necessary

why do we need to check this?
if this is wrong then this is a misconfig issue imo

crates/chainspec/src/spec.rs
```diff
@@ -544,6 +585,28 @@ impl<H: BlockHeader> ChainSpec<H> {
...
+    pub fn evm_limit_params_at_timestamp(&self, timestamp: u64) -> EvmLimitParams {
+        let params = match &self.evm_limit_params {
+            EvmLimitParamsKind::Constant(params) => *params,
+            EvmLimitParamsKind::Variable(ForkEvmLimitParams(fork_params)) => fork_params
+                .iter()
+                .rev()
+                .find(|(fork, _)| {
+                    self.hardforks.is_fork_active_at_timestamp(fork.clone(), timestamp)
+                })
+                .map(|(_, params)| *params)
```
mattsse: this is ne additional lookup, which seems fine but maybe it's better to group all of these together

Cargo.toml
```diff
@@ -747,7 +747,11 @@ vergen-git2 = "9.1.0"
 # networking
 ipnet = "2.11"
 
+# TODO: Remove this patch once alloy-evm is released with with_limits support
 [patch.crates-io]
+alloy-evm = { git = "https://github.com/alloy-rs/evm.git", branch = "main" }
+alloy-op-evm = { git = "https://github.com/alloy-rs/evm.git", branch = "main" }
```
mattsse: these can be removed now

mattsse (commented):
all of this makes sense

left one more q

ptal @klkvr

crates/chainspec/src/spec.rs
rezzmah: when you say "group all of these", do you mean group this with the other fork based lookups in the chain spec - i.e. base fee params + evm limits?

Cargo.toml
rezzmah: maybe im missing something but i don't think it can be removed yet. reth main branch is still targeting 
```toml
alloy-evm = { version = "0.26.3", default-features = false }
```
and https://github.com/alloy-rs/evm doesn't have a new release yet

crates/chainspec/src/spec.rs
rezzmah: You're right. It's a misconfig issue. Fixed in [e0bbf0b](https://github.com/paradigmxyz/reth/pull/21088/commits/e0bbf0b1336a5396af051c211e5457d49e32fabd)

crates/chainspec/src/spec.rs
```diff
@@ -544,6 +595,21 @@ impl<H: BlockHeader> ChainSpec<H> {
...
+        match &self.evm_limit_params {
+            EvmLimitParamsKind::Constant(params) => *params,
+            EvmLimitParamsKind::Variable(ForkEvmLimitParams(fork_params)) => fork_params
+                .iter()
+                .rev()
+                .find(|(fork, _)| {
+                    self.hardforks.is_fork_active_at_timestamp(fork.clone(), timestamp)
+                })
+                .map(|(_, params)| *params)
+                .unwrap_or_default(),
```
mattsse: @klkvr this lookup isnt super ideal, but i dont think this matters much

mattsse (approved):
all of this makes sense

perhaps @klkvr has some additional suggestions

crates/consensus/common/src/validation.rs
```diff
@@ -146,22 +144,23 @@ pub fn validate_block_pre_execution<B, ChainSpec>(
...
         return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
     }
-    // EIP-7825 validation
-    if chain_spec.is_osaka_active_at_timestamp(block.timestamp()) {
+    // EIP-7825 transaction gas limit validation
+    let evm_limits = chain_spec.evm_limit_params_at_timestamp(block.timestamp());
+    if let Some(tx_gas_cap) = evm_limits.tx_gas_limit_cap {
         for tx in block.body().transactions() {
-            if tx.gas_limit() > MAX_TX_GAS_LIMIT_OSAKA {
+            if tx.gas_limit() > tx_gas_cap {
```
klkvr: I think this check is actually redundant because it's already done in revm for each transaction

crates/ethereum/evm/src/lib.rs
```diff
@@ -160,7 +160,8 @@ where
             self.chain_spec(),
             self.chain_spec().chain().id(),
             self.chain_spec().blob_params_at_timestamp(header.timestamp),
-        ))
+        )
+        .with_limits(self.chain_spec().evm_limit_params_at_timestamp(header.timestamp)))
```
klkvr: i'm a bit concerned about this because this now means that if an sdk user needs to modify the execution limits they would have to update both `ChainSpec::evm_limit_params_at_timestamp` and `ConfigureEvm::evm_env` meaning there's no single source of truth

Not sure what's the best way to approach this. I guess ideally we have `CfgEnv` outputted from `evm_env` being source of truth but propagating it to the txpool validation might be tricky

crates/consensus/common/src/validation.rs
rezzmah: yea, i assumed it was valuable as a short-circuit before reaching that point. 

It was introduced in this PR https://github.com/paradigmxyz/reth/pull/18439 with some reasoning about resource exhaustion

crates/ethereum/evm/src/lib.rs
rezzmah: Good point. I see two cases

Case 1: SDK user with a custom `ConfigureEvm` wanting Osaka with default limits
A simple version bump could miss this the `with_limits`. In such a case, they would still get appropriate defaults in revm.
https://github.com/bluealloy/revm/blob/7f3083b9ac5d72f31519a853c3e51ede1ff8641f/crates/context/src/cfg.rs#L378-L385

Case 2: SDK user with a custom `ConfigureEvm` wanting custom limits
Like you said, if they want custom limits, it isn't immediately obvious they have to add `with_limits` here after they modify/configure the chain_spec.

I tried an approach of weaving in the chain_spec as an associated type and then using template method pattern to force the `with_limits` call and keep the chain_spec as the SoT, but it was too verbose for my liking.

rezzmah: poc of template method pattern but https://github.com/rezzmah/reth/pull/7 my gut says its overkill and should be done differently. 
I'd rather figure that out in a different pr

crates/consensus/common/src/validation.rs
```diff
@@ -146,27 +143,14 @@ pub fn validate_block_pre_execution<B, ChainSpec>(
...
-            if tx.gas_limit() > MAX_TX_GAS_LIMIT_OSAKA {
-                return Err(TxGasLimitTooHighErr {
-                    tx_hash: *tx.tx_hash(),
-                    gas_limit: tx.gas_limit(),
-                    max_allowed: MAX_TX_GAS_LIMIT_OSAKA,
-                }
-                .into());
-            }
-        }
-    }
```
klkvr, line 169: removed this for now, imo it's fine to not optimize for this codepath

klkvr (approved):
looks good to me

sorry for completely changing the direction fo the PR but I think simply querying this data from evm should pay off eventually

crates/evm/evm/src/lib.rs
```diff
@@ -203,6 +203,7 @@ pub trait ConfigureEvm: Clone + Debug + Send + Sync + Unpin {
                     + FromRecoveredTx<TxTy<Self::Primitives>>
                     + FromTxWithEncoded<TxTy<Self::Primitives>>,
             Precompiles = PrecompilesMap,
+            Spec: Into<SpecId>,
```
mattsse, line 206: this seems appropirate given that this is kinda enshrined by revm as well

crates/transaction-pool/src/validate/eth.rs
```diff
@@ -774,6 +781,19 @@ where
         }
 
         self.block_gas_limit.store(new_tip_block.gas_limit(), std::sync::atomic::Ordering::Relaxed);
+
+        // Get EVM limits from evm_config.evm_env()
+        let evm_env = self
+            .evm_config
+            .evm_env(new_tip_block)
+            .expect("evm_env should not fail for executed block");
```
mattsse, line 789: this seems fine to do on a per block basis

mattsse (approved):
makes sense, I think this will be easy to migrate actually

PR #21089, opened 2026-01-15
perf(trie): preserve allocations in sparse trie wipe()

doocho:
Modify `wipe()` methods in sparse tries to preserve existing HashMap allocations instead of creating new ones. This follows the same pattern already used in `clear()` methods.

**Note**: Since Cancun (EIP-6780), SELFDESTRUCT only fully destroys contracts created in the same transaction, significantly reducing when wipe() is triggered. The primary benefits are:
1. Historical sync: Processing pre-Cancun blocks where SELFDESTRUCT was more common
2. Code consistency: wipe() now follows the same allocation-preserving pattern as clear()
3. Future-proofing: If wipe() is used for other purposes in the future

Closes #16201


mediocregopher (approved):
This lgtm, pending approval from @Rjected or @shekhirin

crates/trie/sparse/src/trie.rs
```diff
@@ -947,10 +947,14 @@ impl SparseTrieInterface for SerialSparseTrie {
...
-        self.nodes = HashMap::from_iter([(Nibbles::default(), SparseNode::Empty)]);
-        self.values = HashMap::default();
+        // Clear existing allocations instead of creating new ones to preserve capacity
+        self.nodes.clear();
+        self.nodes.insert(Nibbles::default(), SparseNode::Empty);
+        self.branch_node_masks.clear();
+        self.values.clear();
         self.prefix_set = PrefixSetMut::all();
         self.updates = self.updates.is_some().then(SparseTrieUpdates::wiped);
+        self.rlp_buf.clear();
```
mattsse: this does clear more things now, but I think this is appropriate

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -781,9 +765,13 @@ impl SparseTrieInterface for ParallelSparseTrie {
 
     fn wipe(&mut self) {
         self.upper_subtrie.wipe();
-        self.lower_subtries = [const { LowerSparseSubtrie::Blind(None) }; NUM_LOWER_SUBTRIES];
+        // Clear lower subtries instead of replacing with new Blind(None) to preserve allocations
```
mattsse: this comment is totally irrelevant without the context of this pr

doocho: You're right, it's idiomatic Rust and self-explanatory. Removed the comments.

DaniPopes (approved):

PR #21096, opened 2026-01-15
perf(db): Sort transaction hash/nums before inserting them

mediocregopher:
Sorting prior to insertion helps with locality when inserting into MDBX. When we migrate this data to rocksdb later it will also be helpful for perf to have it pre-sorted.

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -443,6 +443,41 @@ where
             Self::RocksDB(batch) => batch.delete::<tables::TransactionHashNumbers>(hash),
         }
     }
+
+    /// Puts multiple transaction hash number mappings from a pre-sorted slice.
+    ///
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/either_writer.rs
```diff
@@ -443,6 +443,41 @@ where
             Self::RocksDB(batch) => batch.delete::<tables::TransactionHashNumbers>(hash),
         }
     }
+
+    /// Puts multiple transaction hash number mappings from a pre-sorted slice.
+    ///
+    /// The entries MUST be pre-sorted by transaction hash for optimal B-tree database performance.
+    /// Uses `cursor.upsert()` which handles arbitrary insertion order and duplicates.
```
pepyakin, line 450: I don't get it, why do we use `upsert` if we already require guarantee that the entries are presorted?

PR #21097, opened 2026-01-15
fix(exex): use max when updating highest_committed_block_height

endorina:
Fix incorrect update of highest_committed_block_height in BlockCache.

Previously, the value was directly assigned from the new committed chain's
tip, which could decrease the maximum when notifications with lower block
numbers were inserted. Now it uses max() to ensure the value only increases,
matching the logic used for lowest_committed_block_height.


mattsse (commented):
this isnt needed

PR #21099, opened 2026-01-15
chore(rpc): return graceful responses for PoW endpoints

0xWeakSheep:
- Adjust PoW RPC endpoints to return graceful responses (eth_mining returns false , eth_getWork/eth_submitWork are unsupported).
- Add a PoW-specific error matcher in tests and update assertions accordingly.
- No major changes beyond PoW RPC behavior and related test assertions.


Rjected (approved):
LGTM

PR #21103, opened 2026-01-15
feat(primitives-traits): add try_recover_signers for parallel batch recovery

mattsse:
## Summary

Add `try_recover_signers` function that combines decoding and signature recovery in a single parallel pass. This is more efficient than sequential recovery for large transaction batches.

## Changes

- Add `try_recover_signers` to `transaction::recover` module
- Uses rayon for parallel recovery when the `rayon` feature is enabled
- Update testing RPC to use batch recovery instead of sequential

## Usage

```rust
let recovered_txs = try_recover_signers(&raw_transactions, |tx| {
    MyTxType::decode_2718_exact(tx.as_ref())
        .map_err(RecoveryError::from_source)
})?;
```

### DIFF
--- a/crates/primitives-traits/src/transaction/recover.rs
+++ b/crates/primitives-traits/src/transaction/recover.rs
@@ -1,68 +1,137 @@
 //! Helpers for recovering signers from a set of transactions
 
+use crate::{transaction::signed::RecoveryError, Recovered, SignedTransaction};
+use alloc::vec::Vec;
+use alloy_consensus::transaction::SignerRecoverable;
+use alloy_primitives::Address;
+
 #[cfg(feature = "rayon")]
-pub use rayon::*;
+use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 
-#[cfg(not(feature = "rayon"))]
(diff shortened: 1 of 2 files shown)


gakonst (approved):
Nice refactor! The new `try_recover_signers` API is a clean abstraction for combining decode + recovery in a single parallel pass.

**Code review:**

1. **Flat module structure** - Good call eliminating the nested `rayon`/`iter` modules. The `#[cfg]` dispatch at the function level is cleaner.

2. **Trait bounds** - The `TryRecoverItems` and `TryRecoverFn` helper traits handle the `Sync` requirement for rayon nicely. The blanket impls keep usage ergonomic.

3. **RPC integration** - The testing.rs change demonstrates the API well. Replacing the sequential `recover_raw_transaction` loop with `try_recover_signers` is a nice win for batch operations.

4. **Error handling** - Using `.or(Err(EthApiError::InvalidTransactionSignature))` loses the original error context. Consider using `.map_err(|_| ...)` to at least log the underlying error, but this is minor for testing endpoints.

LGTM :shipit:

PR #21110, opened 2026-01-15
refactor(chain-state): extract blocks_to_chain helper

mattsse:
## Summary

Refactors `to_chain_notification` to use a dedicated `blocks_to_chain` helper that optimizes chain construction:

- **Single-block case**: Uses `Chain::from_block` directly
- **Multi-block case**: Initializes with first block via `Chain::from_block`, then appends the rest
- Avoids creating an empty `Chain::default()` and repeatedly mutating it
- Deduplicates logic between `Commit` and `Reorg` variants

this removes a few extra/redundant clones per update but needs followup work to fix properly


DaniPopes (approved):

yongkangc (approved):

Rjected (approved):
lgtm

PR #21112, opened 2026-01-16
chore(primitives-traits): remove duplicate feature entry

rejected-l:
Remove duplicated "revm-primitives/serde" entry in the serde feature list.


mattsse (commented):
we can keep this

PR #21113, opened 2026-01-16
feat: bal-devnet-1 branch fix

Soubhik-10:
The current branch is broken


crates/ethereum/evm/src/build.rs
```diff
@@ -112,11 +125,17 @@ where
...
         };
 
         Ok(Block {
             header,
-            body: BlockBody { transactions, ommers: Default::default(), withdrawals },
+            body: BlockBody {
+                transactions,
+                ommers: Default::default(),
+                withdrawals,
+                block_access_list: built_block_access_list.cloned(),
```
mattsse, line 137: I believe this we need to remove now, no more bal in body

crates/payload/validator/src/amsterdam.rs
```diff
@@ -0,0 +1,23 @@
+//! Amsterdam rules for new payloads.
+
+use alloy_rpc_types_engine::PayloadError;
+use reth_primitives_traits::BlockBody;
+
+/// Checks that block body contains withdrawals if Amsterdam is active and vv.
+#[inline]
+pub fn ensure_well_formed_fields<T: BlockBody>(
```
mattsse, line 8: then we also dont need this check

mattsse (changes requested):

jenpaff:
superseded by https://github.com/paradigmxyz/reth/pull/21203

PR #21115, opened 2026-01-16
refactor(db): use hashed state as canonical state representation

gakonst:
## Summary

This PR removes the need for `PlainAccountState` and `PlainStorageState` tables by writing directly to `HashedAccounts` and `HashedStorages` during execution. **This is now configurable via CLI flags.**

## Motivation

The current architecture maintains duplicate state:
- **Plain state tables** (`PlainAccountState`, `PlainStorageState`) - indexed by unhashed addresses/slots
- **Hashed state tables** (`HashedAccounts`, `HashedStorages`) - indexed by keccak256 hashes

The `AccountHashingStage` and `StorageHashingStage` convert from plain to hashed state, which is required for Merkle trie operations. This creates:
- ~2x storage overhead for state data
- Additional I/O during sync
- Pipeline complexity

### CLI Flags

The hashed state mode is now **configurable via CLI**:

```bash
# Start node with hashed state enabled (for new databases only)
reth node --storage.use-hashed-state

# View current storage settings
reth db settings get

# Change storage settings on existing database (requires resync)
reth db settings set use_hashed_state true
```

### StorageSettings

A new `use_hashed_state` field has been added to `StorageSettings`:

```rust
pub struct StorageSettings {
    // ... existing fields ...
    /// Whether to use hashed state tables as canonical state representation
    pub use_hashed_state: bool,
}
```

### StorageArgs

New CLI argument struct with all storage configuration options:

| Flag | Description |
|------|-------------|
| `--storage.use-hashed-state` | Use hashed tables as canonical state |
| `--storage.receipts-in-static-files` | Store receipts in static files |
| `--storage.transaction-senders-in-static-files` | Store senders in static files |
| `--storage.account-changesets-in-static-files` | Store changesets in static files |
| `--storage.storages-history-in-rocksdb` | Store storage history in RocksDB |
| `--storage.transaction-hash-numbers-in-rocksdb` | Store tx hashes in RocksDB |
| `--storage.account-history-in-rocksdb` | Store account history in RocksDB |

### Provider Layer
- `LatestStateProviderRef`: Reads from `HashedAccounts`/`HashedStorages` with keccak256 hashing at lookup boundary (when `use_hashed_state` is enabled)
- `HistoricalStateProviderRef`: Uses hashed tables for "in plain state" lookups
- `DatabaseProvider`: 
  - `write_state_changes` writes directly to hashed tables
  - `write_state_reverts` uses hashed tables
  - `populate_bundle_state` uses hashed cursors

### Stages
- `AccountHashingStage`: Now a no-op when `use_hashed_state` is enabled
- `StorageHashingStage`: Now a no-op when `use_hashed_state` is enabled

The stages are kept for pipeline compatibility but perform no work since execution handles hashing.

### New Files
- `crates/node/core/src/args/storage.rs` - StorageArgs CLI struct
- `crates/storage/db-api/src/models/metadata.rs` - Updated StorageSettings

### Updated Files
- `crates/node/core/src/node_config.rs` - Added `storage: StorageArgs` field
- `crates/cli/commands/src/node.rs` - Wired StorageArgs into NodeCommand
- `crates/cli/commands/src/db/settings.rs` - Added `use_hashed_state` setting management

## Benefits

| Before | After |
|--------|-------|
| Write to plain state → Hash to hashed state | Write directly to hashed state |
| 2 tables per state type | 1 table per state type |
| Hashing stages required | Hashing stages are no-ops |
| Not configurable | Configurable via CLI |

## Testing

- [x] `cargo check -p reth-provider --lib` passes
- [x] `cargo check -p reth-stages --lib` passes
- [x] `cargo check -p reth` passes
- [x] `cargo test -p reth-node-core --lib -- storage` passes (4 tests)
- [ ] Full test suite (requires CI)

## Migration Notes

- Existing databases continue to work with default settings (plain state mode)
- Enable hashed state for **new databases only** with `--storage.use-hashed-state`
- Changing `use_hashed_state` on an existing database requires a full resync
- Plain state tables are no longer read/written when hashed state is enabled

## Follow-up Work

- Remove `PlainAccountState` and `PlainStorageState` from table schema (after deprecation period)
- Update CLI commands that reference plain state tables
- Update benchmarks and examples
- Add database migration for existing nodes

### DIFF
--- a/crates/cli/commands/src/stage/drop.rs
+++ b/crates/cli/commands/src/stage/drop.rs
@@ -134,6 +134,8 @@ impl<C: ChainSpecParser> Command<C> {
                 if provider_rw.cached_storage_settings().use_hashed_state {
                     tx.clear::<tables::HashedAccounts>()?;
                     tx.clear::<tables::HashedStorages>()?;
+                    reset_stage_checkpoint(tx, StageId::AccountHashing)?;
+                    reset_stage_checkpoint(tx, StageId::StorageHashing)?;
                 } else {
                     tx.clear::<tables::PlainAccountState>()?;
                     tx.clear::<tables::PlainStorageState>()?;
--- a/crates/cli/commands/src/stage/drop.rs
+++ b/crates/cli/commands/src/stage/drop.rs
@@ -131,8 +131,13 @@ impl<C: ChainSpecParser> Command<C> {
                 reset_stage_checkpoint(tx, StageId::SenderRecovery)?;
             }
             StageEnum::Execution => {
-                tx.clear::<tables::PlainAccountState>()?;
-                tx.clear::<tables::PlainStorageState>()?;
+                if provider_rw.cached_storage_settings().use_hashed_state {
+                    tx.clear::<tables::HashedAccounts>()?;
+                    tx.clear::<tables::HashedStorages>()?;
+                } else {
+                    tx.clear::<tables::PlainAccountState>()?;
+                    tx.clear::<tables::PlainStorageState>()?;
+                }
                 tx.clear::<tables::AccountChangeSets>()?;
                 tx.clear::<tables::StorageChangeSets>()?;
                 tx.clear::<tables::Bytecodes>()?;
--- a/crates/engine/tree/src/tree/tests.rs
+++ b/crates/engine/tree/src/tree/tests.rs
@@ -221,7 +221,7 @@ impl TestHarness {
             EngineApiKind::Ethereum,
             evm_config,
             changeset_cache,
-            false,
+            provider.cached_storage_settings().use_hashed_state,
         );
 
         let block_builder = TestBlockBuilder::default().with_chain_spec((*chain_spec).clone());
--- a/crates/engine/tree/src/tree/tests.rs
+++ b/crates/engine/tree/src/tree/tests.rs
@@ -221,6 +221,7 @@ impl TestHarness {
             EngineApiKind::Ethereum,
             evm_config,
             changeset_cache,
+            false,
         );
 
         let block_builder = TestBlockBuilder::default().with_chain_spec((*chain_spec).clone());
--- a/crates/node/core/src/args/storage.rs
+++ b/crates/node/core/src/args/storage.rs
@@ -69,8 +69,6 @@ impl StorageArgs {
     /// Fields that are `None` in the CLI args will use the base settings value.
     /// Fields that are `Some` will override the base settings.
     pub fn apply_to(&self, mut base: StorageSettings) -> StorageSettings {
-        base.use_hashed_state = self.use_hashed_state;
-
         if let Some(v) = self.receipts_in_static_files {
             base.receipts_in_static_files = v;
         }
@@ -120,7 +118,6 @@ mod tests {
         let args = CommandParser::<StorageArgs>::parse_from(["reth"]).args;
         assert_eq!(args, default_args);
         assert!(!args.v2);
-        assert!(!args.use_hashed_state);
     }
 
     #[test]
@@ -130,17 +127,9 @@ mod tests {
     }
 
     #[test]
-    fn test_use_hashed_state_flag() {
-        let args =
-            CommandParser::<StorageArgs>::parse_from(["reth", "--storage.use-hashed-state"]).args;
-        assert!(args.use_hashed_state);
-    }
-
-    #[test]
     fn test_storage_settings_override() {
         let args = CommandParser::<StorageArgs>::parse_from([
             "reth",
-            "--storage.use-hashed-state",
             "--storage.receipts-in-static-files",
             "true",
         ])
@@ -148,7 +137,6 @@ mod tests {
 
         let settings = args.storage_settings();
 
-        assert!(settings.use_hashed_state);
         assert!(settings.receipts_in_static_files);
         assert!(!settings.transaction_senders_in_static_files);
     }
@@ -157,7 +145,6 @@ mod tests {
     fn test_all_storage_flags() {
         let args = CommandParser::<StorageArgs>::parse_from([
             "reth",
-            "--storage.use-hashed-state",
             "--storage.receipts-in-static-files",
             "true",
             "--storage.transaction-senders-in-static-files",
@@ -175,7 +162,6 @@ mod tests {
 
         let settings = args.apply_to(StorageSettings::default());
 
-        assert!(settings.use_hashed_state);
         assert!(settings.receipts_in_static_files);
         assert!(settings.transaction_senders_in_static_files);
         assert!(settings.account_changesets_in_static_files);
--- a/crates/node/core/src/args/storage.rs
+++ b/crates/node/core/src/args/storage.rs
@@ -50,15 +50,15 @@ pub struct StorageArgs {
     #[arg(long = "storage.account-changesets-in-static-files")]
     pub account_changesets_in_static_files: Option<bool>,
 
-    /// Store storage history in RocksDB instead of MDBX.
+    /// Store storage history in `RocksDB` instead of MDBX.
     #[arg(long = "storage.storages-history-in-rocksdb")]
     pub storages_history_in_rocksdb: Option<bool>,
 
-    /// Store transaction hash to number mapping in RocksDB instead of MDBX.
+    /// Store transaction hash to number mapping in `RocksDB` instead of MDBX.
     #[arg(long = "storage.transaction-hash-numbers-in-rocksdb")]
     pub transaction_hash_numbers_in_rocksdb: Option<bool>,
 
-    /// Store account history in RocksDB instead of MDBX.
+    /// Store account history in `RocksDB` instead of MDBX.
     #[arg(long = "storage.account-history-in-rocksdb")]
     pub account_history_in_rocksdb: Option<bool>,
 }
@@ -68,7 +68,7 @@ impl StorageArgs {
     ///
     /// Fields that are `None` in the CLI args will use the base settings value.
     /// Fields that are `Some` will override the base settings.
-    pub fn apply_to(&self, mut base: StorageSettings) -> StorageSettings {
+    pub const fn apply_to(&self, mut base: StorageSettings) -> StorageSettings {
         if let Some(v) = self.receipts_in_static_files {
             base.receipts_in_static_files = v;
         }
@@ -94,7 +94,7 @@ impl StorageArgs {
     /// Returns the default storage settings for new nodes, with CLI overrides applied.
     ///
     /// Uses v2 settings if `--storage.v2` is set, otherwise uses v1 settings.
-    pub fn storage_settings(&self) -> StorageSettings {
+    pub const fn storage_settings(&self) -> StorageSettings {
         let base = if self.v2 { StorageSettings::v2() } else { StorageSettings::v1() };
         self.apply_to(base)
     }
--- a/crates/node/core/src/args/storage.rs
+++ b/crates/node/core/src/args/storage.rs
@@ -1,11 +1,12 @@
-//! clap [Args](clap::Args) for storage mode configuration
+//! clap [Args](clap::Args) for storage configuration
 
 use clap::{ArgAction, Args};
+use reth_storage_api::StorageSettings;
 
-/// Parameters for storage mode configuration.
+/// Parameters for storage configuration.
 ///
 /// This controls whether the node uses v2 storage defaults (with `RocksDB` and static file
-/// optimizations) or v1/legacy storage defaults.
+/// optimizations) or v1/legacy storage defaults, plus additional storage settings.
 #[derive(Debug, Args, PartialEq, Eq, Clone, Copy, Default)]
 #[command(next_help_heading = "Storage")]
 pub struct StorageArgs {
@@ -33,6 +34,72 @@ pub struct StorageArgs {
     /// WARNING: Changing this setting on an existing database requires a full resync.
     #[arg(long = "storage.use-hashed-state", default_value_t = false)]
     pub use_hashed_state: bool,
+
+    /// Store receipts in static files instead of the database.
+    ///
+    /// This is enabled by default for new nodes. Existing nodes continue using their
+    /// current storage location unless migrated.
+    #[arg(long = "storage.receipts-in-static-files")]
+    pub receipts_in_static_files: Option<bool>,
+
+    /// Store transaction senders in static files instead of the database.
+    #[arg(long = "storage.transaction-senders-in-static-files")]
+    pub transaction_senders_in_static_files: Option<bool>,
+
+    /// Store account changesets in static files instead of the database.
+    #[arg(long = "storage.account-changesets-in-static-files")]
+    pub account_changesets_in_static_files: Option<bool>,
+
+    /// Store storage history in RocksDB instead of MDBX.
+    #[arg(long = "storage.storages-history-in-rocksdb")]
+    pub storages_history_in_rocksdb: Option<bool>,
+
+    /// Store transaction hash to number mapping in RocksDB instead of MDBX.
+    #[arg(long = "storage.transaction-hash-numbers-in-rocksdb")]
+    pub transaction_hash_numbers_in_rocksdb: Option<bool>,
+
+    /// Store account history in RocksDB instead of MDBX.
+    #[arg(long = "storage.account-history-in-rocksdb")]
+    pub account_history_in_rocksdb: Option<bool>,
+}
+
+impl StorageArgs {
+    /// Applies CLI overrides to the given base settings.
+    ///
+    /// Fields that are `None` in the CLI args will use the base settings value.
+    /// Fields that are `Some` will override the base settings.
+    pub fn apply_to(&self, mut base: StorageSettings) -> StorageSettings {
+        base.use_hashed_state = self.use_hashed_state;
+
+        if let Some(v) = self.receipts_in_static_files {
+            base.receipts_in_static_files = v;
+        }
+        if let Some(v) = self.transaction_senders_in_static_files {
+            base.transaction_senders_in_static_files = v;
+        }
+        if let Some(v) = self.account_changesets_in_static_files {
+            base.account_changesets_in_static_files = v;
+        }
+        if let Some(v) = self.storages_history_in_rocksdb {
+            base.storages_history_in_rocksdb = v;
+        }
+        if let Some(v) = self.transaction_hash_numbers_in_rocksdb {
+            base.transaction_hash_numbers_in_rocksdb = v;
+        }
+        if let Some(v) = self.account_history_in_rocksdb {
+            base.account_history_in_rocksdb = v;
+        }
+
+        base
+    }
+
+    /// Returns the default storage settings for new nodes, with CLI overrides applied.
+    ///
+    /// Uses v2 settings if `--storage.v2` is set, otherwise uses v1 settings.
+    pub fn storage_settings(&self) -> StorageSettings {
+        let base = if self.v2 { StorageSettings::v2() } else { StorageSettings::v1() };
+        self.apply_to(base)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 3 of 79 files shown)


crates/cli/commands/src/db/settings.rs
```diff
@@ -69,6 +69,16 @@ pub enum SetCommand {
         #[clap(action(ArgAction::Set))]
         value: bool,
     },
+    /// Use hashed state tables (HashedAccounts/HashedStorages) as canonical state
+    ///
+    /// When enabled, execution writes directly to hashed tables, eliminating need for
+    /// separate hashing stages. State reads come from hashed tables.
+    ///
+    /// WARNING: Changing this setting requires re-syncing the database.
```
Rjected: changing this from `true` to `false` requires resync but not the other way around

crates/stages/stages/src/stages/hashing_account.rs
```diff
@@ -142,113 +142,25 @@ where
...
-            });
-
-        Ok(ExecOutput { checkpoint, done: true })
+    ///
+    /// NOTE: This stage is now a no-op because the execution stage writes directly to
+    /// HashedAccounts. This stage exists only for backwards compatibility with existing pipelines.
+    fn execute(&mut self, _provider: &Provider, input: ExecInput) -> Result<ExecOutput, StageError> {
+        // Since execution now writes directly to HashedAccounts, this stage is a no-op.
+        // Just report that we're done up to the target block.
+        Ok(ExecOutput::done(input.checkpoint().with_block_number(input.target())))
```
Rjected: what runs in this stage should depend on whether the hashed state storage settings is set, we still need to run this stage if the user is using plain state

crates/stages/stages/src/stages/execution.rs
```diff
@@ -45,22 +45,22 @@ use super::missing_static_data_error;
...
 /// - [`tables::AccountChangeSets`]
 /// - [`tables::StorageChangeSets`]
 ///
 /// For unwinds we are accessing:
 /// - [`tables::BlockBodyIndices`] get tx index to know what needs to be unwinded
 /// - [`tables::AccountsHistory`] to remove change set and apply old values to
-/// - [`tables::PlainAccountState`] [`tables::StoragesHistory`] to remove change set and apply old
-///   values to [`tables::PlainStorageState`]
+/// - [`tables::HashedAccounts`] [`tables::StoragesHistory`] to remove change set and apply old
+///   values to [`tables::HashedStorages`]
```
Rjected: this should take into account that this only updates hashed tables when the hashed state storage setting is set

crates/stages/stages/src/stages/execution.rs
```diff
@@ -894,14 +894,14 @@ mod tests {
...
+            .put::<tables::HashedAccounts>(
+                keccak256(acc1),
                 Account { nonce: 0, balance: U256::ZERO, bytecode_hash: Some(code_hash) },
             )
             .unwrap();
         db_tx
-            .put::<tables::PlainAccountState>(
-                acc2,
+            .put::<tables::HashedAccounts>(
+                keccak256(acc2),
```
Rjected: these `put`s should depend on the new storage setting and retain the previous behavior if unset

crates/stages/stages/src/stages/hashing_account.rs
```diff
@@ -142,113 +142,25 @@ where
...
-            input.checkpoint.account_hashing_stage_checkpoint().unwrap_or_default();
-
-        stage_checkpoint.progress = stage_checkpoint_progress(provider)?;
-
-        Ok(UnwindOutput {
-            checkpoint: StageCheckpoint::new(unwind_progress)
-                .with_account_hashing_stage_checkpoint(stage_checkpoint),
-        })
+        // Since execution now manages HashedAccounts directly during unwind, this is a no-op.
+        Ok(UnwindOutput { checkpoint: StageCheckpoint::new(input.unwind_to) })
```
Rjected: again, this should depend on whether the storage setting is set

crates/stages/stages/src/stages/hashing_storage.rs
```diff
@@ -76,120 +76,29 @@ where
...
+    /// NOTE: This stage is now a no-op because the execution stage writes directly to
+    /// HashedStorages. This stage exists only for backwards compatibility with existing pipelines.
+    fn execute(
+        &mut self,
+        _provider: &Provider,
+        input: ExecInput,
+    ) -> Result<ExecOutput, StageError> {
+        // Since execution now writes directly to HashedStorages, this stage is a no-op.
+        // Just report that we're done up to the target block.
+        Ok(ExecOutput::done(input.checkpoint().with_block_number(input.target())))
```
Rjected: same comment re: keeping old behavior if storage setting is unset

crates/stages/stages/src/stages/hashing_storage.rs
```diff
@@ -76,120 +76,29 @@ where
...
-            input.checkpoint.storage_hashing_stage_checkpoint().unwrap_or_default();
-
-        stage_checkpoint.progress = stage_checkpoint_progress(provider)?;
-
-        Ok(UnwindOutput {
-            checkpoint: StageCheckpoint::new(unwind_progress)
-                .with_storage_hashing_stage_checkpoint(stage_checkpoint),
-        })
+        // Since execution now manages HashedStorages directly during unwind, this is a no-op.
+        Ok(UnwindOutput { checkpoint: StageCheckpoint::new(input.unwind_to) })
```
Rjected: same comment re: keeping old behavior if storage setting is unset

crates/stages/stages/src/stages/mod.rs
```diff
@@ -122,15 +122,15 @@ mod tests {
...
+            .put::<tables::HashedAccounts>(
+                keccak256(address!("0x1000000000000000000000000000000000000000")),
                 Account { nonce: 0, balance: U256::ZERO, bytecode_hash: Some(code_hash) },
             )
             .unwrap();
         provider_rw
             .tx_ref()
-            .put::<tables::PlainAccountState>(
-                address!("0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b"),
+            .put::<tables::HashedAccounts>(
```
Rjected: same comment re: making these `put`s dependent on the storage setting

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -31,6 +31,19 @@ pub struct StorageSettings {
...
+    /// canonical state representation instead of plain state tables
+    /// (`PlainAccountState`/`PlainStorageState`).
+    ///
+    /// When enabled:
+    /// - Execution writes directly to hashed tables, eliminating need for separate hashing stages
+    /// - State reads come from hashed tables
+    /// - `AccountHashingStage` and `StorageHashingStage` become no-ops
+    ///
+    /// WARNING: This setting is only configurable at database creation; changing it later
+    /// requires re-syncing.
```
Rjected: same comment re: nuance for changing true -> false vs false -> true

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -430,13 +431,17 @@ impl<
...
+            HistoryInfo::InPlainState | HistoryInfo::MaybeInPlainState => {
+                let hashed_address = alloy_primitives::keccak256(address);
+                let hashed_slot = alloy_primitives::keccak256(storage_key);
+                Ok(self
+                    .tx()
+                    .cursor_dup_read::<tables::HashedStorages>()?
+                    .seek_by_key_subkey(hashed_address, hashed_slot)?
+                    .filter(|entry| entry.key == hashed_slot)
+                    .map(|entry| entry.value)
+                    .or(Some(StorageValue::ZERO)))
```
Rjected: this needs to depend on the storage setting

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -262,7 +262,8 @@ impl<
                     .map(|account_before| account_before.info)
             }
             HistoryInfo::InPlainState | HistoryInfo::MaybeInPlainState => {
-                Ok(self.tx().get_by_encoded_key::<tables::PlainAccountState>(address)?)
+                let hashed_address = alloy_primitives::keccak256(address);
+                Ok(self.tx().get_by_encoded_key::<tables::HashedAccounts>(&hashed_address)?)
```
Rjected: this needs to depend on the new storage setting

crates/node/core/src/args/storage.rs
```diff
@@ -0,0 +1,169 @@
+//! clap [Args](clap::Args) for storage configuration
```
Rjected, line 1: this entire file doesn't need to exist and is redundant. just add a `--db` flag for the new hashed state storage setting, or whatever we currently do for rocksdb

Rjected (changes requested):
tons of places where this now just looks at the hashed table even though they need to depend on the storage setting for hashed state, this PR needs to change from just replacing plain -> hashed reads and writes, to going to the correct tables depending on the newly introduced storage setting

Rjected:
status: works but unwind causes state root mismatch

crates/node/core/src/args/storage.rs
```diff
@@ -33,28 +34,153 @@ pub struct StorageArgs {
...
+    #[arg(long = "storage.storages-history-in-rocksdb")]
+    pub storages_history_in_rocksdb: Option<bool>,
+
+    /// Store transaction hash to number mapping in `RocksDB` instead of MDBX.
+    #[arg(long = "storage.transaction-hash-numbers-in-rocksdb")]
+    pub transaction_hash_numbers_in_rocksdb: Option<bool>,
+
+    /// Store account history in `RocksDB` instead of MDBX.
+    #[arg(long = "storage.account-history-in-rocksdb")]
+    pub account_history_in_rocksdb: Option<bool>,
```
klkvr: should those just be `bool`s?

crates/node/core/src/args/storage.rs
```diff
@@ -33,28 +34,153 @@ pub struct StorageArgs {
...
+        base
+    }
+
+    /// Returns the default storage settings for new nodes, with CLI overrides applied.
+    ///
+    /// Uses v2 settings if `--storage.v2` is set, otherwise uses v1 settings.
+    pub const fn storage_settings(&self) -> StorageSettings {
+        let base = if self.v2 { StorageSettings::v2() } else { StorageSettings::v1() };
+        self.apply_to(base)
+    }
```
klkvr: i don't think this is ever called outside of tests? also some of the settings here seem redundant to `StaticFilesArgs` — e.g `storage.account-changesets-in-static-files` vs `static-files.account-change-sets`

crates/stages/stages/src/stages/execution.rs
```diff
@@ -464,6 +465,11 @@ where
         // write output
         provider.write_state(&state, OriginalValuesKnown::Yes, StateWriteConfig::default())?;
 
+        if provider.cached_storage_settings().use_hashed_state {
+            let hashed_state = state.hash_state_slow::<KeccakKeyHasher>();
+            provider.write_hashed_state(&hashed_state.into_sorted())?;
+        }
+
```
klkvr: let's add a note on what exactly happens here (i.e `write_state` skips writing state if `use_hashed_state` is set)

also did we consider just making `write_state` handle the hashing to make callsite not care about this?

crates/stages/stages/src/stages/hashing_account.rs
```diff
@@ -234,10 +250,14 @@ where
         provider: &Provider,
         input: UnwindInput,
     ) -> Result<UnwindOutput, StageError> {
+        // NOTE: this runs in both v1 and v2 mode. In v2 mode, execution writes
+        // directly to `HashedAccounts`, but the unwind must still revert those
+        // entries here because `MerkleUnwind` runs after this stage (in unwind
+        // order) and needs `HashedAccounts` to reflect the target block state
+        // before it can verify the state root.
```
klkvr, line 257: should `ExecutionStage` unwind method be modified to account for that? given that this unwinds just the state wouldn't this result in `ExecutionStage` unwinding the same tables again?

crates/trie/db/src/state.rs
```diff
@@ -43,7 +46,10 @@ pub trait DatabaseStateRoot<'a, TX>: Sized {
     ///
     /// The updated state root.
     fn incremental_root(
-        provider: &'a (impl ChangeSetReader + StorageChangeSetReader + DBProvider<Tx = TX>),
+        provider: &'a (impl ChangeSetReader
+                 + StorageChangeSetReader
+                 + StorageSettingsCache
+                 + DBProvider<Tx = TX>),
```
klkvr, line 52: imo could make sense to make `StorageSettingsCache` a supertrait of `StorageChangeSetReader` just to avoid this mess :D

crates/node/core/src/node_config.rs
```diff
@@ -410,6 +416,13 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
 
         s = s.with_use_hashed_state(self.storage.use_hashed_state);
 
+        if s.use_hashed_state {
+            s = s.with_storage_changesets_in_static_files(true);
+        }
+        if s.storage_changesets_in_static_files {
+            s = s.with_use_hashed_state(true);
+        }
```
klkvr, line 412: let's document why those imply each other, and i guess those could just be a single flag?

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -159,17 +159,77 @@ impl<'b, Provider: DBProvider + ChangeSetReader + StorageChangeSetReader + Block
...
+        address: Address,
+        storage_key: StorageSlotKey,
+    ) -> ProviderResult<Option<StorageValue>>
+    where
+        Provider: StorageSettingsCache + RocksDBProviderFactory + NodePrimitivesProvider,
+    {
+        let lookup_key = if self.provider.cached_storage_settings().use_hashed_state {
+            storage_key.to_hashed()
+        } else {
+            storage_key.as_b256()
```
klkvr, line 203: would be nice to assert that this is indeed a plain key at least in debug builds

klkvr (commented):
still reading through but it mostly looks good to me, will be reading blockchain/consistent providers files now

crates/storage/provider/src/providers/consistent.rs
```diff
@@ -257,10 +263,16 @@ impl<N: ProviderNodeTypes> ConsistentProvider<N> {
             };
 
             // match storage.
-            match account_state.2.entry(old_storage.key) {
+            match account_state.2.entry(old_storage.key.as_b256()) {
```
klkvr, line 266: is this right? iiuc this fn is supposed to populate the plain `BundleState` but might insert hashed keys? i guess right now it's not really feasible to reconstruct the plain state changes for latest blocks

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1243,7 +1245,78 @@ impl<TX: DbTx + 'static, N: NodeTypesForProvider> DatabaseProvider<TX, N> {
...
+    /// `HashedAccounts`/`HashedStorages`. Addresses are hashed via `keccak256` for DB lookups,
+    /// while storage keys from changesets are assumed to already be hashed and are used as-is.
+    /// The output `BundleStateInit`/`RevertsInit` structures remain keyed by plain address.
+    fn populate_bundle_state_hashed(
+        &self,
+        account_changeset: Vec<(u64, AccountBeforeTx)>,
+        storage_changeset: Vec<(BlockNumberAddress, ChangesetEntry)>,
+        hashed_accounts_cursor: &mut impl DbCursorRO<tables::HashedAccounts>,
+        hashed_storage_cursor: &mut impl DbDupCursorRO<tables::HashedStorages>,
+    ) -> ProviderResult<(BundleStateInit, RevertsInit)> {
```
klkvr, line 1276: would be cool to have a cursor wrapper that would allow to avoid duplicating methods like this

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2119,23 +2231,47 @@ impl<TX: DbTx + 'static, N: NodeTypes> StorageReader for DatabaseProvider<TX, N>
         &self,
         addresses_with_keys: impl IntoIterator<Item = (Address, impl IntoIterator<Item = B256>)>,
     ) -> ProviderResult<Vec<(Address, Vec<StorageEntry>)>> {
-        let mut plain_storage = self.tx.cursor_dup_read::<tables::PlainStorageState>()?;
+        if self.cached_storage_settings().use_hashed_state {
```
klkvr, line 2246: this fn would also benefit from such wrapper

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2348,11 +2488,18 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
+                    // we can't recover the plain slot keys needed for the changeset format.
+                    // This is safe in practice because SELFDESTRUCT is deprecated (EIP-6780)
+                    // and the explicit `storage_revert` entries already cover all accessed slots.
+                    //
                     // TODO(mediocregopher): This could be rewritten in a way which doesn't require
                     // collecting wiped entries into a Vec like this, see
                     // `write_storage_trie_changesets`.
                     let mut wiped_storage = Vec::new();
-                    if wiped {
+                    if wiped && !use_hashed_state {
```
klkvr: this needs to be updated to actually write hashed changesets

crates/node/core/src/node_config.rs
Rjected: i want to do this https://github.com/paradigmxyz/reth/pull/22042 after we merge this

crates/node/core/src/args/storage.rs
Rjected: just removed all these

crates/storage/provider/src/providers/database/provider.rs
Rjected: updated, see https://github.com/paradigmxyz/reth/pull/21115/changes/3431b6538ad86ba20129b5a4683e1f9b89fcc4a5

crates/storage/provider/src/providers/consistent.rs
Rjected: given https://github.com/paradigmxyz/reth/pull/22090, going to ignore this for now, we can fix properly later

crates/stages/stages/src/stages/execution.rs
Rjected: added a note, have not considered it much

crates/stages/stages/src/stages/hashing_account.rs
Rjected: yes, although we still need to have the code for unwinding the hashed state in execution, in case we are unwinding to the checkpoint due to a bad block in exec

maybe we could check the unwind input and run different code for bad block vs "regular" unwind but this is probably fine for now

crates/node/core/src/args/storage.rs
Rjected: removed the other `storage` args

crates/storage/provider/src/providers/state/historical.rs
Rjected: added a debug_assert

crates/engine/tree/src/tree/tests.rs
```diff
@@ -221,6 +221,7 @@ impl TestHarness {
             EngineApiKind::Ethereum,
             evm_config,
             changeset_cache,
+            false,
```
joshieDo: why false

crates/cli/commands/src/stage/drop.rs
```diff
@@ -131,8 +131,13 @@ impl<C: ChainSpecParser> Command<C> {
                 reset_stage_checkpoint(tx, StageId::SenderRecovery)?;
             }
             StageEnum::Execution => {
-                tx.clear::<tables::PlainAccountState>()?;
-                tx.clear::<tables::PlainStorageState>()?;
+                if provider_rw.cached_storage_settings().use_hashed_state {
+                    tx.clear::<tables::HashedAccounts>()?;
+                    tx.clear::<tables::HashedStorages>()?;
```
joshieDo, line 136: needs to reset the stage checkpoitns similar to stageenum hashing types

crates/storage/storage-api/src/hashing.rs
```diff
@@ -7,6 +7,8 @@ use reth_db_models::AccountBeforeTx;
 use reth_primitives_traits::{Account, StorageEntry};
 use reth_storage_errors::provider::ProviderResult;
 
+use crate::ChangesetEntry;
```
joshieDo: collapse, rm paragraph

crates/storage/storage-api/src/history.rs
```diff
@@ -3,9 +3,10 @@ use auto_impl::auto_impl;
 use core::ops::{RangeBounds, RangeInclusive};
 use reth_db_api::models::BlockNumberAddress;
 use reth_db_models::AccountBeforeTx;
-use reth_primitives_traits::StorageEntry;
 use reth_storage_errors::provider::ProviderResult;
 
+use crate::ChangesetEntry;
```
joshieDo: collapse rm paragraph

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1333,10 +1334,12 @@ impl RocksDBProvider {
 
             // Iterate through storage reverts - these are exactly the slots that have
             // changesets written, ensuring history indices match changeset entries.
+            let use_hashed = ctx.storage_settings.use_hashed_state;
             for storage_block_reverts in reverts.storage {
                 for revert in storage_block_reverts {
                     for (slot, _) in revert.storage_revert {
-                        let key = B256::new(slot.to_be_bytes());
+                        let plain_key = B256::new(slot.to_be_bytes());
+                        let key = if use_hashed { keccak256(plain_key) } else { plain_key };
```
joshieDo: rocksdb always writes hashed, never plain. no need to check storage settings

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -317,7 +317,9 @@ impl RocksDBProvider {
 
             let unique_keys: HashSet<_> = changesets
                 .into_iter()
-                .map(|(block_addr, entry)| (block_addr.address(), entry.key, checkpoint + 1))
+                .map(|(block_addr, entry)| {
+                    (block_addr.address(), entry.key.as_b256(), checkpoint + 1)
```
joshieDo, line 322: add doc above that this is a hashed entry.key

crates/storage/provider/src/providers/state/latest.rs
joshieDo: add a doc that expects a plain storage key slot

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -412,32 +506,23 @@ impl<
             + NodePrimitivesProvider,
     > StateProvider for HistoricalStateProviderRef<'_, Provider>
 {
-    /// Get storage.
     fn storage(
```
joshieDo, line 510: re-add doc mentioning that it expects plain

Rjected (approved):
lgtm

joshieDo (approved):

PR #21116, opened 2026-01-16
fix(payload): correct pattern matching in `is_frozen()`

rejected-l:
`is_frozen()` used struct variant syntax `Self::Freeze { .. }` for a tuple variant `Freeze(Payload)`, causing it to always return false. This meant frozen payloads were never detected, and the builder kept spawning new build jobs even when it shouldn't. Fixed by using the correct tuple pattern `Self::Freeze(..)`.


mattsse (commented):
this is equivalent

PR #21119, opened 2026-01-16
perf(db): remove unnecessary Vec clone in benchmark utils

Aleksandr1732:
The set_up_db function in reth-db benchmark utilities was calling pair.clone() to iterate over test data, which unnecessarily cloned the entire vector including all Bytes fields for each tuple element.


mattsse (commented):
this is only utils, so this is fine o keep as is

PR #21122, opened 2026-01-16
fix(rpc): return error for eth_getLogs on pruned blocks

kustrun:
## Summary

This PR fixes an inconsistency in RPC error handling when querying pruned blocks.

## Problem

Currently, `eth_getLogs` returns an empty array `[]` when querying blocks that have been pruned, while other RPC methods like `eth_getBalance` correctly return an error:

```json
// eth_getBalance on pruned block (current behavior - correct)
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"state at block #12369623 is pruned"}}

// eth_getLogs on pruned block (current behavior - incorrect)
{"jsonrpc":"2.0","id":1,"result":[]}
```

This inconsistency can mislead users and applications into thinking no logs exist for those blocks, when in reality the data is simply unavailable due to pruning.

## Solution

Added a pruning check in `logs_for_filter()` that verifies the requested block range doesn't overlap with pruned receipts data. When it does, the same error as `eth_getBalance` is returned:

```json
// eth_getLogs on pruned block (after fix)
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"state at block #12369623 is pruned"}}
```

The check only verifies the `Receipts` prune segment (not `ContractLogs`), since `ContractLogs` pruning still keeps some receipts and can return partial results.

## Changes

- Added `PruneCheckpointReader` trait bound to filter-related impl blocks
- Added `check_receipts_pruning()` helper method in `EthFilterInner`
- Added pruning check for both block range queries and `AtBlockHash` queries
- Extended `MockEthProvider` to support configurable prune checkpoints for testing
- Added unit tests for the pruning check

## Test Plan

- `cargo test -p reth-rpc` - all 39 tests pass
- `cargo clippy -p reth-rpc --all-features` - no warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10425,6 +10425,7 @@ dependencies = [
  "reth-node-api",
  "reth-primitives-traits",
  "reth-provider",
+ "reth-prune-types",
  "reth-revm",
  "reth-rpc-api",
  "reth-rpc-convert",
--- a/crates/rpc/rpc-eth-api/src/node.rs
+++ b/crates/rpc/rpc-eth-api/src/node.rs
@@ -8,7 +8,8 @@ use reth_node_api::{FullNodeComponents, NodePrimitives, PrimitivesTy};
 use reth_primitives_traits::{BlockTy, HeaderTy, ReceiptTy, TxTy};
 use reth_rpc_eth_types::EthStateCache;
 use reth_storage_api::{
-    BlockReader, BlockReaderIdExt, StageCheckpointReader, StateProviderFactory,
+    BlockReader, BlockReaderIdExt, PruneCheckpointReader, StageCheckpointReader,
+    StateProviderFactory,
 };
 use reth_transaction_pool::{PoolTransaction, TransactionPool};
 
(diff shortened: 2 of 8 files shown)


mattsse (changes requested):
this makes sense directionally but I'd like to change this a bit so that this doesnt require a disk read per request

what we have for block height:

https://github.com/paradigmxyz/reth/blob/346cc0da71a10393a6426aa77952528e44e8dafb/crates/storage/provider/src/providers/static_file/manager.rs#L1675-L1684

we also need to replicate for receipts as well

I suggest we follow this and track smth like receipts height as well

https://github.com/paradigmxyz/reth/blob/346cc0da71a10393a6426aa77952528e44e8dafb/crates/storage/provider/src/providers/static_file/manager.rs#L1207-L1214


a smaller easier fix would be to include earliers: u64 in 

https://github.com/paradigmxyz/reth/blob/346cc0da71a10393a6426aa77952528e44e8dafb/crates/chainspec/src/info.rs#L4-L6

as well, then we can use this in the logs endpoint

mattsse (commented):
superseded by https://github.com/paradigmxyz/reth/pull/21304

PR #21123, opened 2026-01-16
refactor: use `BlockExecutionOutcome` in `ExecutedBlock`

klkvr:
Changes `ExecutedBlock` to use `BlockExecutionOutcome` instead of `ExecutionOutcome`. `ExecutionOutcome` is only useful when dealing with outcomes for multiple blocks and causes issues in engine code where we need to invoke consensus methods taking single block's outcome.

I've renamed `BlockExecutionOutcome::state` -> `BlockExecutionOutcome::bundle` to make it easier to make this change less distruptive.

### DIFF
--- a/crates/evm/execution-types/src/execute.rs
+++ b/crates/evm/execution-types/src/execute.rs
@@ -1,3 +1,5 @@
+use alloy_primitives::{Address, B256, U256};
+use reth_primitives_traits::{Account, Bytecode};
 use revm::database::BundleState;
 
 pub use alloy_evm::block::BlockExecutionResult;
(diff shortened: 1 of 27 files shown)


crates/evm/execution-types/src/execute.rs
```diff
@@ -21,5 +23,24 @@ pub struct BlockExecutionOutput<T> {
     #[deref_mut]
     pub result: BlockExecutionResult<T>,
     /// The changed state of the block after execution.
-    pub state: BundleState,
+    pub bundle: BundleState,
```
mattsse: should we keep this as state perhaps?
fine either way

crates/evm/execution-types/src/execute.rs
```diff
@@ -21,5 +23,24 @@ pub struct BlockExecutionOutput<T> {
...
+    pub fn account(&self, address: &Address) -> Option<Option<Account>> {
+        self.bundle.account(address).map(|a| a.info.as_ref().map(Into::into))
+    }
+
+    /// Get storage if value is known.
+    ///
+    /// This means that depending on status we can potentially return `U256::ZERO`.
+    pub fn storage(&self, address: &Address, storage_key: U256) -> Option<U256> {
+        self.bundle.account(address).and_then(|a| a.storage_slot(storage_key))
+    }
```
mattsse, line 59: I feel like all of these should exist on bundlestate as well

mattsse (approved):
checks out, makes sense

mattsse (approved):

PR #21130, opened 2026-01-16
perf: start saving cache sooner

klkvr:
Based on https://github.com/paradigmxyz/reth/pull/21123

Saving cache might take a long time during which we hold a mutex lock, not allowing any blocks to be processed. We are currently waiting around 150 ms

This PR moves `terminate_caching` call triggering the cache save to be done immediately after execution as soon as we have the state. This makes creation of the cache race the potential state root work left.

We don't want to save cache for invalid blocks so main engine task now notifies the prewarm thread once block validation passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9038,6 +9038,7 @@ dependencies = [
  "reth-ethereum-primitives",
  "reth-evm",
  "reth-evm-ethereum",
+ "reth-execution-types",
  "reth-primitives-traits",
  "reth-provider",
  "reth-revm",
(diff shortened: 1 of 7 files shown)


shekhirin (approved):

DaniPopes (approved):

yongkangc (approved):

mattsse (commented):
this will optimistically commit to the cache, right?

so in case the block is actually invalid, then the cache is polluted and we throw it away, tho imo that is acceptable

mediocregopher (approved):

PR #21132, opened 2026-01-16
feat(trie): add dynamic multiproof worker scaling

yongkangc:
**Summary**

Implements dynamic worker scaling for multiproof computation to reduce latency during big blocks.

**Problem**: Queue saturation (~1800 pending jobs, ~30 workers) → 700ms+ latency  
**Solution**: Elastic worker pools that scale based on predicted workload

**Motivation**

image

- On big blocks, pending queue gets jammed causing proof calculation to take 650ms. this is bad. 
- basically multiproof workers are all busy -> pending queue jams up -> proof calc time increases

**Key Changes**

| Component | Change |
|-----------|--------|
| `WorkloadPredictor` | Lock-free EMA predictor (α=0.3, 1.25x headroom) |
| `ProofWorkerHandle` | Spawn closures as trait objects, CAS-based scaling |
| `MultiProofTask` | Hooks for `prepare_for_block()` / `record_block_workload()` |

Scaling Logic

- Pre-scale workers at block start based on EMA of previous blocks
- Max workers: 4x initial count, scale-up step: 4 workers/decision
- `Shutdown` job variants added for future scale-down support

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -85,25 +85,83 @@ type TrieNodeProviderResult = Result<Option<RevealedNode>, SparseTrieError>;
 /// The handle stores direct senders to both storage and account worker pools,
 /// eliminating the need for a routing thread. All handles share reference-counted
 /// channels, and workers shut down gracefully when all handles are dropped.
-#[derive(Debug, Clone)]
-pub struct ProofWorkerHandle {
+/// Type alias for spawn function closures used in dynamic worker scaling.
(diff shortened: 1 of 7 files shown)


crates/trie/parallel/src/proof_task.rs
```diff
@@ -739,6 +944,8 @@ enum StorageWorkerJob {
         /// Channel to send result back to original caller
         result_sender: Sender<TrieNodeProviderResult>,
     },
+    /// Graceful shutdown signal for dynamic scaling
+    Shutdown,
```
mediocregopher, line 927: we're adding a variant to deadcode?

PR #21133, opened 2026-01-16
perf(engine): defer trie overlay computation with LazyOverlay

mattsse:
## Summary

Introduces lazy resolution of ancestor overlay data to allow block execution to start before the trie input is fully computed.

### New `LazyOverlay` type (`reth-chain-state`)
- Captures `DeferredTrieData` handles without blocking
- Computes `TrieInputSorted` on first access via `get_or_init`
- Fast path: reuses tip block's cached overlay if anchor matches
- Slow path: merges ancestor blocks using existing `merge_batch` logic

### Unified `OverlaySource` enum (`reth-provider`)
Replaces three separate overlay fields with a single enum:
- `OverlaySource::Immediate` - pre-computed overlay data
- `OverlaySource::Lazy` - deferred computation via `LazyOverlay`

### Updated `payload_validator` (`reth-engine-tree`)
- New `get_parent_lazy_overlay()` creates overlay without blocking
- Removed synchronous `compute_trie_input()` and `merge_overlay_trie_input()`
- Removed unused metrics: `trie_input_duration`, `deferred_trie_wait_duration`

## Impact

The optimization allows execution to proceed immediately while overlay computation is deferred to when `OverlayStateProviderFactory` is accessed by payload processor workers, reducing latency on the critical path.

### DIFF
--- a/crates/chain-state/src/lazy_overlay.rs
+++ b/crates/chain-state/src/lazy_overlay.rs
@@ -10,9 +10,9 @@ use reth_trie::{updates::TrieUpdatesSorted, HashedPostStateSorted, TrieInputSort
 use std::sync::{Arc, OnceLock};
 use tracing::{debug, trace};
 
-/// Threshold for switching from extend_ref loop to merge_batch.
+/// Threshold for switching from `extend_ref` loop to `merge_batch`.
 ///
-/// Benchmarked crossover: extend_ref wins up to ~64 blocks, merge_batch wins beyond.
+/// Benchmarked crossover: `extend_ref` wins up to ~64 blocks, `merge_batch` wins beyond.
 const MERGE_BATCH_THRESHOLD: usize = 64;
 
 /// Inputs captured for lazy overlay computation.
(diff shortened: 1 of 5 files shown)


crates/chain-state/src/lazy_overlay.rs
```diff
@@ -0,0 +1,231 @@
...
+///
+/// # Fast Path vs Slow Path
+///
+/// - **Fast path**: If the tip block's cached `anchored_trie_input` is ready and its `anchor_hash`
+///   matches our expected anchor, we can reuse it directly (O(1)).
+/// - **Slow path**: Otherwise, we merge all ancestor blocks' trie data into a new overlay.
+#[derive(Clone)]
+pub struct LazyOverlay {
+    /// Computed result, cached after first access.
+    inner: Arc<OnceLock<TrieInputSorted>>,
```
mattsse, line 41: prehaps this should be a regular mutex instead

mediocregopher (approved):

Rjected (approved):
lgtm

yongkangc (approved):

PR #21135, opened 2026-01-16
fix(exex): prevent ExExManager deadlock when buffer clears after being full

### LINKED ISSUE #19665: ExEx stops receiving notifications if not processed fast enough
### Describe the bug

Sending a bunch of notifications without acknowledging seems to cause the ExExManager to stop sending notifications, even for new blocks.

### Steps to reproduce

1. Checkout this branch: https://github.com/paradigmxyz/reth/compare/main...meyer9:reth:meyer9/demo-exex-deadlock
2. Run the test
3. See that none of the notifications are received
4. Modify the `MAX_CAPACITY` const to a lower number (I think 256 works)
5. See that all of the notifications are properly received

### Platform(s)

_No response_

### Container Type

Not running in a container

Tested on `main` and `v1.8.2`

N/A

N/A

Archive (default)

_No response_

_No response_

### Code of Conduct

dhyaniarun1993:
# Description

This PR fixes a deadlock scenario in the `ExExManager` where the manager could stop processing new notifications permanently after its internal buffer filled up and subsequently cleared.

Closes https://github.com/paradigmxyz/reth/issues/19665

### DIFF
--- a/crates/exex/exex/src/manager.rs
+++ b/crates/exex/exex/src/manager.rs
@@ -36,7 +36,7 @@ use tokio_util::sync::{PollSendError, PollSender, ReusableBoxFuture};
 ///
 /// 1024 notifications in the buffer is 3.5 hours of mainnet blocks,
 /// or 17 minutes of 1-second blocks.
-pub const DEFAULT_EXEX_MANAGER_CAPACITY: usize = 128;
+pub const DEFAULT_EXEX_MANAGER_CAPACITY: usize = 1024;
 
 /// Default maximum number of blocks allowed in the WAL before emitting a warning.
 ///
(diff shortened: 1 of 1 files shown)


crates/exex/exex/src/manager.rs
```diff
@@ -536,6 +537,12 @@ where
         // Update capacity
         this.update_capacity();
 
+        // If the buffer was full and we made space, we need to wake up to accept new notifications
+        if buffer_full && this.buffer.len() < this.max_capacity {
+            debug!(target: "exex::manager", "Buffer has space again, waking up senders");
+            cx.waker().wake_by_ref();
+        }
```
mattsse, line 544: right because we have

> while this.buffer.len() < this.max_capacity { 

in front of the poll loop -.-

mattsse (approved):
very nice find, tysm

this should do it

PR #21137, opened 2026-01-16
perf: make Chain use DeferredTrieData

Rjected:
This makes `Chain` use the `DeferredTrieData` instead of `HashedPostStateSorted` and `TrieUpdatesSorted`.

## Key Changes

- **Deferred Trie Data**: The `Chain` type now stores `BTreeMap<BlockNumber, DeferredTrieData>` instead of separate maps for trie updates and hashed state
- **Non-blocking Construction**: `Chain` can be constructed without blocking on expensive trie computations
- **Lazy Access**: Trie data is computed in background tasks and materialized only when needed

## Accessing Trie Data

The trie data is still fully accessible, but access may block if background computation is in progress:

- `chain.trie_data()` / `chain.trie_data_at(num)` - Returns `DeferredTrieData` handles (non-blocking)
- `chain.trie_updates()` / `chain.trie_updates_at(num)` - **Blocks** until computation completes, then returns `Arc<TrieUpdatesSorted>`
- `chain.hashed_state()` / `chain.hashed_state_at(num)` - **Blocks** until computation completes, then returns `Arc<HashedPostStateSorted>`

For best performance, prefer using `trie_data()` and calling `wait_cloned()` explicitly when you need the computed data.

## Crate Structure

A new `std`-only crate `reth-chain` is added which contains:
- `Chain` type
- `DeferredTrieData` and `ComputedTrieData`
- Bincode-compatible serde implementations for WAL persistence

## Backwards Compatibility

WAL files created with older versions (without trie data) will deserialize correctly - missing fields default to empty maps.


crates/evm/chain/Cargo.toml
```diff
@@ -0,0 +1,69 @@
+[package]
+name = "reth-chain"
```
mattsse, line 2: imo this is fine

mattsse (approved):
makes sense

mattsse (approved):

PR #21140, opened 2026-01-16
perf(storage): batch trie updates across blocks in save_blocks

gakonst:
## Summary

Batches trie updates across all blocks in `save_blocks` instead of writing per-block.

## Problem

Per #eng-perf profiling, `write_trie_updates` was taking **~25% of persistence time**. The current implementation calls `write_trie_updates_sorted` once per block, opening/closing cursors N times.

In back-to-back (b2b) scenarios with 75-250 accumulated blocks, this overhead compounds significantly.

## Solution

Accumulate trie updates across blocks using the existing `extend_ref` method, then write them all in a single batch:

```rust
// Accumulate across blocks
let mut accumulated_trie_updates: Option<TrieUpdatesSorted> = None;

for block in blocks {
    // ... other per-block writes ...
    
    match &mut accumulated_trie_updates {
        Some(acc) => acc.extend_ref(&trie_data.trie_updates),
        None => accumulated_trie_updates = Some((*trie_data.trie_updates).clone()),
    }
}

// Single batch write at end
if let Some(trie_updates) = &accumulated_trie_updates {
    self.write_trie_updates_sorted(trie_updates)?;
}
```

## Expected Impact

- **~50% reduction** in `write_trie_updates` time for b2b scenarios
- Reduces cursor open/close overhead from N to 1
- Reduces MDBX transaction overhead

## Testing

- All existing `reth-provider` tests pass

## Related

- Based on analysis from #eng-perf Slack channel
- Complements `georgios/history-indices-from-memory` (26% history indices optimization)
- Supersedes #21106 and #21139

### DIFF
--- a/crates/engine/tree/Cargo.toml
+++ b/crates/engine/tree/Cargo.toml
@@ -115,6 +115,14 @@ harness = false
 name = "state_root_task"
 harness = false
 
+[[bench]]
+name = "heavy_persistence"
+harness = false
+
+[[bench]]
+name = "execution_cache"
+harness = false
+
 [features]
 test-utils = [
     "reth-chain-state/test-utils",
--- a/crates/engine/tree/benches/execution_cache.rs
+++ b/crates/engine/tree/benches/execution_cache.rs
@@ -10,11 +10,11 @@
 
 #![allow(missing_docs)]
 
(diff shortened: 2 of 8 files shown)


mattsse (commented):
this now only has benches left

PR #21143, opened 2026-01-16
perf(chain-state): share computed overlay between LazyOverlay and DeferredTrieData

gakonst:
## Summary

When `LazyOverlay::merge_blocks` computes an overlay (slow path due to anchor mismatch), it now publishes the result to the tip block's cache via `install_anchored_overlay()`. This allows future calls to `DeferredTrieData::sort_and_build_trie_input` to reuse the overlay via the fast path, avoiding duplicate merge work.

## Problem

Previously, `merge_blocks` and `merge_ancestors_into_overlay` could both be called and perform the same merging work sequentially:
1. `LazyOverlay::compute()` would call `merge_blocks()` when anchor didn't match
2. Later, `DeferredTrieData::sort_and_build_trie_input()` would call `merge_ancestors_into_overlay()` doing the same merge again

## Solution

- Add `DeferredTrieData::install_anchored_overlay()` method to update the cached overlay
- Change `LazyOverlay` to cache `Arc<TrieInputSorted>` for cheap sharing (avoid cloning)
- After slow-path computation in `LazyOverlay::compute()`, publish the result to the tip block's cache
- Now `sort_and_build_trie_input` can hit the fast path and reuse the already-computed overlay

## Changes

- `deferred_trie.rs`: Add `install_anchored_overlay()` method
- `lazy_overlay.rs`: Cache `Arc<TrieInputSorted>`, publish to tip after merge
- Added test `merge_publishes_to_tip_block_on_anchor_mismatch`

## Testing

All 48 tests pass in `reth-chain-state`.

### DIFF
--- a/crates/chain-state/src/lazy_overlay.rs
+++ b/crates/chain-state/src/lazy_overlay.rs
@@ -34,11 +34,13 @@ struct LazyOverlayInputs {
 ///
 /// - **Fast path**: If the tip block's cached `anchored_trie_input` is ready and its `anchor_hash`
 ///   matches our expected anchor, we can reuse it directly (O(1)).
-/// - **Slow path**: Otherwise, we merge all ancestor blocks' trie data into a new overlay.
+/// - **Slow path**: Otherwise, we merge all ancestor blocks' trie data into a new overlay. After
+///   computing, the overlay is published to the tip block's cache so that future calls to
+///   [`DeferredTrieData::sort_and_build_trie_input`] can reuse it.
 #[derive(Clone)]
 pub struct LazyOverlay {
-    /// Computed result, cached after first access.
-    inner: Arc<OnceLock<TrieInputSorted>>,
+    /// Computed result, cached after first access. Wrapped in Arc for cheap sharing.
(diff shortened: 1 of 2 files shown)


crates/chain-state/src/lazy_overlay.rs
```diff
@@ -34,11 +34,13 @@ struct LazyOverlayInputs {
...
-/// - **Slow path**: Otherwise, we merge all ancestor blocks' trie data into a new overlay.
+/// - **Slow path**: Otherwise, we merge all ancestor blocks' trie data into a new overlay. After
+///   computing, the overlay is published to the tip block's cache so that future calls to
+///   [`DeferredTrieData::sort_and_build_trie_input`] can reuse it.
 #[derive(Clone)]
 pub struct LazyOverlay {
-    /// Computed result, cached after first access.
-    inner: Arc<OnceLock<TrieInputSorted>>,
+    /// Computed result, cached after first access. Wrapped in Arc for cheap sharing.
+    inner: Arc<OnceLock<Arc<TrieInputSorted>>>,
```
mattsse, line 43: this doesnt do anything because the internals that we clone are already arc

PR #21150, opened 2026-01-16
fix(engine): request head block download when not buffered after backfill

gakonst:
## Problem

When backfill completes, the engine only triggers further sync if:
1. The finalized block is already buffered (for checking if more backfill is needed), OR
2. The head block's lowest ancestor is buffered (for downloading remaining blocks)

If neither condition is met, the function returns early without triggering any download. This causes the node to stall after backfill completes, waiting indefinitely for a new FCU.

## Solution

When the head block (and its ancestors) is not buffered after backfill, request a download for the FCU head block. This kicks off another sync cycle that will eventually connect the chain.

## Related

Fixes sync stall issue where CL and EL were "staring at walls" for hours after backfill.

Ref: [TOOLS-56](https://linear.app/tempoxyz/issue/TOOLS-56/create-nightly-node-crash-and-restore-test-suite)

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1676,6 +1676,18 @@ where
                 )));
                 return Ok(());
             }
+        } else {
+            // We don't have the head block or any of its ancestors buffered. Request
+            // a download for the head block which will then trigger further sync.
+            debug!(
+                target: "engine::tree",
+                head_hash = %sync_target_state.head_block_hash,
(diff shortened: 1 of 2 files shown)


joshieDo (approved):

mattsse (changes requested):
we need to update the failing test_engine_tree_live_sync_transition_required_blocks_requested test with an additinal event assert before we sent the mocked downloaded block range

mattsse (approved):

PR #21152, opened 2026-01-17
refactor: use ExecutionOutcome::single instead of tuple From

mattsse:
Replace obfuscating tuple `From` conversions with explicit `ExecutionOutcome::single()` calls in `blocks_to_chain` for better readability.

The tuple `From<(BlockExecutionOutput<T>, BlockNumber)>` implementation just delegates to `Self::single(block_number, output)`, so calling the method directly is clearer.


gakonst (approved):
LGTM! Clearer than the tuple From implementation.

DaniPopes (approved):

PR #21153, opened 2026-01-17
refactor(trie): simplify LazyTrieData with SortedTrieData container

mattsse:
## Summary

Simplifies the `LazyTrieData` abstraction introduced in #21139.

## Changes

- Add `SortedTrieData` container bundling `HashedPostStateSorted` + `TrieUpdatesSorted`
- Simplify `LazyTrieData` to use single `Arc<OnceLock<SortedTrieData>>` instead of two separate `OnceLock`s
- Remove `TrieDataCompute` trait - use simple `Arc<dyn Fn() -> SortedTrieData>` instead
- Remove unnecessary `Default` impl for `LazyTrieData`

The previous implementation had two separate `OnceLock` fields and a trait that was only implemented by one type. This consolidates everything into a simpler design while maintaining the same lazy evaluation semantics.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -958,7 +958,7 @@ impl<N: NodePrimitives<SignedTx: SignedTransaction>> NewCanonicalChain<N> {
                         first.block_number(),
                         first.execution_outcome().clone(),
                     ),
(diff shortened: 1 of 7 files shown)


mattsse:
Closing to redo as part of a broader revert of #21137

PR #21161, opened 2026-01-17
feat(download): resumable snapshot downloads with auto-retry

gakonst:
## Summary

Snapshot and ERA downloads now resume from where they left off on failure and automatically retry.

## Motivation

Downloading 100GB+ snapshots over flaky connections sucks. If your connection drops at 90GB you shouldn't have to start over.

## Changes

**crates/cli/commands/src/download.rs**
- Download to .part file, rename on completion
- Send `Range` header to resume from partial
- Retry up to 10x with 5s backoff
- 30s timeout to catch stalled connections

**crates/era-downloader/src/client.rs**
- Add `get_range()` to `HttpClient` trait
- Same .part + retry pattern
- Preserve checksum state across resume for ERA1

## Testing

cargo check, needs manual testing with network interruption

### DIFF
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -331,7 +331,8 @@ fn extract_from_file(path: &Path, format: CompressionFormat, target_dir: &Path) 
 const MAX_DOWNLOAD_RETRIES: u32 = 10;
 const RETRY_BACKOFF_SECS: u64 = 5;
 
-/// Wrapper that tracks download progress while copying data.
+/// Wrapper that tracks download progress while writing data.
+/// Used with [`io::copy`] to display progress during downloads.
 struct ProgressWriter<W> {
     inner: W,
     progress: DownloadProgress,
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -2,14 +2,15 @@ use crate::common::EnvironmentArgs;
 use clap::Parser;
 use eyre::Result;
 use lz4::Decoder;
-use reqwest::Client;
+use reqwest::{blocking::Client as BlockingClient, header::RANGE, Client, StatusCode};
 use reth_chainspec::{EthChainSpec, EthereumHardforks};
 use reth_cli::chainspec::ChainSpecParser;
 use reth_fs_util as fs;
 use std::{
     borrow::Cow,
-    io::{self, Read, Write},
-    path::Path,
+    fs::OpenOptions,
+    io::{self, BufWriter, Read, Write},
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/download.rs
```diff
@@ -327,18 +328,153 @@ fn extract_from_file(path: &Path, format: CompressionFormat, target_dir: &Path)
...
-        eyre::eyre!(
-            "Server did not provide Content-Length header. This is required for snapshot downloads"
-        )
-    })?;
+    info!(target: "reth::cli", "Extracting snapshot...");
+    let file = File::open(&downloaded_path)?;
+    extract_archive(file, total_size, format, target_dir)?;
 
-    extract_archive(response, total_size, format, target_dir)
+    std::fs::remove_file(&downloaded_path)?;
```
mattsse: lets use the fs_common helpers for all fs ops so that we get nicer error message when things go wrong

crates/cli/commands/src/download.rs
```diff
@@ -327,18 +328,153 @@ fn extract_from_file(path: &Path, format: CompressionFormat, target_dir: &Path)
...
+                    let _ = progress.update(n as u64);
+                }
+                Err(e) => {
+                    download_error = Some(e.into());
+                    break;
+                }
+            }
+        }
+
+        let _ = file.flush();
```
mattsse: isnt there a a nicer way to do this, are we primarily doing this for progress updates?

crates/cli/commands/src/download.rs
```diff
@@ -327,18 +328,153 @@ fn extract_from_file(path: &Path, format: CompressionFormat, target_dir: &Path)
...
+            OpenOptions::new().append(true).open(&part_path)?
+        } else {
+            File::create(&part_path)?
+        };
+
+        let start_offset = if is_partial { existing_size } else { 0 };
+        let mut progress = DownloadProgress::new(current_total);
+        progress.downloaded = start_offset;
+
+        let mut buf = [0u8; 64 * 1024];
```
mattsse: this balloons the stack, this should be an actual heap buf

but ideally this uses like an io::copy and or a bufwriter over the file

mattsse (changes requested):

crates/cli/commands/src/download.rs
gakonst: Fixed - now using `fs::metadata`, `fs::rename`, `fs::create_file`, `fs::open`, `fs::remove_file`, and `FsPathError::open` for better error messages.

mattsse (approved):

PR #21162, opened 2026-01-17
feat(execution-types): add receipts_iter helper

mattsse:
Adds `receipts_iter()` method to `ExecutionOutcome` for ergonomic iteration over receipt slices.

```rust
/// Returns an iterator over receipt slices, one per block.
pub fn receipts_iter(&self) -> impl Iterator<Item = &[T]> + '_ {
    self.receipts.iter().map(|v| v.as_slice())
}
```

This is a more ergonomic alternative to `receipts()` that yields slices instead of requiring indexing into a nested `Vec<Vec<T>>`.


mattsse (commented):
.

crates/evm/execution-types/src/execution_outcome.rs
mattsse, line 1: let's only keep this and remove the rest

mattsse (commented):
.

PR #21165, opened 2026-01-17
feat(stages): add RocksDB support for IndexAccountHistoryStage

### LINKED ISSUE #20593: Wire up RocksDB initialization to Stages
### Describe the feature

Pass RocksDB to stages:
   - `TransactionLookupStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - `IndexStorageHistoryStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - `IndexAccountHistoryStage` needs access to `RocksDBProvider` (or `RocksTx`)
   - Options:
     - Add `RocksDBProvider` to stage structs
     - Pass via provider trait
     - Use a shared Arc reference

split from https://github.com/paradigmxyz/reth/issues/20392#issuecomment-3669321629

blocked by https://github.com/paradigmxyz/reth/issues/20392

### Additional context

_No response_

yongkangc:
Add RocksDB support to the account history index stage using the `EitherWriter` abstraction

**Changes**
- Add `RocksDBProviderFactory` trait bound to Stage impl
- Use `EitherWriter::new_accounts_history()` to route writes to MDBX or RocksDB
- Add inline helper functions for loading/flushing shards with proper `u64::MAX` handling
- Add RocksDB-specific tests (execute, incremental sync, unwind checkpoint)

---
## PR Stack

```
main
  ↑
#21165 (rocksdb-index-account-history) ◀ you are here
  ↑
#21175 (rocksdb-index-storage-history)
```
- https://github.com/paradigmxyz/reth/pull/21165 < 
- https://github.com/paradigmxyz/reth/pull/21175

Closes https://github.com/paradigmxyz/reth/issues/20593


crates/stages/stages/src/stages/utils.rs
```diff
@@ -209,13 +209,9 @@ where
             info!(target: "sync::stages::index_history", progress = %format!("{:.2}%", (index as f64 / total_entries as f64) * 100.0), "Writing indices");
         }
 
-        // AccountsHistory: `Address`.
-        // StorageHistory: `Address.StorageKey`.
```
yongkangc, line 212: dont remove this comment

crates/stages/stages/src/stages/utils.rs
```diff
@@ -209,13 +209,9 @@ where
...
         let partial_key = get_partial(sharded_key);
 
         if current_partial != partial_key {
-            // We have reached the end of this subset of keys so
-            // we need to flush its last indice shard.
```
yongkangc, line 217: dont remove this comment

crates/stages/stages/src/stages/utils.rs
```diff
@@ -228,8 +224,6 @@ where
             current_partial = partial_key;
             current_list.clear();
 
-            // If it's not the first sync, there might an existing shard already, so we need to
-            // merge it with the one coming from the collector
```
yongkangc, line 231: dont remove this comment

crates/stages/stages/src/stages/utils.rs
```diff
@@ -249,7 +243,6 @@ where
         )?;
     }
 
-    // There will be one remaining shard that needs to be flushed to DB.
```
yongkangc, line 251: dont remove this comment

crates/storage/provider/src/either_writer.rs
```diff
@@ -518,8 +518,22 @@ impl<'a, CURSOR, N: NodePrimitives> EitherWriter<'a, CURSOR, N>
 where
     CURSOR: DbCursorRW<tables::AccountsHistory> + DbCursorRO<tables::AccountsHistory>,
 {
-    /// Puts an account history entry.
-    pub fn put_account_history(
```
yongkangc, line 522: - Renamed `put_account_history` to `upsert_account_history` to make the semantics clearer (it was already calling `cursor.upsert` internally). Also added a separate `append_account_history` for the first-sync case which is more efficient when we know keys are monotonically increasing.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -430,6 +430,51 @@ impl RocksDBProvider {
...
+                code: -1,
+            }))
+        })?;
+
+        Ok(())
+    }
+
+    /// Returns the smallest key strictly greater than `key` under bytewise comparison.
+    #[inline]
+    fn exclusive_end_after(key: &[u8]) -> Vec<u8> {
```
yongkangc: this seems risky, do we really need it?

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -430,6 +430,51 @@ impl RocksDBProvider {
...
+        let last_key = iter.key().ok_or_else(|| {
+            ProviderError::Database(DatabaseError::Read(DatabaseErrorInfo {
+                message: "invalid iterator key".into(),
+                code: -1,
+            }))
+        })?;
+
+        // Compute exclusive end bound: `last_key || 0x00` is the smallest key > last_key
+        // under bytewise ordering (avoids 0xFF carry issues of increment-last-byte approach)
+        let end_key = Self::exclusive_end_after(last_key);
```
yongkangc: This is necessary for first-sync performance: on first sync we clear the table and rebuild from scratch (same pattern as MDBX's `tx.clear::<T>()`). The implementation is defensive - returns early if table is empty, and uses the safe `exclusive_end_after` helper to compute the range bound.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - restored all comments in the latest commit.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - restored.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - restored.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - restored.

crates/storage/provider/src/either_writer.rs
```diff
@@ -1266,8 +1295,8 @@ mod rocksdb_tests {
         for (highest_block, blocks) in shards {
             let key = ShardedKey::new(address, *highest_block);
             let value = IntegerList::new(blocks.clone()).unwrap();
-            mdbx_writer.put_account_history(key.clone(), &value).unwrap();
-            rocks_writer.put_account_history(key, &value).unwrap();
+            mdbx_writer.upsert_account_history(key.clone(), &value).unwrap();
+            rocks_writer.upsert_account_history(key, &value).unwrap();
```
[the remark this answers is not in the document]
yongkangc: Yes, this is test code that verifies both MDBX and RocksDB backends produce identical results. `upsert` is used here to ensure we can both insert new entries and update existing ones during the test, which exercises the incremental sync path where shards may be merged.

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,153 @@ impl LoadMode {
...
+
+    Ok(())
+}
+
+/// Flushes complete shards for account history, keeping the trailing partial shard buffered.
+///
+/// Only flushes when we have more than one shard's worth of data, keeping the last
+/// (possibly partial) shard for continued accumulation. This avoids writing a shard
+/// that may need to be updated when more indices arrive.
+fn flush_account_history_shards_partial<N, CURSOR>(
```
yongkangc, line 405: why do we need partial flush and full flush functions?

- `flush_account_history_shards_partial`: Called during iteration when accumulating indices. Only flushes complete shards, keeping the last (possibly partial) shard in memory for continued accumulation.

- `flush_account_history_shards`: Called at key boundaries (when address changes) and at the end. Flushes all remaining shards, using `u64::MAX` as the highest block number for the final shard (the invariant that allows `seek_exact(address, u64::MAX)` to find the last shard during incremental sync).

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc: This is a standard technique for computing an exclusive upper bound in RocksDB's `delete_range_cf` which uses `[start, end)` semantics. Appending `0x00` to the last key gives us the minimal key strictly greater than `last_key` under bytewise comparison (safer than incrementing the last byte which has 0xFF overflow edge cases). Only used in `clear<T>()` for first-sync table clearing.

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,153 @@ impl LoadMode {
...
 }
 
+/// Loads account history indices into the database via `EitherWriter`.
+///
+/// Similar to [`load_history_indices`] but works with [`EitherWriter`] to support
+/// both MDBX and `RocksDB` backends.
+///
+/// Uses `Option<Address>` instead of `Address::default()` as the sentinel to avoid
+/// incorrectly treating `Address::ZERO` as "no previous address".
+pub(crate) fn load_account_history_via_writer<N, CURSOR>(
```
yongkangc: Added this function specialized for `EitherWriter`. The original `load_history_indices` is generic over the table type and works with raw cursors, while this one works specifically with `EitherWriter` which abstracts MDBX/RocksDB backends. The key difference is using `Option<Address>` as the sentinel (to handle `Address::ZERO` correctly) and calling `EitherWriter` methods directly instead of through cursor trait bounds.

crates/stages/stages/src/stages/utils.rs
```diff
@@ -125,12 +126,12 @@ where
...
 
     let mut insert_fn = |address: Address, indices: Vec<u64>| {
-        let last = indices.last().expect("qed");
+        let last = indices.last().expect("indices is non-empty");
         collector.insert(
             ShardedKey::new(address, *last),
             BlockNumberList::new_pre_sorted(indices.into_iter()),
         )?;
-        Ok::<(), StageError>(())
+        Ok(())
```
yongkangc, line 133: can we keep

Ok::<(), StageError>(())

crates/stages/stages/src/stages/utils.rs
```diff
@@ -109,7 +110,7 @@ where
     for (address, indices) in cache {
         insert_fn(address, indices)?
     }
-    Ok::<(), StageError>(())
+    Ok(())
```
yongkangc, line 112: > can we keep
> 
> Ok::<(), StageError>(())

crates/stages/stages/src/stages/utils.rs
```diff
@@ -57,15 +58,15 @@ where
...
         for (key, indices) in cache {
-            let last = indices.last().expect("qed");
+            let last = indices.last().expect("indices is non-empty");
             collector.insert(
                 sharded_key_factory(*key, *last),
                 BlockNumberList::new_pre_sorted(indices.iter().copied()),
             )?;
         }
-        Ok::<(), StageError>(())
+        Ok(())
```
yongkangc: can we keep

Ok::<(), StageError>(())?

or its cleaner to not keep?

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -125,14 +133,22 @@ where
...
+        #[cfg(not(all(unix, feature = "rocksdb")))]
+        let rocksdb_batch = ();
+
+        let mut writer = EitherWriter::new_accounts_history(provider, rocksdb_batch)?;
+
+        load_account_history_via_writer(collector, first_sync, &mut writer)?;
+
+        #[cfg(all(unix, feature = "rocksdb"))]
+        if let Some(batch) = writer.into_raw_rocksdb_batch() {
+            provider.set_pending_rocksdb_batch(batch);
```
yongkangc: why do we want to set pending batch here?

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -125,14 +133,22 @@ where
...
+        #[cfg(all(unix, feature = "rocksdb"))]
+        let rocksdb = provider.rocksdb_provider();
+        #[cfg(all(unix, feature = "rocksdb"))]
+        let rocksdb_batch = rocksdb.batch();
+        #[cfg(not(all(unix, feature = "rocksdb")))]
+        let rocksdb_batch = ();
+
+        let mut writer = EitherWriter::new_accounts_history(provider, rocksdb_batch)?;
+
+        load_account_history_via_writer(collector, first_sync, &mut writer)?;
```
yongkangc: not really a fan of using `load_account_history_via_writer` but i guess that would work for now?

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,153 @@ impl LoadMode {
...
+    };
+
+    if shards_to_flush == 0 {
+        return Ok(());
+    }
+
+    let flush_len = shards_to_flush * NUM_OF_INDICES_IN_SHARD;
+    let remainder = list.split_off(flush_len);
+
+    for chunk in list.chunks(NUM_OF_INDICES_IN_SHARD) {
```
yongkangc, line 440: can we add a clear concise comment here to explain what this does?

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,153 @@ impl LoadMode {
...
+    CURSOR: DbCursorRW<reth_db_api::tables::AccountsHistory>
+        + DbCursorRO<reth_db_api::tables::AccountsHistory>,
+{
+    let mut current_address: Option<Address> = None;
+    let mut current_list = Vec::<u64>::new();
+
+    let total_entries = collector.len();
+    let interval = (total_entries / 10).max(1);
+
+    for (index, element) in collector.iter()?.enumerate() {
```
yongkangc, line 353: can we add clear concise comment here to show what this whole chunk of code is doing. its hard to follow

crates/stages/stages/src/stages/utils.rs
yongkangc: add comments throughout this function. else its hard to follow

yongkangc: Done - added inline comments throughout all three functions (load_account_history_via_writer, flush_account_history_shards_partial, flush_account_history_shards) in commit 3213c7216.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - added comments in commit 3213c7216.

crates/stages/stages/src/stages/utils.rs
yongkangc: Done - added comments explaining the u64::MAX invariant and shard flushing logic in commit 3213c7216.

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc: This seems to be a standard technique for computing an exclusive upper bound in RocksDB's `delete_range_cf` which uses `[start, end)` semantics. Appending `0x00` to the last key gives us the minimal key strictly greater than `last_key` under bytewise comparison (safer than incrementing the last byte which has 0xFF overflow edge cases). Only used in `clear<T>()` for first-sync table clearing.

crates/stages/stages/src/stages/utils.rs
yongkangc: Cleaner to not keep. The return type is already specified via the `F: FnMut(...) -> Result<(), StageError>` bound, so Rust infers the type automatically. The explicit annotation `Ok::<(), StageError>(())` is only needed when the compiler can't infer the error type from context.

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,187 @@ impl LoadMode {
...
+    append_only: bool,
+    writer: &mut EitherWriter<'_, CURSOR, N>,
+) -> Result<(), StageError>
+where
+    N: NodePrimitives,
+    CURSOR: DbCursorRW<reth_db_api::tables::AccountsHistory>
+        + DbCursorRO<reth_db_api::tables::AccountsHistory>,
+{
+    // Track current address being processed. Using Option<Address> instead of Address::default()
+    // to correctly handle Address::ZERO (which would otherwise be treated as "no address").
```
yongkangc:
```suggestion
```

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,187 @@ impl LoadMode {
...
+    let mut current_address: Option<Address> = None;
+    // Accumulator for block numbers where the current address changed.
+    let mut current_list = Vec::<u64>::new();
+
+    // Progress reporting setup.
```
yongkangc:
```suggestion
```

crates/stages/stages/src/stages/utils.rs
```diff
@@ -321,6 +322,187 @@ impl LoadMode {
...
+    N: NodePrimitives,
+    CURSOR: DbCursorRW<reth_db_api::tables::AccountsHistory>
+        + DbCursorRO<reth_db_api::tables::AccountsHistory>,
+{
+    // Nothing to flush if we haven't filled a complete shard yet.
+    if list.len() <= NUM_OF_INDICES_IN_SHARD {
+        return Ok(());
+    }
+
+    // Calculate how many complete shards we have.
```
yongkangc:
```suggestion
```

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -102,14 +102,22 @@ where
...
+        // Check if we're using RocksDB for account history
+        let use_rocksdb = provider.cached_storage_settings().account_history_in_rocksdb;
+
         // On first sync we might have history coming from genesis. We clear the table since it's
         // faster to rebuild from scratch.
         if first_sync {
-            provider.tx_ref().clear::<tables::AccountsHistory>()?;
+            if use_rocksdb {
+                #[cfg(all(unix, feature = "rocksdb"))]
+                provider.rocksdb_provider().clear::<tables::AccountsHistory>()?;
```
fgimenez, line 115: this `clear` call executes immediately, operates directly on the db, while mdbx one operates on the tx object https://github.com/paradigmxyz/reth/blob/main/crates/storage/db/src/implementation/mdbx/tx.rs#L429 aiui if the stage crashes after RocksDB clear but before commit the table and the checkpoint would have inconsistent state?

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -102,14 +102,22 @@ where
...
 
+        // Check if we're using RocksDB for account history
+        let use_rocksdb = provider.cached_storage_settings().account_history_in_rocksdb;
+
         // On first sync we might have history coming from genesis. We clear the table since it's
         // faster to rebuild from scratch.
         if first_sync {
-            provider.tx_ref().clear::<tables::AccountsHistory>()?;
+            if use_rocksdb {
+                #[cfg(all(unix, feature = "rocksdb"))]
```
fgimenez, line 114: here, and in other places of the file, if a user configures `account_history_in_rocks_db = true` and builds without `--features rocksdb` the node will be writing only to mdbx siliently, wihtout any feedback, even tough the user configured to use rocksdb, should we error in this case or at least log a message?

yongkangc: @fgimenez  good point, in the future we might remove the flag, but i think for now we can error out - so that nothing bad happens

crates/stages/stages/src/stages/index_account_history.rs
yongkangc: hmm yea, good note, i missed this!  i guess the best way would be to  have `clear()` part of the WriteBatch so it only takes effect when provider.commit() is called, same as the MDBX path.

crates/stages/stages/src/stages/index_account_history.rs
yongkangc: addressed in [6c6d81f](https://github.com/paradigmxyz/reth/pull/21165/commits/6c6d81f98b22ddad3125c518454eb91a33857e7d)

crates/stages/stages/src/stages/index_account_history.rs
yongkangc: good note, although would this be a concern for first sync? 

since its the first sync, the checkpoint only advances after provider.commit(). So if we crash before commit, the checkpoint stays at 0, and we'll rebuild from changesets

fgimenez (approved):

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -143,6 +175,16 @@ where
...
     ) -> Result<UnwindOutput, StageError> {
+        #[cfg(not(all(unix, feature = "rocksdb")))]
+        if provider.cached_storage_settings().account_history_in_rocksdb {
+            return Err(StageError::Fatal(Box::new(ProviderError::other(
+                std::io::Error::new(
+                    std::io::ErrorKind::Unsupported,
+                    "account_history_in_rocksdb is enabled but this binary was built without the 'rocksdb' feature",
+                ),
+            ))));
+        }
```
joshieDo: i wouldnt bother we this kind i think

yongkangc: i see, since its only internal use for now - so we dont really need that 

i will remove for this then

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2970,25 +2970,58 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> HistoryWriter for DatabaseProvi
...
+                        Some(keep_to) => batch.unwind_account_history_to(address, keep_to)?,
+                        None => batch.clear_account_history(address)?,
+                    }
+                }
+                self.pending_rocksdb_batches.lock().push(batch.into_inner());
+            }
+
+            #[cfg(not(all(unix, feature = "rocksdb")))]
+            return Err(ProviderError::UnsupportedProvider);
+        }
```
joshieDo: can we move this to rocksdb provider

yongkangc: makes sense, yup

yongkangc: moved over to `unwind_account_history_indices`

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2970,25 +2970,40 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypes> HistoryWriter for DatabaseProvi
...
-                rem_index,
-                |sharded_key| sharded_key.key == address,
-            )?;
+        let use_rocksdb = self.cached_storage_settings().account_history_in_rocksdb;
 
-            // Check the last returned partial shard.
-            // If it's not empty, the shard needs to be reinserted.
-            if !partial_shard.is_empty() {
-                cursor.insert(
+        if use_rocksdb {
```
```suggestion
        if self.cached_storage_settings().account_history_in_rocksdb {
```
...

else { 

}

crates/stages/stages/src/stages/utils.rs
```diff
@@ -320,6 +321,183 @@ impl LoadMode {
...
+/// both MDBX and `RocksDB` backends.
+///
+/// ## Process
+/// Iterates over elements, grouping indices by their address. It flushes indices to disk
+/// when reaching a shard's max length (`NUM_OF_INDICES_IN_SHARD`) or when the address changes,
+/// ensuring the last previous address shard is stored.
+///
+/// Uses `Option<Address>` instead of `Address::default()` as the sentinel to avoid
+/// incorrectly treating `Address::ZERO` as "no previous address".
+pub(crate) fn load_account_history_via_writer<N, CURSOR>(
```
joshieDo:
```suggestion
pub(crate) fn load_account_history<N, CURSOR>(
```

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -102,17 +102,27 @@ where
...
         let first_sync = input.checkpoint().block_number == 0;
 
+        let use_rocksdb = provider.cached_storage_settings().account_history_in_rocksdb;
+
         // On first sync we might have history coming from genesis. We clear the table since it's
         // faster to rebuild from scratch.
+        //
+        // Note: RocksDB clear() executes immediately (not deferred to commit like MDBX), but this
+        // is safe for first_sync because if we crash before commit, the checkpoint stays at 0 and
+        // we'll just clear and rebuild again on restart. The source data (changesets) is intact.
```
joshieDo: can we move it inside the if use_rocksdb { bracket

crates/storage/provider/src/providers/database/provider.rs
yongkangc: @joshieDo i am guessing we dont want to have use_rocksdb in the log? to me having that in the log makes sense just for a sanity check

crates/stages/stages/src/stages/index_account_history.rs
yongkangc: yup makes sense, thats cleaner

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -430,6 +430,51 @@ impl RocksDBProvider {
...
+    /// Clears all entries from the specified table.
+    ///
+    /// `RocksDB`'s `delete_range` uses an exclusive end bound `[start, end)`. To include the
+    /// last key, we compute `last_key || 0x00` which is the minimal key strictly greater
+    /// than `last_key` under bytewise comparison.
+    pub fn clear<T: Table>(&self) -> ProviderResult<()> {
+        let cf = self.get_cf_handle::<T>()?;
+
+        let mut iter = self.0.db.raw_iterator_cf(cf);
+        iter.seek_to_last();
```
joshieDo: unsure if this will be expensive or not?

since keys are fixed size, we should be able to just pass 0x00..0..0xff..ff no?

crates/storage/provider/src/providers/database/provider.rs
joshieDo: we're not logging here i think? and we could just log `rocksdb_provider.unwind_account_history_indices ` instead

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc: yeap good catch, this makes sense instead of seeking to last

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -719,6 +821,16 @@ impl<'a> RocksDBBatch<'a> {
         self.inner
     }
 
+    /// Gets a value from the database.
+    ///
+    /// **Important constraint:** This reads only committed state, not pending writes in this
+    /// batch or other pending batches in `pending_rocksdb_batches`. This is acceptable for the
+    /// current use case (merging with previously persisted shards at the start of processing
+    /// an address), but callers must not rely on read-your-writes semantics within a batch.
```
```suggestion
    /// **Important constraint:** This reads only committed state, not pending writes in this
    /// batch or other pending batches in `pending_rocksdb_batches`.
```

crates/stages/stages/src/stages/index_account_history.rs
```diff
@@ -101,18 +101,27 @@ where
...
+                provider.tx_ref().clear::<tables::AccountsHistory>()?;
+            }
             range = 0..=*input.next_block_range().end();
         }
 
-        info!(target: "sync::stages::index_account_history::exec", ?first_sync, "Collecting indices");
+        info!(target: "sync::stages::index_account_history::exec", ?first_sync, ?use_rocksdb, "Collecting indices");
 
         let collector = if provider.cached_storage_settings().account_changesets_in_static_files {
-            // Use the provider-based collection that can read from static files.
```
joshieDo, line 115: shouldnt be deleted

joshieDo (approved):

PR #21168, opened 2026-01-18
perf(storage): Avoid rocksdb txs for legacy mdbx only nodes

### LINKED ISSUE #21144: Avoid creating RocksDB transactions for legacy MDBX-only nodes
## Summary

`with_rocksdb_tx` always creates a RocksDB transaction via `rocksdb_provider.tx()` even for legacy MDBX-only nodes where RocksDB isn't used for history indices. This adds unnecessary overhead.

## Problem

When `EitherReader::new_*` constructors are called, they receive a `RocksTxRefArg` and check `storage_settings` to decide whether to use RocksDB. On legacy nodes, these flags are `false`, so the tx is created but never used.

## Proposed Solution

Make `RocksTxRefArg` an `Option<&RocksTx>` and check `storage_settings.any_history_in_rocksdb()` in `with_rocksdb_tx` **before** creating the transaction:

```rust
fn with_rocksdb_tx<F, R>(&self, f: F) -> ProviderResult<R> {
    if !self.cached_storage_settings().any_history_in_rocksdb() {
        return f(None);  // No tx created
    }
    let tx = self.rocksdb_provider().tx();
    f(Some(&tx))
}
```

## Related

Part of #20384

i-m-aditya:
Resolves #21144

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -613,15 +613,15 @@ impl<'a> EitherReader<'a, (), ()> {
     /// Creates a new [`EitherReader`] for storages history based on storage settings.
     pub fn new_storages_history<P>(
         provider: &P,
-        _rocksdb_tx: RocksTxRefArg<'a>,
+        _rocksdb_tx: Option<RocksTxRefArg<'a>>,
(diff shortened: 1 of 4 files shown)


yongkangc:
sorry we will be taking this inhouse

PR #21169, opened 2026-01-18
docs: add missing Engine API documentation for v1.10.0 RPC methods

letmehateu:
Adds documentation for Engine API methods introduced in v1.10.0 that were missing from the docs.


mattsse (commented):
we can keep as is

PR #21173, opened 2026-01-18
fix: remove redundant usize cast in history pruning

forkfury:
Removed redundant as usize cast when comparing BlockNumberList length with Vec length. Rust handles type coercion between u64 and usize in equality comparisons on 64-bit platforms.


mattsse (commented):
we can keep as is

PR #21174, opened 2026-01-18
refactor: remove EthPrimitives re-export from chain-state

avorylli:
Remove the temporary EthPrimitives re-export from reth-chain-state crate.

The generic data primitives integration is complete, so the re-export is no longer needed. 

Update all usages to import directly from reth-ethereum-primitives.


mattsse (commented):
we can keep this

PR #21180, opened 2026-01-19
feat(rpc): add transaction hash caching to EthStateCache

figtracer:
## **Summary**

Adds O(1) transaction hash lookup caching to the RPC layer's `EthStateCache`. This speeds up `eth_getTransactionByHash` and `eth_getTransactionReceipt` for recently cached blocks by avoiding database lookups.

## Changes

- Introduces `CachedTransaction<B, R>` helper struct containing:
  - `block: Arc<RecoveredBlock<B>>` - the cached block (shared with block cache)
  - `tx_index: usize` - transaction index within the block
  - `receipts: Option<Arc<Vec<R>>>` - optional receipts (shared with receipts cache)

- Adds `tx_hash_index: LruMap<TxHash, Rc(B256, usize)>` to `EthStateCacheService` that indexes transaction hashes to their block hash and index

- New `get_transaction_by_hash()` method returns the full `CachedTransaction` with block and optional receipts in a single lookup

- Transactions are indexed when new canonical blocks arrive via `CacheNewCanonicalChain` and removed on reorgs via `RemoveReorgedChain`

- Adds `--rpc-cache.max-cached-tx-hashes` CLI option (default: 30,000) to configure the transaction hash cache size


figtracer:
doesn't show results in practice

mattsse (commented):
@figtracer I like this idea, but perhaps we can solve this on the rpc level, like monitor new blocks and track the index there, see ethrpccache

then we could ptimistaically use the cached block already

figtracer:
> @figtracer I like this idea, but perhaps we can solve this on the rpc level, like monitor new blocks and track the index there, see ethrpccache
> 
> then we could ptimistaically use the cached block already

thanks, will reopen and expand into this

crates/rpc/rpc-eth-types/src/cache/mod.rs
```diff
@@ -255,6 +265,19 @@ impl<N: NodePrimitives> EthStateCache<N> {
...
     }
+
+    /// Looks up a transaction by its hash in the cache index.
+    ///
+    /// Returns the block hash and transaction index if the transaction is in a cached block.
+    ///
+    /// Returns `None` if:
+    /// - The transaction is not in the cache index
+    /// - The cache service is unavailable
+    pub async fn get_transaction_by_hash(&self, tx_hash: TxHash) -> Option<(B256, usize)> {
```
mattsse: should this not return the block as well if we already have it cached?

crates/rpc/rpc-eth-types/src/cache/mod.rs
```diff
@@ -84,6 +87,12 @@ impl<N: NodePrimitives> EthStateCache<N> {
         Provider: BlockReader<Block = N::Block, Receipt = N::Receipt>,
     {
         let (to_service, rx) = unbounded_channel();
+
+        // Size the transaction hash index to accommodate all transactions in cached blocks.
+        // Assuming ~200 transactions per block on average.
+        const TXS_PER_BLOCK: u32 = 300;
+        let max_txs = max_blocks.saturating_mul(TXS_PER_BLOCK);
```
mattsse: let's also add this to the cache config?

mattsse (changes requested):

crates/rpc/rpc-eth-types/src/cache/mod.rs
figtracer: on it

crates/rpc/rpc-eth-types/src/cache/config.rs
```diff
@@ -7,6 +7,9 @@ use reth_rpc_server_types::constants::cache::{
     DEFAULT_RECEIPT_CACHE_MAX_LEN,
 };
 
+/// Default number of transactions per block for sizing the transaction lookup cache.
+pub const DEFAULT_MAX_TXS_PER_BLOCK: u32 = 200;
```
figtracer: should this also be in reth_rpc_server_types?|

crates/node/core/src/args/rpc_server.rs
```diff
@@ -1025,6 +1025,7 @@ mod tests {
                 max_receipts: 2000,
                 max_headers: 1000,
                 max_concurrent_db_requests: 512,
+                max_cached_tx_hashes: 20_000,
```
mattsse: I think we can bump this to 30k by default which should maybe be ~150MB or so

crates/rpc/rpc-eth-types/src/cache/mod.rs
```diff
@@ -24,6 +25,28 @@ use tokio::sync::{
...
+/// Cached data for a transaction lookup.
+#[derive(Debug, Clone)]
+pub struct CachedTransaction<B: Block, R> {
+    /// The block containing this transaction.
+    pub block: Arc<RecoveredBlock<B>>,
+    /// Index of the transaction within the block.
+    pub tx_index: usize,
+    /// Receipts for the block, if available.
+    pub receipts: Option<Arc<Vec<R>>>,
+}
```
mattsse: I'd be nice if this has a helper fn for creating the receipt directly

which could use:

https://github.com/paradigmxyz/reth/blob/4d0e06c338faa3e00f0b6308ca566aa0429abc45/crates/rpc/rpc-eth-types/src/block.rs#L75-L86

so we can move this type there as well I believe

crates/rpc/rpc-eth-types/src/cache/mod.rs
```diff
@@ -317,13 +361,46 @@ pub(crate) struct EthStateCacheService<
...
     rate_limiter: Arc<Semaphore>,
+    /// LRU index mapping transaction hashes to their block hash and index within the block.
+    tx_hash_index: LruMap<TxHash, (Rc<B256>, usize), ByLength>,
+}
+
+// SAFETY: The `Rc<B256>` in `tx_hash_index` is only ever accessed from within the single
+// `EthStateCacheService` task. It is created in `index_block_transactions`, accessed in the
+// `GetTransactionByHash` handler, and removed in `remove_block_transactions` - all within
+// the same task's poll loop.
+unsafe impl<Provider, Tasks, LimitBlocks, LimitReceipts, LimitHeaders> Send
```
mattsse: instead of this, let's do a warpper helper around rc<b256>

mattsse (changes requested):
this makes a lot of sense actually and will be more important once we no longer have the index in mdbx directly


left some suggestions

crates/node/core/src/args/rpc_server.rs
figtracer: how do we get this number?

mattsse (approved):
very nice

PR #21186, opened 2026-01-19
perf(engine): add single-entry cache for canonical header lookups

gakonst:
## Summary

Adds a single-entry `RwLock` cache to avoid repeated database I/O when looking up the canonical tip header on the FCU hot path.

## Problem

During forkchoice updates with payload attributes (validators building blocks), `sealed_header_by_hash` is called for the canonical tip. When the header has been persisted and removed from in-memory `tree_state`, it falls back to a database lookup.

## Solution

Add a single-entry cache keyed by block hash:
- Check in-memory tree state first (fastest path)
- Check single-entry cache second (avoids DB on repeated FCU)  
- DB I/O happens outside any lock to prevent contention
- Cache invalidation is implicit: hash mismatch triggers refill
- Only caches canonical tip lookups (avoids pollution from non-canonical queries)

## Design Decisions

- **Single-entry vs DashMap**: Single-entry cache is sufficient since FCU repeatedly queries the same canonical head. DashMap would add unnecessary overhead (hashing, sharding) for a single-threaded handler with single-entry access pattern.
- **parking_lot::RwLock**: Already used in the codebase, provides fair scheduling, no poisoning
- **No explicit invalidation**: Hash-based keying means stale cache naturally misses when canonical head changes

## Cache Hit Analysis

The cache helps when:
1. FCU has payload attributes (validators building blocks)
2. Head is already canonical
3. Canonical tip was persisted (removed from in-memory `blocks_by_hash`)

With `DEFAULT_MEMORY_BLOCK_BUFFER_TARGET = 0`, canonical head is removed after persistence, making this cache useful for validators building multiple payloads on the same head.

## Expected Impact

- **100-500µs latency reduction** per avoided DB lookup
- **Validators** benefit when building payloads after persistence
- **Zero risk**: Falls back to existing DB path on miss

## Testing

All 111 existing engine-tree tests pass.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1932,10 +1932,10 @@ where
         // Check single-entry cache for canonical tip (avoids DB on repeated FCU)
         {
             let cache = self.canonical_header_cache.read();
-            if let Some((cached_hash, cached_header)) = cache.as_ref() {
-                if *cached_hash == hash {
-                    return Ok(Some(cached_header.clone()));
-                }
+            if let Some((cached_hash, cached_header)) = cache.as_ref() &&
+                *cached_hash == hash
+            {
+                return Ok(Some(cached_header.clone()));
             }
         }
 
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
looks correct, benching now

mattsse (changes requested):
I dont think this is useful because the parent block should be in mem here

this also doesnt include any clean up which we'd need to do any time  we remove from the mem set after persistence or on reorg, so this feels like an invasive change without clear impact

PR #21191, opened 2026-01-19
feat(cli): add --rocksdb.* flags for RocksDB table routing

### LINKED ISSUE #20393: Add CLI flags to enable RocksDB storage

yongkangc:
## Summary
Introduces `RocksDbArgs` struct with dedicated `--rocksdb.*` prefix for RocksDB table routing.

## Flags
- `--rocksdb.tx-hash` - Route tx hash -> number table to RocksDB
- `--rocksdb.storages-history` - Route storages history to RocksDB  
- `--rocksdb.account-history` - Route account history to RocksDB

All flags are genesis-initialization-only (changing later requires re-sync).

## PR Stack
1. [**This PR** - CLI flags](https://github.com/paradigmxyz/reth/pull/21191)
2. [Validation logic (depends on this)](https://github.com/paradigmxyz/reth/pull/21192)
3. [Docs (independent)](https://github.com/paradigmxyz/reth/pull/21190)

closes https://github.com/paradigmxyz/reth/issues/20393

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -68,7 +68,7 @@ use reth_provider::{
     providers::{NodeTypesForProvider, ProviderNodeTypes, RocksDBProvider, StaticFileProvider},
     BlockHashReader, BlockNumReader, DatabaseProviderFactory, ProviderError, ProviderFactory,
     ProviderResult, RocksDBProviderFactory, StageCheckpointReader, StaticFileProviderBuilder,
-    StaticFileProviderFactory, StorageSettings,
+    StaticFileProviderFactory,
 };
 use reth_prune::{PruneModes, PrunerBuilder};
 use reth_rpc_builder::config::RethRpcServerConfig;
@@ -674,37 +674,18 @@ where
         Ok(())
     }
 
-    /// Computes storage settings from CLI args and logs `RocksDB` routing if enabled.
-    fn storage_settings_with_logging(&self) -> StorageSettings {
-        let base_settings = self.node_config().static_files.to_settings();
-        let (settings, grouped_enabled) =
-            self.node_config().rocksdb.apply_to_settings(base_settings);
-
-        if grouped_enabled {
-            let enabled: Vec<&str> = [
-                settings.transaction_hash_numbers_in_rocksdb.then_some("tx-hash"),
-                settings.storages_history_in_rocksdb.then_some("storages-history"),
-                settings.account_history_in_rocksdb.then_some("account-history"),
-            ]
-            .into_iter()
-            .flatten()
-            .collect();
-            info!(target: "reth::cli", "RocksDB routing enabled for tables: {}", enabled.join(", "));
-        }
-
-        settings
-    }
-
     /// Convenience function to [`Self::init_genesis`]
     pub fn with_genesis(self) -> Result<Self, InitStorageError> {
-        let settings = self.storage_settings_with_logging();
+        let base_settings = self.node_config().static_files.to_settings();
+        let (settings, _) = self.node_config().rocksdb.apply_to_settings(base_settings);
         init_genesis_with_settings(self.provider_factory(), settings)?;
         Ok(self)
     }
 
     /// Write the genesis block and state if it has not already been written
     pub fn init_genesis(&self) -> Result<B256, InitStorageError> {
-        let settings = self.storage_settings_with_logging();
+        let base_settings = self.node_config().static_files.to_settings();
+        let (settings, _) = self.node_config().rocksdb.apply_to_settings(base_settings);
         init_genesis_with_settings(self.provider_factory(), settings)
     }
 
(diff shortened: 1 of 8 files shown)


crates/node/core/src/args/rocksdb.rs
```diff
@@ -0,0 +1,158 @@
...
+    use super::*;
+    use clap::Parser;
+
+    #[derive(Parser)]
+    struct CommandParser<T: Args> {
+        #[command(flatten)]
+        args: T,
+    }
+
+    #[test]
```
yongkangc, line 71: default is None

crates/node/builder/src/launch/common.rs
```diff
@@ -676,19 +676,17 @@ where
...
     }
 
     /// Write the genesis block and state if it has not already been written
     pub fn init_genesis(&self) -> Result<B256, InitStorageError> {
-        init_genesis_with_settings(
-            self.provider_factory(),
-            self.node_config().static_files.to_settings(),
-        )
+        let base_settings = self.node_config().static_files.to_settings();
+        let (settings, _) = self.node_config().rocksdb.apply_to_settings(base_settings);
```
joshieDo: could we instead have self.node_config().storage_settings() ?

yongkangc: added [662684a](https://github.com/paradigmxyz/reth/pull/21191/commits/662684a74f14f3c921855d8c271878db1cfcfa95)

yongkangc:
@joshieDo refactored into the function `storage_settings() ` based on your comment

crates/node/core/src/node_config.rs
```diff
@@ -350,6 +357,14 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
         self.pruning.prune_config(&self.chain)
     }
 
+    /// Returns the effective storage settings derived from static-file settings plus
+    /// `RocksDB` table-routing overrides.
+    pub fn storage_settings(&self) -> StorageSettings {
+        let base = self.static_files.to_settings();
+        let (settings, _) = self.rocksdb.apply_to_settings(base);
```
joshieDo: maybe `to_settings().merge(self.rocksdb.to_settings())` ? 

or we just create it here and remove the to_settings from static_files if its not being used anywhere else

yongkangc: maybe `merge()` method to StorageSettings and simplify `RocksDbArgs::apply_to_settings` to just apply overrides

joshieDo: hm, feels unintuitive to me right now, with `merge`  only merging specific fields. 

i'd say, lets just build StorageSettings on the spot.

crates/node/core/src/args/rocksdb.rs
```diff
@@ -0,0 +1,154 @@
...
+            settings.transaction_hash_numbers_in_rocksdb = value;
+        }
+        if let Some(value) = self.storages_history {
+            settings.storages_history_in_rocksdb = value;
+        }
+        if let Some(value) = self.account_history {
+            settings.account_history_in_rocksdb = value;
+        }
+
+        settings
```
Rjected: I don't think I prefer `any_true` to explicitly setting args. While it's nice to have one flag that sets all the rocksdb args to true, it's non intuitive to do this on _any_ flag set. If we are going to do something like this I would rather have some sort of `all` flag which errors on startup if we explicitly set one of these to false.

crates/node/core/src/node_config.rs
yongkangc: yeah that feels more straightfoward instead of indrection

crates/node/core/src/node_config.rs
```diff
@@ -350,6 +357,22 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
         self.pruning.prune_config(&self.chain)
     }
 
+    /// Returns the effective storage settings derived from static-file and `RocksDB` CLI args.
+    pub fn storage_settings(&self) -> StorageSettings {
```
joshieDo, line 361: we can do it in this PR or on a follow-up, but we probably want to remove static_files.to_settings()  i think ?

and we should make sure that the following behaviour introduced here
https://github.com/paradigmxyz/reth/pull/21208/files  also applies to rocksdb

yongkangc: makes sense, think we might want to do it in a separate pr - will be on this

Rjected (approved):
lgtm

PR #21193, opened 2026-01-19
perf(chain-state): parallelize into_sorted with rayon

mattsse:
## Summary

Parallelizes the `into_sorted` conversion for hashed state and trie updates using `rayon::join`.

## Changes

- Use `rayon::join` to convert hashed state and trie updates to sorted form concurrently
- Add `rayon` feature to `reth-chain-state` crate
- Feature-gate parallel implementation with sequential fallback when rayon is disabled

## Follow-up

See #21202 for the parallel merge implementation that builds on this change.

### DIFF
--- a/crates/chain-state/src/deferred_trie.rs
+++ b/crates/chain-state/src/deferred_trie.rs
@@ -163,14 +163,29 @@ impl DeferredTrieData {
         anchor_hash: B256,
         ancestors: &[Self],
     ) -> ComputedTrieData {
-        let sorted_hashed_state = match Arc::try_unwrap(hashed_state) {
-            Ok(state) => state.into_sorted(),
-            Err(arc) => arc.clone_into_sorted(),
-        };
-        let sorted_trie_updates = match Arc::try_unwrap(trie_updates) {
-            Ok(updates) => updates.into_sorted(),
-            Err(arc) => arc.clone_into_sorted(),
-        };
+        #[cfg(feature = "rayon")]
+        let (sorted_hashed_state, sorted_trie_updates) = rayon::join(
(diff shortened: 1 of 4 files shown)


yongkangc (approved):
makes sense to me, but not sure if we want to bench it as if there are not alot of updates it might be more expensive to parallelize


for dan's big block hashpoststate is about 3k, and trieupdatessorted about 15k so that makes sense

crates/chain-state/src/deferred_trie.rs
```diff
@@ -163,14 +163,29 @@ impl DeferredTrieData {
...
     ) -> ComputedTrieData {
-        let sorted_hashed_state = match Arc::try_unwrap(hashed_state) {
-            Ok(state) => state.into_sorted(),
-            Err(arc) => arc.clone_into_sorted(),
-        };
-        let sorted_trie_updates = match Arc::try_unwrap(trie_updates) {
-            Ok(updates) => updates.into_sorted(),
-            Err(arc) => arc.clone_into_sorted(),
-        };
+        #[cfg(feature = "rayon")]
```
DaniPopes, line 166: why is rayon feature gated in the first place?

crates/chain-state/src/deferred_trie.rs
DaniPopes, line 160: can you split changes to this function into another pr, then we can benchmark this pr with the new parallel merge algo?

DaniPopes (changes requested):

crates/chain-state/src/deferred_trie.rs
mattsse: all of these crates are no-std

DaniPopes (approved):

PR #21195, opened 2026-01-19
fix(engine): separate newPayload finish for FCU timing

Fibonacci747:
`forkchoiceUpdated` metrics consumed `latest_finish_at` via `take()`, which cleared the timestamp used to compute `time_between_new_payloads`. As a result, idle time between consecutive `newPayload` calls was regularly missing after FCU events.
Separated the finish timestamp used for FCU timing from the one used for newPayload intervals, and updated FCU metric handling to consume only the dedicated timestamp.

### DIFF
--- a/crates/engine/tree/src/tree/metrics.rs
+++ b/crates/engine/tree/src/tree/metrics.rs
@@ -239,7 +239,7 @@ impl ForkchoiceUpdatedMetrics {
     pub(crate) fn update_response_metrics(
         &self,
         start: Instant,
-        latest_new_payload_at: &mut Option<Instant>,
+        latest_new_payload_finish_for_fcu: &mut Option<Instant>,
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/metrics.rs
```diff
@@ -271,6 +272,9 @@ pub(crate) struct NewPayloadStatusMetrics {
     /// Finish time of the latest new payload call.
     #[metric(skip)]
     pub(crate) latest_finish_at: Option<Instant>,
+    /// Finish time of the latest new payload call for forkchoice timing.
+    #[metric(skip)]
+    pub(crate) latest_finish_for_fcu: Option<Instant>,
```
mattsse, line 277: this isnt even used I believe

PR #21197, opened 2026-01-19
feat(trie): add AsyncAccountValueEncoder for V2 proof computation

mediocregopher:
Building blocks working towards https://github.com/paradigmxyz/reth/issues/19514

* Change the deferred_encoder method signature from &self to &mut self to allow encoders to maintain mutable state during proof generation.

* Add AsyncAccountValueEncoder which implements LeafValueEncoder to dispatch storage proof computation asynchronously while walking the account trie. This enables parallel storage root computation during V2 account proof generation by dispatching work to storage workers and collecting results via crossbeam channels.

* Change StorageProofResultMessage to use StateProofError instead of ParallelStateRootError for cleaner error propagation from storage workers.

### DIFF
--- a/crates/trie/parallel/src/value_encoder.rs
+++ b/crates/trie/parallel/src/value_encoder.rs
@@ -68,7 +68,7 @@ impl DeferredValueEncoder for AsyncAccountDeferredValueEncoder {
 }
 
 /// Implements the [`LeafValueEncoder`] trait for accounts using a [`CrossbeamSender`] to dispatch
-/// and compute storage roots asyncronously. Can also accept a set of already dispatched account
+/// and compute storage roots asynchronously. Can also accept a set of already dispatched account
 /// storage proofs, for cases where it's possible to determine some necessary accounts ahead of
 /// time.
 pub(crate) struct AsyncAccountValueEncoder {
--- a/crates/trie/parallel/src/value_encoder.rs
+++ b/crates/trie/parallel/src/value_encoder.rs
@@ -1,0 +1,185 @@
+use crate::proof_task::{
+    StorageProofInput, StorageProofResult, StorageProofResultMessage, StorageWorkerJob,
+};
+use alloy_primitives::{map::B256Map, B256};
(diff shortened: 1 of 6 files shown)


yongkangc (approved):

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -0,0 +1,185 @@
...
+/// time.
+pub(crate) struct AsyncAccountValueEncoder {
+    storage_work_tx: CrossbeamSender<StorageWorkerJob>,
+    /// Storage proof jobs which were dispatched ahead of time.
+    dispatched: B256Map<CrossbeamReceiver<StorageProofResultMessage>>,
+    /// Storage roots which have already been computed. This can be used only if a storage proof
+    /// wasn't dispatched for an account, otherwise we must consume the proof result.
+    cached_storage_roots: Arc<DashMap<B256, B256>>,
+    /// Tracks storage proof results received from the storage workers. [`Rc`] + [`RefCell`] is
+    /// required because [`DeferredValueEncoder`] cannot have a lifetime.
```
mattsse, line 82: why is that?

is this a refcell because we wanto mutate this on the callsite via `.deferred_encoder()`?

we could maybe try `type DeferredEncoder<'a>` but perhaps not useful here

mattsse (approved):
I think I could follow along here

the rc refcell seems okay, especially now that we changed the trait fn to oparate on mut but there's likely some GAT lifetime hacks we can do here

crates/trie/parallel/src/value_encoder.rs
mediocregopher: > is this a refcell because we wanto mutate this on the callsite via .deferred_encoder()?

yup exactly

I tried adding a lifetime on the associated type, the problem is that we're currently storing the DeferredEncoder on the ProofCalculator as a field, so we end up with self-referential struct which is it's own kind of mess. This seemed like the least bad option.

PR #21202, opened 2026-01-20
perf(trie): parallelize merge_ancestors_into_overlay

### LINKED ISSUE #21187: perf(trie): parallelize merge_ancestors_into_overlay for deferred trie computation
## Description

`merge_ancestors_into_overlay` in `deferred_trie.rs` currently processes ancestors sequentially, which can become a bottleneck when rebuilding overlays with many in-memory ancestors.

## Profiler

See [Firefox Profiler](https://profiler.firefox.com/public/qd1rhjm07s2m0heagj0h2g4448ma8r9ndj9xd6r/flame-graph/?globalTrackOrder=201&hiddenGlobalTracks=01&hiddenLocalTracksByPid=1487349.2-0w2~1487349.1-xgxhxjwxly0wz4zazcwzjzlwAs&localTrackOrderByPid=1487349.1-xkxlzbxizaxhxgzcwzjzlwAsxmwz4Dk0wxfz5wz9zkAtwDixjDj&range=15647m2390&sourceViewIndex=882&symbolServer=http%3A%2F%2F127.0.0.1%3A3001%2F271h2d1w7a4xbb8v1ibqwc4pplmb4kxnb5add5s&thread=Dk&v=12) showing time spent in the merge path.

## Current behavior

```rust
for ancestor in ancestors {
    let ancestor_data = ancestor.wait_cloned();
    state_mut.extend_ref(ancestor_data.hashed_state.as_ref());
    nodes_mut.extend_ref(ancestor_data.trie_updates.as_ref());
}
```

This is O(n) sequential:
- Each `wait_cloned()` may block waiting for deferred computation
- Each `extend_ref()` runs sequentially

## Proposed optimization

1. **Parallel collection**: Call `wait_cloned()` on all ancestors in parallel via `par_iter()`
2. **Horizontal parallelism**: Use `rayon::join()` to merge hashed state and trie updates concurrently
3. **Tree reduction**: Add `merge_parallel()` methods to `HashedPostStateSorted` and `TrieUpdatesSorted` that use tree reduction for O(log n) merge depth

```
Before: A0 → A1 → A2 → ... → An (sequential)

After:
  1. par_iter: collect all ancestor data in parallel
  2. rayon::join:
     ├─ states: tree reduce (log n depth)
     └─ nodes:  tree reduce (log n depth)
```

Image

## Impact

This is a rare fallback path (normally parent overlay is reused), but when triggered with many ancestors it can be significant.

mattsse:
## Summary

Adds parallel merge support for the deferred trie computation fallback path when rebuilding overlays from ancestors.

## Changes

- Add `merge_parallel` methods to `HashedPostStateSorted` and `TrieUpdatesSorted` using recursive divide-and-conquer with `rayon::join`
- Feature-gate parallel implementation behind `rayon` feature with sequential fallback
- Add early exit when ancestors list is empty
- Use `rayon::join` to merge states and updates concurrently
- Use `extend_ref_and_sort` directly for efficiency

## Implementation

The parallel merge uses a recursive divide-and-conquer pattern with `rayon::join`:

```
Level 0:  A0    A1    A2    A3    A4    A5    A6    A7
            \  /        \  /        \  /        \  /
Level 1:   M01          M23         M45         M67    ← parallel via rayon::join
              \        /              \        /
Level 2:      M0123                   M4567            ← parallel via rayon::join
                    \                /
Level 3:            M01234567 (final)
```

This ensures order preservation (critical for "newer overwrites older" semantics) while benefiting from O(log n) parallel depth.

Follow-up to #21193 which adds the parallel `into_sorted` changes.

Closes #21187

### DIFF
--- a/crates/chain-state/src/deferred_trie.rs
+++ b/crates/chain-state/src/deferred_trie.rs
@@ -243,8 +243,53 @@ impl DeferredTrieData {
     /// In normal operation, the parent always has a cached overlay and this
     /// function is never called.
     ///
-    /// Iterates ancestors oldest -> newest, then extends with current block's data,
-    /// so later state takes precedence.
+    /// When the `rayon` feature is enabled, uses parallel collection and merge:
+    /// 1. Collects ancestor data in parallel (each `wait_cloned()` may compute)
+    /// 2. Merges hashed state and trie updates in parallel with each other
+    /// 3. Uses tree reduction within each merge for O(log n) depth
+    #[cfg(feature = "rayon")]
+    fn merge_ancestors_into_overlay(
+        ancestors: &[Self],
+        sorted_hashed_state: &HashedPostStateSorted,
+        sorted_trie_updates: &TrieUpdatesSorted,
+    ) -> TrieInputSorted {
+        // Early exit: no ancestors means just wrap current block's data
+        if ancestors.is_empty() {
+            return TrieInputSorted::new(
+                Arc::new(sorted_trie_updates.clone()),
+                Arc::new(sorted_hashed_state.clone()),
+                Default::default(),
+            );
+        }
+
+        // Collect ancestor data, unzipping states and updates into Arc slices
+        let (states, updates): (Vec<_>, Vec<_>) = ancestors
+            .iter()
+            .map(|a| {
+                let data = a.wait_cloned();
+                (data.hashed_state, data.trie_updates)
+            })
+            .unzip();
+
+        // Merge state and nodes in parallel with each other using tree reduction
+        let (state, nodes) = rayon::join(
+            || {
+                let mut merged = HashedPostStateSorted::merge_parallel(&states);
+                merged.extend_ref_and_sort(sorted_hashed_state);
+                merged
+            },
+            || {
+                let mut merged = TrieUpdatesSorted::merge_parallel(&updates);
+                merged.extend_ref_and_sort(sorted_trie_updates);
+                merged
+            },
+        );
+
+        TrieInputSorted::new(Arc::new(nodes), Arc::new(state), Default::default())
+    }
+
+    /// Merge all ancestors and current block's data into a single overlay (sequential fallback).
+    #[cfg(not(feature = "rayon"))]
     fn merge_ancestors_into_overlay(
         ancestors: &[Self],
         sorted_hashed_state: &HashedPostStateSorted,
--- a/crates/trie/common/src/hashed_state.rs
+++ b/crates/trie/common/src/hashed_state.rs
@@ -714,33 +714,31 @@ impl HashedPostStateSorted {
     /// Parallel batch-merge sorted hashed post states. Slice is **oldest to newest**.
     ///
     /// This is more efficient than sequential `extend_ref` calls when merging many states,
-    /// as it processes all states in parallel with tree reduction.
+    /// as it processes all states in parallel with tree reduction using divide-and-conquer.
     #[cfg(feature = "rayon")]
     pub fn merge_parallel(states: &[Arc<Self>]) -> Self {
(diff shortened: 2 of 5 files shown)


mediocregopher (approved):

crates/chain-state/src/deferred_trie.rs
```diff
@@ -228,8 +243,53 @@ impl DeferredTrieData {
...
+        }
+
+        // Collect ancestor data, unzipping states and updates into Arc slices
+        let (states, updates): (Vec<_>, Vec<_>) = ancestors
+            .iter()
+            .map(|a| {
+                let data = a.wait_cloned();
+                (data.hashed_state, data.trie_updates)
+            })
+            .unzip();
```
Rjected, line 272: this iter is not parallel right? so `Collects ancestor data in parallel (each `wait_cloned()` may compute)` is not accurate?

Rjected (approved):
seems like an easy win, will bench this

crates/trie/common/src/hashed_state.rs
```diff
@@ -710,6 +710,38 @@ impl HashedPostStateSorted {
...
+        use rayon::prelude::*;
+
+        if states.is_empty() {
+            return Self::default();
+        }
+        if states.len() == 1 {
+            return (*states[0]).clone();
+        }
+
+        // Ordered parallel tree reduction: merge pairs in parallel, preserving order
```
DaniPopes: we can do a lot better by manually using rayon::join to continuously work. currently this will stop and synchronize on every iteration, so a large batch in the middle will block all the batches on the ends to be computed until the large one finishes.

DaniPopes (changes requested):

crates/trie/common/src/hashed_state.rs
DaniPopes: For example (clauded) maybe like this?

```rust
fn parallel_merge_tree(states: &[Arc<Self>]) -> Self {
    match states.len() {
        0 => panic!("Cannot merge empty states"),
        1 => states[0].as_ref().clone(),
        2 => Self::merge_batch([
            states[1].as_ref().clone(),
            states[0].as_ref().clone(),
        ].iter().cloned()),
        n => {
            let mid = n / 2;
            let (left, right) = rayon::join(
                || parallel_merge_tree(&states[..mid]),
                || parallel_merge_tree(&states[mid..]),
            );
            // Merge right-then-left to maintain newest-first for merge_batch
            Self::merge_batch([right, left].iter().cloned())
        }
    }
```

yongkangc (approved):

crates/trie/common/src/updates.rs
```diff
@@ -697,6 +698,36 @@ impl TrieUpdatesSorted {
...
+    /// This is more efficient than sequential `extend_ref` calls when merging many updates,
+    /// as it processes all updates in parallel with tree reduction using divide-and-conquer.
+    #[cfg(feature = "rayon")]
+    pub fn merge_parallel(updates: &[Arc<Self>]) -> Self {
+        fn parallel_merge_tree(updates: &[Arc<TrieUpdatesSorted>]) -> TrieUpdatesSorted {
+            match updates.len() {
+                0 => TrieUpdatesSorted::default(),
+                1 => Arc::unwrap_or_clone(Arc::clone(&updates[0])),
+                2 => {
+                    let mut acc = Arc::unwrap_or_clone(Arc::clone(&updates[0]));
```
DaniPopes: Arc::unwrap_or_clone(Arc::clone(x)) -> x.as_ref().clone()

mattsse: ah right because we have a slice here

DaniPopes (approved):

PR #21203, opened 2026-01-20
feat: make changes for Bal devnet 2

Rimeeeeee:
TODO:
*  ~rpc endpoints~ DONE [ref](https://github.com/paradigmxyz/reth/pull/21245)


crates/storage/codecs/src/alloy/block_access_list.rs
mattsse, line 1: we can get away without doing this, so we dont need to store them in the db

and can remove this

crates/storage/db-api/src/tables/mod.rs
```diff
@@ -350,6 +351,12 @@ tables! {
         type Value = StoredBlockWithdrawals;
     }
 
+    /// Stores the block access lists.
+    table BlockAccessLists {
+        type Key = BlockNumber;
+        type Value = StoredBlockAccessList;
+    }
```
mattsse: this we dont need

mattsse (changes requested):

PR #21209, opened 2026-01-20
chore: use nightly for cargo docs

BearSoltokien:
Enforce nightly toolchain for cargo docs commands, as unstable options in RUSTDOCFLAGS require it.


mattsse (commented):
doesnt matter

PR #21211, opened 2026-01-20
feat(cli): add `reth db checksum mdbx/static-file` command

gakonst:
## Summary

Extends the `reth db checksum` command to support both database tables and static file segments, following the same pattern as `db get` and `db clear`.

## Breaking Change

The command now uses `mdbx` and `static-file` subcommands:

```bash
# Before
reth db checksum PlainAccountState

# After  
reth db checksum mdbx PlainAccountState
```

## Usage

```bash
# Database table checksum
reth db checksum mdbx PlainAccountState
reth db checksum mdbx PlainAccountState --start-key 0x... --end-key 0x...
reth db checksum mdbx Transactions --limit 100000

# Static file segment checksum
reth db checksum static-file account-change-sets
reth db checksum static-file transactions --start-block 0 --end-block 1000000
reth db checksum static-file receipts --limit 100000
```

## Supported static file segments
- `headers`
- `transactions`  
- `receipts`
- `transaction-senders`
- `account-change-sets`

## Motivation

When data is moved from DB tables to static files (e.g., AccountChangeSets, TransactionSenders), there was no way to verify the static file contents match between different nodes. This enables direct comparison.

## Implementation

Uses the same hashing approach as the existing mdbx checksum (foldhash with RETHRETH seed). Iterates through static file rows using the existing NippyJar cursor infrastructure.


crates/cli/commands/src/db/checksum.rs
```diff
@@ -4,40 +4,77 @@ use crate::{
...
+use reth_static_file_types::StaticFileSegment;
 use std::{
     hash::{BuildHasher, Hasher},
     sync::Arc,
     time::{Duration, Instant},
 };
 use tracing::{info, warn};
 
+/// Interval for logging progress during checksum computation.
+const PROGRESS_LOG_INTERVAL: usize = 100_000;
```
yongkangc, line 26: i guess this is good enough?

crates/cli/commands/src/db/checksum.rs
```diff
@@ -4,40 +4,77 @@ use crate::{
...
+
+/// Creates a new hasher with the standard seed used for checksum computation.
+fn checksum_hasher() -> impl Hasher {
+    FixedState::with_seed(u64::from_be_bytes(*b"RETHRETH")).build_hasher()
+}
+
```
yongkangc: should we have this function below so that it follows a better file hieracy? 

i.e above `checksum_static_file`

crates/cli/commands/src/db/checksum.rs
```diff
@@ -47,16 +84,107 @@ impl Command {
...
+            if total.is_multiple_of(PROGRESS_LOG_INTERVAL) {
+                info!("Hashed {total} entries.");
+            }
+
+            if total >= limit {
+                break;
+            }
+        }
+
+        drop(jar_provider);
```
yongkangc, line 170: why do we need to drop the provider before the end of the iteration?

crates/cli/commands/src/db/checksum.rs
```diff
@@ -47,16 +84,107 @@ impl Command {
...
+        static_file_provider.remove_cached_provider(segment, fixed_block_range.end());
+
+        if total >= limit {
+            break;
+        }
```
yongkangc: do we still need this outer loop since the inner already breaks?

crates/cli/commands/src/db/checksum.rs
```diff
@@ -47,16 +84,107 @@ impl Command {
...
+            Subcommand::StaticFile { segment, start_block, end_block, limit } => {
+                checksum_static_file(tool, segment, start_block, end_block, limit)?;
+            }
+        }
+
         Ok(())
     }
 }
 
+fn checksum_static_file<N: CliNodeTypes<ChainSpec: EthereumHardforks>>(
```
yongkangc, line 101: should we add tests for this file?

yongkangc (approved):
@joshieDo lgtm!

PR #21212, opened 2026-01-20
dns/testing: deterministic fork id; avoid rng.gen()

AdekunleBamz:
## Summary
- Make DNS discovery ENR fork-id test deterministic by replacing `rand::random()` with a seeded `StdRng`.
- Log invalid ENR `eth` fork-id entries at `trace` instead of silently dropping decode errors.
- Replace `rng.gen()` with `rng.random()` in testing utils to avoid the Rust 2024 `gen` keyword issue.

## Testing
- `cargo test -p reth-dns-discovery -p reth-testing-utils`


mattsse (commented):
no

PR #21216, opened 2026-01-20
fix(rocksdb): correct history index divergence causing nonce errors after unwind

yongkangc:
Fix execution divergence in RocksDB hybrid storage that caused "nonce too high" errors during `engine_newPayload` after unwinding blocks.

**Problem**

When running `reth-bench new-payload-fcu` with RocksDB hybrid storage enabled, the benchmark would fail with nonce errors after unwinding and replaying blocks. The root cause was a mismatch between what RocksDB and MDBX recorded in their history indices.

### 1. Over-approximation of Account History
RocksDB was recording history indices for **all touched accounts**, including those with only storage changes. This diverged from MDBX's `AccountChangeSets` which only records accounts where account-info (nonce, balance, code_hash) changed or the account was created/destroyed.

**Fix:** Use `account.is_info_changed() || account.was_destroyed()` instead of `!account.status.is_not_modified()` to match MDBX semantics exactly.

### 2. Missing Plain State Fallback
When RocksDB history lookup found no entry for an address, it returned `HistoryInfo::NotYetWritten` (account doesn't exist). In hybrid storage, accounts modified **before** RocksDB was enabled exist only in MDBX.

**Fix:** Always fall back to `HistoryInfo::MaybeInPlainState` so pre-existing accounts are correctly looked up in MDBX plain state.

### 3. Architectural Invariant Enforcement (per Joshie's feedback)

The original fix changed commit order (MDBX before RocksDB). Per Joshie's feedback, the proper fix is not changing commit order but **enforcing the architectural invariant**:

> RocksDB history indices (`AccountsHistory`, `StoragesHistory`) are derived data that index into changesets. They can only be safely stored in RocksDB when changesets are in static files (crash-stable), not MDBX (can be unwound).

**Fix:** 
- Reverted commit order back to original: `static_file → RocksDB → MDBX`
- Added `StorageSettings::validate()` to enforce: `account_history_in_rocksdb OR storages_history_in_rocksdb` requires `account_changesets_in_static_files=true`
- Validation runs at **both** genesis init and database load (catches existing invalid configs)
- Added `ProviderError::InvalidStorageSettings` for clear error messages
- Updated invariants.rs documentation

## Additional Improvements

- Improved sentinel shard handling in invariants: pruning now correctly filters block numbers within sentinel shards (`highest_block_number=u64::MAX`)
- Pending history writes are accumulated across `save_blocks()` calls and materialized once at commit time, fixing read-your-writes issues

## Testing

- All 123 RocksDB tests pass, including 16 invariant tests
- Validation tests for `StorageSettings::validate()` function
- Verified that nodes with pre-existing invalid configurations fail with clear error message:
  ```
  Error: invalid storage settings: RocksDB history indices require account_changesets_in_static_files=true; otherwise crash/unwind can diverge RocksDB history from MDBX changesets
  ```

Reference thread:
https://tempoxyz.slack.com/archives/C09FQDW2ZRP/p1765829975230549

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -3,6 +3,19 @@
 //! This module provides consistency checks for tables stored in `RocksDB`, similar to the
 //! consistency checks for static files. The goal is to detect and potentially heal
 //! inconsistencies between `RocksDB` data and MDBX checkpoints.
+//!
+//! # Safety Invariant
+//!
+//! `RocksDB` history indices (`AccountsHistory`, `StoragesHistory`) are derived data that
+//! index into changesets. They can only be safely stored in `RocksDB` when changesets are
+//! in static files (append-only, crash-stable), not MDBX (can be unwound).
+//!
+//! This is because after a crash during unwind, `RocksDB` history could reference block
+//! numbers whose changesets were already removed from MDBX, causing "future state lookups"
+//! or missing data.
+//!
+//! This invariant is enforced by [`StorageSettings::validate`] at configuration time,
+//! which requires `account_changesets_in_static_files=true` when any history is in `RocksDB`.
 
 use super::RocksDBProvider;
 use crate::StaticFileProviderFactory;
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -253,14 +253,21 @@ impl RocksDBProvider {
                     return Ok(None);
                 }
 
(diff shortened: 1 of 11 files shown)


crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -253,46 +253,56 @@ impl RocksDBProvider {
...
+                        {
+                            max_sentinel_block = max_in_shard;
+                        }
+                    } else {
                         found_non_sentinel = true;
                         if highest > max_highest_block {
                             max_highest_block = highest;
                         }
                     }
                 }
```
yongkangc, line 289: add comment for this chunk

PR #21217, opened 2026-01-20
feat(cli): add `reth db checksum rocksdb` command

gakonst:
## Summary

Extends the `reth db checksum` command to support RocksDB tables, following the same pattern as the existing `mdbx` and `static-file` subcommands from #21211.

> **Note**: This PR depends on #21211 and should be merged after it.

## Usage

```bash
reth db checksum rocksdb transaction-hash-numbers
reth db checksum rocksdb accounts-history --limit 100000
reth db checksum rocksdb storages-history
```

## Supported RocksDB tables

- `transaction-hash-numbers` - Transaction hash to transaction number mapping
- `accounts-history` - Account history indices  
- `storages-history` - Storage history indices

## Motivation

When data is stored in RocksDB (e.g., TransactionHashNumbers, AccountsHistory, StoragesHistory based on storage settings), there was no way to verify the RocksDB contents match between different nodes. This enables direct comparison using the same hashing approach as the existing mdbx/static-file checksums.

## Implementation

- Adds `RocksDbTable` enum for the three supported tables
- Adds `Rocksdb` subcommand (cfg-gated on `unix` + `rocksdb` feature)
- Uses the same `foldhash` hasher with `RETHRETH` seed for consistency
- Iterates through RocksDB entries using `RocksDBProvider::iter()`
- Adds `rocksdb` feature to `reth-cli-commands` that enables `reth-provider/rocksdb`

### DIFF
--- a/crates/cli/commands/Cargo.toml
+++ b/crates/cli/commands/Cargo.toml
@@ -131,5 +131,4 @@ arbitrary = [
     "reth-ethereum-primitives/arbitrary",
 ]
 
-edge = ["reth-db-common/edge", "reth-stages/rocksdb", "rocksdb"]
-rocksdb = ["reth-provider/rocksdb"]
+edge = ["reth-db-common/edge", "reth-stages/rocksdb", "reth-provider/rocksdb"]
--- a/crates/cli/commands/Cargo.toml
+++ b/crates/cli/commands/Cargo.toml
@@ -131,4 +131,5 @@ arbitrary = [
     "reth-ethereum-primitives/arbitrary",
 ]
 
-edge = ["reth-db-common/edge", "reth-stages/rocksdb"]
+edge = ["reth-db-common/edge", "reth-stages/rocksdb", "rocksdb"]
+rocksdb = ["reth-provider/rocksdb"]
--- a/crates/cli/commands/src/db/checksum/rocksdb.rs
+++ b/crates/cli/commands/src/db/checksum/rocksdb.rs
@@ -1,0 +1,111 @@
+//! RocksDB checksum implementation.
+
+use super::{checksum_hasher, PROGRESS_LOG_INTERVAL};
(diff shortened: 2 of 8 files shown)


crates/cli/commands/Cargo.toml
```diff
@@ -131,4 +131,5 @@ arbitrary = [
     "reth-ethereum-primitives/arbitrary",
 ]
 
-edge = ["reth-db-common/edge", "reth-stages/rocksdb"]
+edge = ["reth-db-common/edge", "reth-stages/rocksdb", "rocksdb"]
+rocksdb = ["reth-provider/rocksdb"]
```
joshieDo: edge should be enabling rocksdb in downstream crates instead of having rocksdb as a feature

crates/cli/commands/src/db/checksum/rocksdb.rs
```diff
@@ -0,0 +1,106 @@
...
+        table.name(),
+        checksum,
+        total,
+        elapsed
+    );
+
+    Ok(())
+}
+
+/// Computes checksum for a specific RocksDB table by iterating over raw bytes.
```
joshieDo:
```suggestion
/// Computes checksum for a specific RocksDB table by iterating over rows.
```

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -656,6 +656,16 @@ impl RocksDBProvider {
         Ok(RocksDBIter { inner: iter, _marker: std::marker::PhantomData })
     }
 
+    /// Creates a raw iterator over all entries in the specified table.
+    ///
+    /// Returns raw `(key_bytes, value_bytes)` pairs without decoding.
+    /// This is more efficient when you need to process raw bytes directly.
```
```suggestion
```

yongkangc (approved):
lgtm

PR #21200, opened 2026-01-20
perf(db): stack-allocate ShardedKey and StorageShardedKey encoding

gakonst:
## Summary

This PR eliminates heap allocations in hot database paths by changing the `Encode` implementation for `ShardedKey<Address>` and `StorageShardedKey` to return fixed-size stack-allocated arrays instead of `Vec<u8>`.

**Inspired by [Anza Labs' PR #3603](https://github.com/anza-xyz/agave/pull/3603)** on the Solana agave repository, which saved ~20k allocations/sec by moving RocksDB keys from heap to stack.

## Changes

- `ShardedKey<Address>::encode()` now returns `[u8; 28]` (stack-allocated)
- `StorageShardedKey::encode()` now returns `[u8; 60]` (stack-allocated)
- Added criterion benchmarks for measuring encode/decode performance
- Added unit tests for roundtrip encoding

## Impact

These keys are used in the `AccountsHistory` and `StoragesHistory` tables which are frequently accessed during state lookups and history queries. Every get/put/walk operation on these tables previously allocated a `Vec<u8>` for the key encoding.

### Benchmark Results (10k keys per iteration)

| Operation | Throughput |
|-----------|------------|
| `ShardedKey<Address>::encode` | ~168 Melem/s |
| `StorageShardedKey::encode` | ~80 Melem/s |

Run benchmarks with:
```bash
cargo bench -p reth-db-api --bench sharded_key_encode
```

## Background

From Anza's analysis: The hot path of database operations calls `key.encode()` thousands of times per second. Each call that returns `Vec<u8>` triggers a heap allocation and subsequent deallocation. By returning fixed-size stack arrays, we eliminate this allocator pressure entirely.

The key layout is deterministic and bounded:
- `ShardedKey<Address>`: 20 bytes (address) + 8 bytes (block number) = 28 bytes
- `StorageShardedKey`: 20 bytes (address) + 32 bytes (storage key) + 8 bytes (block number) = 60 bytes

cc @yongkangc

PR #21220, opened 2026-01-20
perf(db): use Cow::Borrowed in walk_dup to avoid allocation

gakonst:
## Summary

Avoids heap allocation in `walk_dup` by using `Cow::Borrowed` instead of converting the encoded key to `Vec<u8>` with `Cow::Owned`.

For fixed-size key encodings (like `ShardedKey`), the key is already on the stack, so borrowing avoids an unnecessary allocation.

## Changes

- Changed `walk_dup` to keep the encoded key on the stack and use `Cow::Borrowed`
- Removed the intermediate `Vec<u8>` conversion

## Related

Inspired by #21200 and discussion about MDBX cursor allocation optimization.

cc @joshieDo @rakita @mattsse

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/cursor.rs
+++ b/crates/storage/db/src/implementation/mdbx/cursor.rs
@@ -215,27 +215,26 @@ impl<K: TransactionKind, T: DupSort> DbDupCursorRO<T> for Cursor<K, T> {
     ) -> Result<DupWalker<'_, T, Self>, DatabaseError> {
         let start = match (key, subkey) {
             (Some(key), Some(subkey)) => {
-                // encode key and decode it after.
-                let key: Vec<u8> = key.encode().into();
+                let encoded_key = key.encode();
                 self.inner
-                    .get_both_range(key.as_ref(), subkey.encode().as_ref())
+                    .get_both_range(encoded_key.as_ref(), subkey.encode().as_ref())
                     .map_err(|e| DatabaseError::Read(e.into()))?
-                    .map(|val| decoder::<T>((Cow::Owned(key), val)))
+                    .map(|val| decoder::<T>((Cow::Borrowed(encoded_key.as_ref()), val)))
             }
             (Some(key), None) => {
-                let key: Vec<u8> = key.encode().into();
+                let encoded_key = key.encode();
                 self.inner
-                    .set(key.as_ref())
+                    .set(encoded_key.as_ref())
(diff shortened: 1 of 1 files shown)


crates/storage/db-api/src/table.rs
mattsse, line 1: remove these changes, they arent necessary

crates/storage/db-api/src/tables/raw.rs
mattsse, line 1: same here, undo

crates/storage/db/src/implementation/mdbx/cursor.rs
```diff
@@ -215,27 +215,29 @@ impl<K: TransactionKind, T: DupSort> DbDupCursorRO<T> for Cursor<K, T> {
     ) -> Result<DupWalker<'_, T, Self>, DatabaseError> {
         let start = match (key, subkey) {
             (Some(key), Some(subkey)) => {
-                // encode key and decode it after.
-                let key: Vec<u8> = key.encode().into();
+                // Encode key to stack - no heap allocation for fixed-size keys
+                let encoded_key = key.encode();
```
mattsse: this isnt guaranteed so this comment isnt accurate, can keep it as it was

crates/storage/db/src/implementation/mdbx/cursor.rs
```diff
@@ -215,27 +215,29 @@ impl<K: TransactionKind, T: DupSort> DbDupCursorRO<T> for Cursor<K, T> {
...
-                    .set(key.as_ref())
+                    .set(encoded_key.as_ref())
                     .map_err(|e| DatabaseError::Read(e.into()))?
-                    .map(|val| decoder::<T>((Cow::Owned(key), val)))
+                    .map(|val| decoder::<T>((Cow::Borrowed(encoded_key.as_ref()), val)))
             }
             (None, Some(subkey)) => {
                 if let Some((key, _)) = self.first()? {
-                    let key: Vec<u8> = key.encode().into();
+                    // Encode key to stack - no heap allocation for fixed-size keys
```
mattsse: also not needed

mattsse (changes requested):
remove the trait changes (restore into() where we now have into_vec)

mattsse (approved):

DaniPopes (approved):

joshieDo (approved):

PR #21222, opened 2026-01-20
fix: properly save history indices in pipeline

klkvr:
this was already done for accounts in https://github.com/paradigmxyz/reth/pull/21165, so this change effectively wouldn't change anything (because there are no changes for zero slot of zero address anyway)

however, imo still worth doing it just in case we decide to unify those fns later

resolves RETH-154

### DIFF
--- a/crates/stages/stages/src/stages/utils.rs
+++ b/crates/stages/stages/src/stages/utils.rs
@@ -193,7 +193,7 @@ where
     P: Copy + Default + Eq,
 {
     let mut write_cursor = provider.tx_ref().cursor_write::<H>()?;
-    let mut current_partial = P::default();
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
lgtm!

crates/stages/stages/src/stages/utils.rs
```diff
@@ -193,7 +193,7 @@ where
     P: Copy + Default + Eq,
 {
     let mut write_cursor = provider.tx_ref().cursor_write::<H>()?;
-    let mut current_partial = P::default();
```
mattsse, line 196: oh I see the default could cause bad equality checks otherwise

mattsse (approved):

PR #21226, opened 2026-01-20
refactor(engine): move execution logic from metrics to payload_validator

pepyakin:
While studying the codebase I stumbled upon the fact that currently the transaction dispatch happens in the module responsible for metrics and I just couldn't pass by.

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -7,10 +7,10 @@ use crate::tree::{
     payload_processor::{executor::WorkloadExecutor, PayloadProcessor},
     precompile_cache::{CachedPrecompile, CachedPrecompileMetrics, PrecompileCacheMap},
     sparse_trie::StateRootComputeOutcome,
-    EngineApiMetrics, EngineApiTreeState, ExecutionEnv, PayloadHandle, StateProviderBuilder,
-    StateProviderDatabase, TreeConfig,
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -701,31 +707,117 @@ where
...
+
+            // Send the latest receipt to the background task for incremental root computation
+            if let Some(receipt) = executor.receipts().last() {
+                let tx_index = executor.receipts().len() - 1;
+                let _ = receipt_tx.send(IndexedReceipt::new(tx_index, receipt.clone()));
+            }
+
+            enter.record("gas_used", gas_used);
+        }
+        drop(exec_span);
```
Rjected, line 816: won't this be dropped almost immediately even if we don't explicitly drop here? i think we had this previously because we would call `.finish()` and wanted the span to record just the tx exec

Rjected (approved):
I like this and have wanted to remove execution from being in these `metered` functions for a long time, maybe @mattsse could take a look as well

crates/engine/tree/src/tree/payload_validator.rs
mattsse: doesnt hurt to keep imo

mattsse (approved):
this is long overdue and took a few attempts, thanks this is finally  where it should be lol

PR #21227, opened 2026-01-20
feat(engine): add time_between_forkchoice_updated metric

livehardd:
Follow-up to #21159. 

Noticed forkchoiceUpdated was missing the interval tracking that newPayload got. Added the same two histograms - time_between_forkchoice_updated for idle time and forkchoice_updated_interval for the full start-to-start interval.

Had to change update_response_metrics to &mut self to store the timestamps, same as NewPayloadStatusMetrics does.


Rjected (commented):
lgtm but needs fmt

Rjected (approved):
lgtm, thanks!

PR #21228, opened 2026-01-20
feat(engine): add --engine.disable-cache-metrics flag

gakonst:
## Summary

Add a flag to disable cache metrics recording in `SavedCache::update_metrics()`.

## Problem

The `update_metrics()` method calls `total_storage_slots()` which iterates over all storage cache entries to count them. With large cached state (millions of entries), this can take on the order of **50ms** per call, adding unnecessary latency to block processing.

## Solution

Add `--engine.disable-cache-metrics` CLI flag (default: `false`) that skips the expensive metrics recording when enabled.

### Changes

- Added `disable_cache_metrics` field to `TreeConfig` with getter/setter
- Added CLI flag to `EngineArgs` and `DefaultEngineValues`  
- Added `disable_cache_metrics` field to `SavedCache` with early return in `update_metrics()`
- Updated `SavedCache::split()` to preserve the flag when caches are transferred
- Wired the config through `PayloadProcessor` to all `SavedCache` creation sites

## Usage

```bash
reth node --engine.disable-cache-metrics
```

### DIFF
--- a/crates/engine/primitives/src/config.rs
+++ b/crates/engine/primitives/src/config.rs
@@ -137,6 +137,8 @@ pub struct TreeConfig {
     account_worker_count: usize,
     /// Whether to enable V2 storage proofs.
     enable_proof_v2: bool,
+    /// Whether to disable cache metrics recording (can be expensive with large cached state).
+    disable_cache_metrics: bool,
 }
 
 impl Default for TreeConfig {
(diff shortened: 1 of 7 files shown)


mattsse (approved):
this makes sense

would like another review by @shekhirin here

@gakonst please fix failing ci here

shekhirin (approved):
logic LGTM, not a fan of how we add more return arguments to `split()` but should be ok

PR #21230, opened 2026-01-20
fix: supply a real ptr to mdbx_dbi_flags_ex

pepyakin:
See the comment in diff.


yongkangc (approved):
We should prob run this on bench compare argo

PR #21233, opened 2026-01-20
fix: clear `overlay_cache` in `with_extended_hashed_state_overlay`

klkvr:
If we end up falling back to serial state root, cache is very likely to be already populated by the trie workers, meaning that our hashed state extend won't be used


crates/storage/provider/src/providers/state/overlay.rs
```diff
@@ -178,6 +178,10 @@ impl<F> OverlayStateProviderFactory<F> {
                 });
             }
         }
+        // Clear the overlay cache since the overlay source has been modified.
+        // The cache is keyed by db_tip_block and doesn't account for extensions,
+        // so cached entries would return stale overlays without the extended state.
+        self.overlay_cache.write().clear();
```
mediocregopher: nitpick: since this is a `with_` method and returning a new overlay factory I think this would be better being `self.overlay_cache = Default::default()`, leaving the previous overlay instance intact.

And we should do this for the other `with_` methods as well.

mediocregopher (approved):

PR #21234, opened 2026-01-21
fix(rocksdb): flush memtables before dropping

joshieDo:
## Summary

Ensures memtables are flushed before the RocksDB provider is dropped.
## Changes

- Flush WAL and memtables in `Drop` impl before cancelling background work
- Only applies to read-write mode

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -428,7 +428,8 @@ impl Drop for RocksDBProviderInner {
     fn drop(&mut self) {
         match self {
             Self::ReadWrite { db, .. } => {
-                // Flush all memtables if possible. If not, they will be rebuilt from the WAL on restart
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -427,7 +427,16 @@ impl fmt::Debug for RocksDBProviderInner {
 impl Drop for RocksDBProviderInner {
     fn drop(&mut self) {
         match self {
-            Self::ReadWrite { db, .. } => db.cancel_all_background_work(true),
+            Self::ReadWrite { db, .. } => {
+                // Flush all memtables before shutdown to ensure data durability
```
```suggestion
                // Flush all memtables if possible. If not, they will be rebuilt from the WAL on restart
```

```suggestion
                // Flush all memtables if possible. If not, they will be rebuilt from the WAL on restart
```

DaniPopes (approved):

PR #21236, opened 2026-01-21
perf(pruner): batch pruning with raw iterator optimization

gakonst:
Adds batch pruning methods for RocksDB history tables that use raw iterators with seek optimization.

## Changes

- Add `prune_account_history_batch()` and `prune_storage_history_batch()` methods
- Use raw iterators with conditional seek to skip seeks when targets are adjacent in key order
- Add `PrunedIndices` struct to track batch outcomes
- Add `RocksDBRawIterEnum` wrapper for raw iterators in both read-write and read-only modes

## Performance

When targets are sorted and adjacent in RocksDB key order, the iterator naturally lands on the next prefix after finishing the previous one, so seeks are skipped. This reduces iterator construction overhead and improves pruning performance.

## Safety

- Debug assertions enforce sorted input with no duplicates  
- Iterator status checked after traversal to catch RocksDB errors
- Reuses existing `prune_history_shards_inner` logic for correctness

### DIFF
--- a/crates/prune/prune/src/segments/user/account_history.rs
+++ b/crates/prune/prune/src/segments/user/account_history.rs
@@ -238,8 +238,6 @@ impl AccountHistory {
     where
         Provider: DBProvider + StaticFileProviderFactory + ChangeSetReader + RocksDBProviderFactory,
     {
-        use reth_provider::PruneShardOutcome;
-
         // Unlike MDBX path, we don't divide the limit by 2 because RocksDB path only prunes
         // history shards (no separate changeset table to delete from). The changesets are in
         // static files which are deleted separately.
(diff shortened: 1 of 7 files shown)


crates/stages/stages/src/stages/utils.rs
```diff
@@ -7,22 +7,162 @@ use reth_db_api::{
...
 
+/// Trait for writing sharded history indices to the database.
+pub(crate) trait HistoryShardWriter {
+    /// The prefix key type (e.g., Address for accounts, (Address, B256) for storage).
+    type PrefixKey: Copy + Eq;
+    /// The full sharded key type for the table.
+    type TableKey: Key;
+
+    /// Creates a sharded key from prefix and highest block number.
+    fn make_key(prefix: Self::PrefixKey, highest: u64) -> Self::TableKey;
```
yongkangc: `make_key` is needed to construct the database key when flushing shards. For example, in `flush_shards` (line 157):

```rust
let key = H::make_key(prefix, highest);
writer.write_shard(key, &value, append_only)?;
```

The sharded key combines the prefix (`Address` for accounts, `(Address, B256)` for storage) with the highest block number in that shard. This abstraction lets `load_sharded_history` work generically without knowing whether it is building `ShardedKey<Address>` or `StorageShardedKey`.

crates/stages/stages/src/stages/utils.rs
```diff
@@ -7,22 +7,162 @@ use reth_db_api::{
...
+
+    /// Extracts the prefix key from a sharded key.
+    fn key_prefix(key: &Self::TableKey) -> Self::PrefixKey;
```
yongkangc: also why do we need to extract this function out? feels like a shallow function?

yongkangc: `key_prefix` extracts the grouping key from a sharded key so we can detect when we have moved to a new address (or address+slot). Used on line 78:

```rust
let prefix = H::key_prefix(&sharded_key);
if current_prefix != Some(prefix) {
    // flush previous, start new group
}
```

It is shallow, but necessary because the sharded key types differ:
- `ShardedKey<Address>` → extract `.key` (the `Address`)
- `StorageShardedKey` → extract `(address, sharded_key.key)` (address + slot)

Without this abstraction, `load_sharded_history` would need to know the concrete key structure. Happy to hear alternatives if you have ideas.

PR #21225, opened 2026-01-20
refactor: remove static_files.to_settings() and add edge feature to RocksDB flags

gakonst:
## Summary

This PR addresses the review comment from #21191 ([discussion](https://github.com/paradigmxyz/reth/pull/21191#discussion_r2709676364)):

> we can do it in this PR or on a follow-up, but we probably want to remove static_files.to_settings() i think ?
> and we should make sure that the following behaviour introduced here #21208 also applies to rocksdb

## Changes

1. **Remove `StaticFilesArgs::to_settings()`** - This method is now superseded by `NodeConfig::storage_settings()` which handles both static files and RocksDB settings in one place.

2. **Add edge feature behavior to RocksDB flags** - Following the pattern from #21208:
   - Individual RocksDB flags (`tx_hash`, `storages_history`, `account_history`) now default to `true` when the `edge` feature is enabled
   - Changed flags from `Option<bool>` to `bool` with proper defaults using `default_value_t` and `ArgAction::Set`
   - Added consistent doc comments about genesis-initialization-only behavior

3. **Add RocksDbArgs to EnvironmentArgs** - The `EnvironmentArgs` in `cli/commands` now includes `RocksDbArgs` with a `storage_settings()` method for consistent storage settings across all CLI commands.

4. **Update storage_settings() to use edge feature base** - Both `NodeConfig::storage_settings()` and `EnvironmentArgs::storage_settings()` now start from `StorageSettings::edge()` or `StorageSettings::legacy()` based on the feature flag.

## Files Changed

- `crates/node/core/src/args/rocksdb.rs` - Add edge feature defaults, change Option<bool> to bool
- `crates/node/core/src/args/static_files.rs` - Remove `to_settings()` method
- `crates/node/core/src/node_config.rs` - Update `storage_settings()` to use edge feature base
- `crates/cli/commands/src/common.rs` - Add `RocksDbArgs` and `storage_settings()`
- `crates/node/builder/src/launch/common.rs` - Update callers to use `storage_settings()`

PR #21239, opened 2026-01-21
fix(node): make RocksDbArgs respect edge feature defaults like StaticFilesArgs does

HuberyJulianay:
Fixes the inconsistency where StaticFilesArgs properly uses edge feature defaults but RocksDbArgs was always defaulting to false regardless of the feature flag.

Before this, if you compiled with --features edge, you'd still need to manually pass --rocksdb.tx-hash=true to get the actual edge behavior defined in StorageSettings::edge(). Now it just works.


joshieDo:
sorry, handled by https://github.com/paradigmxyz/reth/pull/21225

the rocksdb/sf will be mostly handled by us, so feel free to pick any other issue from tracker, sorry!

PR #21247, opened 2026-01-21
feat(engine-api): add BAL cache for EIP-7928

mattsse:
## Summary

Introduces an in-memory LRU cache for Block Access Lists (BALs) in the Engine API, building on #21204.

## Changes

- Add `BalCache` with internal `Arc` for cheap cloning
- Store BALs keyed by block hash with block number index for range queries  
- Implement `engine_getBALsByHashV1` and `engine_getBALsByRangeV1` to retrieve from cache
- Cache BALs when `newPayload` returns `VALID` status
- Add metrics for cache inserts/hits/misses

## EIP-7928 Compliance

Per [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928), the EL should retain BALs for at least the weak subjectivity period (~3533 epochs ≈ 113,000 blocks). This initial implementation uses a configurable LRU cache (default 1024 entries) as a starting point.

## Next Steps

- Wire up `ExecutionPayload::block_access_list()` to return actual BAL data (currently returns `None`)
- Consider disk-backed storage for full WSP retention
- Add network-based BAL retrieval for missing entries

Resolves RETH-116


crates/rpc/rpc-engine-api/src/bal_cache.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// missing block. This ensures the caller knows the returned BALs correspond to
+    /// blocks `[start, start + len)`.
+    pub fn get_by_range(&self, start: BlockNumber, count: u64) -> Vec<Bytes> {
+        let entries = self.inner.entries.read();
+        let block_index = self.inner.block_index.read();
+
+        let mut result = Vec::new();
+        for block_num in start..start.saturating_add(count) {
+            let Some(hash) = block_index.get(&block_num) else {
+                break;
```
mediocregopher, line 134: Do we want to increment metrics in this method too?

PR #21249, opened 2026-01-21
fix: fixing unit test errors on local mac

cliff0412:
running `make test-unit` on mac ( a powerful machine), some tests failed. this PR fix those test errors


mattsse (commented):
i think we can keep as is

PR #21251, opened 2026-01-21
fix(rocksdb): flush all column families on drop and show SST/memtable sizes

joshieDo:
The previous fix only called `db.flush()` which flushes only the default column family. This left data in TransactionHashNumbers (and other CFs) stranded in the WAL without being compacted to SST files.

## Changes

1. **Flush all CFs on drop**: Iterate TransactionHashNumbers, AccountsHistory, and StoragesHistory column families and flush each one.

2. **Enhanced db stats output**: Show SST size and memtable size separately to make it clear when data is flushed vs still in memory.

## Before (data stuck in WAL, no compression)

| RocksDB Table          | # Entries | Total Size |
|------------------------|-----------|------------|
| TransactionHashNumbers | 911,422   | 60 MiB     |

## After (flushed to SST with LZ4/Zstd compression)

| RocksDB Table          | # Entries | SST Size | Memtable | Total  |
|------------------------|-----------|----------|----------|--------|
| TransactionHashNumbers | 911,422   | 38.2 MiB | 2 KiB    | 38.2 MiB |

RocksDB is now **14% smaller** than MDBX (38.2 vs 44.4 MiB) instead of 35% larger.

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/metrics.rs
+++ b/crates/storage/provider/src/providers/rocksdb/metrics.rs
@@ -6,7 +6,7 @@ use reth_db::Tables;
 use reth_metrics::Metrics;
 use strum::{EnumIter, IntoEnumIterator};
 
-const ROCKSDB_TABLES: &[&str] = &[
+pub(super) const ROCKSDB_TABLES: &[&str] = &[
     Tables::TransactionHashNumbers.name(),
     Tables::StoragesHistory.name(),
     Tables::AccountsHistory.name(),
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -1,4 +1,4 @@
-use super::metrics::{RocksDBMetrics, RocksDBOperation};
+use super::metrics::{RocksDBMetrics, RocksDBOperation, ROCKSDB_TABLES};
 use crate::providers::{compute_history_rank, needs_prev_shard_check, HistoryInfo};
 use alloy_consensus::transaction::TxHashRef;
 use alloy_primitives::{Address, BlockNumber, TxNumber, B256};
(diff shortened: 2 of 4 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -509,8 +515,17 @@ impl Drop for RocksDBProviderInner {
                 if let Err(e) = db.flush_wal(true) {
                     tracing::warn!(target: "storage::rocksdb", ?e, "Failed to flush WAL on drop");
                 }
-                if let Err(e) = db.flush() {
-                    tracing::warn!(target: "storage::rocksdb", ?e, "Failed to flush memtables on drop");
+                // Flush all column families, not just default
```
```suggestion
```

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -509,8 +515,16 @@ impl Drop for RocksDBProviderInner {
                 if let Err(e) = db.flush_wal(true) {
                     tracing::warn!(target: "storage::rocksdb", ?e, "Failed to flush WAL on drop");
                 }
-                if let Err(e) = db.flush() {
-                    tracing::warn!(target: "storage::rocksdb", ?e, "Failed to flush memtables on drop");
+                for cf_name in [
+                    tables::TransactionHashNumbers::NAME,
+                    tables::AccountsHistory::NAME,
+                    tables::StoragesHistory::NAME,
```
joshieDo: isnt there some iter? i think there was some for mdbx and we should be able to recreate it for rocksdb ?

shekhirin (approved):

mattsse (approved):
lgtm

crates/storage/provider/src/providers/rocksdb/metrics.rs
shekhirin: we should be able to derive an `EnumIter` on `Tables`

DaniPopes (approved):

yongkangc (approved):

PR #21254, opened 2026-01-21
fix(net): limit pending pool imports for broadcast transactions

gakonst:
## Summary

Add backpressure to prevent memory exhaustion when receiving broadcast transactions. Previously, only transaction fetches had capacity checks via `has_capacity_for_fetching_pending_hashes()`. This adds the same check for incoming broadcast transactions (`IncomingTransactions` events).

## Changes

- Extract `has_capacity_for_pending_pool_imports()` helper method from the existing capacity check
- Add capacity check before processing `IncomingTransactions` (broadcast transactions)
- When pending pool imports reach capacity, broadcast transactions are now dropped with a trace log

## Context

Related to sigp audit issue flagged in [CHAIN-489](https://linear.app/tempoxyz/issue/CHAIN-489) - unbounded RLP decoding can exhaust validator memory. While the RLP decoding limit is a separate concern, this ensures the transaction processing pipeline has proper backpressure at the pool import stage.

The `pool_imports` FuturesUnordered queue can grow unbounded when broadcast transactions arrive faster than the pool can process them. With this change, the capacity limit (`DEFAULT_MAX_COUNT_PENDING_POOL_IMPORTS = 4096`) now applies to both:
1. Transactions fetched via GetPooledTransactions (already had check)
2. Transactions received via broadcast (new check added)

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -1347,6 +1347,11 @@ where
             return
         }
 
+        // Early return if we don't have capacity for any imports
+        if !self.has_capacity_for_pending_pool_imports() {
+            return
+        }
+
         let Some(peer) = self.peers.get_mut(&peer_id) else { return };
         let client_version = peer.client_version.clone();
         let mut transactions = transactions.0;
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
can we add a new metric for how many txs we dropped with this

mattsse (approved):
makes sense, this should be sufficient backpressure for broadcast spam, if we drop them then this should only delay ingress because chances are we'll receive the same thing from other peers

crates/net/network/src/transactions/mod.rs
```diff
@@ -1387,6 +1405,18 @@ where
...
+        // that won't be imported anyway.
+        let capacity = self.remaining_pool_import_capacity();
+        if transactions.len() > capacity {
+            let skipped = transactions.len() - capacity;
+            transactions.truncate(capacity);
+            self.metrics
+                .skipped_transactions_pending_pool_imports_at_capacity
+                .increment(skipped as u64);
+            trace!(target: "net::tx", skipped, capacity, "Truncated transactions batch to capacity");
+        }
```
mattsse, line 1418: makes sense

mattsse (approved):

PR #21255, opened 2026-01-21
fix: fix payload file filter prefix in replay-payloads

livehardd:
File filter used `payload_` prefix but parser expected `payload_block_`, causing silent failures. Changed filter to match actual file format.

### DIFF
--- a/bin/reth-bench/src/bench/replay_payloads.rs
+++ b/bin/reth-bench/src/bench/replay_payloads.rs
@@ -180,7 +180,7 @@ impl Command {
             .filter_map(|e| e.ok())
             .filter(|e| {
                 e.path().extension().and_then(|s| s.to_str()) == Some("json") &&
-                    e.file_name().to_string_lossy().starts_with("payload_")
+                    e.file_name().to_string_lossy().starts_with("payload_block_")
             })
             .collect();
 


bin/reth-bench/src/bench/replay_payloads.rs
```diff
@@ -180,7 +180,7 @@ impl Command {
             .filter_map(|e| e.ok())
             .filter(|e| {
                 e.path().extension().and_then(|s| s.to_str()) == Some("json") &&
-                    e.file_name().to_string_lossy().starts_with("payload_")
+                    e.file_name().to_string_lossy().starts_with("payload_block_")
```
mattsse, line 183: payload_ also satifies payload_block

so idk if this is necessary

mattsse (approved):

shekhirin (approved):

bin/reth-bench/src/bench/replay_payloads.rs
livehardd: Right, but `payload_` also matches unwanted files like `payload_test.json`. They'd pass the filter but fail silently at the strip_prefix. Making it stricter avoids the mismatch and matches load_gas_ramp_payloads.

PR #21256, opened 2026-01-21
fix: use full storage settings in init_genesis

HuberyJulianay:
init_genesis was only using `static_files.to_settings()` which ignored all the rocksdb CLI flags like `--rocksdb.tx-hash`, causing user config to be ignored during genesis initialization. switched to` storage_settings()` which properly combines both static files and rocksdb settings.


Rjected (approved):
lgtm, thanks!

PR #21259, opened 2026-01-21
feat(reth-bench): add --target-gas-limit option to gas-limit-ramp

gakonst:
## Summary

Add `--target-gas-limit` as an alternative to `--blocks` for the `gas-limit-ramp` benchmark. When specified, the benchmark will generate blocks until the gas limit reaches or exceeds the target value, rather than a fixed number of blocks.

## Changes

- Made `--blocks` optional
- Added `--target-gas-limit` option (mutually exclusive with `--blocks`)
- Either one or the other must be specified
- Added `RampMode` enum to cleanly handle both modes

## Usage

```bash
# By block count (existing behavior)
reth-bench gas-limit-ramp --blocks 100 --engine-rpc-url ... --jwt-secret ... --output ...

# By target gas limit (new)
reth-bench gas-limit-ramp --target-gas-limit 100000000 --engine-rpc-url ... --jwt-secret ... --output ...
```

### DIFF
--- a/bin/reth-bench/src/bench/gas_limit_ramp.rs
+++ b/bin/reth-bench/src/bench/gas_limit_ramp.rs
@@ -25,9 +25,14 @@ use tracing::info;
 /// `reth benchmark gas-limit-ramp` command.
 #[derive(Debug, Parser)]
 pub struct Command {
-    /// Number of blocks to generate.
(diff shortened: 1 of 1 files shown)


shekhirin (approved):

Rjected (approved):
makes sense to me

PR #21275, opened 2026-01-21
fix(prune): avoid panic in tx lookup

shekhirin:
We recently started setting `deleted_entries_limit` to `usize::MAX`, this now will overflow `u64` when calculating the upper bound of how many entries we can prune.

This PR fixes it by doing saturating addition.


klkvr (approved):
ah makes sense

mattsse (approved):

PR #21279, opened 2026-01-21
perf(db): use ArrayVec for StoredNibbles key encoding

mattsse:
Eliminates heap allocation when encoding `StoredNibbles` keys for trie tables (`AccountsTrie`). Nibbles have a max length of 64, so `ArrayVec<u8, 64>` is sufficient and stack-allocated.

## Changes

- Add `IntoVec` trait for converting encoded types to `Vec<u8>`
- Implement `IntoVec` for `Vec<u8>` (no-op), `[u8; N]`, and `ArrayVec<u8, N>`
- Change `StoredNibbles::Encoded` from `Vec<u8>` to `ArrayVec<u8, 64>`
- Update error paths to use `.into_vec()` instead of `.into()`

This follows the same pattern already used in the `Compact` impl:

https://github.com/paradigmxyz/reth/blob/a62f89b0e06a1dd5f2cc1fe1907bb8f8cd14a124/crates/trie/common/src/nibbles.rs#L25-L31

```rust
#[cfg(any(test, feature = "reth-codec"))]
impl reth_codecs::Compact for StoredNibbles {
    fn to_compact<B>(&self, buf: &mut B) -> usize
    where
        B: bytes::BufMut + AsMut<[u8]>,
    {
        let bytes = self.0.iter().collect::<arrayvec::ArrayVec<u8, 64>>();
```

### DIFF
--- a/crates/storage/db-api/src/models/mod.rs
+++ b/crates/storage/db-api/src/models/mod.rs
@@ -126,13 +126,10 @@ impl Decode for String {
 }
 
 impl Encode for StoredNibbles {
-    type Encoded = Vec<u8>;
+    type Encoded = arrayvec::ArrayVec<u8, 64>;
 
-    // Delegate to the Compact implementation
     fn encode(self) -> Self::Encoded {
-        // NOTE: This used to be `to_compact`, but all it does is append the bytes to the buffer,
-        // so we can just use the implementation of `Into<Vec<u8>>` to reuse the buffer.
-        self.0.to_vec()
+        self.0.iter().collect()
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8099,6 +8099,7 @@ dependencies = [
  "alloy-genesis",
  "alloy-primitives",
  "arbitrary",
+ "arrayvec",
  "bytes",
  "derive_more",
  "metrics",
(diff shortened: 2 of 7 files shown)


crates/storage/db-api/src/models/mod.rs
```diff
@@ -126,13 +126,10 @@ impl Decode for String {
 }
 
 impl Encode for StoredNibbles {
-    type Encoded = Vec<u8>;
+    type Encoded = arrayvec::ArrayVec<u8, 64>;
 
-    // Delegate to the Compact implementation
     fn encode(self) -> Self::Encoded {
-        // NOTE: This used to be `to_compact`, but all it does is append the bytes to the buffer,
-        // so we can just use the implementation of `Into<Vec<u8>>` to reuse the buffer.
```
mattsse, line 134: this comment didnt make any sense to me

DaniPopes (approved):

crates/storage/db-api/src/models/mod.rs
DaniPopes: thats fine
https://github.com/paradigmxyz/reth/commit/68f0b0a08f58723a8897e5e503863620003a61c6

PR #21297, opened 2026-01-22
perf(prune): sort tx hashes for efficient TransactionLookup pruning

gakonst:
## Summary

Sort transaction hashes before passing to `prune_table_with_iterator`.

## Problem

The `TransactionHashNumbers` table is a B+tree keyed by `TxHash` (B256). Previously, hashes were passed in transaction number order (from `transactions_by_tx_range`), causing random O(log n) seeks per hash through the tree.

## Solution

Sort hashes before iteration. This enables the cursor to advance sequentially through the B+tree, significantly reducing seek overhead during pruning.

The `prune_table_with_iterator` docstring already expects a "pre-sorted key iterator" - this change fulfills that contract.


joshieDo (approved):

yongkangc (approved):
makes sense

PR #21299, opened 2026-01-22
fix(cli): clear rocksdb tables in drop-stage command

joshieDo:
## Summary

When dropping stages with `reth drop-stage`, the command now also clears the corresponding tables from RocksDB if the storage settings indicate they are stored there.

Previously, the drop-stage command only cleared MDBX tables, leaving stale data in RocksDB when `--rocksdb.*` flags were used.

## Changes

**TxLookup stage:**
- Clears `TransactionHashNumbers` from RocksDB when `transaction_hash_numbers_in_rocksdb` is enabled

**AccountHistory / StorageHistory stages:**
- Clears `AccountsHistory` from RocksDB when `account_history_in_rocksdb` is enabled
- Clears `StoragesHistory` from RocksDB when `storages_history_in_rocksdb` is enabled

**Stub:**
- Added a `clear<T>()` stub method to the RocksDBProvider stub for non-Unix platforms

## Testing

`cargo clippy -p reth-cli-commands -- -D warnings`

### DIFF
--- a/crates/cli/commands/src/stage/drop.rs
+++ b/crates/cli/commands/src/stage/drop.rs
@@ -15,7 +15,8 @@ use reth_db_common::{
 use reth_node_api::{HeaderTy, ReceiptTy, TxTy};
 use reth_node_core::args::StageEnum;
 use reth_provider::{
-    DBProvider, DatabaseProviderFactory, StaticFileProviderFactory, StaticFileWriter,
+    DBProvider, DatabaseProviderFactory, RocksDBProviderFactory, StaticFileProviderFactory,
+    StaticFileWriter, StorageSettingsCache,
 };
 use reth_prune::PruneSegment;
 use reth_stages::StageId;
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/stage/drop.rs
```diff
@@ -178,6 +179,14 @@ impl<C: ChainSpecParser> Command<C> {
...
             StageEnum::TxLookup => {
                 tx.clear::<tables::TransactionHashNumbers>()?;
+
+                // Also clear from RocksDB if transaction hash numbers are stored there
+                if provider_rw.cached_storage_settings().transaction_hash_numbers_in_rocksdb {
+                    tool.provider_factory
+                        .rocksdb_provider()
+                        .clear::<tables::TransactionHashNumbers>()?;
+                }
+
```
joshieDo, line 200:
```suggestion
                if provider_rw.cached_storage_settings().transaction_hash_numbers_in_rocksdb {
                    tool.provider_factory
                        .rocksdb_provider()
                        .clear::<tables::TransactionHashNumbers>()?;
                } else {
                     tx.clear::<tables::TransactionHashNumbers>()?;
                }
```

yongkangc (approved):
ty for this, think it makes sense. 

tho wondering if we want to refactor into a generic type funciton in the future for all 3 indices?

PR #21300, opened 2026-01-22
fix: remove redundant clone

andrewshab3:
Removes a redundant .clone() when passing `ctx.task_executor` to `repair_trie::Command::execute`


yongkangc (approved):
looks right, thanks!

PR #21310, opened 2026-01-22
perf(rocksdb): disable compression and bloom filters for TransactionHashNumbers CF

joshieDo:
## Summary

This PR optimizes the `TransactionHashNumbers` column family in RocksDB by:

1. **Disabling compression** - Keys are incompressible 32-byte hashes, values are varint-encoded `u64` (a few bytes). Compression wastes CPU cycles for zero space savings.

2. **Disabling bloom filters** - Every lookup expects a hit (we're looking up known transaction hashes). Bloom filters only help when checking for non-existent keys, so they provide no benefit and waste memory.

## Changes

- Added `tx_hash_numbers_column_family_options()` method with specialized CF options
- Modified `build()` to use the specialized options for `TransactionHashNumbers`

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -201,6 +201,32 @@ impl RocksDBBuilder {
         cf_options
     }
 
+    /// Creates optimized column family options for `TransactionHashNumbers`.
+    ///
+    /// This table stores `B256 -> TxNumber` mappings where:
+    /// - Keys are incompressible 32-byte hashes (compression wastes CPU for zero benefit)
+    /// - Values are varint-encoded `u64` (a few bytes - too small to benefit from compression)
+    /// - Every lookup expects a hit (bloom filters only help when checking non-existent keys)
+    fn tx_hash_numbers_column_family_options(cache: &Cache) -> Options {
+        let mut table_options = BlockBasedOptions::default();
+        table_options.set_block_size(DEFAULT_BLOCK_SIZE);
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -201,6 +201,33 @@ impl RocksDBBuilder {
         cf_options
     }
 
+    /// Creates optimized column family options for `TransactionHashNumbers`.
+    ///
+    /// This table stores `B256 -> TxNumber` mappings where:
+    /// - Keys are incompressible 32-byte hashes (compression wastes CPU for zero benefit)
+    /// - Values are SCALE-encoded `u64` (variable-length, typically a few bytes - too small to
+    ///   benefit from compression)
+    /// - Every lookup expects a hit (bloom filters only help when checking non-existent keys)
```
```suggestion
```

yongkangc (approved):

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -262,10 +288,12 @@ impl RocksDBBuilder {
...
-                ColumnFamilyDescriptor::new(
-                    name.clone(),
-                    Self::default_column_family_options(&self.block_cache),
-                )
+                let cf_options = if name == tables::TransactionHashNumbers::NAME {
+                    Self::tx_hash_numbers_column_family_options(&self.block_cache)
+                } else {
+                    Self::default_column_family_options(&self.block_cache)
+                };
+                ColumnFamilyDescriptor::new(name.clone(), cf_options)
```
shekhirin, line 296: hmm, can we move this to table trait level?

PR #21315, opened 2026-01-22
fix(cli): handle invalid hex in `db list --search`

letmehateu:
Replaces a unwrap() in `reth db list` hex search parsing with proper error handling.
Invalid `--search 0x...` values now return a clear eyre error message instead of crashing.

### DIFF
--- a/crates/cli/commands/src/db/list.rs
+++ b/crates/cli/commands/src/db/list.rs
@@ -61,19 +61,21 @@ impl Command {
     }
 
     /// Generate [`ListFilter`] from command.
-    pub fn list_filter(&self) -> ListFilter {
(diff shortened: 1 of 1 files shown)


gakonst (commented):
LGTM! The panic is real and the fix is correct.

Minor suggestion for the error message - since the user already provided the `0x` prefix, the issue is the content being invalid, not the format:

```suggestion
                    hex::decode(search).wrap_err(
                        "Invalid hex content after 0x prefix in --search (expected valid hex like 0xdeadbeef)",
                    )?
```

joshieDo (approved):

mattsse (approved):

PR #21214, opened 2026-01-20
feat(trie): add V2 account proof computation and refactor proof types

mediocregopher:
Refactor proof-related types to support both legacy and V2 proof systems:

• Change ProofResult from struct to enum with Legacy and V2 variants
• Change AccountMultiproofInput from struct to enum with Legacy and V2 variants
• Add compute_v2_account_multiproof method for V2 proof computation
• Add dispatch_v2_storage_proofs for V2 storage proof dispatching
• Add v2_enabled flag to AccountProofWorker to control proof version
• Move cursor_metrics from ParallelTrieTracker to local scope in proof functions

V2 proofs use AsyncAccountValueEncoder for parallel storage root computation during account proof generation. The legacy path retains stats tracking for metrics; V2 currently omits stats.

Closes RETH-167

Closes #19514

PR #21316, opened 2026-01-22
feat(trie): add V2 account proof computation and refactor proof types (reapply)

mediocregopher:
Refactor proof-related types to support both legacy and V2 proof systems:

• Change ProofResult from struct to enum with Legacy and V2 variants
• Change AccountMultiproofInput from struct to enum with Legacy and V2 variants
• Add compute_v2_account_multiproof method for V2 proof computation
• Add dispatch_v2_storage_proofs for V2 storage proof dispatching
• Add v2_enabled flag to AccountProofWorker to control proof version
• Move cursor_metrics from ParallelTrieTracker to local scope in proof functions

V2 proofs use AsyncAccountValueEncoder for parallel storage root computation during account proof generation. The legacy path retains stats tracking for metrics; V2 currently omits stats.

Reapplied after #21214 was reverted


crates/trie/parallel/src/proof_task.rs
mediocregopher, line 1677: This code block was accidentally deleted in the original PR 🤦

mattsse (approved):

PR #21321, opened 2026-01-22
refactor: refactor get_idle_peer_for to use Iterator::find

letmehateu:
Clippy reports this loop as a manual Iterator::find implementation (clippy::manual_find). Using find keeps behavior identical and removes the lint failure.


mattsse (approved):
this is nicer indeed

PR #21324, opened 2026-01-22
fix(txpool): bound transaction batcher channel to prevent DoS

mattsse:
## Summary

This changes the unbounded channel in the transaction pool batcher to a bounded channel with a default capacity of 2048. This prevents potential denial-of-service attacks where malicious users could blast large transactions and quickly cancel requests to bypass request-level rate limiting, causing unbounded memory growth.

## Changes

- Replace `mpsc::unbounded_channel` with `mpsc::channel(2048)`
- Update RPC layer to use bounded `Sender` and `await` on send
- Fix graceful shutdown when channel is closed (processor now terminates cleanly when all senders drop)
- Update tests to use `try_send` for bounded channel compatibility

## Motivation

The unbounded channel allows unbounded memory growth under sustained load. An attacker could:
1. Blast many large transactions via RPC
2. Quickly cancel the requests
3. Bypass request-level rate limiting since requests complete/cancel quickly
4. Cause unbounded memory allocation in the batcher queue

With a bounded channel:
- Backpressure is applied when the queue reaches capacity
- Memory usage is bounded to ~2048 pending requests
- The RPC layer will block (await) when the queue is full, naturally rate-limiting requests

## Testing

- All existing batcher tests pass
- Verified with clippy and fmt

### DIFF
--- a/crates/transaction-pool/src/batcher.rs
+++ b/crates/transaction-pool/src/batcher.rs
@@ -14,7 +14,58 @@ use std::{
 };
 use tokio::sync::{mpsc, oneshot};
 
-/// A single batch transaction request
+/// Default capacity for the transaction batch request channel.
+///
+/// This bounds the number of pending batch requests to prevent unbounded memory growth
+/// under denial-of-service attacks where malicious users blast large transactions and quickly
+/// cancel requests. The value of 2048 provides enough headroom for burst scenarios while ensuring
+/// backpressure is applied before memory usage becomes problematic.
+pub const DEFAULT_BATCH_CHANNEL_CAPACITY: usize = 2048;
+
+/// Default maximum batch size for processing transactions.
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/batcher.rs
```diff
@@ -100,7 +156,12 @@ where
 
         loop {
             // Drain all available requests from the receiver
-            ready!(this.request_rx.poll_recv_many(cx, this.buf, *this.max_batch_size));
+            let n = ready!(this.request_rx.poll_recv_many(cx, this.buf, *this.max_batch_size));
+
+            // Channel closed and no remaining items - shutdown gracefully
+            if n == 0 && this.buf.is_empty() {
+                return Poll::Ready(());
+            }
```
mattsse, line 164: this is equivalent but nicer

PR #21325, opened 2026-01-22
perf: avoid creating RocksDB transactions for legacy MDBX-only nodes

### LINKED ISSUE #21144: Avoid creating RocksDB transactions for legacy MDBX-only nodes

yongkangc:
**Problem:** On legacy nodes where `StorageSettings` has all history flags set to `false`, `with_rocksdb_tx` was still creating a RocksDB transaction that was never used.

**Solution:** Check `any_in_rocksdb()` before creating the transaction. Changed `RocksTxRefArg` to `Option<&RocksTx>` so callers receive `None` when RocksDB isn't needed.

**Changes:**
- `with_rocksdb_tx`: Only creates transaction when `any_in_rocksdb()` returns true, passes `None` otherwise
- `RocksTxRefArg`: Now `Option<&RocksTx>` (or `Option<()>` when feature disabled)
- `EitherReader::new_*`: Unwrap Option, return `ProviderError::InvalidStorageSettings` on mismatch
- Added `StorageSettingsCache` as method-level bound on `with_rocksdb_tx` (no extra impls needed)

**Impact:** Legacy MDBX-only nodes skip RocksDB transaction creation entirely on history lookups.

Closes #21144

### DIFF
--- a/crates/storage/provider/src/providers/blockchain_provider.rs
+++ b/crates/storage/provider/src/providers/blockchain_provider.rs
@@ -20,18 +20,14 @@ use reth_chain_state::{
     MemoryOverlayStateProvider, PersistedBlockNotifications, PersistedBlockSubscriptions,
 };
 use reth_chainspec::ChainInfo;
-use reth_db_api::models::{
-    AccountBeforeTx, BlockNumberAddress, StorageSettings, StoredBlockBodyIndices,
-};
+use reth_db_api::models::{AccountBeforeTx, BlockNumberAddress, StoredBlockBodyIndices};
 use reth_execution_types::ExecutionOutcome;
 use reth_node_types::{BlockTy, HeaderTy, NodeTypesWithDB, ReceiptTy, TxTy};
 use reth_primitives_traits::{Account, RecoveredBlock, SealedHeader, StorageEntry};
 use reth_prune_types::{PruneCheckpoint, PruneSegment};
 use reth_stages_types::{StageCheckpoint, StageId};
 use reth_static_file_types::StaticFileSegment;
-use reth_storage_api::{
-    BlockBodyIndicesProvider, NodePrimitivesProvider, StorageChangeSetReader, StorageSettingsCache,
-};
+use reth_storage_api::{BlockBodyIndicesProvider, NodePrimitivesProvider, StorageChangeSetReader};
 use reth_storage_errors::provider::ProviderResult;
 use reth_trie::{HashedPostState, KeccakKeyHasher};
 use revm_database::BundleState;
@@ -178,16 +174,6 @@ impl<N: ProviderNodeTypes> StaticFileProviderFactory for BlockchainProvider<N> {
         segment: StaticFileSegment,
     ) -> ProviderResult<StaticFileProviderRWRefMut<'_, Self::Primitives>> {
         self.database.get_static_file_writer(block, segment)
-    }
-}
-
-impl<N: ProviderNodeTypes> StorageSettingsCache for BlockchainProvider<N> {
-    fn cached_storage_settings(&self) -> StorageSettings {
-        self.database.cached_storage_settings()
-    }
-
-    fn set_storage_settings_cache(&self, settings: StorageSettings) {
-        self.database.set_storage_settings_cache(settings)
     }
 }
 
--- a/crates/storage/provider/src/providers/blockchain_provider.rs
+++ b/crates/storage/provider/src/providers/blockchain_provider.rs
@@ -20,14 +20,18 @@ use reth_chain_state::{
     MemoryOverlayStateProvider, PersistedBlockNotifications, PersistedBlockSubscriptions,
 };
 use reth_chainspec::ChainInfo;
-use reth_db_api::models::{AccountBeforeTx, BlockNumberAddress, StoredBlockBodyIndices};
+use reth_db_api::models::{
+    AccountBeforeTx, BlockNumberAddress, StorageSettings, StoredBlockBodyIndices,
+};
 use reth_execution_types::ExecutionOutcome;
 use reth_node_types::{BlockTy, HeaderTy, NodeTypesWithDB, ReceiptTy, TxTy};
 use reth_primitives_traits::{Account, RecoveredBlock, SealedHeader, StorageEntry};
 use reth_prune_types::{PruneCheckpoint, PruneSegment};
 use reth_stages_types::{StageCheckpoint, StageId};
 use reth_static_file_types::StaticFileSegment;
-use reth_storage_api::{BlockBodyIndicesProvider, NodePrimitivesProvider, StorageChangeSetReader};
+use reth_storage_api::{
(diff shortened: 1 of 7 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -314,7 +314,7 @@ impl<TX, N: NodeTypes> StaticFileProviderFactory for DatabaseProvider<TX, N> {
     }
 }
 
-impl<TX, N: NodeTypes> RocksDBProviderFactory for DatabaseProvider<TX, N> {
+impl<TX: Send, N: NodeTypes> RocksDBProviderFactory for DatabaseProvider<TX, N> {
```
yongkangc: do we really need send here?

crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -177,6 +181,16 @@ impl<N: ProviderNodeTypes> StaticFileProviderFactory for BlockchainProvider<N> {
...
 
+impl<N: ProviderNodeTypes> StorageSettingsCache for BlockchainProvider<N> {
+    fn cached_storage_settings(&self) -> StorageSettings {
+        self.database.cached_storage_settings()
+    }
+
+    fn set_storage_settings_cache(&self, settings: StorageSettings) {
+        self.database.set_storage_settings_cache(settings)
+    }
+}
```
yongkangc: do we really need these here?

crates/storage/storage-api/src/noop.rs
```diff
@@ -693,3 +695,16 @@ impl<ChainSpec: Send + Sync, N: NodePrimitives> DatabaseProviderFactory
...
+    for NoopProvider<ChainSpec, N>
+{
+    fn cached_storage_settings(&self) -> StorageSettings {
+        StorageSettings::legacy()
+    }
+
+    fn set_storage_settings_cache(&self, _settings: StorageSettings) {
+        // No-op for NoopProvider
+    }
+}
```
yongkangc: why is this a noop?

gakonst (commented):
## Review Summary

Overall this PR correctly addresses #21144 - legacy MDBX-only nodes will now skip RocksDB transaction creation entirely. The approach is sound.

### ✅ What looks good
- Core logic is correct: `with_rocksdb_tx` checks `any_in_rocksdb()` before creating tx
- `Option<&RocksTx>` is the cleanest way to propagate "no tx needed"
- Safety net in `EitherReader::new_*` catches config/tx mismatches
- Tests verify the key paths

### 🔧 Suggested improvements

**1. Test flakiness - use tempdir instead of `/tmp/test`**

The test uses a hardcoded path which can cause issues in parallel test runs:
```rust
// Current
RocksDBProvider::new("/tmp/test").unwrap()

// Suggested - use tempfile::TempDir
let temp_dir = tempfile::TempDir::new().unwrap();
RocksDBProvider::new(temp_dir.path()).unwrap()
```

**2. Add test for the error path**

Missing coverage for when settings require RocksDB but tx is `None`:
```rust
#[test]
#[cfg(all(unix, feature = "rocksdb"))]
fn test_settings_mismatch_returns_error() {
    // Configure with account_history_in_rocksdb=true
    // Call EitherReader::new_accounts_history with None
    // Assert InvalidStorageSettings error
}
```

**3. Consider cfg-gating the test module**

`RocksDBProvider::new` may not exist in non-rocksdb builds. Consider:
```rust
#[cfg(all(test, unix, feature = "rocksdb"))]
mod tests { ... }
```
Or split into rocksdb-only tests that actually open RocksDB, and feature-agnostic tests that only verify `None` is passed.

**4. Minor: PR description mentions `RocksDBTxRequired` error but code uses `InvalidStorageSettings`**

Just a docs alignment issue - the code is correct.

### Performance ✓
The optimization achieves its goal. The extra overhead is just a cached settings read + boolean check, which is negligible compared to avoided RocksDB tx creation.

### Correctness concerns (minor)
- Ensure `any_in_rocksdb()` is updated whenever new RocksDB-backed flags are added in the future
- `StorageSettingsCache` should be stable for the lifetime of an operation to avoid cache mismatches

LGTM pending the test improvements!

crates/storage/provider/src/providers/database/provider.rs
yongkangc: Removed. It was required because `StorageSettingsCache` trait requires `Send`, but the bound only needs to be on the impl where the method is called, not on the `RocksDBProviderFactory` impl.

crates/storage/provider/src/providers/blockchain_provider.rs
yongkangc: Removed. The trait bounds on `with_rocksdb_tx` are method-level bounds (`Self: StorageSettingsCache`), so `BlockchainProvider` doesn't need to impl `StorageSettingsCache` just to implement `RocksDBProviderFactory` - only callers of `with_rocksdb_tx` need that impl.

crates/storage/storage-api/src/noop.rs
yongkangc: Removed. Same reason as above - method-level bounds mean `NoopProvider` doesn't need `StorageSettingsCache` just to implement `RocksDBProviderFactory`.

crates/storage/provider/src/test_utils/noop.rs
```diff
@@ -31,7 +31,5 @@ impl<C: Send + Sync, N: NodePrimitives> RocksDBProviderFactory for NoopProvider<
     }
 
     #[cfg(all(unix, feature = "rocksdb"))]
-    fn set_pending_rocksdb_batch(&self, _batch: rocksdb::WriteBatchWithTransaction<true>) {
-        // No-op for NoopProvider
-    }
+    fn set_pending_rocksdb_batch(&self, _batch: rocksdb::WriteBatchWithTransaction<true>) {}
```
yongkangc:
```suggestion
    fn set_pending_rocksdb_batch(&self, _batch: rocksdb::WriteBatchWithTransaction<true>) {
        // No-op for NoopProvider
    }
```

pepyakin (approved):

PR #21331, opened 2026-01-22
feat(prune): prune rocksdb account and storage history indices

yongkangc:
## Summary

Add `AccountsHistoryPruner` and `StoragesHistoryPruner` segments for pruning RocksDB history indices (`AccountsHistory`, `StoragesHistory`).

Uses changeset-driven pruning: collects affected keys from changesets being pruned, then prunes only those keys from RocksDB history shards. This avoids scanning all RocksDB history entries.

Closes #21292

## Changes

- `prune_account_history_to` and `prune_storage_history_to` on `RocksDBBatch`
- `PruneShardOutcome` enum (`Deleted`, `Updated`, `Unchanged`)
- `AccountsHistoryPruner` and `StoragesHistoryPruner` segments in `reth-prune`
- Support pruning when changesets are stored in static files (deletes jars below pruned block)
- Feature-gated: `#[cfg(all(unix, feature = "rocksdb"))]`
- 21 tests: single/multi-shard scenarios, sentinel promotion, MDBX equivalence

### DIFF
--- a/crates/prune/prune/Cargo.toml
+++ b/crates/prune/prune/Cargo.toml
@@ -42,11 +42,15 @@ rayon.workspace = true
 tokio.workspace = true
 rustc-hash.workspace = true
 
+[features]
+rocksdb = ["reth-provider/rocksdb"]
+
 [dev-dependencies]
 # reth
 reth-db = { workspace = true, features = ["test-utils"] }
-reth-stages = { workspace = true, features = ["test-utils"] }
+reth-stages = { workspace = true, features = ["test-utils", "rocksdb"] }
 reth-primitives-traits = { workspace = true, features = ["arbitrary"] }
+reth-storage-api.workspace = true
 reth-testing-utils.workspace = true
 reth-tracing.workspace = true
 
--- a/crates/prune/prune/src/segments/user/account_history.rs
+++ b/crates/prune/prune/src/segments/user/account_history.rs
@@ -10,20 +10,20 @@ use alloy_primitives::BlockNumber;
 use reth_db_api::{models::ShardedKey, tables, transaction::DbTxMut};
 use reth_provider::{
     changeset_walker::StaticFileAccountChangesetWalker, DBProvider, EitherWriter,
-    StaticFileProviderFactory, StorageSettingsCache,
+    RocksDBProviderFactory, StaticFileProviderFactory,
 };
 use reth_prune_types::{
     PruneMode, PrunePurpose, PruneSegment, SegmentOutput, SegmentOutputCheckpoint,
 };
 use reth_static_file_types::StaticFileSegment;
-use reth_storage_api::ChangeSetReader;
+use reth_storage_api::{ChangeSetReader, StorageSettingsCache};
 use rustc_hash::FxHashMap;
 use tracing::{instrument, trace};
 
 /// Number of account history tables to prune in one step.
 ///
-/// Account History consists of two tables: [`tables::AccountChangeSets`] and
-/// [`tables::AccountsHistory`]. We want to prune them to the same block number.
+/// Account History consists of two tables: [`tables::AccountChangeSets`] (either in database or
+/// static files) and [`tables::AccountsHistory`]. We want to prune them to the same block number.
 const ACCOUNT_HISTORY_TABLES_TO_PRUNE: usize = 2;
 
 #[derive(Debug)]
@@ -42,7 +42,8 @@ where
     Provider: DBProvider<Tx: DbTxMut>
         + StaticFileProviderFactory
         + StorageSettingsCache
-        + ChangeSetReader,
+        + ChangeSetReader
+        + RocksDBProviderFactory,
 {
     fn segment(&self) -> PruneSegment {
         PruneSegment::AccountHistory
@@ -67,7 +68,13 @@ where
         };
         let range_end = *range.end();
 
-        // Check where account changesets are stored
+        // Check where account history indices are stored
+        #[cfg(all(unix, feature = "rocksdb"))]
+        if provider.cached_storage_settings().account_history_in_rocksdb {
+            return self.prune_rocksdb(provider, input, range, range_end);
+        }
+
+        // Check where account changesets are stored (MDBX path)
         if EitherWriter::account_changesets_destination(provider).is_static_file() {
             self.prune_static_files(provider, input, range, range_end)
         } else {
@@ -94,6 +101,8 @@ impl AccountHistory {
             input.limiter
         };
 
+        // The limiter may already be exhausted from a previous segment in the same prune run.
+        // Early exit avoids unnecessary iteration when no budget remains.
         if limiter.is_limit_reached() {
             return Ok(SegmentOutput::not_done(
                 limiter.interrupt_reason(),
@@ -101,11 +110,14 @@ impl AccountHistory {
             ))
         }
 
-        // The size of this map it's limited by `prune_delete_limit * blocks_since_last_run /
-        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with the current defaults it's usually `3500 * 5 /
-        // 2`, so 8750 entries. Each entry is `160 bit + 64 bit`, so the total size should be up to
-        // ~0.25MB + some hashmap overhead. `blocks_since_last_run` is additionally limited by the
-        // `max_reorg_depth`, so no OOM is expected here.
+        // Deleted account changeset keys (account addresses) with the highest block number deleted
+        // for that key.
+        //
+        // The size of this map is limited by `prune_delete_limit * blocks_since_last_run /
+        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with current default it's usually `3500 * 5
+        // / 2`, so 8750 entries. Each entry is `160 bit + 64 bit`, so the total
+        // size should be up to ~0.25MB + some hashmap overhead. `blocks_since_last_run` is
+        // additionally limited by the `max_reorg_depth`, so no OOM is expected here.
         let mut highest_deleted_accounts = FxHashMap::default();
         let mut last_changeset_pruned_block = None;
         let mut pruned_changesets = 0;
@@ -124,8 +136,8 @@ impl AccountHistory {
             limiter.increment_deleted_entries_count();
         }
 
-        // Delete static file jars below the pruned block
-        if let Some(last_block) = last_changeset_pruned_block {
+        // Delete static file jars only when fully processed
+        if done && let Some(last_block) = last_changeset_pruned_block {
             provider
                 .static_file_provider()
                 .delete_segment_below_block(StaticFileSegment::AccountChangeSets, last_block + 1)?;
@@ -209,6 +221,106 @@ impl AccountHistory {
             |a, b| a.key == b.key,
         )
         .map_err(Into::into)
+    }
+
+    /// Prunes account history when indices are stored in `RocksDB`.
+    ///
+    /// Reads account changesets from static files and prunes the corresponding
+    /// `RocksDB` history shards.
+    #[cfg(all(unix, feature = "rocksdb"))]
+    fn prune_rocksdb<Provider>(
+        &self,
+        provider: &Provider,
+        input: PruneInput,
+        range: std::ops::RangeInclusive<BlockNumber>,
+        range_end: BlockNumber,
+    ) -> Result<SegmentOutput, PrunerError>
+    where
+        Provider: DBProvider + StaticFileProviderFactory + ChangeSetReader + RocksDBProviderFactory,
+    {
+        use reth_provider::PruneShardOutcome;
+
+        // Unlike MDBX path, we don't divide the limit by 2 because RocksDB path only prunes
+        // history shards (no separate changeset table to delete from). The changesets are in
+        // static files which are deleted separately.
+        let mut limiter = input.limiter;
+
+        if limiter.is_limit_reached() {
+            return Ok(SegmentOutput::not_done(
+                limiter.interrupt_reason(),
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        let mut highest_deleted_accounts = FxHashMap::default();
+        let mut last_changeset_pruned_block = None;
+        let mut changesets_processed = 0usize;
+        let mut done = true;
+
+        // Walk account changesets from static files using a streaming iterator.
+        // For each changeset, track the highest block number seen for each address
+        // to determine which history shard entries need pruning.
+        let walker = StaticFileAccountChangesetWalker::new(provider, range);
+        for result in walker {
+            if limiter.is_limit_reached() {
+                done = false;
+                break;
+            }
+            let (block_number, changeset) = result?;
+            highest_deleted_accounts.insert(changeset.address, block_number);
+            last_changeset_pruned_block = Some(block_number);
+            changesets_processed += 1;
+            limiter.increment_deleted_entries_count();
+        }
+        trace!(target: "pruner", processed = %changesets_processed, %done, "Scanned account changesets from static files");
+
+        let last_changeset_pruned_block = last_changeset_pruned_block
+            .map(|block_number| if done { block_number } else { block_number.saturating_sub(1) })
+            .unwrap_or(range_end);
+
+        // Prune RocksDB history shards for affected accounts
+        let mut deleted_shards = 0usize;
+        let mut updated_shards = 0usize;
+
+        // Sort by address for better RocksDB cache locality
+        let mut sorted_accounts: Vec<_> = highest_deleted_accounts.into_iter().collect();
+        sorted_accounts.sort_unstable_by_key(|(addr, _)| *addr);
+
+        provider.with_rocksdb_batch(|mut batch| {
+            for (address, highest_block) in &sorted_accounts {
+                let prune_to = (*highest_block).min(last_changeset_pruned_block);
+                match batch.prune_account_history_to(*address, prune_to)? {
+                    PruneShardOutcome::Deleted => deleted_shards += 1,
+                    PruneShardOutcome::Updated => updated_shards += 1,
+                    PruneShardOutcome::Unchanged => {}
+                }
+            }
+            Ok(((), Some(batch.into_inner())))
+        })?;
+        trace!(target: "pruner", deleted = deleted_shards, updated = updated_shards, %done, "Pruned account history (RocksDB indices)");
+
+        // Delete static file jars only when fully processed. During provider.commit(), RocksDB
+        // batch is committed before the MDBX checkpoint. If crash occurs after RocksDB commit
+        // but before MDBX commit, on restart the pruner checkpoint indicates data needs
+        // re-pruning, but the RocksDB shards are already pruned - this is safe because pruning
+        // is idempotent (re-pruning already-pruned shards is a no-op).
+        if done {
+            provider.static_file_provider().delete_segment_below_block(
+                StaticFileSegment::AccountChangeSets,
+                last_changeset_pruned_block + 1,
+            )?;
+        }
+
+        let progress = limiter.progress(done);
+
+        Ok(SegmentOutput {
+            progress,
+            pruned: changesets_processed + deleted_shards + updated_shards,
+            checkpoint: Some(SegmentOutputCheckpoint {
+                block_number: Some(last_changeset_pruned_block),
+                tx_number: None,
+            }),
+        })
     }
 }
 
@@ -538,5 +650,273 @@ mod tests {
         );
         test_prune(998, 2, (PruneProgress::Finished, 1000));
         test_prune(1400, 3, (PruneProgress::Finished, 804));
+    }
+
+    #[cfg(all(unix, feature = "rocksdb"))]
+    #[test]
+    fn prune_rocksdb_path() {
+        use reth_db_api::models::ShardedKey;
+        use reth_provider::{RocksDBProviderFactory, StaticFileProviderFactory};
+
+        let db = TestStageDB::default();
+        let mut rng = generators::rng();
+
+        let blocks = random_block_range(
+            &mut rng,
+            0..=100,
+            BlockRangeParams { parent: Some(B256::ZERO), tx_count: 0..1, ..Default::default() },
+        );
+        db.insert_blocks(blocks.iter(), StorageKind::Database(None)).expect("insert blocks");
+
+        let accounts = random_eoa_accounts(&mut rng, 2).into_iter().collect::<BTreeMap<_, _>>();
+
+        let (changesets, _) = random_changeset_range(
+            &mut rng,
+            blocks.iter(),
+            accounts.into_iter().map(|(addr, acc)| (addr, (acc, Vec::new()))),
+            0..0,
+            0..0,
+        );
+
+        db.insert_changesets_to_static_files(changesets.clone(), None)
+            .expect("insert changesets to static files");
+
+        let mut account_blocks: BTreeMap<_, Vec<u64>> = BTreeMap::new();
+        for (block, changeset) in changesets.iter().enumerate() {
+            for (address, _, _) in changeset {
+                account_blocks.entry(*address).or_default().push(block as u64);
+            }
+        }
+
+        let rocksdb = db.factory.rocksdb_provider();
+        let mut batch = rocksdb.batch();
+        for (address, block_numbers) in &account_blocks {
+            let shard = BlockNumberList::new_pre_sorted(block_numbers.iter().copied());
+            batch
+                .put::<tables::AccountsHistory>(ShardedKey::new(*address, u64::MAX), &shard)
+                .unwrap();
+        }
+        batch.commit().unwrap();
+
+        for (address, expected_blocks) in &account_blocks {
+            let shards = rocksdb.account_history_shards(*address).unwrap();
+            assert_eq!(shards.len(), 1);
+            assert_eq!(shards[0].1.iter().collect::<Vec<_>>(), *expected_blocks);
+        }
+
+        let to_block: BlockNumber = 50;
+        let prune_mode = PruneMode::Before(to_block);
+        let input =
+            PruneInput { previous_checkpoint: None, to_block, limiter: PruneLimiter::default() };
+        let segment = AccountHistory::new(prune_mode);
+
+        db.factory.set_storage_settings_cache(
+            StorageSettings::default()
+                .with_account_changesets_in_static_files(true)
+                .with_account_history_in_rocksdb(true),
+        );
+
+        let provider = db.factory.database_provider_rw().unwrap();
+        let result = segment.prune(&provider, input).unwrap();
+        provider.commit().expect("commit");
+
+        assert_matches!(
+            result,
+            SegmentOutput { progress: PruneProgress::Finished, pruned, checkpoint: Some(_) }
+                if pruned > 0
+        );
+
+        for (address, original_blocks) in &account_blocks {
+            let shards = rocksdb.account_history_shards(*address).unwrap();
+
+            let expected_blocks: Vec<u64> =
+                original_blocks.iter().copied().filter(|b| *b > to_block).collect();
+
+            if expected_blocks.is_empty() {
+                assert!(
+                    shards.is_empty(),
+                    "Expected no shards for address {address:?} after pruning"
+                );
+            } else {
+                assert_eq!(shards.len(), 1, "Expected 1 shard for address {address:?}");
+                assert_eq!(
+                    shards[0].1.iter().collect::<Vec<_>>(),
+                    expected_blocks,
+                    "Shard blocks mismatch for address {address:?}"
+                );
+            }
+        }
+
+        let static_file_provider = db.factory.static_file_provider();
+        let highest_block = static_file_provider.get_highest_static_file_block(
+            reth_static_file_types::StaticFileSegment::AccountChangeSets,
+        );
+        if let Some(block) = highest_block {
+            assert!(
+                block > to_block,
+                "Static files should only contain blocks above to_block ({to_block}), got {block}"
+            );
+        }
+    }
+
+    /// Tests that when a limiter stops mid-block (with multiple changes for the same block),
+    /// the checkpoint is set to `block_number - 1` to avoid dangling index entries.
+    #[test]
+    fn prune_partial_progress_mid_block() {
+        use alloy_primitives::{Address, U256};
+        use reth_primitives_traits::Account;
+        use reth_testing_utils::generators::ChangeSet;
+
+        let db = TestStageDB::default();
+        let mut rng = generators::rng();
+
+        // Create blocks 0..=10
+        let blocks = random_block_range(
+            &mut rng,
+            0..=10,
+            BlockRangeParams { parent: Some(B256::ZERO), tx_count: 0..1, ..Default::default() },
+        );
+        db.insert_blocks(blocks.iter(), StorageKind::Database(None)).expect("insert blocks");
+
+        // Create specific changesets where block 5 has 4 account changes
+        let addr1 = Address::with_last_byte(1);
+        let addr2 = Address::with_last_byte(2);
+        let addr3 = Address::with_last_byte(3);
+        let addr4 = Address::with_last_byte(4);
+        let addr5 = Address::with_last_byte(5);
+
+        let account = Account { nonce: 1, balance: U256::from(100), bytecode_hash: None };
+
+        // Build changesets: blocks 0-4 have 1 change each, block 5 has 4 changes, block 6 has 1
+        let changesets: Vec<ChangeSet> = vec![
+            vec![(addr1, account, vec![])], // block 0
+            vec![(addr1, account, vec![])], // block 1
+            vec![(addr1, account, vec![])], // block 2
+            vec![(addr1, account, vec![])], // block 3
+            vec![(addr1, account, vec![])], // block 4
+            // block 5: 4 different account changes (sorted by address for consistency)
+            vec![
+                (addr1, account, vec![]),
+                (addr2, account, vec![]),
+                (addr3, account, vec![]),
+                (addr4, account, vec![]),
+            ],
+            vec![(addr5, account, vec![])], // block 6
+        ];
+
+        db.insert_changesets(changesets.clone(), None).expect("insert changesets");
+        db.insert_history(changesets.clone(), None).expect("insert history");
+
+        // Total changesets: 5 (blocks 0-4) + 4 (block 5) + 1 (block 6) = 10
+        assert_eq!(
+            db.table::<tables::AccountChangeSets>().unwrap().len(),
+            changesets.iter().flatten().count()
+        );
+
+        let prune_mode = PruneMode::Before(10);
+
+        // Set limiter to stop after 7 entries (mid-block 5: 5 from blocks 0-4, then 2 of 4 from
+        // block 5). Due to ACCOUNT_HISTORY_TABLES_TO_PRUNE=2, actual limit is 7/2=3
+        // changesets. So we'll process blocks 0, 1, 2 (3 changesets), stopping before block
+        // 3. Actually, let's use a higher limit to reach block 5. With limit=14, we get 7
+        // changeset slots. Blocks 0-4 use 5 slots, leaving 2 for block 5 (which has 4), so
+        // we stop mid-block 5.
+        let deleted_entries_limit = 14; // 14/2 = 7 changeset entries before limit
+        let limiter = PruneLimiter::default().set_deleted_entries_limit(deleted_entries_limit);
+
+        let input = PruneInput { previous_checkpoint: None, to_block: 10, limiter };
+        let segment = AccountHistory::new(prune_mode);
+
+        let provider = db.factory.database_provider_rw().unwrap();
+        provider.set_storage_settings_cache(
+            StorageSettings::default().with_account_changesets_in_static_files(false),
+        );
+        let result = segment.prune(&provider, input).unwrap();
+
+        // Should report that there's more data
+        assert!(!result.progress.is_finished(), "Expected HasMoreData since we stopped mid-block");
+
+        // Save checkpoint and commit
+        segment
+            .save_checkpoint(&provider, result.checkpoint.unwrap().as_prune_checkpoint(prune_mode))
+            .unwrap();
+        provider.commit().expect("commit");
+
+        // Verify checkpoint is set to block 4 (not 5), since block 5 is incomplete
+        let checkpoint = db
+            .factory
+            .provider()
+            .unwrap()
+            .get_prune_checkpoint(PruneSegment::AccountHistory)
+            .unwrap()
+            .expect("checkpoint should exist");
+
+        assert_eq!(
+            checkpoint.block_number,
+            Some(4),
+            "Checkpoint should be block 4 (block before incomplete block 5)"
+        );
+
+        // Verify remaining changesets (block 5 and 6 should still have entries)
+        let remaining_changesets = db.table::<tables::AccountChangeSets>().unwrap();
+        // After pruning blocks 0-4, remaining should be block 5 (4 entries) + block 6 (1 entry) = 5
+        // But since we stopped mid-block 5, some of block 5 might be pruned
+        // However, checkpoint is 4, so on re-run we should re-process from block 5
+        assert!(
+            !remaining_changesets.is_empty(),
+            "Should have remaining changesets for blocks 5-6"
+        );
+
+        // Verify no dangling history indices for blocks that weren't fully pruned
+        // The indices for block 5 should still reference blocks <= 5 appropriately
+        let history = db.table::<tables::AccountsHistory>().unwrap();
+        for (key, _blocks) in &history {
+            // All blocks in the history should be > checkpoint block number
+            // OR the shard's highest_block_number should be > checkpoint
+            assert!(
+                key.highest_block_number > 4,
+                "Found stale history shard with highest_block_number {} <= checkpoint 4",
+                key.highest_block_number
+            );
+        }
+
+        // Run prune again to complete - should finish processing block 5 and 6
+        let input2 = PruneInput {
+            previous_checkpoint: Some(checkpoint),
+            to_block: 10,
+            limiter: PruneLimiter::default().set_deleted_entries_limit(100), // high limit
+        };
+
+        let provider2 = db.factory.database_provider_rw().unwrap();
+        provider2.set_storage_settings_cache(
+            StorageSettings::default().with_account_changesets_in_static_files(false),
+        );
+        let result2 = segment.prune(&provider2, input2).unwrap();
+
+        assert!(result2.progress.is_finished(), "Second run should complete");
+
+        segment
+            .save_checkpoint(
+                &provider2,
+                result2.checkpoint.unwrap().as_prune_checkpoint(prune_mode),
+            )
+            .unwrap();
+        provider2.commit().expect("commit");
+
+        // Verify final checkpoint
+        let final_checkpoint = db
+            .factory
+            .provider()
+            .unwrap()
+            .get_prune_checkpoint(PruneSegment::AccountHistory)
+            .unwrap()
+            .expect("checkpoint should exist");
+
+        // Should now be at block 6 (the last block with changesets)
+        assert_eq!(final_checkpoint.block_number, Some(6), "Final checkpoint should be at block 6");
+
+        // All changesets should be pruned
+        let final_changesets = db.table::<tables::AccountChangeSets>().unwrap();
+        assert!(final_changesets.is_empty(), "All changesets up to block 10 should be pruned");
     }
 }
--- a/crates/prune/prune/src/segments/user/storage_history.rs
+++ b/crates/prune/prune/src/segments/user/storage_history.rs
@@ -12,7 +12,7 @@ use reth_db_api::{
     tables,
     transaction::DbTxMut,
 };
-use reth_provider::{DBProvider, EitherWriter, StaticFileProviderFactory};
+use reth_provider::{DBProvider, EitherWriter, RocksDBProviderFactory, StaticFileProviderFactory};
 use reth_prune_types::{
     PruneMode, PrunePurpose, PruneSegment, SegmentOutput, SegmentOutputCheckpoint,
 };
@@ -43,7 +43,8 @@ where
     Provider: DBProvider<Tx: DbTxMut>
         + StaticFileProviderFactory
         + StorageChangeSetReader
-        + StorageSettingsCache,
+        + StorageSettingsCache
+        + RocksDBProviderFactory,
 {
     fn segment(&self) -> PruneSegment {
         PruneSegment::StorageHistory
@@ -68,6 +69,13 @@ where
         };
         let range_end = *range.end();
 
+        // Check where storage history indices are stored
+        #[cfg(all(unix, feature = "rocksdb"))]
+        if provider.cached_storage_settings().storages_history_in_rocksdb {
+            return self.prune_rocksdb(provider, input, range, range_end);
+        }
+
+        // Check where storage changesets are stored (MDBX path)
         if EitherWriter::storage_changesets_destination(provider).is_static_file() {
             self.prune_static_files(provider, input, range, range_end)
         } else {
@@ -94,6 +102,8 @@ impl StorageHistory {
             input.limiter
         };
 
+        // The limiter may already be exhausted from a previous segment in the same prune run.
+        // Early exit avoids unnecessary iteration when no budget remains.
         if limiter.is_limit_reached() {
             return Ok(SegmentOutput::not_done(
                 limiter.interrupt_reason(),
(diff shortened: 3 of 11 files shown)


crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -217,42 +210,32 @@ where
...
+                    }
                 }
             }
-        }
 
-        // Extract and register RocksDB batch for commit at provider level
-        #[cfg(all(unix, feature = "rocksdb"))]
-        if let Some(batch) = writer.into_raw_rocksdb_batch() {
-            provider.set_pending_rocksdb_batch(batch);
-        }
```
yongkangc, line 255: why did we remove this?

crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -158,40 +158,33 @@ where
...
-                for (index, hash_to_number) in hash_collector.iter()?.enumerate() {
-                    let (hash_bytes, number_bytes) = hash_to_number?;
-                    if index > 0 && index.is_multiple_of(interval) {
-                        info!(
-                            target: "sync::stages::transaction_lookup",
-                            ?append_only,
-                            progress = %format!("{:.2}%", (index as f64 / total_hashes as f64) * 100.0),
-                            "Inserting hashes"
-                        );
-                    }
```
yongkangc: why did we remove all these? i dont think we should change it? 

for this pr why are we changing it?

crates/prune/prune/src/segments/mod.rs
```diff
@@ -20,6 +20,8 @@ pub use user::{
     AccountHistory, Bodies, Receipts as UserReceipts, ReceiptsByLogs, SenderRecovery,
     StorageHistory, TransactionLookup,
 };
+#[cfg(all(unix, feature = "rocksdb"))]
+pub use user::{AccountsHistoryPruner, StoragesHistoryPruner};
```
yongkangc: why is it user::?

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
```diff
@@ -0,0 +1,146 @@
...
+/// 1. [`tables::StorageChangeSets`] from MDBX (the authoritative changeset data)
+/// 2. [`tables::StoragesHistory`] from `RocksDB` (the history indices)
+///
+/// The `RocksDB` history pruning is changeset-driven: we only prune `RocksDB` shards
+/// for (address, slot) pairs that had changesets pruned in this run.
+#[derive(Debug)]
+pub struct StoragesHistoryPruner {
+    mode: PruneMode,
+}
+
```
yongkangc: is this a shallow struct? does this follow the pattern of our other pruners? not rocksdb but mdbx

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
```diff
@@ -0,0 +1,146 @@
...
+    PruneMode, PrunePurpose, PruneSegment, SegmentOutput, SegmentOutputCheckpoint,
+};
+use rustc_hash::FxHashMap;
+use tracing::{instrument, trace};
+
+/// Number of storage history tables to prune in one step.
+///
+/// Storage History consists of two tables: [`tables::StorageChangeSets`] (MDBX) and
+/// [`tables::StoragesHistory`] (`RocksDB`). We want to prune them to the same block number.
+const STORAGE_HISTORY_TABLES_TO_PRUNE: usize = 2;
```
yongkangc: why 2?

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    fn mode(&self) -> Option<PruneMode> {
+        Some(self.mode)
+    }
+
+    fn purpose(&self) -> PrunePurpose {
+        PrunePurpose::User
+    }
+
+    #[instrument(target = "pruner", skip(self, provider), ret(level = "trace"))]
```
yongkangc: why did we instrument here? how does mdbx instrument? is it the same way?

crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -158,40 +158,33 @@ where
...
+                let hash_iter = hash_collector.iter()?;
+
+                provider.with_rocksdb_batch(|rocksdb_batch| {
+                    // Create writer that routes to either MDBX or RocksDB based on settings
```
yongkangc: remove all instances of this  

`// Create writer that routes to either MDBX or RocksDB based on settings` 

we dont want this comment

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
```diff
@@ -0,0 +1,146 @@
...
+        let range = match input.get_next_block_range() {
+            Some(range) => range,
+            None => {
+                trace!(target: "pruner", "No account history to prune");
+                return Ok(SegmentOutput::done())
+            }
+        };
+        let range_end = *range.end();
+
+        let mut limiter = if let Some(limit) = input.limiter.deleted_entries_limit() {
```
yongkangc: add a comment here - clear and concise

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn mode(&self) -> Option<PruneMode> {
+        Some(self.mode)
+    }
+
+    fn purpose(&self) -> PrunePurpose {
+        PrunePurpose::User
+    }
+
+    #[instrument(target = "pruner", skip(self, provider), ret(level = "trace"))]
+    fn prune(&self, provider: &Provider, input: PruneInput) -> Result<SegmentOutput, PrunerError> {
```
yongkangc: i feel like we should generalise the pruner of account and storage impl?

crates/stages/stages/src/stages/tx_lookup.rs
yongkangc: Reverted - this was unrelated scope creep for a history-pruner PR. Restored the original batch handling pattern with `set_pending_rocksdb_batch`.

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: Refactored to match the MDBX pattern. Now uses `prune_table_with_range` to delete MDBX changesets, then prunes RocksDB history indices. Added the same Provider bounds (`DBProvider<Tx: DbTxMut>`) and checkpoint semantics.

crates/stages/stages/src/stages/tx_lookup.rs
yongkangc: Done - removed the comment as part of the revert.

crates/stages/stages/src/stages/tx_lookup.rs
yongkangc: Agreed, reverted the tx_lookup batch refactor. The PR is now focused only on the RocksDB history pruner segments.

crates/prune/prune/src/segments/mod.rs
yongkangc: These segments are `PrunePurpose::User` and mirror the existing `AccountHistory`/`StorageHistory` user segments, so they're exported from `segments::user` consistently.

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: Restored `STORAGE_HISTORY_TABLES_TO_PRUNE = 2` constant with doc comment. Same rationale as MDBX: we're pruning two data structures (MDBX changesets + RocksDB indices) to the same checkpoint, so the limiter budget is split evenly.

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
yongkangc: MDBX pruners instrument the `prune` method the same way (`target="pruner"`, `skip(self, provider)`, return logged at `trace`). Kept consistent for parity in tracing.

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
```diff
@@ -0,0 +1,146 @@
...
+//! These operations are **not atomic** across the two databases:
+//!
+//! - If `RocksDB` pruning fails after MDBX changesets are committed, `RocksDB` history indices will
+//!   be stale (pointing to deleted changesets).
+//! - Recovery is problematic because the pruned changeset data (needed to derive affected keys) is
+//!   already gone from MDBX.
+//!
+//! **Operational recommendation**: Treat `RocksDB` pruning errors as serious. If
+//! `RocksDB` fails, consider rebuilding history indices rather than assuming
+//! re-running prune will fix the inconsistency.
```
yongkangc: lets remove these

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
```diff
@@ -0,0 +1,146 @@
...
+//! These operations are **not atomic** across the two databases:
+//!
+//! - If `RocksDB` pruning fails after MDBX changesets are committed, `RocksDB` history indices will
+//!   be stale (pointing to deleted changesets).
+//! - Recovery is problematic because the pruned changeset data (needed to derive affected keys) is
+//!   already gone from MDBX.
+//!
+//! **Operational recommendation**: Treat `RocksDB` pruning errors as serious. If
+//! `RocksDB` fails, consider rebuilding history indices rather than assuming
+//! re-running prune will fix the inconsistency.
```
yongkangc: remove these

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: why are we pruning MDBX and not static files?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -3507,7 +3507,15 @@ impl<TX: DbTx + 'static, N: NodeTypes + 'static> DBProvider for DatabaseProvider
...
-                    self.rocksdb_provider.commit_batch(batch)?;
+                    if let Err(err) = self.rocksdb_provider.commit_batch(batch) {
+                        warn!(
+                            target: "providers::db",
+                            %err,
+                            "RocksDB batch commit failed after MDBX commit succeeded; \
+                             database state may be inconsistent"
+                        );
+                        return Err(err);
+                    }
```
yongkangc: lets revert this part to main. i dont see why we need to add this

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
yongkangc: Both imports are actively used:
- `StaticFileSegment` → used at line 94 (`StaticFileSegment::AccountChangeSets`) and in tests
- `FxHashMap` → used at lines 49 and 129

Cannot remove without breaking compilation.

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: Both imports are actively used:
- `StaticFileSegment` → used at line 94 (`StaticFileSegment::StorageChangeSets`) and in tests
- `FxHashMap` → used at lines 49 and 129

Cannot remove without breaking compilation.

crates/prune/prune/src/segments/user/rocksdb_account_history.rs
yongkangc: Done - added a comment explaining the RocksDB history index pruning logic.

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: Good point - the account and storage pruners share significant logic. Could generalize with a trait or generic implementation in a follow-up PR. Keeping them separate for now ensures the initial implementation is straightforward to review and test independently.

crates/prune/prune/src/segments/user/rocksdb_storage_history.rs
yongkangc: The code handles both:
- **Static files**: When changesets are in static files (line 216-217), we call `prune_static_files()` which reads changesets from static files, prunes RocksDB indices, then deletes the static file jars below the pruned block (line 93-96).
- **MDBX fallback**: When changesets are in MDBX (line 218-219), we call `prune_database()` which reads from MDBX tables.

The `EitherWriter::*_changesets_destination(provider).is_static_file()` check routes to the appropriate path.

crates/storage/provider/src/providers/database/provider.rs
yongkangc: This change adds explicit error logging when a RocksDB batch fails after MDBX already committed - an important warning for debugging potential consistency issues. Happy to revert if you feel strongly, but wanted to flag that this pattern helps surface issues during dual-database operations.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -2767,4 +2879,1021 @@ mod tests {
...
+        let mut batch = provider.batch();
+        let outcome = batch.prune_account_history_to(address, 5).unwrap();
+        batch.commit().unwrap();
+
+        assert_eq!(outcome, PruneShardOutcome::Unchanged);
+
+        let shards = provider.account_history_shards(address).unwrap();
+        assert_eq!(shards.len(), 1);
+        assert_eq!(shards[0].1.iter().collect::<Vec<_>>(), vec![10, 20]);
+    }
```
gakonst, line 3129: [LOW] Consider adding a test for `prune_account_history_to(address, 0)` to verify block 0 edge case handling. The unwind tests cover this scenario, but the prune path may have different semantics.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+            )?;
+        }
+
+        let progress = limiter.progress(done);
+
+        Ok(SegmentOutput {
+            progress,
+            pruned: changesets_processed + deleted_shards + updated_shards,
+            checkpoint: Some(SegmentOutputCheckpoint {
+                block_number: Some(last_changeset_pruned_block),
```
gakonst, line 320: [MEDIUM] No crash recovery or atomicity test. Consider adding a test that simulates a mid-prune failure (e.g., drop batch before commit) and verifies that history remains consistent. The RocksDB batched commit should be atomic, but this isn't verified in tests.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+        let static_file_provider = db.factory.static_file_provider();
+        let highest_block = static_file_provider.get_highest_static_file_block(
+            reth_static_file_types::StaticFileSegment::AccountChangeSets,
+        );
+        if let Some(block) = highest_block {
+            assert!(
+                block > to_block,
+                "Static files should only contain blocks above to_block ({to_block}), got {block}"
+            );
+        }
```
gakonst, line 759: [LOW] Integration test `prune_rocksdb_path` only tests end-to-end success. Consider adding a test that verifies MDBX and RocksDB produce equivalent final state after pruning the same data, using a direct comparison (not just checking that both 'work').

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -79,19 +86,26 @@ impl AccountHistory {
     fn prune_static_files<Provider>(
```
gakonst, line 88: [MINOR] **Code duplication across pruning paths**: The three methods `prune_static_files`, `prune_database`, and `prune_rocksdb` share significant structural duplication: same limiter check pattern, same `highest_deleted_accounts` HashMap pattern, same checkpoint calculation logic, and same `SegmentOutput` construction. Consider extracting the common changeset-walking + checkpoint logic into a helper that returns the collected addresses and last block, letting each path only implement its specific deletion logic.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+            .unwrap()
+            .expect("checkpoint should exist");
+
+        // Should now be at block 6 (the last block with changesets)
+        assert_eq!(final_checkpoint.block_number, Some(6), "Final checkpoint should be at block 6");
+
+        // All changesets should be pruned
+        let final_changesets = db.table::<tables::AccountChangeSets>().unwrap();
+        assert!(final_changesets.is_empty(), "All changesets up to block 10 should be pruned");
+    }
```
gakonst, line 921: [LOW] `prune_partial_progress_mid_block` test verifies checkpoint logic but doesn't test the RocksDB path. The RocksDB prune path doesn't divide limit by 2 (line 271), which may affect limiter behavior differently. Consider adding an equivalent test for the RocksDB flow.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1425,6 +1427,116 @@ impl<'a> RocksDBBatch<'a> {
         Ok(())
     }
 
+    /// Prunes history shards, removing blocks <= `to_block`.
+    ///
+    /// Generic implementation for both account and storage history pruning.
+    /// Mirrors MDBX `prune_shard` semantics. After pruning, the last remaining shard
+    /// (if any) will have the sentinel key (`u64::MAX`).
+    #[allow(clippy::too_many_arguments)]
+    fn prune_history_shards_inner<K>(
```
gakonst, line 1637: [POSITIVE] **Good abstraction**: The `prune_history_shards_inner` generic helper with closure parameters is well-designed - it provides a deep module that hides the complexity of shard iteration, filtering, and sentinel key management behind a simple interface. The 7 closure parameters may look complex at first, but they correctly factor out the minimal differences between account/storage history tables. This follows Ousterhout's principle of hiding complexity behind simple interfaces.

crates/storage/provider/src/providers/rocksdb_types.rs
```diff
@@ -0,0 +1,14 @@
+//! Shared types for `RocksDB` provider.
+//!
+//! These types are used by both the real `RocksDB` implementation and the stub.
+
+/// Outcome of pruning a history shard.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PruneShardOutcome {
```
gakonst: [MINOR] **Enum vs bool/count**: `PruneShardOutcome` with three variants (Deleted/Updated/Unchanged) is appropriate here. It's not over-engineering because: (1) callers use different logic paths based on outcome, (2) using a bool would lose information about whether deletion vs update occurred, (3) a simple count wouldn't distinguish deletion from update. The enum correctly captures the semantic differences needed by the callers.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+        input: PruneInput,
+        range: std::ops::RangeInclusive<BlockNumber>,
+        range_end: BlockNumber,
+    ) -> Result<SegmentOutput, PrunerError>
+    where
+        Provider: DBProvider + StaticFileProviderFactory + ChangeSetReader + RocksDBProviderFactory,
+    {
+        use reth_provider::PruneShardOutcome;
+
+        // Unlike MDBX path, we don't divide the limit by 2 because RocksDB path only prunes
```
gakonst, line 243: [POSITIVE] **Good "why" comment**: This comment explains the reasoning behind not dividing the limit by 2, which is non-obvious behavior that differs from the MDBX path. This is exactly the kind of comment that adds value - explaining "why" rather than "what".

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+        // Total changesets: 5 (blocks 0-4) + 4 (block 5) + 1 (block 6) = 10
+        assert_eq!(
+            db.table::<tables::AccountChangeSets>().unwrap().len(),
+            changesets.iter().flatten().count()
+        );
+
+        let prune_mode = PruneMode::Before(10);
+
+        // Set limiter to stop after 7 entries (mid-block 5: 5 from blocks 0-4, then 2 of 4 from
+        // block 5). Due to ACCOUNT_HISTORY_TABLES_TO_PRUNE=2, actual limit is 7/2=3
```
gakonst, line 819: [MINOR] **Test comment verbosity**: This test has extensive inline comments explaining test scenarios (e.g., "Set limiter to stop after 7 entries (mid-block 5: 5 from blocks 0-4...)"). While helpful for understanding the test logic, consider moving some of this detail into a test docstring at the top of the function. The per-line comments like "// block 0", "// block 1" etc. are fine - they're concise labels. But the multi-line calculation explanations could be more condensed.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+                    PruneShardOutcome::Deleted => deleted_shards += 1,
+                    PruneShardOutcome::Updated => updated_shards += 1,
+                    PruneShardOutcome::Unchanged => {}
+                }
+            }
+            Ok(((), Some(batch.into_inner())))
+        })?;
+        trace!(target: "pruner", deleted = deleted_shards, updated = updated_shards, %done, "Pruned account history (RocksDB indices)");
+
+        // Delete static file jars AFTER RocksDB batch is queued. This ensures that if we crash
```
gakonst: [POSITIVE] **Crash recovery comment**: This is an excellent "why" comment that explains the ordering requirement for crash safety. This is exactly the kind of non-obvious operational constraint that future maintainers need to understand.

gakonst (commented):
## Performance Review

### 1. RocksDB Batch Operations ✅ Good
The `with_rocksdb_batch()` pattern properly batches all puts/deletes into a single `WriteBatchWithTransaction` before committing. The batch is accumulated in memory and written atomically - this is efficient.

### 2. Shard Fetching - Minor Concern 🔸
`account_history_shards(address)` fetches **all shards** for an address into a Vec before pruning. For accounts with many shards (heavy DeFi contracts), this loads all shard data into memory even when we may only need to delete the first few shards.

**Mitigation**: Most accounts have 1-2 shards in practice. For heavy accounts, a streaming iterator that processes shards lazily would be more memory-efficient, but the current approach is acceptable.

### 3. Memory Usage ✅ Bounded
`FxHashMap` for `highest_deleted_accounts` is bounded by the limiter's `deleted_entries_limit`. Per the comment, it's ~8750 entries (0.5MB) under default settings. This is safe.

`sorted_accounts` vector is the same size as the hashmap - also bounded.

### 4. I/O Patterns ✅ Good Design
Sorting by address before RocksDB operations (`sorted_accounts.sort_unstable_by_key`) is effective for RocksDB cache locality since:
- RocksDB keys are ordered by (address, highest_block_number)
- Sequential address access maximizes block cache hits
- Minimizes random I/O during compaction

### 5. Static File Jar Deletion 🔸 Consider
`delete_segment_below_block` calls `initialize_index()` after each jar deletion, which:
- Clears and rebuilds all segment indexes
- Calls `iter_static_files()` to rescan the directory
- Clears the jar cache (`self.map.clear()`)

For many jars, this is O(n²) where n = number of jars. Consider batching: delete all jars first, then call `initialize_index()` once. Though in practice, pruning deletes 1-2 jars at a time, so this is likely fine.

### 6. Compared to MDBX Path ✅ Similar
- Both paths use the same limiter bounds
- RocksDB path doesn't divide limit by 2 (no separate changeset table) - correctly documented
- Static file walker is identical to MDBX's static file path

**Overall**: No performance regressions identified. The implementation follows RocksDB best practices.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+
+        let last_changeset_pruned_block = last_changeset_pruned_block
+            .map(|block_number| if done { block_number } else { block_number.saturating_sub(1) })
+            .unwrap_or(range_end);
+
+        // Prune RocksDB history shards for affected accounts
+        let mut deleted_shards = 0usize;
+        let mut updated_shards = 0usize;
+
+        // Sort by address for better RocksDB cache locality
```
gakonst, line 285: [MINOR] **Comment restates code**: This comment ("Sort by address for better RocksDB cache locality") while explaining the "why" is borderline - the code `sort_unstable_by_key(|(addr, _)| *addr)` already shows sorting by address, but explaining the performance reason is useful. This is acceptable, though the comment could be shortened to just "// better RocksDB cache locality" since "sort by address" is visible in the code.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1425,6 +1427,116 @@ impl<'a> RocksDBBatch<'a> {
...
+        } else {
+            Ok(PruneShardOutcome::Unchanged)
+        }
+    }
+
+    /// Prunes account history for the given address, removing blocks <= `to_block`.
+    ///
+    /// Mirrors MDBX `prune_shard` semantics. After pruning, the last remaining shard
+    /// (if any) will have the sentinel key (`u64::MAX`).
+    pub fn prune_account_history_to(
```
gakonst, line 1701: [MINOR] **Naming consistency**: `prune_account_history_to` and `prune_storage_history_to` are consistent and clear. The parameter name `to_block` is slightly ambiguous - it's the upper bound of blocks to prune (blocks <= to_block are pruned). Consider documenting this in the docstrings more explicitly as "blocks <= to_block will be removed". The current docstring says "removing blocks <= to_block" which is correct.

gakonst (commented):
test comment

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1425,6 +1427,116 @@ impl<'a> RocksDBBatch<'a> {
         Ok(())
     }
 
+    /// Prunes history shards, removing blocks <= `to_block`.
```
gakonst, line 1631: [MINOR] **No AI slop detected**: The comments in this PR are generally technical and purposeful. No signs of AI-generated filler phrases like "This ensures", "It is important to note that", "This allows for", etc. The doc comments explain behavior clearly without excessive verbosity. Good quality overall.

gakonst (commented):
Starting review of RocksDB pruning correctness

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+        let addr3 = Address::with_last_byte(3);
+        let addr4 = Address::with_last_byte(4);
+        let addr5 = Address::with_last_byte(5);
+
+        let account = Account { nonce: 1, balance: U256::from(100), bytecode_hash: None };
+
+        // Build changesets: blocks 0-4 have 1 change each, block 5 has 4 changes, block 6 has 1
+        let changesets: Vec<ChangeSet> = vec![
+            vec![(addr1, account, vec![])], // block 0
+            vec![(addr1, account, vec![])], // block 1
```
gakonst, line 793: [BLOCKER] **Crash consistency violation**: Static file jars are deleted immediately here, but RocksDB batch commit happens later (during provider commit), and MDBX checkpoint commit happens even later. If crash occurs after static file deletion but before MDBX commit:

1. Static file changesets are gone (deleted)
2. Checkpoint is NOT updated  
3. On restart, pruner reads old checkpoint, tries to re-read changesets from deleted static files → data loss/corruption

The code comment says 'RocksDB batch will be committed atomically with the checkpoint' but RocksDB commits BEFORE MDBX in the normal (non-unwind) path.

**Suggested fix**: Defer static file deletion to after MDBX commit, or batch it with provider commit.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -2767,4 +2879,1021 @@ mod tests {
         assert_eq!(shards[1].0.highest_block_number, u64::MAX);
         assert_eq!(shards[1].1.iter().collect::<Vec<_>>(), (51..=75).collect::<Vec<_>>());
     }
+
+    #[test]
+    fn test_prune_account_history_single_shard_truncate() {
```
gakonst: [POSITIVE] **Well-structured tests**: The test suite follows good patterns:
- Clear test names describing the scenario (`test_prune_account_history_single_shard_truncate`, etc.)
- Tests cover edge cases systematically: empty shards, single shard, multi-shard, sentinel promotion, etc.
- Each test is self-contained with clear setup/act/assert
- The equivalence tests comparing RocksDB to MDBX semantics are valuable for correctness verification

The 21 tests provide good coverage of the prune logic's edge cases.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+            .map(|block_number| if done { block_number } else { block_number.saturating_sub(1) })
+            .unwrap_or(range_end);
+
+        // Prune RocksDB history shards for affected accounts
+        let mut deleted_shards = 0usize;
+        let mut updated_shards = 0usize;
+
+        // Sort by address for better RocksDB cache locality
+        let mut sorted_accounts: Vec<_> = highest_deleted_accounts.into_iter().collect();
+        sorted_accounts.sort_unstable_by_key(|(addr, _)| *addr);
```
gakonst, line 287: [PERFORMANCE - INFO] Good use of address-ordered iteration. Since RocksDB keys are ordered by (address, highest_block_number), this sequential access pattern maximizes block cache hits and minimizes random I/O during both reads and any subsequent compaction.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -2767,4 +2879,1021 @@ mod tests {
...
+        assert_eq!(shards[0].0.highest_block_number, 50);
+        assert_eq!(shards[0].1.iter().collect::<Vec<_>>(), vec![30, 40, 50]);
+        // Sentinel unchanged
+        assert_eq!(shards[1].0.highest_block_number, u64::MAX);
+        assert_eq!(shards[1].1.iter().collect::<Vec<_>>(), vec![60, 70]);
+    }
+
+    #[test]
+    fn test_prune_storage_history_single_shard_truncate() {
+        let temp_dir = TempDir::new().unwrap();
```
gakonst: [SUGGESTION] **Test simplification opportunity**: The equivalence tests (e.g., `test_prune_equivalence_*`) could use a parameterized test approach or a helper function to reduce boilerplate. Each test follows the same pattern: create TempDir, build provider, set up shards, prune, verify. A helper like `fn prune_test_case(shards: &[(u64, Vec<u64>)], to_block: u64, expected: &[(u64, Vec<u64>)])` could cut test code by ~50%. However, the current explicit style is also valid for readability - this is a minor suggestion, not blocking.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+
+        db.factory.set_storage_settings_cache(
+            StorageSettings::default()
+                .with_account_changesets_in_static_files(true)
+                .with_account_history_in_rocksdb(true),
+        );
+
+        let provider = db.factory.database_provider_rw().unwrap();
+        let result = segment.prune(&provider, input).unwrap();
+        provider.commit().expect("commit");
```
gakonst, line 721: [BLOCKER] **Same crash consistency issue as in prune_static_files**: Static file jars are deleted immediately here, but RocksDB batch commit and MDBX checkpoint commit happen later during provider.commit(). If crash occurs between these operations, on restart the pruner will try to read changesets from deleted static files.

See longer explanation in the other comment.

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1425,6 +1427,116 @@ impl<'a> RocksDBBatch<'a> {
...
+    /// Prunes account history for the given address, removing blocks <= `to_block`.
+    ///
+    /// Mirrors MDBX `prune_shard` semantics. After pruning, the last remaining shard
+    /// (if any) will have the sentinel key (`u64::MAX`).
+    pub fn prune_account_history_to(
+        &mut self,
+        address: Address,
+        to_block: BlockNumber,
+    ) -> ProviderResult<PruneShardOutcome> {
+        let shards = self.provider.account_history_shards(address)?;
```
gakonst, line 1706: [PERFORMANCE - LOW] This loads all shards for an address into memory before pruning. For accounts with many shards (e.g., heavily-used DeFi contracts), this could load significant BlockNumberList data.

Consider: For the common case where `highest_block_number <= to_block` on the first shard, we could short-circuit by streaming/iterating rather than collecting. However, most accounts have 1-2 shards, so this is acceptable for typical workloads.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+            }
+            Ok(((), Some(batch.into_inner())))
+        })?;
+        trace!(target: "pruner", deleted = deleted_shards, updated = updated_shards, %done, "Pruned account history (RocksDB indices)");
+
+        // Delete static file jars AFTER RocksDB batch is queued. This ensures that if we crash
+        // after jar deletion but before checkpoint commit, we can still recover because the
+        // RocksDB batch will be committed atomically with the checkpoint.
+        if done {
+            provider.static_file_provider().delete_segment_below_block(
```
gakonst, line 308: [PERFORMANCE - INFO] Note: `delete_segment_below_block` calls `initialize_index()` after each jar deletion, which rescans the directory and rebuilds all segment indexes. For bulk jar deletion this would be O(n²), but in practice pruning deletes 1-2 jars at a time so this is acceptable. If future use cases require deleting many jars, consider a `delete_jars_below_block_batch` that defers `initialize_index()` to the end.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+        let mut limiter = input.limiter;
+
+        if limiter.is_limit_reached() {
+            return Ok(SegmentOutput::not_done(
+                limiter.interrupt_reason(),
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        let mut highest_deleted_accounts = FxHashMap::default();
```
gakonst, line 255: [PERFORMANCE - OK] Memory is properly bounded. Unlike the MDBX path (which has an inline comment documenting the ~0.5MB bound), the RocksDB path uses the full limiter (not divided by 2) since there's no separate changeset table. The effective bound is `deleted_entries_limit` entries, which at default settings is ~3500-7000 addresses.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -233,6 +247,101 @@ impl AccountHistory {
...
+
+        // Prune RocksDB history shards for affected accounts
+        let mut deleted_shards = 0usize;
+        let mut updated_shards = 0usize;
+
+        // Sort by address for better RocksDB cache locality
+        let mut sorted_accounts: Vec<_> = highest_deleted_accounts.into_iter().collect();
+        sorted_accounts.sort_unstable_by_key(|(addr, _)| *addr);
+
+        provider.with_rocksdb_batch(|mut batch| {
```
gakonst, line 289: [PERFORMANCE - OK] Correct use of batching. All `prune_account_history_to` calls accumulate in a single `WriteBatchWithTransaction`, which is committed atomically. This avoids per-shard I/O overhead and ensures atomicity.

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -245,6 +260,102 @@ impl StorageHistory {
...
+        for result in walker {
+            if limiter.is_limit_reached() {
+                done = false;
+                break;
+            }
+            let (block_address, entry) = result?;
+            let block_number = block_address.block_number();
+            let address = block_address.address();
+            highest_deleted_storages.insert((address, entry.key), block_number);
+            last_changeset_pruned_block = Some(block_number);
```
gakonst, line 274: [BLOCKER] Same crash consistency issue as in account_history.rs. Static file jars deleted immediately but checkpoint updated later. See comment on account_history.rs for details.

crates/prune/prune/src/segments/user/account_history.rs
gakonst, line 243: [MAJOR] **Missing done check**: This deletes static file jars regardless of whether pruning is complete. If `done == false` (limiter hit mid-block), we delete jars up to `last_block`, but checkpoint is set to `last_block - 1`. On restart, we try to re-read changesets for `last_block` but the jar is already deleted.

Compare with the `prune_rocksdb` path at line 327 which correctly guards deletion with `if done`.

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -245,6 +260,102 @@ impl StorageHistory {
...
+        let mut changesets_processed = 0usize;
+        let mut done = true;
+
+        let walker = provider.static_file_provider().walk_storage_changeset_range(range);
+        for result in walker {
+            if limiter.is_limit_reached() {
+                done = false;
+                break;
+            }
+            let (block_address, entry) = result?;
```
gakonst, line 270: [MAJOR] Same issue as account_history.rs prune_static_files: missing `if done` guard before deleting static file jars. If interrupted mid-block, this deletes jars that are still needed for the next run.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+            PruneInput { previous_checkpoint: None, to_block, limiter: PruneLimiter::default() };
+        let segment = AccountHistory::new(prune_mode);
+
+        db.factory.set_storage_settings_cache(
+            StorageSettings::default()
+                .with_account_changesets_in_static_files(true)
+                .with_account_history_in_rocksdb(true),
+        );
+
+        let provider = db.factory.database_provider_rw().unwrap();
```
gakonst, line 719: [MINOR] This comment is incorrect. Looking at provider.rs commit() (lines 3531-3542), in the normal (non-unwind) path:
1. Static files finalized
2. **RocksDB batches committed** 
3. MDBX committed (which includes checkpoint)

RocksDB is committed BEFORE MDBX, not atomically with it. If crash occurs between RocksDB commit and MDBX commit, the checkpoint won't be updated but RocksDB shards are already pruned.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -562,4 +671,272 @@ mod tests {
...
+
+        assert_matches!(
+            result,
+            SegmentOutput { progress: PruneProgress::Finished, pruned, checkpoint: Some(_) }
+                if pruned > 0
+        );
+
+        for (address, original_blocks) in &account_blocks {
+            let shards = rocksdb.account_history_shards(*address).unwrap();
+
```
gakonst, line 731: [NIT] Inconsistency: MDBX path at line 165 only counts `outcomes.deleted`, but RocksDB path counts `deleted_shards + updated_shards`. Should `updated_shards` be included? The semantics differ from MDBX.

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 120: add back this comment

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -65,7 +66,13 @@ where
...
-        // Check where account changesets are stored
+        // Check where account history indices are stored
+        #[cfg(all(unix, feature = "rocksdb"))]
+        if provider.cached_storage_settings().account_history_in_rocksdb {
+            return self.prune_rocksdb(provider, input, range, range_end);
+        }
+
+        // Check where account changesets are stored (MDBX path)
         if EitherWriter::account_changesets_destination(provider).is_static_file() {
             self.prune_static_files(provider, input, range, range_end)
```
yongkangc, line 79: why did we delete input.limiter here?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 91: why did we change it to prune input?

crates/storage/provider/src/providers/rocksdb_types.rs
```diff
@@ -0,0 +1,14 @@
...
+    /// Shard was deleted entirely.
+    Deleted,
+    /// Shard was updated with filtered block numbers.
+    Updated,
+    /// Shard was unchanged (no blocks <= `to_block`).
+    Unchanged,
+}
```
yongkangc: i dont think this should be another file right? what do you think

crates/prune/prune/src/segments/user/storage_history.rs
yongkangc, line 112: add a clear concise comment for this. when would a limiter be reached?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 91: why did we change it to prune input?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 79: why did we delete input.limiter here?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 120: add back this comment

crates/storage/provider/src/providers/rocksdb_types.rs
yongkangc: i dont think this should be another file right?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 79: why did we delete input.limiter here?

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 91: why did we change it to prune input?

yongkangc: The `prune()` method now routes to backend-specific helpers: `prune_static_files()`, `prune_database()`, and `prune_rocksdb()`. We pass the full `PruneInput` to avoid a long parameter list — each helper extracts what it needs (limiter, checkpoint).

This refactoring exists because the RocksDB path requires `#[cfg(all(unix, feature = "rocksdb"))]` gating and each storage backend has meaningfully different logic. Keeping them as separate methods makes the cfg-gating clean and the main `prune()` a simple router.

crates/prune/prune/src/segments/user/account_history.rs
yongkangc: Same as above — the limiter setup moved into the helper methods (`prune_static_files`, `prune_database`, `prune_rocksdb`). The main `prune()` now just routes to the appropriate backend based on storage settings.

crates/prune/prune/src/segments/user/account_history.rs
yongkangc, line 120: add back this comment

crates/prune/prune/Cargo.toml
```diff
@@ -42,11 +42,15 @@ rayon.workspace = true
 tokio.workspace = true
 rustc-hash.workspace = true
 
+[features]
+rocksdb = ["reth-provider/rocksdb"]
```
yongkangc, line 46: do we need to add this? look at other cargo.toml

crates/prune/prune/Cargo.toml
```diff
@@ -42,11 +42,15 @@ rayon.workspace = true
...
+
 [dev-dependencies]
 # reth
 reth-db = { workspace = true, features = ["test-utils"] }
-reth-stages = { workspace = true, features = ["test-utils"] }
+reth-stages = { workspace = true, features = ["test-utils", "rocksdb"] }
```
yongkangc, line 51: do we need to add this? look at other cargo.toml

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -210,6 +221,103 @@ impl AccountHistory {
...
+        &self,
+        provider: &Provider,
+        input: PruneInput,
+        range: std::ops::RangeInclusive<BlockNumber>,
+        range_end: BlockNumber,
+    ) -> Result<SegmentOutput, PrunerError>
+    where
+        Provider: DBProvider + StaticFileProviderFactory + ChangeSetReader + RocksDBProviderFactory,
+    {
+        use reth_provider::PruneShardOutcome;
```
yongkangc, line 241: move this up to import

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -210,6 +221,103 @@ impl AccountHistory {
...
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        let mut highest_deleted_accounts = FxHashMap::default();
+        let mut last_changeset_pruned_block = None;
+        let mut changesets_processed = 0usize;
+        let mut done = true;
+
+        let walker = StaticFileAccountChangesetWalker::new(provider, range);
```
yongkangc, line 263: add a comment here

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -216,6 +225,104 @@ impl StorageHistory {
...
+        &self,
+        provider: &Provider,
+        input: PruneInput,
+        range: std::ops::RangeInclusive<BlockNumber>,
+        range_end: BlockNumber,
+    ) -> Result<SegmentOutput, PrunerError>
+    where
+        Provider: DBProvider + StaticFileProviderFactory + RocksDBProviderFactory,
+    {
+        use reth_provider::PruneShardOutcome;
```
yongkangc, line 245: move this up

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -216,6 +225,104 @@ impl StorageHistory {
...
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        let mut highest_deleted_storages: FxHashMap<_, _> = FxHashMap::default();
+        let mut last_changeset_pruned_block = None;
+        let mut changesets_processed = 0usize;
+        let mut done = true;
+
+        let walker = provider.static_file_provider().walk_storage_changeset_range(range);
```
yongkangc, line 264: add comment here

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -98,6 +98,13 @@ const DEFAULT_BYTES_PER_SYNC: u64 = 1_048_576;
 /// reducing the first few reallocations without over-allocating.
 const DEFAULT_COMPRESS_BUF_CAPACITY: usize = 4096;
 
+/// Default auto-commit threshold for batch writes (4 GiB).
+///
+/// When a batch exceeds this size, it is automatically committed to prevent OOM
+/// during large bulk writes. The consistency check on startup heals any crash
+/// that occurs between auto-commits.
+const DEFAULT_AUTO_COMMIT_THRESHOLD: usize = 4 * 1024 * 1024 * 1024;
```
yongkangc, line 119: this is from the merge conflict as dan's pr doesnt have main in

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -124,8 +138,7 @@ impl AccountHistory {
             limiter.increment_deleted_entries_count();
         }
 
-        // Delete static file jars below the pruned block
```
yongkangc, line 127: can we restore this comment?

crates/trie/common/src/updates.rs
joshieDo, line 1: if this is not clippy , can we remove these unrelated changes

joshieDo (approved):
lgtm, but would like to take another look once rebased

crates/trie/common/src/updates.rs
yongkangc: sorry @joshieDo the unrelated changes were clippy - will rebase, thanks for taking a look

crates/prune/prune/src/segments/user/storage_history.rs
yongkangc, line 264: can we add a comment here to say what this section does with walker?

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -94,11 +102,13 @@ impl StorageHistory {
...
         };
 
+        // The limiter may already be exhausted from a previous segment in the same prune run.
+        // Early exit avoids unnecessary iteration when no budget remains.
         if limiter.is_limit_reached() {
             return Ok(SegmentOutput::not_done(
                 limiter.interrupt_reason(),
                 input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
-            ))
+            ));
```
yongkangc: remove this ;

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -126,8 +136,7 @@ impl StorageHistory {
             limiter.increment_deleted_entries_count();
         }
 
-        // Delete static file jars below the pruned block
-        if let Some(last_block) = last_changeset_pruned_block {
+        if done && let Some(last_block) = last_changeset_pruned_block {
```
yongkangc, line 140: add back 
`        // Delete static file jars below the pruned block`

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -210,6 +222,101 @@ impl AccountHistory {
...
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        let mut highest_deleted_accounts = FxHashMap::default();
+        let mut last_changeset_pruned_block = None;
+        let mut changesets_processed = 0usize;
+        let mut done = true;
+
+        let walker = StaticFileAccountChangesetWalker::new(provider, range);
```
yongkangc, line 263: can we add a comment here to say what this section does with walker?

PR #21320, opened 2026-01-22
feat(storage): warn storage settings diff at startup

gakonst:
## Summary

Adds validation to detect storage settings mismatches at startup, warning operators when CLI settings differ from stored settings.

Closes #20482

## Changes

- Validate in `init_genesis_with_settings()`: if settings differ from stored, log a warning
- For legacy databases without persisted settings, defaults to `StorageSettings::legacy()` for comparison

## Behavior

| Scenario | Result |
|----------|--------|
| Settings differ from stored | ⚠️ Warning logged, startup proceeds |
| Settings match stored | ✅ OK |
| Legacy DB (no stored settings) | Treated as `legacy()` for comparison |

## Test Plan

2 tests:
- `warn_storage_settings_mismatch` - verifies warning behavior, startup succeeds
- `allow_same_storage_settings` - verifies no warning when settings match

PR #21332, opened 2026-01-22
fix: use unwrap_or_else for lazy evaluation of StorageSettings::legacy

gakonst:
Addresses review comment from https://github.com/paradigmxyz/reth/pull/21320#discussion_r2717711333

Uses `unwrap_or_else(StorageSettings::legacy)` instead of `unwrap_or(StorageSettings::legacy())` to avoid eagerly evaluating the default when not needed.


yongkangc (approved):
this is from a comment that didnt get it

https://github.com/paradigmxyz/reth/pull/21320

PR #21339, opened 2026-01-22
fix(rocksdb): filter history writes to only changed accounts/storage

yongkangc:
**Problem:** RocksDB history index writes included all **touched** accounts/storage from the bundle state, even if only read (not modified). This created false-positive history entries pointing to blocks with no actual changeset, causing `AccountChangesetNotFound` and `StorageChangesetNotFound` errors during historical state queries.

**Root cause:** `write_account_history` iterated `bundle.state().keys()` and `write_storage_history` iterated `account.storage.keys()`. These include all accounts/slots accessed during execution (reads + writes), not just modified ones. History indices must only point to blocks with actual changes, since `HistoricalStateProvider` uses them to look up changesets.

**How found:** During reth-bench historical state queries, `AccountChangesetNotFound` errors surfaced. The history index pointed to blocks with no actual changeset because touched (not just changed) accounts were being indexed.

**Fix:**
- Account history: filter with `account.is_info_changed()`
- Storage history: filter with `slot_value.is_changed()`

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -1078,6 +1078,8 @@ impl RocksDBProvider {
     }
 
     /// Writes account history indices for the given blocks.
+    ///
+    /// Derives history indices from reverts (same source as changesets) to ensure consistency.
     #[instrument(level = "debug", target = "providers::rocksdb", skip_all)]
     fn write_account_history<N: reth_node_types::NodePrimitives>(
         &self,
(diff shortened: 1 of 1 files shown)


pepyakin (approved):

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1089,8 +1089,13 @@ impl RocksDBProvider {
...
             let bundle = &block.execution_outcome().state;
-            for &address in bundle.state().keys() {
-                account_history.entry(address).or_default().push(block_number);
+            for (&address, account) in bundle.state() {
+                // Only index accounts where info actually changed (not just touched/read).
+                // This matches changeset semantics and prevents false-positive history entries
+                // that would cause AccountChangesetNotFound errors during historical lookups.
+                if account.is_info_changed() {
+                    account_history.entry(address).or_default().push(block_number);
+                }
```
Rjected, line 1101: these fns look a bit easy to mess up due to things like this, any reason why we aren't doing either:
```
            let reverts = block.execution_outcome().state.reverts.to_plain_state_reverts();
```

or just calling `changed_accounts_and_blocks_with_range`? or the storage equivalent? since we should save the changesets beforehand

Rjected (changes requested):

Rjected (approved):
lgtm

PR #21334, opened 2026-01-22
fix(rocksdb): periodic batch commits in stages to prevent OOM

joshieDo:
Auto-commit mode for `RocksDBBatch` to prevent OOM during large bulk writes.

**How it works:**
Batch auto-commits and resets when size exceeds 1GB threshold (after `put()`/`delete()`).

**Integration (explicit in each stage):**
- TxLookup: execute
- IndexAccountHistory: execute
- IndexStorageHistory: execute

**Tradeoff:** Breaks stage atomicity. Assumes healing during bootup consistency checks (not happening yet)

PR #21341, opened 2026-01-22
feat(rocksdb): changeset-based crash recovery healing for history indices

joshieDo:
## Summary

Replaces O(n) table scan with changeset-based healing for RocksDB history tables.

## Problem

After a crash with auto-commit mid-stage (PR #21334), RocksDB may have stale data beyond the MDBX checkpoint. Previous approach scanned ALL entries to detect this.

## Solution

1. **Fast path**: `checkpoint == 0` + has data → clear entire table
2. **Slow path**: iterate changesets in `(checkpoint, sf_tip]` batches of 10k, prune via `unwind_*_history_indices()`

## Why not a RocksDB watermark?

History index stages sort data by address (for shard merging), not by block. During writes:
- Address 0x0001 touches blocks [5, 100, 5000]
- Address 0x0002 touches blocks [3, 7, 50000]
- Address 0xFFFF touches blocks [10, 20]

At any point mid-write, there's no consistent block boundary—we've touched a mix of blocks across addresses. A watermark like "committed up to block X" is meaningless because unwritten addresses may reference blocks < X.

## Changes

| Before | After |
|--------|-------|
| `check_accounts_history`: O(n) scan | O(changesets in stale range) |
| `check_storages_history`: O(n) scan | O(changesets in stale range) |
| `prune_*_history_above` functions | Removed (use unwind instead) |

## Tests

18 tests including 15k-block batching tests.

Closes #21313, #21294, #21342

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -24,118 +24,6 @@ use std::collections::HashSet;
 /// Balances memory usage against iteration overhead.
 const HEAL_HISTORY_BATCH_SIZE: u64 = 10_000;
 
-/// Generates a `heal_*_history` method for a history table.
-///
-/// This macro deduplicates the common healing logic between `StoragesHistory` and
-/// `AccountsHistory`, which share identical `checkpoint/sf_tip` handling and batch loop structure.
-macro_rules! impl_heal_history {
-    (
-        $(#[$attr:meta])*
-        fn $fn_name:ident < $provider:ident : $($bound:path),+ $(,)? > (
-            table: $table:ty,
-            stage_id: $stage_id:expr,
-            segment: $segment:expr,
-            table_name: $table_name:literal,
-            get_changesets: |$prov:ident, $range:ident| $get_changesets:expr,
-            extract_indices: |$changesets:ident, $checkpoint:ident| $extract_indices:expr,
-            unwind: |$self:ident, $indices:ident| $unwind:expr $(,)?
-        )
-    ) => {
-        $(#[$attr])*
-        fn $fn_name<$provider>(
-            &self,
-            provider: &$provider,
-        ) -> ProviderResult<Option<BlockNumber>>
-        where
-            $provider: $($bound +)+,
-        {
-            let checkpoint = provider
-                .get_stage_checkpoint($stage_id)?
-                .map(|cp| cp.block_number)
-                .unwrap_or(0);
-
-            // Fast path: if checkpoint is 0 and RocksDB has data, clear everything.
-            if checkpoint == 0 && self.first::<$table>()?.is_some() {
-                tracing::info!(
-                    target: "reth::providers::rocksdb",
-                    concat!($table_name, " has data but checkpoint is 0, clearing all")
-                );
-                self.clear::<$table>()?;
-                return Ok(None);
-            }
-
-            let sf_tip = provider
-                .static_file_provider()
-                .get_highest_static_file_block($segment)
-                .unwrap_or(0);
-
-            if sf_tip < checkpoint {
-                // This should never happen in normal operation - static files are always
-                // committed before RocksDB. If we get here, something is seriously wrong.
-                // The unwind is a best-effort attempt but is probably futile.
-                tracing::warn!(
-                    target: "reth::providers::rocksdb",
-                    sf_tip,
-                    checkpoint,
-                    concat!($table_name, ": static file tip behind checkpoint, unwind needed")
-                );
-                return Ok(Some(sf_tip));
-            }
-
-            if sf_tip == checkpoint {
-                return Ok(None);
-            }
-
-            let total_blocks = sf_tip - checkpoint;
-            tracing::info!(
-                target: "reth::providers::rocksdb",
-                checkpoint,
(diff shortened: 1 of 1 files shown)


gakonst (dismissed):
-

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -14,9 +14,127 @@ use reth_db_api::{tables, transaction::DbTx};
...
+                // This should never happen in normal operation - static files are always
+                // committed before RocksDB. If we get here, something is seriously wrong.
+                // The unwind is a best-effort attempt but is probably futile.
+                tracing::warn!(
+                    target: "reth::providers::rocksdb",
+                    sf_tip,
+                    checkpoint,
+                    concat!($table_name, ": static file tip behind checkpoint, unwind needed")
+                );
+                return Ok(Some(sf_tip));
```
joshieDo: unsure if we should just crash here

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -14,9 +14,127 @@ use reth_db_api::{tables, transaction::DbTx};
...
-    DBProvider, StageCheckpointReader, StorageSettingsCache, TransactionsProvider,
+    ChangeSetReader, DBProvider, StageCheckpointReader, StorageChangeSetReader,
+    StorageSettingsCache, TransactionsProvider,
 };
 use reth_storage_errors::provider::ProviderResult;
+use std::collections::HashSet;
+
+/// Batch size for changeset iteration during history healing.
+/// Balances memory usage against iteration overhead.
+const HEAL_HISTORY_BATCH_SIZE: u64 = 10_000;
```
joshieDo, line 25: maybe it can be bigger? this affects how many changesets we hold in memory to build indices from

yongkangc: would this affect perf (non persitence but more of engine)?

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -221,260 +227,202 @@ impl RocksDBProvider {
...
+            return Ok(Some(sf_tip));
+        }
 
-                // If all entries are sentinel entries (u64::MAX), treat as first-run scenario.
-                // This means no completed shards exist (only sentinel shards with
-                // highest_block_number=u64::MAX), so no actual history has been indexed.
-                if !found_non_sentinel {
-                    return Ok(None);
-                }
+        if sf_tip == checkpoint {
```
yongkangc, line 369: should we have this as the first case? for early exit

yongkangc (approved):
left some comments, but overall lgtm

crates/storage/provider/src/providers/rocksdb/invariants.rs
yongkangc: same for above also

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -221,260 +227,202 @@ impl RocksDBProvider {
...
-                    let (key, _) = result?;
-                    let highest = key.sharded_key.highest_block_number;
-                    if highest != u64::MAX {
-                        found_non_sentinel = true;
-                        if highest > max_highest_block {
-                            max_highest_block = highest;
-                        }
-                    }
-                }
+        if sf_tip < checkpoint {
```
yongkangc, line 262: should we have a comment for these cases similar to txnhashnumbers?

Image

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -221,260 +227,202 @@ impl RocksDBProvider {
...
+            return Ok(Some(sf_tip));
+        }
 
-                // If all entries are sentinel entries (u64::MAX), treat as first-run scenario.
-                // This means no completed shards exist (only sentinel shards with
-                // highest_block_number=u64::MAX), so no actual history has been indexed.
-                if !found_non_sentinel {
-                    return Ok(None);
-                }
+        if sf_tip == checkpoint {
```
yongkangc, line 275: should we have a comment for these cases similar to txnhashnumbers?

Image

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -221,260 +227,202 @@ impl RocksDBProvider {
...
-
-        let mut to_delete: Vec<StorageShardedKey> = Vec::new();
-        for result in self.iter::<tables::StoragesHistory>()? {
-            let (key, _) = result?;
-            let highest_block = key.sharded_key.highest_block_number;
-            if max_block == 0 || (highest_block != u64::MAX && highest_block > max_block) {
-                to_delete.push(key);
-            }
-        }
+        while batch_start <= sf_tip {
```
yongkangc, line 292: should we have a comment for these cases similar to txnhashnumbers?

Image

crates/storage/provider/src/providers/rocksdb/invariants.rs
joshieDo: this is only run during boot-up

Issue #20384, opened 2025-12-15
Wire up RocksDB for TransactionHashNumbers, StoragesHistory, and AccountsHistory tables


fgimenez:
Enable actual usage of RocksDB for storing `TransactionHashNumbers`, `StoragesHistory`, and `AccountsHistory` tables.

Issue #19702, opened 2025-11-13
TransactionHashNumbers Insert Performance Degrades Rapidly with Scale


duyquang6:
## Summary

When attempting to sustain high transaction throughput (~70k TPS) on our devnet with **1-second block time** for approximately 5 minutes, block persistence performance degrades rapidly. The primary bottleneck is identified as inserting records into the `TransactionHashNumbers` table, which uses random hash keys (B256) causing expensive random I/O operations in MDBX's B-tree structure.

**Critical Impact:**
- `TransactionHashNumbers` insertion consumes a significant portion of the block persistence time, which make persistence task often take more than 1 second per block to just persist on high-end hardware
- The random key inserts cause **massive freelist growth** (10x increase), indicating severe page fragmentation
- Cascading delay problem: Under spammer load (~70k TPS), block flushing delays increase from 3 blocks to 6 blocks, and persistence time grows from ~1s to >1s per block, creating a feedback loop that worsens over time. Its is consequence of increasing time on save block & tx commit
- The random key inserts cause severe page fragmentation, evidenced by massive freelist growth:

Dirty idea to isolate issue: Remove `TransactionHashNumbers` out of block persistent to verify

With `TransactionHashNumbers` writes enabled (5-minute spammer test):
- Freelist size growth upto ~250,000 pages
- Indicates frequent page splits and rebalancing
- High fragmentation from random insert patterns -> make commit time more heavy under stress test

With `TransactionHashNumbers` writes disabled:
- Freelist size: ~30,000 pages
- 10x reduction in freelist size
- Reduce persist time 6s -> 2s in bench window 5min

Details:
Freelist before:
<img width="1549" height="299" alt="Image" src="https://github.com/user-attachments/assets/f399f860-a4bd-4cf7-9f3b-c93e54fd0354" />

Freelist after remove insert `TransactionHashNumbers`
<img width="1523" height="381" alt="Image" src="https://github.com/user-attachments/assets/118f5519-b38c-480a-a12a-03b88160fe9a" />

Persistent time before:

<img width="1707" height="413" alt="Image" src="https://github.com/user-attachments/assets/0629e215-0261-451e-86d6-75c0391aceff" />

With default config -> there is 3 blocks flush persistent per interval
At the end of the chart it need persist 6 blocks because persistent cannot keep up with speed of block creation -> so need 6s

Persistent time after removal:

[...]

PR #21343, opened 2026-01-22
feat(rocksdb): static file based healing for TransactionHashNumbers

joshieDo:
## Summary

Replaces MDBX-based detection with static file tip comparison for TransactionHashNumbers healing.

Closes #21313

Related to #20384, #19702

## Problem

After a crash with auto-commit mid-stage, RocksDB may have stale txhash entries beyond the TransactionLookup checkpoint. Previous approach relied on MDBX TransactionBlocks table which could be out of sync with static files.

## Solution

1. **Fast path**: `checkpoint == 0` + has data → clear entire table
2. **Unwind path**: `sf_tip < checkpoint` → return unwind target (shouldn't happen normally)
3. **Heal path**: iterate tx ranges in `(checkpoint, sf_tip]` batches of 10k, prune via hash computation from static files

## Why not a RocksDB watermark?

During batch writes, txhashes from different blocks are written together. If a crash occurs mid-write, there's no clean block boundary:

```
RocksDB writes in progress:
- 0xaaa1... (block 100)
- 0xaaa2... (block 200)
- 0xbbb1... (block 101)
- 0xbbb2... (block 101)
- 0xccc1... (block 102)  ← crash here
- 0xccc2... (block 102)
- 0xddd1... (block 103)
```

After crash: RocksDB has txhashes for blocks 100, 101, and partial 102. A watermark like "committed up to block X" is meaningless—we've written hashes from multiple blocks with no atomic boundary. Static files are the source of truth for what was fully written.

## Changes

| Before | After |
|--------|-------|
| `check_transaction_hash_numbers` | `heal_transaction_hash_numbers` |
| Return type: `Option<BlockNumber>` based on MDBX | Return type: `Option<BlockNumber>` based on sf_tip |
| Used MDBX TransactionBlocks for end tx | Uses `get_highest_static_file_tx(Transactions)` |
| Used MDBX BlockBodyIndices for sf_tip tx | Uses `block_body_indices(checkpoint)` for start tx |

## Tests

19 tests pass including checkpoint=0 clearing and sf_tip < checkpoint unwind tests.

Inspired by #21341


yongkangc (approved):
Makes sense where we should do crash recovery for tx hash lookups in RocksDB by using static files (not MDBX) as the source of truth

PR #21345, opened 2026-01-22
feat: allow setting custom debug block provider

### LINKED ISSUE #20327: Make blockstream in debugclient configurable

klkvr:
Closes https://github.com/paradigmxyz/reth/issues/20327

Adds `with_debug_block_provider` to `DebugNodeLauncherFuture` allowing custom node implementations to specify custom debug provider that takes precedence over etherscan/rpc.

### DIFF
--- a/crates/node/builder/src/launch/debug.rs
+++ b/crates/node/builder/src/launch/debug.rs
@@ -4,10 +4,13 @@ use alloy_consensus::transaction::Either;
 use alloy_provider::network::AnyNetwork;
 use jsonrpsee::core::{DeserializeOwned, Serialize};
 use reth_chainspec::EthChainSpec;
-use reth_consensus_debug_client::{DebugConsensusClient, EtherscanBlockProvider, RpcBlockProvider};
+use reth_consensus_debug_client::{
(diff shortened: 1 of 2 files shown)


mattsse (approved):
cool

PR #21346, opened 2026-01-22
fix(pruner): prune account and storage changeset static files

Rjected:
When account/storage changesets are stored in static files (rather than MDBX), the pruner now correctly reads from the static file provider and deletes the corresponding jars. Previously, pruning would silently skip these changesets since they weren't present in the database tables.

Adds an equivalent `prune` test for static files


crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -67,17 +161,17 @@ where
             return Ok(SegmentOutput::not_done(
                 limiter.interrupt_reason(),
                 input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
-            ))
+            ));
```
joshieDo:
```suggestion
            ))
```

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+        // Check where account changesets are stored
+        if EitherWriter::account_changesets_destination(provider).is_static_file() {
+            self.prune_static_files(provider, input, range, range_end)
+        } else {
+            self.prune_database(provider, input, range, range_end)
+        }
+    }
+}
+
+impl AccountHistory {
```
yongkangc, line 79: one question is whether rocksdb would make use of this impl as well under `AccountHistory`

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -58,11 +66,126 @@ where
...
+            progress,
+            pruned: pruned_changesets + outcomes.deleted,
+            checkpoint: Some(SegmentOutputCheckpoint {
+                block_number: Some(last_changeset_pruned_block),
+                tx_number: None,
+            }),
+        })
+    }
+
+    fn prune_database<Provider>(
```
yongkangc, line 156: i guess this is the legacy implementation? not sure if we wanna add a note `legacy` as doc string

crates/prune/prune/src/segments/user/account_history.rs
Rjected: I think yes it should

Rjected: I think yes it should

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -58,11 +66,126 @@ where
...
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        // The size of this map is limited by `prune_delete_limit * blocks_since_last_run /
+        // STORAGE_HISTORY_TABLES_TO_PRUNE`, and with current defaults it's usually `3500 * 5
+        // / 2`, so 8750 entries. Each entry is `160 bit + 256 bit + 64 bit`, so the total
+        // size should be up to ~0.5MB + some hashmap overhead. `blocks_since_last_run` is
+        // additionally limited by the `max_reorg_depth`, so no OOM is expected here.
+        let mut highest_deleted_storages: FxHashMap<_, _> = FxHashMap::default();
```
yongkangc:
```suggestion
        let mut highest_deleted_storages = FxHashMap::default();
```
could we?

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -58,11 +66,126 @@ where
...
+
+        Ok(SegmentOutput {
+            progress,
+            pruned: pruned_changesets + outcomes.deleted,
+            checkpoint: Some(SegmentOutputCheckpoint {
+                block_number: Some(last_changeset_pruned_block),
+                tx_number: None,
+            }),
+        })
+    }
```
yongkangc: the logic looks really similar as the db one including the comments - not sure if we wanna dedup them in the future. just a note, not this pr tho

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        // The size of this map it's limited by `prune_delete_limit * blocks_since_last_run /
+        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with current default it's usually `3500 * 5
+        // / 2`, so 8750 entries. Each entry is `160 bit + 256 bit + 64 bit`, so the total
+        // size should be up to 0.5MB + some hashmap overhead. `blocks_since_last_run` is
+        // additionally limited by the `max_reorg_depth`, so no OOM is expected here.
+        let mut highest_deleted_accounts: FxHashMap<_, _> = FxHashMap::default();
```
yongkangc:
```suggestion
        let mut highest_deleted_accounts = FxHashMap::default();
```

could we?

yongkangc (approved):
makes sense 

the logic of account, storage changesets looks realy smiliar to mdbx ones, not sure if we can dedup them in the futture

crates/prune/prune/src/segments/user/storage_history.rs
Rjected: could probably add a helper

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+
+        // Delete static file jars below the pruned block
+        if let Some(last_block) = last_changeset_pruned_block {
+            provider
+                .static_file_provider()
+                .delete_segment_below_block(StaticFileSegment::AccountChangeSets, last_block + 1)?;
+        }
+        trace!(target: "pruner", pruned = %pruned_changesets, %done, "Pruned account history (changesets from static files)");
+
+        let last_changeset_pruned_block = last_changeset_pruned_block
```
joshieDo: starting here and  below could be dedupped with the mdbx implementation. but unsure if then @yongkangc would even use it

joshieDo: same for storage history

joshieDo: oops, has been mentioned already

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -303,4 +421,152 @@ mod tests {
...
+                    .iter()
+                    .filter(|(key, _)| key.highest_block_number > last_pruned_block_number)
+                    .map(|(key, blocks)| {
+                        let new_blocks =
+                            blocks.iter().skip_while(|block| *block <= last_pruned_block_number);
+                        (key.clone(), BlockNumberList::new_pre_sorted(new_blocks))
+                    })
+                    .collect::<Vec<_>>();
+
+                assert_eq!(actual_shards, expected_shards);
```
gakonst, line 518: [MINOR] **Missing static file deletion verification**: The test verifies history index pruning but does not assert that static file changesets were actually deleted. Consider adding a check that the static file segment was pruned (e.g., verify the jar count or segment bounds).

**Action:** Should fix

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+        if limiter.is_limit_reached() {
+            return Ok(SegmentOutput::not_done(
+                limiter.interrupt_reason(),
+                input.previous_checkpoint.map(SegmentOutputCheckpoint::from_prune_checkpoint),
+            ))
+        }
+
+        // The size of this map it's limited by `prune_delete_limit * blocks_since_last_run /
+        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with current default it's usually `3500 * 5
+        // / 2`, so 8750 entries. Each entry is `160 bit + 256 bit + 64 bit`, so the total
```
gakonst: [MINOR] **Memory estimate comment is incorrect for account history**

The comment claims each entry is `160 bit + 256 bit + 64 bit` but the map is `FxHashMap<Address, BlockNumber>`:
- Key: `Address` = 160 bits (20 bytes)
- Value: `BlockNumber` = 64 bits (8 bytes)

This is ~28 bytes per entry, not 60 bytes. The comment appears copied from storage_history where the map key is `(Address, B256)` which correctly totals 60 bytes.

For 8750 entries: actual ~244KB vs claimed ~0.5MB. Not a functional issue but misleading documentation.

**Action:** Should fix (update comment to accurately reflect the data structure)

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -303,4 +421,152 @@ mod tests {
...
+            };
+
+        test_prune(
+            998,
+            1,
+            (PruneProgress::HasMoreData(PruneInterruptReason::DeletedEntriesLimitReached), 1000),
+        );
+        test_prune(998, 2, (PruneProgress::Finished, 1000));
+        test_prune(1400, 3, (PruneProgress::Finished, 804));
+    }
```
gakonst, line 541: [MINOR] **Missing empty changeset / no-op test**: Neither `prune_legacy()` nor `prune_static_file()` test the case where no changesets exist to prune (empty static file segment). This is an important edge case for the no-op path at line 61-64 in `prune()`.

**Action:** Should fix

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -306,14 +433,160 @@ mod tests {
...
+                .iter()
+                .filter(|(key, _)| key.sharded_key.highest_block_number > last_pruned_block_number)
+                .map(|(key, blocks)| {
+                    let new_blocks =
+                        blocks.iter().skip_while(|block| *block <= last_pruned_block_number);
+                    (key.clone(), BlockNumberList::new_pre_sorted(new_blocks))
+                })
+                .collect::<Vec<_>>();
+
+            assert_eq!(actual_shards, expected_shards);
```
gakonst, line 532: [MINOR] **Missing static file deletion verification**: Same issue as in account_history.rs - the test does not verify that the static file segment was actually pruned after `prune_static_files()`. Add an assertion on the static file provider's segment bounds or jar count.

**Action:** Should fix

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+            last_changeset_pruned_block = Some(block_number);
+            pruned_changesets += 1;
+            limiter.increment_deleted_entries_count();
+        }
+
+        // Delete static file jars below the pruned block
+        if let Some(last_block) = last_changeset_pruned_block {
+            provider
+                .static_file_provider()
+                .delete_segment_below_block(StaticFileSegment::AccountChangeSets, last_block + 1)?;
```
gakonst, line 131: [MAJOR] **Re-indexing overhead in delete_segment_below_block loop**

The `delete_segment_below_block` function calls `delete_jar` in a loop, and each `delete_jar` call invokes `self.initialize_index()?` which:
1. Acquires a write lock on indexes
2. Clears all indexes
3. Iterates through ALL static files on disk via `iter_static_files`
4. Rebuilds BTreeMaps for every segment
5. Clears the jar cache (`self.map.clear()`)

For pruning old data (potentially many jars), this is O(n²) where n = number of jars deleted. Each jar deletion re-scans the entire static file directory.

Consider either:
1. Batch the deletions and call `initialize_index()` once at the end
2. Incrementally update the index instead of full rebuild

**Action:** Should fix - this could cause significant slowdowns when pruning nodes with large history

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -306,14 +433,160 @@ mod tests {
...
+            );
+        };
+
+        test_prune(
+            998,
+            1,
+            (PruneProgress::HasMoreData(PruneInterruptReason::DeletedEntriesLimitReached), 500),
+        );
+        test_prune(998, 2, (PruneProgress::Finished, 500));
+        test_prune(1200, 3, (PruneProgress::Finished, 202));
```
gakonst, line 554: [NIT] **Missing idempotency test**: Consider adding a test case that runs prune with same parameters twice to verify idempotent behavior (second run should be a no-op with `pruned: 0`).

**Action:** Optional

crates/stages/stages/src/test_utils/test_db.rs
```diff
@@ -473,6 +473,51 @@ impl TestStageDB {
...
+            let block_number = offset + block as u64;
+
+            let mut account_changesets = Vec::new();
+            let mut storage_changesets = Vec::new();
+
+            for (address, old_account, old_storage) in changeset {
+                account_changesets.push(AccountBeforeTx { address, info: Some(old_account) });
+
+                for entry in old_storage {
+                    storage_changesets.push(StorageBeforeTx {
```
gakonst, line 503: [MINOR] **Data model difference from database insertion**: The database `insert_changesets()` inserts `StorageEntry` into `StorageChangeSets`, but this helper uses `StorageBeforeTx` with separate `address`, `key`, `value` fields. Ensure the static file format is correctly mirroring the database semantics - specifically that `StorageBeforeTx.address` matches the block-address key structure used in the database version.

**Action:** Should fix (verify equivalence)

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -303,4 +421,152 @@ mod tests {
...
+                        tx_number: None,
+                        prune_mode
+                    })
+                );
+            };
+
+        test_prune(
+            998,
+            1,
+            (PruneProgress::HasMoreData(PruneInterruptReason::DeletedEntriesLimitReached), 1000),
```
gakonst, line 537: [MINOR] **Prune exactly at threshold boundary not tested**: Both tests prune to block 998 (first run), then 1200/1400 (third run). Consider adding a test case that prunes exactly at `deleted_entries_limit` boundary (e.g., where the limit is reached exactly when finishing the target block) to verify checkpoint handling at precise boundaries.

**Action:** Should fix

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with current default it's usually `3500 * 5
+        // / 2`, so 8750 entries. Each entry is `160 bit + 256 bit + 64 bit`, so the total
+        // size should be up to 0.5MB + some hashmap overhead. `blocks_since_last_run` is
+        // additionally limited by the `max_reorg_depth`, so no OOM is expected here.
+        let mut highest_deleted_accounts: FxHashMap<_, _> = FxHashMap::default();
+        let mut last_changeset_pruned_block = None;
+        let mut pruned_changesets = 0;
+        let mut done = true;
+
+        let walker = StaticFileAccountChangesetWalker::new(provider, range);
```
gakonst, line 114: [MINOR] **Walker I/O pattern: block-by-block fetching**

The `StaticFileAccountChangesetWalker` (used at line 112) fetches changesets block-by-block, calling `account_block_changeset(current_block)` for each block. Each call:
1. Looks up the segment provider via `get_segment_provider_for_block`
2. Gets the changeset offset from the header
3. Creates a new cursor
4. Reads entries one-by-one

For sequential pruning of thousands of blocks within the same jar, this is inefficient compared to keeping a cursor open and iterating.

**Suggestion:** Consider a batch-reading approach that:
- Opens the jar once per static file range
- Iterates through all entries in the jar using a single cursor
- Only seeks when crossing jar boundaries

This is not a blocker for this PR since the walker already exists, but worth noting for future optimization.

**Action:** Optional

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -303,4 +421,152 @@ mod tests {
...
+                    .map(|(block_number, _)| {
+                        (if result.progress.is_finished() {
+                            *block_number
+                        } else {
+                            block_number.saturating_sub(1)
+                        }) as BlockNumber
+                    })
+                    .unwrap_or(to_block);
+
+                let actual_shards = db.table::<tables::AccountsHistory>().unwrap();
```
gakonst, line 506: [NIT] **Cross-backend parity**: The `prune_legacy()` test verifies `AccountChangeSets` table length after pruning (line 296-298), but `prune_static_file()` doesn't have an equivalent assertion for static file segment data count. For complete parity between tests, consider adding such a check.

**Action:** Optional

crates/prune/prune/src/segments/user/account_history.rs
gakonst: [NIT] **Consider pre-sizing FxHashMap**

The `FxHashMap` is created with default capacity. Given the comment explains the expected size (~8750 entries), consider pre-allocating:

```rust
let mut highest_deleted_accounts: FxHashMap<_, _> = 
    FxHashMap::with_capacity_and_hasher(8750, Default::default());
```

This avoids rehashing during insertion. FxHashMap rehashes at ~87% load factor, so without pre-sizing you'll get 3-4 rehash operations (at 14, 28, 56, 112, ..., 8192 buckets).

**Action:** Optional

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -58,11 +66,126 @@ where
...
+            last_changeset_pruned_block = Some(block_number);
+            pruned_changesets += 1;
+            limiter.increment_deleted_entries_count();
+        }
+
+        // Delete static file jars below the pruned block
+        if let Some(last_block) = last_changeset_pruned_block {
+            provider
+                .static_file_provider()
+                .delete_segment_below_block(StaticFileSegment::StorageChangeSets, last_block + 1)?;
```
gakonst, line 133: [MAJOR] **Same re-indexing overhead applies here**

Same issue as in account_history.rs - the `delete_segment_below_block` call at line 131 triggers `initialize_index()` for each jar deleted, resulting in O(n²) complexity.

**Action:** Should fix

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+            // If there's more account changesets to prune, set the checkpoint block number to
+            // previous, so we could finish pruning its account changesets on the next run.
+            .map(|block_number| if done { block_number } else { block_number.saturating_sub(1) })
+            .unwrap_or(range_end);
+
+        // Sort highest deleted block numbers by account address and turn them into sharded keys.
+        // We did not use `BTreeMap` from the beginning, because it's inefficient for hashes.
+        let highest_sharded_keys = highest_deleted_accounts
+            .into_iter()
+            .sorted_unstable() // Unstable is fine because no equal keys exist in the map
```
gakonst: [NIT] **Sorting is necessary and correctly done**

The `.sorted_unstable()` call is necessary for the database cursor to seek efficiently. The history indices table is keyed by `(Address, BlockNumber)` in sorted order, so providing sorted keys allows sequential cursor iteration without random seeks.

The comment correctly explains why `BTreeMap` wasn't used from the start (hash comparison is cheaper for inserts). With 8750 entries, the sort is O(n log n) ≈ 115K comparisons, which is fast.

**No action needed** - this is correct as-is.

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+            .map(|(address, block_number)| {
+                ShardedKey::new(address, block_number.min(last_changeset_pruned_block))
+            });
+        let outcomes = prune_history_indices::<Provider, tables::AccountsHistory, _>(
```
gakonst: [NIT] **Clone in prune_history_indices seek**

At line 43 of history.rs (not in this PR's diff, but called by this code), there's:
```rust
cursor.seek(RawKey::new(sharded_key.clone()))?;
```

The `sharded_key.clone()` is used because the same key is accessed on line 46. For `ShardedKey<Address>` this clones 28 bytes, and for `StorageShardedKey` it clones 60 bytes. This happens 8750 times per prune run.

Consider restructuring to avoid the clone by extracting `to_block` first:
```rust
let to_block = sharded_key.as_ref().highest_block_number;
let mut shard = cursor.seek(RawKey::new(sharded_key))?;
```

**Action:** Optional - this is in existing code, not introduced by this PR

crates/prune/prune/src/segments/user/account_history.rs
gakonst, line 131: [MAJOR] **Crash Consistency Issue**: Static files are deleted via `delete_segment_below_block()` before `prune_history_indices()` completes and before the checkpoint is committed to the database.

If a crash occurs:
1. After static file jar deletion (line 129)
2. But before the database transaction commits (which saves both the history index updates AND the checkpoint)

Then on restart:
- The checkpoint will indicate pruning should resume from a certain block
- But the static file jars containing those blocks may already be deleted
- The walker will return empty results for those blocks
- History indices (`AccountsHistory`) will NOT be pruned for data that no longer exists
- **Result**: Dangling history index entries pointing to non-existent changesets

This is the same atomicity issue that database pruning avoids by having all operations in the same MDBX transaction.

**Suggested fix**: Consider deferring static file deletion until AFTER the database transaction commits, or tracking pending deletions that can be reconciled on restart.

**Action:** Should fix

crates/prune/prune/src/segments/user/storage_history.rs
gakonst, line 133: [MAJOR] **Same crash consistency issue as AccountHistory**: Static file deletion happens before database transaction commits.

See the detailed comment on account_history.rs line 129 for the full analysis.

**Action:** Should fix

crates/prune/prune/src/segments/user/account_history.rs
gakonst, line 114: [MINOR] **Inconsistent walker construction pattern**: AccountHistory uses `StaticFileAccountChangesetWalker::new(provider, range)` directly while StorageHistory uses `provider.static_file_provider().walk_storage_changeset_range(range)`.

Both are functionally equivalent, but the inconsistency is confusing. Consider using the same pattern in both files - either:
- Both use the direct walker constructor, or
- Both use the `static_file_provider().walk_*` method

**Action:** Optional

gakonst, line 114: [NIT] **API Inconsistency**: This file creates the walker directly with `StaticFileAccountChangesetWalker::new(provider, range)`, passing the provider which must implement `ChangeSetReader`. However, in `storage_history.rs` line 112, the equivalent code uses `provider.static_file_provider().walk_storage_changeset_range(range)` instead.

Consider aligning these approaches for consistency. Either:
- Both use the walker constructor directly, or
- Both use the `static_file_provider().walk_*_changeset_range()` method

This works correctly as-is, but the inconsistency may cause confusion during future maintenance.

**Action:** Optional

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -58,11 +66,126 @@ where
...
+    /// Prunes storage history when changesets are stored in static files.
+    fn prune_static_files<Provider>(
+        &self,
+        provider: &Provider,
+        input: PruneInput,
+        range: std::ops::RangeInclusive<BlockNumber>,
+        range_end: BlockNumber,
+    ) -> Result<SegmentOutput, PrunerError>
+    where
+        Provider: DBProvider<Tx: DbTxMut> + StaticFileProviderFactory,
```
gakonst, line 89: [NIT] **Asymmetric trait bounds**: This function requires only `StaticFileProviderFactory` and not `StorageChangeSetReader`, because it uses `provider.static_file_provider().walk_storage_changeset_range(range)` where `StaticFileProvider` itself implements `StorageChangeSetReader`.

In contrast, `account_history.rs` line 87 requires `ChangeSetReader` because it passes the provider directly to `StaticFileAccountChangesetWalker::new(provider, range)`.

Both approaches work correctly, but having different trait bound patterns for analogous functions adds cognitive overhead. Consider aligning either:
- Both require the `*ChangeSetReader` trait directly, or  
- Both use only `StaticFileProviderFactory` and call methods on `static_file_provider()`

**Action:** Optional

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -3,12 +3,18 @@ use crate::{
     segments::{user::history::prune_history_indices, PruneInput, Segment},
     PrunerError,
 };
+use alloy_primitives::BlockNumber;
 use itertools::Itertools;
 use reth_db_api::{models::ShardedKey, tables, transaction::DbTxMut};
-use reth_provider::DBProvider;
+use reth_provider::{
+    changeset_walker::StaticFileAccountChangesetWalker, DBProvider, EitherWriter,
+    StaticFileProviderFactory, StorageSettingsCache,
```
gakonst, line 13: [NIT] **Import inconsistency**: `StorageSettingsCache` is imported from `reth_provider` here, but in `storage_history.rs` line 18, the same trait is imported from `reth_storage_api`. While both work (since `reth_provider` re-exports from `reth_storage_api`), consider using consistent import paths across related files.

**Action:** Optional

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -56,11 +65,120 @@ where
...
+            progress,
+            pruned: pruned_changesets + outcomes.deleted,
+            checkpoint: Some(SegmentOutputCheckpoint {
+                block_number: Some(last_changeset_pruned_block),
+                tx_number: None,
+            }),
+        })
+    }
+
+    fn prune_database<Provider>(
```
yongkangc, line 152: we might want to add a comment here indicating this is for mdbx since we have rocksdb pruning as well?

crates/prune/prune/src/segments/user/account_history.rs
Rjected: btw added reuse for this ptal

joshieDo (approved):
lgtm

PR #21353, opened 2026-01-23
feat(net): add memory-bounded channel for transaction events

mattsse:
## Summary

Implement a semaphore-based memory-bounded channel between `NetworkManager` and `TransactionsManager` to prevent unbounded memory growth when transaction events queue up faster than they can be processed.

## Problem

The current `to_transactions_manager` channel is unbounded, which means during heavy network activity (e.g., transaction announcement floods), messages can accumulate without limit, potentially consuming all available memory.

## Solution

Use a **byte-budget semaphore** pattern:
- Each message must "acquire" semaphore permits equal to its in-memory size before being sent
- Permits are held in a `Budgeted<T>` envelope until the message is dropped by the receiver
- When the budget (default 1GB) is exhausted, `try_send` returns `Full` and the event is dropped with a metric counter

### Key Changes

1. **`reth-metrics`**: Added `MemoryBoundedSender<T>` and `MemoryBoundedReceiver<T>` wrappers
   - `Budgeted<T>` envelope holds owned semaphore permits (RAII pattern)
   - Per-message overhead of 256 bytes added for envelope/permit costs
   
2. **`reth-eth-wire-types`**: Added `InMemorySize` implementations for:
   - `NewPooledTransactionHashes`
   - `GetPooledTransactions`

3. **`reth-network`**:
   - `NetworkTransactionEvent<N>` now implements `InMemorySize`
   - Control plane messages (`GetTransactionsHandle`) return size 0 to avoid blocking
   - `notify_tx_manager` uses `try_send` with backpressure - drops events when at capacity
   - Added `total_dropped_tx_events_at_full_capacity` metric
   - `TransactionsManager` receives `Budgeted<NetworkTransactionEvent<N>>` and unwraps

## Testing

All 119 existing tests in `reth-network` pass.

## Configuration

The memory limit is set to 1GB via:
```rust
const TX_MANAGER_CHANNEL_MEMORY_LIMIT_BYTES: usize = 1024 * 1024 * 1024; // 1GB
```

### DIFF
--- a/crates/metrics/src/common/mpsc.rs
+++ b/crates/metrics/src/common/mpsc.rs
@@ -4,8 +4,13 @@
 use crate::Metrics;
 use futures::Stream;
 use metrics::Counter;
+use reth_primitives_traits::InMemorySize;
 use std::{
     pin::Pin,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
     task::{ready, Context, Poll},
 };
 use tokio::sync::mpsc::{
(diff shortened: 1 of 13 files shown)


crates/metrics/src/common/mpsc.rs
```diff
@@ -399,3 +404,118 @@ struct MeteredPollSenderMetrics {
...
+impl<T> MemoryBoundedReceiver<T> {
+    /// Receives the next message, returning `None` if the channel is closed.
+    pub async fn recv(&mut self) -> Option<Budgeted<T>> {
+        self.inner.recv().await
+    }
+
+    /// Polls to receive the next message on this channel.
+    pub fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Option<Budgeted<T>>> {
+        self.inner.poll_recv(cx)
+    }
```
klkvr, line 495: should this return the unwrapped `T`? i guess it's correct to expect all consumers to always unwrap it immediately

crates/net/network/src/transactions/mod.rs
```diff
@@ -1608,7 +1603,7 @@ where
             "Network transaction events stream",
             DEFAULT_BUDGET_TRY_DRAIN_NETWORK_TRANSACTION_EVENTS,
             this.transaction_events.poll_next_unpin(cx),
-            |event| this.on_network_tx_event(event),
+            |budgeted: Budgeted<NetworkTransactionEvent<N>>| this.on_network_tx_event(budgeted.msg),
```
klkvr, line 1606: i guess for this to work properly we now need some kind of backpressure here, otherwise we'd just end up OOM'ing on tx manager side

not sure how to approach this given different kinds of messages are arriving here

klkvr (approved):
this lgtm and is a useful primitive, though still not sure what's the best approach on the tx manager end

PR #21357, opened 2026-01-23
feat(p2p): add `reth p2p enode` command

gakonst:
Add a new subcommand that prints the enode identifier for a given discovery secret key file.

## Usage

```bash
reth p2p enode --discovery-secret /path/to/key
reth p2p enode --discovery-secret /path/to/key --ip 192.168.1.1
```

## Arguments

- `--discovery-secret` (required): path to the secret key file
- `--ip` (optional): IP address to include in the enode URL (defaults to 0.0.0.0)

## Example Output

```
enode://6f8a80d14311c39f35f516fa664deaaaa13e85b2f7493f37f6144d86991ec012937307647bd3b9a82abe2974e1407241d54947bbb39763a4cac9f77166ad92a0@0.0.0.0:30303
```


crates/cli/commands/src/p2p/enode.rs
```diff
@@ -0,0 +1,35 @@
...
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    path::PathBuf,
+};
+
+/// Print the enode identifier for a given secret key.
+#[derive(Parser, Debug)]
+pub struct Command {
+    /// Path to the secret key file for discovery.
+    #[arg(long, value_name = "PATH", required = true)]
+    pub discovery_secret: PathBuf,
```
mattsse, line 15: I don't think we need this as named, imo it would be more convenient if this would work like

```
reth enode <path> --ip <ip>
```

perhaps even drop the named --ip as well?

mattsse (approved):

PR #21359, opened 2026-01-23
feat(txpool): add Block associated type to TransactionValidator trait

klkvr:
Added `Block` associated type to the `TransactionValidator` and `TransactionPoolExt` traits, making `on_new_head_block` aware of the block header type instead of receiving a generic block.

This would allow to later give `ConfigureEvm` access to the validator which requires concrete headers in its API.

### DIFF
--- a/crates/optimism/txpool/src/transaction.rs
+++ b/crates/optimism/txpool/src/transaction.rs
@@ -325,10 +325,11 @@ mod tests {
     #[tokio::test]
     async fn validate_optimism_transaction() {
         let client = MockEthProvider::default().with_chain_spec(OP_MAINNET.clone());
-        let validator = EthTransactionValidatorBuilder::new(client)
-            .no_shanghai()
-            .no_cancun()
-            .build(InMemoryBlobStore::default());
+        let validator =
+            EthTransactionValidatorBuilder::new(client)
+                .no_shanghai()
+                .no_cancun()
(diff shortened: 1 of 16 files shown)


crates/optimism/txpool/src/transaction.rs
```diff
@@ -325,10 +325,11 @@ mod tests {
...
         let client = MockEthProvider::default().with_chain_spec(OP_MAINNET.clone());
-        let validator = EthTransactionValidatorBuilder::new(client)
-            .no_shanghai()
-            .no_cancun()
-            .build(InMemoryBlobStore::default());
+        let validator =
+            EthTransactionValidatorBuilder::new(client)
+                .no_shanghai()
+                .no_cancun()
+                .build::<_, _, reth_optimism_primitives::OpBlock>(InMemoryBlobStore::default());
```
mattsse, line 332: this is a bit unfortunate but also something we can live with

crates/transaction-pool/src/pool/mod.rs
```diff
@@ -504,10 +503,7 @@ where
     }
 
     /// Updates the entire pool after a new block was executed.
-    pub fn on_canonical_state_change<B>(&self, update: CanonicalStateUpdate<'_, B>)
-    where
-        B: Block,
-    {
+    pub fn on_canonical_state_change(&self, update: CanonicalStateUpdate<'_, V::Block>) {
```
mattsse, line 506: this makes sense

mattsse (approved):

Issue #7820, opened 2024-04-23
Add `Header` fields while maintaining backwards compatibility


joshieDo:
### Describe the feature

With [Prague](https://github.com/paradigmxyz/reth/issues/7363) coming up, we should take this into account.

Generally with `Compact`, the reader first reads a bitflag struct (eg. `HeaderFlags`) and uses that to read the number of bytes necessary for each field.

As long as there are bits available on this bitflag struct, we can keep adding new fields to `Header` under the form of `Option<FIELD>` while maintaing backwards compatability. Each Option<> uses one bit in the bitflag struct. However, we have reached the limit, there's only one bit left.

```rust
#[doc =
        "Fieldset that facilitates compacting the parent type. Used bytes: 4 | Unused bits: 1"]
        #[allow(clippy :: identity_op)]
        pub struct HeaderFlags {
```

The next `Option<N>` has to be its own Header extension struct. N can be whatever we want it to be.
Previous instances of Header will read that bit flag as 0 and ignore that field, and the new Header changes will read that bit as one and decode it

PR #21362, opened 2026-01-23
feat: add ReceiptExt for backwards-compatible receipt field extensions

gakonst:
## Summary

Adds `ReceiptExt` struct following the [`HeaderExt` pattern](https://github.com/paradigmxyz/reth/blob/main/crates/storage/codecs/src/alloy/header.rs#L48-L76) to enable adding new receipt fields (e.g., for Glamsterdam hardfork) without breaking backwards compatibility.

## Problem

The `Receipt` compact encoding uses all 8 bits in its flags byte (0 unused bits):
- `tx_type_len`: 2 bits
- `success_len`: 1 bit
- `cumulative_gas_used_len`: 4 bits
- `__zstd`: 1 bit

This means new fields cannot be added directly to Receipt without a breaking change.

## Solution

Following the established pattern from `HeaderExt` ([#7820](https://github.com/paradigmxyz/reth/issues/7820)), this PR adds:
- `ReceiptExt` struct with `#[derive(Compact)]`
- `into_option()` helper for conditional serialization
- Backwards compatibility test for the bitflags
- **Comprehensive test vectors** generated from main branch to verify old receipts decode correctly

### Test Vectors

Test vectors were generated on `main` branch and verify:
- Legacy receipt (small, no logs)
- EIP-1559 receipt (no logs)  
- EIP-4844 failed receipt (success=false)
- Receipt with logs (zstd compressed)
- All tx types roundtrip

## Next Steps

When the glam field is known:
1. Add the field as `Option<T>` to `ReceiptExt`
2. Update `ReceiptExt::into_option()` to check the new field
3. Wire `extra_fields: Option<ReceiptExt>` into the Compact impl
4. Update the test to expect `UnusedBits::NotZero`

## For OP Receipts

`CompactOpReceipt` already has unused bits (`UnusedBits::NotZero`), so glam fields can be added directly there as `Option<T>` or use a shared `ReceiptExt` for consistency.

cc @mattsse

### DIFF
--- a/crates/ethereum/primitives/src/receipt.rs
+++ b/crates/ethereum/primitives/src/receipt.rs
@@ -490,8 +490,7 @@ mod compact {
     /// All new fields should be added here in the form of a `Option<T>`, since `Option<ReceiptExt>`
     /// is used as an extension field for backwards compatibility.
     ///
-    /// More information: <https://github.com/paradigmxyz/reth/issues/7820> &
-    /// [`reth_codecs_derive::Compact`].
+    /// More information: <https://github.com/paradigmxyz/reth/issues/7820>
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
     #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this is on hold

PR #21364, opened 2026-01-23
feat(cli): add `reth db list rocksdb` subcommand

HuberyJulianay:
Adds support for listing RocksDB table contents via `reth db list rocksdb`, matching the existing MDBX functionality.

Usage:
```
  reth db list rocksdb transaction-hash-numbers --len 10
  reth db list rocksdb accounts-history --search 0xdeadbeef --json
  reth db list rocksdb storages-history --count
```

Supports the same filtering options as MDBX (skip, len, reverse, search, min sizes). This fills the tooling gap - we already have `reth db checksum rocksdb` from #21217 but had no way to actually inspect the data.

Implementation restructures the list command to use subcommands (mdbx/rocksdb) instead of a single table argument.


HuberyJulianay:
@mattsse Hi boss, could this s-stale label be removed? cc @joshieDo

PR #21365, opened 2026-01-23
feat(cli): add `reth db get rocksdb` subcommand

HuberyJulianay:
Adds RocksDB support to the `reth db get` command, completing the tooling for RocksDB tables alongside MDBX and static files.

Usage:
```
  reth db get rocksdb transaction-hash-numbers '"0xabc..."'
  reth db get rocksdb accounts-history '{ "key": "0x...", "highest_block_number": 12345 }'
  reth db get rocksdb storages-history '...' --raw
```

This was the last missing piece - we now have checksum (#21217), list(#21364), and get(this) for RocksDB. Users can finally look up specific entries without iterating through the entire table.

### DIFF
--- a/crates/cli/commands/src/db/checksum/mod.rs
+++ b/crates/cli/commands/src/db/checksum/mod.rs
@@ -24,6 +24,8 @@ use tracing::{info, warn};
 
 #[cfg(all(unix, feature = "edge"))]
 mod rocksdb;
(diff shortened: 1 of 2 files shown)


HuberyJulianay:
@mattsse Hi boss, could this s-stale label be removed? cc @joshieDo

PR #21373, opened 2026-01-23
perf(chain-state): move sorting outside write lock in remove_persiste…

ethfanWilliam:
Moves the expensive sort operation outside the write lock in remove_persisted_blocks to reduce lock contention.

Currently holding the write lock while draining, filtering, cloning, and sorting all blocks which can take many unpersisted blocks during sync. This blocks all readers and writers.

Changed to use a read lock for collecting blocks, sort outside the lock, then acquire write lock only for the fast clear/insert operations. Should reduce write lock hold time by 60-80% in high-load sync scenarios.

The lock ordering comment only requires acquiring numbers lock before blocks lock, doesn't mandate holding locks during sort.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -329,26 +329,31 @@ impl<N: NodePrimitives> CanonicalInMemoryState<N> {
             }
         }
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this unfortunately does gives us anythign because we still have to do the same work regardless

because this is called from the engine context we want to acquire the write lock here

PR #21383, opened 2026-01-23
fix(engine): use LazyTrieData::deferred for chain notification

Rjected:
we did not actually ever use `LazyTrieData::deferred`, so when we called `hashed_state` it would lock and we would spend time in fcu waiting for the hashed state, rather than returning immediately.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -951,29 +951,29 @@ impl<N: NodePrimitives<SignedTx: SignedTransaction>> NewCanonicalChain<N> {
         match blocks {
             [] => Chain::default(),
             [first, rest @ ..] => {
-                let first = first.clone();
+                let trie_data_handle = first.trie_data_handle();
                 let mut chain = Chain::from_block(
                     first.recovered_block().clone(),
                     ExecutionOutcome::from((
                         first.execution_outcome().clone(),
                         first.block_number(),
                     )),
-                    LazyTrieData::deferred(move || SortedTrieData {
-                        hashed_state: first.hashed_state(),
-                        trie_updates: first.trie_updates(),
+                    LazyTrieData::deferred(move || {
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/chain-state/src/in_memory.rs
```diff
@@ -948,22 +948,34 @@ impl<N: NodePrimitives<SignedTx: SignedTransaction>> NewCanonicalChain<N> {
...
-                    LazyTrieData::ready(first.hashed_state(), first.trie_updates()),
+                    LazyTrieData::deferred(move || {
+                        SortedTrieData {
+                            hashed_state: first.hashed_state(),
+                            trie_updates: first.trie_updates(),
+                        }
+                    }),
                 );
                 for exec in rest {
+                    let exec = exec.clone();
```
mattsse: we can replace this with 

https://github.com/paradigmxyz/reth/blob/e2a9f4ca57a71ff1cbbd24a3e8a3c34ef630c2fe/crates/chain-state/src/in_memory.rs#L858-L858

which is slightly cheaper than cloning a few arcs

crates/chain-state/src/in_memory.rs
```diff
@@ -948,22 +948,34 @@ impl<N: NodePrimitives<SignedTx: SignedTransaction>> NewCanonicalChain<N> {
         match blocks {
             [] => Chain::default(),
             [first, rest @ ..] => {
+                let first = first.clone();
```
mattsse: same here

mattsse (changes requested):
cool

smol nits

crates/chain-state/src/in_memory.rs
Rjected: ah true nice

mattsse (approved):

PR #21384, opened 2026-01-24
fix(docker): pass vergen git vars as build args

mattsse:
## Summary
The Docker build excludes `.git` directory for cache efficiency (`COPY --exclude=.git`), which causes vergen to fall back to `VERGEN_IDEMPOTENT_OUTPUT` instead of the actual git SHA.

Current nightly images show:
```
Commit SHA: VERGEN_IDEMPOTENT_OUTPUT
```

## Fix
Pass `VERGEN_GIT_SHA`, `VERGEN_GIT_DESCRIBE`, and `VERGEN_GIT_DIRTY` as build args from the CI workflow where git info is available. The vergen library reads these from environment variables as fallback when `.git` is not present.

## Changes
- `docker-bake.hcl`: Add variables for git info and pass them to build args
- `Dockerfile.depot`: Accept git info as ARGs and set them as ENVs for the build
- `.github/workflows/docker.yml`: Compute git info and export as env vars for bake

## Testing
Verified with `docker buildx bake --print` that variables are properly passed through.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -50,6 +50,13 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Get git info for vergen
+        id: git
+        run: |
+          echo "sha=${{ github.sha }}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


gakonst (approved):
good catch ty

PR #21386, opened 2026-01-24
fix: prevent panic in StaticFileProvider watcher thread

letmehateu:
The background directory watcher in StaticFileProvider::watch_directory could previously panic if watcher creation failed, the watch path was invalid, or file_stem() returned None. Since this thread runs in the background for read-only providers to detect static file changes, a panic causes the index to silently stop updating and users won't see new data.


mattsse (commented):
we can keep this

PR #21387, opened 2026-01-24
feat: change from stable sort to unstable sort

cuiweixie:


mattsse (approved):
we can do this because values are unique

PR #21388, opened 2026-01-24
feat: avoid overflow

cuiweixie:


mattsse (commented):
these are usually gas numbers and we dont expect overflows here

PR #21389, opened 2026-01-24
feat: prealloc vec

cuiweixie:


mattsse (commented):
vec collect already does this internally because specialized

PR #21390, opened 2026-01-24
feat(static-file): add metrics tracking to StaticFileProducer

sconeman:
## Summary

- Add histogram metrics to track StaticFileProducer durations
- Resolve TODO comments from **alexey** in `static_file_producer.rs`

## Test plan

- `cargo check -p reth-static-file`
- `cargo +nightly fmt --all`
- `cargo clippy -p reth-static-file --all-features`
- `cargo test -p reth-static-file`

### DIFF
--- a/crates/static-file/static-file/src/metrics.rs
+++ b/crates/static-file/static-file/src/metrics.rs
@@ -2,13 +2,17 @@
 
 use metrics::Histogram;
 use reth_metrics::Metrics;
+use reth_static_file_types::StaticFileSegment;
 
 /// Metrics for the static file producer.
 #[derive(Metrics, Clone)]
 #[metrics(scope = "static_file_producer")]
 pub(crate) struct StaticFileProducerMetrics {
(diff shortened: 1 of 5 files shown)


crates/static-file/static-file/src/metrics.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use metrics::Histogram;
+use reth_metrics::Metrics;
+
+/// Metrics for the static file producer.
+#[derive(Metrics, Clone)]
+#[metrics(scope = "static_file_producer")]
+pub(crate) struct StaticFileProducerMetrics {
+    /// Histogram of segment copy durations in seconds.
+    pub segment_duration_seconds: Histogram,
```
joshieDo: i think we might want it per segment? cc @shekhirin

sconeman: @joshieDo Done. Added a `"segment"` label to the `segment_duration_seconds` histogram so metrics are tracked per segment type.

PR #21391, opened 2026-01-24
feat: prealloc vec

cuiweixie:


mattsse (approved):
marginally meaningful so this doesnt hurt, must of the time this will be exactsize so this is better

PR #21394, opened 2026-01-24
fix: remove redundant `.clone()` in proptest code generation

viktorking7:
Removes an unnecessary `.clone()` call in the generated proptest code for `Compact` trait roundtrip tests. The `to_compact` method takes `&self`, so cloning the value before calling it is redundant and adds unnecessary performance overhead.


mattsse (commented):
we can keep this

PR #21395, opened 2026-01-24
fix: replace assert with proper error handling in stage dump setup

andrewshab3:
The setup function in stage dump used assert!(from < to, ...) which panics on invalid CLI input instead of returning an error. Additionally, the code performed unchecked from - 1 and to + 1 arithmetic which could underflow/overflow on edge cases.


mattsse (commented):
we can keep this

PR #21398, opened 2026-01-24
fix(engine): only warn for critical capability mismatches

mattsse:
Limit capability mismatch warnings to critical Engine API methods (`engine_forkchoiceUpdated`, `engine_getPayload`, `engine_newPayload`) that are essential for block production and chain synchronization.

Non-critical methods like `engine_getBlobs` are not warned about since not all clients support them and they're not required for operation.

This prevents spurious warnings when CL clients don't support optional methods like `engine_getBlobsV3`.


yongkangc (approved):
this makes sense

PR #21404, opened 2026-01-24
refactor(db): use slice instead of Vec reference in benchmark utils

marukai67:
Removes unnecessary `#[expect(clippy::ptr_arg)]` attribute by fixing the underlying issue.


mattsse (commented):
we can keep this as is

PR #21410, opened 2026-01-25
fix(prune): correct deleted entries count when skip_filter is used

Fallengirl:
Fixed a bug in `prune_table_with_range` where the returned `deleted_entries` count was incremented on every iteration, even when entries were skipped by the filter. This caused inflated counts in logs and statistics for segments using skip filters (e.g., `ReceiptsByLogs`).

The counter now only increments when an entry is actually deleted.


mattsse (commented):

PR #21411, opened 2026-01-25
fix(net): FetchFullBlockRangeFuture can get stuck forever after partial body fetch + error

akolishchak:
If a body range request fails after you already stored some bodies, no new request is started, so the future can stall with pending headers and no in‑flight request.

[`Err(err)` branch only logs](https://github.com/akolishchak/reth/blob/5440d0d89a544d5cd9300cf5ec9fb819087cfa63/crates/net/p2p/src/full_block.rs#L570-L572) and the retry block only runs when `this.bodies.is_empty()` once any bodies exist and a later request errors, `request.bodies` remains `None` and `is_bodies_complete()` stays false forever.

Fixed by checking `this.request.bodies.is_none()` as well as `!this.is_bodies_complete()` to target the actual problem: when an error or completion clears the in‑flight request and the range is not complete, we must kick off a new bodies request for the remaining hashes. That covers the "partial bodies + error" case which was missed.

The new test sets up a failing bodies client to reproduce the stall and then verifies the range completes:
> cargo test -p reth-network-p2p download_full_block_range_retries_after_body_error

### DIFF
--- a/crates/net/p2p/src/full_block.rs
+++ b/crates/net/p2p/src/full_block.rs
@@ -571,8 +571,8 @@ where
                             debug!(target: "downloaders", %err, ?this.start_hash, "Body range download failed");
                         }
                     }
-                    if this.bodies.is_empty() {
-                        // received bad response, re-request headers
+                    if this.request.bodies.is_none() && !this.is_bodies_complete() {
+                        // no pending bodies request (e.g., request error), retry remaining bodies
                         // TODO: convert this into two futures, one which is a headers range
                         // future, and one which is a bodies range future.
                         //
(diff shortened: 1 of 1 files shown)


crates/net/p2p/src/full_block.rs
```diff
@@ -571,8 +571,8 @@ where
                             debug!(target: "downloaders", %err, ?this.start_hash, "Body range download failed");
                         }
                     }
-                    if this.bodies.is_empty() {
-                        // received bad response, re-request headers
+                    if this.request.bodies.is_none() && !this.is_bodies_complete() {
+                        // no pending bodies request (e.g., request error), retry remaining bodies
```
mattsse, line 575: ah right, this now properly checks for completeness

mattsse (approved):
nice, tysm!

checking empty was indeed incorrect here for partial responses

PR #21414, opened 2026-01-25
feat(rpc): implement movePrecompileToAddress for eth_simulateV1

mattsse:
## Summary

This PR adds support for the `movePrecompileToAddress` state override in `eth_simulateV1`, allowing callers to move precompiles from their original addresses to new destinations.

## Dependencies

⚠️ **This PR depends on alloy-rs/alloy#3589** which adds the serde alias for `MovePrecompileToAddress` (the field uses PascalCase in the JSON-RPC spec but alloy expects camelCase).

### crates/rpc/rpc-eth-types/src/simulate.rs
- Add `NotAPrecompile(Address)` error variant with error code `-32000` (matching other clients)
- Add `apply_precompile_overrides()` function that:
  - Extracts `movePrecompileToAddress` entries from state overrides
  - Validates source addresses are actual precompiles before applying changes
  - Moves precompiles to new destinations, clearing the source address
  - Handles self-references (source == dest) as no-ops
  - Allows multiple precompiles to be moved to the same destination (per spec)

### crates/rpc/rpc-eth-api/src/helpers/call.rs
- Add `BlockBuilder` trait import to access `evm_mut()`
- Modified state override handling to keep a reference after applying to database
- Integrate `apply_precompile_overrides()` after EVM/builder creation, before transaction execution (in both inspector and non-inspector code paths)

## Test Cases Addressed

Based on hive test analysis:
| Test | Expected Behavior | Status |
|------|-------------------|--------|
| `ethSimulate-move-ecrecover-and-call-old-and-new` | Move precompile works at new address, original becomes empty | ✅ |
| `ethSimulate-move-ecrecover-and-call` | Precompile works at new address after move | ✅ |
| `ethSimulate-move-ecrecover-twice-and-call` | Sequential moves work correctly | ✅ |
| `ethSimulate-move-two-accounts-to-same-38023` | Multiple precompiles to same destination allowed | ✅ |
| `ethSimulate-move-to-address-itself-reference-38022` | Non-precompile returns error -32000 | ✅ |
| `ethSimulate-try-to-move-non-precompile` | Non-precompile returns error -32000 | ✅ |

## Related

- Depends on: alloy-rs/alloy#3589 (serde alias for `MovePrecompileToAddress`)

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -20,8 +20,8 @@ use alloy_rpc_types_eth::{
 use futures::Future;
 use reth_errors::{ProviderError, RethError};
 use reth_evm::{
-    env::BlockEnvironment, ConfigureEvm, Evm, EvmEnvFor, HaltReasonFor, InspectorFor,
-    TransactionEnv, TxEnvFor,
+    env::BlockEnvironment, execute::BlockBuilder, ConfigureEvm, Evm, EvmEnvFor, HaltReasonFor,
+    InspectorFor, TransactionEnv, TxEnvFor,
 };
 use reth_node_api::BlockBody;
 use reth_primitives_traits::Recovered;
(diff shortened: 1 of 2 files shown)


klkvr (approved):
cool

PR #21422, opened 2026-01-26
perf(db): flatten HashedPostState before persisting

### LINKED ISSUE #20609: Flatten HashedPostState before persisting

gakonst:
## Summary

Accumulate hashed state from all blocks using `HashedPostStateSorted::merge_batch` and write it in a single batch, reducing cursor open/close overhead from N calls to 1.

This mirrors the existing optimization for trie updates (which was already batched).

## Changes

- Move `write_hashed_state` call outside the per-block loop
- Use `HashedPostStateSorted::merge_batch` to combine all hashed states (newest-to-oldest)
- Write the merged hashed state in a single operation

Closes #20609

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -630,27 +630,27 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
                         },
                     )?;
                     timings.write_state += start.elapsed();
-
-                    let trie_data = block.trie_data();
-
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
nice, didnt know that reducing cursor overhead would be so big

mattsse (approved):

PR #21424, opened 2026-01-26
refactor(trie): reuse shared StorageProofCalculator for V2 sync storage roots and add deferred encoder metrics

### LINKED ISSUE #19515: Proof Calc Rewrite: Metrics and Traces

### LINKED ISSUE #21244: Proof Calc Rewrite: Better usage of storage proof workers
### Describe the feature

In the current implementation of V2 account proofs we do the following:

* Dispatch all known storage proofs to proof workers
* Begin account proof, dispatching proofs for storage roots which have not yet been dispatched to storage workers
* Block on storage workers waiting for storage roots

This ends up slowing down the pipeline quite a bit, I suspect because we are waiting for all storage workers to finish the proofs they originally dispatched prior to the account proof being able to continue.

There are a few potential solutions:
* Set up a separate pool of workers just for storage roots which are needed for account proofs but don't have actual storage proof targets
* Don't dispatch storage proofs ahead of time, dispatch them as they are come across in the account proof. This requires ensuring that all accounts for which a storage proof is required have an account target in the same MuliProofTargets.
* Use synchronous account proof handling (basically the same behavior as legacy)

The task is to figure out the best solution and implement it.

related to RETH-170

### Additional context

_No response_

mediocregopher:
After attempting to use the storage worker pool for all storage root calculations, it turns out that synchronous computation of storage roots which are not themselves proof targets performs better. I also messed around with creating a separate worker pool just for these storage roots, and that seemed promising but was taking too much time to optimize.

## Summary

This PR improves the V2 proof computation in the trie parallel module by:

- **No longer dispatching all storage roots to workers**; only storages with actual proof targets are dispatched, this matches the legacy behavior
- **Reusing a shared `StorageProofCalculator`** for synchronous storage root calculations, eliminating redundant cursor creation across multiple accounts
- **Spawning proof workers within a blocking task for each loop** so that account and storage workers can spin up at the same time
- **Adding detailed metrics** for the deferred value encoder to track how storage roots are resolved
- **Adding idle time metrics** for both legacy and v2 codepaths to track how long workers are blocked waiting on channels

Closes #21244
Closes #19515


crates/trie/parallel/src/proof_task.rs
```diff
@@ -93,16 +95,19 @@ pub struct ProofWorkerHandle {
     storage_work_tx: CrossbeamSender<StorageWorkerJob>,
     /// Direct sender to account worker pool
     account_work_tx: CrossbeamSender<AccountWorkerJob>,
+
```
mediocregopher: Remove these extra newlines

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1165,9 +1201,9 @@ where
     /// If this function panics, the worker thread terminates but other workers
     /// continue operating and the system degrades gracefully.
     fn run(mut self) -> ProviderResult<()> {
-        // Create provider from factory
-        let provider = self.task_ctx.factory.database_provider_ro()?;
-        let proof_tx = ProofTaskTx::new(provider, self.worker_id);
+        // Create provider from factory and wrap in Rc for sharing with AsyncAccountValueEncoder
+        let provider_rc: Rc<Factory::Provider> =
```
mediocregopher: Is this change still necessary? I think it's leftover from a previous draft

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -1,44 +1,110 @@
...
-pub(crate) enum AsyncAccountDeferredValueEncoder {
+pub(crate) enum AsyncAccountDeferredValueEncoder<TC, HC> {
+    /// A storage proof job was dispatched to the worker pool.
     Dispatched {
         hashed_address: B256,
         account: Account,
         proof_result_rx: Result<CrossbeamReceiver<StorageProofResultMessage>, DatabaseError>,
-        // None if results shouldn't be retained for this dispatched proof.
+        /// None if results shouldn't be retained for this dispatched proof.
         storage_proof_results: Option<Rc<RefCell<B256Map<Vec<ProofTrieNode>>>>>,
```
mediocregopher: This is always Some, make it not be an Option

proof-v2-trace-analysis.md
```diff
@@ -0,0 +1,86 @@
+# Proof V2 Trace Analysis - Block 24275866
```
mediocregopher: remove this file

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1221,34 +1277,39 @@ where
...
+            total_idle_time_us = total_idle_time.as_micros(),
             "Account worker shutting down"
         );
 
         #[cfg(feature = "metrics")]
         {
             self.metrics.record_account_nodes(account_nodes_processed as usize);
+            self.metrics.record_account_worker_idle_time(total_idle_time);
             self.cursor_metrics.record(&mut cursor_metrics_cache);
+            proof_calculator_metrics_cache.record();
```
mediocregopher: Make the metrics being recorded here part of the metrics on `self.metrics`, and add a method `record_account_proof_calculator_metrics` to it which records the idle time and variant counters, and get rid of this `record` method on the cache

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1293,28 +1354,42 @@ where
...
 
-    fn compute_v2_account_multiproof<Provider>(
+    #[allow(clippy::type_complexity)]
+    fn compute_v2_account_multiproof<'a, Provider>(
         &self,
-        v2_calculator: &mut proof_v2::ProofCalculator<
-            <Provider as TrieCursorFactory>::AccountTrieCursor<'_>,
-            <Provider as HashedCursorFactory>::AccountCursor<'_>,
-            AsyncAccountValueEncoder,
+        v2_account_calculator: &mut proof_v2::ProofCalculator<
```
mediocregopher: Create some type aliases for these proof calculators, to make the code clearer

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -81,39 +160,74 @@ pub(crate) struct AsyncAccountValueEncoder {
     /// Tracks storage proof results received from the storage workers. [`Rc`] + [`RefCell`] is
     /// required because [`DeferredValueEncoder`] cannot have a lifetime.
     storage_proof_results: Rc<RefCell<B256Map<Vec<ProofTrieNode>>>>,
+    /// Shared storage proof calculator for synchronous computation. Reuses cursors and internal
+    /// buffers across multiple storage root calculations.
+    storage_calculator: Rc<RefCell<StorageProofCalculator<TC, HC>>>,
+    /// Accumulated time spent waiting for storage proof results.
+    storage_wait_time: Rc<RefCell<Duration>>,
```
mediocregopher: just keep an `Rc<RefCell<ValueEncoderMetrics>>` and share that with the `AsyncAccountDeferredValueEncoder` as necessary

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -81,39 +160,74 @@ pub(crate) struct AsyncAccountValueEncoder {
...
 
-        // Any remaining dispatched proofs need to have their results collected
+        let mut storage_wait_time = Rc::into_inner(self.storage_wait_time)
+            .expect("no deferred encoders are still allocated")
+            .into_inner();
+
+        // Any remaining dispatched proofs need to have their results collected.
+        // These are proofs that were pre-dispatched but not consumed during proof calculation.
         for (hashed_address, rx) in &self.dispatched {
+            let _guard = debug_span!(
```
mediocregopher: Remove this debug_span, too noisy

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -1,44 +1,110 @@
...
     fn encode(self, buf: &mut Vec<u8>) -> Result<(), StateProofError> {
         let (account, root) = match self {
             Self::Dispatched {
                 hashed_address,
                 account,
                 proof_result_rx,
                 storage_proof_results,
+                storage_wait_time,
             } => {
+                let _guard = debug_span!(
```
mediocregopher: Remove this debug_span, too noisy

crates/trie/parallel/src/proof_task.rs
```diff
@@ -150,91 +168,122 @@ impl ProofWorkerHandle {
...
-        for worker_id in 0..storage_worker_count {
-            let span = debug_span!(target: "trie::proof_task", "storage worker", ?worker_id);
-            let task_ctx_clone = task_ctx.clone();
-            let work_rx_clone = storage_work_rx.clone();
-            let storage_available_workers_clone = storage_available_workers.clone();
-            let cached_storage_roots = cached_storage_roots.clone();
+        // Capture the current span to propagate context to spawned workers
+        let caller_span = tracing::Span::current();
 
+        // Spawn all worker pools in parallel to avoid blocking on sequential spawning
```
mediocregopher: No code changes here, just moved spawning of account/storage workers into their own spawn_blockings

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -109,39 +109,20 @@ where
...
+        let proof = storage_proof_calculator
+            .storage_proof(self.hashed_address, &mut [B256::ZERO.into()])?;
         let storage_root = storage_proof_calculator
-            .storage_proof(self.hashed_address, &mut [B256::ZERO.into()])
-            .map(|nodes| {
-                // Encode the root node to RLP and hash it
-                let root_node =
-                    nodes.first().expect("storage_proof always returns at least the root");
-                root_node.node.encode(buf);
+            .compute_root_hash(&proof)?
```
mediocregopher, line 122: Not really relevant to this PR, but we reuse `compute_root_hash` here to save some code

crates/trie/parallel/src/proof_task.rs
```diff
@@ -888,7 +928,12 @@ where
         // Initially mark this worker as available.
         self.available_workers.fetch_add(1, Ordering::Relaxed);
 
+        let mut total_idle_time = Duration::ZERO;
+        let mut idle_start = Instant::now();
+
         while let Ok(job) = self.work_rx.recv() {
+            total_idle_time += idle_start.elapsed();
```
mattsse, line 904: this seems appropriate

crates/trie/parallel/src/value_encoder.rs
```diff
@@ -1,44 +1,89 @@
...
+    rc::Rc,
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+/// Metrics collected by [`AsyncAccountValueEncoder`] during proof computation.
+///
+/// Tracks time spent waiting for storage proofs and counts of each deferred encoder variant used.
+#[derive(Debug, Default, Clone, Copy)]
+pub(crate) struct ValueEncoderMetrics {
```
mattsse: should we rename this to `Stats`, so that we dont confuse this with the actual metrics container?

crates/trie/parallel/src/proof_task.rs
```diff
@@ -1179,39 +1226,64 @@ where
...
 
-        let mut v2_calculator = if self.v2_enabled {
-            let trie_cursor = proof_tx.provider.account_trie_cursor()?;
-            let hashed_cursor = proof_tx.provider.hashed_account_cursor()?;
-            Some(proof_v2::ProofCalculator::<_, _, AsyncAccountValueEncoder>::new(
-                trie_cursor,
-                hashed_cursor,
-            ))
+        // Create both account and storage calculators for V2 proofs.
+        // The storage calculator is wrapped in Rc<RefCell<...>> for sharing with value encoders.
```
mattsse, line 1199: this makes sense, we have full access control and just dont need to not mess up the mut access hehe

mattsse (approved):
all of this checks out, I have one naming suggestion

otherwise pending @klkvr for vis

crates/trie/parallel/src/proof_task.rs
klkvr: should this record `idle_time/total_time` instead? feels like absolute idle time on its own is not very representative

klkvr (approved):
this looks good to me

yongkangc (approved):

crates/trie/parallel/src/proof_task.rs
mediocregopher: I wanted to avoid recording a metric within the actual work loop itself, and instead do all recordings outside the loop once all work is finished. This metrics is still useful as it tells us roughly how much time the worker spent doing nothing, which ideally is zero, so it's still useful as a signal for whether scheduling improvements even if it's not very fine-grained

PR #21427, opened 2026-01-26
feat(trie): add prune method to SparseTrieInterface

yongkangc:
**Summary**
Add `prune()` method to `SparseTrieInterface` for cross-payload sparse trie caching. Nodes beyond a specified depth are converted to `Hash` nodes, reducing memory usage while preserving the root hash. This enables efficient trie reuse between payload validations.

**Performance Note**
Profiling shows `partition_point` (binary search for descendant detection) takes ~15% of prune time. Alternative approaches were benchmarked but proved slower:
- HashSet bucketing by nibble length: hashing 32+ byte Nibbles keys is slower than lexicographic comparison which short-circuits early
- Prefix trie: construction overhead + terminal tracking negates lookup gains

The current binary search is optimal—log₂(n) cheap comparisons beats O(1) hash lookups for this workload.

**Parallelism Analysis**

1. **`ParallelSparseTrie::prune()`** — Removed parallelism entirely. Rayon dispatch overhead (~80-100µs) always exceeds the trivially fast per-subtrie retain() work (~0.3µs). See analysis: https://gist.github.com/yongkangc/c692e3bb177fb24a05266a8130a112c8

2. **`SparseStateTrie::prune()`** — Kept parallelism but raised threshold from 8 to 16. Benchmarking showed:
   - Small storage tries (~100 leaves, ~3.5µs work): Serial 2-3x faster due to ~15µs rayon overhead
   - Medium/large tries (~1000+ leaves, ~40µs+ work): Parallel 1.3-1.6x faster even at 4 tries
   - Raising threshold reduces worst-case regression while preserving parallel benefits for large workloads

**Expected Impact**
Reduced memory footprint for sparse trie caching across payload executions while maintaining correctness.

Closes RETH-178


crates/trie/sparse/src/metrics.rs
```diff
@@ -15,6 +15,18 @@ pub(crate) struct SparseStateTrieMetrics {
...
+    /// Number of nodes converted to hash stubs during storage trie pruning.
+    pub(crate) prune_storage_nodes_converted: u64,
+    /// Number of storage tries cleared (evicted) during pruning.
+    pub(crate) prune_storage_tries_cleared: u64,
+    /// Number of storage tries retained during pruning.
+    pub(crate) prune_storage_tries_retained: u64,
+    /// Total revealed nodes in account trie after pruning.
+    pub(crate) post_prune_account_nodes: u64,
+    /// Total revealed nodes across retained storage tries after pruning.
+    pub(crate) post_prune_storage_nodes: u64,
```
yongkangc: not 100% sure if we need this, but i am guessing at least initiall to find out the config? 

i can remove from this pr first and have it later for testing

crates/trie/sparse/src/trie.rs
```diff
@@ -1096,6 +1096,116 @@ impl SparseTrieTrait for SerialSparseTrie {
...
+                let candidate = &effective_pruned_roots[idx - 1].0;
+                if p.starts_with(candidate) && p.len() > candidate.len() {
+                    return true;
+                }
+            }
+            false
+        };
+
+        let starts_with_pruned = |p: &Nibbles| -> bool {
+            let idx = effective_pruned_roots.partition_point(|(root, _)| root <= p);
```
yongkangc: this partition point actually is (most) expensive from profiles of prune

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,338 @@ impl SparseTrie for ParallelSparseTrie {
...
+            .map(|s| s.nodes.values().filter(|n| !matches!(n, SparseNode::Hash(_))).count())
+            .sum();
+
+        upper_count + lower_count
+    }
+}
+
+impl SparseTrieExt for ParallelSparseTrie {
+    fn prune(&mut self, max_depth: usize) -> usize {
+        // Always clear transient bookkeeping, even if nothing is pruned.
```
yongkangc: we should use a simpler word than transient thats clearer

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,338 @@ impl SparseTrie for ParallelSparseTrie {
...
+            updates.wiped = false;
+        }
+        self.prefix_set.clear();
+
+        let estimated_capacity = self.upper_subtrie.nodes.len().min(256);
+        let mut effective_pruned_roots = Vec::<(Nibbles, B256)>::with_capacity(estimated_capacity);
+        let mut stack: SmallVec<[(Nibbles, usize); 32]> = SmallVec::new();
+        stack.push((Nibbles::default(), 0));
+
+        while let Some((path, depth)) = stack.pop() {
```
yongkangc, line 935: add comment for whats happening here

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,338 @@ impl SparseTrie for ParallelSparseTrie {
...
+            let subtrie = if SparseSubtrieType::path_len_is_upper(path.len()) {
+                &*self.upper_subtrie
+            } else if let Some(s) = self.lower_subtrie_for_path(&path) {
+                s
+            } else {
+                continue
+            };
+            let Some(node) = subtrie.nodes.get(&path) else { continue };
+
+            match node {
```
yongkangc: for each of the cases in match, add a comment

crates/trie/sparse/src/trie.rs
```diff
@@ -1096,6 +1096,142 @@ impl SparseTrieTrait for SerialSparseTrie {
     fn shrink_values_to(&mut self, size: usize) {
         self.values.shrink_to(size);
     }
+
+    fn revealed_node_count(&self) -> usize {
+        self.nodes.values().filter(|n| !matches!(n, SparseNode::Hash(_))).count()
+    }
+}
+
+impl SparseTrieExt for SerialSparseTrie {
```
yongkangc: lets remove the implementation for serial sparse trie

crates/trie/sparse/src/trie.rs
```diff
@@ -1096,6 +1096,16 @@ impl SparseTrieTrait for SerialSparseTrie {
...
```
mediocregopher: Let's remove this completely I say

crates/trie/sparse/src/traits.rs
```diff
@@ -230,6 +230,37 @@ pub trait SparseTrie: Sized + Debug + Send + Sync {
     /// Shrink the capacity of the sparse trie's value storage to the given size.
     /// This will reduce memory usage if the current capacity is higher than the given size.
     fn shrink_values_to(&mut self, size: usize);
+
+    /// Returns the number of revealed (non-Hash) nodes in the trie.
+    fn revealed_node_count(&self) -> usize;
```
mediocregopher: Let's make this part of SparseTrieExt

crates/trie/sparse/src/state.rs
```diff
@@ -972,6 +972,128 @@ where
...
+        // Collect keys to avoid borrow conflict
+        let tries_to_clear: Vec<B256> = self
+            .storage
+            .tries
+            .keys()
+            .filter(|hash| !tries_to_keep.contains(*hash))
+            .copied()
+            .collect();
+
+        for hash in tries_to_clear {
```
mediocregopher, line 1043: Rather than duplicating this logic you can do a check like [this](https://github.com/paradigmxyz/reth/blob/af3601c65d2b1cbf0ffa04d3f3390c18c997bfb5/crates/trie/sparse-parallel/src/trie.rs#L925)

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+impl SparseTrieExt for ParallelSparseTrie {
+    fn prune(&mut self, max_depth: usize) -> usize {
+        // Pruning is cache eviction, not a state transition.
+        // Clear update tracking and prefix set since the pruned trie is a fresh cache state.
+        if let Some(updates) = self.updates.as_mut() {
+            updates.updated_nodes.clear();
+            updates.removed_nodes.clear();
+            updates.wiped = false;
+        }
+        self.prefix_set.clear();
```
mediocregopher: I don't think these should be done here... prefix_set is reset when `root()` is called, and updates are taken by `take_updates`. It's up to the caller whether or not the sparse trie is "done" with the current set of changes (calling `root` and resetting prefix_set) or if/when they want to take the updates for persisting, `prune` can be agnostic to that.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+
+        // DFS traversal to find nodes at max_depth that can be pruned.
+        // Collects "effective pruned roots" - children of nodes at max_depth with computed hashes.
+        let estimated_capacity = self.upper_subtrie.nodes.len().min(256);
```
mediocregopher: Not sure where this calculation comes from. Afaict `effective_pruned_roots` contains the roots of all pruned subtries (ie all nodes replaced with their hash). This will be up to `16 ^ (max_depth+1)` nodes

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+        let mut effective_pruned_roots = Vec::<(Nibbles, B256)>::with_capacity(estimated_capacity);
+        let mut stack: SmallVec<[(Nibbles, usize); 32]> = SmallVec::new();
+        stack.push((Nibbles::default(), 0));
+
+        while let Some((path, depth)) = stack.pop() {
+            let subtrie = if SparseSubtrieType::path_len_is_upper(path.len()) {
+                &*self.upper_subtrie
+            } else if let Some(s) = self.lower_subtrie_for_path(&path) {
+                s
```
mediocregopher: You can use `subtrie_for_path` here I think. You can expect that it will always return Some, since you are traversing from root and so know which nodes are expected to exist in the overall trie.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+                    child.extend(key);
+                    if depth == max_depth {
+                        let child_subtrie = if SparseSubtrieType::path_len_is_upper(child.len()) {
+                            Some(&*self.upper_subtrie)
+                        } else {
+                            self.lower_subtrie_for_path(&child)
+                        };
+                        if let Some(hash) = child_subtrie
+                            .and_then(|s| s.nodes.get(&child))
+                            .filter(|n| !matches!(n, SparseNode::Hash(_)))
```
mediocregopher: `!SparseNode::is_hash`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+
+            match node {
+                // Terminal nodes: no children to traverse or prune
+                SparseNode::Empty | SparseNode::Hash(_) | SparseNode::Leaf { .. } => {}
+                // Extension: counts as 1 depth regardless of key length
+                SparseNode::Extension { key, .. } => {
+                    let mut child = path;
+                    child.extend(key);
+                    if depth == max_depth {
+                        let child_subtrie = if SparseSubtrieType::path_len_is_upper(child.len()) {
```
mediocregopher: `subtrie_for_path` here as well

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+
+                        let mut child = path;
+                        child.push_unchecked(nibble);
+                        if depth == max_depth {
+                            let child_subtrie = if SparseSubtrieType::path_len_is_upper(child.len())
+                            {
+                                Some(&*self.upper_subtrie)
+                            } else {
+                                self.lower_subtrie_for_path(&child)
+                            };
```
mediocregopher: `subtrie_for_path`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+                            if let Some(hash) = child_subtrie
+                                .and_then(|s| s.nodes.get(&child))
+                                .filter(|n| !matches!(n, SparseNode::Hash(_)))
```
mediocregopher: `!SparseNode::is_hash`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+            {
+                let mut all_roots: Vec<_> = roots_upper
+                    .iter()
+                    .chain(roots_by_lower.iter().flatten())
+                    .map(|(p, _)| p)
+                    .collect();
+                all_roots.sort_unstable();
+                all_roots.windows(2).all(|w| !w[1].starts_with(w[0]))
+            },
+            "prune roots must be prefix-free"
```
mediocregopher, line 1015: nice

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+        for (path, hash) in &effective_pruned_roots {
+            self.subtrie_for_path_mut(path).nodes.insert(*path, SparseNode::Hash(*hash));
+        }
+
+        let nodes_converted = effective_pruned_roots.len();
+
+        // Partition roots by subtrie for efficient per-subtrie pruning.
+        // Upper subtrie roots (path.len() < UPPER_TRIE_MAX_DEPTH) go to roots_upper.
+        // Lower subtrie roots go to roots_by_lower indexed by their subtrie index.
+        let mut roots_upper = Vec::<(Nibbles, B256)>::new();
```
mediocregopher: Check out the logic we use in [reveal_nodes](https://github.com/paradigmxyz/reth/blob/af3601c65d2b1cbf0ffa04d3f3390c18c997bfb5/crates/trie/sparse-parallel/src/trie.rs#L174) to partition a Vec by subtries. We do it by sorting the entire vec and then doing `chunk_by` to separate it by subtrie. I think since rayon isn't being used here this could be used without incurring any allocations at all.

I think trying something like this would work a lot better than doing the multiple for-loops you're doing below, you could call `retain` on the nodes/values of each subtrie using `starts_with_pruned_in` on just the set of roots for each subtrie group, with a bit of special handling for the upper trie (again, see `reveal_nodes` because it also has special handling for the upper trie)

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+        let mut fully_pruned_subtries = [false; NUM_LOWER_SUBTRIES];
+
+        // Case 1: Lower prune roots that exactly match subtrie root paths
+        for (idx, bucket) in roots_by_lower.iter().enumerate() {
+            for (path, hash) in bucket {
+                if let Some(subtrie) = self.lower_subtries[idx].as_revealed_mut() &&
+                    &subtrie.path == path
+                {
+                    subtrie.nodes.clear();
+                    subtrie.inner.values.clear();
```
mediocregopher: `subtrie.clear()` would be cleaner imo

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,6 +906,251 @@ impl SparseTrie for ParallelSparseTrie {
...
+        self.upper_subtrie.nodes.retain(|p, _| !is_strict_descendant_in(&roots_upper, p));
+        // Upper subtrie values can have full paths (values are always stored in upper subtrie).
+        // Check both upper roots (for any path) and the appropriate lower bucket.
+        self.upper_subtrie.inner.values.retain(|p, _| {
+            // Always check upper roots first - they can have descendants at any depth
+            if starts_with_pruned_in(&roots_upper, p) {
+                return false;
+            }
+            // For longer paths, also check the appropriate lower bucket
+            if !SparseSubtrieType::path_len_is_upper(p.len()) {
```
mediocregopher: The path length of keys in `values` is always 64 nibbles, these are the full keys which are mapped to their values.

mediocregopher (changes requested):

crates/trie/sparse-parallel/src/trie.rs
yongkangc: thanks, that is much cleaner

crates/trie/sparse-parallel/src/trie.rs
yongkangc: much nicer

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -906,7 +906,9 @@ impl SparseTrie for ParallelSparseTrie {
             subtrie.shrink_values_to(size_per_subtrie);
         }
     }
+}
 
+impl SparseTrieExt for ParallelSparseTrie {
     fn revealed_node_count(&self) -> usize {
```
yongkangc, line 913: we should add a docstring here

crates/trie/sparse/src/state.rs
yongkangc: ah nice, thanks for that. Extracted the parallelism check into a helper function `is_prune_parallelism_enabled()`

crates/trie/sparse/src/state.rs
```diff
@@ -972,6 +972,119 @@ where
...
+    A: SparseTrieTrait + SparseTrieExt + Default,
+    S: SparseTrieTrait + SparseTrieExt + Default + Clone,
+{
+    /// Prunes the account trie and selected storage tries to reduce memory usage.
+    ///
+    /// # Arguments
+    ///
+    /// * `max_depth` - Maximum depth to retain in pruned tries
+    /// * `max_storage_tries` - Maximum number of storage tries to keep (largest by node count)
+    ///
```
yongkangc:
```suggestion
```

crates/trie/sparse/src/state.rs
yongkangc: or do u mean simplifying @mediocregopher 

```
  for hash in tries_to_clear { 1Code has comments. Press enter to view.
            if let Some(trie) = self.storage.tries.remove(&hash) {
                self.storage.cleared_tries.push(trie.clear());
            }
            if let Some(mut paths) = self.storage.revealed_paths.remove(&hash) {
                paths.clear();
                self.storage.cleared_revealed_paths.push(paths);
            }
        }
```

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -908,6 +908,183 @@ impl SparseTrie for ParallelSparseTrie {
...
+                }
+            }
+        }
+
+        if effective_pruned_roots.is_empty() {
+            return 0;
+        }
+
+        for (path, hash) in &effective_pruned_roots {
+            self.subtrie_for_path_mut(path).nodes.insert(*path, SparseNode::Hash(*hash));
```
mediocregopher: Not sure if it's possible due to bounds checking, but maybe: could we use `subtrie_for_path_mut` in the while-loop over the stack above, and replace the node with a `SparseNode::Hash` in that loop? That would avoid this loop completely

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -908,6 +908,183 @@ impl SparseTrie for ParallelSparseTrie {
...
+
+            // Skip if an upper trie extension already covers this subtrie (replaced with Blind)
+            if fully_pruned_subtries[subtrie_idx] {
+                continue;
+            }
+
+            // Skip unrevealed subtries - nothing to prune
+            let Some(subtrie) = self.lower_subtries[subtrie_idx].as_revealed_mut() else {
+                continue;
+            };
```
mediocregopher: I think these two checks are redundant, if the lower subtrie was fully pruned then it was also blinded. My be possible to factor out `fully_pruned_subtries` completely.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -908,6 +908,183 @@ impl SparseTrie for ParallelSparseTrie {
...
+        // Upper prune roots that are prefixes of lower subtrie root paths cause the entire
+        // subtrie to be replaced with Blind(None).
+        if !roots_upper.is_empty() {
+            for (idx, subtrie) in self.lower_subtries.iter_mut().enumerate() {
+                let should_clear = subtrie.as_revealed_ref().is_some_and(|s| {
+                    let search_idx = roots_upper.partition_point(|(root, _)| root <= &s.path);
+                    search_idx > 0 && s.path.starts_with(&roots_upper[search_idx - 1].0)
+                });
+                if should_clear {
+                    *subtrie = LowerSparseSubtrie::Blind(None);
```
mediocregopher: Rather than setting to `Blind`, just call `clear` on the subtrie and it will automatically blind it while preserving the previous allocations

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -908,6 +908,183 @@ impl SparseTrie for ParallelSparseTrie {
...
+            }
+
+            // Skip unrevealed subtries - nothing to prune
+            let Some(subtrie) = self.lower_subtries[subtrie_idx].as_revealed_mut() else {
+                continue;
+            };
+
+            // Retain only nodes/values not descended from any pruned root.
+            subtrie.nodes.retain(|p, _| {
+                !is_strict_descendant_in(roots_group, p) && !is_strict_descendant_in(roots_upper, p)
```
mediocregopher: I'm trying to think if these calls to `is_strict_descendant_in(roots_upper, p)` would ever return false... if `roots_upper` had a node then all lower subtries which are children to it would already be cleared/blinded, so we wouldn't be at this point at all. Am I missing something?

mediocregopher (changes requested):

mediocregopher (approved):

PR #21428, opened 2026-01-26
docs(chain-state): correct iteration order comments

reallesee:
Comments in `canonical_hashes_range` incorrectly described `in_memory` as storing blocks in ascending order. The struct documentation and `trie_input()` logic confirm descending order (newest to oldest).


mattsse (commented):

PR #21433, opened 2026-01-26
feat(engine): slow block logs

shekhirin:
Ref: https://hackmd.io/dg7rizTyTXuCf2LSa2LsyQ, https://ethresear.ch/t/data-driven-analysis-on-eip-7907/23850

Reference implementations in other clients: https://github.com/NethermindEth/nethermind/pull/10288, https://github.com/ethereum/go-ethereum/pull/33655

Adds configurable slow block logging via `--engine.slow-block-threshold` that emits a structured `warn!` log with detailed timing, state-operation counts, and cache hit-rate metrics for blocks whose total processing time exceeds the threshold.

When the threshold is configured, block validation instruments the state provider to collect fetch counts and latencies, and tracks cache hits/misses. After persistence commits, blocks exceeding the threshold emit a `ConsensusEngineEvent::SlowBlock` carrying an `ExecutionTimingStats` snapshot. The node events handler formats this into a single structured log line with execution time, state root time, commit time, mgas/s, account/storage/code fetch counts, and cache hit rates.

When no threshold is set, no stats are collected and no additional allocations occur.

Key types introduced:
- `ExecutionTimingStats` (reth-chain-state): 24-field struct capturing block number/hash, execution/state-root/commit durations, state operation counts, and cache metrics
- `SlowBlockInfo` (reth-engine-primitives): wrapper carrying stats + commit duration + total duration
- `CacheStats` (engine-tree): atomic counters for cache hit/miss tracking
- `StateProviderStats` (engine-tree): atomic counters for state provider fetch counts and latencies
- `ValidationOutcome` (engine-tree): result type alias for block validation with optional timing stats

Also refactors `PersistenceResult` to carry commit duration alongside the last persisted block, replacing the separate `SaveBlocksResult` type.


crates/engine/tree/src/tree/instrumented_state.rs
```diff
@@ -308,3 +258,113 @@ impl<S: HashedPostStateProvider> HashedPostStateProvider for InstrumentedStatePr
         self.state_provider.hashed_post_state(bundle_state)
     }
 }
+
+/// A handle to [`InstrumentedStateProvider`] that can be used to fetch recorded metrics.
```
mattsse: i find handle here a bit confusing, isnt this now just a container for metrics?

crates/node/core/src/args/debug.rs
```diff
@@ -236,7 +236,7 @@ impl FromStr for InvalidBlockSelection {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         if s.is_empty() {
-            return Ok(Self(Default::default()))
+            return Ok(Self(Default::default()));
```
mattsse: unrelated

crates/node/builder/src/builder/mod.rs
```diff
@@ -703,7 +703,6 @@ where
         let Self { builder, task_executor } = self;
 
         let engine_tree_config = builder.config.engine.tree_config();
-
```
mattsse, line 708: unrelated

crates/ethereum/evm/Cargo.toml
```diff
@@ -34,9 +34,15 @@ derive_more = { workspace = true, optional = true }
...
-reth-evm = { workspace = true, features = ["test-utils"] }
+reth-evm = { workspace = true, features = ["test-utils", "metrics"] }
+reth-revm.workspace = true
+reth-execution-types.workspace = true
 secp256k1.workspace = true
 alloy-genesis.workspace = true
+revm-bytecode.workspace = true
+revm-database.workspace = true
+revm-primitives.workspace = true
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
```
mattsse: why do we need all of this here now?

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -99,6 +103,28 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
...
+        trie_output: TrieUpdates,
+        overlay_factory: OverlayStateProviderFactory<P>,
+    ) -> Self {
+        Self {
+            hashed_state: Arc::new(hashed_state),
+            trie_output: Arc::new(trie_output),
+            overlay_factory,
+        }
+    }
+}
```
mattsse: im losing it with amp, every single type it puts new types at the wrong place, this does not belong above the main type

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -99,6 +103,28 @@ impl<'a, N: NodePrimitives> TreeCtx<'a, N> {
     }
 }
 
+/// Trie context needed for [`BasicEngineValidator::spawn_deferred_trie_task`]
+#[derive(Debug)]
+struct TrieCtx<P> {
+    hashed_state: Arc<HashedPostState>,
+    trie_output: Arc<TrieUpdates>,
+    overlay_factory: OverlayStateProviderFactory<P>,
+}
```
mattsse: is this an unrelated change? i dont see how this is related to metrics

crates/engine/tree/src/persistence.rs
```diff
@@ -141,26 +153,117 @@ where
...
+                stats.state_hash_duration +
+                commit_duration;
+
+            // Check if total time exceeds threshold
+            if let Some(threshold) = self.slow_block_threshold &&
+                total_duration > threshold
+            {
+                self.log_slow_block(stats, commit_duration, total_duration);
+            }
+        }
```
mattsse: can we extract this to a helper fn perhaps?

crates/engine/tree/src/persistence.rs
```diff
@@ -141,26 +153,117 @@ where
...
+        let commit_duration = total_commit_duration.div_f64(block_count as f64);
 
         debug!(target: "engine::persistence", first=?first_block, last=?last_block, "Saved range of blocks");
 
         self.metrics.save_blocks_block_count.record(block_count as f64);
-        self.metrics.save_blocks_duration_seconds.record(start_time.elapsed());
+        self.metrics.save_blocks_duration_seconds.record(total_commit_duration);
+
+        // Emit unified slow block logs after commit completes
+        for stats in timing_stats {
```
mattsse: this loop is a noop if no slow_block_threshold is set right?

crates/chain-state/src/in_memory.rs
```diff
@@ -760,6 +761,12 @@ pub struct ExecutedBlock<N: NodePrimitives = EthPrimitives> {
     /// This allows deferring the computation of the trie data which can be expensive.
     /// The data can be populated asynchronously after the block was validated.
     pub trie_data: DeferredTrieData,
+    /// Optional timing statistics for slow block logging.
+    ///
+    /// When slow block logging is enabled, these statistics are populated during
+    /// block validation and used by the persistence service to emit a unified log
+    /// after database commit.
+    pub timing_stats: Option<ExecutionTimingStats>,
```
mattsse: hmm, not sure how i feel about this being tracked here

can we perhaps try keeping this internal to the tree?

crates/engine/tree/src/persistence.rs
```diff
@@ -141,26 +153,117 @@ where
...
+            (stats.execution_duration.as_secs_f64() + stats.state_hash_duration.as_secs_f64());
+
+        // Calculate cache hit rates
+        let account_hit_rate =
+            calculate_hit_rate(stats.account_cache_hits, stats.account_cache_misses);
+        let storage_hit_rate =
+            calculate_hit_rate(stats.storage_cache_hits, stats.storage_cache_misses);
+        let code_hit_rate = calculate_hit_rate(stats.code_cache_hits, stats.code_cache_misses);
+
+        warn!(
```
mattsse: is this the expected format we need to emit?

crates/engine/tree/src/persistence.rs
```diff
@@ -141,26 +153,117 @@ where
         &self,
         blocks: Vec<ExecutedBlock<N::Primitives>>,
     ) -> Result<Option<BlockNumHash>, PersistenceError> {
```
mattsse: one idea, what if we return the duration as part of the update, then we could emit this as part of the general purpose events channel and we don't have most of it in the tree itself

mattsse (changes requested):
this is quite a lot of stuff -.-

but I guess all of these make sense, I'd like to avoid having this leak as part of the block type

also have some suggestions for how to avoid adding this to the persistence task as well.

crates/ethereum/evm/Cargo.toml
CPerezz: To enable testing. I needed all of the dev depencencies in order to be able to run the tests.

crates/engine/tree/src/persistence.rs
CPerezz: You can cross-check with Geth. I'd prefer to have indentation but with the logging lib used here it's not going to work or going to be ugly. At least having it as JSON makes it parseable

crates/engine/tree/src/tree/payload_validator.rs
CPerezz: This wasn't in my original PR. @shekhirin I assume is your addition?

crates/engine/tree/src/persistence.rs
CPerezz: Yes. The format can be found in the Geth PR too: https://github.com/ethereum/go-ethereum/pull/33655#issuecomment-3809755085

crates/chain-state/src/execution_stats.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::time::Duration;
+
+use alloy_primitives::B256;
+
+/// Statistics collected during block execution for cross-client performance analysis.
+///
+/// These statistics are populated during block validation and carried through to
+/// persistence, where they are used to emit a single unified log entry that includes
+/// complete timing information (including commit time).
+#[derive(Debug, Clone, Copy, Default)]
```
DaniPopes: don't make this copy

crates/chain-state/src/execution_stats.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct ExecutionTimingStats {
```
DaniPopes, line 17: this is quite big, maybe box or arc internally?

crates/engine/tree/src/persistence.rs
```diff
@@ -142,18 +148,28 @@ where
...
+            blocks.iter().filter_map(|b| b.timing_stats()).collect()
+        } else {
+            Vec::new()
+        };
 
-        if let Some(last) = last_block {
+        let total_commit_duration = {
+            let start = Instant::now();
             let provider_rw = self.provider.database_provider_rw()?;
             provider_rw.save_blocks(blocks, SaveBlocksMode::Full)?;
```
mattsse: instead of doing all of thise here, we should return the metrics we need from persistince as part of the response message we send, currently we only return Option<BlockNumHash>

crates/chain-state/src/in_memory.rs
mattsse: we shouldnt track this as part of the block itself, and instead track this internally in the engine treeblock buffer.

once we've persisted a block we should emit a message as part of

pub enum ConsensusEngineEvent<N: NodePrimitives = EthPrimitives> {

that has the metrics included, then we can log the slow blockresponses on the event task and dont need to block on this in the engine tiself.

something like

```
ConsensusEngineEvent::Persisted(Vec<PersitedInfo>)

struct PersitedInfo {
   block: NumHash,
   timings: Option<Timings>
}
```

mattsse (changes requested):

crates/engine/tree/src/persistence.rs
```diff
@@ -157,18 +165,10 @@ where
...
-                if finalized > last.number {
-                    self.pending_finalized_block = Some(finalized);
-                }
+                provider_rw.save_finalized_block_number(finalized)?;
             }
             if let Some(safe) = pending_safe {
-                provider_rw.save_safe_block_number(safe.min(last.number))?;
-                if safe > last.number {
-                    self.pending_safe_block = Some(safe);
-                }
```
shekhirin, line 171: why was this removed?

mattsse (approved):

PR #21435, opened 2026-01-26
refactor(trie): always use ParallelSparseTrie, deprecate config flags

mediocregopher:
- Deprecate --engine.disable-parallel-sparse-trie CLI flag (now no-op)
- Deprecate --engine.parallel-sparse-trie CLI flag (now no-op)
- Delete ConfiguredSparseTrie enum, use ParallelSparseTrie directly
- Rename SparseTrie<T> enum to RevealableSparseTrie<T>
- Rename SparseTrieInterface trait to SparseTrie


shekhirin (approved):
nice

PR #21439, opened 2026-01-26
perf(trie): add `has_non_zero` flag to `HashedStorageSorted` for O(1) emptiness checks

figtracer:
## Overview

Optimizes `is_storage_empty()` by pre-computing a `has_non_zero` flag during storage sorting, replacing an O(n) scan with an O(1) check.

## Problem

MDBX profiling on mainnet identified `is_storage_empty()` as a bottleneck during parallel trie proof generation:

- 12,917 slow operations at 5.37ms average
- ~69 seconds total across 230 blocks
- High major page fault rate (~47%) indicating disk I/O pressure

The current implementation scans all storage slots on every call:
```rust
self.post_state_cursor.has_any(|(_, value)| value.into_option().is_some())
```

## Solution

Add a `has_non_zero` boolean to `HashedStorageSorted` that is:

1. Computed once during `into_sorted()` (which already iterates all slots)
2. Maintained through `extend_ref()` and `merge_batch()` operations  
3. Used for O(1) short-circuit in `is_storage_empty()`

For accounts with non-zero storage in post-state, this skips both the O(n) scan and the MDBX `seek_exact` call entirely.

## Changes

- Add `has_non_zero: bool` field to `HashedStorageSorted`
- Update construction sites to compute the flag
- Replace `has_any()` scan with flag check in `is_storage_empty()`


crates/trie/trie/src/hashed_cursor/post_state.rs
```diff
@@ -350,8 +358,8 @@ where
     /// This function should be called before attempting to call [`HashedCursor::seek`] or
     /// [`HashedCursor::next`].
     fn is_storage_empty(&mut self) -> Result<bool, DatabaseError> {
-        // Storage is not empty if it has non-zero slots.
-        if self.post_state_cursor.has_any(|(_, value)| value.into_option().is_some()) {
+        // O(1) check: storage is not empty if post-state has any non-zero slots.
+        if self.has_non_zero {
```
mattsse, line 362: im unsure that avoding this any check is better than having to always check all values

PR #21444, opened 2026-01-26
fix(stages): retain RocksDB TempDir in TestStageDB to prevent premature deletion

ethfanWilliam:
The RocksDB temp directory was being dropped immediately after creation in both `TestStageDB::default()` and `TestStageDB::new()`, while the path was still being used by RocksDBProvider. This could cause the directory to be deleted before RocksDB finished using it, leading to undefined behavior.                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                       Added `temp_rocksdb_dir` field to TestStageDB to keep the TempDir alive for the struct's lifetime, matching how we already handle `temp_static_files_dir`. Also matches the pattern used in `provider/src/test_utils/mod.rs` and `exex/test-utils/src/lib.rs`.


joshieDo (approved):
i think RocksDB was creating it anyway, but it makes sense, ty

PR #21447, opened 2026-01-26
refactor(storage): remove dead test helper functions from LatestStateProvider

Aleksandr1732:
Removes unused compile-time test helpers from `LatestStateProvider` that were marked with `#[expect(dead_code)]` and never invoked.


joshieDo (approved):

mattsse (commented):
we can keep these because these are free

PR #21448, opened 2026-01-26
chore(engine): remove deprecated config flags

ethfanWilliam:
Removes the deprecated `caching_and_prewarming_enabled`, `precompile_cache_enabled`, `parallel_sparse_trie_enabled`, and `parallel_sparse_trie_disabled` fields from EngineArgs. These were already marked deprecated and hidden and they had no effect and weren't used in tree_config().


mattsse (commented):
We need to keep these deprecated flags until the next major release.

PR #21463, opened 2026-01-27
fix(node): ethstats conn/last_ping deadlock

akolishchak:
The deadlock setup:
task 1 (`send_ping()` - timeout spawn): holds `last_ping`, waits `conn.write()`
task 2 (`report_latency()` - on pong): holds `conn.read()`, waits `last_ping`

Fixed by never holding `last_ping` while awaiting on `conn` (and vice‑versa), to remove the lock‑order inversion.


mattsse (approved):
I see, this checks out, tysm, sorry for delay

PR #21464, opened 2026-01-27
chore: replace Option::map().unwrap_or(false) with is_some_and

davidjsonn:
Simplify boolean checks on `Option` using `is_some_and`, aligning with existing codebase style.


mattsse (commented):

PR #21468, opened 2026-01-27
fix(provider): skip storage changeset writes when routed to static files

yongkangc:
## Summary

Fixed a bug where storage changesets were written twice when static file storage changesets are enabled, causing `UnexpectedStaticFileBlockNumber` errors.

## Problem

PR #20896 added static file routing for storage changesets but forgot to add `write_storage_changesets` to `StateWriteConfig`. This caused storage changesets to be written twice:

1. First write: via `write_blocks_data` → static file manager
2. Second write: via `write_state` → `write_state_reverts` → `EitherWriter::new_storage_changesets` (also routed to static files)

The second write fails because block N was already written:
```
ProviderError(UnexpectedStaticFileBlockNumber(StorageChangeSets, 1, 2))
```

This manifests with `persistence_threshold(0)` and static file changesets enabled.

## Solution

1. Add `write_storage_changesets` field to `StateWriteConfig`
2. Pass `!sf_ctx.write_storage_changesets` to the config in `save_blocks`
3. Guard storage changeset writes in `write_state_reverts` with the config

This matches the existing pattern for `write_receipts` and `write_account_changesets`.

Closes #21467

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -44,3 +44,24 @@ jobs:
             --exclude 'op-reth' \
             --exclude 'reth' \
             -E 'binary(e2e_testsuite)'
+
+  rocksdb:
+    name: e2e-rocksdb
+    runs-on: depot-ubuntu-latest-4
+    env:
+      RUST_BACKTRACE: 1
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: mozilla-actions/sccache-action@v0.0.9
+      - uses: taiki-e/install-action@nextest
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+      - name: Run RocksDB e2e tests
+        run: |
+          cargo nextest run \
+            --locked --features "edge" \
+            -p reth-e2e-test-utils \
+            -E 'binary(rocksdb)'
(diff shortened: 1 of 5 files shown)


crates/e2e-test-utils/tests/rocksdb/main.rs
```diff
@@ -98,14 +98,8 @@ fn test_attributes_generator(timestamp: u64) -> EthPayloadBuilderAttributes {
...
 /// Enables `RocksDB` for `TransactionHashNumbers` table.
-///
-/// Note: Static file changesets are disabled because `persistence_threshold(0)` causes
-/// a race where the static file writer expects sequential block numbers but receives
-/// them out of order, resulting in `UnexpectedStaticFileBlockNumber` errors.
 fn with_rocksdb_enabled<C>(mut config: NodeConfig<C>) -> NodeConfig<C> {
     config.rocksdb = RocksDbArgs { tx_hash: true, ..Default::default() };
-    config.static_files.storage_changesets = false;
-    config.static_files.account_changesets = false;
     config
```
joshieDo: can we be explicit : true instead of default

joshieDo: if edge feature is enabled

yongkangc: much better yup

yongkangc: addressed

.github/workflows/e2e.yml
```diff
@@ -44,3 +44,24 @@ jobs:
...
+      - uses: taiki-e/install-action@nextest
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+      - name: Run RocksDB e2e tests
+        run: |
+          cargo nextest run \
+            --locked --features "edge" \
+            -p reth-e2e-test-utils \
+            -E 'binary(rocksdb)'
```
yongkangc, line 67: not 100% sure if we want this to run all the time, but makes sense to me

joshieDo (approved):

PR #21470, opened 2026-01-27
fix(db): write genesis history to correct storage backend

gakonst:
## Summary

When `account_history_in_rocksdb` or `storages_history_in_rocksdb` is enabled, `insert_genesis_history` now writes to RocksDB instead of MDBX.

## Problem

Previously, genesis account/storage history was always written to MDBX regardless of storage settings. This caused ~8893 stale entries in MDBX `AccountsHistory` table on reth-edge nodes configured for RocksDB history (the ~8893 entries correspond to Ethereum mainnet genesis allocation accounts).

## Solution

Updated `insert_genesis_history` and `insert_history` to check storage settings and write to the appropriate backend:
- When `account_history_in_rocksdb` is true → write to RocksDB
- When `account_history_in_rocksdb` is false → write to MDBX (existing behavior)
- Same logic for `storages_history_in_rocksdb`

This ensures history indices are written to exactly one location based on the storage configuration.

### DIFF
--- a/crates/storage/db-common/src/init.rs
+++ b/crates/storage/db-common/src/init.rs
@@ -15,9 +15,10 @@ use reth_primitives_traits::{
 use reth_provider::{
     errors::provider::ProviderResult, providers::StaticFileWriter, BlockHashReader, BlockNumReader,
     BundleStateInit, ChainSpecProvider, DBProvider, DatabaseProviderFactory, ExecutionOutcome,
-    HashingWriter, HeaderProvider, HistoryWriter, MetadataWriter, OriginalValuesKnown,
-    ProviderError, RevertsInit, StageCheckpointReader, StageCheckpointWriter, StateWriter,
-    StaticFileProviderFactory, StorageSettings, StorageSettingsCache, TrieWriter,
+    HashingWriter, HeaderProvider, HistoryWriter, MetadataWriter, NodePrimitivesProvider,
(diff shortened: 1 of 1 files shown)


crates/storage/db-common/src/init.rs
```diff
@@ -369,37 +370,71 @@ where
...
-    provider.insert_storage_history_index(storage_transitions)?;
-
-    trace!(target: "reth::cli", "Inserted storage history");
+    use reth_db_api::{
+        models::{storage_sharded_key::StorageShardedKey, ShardedKey},
+        BlockNumberList,
+    };
+    use reth_provider::EitherWriter;
+
+    // Write account history using EitherWriter
```
joshieDo:
```suggestion
```

crates/storage/db-common/src/init.rs
```diff
@@ -369,37 +370,71 @@ where
...
+        let mut writer = EitherWriter::new_accounts_history(provider, batch)?;
+        let list = BlockNumberList::new([block]).expect("single block always fits");
+        for (addr, _) in alloc.clone() {
+            writer.put_account_history(ShardedKey::last(*addr), &list)?;
+        }
+        trace!(target: "reth::cli", "Inserted account history");
+        Ok(((), writer.into_raw_rocksdb_batch()))
+    })?;
+
+    // Write storage history using EitherWriter
```
joshieDo:
```suggestion
```

PR #21475, opened 2026-01-27
perf(tree): optimistically prepare canonical overlay

mattsse:
## Summary

Pre-compute the `LazyOverlay` for the canonical head after persistence completes, so subsequent payloads building on the canonical head can reuse the cached overlay immediately without re-traversing in-memory blocks.

## Changes

- Add `PreparedCanonicalOverlay` struct to `TreeState` to cache the overlay with parent and anchor hashes
- Add `prepare_canonical_overlay()` that returns the overlay clone for background computation  
- Spawn rayon task to trigger `overlay.get()` for eager computation
- Add anchor hash verification in `get_cached_overlay()` to prevent using stale overlays after persistence advances
- Invalidate overlay in `remove_until()` and after persistence changes the anchor

## How it works

1. After persistence completes, `prepare_canonical_overlay()` is called
2. It creates a `LazyOverlay` with handles to deferred trie data from in-memory blocks
3. A rayon task is spawned to call `.get()` which triggers the actual overlay computation
4. When the next payload arrives building on the canonical head, `get_parent_lazy_overlay()` checks the cache
5. If parent hash and anchor hash match, the pre-computed overlay is reused

## Safety

- Both `parent_hash` and `anchor_hash` must match for cache hit (prevents stale overlay usage)
- Overlay is invalidated on `reset()`, `remove_until()`, and after persistence
- `LazyOverlay` uses `Arc<OnceLock>` so the spawned task and cached copy share the same result

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -1406,7 +1406,20 @@ where
         );
         self.changeset_cache.evict(eviction_threshold);
 
+        // Invalidate cached overlay since the anchor has changed
+        self.state.tree_state.invalidate_cached_overlay();
+
         self.on_new_persisted_block()?;
+
+        // Re-prepare overlay for the current canonical head with the new anchor.
+        // Spawn a background task to trigger computation so it's ready when the next payload
+        // arrives.
+        if let Some(overlay) = self.state.tree_state.prepare_canonical_overlay() {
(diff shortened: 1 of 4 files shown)


mediocregopher (approved):

shekhirin (approved):

gakonst:
What was the improvement?

PR #21478, opened 2026-01-27
fix: ensure edge enables history in rocksdb

joshieDo:


yongkangc (approved):

gakonst (approved):
LGTM - correctly enables history indices in RocksDB for edge mode

PR #21485, opened 2026-01-27
perf(net): avoid double clone in block range fetch

dizer-ti:
Was cloning headers twice - once for validation, once for pending_headers.
Now clone once and reverse in-place instead.


mattsse (commented):
this doesnt safe anything

PR #21487, opened 2026-01-27
chore: make engine tree crate items public

meyer9:
This makes the required items public for users of the Reth API to implement engine validator.

The goal is to be able to control the state root task and transaction execution more closely. Specifically, we want to reuse results from flashblock execution during NewPayload execution without having the entire block executed.

We also eventually want to control the state root task more closely by incrementally building the state root as flashblocks are processed instead of only when we receive the NewPayload call.

We recognize that the Reth team probably wants to implement these features upstream at some point, but it may not be an immediate priority, so this is a good compromise for now to allow us to implement these features without including them upstream yet.

Also, it makes sense that the engine validator should be implementable externally considering it's already part of the add-ons API to Reth. Otherwise, it doesn't really make sense for EngineValidator to be a trait that can be switched out via the Reth API.

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -76,7 +76,8 @@ impl CacheConfig for EpochCacheConfig {
 type FixedCache<K, V, H = DefaultHashBuilder> = fixed_cache::Cache<K, V, H, EpochCacheConfig>;
 
 /// A wrapper of a state provider and a shared cache.
-pub(crate) struct CachedStateProvider<S> {
+#[derive(Debug)]
+pub struct CachedStateProvider<S> {
     /// The state provider
     state_provider: S,
 
@@ -96,7 +97,7 @@ where
 {
     /// Creates a new [`CachedStateProvider`] from an [`ExecutionCache`], state provider, and
     /// [`CachedStateMetrics`].
-    pub(crate) const fn new(
+    pub const fn new(
         state_provider: S,
         caches: ExecutionCache,
         metrics: CachedStateMetrics,
@@ -114,7 +115,7 @@ impl<S> CachedStateProvider<S> {
     /// [`State`](revm::database::State) also caches internally during block execution and the cache
     /// is then updated after the block with the entire [`BundleState`] output of that block which
     /// contains all accessed accounts,code,storage. See also [`ExecutionCache::insert_state`].
-    pub(crate) const fn prewarm(mut self) -> Self {
+    pub const fn prewarm(mut self) -> Self {
         self.prewarm = true;
         self
     }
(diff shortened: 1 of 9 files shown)


mediocregopher (commented):
This is fine by me, just needs CI fixed 👍 

@meyer9 just fyi we're currently working on a fairly large overhaul to how the multiproof task and sparse trie task interact, so that's something to look out for.

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -662,7 +658,8 @@ impl ExecutionCache {
     ///
     /// Returns an error if the state updates are inconsistent and should be discarded.
     #[instrument(level = "debug", target = "engine::caching", skip_all)]
-    pub(crate) fn insert_state(&self, state_updates: &BundleState) -> Result<(), ()> {
+    #[expect(clippy::result_unit_err)]
+    pub fn insert_state(&self, state_updates: &BundleState) -> Result<(), ()> {
```
meyer9, line 662: This needs to be public because it's referenced in a doc that's public. I didn't want to change this interface, so added this lint disable for now.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -134,14 +136,14 @@ where
...
-    tree_state: TreeState<N>,
+    pub tree_state: TreeState<N>,
     /// Tracks the forkchoice state updates received by the CL.
     forkchoice_state_tracker: ForkchoiceStateTracker,
     /// Buffer of detached blocks.
     buffer: BlockBuffer<N::Block>,
     /// Tracks the header of invalid payloads that were rejected by the engine because they're
     /// invalid.
-    invalid_headers: InvalidHeaderCache,
+    pub invalid_headers: InvalidHeaderCache,
```
mattsse: I'd like to avoid mixing pub fields, can we instead add getters for this?

crates/engine/tree/src/tree/state.rs
mattsse, line 24: I dont want to leak private field access, because this would allow for modifications that could break things

would accept (mut) getters tho

mattsse (changes requested):
not opposed, seems reasonable,

I would however avoid giving pub field access because this makes it harder to reason about access etc

crates/engine/tree/src/tree/state.rs
meyer9: nice, not sure why those were changed to public actually since they aren't required anywhere. removed this change

crates/engine/tree/src/tree/mod.rs
meyer9: yup, done! I added `has_invalid_header` instead of returning a reference to the header cache.

mattsse (approved):

PR #21489, opened 2026-01-27
fix(rpc): add block sequence validation to debug_traceCallMany

livehardd:
Same validation as eth_simulateV1 - reject non-monotonic block numbers and timestamps when user provides explicit overrides.


mattsse (commented):
this isnt specced for callmany and i dont think these checks really matter anyway

PR #21490, opened 2026-01-27
refactor(db): make Tx::inner field private with accessor

mattsse:
Makes the `Tx::inner` field private and exposes a `pub const fn inner()` accessor instead.

This improves encapsulation while maintaining the same functionality.


Rjected (approved):
lgtm

PR #21497, opened 2026-01-27
feat(rpc): add debug_verbosity/vmodule

figtracer:
this pr adds the capability to change log verbosity at runtime via debug_verbosity/debug_vmodule rpc calls (#21271)

reopens #21314

ref
https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugverbosity
https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugvmodule

log levels:
- 0: OFF (silent)
- 1: ERROR
- 2: WARN
- 3: INFO
- 4: DEBUG
- 5+: TRACE

## usage
changing log verbosity to INFO
```bash
root@ns3258516:/home/ubuntu/reth# curl -X POST \
 --data '{"jsonrpc":"2.0","method":"debug_verbosity","params":[3],"id":1}' \
http://localhost:8545
{"jsonrpc":"2.0","id":1,"result":null}
```
debug_vmodule overrides every filter
```bash
curl -X POST -H "Content-Type: application/json" \
  --data '{"jsonrpc":"2.0","method":"debug_vmodule","params":["warn,engine=trace,reth_node_events=info"],"id":1}' \
  http://localhost:8545
{"jsonrpc":"2.0","id":1,"result":null}
```


crates/tracing/src/lib.rs
```diff
@@ -65,9 +65,95 @@ mod layers;
...
+
+/// Global log level handle for runtime filter changes.
+static GLOBAL_LOG_HANDLE: OnceLock<Arc<LogFilterReloadHandle>> = OnceLock::new();
+
+/// Type alias for the reload handle used to dynamically update log filters.
+pub type LogFilterReloadHandle = reload::Handle<EnvFilter, Registry>;
+
+/// Installs the global log level handle.
+/// Returns `true` if the handle was installed, `false` if one was already installed.
+pub fn install_log_handle(handle: LogFilterReloadHandle) -> bool {
```
DaniPopes: please put this stuff into another module

crates/tracing/src/lib.rs
```diff
@@ -65,9 +65,95 @@ mod layers;
...
+    }
+
+    let filter = EnvFilter::try_new(pattern).map_err(|e| format!("Invalid filter pattern: {e}"))?;
+
+    handle.reload(filter).map_err(|e| e.to_string())
+}
+
+/// Result of initializing tracing.
+#[derive(Debug, Default)]
+pub struct TracingInitResult {
```
DaniPopes: why is this needed?

crates/tracing/src/layers.rs
```diff
@@ -110,6 +110,33 @@ impl Layers {
...
+    ///
+    /// # Arguments
+    /// * `format` - The log message format.
+    /// * `default_directive` - Directive for the default logging level.
+    /// * `filters` - Additional filter directives as a string.
+    /// * `color` - Optional color configuration for the log messages.
+    ///
+    /// # Returns
+    /// An `eyre::Result<LogFilterReloadHandle>` containing the handle to reload the filter.
+    pub(crate) fn stdout_reloadable(
```
DaniPopes: please just pass a bool to the stdout method instead of making a new one

DaniPopes (changes requested):

crates/tracing/src/lib.rs
figtracer: on it

crates/tracing/src/layers.rs
figtracer: on it, didn't catch that!

crates/tracing/src/lib.rs
figtracer: not needed, will keep the Option and remove this wrapper

crates/node/core/src/args/rpc_server.rs
```diff
@@ -776,6 +776,27 @@ impl RpcServerArgs {
         self
     }
 
+    /// Returns `true` if the debug RPC namespace is enabled on any transport.
+    ///
+    /// This is used to determine whether to enable runtime log level changes
+    /// via `debug_verbosity` and `debug_vmodule` RPC methods.
+    pub fn debug_namespace_enabled(&self) -> bool {
```
DaniPopes: can use as arg instead of hardcoding debug 

, ns: RethRpcModule

DaniPopes (commented):
lgtm but only reloads stdout

crates/node/core/src/args/rpc_server.rs
figtracer: nice, on it

crates/node/core/src/args/rpc_server.rs
```diff
@@ -777,6 +776,22 @@ impl RpcServerArgs {
...
+
+        // Check WS API
+        if self.ws_api.as_ref().is_some_and(|api| api.contains(&ns)) {
+            return true;
+        }
+
+        // IPC exposes all modules when enabled
+        !self.ipcdisable
+    }
+
```
figtracer, line 791: this is a bit useless atm

figtracer:
@DaniPopes should we make stdout and file logs independent? might be more useful

crates/tracing/src/log_handle.rs
```diff
@@ -0,0 +1,133 @@
...
+        for handle in &self.handles {
+            let filter = make_filter()?;
+            handle.reload(filter).map_err(|e| e.to_string())?;
+        }
+        Ok(())
+    }
+}
+
+/// Single global log handle shared by all reloadable layers.
+static LOG_HANDLE: OnceLock<std::sync::Mutex<LogFilterHandle>> = OnceLock::new();
```
DaniPopes: this doesn't need the oncelock

DaniPopes (approved):

DaniPopes (approved):

PR #21500, opened 2026-01-27
refactor(engine): buffer state updates and process in parallel

figtracer:
reopens #20844

this buffers state updates and processes all in parallel

less granularity on parallelization as compared to #20635 and following #20842 

avg transaction count per block seems to be around 300, so the lower bound of the benchmarks doesn't look concerning
 
benchmarks:
https://gist.github.com/figtracer/e4e146d1a9d5817139a7c25aebb0c9bd


figtracer:
this has been changed quite a lot

PR #21504, opened 2026-01-28
 docs: add missing debug RPC methods to documentation

eeemmmmmm:
These methods are listed in pruning.mdx tables but were missing from debug.mdx. Only added methods with working implementations, skipped Geth compatibility stubs.


mattsse (commented):

PR #21505, opened 2026-01-28
perf: use FxHashMap in BlockBuffer

davidjsonn:
Replaced std HashMap/HashSet with FxHashMap/FxHashSet in BlockBuffer. Added with_capacity() to avoid reallocations during buffer initialization and in hot paths (remove_old_blocks, remove_children).

This pattern is already used across the codebase in txpool, network, and storage layers. BlockHash keys are ideal for FxHash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8452,6 +8452,7 @@ dependencies = [
  "revm",
  "revm-primitives",
  "revm-state",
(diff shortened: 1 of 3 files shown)


mattsse (commented):
we can keep this as is

PR #21507, opened 2026-01-28
perf(trie): skip empty storage proofs

gakonst:
## Summary

Move the empty storage check earlier in the proof dispatch flow, **before** calling `dispatch_storage_proofs`. This avoids unnecessary work for accounts with empty storage roots by filtering them out upfront, rather than checking inside the dispatch loop.

### Changes

- Filter accounts with empty storage (via `storage_filter.may_have_storage`) before calling `dispatch_storage_proofs`
- Create immediate empty proof responses for filtered accounts
- Pass only non-empty storage accounts to the dispatch function
- Merge empty storage receivers back after dispatching

### Benefits

Accounts known to have no storage are now filtered at the earliest possible point, avoiding:
- Iteration in the dispatch loop
- Channel creation/allocation overhead
- Any potential chunking logic

---

*This is a continuation of the zero-storage-filter optimization work.*


gakonst:
This seems like low hanging fruit from here? Did we see any numbers? @yongkangc plz run bench / try to drive this to completion before Europe is up

PR #21513, opened 2026-01-28
fix: remove redundant receipt vector clone in receipt root task test

prestoalvarez:
Remove unnecessary `.clone()` call in `test_receipt_root_task_single_receipt` by computing the expected root before moving `receipts` into the iterator.


mediocregopher:
test perf doesn't matter

PR #21518, opened 2026-01-28
fix: remove redundant clone

maximevtush:
In `walk_nodes_with_common_prefix`, the loop iterates over `&inputs`, so `v` is already a reference to the value type. Calling `v.clone()` and then immediately taking a reference `&v.clone()` performs an unnecessary copy that does not affect ownership, borrowing, or the data written to the database.


mattsse (commented):
this is just test code we can keep as is

PR #21520, opened 2026-01-28
fix(prune): add minimum 64 block retention for receipts and bodies

joshieDo:
## Problem

With `PruneMode::Full` for receipts (`min_blocks() == 0`), receipts are pruned immediately after persistence. During a same-height reorg, `canonical_block_by_hash` fails with `no receipt found` because receipts were already pruned.

## Fix

- Add `MINIMUM_DISTANCE = 64` for `Receipts` and `Bodies` segments
- Rename `MINIMUM_PRUNING_DISTANCE` → `MINIMUM_UNWIND_SAFE_DISTANCE`

### DIFF
--- a/crates/node/core/src/args/pruning.rs
+++ b/crates/node/core/src/args/pruning.rs
@@ -5,7 +5,9 @@ use alloy_primitives::{Address, BlockNumber};
 use clap::{builder::RangedU64ValueParser, Args};
 use reth_chainspec::EthereumHardforks;
 use reth_config::config::PruneConfig;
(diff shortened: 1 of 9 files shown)


mattsse (approved):

mediocregopher (approved):

crates/node/core/src/args/pruning.rs
```diff
@@ -80,9 +82,9 @@ impl Default for DefaultPruningValues {
                 sender_recovery: Some(PruneMode::Full),
                 transaction_lookup: Some(PruneMode::Full),
                 receipts: Some(PruneMode::Full),
```
shekhirin, line 84: should we also change this?

PR #21524, opened 2026-01-28
fix: simplify UTF-8 decoding in `StreamCodec` by using `Result::ok`

yyakmv:
The previous code manually mapped `Ok(val)` to `Ok(Some(val))` and any `Err(_)` to `Ok(None)`.  
`Result::ok()` already encodes this behavior, so the explicit `match` was redundant and added noise without changing semantics.


mattsse (approved):
fair

PR #21525, opened 2026-01-28
feat(trie): add update_leaves method to SparseTrieExt

yongkangc:
## Summary

Implements RETH-177: Add `update_leaves` method to `SparseTrieExt` that enables batch leaf updates with blinded node handling.

Closes https://linear.app/tempoxyz/issue/RETH-177


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1062,6 +1062,124 @@ impl SparseTrieExt for ParallelSparseTrie {
...
+        let num_upper = ops
+            .iter()
+            .position(|(_, path, _, _)| !SparseSubtrieType::path_len_is_upper(path.len()))
+            .unwrap_or(ops.len());
+
+        let (upper_ops, lower_ops) = ops.split_at(num_upper);
+
+        // Track deduplicated proof requests and failed updates to restore later.
+        let mut requested: HashSet<(Nibbles, u8)> = HashSet::default();
+        let mut failures: Vec<(B256, LeafUpdate)> = Vec::new();
```
mediocregopher: Instead of tracking failures like this, we can just re-insert them directly into `updates`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1094,6 +1212,110 @@ impl ParallelSparseTrie {
...
+            // Removal: empty value triggers leaf deletion. Snapshot the value first so we can
+            // restore it if removal fails due to a blinded node.
+            LeafUpdate::Changed(value) if value.is_empty() => {
+                let old_value = self.get_leaf_value(full_path).cloned();
+                match self.remove_leaf(full_path, NoRevealProvider) {
+                    Ok(()) => Ok(()),
+                    Err(e) => {
+                        if let SparseTrieErrorKind::BlindedNode { path, .. } = e.kind() {
+                            if let Some(old) = old_value {
+                                self.subtrie_for_path_mut(full_path)
```
mediocregopher: I think let's move this into `remove_leaf` itself... that method should handle its own cleanup on error anyway, doing it here is going to result in a bug two years from now when someone modifies `remove_leaf` and doesn't realize cleanup is required 😅

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1094,6 +1212,110 @@ impl ParallelSparseTrie {
...
+            // Insert/update: non-empty value creates or updates the leaf. Track whether the leaf
+            // existed so we can clean up if update fails due to a blinded node.
+            LeafUpdate::Changed(value) => {
+                let existed = self.get_leaf_value(full_path).is_some();
+                match self.update_leaf(*full_path, value, NoRevealProvider) {
+                    Ok(()) => Ok(()),
+                    Err(e) => {
+                        if let SparseTrieErrorKind::BlindedNode { path, .. } = e.kind() {
+                            if !existed {
+                                self.upper_subtrie.inner.values.remove(full_path);
```
mediocregopher: same thing here, it's more of a pain because `update_leaf` has two different spots which can error, but if it errors it should leave the trie in the same state it found it in

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1094,6 +1212,115 @@ impl ParallelSparseTrie {
...
+
+        match update {
+            // Removal: empty value triggers leaf deletion. Snapshot the value first so we can
+            // restore it if removal fails due to a blinded node.
+            LeafUpdate::Changed(value) if value.is_empty() => {
+                let old_value = self.get_leaf_value(full_path).cloned();
+                match self.remove_leaf(full_path, NoRevealProvider) {
+                    Ok(()) => Ok(()),
+                    Err(e) => {
+                        if let SparseTrieErrorKind::BlindedNode { path, .. } = e.kind() {
```
mediocregopher: If `trie_node` returns None then the returned error is `SparseTrieErrorKind::NodeNotFoundInProvider`, not BlindedNode. Same for `update_leaf`

mediocregopher (commented):
Directionally looks correct, just a few nitpicks that are more about style than anything but this is basically right

crates/trie/sparse-parallel/src/trie.rs
yongkangc: ah thats nicer! thanks :) addressed in https://github.com/paradigmxyz/reth/commit/c6c2fe54d38f5b2bd9a77420752f9a4b12481979

crates/trie/sparse-parallel/src/trie.rs
yongkangc: ahh thanks for catch, added new helper `get_retriable_path` to handle both this cases. addressed in c6c2fe54d3

crates/trie/sparse-parallel/src/trie.rs
yongkangc: thats very true 😆 brings back trauma - makes alot of sense

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1094,6 +1298,46 @@ impl ParallelSparseTrie {
...
+            SparseTrieErrorKind::NodeNotFoundInProvider { path } => Some(*path),
+            _ => None,
+        }
+    }
+
+    /// Rolls back a partial update by removing the value, removing any inserted nodes,
+    /// and restoring any modified original node.
+    /// This ensures `update_leaf` is atomic - either it succeeds completely or leaves the trie
+    /// unchanged.
+    fn rollback_update(
```
mediocregopher: super duper nit: let's call this `rollback_insert`

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1015,16 +1127,19 @@ impl SparseTrieExt for ParallelSparseTrie {
             "prune roots must be prefix-free"
         );
 
+        let mut fully_pruned_subtries = [false; NUM_LOWER_SUBTRIES];
```
mediocregopher: Curious why this and the checks to `is_strict_descendant_in(roots_upper, p)` got added back in

mediocregopher (approved):

crates/trie/sparse-parallel/src/trie.rs
yongkangc: sorry about that 🙏🏻

PR #21526, opened 2026-01-28
fix: avoid redundant clone when constructing `TransactionFetcherInfo`

andreypryjmak:
This PR removes an unnecessary `clone().into()` when initializing `TransactionFetcherInfo` in `TransactionFetcher::with_transaction_fetcher_config`.
Instead of cloning `TransactionFetcherConfig` and relying on `From`, `TransactionFetcherInfo` is now constructed directly from the existing scalar fields of `TransactionFetcherConfig`. This keeps the logic equivalent while avoiding a redundant data copy and makes the relationship between the config and its derived info more explicit.


mattsse (commented):
this clone is negligible here

PR #21529, opened 2026-01-28
fix(rpc): update broken execution-apis documentation links

hazeezs:
The old /api-documentation links are dead and return 404. Replaced them with the current execution-apis ETH RPC docs.


mattsse (commented):

PR #21534, opened 2026-01-28
feat(engine): preserve sparse trie across payload validations

mattsse:
## Summary

Implement sparse trie preservation to reduce memory allocations when validating consecutive blocks with parent-child relationship.

### New Types
- `PreservedSparseTrie` - Stores trie with block hash it was computed for
- `SharedPreservedSparseTrie` - Thread-safe handle for passing between tasks

### TreeConfig
- Add `sparse_trie_prune_depth` (default: 4) - depth for trie pruning
- Add `sparse_trie_max_storage_tries` (default: 100) - max storage tries to retain

### SparseStateTrie
- Add `clear()` method - clears state while preserving allocations
- Add `shrink_to()` method - shrinks allocations to limit memory

### SparseTrieTask
- `run()` now takes `&mut self` and returns result only
- `into_trie_for_reuse()` - prunes and shrinks for successful payloads
- `into_cleared_trie()` - clears for failed/cancelled payloads

### Preservation Logic
1. After successful state root computation, prune trie and store with block hash
2. On next payload validation:
   - If `parent_hash == stored_block_hash`: reuse pruned trie (continuation)
   - Otherwise: clear trie but keep allocations
3. On failure/cancellation: clear trie to avoid preserving inconsistent state

## Performance
- Reduces memory allocations for consecutive block validations
- Pruning/cleanup happens after sending result (non-blocking critical path)

Resolves RETH-183


mediocregopher (approved):

crates/engine/primitives/src/config.rs
```diff
@@ -47,6 +47,17 @@ pub const DEFAULT_RESERVED_CPU_CORES: usize = 1;
...
+///
+/// Nodes at this depth and below are converted to hash stubs to reduce memory.
+/// Depth 4 means we keep roughly 16^4 = 65536 potential branch paths at most.
+pub const DEFAULT_SPARSE_TRIE_PRUNE_DEPTH: usize = 4;
+
+/// Default maximum number of storage tries to keep after pruning.
+///
+/// Storage tries beyond this limit are cleared (but allocations preserved).
+pub const DEFAULT_SPARSE_TRIE_MAX_STORAGE_TRIES: usize = 100;
+
```
klkvr, line 60: those are already defined here i think?

https://github.com/paradigmxyz/reth/blob/e9fe0283a9d60b26e733fef0f0b7b3b9b6b88341/crates/trie/sparse/src/lib.rs#L8-L12

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -902,7 +902,16 @@ impl SparseTrie for ParallelSparseTrie {
...
+        match self.updates.take() {
+            Some(updates) => {
+                self.updates = Some(SparseTrieUpdates::with_capacity(
+                    updates.updated_nodes.len(),
+                    updates.removed_nodes.len(),
+                ));
+                updates
+            }
+            None => SparseTrieUpdates::default(),
+        }
```
klkvr, line 915: should we also change this here? https://github.com/paradigmxyz/reth/blob/e9fe0283a9d60b26e733fef0f0b7b3b9b6b88341/crates/trie/sparse/src/trie.rs#L982-L984

klkvr (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -497,72 +512,108 @@ where
...
+                guard.store(PreservedSparseTrie::new(trie, block_hash));
+                return;
+            }
 
-            cleared_sparse_trie.lock().replace(cleared_trie);
+            // Only preserve the trie if computation succeeded - a failed computation
+            // may have left the trie in a partially updated/inconsistent state.
+            let _enter =
+                debug_span!(target: "engine::tree::payload_processor", "preserve").entered();
+            if succeeded {
```
mediocregopher: note that `succeeded` here is only checking that the state root was computed without errors, it's not checking that the computed state root matches the payload's state root. I think we'll want to pass in the expected state root into `spawn_sparse_trie_task` as well

crates/trie/sparse/src/traits.rs
```diff
@@ -252,6 +252,17 @@ pub trait SparseTrieExt: SparseTrie {
...
     fn revealed_node_count(&self) -> usize;
 
+    /// Returns a cheap O(1) size hint for the trie.
+    ///
+    /// This is used as a heuristic for prioritizing which storage tries to keep
+    /// during pruning. Larger values indicate larger tries that are more valuable to preserve.
+    ///
+    /// The default implementation returns `revealed_node_count()`, but implementations
+    /// should override this with a cheaper metric like `nodes.len()` when available.
+    fn size_hint(&self) -> usize {
```
mediocregopher: imo we can get rid of revealed_node_count completely in favor of this, it was only introduced for pruning

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2036,7 +2092,90 @@ impl ParallelSparseTrie {
...
 
             self.lower_subtries[index] = LowerSparseSubtrie::Revealed(subtrie);
+            self.subtrie_heat.mark_modified(index);
+        }
+    }
+}
+
+/// Bitset tracking which of the 256 lower subtries were modified in the current cycle.
+#[derive(Clone, Default, PartialEq, Eq, Debug)]
+struct HotSubtries([u64; 4]);
```
mediocregopher: Super duper nit: can we name this `ModifiedSubtries` or `DirtySubtries` or something? "Hot" seems related to frequency of access when really this is just saying if it was accessed at all.

crates/engine/tree/src/tree/payload_processor/mod.rs
mattsse: I think technically this is already prevented by the blockhash guard

but we could accept the expected root, to avoid doing reuse work

mediocregopher: If state root task failed and but the fallback succeeded we would still end up validating the block, so I don't think the block hash guard is enough

mediocregopher (approved):

mediocregopher (approved):

PR #21535, opened 2026-01-28
fix(prune): respect rocksdb storage settings for history pruning

HuberyJulianay:
History pruning was hardcoded to use MDBX cursors, ignoring `account_history_in_rocksdb` and `storages_history_in_rocksdb` settings. This meant RocksDB history data would never get cleaned up when pruning was enabled.                                                            
                                                                                                                                                                                                                                                                                       Added `prune_account_history_up_to` and `prune_storage_history_up_to` to RocksDBBatch, and updated the prune segments to check storage settings before choosing the backend.

### DIFF
--- a/crates/prune/prune/Cargo.toml
+++ b/crates/prune/prune/Cargo.toml
@@ -51,3 +51,7 @@ reth-testing-utils.workspace = true
 reth-tracing.workspace = true
 
 assert_matches.workspace = true
+
+[features]
+default = []
+rocksdb = ["reth-provider/rocksdb"]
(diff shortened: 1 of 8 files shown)


gakonst:
Closing this PR after review. The approach has several issues:

1. **Trait bound leakage**: The `RocksDBProviderFactory` bound appears to be required unconditionally in some code paths, which would break MDBX-only builds.

2. **Excessive cfg fragmentation**: +842 lines across 8 files with duplicated cfg-gated implementations creates maintenance burden and increases risk of cfg mismatch bugs across the feature matrix.

3. **API stability**: Having two different `build()` methods under different cfgs can cause downstream breakage when call sites need their own conditional compilation.

4. **Unclear runtime behavior**: Not evident that the runtime paths actually use the new RocksDB provider factory when settings indicate RocksDB history storage.

If you'd like to pursue this fix, consider:
- Centralizing backend selection behind a single `HistoryBackendFactory` abstraction
- Ensuring all RocksDB-specific code is strictly behind cfg gates
- Adding integration tests for both rocksdb and non-rocksdb build configurations

PR #21540, opened 2026-01-28
feat(node): WAL Blocks Warning Threshold in AddOns

refcell:
## Summary

This PR extends [#20867](https://github.com/paradigmxyz/reth/pull/20867) by exposing the WAL block warning threshold configuration through the node builder, enabling programmatic configuration for custom node implementations.

### Problem

While #20867 added the ability to configure the WAL warning threshold on `ExExLauncher`, nodes that use the standard builder pattern cannot currently pass this configuration through. The `EngineNodeLauncher` internally calls `ctx.launch_exex()` which uses the default threshold.

For L2 chains with faster block times (e.g., Base with ~2 second blocks vs Ethereum's ~12 seconds), the default threshold of 128 blocks triggers warnings too frequently, creating noise in logs.

### Solution

Add a `wal_blocks_warning` field to the `AddOns` struct and a corresponding builder method on `WithLaunchContext<NodeBuilderWithComponents>`. The `EngineNodeLauncher` reads this value from `AddOns` during launch and applies it to the underlying `ExExLauncher`.

### Usage

```rust
let builder = builder
    .with_types_and_provider::<MyNode, BlockchainProvider<_>>()
    .with_components(node.components())
    .with_add_ons(node.add_ons())
    .with_wal_blocks_warning(768)  // 6x default for ~6x faster block times
    .on_component_initialized(|_ctx| Ok(()));
```

###  Changes

- Added wal_blocks_warning: Option<usize> field to AddOns struct
- Added with_wal_blocks_warning(threshold: usize) method to AddOns
- Added with_wal_blocks_warning(threshold: usize) builder method to WithLaunchContext<NodeBuilderWithComponents>
- Updated EngineNodeLauncher::launch_node() to read the threshold from AddOns and apply it to exex_launcher() before launching
- Added documentation explaining the use case for L2 chains

### DIFF
--- a/crates/node/builder/src/builder/add_ons.rs
+++ b/crates/node/builder/src/builder/add_ons.rs
@@ -4,6 +4,32 @@ use reth_node_api::{FullNodeComponents, NodeAddOns};
 
 use crate::{exex::BoxedLaunchExEx, hooks::NodeHooks};
 
+/// Configuration for node add-ons.
+///
+/// This struct holds additional configuration options that can be extended
+/// with new settings without breaking the `AddOns` struct.
+#[derive(Debug, Clone, Default)]
+pub struct AddOnsConfig {
+    /// The threshold for the number of blocks in the WAL before emitting a warning.
+    ///
+    /// For L2 chains with faster block times, this value should be increased proportionally
+    /// to avoid excessive warnings. For example, a chain with 2-second block times might use
+    /// a value 6x higher than the default (768 instead of 128).
+    pub wal_blocks_warning: Option<usize>,
+}
+
+impl AddOnsConfig {
+    /// Sets the threshold for the number of blocks in the WAL before emitting a warning.
+    ///
+    /// For L2 chains with faster block times, this value should be increased proportionally
+    /// to avoid excessive warnings. For example, a chain with 2-second block times might use
+    /// a value 6x higher than the default (768 instead of 128).
+    pub const fn with_wal_blocks_warning(mut self, threshold: usize) -> Self {
+        self.wal_blocks_warning = Some(threshold);
+        self
+    }
+}
+
 /// Additional node extensions.
 ///
 /// At this point we consider all necessary components defined.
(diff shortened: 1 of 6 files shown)


refcell:
FWIW, this may not be the perfect place to put the wal warning threshold field so happy to change if needed!

crates/node/builder/src/builder/add_ons.rs
```diff
@@ -14,4 +14,22 @@ pub struct AddOns<Node: FullNodeComponents, AddOns: NodeAddOns<Node>> {
     pub exexs: Vec<(String, Box<dyn BoxedLaunchExEx<Node>>)>,
     /// Additional captured addons.
     pub add_ons: AddOns,
+    /// The threshold for the number of blocks in the WAL before emitting a warning.
+    ///
+    /// For L2 chains with faster block times, this value should be increased proportionally
+    /// to avoid excessive warnings. For example, a chain with 2-second block times might use
+    /// a value 6x higher than the default (768 instead of 128).
+    pub wal_blocks_warning: Option<usize>,
```
mattsse, line 18: this will likely break eventually, if we want to introduce additional configs,

breaking this is fine because this doesn't really leak all over, but I think we can make this an extra Config type instead that we then extend with additional settings if any

crates/node/builder/src/launch/engine.rs
```diff
@@ -125,7 +125,11 @@ impl EngineNodeLauncher {
             .with_components(components_builder, on_component_initialized).await?;
 
         // spawn exexs if any
-        let maybe_exex_manager_handle = ctx.launch_exex(installed_exex).await?;
+        let mut exex_launcher = ctx.exex_launcher(installed_exex);
+        if let Some(threshold) = wal_blocks_warning {
+            exex_launcher = exex_launcher.with_wal_blocks_warning(threshold);
+        }
```
mattsse: can we make this fn accept an option instead, then we dont need the if let some here

mattsse (changes requested):
smol nits

crates/node/builder/src/builder/add_ons.rs
refcell: On it!

PR #21544, opened 2026-01-28
perf(engine): make prewarm max concurrency adaptive to CPU cores

gancerlory:
Same pattern as storage_worker_count - scales with available parallelism instead of hardcoded 16. Fixes the inconsistency where storage/account workers already adapt to CPU cores but prewarm concurrency didn't.


mattsse (commented):
This is already controlled via configs, we can keep as is.

PR #21552, opened 2026-01-28
fix: simplify `ToRpcResult` impl by using `Result::map_err`

yyakmv:
Replace `match self { Ok(t) => Ok(t), Err(err) => ... }` with `self.map_err(|err| ...)` in `map_rpc_err`, `map_internal_err_with_data`, and `with_message`.

`map_err` already encodes the same invariant used throughout this helper: preserve `Ok` as-is and only map `Err` into a JSON-RPC `ErrorObject`.


mattsse:
we can keep this as is

PR #21553, opened 2026-01-28
test(stages): add pipeline forward sync and unwind test

joshieDo:
## Summary

Adds a comprehensive pipeline e2e test to `reth-stages` that exercises all stages including account hashing, storage hashing, and history indexing. **This test validates the fix from #21561** - without that fix, this test fails with state root mismatch when run with `--features edge`.

## Test Details

- Pre-funds a signer account and deploys a Counter contract in genesis
- Each block contains two transactions:
  - ETH transfer to a recipient (account state changes)
  - Counter `increment()` call (storage state changes)
- Runs the full pipeline with all stages enabled
- Forward syncs to block 5, then unwinds to block 2
- Verifies all stage checkpoints after sync and unwind
- **With `--features edge`**: Merkle stage validates state root after unwind, catching any issues with storage hashing unwind reading from wrong storage location

### Pipeline Test
- Added `FileClient::from_blocks()` constructor for creating file clients from sealed blocks
- Added pipeline test at `crates/stages/stages/tests/pipeline.rs`
- Added dev-dependencies: `alloy-genesis`, `reth-db-common`, `reth-tracing`
- Enabled `file-client` feature for `reth-downloaders` in dev-deps

### Edge Feature Fix
- Added `reth-db-common/edge` to `reth-stages` edge feature so `init_genesis` uses `StorageSettings::edge()` which writes changesets to static files
- Fixed `create_test_provider_factory_with_node_types` to keep RocksDB temp directory alive by calling `.keep()` (similar to static files dir)

## Validation

To verify the test catches the #21561 bug:
1. Revert commit `effa0ab4c7` (the fix for storage-aware changeset reading)
2. Run `cargo test -p reth-stages --test pipeline --features edge test_pipeline`
3. Test fails with: `mismatched block state root` during Merkle stage unwind

Without the edge feature, the test passes even without the fix because changesets go to MDBX (legacy path).


crates/storage/provider/src/test_utils/mod.rs
```diff
@@ -58,11 +58,13 @@ pub fn create_test_provider_factory_with_node_types<N: NodeTypesForProvider>(
     let (static_dir, _) = create_test_static_files_dir();
     let (rocksdb_dir, _) = create_test_rocksdb_dir();
     let db = create_test_rw_db();
+    // Keep the rocksdb_dir alive for the test duration (similar to static_dir.keep())
```
joshieDo:
```suggestion
```

mattsse (approved):

emmajam (approved):

crates/storage/db-common/src/init.rs
```diff
@@ -254,6 +249,22 @@ where
             .set_block_range(genesis_block_number, genesis_block_number);
     }
 
+    // Changesets are written by `write_state_reverts`, but for non-zero genesis blocks we must
+    // initialize the static file block range here so the writer expects the correct starting block.
```
joshieDo:
```suggestion
```

PR #21561, opened 2026-01-29
fix(provider): use storage-aware methods in unwind_trie_state_from

yongkangc:
## Summary

Fixes UnsortedInput error in `write_storage_history` after stage unwind on RocksDB hot/cold nodes.

## Problem

`unwind_trie_state_from` reads changesets directly from MDBX tables:
```rust
let changed_accounts = self.tx.cursor_read::<tables::AccountChangeSets>()?...
let changed_storages = self.tx.cursor_read::<tables::StorageChangeSets>()?...
```

On RocksDB hot/cold nodes where changesets are stored in **static files**, this returns empty results. This causes:
- History indices not unwound
- Stale keys remain in database
- `UnsortedInput` error when replaying payloads

## Solution

Replace direct MDBX reads with storage-aware trait methods:
- `account_changesets_range(from..)`
- `storage_changesets_range(from..)`

These methods check `storage_changesets_in_static_files` and read from the appropriate source (static files vs MDBX).

RETH-285

### DIFF
--- a/crates/e2e-test-utils/tests/rocksdb/main.rs
+++ b/crates/e2e-test-utils/tests/rocksdb/main.rs
@@ -469,3 +469,123 @@ async fn test_rocksdb_pending_tx_not_in_storage() -> Result<()> {
 
     Ok(())
 }
+
+/// Reorg with `RocksDB`: verifies that unwind correctly reads changesets from
+/// storage-aware locations (static files vs MDBX) rather than directly from MDBX.
+///
+/// This test exercises `unwind_trie_state_from` which previously failed with
+/// `UnsortedInput` errors because it read changesets directly from MDBX tables
+/// instead of using storage-aware methods that check `storage_changesets_in_static_files`.
+#[tokio::test]
+async fn test_rocksdb_reorg_unwind() -> Result<()> {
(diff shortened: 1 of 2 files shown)


gakonst (approved):
**Type Compatibility**
- `account_changesets_range` returns `Vec<(BlockNumber, AccountBeforeTx)>` - matches `unwind_account_hashing` and `unwind_account_history_indices` signatures
- `storage_changesets_range` returns `Vec<(BlockNumberAddress, StorageEntry)>` - matches `unwind_storage_hashing` signature

**Impact on MDBX-only nodes (no static files)**
The trait methods check `storage_changesets_in_static_files` / `account_changesets_in_static_files` flags:
- When **false** (MDBX mode), they fall back to the exact same cursor reads that were previously hardcoded
- No behavioral change for pure MDBX setups

**One minor observation**
The old code used `cursor_read` for storage, the new trait method uses `cursor_dup_read`. Both work for `StorageChangeSets` since it's a `DUPSORT` table - `cursor_dup_read` is technically more correct.

Pending @joshieDo

joshieDo (approved):

PR #21562, opened 2026-01-29
fix(engine): add progress detection to persistence loop

BearSoltokien:
The `persist_until_complete` function could loop indefinitely if persistence tasks fail silently. Add progress tracking to detect when the same number of blocks need to be persisted after multiple iterations, indicating the persistence task is not making progress. Return error after 3 iterations without progress to prevent infinite loops during shutdown.


mattsse (commented):

PR #21571, opened 2026-01-29
fix(provider): check executed block before returning historical state

gakonst:
## Summary

Fixes state provider returning invalid data for blocks that haven't been executed yet.

## Problem

During pipeline sync, `HistoricalStateProvider` could return wrong state for blocks where headers/bodies exist but execution hasn't run. The lookup falls through to `HistoryInfo::InPlainState` and reads from `PlainAccountState` - returning **current** state instead of erroring.

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -825,22 +825,25 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
         self,
         mut block_number: BlockNumber,
     ) -> ProviderResult<StateProviderBox> {
-        // if the block number is the same as the currently best block number on disk we can use the
-        // latest state provider here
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -825,10 +825,19 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
         self,
         mut block_number: BlockNumber,
     ) -> ProviderResult<StateProviderBox> {
-        // if the block number is the same as the currently best block number on disk we can use the
-        // latest state provider here
-        if block_number == self.best_block_number().unwrap_or_default() {
-            return Ok(Box::new(LatestStateProvider::new(self)))
+        let best_block = self.best_block_number()?;
```
```suggestion
        let best_block = self.best_block_number().unwrap_or_default();
```

mattsse (approved):
this makes sense

PR #21579, opened 2026-01-29
fix(engine): Try to always compute storage root in V2 proofs when account proof is present, fallback if not

mediocregopher:
The `AsyncAccountValueEncoder` checks for the presence of a dispatched storage proof channel when it's encoding the storage root for an account. However with partial proofs the storage proof's results might not contain the root node to compute the storage root with. The fix is two-part:

* When the multiproof contains an account proof and storage proof for the same account, ensure the storage proof has at least one proof with min_len == 0 (which will guarantee that the root node is included)

* For cases where an account needs to be encoded but there is not an account proof for it, and there _is_ a storage proof for it which does not include the root, fallback to a synchronous storage proof just to get that proof. A metric is added to track this, it should only happen rarely.


klkvr (approved):

shekhirin (approved):

yongkangc (approved):
Lfg

PR #21583, opened 2026-01-29
feat: sparse trie as cache

klkvr:
Adds a new `SparseTrieCacheTask` that uses the in-memory sparse trie to drive proof fetching instead of relying on the `MultiProofTask`


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -129,6 +148,359 @@ where
...
+    /// Account trie updates.
+    account_updates: B256Map<LeafUpdate>,
+    /// Storage trie updates. hashed address -> slot -> update.
+    storage_updates: B256Map<B256Map<LeafUpdate>>,
+    /// Account updates that are blocked by storage root calculation or account reveal.
+    ///
+    /// Those are being moved into `account_updates` once storage roots
+    /// are revealed and/or calculated.
+    ///
+    /// Invariant: for each entry in `pending_account_updates` account must either be already
```
mediocregopher, line 172: Is this comment right? I thought the account couldn't be in account_updates until we had a storage root, and if we had a storage root the update wouldn't be pending

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -129,6 +148,359 @@ where
...
+                    };
+
+                    match update {
+                        MultiProofMessage::PrefetchProofs(targets) => {
+                            self.on_prewarm_targets(targets);
+                        }
+                        MultiProofMessage::StateUpdate(_, state) => {
+                            self.on_state_update(state);
+                        }
+                        MultiProofMessage::EmptyProof { sequence_number: _, state } => {
```
mediocregopher, line 265: are empty proofs possible if we're bypassing the multiproof task?

mediocregopher (approved):
Two small things but LGTM for merging, we can optimize after

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
klkvr: yeah it could be there as `Touched` or `Changed` with outdated storage root

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
klkvr: yeah it's unreachable now. we can clean it up later via a new message enum or by removing multiprooftask and message entirely

PR #21584, opened 2026-01-29
chore(deps): breaking bumps

DaniPopes:


crates/optimism/rpc/Cargo.toml
```diff
@@ -61,7 +61,7 @@ op-revm.workspace = true
 tokio.workspace = true
 futures.workspace = true
 tokio-stream.workspace = true
-reqwest = { workspace = true, features = ["rustls-tls-native-roots"] }
```
mattsse, line 64: I believe we need to keep rustls-tls-native-roots here because that was an issue with tx forwarding

mattsse (changes requested):

mattsse (approved):

PR #21586, opened 2026-01-29
refactor(bench): share CSV row types between reth-bench and reth-bench-compare

livehardd:
Extract `TotalGasCsvRow` and `CombinedLatencyCsvRow` into reth-bench as shared types, removing duplicate struct definitions from reth-bench-compare. This ensures type safety between CSV writer and reader.


mattsse (commented):
this might break our set up

PR #21592, opened 2026-01-29
feat(dns): add test for bootstrap() with multiple discovery links

conomist:
**Added:** `test_bootstrap_calls_sync_for_all_links` in `reth/crates/net/dns/src/lib.rs`
  - Uses `DnsDiscoveryConfig` with `bootstrap_dns_networks: Some([link1, link2])`
  - Calls `service.bootstrap()` instead of manual `sync_tree_with_link`
  - Asserts both links are synced via `contains_key` with descriptive messages


mattsse (commented):
no reason to change

PR #21593, opened 2026-01-29
feat(static-file): incremental changeset offset storage

gakonst:
## Summary

Replace inline `Vec<ChangesetOffset>` in `SegmentHeader` with a separate `.csoff` sidecar file for **incremental append/prune operations**.

## Problem

Previously, changeset offsets were stored as `Vec<ChangesetOffset>` in `SegmentHeader` and fully rewritten on every commit. For segments with 500k+ blocks, this meant **~8MB rewritten per commit** even when appending a single block.

## Solution

- Store offsets in a separate `.csoff` sidecar file (fixed 16-byte records)
- `SegmentHeader` now stores only `changeset_offsets_len: u64` (count)
- Crash consistency: sidecar is synced before header commit

## Performance Impact

| Operation | Before | After |
|-----------|--------|-------|
| Append 1 block | O(total_blocks) | O(1) (16 bytes) |
| Commit overhead | ~8MB (500k blocks) | ~100 bytes |
| Prune | O(remaining_blocks) | O(1) |
| Lookup | O(1) from memory | O(1) from mmap/pread |

### reth-static-file-types
- Add `ChangesetOffsetsWriter`/`ChangesetOffsetsReader` for sidecar I/O
- Replace `changeset_offsets: Vec` with `changeset_offsets_len: u64` in `SegmentHeader`
- Update serialization to write u64 count instead of Vec

### reth-nippy-jar
- Add `changeset_offsets_path()` for `.csoff` file paths
- Update `delete()` to clean up sidecar files

### reth-provider
- Add `ChangesetOffsetsWriter` to `StaticFileProviderRW`
- Track current block offset during writes, write to sidecar on block completion
- Sync sidecar before header commit (crash consistency)
- Add `read_changeset_offset()`/`read_changeset_offsets()` to `StaticFileJarProvider`
- Update `manager.rs` to read offsets from sidecar file
- Update `truncate_changesets` to truncate sidecar file

## Breaking Change

This changes the static file format for changeset segments. Existing changeset static files are not backwards compatible and must be regenerated.

## Testing

```bash
cargo test -p reth-static-file-types
cargo check -p reth-provider
```

cc @joshieDo @mattsse

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10896,6 +10896,7 @@ dependencies = [
  "serde",
  "serde_json",
  "strum 0.27.2",
+ "tempfile",
 ]
 
 [[package]]
--- a/crates/static-file/types/Cargo.toml
+++ b/crates/static-file/types/Cargo.toml
@@ -24,6 +24,7 @@ strum = { workspace = true, features = ["derive"] }
 reth-nippy-jar.workspace = true
 serde_json.workspace = true
 insta.workspace = true
+tempfile.workspace = true
 
 [features]
 default = ["std"]
--- a/crates/static-file/types/src/changeset_offsets.rs
+++ b/crates/static-file/types/src/changeset_offsets.rs
@@ -1,0 +1,215 @@
+//! Changeset offset sidecar file I/O.
+//!
+//! Provides append-only writing and O(1) random-access reading for changeset offsets.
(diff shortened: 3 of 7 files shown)


gakonst:
Closing due to inconsistencies in base commit. Will re-open after fixing upstream issues.

PR #21595, opened 2026-01-29
fix(ethstats): WSS Handling

MysticRyuujin:
The current ethstats flag hard codes ws:// but if you need to use wss:// you can't and if you try to use a 301 redirect reth will panic and crash:

```text
2026-01-29T19:46:04.870116Z ERROR shutting down due to error
Error: WebSocket error: HTTP error: 301 Moved Permanently

Caused by:
    HTTP error: 301 Moved Permanently

Location:
    /home/runner/work/reth/reth/crates/node/builder/src/launch/common.rs:1131:24
```

This change allows you to add wss:// to the host portion to explicitly use WSS


crates/node/ethstats/src/credentials.rs
```diff
@@ -42,6 +49,16 @@ impl FromStr for EthstatsCredentials {
...
-        Ok(Self { node_id, secret, host })
+        // Detect and strip protocol prefix if present
+        let mut use_tls = false;
+        if let Some(stripped) = host.strip_prefix("wss://") {
+            use_tls = true;
+            host = stripped.to_string();
+        } else if let Some(stripped) = host.strip_prefix("ws://") {
+            use_tls = false;
+            host = stripped.to_string();
+        }
```
mattsse, line 61: oh I see, this would previously only support   hosts like @ethstats-server:3000

checking prefix here makes sense

crates/node/ethstats/src/ethstats.rs
```diff
@@ -102,13 +102,15 @@ where
...
+    /// connection timeouts and errors. Uses either `ws://` or `wss://` based
+    /// on the credentials configuration.
     async fn connect(&self) -> Result<(), EthStatsError> {
         debug!(
             target: "ethstats",
             "Attempting to connect to EthStats server at {}", self.credentials.host
         );
-        let full_url = format!("ws://{}/api", self.credentials.host);
+        let protocol = if self.credentials.use_tls { "wss" } else { "ws" };
+        let full_url = format!("{}://{}/api", protocol, self.credentials.host);
```
mattsse, line 113: we could maybe do the host check here, but doesnt really matter

mattsse (approved):
makes sense, ty

PR #21596, opened 2026-01-29
feat(static-file): incremental changeset offset storage

gakonst:
## Summary

Moves changeset offset storage from inline `Vec<ChangesetOffset>` in `SegmentHeader` to a separate `.csoff` sidecar file. This makes static file appends and commits O(1) instead of O(n) where n is the number of blocks.

## Problem

Previously, each `SegmentHeader` for changeset segments (AccountChangeSets, StorageChangeSets) contained a `Vec<ChangesetOffset>` that grew linearly with block count. Every commit required serializing this entire vector, making appends increasingly slow as the chain grew.

### New Sidecar File Format (`.csoff`)
- Fixed-width 16-byte records: `[offset: u64 LE][num_changes: u64 LE]`
- Append-only writes with O(1) commits
- O(1) random access reads via file seeking

### Changes

**reth-static-file-types**
- New `changeset_offsets.rs` module with `ChangesetOffsetWriter` and `ChangesetOffsetReader`
- `SegmentHeader`: Replace `changeset_offsets: Option<Vec<ChangesetOffset>>` with `changeset_offsets_len: u64`
- Partial record recovery on open (truncates incomplete records after crash)
- Crash-safe reads via `with_len()` that bounds by committed header length

**reth-provider**
- Add `ChangesetOffsetWriter` to `StaticFileProviderRW` for sidecar writes
- Track current block offset during writes, flush to sidecar on block completion
- `flush_current_changeset_offset()` helper ensures last block is persisted before commit
- Sync sidecar **before** header commit (crash consistency ordering)
- Add `read_changeset_offset()` / `read_changeset_offsets()` to `StaticFileJarProvider`
- Update `truncate_changesets()` to truncate sidecar file
- Update `delete_jar()` and `delete_current_and_open_previous()` to handle `.csoff` cleanup
- Fix `finalize()` to call `self.sync_all()` instead of `self.writer.sync_all()`

**reth-nippy-jar**
- Add `CHANGESET_OFFSETS_FILE_EXTENSION = "csoff"` 
- `delete()` now also removes `.csoff` sidecar files

**Feature propagation**
- Add `edge` feature to `reth-exex`, `reth-stages`, `reth-static-file` for proper feature flag propagation

## Testing
- New unit tests: `test_partial_record_recovery`, `test_with_len_bounds_reads`
- New integration test: `test_last_block_flushed_on_commit`
- All existing static-file tests pass
- `test_backfill_with_batch_threshold` passes with `--features edge`

## Breaking Change

This changes the static file format for changeset segments. Existing changeset static files with embedded offsets are not backwards compatible and must be regenerated.

cc @joshieDo @mattsse


yongkangc (commented):
looking through crash/heal scenarios - left some questions

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -336,7 +354,95 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
+                    // Truncate sidecar to valid blocks
+                    csoff_writer.truncate(valid_blocks).map_err(ProviderError::other)?;
+
+                    // Update header to match
+                    self.writer.user_header_mut().set_changeset_offsets_len(valid_blocks);
+
+                    // Also need to update block range if we removed blocks
+                    if valid_blocks < committed_len {
+                        let blocks_removed = committed_len - valid_blocks;
+                        self.writer.user_header_mut().prune(blocks_removed);
```
yongkangc: Since [`prune()`](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/static-file/types/src/segment.rs#L501) already decrements `changeset_offsets_len` internally, doesn't calling `set_changeset_offsets_len(valid_blocks)` followed by `prune(blocks_removed)` result in a double-decrement? i.e. `len = valid_blocks - blocks_removed` instead of just `valid_blocks`?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -684,6 +833,14 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
         // Sync changeset offsets to match the new block range
         self.writer.user_header_mut().sync_changeset_offsets();
 
+        // Truncate the sidecar file to match the new block count
+        if let Some(writer) = &mut self.changeset_offsets {
+            writer.truncate(blocks_to_keep).map_err(ProviderError::other)?;
```
yongkangc, line 936: If a crash occurs after [truncating the sidecar](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/storage/provider/src/providers/static_file/writer.rs#L838) but before [`commit()`](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/storage/provider/src/providers/static_file/writer.rs#L845), won't the sidecar be shorter than the committed header on disk? This would cause `ChangesetOffsetWriter::new()` to error on restart since it expects `sidecar_len >= committed_len`.

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -764,16 +921,36 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
         self.writer = previous_writer;
         self.writer.set_dirty();
-        self.data_path = data_path;
+        self.data_path = data_path.clone();
+
+        // Delete the sidecar file for changeset segments before deleting the main jar
+        if segment.is_change_based() {
+            let csoff_path = current_path.with_extension("csoff");
+            if csoff_path.exists() {
+                std::fs::remove_file(&csoff_path).map_err(ProviderError::other)?;
```
yongkangc, line 1038: If a crash occurs after [deleting the sidecar](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/storage/provider/src/providers/static_file/writer.rs#L940) but before [deleting the main jar](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/storage/provider/src/providers/static_file/writer.rs#L954-L957), won't we have a jar with `changeset_offsets_len > 0` but a missing sidecar file? Should the jar deletion happen first?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -336,7 +354,95 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
+        let csoff_path = self.data_path.with_extension("csoff");
+        let committed_len = self.writer.user_header().changeset_offsets_len();
+        let actual_rows = self.writer.rows() as u64;
+
+        // Open the sidecar writer (this handles partial records and uncommitted tail)
+        let mut csoff_writer =
+            ChangesetOffsetWriter::new(&csoff_path, committed_len).map_err(ProviderError::other)?;
+
+        // If we have offsets, validate the last one doesn't exceed actual rows
+        if committed_len > 0 && actual_rows > 0 {
```
yongkangc: If NippyJar healed down to 0 rows but the sidecar still has offsets pointing to deleted rows, won't this condition skip the necessary truncation? Should the `&& actual_rows > 0` be removed?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -631,32 +780,32 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
+        } else if blocks_to_keep >= changeset_offsets_len {
+            // Keep all rows in this file
             self.writer.rows() as u64
         } else {
-            // Find the offset for the block after last_block
-            // This gives us the number of rows to keep
-            changeset_offsets[blocks_to_keep as usize].offset()
+            // Read offset for the block after last_block from sidecar
+            let mut reader =
+                ChangesetOffsetReader::new(&csoff_path).map_err(ProviderError::other)?;
```
yongkangc: Should this use [`ChangesetOffsetReader::with_len()`](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/static-file/types/src/changeset_offsets.rs#L170) instead? Using `new()` reads physical file length which could include uncommitted tail records from a prior crash.

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -548,7 +674,16 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
                 self.writer = writer;
-                self.data_path = data_path;
+                self.data_path = data_path.clone();
+
+                // Update changeset offsets writer for the new file (starts empty)
+                if segment.is_change_based() {
+                    let csoff_path = data_path.with_extension("csoff");
+                    self.changeset_offsets = Some(
+                        ChangesetOffsetWriter::new(&csoff_path, 0)
+                            .map_err(ProviderError::other)?,
```
yongkangc: If the "next range" jar already exists (e.g., prior crash after partial writes), passing `committed_len=0` here will truncate any valid offsets to zero. Shouldn't this use the existing jar's [`changeset_offsets_len()`](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/static-file/types/src/segment.rs#L373) instead, similar to how `heal_changeset_sidecar()` works on initial open?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -631,32 +780,32 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
         // Find the number of rows to keep (up to and including last_block)
         let blocks_to_keep = if last_block >= expected_block_start {
             last_block - expected_block_start + 1
         } else {
             0
         };
 
+        // Read changeset offsets from sidecar file to find where to truncate
+        let csoff_path = self.data_path.with_extension("csoff");
+        let changeset_offsets_len = self.writer.user_header().changeset_offsets_len();
```
yongkangc, line 885: If [`current_changeset_offset`](https://github.com/paradigmxyz/reth/blob/9edc72e278c5d898115fadc65db91e2985a31776/crates/storage/provider/src/providers/static_file/writer.rs#L228) has unflushed data (in-flight block), the sidecar read here won't include it. Should this call `flush_current_changeset_offset()` first to ensure the sidecar is up-to-date before computing `rows_to_keep`?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -2595,9 +2608,17 @@ impl<N: NodePrimitives> StorageChangeSetReader for StaticFileProvider<N> {
...
-                    for offset in changeset_offsets {
+            for (block_range, header) in changeset_segments {
+                let csoff_path = self
+                    .path
+                    .join(StaticFileSegment::StorageChangeSets.filename(block_range))
+                    .with_extension("csoff");
+                if csoff_path.exists() {
+                    let len = header.changeset_offsets_len();
+                    let mut reader = ChangesetOffsetReader::new(&csoff_path, len)
+                        .map_err(ProviderError::other)?;
```
Rjected, line 2619: nit / possible followup: wonder if we should add a helper for returning a reader for a segment

crates/static-file/types/src/segment.rs
```diff
@@ -608,21 +592,22 @@ impl SegmentHeader {
...
 
-    /// Returns the `ChangesetOffset` corresponding for the given block, if it's in the block
-    /// range.
+    /// Returns the index into the sidecar file for a given block's changeset offset.
     ///
-    /// If it is not in the block range or the changeset list in the header does not contain a
-    /// value for the block, this returns `None`.
-    pub fn changeset_offset(&self, block: BlockNumber) -> Option<&ChangesetOffset> {
+    /// Returns `None` if the block is not in the block range.
+    /// The caller must read the actual offset from the sidecar file at this index.
```
Rjected:
```suggestion
    /// To get the changeset offset, the caller must read the offset from the sidecar file at this index.
```

crates/static-file/types/src/segment.rs
```diff
@@ -393,9 +400,20 @@ impl SegmentHeader {
         self.tx_range
     }
 
-    /// Returns the changeset offsets.
-    pub const fn changeset_offsets(&self) -> Option<&Vec<ChangesetOffset>> {
-        self.changeset_offsets.as_ref()
+    /// Returns the number of changeset offset entries.
+    /// The actual offsets are stored in the .csoff sidecar file.
```
Rjected:
```suggestion
    /// The actual offsets are stored in the `.csoff` sidecar file.
```

Rjected (approved):
just have nits, looked hard at `set_len` / `sync_all` / healing to see if we would have any races or other issues due to crashing mid-write or something similar, but could not find anything

yongkangc (approved):

PR #21599, opened 2026-01-29
feat(reth-bench): add wait options to new-payload-only command

HuberyJulianay:
Adds `--wait-time`, `--wait-for-persistence`, and `--persistence-threshold` options to `new-payload-only` command, Following the existing options in `new-payload-fcu`. 

Allows benchmarking newPayload-only scenarios with persistence waiting.


Rjected:
`new-payload-only` will never cause the node to persist so these don't really make sense i think

PR #21605, opened 2026-01-29
fix(provider): prune account changesets from static files in remove_state_above

joshieDo:
## Summary

Fixes a bug where `remove_state_above` only pruned account changesets from MDBX tables but not from static files when the `edge` feature was enabled.

## Problem

When account changesets are configured to be stored in static files (via `edge` feature), the `remove_state_above` function was not properly pruning them during reorgs. This caused `UnexpectedStaticFileBlockNumber` errors:

```
err=ProviderError(UnexpectedStaticFileBlockNumber(AccountChangeSets, 21, 100))
```

The sequence:
1. Chain advances to block 100, writing account changesets to static files
2. Reorg happens, `remove_state_above(20)` is called
3. Static file writer still has `block_end = 99`, expecting next block to be 100
4. New chain tries to write block 21 changesets
5. `increment_block(21)` fails because it expects block 100

## Solution

This fix mirrors the existing storage changeset handling (lines 2501-2512) to also check if account changesets are in static files and call `prune_account_changesets(block)` to properly truncate them during unwinding.

## Testing

- All existing changeset tests pass with `--features edge`
- Fixes the `test_long_reorg` test failure in CI

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -2498,10 +2498,7 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
             block_bodies.first().expect("already checked if there are blocks").first_tx_num();
 
         let storage_range = BlockNumberAddress::range(range.clone());
-        let storage_changeset = if let Some(_highest_block) = self
-            .static_file_provider
-            .get_highest_static_file_block(StaticFileSegment::StorageChangeSets) &&
-            self.cached_storage_settings().storage_changesets_in_static_files
+        let storage_changeset = if self.cached_storage_settings().storage_changesets_in_static_files
         {
             let changesets = self.storage_changesets_range(range.clone())?;
             let mut changeset_writer =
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -2511,7 +2511,19 @@ impl<TX: DbTxMut + DbTx + 'static, N: NodeTypesForProvider> StateWriter
...
+            self.cached_storage_settings().account_changesets_in_static_files
+        {
+            let changesets = self.account_changesets_range(range)?;
+            let mut changeset_writer =
+                self.static_file_provider.latest_writer(StaticFileSegment::AccountChangeSets)?;
+            changeset_writer.prune_account_changesets(block)?;
+            changesets
+        } else {
+            self.take::<tables::AccountChangeSets>(range)?
+        };
```
Rjected: lets just use `self.cached_storage_settings().account_changesets_in_static_files` instead of the highest block && condition? since we dont even use the highest block here

Rjected (approved):
lgtm

PR #21612, opened 2026-01-29
perf: cache fetched proof targets in `SparseTrieCacheTask`

klkvr:
Based on https://github.com/paradigmxyz/reth/pull/21611

Implements cache for fetched proof targets so that we don't fetch same account/storage proofs over and over.


mediocregopher:
this lgtm as a first step, one thing to note though is that one proof can be covered by others with different keys. E.g. given these already dispatched targets:

- 0x123abc (min_len:3)
- 0x123def (min_len:4)

If the following proof target came in:
- 0x123def (min_len:3)

The 0x123abc proof target is already covering 0x123, and the existing 0x123def target is covering the rest of the key, so 0x123def (min_len:3) can be ignored.

I don't think this is an optimization we should make right now, but just something to keep in mind.

mediocregopher (approved):

PR #21614, opened 2026-01-30
feat(prune): add --delete-limit flag to control memory usage (temp fix for prune)

yongkangc:
## Summary

Adds a configurable `--delete-limit` flag to the `reth prune` command to control memory usage during pruning operations.

## Motivation

The `reth prune` command causes OOM with ~50GB RSS on archive nodes. The root cause is that `delete_limit(usize::MAX)` causes walker loops to collect all entries into memory before processing, which is problematic for nodes with hundreds of millions of entries.

## Solution

- Add `--delete-limit` CLI argument with a sensible default of 100,000 entries per segment
- Use `0` to mean unlimited (backwards compatible with legacy behavior)
- Log the delete limit for visibility during pruning

## Changes

- Added `const DEFAULT_DELETE_LIMIT: usize = 100_000`
- Added `--delete-limit` CLI argument with documentation
- Updated pruner builder to use configurable limit
- Enhanced logging to show the configured delete limit

## Testing

Tested on dev-yk with 2.5TB RocksDB datadir - memory usage stayed under control with the new default limit.

### DIFF
--- a/crates/cli/commands/src/prune.rs
+++ b/crates/cli/commands/src/prune.rs
@@ -1,4 +1,4 @@
-//! Command that runs pruning without any limits.
+//! Command that runs pruning with a configurable delete limit.
 use crate::common::{AccessRights, CliNodeTypes, EnvironmentArgs};
 use clap::Parser;
 use reth_chainspec::{ChainSpecProvider, EthChainSpec, EthereumHardforks};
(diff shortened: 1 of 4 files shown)


crates/cli/commands/src/prune.rs
```diff
@@ -68,14 +68,50 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> PruneComma
 
         // Delete data which has been copied to static files.
         if let Some(prune_tip) = lowest_static_file_height {
-            info!(target: "reth::cli", ?prune_tip, ?config, "Pruning data from database...");
-            // Run the pruner according to the configuration, and don't enforce any limits on it
+            let limit_display = if delete_limit == usize::MAX {
+                "unlimited".to_string()
```
yongkangc: this feels kinda weird, isnt there a more idiomatic way to do that?

crates/cli/commands/src/prune.rs
```diff
@@ -1,62 +1,62 @@
...
-                    target_triple: version_metadata().vergen_cargo_target_triple.as_ref(),
-                    build_profile: version_metadata().build_profile_name.as_ref(),
-                },
-                ChainSpecInfo { name: provider_factory.chain_spec().chain().to_string() },
-                ctx.task_executor,
-                metrics_hooks(&provider_factory),
-                data_dir.pprof_dumps(),
-            );
+        // Check if static files are involved in this node's storage configuration
+        let settings = provider_factory.cached_storage_settings();
```
yongkangc: cant we use like reth edge kind of flag? see the codebase for reference for how this is usually done? this flag

PR #21618, opened 2026-01-30
fix(prune): use internal batching to prevent OOM in history pruning

yongkangc:
## Summary

Fixes OOM when running `reth prune` with edge flag (static files). **Fixes RETH-291**

## Problem

The code comments claim OOM is impossible because HashMap size is bounded by `prune_delete_limit * blocks_since_last_run / TABLES_TO_PRUNE` (typically `3500 * 5 / 2 = 8750` entries).

However, `reth prune` CLI sets `delete_limit(usize::MAX)`:

1. **Limit calculation**: `usize::MAX / 2` remains effectively unlimited
2. **Loop behavior**: `limiter.is_limit_reached()` never returns true
3. **Iteration**: Loop processes all changesets in range without breaking

**Measured from mainnet static files (blocks 0-24M via `reth db stats`):**

| Metric | Account History | Storage History |
|--------|-----------------|-----------------|
| Static file changeset entries | 4.24 billion | 7.83 billion |
| History index shards (RocksDB) | 397 million | 1.92 billion |
| Sharding factor | ~10.7x | ~4.1x |
| **Estimated unique keys** | **~37M addresses** | **~469M (addr,slot) pairs** |
| HashMap entry size | 28 bytes | 60 bytes |
| **Estimated memory** | **~1.0 GB** | **~28.1 GB** |

Total estimated HashMap memory: **~29 GB** (actual OOM observed at 55GB includes other allocations and HashMap overhead).

## Solution

Add `MAX_*_HISTORY_ENTRIES_PER_RUN` constants (200,000) to cap entries per batch:

```rust
let capped = limit.min(MAX_ACCOUNT_HISTORY_ENTRIES_PER_RUN);
input.limiter.set_deleted_entries_limit(capped / ACCOUNT_HISTORY_TABLES_TO_PRUNE)
```

This bounds HashMap to ~100k entries (~4MB) per `prune()` call. The outer pruner loop calls `prune()` repeatedly until done.

Only `prune_static_files()` is modified. `prune_database()` already has bounded iteration via `prune_table_with_range`.


crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -553,4 +563,131 @@ mod tests {
         test_prune(998, 2, (PruneProgress::Finished, 500));
         test_prune(1200, 3, (PruneProgress::Finished, 202));
     }
+
+    /// Test that verifies the pruner caps HashMap growth even when delete_limit is unbounded.
```
yongkangc: this test willl be removed

crates/prune/prune/src/segments/user/account_history.rs
```diff
@@ -101,52 +109,100 @@ impl AccountHistory {
...
-        // The size of this map it's limited by `prune_delete_limit * blocks_since_last_run /
-        // ACCOUNT_HISTORY_TABLES_TO_PRUNE`, and with the current defaults it's usually `3500 * 5 /
-        // 2`, so 8750 entries. Each entry is `160 bit + 64 bit`, so the total size should be up to
-        // ~0.25MB + some hashmap overhead. `blocks_since_last_run` is additionally limited by the
-        // `max_reorg_depth`, so no OOM is expected here.
-        let mut highest_deleted_accounts = FxHashMap::default();
-        let mut last_changeset_pruned_block = None;
-        let mut pruned_changesets = 0;
-        let mut done = true;
+        const BATCH_SIZE: usize = MAX_ACCOUNT_HISTORY_ENTRIES_PER_RUN;
```
yongkangc: should we mve this up? or remove `BATCH_SIZE` const seems uncessary

crates/prune/prune/src/segments/user/storage_history.rs
```diff
@@ -101,56 +110,107 @@ impl StorageHistory {
...
-        // The size of this map is limited by `prune_delete_limit * blocks_since_last_run /
-        // STORAGE_HISTORY_TABLES_TO_PRUNE`, and with current defaults it's usually `3500 * 5
-        // / 2`, so 8750 entries. Each entry is `160 bit + 256 bit + 64 bit`, so the total
-        // size should be up to ~0.5MB + some hashmap overhead. `blocks_since_last_run` is
-        // additionally limited by the `max_reorg_depth`, so no OOM is expected here.
-        let mut highest_deleted_storages = FxHashMap::default();
-        let mut last_changeset_pruned_block = None;
-        let mut pruned_changesets = 0;
-        let mut done = true;
+        const BATCH_SIZE: usize = MAX_STORAGE_HISTORY_ENTRIES_PER_RUN;
```
yongkangc: should we mve this up? or remove BATCH_SIZE const seems uncessary

Rjected (changes requested):
hmm I wonder if it's simpler just to change the `prune` command to run in a loop, but use the limiter, something like:
```rust

// Run the pruner according to the configuration, enforce a limit to limit memory
let mut pruner = PrunerBuilder::new(config)
    .delete_limit(new_limit)
    .build_with_provider_factory(provider_factory);

// loop which runs prune until it doesn't return HasMoreData
```

yongkangc:
Closing this for a simpler fix
https://github.com/paradigmxyz/reth/pull/21649

PR #21622, opened 2026-01-30
feat(txpool): add EIP-7594 blob sidecar toggle

MatusKysel:
Add `no_eip7594()` and `set_eip7594()` builder methods to                                                                                                                                                       `EthTransactionValidatorBuilder`, allowing chains to disable                                                                                                                                                       
  EIP-7594 (`PeerDAS`) blob sidecar support.                                                                                                                                                                         
                                                                                                                                                                                                                   
  When disabled, EIP-7594 (v1) blob sidecars are always rejected                                                                                                                                                   and EIP-4844 (v0) sidecars are always accepted, regardless of                                                                                                                                                   Osaka fork activation.                                                                                                                                                                                           
                                                                                                                                                                                                                   
  This is useful for chains that do not adopt EIP-7594.

### DIFF
--- a/crates/transaction-pool/src/validate/eth.rs
+++ b/crates/transaction-pool/src/validate/eth.rs
@@ -99,6 +99,10 @@ pub struct EthTransactionValidator<Client, T, Evm> {
     validation_metrics: TxPoolValidationMetrics,
     /// Bitmap of custom transaction types that are allowed.
     other_tx_types: U256,
+    /// Whether EIP-7594 blob sidecars are accepted.
+    /// When false, EIP-7594 (v1) sidecars are always rejected and EIP-4844 (v0) sidecars
+    /// are always accepted, regardless of Osaka fork activation.
+    eip7594: bool,
 }
 
 impl<Client, Tx, Evm> EthTransactionValidator<Client, Tx, Evm> {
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/validate/eth.rs
```diff
@@ -690,16 +694,27 @@ where
...
                             return Err(InvalidPoolTransactionError::Eip4844(
-                                Eip4844PoolTransactionError::UnexpectedEip4844SidecarAfterOsaka,
+                                Eip4844PoolTransactionError::UnexpectedEip7594SidecarBeforeOsaka,
+                            ))
+                        }
+                    } else {
+                        // EIP-7594 disabled: always reject v1 sidecars, accept v0
+                        if sidecar.is_eip7594() {
+                            return Err(InvalidPoolTransactionError::Eip4844(
+                                Eip4844PoolTransactionError::UnexpectedEip7594SidecarBeforeOsaka,
```
mattsse: this error isn't quite accurate, can we introduce a dedicated error for this, Eip7594Sidecar disallowed

mattsse (changes requested):
makes sense, smol nit

mattsse (approved):

PR #21623, opened 2026-01-30
feat(rpc): add EIP-7934 block size validation to testing_buildBlockV1

gakonst:
## Summary

Add Osaka block size limit (EIP-7934) validation to the `testing_buildBlockV1` RPC endpoint.

## Motivation

EIP-7934 introduces a maximum RLP block size of 8 MiB (10 MiB total minus 2 MiB safety margin) starting with the Osaka hardfork. This limit is already enforced in the ethereum payload builder during transaction addition, but was missing from the testing RPC endpoint.

## Changes

- Added `ChainSpecProvider` trait bound to access chain spec and check Osaka hardfork status
- Track `block_transactions_rlp_length` across iterations during block building
- Before executing each transaction, estimate block size with `block_transactions_rlp_length + tx_rlp_len + withdrawals_rlp_length + 1024` (1KB header overhead)
- Skip/reject transactions that would exceed `MAX_RLP_BLOCK_SIZE`
- After successful execution, increment `block_transactions_rlp_length`

This follows the same pattern as the ethereum payload builder in `crates/ethereum/payload/src/lib.rs`.

## Testing

- Existing E2E test in `crates/ethereum/node/tests/it/testing.rs` validates the endpoint works

## References

- [EIP-7934: RLP Execution Block Size Limit](https://eips.ethereum.org/EIPS/eip-7934)

### DIFF
--- a/crates/rpc/rpc/src/testing.rs
+++ b/crates/rpc/rpc/src/testing.rs
@@ -16,11 +16,14 @@
 
 use alloy_consensus::{Header, Transaction};
 use alloy_eips::eip2718::Decodable2718;
-use alloy_evm::Evm;
+use alloy_evm::{Evm, RecoveredTx};
 use alloy_primitives::{map::HashSet, Address, U256};
+use alloy_rlp::Encodable;
 use alloy_rpc_types_engine::ExecutionPayloadEnvelopeV5;
 use async_trait::async_trait;
 use jsonrpsee::core::RpcResult;
+use reth_chainspec::{ChainSpecProvider, EthereumHardforks};
+use reth_consensus_common::validation::MAX_RLP_BLOCK_SIZE;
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/testing.rs
```diff
@@ -123,6 +136,36 @@ where
                         continue;
                     }
 
+                    // EIP-7934: Check estimated block size before adding transaction
+                    let tx_rlp_len = tx.tx().length();
+                    if is_osaka {
+                        // 1KB overhead for block header
+                        let estimated_block_size = block_transactions_rlp_length +
```
mattsse, line 143: we could avoid recomputing these per tx, but also doesnt really matter

mattsse (approved):

PR #21624, opened 2026-01-30
feat(rpc): make blob sidecar upcasting opt-in

gakonst:
## Summary

Makes the forced EIP-4844 to EIP-7594 blob sidecar upcasting opt-in via a new CLI flag.

## Motivation

Currently, blob transactions with EIP-4844 sidecars are automatically converted to EIP-7594 format when Osaka is active. This change makes that behavior opt-in, allowing users to choose whether they want automatic conversion.

## Changes

- Add `--rpc.force-blob-sidecar-upcasting` flag to `RpcServerArgs` (defaults to `false`)
- Add `force_blob_sidecar_upcasting` field to `EthConfig` with builder method
- Wire the setting through `EthApiBuilder` to `EthApiInner`
- Check the flag in `send_transaction` before performing the conversion

## Testing

The change is additive and preserves existing behavior when the flag is enabled.


mattsse (approved):
lgtm

PR #21625, opened 2026-01-30
perf(net): optimize block fetcher queue to O(1) for high-priority requests

bobtajson:
Replaces the single request queue in `StateFetcher` with separate high-priority and normal-priority queues to eliminate O(n) insertion cost when handling high-priority block download requests under load.


mattsse (commented):
we can keep as is

PR #21629, opened 2026-01-30
fix(trie): cleanup modified branch masks in update_leaf on reveal failure

mediocregopher:
Another atomicity fix for update_leaf


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -1316,6 +1328,10 @@ impl ParallelSparseTrie {
                 }
             }
         }
+        // Remove any branch masks that were inserted
```
mattsse, line 1331: I see

mattsse (approved):

PR #21639, opened 2026-01-30
chore(trie): add set_* methods alongside with_* builders

DaniPopes:
Adds `set_*` methods that take `&mut self` alongside the existing `with_*` builder methods on `SparseTrie` trait and `SparseStateTrie`:

- `set_root` / `with_root`
- `set_updates` / `with_updates`  
- `set_accounts_trie` / `with_accounts_trie`
- `set_default_storage_trie` / `with_default_storage_trie`

The `with_*` methods are now default trait implementations that call the `set_*` methods, allowing in-place mutation when the builder pattern isn't needed.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -152,27 +152,25 @@ impl Default for ParallelSparseTrie {
 }
 
 impl SparseTrie for ParallelSparseTrie {
-    fn with_root(
-        mut self,
+    fn set_root(
+        &mut self,
         root: TrieNode,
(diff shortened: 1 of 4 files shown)


mattsse (commented):
why do we need this?

DaniPopes:
with_ on big structs is not efficient, copies all the bytes all the time and looks odd when not in a builder situation, like here
```rust
*revealed_trie = revealed_trie.with_root(root, masks, retain_updates)?;
```
vs
```rust
revealed_trie.set_root(root, masks, retain_updates)?;
```

mattsse (approved):

PR #21640, opened 2026-01-30
fix(engine): wait for persistence service thread before RocksDB drop

gakonst:
## Summary

Fixes flaky persistence tests with `pthread lock: Invalid argument` errors on the `edge` feature (RocksDB backend).

## Root Cause

Race condition during test cleanup:
1. Test ends, `PersistenceHandle` (sender) is dropped
2. `ProviderFactory` gets dropped, triggering RocksDB destructor  
3. RocksDB calls `cancel_all_background_work(true)`
4. Meanwhile the persistence service thread is still winding down
5. Race between RocksDB shutdown and persistence thread accessing mutexes -> pthread error

## Solution

`PersistenceHandle` now holds an `Arc<ServiceGuard>` that joins the service thread when dropped:

```rust
pub struct PersistenceHandle<N: NodePrimitives = EthPrimitives> {
    sender: Sender<PersistenceAction<N>>,
    _service_guard: Arc<ServiceGuard>,
}

struct ServiceGuard(Option<JoinHandle<()>>);

impl Drop for ServiceGuard {
    fn drop(&mut self) {
        if let Some(join_handle) = self.0.take() {
            let _ = join_handle.join();
        }
    }
}
```

When all clones of `PersistenceHandle` are dropped (e.g., when the tree handler that holds it is dropped), the `Arc` refcount hits zero, `ServiceGuard::drop` runs, and the service thread is joined before RocksDB is released.

## Caveat

`join()` can block indefinitely if the persistence service hangs, and thread panics are silently ignored. This is acceptable since a hung service indicates a bug, but could be improved by logging panics.

## Changes

- Added `ServiceGuard` struct that joins the thread on drop
- `PersistenceHandle` now holds `Arc<ServiceGuard>` (remains `Clone`)
- `spawn_service` returns `PersistenceHandle` directly (no wrapper type)
- `PersistenceHandle::new()` preserved for test mocking (creates empty guard)

Closes the flaky test issue in https://github.com/paradigmxyz/reth/actions/runs/21525951607/job/62029485190#step:9:1240

### DIFF
--- a/crates/engine/service/src/service.rs
+++ b/crates/engine/service/src/service.rs
@@ -61,9 +61,6 @@ where
     Client: BlockClient<Block = BlockTy<N>> + 'static,
 {
     orchestrator: EngineServiceType<N, Client>,
-    /// Handle to the persistence service - kept to ensure graceful shutdown.
-    /// The handle joins the service thread when dropped.
-    _persistence_handle: PersistenceHandle<N::Primitives>,
 }
 
 impl<N, Client> EngineService<N, Client>
@@ -108,7 +105,7 @@ where
             blockchain_db,
             consensus,
             payload_validator,
-            persistence_handle.clone(),
+            persistence_handle,
             payload_builder,
             canonical_in_memory_state,
             tree_config,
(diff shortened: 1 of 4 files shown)


joshieDo (commented):
.

crates/engine/service/src/service.rs
```diff
@@ -14,7 +14,7 @@ pub use reth_engine_tree::{
     chain::{ChainEvent, ChainOrchestrator},
     engine::EngineApiEvent,
 };
-use reth_ethereum_primitives::EthPrimitives;
+
```
joshieDo:
```suggestion
```

DaniPopes (approved):

mattsse (approved):

PR #21641, opened 2026-01-30
refactor(db): derive Clone for DatabaseEnv

gakonst:
## Summary
Derives Clone for DatabaseEnv and replaces all Arc<DatabaseEnv> with DatabaseEnv.

## Motivation
DatabaseEnv already uses Arc internally for its Environment and dbis fields, so there's no need to wrap it in Arc<> externally.

## Changes
- Added derive(Clone) to DatabaseEnv
- Replaced all Arc<DatabaseEnv> with DatabaseEnv throughout the codebase (23 files)

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/mod.rs
+++ b/crates/storage/db/src/implementation/mdbx/mod.rs
@@ -645,10 +645,7 @@ mod tests {
 
     /// Create database for testing
     fn create_test_db(kind: DatabaseEnvKind) -> DatabaseEnv {
-        Arc::new(create_test_db_with_path(
-            kind,
-            &tempfile::TempDir::new().expect(ERROR_TEMPDIR).keep(),
-        ))
+        create_test_db_with_path(kind, &tempfile::TempDir::new().expect(ERROR_TEMPDIR).keep())
     }
 
     /// Create database for testing with specified path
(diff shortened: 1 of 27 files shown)


crates/storage/db/src/implementation/mdbx/mod.rs
```diff
@@ -222,7 +222,7 @@ impl DatabaseArguments {
 }
 
 /// Wrapper for the libmdbx environment: [Environment]
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct DatabaseEnv {
     /// Libmdbx-sys environment.
     inner: Environment,
```
mattsse, line 228: this clone is now a bunch of individual arc clones

dont think this matters really, we should rather wrap all fields in 

https://github.com/paradigmxyz/reth/blob/29072639d6cf38063fea9ff8505c44c5ac4f3b95/crates/storage/provider/src/providers/blockchain_provider.rs#L48-L52

mattsse (approved):
I like this more, the clone overhead is kinda negligible I think


up to @DaniPopes

crates/storage/db/src/implementation/mdbx/mod.rs
DaniPopes: thats fine

DaniPopes (approved):

PR #21644, opened 2026-01-30
perf(trie): reuse account RLP buffer in SparseTrieCacheTask

gakonst:
## Summary
Adds a reusable `account_rlp_buf` field to `SparseTrieCacheTask`, pre-allocated with `TRIE_ACCOUNT_RLP_MAX_SIZE`, to avoid repeated heap allocations when encoding accounts.

## Changes
- Added `account_rlp_buf: Vec<u8>` field to `SparseTrieCacheTask`
- Pre-allocate buffer with `TRIE_ACCOUNT_RLP_MAX_SIZE` in constructor
- Replace `alloy_rlp::encode()` calls with `clear(); encode(&mut buf); buf.clone()`
- Removed TODO comments

## Testing
`cargo check -p reth-engine-tree` and clippy pass.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -530,21 +530,18 @@ where
         }
 
         // Now handle pending account updates that can be upgraded to a proper update.
-        let storage_updates = &self.storage_updates;
-        let account_updates = &mut self.account_updates;
-        let trie = &mut self.trie;
         let account_rlp_buf = &mut self.account_rlp_buf;
         self.pending_account_updates.retain(|addr, account| {
             // If account has pending storage updates, it is still pending.
-            if storage_updates.get(addr).is_some_and(|updates| !updates.is_empty()) {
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -513,28 +517,34 @@ where
...
                     self.account_updates.insert(*addr, LeafUpdate::Changed(encoded));
                 }
             }
         }
 
         // Now handle pending account updates that can be upgraded to a proper update.
+        let storage_updates = &self.storage_updates;
+        let account_updates = &mut self.account_updates;
+        let trie = &mut self.trie;
+        let account_rlp_buf = &mut self.account_rlp_buf;
```
mattsse: why do we need to do this, this shouldnt conflict with the borrow checker because we can access the fields separately

mattsse (changes requested):

mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):

PR #21647, opened 2026-01-30
fix(cli): delete all static files when PruneModes::Full is configured

Rjected:
Right now `reth prune` does not actually prune all static files where `PruneModes::Full` is configured

Also skips the consistency check when there is a prune checkpoint with `Full`

### DIFF
--- a/crates/prune/prune/src/segments/receipts.rs
+++ b/crates/prune/prune/src/segments/receipts.rs
@@ -15,13 +15,16 @@ use reth_provider::{
     NodePrimitivesProvider, PruneCheckpointWriter, StaticFileProviderFactory, StorageSettingsCache,
     TransactionsProvider,
 };
-use reth_prune_types::{PruneCheckpoint, PruneSegment, SegmentOutput, SegmentOutputCheckpoint};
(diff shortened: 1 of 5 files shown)


crates/prune/prune/src/segments/receipts.rs
```diff
@@ -33,6 +36,16 @@ where
 {
     if EitherWriter::receipts_destination(provider).is_static_file() {
         debug!(target: "pruner", "Pruning receipts from static files.");
+
+        if mode.is_full() {
```
joshieDo: https://github.com/paradigmxyz/reth/blob/3d699ac9c6676ca7af669dd41819f53bfa779285/crates/prune/types/src/segment.rs#L66-L77

i think for receipts we cant really do it, because we expect some to exist

joshieDo: https://github.com/paradigmxyz/reth/pull/21677/files relevant as well, so this is never even reachable

Rjected: will remove this branch then

joshieDo (approved):

PR #21648, opened 2026-01-30
perf(trie): reuse proof nodes buffer in `reveal_nodes`

gakonst:
## Summary

Changes `reveal_nodes` signature from `Vec<ProofTrieNode>` to `&mut [ProofTrieNode]` to enable caller-side buffer preservation and ensures expensive drops happen after lock release.

Adds `DeferredDrops` struct that collects all `Vec<ProofTrieNode>` accumulated during the block, and drops them all at once between blocks.

## Motivation

The previous signature required callers to pass owned Vec<ProofTrieNode>, which meant:
* No buffer reuse across payload runs
* Expensive deallocations happened while holding the `preserved_sparse_trie` lock

## Lock/drop ordering

`into_trie_for_reuse` and `into_cleared_trie` now return `(SparseStateTrie, DeferredDrops)`.

The guard is explicitly dropped before `DeferredDrops`, ensuring expensive deallocations happen outside the lock.

## Benchmark
```
Metric        | Baseline         | Feature          | Change    | Threshold
--------------+------------------+------------------+-----------+-----------
Mean Latency  |          33.07ms |          32.24ms |   -2.5% ≈ |    ±2.97%
Std Dev       |          15.83ms |          15.23ms |           |
P50 Latency   |          29.04ms |          28.57ms |  -1.65% ≈ |    ±4.24%
P90 Latency   |          54.99ms |          52.65ms |  -4.25%   |
P99 Latency   |          79.17ms |          77.16ms |  -2.53%   |
--------------+------------------+------------------+-----------+-----------
Gas/sec       |    921.51 Mgas/s |     945.1 Mgas/s |  +2.56% ≈ |    ±2.97%
```

### DIFF
--- a/crates/trie/sparse/src/state.rs
+++ b/crates/trie/sparse/src/state.rs
@@ -54,7 +54,7 @@ pub struct SparseStateTrie<
     ///
     /// Each reveal operation creates a new buffer, uses it, and pushes it here.
     /// These are collected via [`Self::take_deferred_drops`] and dropped after locks are released.
-    proof_nodes_bufs: Vec<Vec<ProofTrieNode>>,
+    deferred_drops: DeferredDrops,
     /// Metrics for the sparse state trie.
     #[cfg(feature = "metrics")]
     metrics: crate::metrics::SparseStateTrieMetrics,
@@ -72,7 +72,7 @@ where
             storage: Default::default(),
             retain_updates: false,
             account_rlp_buf: Vec::with_capacity(TRIE_ACCOUNT_RLP_MAX_SIZE),
-            proof_nodes_bufs: Vec::new(),
+            deferred_drops: DeferredDrops::default(),
             #[cfg(feature = "metrics")]
             metrics: Default::default(),
         }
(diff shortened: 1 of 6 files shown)


crates/trie/sparse/src/state.rs
```diff
@@ -81,6 +92,8 @@ pub struct SparseStateTrie<
     retain_updates: bool,
     /// Reusable buffer for RLP encoding of trie accounts.
     account_rlp_buf: Vec<u8>,
+    /// Reusable buffer for proof nodes, to avoid allocations across payload runs.
+    proof_nodes_buf: Vec<ProofTrieNode>,
```
mediocregopher: If we tracked a Vec<ProofTrieNodes> for each storage trie we could then reuse the vecs across cleared/reused SparseStateTries, rather than having to deallocate them. Even if we want to keep deallocating them we should at least also defer the dealloc of the storage node vecs, right now they are still dealloc'd synchronously.

Also right now `proof_nodes_buf` is never being written to 🙈

DaniPopes (approved):

crates/trie/sparse/src/trie.rs
```diff
@@ -623,10 +624,14 @@ impl SparseTrieTrait for SerialSparseTrie {
...
+    fn reveal_nodes(&mut self, nodes: &mut [ProofTrieNode]) -> SparseTrieResult<()> {
         nodes.sort_unstable_by_key(|node| node.path);
-        for node in nodes {
-            self.reveal_node(node.path, node.node, node.masks)?;
+        for node in nodes.iter_mut() {
+            self.reveal_node(
+                node.path,
+                core::mem::replace(&mut node.node, TrieNode::EmptyRoot),
+                node.masks.take(),
+            )?;
```
mattsse, line 629: this is fine because for the serial impl, we dont need this

mattsse (approved):

mediocregopher (approved):

PR #21649, opened 2026-01-31
fix(prune): use batched pruning loop with edge feature to prevent OOM

gakonst:
## Summary

Alternative approach to #21618 based on Dan's review feedback. Instead of internal batching within the pruner segments, this implements the batching at the CLI level by running the pruner in a loop with a bounded limit.

## Motivation

When running `reth prune` with the edge flag (static files), the pruner could OOM because the `FxHashMap` tracking highest deleted entries grew unboundedly with `delete_limit(usize::MAX)`.

## Solution

For edge builds only:
- Use a bounded `DELETE_LIMIT` of 200,000 entries per batch
- Run the pruner in a loop until `progress.is_finished()` returns true
- Log progress after each batch

Non-edge builds continue with the existing unbounded single-run behavior.

## Changes

- `prune.rs`: Conditional compilation for edge vs non-edge pruning strategy
- `Cargo.toml`: Add `reth-prune-types` dependency for edge feature

## Testing

Existing tests pass. The batched execution is transparent to callers - the command runs to completion just like before, only the internal execution strategy differs.

---

Fixes RETH-291

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2927,7 +2927,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.114",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


yongkangc (approved):

joshieDo (approved):
lets just make the batching the default option, no need for edge feature imo

PR #21652, opened 2026-01-31
Fix(#21482)  Invalid Prometheus metric for batch size

ranjanmukesh:
Fixes invalid Prometheus exposition for the persistence batch size metric that caused scrape failures.
The metric  used a reserved _count suffix.
This PR:
	•	Renames the metric to avoid _count collisions


mattsse (commented):
this isnt smth we can review

PR #21655, opened 2026-01-31
trie(sparse): avoid cloning update map in fuzz test

0xcharry:
Iterate over &update instead of update.clone() in sparse_trie_fuzz. Clone only the key and account needed for update_leaf.


mattsse (commented):
not needed

PR #21659, opened 2026-01-31
perf(trie): preallocate capacity for removed_accounts vector

Noisyxl:
Preallocates `removed_accounts` vector to `state.accounts.len() + results.len()` to avoid reallocations during sparse trie updates. Upper bound is known upfront, so this reduces allocations in the hot path with zero risk.


mattsse (commented):
we dont know how mayn

PR #21662, opened 2026-02-01
chore(trie): remove redundant Vec::<_>:: turbofish

dennsikl:
Remove unnecessary turbofish syntax `Vec::<_>::` in `ProofCalculator::new()`. The type parameter is already inferred from the struct field annotations, making the explicit `<_>` placeholder redundant.


mattsse (commented):
Thanks for the contribution, but we can't accept purely cosmetic changes like removing turbofish syntax. These provide no functional benefit and add review overhead. We're looking for contributions that fix real bugs, add features, or improve production code performance.

PR #21664, opened 2026-02-01
fix(storage): use correct dictionary in create_tx_compressor

Aleksandr1732:
Fixes incorrect dictionary usage in `create_tx_compressor()` function.


mattsse (commented):
unsure we can safely do this

PR #21666, opened 2026-02-01
perf(provider): remove redundant clone in static file test iteration

splinter012:
Removes unnecessary Vec::clone() when iterating over headers in the test_static_files test. The vector is only read during iteration, so borrowing is sufficient and avoids an unnecessary heap allocation of the entire Vec<SealedHeader>.


mattsse (commented):
Thanks for the contribution, but we can't accept PRs that only remove clones in test code - these provide no meaningful value to the project. We're looking for contributions that fix real bugs, add features, or improve production code performance.

PR #21667, opened 2026-02-01
perf(trie): remove unnecessary clones in test_account_trie_order

marukai67:
Removes redundant allocations in `test_account_trie_order` test function:


mattsse (commented):
Thanks for the contribution, but we can't accept PRs that only remove clones in test code - these provide no meaningful value to the project. We're looking for contributions that fix real bugs, add features, or improve production code performance.

PR #21670, opened 2026-02-01
refactor(provider): deduplicate segment-to-stage mapping in static file manager

letmehateu:
Replace two identical match expressions in `ensure_unwind_consistency` and `stage_checkpoint_matches_static_file`

### DIFF
--- a/crates/static-file/types/src/segment.rs
+++ b/crates/static-file/types/src/segment.rs
@@ -199,7 +199,7 @@ impl StaticFileSegment {
         self.is_block_based() || self.is_change_based()
     }
 
-    /// Maps this segment to the [`StageId`] responsible for it.                              
+    /// Maps this segment to the [`reth_stages_types::StageId`] responsible for it.                              
     pub const fn to_stage_id(&self) -> reth_stages_types::StageId {
         use reth_stages_types::StageId;
         match self {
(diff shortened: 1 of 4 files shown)


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -3032,6 +3018,18 @@ where
     Ok((keccak256(rlp_buf), tx_id))
 }
 
+/// Maps a [`StaticFileSegment`] to the [`StageId`] responsible for it.
+const fn segment_stage_id(segment: StaticFileSegment) -> StageId {
```
joshieDo: could instead be a method of StaticFileSegment, to_stage_id()

crates/static-file/types/src/segment.rs
```diff
@@ -198,6 +198,19 @@ impl StaticFileSegment {
     pub const fn is_block_or_change_based(&self) -> bool {
         self.is_block_based() || self.is_change_based()
     }
+
+    /// Maps this segment to the [`reth_stages_types::StageId`] responsible for it.
+    pub const fn to_stage_id(&self) -> reth_stages_types::StageId {
+        use reth_stages_types::StageId;
```
joshieDo: nit: can you move the import top level, and use it in the method sig

letmehateu: > nit: can you move the import top level, and use it in the method sig

Done, thank you

joshieDo (approved):
ty

PR #21671, opened 2026-02-01
fix(cli): correct `--trusted-peer` to `--trusted-peers` in error message

HuberyJulianay:
Hi when I run the error message in p2p command referenced `--trusted-peer` but dosent work, And the actual CLI flag is `--trusted-peers`. This caused confusion......


mattsse (commented):

PR #21672, opened 2026-02-01
fix(rpc): return error when all forks are in the future in eth_config

HuberyJulianay:
Fixed an edge case in `eth_config` where `position(0).checked_sub(1)` falls back to the last fork index instead of returning an error when all timestamp forks are in the future.


mattsse (commented):

PR #21674, opened 2026-02-01
perf: add ParallelBridgeBuffered trait to replace par_bridge

DaniPopes:
Replace all par_bridge calls with a new utility extension method called
par_bridge_buffered. This collects the iterator into a Vec first, then
calls par_iter on it.

This is better in some cases because par_bridge wraps the iterator in a
mutex, which can create contention if either the iterator's next() is
fast or the tasks that consume the items are fast.


mattsse (approved):
makes sense

PR #21684, opened 2026-02-01
perf(trie): dispatch V2 storage proofs in lexicographical order

HuberyJulianay:
dispatch_v2_storage_proofs iterates over B256Map in arbitrary order, but storage proofs are consumed in lexicographical order during the account trie walk. This mismatch causes head-of-line blocking. This sorts storage_targets by hashed address before dispatching to align production order with consumption order.

References:
Dispatch: crates/trie/parallel/src/proof_task.rs#L1810
Consumption: crates/trie/parallel/src/proof_task.rs#L1379

Dont think this is low value? All refer to existing patten.

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -1828,8 +1828,14 @@ fn dispatch_v2_storage_proofs(
         }
     }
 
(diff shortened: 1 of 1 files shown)


mediocregopher (approved):

mediocregopher:
Forgot to do this for V2, good catch, thanks!

PR #21687, opened 2026-02-02
perf(trie): add dedicated thread spawning for proof workers

gakonst:
## Summary

Adds `ProofWorkerHandle::new_with_dedicated_threads()` which spawns workers using `std::thread::spawn` instead of Tokio's `spawn_blocking`, avoiding blocking pool scheduler overhead.

## Problem

Workers were spawned via `executor.spawn_blocking()` which goes through Tokio's blocking pool scheduler. This adds latency before workers can start processing proof tasks.

## Solution

- Add `new_with_dedicated_threads` method that uses `std::thread::Builder::spawn`
- Workers are spawned as named threads (`proof-storage-N`, `proof-account-N`) for debugging
- Update PayloadProcessor to use the new method
- Add `worker_pool` module with pool configuration types

## Limitation

This is a stepping stone toward a fully persistent worker pool. Currently, threads are still spawned per-payload because workers need a fresh database provider per payload (the provider is created at the start of `run()` and lives for the worker's lifetime).

A future optimization could implement session-based workers that:
1. Persist across payloads
2. Receive new database factory contexts per block
3. Create fresh providers per session while reusing threads

That would require restructuring to separate thread lifecycle from per-payload database context.

Request from @mattsse: _"we need to refactor how we spawn the proof workers, these should use a dedicated pool to avoid any spawn overhead before we start execution"_

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -38,6 +38,7 @@ use reth_trie::{hashed_cursor::HashedCursorFactory, trie_cursor::TrieCursorFacto
 use reth_trie_parallel::{
     proof_task::{ProofTaskCtx, ProofWorkerHandle},
     root::ParallelStateRootError,
+    worker_pool::ProofWorkerPool,
 };
 use reth_trie_sparse::{ClearedSparseStateTrie, RevealableSparseTrie, SparseStateTrie};
 use reth_trie_sparse_parallel::{ParallelSparseTrie, ParallelismThresholds};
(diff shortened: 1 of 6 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -417,14 +434,16 @@ where
...
+        (prewarm_rx, execute_rx, transaction_count_hint)
     }
 
     /// Spawn prewarming optionally wired to the multiproof task for target updates.
+    #[expect(clippy::too_many_arguments)]
     fn spawn_caching_with<P>(
         &self,
         env: ExecutionEnv<Evm>,
         mut transactions: mpsc::Receiver<impl ExecutableTxFor<Evm> + Clone + Send + 'static>,
+        transaction_count_hint: usize,
```
DaniPopes: this is wrong. don't revert of a previous pr

PR #21688, opened 2026-02-02
test(provider): add safe-no-sync crash recovery tests

gakonst:
## Summary

Adds tests verifying that reth can recover from a crash when using `--db.sync-mode safe-no-sync`.

## Background

In `safe-no-sync` mode, MDBX commits may be lost on crash because fsync is skipped. The recovery mechanism relies on:

1. **Static files being written and synced BEFORE MDBX commit** (see `provider.rs` L3532-3551)
2. **Changesets being stored in static files** (not MDBX) - `AccountChangeSets`, `StorageChangeSets`
3. **`check_consistency` detecting when static files are ahead of MDBX checkpoints**
4. **Pipeline unwind using changesets from static files, then re-execution**

This confirms that the architecture already supports safe crash recovery with `safe-no-sync` mode.

## Tests Added

| Test | Scenario | Validates |
|------|----------|-----------|
| `test_safe_no_sync_crash_recovery_static_files_ahead` | SF at block 5, MDBX rolled back to 3 | Recovery triggered correctly |
| `test_safe_no_sync_consistent_state_no_panic` | SF and MDBX consistent | No panic on consistent state |
| `test_safe_no_sync_mdbx_far_behind_static_files` | SF at block 2, MDBX at 0 | Handles severe data loss |

## Related Discussion

From #eng-perf: Making `--db.sync-mode safe-no-sync` safe to avoid MDBX fsync overhead in b2b case.

cc @mattsse

### DIFF
--- a/crates/storage/provider/src/providers/static_file/mod.rs
+++ b/crates/storage/provider/src/providers/static_file/mod.rs
@@ -7,6 +7,9 @@ pub use manager::{
 mod jar;
 pub use jar::StaticFileJarProvider;
 
+#[cfg(test)]
+mod safe_no_sync_test;
+
 mod writer;
 pub use writer::{StaticFileProviderRW, StaticFileProviderRWRefMut};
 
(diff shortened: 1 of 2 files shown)


gakonst (commented):
@mattsse this makes sense to me? if this is safe, is it worth making this db mode the default? and we'd expect to see impact on commit times when?

PR #21694, opened 2026-02-02
fix(prune): count unique addresses instead of changesets in RocksDB pruner limiter

gakonst:
## Summary

Fixes RETH-292: The RocksDB pruning path was incrementing the limiter for every changeset scanned from static files, not for each unique address that corresponds to actual RocksDB shard work.

## Problem

The pruner limiter counted **changesets scanned** instead of **RocksDB operations performed**:

```rust
// Before: increments for every changeset
for result in walker {
    // ...
    limiter.increment_deleted_entries_count();  // Called 7k times per block!
}
```

With ~7k changesets/block but fewer unique addresses, the limiter stopped prematurely.

## Solution

Use `HashMap::entry()` to only increment the limiter on **first occurrence** of each unique key:

```rust
// After: increments only for new unique addresses
match highest_deleted_accounts.entry(changeset.address) {
    Entry::Vacant(v) => {
        v.insert(block_number);
        limiter.increment_deleted_entries_count(); // Only for new addresses
    }
    Entry::Occupied(mut o) => {
        if block_number > *o.get() {
            o.insert(block_number);
        }
    }
}
```

## Test Results

| Scenario | Before (bug) | After (fix) |
|----------|--------------|-------------|
| 10k changesets, 10 unique addresses, limit=100 | Stops at <1 block | Processes all 10 blocks |

```
=== RETH-292 Fix Verified ===
Limit: 100
Changesets per block: 1000
Unique addresses (RocksDB shards): 10
Blocks processed: 9
Finished: true

FIX: Limiter now counts 10 unique addresses (shards), not 10000 changesets.
```

## Changes

- `account_history.rs`: Increment limiter only on new unique addresses
- `storage_history.rs`: Increment limiter only on new unique (address, slot) pairs
- Added regression test that verifies the fix

## All prune tests pass

```
test result: ok. 39 passed; 0 failed; 0 ignored
```


gakonst:
Closing - the underlying issue is mitigated by PR #19141 which sets `delete_limit: usize::MAX` by default. With unlimited limit, the limiter accounting difference has no practical effect.

The real concern (throughput for large backlogs) is tracked in RETH-296.

PR #21702, opened 2026-02-02
perf(trie): avoid clearing already-cached sparse trie

mattsse:
Add `ClearedSparseStateTrie::cached()` that wraps a trie without clearing it, for use when the trie is retrieved from cache and already in a clean state. This avoids redundant clearing operations on cached tries.

Resolves RETH-328

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -39,7 +39,7 @@ use reth_trie_parallel::{
     proof_task::{ProofTaskCtx, ProofWorkerHandle},
     root::ParallelStateRootError,
 };
-use reth_trie_sparse::{ClearedSparseStateTrie, RevealableSparseTrie, SparseStateTrie};
+use reth_trie_sparse::{RevealableSparseTrie, SparseStateTrie};
 use reth_trie_sparse_parallel::{ParallelSparseTrie, ParallelismThresholds};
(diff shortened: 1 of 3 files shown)


shekhirin (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -556,7 +556,7 @@ where
                     from_multi_proof,
                     proof_worker_handle,
                     trie_metrics.clone(),
-                    ClearedSparseStateTrie::from_state_trie(sparse_state_trie),
+                    ClearedSparseStateTrie::cached(sparse_state_trie),
```
mediocregopher: I feel like a better fix would be to not pass a ClearedSparseStateTrie into SparseTrieCacheTask? We could pass the SparseStateTrie directly

crates/trie/sparse/src/state.rs
```diff
@@ -23,48 +23,6 @@ use reth_trie_common::{
 use tracing::debug;
 use tracing::{instrument, trace};
 
-/// Provides type-safe re-use of cleared [`SparseStateTrie`]s, which helps to save allocations
-/// across payload runs.
-#[derive(Debug)]
-pub struct ClearedSparseStateTrie<
```
mediocregopher, line 29: even better!

mediocregopher (approved):

PR #21718, opened 2026-02-02
fix(rpc): validate toBlock in trace_filter

gancerlory:
## Summary
Add validation for `toBlock` parameter in `trace_filter` to return `HeaderNotFound` error when it exceeds the latest block.

## Problem
When `toBlock > latest_block`, `recovered_block_range` silently returns partial results instead of an error.

Example (latest_block = 1000):
- Request: `trace_filter({ fromBlock: 950, toBlock: 1050 })`  
- Before: Returns traces for 950-1000 only, **no error**
- After: Returns `HeaderNotFound(1050)` error


mattsse (approved):
makes sense

PR #21724, opened 2026-02-03
chore: simplify changeset range usage

BearSoltokien:
Removes unnecessary intermediate variables and simplifies error conversions in `storage_changeset` and `account_block_changeset` methods, following the same pattern established in #21457.


mattsse (commented):
can keep as is

PR #21741, opened 2026-02-03
fix(net): prevent unintended MRU promotion of tx hashes in pending fetch queue

niran:
## Summary

Fixes transaction starvation caused by LRU promotion when re-buffering hashes.

### Problem

When hashes are re-buffered (surplus from announcement or after failed request), calling `insert()` on `hashes_pending_fetch` promotes them to MRU position. This causes old, widely-propagated transactions to continuously jump to the front of the queue, starving newer transactions.

### Solution

Use `hashes_fetch_inflight_and_pending_fetch` for iteration ordering instead of `hashes_pending_fetch`. Unlike the pending cache, this LruMap is designed to remain stable during retries—its order only changes when new hashes are inserted or old ones are evicted.

To fully avoid promotion, we also switch from `get` to `peek_mut` when updating metadata in the LruMap. Since `hashes_pending_fetch` no longer needs ordering, it becomes a `HashSet` for simple pending/inflight membership tracking.

Key changes:
- Add `pending_hashes_iter()` that iterates the LruMap filtered by HashSet membership
- Use `peek_mut` instead of `get` when updating metadata
- Change `hashes_pending_fetch` from `LruCache` to `HashSet`

## Test Plan

- [x] `pending_hashes_iter_skips_inflight` - verifies iterator excludes inflight hashes
- [x] `buffer_hashes_does_not_promote_existing_to_mru` - verifies re-buffered hashes maintain position
- [x] Existing tests pass: `test_on_fetch_pending_hashes`, `test_max_retries_tx_request`

### DIFF
--- a/crates/net/network/src/transactions/fetcher.rs
+++ b/crates/net/network/src/transactions/fetcher.rs
@@ -1563,7 +1563,7 @@ mod test {
 
         // re-buffer hash1 (simulating retry after failed request)
         // this should NOT promote hash1 to MRU position
-        let retry_hashes = RequestTxHashes::new([hash1].into_iter().collect());
+        let retry_hashes = RequestTxHashes::new(std::iter::once(hash1).collect());
         tx_fetcher.buffer_hashes(retry_hashes, None);
 
         // hash1 should still be at the end (not promoted to MRU)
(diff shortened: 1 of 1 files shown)


mattsse (commented):
unclear where the fix is and what is refactor

so I dont know how to review this really

PR #21742, opened 2026-02-03
feat(txpool): add `is_transaction_ready` to `TransactionPool` trait

mininny:
Adds `is_transaction_ready(&self, sender: &Address, nonce: u64) -> bool` to the `TransactionPool` trait.

## Motivation
External builder implementation needs to check whether a specific transaction is ready for inclusion. The existing `best_transactions()` iterator only supports sequential consumption - there's no API to check if a specific transaction is ready to be included. This can be easily done by using the `independent_transactions` set that the txpool is managing internally. 

## Solution
- Add `is_transaction_ready` to `TransactionPool` trait with a default implementation that scans `pending_transactions()` for the sender's minimum nonce (preserve backwards compatibility of trait).
- Override in `Pool<V,T,S>` with an implementation that checks the pending pool's `independent_transactions` set directly

The `independent_transactions` set contains exactly the transactions whose nonce is next-expected for their sender, making this an `O(k)` lookup where k is the number of unique senders with pending txs.

### DIFF
--- a/crates/transaction-pool/src/traits.rs
+++ b/crates/transaction-pool/src/traits.rs
@@ -405,6 +405,26 @@ pub trait TransactionPool: Clone + Debug + Send + Sync {
     /// Consumer: RPC
     fn pending_transactions(&self) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>>;
 
+    /// Returns `true` if the transaction with the given sender and nonce is ready to be executed.
+    ///
+    /// A transaction is "nonce-ready" if it is the next expected transaction for its sender,
+    /// meaning all prior nonces have already been consumed or are not required.
+    ///
+    /// This is useful for checking whether a transaction can be directly included in a block
+    /// without violating nonce ordering constraints.
+    ///
+    /// The default implementation scans pending transactions to find the lowest nonce for the
+    /// sender and checks if it matches the given nonce.
+    fn is_transaction_ready(&self, sender: &Address, nonce: u64) -> bool {
+        use alloy_consensus::Transaction;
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/traits.rs
```diff
@@ -405,6 +405,26 @@ pub trait TransactionPool: Clone + Debug + Send + Sync {
...
+    ///
+    /// A transaction is "nonce-ready" if it is the next expected transaction for its sender,
+    /// meaning all prior nonces have already been consumed or are not required.
+    ///
+    /// This is useful for checking whether a transaction can be directly included in a block
+    /// without violating nonce ordering constraints.
+    ///
+    /// The default implementation scans pending transactions to find the lowest nonce for the
+    /// sender and checks if it matches the given nonce.
+    fn is_transaction_ready(&self, sender: &Address, nonce: u64) -> bool {
```
mattsse: I'd pending to change this slightly to

`fn get_pending_transaction_by_sender_and_nonce`
similar to:

https://github.com/paradigmxyz/reth/blob/6946f26d77be5b1defd93f14af5a8e02064f87ad/crates/transaction-pool/src/traits.rs#L561-L566

then this is more useful

mattsse (changes requested):

crates/transaction-pool/src/traits.rs
mininny: Thank you for the quick review! I've changed it to `get_pending_transaction_by_sender_and_nonce` like your suggestion 🙏

mattsse (approved):

PR #21753, opened 2026-02-03
feat(cli): add range query support for `db list` command

### LINKED ISSUE #21682: should we add `--start-key`/`--end-key` range query support to `reth db list`  ?
### Describe the feature

The `reth db list` command currently uses `skip/len` pagination which requires O(n) traversal. For large tables, querying middle ranges is very slow.                                                                                                                                
                                                                                                                                                                                                                                                                                       
Since `reth db get` already supports `--start-key`/`--end-key` range queries, would it be useful to add the same capability to `list`?                                                                                                                                               
                                                                                                                                                                                                                                                                                       
Inspired by: https://github.com/paradigmxyz/reth/blob/main/crates/storage/db-common/src/db_tool/mod.rs#L98-L100 

### Additional context

_No response_

HuberyJulianay:
closes https://github.com/paradigmxyz/reth/issues/21682

The `list` command currently only supports `skip/len` pagination, which requires scanning from the beginning of the table. For large tables, querying middle data (e.g., `--skip 1000000`) is extremely slow as it needs to traverse 1 million records.     

Add range query support similar to `db get` and `db checksum` commands:                                                                                                                                                                                                              
  - `--start-key`: Start key for range query (inclusive), enables O(log n) seek                                                                                                                                                                                                        
  - `--end-key`: End key for range query (exclusive)                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                       
Example Usage:                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                       
  ```bash                                                                                                                                                                                                                                                                              
  # Query Headers table for blocks 1000 to 2000                                                                                                                                                                                                                                        
  reth db list Headers --start-key 1000 --end-key 2000 --json                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                       
  # Query specific address range in HashedAccounts                                                                                                                                                                                                                                     
  reth db list HashedAccounts --start-key '"0x0000..."' --end-key '"0xffff..."' --json                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                       
  # Combine with other options                                                                                                                                                                                                                                                         
  reth db list Headers --start-key 1000000 --len 10 --json

### DIFF
--- a/crates/cli/commands/src/db/list.rs
+++ b/crates/cli/commands/src/db/list.rs
@@ -19,7 +19,8 @@ use tracing::error;
 pub struct Command {
     /// The table name
     table: Tables,
-    /// Skip first N entries
+    /// Skip first N entries. When used with --start-key, skips relative to the seek position
+    /// rather than the table start.
     #[arg(long, short, default_value_t = 0)]
     skip: usize,
     /// Reverse the order of the entries. If enabled last table entries are read.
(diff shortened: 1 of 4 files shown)


emmajam:
Thanks for working on this — the feature itself is useful. A few things to address:

1. `--reverse + --start-key/--end-key` is broken. `walk_back(start_key)` starts at `start_key` and walks backwards (toward lower keys), so `--start-key 1000 --end-key 2000 --reverse` would walk from 1000 toward 0 and never reach 2000. The `end_key check (&key >= end)` is also wrong in reverse — it would need to be <=. Either swap the roles of start/end when reversed, or disallow combining them for now with a clear error.
2. Should validate `start_key < end_key` — currently `--start-key 2000 --end-key 1000` silently returns nothing.
3. `skip + start_key` interaction should be documented — with `--start-key`, skip is relative to the seek position, not the table start. That's probably the right behaviour but worth calling out in the help text.
4. Minor: the `let end_key = filter.end_key.clone()` before the closure could just use `filter.end_key.as_ref()` inside to avoid the clone.

PR #21763, opened 2026-02-03
fix(rpc): use consistent sidecar check in fill_transaction for EIP-7594 support

gancerlory:
## Summary

- Fix `populate_blob_hashes()` not being called when EIP-7594 (v1) sidecar is provided
- Change condition from `blob_sidecar().is_some()` to `sidecar.is_some()`

## Problem

In `fill_transaction`, the condition for populating blob hashes used `blob_sidecar().is_some()`, which only returns `Some` for EIP-4844 (v0) sidecars. When users provide an EIP-7594 (v1) sidecar, this check returns `false`, causing `populate_blob_hashes()` to be skipped.

## Solution

Use `sidecar.is_some()` which checks the raw `Option<BlobTransactionSidecarVariant>` field directly, correctly handling both v0 and v1 formats.


mattsse (approved):
this is actually valid

PR #21764, opened 2026-02-03
perf: use separate pool for save_blocks

DaniPopes:
~~blocked on CI being weird~~

CI failure was due to rayon recursion in statetests. Even if in a different thread pool, rayon will still try to run jobs while .scope is waiting or whatever, so run test -> save blocks -> run a new test and so on until memmap exhaustion from open_db, or stack overflow.


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -694,70 +689,87 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
-                    |w| Self::write_transaction_senders(w, blocks, tx_nums),
-                )
+        let mut r_headers = None;
+        let mut r_txs = None;
+        let mut r_senders = None;
+        let mut r_receipts = None;
+        let mut r_account_changesets = None;
+        let mut r_storage_changesets = None;
+
+        STORAGE_POOL.in_place_scope(|s| {
```
mattsse, line 699: can we add some docs here that this is the only thing that accesses the pool so that we always have the threads available that we need?

DaniPopes: already in STORAGE_POOL docs

joshieDo (approved):

PR #21765, opened 2026-02-03
feat: add a public `prune_transactions` method to the TransactionPool

haardikk21:
For Flashblocks built beyond the first one, we need to reset the best transactions iterator over the TxPool on a per flashblock basis to reset the priority boundaries. However, since the TxPool has no idea of which transactions have already been committed in past flashblocks, the new iterator has all the same transactions again.

While we can trivially skip transactions that have already been committed, it does mess up the ordering of best transactions since descendant txns of previously committed txns can have different priority fees and should have been yielded earlier in the new flashblock building process but weren't because we were waiting on it's ancestor to first be yielded and skipped before it even moves the descendant to the independent set.

This PR adds a public `prune_transactions` method to the trait that will allow consumers of the Reth SDK to manually prune a transaction from the transaction pool even if a canonical state update has not been received, allowing us to cleanly mark previously committed txns as pruned and bump their descendants in the middle of the Flashblocks building process.

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -644,8 +644,8 @@ where
                     let (result, meta) = self.add_transaction(&mut pool, origin, tx);
 
                     // Only collect metadata for successful insertions
-                    if result.is_ok()
-                        && let Some(meta) = meta
+                    if result.is_ok() &&
+                        let Some(meta) = meta
                     {
                         added_metas.push(meta);
                     }
@@ -679,8 +679,8 @@ where
             // A newly added transaction may be immediately discarded, so we need to
             // adjust the result here
             for res in &mut results {
-                if let Ok(AddedTransactionOutcome { hash, .. }) = res
-                    && discarded_hashes.contains(hash)
+                if let Ok(AddedTransactionOutcome { hash, .. }) = res &&
+                    discarded_hashes.contains(hash)
                 {
                     *res = Err(PoolError::new(*hash, PoolErrorKind::DiscardedOnInsert))
                 }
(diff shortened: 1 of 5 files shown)


crates/transaction-pool/src/traits.rs
```diff
@@ -496,6 +496,26 @@ pub trait TransactionPool: Clone + Debug + Send + Sync {
         sender: Address,
     ) -> Vec<Arc<ValidPoolTransaction<Self::Transaction>>>;
 
+    /// Prunes a single transaction corresponding to the given hash.
+    ///
+    /// Returns the pruned transaction if it was found in the pool.
+    ///
+    /// Consumer: Utility
+    fn prune_transaction(
```
mattsse: can we clarify what the difference is compared to

https://github.com/paradigmxyz/reth/blob/c9cc118def7f1783b3cfa9d5611b3ca16d102b42/crates/transaction-pool/src/traits.rs#L459-L464

crates/transaction-pool/src/pool/mod.rs
```diff
@@ -774,7 +774,7 @@ where
                         needs_cleanup = true;
                     }
                     // Skip non-propagate transactions for propagate-only listeners
-                    continue
+                    continue;
```
mattsse: let's undo all of these please :)

mattsse (changes requested):
makes sense

because this can be a bit confusing, I'd like some extra docs what this exactly does and how this behaves compared to remove

crates/transaction-pool/src/pool/mod.rs
haardikk21: I ran the formatter and it did all of this lol, looked weird to me as well. I'll go and manually undo it if it still passes CI

mattsse (approved):

PR #21772, opened 2026-02-03
fix(test): clean up test temp directories on drop

### LINKED ISSUE #21769: test tmpdirs are not cleaned up
### Describe the feature

https://github.com/paradigmxyz/reth/blob/87bae740949ff1b91e774de99ba5c96e5258153d/crates/storage/provider/src/test_utils/mod.rs#L58

these are not cleaned up so running tests a couple times will fill up /tmp to the max

### Additional context

_No response_

HuberyJulianay:
Closes https://github.com/paradigmxyz/reth/issues/21769 cc @DaniPopes 

The test helper `create_test_provider_factory_with_node_types` was creating three separate temp directories and calling `.keep()` on each, which tells tempfile not to delete them. This meant running tests a few times would leave hundreds of directories in /tmp.

Changed it to use a single parent directory containing db/, static_files/, and rocksdb/ subdirs. Now TempDatabase cleans up the whole thing when it's dropped.

Also fixed a test that was accidentally dropping the factory too early - it was doing `factory.with_prune_modes(...).provider_rw()` which consumes the factory, so TempDatabase got dropped while the provider was still in use.

Oh and thanks to Claude, help to located the key point quickly.

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/mod.rs
+++ b/crates/storage/db/src/implementation/mdbx/mod.rs
@@ -643,9 +643,11 @@ mod tests {
     use std::str::FromStr;
     use tempfile::TempDir;
 
-    /// Create database for testing
-    fn create_test_db(kind: DatabaseEnvKind) -> DatabaseEnv {
-        create_test_db_with_path(kind, &tempfile::TempDir::new().expect(ERROR_TEMPDIR).keep())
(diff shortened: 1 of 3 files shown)


DaniPopes:
Thanks this should work. Can you also look at other ".keep" calls? I see there's a few more similar to this one

DaniPopes (approved):

mattsse (approved):

PR #21773, opened 2026-02-03
chore: don't format fields in tracy spans

DaniPopes:


mattsse (commented):
what does this do?

DaniPopes:
format_fields_in_zone_name true -> false
as title

mediocregopher (approved):

PR #21777, opened 2026-02-03
chore: name tokio rt threads

DaniPopes:


crates/engine/tree/src/tree/payload_processor/executor.rs
```diff
@@ -45,29 +42,6 @@ struct WorkloadExecutorInner {
 
 impl WorkloadExecutorInner {
     fn new() -> Self {
-        fn get_runtime_handle() -> Handle {
-            Handle::try_current().unwrap_or_else(|_| {
-                // Create a new runtime if no runtime is available
```
mattsse, line 50: I'd prefer having this here

crates/trie/parallel/src/root.rs
```diff
@@ -98,7 +98,7 @@ where
         let mut storage_roots = HashMap::with_capacity(storage_root_targets.len());
 
         // Get runtime handle once outside the loop
-        let handle = get_runtime_handle();
+        let handle = get_tokio_runtime_handle();
```
mattsse, line 101: unsure this even even used with v2

crates/cli/runner/src/lib.rs
```diff
@@ -219,7 +219,14 @@ impl CliRunnerConfig {
 /// Creates a new default tokio multi-thread [Runtime](tokio::runtime::Runtime) with all features
 /// enabled
 pub fn tokio_runtime() -> Result<tokio::runtime::Runtime, std::io::Error> {
-    tokio::runtime::Builder::new_multi_thread().enable_all().build()
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        // Keep the threads alive for at least the block time (12 seconds) plus buffer.
+        // This prevents the costly process of spawning new threads on every
+        // new block, and instead reuses the existing threads.
```
mattsse, line 226: unsure about this one, this wasnt the case before, this is used for all other things, that dont have a real tick rate, so this is perhaps a bit misleading

mattsse (changes requested):

crates/engine/tree/src/tree/payload_processor/executor.rs
DaniPopes: why? theres not much benefit to having multiple tokio runtimes

crates/cli/runner/src/lib.rs
mattsse: actually, this seems fine

crates/trie/parallel/src/root.rs
```diff
@@ -270,17 +270,19 @@ impl From<StateProofError> for ParallelStateRootError {
...
         static RT: OnceLock<Runtime> = OnceLock::new();
 
         let rt = RT.get_or_init(|| {
             Builder::new_multi_thread()
+                .enable_all()
                 // Keep the threads alive for at least the block time (12 seconds) plus buffer.
                 // This prevents the costly process of spawning new threads on every
                 // new block, and instead reuses the existing threads.
                 .thread_keep_alive(Duration::from_secs(15))
+                .thread_name("trie-tokio-rt")
```
mattsse, line 285: this will now also have all the prewarm or any of the tasks we spawn in payload processing so this is a bit misleading, we could name it -payload, but then this doesnt really fit in here

mattsse (approved):

PR #21780, opened 2026-02-04
perf: spawn proof workers in a separate thread

DaniPopes:
Same as https://github.com/paradigmxyz/reth/pull/20575 but for storage and account workers.

Can be up to 5% of a single block validation time.

Review with "hide whitespace".

For example, this is very significant for the first few ever blocks processed due to it spawning the actual OS threads: image

Otherwise, it can be a couple % **before** execution:
image


Rjected (approved):
lgtm

PR #21786, opened 2026-02-04
perf(rocksdb): use single iterator for batch history unwind

yongkangc:
## Summary

Applies the same optimization from #21767 to the `unwind_*` methods.

Previously, `unwind_account_history_indices` and `unwind_storage_history_indices` created a new `iterator_cf` for every address/storage key by calling `account_history_shards`/`storage_history_shards` in a loop.

The batch methods reuse a single raw iterator and skip seeks when the iterator is already positioned correctly (for sorted targets in key order). This reduces RocksDB seek overhead for large unwind operations.

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -2694,13 +2694,19 @@ impl Iterator for RocksDBIterEnum<'_> {
     }
 }
 
-/// Raw iterator wrapper supporting seek operations across DB modes.
+/// Wrapper enum for raw `RocksDB` iterators that works in both read-write and read-only modes.
+///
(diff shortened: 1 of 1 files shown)


yongkangc (commented):
Add `iter.status()` check after the iteration loops to catch RocksDB errors during iteration.

In both `unwind_account_history_batch` and `unwind_storage_history_batch`, add after the `while iter.valid()` loop:

```rust
iter.status().map_err(|e| {
    ProviderError::Database(DatabaseError::Read(DatabaseErrorInfo {
        message: e.to_string().into(),
        code: -1,
    }))
})?;
```

Currently `iter.status()` is only checked after `seek()`, but errors during iteration (e.g., I/O errors while calling `next()`) may go undetected.

PR #21795, opened 2026-02-04
feat(eth-wire): introduce ProtocolMessage::decode_status for handshake

gakonst:
## Summary

During the eth handshake, only Status messages are valid responses. Previously, we used `decode_message` which would decode any message type, then match on the result to check if it was a Status message.

This PR introduces a dedicated `decode_status` method on `ProtocolMessage` that:

- Returns `StatusMessage` directly instead of `ProtocolMessage`
- Returns `MessageError::UnexpectedStatusMessage` if a non-Status message is received
- Simplifies the handshake code by eliminating the match arm for non-Status messages

The handshake now uses `ProtocolBreach` disconnect reason on decode errors since receiving a non-Status message during handshake is a protocol violation.

## Changes

- Added `MessageError::UnexpectedStatusMessage` variant for when a non-Status message is received during decode_status
- Added `ProtocolMessage::decode_status()` method that specifically decodes only Status messages
- Updated `EthereumEthHandshake::eth_handshake()` to use the new `decode_status` method
- Changed disconnect reason from `DisconnectRequested` to `ProtocolBreach` for decode errors (since non-Status is a protocol violation)
- Added tests for both successful decode and rejection of non-Status messages

## References

https://github.com/paradigmxyz/reth/blob/main/crates/net/eth-wire/src/handshake.rs#L122-L125
https://github.com/paradigmxyz/reth/blob/main/crates/net/eth-wire/src/handshake.rs#L224-L232

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,9 +106,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.24"
+version = "0.2.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b163ff4acf0eac29af05a911397cc418a76e153467b859398adc26cb9335a611"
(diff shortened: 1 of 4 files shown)


gakonst:
Recreating with clean fork

PR #21801, opened 2026-02-04
fix(prune): handle RocksDB TransactionHashNumbers pruning when MDBX has no data

yongkangc:
## Summary

Fixes RocksDB `TransactionHashNumbers` entries never being pruned when MDBX has no transaction data.

## Problem

When `get_next_tx_num_range()` returns `None` (MDBX `BlockBodyIndices` empty/pruned), the function returns early **before** reaching the RocksDB pruning check.

## Solution

Move the RocksDB check inside the `None` branch. For `PruneMode::Full`, add `prune_rocksdb_full()` that clears the table using static files for checkpoint data.

### DIFF
--- a/crates/prune/prune/src/segments/user/transaction_lookup.rs
+++ b/crates/prune/prune/src/segments/user/transaction_lookup.rs
@@ -224,18 +224,26 @@ impl TransactionLookup {
         rocksdb.clear::<tables::TransactionHashNumbers>()?;
         trace!(target: "pruner", "Cleared transaction lookup table (RocksDB, no MDBX range)");
 
-        // Get the last transaction number from static files for the checkpoint
+        // Get checkpoint from static files. Use the actual highest block/tx from static files
+        // to ensure consistency, clamped to to_block (we shouldn't report pruning beyond to_block).
         let static_file_provider = provider.static_file_provider();
-        let highest_tx = static_file_provider
-            .get_highest_static_file_tx(StaticFileSegment::Transactions)
-            .unwrap_or(0);
+        let highest_static_block =
+            static_file_provider.get_highest_static_file_block(StaticFileSegment::Transactions);
+        let highest_tx =
+            static_file_provider.get_highest_static_file_tx(StaticFileSegment::Transactions);
+
+        // Use the minimum of to_block and highest static block for consistency
+        let checkpoint_block = match highest_static_block {
+            Some(static_block) => Some(to_block.min(static_block)),
+            None => Some(to_block),
+        };
 
         Ok(SegmentOutput {
             progress: PruneProgress::Finished,
             pruned: 0, // RocksDB clear doesn't return count
             checkpoint: Some(SegmentOutputCheckpoint {
-                block_number: Some(to_block),
-                tx_number: Some(highest_tx),
+                block_number: checkpoint_block,
+                tx_number: highest_tx,
             }),
         })
     }
@@ -659,10 +667,24 @@ mod tests {
         let result = segment.prune(&provider, input).unwrap();
         provider.commit().expect("commit");
 
-        // Verify pruning completed successfully
+        // Verify pruning completed successfully with correct checkpoint
         assert_matches!(
             result,
             SegmentOutput { progress: PruneProgress::Finished, checkpoint: Some(_), .. }
+        );
+
+        // Verify checkpoint values are consistent with static files
+        let checkpoint = result.checkpoint.unwrap();
+        let highest_tx = tx_hash_numbers.len() as u64 - 1; // 0-indexed
+        assert_eq!(
+            checkpoint.block_number,
+            Some(to_block),
+            "Checkpoint block should match to_block (which equals highest static block)"
+        );
+        assert_eq!(
+            checkpoint.tx_number,
+            Some(highest_tx),
+            "Checkpoint tx should match highest static file tx"
         );
 
         // Verify RocksDB is now empty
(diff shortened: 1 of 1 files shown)


crates/prune/prune/src/segments/user/transaction_lookup.rs
```diff
@@ -80,18 +80,30 @@ where
...
+        let use_rocksdb = provider.cached_storage_settings().transaction_hash_numbers_in_rocksdb;
+        #[cfg(not(all(unix, feature = "rocksdb")))]
+        let _use_rocksdb = false;
+
         let (start, end) = match input.get_next_tx_num_range(provider)? {
-            Some(range) => range,
+            Some(range) => range.into_inner(),
             None => {
+                // When using RocksDB storage with PruneMode::Full, we can still clear the table
+                // even if MDBX has no transaction range data (e.g., MDBX was already pruned).
```
joshieDo: this is kinda incorrect, i dont think we ever prune the table (BlockBodyIndices) used on the `get_next_tx_num_range`

crates/prune/prune/src/segments/user/transaction_lookup.rs
```diff
@@ -80,18 +80,30 @@ where
...
+                // We get the end transaction number from static files instead.
+                #[cfg(all(unix, feature = "rocksdb"))]
+                if use_rocksdb && self.mode.is_full() {
+                    return self.prune_rocksdb_full(provider, input.to_block);
+                }
```
joshieDo, line 99: i like the idea (and we could do it for mdbx too), but i don't it can exist here. If the table is big for whatever reason, then this might take a lot of time? and slow down considerable the persistence task. 

but for reth prune cmd, i like it

Rjected: we already do this https://github.com/paradigmxyz/reth/blob/1340d732ef9ced37ffc29de09a10365af78428f5/crates/prune/prune/src/segments/user/transaction_lookup.rs#L211

yongkangc:
Addressed the PR comments:

**@joshieDo (line 93):** You're right - BlockBodyIndices is never pruned. Updated the comment to clarify that `get_next_tx_num_range()` returns `None` when MDBX has no `BlockBodyIndices` for `to_block` (e.g., data only exists in static files, or indices not yet populated), not because "MDBX was already pruned".

**@Rjected (line 98):** Yes, MDBX already has this `PruneMode::Full` fast-path that clears the table. The issue is that the RocksDB path never reaches the existing MDBX table-clear code when `get_next_tx_num_range()` returns `None` - it early-returns before we can even check the storage backend. This PR adds the equivalent behavior for RocksDB in that edge case.

**@joshieDo (line 98 - performance concern):** I upgraded the log from `trace` to `debug` to make it more visible. The `rocksdb.clear()` is a column-family drop/recreate (fast operation), not key-by-key deletion. It only triggers when all of:
- `use_rocksdb == true`
- `PruneMode::Full` (explicit operator choice)
- MDBX indices unavailable for `to_block`

The checkpoint ensures this only runs once, not on every persistence tick. This is the same semantic as MDBX Full mode - just clearing the table.

yongkangc:
Addressed the comments:

1. **@joshieDo (line 93)**: Fixed the comment - you're right that `BlockBodyIndices` is never pruned. Updated to: "MDBX has no `BlockBodyIndices` for `to_block` (e.g., data only exists in static files, or indices not yet populated)"

2. **@joshieDo (line 98)**: Added a comment explaining that `rocksdb.clear()` uses `delete_range_cf` which is O(1) to issue - actual deletion happens asynchronously during compaction. This mirrors MDBX's `clear_table` behavior, so performance should be comparable.

3. **@Rjected**: Yes, the MDBX path at line ~111 already does `clear_table`. This PR adds the RocksDB equivalent for the case where `get_next_tx_num_range()` returns `None` (BlockBodyIndices missing) but we still need to clear the RocksDB table.

crates/prune/prune/src/segments/user/transaction_lookup.rs
yongkangc: @Rjected i think the existing `prune_rocksdb` full-clear path https://github.com/paradigmxyz/reth/blob/1340d732ef9ced37ffc29de09a10365af78428f5/crates/prune/prune/src/segments/user/transaction_lookup.rs#L257-L274 requires `end` from `get_next_tx_num_range()` to build the checkpoint. 

But right, if MDBX block indices are missing ->  we never reach `prune_rocksdb`. So `prune_rocksdb_full` is the fallback for this case. it derives the checkpoint from static files instead of MDBX, since that's all we have available.

crates/prune/prune/src/segments/user/transaction_lookup.rs
yongkangc: Fixed the comment - BlockBodyIndices is never pruned. Updated to explain this happens when data only exists in static files or indices aren't populated yet.

crates/prune/prune/src/segments/user/transaction_lookup.rs
yongkangc: @joshieDo `prune_rocksdb_full()` calls `rocksdb.clear()` which uses `delete_range_cf` under the hood - writes a range tombstone to WAL + memtable (no scanning), actual deletion happens during compaction. Table size doesn't affect blocking time. 

Same as MDBX `clear_table`  which we already use for `PruneMode::Full`.

joshieDo: >But right, if MDBX block indices are missing -> we never reach prune_rocksdb. So prune_rocksdb_full is the fallback for this case.

this should never happen in a correct node , because we dont prune block body indices

PR #21803, opened 2026-02-04
perf: switch to unstable sort

cuiweixie:


mattsse (approved):
fine, doesnt really matter here

PR #21805, opened 2026-02-04
perf: prealloc vector

cuiweixie:


mattsse (commented):
we can keep this as is, we only expect this to be used with 1-2 blocks and I'd like avoid ..u64::Max misuse

PR #21806, opened 2026-02-04
perf(dns): avoid cloning entire HashSet in bootstrap method

Aleksandr1732:
Optimizes memory allocation in `DnsDiscoveryService::bootstrap` by replacing collection cloning with iterator-based element cloning.


mattsse (approved):

mattsse:
doesnt work

PR #17959, opened 2025-08-20
perf(trie): Use ParallelSparseTrie (if enabled) for storage tries

mediocregopher:

PR #21808, opened 2026-02-04
feat(trie): remove `SerialSparseTrie`

DaniPopes:
## Summary

Remove `SerialSparseTrie` and make `ParallelSparseTrie` the only sparse trie implementation.

## Background

`ParallelSparseTrie` has been the default and only used implementation in production since:
- **Aug 2025**: Enabled by default for account tries (#17956)
- **Sep 2025**: Extended to storage tries (#17959)

The `--disable-parallel-sparse-trie` flag was kept for fallback, but `SerialSparseTrie` has been effectively dead code.

## Changes

- Remove `SerialSparseTrie` from `reth-trie-sparse`
- Move `ParallelSparseTrie` from `reth-trie-sparse-parallel` into `reth-trie-sparse`
- Delete `reth-trie-sparse-parallel` crate
- Update `SparseStateTrie` default type parameter to `ParallelSparseTrie`
- Remove sparse trie benchmarks (were comparing serial vs parallel)
- Update tests to use `find_leaf()` instead of `nodes_ref()` (internal API not exposed by `ParallelSparseTrie`)

## Simplifications

- One less crate to maintain
- No more feature flag complexity for parallel trie selection
- Clearer code path in engine tree


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -211,7 +211,7 @@ where
 const MAX_PENDING_UPDATES: usize = 100;
 
 /// Sparse trie task implementation that uses in-memory sparse trie data to schedule proof fetching.
-pub(super) struct SparseTrieCacheTask<A = SerialSparseTrie, S = SerialSparseTrie> {
+pub(super) struct SparseTrieCacheTask<A = ParallelSparseTrie, S = ParallelSparseTrie> {
```
mattsse, line 215: maybe we dont actually need these anymore, but can keep for now

mattsse (approved):
rip bozo

DaniPopes:
will do rename and cleanup in a followup so that this one is actually reviewable

PR #21811, opened 2026-02-04
perf(trie): Don't filter proofs in v2 if sparse trie as cache is enabled

mediocregopher:
The sparse trie will already ignore already-revealed nodes, so there's no need for an extra filtering step.

Closes RETH-326


yongkangc (approved):
👍🏻 nice

mattsse (approved):
makes sense, looking forward to deleting the legacy stuff 🫡

PR #21821, opened 2026-02-04
fix: EIP-7778 gas accounting in receipts and block validation

qu0b:
## Summary

- **Receipt gas fix**: `RethReceiptBuilder` now uses `gas_spent` (after-refund cumulative gas) for `cumulative_gas_used` when EIP-7778 is active, matching the alloy-evm fix in alloy-rs/evm#279
- **Block validation fix**: `validate_block_post_execution` now accepts `execution_gas_used` from `BlockExecutionResult` and uses it for the header gas check when Amsterdam is active — post-EIP-7778, `header.gas_used` (before-refund) diverges from `receipt.cumulative_gas_used` (after-refund) by design

Without these fixes, reth rejects valid post-Amsterdam blocks due to gas_used mismatches at the fork boundary.

## Dependencies

- Requires alloy-rs/evm#279 for the matching `AlloyReceiptBuilder` fix

## Test plan

- [x] Tested locally with Kurtosis (geth+lighthouse supernode, reth+lodestar under test)
- [x] Pre-fork blocks sync correctly between geth and reth
- [ ] Verify post-fork blocks accepted by both clients
- [ ] Verify receipt roots match across clients

### DIFF
--- a/crates/ethereum/consensus/src/validation.rs
+++ b/crates/ethereum/consensus/src/validation.rs
@@ -15,6 +15,11 @@ use reth_primitives_traits::{
 ///
 /// If `receipt_root_bloom` is provided, the pre-computed receipt root and logs bloom are used
 /// instead of computing them from the receipts.
+///
+/// `gas_spent` is the gas_used value from the block execution result. When EIP-7778
+/// (Amsterdam) is active, block header gas_used tracks gas before refunds while receipt
+/// cumulative_gas_used tracks gas after refunds. In that case, the header must be validated
+/// against the execution result's gas_used rather than the receipt value.
 pub fn validate_block_post_execution<B, R, ChainSpec>(
     block: &RecoveredBlock<B>,
     chain_spec: &ChainSpec,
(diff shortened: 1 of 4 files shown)


crates/ethereum/consensus/src/validation.rs
```diff
@@ -15,6 +15,11 @@ use reth_primitives_traits::{
 ///
 /// If `receipt_root_bloom` is provided, the pre-computed receipt root and logs bloom are used
 /// instead of computing them from the receipts.
+///
+/// `gas_spent` is the gas_used value from the block execution result. When EIP-7778
+/// (Amsterdam) is active, block header gas_used tracks gas before refunds while receipt
+/// cumulative_gas_used tracks gas after refunds. In that case, the header must be validated
+/// against the execution result's gas_used rather than the receipt value.
```
mattsse, line 22: ty, this will be useful in the future

mattsse (approved):

PR #21836, opened 2026-02-05
perf: hash state updates in parallel

klkvr:
right now `EvmState` -> `HashedPostState` conversion is on the main path of state root task. this PR moves hashing to a separate task allowing main loop to do any potentially more useful work instead of hashing

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -516,7 +516,7 @@ where
     #[instrument(
         level = "debug",
         target = "engine::tree::payload_processor::sparse_trie",
-        skip_all,
+        skip_all
     )]
     fn on_hashed_state_update(&mut self, hashed_state_update: HashedPostState) {
         for (address, storage) in hashed_state_update.storages {
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -211,16 +211,6 @@ where
 /// Maximum number of pending/prewarm updates that we accumulate in memory before actually applying.
 const MAX_PENDING_UPDATES: usize = 100;
 
-/// Message type for the sparse trie task.
-enum SparseTrieTaskMessage {
-    /// A hashed state update ready to be processed.
-    HashedState(HashedPostState),
-    /// Prefetch proof targets (passed through directly).
-    PrefetchProofs(VersionedMultiProofTargets),
-    /// Signals that all state updates have been received.
-    FinishedStateUpdates,
-}
-
 /// Sparse trie task implementation that uses in-memory sparse trie data to schedule proof fetching.
 pub(super) struct SparseTrieCacheTask<A = SerialSparseTrie, S = SerialSparseTrie> {
     /// Sender for proof results.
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -211,14 +211,24 @@ where
...
 
+/// Message type for the sparse trie task.
+enum SparseTrieTaskMessage {
+    /// A hashed state update ready to be processed.
+    HashedState(HashedPostState),
+    /// Prefetch proof targets (passed through directly).
+    PrefetchProofs(VersionedMultiProofTargets),
+    /// Signals that all state updates have been received.
+    FinishedStateUpdates,
+}
```
mattsse: can we move this below the SparseTrieCacheTask impl block :)

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -211,14 +211,24 @@ where
 /// Maximum number of pending/prewarm updates that we accumulate in memory before actually applying.
 const MAX_PENDING_UPDATES: usize = 100;
 
+/// Message type for the sparse trie task.
+enum SparseTrieTaskMessage {
+    /// A hashed state update ready to be processed.
+    HashedState(HashedPostState),
```
mattsse: this makes sense

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -282,25 +287,28 @@ where
...
         updates: CrossbeamReceiver<MultiProofMessage>,
         proof_worker_handle: ProofWorkerHandle,
         metrics: MultiProofTaskMetrics,
         trie: SparseStateTrie<A, S>,
         chunk_size: Option<usize>,
     ) -> Self {
         let (proof_result_tx, proof_result_rx) = crossbeam_channel::unbounded();
+        let (hashed_state_tx, hashed_state_rx) = crossbeam_channel::unbounded();
+
+        executor.spawn_blocking(move || Self::run_hashing_task(updates, hashed_state_tx));
```
mattsse: ah I see, this simply forwards the updates processing

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -312,6 +320,34 @@ where
...
+    /// Runs the hashing task that drains updates from the channel and converts them to
+    /// `HashedPostState` in parallel.
+    fn run_hashing_task(
+        updates: CrossbeamReceiver<MultiProofMessage>,
+        hashed_state_tx: CrossbeamSender<SparseTrieTaskMessage>,
+    ) {
+        while let Ok(message) = updates.recv() {
+            let msg = match message {
+                MultiProofMessage::PrefetchProofs(targets) => {
+                    SparseTrieTaskMessage::PrefetchProofs(targets)
```
mattsse, line 333: these technically do one redundant hop but I think this should be fine especially since we receive all of them first and we're busy on the stac anyways

mattsse (approved):
smol pedantic nit

yongkangc (approved):

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -312,6 +317,34 @@ where
         }
     }
 
+    /// Runs the hashing task that drains updates from the channel and converts them to
+    /// `HashedPostState` in parallel.
+    fn run_hashing_task(
+        updates: CrossbeamReceiver<MultiProofMessage>,
+        hashed_state_tx: CrossbeamSender<SparseTrieTaskMessage>,
+    ) {
+        while let Ok(message) = updates.recv() {
```
DaniPopes, line 330: can we instrument/span this?

mediocregopher (approved):

PR #21837, opened 2026-02-05
perf(rocksdb): avoid to_plain_state_reverts() in history indexing loop

yongkangc:
## Summary

Eliminates expensive per-block allocations in RocksDB history indexing by iterating raw `Reverts` directly instead of calling `to_plain_state_reverts()`.

## Problem

In `write_account_history` and `write_storage_history`, `to_plain_state_reverts()` was called inside a per-block loop. This is expensive because it:

- Allocates new `Vec`s using `with_capacity()` for every call
- Clones `AccountInfo` for every account revert
- Collects storage reverts into new `Vec`s via `.collect()`
- Both methods called this independently, so conversion happened **twice per block**

History indexing only needs keys (addresses/slots), so the full data conversion was wasted work.

RETH-360

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/provider.rs
+++ b/crates/storage/provider/src/providers/rocksdb/provider.rs
@@ -27,6 +27,7 @@ use reth_storage_errors::{
     db::{DatabaseErrorInfo, DatabaseWriteError, DatabaseWriteOperation, LogLevel},
     provider::{ProviderError, ProviderResult},
 };
+use revm_database::states::reverts::AccountInfoRevert;
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamilyDescriptor, CompactionPri, DBCompressionType,
     DBRawIteratorWithThreadMode, IteratorMode, OptimisticTransactionDB,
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1283,6 +1288,7 @@ impl RocksDBProvider {
     /// Writes storage history indices for the given blocks.
     ///
     /// Derives history indices from reverts (same source as changesets) to ensure consistency.
+    /// Iterates raw reverts directly to avoid expensive `to_plain_state_reverts()` conversion.
```
yongkangc:
```suggestion
```

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1250,6 +1248,7 @@ impl RocksDBProvider {
     /// Writes account history indices for the given blocks.
     ///
     /// Derives history indices from reverts (same source as changesets) to ensure consistency.
+    /// Iterates raw reverts directly to avoid expensive `to_plain_state_reverts()` conversion.
```
yongkangc: remove this comment

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1294,18 +1299,18 @@ impl RocksDBProvider {
...
             let block_number = ctx.first_block_number + block_idx as u64;
-            let reverts = block.execution_outcome().state.reverts.to_plain_state_reverts();
+            let reverts = &block.execution_outcome().state.reverts;
 
-            // Iterate through storage reverts - these are exactly the slots that have
-            // changesets written, ensuring history indices match changeset entries.
-            for storage_block_reverts in reverts.storage {
-                for revert in storage_block_reverts {
-                    for (slot, _) in revert.storage_revert {
+            // Index all modified storage slots
```
yongkangc: can we keep this comment

```
            // Iterate through storage reverts - these are exactly the slots that have
            // changesets written, ensuring history indices match changeset entries.
```

crates/storage/provider/src/providers/rocksdb/provider.rs
yongkangc:
```suggestion
```

gakonst (approved):
Reviewed for correctness and performance — LGTM.

**Correctness:**
- Account history: skipping `DoNothing` matches exactly what `to_plain_state_reverts()` did (lines 60-68 of revm-database reverts.rs). Only `RevertTo` and `DeleteIt` produce account entries.
- Storage history: iterating `account_revert.storage.keys()` produces the same set of slots as the old path. The edge case where `wipe_storage=true` but `storage` is empty correctly indexes nothing in both old and new code (old code created a `PlainStorageRevert` but with empty `storage_revert` vec, so the inner loop was a no-op).
- Intra-block dedup (`v.last() != Some(block_number)`) is correct and actually better — history indices are sets of block numbers, and `append_*_history_shard` expects strictly increasing values.
- Slot encoding: `B256::new(slot.to_be_bytes())` is correct since revm's `StorageKey = U256`.

**Performance:** Eliminates per-block `Vec` allocations, `AccountInfo` clones, and `.collect()` calls that were immediately discarded. Clean win.

PR #21839, opened 2026-02-05
fix(db stat): skip over missing static files

Vui-Chee:
Before this, when running `db stat` over a snapshot with custom genesis number, the first static file may contain a header that begins from custom genesis number (does not begin with a multiple of X), resulting in a segment range of `{genesis}-xxx`. The stat tool tries to look for a file with this segment range, and since the static files tend to be named with names beginning with prefix of multiples of 500000 (default), the tool panics (at `get_segment_provider_for_range`) and shutdown.

Instead, perhaps it could skip over missing files and print whatever information is present in the snapshot.


Vui-Chee:
@mattsse Hope you can provide feedback :)

PR #21844, opened 2026-02-05
fix: optimize bootstrap node configuration

andrewshab3:
Refactored bootstrap node extension to iterate directly instead of creating a copy.


mattsse (commented):
we can keep as is

PR #21849, opened 2026-02-05
refactor(reth-bench): extract waiter setup into shared helper

merge-script:
The waiter initialization logic was duplicated across `new_payload_fcu` and `replay_payloads`. Pulled it out into `create_waiter()` and `log_waiter_config()` in the persistence_waiter module.


mattsse (commented):
no idea

PR #21855, opened 2026-02-05
fix: support EIP-1559 params configuration for Optimism dev mode

xzav3r:
## Description

Fixes Optimism dev mode startup issue by properly configuring EIP-1559 parameters required after Jovian fork activation.

## Problem

Running `op-reth node --dev --dev.block-time 1s` fails with error:
```
failed to resolve pending payload err=Minimum base fee cannot be None after Jovian
Error advancing the chain: No payload

Location:
    /data/reth/crates/engine/local/src/miner.rs:217:13
```

This occurs because Jovian fork requires EIP-1559 parameters and minimum base fee to be set, but they were missing in dev mode.

## Changes

- Add environment variable support for EIP-1559 params (`OP_DEV_EIP1559_DENOMINATOR`, `OP_DEV_EIP1559_ELASTICITY`, `OP_DEV_GAS_LIMIT`)
- Encode base fee parameters in genesis header `extra_data` when Jovian fork is active at genesis
- Activate Jovian fork at timestamp 0 in dev mode
- Set `min_base_fee` in payload attributes (required after Jovian)
- Add Osaka fork to Ethereum dev hardforks

### DIFF
--- a/crates/engine/local/src/payload.rs
+++ b/crates/engine/local/src/payload.rs
@@ -83,8 +83,11 @@ where
             "7ef8f8a0683079df94aa5b9cf86687d739a60a9b4f0835e520ec4d664e2e415dca17a6df94deaddeaddeaddeaddeaddeaddeaddeaddead00019442000000000000000000000000000000000000158080830f424080b8a4440a5e200000146b000f79c500000000000000040000000066d052e700000000013ad8a3000000000000000000000000000000000000000000000000000000003ef1278700000000000000000000000000000000000000000000000000000000000000012fdf87b89884a61e74b322bbcf60386f543bfae7827725efaaf0ab1de2294a590000000000000000000000006887246668a3b87f54deb3b94ba47a6f63f32985"
         );
 
+        // Configure EIP-1559 parameters for dev mode. These can be overridden via environment
+        // variables (OP_DEV_EIP1559_DENOMINATOR, OP_DEV_EIP1559_ELASTICITY, OP_DEV_GAS_LIMIT),
+        // otherwise defaults from Optimism's BaseFeeParams are used. The parameters are encoded
+        // as an 8-byte value (denominator + elasticity) required by Optimism's Jovian fork.
         let default_eip_1559_params = BaseFeeParams::optimism();
-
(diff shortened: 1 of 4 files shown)


crates/optimism/chainspec/src/lib.rs
mattsse, line 1: can you submit this to https://github.com/op-rs/op-reth instead, this code is scheduled to be removed, so it is currently frozen

crates/ethereum/hardforks/src/hardforks/dev.rs
mattsse, line 1: this makes sense

crates/engine/local/src/payload.rs
```diff
@@ -80,14 +83,30 @@ where
...
+
+        let denominator = env::var("OP_DEV_EIP1559_DENOMINATOR")
+            .ok()
+            .and_then(|v| v.parse::<u32>().ok())
+            .unwrap_or(default_eip_1559_params.max_change_denominator as u32);
+        let elasticity = env::var("OP_DEV_EIP1559_ELASTICITY")
+            .ok()
+            .and_then(|v| v.parse::<u32>().ok())
+            .unwrap_or(default_eip_1559_params.elasticity_multiplier as u32);
+        let gas_limit = env::var("OP_DEV_GAS_LIMIT").ok().and_then(|v| v.parse::<u64>().ok());
```
mattsse, line 99: since this is for dev node support this seems fine, idk about the envars, but seems fine, can we document these on the impl block?

mattsse (changes requested):
ty, this makes sense, but I think we should split this

fyi @theochap

crates/optimism/chainspec/src/lib.rs
xzav3r: I can submit another PR to [op-rs/op-reth](https://github.com/op-rs/op-reth). However, I currently have a project that depends on op-reth, and dev mode is currently broken and cannot run. Would it be possible to fix this issue before freezing the code?

crates/engine/local/src/payload.rs
xzav3r: done

mattsse (approved):

PR #21858, opened 2026-02-05
fix(downloaders): reset metrics on clear

cuiweixie:


crates/net/downloaders/src/bodies/queue.rs
```diff
@@ -52,6 +52,7 @@ where
     pub(crate) fn clear(&mut self) {
         self.inner.clear();
         self.last_requested_block_number.take();
+        self.metrics = BodyDownloaderMetrics::default();
```
mattsse: I think we want to be more selective here, and only reset the gauges

maybe we add BodyDownloaderMetrics::clear as well or something, because we want to preserve the counters

mattsse (dismissed):

DaniPopes (approved):

shekhirin (approved):

PR #21859, opened 2026-02-05
perf: avoid collect to vector and sort

cuiweixie:


mattsse (commented):
this isnt quite the same, because the logic actually ensures that we have canonical entries

PR #21862, opened 2026-02-05
feat(examples): add engine-persistence-threshold example

### LINKED ISSUE #16673: Add example that demonstrates and documents --engine.persistence-threshold 0 and its impacts
### Describe the feature

ref #16511

the way these settings work is that they control how early we flush a new payload to disk, the default threshold is 2 blocks, motivated by efficient reorg handling, meaning that 2 block deep reorgs can always be handled in memory.
The impact of this is however that if reth db's is accessed externally, the disk will always trail by two blocks, this is what #16511 describes.

https://github.com/paradigmxyz/reth/blob/63cc4eccadc5149ac725746cb2ef1c629407de10/crates/node/core/src/args/engine.rs#L15-L21

we want an example crate that is similar to the db-access example

https://github.com/paradigmxyz/reth/blob/63cc4eccadc5149ac725746cb2ef1c629407de10/examples/db-access/src/main.rs#L22-L22

so that we can document this behaviour properly.

## TODO
* add an example and additional docs for this behaviour

### Additional context

_No response_

illegalcall:
Demonstrates `--engine.persistence-threshold 0` behavior for external database readers, addressing the lag issue described in #16511.

Closes #16673

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3772,6 +3772,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "example-engine-persistence-threshold"
+version = "0.0.0"
+dependencies = [
+ "alloy-genesis",
+ "alloy-network",
+ "alloy-primitives",
+ "alloy-signer-local",
+ "clap",
+ "eyre",
+ "futures-util",
+ "reth-ethereum",
+ "serde_json",
+ "tempfile",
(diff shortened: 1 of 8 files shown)


mattsse (commented):
ty, but I think we dont need this anymore really

PR #21868, opened 2026-02-05
chore: introduce v2 storage flag and remove edge flag

Rjected:
Removes the `edge` feature flag. Adds `rocksdb` as a default feature for `reth`. Adds a v2 storage flag so that we can enable this via cli.


yongkangc (approved):
Don't see any issues here

joshieDo (approved):

PR #21871, opened 2026-02-05
chore: simplify subtrie hashes parallel loop

DaniPopes:
`Self -> Self` collecting can be changed to `iter_mut`.


mattsse (approved):
right, this is equivalent

PR #21877, opened 2026-02-06
fix: only exit when both channels are drained

klkvr:
With https://github.com/paradigmxyz/reth/pull/21768 it might rarely happen that accounts present in `pending_account_updates` are ending up revealed in the accounts trie but not promoted to a proper update immediately

This in combination with `updates` or `proof_result_rx` channels not being fully drained might result in those account updates never being applied to the trie

With this PR we will only exit the loop if both channels are drained

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -436,6 +436,14 @@ where
                 self.dispatch_pending_targets();
                 self.process_new_updates()?;
                 self.promote_pending_account_updates()?;
+
+                if self.finished_state_updates &&
+                    self.account_updates.is_empty() &&
+                    self.storage_updates.iter().all(|(_, updates)| updates.is_empty())
+                {
+                    break;
+                }
+
                 self.dispatch_pending_targets();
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -436,6 +436,14 @@ where
                 self.dispatch_pending_targets();
                 self.process_new_updates()?;
                 self.promote_pending_account_updates()?;
+
+                if self.finished_state_updates &&
+                    self.account_updates.is_empty() &&
+                    self.storage_updates.iter().all(|(_, updates)| updates.is_empty())
+                {
+                    break;
+                }
```
mattsse, line 445: I have a feeling that we'll change this logic again

for example currently this if chain mstly checks self.updates.is_empty() twice which is bit redundant, we could store this.

we could extract to something like fn has_pending_work or someting that we can document the exit condtion here

mattsse (approved):
this makes sense, id prefer this as a standalone fn, but we can take care of this separately

mediocregopher (approved):

PR #21880, opened 2026-02-06
perf(storage): Static file provider delete_segment should use unwind_queue logic

yongkangc:
**Summary**
`delete_segment` and `delete_segment_below_block` previously deleted static file jars immediately during the pruner run, before `DatabaseProvider::commit()`. 

This bypassed the unwind commit ordering (DB first, then static files) that ensures crash consistency, a crash after jar deletion but before DB commit left the database referencing jars that no longer existed on disk.

**Solution:**
This PR defers jar deletions to `commit()` time, matching the existing `prune_on_commit` pattern. `has_unwind_queued()` now returns `true` when deletions are queued, routing `DatabaseProvider::commit()` through the unwind path (DB commits first, then static files are deleted). 

A single `initialize_index()` call replaces the per-jar reindex loop.

**Expected Impact**
Crash-consistent commit ordering for segment deletions, and O(1) reindexing instead of O(n) per-jar reindex during pruning.

Resolves [RETH-343](https://linear.app/tempoxyz/issue/RETH-343)

### DIFF
--- a/crates/storage/provider/src/providers/static_file/manager.rs
+++ b/crates/storage/provider/src/providers/static_file/manager.rs
@@ -928,17 +928,19 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
     /// Files are removed from disk when [`StaticFileWriter::commit`] is called. Until then the
     /// index and on-disk data remain unchanged.
     ///
-    /// For example if block is 1M and the blocks per file are 500K this will queue files 0-499K
-    /// and 500K-999K for deletion.
+    /// Only **full jars whose expected block range ends strictly below `block`** are queued.
+    /// The jar that *contains* `block` is never deleted because jars can only be removed
+    /// entirely.
     ///
-    /// This will not delete the file that contains the block itself, because files can only be
-    /// removed entirely.
+    /// For example, with 500K blocks per file and `block = 1_000_000`, jars 0-499K and
+    /// 500K-999K are queued while the 1M-1.499M jar is kept.
     ///
-    /// # Safety
+    /// # Highest-jar preservation
     ///
-    /// This method will never delete the highest static file for the segment, even if the
-    /// requested block is higher than the highest block in static files. This ensures we always
-    /// maintain at least one static file if any exist.
+    /// The highest jar for the segment is **always preserved**, even when its blocks are
+    /// entirely below the requested threshold. This means that if only one jar exists, or
+    /// if `block` is higher than every block in static files, no jars are deleted. Use
+    /// [`Self::delete_segment`] to remove all jars including the highest.
     ///
     /// Returns a list of `SegmentHeader`s from the jars that will be deleted.
     pub fn delete_segment_below_block(
@@ -2420,6 +2422,15 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
     ///
     /// If a deletion fails mid-way, unprocessed operations are re-queued so a
     /// subsequent `commit()` can retry them.
+    ///
+    /// # Error semantics
+    ///
+    /// When used through [`DatabaseProvider::commit`](crate::providers::DatabaseProvider),
+    /// the database transaction is committed **before** this method runs. An error
+    /// returned here therefore means the DB changes are already durable but some
+    /// static-file jar deletions did not complete. Callers must **not** attempt to
+    /// roll back the database in response to this error — the unprocessed deletions
+    /// are re-queued and will be retried on the next `commit()` call.
     fn commit(&self) -> ProviderResult<()> {
         // Flush pending writer data.
         self.writers.commit()?;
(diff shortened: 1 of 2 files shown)


yongkangc (commented):
## Review: Deferred `delete_segment` / `delete_segment_below_block`

### 1. Thread-safety of the Mutex-based delete queue

The `Mutex<Vec<(StaticFileSegment, Vec<BlockNumber>)>>` is fine. All three methods (`queue_delete`, `has_delete_queued`, `take_delete_queue`) acquire the lock, do a fast operation, and release. No risk of deadlock since the `Mutex` is never held while acquiring other locks. **No issues.**

### 2. Correctness of computing deletions from index snapshot vs the old loop-and-recheck approach

The old code used a loop that deleted the lowest jar, re-initialized the index, then checked the new lowest. The new code snapshots all keys from `expected_block_ranges_by_max_block` (a `BTreeMap`, so ascending order), which is behaviorally equivalent for fully-filled jars.

**Bug: highest jar exclusion fails for partially-filled jars** in `delete_segment_below_block`.

The old code compared `get_lowest_range_end()` (actual `block_range.end()`) against `get_highest_static_file_block()` (actual `index.max_block`). Both use actual block numbers, so the comparison always works.

The new code compares BTreeMap keys (`expected_block_end()` — the *expected* range end) against `highest_block` (`index.max_block` — the *actual* max block). For a partially-filled highest jar, these differ:

```
Example: 25 blocks, blocks_per_file=10
- Jar 0-9:   expected_block_end = 9,  actual max = 9
- Jar 10-19: expected_block_end = 19, actual max = 19  
- Jar 20-29: expected_block_end = 29, actual max = 24 (only 5 blocks written)

index.max_block = 24
BTreeMap key for last jar = 29

Check: Some(29) == Some(24) → false → highest jar is NOT excluded → BUG
```

If `block >= 30` (or any value above 29), the `max_block >= block` check would also fail, so the partially-filled highest jar gets queued for deletion when it should be preserved.

**Fix**: Compare against the *expected* block end of the highest jar instead of `index.max_block`. For example:
```rust
let highest_expected_end = index.expected_block_ranges_by_max_block.last_key_value().map(|(&k, _)| k);
```
And then compare `Some(max_block) == highest_expected_end`.

### 3. Proper handling of the highest jar exclusion in `delete_segment_below_block`

As noted above, the exclusion is broken for partially-filled jars. The `delete_segment` function does NOT have this bug because it intentionally includes all jars (no exclusion needed).

### 4. Commit ordering (writers commit before delete queue drains)

The ordering in `commit()` is correct:
1. `self.writers.commit()` — flush pending writes
2. `self.writers.reset(segment)` — drop writer handles for segments about to be deleted
3. `delete_jar_no_reindex` — delete files
4. `self.initialize_index()` — rebuild index

The `reset()` call before deletion is important to release file handles. **No issues.**

### 5. Edge cases

**Calling `delete_segment` or `delete_segment_below_block` twice before `commit()`:**
The index is not updated between calls (deletions are deferred), so the second call would compute the same set of jars and queue them again. On `commit()`, the first pass of `delete_jar_no_reindex` deletes the files; the second pass would fail trying to load an already-deleted jar from disk. **This is a bug** — either deduplicate the queue in `commit()` or guard against double-queuing.

**Empty queue on commit:** Handled correctly — early return after `self.writers.commit()`.

**Empty segment:** `expected_block_ranges_by_max_block` would have no entries, so `range_ends_to_delete` stays empty, `queue_delete` skips the push (empty check), and nothing happens. **Correct.**

### 6. `finalize()` guard

The guard is correct. If `has_delete_queued()` returns true, `finalize()` returns an error, forcing the caller to use `commit()` instead. This works because `DatabaseProvider::commit()` checks `has_unwind_queued()` (which now includes `has_delete_queued()`) and routes to the `commit()` path rather than `finalize()`. **No issues.**

### 7. Caller chain: do prune callers eventually call `commit()`?

The prune segments (`prune_static_files`, `delete_static_files_segment`, `AccountHistory::prune_static_files`, `StorageHistory::prune_static_files`) call `delete_segment_below_block` / `delete_segment` but do NOT call `commit()` themselves.

The caller chain is:
1. `Segment::prune()` returns `SegmentOutput`
2. `Pruner::run_with_provider()` orchestrates all segments
3. `Pruner::run()` calls `provider.commit()` which calls `DatabaseProvider::commit()`
4. `DatabaseProvider::commit()` checks `has_unwind_queued()` → now returns `true` due to queued deletes → takes the unwind path → calls `self.static_file_provider.commit()`

**This is correct.** The deferred deletions will be flushed when `Pruner::run()` commits the `DatabaseProviderRW`.

### 8. Minor observations

- Tracing target changed from `"providers::static_file"` to `"provider::static_file"` throughout. This is a breaking change for anyone filtering logs by the old target. Consider doing this in a separate PR or mentioning it in the PR description.
- The `#[instrument]` macros were removed from `StaticFileWriters::commit`, `StaticFileWriters::finalize`, `StaticFileProviderRW::finalize`, and `StaticFileProviderRW::commit`. This reduces structured tracing span information. Was this intentional?
- `segments_seen` in `commit()` uses a `Vec` with linear search. A `HashSet` or `SmallVec` would be more idiomatic, though the segment count is small enough that it does not matter for performance.
- Header reading in both `delete_segment` and `delete_segment_below_block` duplicates the same load-from-map-or-disk pattern. Consider extracting a helper like `fn read_segment_header(&self, segment, range_end) -> ProviderResult<SegmentHeader>`.

### Summary

Two bugs found:
1. **P1**: Highest jar exclusion in `delete_segment_below_block` is broken for partially-filled jars due to comparing expected block end (BTreeMap key) with actual max block (`index.max_block`).
2. **P2**: Calling `delete_segment` / `delete_segment_below_block` twice before `commit()` will queue duplicate deletions, causing the second `delete_jar_no_reindex` call to fail when files are already deleted.

Tests pass but only cover fully-filled jars. Add a test with a partially-filled last jar (e.g., 25 blocks with blocks_per_file=10) to catch bug #1.

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -979,39 +1005,60 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
         let header = jar.user_header().clone();
         jar.delete().map_err(ProviderError::other)?;
 
-        // SAFETY: this is currently necessary to ensure that certain indexes like
-        // `static_files_min_block` have the correct values after pruning.
-        self.initialize_index()?;
-
```
yongkangc, line 985: Is it okay tot remove this safety?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -2353,14 +2400,55 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
...
+        let remaining = ops.split_off(completed);
+        if !remaining.is_empty() {
+            self.0.queue_delete_raw(remaining);
+        }
+
+        self.initialize_index()?;
+
+        if let Some(err) = delete_err {
+            return Err(err);
+        }
```
yongkangc, line 2475: add docstring for this function and for each section addd comments to allow easier to understand

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -2353,14 +2400,55 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
...
+
+        Ok(())
     }
 
     fn has_unwind_queued(&self) -> bool {
```
yongkangc, line 2481: add docstring - 1 liner for this

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -2353,14 +2400,55 @@ impl<N: NodePrimitives> StaticFileWriter for StaticFileProvider<N> {
...
-        self.writers.has_unwind_queued()
+        self.writers.has_unwind_queued() || self.has_queued_deletions()
     }
 
     fn finalize(&self) -> ProviderResult<()> {
```
yongkangc, line 2488: add docstring - 1 liner for this

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -188,6 +188,18 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
...
+        let mut writer = match segment {
+            StaticFileSegment::Headers => self.headers.write(),
+            StaticFileSegment::Transactions => self.transactions.write(),
+            StaticFileSegment::Receipts => self.receipts.write(),
+            StaticFileSegment::TransactionSenders => self.transaction_senders.write(),
+            StaticFileSegment::AccountChangeSets => self.account_change_sets.write(),
+            StaticFileSegment::StorageChangeSets => self.storage_change_sets.write(),
+        };
+        *writer = None;
+    }
```
yongkangc, line 203: why do we need this? add a docstring comment for this

PR #21886, opened 2026-02-06
perf(engine): merge pruning into save_blocks write transaction

yongkangc:
**Summary**
After `on_save_blocks` committed blocks (fsync #1), the persistence thread ran pruning in a separate MDBX write transaction with its own commit (fsync #2). During this entire pruning pass, the persistence thread could not process new requests. This merges pruning into the same write transaction by calling `Pruner::run_with_provider()` with the existing `provider_rw` before `commit()`, eliminating the second fsync entirely.

**Changes**
- Move pruning into `on_save_blocks` before `provider_rw.commit()` using the existing `run_with_provider` API
- Catch and log prune errors without preventing block persistence (preserves existing guarantee)
- Remove standalone `prune_before` method and post-save pruning from `run()` loop
- Clean up unused `PrunerOutput` import

**Expected Impact**
Eliminates ~128ms of redundant fsync latency on every prune cycle (fires every other save). Based on bench metrics: prune accounts for 14.9% of total persistence wall time (53s / 356s). Each persistence cycle becomes a single atomic DB update with one fsync instead of two.

**Notes**
The `PersistenceError::PrunerError` variant is retained for API compatibility even though it is no longer constructed in this code path. The pruner's built-in `delete_limit` and `timeout` bound the additional transaction duration.

### DIFF
--- a/crates/engine/tree/src/persistence.rs
+++ b/crates/engine/tree/src/persistence.rs
@@ -9,7 +9,7 @@ use reth_provider::{
     providers::ProviderNodeTypes, BlockExecutionWriter, BlockHashReader, ChainStateBlockWriter,
     DBProvider, DatabaseProviderFactory, ProviderFactory, SaveBlocksMode,
 };
-use reth_prune::{PrunerError, PrunerOutput, PrunerWithFactory};
+use reth_prune::{PrunerError, PrunerWithFactory};
 use reth_stages_api::{MetricEvent, MetricEventsSender};
 use reth_tasks::spawn_os_thread;
 use std::{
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/persistence.rs
```diff
@@ -74,18 +74,6 @@ where
...
-    /// configuration.
-    #[instrument(level = "debug", target = "engine::persistence", skip_all, fields(block_num))]
-    fn prune_before(&mut self, block_num: u64) -> Result<PrunerOutput, PrunerError> {
-        debug!(target: "engine::persistence", ?block_num, "Running pruner");
-        let start_time = Instant::now();
-        // TODO: doing this properly depends on pruner segment changes
-        let result = self.pruner.run(block_num);
-        self.metrics.prune_before_duration_seconds.record(start_time.elapsed());
-        result
-    }
```
Rjected, line 88: would like to keep a helper method for this if possible and the instrumentation

Rjected (changes requested):
I do think it would be worth adding helpers for running the pruner with the same provider_rw, so that we only commit once per batch of blocks, would like the cached state / unrelated changes removed though.

PR #21898, opened 2026-02-06
test(e2e): add eth_simulateV1 test for maxFeePerBlobGas without blob data

gakonst:
Adds an e2e test that verifies `eth_simulateV1` correctly handles transactions with `maxFeePerBlobGas` set but no `blob_versioned_hashes` or sidecar. The transaction should be treated as EIP-1559, not EIP-4844.

This test reproduces [#21809](https://github.com/paradigmxyz/reth/issues/21809) where reth returned `Transaction is not buildable` while geth, besu, nethermind, and erigon all succeeded.

**Root cause fix:** alloy-rs/alloy#3651 - `preferred_type()` and `minimal_tx_type()` were returning `TxType::Eip4844` when only `max_fee_per_blob_gas` was set, even without `blob_versioned_hashes`. This matches go-ethereum behavior where only `BlobHashes != nil` triggers blob tx type.

**This test will pass once the alloy dependency is bumped** to include the fix.

Closes #21809


gakonst:
Reopening as a clean PR without polluted .github changes

PR #21904, opened 2026-02-06
fix(evm): add range validation in ExecutionOutcome::split_at

andrewshab3:
When calling `split_at(at)` with `at < self.first_block`, a panic occurred in `at.checked_sub(1).unwrap()` before range validation. This resulted in an uninformative error message and violated the method's contract documented in the docstring.


mattsse (commented):
no need for this

PR #21905, opened 2026-02-06
fix: add more safety checks to reveals of upper subtrie nodes

klkvr:
Makes sure we don't process lower subtrie reveals anymore if node was already present in upper subtrie

Additionally, now we will always attempt to find the immediate parent of the node and make sure that it is either a leaf or branch node that knows about the child we are currently revealing. This is needed to correctly skip revealing nodes that were already removed from the subtrie


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2072,9 +2072,48 @@ impl ParallelSparseTrie {
...
     ) -> SparseTrieResult<()> {
-        // If there is no subtrie for the path it means the path is UPPER_TRIE_MAX_DEPTH or less
-        // nibbles, and so belongs to the upper trie.
-        self.upper_subtrie.reveal_node(path, node, masks)?;
+        // For upper subtrie ensure that the node was not removed before by finding its parent and
+        // making sure it knows about the child we are revealing.
+        if !self.upper_subtrie.nodes.contains_key(&path) && !path.is_empty() {
+            // Find the closest parent node walking up the path.
+            let mut parent = path.slice(..path.len() - 1);
+            while !self.upper_subtrie.nodes.contains_key(&parent) && !parent.is_empty() {
```
mediocregopher: `parent` could be the root node, so not sure if the `is_empty` check is right

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2072,9 +2072,48 @@ impl ParallelSparseTrie {
         node: &TrieNode,
         masks: Option<BranchNodeMasks>,
     ) -> SparseTrieResult<()> {
-        // If there is no subtrie for the path it means the path is UPPER_TRIE_MAX_DEPTH or less
-        // nibbles, and so belongs to the upper trie.
-        self.upper_subtrie.reveal_node(path, node, masks)?;
+        // For upper subtrie ensure that the node was not removed before by finding its parent and
+        // making sure it knows about the child we are revealing.
+        if !self.upper_subtrie.nodes.contains_key(&path) && !path.is_empty() {
```
mediocregopher: Right now this check is being made for revealing a node in the upper trie, but what we wanted was to make this check when revealing the root of a _lower_ trie. So I think this would belong at the top of `SparseSubtrie::reveal_node` if `self.path == path`.

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -214,13 +214,20 @@ impl SparseTrie for ParallelSparseTrie {
...
+
         if !self.is_reveal_parallelism_enabled(lower_nodes.len()) {
             for node in lower_nodes {
-                if let Some(subtrie) = self.lower_subtrie_for_path_mut(&node.path) {
-                    subtrie.reveal_node(node.path, &node.node, node.masks)?;
-                } else {
-                    panic!("upper subtrie node {node:?} found amongst lower nodes");
+                let idx = path_subtrie_index_unchecked(&node.path);
+                if !reachable_subtries.get(idx) {
+                    continue;
```
mediocregopher, line 228: nit: let's `trace!` here, i feel like we'll want to know this is happening if debugging something in the future

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2650,17 +2737,7 @@ impl SparseSubtrie {
...
-                        SparseNode::Branch { .. } | SparseNode::Extension { .. } => {}
-                        // All other node types can't be handled.
-                        node @ (SparseNode::Empty | SparseNode::Leaf { .. }) => {
-                            return Err(SparseTrieErrorKind::Reveal {
-                                path: *entry.key(),
-                                node: Box::new(node.clone()),
-                            }
-                            .into())
-                        }
+                        _ => unreachable!("checked that node is either a hash or non-existent"),
```
mediocregopher, line 2745: Not 100% this is unreachable, we've only done this check for root of lower subtries right?

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -247,19 +254,24 @@ impl SparseTrie for ParallelSparseTrie {
...
+                .filter_map(|nodes| {
                     // NOTE: chunk_by won't produce empty groups
                     let node = &nodes[0];
                     let idx =
                         SparseSubtrieType::from_path(&node.path).lower_index().unwrap_or_else(
                             || panic!("upper subtrie node {node:?} found amongst lower nodes"),
                         );
+
+                    if !reachable_subtries.get(idx) {
+                        return None;
```
mediocregopher, line 276: `trace!` here too

crates/trie/sparse-parallel/src/trie.rs
klkvr: i think this check at the beginning of `reveal_node` covers us?

https://github.com/paradigmxyz/reth/blob/481ff37033c057c07bf4ef0e6a06c5e6ab7880a3/crates/trie/sparse-parallel/src/trie.rs#L2699-L2702

mediocregopher: Ah ok, so this was just never reachable even before

mediocregopher (approved):

PR #21906, opened 2026-02-06
feat(trie): Add root_node method to v2 ProofCalculator

mediocregopher:
We were previously using a bit of a hack to calculate the root hash of the tries using proof v2, passing in `B256::ZERO` as a target. This ended up being a contributor to SR mismatches we were seeing. This PR properly implements a `root_node` method on the ProofCalculator, and a `storage_root_node` method for the StorageProofCalculator, and uses those instead of the hack.

We return the node and not just the hash because we need the node itself to reveal the storage tries, in cases where the account has been updated but no slots have been.

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -539,9 +539,11 @@ where
 
         let proof_start = Instant::now();
 
-        // If targets is empty it means the caller only wants the root hash.
+        // If targets is empty it means the caller only wants the root node.
         let (proof, root) = if targets.is_empty() {
-            (Vec::new(), Some(calculator.storage_root(hashed_address)?))
+            let root_node = calculator.storage_root_node(hashed_address)?;
+            let root = calculator.compute_root_hash(std::slice::from_ref(&root_node))?;
+            (vec![root_node], root)
         } else {
             let proof = calculator.storage_proof(hashed_address, &mut targets)?;
             let root = calculator.compute_root_hash(&proof)?;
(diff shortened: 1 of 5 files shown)


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -1333,6 +1337,75 @@ where
...
         self.proof_inner(value_encoder, targets)
     }
+
+    /// Computes the root hash from a set of proof nodes.
+    ///
+    /// Returns `None` if there is no root node (partial proof), otherwise returns the hash of the
+    /// root node.
+    ///
+    /// This method reuses the internal RLP encode buffer for efficiency.
+    pub fn compute_root_hash(
```
mediocregopher, line 1347: This was moved up from below, it hasn't changed

crates/trie/parallel/src/proof_task.rs
```diff
@@ -545,7 +538,15 @@ where
         let _span_guard = span.enter();
 
         let proof_start = Instant::now();
-        let proof = calculator.storage_proof(hashed_address, &mut targets)?;
+
+        // If targets is empty it means the caller only wants the root node.
+        let proof = if targets.is_empty() {
+            let root_node = calculator.storage_root_node(hashed_address)?;
+            vec![root_node]
```
shekhirin, line 545: nit for another PR: we can avoid allocating a vec here, IIUC we always calculate storage proofs for EOAs

shekhirin (approved):

mattsse (approved):
afaict this makes sense

pending @klkvr

PR #21913, opened 2026-02-06
fix(trie): cleanup branch_node_masks on reveal_node failure

gancerlory:
`SerialSparseTrie::reveal_node` inserts into `branch_node_masks` at the start of the method, but if any subsequent operation fails (like `reveal_node_or_hash` or node type conflicts), the method returns early without cleaning up. This leaves the trie in an inconsistent state.

Fixed by tracking whether we inserted new masks and removing them on error.

-Extract core logic into reveal_node_inner
-Add rollback for branch_node_masks on error

### DIFF
--- a/crates/trie/sparse/src/trie.rs
+++ b/crates/trie/sparse/src/trie.rs
@@ -490,133 +490,17 @@ impl SparseTrieTrait for SerialSparseTrie {
             return Ok(())
         }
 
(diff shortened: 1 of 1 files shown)


Rjected:
we're actually going to remove `SerialSparseTrie`: https://github.com/paradigmxyz/reth/pull/21808

PR #21920, opened 2026-02-06
perf(provider): remove redundant clones in static_file tests

0xcharry:
Removes two unnecessary `.clone()` calls in static file provider tests that were causing redundant allocations.


Rjected:
i dont think we should bother removing clones from tests + the range clone is required

PR #21924, opened 2026-02-06
fix(trie): Do not reveal disconnected leaves

mediocregopher:
When a leaf is revealed, and then later it's path changes (e.g. due to a branch node collapse) we need to prevent re-reveals of the original leaf from overwriting the `values` HashMap. To do this we check that all leaves are connectable from their branch prior to revealing them.

This check has to be done in a special way for leaves sitting on the upper/lower border, as they will be on the lower subtrie and their branch on the upper, so for these we can't easily do the check from the SparseSubtrie.

### DIFF
--- a/crates/trie/sparse-parallel/src/trie.rs
+++ b/crates/trie/sparse-parallel/src/trie.rs
@@ -2795,18 +2795,6 @@ impl SparseSubtrie {
                 self.nodes.insert(path, SparseNode::Empty);
             }
             TrieNode::Branch(branch) => {
-                // For a branch node, iterate over all children
-                let mut stack_ptr = branch.as_ref().first_child_index();
-                for idx in branch.state_mask.iter() {
-                    let mut child_path = path;
-                    child_path.push_unchecked(idx);
-                    if Self::is_child_same_level(&path, &child_path) {
-                        // Reveal each child node or hash it has, but only if the child is on
-                        // the same level as the parent.
-                        self.reveal_node_or_hash(child_path, &branch.stack[stack_ptr])?;
-                    }
-                    stack_ptr += 1;
-                }
                 // Update the branch node entry in the nodes map, handling cases where a blinded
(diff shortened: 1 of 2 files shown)


crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2224,6 +2258,31 @@ impl ParallelSparseTrie {
...
+            return true
+        }
+
+        let parent_path = path.slice(..path.len() - 1);
+        let leaf_nibble = path.get_unchecked(path.len() - 1);
+
+        match upper_nodes.get(&parent_path) {
+            Some(SparseNode::Branch { state_mask, .. }) => state_mask.is_bit_set(leaf_nibble),
+            _ => false,
+        }
```
klkvr, line 2283: i see, i guess this is enough to just check that parent exists and has this child's nibble because no orphan nodes can be present in upper subtrie because we now do this check https://github.com/paradigmxyz/reth/blob/afa3b6eaecfec2253450962e37b619ef61aa5a77/crates/trie/sparse-parallel/src/trie.rs#L2126-L2130

crates/trie/sparse-parallel/src/trie.rs
```diff
@@ -2396,6 +2455,28 @@ impl SparseSubtrie {
...
+    /// for this leaf's nibble, meaning the leaf is not reachable.
+    fn is_leaf_reachable_from_parent(&self, path: &Nibbles, _node: &TrieNode) -> bool {
+        if path.is_empty() {
+            return true
+        }
+
+        let parent_path = path.slice(..path.len() - 1);
+        let leaf_nibble = path.get_unchecked(path.len() - 1);
+
+        match self.nodes.get(&parent_path) {
```
klkvr, line 2474: hmm i'm not sure if we can make such assumptions here? iiuc technically we could still have disconnected nodes in lower subtries

klkvr (approved):

PR #21925, opened 2026-02-06
refactor(trie): separate leaf data from structural nodes in parallel sparse trie

mattsse:
Removes the `Leaf` variant from the parallel sparse trie's `SparseNode` and stores leaf data in a separate `LeafValue` struct (`node.rs`). Leaves are now managed through `values` (full_path → LeafValue) and `leaf_prefixes` (prefix_path → full_path) maps on `SparseSubtrieInner`.

This makes the upper subtrie a stable routing skeleton that doesn't change shape on leaf updates — a prerequisite for parallel `update_leaves`.

### DIFF
--- a/crates/trie/sparse-parallel/src/node.rs
+++ b/crates/trie/sparse-parallel/src/node.rs
@@ -1,0 +1,122 @@
+use alloy_primitives::B256;
+use alloy_trie::TrieMask;
+use reth_trie_common::Nibbles;
+
+/// Data stored for each leaf in the sparse trie.
+///
+/// This combines the short key (suffix), value, and cached hash that were previously
+/// split between `SparseNode::Leaf` (in nodes map) and `values` map.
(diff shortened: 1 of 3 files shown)


crates/trie/sparse-parallel/src/node.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+/// Sparse trie node type for the parallel trie.
+///
+/// Unlike the upstream `SparseNode`, this type does **not** have a `Leaf` variant.
+/// Leaves are stored externally in [`LeafValue`] via the `leaf_prefixes` / `values` maps
+/// on `SparseSubtrieInner`, so structural nodes and leaf data are fully separated.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub(crate) enum SparseNode {
```
mattsse, line 33: this is duplicated from spars-trie minues the leaf variant

PR #21928, opened 2026-02-07
perf(engine): reduce save_cache write lock hold time

gakonst:
## Summary
Moves `insert_state` and `valid_block_rx.recv()` outside the `PayloadExecutionCache` write lock, reducing lock hold time from O(state_size + validation_wait) to O(1) pointer swap.

## Motivation
The `save_cache` method held a write lock on `PayloadExecutionCache` during the entire `insert_state()` iteration over all accounts/storage, plus a blocking `recv()` waiting for state root validation. Any `get_cache_for()` call (new block processing) would block until completion.

## Changes
- `saved_cache.split()` + `SavedCache::new()` + `insert_state()` now happen outside the lock on a thread-local `new_cache`
- `valid_block_rx.recv()` blocking wait moved outside the lock
- Write lock only acquired for the final `Option<SavedCache>` pointer swap

## Testing
- `cargo check -p reth-engine-tree` passes
- CI benchmark submitted (run `9ede220a`)

Thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770428917299909

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -222,36 +222,36 @@ where
 
         if let Some(saved_cache) = saved_cache {
             debug!(target: "engine::caching", parent_hash=?hash, "Updating execution cache");
-            // Perform all cache operations atomically under the lock
-            execution_cache.update_with_guard(|cached| {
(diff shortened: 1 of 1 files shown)


gakonst:
Closing - benchmark showed results within noise, not promising enough

PR #21932, opened 2026-02-07
perf(caching): move blocking recv outside RwLock write guard in save_cache

gakonst:
## Summary
Reduce save_cache write lock hold time from unbounded (blocking on validation) to O(1) pointer swap.

## Motivation
save_cache held the RwLock write guard while calling insert_state(), update_metrics(), AND valid_block_rx.recv() (a blocking channel wait for state root validation). This blocked all get_cache_for readers — meaning prewarming for the next block was stalled for potentially hundreds of milliseconds while validation completed.

Slack thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770425974908369

## Changes
- Move saved_cache.split(), insert_state(), update_metrics(), and valid_block_rx.recv() outside the write lock
- Only hold the write lock for the final pointer swap
- Add save_cache_validation_wait_duration and save_cache_lock_hold_duration histogram metrics

## Safety
- insert_state operates on lock-free fixed_cache::Cache internally
- new_cache is not in the shared slot during prep — readers see old cache
- Prewarm workers guaranteed terminated before save_cache runs
- Final swap still atomic via write lock

## Testing
- cargo check -p reth-engine-tree passes
- Submitting to reth-bench-compare for perf validation

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -222,36 +222,43 @@ where
 
         if let Some(saved_cache) = saved_cache {
             debug!(target: "engine::caching", parent_hash=?hash, "Updating execution cache");
-            // Perform all cache operations atomically under the lock
+
+            // Phase 1: ALL prep work OUTSIDE the lock
+            // consumes the `SavedCache` held by the prewarming task, which releases its usage
+            // guard
+            let (caches, cache_metrics, disable_cache_metrics) = saved_cache.split();
+            let new_cache = SavedCache::new(hash, caches, cache_metrics)
+                .with_disable_cache_metrics(disable_cache_metrics);
+
+            // insert_state outside lock (fixed-cache is lock-free internally)
+            if new_cache.cache().insert_state(&execution_outcome.state).is_err() {
+                debug!(target: "engine::caching", "cleared execution cache on update error");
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -222,36 +222,43 @@ where
...
-                new_cache.update_metrics();
+            // Phase 1: ALL prep work OUTSIDE the lock
+            // consumes the `SavedCache` held by the prewarming task, which releases its usage
+            // guard
+            let (caches, cache_metrics, disable_cache_metrics) = saved_cache.split();
+            let new_cache = SavedCache::new(hash, caches, cache_metrics)
+                .with_disable_cache_metrics(disable_cache_metrics);
+
+            // insert_state outside lock (fixed-cache is lock-free internally)
+            if new_cache.cache().insert_state(&execution_outcome.state).is_err() {
```
mattsse, line 234: I'm not entirely sure this is safe, because for alternative blocks we can end up here

https://github.com/paradigmxyz/reth/blob/e72e85632b33457798e7c9d58ed68a32e013e0c1/crates/engine/tree/src/tree/payload_processor/mod.rs#L886-L889

while we're polluting the shared cache

mattsse (changes requested):

PR #21936, opened 2026-02-07
perf(engine): skip precompile cache for identity

gakonst:
## Summary

Skip caching for the identity precompile by using the new `Precompile::supports_caching` trait method and `map_cacheable_precompiles` from alloy-evm.

## Motivation

The identity precompile simply copies its input to output, which is an O(n) operation. Caching it is counterproductive because computing the cache key hash is also O(n), so we'd be doing O(n) work to potentially save O(n) work with no net benefit.

## Changes

- Patched alloy-evm to use `map_cacheable_precompiles` (replaces `map_pure_precompiles`)
- Updated `payload_validator.rs` and `prewarm.rs` to use new API
- Removed reth-side `supports_caching` helper (now handled by the trait)

## Dependencies

- alloy-rs/evm#284

## Testing

```
cargo test -p reth-engine-tree precompile_cache
```

### DIFF
--- a/crates/engine/tree/src/tree/precompile_cache.rs
+++ b/crates/engine/tree/src/tree/precompile_cache.rs
@@ -11,6 +11,24 @@ use std::{hash::Hash, sync::Arc};
 /// Default max cache size for [`PrecompileCache`]
 const MAX_CACHE_SIZE: u32 = 10_000;
 
+/// Identity precompile address (0x04).
+///
+/// The identity precompile simply copies its input to output, which is an O(n) operation.
+/// Caching it is counterproductive because computing the cache key hash is also O(n),
+/// so we'd be doing O(n) work to potentially save O(n) work with no net benefit.
+const IDENTITY_PRECOMPILE: Address =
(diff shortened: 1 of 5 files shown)


crates/engine/tree/src/tree/precompile_cache.rs
```diff
@@ -11,6 +11,24 @@ use std::{hash::Hash, sync::Arc};
 /// Default max cache size for [`PrecompileCache`]
 const MAX_CACHE_SIZE: u32 = 10_000;
 
+/// Identity precompile address (0x04).
+///
+/// The identity precompile simply copies its input to output, which is an O(n) operation.
+/// Caching it is counterproductive because computing the cache key hash is also O(n),
+/// so we'd be doing O(n) work to potentially save O(n) work with no net benefit.
+const IDENTITY_PRECOMPILE: Address =
+    Address::new([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x04]);
```
mattsse: use the address! macro for this

mattsse (changes requested):

PR #21938, opened 2026-02-07
chore(trie): move revm-database to dev-dependencies

outlineeee:
`revm-database` is only used in `benches/hash_post_state.rs`, so it should be a dev-dependency.

Also removes duplicate entries in the `serde` feature and fixes `parking_lot?/serde` syntax for optional dep.


mattsse:
Closing — while the changes are technically correct, we'll handle this as part of a broader dependency cleanup. Thanks for the contribution.

PR #21942, opened 2026-02-07
perf(engine/tree): avoid redundant hash lookups in tree state and block buffer

Abhisheklearn12:
## Summary                                                                                                                              
 - Replace `max_by_key` with `last()` in `persist_blocks` , input is always ascending from `get_canonical_blocks_to_persist`, so `last()` is O(1) instead of O(n)                                                                                                                         
 - Use entry API in `TreeState::insert_executed` to avoid double hash+probe on `B256Map` (`contains_key` + `insert` → single                                                                                              `entry`)                                                                                                                                                 
 - Use entry API in `BlockBuffer::remove_from_parent` to avoid redundant HashMap lookup on empty-set cleanup path                                                                                                       
                                                                                                                                                                          
## Test plan                                                                                                                                               
- [x] Existing tests pass (`cargo nextest run -p reth-engine-tree`)                                                                                        
- [x] No behavior change, only fewer hash computations per call

### DIFF
--- a/crates/engine/tree/src/tree/block_buffer.rs
+++ b/crates/engine/tree/src/tree/block_buffer.rs
@@ -2,7 +2,7 @@ use crate::tree::metrics::BlockBufferMetrics;
 use alloy_consensus::BlockHeader;
 use alloy_primitives::{BlockHash, BlockNumber};
 use reth_primitives_traits::{Block, SealedBlock};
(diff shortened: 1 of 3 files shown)


mattsse:
Closing — these are trivial micro-optimizations on tiny collections (block buffer is a handful of entries) with no measurable performance impact. Thanks for the interest, but we'd prefer contributions that address real issues or come with profiling data showing actual impact.

PR #21949, opened 2026-02-07
perf(prewarm): route transactions to workers by sender address

gakonst:
## Summary
Routes same-sender transactions to the same prewarm worker via per-worker channels instead of a shared work-stealing channel.

## Motivation
Same-sender transactions share account state (balance, nonce). With the previous shared-channel model, any worker could pick up any transaction, meaning the same sender's account would be fetched from disk/cache independently by multiple workers. Sender-affinity routing keeps that state warm in a single worker's EVM instance.

## Changes
- `spawn_workers` now creates per-worker `crossbeam_channel`s instead of a single shared channel
- `spawn_all` routes each tx to a worker via `sender_to_worker_index()` (address low-bytes mod num_workers)
- Added `.max(1)` guard on `workers_needed` to prevent divide-by-zero

## Testing
`cargo check -p reth-engine-tree` and `cargo +nightly clippy -p reth-engine-tree` pass clean.

Thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770485728341129

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -131,6 +131,10 @@ where
 
     /// Spawns all pending transactions as blocking tasks by first chunking them.
     ///
+    /// Transactions from the same sender are routed to the same worker so that
+    /// sender account state (balance, nonce) is warm in that worker's local EVM
+    /// cache, reducing redundant state lookups.
+    ///
     /// For Optimism chains, special handling is applied to the first transaction if it's a
     /// deposit transaction (type 0x7E/126) which sets critical metadata that affects all
     /// subsequent transactions in the block.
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -736,6 +750,18 @@ where
...
+/// Maps a sender address to a worker index using the low bytes of the address.
+///
+/// Addresses have good entropy in their low bytes (they're derived from keccak of the public
+/// key), so a simple modulo gives reasonable distribution across workers.
+#[inline]
+fn sender_to_worker_index(sender: &alloy_primitives::Address, num_workers: usize) -> usize {
+    debug_assert!(num_workers > 0);
+    let bytes = sender.as_slice();
+    let val = u64::from_le_bytes(bytes[12..20].try_into().expect("20-byte address"));
+    val as usize % num_workers
```
mattsse, line 762: this likely doesnt work, because this can route txs to workers that are spawned later, delaying them

PR #21950, opened 2026-02-07
refactor(engine): wrap ExecutionCache internals in single Arc

gakonst:
## Summary
Wraps all `ExecutionCache` fields in a single `Arc<ExecutionCacheInner>` instead of wrapping each field individually in its own `Arc`.

## Changes
- Introduced `ExecutionCacheInner` struct holding all cache fields directly
- `ExecutionCache` is now a newtype around `Arc<ExecutionCacheInner>`
- Reduces cloning overhead from 7 atomic ref-count increments to 1

Note: `CacheStatsHandler` fields remain `Arc` because they're shared with the `FixedCache` via `Stats::new()`.

## Testing
`cargo test -p reth-engine-tree --lib cached_state` — all 9 tests pass.


DaniPopes (approved):

mattsse (approved):
makes sense

PR #21952, opened 2026-02-07
refactor(ethereum-evm): extract excess blob gas calculation into help…

gap-editor:
### Summary

Extracts the excess blob gas calculation logic from `EthBlockAssembler::assemble_block` into a dedicated `excess_blob_gas_for_block` helper method.

### Changes

- Added a new `impl` block for `EthBlockAssembler<ChainSpec>` with `EthChainSpec + EthereumHardforks` bounds
- Extracted `excess_blob_gas_for_block` method that handles both cases:
  - Parent is already post-Cancun: derives value from parent header
  - First post-fork block: treats `blob_gas_used` and `excess_blob_gas` as zero per EIP-4844
- Added doc comments explaining the method's behavior
- Simplified the Cancun fields block in `assemble_block` to a single method call

### Motivation

The nested `if/else` inside the Cancun block made `assemble_block` slightly harder to scan. Pulling the logic into a named method improves readability and documents the two distinct cases. This also makes the blob gas logic easier to unit test independently in the future.
unit tests for the different hardfork scenarios in `assemble_block` would be a good follow-up.

### DIFF
--- a/crates/ethereum/evm/src/build.rs
+++ b/crates/ethereum/evm/src/build.rs
@@ -25,6 +25,30 @@ impl<ChainSpec> EthBlockAssembler<ChainSpec> {
     }
 }
 
+impl<ChainSpec: EthChainSpec + EthereumHardforks> EthBlockAssembler<ChainSpec> {
+    /// Calculates the excess blob gas for the next block.
+    ///
+    /// If the parent block was already post-Cancun, derives the value from the parent header.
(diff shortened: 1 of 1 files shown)


mattsse (commented):
Thanks for the PR! This is a clean extraction, but the helper is private, called from a single site, and doesn't enable any new reuse — it just moves the code into a separate `impl` block. The inline version is already short and readable enough, so this doesn't clear the bar for a refactor-only change.

Closing as a needless refactor.

PR #21954, opened 2026-02-07
fix(reth-bench): add WS keepalive heartbeat to persistence subscription

gakonst:
## Summary
Fixes reth-bench persistence subscription disconnecting during heavy I/O.

## Motivation
During benchmarks, the MDBX persistence commit can take 29+ seconds (all fsync time). While this runs on a dedicated OS thread, the resulting I/O pressure / CPU throttling in k8s can prevent the reth WS server (jsonrpsee/soketto) from responding to WebSocket pings in time.

`alloy-transport-ws` sends a WS ping every 10s and expects a pong within the next 10s. If the pong doesn't arrive, the connection is killed with `WS server missed a pong`, the persistence subscription dies, and reth-bench fails with `Failed to send block data: channel closed`.

## Changes
- Spawn a background keepalive task in `PersistenceSubscription` that sends a lightweight RPC request (`eth_chainId`) every 6 seconds through the WS connection
- This resets alloy's keepalive timer on the client side, so pings are never sent and the pong timeout never triggers
- The keepalive task is aborted on drop

## Testing
- Existing tests pass
- `cargo +nightly clippy -p reth-bench -- -D warnings` clean

Observed in: https://tempoxyz.slack.com/archives/C09R81BF3J5/p1770501497159599

### DIFF
--- a/bin/reth-bench/src/bench/persistence_waiter.rs
+++ b/bin/reth-bench/src/bench/persistence_waiter.rs
@@ -16,10 +16,19 @@ use alloy_transport_ws::WsConnect;
 use eyre::Context;
 use futures::StreamExt;
 use std::time::Duration;
+use tokio::task::JoinHandle;
 use tracing::{debug, info};
 
 /// Default `WebSocket` RPC port for reth.
 const DEFAULT_WS_RPC_PORT: u16 = 8546;
+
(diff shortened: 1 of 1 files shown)


gakonst:
Closing in favor of adding keepalive configuration upstream in alloy

PR #21961, opened 2026-02-08
chore(engine): remove biased select in engine service loop

gakonst:
Removes the `biased;` from the `tokio::select!` in the engine service loop.


mattsse (commented):
@joshieDo should we maybe remove the bias as well?

DaniPopes (approved):

PR #21967, opened 2026-02-08
feat(engine): add --engine.disable-sparse-trie-cache-pruning flag

gakonst:
## Summary
Adds a CLI flag to fully disable sparse trie cache pruning between blocks for benchmarking.

## Motivation
When benchmarking sparse trie cache performance, we need to isolate the effects of pruning vs retaining the full cache. Currently there's no way to skip pruning entirely — `sparse-trie-prune-depth` and `sparse-trie-max-storage-tries` can be tuned but not disabled.

## Changes
- Added `--engine.disable-sparse-trie-cache-pruning` CLI flag (requires `--engine.enable-sparse-trie-as-cache`)
- Plumbed through `DefaultEngineValues` → `EngineArgs` → `TreeConfig` → `PayloadProcessor`
- When set, `SparseTrieCacheTask::into_trie_for_reuse` skips the `trie.prune()` call but still runs `shrink_to` for capacity management
- Updated existing CLI arg test

## Testing
```
cargo test -p reth-node-core -- engine
```
- `test_parse_engine_args`: verifies defaults parse correctly
- `engine_args`: verifies the new flag parses from CLI args

Note: this flag will cause memory to grow unboundedly since no trie nodes are evicted. It is intended strictly for benchmarking.


shekhirin (approved):

mattsse (approved):
I think this makes sense for now, but this will giga balloon mem

PR #21972, opened 2026-02-09
chore(engine): add engine_newPayloadV5 and engine_getPayloadV6 to capabilities list

Elaela22soL:
These methods are implemented but missing from the capabilities list, causing engine_exchangeCapabilities to omit them from the response.


mattsse (commented):
these are upcoming cant include them yet

PR #21973, opened 2026-02-09
chore(trie): Spans and traces for sparse trie

mediocregopher:
* Propagate spans to storage trie rayon threads in SparseTrieCacheTask, so that the hashed address of the storage trie shows up in its traces.

* Add traces to ParallelSparseTrie which are useful when debugging. These are the traces I end up adding to the codebase manually everytime I have to debug something, imo we should just commit them.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -623,6 +623,7 @@ where
             if new { &mut self.new_storage_updates } else { &mut self.storage_updates };
 
         // Process all storage updates in parallel, skipping tries with no pending updates.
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -623,6 +623,7 @@ where
             if new { &mut self.new_storage_updates } else { &mut self.storage_updates };
 
         // Process all storage updates in parallel, skipping tries with no pending updates.
+        let span = tracing::Span::current();
```
klkvr, line 626: should we drop this before processing account leaf updates?

klkvr (approved):

PR #21982, opened 2026-02-09
perf(engine): parallelize transaction root computation for payloads

gakonst:
## Summary
Spawn a background task to compute the transaction root from the raw encoded payload transactions concurrently with block execution, avoiding the expensive re-encode + MPT build during post-execution validation.

## Motivation
The `calculate_tx_root` in `validate_block_pre_execution` re-encodes all transactions to RLP and builds a Merkle Patricia Trie (~2-8ms on mainnet blocks). By computing it from the raw payload bytes in parallel with execution, we eliminate this from the critical path.

Resolves reth-384.

## Changes
- Add `raw_transactions()` to `ExecutionPayload` trait to expose pre-encoded tx bytes
- Add optional `transactions_root: Option<B256>` to `Consensus::validate_block_pre_execution` (follows the same pattern as `receipt_root_bloom` in `validate_block_post_execution`)
- Spawn tx root task using `ordered_trie_root_encoded` at the top of `validate_block_with_state` (payloads only, not regular blocks)
- Plumb pre-computed root through `validate_post_execution` → `validate_block_inner` → `validate_block_pre_execution`

## Testing
```
cargo check -p reth-engine-tree --all-targets
cargo check -p reth-node-builder -p reth-node-ethereum -p reth-downloaders -p reth-rpc -p reth-stateless
```

### DIFF
--- a/crates/consensus/common/src/validation.rs
+++ b/crates/consensus/common/src/validation.rs
@@ -2,6 +2,7 @@
 
 use alloy_consensus::{BlockHeader as _, EMPTY_OMMER_ROOT_HASH};
 use alloy_eips::{eip4844::DATA_GAS_PER_BLOB, eip7840::BlobParams};
+use alloy_primitives::B256;
 use reth_chainspec::{EthChainSpec, EthereumHardfork, EthereumHardforks};
 use reth_consensus::ConsensusError;
 use reth_primitives_traits::{
@@ -137,9 +138,13 @@ where
 /// - Compares the ommer hash in the block header to the block body
 /// - Compares the transactions root in the block header to the block body
 /// - Pre-execution transaction validation
+///
+/// If `transactions_root` is provided, the pre-computed transaction root is used instead of
+/// recomputing it from the block body.
 pub fn validate_block_pre_execution<B, ChainSpec>(
     block: &SealedBlock<B>,
     chain_spec: &ChainSpec,
+    transactions_root: Option<B256>,
 ) -> Result<(), ConsensusError>
 where
     B: Block,
@@ -148,7 +153,13 @@ where
     post_merge_hardfork_fields(block, chain_spec)?;
 
     // Check transaction root
-    if let Err(error) = block.ensure_transaction_root_valid() {
+    if let Some(root) = transactions_root {
+        if block.header().transactions_root() != root {
+            return Err(ConsensusError::BodyTransactionRootDiff(
+                GotExpected { got: root, expected: block.header().transactions_root() }.into(),
+            ))
+        }
+    } else if let Err(error) = block.ensure_transaction_root_valid() {
         return Err(ConsensusError::BodyTransactionRootDiff(error.into()))
     }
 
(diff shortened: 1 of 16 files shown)


crates/consensus/common/src/validation.rs
```diff
@@ -137,9 +138,13 @@ where
...
 /// - Compares the transactions root in the block header to the block body
 /// - Pre-execution transaction validation
+///
+/// If `transactions_root` is provided, the pre-computed transaction root is used instead of
+/// recomputing it from the block body.
 pub fn validate_block_pre_execution<B, ChainSpec>(
     block: &SealedBlock<B>,
     chain_spec: &ChainSpec,
+    transactions_root: Option<B256>,
 ) -> Result<(), ConsensusError>
```
mattsse, line 143: I think we should keep this fn as is but instead intrduce another fn that accepts the option, then this wont break a ton of code, like the net/p2p stuff

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -341,6 +341,23 @@ where
...
+        // For payloads, spawn a background task to compute the transaction root from the
+        // raw encoded transactions. This runs concurrently with setup + execution, avoiding
+        // the cost of re-encoding all transactions into an MPT during validation.
+        // Regular blocks skip this since they were already validated when downloaded.
+        let tx_root_rx = match &input {
+            BlockOrPayload::Payload(payload) => {
+                let raw_txs = payload.raw_transactions();
+                let (tx, rx) = tokio::sync::oneshot::channel();
+                self.payload_processor.executor().spawn_blocking(move || {
+                    let root = alloy_trie::root::ordered_trie_root_encoded(raw_txs.as_slice());
```
mattsse, line 353: I think we should spawn this on rayon instead, this blocking task could interfere with the prewarm/storage workers

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -341,6 +341,23 @@ where
...
         Evm: ConfigureEngineEvm<T::ExecutionData, Primitives = N>,
     {
+        // For payloads, spawn a background task to compute the transaction root from the
+        // raw encoded transactions. This runs concurrently with setup + execution, avoiding
+        // the cost of re-encoding all transactions into an MPT during validation.
+        // Regular blocks skip this since they were already validated when downloaded.
+        let tx_root_rx = match &input {
+            BlockOrPayload::Payload(payload) => {
+                let raw_txs = payload.raw_transactions();
+                let (tx, rx) = tokio::sync::oneshot::channel();
```
mattsse, line 351: we use crossbeam for the receipts root, I think we should make this consistent and also use that

crates/engine/tree/src/tree/payload_validator.rs
mattsse, line 353: this is definitely overhead for blocks with < X txs, so maybe we pick a cutoff that is 25txs or smth

crates/payload/primitives/src/payload.rs
```diff
@@ -58,6 +58,12 @@ pub trait ExecutionPayload:
 
     /// Returns the number of transactions in the payload.
     fn transaction_count(&self) -> usize;
+
+    /// Returns the raw RLP-encoded transactions from the payload.
+    ///
+    /// These are the pre-encoded transaction bytes that can be used to compute the
+    /// transaction root without decoding and re-encoding.
+    fn raw_transactions(&self) -> Vec<Bytes>;
```
mattsse, line 66: let's make this return a ref instead then the caller can decide when to clone

mattsse (changes requested):
smol suggestions

PR #21987, opened 2026-02-09
perf: better scheduling for storage roots computation

klkvr:
Changes `SparseTrieCacheTask` so that storage roots are only computed in parallel for tries that don't have root cached. This allows to eliminate/reduce rayon overhead when few or none roots actually need recomputation

Updated `ParallelSparseTrie` `root` method to short-circuit if root node is already cached and added a helper `is_root_cached` to determine if that's the case


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -718,53 +723,50 @@ where
...
             self.pending_account_updates.retain(|addr, account| {
-                // If account has pending storage updates, it is still pending.
-                if self.storage_updates.get(addr).is_some_and(|updates| !updates.is_empty()) {
-                    return true;
+                if let Some(updates) = self.storage_updates.get(addr) {
+                    if !updates.is_empty() {
+                        // If account has pending storage updates, it is still pending.
+                        return true;
+                    } else if let Some(account) = account.take() {
+                        let storage_root = self.trie.storage_root(addr).expect("updates are drained, trie should be revealed by now");
```
mediocregopher:
```suggestion
                        let storage_root = self.trie.storage_root(addr).expect("updates are drained, storage trie should be revealed by now");
```

mediocregopher (approved):

mattsse (approved):

PR #21990, opened 2026-02-09
perf(engine): implement BAL handler for SparseTrieCacheTask

yongkangc:
**Problem**
The STAC path ignores `BlockAccessList` messages — BAL-based state root computation only works with the legacy `MultiProofTask`, not `SparseTrieCacheTask`.

**Solution**
Handle BAL in `run_hashing_task` by converting it to `HashedPostState` and forwarding it as a single state update + `FinishedStateUpdates`, reusing the existing STAC machinery.

**Changes**
- Implement `BlockAccessList` handler in `run_hashing_task` (BAL → `HashedPostState` → sparse trie)
- Plumb `CachedStateProvider`-wrapped provider into STAC task so account reads populate the execution cache
- Add unit test for the full BAL → STAC round-trip

Closes RETH-339


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -340,9 +346,41 @@ where
...
                 }
-                MultiProofMessage::EmptyProof { .. } | MultiProofMessage::BlockAccessList(_) => {
-                    continue
+                // BAL provides the full post-state; convert and send it in one shot.
+                MultiProofMessage::BlockAccessList(bal) => {
+                    let Some(ref provider) = provider else {
+                        error!(target: "engine::tree::payload_processor::sparse_trie", "Received BAL but no provider available");
+                        break;
+                    };
+                    match bal_to_hashed_post_state(&bal, provider.as_ref()) {
```
klkvr: i'm wondering if instead of taking provider here it could make sense to track partial account updates (only nonce/balance/code) and dispatch them as proof fetches instead so that we're not bottlenecked by single provider's Io here and can also immediately apply e.g storage updates to the trie

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -509,6 +523,7 @@ where
         from_multi_proof: CrossbeamReceiver<MultiProofMessage>,
         config: &TreeConfig,
         parent_state_root: B256,
+        provider: Option<StateProviderBox>,
```
mattsse: instead of doing this here, what if we, spawn this where we also spawn the pre warm bal worker, that should already have access to the MultiProofMessage message sender

mattsse (changes requested):

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -340,9 +340,10 @@ where
                 MultiProofMessage::FinishedStateUpdates => {
                     SparseTrieTaskMessage::FinishedStateUpdates
                 }
-                MultiProofMessage::EmptyProof { .. } | MultiProofMessage::BlockAccessList(_) => {
-                    continue
+                MultiProofMessage::EmptyProof { state, .. } => {
+                    SparseTrieTaskMessage::HashedState(state)
```
mattsse: why are we repurposing emptyproof variant for this?

mattsse (approved):
lgtm, pending @klkvr

mediocregopher (approved):
LGTM as an MVP, but @klkvr 's suggestion is good for a followup

PR #21992, opened 2026-02-09
feat(engine): reorg depth commitment metric

shekhirin:
Splits the single `blockchain_tree_reorgs` counter into three counters distinguished by a commitment label (head, safe, finalized), based on where the first reorged block falls relative to the safe and finalized checkpoints.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -38,6 +38,7 @@ use reth_revm::database::StateProviderDatabase;
 use reth_stages_api::ControlFlow;
 use reth_tasks::spawn_os_thread;
 use reth_trie_db::ChangesetCache;
+use revm::interpreter::debug_unreachable;
 use state::TreeState;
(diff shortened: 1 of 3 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2398,9 +2399,23 @@ where
...
+                self.metrics.tree.finalized_reorgs.increment(1);
+            } else if let Some(safe) = self.canonical_in_memory_state.get_safe_num_hash() &&
+                first_reorged_block <= safe.number
+            {
+                self.metrics.tree.safe_reorgs.increment(1);
+            } else {
+                self.metrics.tree.head_reorgs.increment(1);
+            }
+        } else {
+            debug_unreachable!("Reorged chain doesn't have any blocks");
```
Rjected, line 2417: do we need this?

shekhirin: this shouldn't happen but I don't feel comfortable doing an unwrap just for the metric

Rjected (approved):

PR #21994, opened 2026-02-09
fix: enforce zstd compression at TransactionSigned level only

CreeptoGengar:
Add validation to prevent zstd compression at nested Transaction levels.

The check ensures that decompressed transaction data doesn't contain zstd flags, which would indicate incorrect nested compression. This prevents potential conflicts if zstd support is added to Transaction types in the future.


mattsse (commented):
can keep as is

PR #21996, opened 2026-02-09
refactor: remove unnecessary Clone bound from `PayloadTransactionsFixed`

Snezhkko:
`PayloadTransactionsFixed` had exclusive ownership of transactions in a `Vec<T>` but was cloning each element on `next()` via indexed access, requiring a `T: Clone` bound. Since the iterator only moves forward and never revisits elements, cloning was unnecessary overhead.
Replaced `Vec<T>` + `index: usize` with `std::vec::IntoIter<T>`, which yields owned values directly. This removes the Clone trait bound, eliminates per-element cloning, and drops the redundant index field.


mattsse (commented):
can keep as is

PR #22006, opened 2026-02-09
fix: implement missing use_hashed_state support across codebase

joshieDo:
## Summary

Fills in the remaining gaps from the `use_hashed_state` implementation in #21115.

### Changes

**Storage key tagging (`StorageSlotKey` / `ChangesetEntry`)**
- Introduced `ChangesetEntry` in `reth-storage-api` — a changeset entry whose key is tagged as `StorageSlotKey::Plain` or `StorageSlotKey::Hashed`, so downstream code can call `to_hashed()` without consulting `StorageSettings`
- Changed all `StorageChangeSetReader` trait methods (`storage_changeset`, `get_storage_before_block`, `storage_changesets_range`) to return `ChangesetEntry` instead of raw `StorageEntry`
- Updated all implementations: `DatabaseProvider`, `ConsistentProvider`, `BlockchainProvider`, `StaticFileProvider`, `MockEthProvider`, `NoopProvider`

**Trie / hashing (correct key handling)**
- `from_reverts()` no longer takes a `KeyHasher` generic — storage keys are now tagged by the changeset reader, so `to_hashed()` handles both plain (keccak256) and pre-hashed (identity) keys
- `load_prefix_sets_with_provider()` drops its `KH: KeyHasher` generic for the same reason
- `hashed_storage_from_reverts_with_provider()` uses `to_hashed()` instead of unconditionally calling `keccak256`
- `unwind_storage_hashing()` and `unwind_storage_history_indices()` accept `ChangesetEntry` iterators

**Read paths**
- Added `StateProvider::storage_by_hashed_key()` for looking up storage when the key is already hashed (avoids double-hashing in `use_hashed_state` mode)
- `ConsistentProvider::populate_bundle_state()` dispatches between `storage()` and `storage_by_hashed_key()` based on the key tag
- `DatabaseProvider::populate_bundle_state_hashed()` uses pre-hashed storage keys directly for `HashedStorages` lookups instead of re-hashing

**Write paths**
- `write_storage_changesets()` in the static file manager always hashes slot keys with `keccak256` (write path implies hashed state)
- `write_state_reverts()` MDBX path hashes slot keys in changesets when `use_hashed_state`
- `StorageSettings::v2()` now sets `use_hashed_state: true`

**Trait bounds**
- Propagated `StorageSettingsCache` through `FullProvider`, `DatabaseStateRoot`, changeset cache, and related traits
- Implemented `StorageSettingsCache` for `MockEthProvider` and `NoopProvider`

**CLI**
- Added `--hashed` flag to `db state` and `db account-storage` commands to read from `HashedAccounts`/`HashedStorages`

**Pipeline tests**
- Refactored `test_pipeline` / `test_pipeline_v2` into a shared `run_pipeline_forward_and_unwind(storage_settings, num_blocks, unwind_target)` helper
- The helper now runs forward sync → unwind → re-sync, verifying all stage checkpoints and counter storage at each phase
- Added `assert_changesets_queryable()` to verify changesets are readable from the correct source (static files vs MDBX) based on storage settings

**Unit tests**
- `ConsistentProvider`: tests for `get_state` with hashed/plain storage, `storage_changeset` key consistency between DB and in-memory across both modes, `storage_changesets_range` consistency
- `HashedPostStateSorted::from_reverts`: tests for hashed-state mode (pre-hashed keys in static files) and legacy mode (plain keys in MDBX)
- `hashed_storage_from_reverts_with_provider`: tests for both legacy and hashed-state modes
- `KeccakKeyHasher`: basic correctness test
- `load_prefix_sets_with_provider`: removed `KeyHasher` generic requirement


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -747,7 +751,7 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                     r_storage_changesets = Some(self.write_segment(
                         StaticFileSegment::StorageChangeSets,
                         first_block_number,
-                        |w| Self::write_storage_changesets(w, blocks),
+                        |w| Self::write_storage_changesets(w, blocks, ctx.use_hashed_state),
```
joshieDo: should just take ctx

crates/trie/db/src/changesets.rs
```diff
@@ -76,15 +78,27 @@ where
...
     // Step 1: Collect/calculate state reverts
 
+    let use_hashed = provider.cached_storage_settings().use_hashed_state;
+
     // This is just the changes from this specific block
-    let individual_state_revert = HashedPostStateSorted::from_reverts::<KeccakKeyHasher>(
-        provider,
-        block_number..=block_number,
-    )?;
+    let individual_state_revert = if use_hashed {
```
joshieDo: inline here  and all other similar places here when its not reused in the context

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -95,6 +95,8 @@ pub struct StaticFileWriteCtx {
     pub write_account_changesets: bool,
     /// Whether storage changesets should be written to static files.
     pub write_storage_changesets: bool,
+    /// Whether to hash storage keys with keccak256.
+    pub use_hashed_state: bool,
```
joshieDo: shouldnt need, if write_storage_changesets THEN its always hashed state

crates/cli/commands/src/db/state.rs
```diff
@@ -39,6 +39,10 @@ pub struct Command {
     /// Output format (table, json, csv)
     #[arg(long, short, default_value = "table")]
     format: OutputFormat,
+
+    /// Read from HashedAccounts/HashedStorages instead of PlainAccountState/PlainStorageState
+    #[arg(long)]
+    hashed: bool,
```
joshieDo: shouldnt need, check storage settings

crates/cli/commands/src/db/account_storage.rs
```diff
@@ -16,58 +16,100 @@ const LOG_INTERVAL: Duration = Duration::from_secs(5);
 pub struct Command {
     /// The account address to check storage for
     address: Address,
+
+    /// Read from HashedStorages instead of PlainStorageState
+    #[arg(long)]
+    hashed: bool,
```
joshieDo: shouldnt need, check storage settings

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -1494,13 +1494,19 @@ impl<TX: DbTx, N: NodeTypes> StorageChangeSetReader for DatabaseProvider<TX, N>
         address: Address,
         storage_key: B256,
     ) -> ProviderResult<Option<StorageEntry>> {
+        let lookup_key = if self.cached_storage_settings().use_hashed_state {
```
Rjected: should we guard against `use_hashed_state: true` and `storage_changesets_in_static_files: false`?

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -641,9 +641,10 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
                 .flatten()
                 .flat_map(|revert| {
                     revert.storage_revert.into_iter().map(move |(key, revert_to_slot)| {
+                        let raw_key = B256::new(key.to_be_bytes());
                         StorageBeforeTx {
                             address: revert.address,
-                            key: B256::new(key.to_be_bytes()),
+                            key: keccak256(raw_key),
```
Rjected: this is not gated by the storage setting - needs to be i think

crates/cli/commands/src/db/state.rs
```diff
@@ -63,39 +63,63 @@ impl Command {
...
+        let entries = tool.provider_factory.db_ref().view(|tx| {
+            let (account, walker_entries) = if use_hashed_state {
+                let hashed_address = keccak256(address);
+                let account = tx.get::<tables::HashedAccounts>(hashed_address)?;
+                let mut cursor = tx.cursor_dup_read::<tables::HashedStorages>()?;
+                let walker = cursor.walk_dup(Some(hashed_address), None)?;
+                let mut entries = Vec::new();
+                let mut last_log = Instant::now();
+                for (idx, entry) in walker.enumerate() {
+                    let (_, storage_entry) = entry?;
```
Rjected, line 77: ah i see, this is not changeset / history related

Rjected (commented):
looked at it, I want to double check all the places we write, to make sure this is all of them, looks fine so far

crates/storage/provider/src/providers/database/provider.rs
joshieDo: ye i think, we could just add some runtime check that both have to be true?  or enforce

crates/storage/provider/src/providers/static_file/manager.rs
joshieDo: if we're writing storage changesets to static files, the keys are always hashed. Added a runtime enforcement that `use_hashed_state` implies `storage_changesets_in_static_files = true`.

crates/storage/provider/src/providers/database/provider.rs
joshieDo: enforce: https://github.com/paradigmxyz/reth/pull/22006/commits/53b6f373c98dc6d8b5e44615d335f57e73956140

crates/trie/db/src/prefix_set.rs
```diff
@@ -69,3 +69,15 @@ where
         destroyed_accounts,
     })
 }
+
+/// Calls [`load_prefix_sets_with_provider`] with the correct [`KeyHasher`] based on storage settings.
+pub fn load_prefix_sets_auto(
```
joshieDo: no need for this one, inline

crates/storage/storage-api/src/state.rs
```diff
@@ -41,12 +41,29 @@ pub trait StateProvider:
...
+    /// Unlike [`Self::storage`], `hashed_storage_key` must already be keccak256-hashed.
+    /// The `address` remains unhashed (plain) since history indices are keyed by plain address.
+    /// This is used when changeset keys are pre-hashed (e.g., `use_hashed_state` mode)
+    /// to avoid double-hashing.
+    fn storage_by_hashed_key(
+        &self,
+        _address: Address,
+        _hashed_storage_key: StorageKey,
+    ) -> ProviderResult<Option<StorageValue>> {
+        unreachable!("storage_by_hashed_key is only supported by providers with hashed state")
```
Rjected: dont need unreachable in trait def

Rjected (approved):
lgtm

PR #22009, opened 2026-02-09
feat(net): add reason label to backed_off_peers metric

emmajam:
## Summary

Adds a `reason` label to the `backed_off_peers` gauge so operators can distinguish **why** peers are backed off. This addresses the issue where `reth_network_too_many_peers` fires even when the node isn't at peering limits — the metric was conflating remote `TooManyPeers` disconnects with other backoff reasons.

## New metric

```
network_backed_off_peers{reason="too_many_peers"}
network_backed_off_peers{reason="graceful_close"}
network_backed_off_peers{reason="connection_error"}
```

The existing `network_backed_off_peers` total gauge is preserved.

## Changes

- **peers.rs**: Store a `BackoffReason` alongside the backoff expiry in the `backed_off_peers` map. Reason is derived from the concrete `DisconnectReason` at each callsite (`TooManyPeers` vs graceful close vs other errors).
- **metrics.rs**: Add `BackedOffPeersMetrics` struct with labeled gauges.
- **manager.rs**: Report per-reason counts via `update_backed_off_peers_metrics()` helper.

### DIFF
--- a/crates/net/network/src/manager.rs
+++ b/crates/net/network/src/manager.rs
@@ -29,7 +29,7 @@ use crate::{
         PendingSessionFailureMetrics, NETWORK_POOL_TRANSACTIONS_SCOPE,
     },
     network::{NetworkHandle, NetworkHandleMessage},
-    peers::PeersManager,
+    peers::{BackoffReason, PeersManager},
     poll_nested_stream_with_budget,
     protocol::IntoRlpxSubProtocol,
     required_block_filter::RequiredBlockFilter,
@@ -872,10 +872,15 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
                         &peer_id,
                         err,
                     );
+                    self.backed_off_peers_metrics.increment_for_reason(
+                        BackoffReason::from_disconnect(err.as_disconnected()),
+                    );
                     err.as_disconnected()
                 } else {
                     // Gracefully disconnected
                     self.swarm.state_mut().peers_mut().on_active_session_gracefully_closed(peer_id);
+                    self.backed_off_peers_metrics
+                        .increment_for_reason(BackoffReason::GracefulClose);
                     None
                 };
                 self.closed_sessions_metrics.active.increment(1);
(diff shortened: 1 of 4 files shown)


crates/net/network/src/metrics.rs
```diff
@@ -77,6 +77,36 @@ pub struct NetworkMetrics {
...
+    /// Peers backed off due to connection or protocol errors.
+    pub connection_error: Gauge,
+}
+
+impl Default for BackedOffPeersMetrics {
+    fn default() -> Self {
+        Self {
+            too_many_peers: metrics::gauge!("network_backed_off_peers", "reason" => "too_many_peers"),
+            graceful_close: metrics::gauge!("network_backed_off_peers", "reason" => "graceful_close"),
+            connection_error: metrics::gauge!("network_backed_off_peers", "reason" => "connection_error"),
```
mattsse: @DaniPopes same q, can we now get this with new metrics macro now?

crates/net/network/src/manager.rs
```diff
@@ -872,9 +877,7 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
                 if let Some(reason) = reason {
                     self.disconnect_metrics.increment(reason);
                 }
-                self.metrics
-                    .backed_off_peers
-                    .set(self.swarm.state().peers().num_backed_off_peers() as f64);
+                self.update_backed_off_peers_metrics();
```
mattsse: this is quite expensive to do per response, because this loops over all existing backed off peers,

I think we should just incr based on the reason we got here

mattsse (dismissed):

DaniPopes (approved):

crates/net/network/src/peers.rs
```diff
@@ -70,8 +70,8 @@ pub struct PeersManager {
     connection_info: ConnectionInfo,
     /// Tracks unwanted ips/peer ids.
     ban_list: BanList,
-    /// Tracks currently backed off peers.
-    backed_off_peers: HashMap<PeerId, std::time::Instant>,
+    /// Tracks currently backed off peers and the reason for backing off.
+    backed_off_peers: HashMap<PeerId, (std::time::Instant, BackoffReason)>,
```
mattsse: do we even need to store the reason here?

PR #22012, opened 2026-02-09
fix(exex): reset finished_height after sending revert/reorg notification

### LINKED ISSUE #19755: ExEx drops `ChainCommitted` notifications during reorg due to stale `finished_height`, causing missed re-execution and inconsistent state
### Describe the bug

https://github.com/paradigmxyz/reth/blob/26f575440d08227f20f8453b7526d33e3d869eb9/crates/exex/exex/src/manager.rs#L138-L151

This comparison becomes invalid during a reorg/revert, because `finished_height` is still pointing to the old chain while the ExEx has not yet processed the ChainReorged / ChainReverted event. This could lead to race condition.

### Steps to reproduce

NA

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Not running in a container

NA

NA

Testing with OP devnet

Archive (default)

_No response_

_No response_

### Code of Conduct

000boil:
During a chain reorg, `finished_height` was not being reset after delivering a `ChainReverted` or `ChainReorged` notification. This caused subsequent `ChainCommitted` notifications to be incorrectly skipped when the new chain's tip was at or below the stale `finished_height` from the old chain, leading to missed re-execution and inconsistent ExEx state.

Closes #19755

### DIFF
--- a/crates/exex/exex/src/manager.rs
+++ b/crates/exex/exex/src/manager.rs
@@ -181,6 +181,15 @@ impl<N: NodePrimitives> ExExHandle<N> {
             Ok(()) => {
                 self.next_notification_id = notification_id + 1;
                 self.metrics.notifications_sent_total.increment(1);
+
+                if matches!(
+                    notification,
+                    ExExNotification::ChainReorged { .. } |
+                        ExExNotification::ChainReverted { .. }
+                ) {
+                    self.finished_height = None;
+                }
+
                 Poll::Ready(Ok(()))
             }
             Err(err) => Poll::Ready(Err(err)),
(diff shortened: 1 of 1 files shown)


crates/exex/exex/src/manager.rs
```diff
@@ -181,6 +181,15 @@ impl<N: NodePrimitives> ExExHandle<N> {
...
                 self.next_notification_id = notification_id + 1;
                 self.metrics.notifications_sent_total.increment(1);
+
+                if matches!(
+                    notification,
+                    ExExNotification::ChainReorged { .. } |
+                        ExExNotification::ChainReverted { .. }
+                ) {
+                    self.finished_height = None;
+                }
```
mattsse, line 191: I see, this makes sense,

but should we instead do this here

https://github.com/paradigmxyz/reth/blob/32466fe2232fec8d78ebfea074dcd3ae073d009b/crates/exex/exex/src/manager.rs#L159-L159

and/or check the hash here

https://github.com/paradigmxyz/reth/blob/32466fe2232fec8d78ebfea074dcd3ae073d009b/crates/exex/exex/src/manager.rs#L138-L141

mattsse (changes requested):
this checks out, only have some followup questions

PR #22013, opened 2026-02-10
feat(net): enforce EIP-868 fork ID for discovered peers

gakonst:
## Summary

Enforce EIP-868 ENR fork ID validation for discovered peers to filter out nodes from other networks that pollute the discovery table.

## Motivation

Discovery pollution: nodes accumulate 30-45k tracked peers, mostly from other networks. They burn through outgoing connection slots dialing these garbage peers. This PR adds a mechanism to defer peer admission until fork ID is confirmed via EIP-868.

## Changes

- `DiscoveryEvent::EnrForkId` now carries full `NodeRecord` instead of just `PeerId`, bubbling up the complete record as requested
- `StateAction::DiscoveredEnrForkId` includes `addr: PeerAddr` extracted from the record
- `EnrForkId` handler now calls `add_peer(peer_id, addr, Some(fork_id))` instead of `set_discovered_fork_id` — so peers discovered via ENR get properly added (not just updated)
- Added `enforce_enr_fork_id: bool` to `PeersConfig` — when enabled, `DiscoveredNode` events with `fork_id: None` are skipped; peers only get added once their fork ID is confirmed via EIP-868
- Added tcp_port == 0 guard on the EnrForkId path to match the existing NewNode check

## Testing

```bash
cargo check -p reth-network -p reth-network-api -p reth-network-types
cargo clippy -p reth-network -p reth-network-api -p reth-network-types
```


crates/net/network-api/src/events.rs
```diff
@@ -153,7 +153,7 @@ pub enum DiscoveryEvent {
     /// Discovered a node
     NewNode(DiscoveredEvent),
     /// Retrieved a [`ForkId`] from the peer via ENR request, See <https://eips.ethereum.org/EIPS/eip-868>
-    EnrForkId(PeerId, ForkId),
+    EnrForkId(NodeRecord, ForkId),
```
mattsse: we should update the docs here

crates/net/network/src/swarm.rs
```diff
@@ -250,14 +250,18 @@ impl<N: NetworkPrimitives> Swarm<N> {
...
                     return None
                 }
-                // Insert peer only if no fork id or a valid fork id
-                if fork_id.map_or_else(|| true, |f| self.sessions.is_valid_fork_id(f)) {
+                let enforce = self.state().peers().enforce_enr_fork_id();
+                let allow = match fork_id {
+                    Some(f) => self.sessions.is_valid_fork_id(f),
+                    None => !enforce,
+                };
+                if allow {
```
mattsse, line 264: this seems a bit complicated, can we at least document this properly?

mattsse (changes requested):

mattsse (approved):
lgtm

PR #22014, opened 2026-02-10
feat(net): add direction labels to closed_sessions and pending_session_failures metrics

emmajam:
## Summary

Adds `direction` labels to the `closed_sessions` and `pending_session_failures` counters so operators can distinguish **where** sessions are failing. Follows the same pattern as #22009 (backed_off_peers reason labels).

## New metrics

```
network_closed_sessions{direction="active"}
network_closed_sessions{direction="incoming_pending"}
network_closed_sessions{direction="outgoing_pending"}

network_pending_session_failures{direction="inbound"}
network_pending_session_failures{direction="outbound"}
```

The existing aggregate `network_closed_sessions` and `network_pending_session_failures` counters are preserved for backward compatibility.

## Changes

- **metrics.rs**: Add `ClosedSessionsMetrics` and `PendingSessionFailureMetrics` structs with labeled counters.
- **manager.rs**: Increment the labeled counters alongside the existing aggregate counters at each swarm event callsite.

### DIFF
--- a/crates/net/network/src/manager.rs
+++ b/crates/net/network/src/manager.rs
@@ -875,7 +875,6 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
                     self.swarm.state_mut().peers_mut().on_active_session_gracefully_closed(peer_id);
                     None
                 };
-                self.metrics.closed_sessions.increment(1);
                 self.closed_sessions_metrics.active.increment(1);
                 self.update_active_connection_metrics();
 
@@ -901,7 +900,6 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
                         .state_mut()
                         .peers_mut()
                         .on_incoming_pending_session_dropped(remote_addr, err);
-                    self.metrics.pending_session_failures.increment(1);
                     self.pending_session_failure_metrics.inbound.increment(1);
                     if let Some(reason) = err.as_disconnected() {
                         self.disconnect_metrics.increment(reason);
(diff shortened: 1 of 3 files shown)


crates/net/network/src/manager.rs
```diff
@@ -925,6 +937,7 @@ impl<N: NetworkPrimitives> NetworkManager<N> {
                         err,
                     );
                     self.metrics.pending_session_failures.increment(1);
```
mattsse: do we then still need those? because now this is the same as the sum of the directional ones?

crates/net/network/src/metrics.rs
```diff
@@ -77,6 +77,45 @@ pub struct NetworkMetrics {
...
+    /// Sessions closed from outgoing pending connections.
+    pub outgoing_pending: Counter,
+}
+
+impl Default for ClosedSessionsMetrics {
+    fn default() -> Self {
+        Self {
+            active: metrics::counter!("network_closed_sessions", "direction" => "active"),
+            incoming_pending: metrics::counter!("network_closed_sessions", "direction" => "incoming_pending"),
+            outgoing_pending: metrics::counter!("network_closed_sessions", "direction" => "outgoing_pending"),
```
mattsse, line 90: @DaniPopes is this now supported with new metrics macro?

mattsse (changes requested):

mattsse (approved):
lgtm

PR #22018, opened 2026-02-10
rpc: fix eth_simulateV1 revert error code

Sahil-4555:
### What

Align `eth_simulateV1` revert error handling with the Execution APIs specification.

EVM reverts now return JSON-RPC error code `3` instead of the generic `-32000`.
VM execution failures (e.g. OOG, invalid opcode) continue to return `-32015`.

### Why

The Execution APIs spec defines:
- `code = 3` for EVM `REVERT`
- `-32015` for VM execution errors

Returning `-32000` made it impossible for clients and tooling to reliably
distinguish reverts from generic RPC failures.

refrence: https://github.com/ethereum/execution-apis/pull/748

### DIFF
--- a/.changelog/eth-simulatev1-revert-error-code.md
+++ b/.changelog/eth-simulatev1-revert-error-code.md
@@ -1,0 +1,10 @@
+---
+reth-rpc-eth-types: minor
+reth-rpc-e2e-tests: patch
+---
+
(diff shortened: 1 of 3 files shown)


mattsse (commented):
I'd like this, lets wait until the pr clears

maybe we can further simplify revert handling then

PR #22020, opened 2026-02-10
feat(rpc): implement debug_getBlockAccessList (EIP-7928)

joohhnnn:
Spotted this while reading the debug RPC code — the trait and alloy-eip7928 types are all there but the handler just bails with "unimplemented".

The implementation replays the block tx-by-tx on the parent state (same approach as execution witness), diffs each transaction and accumulates per-account balance/nonce/code/storage changes with their tx index.

Ran into one thing worth noting: revm filters out read-only storage slots in its state diff (the is_changed() gate in finalize), so storage_reads would always be empty if you only look at the diff. Plugged in an AccessListInspector per tx to grab SLOAD-only accesses separately.

No new deps needed.

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -681,10 +681,7 @@ where
                                     .as_ref()
                                     .map(|i| i.balance)
                                     .unwrap_or_default(),
-                                last_nonce: pre_info
-                                    .as_ref()
-                                    .map(|i| i.nonce)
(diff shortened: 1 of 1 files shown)


mattsse (commented):
thanks for this but this will soon work out of the box once we bump to new revm, see https://github.com/paradigmxyz/reth/pull/21245

joohhnnn:
> thanks for this but this will soon work out of the box once we bump to new revm, see #21245

Nice!

PR #22021, opened 2026-02-10
feat(trie): Combine extension and branch nodes in output from proof v2

mediocregopher:
This PR introduces a new `TrieNodeV2` type which does not treat extension nodes as a separate node type from branches, but rather tracks a short key on branches just like leaves have, letting the extension node be a serialization detail.

This allows us to:
* Remove the extension node type in the sparse trie, removing a hop during traversal
* Know the branch node's parent path during revealing in the sparse trie, allowing for better connectivity checks
* Traverse the sparse trie inwards from the leaves, useful for both removing leaves and pruning


crates/trie/common/src/trie_node_v2.rs
```diff
@@ -0,0 +1,112 @@
...
+impl BranchNodeV2 {
+    /// Creates a new branch node with the given short key, stack, and state mask.
+    pub const fn new(key: Nibbles, stack: Vec<RlpNode>, state_mask: TrieMask) -> Self {
+        Self { key, stack, state_mask }
+    }
+
+    /// Converts this node into its RLP node representation, encoding into the provided buffer.
+    ///
+    /// Returns the [`RlpNode`] pointing to the encoded data.
+    pub fn encode_rlp(&self, buf: &mut Vec<u8>) {
```
mediocregopher: The `Encodable` trait requires the use of a BufMut here, which doesn't allow for clearing/overwriting. In order to support that this implementation would essentially have to allocate a new Vec on-the-fly every time, so I introduced a new method which accepts `&mut Vec` instead.

crates/trie/trie/src/proof/mod.rs
```diff
@@ -40,6 +40,8 @@ pub struct Proof<T, H> {
     prefix_sets: TriePrefixSetsMut,
     /// Flag indicating whether to include branch node masks in the proof.
     collect_branch_node_masks: bool,
+    /// Added and removed keys for proof retention.
+    added_removed_keys: Option<K>,
```
mediocregopher, line 44: Nothing outside of the proof v2 proptests is using this.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -540,19 +541,12 @@ where
             self.branch_path.len() - branch.ext_len as usize,
         );
 
-        // Wrap the `BranchNode` so it can be pushed onto the child stack.
-        let mut branch_as_child = ProofTrieBranchChild::Branch {
-            node: BranchNode::new(rlp_nodes_buf, branch.state_mask),
+        // Wrap the `BranchNodeV2` so it can be pushed onto the child stack.
```
mediocregopher, line 554: This is the main logical change here. Since proof v2 was already representing branches internally as having a short key the overall change is pretty small.

klkvr (approved):
nice

crates/trie/common/src/nibbles.rs
```diff
@@ -1,7 +1,28 @@
...
+use core::cmp::Ordering;
 use derive_more::Deref;
 pub use nybbles::Nibbles;
 
+/// Compares two [`Nibbles`] in depth-first order.
+///
+/// In depth-first ordering:
+/// - Descendants come before their ancestors (children before parents)
+/// - Siblings are ordered lexicographically
+pub fn depth_first_cmp(a: &Nibbles, b: &Nibbles) -> Ordering {
```
mediocregopher: Just FYI there's already an implementation [here](https://github.com/paradigmxyz/reth/blob/217a337d8c5330407b7915370c046976a39eb71f/crates/trie/trie/src/trie_cursor/depth_first.rs#L22), maybe better to keep that one? Or delete that one and use this one instead, this might be a more sensible place

crates/trie/sparse/src/state.rs
```diff
@@ -274,83 +271,7 @@ where
...
-                    any_err = result.map(|_| ());
-                }
-
-                // Keep buffers for deferred dropping
-                self.deferred_drops.proof_nodes_bufs.extend(bufs);
-            }
-
-            any_err
-        }
+        self.reveal_decoded_multiproof_v2(multiproof.into())
```
mediocregopher, line 269: nice

crates/trie/sparse/src/parallel.rs
```diff
@@ -476,7 +500,7 @@ impl SparseTrie for ParallelSparseTrie {
                             };
                             if let Some(RevealedNode { node, tree_mask, hash_mask }) = revealed_node
                             {
-                                let decoded = match TrieNode::decode(&mut &node[..]) {
+                                let decoded = match TrieNodeV2::decode(&mut &node[..]) {
```
mediocregopher: fwiw we don't actually hit this codepath anymore, once we get rid of the old sparse trie task and multiproof task code this can be deleted completely, and we probably don't need `decode`

crates/trie/sparse/src/parallel.rs
```diff
@@ -674,6 +699,15 @@ impl SparseTrie for ParallelSparseTrie {
             match Self::find_next_to_leaf(&curr_path, curr_node, full_path) {
                 FindNextToLeafOutcome::NotFound => return Ok(()), // leaf isn't in the trie
                 FindNextToLeafOutcome::BlindedNode(hash) => {
+                    // If we hit a blinded node that is a child of an extension node, we need to
```
mediocregopher: How would we get into a situation where the extension is revealed but the branch isn't? If proof v2 is always revealing both simultaneously now?

klkvr: i think main such scenario is branch node being replaced with a hash during pruning

crates/trie/common/src/trie_node_v2.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Variant representing a [`BranchNodeV2`].
+    Branch(BranchNodeV2),
+    /// Variant representing a [`LeafNode`].
+    Leaf(LeafNode),
+    /// Variant representing an [`ExtensionNode`].
+    ///
+    /// This will only be used for extension nodes for which child is not inlined. This variant
+    /// will never be produced by proof workers that will always reveal a full path to a requested
+    /// leaf.
+    Extension(ExtensionNode),
```
mediocregopher, line 110: Once we can remove the inline-reveal code I think we can also remove this variant, since that's the only place using the Decode afaict

crates/trie/common/src/trie_node_v2.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub stack: Vec<RlpNode>,
+    /// The bitmask indicating the presence of children at the respective nibble positions
+    pub state_mask: TrieMask,
+    /// The hash of the branch node.
+    ///
+    /// This is [`None`] in 2 cases:
+    ///   - When the `key` is empty, i.e this branch is not coupled with a child extension node.
+    ///   - When the branch encoding is short enough that it can be stored in the database without
+    ///     hashing.
+    pub hash: Option<B256>,
```
mediocregopher: let's make this branch_rlp_node, rather than hash, so it's only unavailable if `key` is empty

crates/trie/common/src/trie_node_v2.rs
```diff
@@ -0,0 +1,239 @@
...
+        let branch_ref = BranchNodeRef::new(&self.stack, self.state_mask);
+        branch_ref.encode(buf);
+
+        // If key is empty then there is no parent extension, the encoded branch is the result
+        if self.key.is_empty() {
+            return
+        }
+
+        // Convert branch to `RlpNode`. This will hash it if it's >32 bytes encoded.
+        let branch_rlp_node = RlpNode::from_rlp(buf);
```
mediocregopher: If the RlpNode is already cached on the BranchNodeV2 we should use it, rather than re-encoding.

crates/trie/sparse/src/state.rs
```diff
@@ -647,178 +515,15 @@ where
...
-        // Keep buffer for deferred dropping
-        bufs.push(nodes);
-
-        Ok(metric_values)
-    }
-
     /// Reveal state witness with the given state root.
     /// The state witness is expected to be a map of `keccak(rlp(node)): rlp(node).`
     /// NOTE: This method does not extensively validate the witness.
     pub fn reveal_witness(
```
mediocregopher: This can be deleted completely now https://github.com/paradigmxyz/reth/pull/22115

crates/trie/sparse/src/parallel.rs
```diff
@@ -2165,38 +2258,64 @@ impl ParallelSparseTrie {
...
                         child_path.push_unchecked(idx);
                         self.lower_subtrie_for_path_mut(&child_path)
                             .expect("child_path must have a lower subtrie")
                             .reveal_node_or_hash(child_path, &branch.stack[stack_ptr])?;
-                        stack_ptr += 1;
                     }
                 }
             }
-            TrieNode::Extension(ext) => {
+            TrieNodeV2::Extension(ext) => {
```
mediocregopher, line 2077: Let's panic here, we should never see an Extension node from the workers

crates/trie/sparse/src/parallel.rs
```diff
@@ -2830,51 +3046,71 @@ impl SparseSubtrie {
...
+                self.reveal_branch(
+                    branch_path,
+                    branch.state_mask,
+                    &branch.stack,
+                    masks,
+                    branch.hash,
+                )?;
             }
-            TrieNode::Extension(ext) => match self.nodes.entry(path) {
+            TrieNodeV2::Extension(ext) => match self.nodes.entry(path) {
```
mediocregopher, line 2784: Another Extension match we should just panic on, we should never see it

crates/trie/sparse/src/parallel.rs
```diff
@@ -2998,7 +3234,7 @@ impl SparseSubtrie {
             return Ok(())
         }
 
-        self.reveal_node(path, &TrieNode::decode(&mut &child[..])?, None)?;
+        self.reveal_node(path, &TrieNodeV2::decode(&mut &child[..])?, None)?;
```
mediocregopher, line 2908: Just a note here: it's not possible for decode to return an Extension here: `child` is from an RlpNode, so must be <32B here. `TrieNodeV2::decode` only returns Extension if the child is >=32B, but for that to be the case the overall extension would be >32B and therefore the RlpNode would be a hash. So an extension here must have an embedded branch (which _can_ happen), but we return `TrieNodeV2::Branch` in that case.

PR #22022, opened 2026-02-10
ci: skip scheduled workflows on forks

joohhnnn:
My inbox was getting bombed by CI failure emails from fork scheduled runs😧 — sync tests, docker nightly builds, kurtosis.....  On top of that, the cargo update workflow was auto-creating PRs on forks which makes no sense.

Noticed `hive.yml` already had a `github.repository == 'paradigmxyz/reth'` guard, so just applied the same pattern to the remaining 7 scheduled workflows.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -15,6 +15,7 @@ permissions:
 
 jobs:
   update:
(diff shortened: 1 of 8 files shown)


mattsse (approved):
makes sense, ty

PR #22024, opened 2026-02-10
feat: implement variants for BAL devp2p variants

### LINKED ISSUE #22015: Implement variants for BAL devp2p variants
### Describe the feature

for this upcoming change we need additional message variants for the eth71 protocol (we can assume this will be eth71) see

https://github.com/ethereum/devp2p/pull/264/changes
we need to implment this, similar to receipts70 and add the necessary versioned checks and also do the message variant conversions properly,
this can also be implemented in the ethrequesthandler in the networking stuff, but for now just return empty responses

@0xKarl98 

### Additional context

_No response_

0xKarl98:
This closes #22015 

### Summary 
Implement eth/71 BAL devp2p variants and versioned request handling for reth networking, including temporary empty BAL responses as requested by the issue.

### Motivation 
Upcoming BAL support requires new eth protocol variants and strict version-aware behavior. Without explicit gating, BAL request IDs can conflict with snap multiplexing on lower negotiated eth versions (notably eth/70). This PR adds the required variants, conversions, and guards to keep decoding/dispatch correct across versions.

cc @mattsse


crates/net/eth-wire-types/src/block_access_lists.rs
```diff
@@ -0,0 +1,27 @@
...
+);
+
+/// Response for [`GetBlockAccessLists`] containing one BAL per requested block hash.
+#[derive(Clone, Debug, PartialEq, Eq, RlpEncodableWrapper, RlpDecodableWrapper, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[add_arbitrary_tests(rlp)]
+pub struct BlockAccessLists(
+    /// The requested block access lists. Unavailable entries are represented by empty BALs.
+    pub Vec<BlockAccessList>,
```
mattsse: I think these should just be Bytes, right?

mattsse (commented):
cool, this looks already pretty good

crates/net/eth-wire-types/src/block_access_lists.rs
0xKarl98: You are so quick , yeah it should be

crates/net/eth-wire-types/src/version.rs
```diff
@@ -29,14 +29,17 @@ pub enum EthVersion {
...
 
 impl EthVersion {
     /// The latest known eth version
-    pub const LATEST: Self = Self::Eth69;
+    pub const LATEST: Self = Self::Eth70;
 
     /// All known eth versions
-    pub const ALL_VERSIONS: &'static [Self] = &[Self::Eth69, Self::Eth68, Self::Eth67, Self::Eth66];
+    pub const ALL_VERSIONS: &'static [Self] =
+        &[Self::Eth70, Self::Eth69, Self::Eth68, Self::Eth67, Self::Eth66];
```
mattsse: this should remain unchanged for now, because eth70 is still being tested

mattsse (changes requested):
nice, all of this looks great, smol pedantic nit

mattsse (approved):

PR #22025, opened 2026-02-10
fix(stages): use actual highest block number instead of entry count in bodies ensure_consistency

MozirDmitriy:
`ensure_consistency` in the bodies stage used `entries()` to get the highest database block number for passing to `prune_transactions`. The problem is that `entries()` returns the total row count, not the last key — for blocks `[0..=N]` it returns `N+1` instead of `N`. This caused the static file block range to be set one block higher than what actually exists in the database.

Replace `entries()` with `cursor_read().last()` to retrieve the real highest block number. Add a regression test covering this scenario.


emmajam:

PR #22030, opened 2026-02-10
fix(txpool): correct swapped args in blob_tx_priority calls

emmajam:
## Summary

The calls to `blob_tx_priority` in `BlobTransaction::new` and `update_priority` had the arguments swapped in pairs: network fees (`blob_fee`, `base_fee`) were passed where the transaction's fee caps should be, and vice versa.

This inverted the sign of `fee_delta`, causing:
- **Low-fee-cap** blob transactions to appear as **highest priority** (0)
- **High-fee-cap** blob transactions to get **negative priority** (evicted first)

This enables blob sub-pool mempool poisoning by filling it with non-executable transactions that never get evicted.

## Fix

Swap each pair of arguments in both call sites to match the function signature `(blob_fee_cap, blob_fee, max_priority_fee, base_fee)`:

```rust
// Before (buggy):
blob_tx_priority(
    pending_fees.blob_fee,                              // network fee in cap slot
    transaction.max_fee_per_blob_gas().unwrap_or_default(), // tx cap in network slot
    pending_fees.base_fee as u128,                      // network fee in cap slot
    transaction.max_fee_per_gas(),                      // tx cap in network slot
);

// After (fixed):
blob_tx_priority(
    transaction.max_fee_per_blob_gas().unwrap_or_default(), // tx cap ✓
    pending_fees.blob_fee,                              // network fee ✓
    transaction.max_fee_per_gas(),                      // tx cap ✓
    pending_fees.base_fee as u128,                      // network fee ✓
);
```

Updated `test_blob_ordering` expected ordering to match correct priority computation.

Closes RETH-396


mattsse (approved):
I see, 

I think this checks out

yongkangc (approved):

PR #22032, opened 2026-02-10
feat: add StorageSlotKey enum and thread through StorageChangeSetReader

joshieDo:
## Summary

Introduces a `StorageSlotKey` enum (`Plain(B256)` / `Hashed(B256)`) that carries provenance with storage slot keys, eliminating scattered `use_hashed_state` boolean checks across consumers.

### The problem

In v2 mode, `StorageChangeSets` store keccak256-hashed slot keys. Every consumer that reads changesets had to independently check `use_hashed_state` and conditionally hash:

```rust
// Before: scattered boolean checks, easy to double-hash
let hashed = if use_hashed_state { key } else { keccak256(key) };
```

### The solution

`StorageSlotKey::to_hashed()` is **idempotent** — it keccak256-hashes `Plain`, passes through `Hashed`. Double-hashing is impossible:

```rust
// After: tagged at the source, safe everywhere
let hashed = entry.key.to_hashed(); // always correct
```

### New types (`reth-primitives-traits`)
- **`StorageSlotKey`** enum with `Plain(B256)` and `Hashed(B256)` variants
- `to_hashed()` — idempotent keccak256 (no-op on `Hashed`, hashes `Plain`)
- `to_changeset_key(use_hashed_state)` — for REVM→changeset boundaries (debug-asserts `Plain`)
- `from_u256()`, `from_raw()`, `as_b256()` helpers

### Trait change (`reth-storage-api`)
- **`ChangesetEntry`** struct with `key: StorageSlotKey` + `value: U256`
- **`StorageChangeSetReader`** trait methods now return `ChangesetEntry` instead of `StorageEntry`
- Tagging happens at the source — implementations set `Plain`/`Hashed` based on storage mode

### Hardened key provenance
- **MDBX reads**: `StorageSlotKey::plain()` — MDBX changesets always store plain keys
- **Static file reads**: `StorageSlotKey::hashed()` — static file changesets always store hashed keys
- **Static file provider**: removed `AtomicBool use_hashed_state` and `use_hashed_state` from `StaticFileWriteCtx` — static file changesets are always v2/hashed, no flag needed
- **`insert_storage_for_hashing`**: added `debug_assert!(!use_hashed_state)` to guard against v2 misuse

### Removed `use_hashed_state` plumbing
These no longer need the boolean — the tag is in the entry:
- `DatabaseHashedPostState::from_reverts`
- `load_prefix_sets_with_provider`
- `hashed_storage_from_reverts_with_provider`
- `StaticFileProviderInner::use_hashed_state` / `set_use_hashed_state`
- `StaticFileWriteCtx::use_hashed_state`
- `ProviderFactory::set_storage_settings_cache` no longer propagates to static file provider

### DRY-up
- Consolidated 3x repeated `Plain→Hashed` tagging pattern in `ConsistentProvider` `StorageChangeSetReader` impl

## Encoding
**No on-disk format changes.** `StorageEntry` (the DB row type) keeps `key: B256`. The enum is purely in-memory — readers tag after decode, the tag is never serialized.

## Testing
All 8 hashed state tests pass. All 21 trie-db tests pass.


crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -395,6 +395,8 @@ pub struct StaticFileProviderInner<N> {
     _lock_file: Option<StorageLock>,
     /// Genesis block number, default is 0;
     genesis_block_number: u64,
+    /// Whether storage keys in changesets are keccak256-hashed.
+    use_hashed_state: std::sync::atomic::AtomicBool,
```
joshieDo: why is this atomic

Rjected (approved):
lgtm

PR #22038, opened 2026-02-10
feat(reth-bench): accept bare integers as milliseconds for --wait-time

gakonst:
## Summary

`--wait-time 400` now works the same as `--wait-time 400ms`. Bare integers are treated as milliseconds.

## Changes

- Added `parse_duration` helper in `helpers.rs` that falls back to parsing bare integers as ms when `humantime::parse_duration` fails
- Replaced `humantime::parse_duration` imports in `new_payload_fcu.rs` and `replay_payloads.rs`
- Updated doc comments and README

## Testing

`cargo test -p reth-bench` — all 12 tests pass, including 3 new `parse_duration` tests.


mattsse (approved):
this makes sense, i had the same issue myself

PR #22042, opened 2026-02-10
chore(storage): rm storage settings, use only one

Rjected:


crates/cli/commands/src/db/settings.rs
```diff
@@ -39,38 +39,11 @@ enum Subcommands {
...
-    AccountHistory {
-        #[clap(action(ArgAction::Set))]
-        value: bool,
-    },
-    /// Store storage changesets in static files instead of the database
-    StorageChangesets {
+    /// Enable or disable v2 storage layout
+    ///
+    /// When enabled, uses static files for receipts/senders/changesets and RocksDB for
+    /// history indices. When disabled, uses v1/legacy layout (everything in MDBX).
```
joshieDo:
```suggestion
    /// When enabled, uses static files for receipts/senders/changesets and RocksDB for
    /// history indices and transaction hashes. When disabled, uses v1/legacy layout (everything in MDBX).
```

crates/node/core/src/args/rocksdb.rs
```diff
@@ -1,160 +1,8 @@
...
-        let args = RocksDbArgs {
-            all: true,
-            tx_hash: None,
-            storages_history: None,
-            account_history: Some(false),
-        };
-        assert_eq!(args.validate(), Err(RocksDbArgsError::ConflictingFlags("account-history")));
-    }
-}
+pub struct RocksDbArgs {}
```
joshieDo: can we remove this

crates/storage/db-api/src/models/metadata.rs
```diff
@@ -62,159 +48,133 @@ impl StorageSettings {
...
+    pub const fn transaction_senders_in_static_files(&self) -> bool {
+        self.storage_v2
     }
 
-    /// Sets the `transaction_hash_numbers_in_rocksdb` flag to the provided value.
-    pub const fn with_transaction_hash_numbers_in_rocksdb(mut self, value: bool) -> Self {
-        self.transaction_hash_numbers_in_rocksdb = value;
-        self
+    /// Whether account changesets are stored in static files.
+    pub const fn account_changesets_in_static_files(&self) -> bool {
```
joshieDo: should we remove all these methods

Rjected: removed

crates/node/core/src/args/rocksdb.rs
Rjected: removed

crates/node/core/src/node_config.rs
```diff
@@ -369,49 +363,17 @@ impl<ChainSpec> NodeConfig<ChainSpec> {
...
-
-        if s.use_hashed_state {
-            s = s.with_storage_changesets_in_static_files(true);
-        }
-        if s.storage_changesets_in_static_files {
-            s = s.with_use_hashed_state(true);
+        if self.storage.v2 {
+            StorageSettings::v2()
+        } else {
+            StorageSettings::base()
```
joshieDo, line 375: could rename it to v1

joshieDo (approved):

crates/node/core/src/node_config.rs
Rjected: we should but let's remove the `edge` feature first from db-api, will do a followup for this

PR #22045, opened 2026-02-10
chore(evm): remove unnecessary unused_imports suppression

splinter012:
Consolidates std-specific imports under #[cfg(feature = "std")] instead of using #[allow(unused_imports)] suppression.


mattsse (commented):
can keep as is

PR #22049, opened 2026-02-11
refactor: remove TypesAnd1-5 staging types from ProviderFactoryBuilder

mattsse:
Remove the unnecessary `TypesAnd1`-`TypesAnd5` generic staging types and the `db()` method from `ProviderFactoryBuilder`. These types were only used internally to chain the builder steps in `open_read_only`, which now calls `ProviderFactory::new` directly.

-186 lines

### DIFF
--- a/.changelog/quiet-frogs-whisper.md
+++ b/.changelog/quiet-frogs-whisper.md
@@ -1,0 +1,5 @@
+---
+reth-provider: patch
+---
+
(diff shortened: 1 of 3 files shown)


DaniPopes (approved):
nuke it

shekhirin (approved):

joshieDo (approved):

PR #22051, opened 2026-02-11
perf: use dedicated trie rayon pool for proof workers

DaniPopes:
Migrate trie proof workers and storage root computation from tokio::spawn_blocking to the dedicated trie rayon thread pool, removing the tokio blocking thread overhead and unnecessary nesting of spawn_blocking calls for worker spawning.


yongkangc (approved):
lgtm, we should bench

shekhirin (approved):

PR #22056, opened 2026-02-11
chore: remove unused hive Dockerfile

merge-script:
## Summary
- Remove `.github/scripts/hive/Dockerfile` which is no longer referenced anywhere in the codebase
- All hive builds have migrated to `depot/bake-action` with `docker-bake.hcl` targets (`hive-stable`, `hive-edge`) pointing to `Dockerfile.depot`


mattsse (commented):
can keep no idea

PR #22058, opened 2026-02-11
feat(reth-bench): add gas-range measurement filter for small blocks benchmarking

gakonst:
## Summary

Adds `--measure-gas-min` and `--measure-gas-max` flags to `reth-bench new-payload-fcu` and `reth-bench-compare` for focused benchmarking of specific gas ranges.

## Motivation

We need to benchmark reth's performance specifically on small blocks (0-20M gas) to identify tuning opportunities for typical mainnet block sizes. The existing `new-payload-fcu` benchmark replays all blocks indiscriminately, making it hard to isolate small block performance.

## Usage

```bash
# Benchmark only small blocks (0-20M gas)
reth-bench new-payload-fcu --measure-gas-max 20M ...

# Benchmark medium blocks (10-30M gas)  
reth-bench new-payload-fcu --measure-gas-min 10M --measure-gas-max 30M ...

# Via reth-bench-compare
reth-bench-compare --measure-gas-max 20M --baseline-ref main --feature-ref feat/optimize ...
```

### DIFF
--- a/.changelog/evil-clams-read.md
+++ b/.changelog/evil-clams-read.md
@@ -1,0 +1,6 @@
+---
+reth-bench-compare: minor
+reth-bench: minor
+---
+
+Added gas range filtering for benchmark measurements. Blocks with gas usage outside the specified `--measure-gas-min` and `--measure-gas-max` range are now executed but excluded from timing statistics, allowing more focused performance analysis of blocks within specific gas usage ranges.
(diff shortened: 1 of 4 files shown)


gakonst:
Closing per discussion — we'll filter the output with a Python script instead of adding a reth-bench flag. If this becomes a common need, we can revisit adding it as a feature.

PR #22059, opened 2026-02-11
perf(engine): skip transaction prewarming for small blocks

gakonst:
## Summary
Skip transaction prewarming for blocks under 20M gas to reduce `new_payload` latency on small blocks.

## Motivation
Pandaops Lab data over the last 10k mainnet blocks shows reth's winrate drops from **88% (40-50M gas blocks)** to **49% (0-10M gas blocks)**. The root cause is the fixed overhead of spawning prewarm workers — each worker builds state providers, creates EVM instances, and wraps precompiles — which exceeds the execution time of small blocks.

~23% of recent mainnet blocks fall under 20M gas where this overhead dominates.

## Changes
- Added `SMALL_BLOCK_GAS_THRESHOLD = 20_000_000` constant
- Added `gas_used` field to `ExecutionEnv`, plumbed from payload header via `BlockOrPayload::gas_used()`
- Modified `spawn_caching_with` to skip transaction prewarming when `gas_used > 0 && gas_used < 20M`
- State root task and cross-block caching remain active (only prewarming is skipped)

## Testing
- `cargo check -p reth-engine-tree` — clean
- Needs benchmarking on real blocks to validate latency improvement

Prompted by: yk

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -94,6 +94,17 @@ pub const SPARSE_TRIE_MAX_NODES_SHRINK_CAPACITY: usize = 1_000_000;
 /// 144MB.
 pub const SPARSE_TRIE_MAX_VALUES_SHRINK_CAPACITY: usize = 1_000_000;
 
+/// Gas used threshold below which prewarming is skipped for small blocks.
+///
+/// Based on analysis of 100 mainnet blocks (starting at block 24,425,210), reth's `new_payload`
+/// winrate drops from 88% on 40-50M gas blocks to 49% on 0-10M gas blocks. The root cause is the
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -431,9 +441,10 @@ where
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
     {
-        if self.disable_transaction_prewarming {
-            // if no transactions should be executed we clear them but still spawn the task for
-            // caching updates
+        let skip_prewarm = self.disable_transaction_prewarming ||
```
yongkangc, line 445: can we have a good comment here like previously? 

```
            // if no transactions should be executed we clear them but still spawn the task for
            // caching updates
```

shekhirin (approved):

DaniPopes (approved):

PR #22063, opened 2026-02-11
fix: don't drop node

DaniPopes:
Fixes the RPC server immediately stopping after startup.

In #22057, `main.rs` changed from `let NodeHandle { node, node_exit_future }` to `let NodeHandle { node_exit_future, .. }`, which drops the `node` field immediately after launch. This drops the only `RpcServerHandle`, whose inner `ServerHandle(Arc<watch::Sender<()>>)` closes the jsonrpsee stop channel, causing the server accept loop to exit before any connections can be accepted.

The fix keeps the full `NodeHandle` alive by using `handle.wait_for_node_exit()` instead of destructuring.


shekhirin (approved):
Clean.

PR #22066, opened 2026-02-11
perf(engine): add PrewarmMode::Skipped to avoid spawning idle workers

yongkangc:
**Summary**
When prewarming is disabled, workers were still spawned via a dead-channel hack (`mpsc::channel().1`) only to exit immediately. This replaces that with a `PrewarmMode::Skipped` variant that sends `FinishedTxExecution` directly — no workers spawned, cache saving still works.

**Changes**
- Add `PrewarmMode::Skipped` — skips worker spawning, signals completion immediately
- Remove dead-channel hack and `mut` on `transactions` param
- Remove redundant `drop(actions_tx)`

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -464,6 +464,8 @@ where
                     PrewarmMode::Skipped
                 } else if let Some(bal) = bal {
                     PrewarmMode::BlockAccessList(bal)
+                } else if skip_prewarm {
+                    PrewarmMode::Skipped
                 } else {
                     PrewarmMode::Transactions(transactions)
                 };
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -422,7 +422,7 @@ where
     fn spawn_caching_with<P>(
         &self,
         env: ExecutionEnv<Evm>,
-        mut transactions: mpsc::Receiver<impl ExecutableTxFor<Evm> + Clone + Send + 'static>,
+        transactions: mpsc::Receiver<impl ExecutableTxFor<Evm> + Clone + Send + 'static>,
         provider_builder: StateProviderBuilder<N, P>,
         to_multi_proof: Option<CrossbeamSender<MultiProofMessage>>,
         bal: Option<Arc<BlockAccessList>>,
@@ -431,11 +431,7 @@ where
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
     {
-        if self.disable_transaction_prewarming {
-            // if no transactions should be executed we clear them but still spawn the task for
-            // caching updates
-            transactions = mpsc::channel().1;
-        }
+        let skip_prewarm = self.disable_transaction_prewarming;
 
         let saved_cache = self.disable_state_cache.not().then(|| self.cache_for(env.parent_hash));
 
@@ -464,7 +460,9 @@ where
         {
             let to_prewarm_task = to_prewarm_task.clone();
             self.executor.spawn_blocking(move || {
-                let mode = if let Some(bal) = bal {
+                let mode = if skip_prewarm {
+                    PrewarmMode::Skipped
+                } else if let Some(bal) = bal {
                     PrewarmMode::BlockAccessList(bal)
                 } else {
                     PrewarmMode::Transactions(transactions)
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -464,8 +464,6 @@ where
                     PrewarmMode::Skipped
                 } else if let Some(bal) = bal {
                     PrewarmMode::BlockAccessList(bal)
-                } else if skip_prewarm {
-                    PrewarmMode::Skipped
                 } else {
                     PrewarmMode::Transactions(transactions)
                 };
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -422,7 +422,6 @@ where
             PrewarmMode::Skipped => {
                 let _ = actions_tx
                     .send(PrewarmTaskEvent::FinishedTxExecution { executed_transactions: 0 });
-                drop(actions_tx);
             }
         }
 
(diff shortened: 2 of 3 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -431,11 +431,7 @@ where
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
     {
-        if self.disable_transaction_prewarming {
-            // if no transactions should be executed we clear them but still spawn the task for
-            // caching updates
-            transactions = mpsc::channel().1;
-        }
```
yongkangc, line 438: mpsc::channel().1 creates a channel and drops the sender — making a receiver that instantly returns Err on recv(). It's a hack to make workers spawn and immediately exit.

Before: spawn max_concurrency workers → each hits dead receiver → exits immediately → save_cache() still runs.

After (Skipped): send FinishedTxExecution directly, spawn zero workers → save_cache() still runs.

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -416,6 +418,10 @@ where
             PrewarmMode::BlockAccessList(bal) => {
                 self.run_bal_prewarm(bal, actions_tx);
             }
+            PrewarmMode::Skipped => {
+                let _ = actions_tx
+                    .send(PrewarmTaskEvent::FinishedTxExecution { executed_transactions: 0 });
```
yongkangc, line 424: this is so that it doesnt stall

gakonst (dismissed):
LGTM — clean fix. Skipped variant avoids spawning idle workers, correctly sends FinishedTxExecution so the event loop exits cleanly, cache path unaffected.

yongkangc:
@DaniPopes i see good pt, u mean like we still spawn `PrewarmCacheTask itself` ?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -431,13 +435,19 @@ where
     where
         P: BlockReader + StateProviderFactory + StateReader + Clone + 'static,
     {
-        if self.disable_transaction_prewarming {
-            // if no transactions should be executed we clear them but still spawn the task for
-            // caching updates
-            transactions = mpsc::channel().1;
+        let saved_cache = self.disable_state_cache.not().then(|| self.cache_for(env.parent_hash));
```
yongkangc: why did we do this?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -813,9 +850,35 @@ impl<R: Send + Sync + 'static> CacheTaskHandle<R> {
...
+            if new_cache.cache().insert_state(&execution_outcome.state).is_err() {
+                *cached = None;
+                debug!(target: "engine::caching", "cleared execution cache on inline commit error");
+                return;
+            }
+
+            new_cache.update_metrics();
+            *cached = Some(new_cache);
+        });
+    }
```
yongkangc: this seems like duplicated code? where are we referenceing? can we add comments for this?

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: This is now gone entirely — we no longer create a dead channel or spawn idle workers. When `disable_transaction_prewarming && bal.is_none()`, `spawn_caching_with` returns early with `to_prewarm_task: None` and no background task is spawned at all.

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -431,13 +435,19 @@ where
...
+                saved_cache,
+                to_prewarm_task: None,
+                execution_cache: self.execution_cache.clone(),
+                block_hash: env.hash,
+                pending_outcome: None,
+            };
         }
 
-        let saved_cache = self.disable_state_cache.not().then(|| self.cache_for(env.parent_hash));
+        let block_hash = env.hash;
```
yongkangc: why did we extract his out?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
yongkangc: Removed — `PrewarmMode::Skipped` no longer exists. Since no background task is spawned when prewarming is disabled, there is no event loop to stall.

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: This moved up because the early return needs `saved_cache`. When `disable_transaction_prewarming && bal.is_none()`, we return a `CacheTaskHandle` with `to_prewarm_task: None` immediately — no `PrewarmContext`, no `PrewarmCacheTask`, no `spawn_blocking`. The cache is saved inline via `commit_cache()` after block validation succeeds.

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: Yes, this mirrors the core logic of `PrewarmCacheTask::save_cache` in `prewarm.rs` (L237-268). The duplication is intentional — `save_cache` runs in a background task and coordinates via `valid_block_rx` channel, while `commit_cache` runs inline after validation succeeds (no channel needed). The doc comment already references the source: "This performs the same cache update that `PrewarmCacheTask::save_cache` would do in the background task path."

Extracting a shared helper would require threading `PayloadExecutionCache` through `PrewarmCacheTask` differently and the two paths have different ownership semantics (`self` consuming vs `&mut self`), so I kept them separate for clarity.

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -760,13 +776,21 @@ impl<Tx, Err, R: Send + Sync + 'static> PayloadHandle<Tx, Err, R> {
     /// path without cloning the expensive `BundleState`.
     ///
     /// Returns a sender for the channel that should be notified on block validation success.
+    /// Returns `None` when no background task was spawned (prewarming disabled).
```
yongkangc: waht does this mean? if no perwarming, we also dont cache?

yongkangc: We still cache — the `None` just means there is no background task channel to signal.

Two paths:
- **Background task** (prewarming enabled): `terminate_caching` returns `Some(valid_block_tx)` → background task waits on that channel → validator sends `valid_block_tx.send(())` on success → background task commits cache
- **Inline** (prewarming disabled): `terminate_caching` returns `None` and stashes the outcome in `pending_outcome` → validator calls `handle.commit_cache()` on success → cache committed inline on the same thread

The cache is always updated after a valid block, regardless of whether prewarming is enabled.

crates/engine/tree/src/tree/payload_processor/mod.rs
yongkangc: Because `env` is moved into `PrewarmContext` on L453, so we need to capture `env.hash` before that move. The early-return path above (L440-448) can access `env.hash` directly since it returns before the move.

DaniPopes (approved):

PR #22070, opened 2026-02-11
feat(network): add inbound / outbound scopes for disconnect reasons

Rjected:
This lets us determine why we disconnect from peers vs why they disconnect from us. Keeps the old cumulative metrics as well


mattsse (approved):
lgtm, pending @emmajam

emmajam (approved):
send it

PR #22072, opened 2026-02-11
feat: make PayloadAttributesBuilder::build async

gakonst:
## Summary
Make `PayloadAttributesBuilder::build` return a boxed future (`Pin<Box<dyn Future<Output = Attributes> + Send + 'a>>`) instead of returning attributes synchronously.

## Changes
- `crates/payload/primitives/src/traits.rs`: trait method now returns `Pin<Box<dyn Future + Send + 'a>>`, updated blanket impls for closures, `Either`, and `Box<dyn ...>`
- `crates/engine/local/src/payload.rs`: `LocalPayloadAttributesBuilder` impls wrap body in `Box::pin(async move { ... })`, OP impl uses UFCS to call the Eth build
- `crates/engine/local/src/miner.rs`: `.await` the build call
- `crates/node/builder/src/launch/debug.rs`: mapping closure uses `Arc` for cloneability and returns an async block

## Testing
Targeted crate checks pass: `reth-payload-primitives`, `reth-engine-local --all-features`, `reth-node-builder`, `reth-node-ethereum`.

### DIFF
--- a/crates/payload/primitives/src/traits.rs
+++ b/crates/payload/primitives/src/traits.rs
@@ -8,7 +8,7 @@ use alloy_eips::{
 };
 use alloy_primitives::{Address, B256, U256};
 use alloy_rpc_types_engine::{PayloadAttributes as EthPayloadAttributes, PayloadId};
-use core::fmt;
+use core::{fmt, future::Future, pin::Pin};
 use either::Either;
 use reth_chain_state::ComputedTrieData;
 use reth_execution_types::BlockExecutionOutput;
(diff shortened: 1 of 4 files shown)


crates/payload/primitives/src/traits.rs
```diff
@@ -200,17 +200,25 @@ impl PayloadAttributes for op_alloy_rpc_types_engine::OpPayloadAttributes {
 pub trait PayloadAttributesBuilder<Attributes, Header = alloy_consensus::Header>:
     Send + Sync + 'static
 {
-    /// Constructs new payload attributes for the given timestamp.
-    fn build(&self, parent: &SealedHeader<Header>) -> Attributes;
+    /// Constructs new payload attributes for the given parent header.
+    fn build<'a>(
+        &'a self,
+        parent: &'a SealedHeader<Header>,
+    ) -> Pin<Box<dyn Future<Output = Attributes> + Send + 'a>>;
```
mattsse, line 207: I think this doesnt need Pin Box and can instead return impl Future so that trait impls can use async fn @gakonst ?

mattsse (changes requested):

mattsse (commented):
actually we can keep as is

PR #22079, opened 2026-02-11
perf: inline receipt root computation for small blocks

yongkangc:
**Summary**
Every block spawns a `spawn_blocking` task + crossbeam channel + per-receipt clone for receipt root computation, even when the block has few transactions. For small blocks (<30 txs), the pipelining overhead exceeds the benefit. This PR computes the receipt root inline after execution completes — no task spawn, no channel, no per-receipt clone.

**Changes**
- Add `INLINE_RECEIPT_ROOT_TX_THRESHOLD` (30 txs)
- Small blocks: execute without streaming receipts, compute receipt root inline from final receipts vec
- Large blocks: keep existing streaming background task

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -17,7 +17,11 @@ use alloy_evm::Evm;
 use alloy_primitives::B256;
 
 use crate::tree::payload_processor::receipt_root_task::{IndexedReceipt, ReceiptRootTaskHandle};
+use alloy_consensus::TxReceipt;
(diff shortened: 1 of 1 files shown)


yongkangc:
Closing this PR — after analysis, the motivation doesn't hold up. The pipelining overhead (channel + spawn_blocking) is ~5–20μs, while the inline path adds receipt root computation (~50–200μs for 30 txs) to the critical path *after* finish/merge_transitions. The pipelined version hides this behind execution, so it's net-faster even for small blocks. The PR also introduces significant code duplication in a critical path without benchmark data showing improvement.

PR #22082, opened 2026-02-11
perf: batch multiproof dispatch for small blocks

yongkangc:
**Summary**
The multiproof dispatch path sends a separate proof request per transaction state update: partition, Arc clone, chunking check, channel send, worker wakeup — repeated N times per block. For a 30-tx block, that's 30 dispatches when a single merged request would suffice. This PR buffers `HashedPostState` updates and dispatches a single merged proof request on `FinishedStateUpdates` for small blocks (<30 state updates), falling back to per-update dispatch for larger blocks.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -36,6 +36,16 @@ use tracing::{debug, debug_span, error, instrument};
 /// Maximum number of pending/prewarm updates that we accumulate in memory before actually applying.
 const MAX_PENDING_UPDATES: usize = 100;
 
+/// Maximum number of state updates to buffer for small-block batching in the hashing task.
+///
+/// For blocks with fewer state updates than this threshold, all hashed updates are merged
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/multiproof.rs
```diff
@@ -61,6 +61,16 @@ const PREFETCH_MAX_BATCH_TARGETS: usize = 512;
...
 
+/// Maximum number of state updates to buffer for small-block batching.
+///
+/// For blocks with fewer state updates than this threshold, all updates are buffered and
+/// dispatched as a single merged proof request at the end. This eliminates per-update overhead
+/// from partition, dispatch, Arc cloning, and channel operations.
+///
+/// Set to 30 to cover small blocks (< 30 txs). Above this, the per-update pipeline
+/// activates to overlap proof computation with execution.
+const SMALL_BLOCK_STATE_UPDATE_BATCH_THRESHOLD: usize = 30;
```
yongkangc:
```suggestion
const SMALL_BLOCK_STATE_UPDATE_BATCH_THRESHOLD: usize = 50;
```

PR #22083, opened 2026-02-11
feat(cli): split account-history and storage-history stage drops

gakonst:
## Summary

`reth stage drop account-history` and `reth stage drop storage-history` can now be used independently to drop only one of the two history index tables.

## Motivation

Previously both variants were handled by the same match arm, so dropping either one would always clear **both** `AccountsHistory` and `StoragesHistory` tables and reset both stage checkpoints. This makes it impossible to selectively re-index just accounts or just storage.

## Changes

- **`crates/cli/commands/src/stage/drop.rs`** – split the `AccountHistory | StorageHistory` match arm into two independent arms, each clearing only its own table and checkpoint.
- **`crates/storage/db-common/src/init.rs`** – add `insert_genesis_account_history` / `insert_genesis_storage_history` (and their non-genesis counterparts `insert_account_history` / `insert_storage_history`). The existing `insert_genesis_history` / `insert_history` now delegate to these, so there is no breaking change.

## Testing

```
cargo check -p reth-cli-commands -p reth-db-common  # clean
```

### DIFF
--- a/crates/cli/commands/src/stage/drop.rs
+++ b/crates/cli/commands/src/stage/drop.rs
@@ -9,7 +9,10 @@ use reth_db_api::{
     transaction::{DbTx, DbTxMut},
 };
 use reth_db_common::{
-    init::{insert_genesis_header, insert_genesis_history, insert_genesis_state},
+    init::{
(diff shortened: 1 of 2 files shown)


joshieDo (approved):
l
g
t
m

PR #22089, opened 2026-02-11
perf: treat hashes as bytes in BranchNodeCompact

DaniPopes:
ser/de all hash bytes at once instead of one at a time

### DIFF
--- a/crates/storage/codecs/src/alloy/trie.rs
+++ b/crates/storage/codecs/src/alloy/trie.rs
@@ -2,7 +2,7 @@
 
 use crate::Compact;
 use alloc::vec::Vec;
-use alloy_primitives::B256;
+use alloy_primitives::{FixedBytesSliceExt, B256};
 use alloy_trie::{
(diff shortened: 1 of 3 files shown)


crates/storage/codecs/src/alloy/trie.rs
```diff
@@ -67,10 +67,9 @@ impl Compact for BranchNodeCompact {
             buf.put_slice(root_hash.as_slice());
         }
 
-        for hash in self.hashes.iter() {
-            buf_size += B256::len_bytes();
-            buf.put_slice(hash.as_slice());
-        }
+        let hashes = self.hashes.as_slice().as_flattened();
+        buf_size += hashes.len();
```
Rjected, line 71: should this be `hashes.len() * B256::len_bytes()`?

DaniPopes: hashes is &[u8] already (as_flattened [B256] -> [u8])

Rjected (approved):

PR #22093, opened 2026-02-11
perf(engine): use transaction count threshold for prewarm skip

gakonst:
## Summary
Switch prewarming skip threshold from gas-based (20M gas) to transaction-count-based (30 txns).

## Motivation
Follow-up to #22059. Transaction count is a more direct proxy for prewarm overhead than gas, since the overhead comes from per-worker setup (state providers, EVM instances, precompile wrapping) which scales with transaction count, not gas.

## Changes
- Replace `SMALL_BLOCK_GAS_THRESHOLD` (20M gas) with `SMALL_BLOCK_TX_THRESHOLD` (30 txns)
- Remove `gas_used` field from `ExecutionEnv` (no longer needed)
- Remove `BlockOrPayload::gas_used()` method

## Testing
`cargo check -p reth-engine-tree` passes.

Prompted by: yk


gakonst:
Superseded — recreating with the original commit included.

PR #22101, opened 2026-02-11
perf: reduce tracing span noise in prewarm and proof workers

DaniPopes:
Reduces profiler trace overhead by cutting redundant/noisy tracing spans that contribute to massive trace file sizes (~286MB string overhead from B256 formatting, ~10.5M zones from trie machinery alone).

**Metric spans to trace level**: Trie cursor and hashed cursor metric spans (`Trie cursor metrics`, `Hashed cursor metrics`) are pure noise at debug level — they dump counter data into the timeline after every proof calculation (~304k zones). Moved to trace.

**Prewarm span cleanup**: The `prewarm tx` → `execute` → `prewarm outcome` span triple is redundant per transaction (~1.65M zones each). Removed the `prewarm outcome` span entirely and dropped `tx_hash`, `is_success`, and `gas_used` fields from `prewarm tx`. Similarly dropped `tx_hash` and `gas_used` from `execute tx` in the payload validator.

**Proof span field removal**: Removed `worker_id` from all proof worker spans (storage/account worker, storage/account/V2 multiproof calculation, blinded node). Removed `hashed_address` from storage proof calculation and waiting-for-storage-proof spans. These B256 Debug-format fields are the primary driver of the 6.2M strings in traces.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -602,9 +602,6 @@ where
                 target: "engine::tree::payload_processor::prewarm",
                 "prewarm tx",
                 index,
-                tx_hash = %tx.tx().tx_hash(),
-                is_success = tracing::field::Empty,
-                gas_used = tracing::field::Empty,
             )
             .entered();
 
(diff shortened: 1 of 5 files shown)


yongkangc (approved):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -598,13 +598,10 @@ where
         };
 
         while let Ok(IndexedTransaction { index, tx }) = txs.recv() {
-            let enter = debug_span!(
+            let _enter = debug_span!(
                 target: "engine::tree::payload_processor::prewarm",
                 "prewarm tx",
                 index,
```
DaniPopes, line 604: index is enough to map it to block's tx hash if you look hard enough

crates/trie/parallel/src/proof_task.rs
DaniPopes, line 1534: ID at the top level is enough, spans are stacked so just look up the stack

PR #22108, opened 2026-02-12
perf: add dedicated prewarming rayon pool

DaniPopes:
Add a dedicated rayon thread pool for execution prewarming, replacing tokio's spawn_blocking. This isolates prewarming work from the general blocking pool to avoid contention.

Changes:
- Add prewarming_pool to Runtime with configurable thread count
- Use prewarming_pool for all prewarm task spawning (tx workers, BAL)
- Simplify BAL prewarming to use par_iter().for_each_init() with per-thread lazy provider construction instead of manual worker spawning and range splitting
- Flatten spawn_workers to spawn directly on the pool


gakonst (approved):

crates/tasks/src/runtime.rs
```diff
@@ -862,12 +883,19 @@ impl RuntimeBuilder {
                 .thread_name(|i| format!("proof-acct-{i:02}"))
                 .build()?;
 
+            let prewarming_threads = config.rayon.prewarming_threads.unwrap_or(default_threads);
+            let prewarming_pool = rayon::ThreadPoolBuilder::new()
+                .num_threads(prewarming_threads)
+                .thread_name(|i| format!("prewarm-{i}"))
```
DaniPopes:
```suggestion
                .thread_name(|i| format!("prewarm-{i:02}"))
```

PR #22115, opened 2026-02-12
chore(stateless): Remove reth-stateless crate

mediocregopher:
Removes reth-stateless crate, and modifies the handful of tests which were using it to no longer do so.


mattsse (approved):
fyi @kevaundray moving all of this to paradigmxyz/stateless to maintain seprately

PR #22119, opened 2026-02-12
feat(download): optional chain-aware snapshot url

johnletey:
## Summary

- Adds an optional `default_chain_aware_base_url` to `DownloadDefaults` so snapshot providers can serve snapshots keyed by chain ID (`{base_url}/{chain_id}/latest.txt`)
- When configured, the download command automatically resolves the correct snapshot for the active chain
- Falls back to the existing `default_base_url` when no chain-aware URL is set

### DIFF
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -37,7 +37,13 @@ pub struct DownloadDefaults {
     pub available_snapshots: Vec<Cow<'static, str>>,
     /// Default base URL for snapshots
     pub default_base_url: Cow<'static, str>,
-    /// Default base URL for chain-aware snapshots
+    /// Default base URL for chain-aware snapshots.
+    ///
+    /// When set, the chain ID is appended to form the full URL: `{base_url}/{chain_id}`.
+    /// For example, given a base URL of `https://snapshots.example.com` and chain ID `1`,
+    /// the resulting URL would be `https://snapshots.example.com/1`.
+    ///
+    /// Falls back to [`default_base_url`](Self::default_base_url) when `None`.
     pub default_chain_aware_base_url: Option<Cow<'static, str>>,
     /// Optional custom long help text that overrides the generated help
     pub long_help: Option<String>,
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/download.rs
```diff
@@ -509,8 +518,12 @@ async fn stream_and_extract(url: &str, target_dir: &Path) -> Result<()> {
...
 
 // Builds default URL for latest mainnet archive snapshot using configured defaults
-async fn get_latest_snapshot_url() -> Result<String> {
-    let base_url = &DownloadDefaults::get_global().default_base_url;
+async fn get_latest_snapshot_url(chain_id: u64) -> Result<String> {
+    let defaults = DownloadDefaults::get_global();
+    let base_url = match &defaults.default_chain_aware_base_url {
+        Some(url) => format!("{url}/{chain_id}"),
+        None => defaults.default_base_url.to_string(),
+    };
```
mattsse, line 534: oh I see, this makes sense, could we highlight what the expected chain url then is for this in the docs?

crates/cli/commands/src/download.rs
```diff
@@ -37,6 +37,8 @@ pub struct DownloadDefaults {
     pub available_snapshots: Vec<Cow<'static, str>>,
     /// Default base URL for snapshots
     pub default_base_url: Cow<'static, str>,
+    /// Default base URL for chain-aware snapshots
+    pub default_chain_aware_base_url: Option<Cow<'static, str>>,
```
mattsse: here

mattsse (changes requested):
makes sense, pedantic doc nit

crates/cli/commands/src/download.rs
johnletey: https://github.com/paradigmxyz/reth/pull/22119/commits/f3f052c350e9140540a986590bc1906cbb5725c7

mattsse (approved):
makes sense, I think chainid is a reasonable identifier

PR #22122, opened 2026-02-12
storage-api: gate RangeBounds import behind db-api feature

conomist:
Summary
- Remove an unused `RangeBounds` import in `storage-api`.
- Keep `RangeBounds` imported only when `db-api` feature is enabled.
- Reduce warning noise and keep imports aligned with actual feature-gated usage.
Why
`RangeBounds` is only used in `StorageChangeSetReader`, which is compiled under `db-api`.  
Importing it unconditionally triggers an `unused_imports` warning in builds without that feature.
Changes
- `crates/storage/storage-api/src/storage.rs`
  - Split `core::ops` import into:
    - unconditional `RangeInclusive`
    - `#[cfg(feature = "db-api")]` `RangeBounds`

### DIFF
--- a/crates/storage/storage-api/src/storage.rs
+++ b/crates/storage/storage-api/src/storage.rs
@@ -3,7 +3,9 @@ use alloc::{
     vec::Vec,
 };
 use alloy_primitives::{Address, BlockNumber, B256};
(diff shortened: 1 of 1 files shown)


mattsse (commented):
no, id rather use the full path for this instead where we need to use it

PR #22123, opened 2026-02-12
Mediocregopher/pst compact leaves

mediocregopher:
RETH-354

### DIFF
--- a/crates/trie/sparse/src/traits.rs
+++ b/crates/trie/sparse/src/traits.rs
@@ -11,7 +11,7 @@ use alloy_trie::BranchNodeCompact;
 use reth_execution_errors::SparseTrieResult;
 use reth_trie_common::{BranchNodeMasks, Nibbles, ProofTrieNode, TrieNode};
 
-use crate::{provider::TrieNodeProvider, SparseNodeLeaf};
+use crate::provider::TrieNodeProvider;
 
 /// Describes an update to a leaf in the sparse trie.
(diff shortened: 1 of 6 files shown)


crates/trie/sparse/src/trie.rs
mediocregopher: Can we remove this method? It cannot accurately produce a Self::Branch, whatever is using it should be using a specific new_* method instead.

crates/trie/sparse/src/traits.rs
```diff
@@ -215,7 +215,7 @@ pub trait SparseTrie: Sized + Debug + Send + Sync {
     /// - The value does not exists in the trie, so it cannot be revealed
     /// - The value has not yet been revealed. In order to determine which is true, one would need
     ///   an exclusion proof.
-    fn get_leaf_value(&self, full_path: &Nibbles) -> Option<&Vec<u8>>;
+    fn get_leaf(&self, full_path: &Nibbles) -> Option<&SparseNodeLeaf>;
```
mediocregopher: I don't think this was a useful change, all usages of `get_leaf` are either just checking the option or immediately extracting the value. Revert this back to get_leaf_value

crates/trie/sparse/src/trie.rs
mediocregopher: The Leaf variant should be completely removed, we should never be storing leaf data in `nodes`, only in `values`

PR #22133, opened 2026-02-12
feat(rpc, reth-bench):  `reth_newPayload` methods for reth-bench

shekhirin:
Adds `reth_newPayload*` RPC methods with server-side timing breakdown.

These methods mirror the standard `engine_newPayload*` API but return server-side execution timings — new payload latency, persistence wait, execution cache wait, and sparse trie wait — enabling precise benchmarking without relying on client-side round-trip measurements.

This allows us to send new payload requests to Reth back-to-back without any artificial wait time between payloads, while simulating enough time between blocks (Ethereum's 12s) to do persistence, update caches, etc.


yongkangc (approved):

crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -1283,14 +1313,40 @@ where
     }
 }
 
+/// Implementation of `RethEngineApiServer` under the `reth_` namespace.
+///
+/// Waits for execution cache and sparse trie locks before processing.
+#[async_trait]
+impl<Provider, EngineT, Pool, Validator, ChainSpec> RethEngineApiServer
+    for EngineApi<Provider, EngineT, Pool, Validator, ChainSpec>
```
klkvr, line 1321: I think we might want to extract this from `EngineApi` so that it's enabled by default even on clients that use custom engine API impl (like OP or Tempo)

i.e i could see it being enabled by just `--http.api reth`

klkvr (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -62,6 +63,26 @@ pub mod sparse_trie;
...
+/// Trait for types that can wait for execution cache and sparse trie locks to become available.
+///
+/// This is used by `reth_newPayload` endpoint to ensure that payload processing
+/// waits for any ongoing operations to complete before starting.
+pub trait WaitForCaches {
+    /// Waits for persistence and cache updates to complete.
+    ///
+    /// Returns the time spent waiting for each cache separately.
+    fn wait_for_caches(&self) -> CacheWaitDurations;
+}
```
mattsse, line 84: this looks a bit weird, why do we need this?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -62,6 +63,26 @@ pub mod sparse_trie;
...
 use preserved_sparse_trie::{PreservedSparseTrie, SharedPreservedSparseTrie};
 
+/// Result of waiting for caches to become available.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct CacheWaitDurations {
+    /// Time spent waiting for the execution cache lock.
+    pub execution_cache: Duration,
+    /// Time spent waiting for the sparse trie lock.
+    pub sparse_trie: Duration,
+}
```
mattsse, line 73: these types are out of order

PR #22137, opened 2026-02-12
fix(cli): correct error message for non-DUPSORT table arguments

letmehateu:
The error message said "Only END_KEY can be given" but the code actually rejects BOTH end_key and end_subkey for non-DUPSORT tables. For non-DUPSORT tables, subkey is used as the end key for range queries.


mattsse (commented):
can keep as is

PR #22138, opened 2026-02-12
perf: release execution cache lock before waiting for state root validation

yongkangc:
**Summary**
Traced via Grafana Tempo that `save_cache` is the #1 bottleneck in the `new_payload` pipeline, blocking for ~95-131ms (up to 80% of NP time) while holding the `ExecutionCache` mutex. The root cause is that `valid_block_rx.recv()` (waiting for state root validation) is called *inside* `update_with_guard`, holding the lock for the entire state root duration. This blocks the next block's prewarming from accessing the cache.

Split `save_cache` from a single locked phase into three phases:
1. **Phase 1 (locked, ~1ms):** consume the `SavedCache`, run `insert_state`, and optimistically publish the warmed cache
2. **Phase 2 (unlocked, ~100ms+):** wait for `valid_block_rx` without holding the lock — the next block's prewarming can start immediately
3. **Phase 3 (locked, only if invalid):** re-acquire the lock and clear the cache only if it still belongs to this block (hash guard prevents clearing a newer valid cache from a subsequent block)

**Changes**
- Restructured `save_cache` to release the `ExecutionCache` lock before blocking on `valid_block_rx.recv()`
- Added hash-guarded rollback: on invalid blocks, only clears cache if `executed_block_hash() == hash`

**Expected Impact**
Reduces execution cache lock hold time from ~100ms+ to ~1ms per block, unblocking prewarming for subsequent blocks and improving overall NP throughput.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/prewarm.rs
+++ b/crates/engine/tree/src/tree/payload_processor/prewarm.rs
@@ -211,15 +211,16 @@ where
         });
     }
 
-    /// This method calls `ExecutionCache::update_with_guard` which requires exclusive access.
-    /// It should only be called after ensuring that:
-    /// 1. All prewarming tasks have completed execution
-    /// 2. No other concurrent operations are accessing the cache
-    ///
     /// Saves the warmed caches back into the shared slot after prewarming completes.
     ///
-    /// This consumes the `SavedCache` held by the task, which releases its usage guard and allows
-    /// the new, warmed cache to be inserted.
+    /// Waits for block validation without any lock held, then only on success splits the
+    /// saved cache, inserts state, and publishes under a brief write lock. This avoids
+    /// the ~100ms+ lock hold that previously blocked concurrent readers during
+    /// `valid_block_rx.recv()`.
+    ///
+    /// The ordering is critical: `split()` releases the usage guard and `insert_state()`
+    /// mutates the shared fixed-caches in-place, so both must happen only after
+    /// validation to prevent concurrent readers from observing unvalidated state.
     ///
     /// This method is called from `run()` only after all execution tasks are complete.
     #[instrument(level = "debug", target = "engine::tree::payload_processor::prewarm", skip_all)]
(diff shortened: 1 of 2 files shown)


mattsse (commented):
I'm pretty sure I reviewed the same change somewhere else

this change is problematic because this doesnt guard against re-use for alternative blocks, e.g. we just executed `N + 1 == A` what happens when we get a different `A'` with N as parent.

now the cache can already point to `A` but we haven't adjusted the anchor yet

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -236,36 +237,37 @@ where
...
-                    *cached = None;
-                    debug!(target: "engine::caching", "cleared execution cache on update error");
-                    return;
-                }
 
+            // Wait for state root validation WITHOUT holding the cache lock.
+            // This is the key optimization: the original code held the lock across this
+            // blocking recv(), which blocked the next block's prewarming from accessing
+            // the cache for ~100ms+.
+            if valid_block_rx.recv().is_err() {
```
mattsse, line 245: isnt this unproblematic anyways, because this fires once we have validated the block and the remaining ops here are cheap

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -211,15 +211,16 @@ where
...
-    /// 2. No other concurrent operations are accessing the cache
-    ///
     /// Saves the warmed caches back into the shared slot after prewarming completes.
     ///
-    /// This consumes the `SavedCache` held by the task, which releases its usage guard and allows
-    /// the new, warmed cache to be inserted.
+    /// Waits for block validation without any lock held, then only on success splits the
+    /// saved cache, inserts state, and publishes under a brief write lock. This avoids
+    /// the ~100ms+ lock hold that previously blocked concurrent readers during
+    /// `valid_block_rx.recv()`.
```
mattsse, line 219: there are no concurrent readers here, because the main exec thread is busy validating the block

mattsse (changes requested):
I don't think the assessment here is correct.

this change doesnt make it faster overall and just moves around the locking which is now problematic

the meassurements of 

https://github.com/paradigmxyz/reth/blob/4b444069a5cbbd21c4c0d17a7c00b19c476345c0/crates/engine/tree/src/tree/payload_processor/prewarm.rs#L271-L271

are flawed because this includes both cache updates + waiting until the block was validated

PR #22144, opened 2026-02-12
perf(engine): replace channel+BTreeMap reorder with lock-free for_each_ordered

gakonst:
## Summary
Replace the 3-channel + BTreeMap reorder task in payload tx processing with a lock-free `ForEachOrdered` trait that streams rayon parallel iterator results in index order using an atomic bitmap buffer.

## Changes
- Add `rayon_ordered` module to `reth-tasks` with `ForEachOrdered` extension trait on `IndexedParallelIterator`
- `AtomicBitmap`: packed `AtomicU64` readiness flags (1 bit per slot, `fetch_or` / `fetch_and`)
- `Shared<T>`: pre-allocated `MaybeUninit` buffer, producers write by index, consumer drains in order
- Uses `rayon::in_place_scope` + `rayon::yield_now()` for work-stealing while the consumer waits
- Replace `spawn_tx_iterator` parallel path: eliminates ooo channel, reorder task, and BTreeMap
- Sequential path simplified (no longer needs index tracking)

## Testing
Existing e2e and unit tests cover the payload processing flow. The `ForEachOrdered` implementation has been tested separately with 32 unit tests + 16 miri tests (https://github.com/tempoxyz/rayon-ordered).

Prompted by: danipopes


crates/tasks/src/for_each_ordered.rs
```diff
@@ -0,0 +1,143 @@
...
+        ordered_impl(self, f);
+    }
+}
+
+fn ordered_impl<I, F>(iter: I, mut f: F)
+where
+    I: IndexedParallelIterator,
+    I::Item: Send,
+    F: FnMut(I::Item),
+{
```
mattsse, line 113: can we document what this does exactly and why this works

crates/tasks/src/rayon_ordered/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+            unsafe {
+                (*self.slots[idx].0.get()).assume_init_drop();
+            }
+        }
+    }
+}
+
+/// Extension trait for [`IndexedParallelIterator`]
+/// that streams results to a sequential consumer in index order.
+pub trait ForEachOrdered: IndexedParallelIterator {
```
mattsse: can we reorder this so that this is top of file

crates/tasks/src/rayon_ordered/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+                // Yield to rayon's work-stealing so the producer can make progress,
+                // especially important when the thread pool is small.
+                if rayon::yield_now() == Some(rayon::Yield::Executed) {
+                    continue;
+                }
+                std::hint::spin_loop();
+                spins = spins.wrapping_add(1);
+                if spins.is_multiple_of(128) {
+                    std::thread::yield_now();
+                }
```
mattsse: where is this coming from

crates/tasks/src/rayon_ordered/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+                if shared.ready.is_set(i, Ordering::Acquire) {
+                    break;
+                }
+                if shared.panicked.load(Ordering::Relaxed) {
+                    return;
+                }
+                // Yield to rayon's work-stealing so the producer can make progress,
+                // especially important when the thread pool is small.
+                if rayon::yield_now() == Some(rayon::Yield::Executed) {
+                    continue;
```
mattsse: can we label the loops here, I found this continue confusing at first

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -409,53 +409,32 @@ where
...
+            // Parallel path — recover signatures in parallel on rayon, stream results
+            // to execution in order via `for_each_ordered` (lock-free atomic bitmap reorder).
             rayon::spawn(move || {
                 let (transactions, convert) = transactions.into_parts();
-                transactions.into_par_iter().enumerate().for_each_with(
-                    ooo_tx,
-                    |ooo_tx, (idx, tx)| {
+                transactions
+                    .into_par_iter()
+                    .map(|tx| {
```
mattsse, line 430: this is great, this will remove the ordering on a different task overhead entirely

crates/tasks/src/for_each_ordered.rs
mattsse, line 1: can we add some basic tests for this, ideally also with a cases where idx n takes a long time to execute but is still returned in order

mattsse (dismissed):
this is pretty clever

afaict the impl is sound, just have a few nits

DaniPopes (approved):

DaniPopes (approved):

PR #22152, opened 2026-02-12
perf(cli): remove clone in trie repair

andrewshab3:
When upserting storage trie entries, `nibbles` was cloned twice: once for creating the entry struct and once for seek_by_key_subkey.
By moving entry creation after the seek operation, we can move `nibbles` into the entry instead of cloning it.


mattsse (approved):
fair

PR #22172, opened 2026-02-13
fix(pruning): trim spaces in receipts log filter parsing

gancerlory:
Ran into this while testing pruning config - spaces after commas in `--prune.receiptslogfilter` cause address parsing to fail:

"0x1234:full, 0x5678:distance:1000" - the space before 0x5678 breaks it

Other parsers in the codebase (RPC modules, debug args) already handle this with `.map(str::trim)`. Just added the same here.


mattsse (approved):
makes sense

PR #22174, opened 2026-02-14
feat: impl bal_store for eip-7528

0xKarl98:
###Summary 
Productionizes BAL retention in the Engine API by introducing a disk-backed BAL store and layering BalCache as an in-memory L1 cache on top of persistent storage.
This extends the in-memory BAL cache work in #21247.

###Changes
- Add a BalStore abstraction for BAL persistence and retrieval (insert, get_by_hashes, get_by_range)
- Implement DiskFileBalStore 
- Keep BalCache as the fast in-memory layer and implement BalStore for API compatibility
- Refactor EngineApi BAL wiring
- cache-first reads with partial fallback to store
- unified BAL write path (VALID payloads persisted to store, then cached)

cc @mattsse


crates/net/bal-store/src/disk.rs
```diff
@@ -0,0 +1,387 @@
...
+    index_dir: PathBuf,
+    max_entries: u32,
+    state: Mutex<IndexState>,
+}
+
+#[derive(Debug, Default)]
+struct IndexState {
+    block_to_hash: BTreeMap<BlockNumber, BlockHash>,
+    hash_to_block: HashMap<BlockHash, BlockNumber>,
+}
```
mattsse, line 116: these types are defined out of order

mattsse (changes requested):
we also need to share access to this with the network stacke (your last pr)

so we should move this implementation to somewhere else
looks like we can even make this a standalone crate for this, or add to network types

crates/net/bal-store/src/disk.rs
0xKarl98: Have extracted this part of logic to another crate and reorg the struct

crates/net/bal-store/src/disk.rs
```diff
@@ -0,0 +1,413 @@
...
+
+#[derive(Debug)]
+struct DiskFileBalStoreInner {
+    root_dir: PathBuf,
+    entries_dir: PathBuf,
+    index_dir: PathBuf,
+    max_entries: u32,
+    state: Mutex<IndexState>,
+}
```
mattsse, line 116: these types are defined out of order

crates/net/bal-store/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+mod noop;
+pub use disk::{DiskFileBalStore, DiskFileBalStoreConfig, DEFAULT_MAX_BAL_STORE_ENTRIES};
+pub use noop::NoopBalStore;
+
+/// A store for EIP-7928 Block Access Lists (BALs).
+///
+/// The store is keyed by block hash and maintains a block-number index for range queries.
+/// Implementations should preserve contiguous-range semantics:
+/// queries by range stop at the first missing block.
+pub trait BalStore: Send + Sync + 'static {
```
mattsse: lets require this to be debug as well

crates/node/builder/src/rpc.rs
```diff
@@ -1397,7 +1398,7 @@ where
             commit: version_metadata().vergen_git_sha.to_string(),
         };
 
-        Ok(EngineApi::new(
+        Ok(EngineApi::with_bal_store(
```
mattsse: lets just update the new fn with this additional arg

crates/net/bal-store/src/disk.rs
mattsse, line 1: we can expect that the most recent blocks are requested mor often, so lets also add a simple cache for the last 1024 blocks or so

mattsse (changes requested):
cool, this is looking promising already.

I think instead of integrating this into the engine api, we can also integrate this directly into the ethapi because we also need this for some endpoints there

crates/node/builder/src/rpc.rs
0xKarl98: Makes sense , this arc is more clean with only new as the entry

crates/net/bal-store/src/disk.rs
0xKarl98: Yeah this makes sense

PR #22186, opened 2026-02-14
feat(rpc): add eth_getStorageValues batch storage slot retrieval

gakonst:
## Summary
Implements `eth_getStorageValues` per [ethereum/execution-apis#752](https://github.com/ethereum/execution-apis/issues/752).

## Motivation
`eth_getStorageAt` fetches one storage slot per call. Consumers doing heavy local simulations often need hundreds of slots across multiple contracts at the same block. A batch method eliminates repeated state lookups and round-trips.

## Changes
- `crates/rpc/rpc-server-types/src/constants.rs`: add `DEFAULT_MAX_STORAGE_VALUES_SLOTS = 1024`
- `crates/rpc/rpc-eth-api/src/core.rs`: add `getStorageValues` to `EthApi` trait + server impl
- `crates/rpc/rpc-eth-api/src/helpers/state.rs`: add `storage_values` to `EthState` trait with slot cap enforcement

## Testing
Compilation verified. Method follows existing `storage_at` pattern exactly.

Prompted by: mattsse

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/state.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/state.rs
@@ -94,36 +94,31 @@ pub trait EthState: LoadState + SpawnBlocking {
         requests: HashMap<Address, Vec<JsonStorageKey>>,
         block_id: Option<BlockId>,
     ) -> impl Future<Output = Result<HashMap<Address, Vec<B256>>, Self::Error>> + Send {
-        async move {
+        self.spawn_blocking_io_fut(move |this| async move {
             let total_slots: usize = requests.values().map(|slots| slots.len()).sum();
             if total_slots > DEFAULT_MAX_STORAGE_VALUES_SLOTS {
-                return Err(Self::Error::from_eth_err(EthApiError::InvalidParams(
-                    format!(
-                        "total slot count {total_slots} exceeds limit {DEFAULT_MAX_STORAGE_VALUES_SLOTS}",
-                    ),
-                )));
+                return Err(Self::Error::from_eth_err(EthApiError::InvalidParams(format!(
+                    "total slot count {total_slots} exceeds limit {DEFAULT_MAX_STORAGE_VALUES_SLOTS}",
+                ))));
             }
 
-            let state = self.state_at_block_id_or_latest(block_id).await?;
+            let state = this.state_at_block_id_or_latest(block_id).await?;
 
-            self.spawn_blocking_io(move |_| {
-                let mut result = HashMap::with_capacity(requests.len());
-                for (address, slots) in requests {
-                    let mut values = Vec::with_capacity(slots.len());
-                    for slot in &slots {
-                        let value = state
-                            .storage(address, slot.as_b256())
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -83,6 +85,44 @@ pub trait EthState: LoadState + SpawnBlocking {
...
+                        "total slot count {total_slots} exceeds limit {DEFAULT_MAX_STORAGE_VALUES_SLOTS}",
+                    ),
+                )));
+            }
+
+            let state = this.state_at_block_id_or_latest(block_id).await?;
+
+            let mut result = HashMap::with_capacity(requests.len());
+            for (address, slots) in requests {
+                let mut values = Vec::with_capacity(slots.len());
```
mattsse: we need to perform this on a spawned blocking task because this can end up doing a lot of io, see spawn_blocking_io_fut

mattsse (changes requested):

crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -83,6 +85,47 @@ pub trait EthState: LoadState + SpawnBlocking {
...
+                return Err(Self::Error::from_eth_err(EthApiError::InvalidParams(
+                    format!(
+                        "total slot count {total_slots} exceeds limit {DEFAULT_MAX_STORAGE_VALUES_SLOTS}",
+                    ),
+                )));
+            }
+
+            let state = self.state_at_block_id_or_latest(block_id).await?;
+
+            self.spawn_blocking_io(move |_| {
```
mattsse: I think this can be comvined like:

```
        self.spawn_blocking_io_fut(move |this| async move {
            let state = this.state_at_block_id(at).await?;
```

mattsse (changes requested):

crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -83,6 +85,42 @@ pub trait EthState: LoadState + SpawnBlocking {
...
+        requests: HashMap<Address, Vec<JsonStorageKey>>,
+        block_id: Option<BlockId>,
+    ) -> impl Future<Output = Result<HashMap<Address, Vec<B256>>, Self::Error>> + Send {
+        self.spawn_blocking_io_fut(move |this| async move {
+            let total_slots: usize = requests.values().map(|slots| slots.len()).sum();
+            if total_slots > DEFAULT_MAX_STORAGE_VALUES_SLOTS {
+                return Err(Self::Error::from_eth_err(EthApiError::InvalidParams(format!(
+                    "total slot count {total_slots} exceeds limit {DEFAULT_MAX_STORAGE_VALUES_SLOTS}",
+                ))));
+            }
```
mattsse: this validation should happen before we spawn to fail early

mattsse (changes requested):

mattsse (approved):

mattsse (approved):

PR #22188, opened 2026-02-14
perf(storage): remove redundant clone in ConsistentProvider range fetch

marukai67:
Remove unnecessary clone() of storage_range in ConsistentProvider::get_range_with_predicate by calculating the expected length before consuming the range.


mattsse (commented):
can keep asis

PR #22190, opened 2026-02-14
test(reth-bench): add CLI integration tests

HuberyJulianay:
## Summary
- Add CLI integration tests for `reth-bench` binary covering all 7 subcommands
- Tests verify `--help` output, unknown subcommand/flag rejection, and that each subcommand's clap configuration is wired correctly
- Follows the same pattern as `bin/reth/tests/it/main.rs`


mattsse (commented):
no need for this

PR #22191, opened 2026-02-14
test(reth-bench-compare): add CLI integration tests

HuberyJulianay:
## Summary
- Add CLI integration tests for `reth-bench-compare` binary
- Tests verify `--help`, `--version`, missing required args (`--baseline-ref`, `--feature-ref`), and clap configuration correctness
- Follows the same pattern as `bin/reth/tests/it/main.rs`


mattsse (commented):
no need for this

PR #22193, opened 2026-02-14
test(reth-bench): add unit tests for InvalidationConfig and parse_gas_limit overflow

HuberyJulianay:
## Summary
- Add 11 unit tests for `InvalidationConfig` in `send_invalid_payload/invalidation.rs`:
  - Default config produces no changes
  - Explicit field override (state_root)
  - Auto-invalidation behaviors: gas_used exceeds limit, timestamp zeroed, invalid RLP prepended
  - V2 withdrawal injection
  - V3 blob gas explicit override and auto-invalidation
  - `should_skip_hash_recalc` for both explicit hash and auto-invalidate triggers
- Add 1 test for `parse_gas_limit` overflow via `checked_mul` in `helpers.rs`

### DIFF
--- a/bin/reth-bench/src/bench/helpers.rs
+++ b/bin/reth-bench/src/bench/helpers.rs
@@ -291,6 +291,14 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_gas_limit_overflow() {
+        // u64::MAX * 1G overflows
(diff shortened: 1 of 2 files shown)


mattsse (commented):
no need for this

PR #22194, opened 2026-02-14
test(reth-bench-compare): add unit tests for build_additional_args and arg parsing

HuberyJulianay:
## Summary
- Add 15 unit tests to `cli.rs` covering:
  - `DisableStartupSyncStateIdle` enum: `FromStr` parsing, case insensitivity, invalid input, `Display` roundtrip (4 tests)
  - `parse_args_string`: simple splitting, quoted args, empty input (3 tests)
  - `build_additional_args`: default behavior adds flag to all ref types, `--disable-startup-sync-state-idle` with `all`/`baseline`/`feature` variants, base args passthrough (5 tests)
  - `Args` helpers: `get_warmup_blocks` default and override, `output_dir_path` tilde expansion (3 tests)

### DIFF
--- a/bin/reth-bench-compare/src/cli.rs
+++ b/bin/reth-bench-compare/src/cli.rs
@@ -1031,3 +1031,187 @@ async fn get_samply_path() -> Result<String> {
 
     Ok(samply_path)
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
no need for this

PR #22196, opened 2026-02-14
feat(stages): add unified BlockProcessingStage

gakonst:
## Summary
Unified pipeline stage that replaces the sequential Bodies → SenderRecovery → TransactionLookup chain with a single `BlockProcessingStage`.

## Motivation
The three stages currently run sequentially with a static file commit between body writing and sender recovery. This commit is required because the recovery OS thread reads transactions back from static files. By extracting transactions from the in-memory download buffer instead, we eliminate the commit entirely and can parallelize recovery + tx lookup.

## Changes
- `crates/stages/stages/src/stages/block_processing.rs` — new unified stage
- `crates/stages/stages/src/stages/mod.rs` — register module + re-export

## Architecture
1. **Phase 1 (sequential)**: Write block bodies via `append_block_bodies`. Extract `(tx_num, tx)` pairs from the in-memory buffer *before* writing.
2. **Phase 2 (parallel)**: `rayon::join` runs ECDSA sender recovery (`par_chunks` + `recover_unchecked_with_buf`) and tx hash extraction concurrently. All DB writes happen on the main thread after both complete.
3. **Checkpointing**: Reports primary checkpoint under `StageId::Bodies` and saves `SenderRecovery` + `TransactionLookup` checkpoints so downstream stages see consistent state.
4. **Unwind**: Reverses in dependency order — tx lookup → senders → bodies.

Key difference from the old approach: no static file commit between writing bodies and recovering senders. The transactions are already in memory from the downloader response.

## Testing
```
cargo test -p reth-stages --test pipeline test_pipeline  # 2/2 pass (v1 + v2)
```

Prompted by: joshie

### DIFF
--- a/crates/stages/stages/src/stages/block_processing.rs
+++ b/crates/stages/stages/src/stages/block_processing.rs
@@ -1,0 +1,496 @@
+use super::bodies::ensure_consistency;
+use alloy_eips::eip2718::Encodable2718;
+use alloy_primitives::{Address, TxHash, TxNumber};
+use futures_util::TryStreamExt;
+use num_traits::Zero;
+use reth_config::config::StageConfig;
+#[cfg(all(unix, feature = "rocksdb"))]
+use reth_db_api::Tables;
+use reth_db_api::{
+    table::{Decode, Decompress, Value},
+    tables,
+    transaction::DbTxMut,
+};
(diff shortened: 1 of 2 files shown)


emmajam:

PR #22199, opened 2026-02-15
fix: defer insert_state until after block validation to prevent cache race

qu0b:
## Summary

- Move `insert_state` in `save_cache` to **after** `valid_block_rx.recv()` to prevent a race condition with the multiproof task that causes permanent node stalls

## Problem

The `save_cache` method in `prewarm.rs` calls `insert_state` before waiting for block validation (`valid_block_rx.recv()`). At this point, the multiproof task is still actively reading from the same shared `Arc<FixedCache>` while computing state root proofs.

The `FixedCache` uses non-blocking inserts (`try_lock_ret(None)`) that **silently drop writes** when a bucket is locked by a concurrent reader. If the multiproof task happens to read any address that maps to the same FixedCache bucket as an account being updated by `insert_state`, that account's state update (e.g. nonce increment) is silently lost.

The poisoned cache is then stored and reused for the next block, which reads the stale nonce and permanently fails with `nonce too high`. The node cannot recover without a data wipe.

### Observed on bal-devnet-2

The lighthouse-reth node got permanently stuck at block 52265. Every attempt to execute block 52266 failed at tx 148/235 with `nonce 26754 too high, expected 26753` — repeating 1758+ times. The RPC correctly returned nonce 26754 at block 52265, but the execution engine's `CachedStateProvider` consistently returned 26753 from the FixedCache.

## Fix

Move `insert_state` to after `valid_block_rx.recv()`. By this point:
1. The main thread has received the state root from the multiproof task
2. The multiproof task has finished its main loop and is no longer reading from the FixedCache
3. All FixedCache inserts succeed without contention

This also simplifies the invalid-block path — we no longer mutate the cache before knowing whether the block is valid, so there's no need to clear a "polluted" cache on failure.

## Test plan

- [x] `cargo check -p reth-engine-tree` passes
- [x] Deployed fix image to live bal-devnet-2 lighthouse-reth node
- [x] Node synced to chain tip and processes blocks without nonce errors
- [x] All 10 network nodes in sync after fix deployment


emmajam (changes requested):
needs some changes

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -280,24 +268,27 @@ where
...
+                //
+                // This ordering is critical: `insert_state` must happen AFTER the block is
+                // validated (and thus after the multiproof/state-root task has finished reading
+                // from the shared FixedCache). The FixedCache uses non-blocking inserts that
+                // silently drop writes when a bucket is locked by a concurrent reader. If
+                // `insert_state` runs while the multiproof task is still reading from the same
+                // cache, account updates (e.g. nonce increments) can be silently lost, causing
+                // the next block to execute against stale state.
+                if valid_block_rx.recv().is_ok() {
+                    if new_cache.cache().insert_state(&execution_outcome.state).is_err() {
```
emmajam, line 276: `insert_state` + `update_metrics` are now nested inside if `valid_block_rx.recv().is_ok()`, meaning metrics are skipped on invalid blocks. 

Previously `update_metrics()` ran unconditionally (L292 in base). 

if the cache is thrown away on invalid blocks, it's fine, but it's an undocumented behavioural change?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -280,24 +268,27 @@ where
...
 
-                // Insert state into cache while holding the lock
-                // Access the BundleState through the shared ExecutionOutcome
-                if new_cache.cache().insert_state(&execution_outcome.state).is_err() {
-                    // Clear the cache on error to prevent having a polluted cache
-                    *cached = None;
-                    debug!(target: "engine::caching", "cleared execution cache on update error");
-                    return;
-                }
+                // Wait for block validation before inserting state into the cache.
```
emmajam: this comment block is good but verbose for the project's style. 

condense to 2-3 lines per CLAUDE.md ("explain WHY, not what").

crates/engine/tree/src/tree/cache_race_poc.rs
```diff
@@ -0,0 +1,121 @@
...
+        let final_nonce = cache.get(&1).unwrap().nonce;
+
+        eprintln!(
+            "\n=== FixedCache Race Condition PoC ===\n\
+             Reader threads:  {num_reader_threads}\n\
+             Clone hold:      5ms\n\
+             Insert attempts: {total_inserts}\n\
+             Final nonce:     {final_nonce} (expected: {total_inserts})\n\
+             Write dropped:   {}\n",
+            final_nonce != total_inserts
```
emmajam: If `FixedCache` is ever fixed to retry/block on insert, this test will start failing. It should either be an `#[ignore]` demonstration or assert the fix rather than the bug.

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -1082,7 +1108,165 @@ mod tests {
...
+    /// silent nonce corruption via FixedCache's non-blocking bucket locks.
+    ///
+    /// This reproduces the production bug: multiproof workers call `account_cache.get()`
+    /// which holds the per-bucket lock during `Account` clone. A concurrent
+    /// `insert_state()` call sees `LOCKED_BIT` set and silently drops the nonce update.
+    ///
+    /// The test is probabilistic — `Account` clone is fast (~nanoseconds), so the race
+    /// window is small. 4 reader threads maximize the chance of hitting it.
+    #[test]
+    fn test_insert_state_nonce_dropped_during_concurrent_reads() {
```
emmajam: this test has no assertion on the outcome - it just `eprintln!`s

This test always passes regardless of whether the race is hit, so it's dead weight — either remove it or make it deterministic with the slow-clone trick from the PoC

crates/engine/tree/src/tree/cache_race_poc.rs
qu0b: I can remove the tests, they were just to demonstrate the issue such that the developer could reproduce the issue. Let me know how you want to proceed.

crates/engine/tree/src/tree/mod.rs
```diff
@@ -57,19 +55,20 @@ pub mod error;
 pub mod instrumented_state;
 mod invalid_headers;
 mod metrics;
-pub mod payload_processor;
+mod payload_processor;
 pub mod payload_validator;
 mod persistence_state;
 pub mod precompile_cache;
 #[cfg(test)]
+mod cache_race_poc;
```
emmajam: `cache_race_poc` is registered under `crates/engine/tree/src/tree/mod.rs` (sibling of `cached_state.rs`) rather than under `payload_processor/mod.rs`

the PoC tests the `FixedCache` primitive, not the tree module — consider placing it closer to where `FixedCache` is defined, or consolidating into `cached_state.rs` tests

emmajam (commented):
some comments ty

crates/engine/tree/src/tree/mod.rs
qu0b: I didn't plan on merging the tests

crates/engine/tree/src/tree/payload_processor/prewarm.rs
qu0b: can do

crates/engine/tree/src/tree/payload_processor/prewarm.rs
qu0b: moved updated_metrics outside

emmajam (approved):

PR #22201, opened 2026-02-15
fix(rpc): enforce timeout in debug trace methods

### LINKED ISSUE #22027: debug_traceCallMany: timeout option does not interrupt execution
### Describe the bug

It appears that `debug_traceCallMany` in reth does **not enforce `timeout` as a hard execution limit**.

After reviewing the implementation, the tracing execution path does not check for:
- elapsed time
- request cancellation
- deadline expiration

As a result, even when a timeout is provided in `GethDebugTracingOptions`, the node continues executing the full trace.

### Steps to reproduce

1. Run a reth node with debug RPC enabled.

2. Construct a computationally expensive call bundle and large gaslimit for `debug_traceCallMany`, for example:
   - A transaction that executes multiple Uniswap v4 `swap` operations in a single call
   - The call involves deep call stacks, multiple `delegatecall`s, and repeated pool state updates
   - Gas limit is set sufficiently high to allow full execution

3. Call `debug_traceCallMany` with  timeout (e.g. 5s) and the request doesn't stop as expected.

### Node logs

```text

```

### Platform(s)

Linux (ARM)

### Container Type

Docker

Base Reth Node Version: 1.9.3-dev
Commit SHA: https://github.com/paradigmxyz/reth/commit/27a8c0f5a6dfb27dea84c5751776ecabdd069646
Build Timestamp: 2026-02-09T05:53:30.562535116Z
Build Features:
Build Profile: maxperf

Current database version: 2
Local database version: 2

Base mainnet

Archive (default)

_No response_

_No response_

### Code of Conduct

0xMars42:
## Summary

GethDebugTracingOptions.timeout is silently dropped by DebugInspector::new() (destructures with ..). This adds enforcement across all debug trace methods.

Parses Go-style duration strings (5s, 300ms, 1.5s, 2m, 1h), creates an Instant deadline and checks it between per-tx executions. Returns ExecutionTimedOut when exceeded. Defaults to 5s when unset.

Affected methods: trace_block, debug_trace_transaction, debug_trace_call, debug_trace_call_at_tx_index, debug_trace_call_many.

Timeout granularity is per-transaction since we can't interrupt the EVM mid-opcode.

Fixes #22027

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -40,9 +40,61 @@ use reth_trie_common::{updates::TrieUpdates, HashedPostState};
 use revm::DatabaseCommit;
 use revm_inspectors::tracing::{DebugInspector, TransactionContext};
 use serde::{Deserialize, Serialize};
-use std::{collections::VecDeque, sync::Arc};
+use std::{
+    collections::VecDeque,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 use tokio::sync::{AcquireError, OwnedSemaphorePermit};
 use tokio_stream::StreamExt;
+
+/// Default timeout for debug trace methods, matching Geth's default of 5 seconds.
+const DEFAULT_TRACE_TIMEOUT: Duration = Duration::from_secs(5);
+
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this doesnt quite work like this

PR #22203, opened 2026-02-15
fix(engine): prevent prewarming cache poisoning from snapshot divergence

### LINKED ISSUE #22159: bug: prewarming can poison execution cache when persistence commits between provider builds
### Describe the bug

Hey, I'm building on the reth SDK for a chain with ~10s block times and I've been hitting sporadic gas mismatch errors on `newPayload`. I think I've tracked it down to a race condition but wanted to get your eyes on it before I go too far down the wrong path.

In `payload_validator.rs`, `validate_block_with_state` creates the execution state provider:

```rust
// payload_validator.rs:386
let mut state_provider = ensure_ok!(provider_builder.build());
``` 
That same `provider_builder` then gets passed into `spawn_payload_processor`:
```rust
// payload_validator.rs:438-445
let mut handle = ensure_ok!(self.spawn_payload_processor(
    env.clone(),
    txs,
    provider_builder,
    overlay_factory.clone(),
    strategy,
    block_access_list,
));
``` 
Inside `spawn()`, the builder gets cloned for prewarming at lines 258 and 268:
```rust
// payload_processor/mod.rs:258
provider_builder.clone(),
``` 
The prewarming workers each clone the `PrewarmContext` (which contains the builder) and call `.build()` in their own threads:
```rust
// prewarm.rs:509-519 (inside evm_for_ctx, called per worker)
let mut state_provider = match provider.build() {
    Ok(provider) => provider,
    Err(err) => {
        trace!(
            target: "engine::tree::payload_processor::prewarm",
            %err,
            "Failed to build state provider in prewarm thread"
        );
        return None
    }
};
``` 
Same thing happens for BAL workers:
```rust
// prewarm.rs:734-745
let state_provider = match provider.build() {
    Ok(provider) => provider,
    Err(err) => {
        trace!(
            target: "engine::tree::payload_processor::prewarm",
            %err,
            "Failed to build state provider in BAL prewarm thread"
        );
        let _ = done_tx.send(());
        return;
    }
};
``` 
And the multiproof task:
```rust
// payload_processor/mod.rs:301
let provider = provider_builder.build().expect("failed to build provider");
``` 
Each `.build()` call goes through `StateProviderBuilder::build()` which opens a fresh MDBX read transaction every time:
```rust
// tree/mod.rs:115-121
pub fn build(&self) -> ProviderResult<StateProviderBox> {
    let mut provider = self.provider_factory.state_by_block_hash(self.historical)?;
    if let Some(overlay) = self.overlay.clone() {
        provider = Box::new(MemoryOverlayStateProvider::new(provider, overlay))
    }
    Ok(provider)
}
``` 
If the persistence thread commits a block to MDBX between the execution `.build()` (line 386) and any worker `.build()` (lines 509, 734, 301), the workers end up with a newer snapshot than execution. Prewarming writes values from that newer snapshot into the shared `ExecutionCache`, and execution reads them as if they were from its own snapshot.
Example: account 0xABC has nonce 5 in execution's snapshot, nonce 6 in prewarming's snapshot (because persistence committed block N-1 which included a tx from that account). Prewarming caches nonce=6. Execution reads nonce=6 from cache instead of nonce=5 from its own provider. Wrong nonce, different gas calc, block fails validation.
The cache hit path in `cached_state.rs` returns cached values without checking against the provider:
```rust
// cached_state.rs:139-154
impl<S: AccountReader> AccountReader for CachedStateProvider<S> {
    fn basic_account(&self, address: &Address) -> ProviderResult<Option<Account>> {
        if let Some(res) = self.caches.account_cache.get(address) {
            self.metrics.account_cache_hits.increment(1);
            return Ok(res)
        }

        self.metrics.account_cache_misses.increment(1);

        let res = self.state_provider.basic_account(address)?;

        if self.is_prewarm() {
            self.caches.account_cache.insert(*address, res);
        }
        Ok(res)
    }
}
``` 
Cache misses are only inserted when `self.is_prewarm()` is true (line 150), so prewarming populates the cache and execution trusts it blindly.

### our workaround
We modified `basic_account` (and the storage equivalent) to validate cache hits against the provider during execution:
```rust
fn basic_account(&self, address: &Address) -> ProviderResult<Option<Account>> {
    if let Some(cached) = self.caches.account_cache.get(address) {
        if !self.prewarm {
            let actual = self.state_provider.basic_account(address)?;
            if cached != actual {
                self.caches.account_cache.insert(*address, actual);
                return Ok(actual)
            }
        }
        return Ok(cached)
    }
    // etc
}
``` 
It works and we haven't had a mismatch since, but it kills the cache benefit during execution since every hit does a provider read anyway. Not great but it keeps us running.

### what I'm hoping to understand
Is my analysis of the race correct, or am I missing something about how the overlay/persistence interaction is supposed to prevent this? The `OverlayStateProviderFactory` caching from #19752 covers the in-memory overlay, but if the divergence happens at the MDBX layer below the overlay (which is what I think is happening), it wouldn't help.
If the race is real, what's the right way to fix this? I have some ideas but I don't want to go build something that fights the architecture. Happy to put together a PR if someone can point me in the right direction.

### Steps to reproduce

not reliably reproducible on demand. It depends on the persistence thread committing at the exact moment between the execution and prewarming `.build()` calls.

**what we observed:**

1.) synced a full node on a chain with ~10s block times

2.) once the node caught up to tip and started processing live blocks with prewarming enabled, we hit gas mismatch errors on `newPayload`

3.) it only happened once during catchup, likely because the persistence thread was more active while processing a backlog of blocks in quick succession

4.) after applying the workaround (validating cache hits against the provider), we haven't seen it since

### Node logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

reth v1.10.2

2

N/a

Full via --full flag

_No response_

_No response_

### Code of Conduct

0xMars42:
## Summary

Pin the static file tip on first `StateProviderBuilder::build()` call, reject subsequent builds if persistence advanced. Prevents prewarming workers from poisoning the shared `ExecutionCache` with data from a newer MDBX snapshot.

Also replaces `.expect()` on the multiproof task's `build()` with graceful error handling (was a panic waiting to happen regardless).

## Problem

`StateProviderBuilder.build()` opens a fresh MDBX read transaction each call. The builder is cloned to prewarm workers and the multiproof task, which each call `build()` on their own thread. If the persistence thread commits between the execution build (`payload_validator.rs:376`) and any worker build (`prewarm.rs:522/732`, `mod.rs:308`), the worker sees newer state and caches it. Execution then reads those values from the shared cache instead of from its own provider.

Reported as sporadic gas mismatch errors on `newPayload` with ~10s block times.

## Fix

Add `tip: Arc<OnceLock<(B256, u64)>>` to `StateProviderBuilder`. First `build()` pins the static file tip (hash + block number via `sealed_header`). Later builds check it hasn't changed, if it has, return `ConsistentViewError::Reorged`. All prewarm call sites already handle `build()` errors gracefully (log + skip), so the cache stays clean and execution falls through to its own provider.

Static files commit before MDBX (`consistent_view.rs` documents this), so the check is conservative: if static files haven't moved, MDBX hasn't either. Same detection pattern as `ConsistentDbView`.

Fixes #22159

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -29,10 +29,10 @@ use reth_payload_primitives::{
 };
 use reth_primitives_traits::{NodePrimitives, RecoveredBlock, SealedBlock, SealedHeader};
 use reth_provider::{
-    BlockExecutionOutput, BlockExecutionResult, BlockReader, ChangeSetReader,
-    DatabaseProviderFactory, HashedPostStateProvider, ProviderError, StageCheckpointReader,
-    StateProviderBox, StateProviderFactory, StateReader, StorageChangeSetReader,
-    StorageSettingsCache, TransactionVariant,
+    providers::ConsistentViewError, BlockExecutionOutput, BlockExecutionResult, BlockReader,
+    ChangeSetReader, DatabaseProviderFactory, HashedPostStateProvider, ProviderError,
+    StageCheckpointReader, StateProviderBox, StateProviderFactory, StateReader,
+    StorageChangeSetReader, StorageSettingsCache, TransactionVariant,
 };
 use reth_revm::database::StateProviderDatabase;
 use reth_stages_api::ControlFlow;
@@ -40,7 +40,12 @@ use reth_tasks::spawn_os_thread;
 use reth_trie_db::ChangesetCache;
 use revm::interpreter::debug_unreachable;
 use state::TreeState;
-use std::{fmt::Debug, ops, sync::Arc, time::Instant};
+use std::{
+    fmt::Debug,
+    ops,
+    sync::{Arc, OnceLock},
+    time::Instant,
+};
 
 use crossbeam_channel::{Receiver, Sender};
 use tokio::sync::{
@@ -93,6 +98,11 @@ pub(crate) const MIN_BLOCKS_FOR_PIPELINE_RUN: u64 = EPOCH_SLOTS;
 const CHANGESET_CACHE_RETENTION_BLOCKS: u64 = 64;
 
 /// A builder for creating state providers that can be used across threads.
+///
+/// The builder tracks the database tip at the time of the first [`Self::build`] call and rejects
+/// subsequent builds if persistence has advanced. This prevents prewarming workers from opening a
+/// newer MDBX snapshot than the execution thread, which would poison the shared
+/// [`ExecutionCache`] with stale data.
 #[derive(Clone, Debug)]
 pub struct StateProviderBuilder<N: NodePrimitives, P> {
     /// The provider factory used to create providers.
@@ -101,17 +111,21 @@ pub struct StateProviderBuilder<N: NodePrimitives, P> {
     historical: B256,
     /// The blocks that form the chain from historical to target and are in memory.
     overlay: Option<Vec<ExecutedBlock<N>>>,
+    /// Pinned tip `(hash, block_number)` from the first [`Self::build`] call. Shared across
+    /// clones via [`Arc`] so that prewarm workers detect if persistence committed between the
+    /// execution build and their own build, preventing cache poisoning from snapshot divergence.
+    tip: Arc<OnceLock<(B256, u64)>>,
 }
 
 impl<N: NodePrimitives, P> StateProviderBuilder<N, P> {
     /// Creates a new state provider from the provider factory, historical block hash and optional
     /// overlaid blocks.
-    pub const fn new(
+    pub fn new(
         provider_factory: P,
         historical: B256,
         overlay: Option<Vec<ExecutedBlock<N>>>,
     ) -> Self {
-        Self { provider_factory, historical, overlay }
+        Self { provider_factory, historical, overlay, tip: Arc::new(OnceLock::new()) }
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -120,7 +134,44 @@ where
...
+    /// committed before the database, so a change in the static file tip is a conservative
+    /// indicator that the MDBX snapshot may have also advanced.
+    ///
+    /// [`ConsistentDbView`]: reth_provider::providers::ConsistentDbView
     pub fn build(&self) -> ProviderResult<StateProviderBox> {
+        // Pin-and-check: detect if persistence committed between builds.
+        // Both `last_block_number` and `sealed_header` read from static files, which are
+        // committed before MDBX. If these haven't changed, MDBX hasn't advanced either.
+        let current_num = self.provider_factory.last_block_number()?;
+        let current_hash = match self.provider_factory.sealed_header(current_num)? {
```
mattsse, line 153: unsure if this is correct

this is definitely not atomic and doesnt not look up header by hash

mattsse (changes requested):
unclear what this fixes really

dont have context re looking up history state @joshieDo

crates/engine/tree/src/tree/mod.rs
0xMars42: You're right it's not atomic I based this on ConsistentDbView which does the same two separate reads in new_with_latest_tip(), and a similar check in provider_ro(). My understanding is that static files are append-only so sealed_header(N) always returns the same result even if persistence appends N+1 between the two calls, which is why ConsistentDbView doesn't need atomicity either.

For the hash lookup  I avoided it because of the race condition documented in provider_ro() (L52-L69): hash-to-number indexes are in MDBX which commits after static files, so fetching by hash during persistence could hit uncommitted state.

Happy to rework if I'm wrong on either of these assumptions, and definitely interested in @joshieDo's perspective.

PR #22213, opened 2026-02-16
fix(net): fix flaky test_trusted_peer_only test

gakonst:
## Summary
Fix flaky `test_trusted_peer_only` integration test.

## Motivation
The test intermittently fails at the final `assert_eq!(handle.num_connected_peers(), 2)` with count 1 instead of 2.

## Root Cause
After `handle1.add_peer(handle)` is rejected (handle only accepts trusted peers), handle1 still has handle in its peer list and retries connecting. When `handle.add_trusted_peer(handle1)` is called, both sides attempt connections simultaneously — handle's outgoing to handle1, and handle1's retry to handle. This creates duplicate sessions that get resolved by dropping one, leaving the peer count at 1.

## Fix
Call `handle1.remove_peer()` after confirming the untrusted connection was rejected, before initiating the trusted connection. This eliminates the competing connection attempt.

Prompted by: DaniPopes

### DIFF
--- a/crates/net/network/tests/it/connect.rs
+++ b/crates/net/network/tests/it/connect.rs
@@ -12,7 +12,7 @@ use reth_network::{
 };
 use reth_network_api::{
     events::{PeerEvent, SessionInfo},
-    NetworkInfo, Peers, PeersInfo,
+    NetworkInfo, PeerKind, Peers, PeersInfo,
 };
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

DaniPopes:
cc @mattsse

PR #22219, opened 2026-02-16
perf: use synchronous state root for small blocks

yongkangc:
**Summary**
Reduce newPayload latency for small blocks by avoiding parallel state root tasks when transaction counts are low. This targets fixed overhead from sparse trie task setup that dominates sub-50-tx payloads.

**Changes**
- Choose synchronous state root computation for blocks with <=50 transactions
- Leave config-driven state root strategies unchanged for larger blocks

**Expected Impact**
Lower fixed per-block overhead for small payloads by skipping sparse trie task setup and waiting.


yongkangc:
Closing in favor of #22129 which already addresses this fast-path.

PR #22222, opened 2026-02-16
fix(net): use test backoff durations in Testnet PeerConfig

gakonst:
## Summary
Fix flaky `test_reconnect_trusted` by using `PeersConfig::test()` in `PeerConfig::network_config_builder`.

## Motivation
`Testnet::create` used default `PeersConfig` with a 30s low backoff for trusted peer reconnection, but `test_reconnect_trusted` only waits 10s — a race that fails under CI load.

## Changes
- Add `.peer_config(PeersConfig::test())` to `PeerConfig::network_config_builder`, giving all `Testnet::create` peers 200ms backoff durations instead of 30s

## Testing
`cargo check -p reth-network --tests`

Prompted by: DaniPopes

### DIFF
--- a/crates/net/network/src/test_utils/testnet.rs
+++ b/crates/net/network/src/test_utils/testnet.rs
@@ -10,7 +10,7 @@ use crate::{
         policy::NetworkPolicies,
         TransactionsHandle, TransactionsManager, TransactionsManagerConfig,
     },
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

DaniPopes:
cc @mattsse

PR #22226, opened 2026-02-16
perf(trie): reduce nibble comparisons in prune scan

yongkangc:
## Summary
- replace per-subtrie partition_point search with a single linear scan in prune roots
- reduces repeated nibble comparisons in the prune pass

## Performance
- targets nibble/partition_point hotspots seen in profile 40dZkvo

## Testing
- cargo +nightly fmt --all


yongkangc:
Closing per request; cache path already filters targets upstream so this optimization is unnecessary.

PR #22228, opened 2026-02-16
perf: reduce update_leaves key cloning

yongkangc:
**Summary**
Reduce key cloning in sparse trie update processing by draining the updates map instead of collecting keys, keeping cache-path leaf updates lighter while preserving the map's allocation.

**Changes**
- Replace key collection + per-entry remove with `updates.drain()` to iterate entries directly
- Preserve update map capacity by avoiding `mem::take` reallocation

**Expected Impact**
Lower per-update allocation and key clone overhead during sparse trie cache updates.


DaniPopes (approved):
good catch, can we do this anywhere else?

PR #22241, opened 2026-02-16
fix: Use `Option::transpose` in `StaticFileCursor` to remove duplicated Option/Result plumbing

prestoalvarez:
Refactor `StaticFileCursor::{get_one, get_two, get_three}` to use `Option::map(...).transpose()?` instead of three hand-written `match` blocks converting `Option<Row>` and fallible `Decompress` calls into `ProviderResult<Option<_>>`.


mattsse (commented):
I find what we have rn easier to look at because that doesnt require the result hint

PR #22242, opened 2026-02-16
feat(re-execute): work-stealing parallelization

gakonst:
## Summary

Replace static block range partitioning with work-stealing so workers never sit idle.

## Motivation

Previously, the total block range was divided equally among workers upfront. If one worker finished its chunk early (e.g., lighter blocks with fewer transactions), it would idle until all others finished. This wastes CPU.

## Changes

- Replace per-worker `start_block..end_block` ranges with a shared `AtomicU64` counter
- Workers atomically grab the next chunk of blocks via `fetch_add`, process them, then loop for more
- Each chunk gets a fresh executor (blocks aren't contiguous per-worker anyway since workers finish at different rates)
- Add `--blocks-per-worker` flag (default: 1000) to control chunk size
- Remove `blocks_per_task` static partitioning

## Testing

Compiles clean, passes clippy. Functional behavior unchanged — same validation logic, same stats reporting.

Prompted by: arsenii

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -50,8 +50,8 @@ pub struct Command<C: ChainSpecParser> {
     num_tasks: Option<u64>,
 
     /// Number of blocks each worker processes before grabbing the next chunk.
-    #[arg(long, default_value = "1000")]
-    blocks_per_worker: u64,
+    #[arg(long, default_value = "5000")]
+    blocks_per_chunk: u64,
 
     /// Continues with execution when an invalid block is encountered and collects these blocks.
     #[arg(long)]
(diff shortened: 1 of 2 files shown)


shekhirin (approved):

crates/cli/commands/src/re_execute.rs
```diff
@@ -46,6 +49,10 @@ pub struct Command<C: ChainSpecParser> {
     #[arg(long)]
     num_tasks: Option<u64>,
 
+    /// Number of blocks each worker processes before grabbing the next chunk.
+    #[arg(long, default_value = "5000")]
```
mattsse, line 53: ca use default_value_t here

mattsse (approved):

PR #22243, opened 2026-02-16
fix(txpool): notify subscribers when set_block_info promotes transaction

gancerlory:
`set_block_info` passes empty closures to `update_basefee`/`update_blob_fee`, so transactions promoted due to fee changes are not notified to subscribers. This affects node startup and deep sync skip scenarios.

Now collects promoted transactions and calls `notify_on_transaction_updates`, consistent with `on_canonical_state_change`.

### DIFF
--- a/crates/transaction-pool/src/pool/mod.rs
+++ b/crates/transaction-pool/src/pool/mod.rs
@@ -204,9 +204,15 @@ where
     pub fn block_info(&self) -> BlockInfo {
         self.get_pool_data().block_info()
     }
-    /// Sets the currently tracked block
+    /// Sets the currently tracked block.
+    ///
+    /// This will also notify subscribers about any transactions that were promoted to the pending
+    /// pool due to fee changes.
     pub fn set_block_info(&self, info: BlockInfo) {
-        self.pool.write().set_block_info(info)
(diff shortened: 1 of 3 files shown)


mattsse (commented):
Thanks for the fix! The core idea is correct — `set_block_info` was silently dropping promoted transaction notifications which matters at startup and during deep sync skips.

Two things to address:

1. **Breaking visibility change**: `TxPool::set_block_info` was `pub` and is now `pub(crate)`. Since `TxPool` is a `pub struct` in `pub mod txpool`, this is a breaking change for downstream consumers. Please keep it `pub`.

2. **Discarded transactions are never populated**: `outcome.discarded` is always empty because neither `update_basefee` nor `update_blob_fee` collect demoted transactions. When the basefee *increases*, `update_basefee` moves transactions from pending → basefee/queued pools, but these demotions aren't captured. The call to `notify_on_transaction_updates(outcome.promoted, outcome.discarded)` will therefore never notify about discarded txs. This may be intentional (since `set_block_info` is only called at startup/sync-skip where there are no prior subscribers), but if so it should be documented.

mattsse (approved):
makes sense, ty!

PR #22245, opened 2026-02-16
feat(stages): support deferred history indexing in pipeline

gakonst:
## Summary

This PR turns history indexing from a startup blocker into a background catch-up path. With `deferred_history_indexing` enabled, the node can start following head without waiting for `TransactionLookup`, `IndexStorageHistory`, and `IndexAccountHistory` to finish.

The main design change is that deferred indexing runs inside the persistence service thread.

## Operator experience

Operators can now bootstrap from snapshots that do not include RocksDB history indices and still enter live sync quickly. While the node is catching up on deferred indexing, historical RPC coverage improves progressively as stage checkpoints advance. Once catch-up is complete, writes and RPC behavior converge to the same fully-indexed state as before.

## Important details

Deferred stages run in bounded round-robin batches, with block persistence always taking priority so tip-following remains responsive. Reorgs force the indexer back into catch-up mode. Startup consistency checks explicitly recognize deferred-catch-up mode so expected lag after a snapshot download does not trigger healing.

If transaction lookup is fully pruned, tx-hash healing is skipped.

## Operator notes

- `deferred_history_indexing` now defaults to `true`.
- Set it to `false` if you need the previous "fully indexed before live sync" behavior, such as deterministic import/test workflows.

## Testing focus

Coverage focuses on save-mode checkpoint semantics, deferred catch-up and reorg handling, RocksDB consistency behavior in deferred/pruned modes, and idempotent account/storage history shard appends.


crates/engine/tree/src/persistence.rs
```diff
@@ -62,6 +71,8 @@ where
         incoming: Receiver<PersistenceAction<N::Primitives>>,
         pruner: PrunerWithFactory<ProviderFactory<N>>,
         sync_metrics_tx: MetricEventsSender,
+        deferred_indexer_config: Option<DeferredHistoryIndexerConfig>,
+        deferred_indexing_enabled: bool,
```
joshieDo, line 75: is it possible to be true and none ? if not should we just collapse this into DeferredHistoryIndexerConfig { enabled: bool, ... }

crates/stages/stages/src/sets.rs
```diff
@@ -347,10 +347,11 @@ where
         .add_set(HistoryIndexingStages {
             stages_config: self.stages_config.clone(),
             prune_modes: self.prune_modes.clone(),
-        })
+        });
+
         // Prune stage should be added after all hashing stages, because otherwise it will
         // delete
-        .add_stage(PruneStage::new(
+        builder.add_stage(PruneStage::new(
```
joshieDo, line 354: necessary?

crates/stages/stages/src/stages/tx_lookup.rs
```diff
@@ -54,7 +60,18 @@ impl TransactionLookupStage {
         etl_config: EtlConfig,
         prune_mode: Option<PruneMode>,
     ) -> Self {
-        Self { chunk_size: config.chunk_size, etl_config, prune_mode }
+        Self { chunk_size: config.chunk_size, etl_config, prune_mode, log_progress: true }
+    }
+
+    /// Enables or disables progress/info logging for this stage execution.
+    pub const fn with_progress_logging(mut self, log_progress: bool) -> Self {
```
joshieDo, line 67: why not always true during the stage execution?

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1552,6 +1562,16 @@ impl<'a> RocksDBBatch<'a> {
...
+        // Make appends idempotent for restart/replay scenarios: if the last shard already contains
+        // newer indices (e.g. after a crash between RocksDB and MDBX checkpoint commit), keep only
+        // strictly newer input values.
+        if let Some(current_max) = last_shard.iter().next_back() {
+            indices.retain(|index| *index > current_max);
+            if indices.is_empty() {
+                return Ok(());
+            }
+        }
+
```
joshieDo, line 1574: not entirely sure i understand why this is here - healing should have fix any restart/crash issue no ?

joshieDo: same applies for others

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -361,6 +406,26 @@ impl RocksDBProvider {
...
 
+        let sf_tip = provider
+            .static_file_provider()
+            .get_highest_static_file_block(StaticFileSegment::AccountChangeSets)
+            .unwrap_or(0);
+
+        // During deferred indexing, `Finish` is already at tip while history stages trail.
+        // In that mode, startup-time healing scans are unnecessary and expensive.
+        let finish_checkpoint =
+            provider.get_stage_checkpoint(StageId::Finish)?.map(|cp| cp.block_number).unwrap_or(0);
```
joshieDo, line 417: hm i understand now, but this makes me nervous ngl

crates/storage/provider/src/providers/rocksdb/provider.rs
joshieDo: ok we're adding this because during the deferring period we skip the healing inconsistencies, can we document it here

yongkangc (commented):
Should we consider adding metrics for status on deferred indexing?

For both users and debugging purposes 

Maybe
• Deferred indexer lag (blocks behind tip)
• Batch processing time per tick
• Catch-up completion event

PR #22246, opened 2026-02-16
feat(download): modular snapshot downloads with interactive TUI and config generation

gakonst:
## Summary

Redesigns `reth download` into a modular, component-based snapshot system. Users select what they need via an interactive TUI or CLI flags. Downloads run in parallel with resume support, and a matching `reth.toml` is auto-generated with the correct pruning config. Also adds a new `reth snapshot-manifest` command for manifest generation.

## Motivation

Full archive snapshots are massive. Most users only need state + headers. This makes downloads modular — pick exactly the components you want, download them in parallel, and get a pruning config that matches.

## Components

| Component | Manifest Key | Type | Required | Minimal | Description |
|---|---|---|---|---|---|
| State | `state` | Single archive | Yes | Yes | mdbx database |
| Headers | `headers` | Chunked | Yes | Yes | Block header static files |
| Transactions | `transactions` | Chunked | No | Yes | Transaction static files |
| Receipts | `receipts` | Chunked | No | No | Receipt static files |
| Account Changesets | `account_changesets` | Chunked | No | Yes | Account changeset static files |
| Storage Changesets | `storage_changesets` | Chunked | No | Yes | Storage changeset static files |
| RocksDB Indices | `rocksdb_indices` | Single archive | No | No | Archive node RocksDB indices (transaction lookup, account/storage history). Auto-included when all data components are set to `All` (archive selection). Skip with `--without-rocksdb`. |

Chunked components are split by block range (`blocks_per_file`, default 500k) matching reth static file segments, enabling parallel downloads and granular resume. The TUI bundles account + storage changesets into a single "State History" row.

**RocksDB indices** are a hidden archive-only component — they're automatically injected when all data components (transactions, receipts, account changesets, storage changesets) are selected as `All`. This includes `--archive`/`--all` and the TUI archive preset (`a`). When RocksDB indices are downloaded, index stage checkpoints are preserved from the source node so the pipeline doesn't rebuild them. When they're not downloaded, `TransactionLookup`, `IndexAccountHistory`, and `IndexStorageHistory` stage checkpoints are reset to block 0 so the pipeline rebuilds indices from scratch. Use `--without-rocksdb` to skip RocksDB indices even for archive selections.

### Downloading

```bash
# Interactive TUI — select components with arrow keys, cycle through None/Last N/All
reth download

# Non-interactive, minimal set (state + headers + txs + changesets at distance 10064)
reth download -y

# State + headers + transactions + receipts (parallel download)
reth download --with-txs --with-receipts

# Include state history (account + storage changesets)
reth download --with-state-history

# Full archive — all components
reth download --all

# Custom manifest URL
reth download --manifest-url https://example.com/manifest.json

# Skip reth.toml generation
reth download --no-config

# Legacy single-archive mode (unchanged)
reth download --url https://example.com/snapshot.tar.zst
```

### Generating a manifest

```bash
reth snapshot-manifest \
  -d ./archives \
  --base-url https://snapshots.reth.rs/1 \
  --block 24500000 \
  --chain-id 1 \
  --blocks-per-file 500000 \
  -o manifest.json
```

Scans the archive directory for files matching `state.tar.zst` and `{component}-{start}-{end}.tar.zst`, computes per-chunk sizes, and writes the manifest.

## Manifest format

```json
{
  "block": 24500000,
  "chain_id": 1,
  "storage_version": 2,
  "timestamp": 1700000000,
  "base_url": "https://snapshots.reth.rs/1",
  "components": {
    "state": {
      "file": "state.tar.zst",
      "size": 451000000000
    },
    "headers": {
      "blocks_per_file": 500000,
      "total_blocks": 24500000,
      "chunk_sizes": [40000000, 42000000, ...]
    },
    "transactions": {
      "blocks_per_file": 500000,
      "total_blocks": 24500000,
      "chunk_sizes": [16000000000, 18000000000, ...]
    }
  }
}
```

- **Single archives** (`state`): `file` (relative to `base_url`) + `size` in bytes
- **Chunked archives**: `blocks_per_file` + `total_blocks` + per-chunk `chunk_sizes` array. Archive filenames follow `{key}-{start}-{end}.tar.zst`

## Config generation

After download, a `reth.toml` is generated with pruning config matching what was downloaded. `sender_recovery` and `transaction_lookup` are always pruned full (non-archive). Prune checkpoints are written to the database so the pruner knows data up to the snapshot block is already in the expected state.

| Missing Component | Prune Segments |
|---|---|
| Transactions | `bodies_history = "full"` |
| Receipts | `receipts = { distance = 64 }` |
| Changesets | `account_history = { distance = 10064 }`, `storage_history = { distance = 10064 }` |

When all data components are present → archive node, no pruning.

The TUI also supports partial distance selections (Last 10064 / Last 100k / Last 1M) which map to `PruneMode::Distance(N)`.

## Post-download: index stage checkpoint reset

The snapshot mdbx comes from a fully synced node, so `TransactionLookup`, `IndexAccountHistory`, and `IndexStorageHistory` stage checkpoints are at the tip. Since we don't distribute the rocksdb indices, these checkpoints must be reset to block 0 so the pipeline (or background indexer) rebuilds the indices from scratch. The corresponding prune checkpoints (`TransactionLookup`, `AccountHistory`, `StorageHistory`) are also cleared to avoid inheriting state from the source node.

## Interactive TUI

The TUI lets users cycle through selection presets per component:

- **None** → don't download
- **Last 10,064** / **Last 100,000** / **Last 1,000,000** → partial download (tail chunks only)
- **All** → full download

Keyboard: `←/→` cycle selection, `m` minimal preset, `a` all preset, `Enter` confirm, `Esc` cancel.

## Download behavior

- Parallel downloads (4 concurrent via `buffer_unordered`)
- Resumable downloads with HTTP Range requests (`.part` files)
- Up to 10 retries with 5s backoff
- Supports both `.tar.lz4` and `.tar.zst` compression
- `file://` URLs for local extraction
- Aggregated progress display for parallel downloads (every 3s via tracing)

## Testing

```bash
cargo test -p reth-cli-commands --lib download
cargo check -p reth-ethereum-cli
```

Prompted by: dcline

### DIFF
--- a/crates/cli/commands/src/download/config_gen.rs
+++ b/crates/cli/commands/src/download/config_gen.rs
@@ -4,12 +4,20 @@ use reth_prune_types::PruneMode;
 use std::path::Path;
 use tracing::info;
 
+/// Minimum blocks to keep for receipts, matching `--minimal` prune settings.
+const MINIMUM_RECEIPTS_DISTANCE: u64 = 64;
+
+/// Minimum blocks to keep for history/bodies, matching `--minimal` prune settings
+/// (`MINIMUM_UNWIND_SAFE_DISTANCE`).
+const MINIMUM_HISTORY_DISTANCE: u64 = 10064;
+
 /// Generates an appropriate [`Config`] based on which snapshot components were downloaded.
 ///
-/// If a data category wasn't downloaded, the corresponding prune segments are set to avoid
-/// indexing or serving that data. Transaction lookup and sender recovery are pruned whenever
-/// transaction bodies are not present since the lookup index is useless without the underlying
-/// data.
+/// The generated prune config mirrors reth's `--minimal` prune settings for components
+/// that weren't downloaded:
+/// - Missing transactions: prune sender recovery, transaction lookup, and bodies history
+/// - Missing receipts: prune receipts to last 64 blocks
+/// - Missing changesets: prune account/storage history to last 10,064 blocks
 pub fn config_for_components(selected: &[SnapshotComponentType]) -> Config {
     let has_txs = selected.contains(&SnapshotComponentType::Transactions);
     let has_receipts = selected.contains(&SnapshotComponentType::Receipts);
@@ -25,15 +33,16 @@ pub fn config_for_components(selected: &[SnapshotComponentType]) -> Config {
         if !has_txs {
             prune.segments.transaction_lookup = Some(PruneMode::Full);
             prune.segments.sender_recovery = Some(PruneMode::Full);
+            prune.segments.bodies_history = Some(PruneMode::Distance(MINIMUM_HISTORY_DISTANCE));
         }
 
         if !has_receipts {
-            prune.segments.receipts = Some(PruneMode::Full);
+            prune.segments.receipts = Some(PruneMode::Distance(MINIMUM_RECEIPTS_DISTANCE));
         }
 
         if !has_changesets {
-            prune.segments.account_history = Some(PruneMode::Distance(10064));
-            prune.segments.storage_history = Some(PruneMode::Distance(10064));
+            prune.segments.account_history = Some(PruneMode::Distance(MINIMUM_HISTORY_DISTANCE));
+            prune.segments.storage_history = Some(PruneMode::Distance(MINIMUM_HISTORY_DISTANCE));
         }
 
         config.prune = prune;
@@ -86,15 +95,16 @@ pub fn describe_prune_config(selected: &[SnapshotComponentType]) -> Vec<String> 
     if !has_txs {
         lines.push("transaction_lookup = \"full\"".to_string());
         lines.push("sender_recovery = \"full\"".to_string());
+        lines.push(format!("bodies_history = {{ distance = {MINIMUM_HISTORY_DISTANCE} }}"));
     }
 
     if !has_receipts {
-        lines.push("receipts = \"full\"".to_string());
+        lines.push(format!("receipts = {{ distance = {MINIMUM_RECEIPTS_DISTANCE} }}"));
     }
 
     if !has_changesets {
-        lines.push("account_history = { distance = 10064 }".to_string());
-        lines.push("storage_history = { distance = 10064 }".to_string());
+        lines.push(format!("account_history = {{ distance = {MINIMUM_HISTORY_DISTANCE} }}"));
+        lines.push(format!("storage_history = {{ distance = {MINIMUM_HISTORY_DISTANCE} }}"));
     }
 
     lines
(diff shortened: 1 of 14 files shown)


crates/cli/commands/src/download/mod.rs
```diff
@@ -0,0 +1,1230 @@
...
+        // snapshot block is already in the expected pruned state
+        if config.prune.segments != Default::default() {
+            write_prune_checkpoints(&db, &config, manifest.block)?;
+        }
+
+        // Reset stage checkpoints for history indexing stages. The snapshot mdbx
+        // comes from a fully synced node with these at the tip, but we don't
+        // distribute the rocksdb indices. Resetting to block 0 ensures the
+        // pipeline or background indexer rebuilds them.
+        reset_index_stage_checkpoints(&db)?;
```
joshieDo: should we pass tx_mut instead so it gets committed at same time with write_prune_checkpoints

crates/cli/commands/src/download/config_gen.rs
```diff
@@ -0,0 +1,725 @@
...
+) -> eyre::Result<()> {
+    let segments = &config.prune.segments;
+
+    // Collect (segment, mode) pairs for all configured prune segments
+    let checkpoints: Vec<(PruneSegment, PruneMode)> = [
+        (PruneSegment::SenderRecovery, segments.sender_recovery),
+        (PruneSegment::TransactionLookup, segments.transaction_lookup),
+        (PruneSegment::Receipts, segments.receipts),
+        (PruneSegment::AccountHistory, segments.account_history),
+        (PruneSegment::StorageHistory, segments.storage_history),
```
joshieDo, line 64: should this be changesets?

crates/cli/commands/src/download/mod.rs
```diff
@@ -0,0 +1,1230 @@
...
+            let desc = config_gen::describe_prune_config_from_selections(&selections, &manifest);
+            info!(target: "reth::cli", "{}", desc.join(", "));
+        }
+
+        // Open the DB to write checkpoints
+        let db_path = data_dir.db();
+        let db = init_db(&db_path, self.env.db.database_args())?;
+
+        // Write prune checkpoints to the DB so the pruner knows data before the
+        // snapshot block is already in the expected pruned state
```
joshieDo, line 411: confirming: this is true because when we look at the manifest we download only what we need

crates/cli/commands/src/download/config_gen.rs
joshieDo: not a big deal i think, because we wont have them anyway, so the first run will essentially bump the prune checkpoint

crates/cli/commands/src/download/config_gen.rs
```diff
@@ -0,0 +1,725 @@
...
+    db: &DatabaseEnv,
+    config: &Config,
+    snapshot_block: u64,
+) -> eyre::Result<()> {
+    let segments = &config.prune.segments;
+
+    // Collect (segment, mode) pairs for all configured prune segments
+    let checkpoints: Vec<(PruneSegment, PruneMode)> = [
+        (PruneSegment::SenderRecovery, segments.sender_recovery),
+        (PruneSegment::TransactionLookup, segments.transaction_lookup),
```
joshieDo, line 61: we delete its prune checkpoint after on reset_* right?

crates/cli/commands/src/download/config_gen.rs
Rjected: `PruneSegment::AccountHistory` covers account history + changesets, and `PruneSegment::StorageHistory` storage history + changesets

crates/cli/commands/src/download/config_gen.rs
Rjected: yep https://github.com/paradigmxyz/reth/blob/1f34a16d54482421584ac91571c9122f4c2cb903/crates/cli/commands/src/download/config_gen.rs#L140

crates/cli/commands/src/download/mod.rs
Rjected: yes

joshieDo (approved):

yongkangc (approved):
Lgtm

shekhirin (approved):

PR #22247, opened 2026-02-16
refactor(net/tx): rewrite TransactionFetcher as two-stage pipeline

figtracer:
Rewrites `TransactionFetcher` from scratch as a **two-stage pipeline** (QUEUE → FETCH), replacing the previous flat two-state model.

Propagation improvements (deterministic peer selection via `BroadcastChoice`, `TxPeerTracking`, full broadcast size gate) have been split into a separate PR: https://github.com/paradigmxyz/reth/pull/22437

## Problem

The previous `TransactionFetcher` used a flat two-state model (pending fetch vs inflight) with several limitations:

- **Budget-based LRU scanning.** Finding hashes for a given peer required scanning an LRU cache with a search budget — O(n) per poll cycle, and hashes could be skipped entirely if the budget ran out.
- **Single response per poll.** The `Stream` implementation processed only one inflight response per `poll_next` call, creating a bottleneck under load.
- **No fetch timeout.** Inflight requests could hang indefinitely with no mechanism to reclaim hashes and retry with alternate peers.
- **No per-peer DoS caps.** A peer could flood announcements without limit.
- **Retry/fallback complexity.** Multiple constants and budget-constrained searches for retries, fallback peers, and intersection finding added complexity with marginal benefit.

## Solution

Two-stage pipeline modeled after geth's `TxFetcher`:

- **Stage 1 (QUEUE):** Hash is ready to fetch. `schedule_fetches()` picks idle peers via a per-peer index (`announces: HashMap<PeerId, HashMap<TxHash, Metadata>>`) — O(1) lookup instead of budget scanning. Hashes are sorted by sequence number for FIFO ordering.
- **Stage 2 (FETCH):** Hash is actively being fetched. Includes 5s fetch timeout with dangling request tracking (timed-out peer's slot stays blocked until response or sweep at 4× timeout). Alternate peers are tracked for retry on failure.

## Key Design Decisions

- **Two-phase commit in `schedule_fetches`:** The request is sent to the peer *before* mutating stage maps. If the channel is full, the hash stays in Stage 1 with no state corruption.
- **`announces` as superset index:** The per-peer `announces` map is kept across both stages — entries are only removed when a hash leaves the system entirely (fetched, received via broadcast, or peer disconnects). This avoids losing metadata during stage transitions.
- **Cutoff heuristic for truncated responses:** When a peer returns only some of the requested hashes, hashes before the last delivered one are treated as intentionally skipped (peer removed from their candidate set). Hashes after the cutoff are treated as benign truncation (size limit).
- **Underpriced import cache:** Transactions rejected as underpriced are cached separately from bad imports. Peers are not penalized for underpriced transactions since pricing is a local/dynamic condition.

### DIFF
--- a/crates/net/network/src/metrics.rs
+++ b/crates/net/network/src/metrics.rs
@@ -178,10 +178,6 @@ pub struct TransactionsManagerMetrics {
     /* ================ POOL IMPORTS ================ */
     /// Number of transactions about to be imported into the pool.
     pub(crate) pending_pool_imports: Gauge,
-    /// Total number of bad imports, imports that fail because the transaction is badly formed
-    /// (i.e. have no chance of passing validation, unlike imports that fail due to e.g. nonce
-    /// gaps).
-    pub(crate) bad_imports: Counter,
     /// Number of inflight requests at which the
     /// [`TransactionPool`](reth_transaction_pool::TransactionPool) is considered to be at
     /// capacity. Note, this is not a limit to the number of inflight requests, but a health
@@ -245,6 +241,9 @@ pub struct TransactionsManagerMetrics {
     ///
     /// Duration in seconds.
     pub(crate) acc_duration_poll_commands: Gauge,
+    /// Number of entries in the `transactions_by_peers` map, tracking pending pool imports
+    /// attributed to their originating peers.
+    pub(crate) transactions_by_peers_len: Gauge,
 }
 
 /// Metrics for the [`TransactionsManager`](crate::transactions::TransactionsManager).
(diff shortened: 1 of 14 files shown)


crates/net/network/src/transactions/constants.rs
```diff
@@ -53,10 +53,19 @@ pub mod tx_manager {
     ///
     /// Default is 100 KiB, i.e. 3 200 transaction hashes.
     pub const DEFAULT_MAX_COUNT_BAD_IMPORTS: u32 = 100 * 1024 / 32;
+
+    /// Default limit for number of underpriced transaction hashes to cache.
+    /// Transactions rejected as underpriced are cached to avoid re-fetching on subsequent
+    /// announcements.
+    ///
+    /// Default is 32 768 transaction hashes.
+    pub const DEFAULT_MAX_COUNT_UNDERPRICED_IMPORTS: u32 = 32_768;
```
figtracer: think abt this value

crates/net/network/src/metrics.rs
```diff
@@ -259,31 +258,44 @@ pub struct TransactionFetcherMetrics {
...
+    /// Latency of dangling (timed-out) responses that eventually arrived, in seconds.
+    pub(crate) dangling_response_latency_seconds: Histogram,
+    /// Number of dangling responses that contained transactions.
+    pub(crate) dangling_completed_useful: Counter,
+    /// Number of dangling responses that were empty or errored.
+    pub(crate) dangling_completed_empty: Counter,
+    /// Broadcasts arriving for hashes in Stage 2 (after waitlist expired, before fetch).
+    pub(crate) late_broadcast_announced: Counter,
+    /// Broadcasts arriving for hashes in Stage 3 (during active fetch).
+    pub(crate) late_broadcast_inflight: Counter,
```
figtracer, line 298: i want to remove most of this post-testing

PR #22248, opened 2026-02-16
fix(rpc): enforce timeout option in debug trace methods

### LINKED ISSUE #22027: debug_traceCallMany: timeout option does not interrupt execution

jose-compu:
## Summary
- `debug_traceCallMany` (and `debug_traceCall`, `debug_traceTransaction`, `debug_traceBlock`) accepted the `timeout` field in `GethDebugTracingOptions` but never enforced it — `DebugInspector::new()` discards it with `..`, and no reth code checked elapsed time or deadline expiration
- Wraps blocking trace futures with `tokio::time::timeout`, returning `EthApiError::ExecutionTimedOut` when the configured timeout elapses (same pattern used by `sim_bundle`)
- Adds a Go-style duration string parser to handle the timeout format (`"5s"`, `"100ms"`, `"5000ms"` as produced by alloy's `with_timeout`)

## Test plan
- [x] 8 unit tests for `parse_go_duration` covering seconds, milliseconds, fractional, compound, nanosecond strings, all units, invalid inputs, whitespace
- [x] Round-trip test verifying parser handles alloy's `GethDebugTracingOptions::with_timeout` output format
- [x] Async test verifying `tokio::time::timeout` aborts a slow future
- [x] `cargo clippy -p reth-rpc` clean
- [x] `cargo test -p reth-rpc --lib debug::tests` — 11/11 passing

Closes #22027

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -40,9 +40,78 @@ use reth_trie_common::{updates::TrieUpdates, HashedPostState};
 use revm::DatabaseCommit;
 use revm_inspectors::tracing::{DebugInspector, TransactionContext};
 use serde::{Deserialize, Serialize};
-use std::{collections::VecDeque, sync::Arc};
+use std::{collections::VecDeque, sync::Arc, time::Duration};
 use tokio::sync::{AcquireError, OwnedSemaphorePermit};
 use tokio_stream::StreamExt;
+
+/// Parses a Go-style duration string into a [`Duration`].
+///
+/// Go duration strings consist of decimal numbers, each with an optional fractional part
+/// and a unit suffix. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+///
+/// Examples: "5s", "100ms", "1.5s", "2h45m", "300ms"
+///
+/// A pure numeric string is treated as nanoseconds.
+fn parse_go_duration(s: &str) -> Option<Duration> {
+    let s = s.trim();
+    if s.is_empty() {
+        return None;
+    }
+
+    // Pure number is treated as nanoseconds
+    if let Ok(nanos) = s.parse::<u64>() {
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -270,8 +353,16 @@ where
                     .map_err(Eth::Error::from_eth_err)?;
 
                 Ok(trace)
-            })
-            .await
+            });
+
+        if let Some(timeout_duration) = timeout {
+            match tokio::time::timeout(timeout_duration, trace_fut).await {
```
mattsse, line 359: this does not work, all this does is cancel the request side

PR #22273, opened 2026-02-17
fix: avoid redundant allocations when parsing ERA1 filenames

anim001k:
Avoid redundant `String` and `Vec` allocations when parsing ERA1 filenames in `read_dir`, keeping the existing filename semantics unchanged.


mattsse (commented):
I think we can keep as is

PR #22276, opened 2026-02-17
feat(rpc): update testing_buildBlockV1 to match latest spec

gakonst:
## Summary
Updates `testing_buildBlockV1` RPC to match the [latest execution-apis spec](https://github.com/ethereum/execution-apis/pull/747).

## Motivation
The spec changed to use positional params and nullable transactions (null = build from mempool). Changing the alloy `TestingBuildBlockRequestV1` type would be a breaking change, so we duplicate it locally in reth.

## Changes
- `crates/rpc/rpc-api/src/testing.rs`: Define local `TestingBuildBlockRequestV1` with `transactions: Option<Vec<Bytes>>` instead of re-exporting from alloy. RPC trait now uses 4 separate positional params.
- `crates/rpc/rpc/src/testing.rs`: Update impl to take separate params. When `transactions=null`, build block from mempool using `pool.best_transactions_with_attributes()` — same tx selection logic as pending block building (gas limit checks, blob gas, EIP-7934 size limits, nonce-too-low handling, private tx filtering).
- `crates/ethereum/node/src/node.rs`: Pass pool to `TestingApi::new()` (`TestingApi` now takes a `Pool` generic).
- All call sites (`e2e-test-utils`, `reth-bench`, integration tests) updated to pass positional params tuple and use `Some(vec![...])` for transactions.

## Testing
`cargo check` passes for all affected crates (`reth-rpc-api`, `reth-rpc`, `reth-node-ethereum --tests`, `reth-e2e-test-utils`, `reth-bench`).

Prompted by: mattsse

### DIFF
--- a/crates/rpc/rpc-api/src/testing.rs
+++ b/crates/rpc/rpc-api/src/testing.rs
@@ -5,10 +5,34 @@
 //! disabled by default and never be exposed on public-facing RPC without an
 //! explicit operator flag.
 
-use alloy_rpc_types_engine::ExecutionPayloadEnvelopeV5;
+use alloy_primitives::{Bytes, B256};
+use alloy_rpc_types_engine::{ExecutionPayloadEnvelopeV5, PayloadAttributes};
 use jsonrpsee::proc_macros::rpc;
 
-pub use alloy_rpc_types_engine::{TestingBuildBlockRequestV1, TESTING_BUILD_BLOCK_V1};
+pub use alloy_rpc_types_engine::TESTING_BUILD_BLOCK_V1;
+
+/// Request payload for `testing_buildBlockV1`.
+///
+/// This is a reth-local duplicate of the alloy type, updated to match the latest
+/// spec changes where `transactions` can be `null` (to build from mempool).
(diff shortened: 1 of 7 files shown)


crates/rpc/rpc-api/src/testing.rs
```diff
@@ -5,10 +5,34 @@
...
-pub use alloy_rpc_types_engine::{TestingBuildBlockRequestV1, TESTING_BUILD_BLOCK_V1};
+pub use alloy_rpc_types_engine::TESTING_BUILD_BLOCK_V1;
+
+/// Request payload for `testing_buildBlockV1`.
+///
+/// This is a reth-local duplicate of the alloy type, updated to match the latest
+/// spec changes where `transactions` can be `null` (to build from mempool).
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TestingBuildBlockRequestV1 {
```
mattsse, line 20: if this container is not part of the api, do we even need it? if not lets remove

PR #22277, opened 2026-02-17
fix(cli): handle hex-encoded `--builder.extradata` values

### LINKED ISSUE #22274: --builder.extradata rejects valid 32-byte hex-encoded values
## Description

The `--builder.extradata` CLI flag validates the input by checking the **string length** against `MAXIMUM_EXTRA_DATA_SIZE` (32), but the Ethereum protocol defines `extraData` as a **byte array** with a 32-byte limit.

This means valid 32-byte values cannot be passed as hex strings. For example, a SHA256 hash is exactly 32 bytes but its hex representation is 64 characters (or 66 with `0x` prefix), which gets rejected by the CLI parser.

## Steps to reproduce

```bash
# This fails despite being exactly 32 bytes when decoded:
reth node --builder.extradata "0x9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
# Error: Payload builder extradata size exceeds 32-byte limit
```

## Root cause

`ExtraDataValueParser::parse_ref()` in `crates/node/core/src/args/payload_builder.rs`:

```rust
if val.len() > MAXIMUM_EXTRA_DATA_SIZE {
    return Err(...)
}
```

This checks the string character count, not the decoded byte count. Additionally, `extra_data_bytes()` in `crates/node/core/src/cli/config.rs` converts via `as_bytes()` (UTF-8), so even if the validator were bypassed, the 64-char hex string would produce 64 bytes in the block header.

## Expected behavior

When the value starts with `0x`, the parser should:
1. Validate the hex characters
2. Check the **decoded** byte length against the 32-byte limit
3. `extra_data_bytes()` should hex-decode `0x`-prefixed strings to raw bytes

Plain strings (without `0x` prefix) should continue working as before.

## Context

The genesis JSON `extraData` field already supports hex encoding naturally (`"0x..."` is decoded to raw bytes). The CLI flag should be consistent with this behavior.

gakonst:
## Summary
Fixes `--builder.extradata` rejecting valid 32-byte hex-encoded values.

## Motivation
Closes #22274 — the CLI parser validated string character count instead of decoded byte count, so a 32-byte SHA256 hash (`0x` + 64 hex chars) was rejected despite being exactly at the limit.

## Changes
- `ExtraDataValueParser::parse_ref()` in `crates/node/core/src/args/payload_builder.rs`: detect `0x` prefix, validate even length, pre-check decoded byte count before allocating, then hex-decode and validate
- `extra_data_bytes()` in `crates/node/core/src/cli/config.rs`: hex-decode `0x`-prefixed input instead of treating it as raw UTF-8 bytes
- Added tests for valid hex, oversized hex, invalid hex, and odd-length hex

## Testing
```
cargo test -p reth-node-core
# 100 passed; 0 failed
```

Prompted by: matthias

### DIFF
--- a/crates/node/core/src/args/payload_builder.rs
+++ b/crates/node/core/src/args/payload_builder.rs
@@ -181,7 +181,39 @@ impl TypedValueParser for ExtraDataValueParser {
     ) -> Result<Self::Value, clap::Error> {
         let val =
             value.to_str().ok_or_else(|| clap::Error::new(clap::error::ErrorKind::InvalidUtf8))?;
-        if val.len() > MAXIMUM_EXTRA_DATA_SIZE {
+
+        // Determine the byte length: if the value is a 0x-prefixed hex string, validate the
+        // decoded byte length; otherwise, validate the raw UTF-8 byte length.
+        let byte_len = if let Some(hex_str) = val.strip_prefix("0x") {
+            if hex_str.len() % 2 != 0 {
+                return Err(clap::Error::raw(
+                    clap::error::ErrorKind::InvalidValue,
+                    "Hex extradata must have an even number of digits",
+                ))
+            }
+            // Reject before decoding to avoid unnecessary allocation
+            let decoded_len = hex_str.len() / 2;
+            if decoded_len > MAXIMUM_EXTRA_DATA_SIZE {
+                return Err(clap::Error::raw(
+                    clap::error::ErrorKind::InvalidValue,
+                    format!(
+                        "Payload builder extradata size exceeds {MAXIMUM_EXTRA_DATA_SIZE}-byte limit"
+                    ),
+                ))
+            }
+            // Validate that it's actually valid hex
+            alloy_primitives::hex::decode(hex_str).map_err(|e| {
+                clap::Error::raw(
+                    clap::error::ErrorKind::InvalidValue,
+                    format!("Invalid hex in extradata: {e}"),
+                )
+            })?;
+            decoded_len
+        } else {
+            val.len()
+        };
+
+        if byte_len > MAXIMUM_EXTRA_DATA_SIZE {
             return Err(clap::Error::raw(
                 clap::error::ErrorKind::InvalidValue,
                 format!(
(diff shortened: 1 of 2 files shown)


crates/node/core/src/args/payload_builder.rs
```diff
@@ -181,7 +181,39 @@ impl TypedValueParser for ExtraDataValueParser {
     ) -> Result<Self::Value, clap::Error> {
         let val =
             value.to_str().ok_or_else(|| clap::Error::new(clap::error::ErrorKind::InvalidUtf8))?;
-        if val.len() > MAXIMUM_EXTRA_DATA_SIZE {
+
+        // Determine the byte length: if the value is a 0x-prefixed hex string, validate the
+        // decoded byte length; otherwise, validate the raw UTF-8 byte length.
+        let byte_len = if let Some(hex_str) = val.strip_prefix("0x") {
```
DaniPopes: all of this is hex::check

DaniPopes (changes requested):

crates/node/core/src/args/payload_builder.rs
```diff
@@ -181,7 +181,22 @@ impl TypedValueParser for ExtraDataValueParser {
     ) -> Result<Self::Value, clap::Error> {
         let val =
             value.to_str().ok_or_else(|| clap::Error::new(clap::error::ErrorKind::InvalidUtf8))?;
-        if val.len() > MAXIMUM_EXTRA_DATA_SIZE {
+
+        // If the value is a 0x-prefixed hex string, validate it and check the decoded byte
+        // length; otherwise, validate the raw string length.
+        let byte_len = if val.starts_with("0x") {
+            alloy_primitives::hex::check(val).map_err(|e| {
```
mattsse, line 188: this doesnt look right, I guess this should just try to decode it here and then check the length

or use check_raw here

mattsse (commented):
defer to @DaniPopes

crates/node/core/src/args/payload_builder.rs
DaniPopes: ya prob just hexdecode

PR #22283, opened 2026-02-17
fix(rpc): resolve `AtBlockHash` to single block in `eth_getFilterChanges`

merge-script:
When a log filter uses `blockHash`, `filter_changes` should only query that one block.
But it was returning `(start_block, best_number)` — scanning the entire range instead.

The comment right above even says "blockHash is equivalent to fromBlock = toBlock = the block
number with hash blockHash", the code just... didn't do that.

The fix resolves the hash to its block number and uses `(block_number, block_number)`,
matching how `logs_for_filter` already handles `AtBlockHash`.

### DIFF
--- a/crates/rpc/rpc/src/eth/filter.rs
+++ b/crates/rpc/rpc/src/eth/filter.rs
@@ -271,11 +271,15 @@ where
                             .flatten();
                         logs_utils::get_filter_block_range(from, to, start_block, info)?
                     }
-                    FilterBlockOption::AtBlockHash(_) => {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
LGTM — the fix is correct per spec. Worth noting this code path barely matters in practice: `AtBlockHash` + polling `filter_changes` is an odd combination. The first poll returns the block's logs, every subsequent poll returns `Empty` (because `filter.block` advances past `best_number`). The real `AtBlockHash` handling lives in `logs_for_filter` which was already correct. Still, this makes the code match the comment right above it, so 👍

PR #22285, opened 2026-02-17
perf(storage): merge-scan HashedStorages writes in write_hashed_state

i-m-aditya:
## Summary

Apply the merge-scan pattern from #22157 to `HashedStorages` writes in `write_hashed_state`. 

## Motivation

`write_hashed_state` calls `seek_by_key_subkey` per storage slot update to check if the entry exists before deleting it, then calls `upsert` to insert the new value. Both are full B-tree traversals (primary key + DupSort sub-tree). Since the incoming updates (`HashedStorageSorted`) and the existing DB entries are both sorted, we can replace the per-entry seek with a linear cursor scan.

## Testing
`cargo test -p reth-provider`

All existing + 1 new test pass.


shekhirin:
derek bench

shekhirin:
derek bench

i-m-aditya:
One optimization that should be a safe win: when is_wiped() is true, the original still does seek_by_key_subkey + upsert per slot even though all entries were just deleted. Replacing that with append_dup (O(1) amortized, no B-tree seek) would eliminate U * log(S) wasted seeks on the wiped path.

PR #22289, opened 2026-02-17
feat(stateless): make witness generation conform to the draft specs

jsign:
This PR contains changes to Reth to conform to the witness generations specs draft. These specs are still in flux ([PR 1](https://github.com/jsign/execution-specs/pull/1) and [PR 3](https://github.com/jsign/execution-specs/pull/3)). We are working with the EF STEEL team to officialize them.

I'm leaving this PR in draft for a bit to get some feedback about the changes, mostly reg the last paragraph of this PR description.

You don't need to understand the specs fully to review this PR, since all changes must make sense on their own, which I'll help to explain in this PR description and comments.

Note that other required fixes aren't in this repo but in the newly created [paradigmxyz/stateless](https://github.com/paradigmxyz/stateless) (PR there to be created soon).

Below, I’ll summarise the changes, but I’ll also create PR comments extending the explanation a bit more for each case:
- Avoid including empty bytecode case in the witness (i.e. rlp encoded `0x80` ).
- Avoid including empty MPT node references (i.e. rlp encoded `0x80`).
- Avoid including bytecodes generated during execution (i.e., only pre-state bytecodes are needed by a stateless validator).
- Avoid including storage tries root node if the accessed account never accessed any storage slots.
- Change both ParallelTrie and StatelessTrie (but this now lies in `stateless` repo thus in other PR) post-root calculation to apply insertions/updates first and deletions after to minimise MPT nodes included in the witness — this is related to minimising cases of sibling branch compression inclusions.
- Conform to the ordering of `nodes`, `bytecodes` and `ancestors` in `ExecutionWitness` as defined in the spec (basically, lexicographic order for the first two, and `blockNum asc` for the latter).

There is only one change that makes this new execution witness version a breaking change: the new post-state root update ordering (5th bullet point above). If someone uses an execution witness generated by Reth done with this PR, with an older stateless trie implementation, it might find that the post-state root can’t be calculated because the older version of the trie might ask for more MPT nodes than strictly needed. 

We could fix this by adding more code to still support both witness generation styles for a while until people update their code (prob also involves modifying RPC `debug_executionWitness` to have an extra optional param to ask for “canonical form” or the current/mainnet style, or create a new one). If we also want to keep the behavior for other fields (i.e. still keep the bytecodes bloating and other cases, it can become a bit messy since basically all code fixes in this PR should be guarded in some kind of `if canonical_witness` or similar. I guess doable but can get tricky.

Note that the reverse is okay — the new stateless trie (that now lives in paradigmxyz/stateless) can support older-style execution witnesses, since these extra MPT nodes are totally okay to be there (they won’t be used). I want to gauge a bit what the Reth team prefers to do on this front.

### DIFF
--- a/crates/chain-state/src/in_memory.rs
+++ b/crates/chain-state/src/in_memory.rs
@@ -1169,6 +1169,7 @@ mod tests {
             &self,
             _input: TrieInput,
             _target: HashedPostState,
+            _mode: reth_trie::ExecutionWitnessMode,
         ) -> ProviderResult<Vec<Bytes>> {
             Ok(Vec::default())
         }
--- a/crates/revm/src/witness.rs
+++ b/crates/revm/src/witness.rs
@@ -1,5 +1,5 @@
 use alloc::vec::Vec;
-use alloy_primitives::{keccak256, map::B256Map, Bytes, B256};
+use alloy_primitives::{keccak256, Bytes, B256};
 use reth_trie::{ExecutionWitnessMode, HashedPostState, HashedStorage};
 use revm::database::State;
 
--- a/crates/revm/src/witness.rs
+++ b/crates/revm/src/witness.rs
@@ -1,6 +1,6 @@
 use alloc::vec::Vec;
-use alloy_primitives::{keccak256, Bytes, B256};
-use reth_trie::{HashedPostState, HashedStorage};
+use alloy_primitives::{keccak256, map::B256Map, Bytes, B256};
+use reth_trie::{ExecutionWitnessMode, HashedPostState, HashedStorage};
 use revm::database::State;
 
 /// Tracks state changes during execution.
@@ -30,21 +30,36 @@ pub struct ExecutionWitnessRecord {
 }
 
 impl ExecutionWitnessRecord {
-    /// Records the state after execution.
-    pub fn record_executed_state<DB>(&mut self, statedb: &State<DB>) {
-        self.codes = statedb
-            .cache
-            .contracts
-            .values()
-            .map(|code| code.original_bytes())
-            .chain(
-                // cache state does not have all the contracts, especially when
-                // a contract is created within the block
-                // the contract only exists in bundle state, therefore we need
-                // to include them as well
-                statedb.bundle_state.contracts.values().map(|code| code.original_bytes()),
-            )
-            .collect();
+    /// Records the state after execution using the given witness generation mode.
+    pub fn record_executed_state<DB>(&mut self, statedb: &State<DB>, mode: ExecutionWitnessMode) {
+        self.codes = match mode {
+            ExecutionWitnessMode::Legacy => statedb
+                .cache
+                .contracts
+                .values()
+                .map(|code| code.original_bytes())
+                .chain(
+                    // cache state does not have all the contracts, especially when
+                    // a contract is created within the block
+                    // the contract only exists in bundle state, therefore we need
+                    // to include them as well
+                    statedb.bundle_state.contracts.values().map(|code| code.original_bytes()),
+                )
+                .collect(),
+            ExecutionWitnessMode::Canonical => {
+                let mut accessed_codes = B256Map::default();
+                for code in statedb.cache.contracts.values() {
+                    let code = code.original_bytes();
+                    if code.is_empty() {
+                        continue;
+                    }
+                    accessed_codes.entry(keccak256(&code)).or_insert(code);
+                }
+                let mut codes: Vec<_> = accessed_codes.into_values().collect();
+                codes.sort_unstable();
+                codes
+            }
+        };
 
         for (address, account) in &statedb.cache.accounts {
             let hashed_address = keccak256(address);
@@ -75,9 +90,9 @@ impl ExecutionWitnessRecord {
     }
 
     /// Creates the record from the state after execution.
-    pub fn from_executed_state<DB>(state: &State<DB>) -> Self {
+    pub fn from_executed_state<DB>(state: &State<DB>, mode: ExecutionWitnessMode) -> Self {
         let mut record = Self::default();
-        record.record_executed_state(state);
+        record.record_executed_state(state, mode);
         record
     }
 
@@ -93,6 +108,7 @@ impl ExecutionWitnessRecord {
         state_provider: &SP,
         headers_provider: &HP,
         block_number: u64,
+        mode: ExecutionWitnessMode,
     ) -> reth_storage_errors::provider::ProviderResult<alloy_rpc_types_debug::ExecutionWitness>
     where
         SP: reth_storage_api::StateProofProvider + ?Sized,
@@ -101,7 +117,7 @@ impl ExecutionWitnessRecord {
     {
         let Self { hashed_state, codes, keys, lowest_block_number } = self;
 
-        let state = state_provider.witness(Default::default(), hashed_state)?;
+        let state = state_provider.witness(Default::default(), hashed_state, mode)?;
         let mut exec_witness =
             alloy_rpc_types_debug::ExecutionWitness { state, codes, keys, ..Default::default() };
 
--- a/crates/revm/src/witness.rs
+++ b/crates/revm/src/witness.rs
@@ -47,15 +47,13 @@ impl ExecutionWitnessRecord {
                 )
                 .collect(),
             ExecutionWitnessMode::Canonical => {
-                let mut accessed_codes = B256Map::default();
-                for code in statedb.cache.contracts.values() {
-                    let code = code.original_bytes();
-                    if code.is_empty() {
-                        continue;
-                    }
-                    accessed_codes.entry(keccak256(&code)).or_insert(code);
-                }
-                let mut codes: Vec<_> = accessed_codes.into_values().collect();
+                let mut codes: Vec<_> = statedb
+                    .cache
+                    .contracts
+                    .values()
+                    .map(|c| c.original_bytes())
+                    .filter(|code| !code.is_empty())
+                    .collect();
                 codes.sort_unstable();
                 codes
             }
--- a/crates/storage/provider/src/providers/state/historical.rs
+++ b/crates/storage/provider/src/providers/state/historical.rs
@@ -23,9 +23,9 @@ use reth_trie::{
     trie_cursor::InMemoryTrieCursorFactory,
     updates::TrieUpdates,
     witness::TrieWitness,
-    AccountProof, HashedPostState, HashedPostStateSorted, HashedStorage, KeccakKeyHasher,
-    MultiProof, MultiProofTargets, StateRoot, StorageMultiProof, StorageRoot, TrieInput,
-    TrieInputSorted,
+    AccountProof, ExecutionWitnessMode, HashedPostState, HashedPostStateSorted, HashedStorage,
+    KeccakKeyHasher, MultiProof, MultiProofTargets, StateRoot, StorageMultiProof, StorageRoot,
+    TrieInput, TrieInputSorted,
 };
 use reth_trie_db::{
     hashed_storage_from_reverts_with_provider, DatabaseProof, DatabaseStateRoot,
@@ -524,13 +524,18 @@ impl<
         })
     }
 
-    fn witness(&self, input: TrieInput, target: HashedPostState) -> ProviderResult<Vec<Bytes>> {
+    fn witness(
+        &self,
+        input: TrieInput,
+        target: HashedPostState,
+        mode: ExecutionWitnessMode,
+    ) -> ProviderResult<Vec<Bytes>> {
         reth_trie_db::with_adapter!(self.provider, |A| {
             let mut input = input;
             input.prepend(self.revert_state()?.into());
             let nodes_sorted = input.nodes.into_sorted();
             let state_sorted = input.state.into_sorted();
-            TrieWitness::new(
+            let witness = TrieWitness::new(
                 InMemoryTrieCursorFactory::new(
                     reth_trie_db::DatabaseTrieCursorFactory::<_, A>::new(self.tx()),
                     &nodes_sorted,
@@ -541,10 +546,16 @@ impl<
                 ),
             )
             .with_prefix_sets_mut(input.prefix_sets)
-            .always_include_root_node()
-            .compute(target)
-            .map_err(ProviderError::from)
-            .map(|hm| hm.into_values().collect())
+            .with_execution_witness_mode(mode);
+            let witness =
+                if mode.is_canonical() { witness } else { witness.always_include_root_node() };
+            witness.compute(target).map_err(ProviderError::from).map(|hm| {
+                let mut values: Vec<_> = hm.into_values().collect();
+                if mode.is_canonical() {
+                    values.sort_unstable();
+                }
+                values
+            })
         })
     }
 }
--- a/crates/trie/common/src/execution_witness.rs
+++ b/crates/trie/common/src/execution_witness.rs
@@ -1,0 +1,18 @@
+/// Controls how execution witnesses are generated.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
+pub enum ExecutionWitnessMode {
+    /// Preserve the current witness shape for compatibility with existing consumers.
+    #[default]
+    Legacy,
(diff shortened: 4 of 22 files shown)


crates/revm/src/witness.rs
```diff
@@ -32,19 +32,22 @@ pub struct ExecutionWitnessRecord {
...
-            .cache
-            .contracts
-            .values()
-            .map(|code| code.original_bytes())
-            .chain(
-                // cache state does not have all the contracts, especially when
-                // a contract is created within the block
-                // the contract only exists in bundle state, therefore we need
-                // to include them as well
-                statedb.bundle_state.contracts.values().map(|code| code.original_bytes()),
```
jsign, line 45: This was interesting since including bytecode created during block execution apparently was intentional.

From my perspective, this should be actually the opposite. Bytecode created during block execution isn't required in the `ExecutionWitness` since the stateless execution doesn't require it as data.

crates/revm/src/witness.rs
```diff
@@ -32,19 +32,22 @@ pub struct ExecutionWitnessRecord {
...
+        for code in statedb.cache.contracts.values() {
+            let code = code.original_bytes();
+            if code.is_empty() {
+                continue;
+            }
+
+            accessed_codes.entry(keccak256(&code)).or_insert(code);
+        }
+        self.codes = accessed_codes.into_values().collect();
+        self.codes.sort();
```
mattsse: I believe all of this can use sort_unstable which is slightly faster

crates/revm/src/witness.rs
```diff
@@ -32,19 +32,22 @@ pub struct ExecutionWitnessRecord {
...
+        // not required in the witness.
+        let mut accessed_codes = B256Map::default();
+        for code in statedb.cache.contracts.values() {
+            let code = code.original_bytes();
+            if code.is_empty() {
+                continue;
+            }
+
+            accessed_codes.entry(keccak256(&code)).or_insert(code);
+        }
```
mattsse: isnt this a map of codehash already?

so i dont think we need to do any filtering + keccak here

crates/revm/src/witness.rs
```diff
@@ -32,19 +32,22 @@ pub struct ExecutionWitnessRecord {
...
+        // during execution. `cache.contracts` tracks bytecode lookups
+        // performed by revm. Note `bundle_state.contracts` is not considered
+        // since it contains also bytecodes created during execution which are
+        // not required in the witness.
+        let mut accessed_codes = B256Map::default();
+        for code in statedb.cache.contracts.values() {
+            let code = code.original_bytes();
+            if code.is_empty() {
+                continue;
+            }
```
jsign: Adding this extra check. If the account `code_hash` is the empty bytecode, then we don't need to include the RLP-enocded bytecode in the witness (i.e. 0x80).

The stateless validator can see that if the `code_hash` of the account is the empty code, then no bytecode is required to query in the execution witness.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -483,9 +483,13 @@ impl<
 
     fn witness(&self, mut input: TrieInput, target: HashedPostState) -> ProviderResult<Vec<Bytes>> {
         input.prepend(self.revert_state()?.into());
-        TrieWitness::overlay_witness(self.tx(), input, target)
-            .map_err(ProviderError::from)
-            .map(|hm| hm.into_values().collect())
+        TrieWitness::overlay_witness(self.tx(), input, target).map_err(ProviderError::from).map(
+            |hm| {
+                let mut v: Vec<_> = hm.into_values().collect();
+                v.sort();
```
jsign: Just sorting for make the ExecutionWitness canonical.

mattsse (changes requested):
ty, some suggestions

also please let me know if you have any suggestions for how to make this easier, will try my best to keep the stateless crate and this here in sync

crates/trie/db/tests/witness.rs
```diff
@@ -26,16 +25,14 @@ fn includes_empty_node_preimage() {
     let hashed_address = keccak256(address);
     let hashed_slot = B256::random();
 
-    // witness includes empty state trie root node
```
jsign: This test was a bit surprising since it seems it was intentional to make the witness include the root node of an empty state trie.

From my perspective, if `code_hash` is the empty trie, then I see no need to include it. (I fixed the code for this case later, I'll add a comment)

crates/trie/trie/src/witness.rs
```diff
@@ -140,9 +140,17 @@ where
...
-        for storage_node in multiproof.storages.values().flat_map(|s| s.subtree.values()) {
-            if let Entry::Vacant(entry) = self.witness.entry(keccak256(storage_node.as_ref())) {
-                entry.insert(storage_node.clone());
+        for (hashed_address, storage_mp) in &multiproof.storages {
+            // Skip storage proof nodes for accounts that don't target any storage slots.
+            // These nodes may be produced while deriving account storage roots, but they are
+            // unnecessary for stateless execution witness verification.
+            if !proof_targets.get(hashed_address).is_some_and(|slots| !slots.is_empty()) {
+                continue;
+            }
```
jsign: This is an interesting fix required since the Multiproofs always include the root node in storage tries even if you didn't ask for any storage slot.
I think this happens since it tries to calculate the `code_hash` by pulling the node which indirectly adds it to the witness.

Here I put a guard that if we didn't ask for any storage slot in the proof, then we don't need any storage trie MPT node.

crates/trie/trie/src/witness.rs
```diff
@@ -167,21 +175,38 @@ where
...
-            for hashed_slot in hashed_slots.into_iter().sorted_unstable() {
-                let storage_nibbles = Nibbles::unpack(hashed_slot);
-                let maybe_leaf_value = storage
-                    .and_then(|s| s.storage.get(&hashed_slot))
-                    .filter(|v| !v.is_zero())
-                    .map(|v| alloy_rlp::encode_fixed_size(v).to_vec());
+            let sorted_slots = hashed_slots.into_iter().sorted_unstable().collect_vec();
+            // Match execution-specs ordering to minimize witness nodes:
+            // apply updates/inserts first, then deletions.
+            for is_delete_pass in [false, true] {
```
jsign: The change here is that while doing post-state root calculation, we do two passes:
- First, the insertions+updates
- Then the deletions.

This minimizes the branch compressions during post-state root calculation. Branch compressions require sibilng nodes to be included in the witness since we have to "move them up".

By doing the deletions last, we minimize including sibilings.

crates/trie/trie/src/witness.rs
```diff
@@ -201,6 +226,10 @@ where
             }
         }
 
+        // Filter out empty trie root nodes — they carry no useful witness information
+        // since any verifier can reconstruct an empty trie root trivially.
+        self.witness.retain(|_, v| v.as_ref() != [EMPTY_STRING_CODE]);
```
jsign: The MPT nodes `nodes` in the execution witness had `0x80` which are basically "empty childrens" in branch nodes.

We don't need them since we know are empty children.

jsign (commented):
@mattsse adding some extra comments to help understanding.

crates/revm/src/witness.rs
jsign: Ah yeah, I can change it.

crates/revm/src/witness.rs
jsign: Ah added a comment above, lmk if make sense to you the rationale.

mattsse (approved):
makes sense

maybe @mediocregopher wants to do a quick sanity review

crates/chain-state/src/in_memory.rs
```diff
@@ -1169,6 +1169,7 @@ mod tests {
             &self,
             _input: TrieInput,
             _target: HashedPostState,
+            _mode: reth_trie::ExecutionWitnessMode,
```
jsign, line 1172: `ExecutionWitnessMode` is a new enum to allow generating/using both witness types: `Legacy` (current), and `Canonical` (new).

At the `debug_executionWitness` level there is a new parameter that allows to select this. It is an `Option` so not providing it defaults to `Legacy` for backwards compat.

crates/revm/src/witness.rs
```diff
@@ -30,21 +30,36 @@ pub struct ExecutionWitnessRecord {
...
-            .chain(
-                // cache state does not have all the contracts, especially when
-                // a contract is created within the block
-                // the contract only exists in bundle state, therefore we need
-                // to include them as well
-                statedb.bundle_state.contracts.values().map(|code| code.original_bytes()),
-            )
-            .collect();
+    /// Records the state after execution using the given witness generation mode.
+    pub fn record_executed_state<DB>(&mut self, statedb: &State<DB>, mode: ExecutionWitnessMode) {
```
jsign, line 34: Now we have both.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -541,10 +546,16 @@ impl<
                 ),
             )
             .with_prefix_sets_mut(input.prefix_sets)
-            .always_include_root_node()
-            .compute(target)
-            .map_err(ProviderError::from)
-            .map(|hm| hm.into_values().collect())
+            .with_execution_witness_mode(mode);
+            let witness =
+                if mode.is_canonical() { witness } else { witness.always_include_root_node() };
```
jsign, line 551: Note `always_include_root_node()` now depends on mode.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -541,10 +546,16 @@ impl<
...
+            witness.compute(target).map_err(ProviderError::from).map(|hm| {
+                let mut values: Vec<_> = hm.into_values().collect();
+                if mode.is_canonical() {
```
jsign, line 554: Ditto, reg sorting.

crates/trie/db/tests/witness.rs
```diff
@@ -34,20 +37,33 @@ fn includes_empty_node_preimage() {
...
+            DatabaseTrieCursorFactory::<_, A>::new(provider.tx_ref()),
+            DatabaseHashedCursorFactory::new(provider.tx_ref()),
+        )
+        .with_execution_witness_mode(ExecutionWitnessMode::Canonical)
+        .compute(HashedPostState {
+            accounts: HashMap::from_iter([(hashed_address, Some(Account::default()))]),
+            storages: HashMap::default(),
+        })
+        .unwrap();
+        assert!(canonical_empty_witness.is_empty());
```
jsign, line 65: As a refresher: compared to `Legacy` above, the empty root hash must not be in the witness since only by the hash clients can know which is the RLP-representation.

crates/trie/db/tests/witness.rs
```diff
@@ -185,3 +220,163 @@ fn correctly_decodes_branch_node_values() {
...
+            DatabaseHashedCursorFactory::new(provider.tx_ref()),
+        )
+        .with_execution_witness_mode(ExecutionWitnessMode::Canonical)
+        .compute(target_state)
+        .unwrap();
+        assert!(canonical_witness.contains_key(&state_root));
+        for node in multiproof.account_subtree.values() {
+            assert_eq!(canonical_witness.get(&keccak256(node)), Some(node));
+        }
+        assert!(!canonical_witness.contains_key(&storage_root));
```
jsign, line 284: This test is covering another behavior difference between `Legacy` and `Canonical`: if the storage trie of an account doesn't change:
- `Legacy` always include the storage trie root node RLP (not sure why, but that's how it works)
- `Canonical`, doesn't since isn't required for anything.

crates/trie/db/tests/witness.rs
```diff
@@ -185,3 +220,163 @@ fn correctly_decodes_branch_node_values() {
...
+    });
+}
+
+#[test]
+fn canonical_mode_handles_mixed_storage_inserts_and_removals() {
```
jsign, line 289: The last remaining diff between `Legacy` and `Canonical`: how apply state diffs when calculating post-state root.

As quick context:
- `Legacy`: applies removals first and then insert/updates -- my interpretation is that this is done to provide the most "compatibility" with users since it will maximize the number of potential nodes that might be required.
- `Canonical`: applies insert/updates first and then removals. This is done to minimize branch compressions, thus minimize the witness size as much as possible. This implies that whoever is using this witness should also be a "canonical" one in the sense (i.e. apply insert/updates firsta nd removals after).

crates/trie/db/tests/witness.rs
```diff
@@ -185,3 +220,163 @@ fn correctly_decodes_branch_node_values() {
...
+    let factory = create_test_provider_factory();
+    let provider = factory.provider_rw().unwrap();
+
+    let address = Address::random();
+    let hashed_address = keccak256(address);
+
+    // Pre-state storage root is a branch with two children at nibbles 1 and 2.
```
jsign, line 296: TL;DR: we have a branch node with two children, and we have some state-diff that will add a new children and remove an existing one.

This setup allows to see if the sibiling is or isn't in the witness depending on the mode.

jsign (commented):
@mediocregopher, this is ready for new eyes:
- Rebased on top of `master`
- Adjusted for the new V2 stuff
- Added `ExecutionWitnessMode` so both styles `Legacy` (current) and `Canonical` (new) behavior can co-exist.

I resolved all previous comments, and created new ones so you only see fresh ones after this re-work.

cc @figtracer

crates/revm/src/witness.rs
```diff
@@ -30,21 +30,36 @@ pub struct ExecutionWitnessRecord {
...
+                for code in statedb.cache.contracts.values() {
+                    let code = code.original_bytes();
+                    if code.is_empty() {
+                        continue;
+                    }
+                    accessed_codes.entry(keccak256(&code)).or_insert(code);
+                }
+                let mut codes: Vec<_> = accessed_codes.into_values().collect();
+                codes.sort_unstable();
+                codes
```
mediocregopher: I think this could be simplified to eliminate the keccak by doing something like:

```
statedb.cache.contracts.values.filter_map(|c| c.original_bytes()).sorted_unstable().dedup().collect()
```

crates/trie/common/src/execution_witness.rs
```diff
@@ -0,0 +1,18 @@
+/// Controls how execution witnesses are generated.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
+pub enum ExecutionWitnessMode {
+    /// Preserve the current witness shape for compatibility with existing consumers.
```
mediocregopher: nit: Let's assume we're going to have this enum for a while, describe the current witness shape, don't refer to "existing consumers" as those will change over time.

crates/trie/common/src/execution_witness.rs
```diff
@@ -0,0 +1,18 @@
...
+    #[default]
+    Legacy,
+    /// Produce the minimized, draft-spec witness form.
```
mediocregopher: nit: Similarly here, can we link to the draft spec, or describe it?

crates/revm/src/witness.rs
jsign: Can't do that exactly since `original_bytes()` doesn't return an option. But did something to avoid the keccak which should be equivalent to what we need. LMK if makes sense.

crates/trie/common/src/execution_witness.rs
jsign: Extended with the specs branch link and difference highlights.

crates/trie/common/src/execution_witness.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// ethereum/execution-specs@projects/zkevm. The main differences with the
+    /// legacy format are:
+    /// - For the `bytecode` field:
+    ///     - It contains only bytecodes required for execution, compared to Legacy which also
+    ///       contains created bytecode.
+    ///     - Compared to Legacy, it does not include empty bytecodes (i.e. 0x80).
+    ///     - Values are sorted lexicographically ascending.
+    /// - For the `state` field:
+    ///     - Avoids including empty nodes (i.e. 0x80).
+    ///     - Compared to legacy, it does not include storage trie root nodes if no storae is
```
mediocregopher:
```suggestion
    ///     - Compared to legacy, it does not include storage trie root nodes if no storage is
```

mediocregopher (approved):
One small nit but lgtm

mattsse (approved):

PR #22290, opened 2026-02-17
perf: store blinded node hashes on `SparseNode::Branch`

klkvr:
This PR removes `SparseNode::Hash` variant and moves all blinded hashes onto the `SparseNode::Branch`


crates/trie/sparse/src/parallel.rs
```diff
@@ -268,10 +266,52 @@ impl SparseTrie for ParallelSparseTrie {
...
+                    let SparseNode::Branch { blinded_mask, blinded_hashes, .. } = self
+                        .upper_subtrie
+                        .nodes
+                        .get_mut(&node.path.slice(0..UPPER_TRIE_MAX_DEPTH - 1))
+                        .unwrap()
+                {
+                    let nibble = node.path.get_unchecked(UPPER_TRIE_MAX_DEPTH - 1);
+                    blinded_mask.unset_bit(nibble);
+                    if let Some(trie_node) = subtrie.nodes.get_mut(&node.path) {
+                        let hash = blinded_hashes[nibble as usize];
```
klkvr: this is the ugliest part of this change — for boundary nodes we need to update their parents in upper subtrie and remove the revealed hashes from them

crates/trie/sparse/src/parallel.rs
```diff
@@ -2719,7 +2781,12 @@ impl SparseSubtrie {
                         },
                         None => SparseNodeState::Dirty,
                     };
-                entry.insert(SparseNode::Branch { state_mask, state });
+                entry.insert(SparseNode::Branch {
```
mediocregopher, line 2729: Ideally we wouldn't construct the blinded_hashes until this point, so we don't allocate the Box unless it's definitely going to be kept

crates/trie/sparse/src/parallel.rs
mediocregopher: Maybe instead of doing this after calling `subtrie.reveal_node`, you could pull the hash out of the parent beforehand and pass it into `subtrie.reveal_node` as an Option, so the method can populate the cached state while it's constructing the node.

klkvr: ah right good point, done in ca62b86733

mediocregopher (approved):

PR #22294, opened 2026-02-18
perf(storage): skip plain state conversion in write_state for storage v2

gakonst:
## Summary
Skip the expensive `BundleState::to_plain_state_and_reverts` conversion in `write_state` when running with `--storage.v2`.

## Motivation
In storage v2, plain state and changesets are written to static files — not MDBX. When `save_blocks` calls `write_state` with all config flags disabled (`write_receipts=false`, `write_account_changesets=false`, `write_storage_changesets=false`), the only useful work is writing bytecodes. But `to_plain_state_and_reverts` still iterates every changed account and storage slot to build `StateChangeset` and `PlainStateReverts` structs that are immediately discarded. This shows up as non-trivial time on the `save_blocks_write_state` metric.

## Changes
- Added early return in `write_state` when `use_hashed_state()` is true and all config outputs are disabled
- Added `write_bytecodes` helper that extracts contracts directly from `BundleState.contracts`, skipping the full plain state conversion
- The optimization only triggers when called from `save_blocks` (config flags all false); other callers passing `StateWriteConfig::default()` are unaffected

## Testing
`cargo test -p reth-provider` — 110 passed, 0 failed

Prompted by: joshie

### DIFF
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -28,7 +28,7 @@ use alloy_eips::BlockHashOrNumber;
 use alloy_primitives::{
     keccak256,
     map::{hash_map, AddressSet, B256Map, HashMap},
-    Address, BlockHash, BlockNumber, TxHash, TxNumber, B256,
+    Address, BlockHash, BlockNumber, TxHash, TxNumber, B256, KECCAK256_EMPTY,
 };
 use itertools::Itertools;
 use parking_lot::RwLock;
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -843,6 +844,24 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
 
         Ok(())
     }
+
+    /// Writes only bytecodes from the bundle state, skipping plain state and changeset conversion.
+    ///
+    /// Used in storage v2 where plain state and changesets are written to static files, making
+    /// the full [`BundleState::to_plain_state_and_reverts`] conversion unnecessary.
```
```suggestion
    /// Writes only bytecodes from the bundle state.
```

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -843,6 +843,22 @@ impl<TX: DbTx + DbTxMut + 'static, N: NodeTypesForProvider> DatabaseProvider<TX,
 
         Ok(())
     }
+
+    /// Writes bytecodes to MDBX.
+    ///
+    /// Also used as a fast path in [`StateWriter::write_state`] for storage v2 where plain state
+    /// and changesets are written to static files, making the full
+    /// [`BundleState::to_plain_state_and_reverts`] conversion unnecessary.
```
```suggestion
    /// Writes bytecodes to MDBX.
```

```suggestion
    /// Writes bytecodes to MDBX.
```

Rjected (approved):
lgtm

PR #22299, opened 2026-02-18
docs: clarify StateWriteConfig is about database (MDBX) writes vs static files

gakonst:
## Summary
Clarifies `StateWriteConfig` doc comments to make it obvious this struct controls what gets written to the **database (MDBX)**, not to static files.

## Motivation
The existing docs just said "skip writing certain data types" without explaining the DB vs static file distinction. During pipeline execution, receipts and changesets can be written to static files instead of MDBX, and this config is how the caller signals that — the docs should reflect that.

## Changes
- Updated struct-level doc to mention MDBX and the static file relationship
- Updated each field's doc to explain when `false` is used (static file path)

Prompted by: Dan Cline

### DIFF
--- a/crates/storage/storage-api/src/state_writer.rs
+++ b/crates/storage/storage-api/src/state_writer.rs
@@ -127,16 +127,25 @@ pub trait StateWriter {
     ) -> ProviderResult<ExecutionOutcome<Self::Receipt>>;
 }
 
-/// Configuration for what to write when calling [`StateWriter::write_state`].
+/// Configuration for what to write to the database (MDBX) when calling
+/// [`StateWriter::write_state`].
 ///
-/// Used to skip writing certain data types, when they are being written separately.
+/// During pipeline execution, some data types (receipts, changesets) may be written directly to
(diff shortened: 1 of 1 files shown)


crates/storage/storage-api/src/state_writer.rs
```diff
@@ -127,16 +127,25 @@ pub trait StateWriter {
...
 }
 
-/// Configuration for what to write when calling [`StateWriter::write_state`].
+/// Configuration for what to write to the database (MDBX) when calling
+/// [`StateWriter::write_state`].
 ///
-/// Used to skip writing certain data types, when they are being written separately.
+/// During pipeline execution, some data types (receipts, changesets) may be written directly to
+/// static files instead of the database. This config allows skipping those data types in the
+/// database write to avoid duplication.
```
Rjected:
```suggestion
/// Some types (receipts, changesets) may be written directly to
/// static files instead of the database depending on the storage settings. This config allows skipping those types in the
/// database write to avoid duplication.
```

Rjected (approved):

PR #22305, opened 2026-02-18
perf(engine): prefetch first txs sequentially to avoid rayon scheduling stall

gakonst:
## Summary
Recovers the first 4 transactions sequentially before entering rayon parallel recovery to eliminate ~1ms first-tx latency.

## Motivation
The `record_transaction_wait` metric shows that execution sometimes waits ~1ms for the first transaction from the parallel signature recovery iterator. This happens because:
1. `spawn_blocking` needs tokio to schedule the blocking task
2. `rayon::in_place_scope` + `s.spawn` inside `ForEachOrdered` doesn't guarantee index 0 is processed first — rayon's work-stealing can pick any chunk

The ordered consumer blocks on slot 0's condvar until rayon happens to process it, which can easily take ~1ms under load.

## Changes
- Added `PARALLEL_PREFETCH_COUNT = 4` constant in `PayloadProcessor`
- In `spawn_tx_iterator`'s parallel path (≥30 txs), the first 4 transactions are now recovered sequentially and sent to both the prewarm and execute channels immediately, before entering `into_par_iter().for_each_ordered()` for the remainder
- The parallel iterator starts at an offset, so ordering is preserved and no double-recovery occurs

## Testing
- `cargo check -p reth-engine-tree` ✅
- `cargo +nightly clippy -p reth-engine-tree` ✅
- `cargo +nightly fmt --all --check` ✅

Prompted by: DaniPopes


DaniPopes (approved):

DaniPopes:
derek bench

PR #22308, opened 2026-02-18
fix(txpool): ensure transactions are added to pending subpool in nonce order

gakonst:
## Summary

Fixes nonce-disordered iteration in live `BestTransactions` iterators when a newly added transaction triggers promotion of same-sender lower-nonce transactions.

## Motivation

When `TxPool::add_transaction` inserts a tx that also causes a lower-nonce same-sender tx to be promoted (via `updates`), the new tx was added to the pending subpool *before* the promoted tx. Live `BestTransactions` iterators then received nonce 2 before nonce 1, violating the partial nonce ordering contract and causing nonce gaps during block building.

Supersedes #17702.

## Changes

- Partition `updates` in `add_transaction`: same-sender lower-nonce promotions are processed *before* adding the new tx; everything else is processed after.
- Added `best_transactions_nonce_order_on_promotion` test that reproduces the exact scenario from #17701.

## Testing

```
cargo test -p reth-transaction-pool --lib  # 222 passed
```

Prompted by: mattsse

### DIFF
--- a/.changelog/warm-pandas-cook.md
+++ b/.changelog/warm-pandas-cook.md
@@ -1,0 +1,5 @@
+---
+reth-transaction-pool: patch
+---
+
+Fixed a bug where transactions from the same sender were added to the pending subpool out of nonce order. Ensured `process_updates` runs before `add_new_transaction` so that lower-nonce promotions are enqueued before the newly inserted higher-nonce transaction, preserving correct ordering for live `BestTransactions` iterators.
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -761,15 +761,40 @@ impl<T: TransactionOrdering> TxPool<T> {
             .update(on_chain_nonce, on_chain_balance);
 
         match self.all_transactions.insert_tx(tx, on_chain_balance, on_chain_nonce) {
-            Ok(InsertOk { transaction, move_to, replaced_tx, updates, state }) => {
-                // Process updates before adding the new transaction so that same-sender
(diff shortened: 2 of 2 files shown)


mattsse (approved):

mattsse (changes requested):
## Review: Breaks the gap-fill scenario

The swap-only approach fixes the reported issue (#17701) but introduces the **same nonce-disorder bug in reverse** for gap-fills — the more common promotion path.

### Example

1. Add `tx1` (nonce 1) → goes to **Queued** (nonce gap)
2. Create a `BestTransactions` iterator
3. Add `tx0` (nonce 0) → fills the gap, `tx1` gets promoted via `updates`

**With this PR:** `process_updates` promotes `tx1` first → listener hears `[1, 0]` → nonce disorder.

### Suggested fix

Partition `updates` by the new transaction's nonce and interleave `add_new_transaction` between the two halves:

```rust
let new_nonce = transaction.nonce();
let (before, after): (Vec<_>, Vec<_>) = updates.into_iter()
    .partition(|u| u.id.nonce < new_nonce);

let mut outcome = self.process_updates(before);  // promote lower-nonce txs first
self.add_new_transaction(transaction.clone(), replaced_tx.clone(), move_to);
self.metrics.inserted_transactions.increment(1);
let after_outcome = self.process_updates(after);  // then promote higher-nonce txs

outcome.promoted.extend(after_outcome.promoted);
outcome.discarded.extend(after_outcome.discarded);
```

Also needs a gap-fill test (insert nonce 1, then nonce 0, assert `best` yields `[0, 1]`).

mattsse (approved):

PR #22311, opened 2026-02-18
fix(net): stop penalizing peers for empty block body responses

gakonst:
## Summary
Stop penalizing peers for returning empty block body responses, which can be legitimate when a block body exceeds the peer's soft response size limit (~2MB).

## Motivation
With increasing gas limits, block bodies can grow large enough that they exceed the ~2MB soft response limit applied by peers. In this case, a peer legitimately returns an empty response because it can't fit even a single body. Previously, `BodiesRequestFuture` treated all errors (including `EmptyResponse`) as bad messages and penalized the peer via `report_bad_message`. This is too aggressive — the peer did nothing wrong.

The `StateFetcher` already handles this somewhat correctly by marking peers as `last_response_likely_bad` (de-ranking, not penalizing), but the downloader layer was additionally calling `report_bad_message` which applies reputation damage.

## Changes
- `crates/net/p2p/src/error.rs`: Add `DownloadError::is_malicious_response()` which returns `true` only for clearly invalid/malicious responses (`TooManyBodies`, `BodyValidation`, `HeaderValidation`, `InvalidTip`)
- `crates/net/downloaders/src/bodies/request.rs`:
  - Only call `report_bad_message` when `error.is_malicious_response()` is true
  - Track consecutive empty response retries with `empty_response_retries` counter
  - Emit a warning after exceeding `MAX_EMPTY_RESPONSE_RETRIES` (5) to surface blocks that are too large for any peer to serve
  - Reset retry counter on successful (non-empty) response

## Testing
- `cargo test -p reth-downloaders --lib bodies` — 12/12 tests pass
- `cargo test -p reth-network-p2p` — 7/7 tests pass

Prompted by: mattsse

### DIFF
--- a/crates/net/downloaders/src/bodies/request.rs
+++ b/crates/net/downloaders/src/bodies/request.rs
@@ -17,6 +17,11 @@ use std::{
     task::{ready, Context, Poll},
 };
 
+/// Maximum number of times we'll retry after receiving empty responses from different peers
+/// before giving up on a request. This prevents infinite retry loops when a block body is
+/// truly unservable (e.g. exceeds all peers' response size limits).
+const MAX_EMPTY_RESPONSE_RETRIES: usize = 5;
+
 /// Body request implemented as a [Future].
 ///
 /// The future will poll the underlying request until fulfilled.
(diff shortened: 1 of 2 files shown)


mattsse (commented):
this isnt right

PR #22312, opened 2026-02-18
fix(prune): correct deleted entries count when skip_filter is used

reallesee:
`prune_table_with_range` incremented `deleted_entries` on every iteration, even when the entry was skipped and not deleted. This caused inflated pruning stats in `ReceiptsByLogs` where skip_filter preserves certain receipts.

### DIFF
--- a/crates/prune/prune/src/db_ext.rs
+++ b/crates/prune/prune/src/db_ext.rs
@@ -10,7 +10,7 @@ use tracing::debug;
 
 /// Result of a single prune step in [`DbTxPruneExt::prune_table_with_range_step`].
 #[derive(Debug, Clone, Copy)]
-struct PruneStepResult {
+pub(crate) struct PruneStepResult {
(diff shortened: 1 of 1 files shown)


crates/prune/prune/src/db_ext.rs
```diff
@@ -120,18 +125,19 @@ pub(crate) trait DbTxPruneExt: DbTxMut + DbTx {
         limiter: &mut PruneLimiter,
         skip_filter: &mut impl FnMut(&TableRow<T>) -> bool,
         delete_callback: &mut impl FnMut(TableRow<T>),
-    ) -> Result<bool, DatabaseError> {
-        let Some(res) = walker.next() else { return Ok(true) };
+    ) -> Result<(bool, bool), DatabaseError> {
```
DaniPopes: please use a struct, this is not readable

DaniPopes (approved):

PR #22322, opened 2026-02-18
perf(trie): replace Vec<u8> with SmallVec for leaf values

yongkangc:
**Summary**
This PR replaces trie leaf update values from `Vec<u8>` to `LeafValue` (`SmallVec<[u8; 128]>`) in `reth-trie-sparse` hot update paths to reduce per-update heap allocations. The previous path repeatedly allocated via `encode_fixed_size(...).to_vec()` and `account_rlp_buf.clone()` during state-root processing; this keeps the same behavior while avoiding those allocations in the common case.

**Expected Impact**
- Lower allocation/allocator overhead in sparse-trie account and storage leaf updates.
- No intended correctness or semantic changes.
- `LeafValue` capacity is chosen to fit typical leaf payloads inline (account RLP up to 110 bytes, storage slot RLP up to 33 bytes).

**Notes**
`proofs.rs` verification path still uses `Option<Vec<u8>>` in `reth-trie-common::verify_proof` and is intentionally out of scope for this PR.

**Microbench**
- `account_materialize`: ~73% faster
- `clone_retry_insert`: ~51% faster
- `storage_update` (end-to-end trie update): ~flat (avg -0.6%, within noise)

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -27,7 +27,7 @@ use reth_trie_parallel::{
 #[cfg(feature = "trie-debug")]
 use reth_trie_sparse::debug_recorder::TrieDebugRecorder;
 use reth_trie_sparse::{
-    errors::SparseTrieResult, DeferredDrops, LeafUpdate, ParallelSparseTrie, SparseStateTrie,
+    errors::SparseTrieResult, DeferredDrops, LeafUpdate, LeafValue, ParallelSparseTrie, SparseStateTrie,
     SparseTrie,
 };
 use revm_primitives::{hash_map::Entry, B256Map};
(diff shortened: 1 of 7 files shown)


yongkangc:
## Closing rationale

Thanks for the effort here! After analysis, we think this needs to be reworked before merging:

### CI Issues (fixable)
- 4× `clippy::useless_conversion` — `value_encoded().into()` where `value_encoded()` already returns `LeafValue`
- 1× `clippy::redundant_clone` — `wrong_value.clone()` in test assertion
- e2e failure is infra/network noise (connect timeout), unrelated

### Design Concern (blocking)
`SmallVec<[u8; 128]>` makes each entry ~137 bytes vs 24 bytes for `Vec<u8>`. In a `B256Map<LeafUpdate>` with ~2000 storage entries:
- **Vec**: 2000 × 24 = 48 KB (fits L1/L2 cache)
- **SmallVec<128>**: 2000 × 137 = 274 KB (spills L2, hits L3)

That is 5.7× worse cache footprint on every HashMap insert/lookup/iteration, which likely offsets the allocator savings (~75µs from ~2500 small allocs).

### Suggested path forward
Split into two smaller PRs:

1. **PR A (safe win)**: Replace `account_rlp_buf.clone()` with `LeafValue::from_slice()` — this eliminates the `BytesMut` clone per account update with no structural changes.

2. **PR B (experimental)**: If retrying the SmallVec approach, use `SmallVec<[u8; 33]>` (42 bytes/entry → 84 KB for 2000 entries). This inlines ~80% of updates (storage slots ≤33 bytes) while only spilling accounts (~20%) to heap. **Must include reth-bench comparison** to prove net benefit.

The theoretical savings are ~50-75µs per block against a ~30ms payload (~0.2%), so benchmarks are needed to justify the complexity.

yongkangc:
derek bench

PR #22327, opened 2026-02-18
refactor(trie): remove update_leaf and remove_leaf from SparseTrie trait

mediocregopher:
Remove update_leaf and remove_leaf from the SparseTrie trait and make them private methods on ParallelSparseTrie that no longer accept a TrieNodeProvider. Where the provider would previously have been used to reveal blinded nodes inline, the methods now return BlindedNode or NodeNotFoundInProvider errors directly, matching the behavior of NoRevealProvider.

- Remove update_leaf/remove_leaf from SparseTrie trait in traits.rs
- Move implementations to private inherent methods on ParallelSparseTrie
- Simplify pre_validate_reveal_chain and reveal_remaining_child_on_leaf_removal to return errors immediately for blinded nodes instead of attempting provider-based reveals
- Remove update_leaf/remove_leaf from RevealableSparseTrie
- Remove update_account_leaf, update_storage_leaf, remove_account_leaf, remove_storage_leaf from SparseStateTrie
- Update witness.rs to use update_leaves API and remove WitnessTrieNodeProviderFactory/WitnessTrieNodeProvider
- Remove tests for inline node revealing (no longer supported)
- Update all remaining tests to use new signatures

NOTE: I highly recommend reviewing this PR by checking it out and comparing it to `main` using `--patience`:

```
git checkout origin/mediocregopher/more-trie-cleanup
git diff origin/main..HEAD --patience
```

The update_leaf/remove_leaf definitions have been moved out of the `SparseTrie` impl essentially unchanged, but by default `git diff` (and therefore github) doesn't handle that very well.

### DIFF
--- a/crates/trie/trie/src/witness.rs
+++ b/crates/trie/trie/src/witness.rs
@@ -1,13 +1,13 @@
 use crate::{
     hashed_cursor::{HashedCursor, HashedCursorFactory},
     prefix_set::TriePrefixSetsMut,
-    proof::{Proof, ProofTrieNodeProviderFactory},
+    proof::Proof,
     trie_cursor::TrieCursorFactory,
 };
 use alloy_rlp::EMPTY_STRING_CODE;
 use alloy_trie::EMPTY_ROOT_HASH;
 use reth_trie_common::HashedPostState;
-use reth_trie_sparse::SparseTrie;
+use reth_trie_sparse::{LeafUpdate, SparseTrie};
 
 use alloy_primitives::{
     keccak256,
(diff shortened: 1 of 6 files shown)


crates/trie/trie/src/witness.rs
```diff
@@ -159,45 +149,64 @@ where
...
+                alloy_rlp::Encodable::encode(
+                    &account.into_trie_account(storage_root),
+                    &mut account_rlp,
+                );
+                let mut account_updates = B256Map::default();
+                account_updates.insert(hashed_address, LeafUpdate::Changed(account_rlp));
+                sparse_trie
+                    .trie_mut()
+                    .update_leaves(&mut account_updates, |_, _| {})
+                    .map_err(SparseStateTrieErrorKind::from)?;
```
klkvr, line 201: let's add a note that we assume that multiproof reveal above should've revealed all required nodes?

also should we update the multiproof invocation above to configure `AddedRemovedKeys` with `assume_added = true`?

also do we even need to perform the trie updates here? it doesn't look like we even read anything from `sparse_trie` anymore, iiuc before it was only used so that provider could accumulate extra nodes that are needed in the witness

klkvr (commented):
nice, only have a question

PR #22333, opened 2026-02-18
refactor: remove duplicate `apply_pre_execution_changes` from `Trace` trait

stevencartavia:
Removes the hand-rolled `apply_pre_execution_changes` method from the `Trace` trait, which manually called `SystemCaller` to apply EIP-2935/EIP-4788 system calls.

All 5 call sites (2 in `trace.rs`, 3 in `debug.rs`) now delegate to `ConfigureEvm::executor_for_block`, reusing the canonical `BlockExecutor` path used by payload building and block validation.

Fixes old code was missing `set_state_clear_flag` (EIP-161) that `EthBlockExecutor::apply_pre_execution_changes` handles.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -1,25 +1,24 @@
 //! Loads a pending block from database. Helper trait for `eth_` call and trace RPC methods.
 
 use super::{Call, LoadBlock, LoadState, LoadTransaction};
-use crate::FromEvmError;
+use crate::{FromEthApiError, FromEvmError};
 use alloy_consensus::{transaction::TxHashRef, BlockHeader};
 use alloy_primitives::B256;
 use alloy_rpc_types_eth::{BlockId, TransactionInfo};
 use futures::Future;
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -408,26 +417,4 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
     {
         self.trace_block_until_with_inspector(block_id, block, None, insp_setup, f)
     }
-
-    /// Applies chain-specific state transitions required before executing a block.
-    ///
-    /// Note: This should only be called when tracing an entire block vs individual transactions.
-    /// When tracing transaction on top of an already committed block state, those transitions are
-    /// already applied.
-    fn apply_pre_execution_changes<DB: Send + Database + DatabaseCommit>(
```
DaniPopes, line 417: why can't we just keep this method and simplify it here instead of 3 map_errs at all callsites?

stevencartavia: true, better!

DaniPopes (approved):

PR #22342, opened 2026-02-19
perf(nippy-jar): enable MADV_HUGEPAGE on static file mmaps

gakonst:
## Summary

Enable Transparent Huge Pages on nippy-jar (static file) mmap regions via `madvise(MADV_HUGEPAGE)` to reduce TLB misses on large sequential scans.

## Motivation

libmdbx explicitly disables huge pages (`MADV_NOHUGEPAGE`) on its B-tree mmaps because random 4KB page access makes THP counterproductive there. Static files are the opposite: large, immutable, append-only, and frequently scanned sequentially (e.g. `eth_getLogs`, historical RPC queries, sync). THP reduces TLB pressure from ~26M page table entries (100GB / 4KB) to ~50K (100GB / 2MB).

## Changes

- Call `Mmap::advise(Advice::HugePage)` on data and offset mmaps after creation in `DataReader::new()`
- Linux-only (`#[cfg(target_os = "linux")]`), no-op on other platforms
- Errors silently ignored (kernel may not have `CONFIG_TRANSPARENT_HUGEPAGE`)

## Testing

Needs benchmark on archive node with heavy RPC load. `perf stat -e dTLB-load-misses` would show the difference.

Prompted by: DaniPopes


decofe:
cc @DaniPopes

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/22480141637/job/65116059832)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `danipopes/hugepage-static-files`

PR #22344, opened 2026-02-19
fix(cli): store extradata as Bytes, decode hex in parser

### LINKED ISSUE #22274: --builder.extradata rejects valid 32-byte hex-encoded values

gakonst:
## Summary
Fixes `--builder.extradata` rejecting valid 32-byte hex-encoded values. Rewrites #22277 to store parsed bytes directly instead of deferring hex decoding to `extra_data_bytes()`.

## Motivation
Closes #22274 — the CLI parser validated string character count instead of decoded byte count, so a 32-byte SHA256 hash (`0x` + 64 hex chars) was rejected.

## Changes
- `PayloadBuilderArgs::extra_data` is now `Bytes` instead of `String`
- `ExtraDataValueParser` decodes at parse time: `0x`-prefixed → `hex::decode`, otherwise → `to_vec()`
- `PayloadBuilderConfig::extra_data() -> Cow<str>` removed, `extra_data_bytes() -> Bytes` is now the required method (just returns a clone)
- Added tests for valid hex, oversized hex, invalid hex, and odd-length hex

## Testing
```
cargo test -p reth-node-core
# 98 passed; 0 failed
```

Prompted by: danipopes

### DIFF
--- a/crates/ethereum/node/src/payload.rs
+++ b/crates/ethereum/node/src/payload.rs
@@ -55,7 +55,7 @@ where
             EthereumBuilderConfig::new()
                 .with_gas_limit(gas_limit)
                 .with_max_blobs_per_block(conf.max_blobs_per_block())
-                .with_extra_data(conf.extra_data_bytes()),
+                .with_extra_data(conf.extra_data()),
         ))
     }
 }
--- a/crates/node/core/src/args/payload_builder.rs
+++ b/crates/node/core/src/args/payload_builder.rs
@@ -145,7 +145,7 @@ impl Default for PayloadBuilderArgs {
 }
 
 impl PayloadBuilderConfig for PayloadBuilderArgs {
-    fn extra_data_bytes(&self) -> Bytes {
+    fn extra_data(&self) -> Bytes {
         self.extra_data.clone()
     }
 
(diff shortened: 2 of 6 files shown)


DaniPopes (approved):

mattsse (approved):
makes sense

PR #22347, opened 2026-02-19
perf(tasks): add LazyHandle<T>, use for hash post state

gakonst:
## Summary

Add `LazyHandle<T>` — a generic, Arc'd handle returned by `spawn_blocking_named` that resolves lazily on first `.get()` call and caches the result via `OnceLock`.

Use it to defer `hashed_post_state` computation (keccak256 hashing of all changed addresses/slots) — spawned in the background right after execution, resolved lazily when first needed.

## Changes

**`reth-tasks`**:
- New `LazyHandle<T>` type in `crates/tasks/src/lazy.rs`
  - Arc'd internally with `OnceLock` + `Mutex<Option<Receiver>>`
  - `.get(&self) -> &T` — blocks on first call, cached on subsequent
  - `.into_inner(self) -> T` — avoids clone when sole owner via `Arc::try_unwrap`
  - `::ready(value)` — pre-resolved constructor for testing and immediate values
  - `Clone` + `Debug` implemented
- `Runtime::spawn_blocking_named` now returns `LazyHandle<R>`

**`reth-engine-primitives`**:
- `PayloadValidator::validate_block_post_execution_with_hashed_state` now takes `&LazyHandle<HashedPostState>` instead of `&HashedPostState`

**`reth-chain-state`**:
- `DeferredTrieData::pending` now takes `LazyHandle<HashedPostState>` instead of `Arc<HashedPostState>`
- `sort_and_build_trie_input` takes `LazyHandle<HashedPostState>`, resolves at the point of consumption

**`reth-engine-tree`**:
- `hashed_post_state` spawned in background immediately after execution
- `type LazyHashedPostState = LazyHandle<HashedPostState>` threaded through all validator methods
- `.get()` only called at actual consumption points: `compute_state_root_parallel`, `compute_state_root_serial`, `sort_and_build_trie_input`
- `spawn_deferred_trie_task` passes the lazy handle directly to `DeferredTrieData` without resolving

## Testing

`cargo test -p reth-tasks -- lazy` (5 tests), `cargo test -p reth-chain-state -- deferred` (14 tests), `cargo check -p reth-node-ethereum`

Prompted by: DaniPopes


DaniPopes (approved):
LGTM!

PR #22348, opened 2026-02-19
perf(engine): multi-slot preserved sparse trie cache

yongkangc:
## Problem

`SharedPreservedSparseTrie` stores only a single trie (`Option<PreservedSparseTrie>`). Any continuity break or race drops to a cold/cleared trie, losing all cached trie node data. This is particularly costly during short reorgs and parallel payload validations where multiple parent state roots may be active.

## Solution

Replace the single-slot `Option<PreservedSparseTrie>` with a bounded `VecDeque` cache (`PreservedTrieCache`) keyed by state root. The cache stores up to 3 anchored tries and one cleared fallback.

### Cache lookup priority (`take_for_parent()`):
1. **Exact anchored match** — state root == parent state root (warm continuation)
2. **Cleared trie** — allocations preserved, data cleared (allocation reuse)
3. **Evict oldest anchored** — clear and reuse oldest entry's allocations

### Key changes:
- **`preserved_sparse_trie.rs`**: New `PreservedTrieCache` struct with `VecDeque<(B256, SparseTrie)>` for anchored tries plus `Option<SparseTrie>` for cleared fallback. `PreservedSparseTrie` enum removed — its logic is absorbed into the cache.
- **`mod.rs`**: Call sites updated from `take()`/`store()` to `take_for_parent(parent_state_root)`/`store_anchored()`/`store_cleared()`.
- **Tests**: 6 unit tests covering exact match, cleared fallback, eviction, capacity bounds, empty cache, and deduplication.

## Verification
- `cargo +nightly fmt --all` ✅
- `cargo clippy -p reth-engine-tree --all-features` ✅ (0 warnings)
- `cargo nextest run -p reth-engine-tree` ✅ (114/114 passed)

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -60,7 +60,7 @@ pub mod prewarm;
 pub mod receipt_root_task;
 pub mod sparse_trie;
 
-use preserved_sparse_trie::{PreservedSparseTrie, SharedPreservedSparseTrie};
+use preserved_sparse_trie::SharedPreservedSparseTrie;
 
 /// Default parallelism thresholds to use with the [`ParallelSparseTrie`].
 ///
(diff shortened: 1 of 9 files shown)


yongkangc:
Closing — after deeper analysis, this is **net negative for mainnet steady-state**.

CPU overhead of the VecDeque scan is negligible, but retaining 3 full SparseStateTrie instances (~250MB+ each) is an always-on memory tax:
- Reduces OS page cache → more disk I/O for proof/state reads
- Increases TLB/cache pressure across the working set
- Increases allocator fragmentation

Mainnet reorgs are extremely rare (<<1%), so the benefit of multi-slot caching almost never materializes, while the memory cost is paid every block.

PR #22350, opened 2026-02-19
perf(engine): release preserved trie lock during expensive prepare-for-reuse work

yongkangc:
## Problem

The preserved sparse trie mutex is held during expensive `into_trie_for_reuse()` work (prune + shrink), serializing the next payload on trie cleanup. This means the next block's `take()` blocks on the mutex for the entire duration of trie preparation, preventing any overlap between payload processing and trie cleanup.

## Solution

Replace the mutex-held pattern with a 2-phase Condvar-backed protocol:

1. **Lock phase** (brief): Acquire mutex → `send()` result → set `preparing` flag → release mutex
2. **Prepare phase** (expensive, unlocked): Run `into_trie_for_reuse()` (prune + shrink) outside the lock → `store()` result and clear flag via `PreparingGuard`

The next payload's `take()` and `wait_for_availability()` now block on the Condvar while `preparing` is true, rather than contending on the mutex itself.

### Key design decisions

- **`PreparingGuard` RAII safety**: If the task panics or returns early after marking `preparing`, the Drop impl clears the flag and notifies waiters, preventing deadlock.
- **`mark_preparing_and_release()`**: Single method on `PreservedTrieGuard` that atomically sets the flag and drops the mutex guard, returning a `PreparingGuard`.
- **`PreservedTrieInner`**: New struct wrapping `preparing: bool` + `slot: Option<PreservedSparseTrie>` behind `Arc<(Mutex<_>, Condvar)>`.

### Changes

- `preserved_sparse_trie.rs`: New `PreservedTrieInner`, `PreparingGuard`, Condvar-based blocking in `take()`/`wait_for_availability()`, `mark_preparing_and_release()` on guard, plus unit tests
- `mod.rs`: Restructured lock usage to release before expensive work

### Tests

4 new unit tests covering:
- `take()` blocks while preparing, unblocks after `store()`
- `wait_for_availability()` blocks while preparing
- `PreparingGuard` drop (without commit) clears flag
- `mark_preparing_and_release()` round-trip

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -621,11 +621,17 @@ where
                 return;
             }
 
+            // Mark preparing and release the lock so the next payload's take() blocks
+            // on the condvar instead of the mutex. Expensive trie preparation runs
+            // unlocked below; the PreparingGuard ensures the flag is cleared even on panic.
+            let preparing_guard =
+                guard.mark_preparing_and_release(&preserved_sparse_trie);
+
             // Only preserve the trie as anchored if computation succeeded.
             // A failed computation may have left the trie in a partially updated state.
             let _enter =
                 debug_span!(target: "engine::tree::payload_processor", "preserve").entered();
-            let deferred = if let Some(state_root) = computed_state_root {
+            if let Some(state_root) = computed_state_root {
                 let start = Instant::now();
                 let (trie, deferred) = task.into_trie_for_reuse(
                     prune_depth,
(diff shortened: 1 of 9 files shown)


yongkangc:
Closing — after deeper analysis, this is **neutral to slightly negative** in the common sequential mainnet case.

The next payload still blocks on the Condvar for the full duration of prune+shrink — the blocking primitive changed but the blocking duration didn't. The only win would be if other operations could use the mutex while preparing runs, but nothing else contends for it.

Added costs:
- Extra lock/unlock cycle (futex round-trip if contended)
- Condvar bookkeeping + potential syscall overhead
- Added complexity / deadlock surface area

Net: no reduction in critical-path latency for next payload start.

PR #22352, opened 2026-02-19
docs(engine): document payload validation flow with spawn/await map

gakonst:
Documents the current payload validation flow in `payload_validator.rs`, including BAL handling, state-root strategies, spawned background work, and payload attributes validation.

Replaces the inline ASCII timeline with a Mermaid sequence diagram and refreshes the spawn/await map after merging current `main`.


yongkangc (commented):
generally think this is a good idea

emmajam (approved):

emmajam (approved):

PR #22355, opened 2026-02-19
perf(trie): cache pending sparse trie target count

yongkangc:
Removes repeated O(n) rescans of pending_targets in a hot loop and replaces them with O(1) counter reads.

**Summary**
Sparse trie processing repeatedly computed `pending_targets.chunking_length()` in the run loop and dispatch path, rescanning queued storage targets each time. This change keeps an incremental `pending_targets_len` counter updated as targets are enqueued and uses the cached value for threshold checks and dispatch sizing.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -90,10 +90,8 @@ pub(super) struct SparseTrieCacheTask<A = ParallelSparseTrie, S = ParallelSparse
     account_rlp_buf: Vec<u8>,
     /// Whether the last state update has been received.
     finished_state_updates: bool,
-    /// Pending targets to be dispatched to the proof workers.
-    pending_targets: MultiProofTargetsV2,
-    /// Cached number of account + storage proof targets currently queued in `pending_targets`.
-    pending_targets_len: usize,
+    /// Pending proof targets queued for dispatch to proof workers.
+    pending_targets: PendingTargets,
     /// Number of pending execution/prewarming updates received but not yet passed to
     /// `update_leaves`.
     pending_updates: usize,
@@ -143,7 +141,6 @@ where
             account_rlp_buf: Vec::with_capacity(TRIE_ACCOUNT_RLP_MAX_SIZE),
             finished_state_updates: Default::default(),
             pending_targets: Default::default(),
-            pending_targets_len: Default::default(),
             pending_updates: Default::default(),
             metrics,
         }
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -92,6 +92,8 @@ pub(super) struct SparseTrieCacheTask<A = ParallelSparseTrie, S = ParallelSparse
     finished_state_updates: bool,
     /// Pending targets to be dispatched to the proof workers.
     pending_targets: MultiProofTargetsV2,
+    /// Cached number of account + storage proof targets currently queued in `pending_targets`.
+    pending_targets_len: usize,
```
mattsse: I find cached a bit confusing here, 

I also worry about messing up the tracking here.

unsure how we currently track `pending_targets`, but wouldnt it make sense to do something equivalent to `Option<struct PendingTargets {targets, count}`

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -669,8 +675,11 @@ where
...
     fn dispatch_pending_targets(&mut self) {
-        if !self.pending_targets.is_empty() {
-            let chunking_length = self.pending_targets.chunking_length();
+        // Guard against cached-length drift if future enqueue paths miss counter updates.
+        debug_assert_eq!(self.pending_targets_len, self.pending_targets.chunking_length());
+        if self.pending_targets_len > 0 {
+            let chunking_length = self.pending_targets_len;
+            self.pending_targets_len = 0;
             dispatch_with_chunking(
                 std::mem::take(&mut self.pending_targets),
```
mattsse: then we can replace this with a take

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
yongkangc: @mattsse  yeah good point, rn we have a counter, but that might be easy to missed. 

Think with ur suggestion `PendingTargets` much cleaner now. We dont need `Option` because the targets are always enqueued and drained in the main loop - no none state, but theres a empty state i.e 2 empty vecs

addressed in : b17fc87

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
yongkangc: addressed in : https://github.com/paradigmxyz/reth/commit/b17fc87d1c194f78b32fc6414d799763bbf3d82c

PR #22374, opened 2026-02-19
fix(rpc): return -32602 for PayloadAttributes structure validation errors

### LINKED ISSUE #8732: FCU v2 with `null` withdrawals causes reth to return `-38003 Invalid payload attributes` instead of `-32602 Invalid params`
Running the following tests:
```
./hive --sim ethereum/engine --client reth --sim.limit "engine-withdrawals/"
```

The simulator returns:
```
>> (c8ffb2c6) {"jsonrpc":"2.0","id":1,"method":"engine_forkchoiceUpdatedV2","params":[{"headBlockHash":"0xf62b6ae287c6cdc2f1bda80d0ce83cff49aa8665d0477605f9c157355885f511","safeBlockHash":"0x0000000000000000000000000000000000000000000000000000000000000000","finalizedBlockHash":"0x0000000000000000000000000000000000000000000000000000000000000000"},{"timestamp":"0x1235","prevRandao":"0x0000000000000000000000000000000000000000000000000000000000000000","suggestedFeeRecipient":"0x0000000000000000000000000000000000000000","withdrawals":null,"parentBeaconBlockRoot":null}]}
<< (c8ffb2c6) {"jsonrpc":"2.0","error":{"code":-38003,"message":"Invalid payload attributes","data":{"err":"Payload attributes validation error: no withdrawals post-Shanghai"}},"id":1}
DEBUG (Withdrawals Fork On Genesis): Failed `Expect*` routine called from: file=/source/suites/withdrawals/tests.go, line=1284
DEBUG (Withdrawals Fork On Genesis): Sent shanghai fcu using PayloadAttributesV1, error is expected
FAIL (Withdrawals Fork On Genesis): Expected error code on EngineForkchoiceUpdatedV2: want=-32602, got=-38003
```

This is because we return `-38003 Invalid payload attributes`. IMO this makes sense to return, but it is out of spec, which is mainly why we do not pass the `engine-withdrawals` suite.

We may not need to change this, depending on what happens to the spec.

emmajam:
Per the Engine API spec, structure validation errors for PayloadAttributes (e.g., missing withdrawals post-Shanghai, missing parentBeaconBlockRoot post-Cancun) should return `-32602` ("Invalid params"), not `-38003`.

The `-38003` code is reserved for semantic validation errors that occur after structure validation passes (e.g., `ForkchoiceUpdateError::UpdatedInvalidPayloadAttributes`).

This fixes 10+ hive `engine-withdrawals` test failures:
```
Expected error code on EngineForkchoiceUpdatedV2: want=-32602, got=-38003
```

Ref: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md
Fixes #8732

### DIFF
--- a/crates/rpc/rpc-engine-api/src/error.rs
+++ b/crates/rpc/rpc-engine-api/src/error.rs
@@ -14,8 +14,6 @@ use thiserror::Error;
 /// The Engine API result type
 pub type EngineApiResult<Ok> = Result<Ok, EngineApiError>;
 
-/// Invalid payload attributes code.
-pub const INVALID_PAYLOAD_ATTRIBUTES: i32 = -38003;
 /// Payload unsupported fork code.
 pub const UNSUPPORTED_FORK_CODE: i32 = -38005;
 /// Payload unknown error code.
@@ -25,9 +23,6 @@ pub const REQUEST_TOO_LARGE_CODE: i32 = -38004;
 
 /// Error message for the request too large error.
 const REQUEST_TOO_LARGE_MESSAGE: &str = "Too large request";
-
-/// Error message for the request too large error.
-const INVALID_PAYLOAD_ATTRIBUTES_MSG: &str = "Invalid payload attributes";
 
 /// Error returned by [`EngineApi`][crate::EngineApi]
 ///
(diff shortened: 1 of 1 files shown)


mattsse (approved):
> The -38003 code is reserved for semantic validation errors that occur after structure validation passes (e.g., ForkchoiceUpdateError::UpdatedInvalidPayloadAttributes).



now this finally makes sense

PR #22379, opened 2026-02-19
feat(storage): slot preimage DB for plain changeset keys in v2

gakonst:
## Summary

In `--storage.v2` mode, storage changesets record hashed slot keys. This PR adds a slot preimage database (`keccak256(slot) → slot`) that the `ExecutionStage` populates, enabling reverse lookups so changesets can always use plain (unhashed) keys.

### Key changes

- **Slot preimage MDBX**: A separate auxiliary database (`preimage.dat`) written during execution, storing `keccak256(slot) → plain_slot` mappings
- **`StoragePath` trait**: Derives the preimage DB path from the provider at runtime (via `Database::path()`), removing the need to thread a `PathBuf` through `ExecutionStage`
- **Remove `StorageSlotKey` / `ChangesetEntry`**: With all changeset keys now plain, the hashed-key abstraction is no longer needed
- **Remove `storage_by_hashed_key`**: No longer needed since storage lookups always use plain keys
- **Restore `populate_bundle_state_hashed`**: Reads current state from `HashedAccounts`/`HashedStorages` during v2 unwind (instead of empty `PlainAccountState`), with storage keys hashed via `keccak256` for DB lookups

## Test plan

- [x] `test_write_and_remove_state_roundtrip_v2` — v2 unwind correctly removes accounts/storage
- [x] `test_get_state_storage_value_plain_state` — plain state storage lookups
- [x] `cargo +nightly clippy --workspace --all-features`
- [x] `no_std` / wasm compatibility (`StoragePath` gated behind `std` feature)


crates/stages/stages/src/stages/slot_preimages.rs
```diff
@@ -0,0 +1,106 @@
...
+        builder.set_max_dbs(1);
+        builder.set_geometry(Geometry {
+            size: Some(256 * 1024..64 * 1024 * 1024 * 1024),
+            ..Default::default()
+        });
+        builder.write_map();
+        builder.set_flags(EnvironmentFlags {
+            mode: Mode::ReadWrite { sync_mode: SyncMode::Durable },
+            ..Default::default()
+        });
```
joshieDo: just replicate the existing mdbx settings/env

crates/stages/stages/src/stages/slot_preimages.rs
joshieDo, line 1: stages/execution/slot_preimages.rs
stages/execution/mod.rs

crates/stages/stages/src/stages/execution.rs
```diff
@@ -91,6 +99,13 @@ where
     exex_manager_handle: ExExManagerHandle<E::Primitives>,
     /// Executor metrics.
     metrics: ExecutorMetrics,
+    /// Path for the auxiliary slot-preimage MDBX database.
+    ///
+    /// When set, enables storage-slot preimage tracking for pre-Cancun selfdestruct handling
+    /// in `--storage.v2` mode so that changesets always contain plain (unhashed) storage keys.
+    slot_preimages_path: Option<PathBuf>,
+    /// Lazily-opened slot preimage store.
+    slot_preimages: Option<SlotPreimages>,
```
joshieDo: could this NOT be here ? we can store it in the existing datadir/db path with a different name.

we can lazily init inside the run execution stage (no need to handle it for unwind), if block start is pre cancun.

if block start is post cancun - rm -rf it

crates/stages/stages/src/stages/execution.rs
```diff
@@ -254,6 +281,122 @@ where
...
     }
+
+    /// Collects `keccak256(slot) → slot` preimage entries from the bundle state and stores
+    /// them in the auxiliary preimage database, then rewrites wipe reverts for self-destructed
+    /// accounts to use plain slot keys instead of relying on the hashed-storage DB walk.
+    ///
+    /// This eliminates the need for the changeset writer to read from `HashedStorages` during
+    /// storage wipes, keeping all changeset keys in plain format.
+    fn inject_plain_wipe_slots<P: DBProvider>(
+        &mut self,
```
joshieDo: shouldnt need to be part of self i think and can go to slot_image.rs

crates/stages/stages/src/stages/slot_preimages.rs
joshieDo: also can we not use current Database and cursor semantics here ?

crates/stages/stages/src/stages/slot_preimages.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        trace!(target: "stages::slot_preimages", ?path, "Opened slot-preimage store");
+
+        Ok(Self { env })
+    }
+
+    /// Batch-insert `hashed_slot → plain_slot` preimage entries.
+    ///
+    /// Duplicate keys are silently skipped via [`WriteFlags::NO_OVERWRITE`].
+    pub(super) fn insert_preimages(&self, entries: &[(B256, B256)]) -> eyre::Result<()> {
```
joshieDo: pre sort them

crates/trie/db/src/state.rs
```diff
@@ -267,8 +265,7 @@ impl<'a, TX: DbTx> DatabaseStateRoot<'a, TX>
 
 /// Calls [`HashedPostStateSorted::from_reverts`].
 ///
-/// This is a convenience wrapper kept for backward compatibility. The storage
-/// key tagging is now handled internally by the changeset reader.
+/// This is a convenience wrapper kept for backward compatibility.
```
joshieDo: can rm comment

crates/trie/db/src/state.rs
```diff
@@ -143,10 +143,8 @@ pub trait DatabaseHashedPostState: Sized {
     /// Initializes [`HashedPostStateSorted`] from reverts. Iterates over state reverts in the
     /// specified range and aggregates them into sorted hashed state.
     ///
-    /// Storage keys from changesets are tagged as
-    /// [`Plain`](reth_primitives_traits::StorageSlotKey::Plain) or
-    /// [`Hashed`](reth_primitives_traits::StorageSlotKey::Hashed) by the reader, so no
-    /// `use_hashed_state` flag is needed. Addresses are always hashed.
+    /// Storage keys in changesets are always plain (unhashed) and are hashed via
+    /// `keccak256` for trie use. Addresses are always hashed.
```
joshieDo: can rm doc

crates/trie/db/src/prefix_set.rs
```diff
@@ -20,10 +20,8 @@ use reth_trie::{
 /// Load prefix sets using a provider that implements [`ChangeSetReader`]. This function can read
 /// changesets from both static files and database.
 ///
-/// Storage keys from changesets are tagged as
-/// [`Plain`](reth_primitives_traits::StorageSlotKey::Plain)
-/// or [`Hashed`](reth_primitives_traits::StorageSlotKey::Hashed) by the reader, so callers need
-/// not pass a `use_hashed_state` flag. Addresses are always hashed.
+/// Storage keys in changesets are always plain (unhashed) and are hashed via `keccak256` for
+/// trie prefix computation. Addresses are always hashed.
```
joshieDo: can rm doc

crates/stages/stages/src/stages/execution/mod.rs
```diff
@@ -91,6 +94,11 @@ where
     exex_manager_handle: ExExManagerHandle<E::Primitives>,
     /// Executor metrics.
     metrics: ExecutorMetrics,
+    /// Path for the auxiliary slot-preimage MDBX database.
+    ///
+    /// When set, enables storage-slot preimage tracking for pre-Cancun selfdestruct handling
+    /// in `--storage.v2` mode so that changesets always contain plain (unhashed) storage keys.
+    slot_preimages_path: Option<PathBuf>,
```
joshieDo: do we still need this ? is that how ETL has the path eg. ?

joshieDo: couldnt we get the one from the main db, so we dont need another field on the stage?

crates/stages/stages/src/stages/execution/slot_preimages.rs
```diff
@@ -0,0 +1,218 @@
...
+impl SlotPreimages {
+    /// Opens (or creates) the slot-preimage MDBX environment at the given `path`.
+    ///
+    /// Uses `WriteMap` mode for performance. The geometry allows growth up to 64 GB to
+    /// accommodate chains with many unique storage keys (mapping-derived keys especially).
+    fn open(path: &Path) -> eyre::Result<Self> {
+        let mut builder = Environment::builder();
+        builder.set_max_dbs(1);
+        builder.set_geometry(Geometry {
+            size: Some(256 * 1024..64 * 1024 * 1024 * 1024),
```
joshieDo: set the same exact environment builder fields as the main mdbx - is this it? i

crates/stages/stages/src/stages/execution/slot_preimages.rs
```diff
@@ -0,0 +1,218 @@
...
+            let plain = B256::from(slot_key.to_be_bytes());
+            let hashed = keccak256(plain);
+            if seen_hashes.insert(hashed) {
+                preimage_entries.push((hashed, plain));
+            }
+        }
+    }
+
+    // Pre-sort entries by hash key for optimal MDBX insert performance.
+    preimage_entries.sort_unstable_by_key(|(hash, _)| *hash);
```
joshieDo: par_sort_unstable

crates/stages/stages/src/stages/execution/slot_preimages.rs
```diff
@@ -0,0 +1,223 @@
...
+/// storage wipes, keeping all changeset keys in plain format.
+pub(super) fn inject_plain_wipe_slots<P: DBProvider, R>(
+    slot_preimages_path: &Path,
+    provider: &P,
+    state: &mut ExecutionOutcome<R>,
+) -> Result<(), StageError> {
+    // Collect preimage entries from all storage changes in the bundle.
+    // StorageKey in revm is U256, representing a plain EVM slot index.
+    let mut preimage_entries = Vec::new();
+    let mut seen_hashes = HashSet::new();
```
yongkangc, line 133: nit: should we consider using `FxHashSet` here?


> the keys are B256 (keccak256 outputs), which are already uniformly distributed. SipHash's anti-collision protection is wasted on them. FxHash will be faster on every insert/lookup.

> It's ~2-3x faster for small fixed-size keys like B256/[u8; 32] because it skips the DOS-resistance overhead of SipHash that you don't need for non-user-facing internal data structures.

yongkangc (approved):
nice, guess most of the file changes is just removing code for hash -> plainstate type conversion

crates/stages/stages/src/stages/execution/mod.rs
```diff
@@ -462,6 +467,26 @@ where
...
+                .header_by_number(start_block)?
+                .ok_or_else(|| ProviderError::HeaderNotFound(start_block.into()))?;
+
+            let path = provider.storage_path().join("preimage");
+            if !provider.chain_spec().is_cancun_active_at_timestamp(start_header.timestamp()) {
+                slot_preimages::inject_plain_wipe_slots(&path, provider, &mut state)?;
+            } else if path.exists() {
+                // Post-Cancun: no more self-destructs, preimage db is no longer needed.
+                let _ = std::fs::remove_dir_all(&path);
+            }
```
klkvr, line 487: is it cheap enough to do this on every iteration?

crates/stages/stages/tests/pipeline.rs
```diff
@@ -90,9 +90,14 @@ fn assert_changesets_queryable(
...
+        // Verify keys are plain (not hashed) — v2 changesets use plain keys like v1
         for (_, entry) in &storage_changesets {
-            assert!(entry.key.is_hashed(), "v2: storage changeset keys should be tagged as hashed");
+            let key = entry.key;
+            assert_ne!(
+                key,
+                keccak256(key),
+                "v2: storage changeset key should be plain (not its own keccak256)"
+            );
         }
```
klkvr: should this check be removed entirely?

crates/stages/stages/src/stages/execution/slot_preimages.rs
```diff
@@ -0,0 +1,223 @@
...
+    for account in state.bundle.state().values() {
+        for &slot_key in account.storage.keys() {
+            let plain = B256::from(slot_key.to_be_bytes());
+            let hashed = keccak256(plain);
+            if seen_hashes.insert(hashed) {
+                preimage_entries.push((hashed, plain));
+            }
+        }
+    }
```
klkvr, line 142: i'm wondering if there are some edge cases when slots are present in reverts but not in the state — e.g if slot value was changed in-between but then returned back to original value

crates/stages/stages/src/stages/execution/slot_preimages.rs
```diff
@@ -0,0 +1,223 @@
...
+            if !revert.wipe_storage {
+                continue;
+            }
+
+            if !already_wiped.insert(*address) {
+                // Second (or subsequent) destruction within the same batch: skip the DB walk.
+                // All slots from re-creation are already in `revert.storage` as explicit
+                // changes from execution.
+                continue;
+            }
```
klkvr: just confirming that this covers the case when account is destroyed at block N, re-created at N+1, writes some new storage, and then is destroyed again at N+2?

would `BundleState` include reverts for all new slots created in N+1?

klkvr (approved):

PR #22381, opened 2026-02-19
feat(trie): ArenaParallelSparseTrie

mediocregopher:
Adds `ArenaParallelSparseTrie`, an alternative `SparseTrie` implementation backed by a thunderdome arena allocator with subtrie-level parallelism.

The upper trie (first 2 nibble levels) lives in a shared arena, while each subtrie at depth 2 gets its own `ArenaSparseSubtrie` with an independent arena. This allows parallel hashing, leaf updates, and pruning across subtries using rayon, controlled by configurable `ArenaParallelismThresholds`.

Key components:
- **`ArenaSparseNode`** — branch, leaf, subtrie, and placeholder node types stored in arenas
- **`ArenaCursor`** — stack-based cursor for depth-first traversal with dirty-state propagation, supporting `seek`, `next`, `push`/`pop`, and `drain` operations
- **`ArenaSparseSubtrie`** — independent subtrie with its own arena, buffers, and leaf counters

Operations implemented: `set_root`, `reveal_nodes`, `update_leaves` (with proof-request loop), `root` (parallel subtrie hashing), `prune` (parallel retained-leaf filtering), `get_leaf_value`, `find_leaf`, `wipe`, `take_updates`.

Includes a proptest harness that validates root hashes and trie updates against `StorageRoot` across random initial states, changesets with mixed inserts/updates/deletions, and prune-then-update sequences.

## CLI arg

A new CLI argument `--engine.enable-arena-sparse-trie` is added which enables the APST during payload validation. ConfigurableSparseTrie has been resurrected to support this.


mediocregopher:
derek bench

crates/trie/sparse/src/arena/cursor.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /// Returns the entry below the top of the stack (the parent of the head), or `None`.
+    pub(super) fn parent(&self) -> Option<&ArenaCursorStackEntry> {
+        let len = self.stack.len();
+        (len >= 2).then(|| &self.stack[len - 2])
+    }
+
+    /// Returns the number of entries on the stack.
+    pub(super) fn len(&self) -> usize {
```
pepyakin: Nit: reading the callers it's kind of confusing to see checking the "length" of a cursor. Maybe depth?

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3039 @@
...
+/// depth or deeper belong to lower subtries.
+const UPPER_TRIE_MAX_DEPTH: usize = 2;
+
+/// Returns the index into a densely-packed children array for a given nibble, based on the
+/// state mask. Returns `None` if the nibble's bit is not set.
+const fn child_dense_index(state_mask: TrieMask, nibble: u8) -> Option<usize> {
+    if !state_mask.is_bit_set(nibble) {
+        return None;
+    }
+    Some((state_mask.get() & ((1u16 << nibble) - 1)).count_ones() as usize)
```
pepyakin: Nit

```rust
/// Counts the number of occupied child slots below `nibble` in the dense children array.
fn dense_index_of(state_mask: TrieMask, nibble: u8) -> usize {
    (state_mask.get() & ((1u16 << nibble) - 1)).count_ones() as usize
}
```

then reuse it here 
```suggestion
    Some(dense_index_of(state_mask, nibble))
```

and down `upsert_leaf`'s `SeekResult::NoChild { .. }` arm.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3049 @@
...
+            .iter()
+            .filter_map(|(_, node)| match node {
+                ArenaSparseNode::Subtrie(s) => Some(s.num_leaves as usize),
+                _ => None,
+            })
+            .sum()
+    }
+
+    fn memory_size(&self) -> usize {
+        // Approximate: count arena entries across all subtries.
```
pepyakin: nah brother this aint going to cut it

that being said, we are in a better position here: arenas are the biggest contributors of the resident memory amount and there we can multiply size_of<Node> by the capacity of the arena.

we should also account for the cleared subtries and the buffers, etc.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3049 @@
...
+        let rlp_node = self.arena[idx].state_ref().and_then(|s| s.cached_rlp_node()).cloned();
+
+        self.arena.remove(idx);
+
+        // Blind the parent's child slot with the pruned node's cached RLP. At the pruning
+        // boundary, the parent is a retained branch whose child (this node) had no retained
+        // descendants. Deeper removals have parents that will also be removed, but blinding
+        // them is harmless. We must blind regardless of whether the RLP is a hash or inline,
+        // otherwise the parent retains a `Revealed(idx)` pointing to the removed arena entry.
+        if let Some(rlp_node) = rlp_node {
```
pepyakin: I think this is fine when the `root()` was called prior calling `prune(..)`, but if not this would leave dangling pointers.

pepyakin (commented):
First pass.

shekhirin:
derek bench samply feature-args="--engine.enable-arena-sparse-trie"

crates/trie/sparse/src/traits.rs
```diff
@@ -406,3 +406,158 @@ pub enum LeafLookup {
...
+    }
+
+    macro_rules! delegate {
+    ($self:ident, $method:ident $(, $arg:expr)*) => {
+        match $self {
+            Self::Arena(inner) => inner.$method($($arg),*),
+            Self::HashMap(inner) => inner.$method($($arg),*),
+        }
+    };
+}
```
DaniPopes: formatting is off here

Cargo.toml
```diff
@@ -544,6 +544,7 @@ strum = { version = "0.27", default-features = false }
 strum_macros = "0.27"
 syn = "2.0"
 thiserror = { version = "2.0.0", default-features = false }
+thunderdome = "0.6"
```
DaniPopes, line 549: should probably be using [slotmap](https://docs.rs/slotmap/latest/slotmap/).

crates/trie/sparse/src/arena/nodes.rs
```diff
@@ -0,0 +1,280 @@
...
+                        rlp_node.as_hash().expect("blinded child must be a hash")
+                    }
+                    ArenaSparseNodeBranchChild::Revealed(child_idx) => {
+                        arena[*child_idx].cached_hash()
+                    }
+                };
+                hashes.push(hash);
+            }
+        }
+        *compact = BranchNodeCompact::new(
```
DaniPopes: should reuse memory here for the hashes vec?

crates/trie/sparse/src/arena/nodes.rs
```diff
@@ -0,0 +1,280 @@
...
+            TrieNodeV2::Leaf(leaf) => Self::Leaf {
+                state: ArenaSparseNodeState::Revealed,
+                key: leaf.key,
+                value: leaf.value,
+            },
+            TrieNodeV2::Branch(branch) => {
+                let mut children = SmallVec::with_capacity(branch.state_mask.count_bits() as usize);
+                for (stack_ptr, _nibble) in branch.state_mask.iter().enumerate() {
+                    children
+                        .push(ArenaSparseNodeBranchChild::Blinded(branch.stack[stack_ptr].clone()));
```
DaniPopes: this is just `branch.stack[..branch.state_mask.len()].iter().map().collect()`, `_nibble` is discarded and `stack_ptr` is `0..n`

crates/trie/sparse/src/arena/nodes.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+impl From<ProofTrieNodeV2> for ArenaSparseNode {
+    /// Converts a [`ProofTrieNodeV2`] into an [`ArenaSparseNode`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the node is an `Extension`, which should have been merged into a branch
+    /// by [`TrieNodeV2`].
```
DaniPopes, line 259: From impls recommend against panicking. Possibly move to a separate method

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3049 @@
...
+            let parent_idx = self.buffers.cursor.parent().expect("pruned child has parent").index;
+            let child_nibble = nibble.expect("non-root child");
+            let parent_branch = self.arena[parent_idx].branch_mut();
+            let dense_idx = child_dense_index(parent_branch.state_mask, child_nibble)
+                .expect("child nibble not found in parent state_mask");
+            parent_branch.children[dense_idx] = ArenaSparseNodeBranchChild::Blinded(rlp_node);
```
DaniPopes: I would recommend encapsulating this to a separate BranchChildIdx type, which can also make it be a u8, and move all these free fns to that type as constructor / getters etc. and also have a wrapper iter type for the mask.iter().enumerate() to return this new type

crates/trie/sparse/src/arena/mod.rs
pepyakin: I was wrong since the branch nodes can retain even more memory due to children. 

Not entirely sure what to do here though.

1. if we correct for this and iterate over nodes then there is a noticeable slowdown.
1. we can do O(1) estimation. The average node retained size is around 500B. The error bounds are like ±8-10%.
1. we can make metrics opt-in enabled by a flag. Logistically, this would be better made in a follow-up PR>

mediocregopher: imo for MVP we can just do size_of * capacity and refine from there in followups

crates/trie/sparse/src/arena/nodes.rs
mediocregopher: The problem is that TrieUpdates get taken out of the sparse trie and held in the chain-state until persistence, so closing the loop to re-use them becomes a larger architectural change

DaniPopes: this should then return BranchNodeCompact instead of taking as input

crates/trie/sparse/src/arena/mod.rs
pepyakin: agreed!

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3049 @@
...
+        }
+        self.upper_arena.clear();
+        self.root = self.upper_arena.insert(ArenaSparseNode::EmptyRoot);
+        if let Some(updates) = self.updates.as_mut() {
+            updates.clear()
+        }
+        self.buffers.clear();
+    }
+
+    fn shrink_nodes_to(&mut self, _size: usize) {
```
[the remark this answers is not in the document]
mediocregopher: We can address this if/when https://github.com/orlp/slotmap/pull/148 is merged, or if we implement our own arena

deny.toml
```diff
@@ -88,4 +88,5 @@ allow-git = [
     "https://github.com/alloy-rs/evm",
     "https://github.com/alloy-rs/hardforks",
     "https://github.com/paradigmxyz/jsonrpsee",
+    "https://github.com/DaniPopes/slotmap.git",
```
mediocregopher, line 91: Highlighting this in case it's a problem

crates/trie/sparse/tests/suite/set_root.rs
```diff
@@ -131,86 +131,6 @@ pub(super) fn test_set_root_does_not_retain_updates_when_not_requested<T: Sparse
     );
 }
 
-/// Branch masks influence `removed_nodes` detection.
-///
-/// When proof nodes carry `BranchNodeMasks`, `reveal_nodes` registers them in the
-/// trie's internal `branch_node_masks`. After mutations that remove leaves (causing
-/// branch nodes to disappear), `take_updates().removed_nodes` should correctly
-/// report nodes whose masks were previously registered but are now gone.
-pub(super) fn test_set_root_with_branch_masks<T: SparseTrie + Default>() {
```
mediocregopher, line 140: This test didn't actually test anything related to set_root specifically, and what it did test was redundant with other tests

crates/trie/sparse/tests/suite/take_updates.rs
```diff
@@ -76,29 +76,49 @@ pub(super) fn test_take_updates_resets_after_take<T: SparseTrie + Default>() {
 
 /// `take_updates` contains both updated and removed nodes, mutually exclusive.
 ///
-/// After mutations that cause branch node changes and deletions, `take_updates` should
-/// contain both updated and removed nodes. The two sets must be mutually exclusive.
+/// Uses a 3-level branching structure so that intermediate branches are "real" DB nodes
```
mediocregopher, line 79: The previous version of this test wasn't well formed, the db data it generated didn't contain any db nodes. This new version fixes this.

Cargo.toml
DaniPopes: unused

crates/trie/sparse/src/arena/nodes.rs
```diff
@@ -0,0 +1,287 @@
...
+use alloy_primitives::{keccak256, B256};
+use alloy_trie::{BranchNodeCompact, TrieMask};
+use reth_trie_common::{BranchNodeMasks, Nibbles, ProofTrieNodeV2, RlpNode, TrieNodeV2};
+use slotmap::{DefaultKey, SlotMap};
+use smallvec::SmallVec;
+
+/// Alias for the slotmap key type used as node references throughout the arena trie.
+type Index = DefaultKey;
+/// Alias for the slotmap used as the node arena throughout the arena trie.
+type NodeArena = SlotMap<Index, ArenaSparseNode>;
```
pepyakin, line 12: Nit:

1. why is this defined multiple types instead of being reused?
2. slotmap::new_key_type! > type Index

crates/trie/sparse/src/arena/cursor.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        parent_branch.children[child_idx] = ArenaSparseNodeBranchChild::Revealed(new_idx);
+    }
+
+    /// Advances the DFS traversal to the next actionable node.
+    ///
+    /// If a previous call returned [`NextResult::NonBranch`] or [`NextResult::Branch`],
+    /// the head entry is automatically popped (with dirty-state propagation) before
+    /// descending further. This means callers never need to call [`Self::pop`] after
+    /// `next` — it is handled internally on the subsequent call.
```
pepyakin, line 244: This works only when you exhaust the cursor (next-ing until it returns Done) and even then you will get a debug_assert for that. So not sure if even worth mentioning.

crates/trie/sparse/src/arena/cursor.rs
```diff
@@ -0,0 +1,387 @@
...
+            );
+        }
+
+        if let Some(parent) = self.stack.last() {
+            let child_is_dirty = arena.get(entry.index).is_some_and(|node| match node {
+                ArenaSparseNode::Branch(b) => matches!(b.state, ArenaSparseNodeState::Dirty),
+                ArenaSparseNode::Leaf { state, .. } => matches!(state, ArenaSparseNodeState::Dirty),
+                ArenaSparseNode::Subtrie(s) => {
+                    let root = &s.arena[s.root];
+                    matches!(root, ArenaSparseNode::EmptyRoot) ||
```
pepyakin, line 147: Nit: this makes sense but non-obvious. Perhaps a doc?

crates/trie/sparse/src/arena/cursor.rs
```diff
@@ -0,0 +1,387 @@
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the stack is empty.
+    pub(super) const fn depth(&self) -> usize {
+        self.stack.len() - 1
+    }
+
+    /// Returns `true` if the stack is empty.
+    pub(super) const fn is_empty(&self) -> bool {
```
pepyakin, line 97: is root?

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+                )
+            }
+            SeekResult::Diverged => {
+                let head_path = head.path;
+                let full_path_from_head = full_path.slice(head_path.len()..);
+
+                let split_dirtied_existing =
+                    Self::split_and_insert_leaf(arena, cursor, root, full_path_from_head, value);
+
+                let result = if cursor.depth() >= 1 {
```
pepyakin, line 1365: Nit: cursor.is_empty() ⇒ cursor.is_root() ? A comment would be good.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+                    key: leaf_key,
+                    value: value.to_vec(),
+                });
+
+                let branch = arena[head_idx].branch_mut();
+                branch.state_mask.set_bit(child_nibble);
+                branch
+                    .children
+                    .insert(insert_pos.get(), ArenaSparseNodeBranchChild::Revealed(new_leaf));
+                branch.state = ArenaSparseNodeState::Dirty;
```
pepyakin, line 1397: Nit: this looks as if it could live on `ArenaSparseNodeBranch::set_child(nibble, child_idx)`.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+                let parent_branch = arena[parent_idx].branch_ref();
+                let child_idx = BranchChildIdx::new(parent_branch.state_mask, child_nibble)
+                    .expect("leaf nibble not found in parent state_mask");
+
+                // Remove the leaf from the arena and from the parent's children.
+                arena.remove(head_idx);
+                let parent_branch = arena[parent_idx].branch_mut();
+                parent_branch.children.remove(child_idx.get());
+                parent_branch.unset_child_bit(child_nibble);
+                parent_branch.state = ArenaSparseNodeState::Dirty;
```
pepyakin, line 1526: Nit: Similarly, this could be a method `ArenaSparseNodeBranch::remove_child(nibble)`.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+                trie_updates.removed_nodes.insert(logical_path);
+            }
+        }
+
+        // Build the prefix: branch's short_key + remaining child's nibble.
+        let mut prefix = branch_short_key;
+        prefix.push_unchecked(remaining_nibble);
+
+        let ArenaSparseNodeBranchChild::Revealed(child_idx) = branch.children[0] else {
+            unreachable!()
```
pepyakin, line 1665: Nit: technically it's a precondition.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+
+        let parent_entry = cursor.parent()?;
+        let parent_branch = arena[parent_entry.index].branch_ref();
+        if parent_branch.state_mask.count_bits() != 2 {
+            return None;
+        }
+
+        let child_idx = BranchChildIdx::new(parent_branch.state_mask, child_nibble)
+            .expect("child nibble not in parent state_mask");
+        let sibling_dense_idx = 1 - child_idx.get();
```
pepyakin, line 1587: Nit: I really appreciated the comment last time I saw this trick. Perhaps copy it here as well?

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+    #[instrument(level = "trace", target = TRACE_TARGET, skip_all)]
+    fn wipe(&mut self) {
+        trace!(target: TRACE_TARGET, "Wiping arena trie");
+        self.clear();
+        self.updates = self.updates.is_some().then(SparseTrieUpdates::wiped);
+    }
+
+    #[instrument(level = "trace", target = TRACE_TARGET, skip_all)]
+    fn clear(&mut self) {
+        for idx in self.all_subtries() {
```
pepyakin, line 2351: Nit: could be `mem::take(&mut self.upper_arena)`? 

`all_subtries` can be then removed since this is a single call-site.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -0,0 +1,3242 @@
...
+                let (_, ref mut subtrie, ref range) = taken[0];
+                pruned += subtrie.prune(&retained_leaves[range.clone()]);
+            } else {
+                use rayon::iter::{IntoParallelRefMutIterator, ParallelIterator};
+
+                let parallel_pruned: Vec<usize> = taken
+                    .par_iter_mut()
+                    .map(|(_, subtrie, range)| subtrie.prune(&retained_leaves[range.clone()]))
+                    .collect();
+                pruned += parallel_pruned.into_iter().sum::<usize>();
```
pepyakin, line 2545: Nit: I think the intermediate vector can be avoided by just using par_iter's sum directly?

Cargo.toml
```diff
@@ -539,11 +539,14 @@ serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 serde_with = { version = "3", default-features = false, features = ["macros"] }
 sha2 = { version = "0.10", default-features = false }
 shlex = "1.3"
+# https://github.com/orlp/slotmap/pull/148
+slotmap = { git = "https://github.com/DaniPopes/slotmap.git", branch = "dani/shrink-methods" }
```
pepyakin, line 543: I don't think it does what it claims it does.

Consider that it is a vector with holes and a freelist. When you remove an item, it's position is inserted into the freelist.

After removing some arbitrary items the freespace is going to be scattered all over. At the same time, the length of the underlying vector will be the same or close to the capacity.

Then, even when you `clear` the slotmap you don't really change the length of the underlying vector.

shrink_to then cannot shrink below the length of the vector.

Therefore, it's probably no-op in all cases.

Am I missing something?

cc @DaniPopes

pepyakin (approved):
LGTM. Some nits but none of them are blocking!

PR #22386, opened 2026-02-19
fix: catch panics of named tasks

klkvr:
right now if a task sent to a `WorkerThread` panics, the thread will die and `rx` side will be lost causing any new tasks with the same name to not be executed


shekhirin (approved):

shekhirin:
derek bench

PR #22387, opened 2026-02-19
feat(primitives): add Block::num_hash_slow

gakonst:
Add `num_hash_slow` to the `Block` trait — returns `BlockNumHash` by computing `header().number()` + `header().hash_slow()`.

Prompted by: matthias


mattsse (commented):
not really needed

PR #22389, opened 2026-02-19
fix: align static-file changeset checksum with MDBX semantics

joshieDo:
## Summary
- Aligns `reth db checksum static-file` behavior for change-based segments with MDBX by hashing block key bytes together with row value bytes.
- Uses changeset offset sidecar metadata to map each static-file row to its block number during checksum calculation.
- Improves the checksum loop flow so limit handling is clearer and static-file provider cache cleanup remains deterministic.
- Adds consistency guards that error when changeset offsets are missing or do not fully cover stored rows.

## Validation
- `cargo test -p reth-cli-commands checksum --no-run`

### DIFF
--- a/crates/cli/commands/src/db/checksum/mod.rs
+++ b/crates/cli/commands/src/db/checksum/mod.rs
@@ -115,136 +115,6 @@ fn checksum_hasher() -> impl Hasher {
     FixedState::with_seed(u64::from_be_bytes(*b"RETHRETH")).build_hasher()
 }
 
-/// Accumulates a checksum over key-value entries, tracking count and limit.
-struct Checksummer<H> {
-    hasher: H,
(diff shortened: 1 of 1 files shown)


joshieDo:
```
Ran:
v1 MDBX against ~/preimage/nodev1
v2 static-file against ~/preimage/nodev2 (joshie/slot-preimages)
Results:

mdbx AccountChangeSets: 0xda94866113e047a4
mdbx StorageChangeSets: 0xe74ae545a6eb0ba7
static-file account-change-sets: 0xda94866113e047a4 (11282073 entries)
static-file storage-change-sets: 0xe74ae545a6eb0ba7 (1514386 entries)
```

Rjected (approved):
lgtm

PR #22397, opened 2026-02-20
feat(rpc): add `reth_getBlockExecutionOutcome` endpoint

stevencartavia:
Adds a new RPC endpoint `reth_getBlockExecutionOutcome` that re-executes an arbitrary block against its parent state and returns the execution outcome (receipts, state changes, deployed contracts, and EIP-7685 requests).

## Changes

- **`reth-rpc-api`**: New method on `RethApi` trait with `BlockExecutionOutcomeResponse` and `AccountStateChanges` response types
- **`reth-rpc`**: Implementation that fetches the block, loads parent state, and re-executes via `evm_config().executor(db).execute(&block)`
- **`reth-rpc-builder`**: Wires `evm_config` and `BlockingTaskGuard` into `RethApi` construction

Concurrency is limited via `BlockingTaskGuard`. Genesis block (no parent state) returns an explicit error. Includes 5 unit tests.


crates/rpc/rpc-api/src/reth.rs
```diff
@@ -1,6 +1,7 @@
-use alloy_eips::BlockId;
-use alloy_primitives::{map::AddressMap, U256};
+use alloy_eips::{eip7685::Requests, BlockId};
+use alloy_primitives::{map::AddressMap, Address, Bytes, B256, U256};
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use std::collections::HashMap;
```
mattsse: lets use alloy primitive map types for this like AddressMap

crates/rpc/rpc-api/src/reth.rs
```diff
@@ -47,3 +56,37 @@ pub trait RethApi {
...
+/// Response type for `reth_getBlockExecutionOutcome`.
+///
+/// Contains the execution result of re-executing a block against its parent state.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockExecutionOutcomeResponse {
+    /// Gas used by the block.
+    pub gas_used: u64,
+    /// Blob gas used by the block.
+    pub blob_gas_used: u64,
```
mattsse: I'd like to use U64 here

crates/rpc/rpc-api/src/reth.rs
```diff
@@ -47,3 +56,37 @@ pub trait RethApi {
...
+    pub contracts: HashMap<B256, Bytes>,
+}
+
+/// State changes for a single account.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AccountStateChanges {
+    /// The account nonce after execution, if the account exists.
+    pub nonce: Option<u64>,
+    /// The account balance after execution, if the account exists.
```
mattsse: same here

crates/rpc/rpc-api/src/reth.rs
```diff
@@ -47,3 +56,37 @@ pub trait RethApi {
...
+    pub balance: Option<U256>,
+    /// The account code hash after execution, if the account exists.
+    pub code_hash: Option<B256>,
+    /// Storage changes: slot -> present value.
+    pub storage: HashMap<B256, U256>,
```
mattsse: this can be B256Map

crates/rpc/rpc/src/reth.rs
```diff
@@ -11,39 +11,100 @@ use reth_chain_state::{
...
+    /// [`BlockExecutionOutcomeResponse`].
+    fn convert_execution_output<R: Serialize>(
+        output: BlockExecutionOutput<R>,
+    ) -> Result<BlockExecutionOutcomeResponse, serde_json::Error> {
+        let receipts = output
+            .result
+            .receipts
+            .iter()
+            .map(serde_json::to_value)
+            .collect::<Result<Vec<_>, _>>()?;
```
mattsse: maybe we can even avoid doing this conversion entirely, because BlockExecutionOutput should already be serde?

if not, lets add

crates/rpc/rpc-api/src/reth.rs
```diff
@@ -47,3 +56,37 @@ pub trait RethApi {
         &self,
     ) -> jsonrpsee::core::SubscriptionResult;
 }
+
+/// Response type for `reth_getBlockExecutionOutcome`.
+///
+/// Contains the execution result of re-executing a block against its parent state.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockExecutionOutcomeResponse {
```
mattsse: I kinda like having a dedicated serde type for this but in practice this means that a caller likely needs do another conversion to get back the regular ExecutionOutcome, so maybe we can send this directly.

this type is more or less just

https://github.com/paradigmxyz/reth/blob/16b7dffe9119e79422590af950fa7df4b612c1e9/crates/evm/execution-types/src/execution_outcome.rs#L49-L49

so maybe lets just send this?

crates/rpc/rpc-api/src/reth.rs
```diff
@@ -16,6 +17,14 @@ pub trait RethApi {
         block_id: BlockId,
     ) -> RpcResult<AddressMap<U256>>;
 
+    /// Re-executes a block and returns the execution outcome including receipts, state changes,
+    /// and EIP-7685 requests.
+    #[method(name = "getBlockExecutionOutcome")]
+    async fn reth_get_block_execution_outcome(
+        &self,
+        block_id: BlockId,
+    ) -> RpcResult<Option<BlockExecutionOutcomeResponse>>;
```
mattsse: cool, lets also add another optional argument that is count: option U64

then we can return a batch of 

https://github.com/paradigmxyz/reth/blob/16b7dffe9119e79422590af950fa7df4b612c1e9/crates/evm/execution-types/src/execution_outcome.rs#L43-L49

see

https://github.com/paradigmxyz/reth/blob/16b7dffe9119e79422590af950fa7df4b612c1e9/crates/evm/execution-types/src/execution_outcome.rs#L135-L135

mattsse (changes requested):
very nice, some comments, I think we can make this even more powerful by changeing the response value

crates/rpc/rpc-api/src/reth.rs
stevencartavia: removed that type

crates/rpc/rpc-api/src/reth.rs
stevencartavia: removed that type

crates/rpc/rpc-api/src/reth.rs
stevencartavia: this one too

crates/rpc/rpc-api/src/reth.rs
stevencartavia: done

crates/rpc/rpc/src/reth.rs
stevencartavia: avoided

crates/rpc/rpc-api/src/reth.rs
stevencartavia: 🫡

mattsse (approved):

PR #22406, opened 2026-02-20
feat: implement bal specific rpc methods

Soubhik-10:


crates/ethereum/node/tests/e2e/debug.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    // Mine an empty block
+    node.advance_block().await?;
+
+    // Get BAL for block 1 (empty block)
+    let bal: BlockAccessList =
+        provider.client().request("debug_getBlockAccessList", (1u64,)).await?;
+
+    // Empty block should return empty BAL
+    assert!(bal.is_empty(), "Empty block should have empty BAL");
```
emmajam: This chain spec has `.cancun_activated()`, which enables EIP-4788 (beacon block root). The system call at the start of each block writes to the beacon roots contract, so the BAL may not actually be empty even for a block with no user transactions. This assertion could be flaky or wrong — verify whether the BAL builder captures pre-execution system calls and adjust accordingly.

crates/rpc/rpc-eth-api/src/core.rs
```diff
@@ -913,17 +921,68 @@ where
...
+            self.evm_env_at(block_hash.into()),
+            self.recovered_block(block_hash.into()),
+        )?;
+
+        let block = block
+            .ok_or_else(|| EthApiError::Internal(reth_errors::RethError::msg("Block Not Found")))?;
+
+        let json = self
+            .spawn_blocking_io(move |eth_api| {
+                let state = eth_api
```
emmajam: Missing post-execution changes. The loop applies pre-execution and per-tx execution but never calls post-execution (e.g. Shanghai withdrawals). Balance updates from withdrawals won't be captured in the BAL. Need to apply post-execution changes before `take_built_alloy_bal()`

crates/rpc/rpc-api/src/debug.rs
```diff
@@ -157,9 +156,10 @@ pub trait DebugApi<TxReq: RpcObject> {
         hash: B256,
     ) -> RpcResult<ExecutionWitness>;
 
-    /// Re-executes a block and returns the Block Access List (BAL) as defined in EIP-7928.
+    /// Re-executes a block and returns the RLP-encoded Block Access List (BAL) as defined in
+    /// EIP-7928.
     #[method(name = "getBlockAccessList")]
-    async fn debug_get_block_access_list(&self, block_id: BlockId) -> RpcResult<BlockAccessList>;
+    async fn debug_get_block_access_list(&self, block_id: BlockId) -> RpcResult<Bytes>;
```
emmajam, line 162: `debug_getBlockAccessList` returns `Bytes` (RLP-encoded), but the E2E test at `crates/ethereum/node/tests/e2e/debug.rs:96` deserializes the response as `BlockAccessList` (JSON struct). These are incompatible — the test will fail. Either return the structured `BlockAccessList` or update the test to decode RLP.

crates/rpc/rpc-eth-api/src/core.rs
```diff
@@ -406,14 +414,14 @@ pub trait EthApi<
 
     /// Returns the EIP-7928 block access list for a block by hash.
     #[method(name = "getBlockAccessListByBlockHash")]
-    async fn block_access_list_by_block_hash(&self, hash: B256) -> RpcResult<Option<Bytes>>;
+    async fn block_access_list_by_block_hash(&self, hash: B256) -> RpcResult<Option<Value>>;
```
emmajam, line 411: `eth_getBlockAccessListByBlockHash` and `eth_getBlockAccessListByBlockNumber` return `Option<Value>` — returning `Option<BlockAccessList>` would preserve type safety and avoid the new `serde_json` dependency added to this crate.

crates/rpc/rpc-eth-api/src/core.rs
```diff
@@ -913,17 +921,68 @@ where
...
-        Err(internal_rpc_err("unimplemented"))
+    async fn block_access_list_by_block_hash(&self, block_hash: B256) -> RpcResult<Option<Value>> {
+        trace!(target: "rpc::eth", ?block_hash, "Serving eth_getBlockAccessListByBlockHash");
+        let ((evm_env, _), block) = futures::try_join!(
+            self.evm_env_at(block_hash.into()),
+            self.recovered_block(block_hash.into()),
+        )?;
+
+        let block = block
+            .ok_or_else(|| EthApiError::Internal(reth_errors::RethError::msg("Block Not Found")))?;
```
emmajam: Nit: use `EthApiError::HeaderNotFound(block_hash.into())` to return the correct JSON-RPC error code, consistent with the rest of the codebase.

crates/rpc/rpc-eth-api/Cargo.toml
```diff
@@ -58,6 +64,7 @@ tokio.workspace = true
 auto_impl.workspace = true
 dyn-clone.workspace = true
 tracing.workspace = true
+serde_json.workspace = true
```
emmajam, line 61: `serde_json` was added only to support the `Value` return type. If switching to `BlockAccessList` (comment #3), this dependency can be dropped.

emmajam (commented):
ty !!

crates/rpc/rpc-eth-api/src/helpers/block_access_list.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Retrieves the block access list for a block identified by its hash.
+    fn get_block_access_list(
+        &self,
+        block_hash: B256,
+    ) -> impl Future<Output = Result<Option<BlockAccessList>, Self::Error>> + Send {
+        async move {
+            let ((evm_env, _), block) = futures::try_join!(
+                self.evm_env_at(block_hash.into()),
+                self.recovered_block(block_hash.into()),
+            )?;
```
mattsse: we can simplify this like:

https://github.com/paradigmxyz/reth/pull/22747

crates/rpc/rpc-eth-api/src/helpers/block_access_list.rs
```diff
@@ -0,0 +1,60 @@
...
+
+                eth_api.apply_pre_execution_changes(&block, &mut db)?;
+
+                for tx in block.transactions_recovered() {
+                    let tx_env = eth_api.evm_config().tx_env(tx);
+                    let res = eth_api.transact(&mut db, evm_env.clone(), tx_env)?;
+                    db.commit(res.state);
+                }
+
+                eth_api.apply_post_execution_changes(&block, &mut db)?;
```
mattsse: I think we can do it like this here:

https://github.com/paradigmxyz/reth/blob/ed379f432bd1af8af53a8771b6a19636fc7534c8/crates/rpc/rpc-eth-api/src/helpers/call.rs#L747-L759

crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -426,4 +426,19 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
...
+        db: &mut StateCacheDb,
+    ) -> Result<(), Self::Error> {
+        self.evm_config()
+            .executor_for_block(db, block.sealed_block())
+            .map_err(RethError::other)
+            .map_err(Self::Error::from_eth_err)?
+            .apply_post_execution_changes()
+            .map_err(Self::Error::from_eth_err)?;
+        Ok(())
+    }
```
mattsse: then we dont need this fn really

mattsse (changes requested):
nice, smol suggestions

emmajam (approved):

mattsse (approved):

emmajam (approved):

mattsse (approved):

PR #22407, opened 2026-02-20
perf(engine): release cache lock before waiting for state root validation

yongkangc:
## Summary

Moves `valid_block_rx.recv()` out from under the execution cache write lock in `save_cache()`. The original code held the write lock across this blocking recv, stalling all concurrent readers (`get_cache_for`) for ~100ms+ during state root validation. This prevented the next block's prewarming from accessing the cache.

## Changes

- `save_cache()` now waits for block validation **without** any lock held
- On success, `insert_state()` mutates the caches while the usage guard is still held (`is_available() == false`), preventing concurrent readers
- `split()` consumes the `SavedCache` (releasing the usage guard) and publishes the new cache under a brief write lock
- Updated `update_with_guard` doc comment to accurately describe the safety contract

## Safety

The correctness relies on the usage guard (`Arc<()>` strong count):
1. While the prewarming task holds its `SavedCache` clone, `is_available()` returns false → `get_cache_for()` returns `None`
2. `insert_state()` operates on the shared `FixedCache` (concurrent-safe) with no other readers possible
3. `split()` is called **inside** `update_with_guard` (write lock held), so the guard release and new cache publication are atomic

## Expected Impact

Eliminates ~100ms+ of write lock contention per block in `save_cache()`, unblocking the next block's prewarming from starting cache reads earlier.

## Testing

```
cargo nextest run -p reth-engine-tree --lib  # 96/96 passed
cargo clippy -p reth-engine-tree --lib --all-features  # zero warnings
```

Split from #22149 — this PR contains only the cache lock optimization.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -1019,16 +1019,10 @@ impl PayloadExecutionCache {
     /// Updates the cache with a closure that has exclusive access to the guard.
     /// This ensures that all cache operations happen atomically.
     ///
-    /// ## CRITICAL SAFETY REQUIREMENT
-    ///
-    /// **Before calling this method, you MUST ensure there are no other active cache users.**
-    /// This includes:
-    /// - No running [`PrewarmCacheTask`] instances that could write to the cache
-    /// - No concurrent transactions that might access the cached state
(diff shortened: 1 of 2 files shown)


yongkangc:
Superseded by #22413 from branch yk/cache-lock.

PR #22425, opened 2026-02-20
feat(rpc): move reth_newPayload from EngineApi to RethApi

gakonst:
## Summary
Moves `reth_newPayload` from the auth-only `EngineApi` to the public `RethApi`, making it available by default when the `reth` namespace is enabled.

## Changes
- Add `reth_newPayload` method to the `RethApi` trait (`crates/rpc/rpc-api/src/reth.rs`)
- Propagate `beacon_engine_handle` through `RpcRegistryInner` similarly to `engine_events` (`crates/rpc/rpc-builder/src/lib.rs`)
- Implement `reth_newPayload` on `RethApi` using the engine handle (`crates/rpc/rpc/src/reth.rs`)
- Remove `RethEngineApiServer` trait and its `EngineApi` implementation (`crates/rpc/rpc-engine-api/src/engine_api.rs`)
- Remove `reth_newPayload` merge from `EngineApi::into_rpc_module`

## Testing
- Full workspace compiles: `cargo check --workspace`
- Clippy passes on affected crates
- Formatting verified with `cargo +nightly fmt --all --check`

Prompted by: arsenii


mattsse (approved):
pending @klkvr

shekhirin (approved):

PR #22426, opened 2026-02-20
chore(txpool): use `to_consensus` helper instead of reaching into inner field

figtracer:


mattsse (approved):
this is equivalent, yes

PR #22436, opened 2026-02-20
fix(transaction-pool): clean up closed blob sidecar listeners on add

Bashmunta:
Previously, `add_blob_sidecar_listener` did not clean up closed channel senders before adding new listeners, unlike `add_pending_listener` and `add_new_transaction_listener`. This caused closed blob listeners to accumulate in memory indefinitely if no new blob transactions were added after a receiver was dropped.

Added `listeners.retain(|l| !l.sender.is_closed())` before pushing new blob sidecar listeners, matching the existing pattern for other listener types.


mattsse (commented):
this isnt needed, this happens with the next event

PR #22434, opened 2026-02-20
ci(bench): switch to @decofe bot and new secret names

shekhirin:
Changed benchmark trigger from `derek bench` to `@decofe bench` and updated GitHub token secrets to use @decofe's PAT instead of generic GITHUB_TOKEN for reactions and comments. Renamed BENCH_BOT_PAT to DEREK_TOKEN for consistency. All 9 github-script steps that interact with comments/reactions now use DEREK_PAT with appropriate permissions (Issues: read+write, PRs: read, Actions: read).

Note: Requires new secrets to be added:
- `DEREK_PAT`: Fine-grained PAT scoped to paradigmxyz/reth with Issues (read+write), Pull requests (read), Actions (read), Metadata (read)
- `DEREK_TOKEN`: Fine-grained PAT scoped to decofe/reth-bench-charts with Contents (read+write)

PR #22439, opened 2026-02-20
ci(bench): add median lines to benchmark charts

shekhirin:
Add horizontal dashed median lines to the latency and throughput benchmark charts for easier visual comparison.

Each series (baseline and feature) now displays its median as a dashed line with a label in the legend. This makes it easier to spot performance differences at a glance by comparing the median lines rather than trying to visually parse the noisy per-block data.

The change also ensures the legend is always displayed on these charts, even when there's no baseline to compare against.

Closes #22434


shekhirin:
@decofe bench samply

gakonst (approved):

PR #22441, opened 2026-02-20
refactor(net): add methods to PropagatedTransactions instead of exposing .0

figtracer:
## Summary

Add proper methods to `PropagatedTransactions` and make the inner `HashMap` field private, replacing ~15 direct `.0` tuple field accesses across the codebase.

## Details

`PropagatedTransactions` is a newtype around `HashMap<TxHash, Vec<PropagateKind>>`, but callers were drilling into `.0` directly everywhere — defeating the purpose of the newtype pattern.

### New methods:
- `record(&mut self, hash: TxHash, kind: PropagateKind)` — replaces `.0.entry(hash).or_default().push(kind)`
- `len(&self) -> usize` — replaces `.0.len()`
- `is_empty(&self) -> bool` — replaces `.0.is_empty()`
- `get(&self, hash: &TxHash) -> Option<&Vec<PropagateKind>>` — replaces `.0.get(hash)`
- `IntoIterator` impl — replaces `.0.into_iter()`

### Files changed:
- `crates/transaction-pool/src/traits.rs` — added methods, made field private
- `crates/transaction-pool/src/pool/mod.rs` — updated `on_propagated`
- `crates/net/network/src/transactions/mod.rs` — updated all propagation call sites and tests

### DIFF
--- a/crates/transaction-pool/src/traits.rs
+++ b/crates/transaction-pool/src/traits.rs
@@ -821,7 +821,38 @@ impl<T: PoolTransaction> IntoIterator for AllPoolTransactions<T> {
 
 /// Represents transactions that were propagated over the network.
 #[derive(Debug, Clone, Eq, PartialEq, Default)]
-pub struct PropagatedTransactions(pub HashMap<TxHash, Vec<PropagateKind>>);
+pub struct PropagatedTransactions(HashMap<TxHash, Vec<PropagateKind>>);
+
+impl PropagatedTransactions {
+    /// Records a propagation of a transaction to a peer.
+    pub fn record(&mut self, hash: TxHash, kind: PropagateKind) {
+        self.0.entry(hash).or_default().push(kind);
+    }
+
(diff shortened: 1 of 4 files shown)


crates/transaction-pool/src/traits.rs
```diff
@@ -821,7 +821,38 @@ impl<T: PoolTransaction> IntoIterator for AllPoolTransactions<T> {
 
 /// Represents transactions that were propagated over the network.
 #[derive(Debug, Clone, Eq, PartialEq, Default)]
-pub struct PropagatedTransactions(pub HashMap<TxHash, Vec<PropagateKind>>);
+pub struct PropagatedTransactions(HashMap<TxHash, Vec<PropagateKind>>);
```
mattsse:
```suggestion
pub struct PropagatedTransactions(pub HashMap<TxHash, Vec<PropagateKind>>);
```

mattsse (approved):

PR #22448, opened 2026-02-21
perf(trie): inline single-target storage proofs in account worker

gakonst:
## Summary
Inline storage proofs with `targets.len() <= 1` into the account proof worker instead of dispatching to the storage worker pool.

## Motivation
Single-target storage proofs (and root-only requests) incur channel allocation, cross-thread dispatch, and synchronization overhead that dominates the actual trie work. By computing these inline in the account worker, we eliminate this overhead for the most common case.

Related: [RETH-418](https://linear.app/tempoxyz/issue/RETH-418)

## Changes
- `dispatch_v2_storage_proofs` (`proof_task.rs`): separates single-target proofs into an `inline_single` map instead of dispatching them to storage workers
- `AsyncAccountDeferredValueEncoder` (`value_encoder.rs`): new `InlineSingle` variant that computes the storage proof synchronously using the account worker's existing `StorageProofCalculator`
- `AsyncAccountValueEncoder::deferred_encoder`: checks `inline_single` map before falling through to `Sync`/`FromCache` paths
- `AsyncAccountValueEncoder::finalize`: handles remaining unconsumed inline proofs
- `ProofTaskTrieMetrics` (`proof_task_metrics.rs`): new `deferred_encoder_inline_single` histogram

## Testing
```
cargo test -p reth-trie-parallel  # 2 passed
cargo +nightly clippy -p reth-trie-parallel --all-features  # clean
```

Needs reth-bench A/B comparison (baseline=main, feature=this branch) to measure throughput impact.

Prompted by: yk

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -1062,12 +1062,11 @@ where
 
         trace!(target: "trie::proof_task", "Processing V2 account multiproof");
 
-        let dispatched =
+        let storage_proof_receivers =
             dispatch_v2_storage_proofs(&self.storage_work_tx, &account_targets, storage_targets)?;
 
         let mut value_encoder = AsyncAccountValueEncoder::new(
-            dispatched.receivers,
-            dispatched.inline_single,
+            storage_proof_receivers,
             self.cached_storage_roots.clone(),
             v2_storage_calculator,
         );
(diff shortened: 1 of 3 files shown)


yongkangc:
derek bench

mattsse:
Closing this draft because the benchmark did not demonstrate a performance improvement: mean latency regressed by 1.53%, P90 latency by 3.17%, and throughput by 1.01%. The experiment has also been inactive since February, so keeping the PR open no longer appears useful. We can revisit the approach if new profiling data suggests a viable direction.

PR #22449, opened 2026-02-21
perf: avoid repeated sort in PendingPool::remove_to_limit

Serdnad:
Sort highest-nonce transactions once and reuse the sender ordering across eviction rounds in `PendingPool::remove_to_limit`, replacing the per-round `collect + sort` with a per-sender HashMap lookup


mattsse:
Thanks for looking into this. The optimization target makes sense, but this version changes eviction ordering after the first round: the initial sender order is reused even though each sender's current highest-nonce transaction can have a different priority after removals. That can evict a better remaining transaction and keep a worse one compared with the current per-round sort.

Closing this PR for now; I'll look at a lower-risk way to reduce the current implementation's overhead while preserving the existing semantics.

PR #22456, opened 2026-02-21
fix(net): correct high-priority request queue insertion order

andrewshab3:
Fix FIFO violation for high-priority download requests when the queue has no normal-priority items. `position(is_normal)` returns None and unwrap_or(0) inserts at the front instead of the back, reversing the order of high-priority requests.


mattsse (commented):
no

PR #22465, opened 2026-02-21
fix(txpool): update AllTransactions size metrics on remove paths

HuberyJulianay:
`AllTransactions::update_size_metrics()` was only called from `insert_tx()`, so the `all_transactions_by_hash` and `all_transactions_by_id` gauges would go stale after every remove until the next insert. The other sub-pool gauges (pending/basefee/queued/blob) were already being refreshed correctly through `TxPool::update_size_metrics()` — this just adds the missing call for `all_transactions` in the same place.


mattsse (commented):
can keep as is

PR #22466, opened 2026-02-21
perf(storage): add bloom filter to skip MDBX seeks for empty slots

0xMars42:
## Summary

Opt-in bloom filter that skips MDBX seeks for storage slots that were never written. ~30-40% of `SLOAD` ops hit empty slots ([Nethermind data](https://github.com/NethermindEth/nethermind/issues/9854)) — each one currently walks the B-tree for nothing. The bloom answers "definitely not present" in <400ns and avoids the seek entirely.

Closes #21184

```
SLOAD(address, slot)
  → bloom.might_contain(address, slot)?
    false → Ok(None)                       ← <400ns, no MDBX touch
    true  → PlainStorageState seek          ← same path as before
```

## Design

Lock-free `AtomicU64` bit array, Kirsch-Mitzenmacher double hashing from a single `keccak256(address || slot)`, K=3 probes. Insert-only — deleted slots just become false positives (harmless, falls back to MDBX). Persisted to `<datadir>/storage_bloom.bin` with block number so restarts only need an incremental catchup via `StorageChangeSets` instead of a full table scan.

Default size is 1024 MB (~700M capacity at <1% FP, mainnet has ~500M non-zero slots today).

## Changes

**New crate `reth-storage-bloom`** — the filter itself, `no_std` compatible, with persistence (`save_to_file`/`load_from_file`, atomic rename, format versioning). Includes criterion benchmarks and 11 unit tests.

**Provider integration** — `BloomStateProvider<S>` wraps any `StateProvider` and intercepts `storage()` with a bloom check (same pattern as `InstrumentedStateProvider`). `ProviderFactory` wraps `latest()` / `history_by_block_*()` providers. `DatabaseProvider` inserts into the bloom on every non-zero storage write so new slots are immediately visible.

**Startup** — first run does a full `PlainStorageState` walk, subsequent runs load from disk and catch up from `StorageChangeSets`. Graceful fallback if the file is missing or corrupted.

## Benchmarks

256 MB filter, K=3, x86-64:

| Operation | Latency | |
|---|---|---|
| insert | ~359ns | keccak + 3 atomic ORs |
| true negative | ~391ns | early exit on first 0 bit |
| true positive | ~413ns | all 3 bits checked |
| mixed 35/65 | ~437ns | realistic empty/occupied ratio |

vs MDBX seek 1-10µs → 2.5-25x speedup on the empty-slot fraction.

### DIFF
--- a/crates/storage/bloom/benches/bloom_filter.rs
+++ b/crates/storage/bloom/benches/bloom_filter.rs
@@ -1,0 +1,157 @@
+//! Benchmarks for `StorageBloomFilter`.
+//!
+//! Measures insert and lookup throughput to validate that the bloom filter
+//! adds negligible overhead compared to MDBX seeks (~1-10us).
+
+use alloy_primitives::{address, Address, B256, U256};
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use reth_storage_bloom::StorageBloomFilter;
+
+/// Generate deterministic (address, slot) pairs for benchmarking.
+fn generate_pairs(n: u64) -> Vec<(Address, B256)> {
+    let addr = address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045");
+    (0..n).map(|i| (addr, B256::from(U256::from(i)))).collect()
+}
+
+/// Generate absent pairs (different address, so guaranteed not in bloom after inserting
+/// pairs from `generate_pairs`).
+fn generate_absent_pairs(n: u64) -> Vec<(Address, B256)> {
+    let addr = address!("1111111111111111111111111111111111111111");
+    (0..n).map(|i| (addr, B256::from(U256::from(i)))).collect()
+}
+
+fn bench_insert(c: &mut Criterion) {
+    let mut group = c.benchmark_group("bloom_insert");
+
+    for size_mb in [16, 128, 256] {
+        let bloom = StorageBloomFilter::with_size_mb(size_mb);
+        let pairs = generate_pairs(10_000);
+
+        group.bench_with_input(
+            BenchmarkId::new("insert", format!("{size_mb}MB")),
+            &pairs,
+            |b, pairs| {
+                b.iter(|| {
+                    for (addr, slot) in pairs {
+                        bloom.insert(addr, slot);
+                    }
+                });
+            },
+        );
+    }
+
+    group.finish();
+}
+
+fn bench_might_contain_true_negative(c: &mut Criterion) {
+    let mut group = c.benchmark_group("bloom_true_negative");
+
+    for size_mb in [16, 128, 256] {
+        let bloom = StorageBloomFilter::with_size_mb(size_mb);
+        let pairs = generate_pairs(100_000);
+        for (addr, slot) in &pairs {
+            bloom.insert(addr, slot);
+        }
+
+        let absent = generate_absent_pairs(10_000);
+
+        group.bench_with_input(
+            BenchmarkId::new("might_contain_miss", format!("{size_mb}MB")),
+            &absent,
+            |b, absent| {
+                b.iter(|| {
+                    for (addr, slot) in absent {
+                        std::hint::black_box(bloom.might_contain(addr, slot));
+                    }
+                });
+            },
+        );
+    }
+
+    group.finish();
+}
+
+fn bench_might_contain_true_positive(c: &mut Criterion) {
+    let mut group = c.benchmark_group("bloom_true_positive");
+
+    for size_mb in [16, 128, 256] {
+        let bloom = StorageBloomFilter::with_size_mb(size_mb);
+        let pairs = generate_pairs(100_000);
+        for (addr, slot) in &pairs {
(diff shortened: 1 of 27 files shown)


crates/storage/bloom/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    /// Returns the total number of bits set to 1.
+    pub fn count_set_bits(&self) -> u64 {
+        self.bits.iter().map(|w| w.load(Ordering::Relaxed).count_ones() as u64).sum()
+    }
+
+    /// Saves the bloom filter to a file.
+    ///
+    /// File format: `[magic: 8B][version: 1B][num_hashes: 1B][block_number: 8B][num_words: 8B][raw
+    /// bits...]`
```
DeshErBojhaa, line 143: nit: Can be in one line.

crates/storage/bloom/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        Ok(())
+    }
+
+    /// Loads a bloom filter from a file previously written by [`save_to_file`](Self::save_to_file).
+    ///
+    /// Returns `(bloom, block_number)` where `block_number` is the tip at which the filter
+    /// was saved. The caller should catch up from `block_number + 1` to the current tip
+    /// using storage changesets.
+    ///
+    /// Returns `Ok(None)` if the file does not exist or has an incompatible format/size.
```
DeshErBojhaa, line 181: Sorry for the dumb question, but why we are returning `Ok(None)` instead of someting like `ErrCorruptedFile`

crates/storage/bloom/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// Creates a new bloom filter with the given size in bytes and hash count.
+    ///
+    /// # Arguments
+    /// - `size_bytes`: Total memory for the bit array. Must be >= 8 (one u64).
+    /// - `num_hashes`: Number of probe positions per key (K). Typical: 3.
+    ///
+    /// # Panics
+    /// Panics if `size_bytes < 8`.
+    pub fn new(size_bytes: usize, num_hashes: u8) -> Self {
+        assert!(size_bytes >= 8, "bloom filter must be at least 8 bytes");
```
DeshErBojhaa, line 68: Maybe we should also consider assertion on `num_hashes`. Make sure it's not too big (even if fits into u8).

decofe:
cc @shekhirin

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/22274913452/job/64435376532)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `perf-storage-bloom-filter`

crates/storage/bloom/src/lib.rs
0xMars42: Good catch, will fix.

crates/storage/bloom/src/lib.rs
0xMars42: Good question. The bloom filter is a derived cache, not a source of truth, it's always rebuildable from `StorageChangeSets`. Both a missing file and a corrupted file lead to the same caller action: rebuild from scratch. Returning `Ok(None)` keeps the caller logic simple (one code path for "no usable filter") instead of forcing it to distinguish between "missing" and "corrupted" when the recovery is identical.

Good idea on surfacing it though. I'll add a `tracing::warn!` when the provider integration lands (this crate doesn't depend on tracing yet) so operators can notice corruption.

crates/storage/bloom/src/lib.rs
0xMars42: `num_hashes` is already a `u8` so it caps at 255. The real risk is 0 (no bits set, everything returns false-negative). I'll add `assert!(num_hashes > 0)` next to the existing size assert. For an upper bound, any value beyond ~20 is impractical but not dangerous (just slower lookups and faster saturation), so I'd keep it uncapped and let the caller choose.

crates/storage/bloom/benches/bloom_filter.rs
DeshErBojhaa, line 1: I am wondering whether it is also worth benchmarking on `num_hashes`. Maybe we can collect metrics on how the false-positive rate decreases as `num_hashes` increases.

For example, if we find some stat like false positive drops from 30% -> 10% if we increase `num_hashes` from 3 -> 4. Then we can discuss if it's worth the one extra hash.

0xMars42: Good idea. I added a test that sweeps K=1..8 at fixed capacity (1MB filter, 100K inserts) and prints measured FP rates. For reference, the theoretically optimal K minimizing FP rate for our default config (1GB filter, ~700M capacity) is ~8-9, but K=3 already achieves <0.5% FP while keeping per-query cost low (~400ns). The test confirms diminishing returns beyond K=3 — you can run it with `cargo test -p reth-storage-bloom false_positive_rate_vs_num_hashes -- --nocapture`.

PR #22470, opened 2026-02-22
refactor(rpc): deduplicate debug_trace_transaction using shared tracing

stevencartavia:
`debug_trace_transaction` had its own hand-rolled block replay implementation duplicating the logic already in `spawn_trace_transaction_in_block_with_inspector`. This was necessary because the shared helper required `Insp: Send`, which`DebugInspector` can't satisfy when the `js-tracer` feature is enabled (boa's `Context` holds `Rc` fields).

This PR changes the helper to accept an inspector factory closure instead of a pre-built inspector, so the inspector is created inside the blocking task and never needs to be `Send`. With that constraint lifted, `debug_trace_transaction`can use the shared helper directly, removing the duplicated code and one redundant DB round-trip.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -132,7 +132,11 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
             + 'static,
         R: Send + 'static,
     {
-        self.spawn_trace_transaction_in_block_with_inspector(hash, TracingInspector::new(config), f)
+        self.spawn_trace_transaction_in_block_with_inspector(
+            hash,
+            move || Ok(TracingInspector::new(config)),
+            move |tx_info, inspector, res, db, _evm_env, _tx_env| f(tx_info, inspector, res, db),
+        )
     }
 
     /// Retrieves the transaction if it exists and returns its trace.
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -157,10 +164,13 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
                 Insp,
                 ResultAndState<HaltReasonFor<Self::Evm>>,
                 StateCacheDb,
+                EvmEnvFor<Self::Evm>,
+                TxEnvFor<Self::Evm>,
             ) -> Result<R, Self::Error>
             + Send
             + 'static,
-        Insp: for<'a> InspectorFor<Self::Evm, &'a mut StateCacheDb> + Send + 'static,
+        Setup: FnOnce() -> Result<Insp, Self::Error> + Send + 'static,
```
klkvr, line 172: given that this always invokes inspector just once i think we can keep this taking the inspector itself instead of the closure?

klkvr (approved):

crates/rpc/rpc-eth-api/src/helpers/trace.rs
stevencartavia: i think that would undo the changes

PR #22473, opened 2026-02-22
perf(trie): inline len=1 storage proofs on account worker

gakonst:
## Summary
Inline single-target storage proofs on the account worker instead of dispatching them to the storage worker pool, avoiding channel send/recv and worker scheduling overhead for the majority of storage proof calls.

## Motivation
Tracy profiling shows ~82% of storage proof calls (2.18M out of 2.65M) have `targets.len() == 1`. Each of these currently goes through the full dispatch path: crossbeam channel send → storage worker recv → compute → channel send result → account worker recv. For a single-target proof, the dispatch overhead dominates the actual computation.

## Changes
- In `compute_v2_account_multiproof`, partition storage targets: inline len=1 proofs for accounts that are also in `account_targets`, dispatch the rest
- Compute inlined proofs using the account worker's existing `v2_storage_calculator`, cache the storage root immediately
- Pass `inlined_accounts` set to `dispatch_v2_storage_proofs` to skip synthetic root-only dispatches for already-covered accounts
- Add `inlined_count` metric to `ValueEncoderStats` and `ProofTaskTrieMetrics` for observability

## Design decisions
- **Inline only when address is in `account_targets`**: these are the proofs that gate account-leaf encoding (storage root needed). Non-overlapping storage proofs stay dispatched to preserve concurrency.
- **Conservative scope**: only targets len=1, not a broader threshold, to avoid turning the account worker into a bottleneck for heavier storage proofs.

## Testing
- `cargo test -p reth-trie-parallel` — all pass
- `cargo test -p reth-trie --lib proof_v2` — all pass
- Compiles with and without `metrics` feature

Prompted by: yk

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -1062,8 +1062,43 @@ where
 
         trace!(target: "trie::proof_task", "Processing V2 account multiproof");
 
-        let storage_proof_receivers =
-            dispatch_v2_storage_proofs(&self.storage_work_tx, &account_targets, storage_targets)?;
+        // Inline len=1 storage proofs on the account worker to avoid dispatcher overhead.
+        // Tracy shows ~82% of storage proof calls have a single target.
+        let account_target_addresses: B256Set = account_targets.iter().map(|t| t.key()).collect();
+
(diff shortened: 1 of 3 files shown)


yongkangc:
derek bench

PR #22486, opened 2026-02-23
perf(trie): skip buffer allocation for in-order receipt root items

yongkangc:
Skips a heap alloc+copy+free per receipt by feeding bytes directly to the hash builder when receipts arrive in order. Skips ~N-1 allocs per block

**Summary:**

- When the pushed item is next in trie order, feed bytes directly to `HashBuilder::add_leaf` — skipping the `to_vec()` + buffer + flush cycle
- 199/200 receipts hit this fast path; only index 0 needs buffering
- A `pushed: Vec<bool>` bitmap handles duplicate detection since fast-path items bypass `pending`

**Per-receipt savings (fast path hit)**:

| Operation | Saved |
|---|---|
| `bytes.to_vec()` | 1 heap alloc + memcpy (~200 bytes) |
| `pending[index] = Some(...)` | 1 pointer-sized write |
| `flush() → .take()` | 1 read + reset |
| `flush() → add_leaf(&value)` | 1 extra memcpy (~200 bytes) |
| `drop(Vec<u8>)` | 1 heap dealloc |

Branch cost: ~2-3 cycles of inlined comparisons per push, predicted correctly 199/200 times.

**Bonus bug fix**: The bitmap fixes a latent bug where `flush()` calling `.take()` reset pending slots to `None`, allowing silent duplicate acceptance after flush.

### Notes

RLP encodes `0` as `0x80` but `1`-`127` as raw bytes, so the trie inserts index 0 *after* 1-127:

```
Execution order:  0   1   2  ...  127   128  ...  199
Trie order:       1   2   3  ...  127   0    128  ...  199
```

Receipts arrive sequentially (0, 1, 2, ...) but the trie expects 1 first:

```
Push 0:   trie wants 1   → buffer          ← only slow path
Push 1:   trie wants 1   → fast path ✓
Push 2:   trie wants 2   → fast path ✓
  ...
Push 127: trie wants 127 → fast path ✓     → flushes buffered 0
Push 128: trie wants 128 → fast path ✓
  ...
Push 199: trie wants 199 → fast path ✓     → 199/200 skip the buffer
```

### DIFF
--- a/crates/trie/common/src/ordered_root.rs
+++ b/crates/trie/common/src/ordered_root.rs
@@ -138,6 +138,8 @@ pub struct OrderedTrieRootEncodedBuilder {
     next_insert_i: usize,
     /// Buffer for pending items, indexed by execution index.
     pending: Vec<Option<Vec<u8>>>,
+    /// Tracks which indices have been pushed (for duplicate detection).
+    pushed: Vec<bool>,
     /// The underlying hash builder.
     hb: HashBuilder,
 }
@@ -150,6 +152,7 @@ impl OrderedTrieRootEncodedBuilder {
             received: 0,
             next_insert_i: 0,
             pending: alloc::vec![None; len],
+            pushed: alloc::vec![false; len],
             hb: HashBuilder::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


decofe:
cc @yongkangc

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/22299815642/job/64504684359)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `codex/receipt-root-fast-path-1771836752`

PR #22494, opened 2026-02-23
feat(rpc): accept optional timeout parameter in eth_sendRawTransactionSync per EIP-7966

merge-script:
The current eth_sendRawTransactionSync signature only accepts the raw tx bytes. Clients following EIP-7966 that send a second timeout_ms parameter (e.g. params: ["0xf86c...", 5000]) get an invalid params error from jsonrpsee because the arity doesn't match.

This adds the optional timeout_ms parameter and caps it against the server-configured max:

effective_timeout = min(client_timeout_ms, server_max_timeout)
When omitted, behavior is unchanged — the server default (30s) applies. When provided, the client can request a shorter wait but never exceed the server cap.

Note: EIP-7966 is still in Draft status, but since we already ships eth_sendRawTransactionSync, we should match the spec's actual interface to avoid breaking compliant callers.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/core.rs
+++ b/crates/rpc/rpc-eth-api/src/core.rs
@@ -374,7 +374,7 @@ pub trait EthApi<
     async fn send_raw_transaction_sync(
         &self,
         bytes: Bytes,
-        timeout_ms: Option<u64>,
+        timeout_ms: Option<U64>,
     ) -> RpcResult<R>;
 
     /// Returns an Ethereum specific signature with: sign(keccak256("\x19Ethereum Signed Message:\n"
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-eth-api/src/core.rs
```diff
@@ -367,8 +367,15 @@ pub trait EthApi<
...
     /// This will return a timeout error if the transaction isn't included within some time period.
+    ///
+    /// An optional `timeout_ms` parameter (in milliseconds) allows the client to specify a
+    /// per-request timeout, which is capped by the server-configured maximum.
     #[method(name = "sendRawTransactionSync")]
-    async fn send_raw_transaction_sync(&self, bytes: Bytes) -> RpcResult<R>;
+    async fn send_raw_transaction_sync(
+        &self,
+        bytes: Bytes,
+        timeout_ms: Option<u64>,
```
mattsse: should this be U64?

mattsse (commented):
I find this arg a bit redundant, because the caller can control this as well

crates/rpc/rpc-eth-api/src/core.rs
merge-script: The thing is, client-side timeout just drops the connection — the server doesn't know about it and keeps waiting on the canonical state stream for the full 30s doing nothing. With server-side awareness we can cut that short and free up resources earlier.

But yeah it's a draft EIP so I get it if you don't think it's worth the extra surface. Let me know either way.

On the type — yeah U64 makes more sense, Fixed.

PR #22501, opened 2026-02-23
fix(net): terminate discv4 background task when discv5 initialization fails

HuberyJulianay:
- Fix resource leak where discv4 background tasks and UDP socket persist indefinitely when discv5 initialization fails
- Replace tokio::try_join! with tokio::join! in Discovery::new so we can inspect individual results and clean up on partial failure
- Explicitly call discv4.terminate() before returning the error, which sends Discv4Command::Terminated to gracefully shut down the spawned service


mattsse (commented):
we can keep this as is

PR #22503, opened 2026-02-23
refactor(rpc): fetch block before tracing to avoid double lookups

stevencartavia:
Fetches the block once upfront in `trace_block`, `trace_block_opcode_gas`, and `trace_block_storage_access`, then passes it via `Some(block)` to avoid redundant `recovered_block` calls inside `trace_block_with` / `trace_block_inspector`.

### DIFF
--- a/crates/rpc/rpc/src/trace.rs
+++ b/crates/rpc/rpc/src/trace.rs
@@ -476,26 +476,28 @@ where
         &self,
         block_id: BlockId,
     ) -> Result<Option<Vec<LocalizedTransactionTrace>>, Eth::Error> {
-        let traces = self.eth_api().trace_block_with(
-            block_id,
-            None,
-            TracingInspectorConfig::default_parity(),
-            |tx_info, mut ctx| {
-                let traces = ctx
-                    .take_inspector()
-                    .into_parity_builder()
-                    .into_localized_transaction_traces(tx_info);
-                Ok(traces)
-            },
-        );
+        let Some(block) = self.eth_api().recovered_block(block_id).await? else {
+            return Ok(None);
+        };
 
-        let block = self.eth_api().recovered_block(block_id);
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

crates/rpc/rpc/src/trace.rs
```diff
@@ -583,11 +588,15 @@ where
...
+        let Some(block) = self.eth_api().recovered_block(block_id).await? else {
+            return Err(EthApiError::HeaderNotFound(block_id).into());
+        };
+
+        let Some(transactions) = self
             .eth_api()
             .trace_block_inspector(
                 block_id,
-                None,
+                Some(block.clone()),
```
mattsse, line 599: this clones seems wasteful?

if I see this correctly then we only need numhash which we can store before calling this

crates/rpc/rpc/src/trace.rs
```diff
@@ -476,26 +476,28 @@ where
...
+            return Err(EthApiError::HeaderNotFound(block_id).into());
+        };
 
-        let mut maybe_traces =
-            maybe_traces.map(|traces| traces.into_iter().flatten().collect::<Vec<_>>());
+        let mut traces = self
+            .eth_api()
+            .trace_block_with(
+                block_id,
+                Some(block.clone()),
```
mattsse, line 487: similar here?

mattsse (changes requested):

crates/rpc/rpc/src/trace.rs
DaniPopes: i believe it's arc'd

stevencartavia: yes, they are arc'd

mattsse (approved):

PR #22516, opened 2026-02-24
perf(trie): replace Box clone with unsafe reborrow in prune

gakonst:
## Summary
Eliminates a 512-byte heap allocation (`Box<[B256; 16]>::clone()`) per branch node at `max_depth` during `ParallelSparseTrie::prune` by using unsafe lifetime decoupling to mutate `blinded_mask` and `blinded_hashes` in place.

## Changes
- Replace `blinded_hashes.clone()` + write-back pattern with `decouple_lt_mut` to mutate branch fields in-place
- Remove the second `subtrie_for_path_mut_untracked(&path)` lookup that was only needed to write back the cloned values
- Add `decouple_lt` / `decouple_lt_mut` unsafe helpers for lifetime decoupling

## Safety
The parent branch node at `path` (depth = `max_depth`) lives in the upper subtrie, while child nodes at `child` (depth = `max_depth + 1`) are in lower subtries. `subtrie_for_path_mut_untracked(&child)` returns a different subtrie, so the mutable references do not alias.

## Testing
All 95 tests in `reth-trie-sparse` pass.

Prompted by: danipopes

### DIFF
--- a/crates/trie/sparse/src/parallel.rs
+++ b/crates/trie/sparse/src/parallel.rs
@@ -1235,8 +1235,14 @@ impl SparseTrie for ParallelSparseTrie {
                 SparseNode::Branch { state_mask, blinded_mask, blinded_hashes, .. } => {
                     // For branch nodes at max depth, collapse all children onto them,
                     if depth == max_depth {
-                        let mut blinded_mask = *blinded_mask;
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
lgtm cc @klkvr

PR #22520, opened 2026-02-24
refactor(engine): add payload validator hook for state root validation

panos-xyz:
## Summary
- add a `PayloadValidator` hook for validating computed state roots
- route engine-tree state-root checks through the hook (both validation points)
- keep default behavior unchanged (strict equality)
- add a test covering default-vs-skipped policy behavior

## Motivation
Some networks and fork transitions need a custom state-root acceptance policy while still reusing the engine tree execution/validation pipeline. Today the engine tree hard-codes a strict equality check in multiple places, which makes this customization require duplicating large parts of `BasicEngineValidator`.

This change introduces a small extension point in `PayloadValidator` so downstream implementations can customize the policy without forking engine-tree internals.

## Behavior
Default behavior is unchanged:
- computed state root must equal the header state root
- mismatches return `ConsensusError::BodyStateRootDiff`

Custom validators can override the hook to accept or skip comparison for network-specific cases.


crates/engine/tree/src/tree/tests.rs
```diff
@@ -58,6 +70,25 @@ impl reth_engine_primitives::PayloadValidator<EthEngineTypes> for MockEngineVali
         })?;
         Ok(block.seal_slow())
     }
+
+    fn validate_computed_state_root(
+        &self,
+        block: &reth_primitives_traits::RecoveredBlock<Self::Block>,
+        computed_state_root: B256,
```
mattsse: wouldnt this be giga wasteful if the payload validatin computes the SR but then always skips it?

mattsse (commented):
this is def something we want to make more ergonomic

currently the way this is intended to be modified is via:

https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/engine/tree/src/tree/payload_validator.rs#L1715-L1715

specifically:

https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/engine/tree/src/tree/payload_validator.rs#L1754-L1760

however, currently it would require a bunch of code duplication from  
https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/engine/tree/src/tree/payload_validator.rs#L1754-L1760 

I think there's room to abstract the SR validation inside the basic type.

it'd be helpful if you could experience with this a bit, if it's easier to write oyur own

https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/engine/tree/src/tree/payload_validator.rs#L1715-L1715

or how to configure htis inside the BasicEngineValidator how the SR should be computed+validated

crates/engine/tree/src/tree/tests.rs
panos-xyz: fixed

PR #22524, opened 2026-02-24
feat(db): warmup branch pages for key tables on startup

gakonst:
## Summary

Adds `mdbx_dbi_warmup(txn, dbi)` that walks a table's B-tree touching only branch (internal) pages, then calls it during `init_db` for the hot-path tables.

## Motivation

On cold start, the first block executions and state root computations hit page faults on every B-tree internal node lookup. Branch pages are a tiny fraction of total data (~0.1% for large tables) but sit on every lookup path. Preloading them into the OS page cache eliminates these faults.

## Changes

- `mdbx.h` / `mdbx.c`: new `mdbx_dbi_warmup()` C function. Uses tree height to walk only branch levels (depths 0..height-2), never touches leaf pages. Recursive with max depth bounded by `CURSOR_STACK` (~20 levels).
- `libmdbx-rs/src/transaction.rs`: `Transaction::warmup_branch_pages(dbi)` Rust binding.
- `db/src/implementation/mdbx/mod.rs`: `DatabaseEnv::warmup_branch_pages()` that opens tables, calls warmup, logs branch page count per table.
- `db/src/mdbx.rs`: calls warmup in `init_db_for()` for `AccountsTrie`, `StoragesTrie`, `HashedAccounts`, `HashedStorages`, `PlainAccountState`, `PlainStorageState`.

## Testing

```
cargo test -p reth-libmdbx  # 11 passed
cargo test -p reth-db       # 38 passed
```

Prompted by: alexey

### DIFF
--- a/crates/storage/db/src/implementation/mdbx/mod.rs
+++ b/crates/storage/db/src/implementation/mdbx/mod.rs
@@ -615,6 +615,55 @@ impl DatabaseEnv {
         }
     }
 
+    /// Warms up the branch pages of the specified tables by faulting them into the mmap.
+    pub fn warmup_branch_pages(&self, table_names: &[&str]) {
+        let tx = match self.inner.begin_ro_txn() {
+            Ok(tx) => tx,
+            Err(e) => {
+                reth_tracing::tracing::warn!(
+                    target: "storage::db::mdbx",
+                    %e,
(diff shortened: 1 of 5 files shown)


shekhirin:
@decofe bench warmup=500 blocks=2000 samply

PR #22528, opened 2026-02-24
perf(txpool): use BTree range queries in pending_txs_by_sender/queued_txs_by_sender

figtracer:


crates/transaction-pool/src/pool/parked.rs
```diff
@@ -151,6 +151,18 @@ impl<T: ParkedOrd> ParkedPool<T> {
...
 
+    /// Returns all transactions for the given sender, using a BTree range query.
+    pub(crate) fn txs_by_sender(
+        &self,
+        sender: SenderId,
+    ) -> Vec<Arc<ValidPoolTransaction<T::Transaction>>> {
+        self.by_id
+            .range((sender.start_bound(), Unbounded))
+            .take_while(move |(other, _)| sender == other.sender)
+            .map(|(_, tx)| tx.transaction.clone().into())
```
mattsse: this should be a .tx.clone() instead, right?

mattsse (changes requested):

crates/transaction-pool/src/pool/parked.rs
figtracer: tx is BasefeeOrd<T> here cant .clone() wrapped it

mattsse (approved):

PR #22529, opened 2026-02-24
fix(net/tx): move transactions_by_peers removal after guard in on_bad_import

figtracer:


crates/net/network/src/transactions/mod.rs
```diff
@@ -489,15 +489,14 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
     /// - not blob transaction (tx type mismatch)
     /// - wrong versioned kzg commitment hash
     fn on_bad_import(&mut self, err: PoolError) {
-        let peers = self.transactions_by_peers.remove(&err.hash);
-
         // if we're _currently_ syncing, we ignore a bad transaction
         if !err.is_bad_transaction() || self.network.is_syncing() {
             return
         }
```
mattsse, line 495: this will bypass the cleanup here, so we always need to remove the peers by hash

PR #22532, opened 2026-02-24
fix(rpc): prevent u64 underflow when re-executing genesis block

MagicJoshh:
Was running a script locally to pull execution outcomes for a range of blocks on a debug build, and it panicked at block 0 with "attempt to subtract with overflow". And realize it's just `start_block - 1` on line 158 not handling the genesis case.

In release mode it wraps to u64::MAX and you get some confusing provider error that doesn't tell you what actually went wrong. Added an early return with a clear error message for block 0 since there's no prior state to re-execute against anyway.

### DIFF
--- a/crates/rpc/rpc/src/reth.rs
+++ b/crates/rpc/rpc/src/reth.rs
@@ -156,9 +156,7 @@ where
         };
 
         if start_block == 0 {
-            return Err(EthApiError::InvalidParams(
-                "cannot re-execute genesis block".into(),
-            ));
+            return Ok(Some(ExecutionOutcome::default()))
         }
 
         let state_provider = self.provider().history_by_block_number(start_block - 1)?;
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc/src/reth.rs
```diff
@@ -155,6 +155,12 @@ where
             return Ok(None)
         };
 
+        if start_block == 0 {
+            return Err(EthApiError::InvalidParams(
+                "cannot re-execute genesis block".into(),
+            ));
+        }
```
mattsse, line 160: should this just return an empty outcome instead? because technically this block can be executed I guess, even if 0 txs

MagicJoshh: Make sense, changed to return an empty execution outcome for block 0 instead of erroring.

mattsse (approved):

PR #22534, opened 2026-02-24
Chore/remove dead code sparse trie

Kinaqu:
Removed stale #[allow(dead_code)] attributes in sparse trie state; no functional change.


crates/trie/sparse/src/state.rs
```diff
@@ -415,7 +415,7 @@ where
             trie.reveal_nodes(&mut nodes)?;
 
             self.deferred_drops.proof_nodes_bufs.push(nodes);
-            return Ok(())
```
shekhirin, line 418: please revert semicolon changes, otherwise looks good

Kinaqu: Reverted the semicolon-only changes in state.rs. Thanks!

PR #22533, opened 2026-02-24
feat(rpc): accept RLP-encoded blocks in reth_newPayload

gakonst:
## Summary

Accept either `ExecutionData` or an RLP-encoded block in `reth_newPayload`.

## Changes

- Add `RethNewPayloadInput<ExecutionData>` untagged enum with two variants: `ExecutionData` (existing JSON object) and `RlpEncodedBlock` (hex-encoded RLP bytes)
- Update `RethEngineApi` / `RethEngineApiServer` trait to take the new enum
- Decode RLP blocks via `Block::decode` and convert to `ExecutionData` via `PayloadTypes::block_to_payload`
- Fully backward-compatible: existing callers sending `ExecutionData` JSON will match the first untagged variant

Prompted by: arsenii

PR #22537, opened 2026-02-24
feat(bench): support RLP-encoded blocks via debug_getRawBlock

gakonst:
## Summary

Add `--rlp-blocks` flag to reth-bench that fetches blocks via `debug_getRawBlock` and sends the raw RLP bytes directly to `reth_newPayload`.

Depends on #22533 for the server-side `RlpEncodedBlock` variant in `reth_newPayload`.

## Changes

- Add `--rlp-blocks` flag to `BenchmarkArgs` (implies `--reth-new-payload`)
- Add `fetch_raw_block` helper that calls `debug_getRawBlock`
- Add `call_reth_new_payload_rlp` that sends raw RLP bytes via `reth_newPayload`
- Update `new-payload-only` and `new-payload-fcu` commands to support the RLP path
- RLP blocks are decoded locally only to extract header metadata (hash, gas, tx count) for metrics and FCU

Prompted by: arsenii

### DIFF
--- a/bin/reth-bench/src/bench/new_payload_fcu.rs
+++ b/bin/reth-bench/src/bench/new_payload_fcu.rs
@@ -23,13 +23,11 @@ use crate::{
     },
     valid_payload::{
         block_to_new_payload, call_forkchoice_updated, call_new_payload_with_reth,
-        call_reth_new_payload_rlp, fetch_raw_block,
+        call_reth_new_payload_rlp, fetch_and_decode_raw_block,
     },
 };
-use alloy_consensus::BlockHeader;
 use alloy_primitives::{Bytes, B256};
 use alloy_provider::{network::AnyRpcBlock, Provider};
-use alloy_rlp::Decodable;
 use alloy_rpc_types_engine::ForkchoiceState;
 use clap::Parser;
 use eyre::{Context, OptionExt};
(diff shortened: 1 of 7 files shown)


bin/reth-bench/src/valid_payload.rs
```diff
@@ -368,6 +368,55 @@ pub(crate) async fn call_new_payload_with_reth<N: Network, P: Provider<N>>(
     }
 }
 
+/// Sends an RLP-encoded block to the `reth_newPayload` endpoint.
+///
+/// The server decodes the RLP and converts it to `ExecutionData` via `block_to_payload`.
+/// Returns server-side timing breakdown.
+pub(crate) async fn call_reth_new_payload_rlp<N: Network, P: Provider<N>>(
```
klkvr, line 375: can we just reuse `call_new_payload_with_reth` and make `reth_execution_data` data take the rlp or `ExecutionData`?

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -170,46 +181,24 @@ impl Command {
...
-
-                next_block += 1;
-                if let Err(e) = sender
-                    .send((block, head_block_hash, safe_block_hash, finalized_block_hash))
-                    .await
-                {
-                    tracing::error!(target: "reth-bench", "Failed to send block data: {e}");
-                    break;
                 }
             }
```
klkvr, line 200: ideally we keep this mostly unchanged. it's fine to fetch block hashes from the provider in the rlp case

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -219,33 +208,69 @@ impl Command {
...
+                        let timings = call_reth_new_payload_rlp(&auth_provider, rlp).await?;
+                        (
+                            block_number,
+                            transaction_count,
+                            gas_used,
+                            gas_limit,
+                            Some(timings),
+                            EngineApiMessageVersion::V4,
+                        )
+                    }
```
klkvr, line 241: same here, let's keep diff minimal

PR #22539, opened 2026-02-24
perf: Remove unnecessary single-target storage proofs

mediocregopher:
Two small cleanups to the proof task infrastructure:

**Remove `sequence_number` from `ProofResultMessage` and `ProofResultContext`**

The `sequence_number` field was carried through the proof pipeline but never read by any consumer. This removes it from both structs and all construction/destructuring sites.

Note: the unrelated `sequence_number` in `MultiProofMessage::EmptyProof` is a different field and remains unchanged.

**Stop generating single-target storage proofs for accounts without storage targets**

Previously, `dispatch_v2_storage_proofs` would generate a synthetic `B256::ZERO` storage proof target for every account that had an account proof target but no storage targets, just to obtain its storage root. This is unnecessary — the storage root for these accounts can be computed synchronously from the sparse trie instead. Removing this avoids dispatching extra work to the storage worker pool and waiting on results that aren't needed.


mediocregopher:
derek bench

klkvr:
derek bench

klkvr (approved):
nice

DaniPopes (approved):

PR #22540, opened 2026-02-24
chore(tasks): remove quanta upkeep from runtime

gakonst:
Removes the quanta upkeep thread from `RuntimeInner`. Drops the `quanta` dep from `reth-tasks`.

Prompted by: DaniPopes


DaniPopes (approved):
We don't make use of Instant::recent; ::now is good enough.

PR #22545, opened 2026-02-24
feat(net): use fork_id as tiebreaker in peer selection

mattsse:
When selecting the best unconnected peer for outbound connections, use the presence of a discovered `fork_id` to break ties between peers with equal reputation. A peer with a `fork_id` is preferred since it indicates the peer is on a compatible fork.


Rjected (approved):
Lgtm, makes sense

PR #22547, opened 2026-02-24
perf(discv4): cache signed FindNode packets during Kademlia lookups

figtracer:
## Summary

During a Kademlia lookup, `find_node()` constructs a fresh `FindNode` message for each peer and passes it through `send_packet()` → `Message::encode()`, which performs a full `keccak256` + `secp256k1::sign_ecdsa_recoverable` for every single packet.

With `ALPHA = 3` FindNode messages per round and multiple recursive rounds triggered by `on_neighbours()`, a single lookup cycle produces tens of redundant ECDSA sign operations.

**Key insight:** All FindNode messages within a lookup share the **same target ID**. The `FindNode` struct only contains `(id, expire)` — no per-destination fields. Two packets sent milliseconds apart will have nearly identical expiration values, so there is no need to re-sign.

## Changes

Adds a `CachedFindNode` field to `Discv4Service` that caches the last signed FindNode packet. When `find_node()` is called:

1. Check if a cached packet exists with the same target and has not expired (TTL = `request_timeout / 4`)
2. **Cache hit:** reuse the pre-signed `Bytes` payload directly, skip `encode()`/ECDSA sign entirely
3. **Cache miss:** encode + sign as usual, store the result

The TTL is derived from `request_timeout / 4` (default: 5s with 20s timeout) to ensure the `expire` timestamp embedded in the cached packet always has sufficient remaining validity at the receiver.

## Safety

- `FindNode` has no per-destination fields, so the same signed bytes are valid for any peer
- The packet hash return from `send_packet()` is already discarded in `find_node()`, so caching does not change any observable behavior
- Error handling on `egress.try_send()` matches the existing `send_packet()` pattern (logged with `debug!`)

### DIFF
--- a/crates/net/discv4/src/lib.rs
+++ b/crates/net/discv4/src/lib.rs
@@ -503,6 +503,8 @@ pub struct Discv4Service {
     received_pongs: PongTable,
     /// Interval used to expire additionally tracked nodes
     expire_interval: Interval,
+    /// Cached signed `FindNode` packet to avoid redundant ECDSA signing during lookups.
+    cached_find_node: Option<CachedFindNode>,
 }
 
 impl Discv4Service {
@@ -603,6 +605,7 @@ impl Discv4Service {
             queued_events: Default::default(),
             received_pongs: Default::default(),
             expire_interval: tokio::time::interval(EXPIRE_DURATION),
+            cached_find_node: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
makes sense

can we make this a standalone fn that returns the message, then this is easier to understand

crates/net/discv4/src/lib.rs
```diff
@@ -811,9 +814,23 @@ impl Discv4Service {
     fn find_node(&mut self, node: &NodeRecord, ctx: LookupContext) {
         trace!(target: "discv4", ?node, lookup=?ctx.target(), "Sending FindNode");
         ctx.mark_queried(node.id);
-        let id = ctx.target();
-        let msg = Message::FindNode(FindNode { id, expire: self.find_node_expiration() });
-        self.send_packet(msg, node.udp_addr());
```
mattsse, line 816: why are we no longer using send_packet and are now doing let _ = self.egress.try_send

mattsse, line 816: what I wanted this to look like is

```
let msg = self.find_node_message()
```

which then does the cache check

mattsse (changes requested):
pedantic style nit

crates/net/discv4/src/lib.rs
figtracer: send_packet goes through 

https://github.com/paradigmxyz/reth/blob/0df9791bea08fc47d6693a4dc7ba6057c6b3416d/crates/net/discv4/src/proto.rs#L117

and we should skip if using the cache

we could make send_packet accept pre-encoded payloads instead and not do this

mattsse (approved):
makes a lot of sense when we send a lot of messages in a row

PR #22551, opened 2026-02-24
perf(discv4): trigger immediate lookup on first bootnode pong

figtracer:
On cold start, the first `lookup_interval` tick fires at t=0 before any bootnode has proven its endpoint, wasting it. The next tick doesn't fire until t=20s, leaving a ~20s dead window where the node has proven bootnodes but isn't asking them about peers.
> 
This adds a `pending_lookup_reset` flag that resets the `lookup_interval` as soon as the first bootnode pong arrives, so the next `poll_tick` fires immediately. This cuts first-peer discovery from ~20s to ~200ms on fresh nodes.

### DIFF
--- a/crates/net/discv4/src/lib.rs
+++ b/crates/net/discv4/src/lib.rs
@@ -489,6 +489,8 @@ pub struct Discv4Service {
     lookup_interval: Interval,
     /// Used to rotate targets to lookup
     lookup_rotator: LookupTargetRotator,
+    /// Whether we still need to reset the lookup interval on the first bootnode pong.
+    pending_lookup_reset: bool,
     /// Interval when to recheck active requests
     evict_expired_requests_interval: Interval,
     /// Interval when to resend pings.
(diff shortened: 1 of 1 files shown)


crates/net/discv4/src/lib.rs
```diff
@@ -1295,9 +1298,15 @@ impl Discv4Service {
         match reason {
             PingReason::InitialInsert => {
                 self.update_on_pong(node, pong.enr_sq);
+                // Trigger an immediate self-lookup once the first bootnode proves its
+                // endpoint, so we don't wait for the next lookup_interval tick (~20s).
+                if self.pending_initial_lookup && self.config.bootstrap_nodes.contains(&node) {
+                    self.pending_initial_lookup = false;
+                    self.lookup_self();
+                }
```
mattsse: hmm, could we instead reset the the lookup interval here?

mattsse (changes requested):
this makes a lot of sense actually, can we try resetting the interval instead

another idea would be that we use a lower interval for the first tick?

crates/net/discv4/src/lib.rs
figtracer: nice

> another idea would be that we use a lower interval for the first tick?

lower interval for first tick is also a good idea and probably simpler but imo thhis approach is more reliable

mattsse (approved):
the bool check is pretty cheap here so this makes sense

PR #22553, opened 2026-02-25
feat(exex): expose ExExManager buffer capacity in ExExLauncher

Elaela22soL:
Fixes #22509

`ExExLauncher` hardcoded `DEFAULT_EXEX_MANAGER_CAPACITY` (1024) when creating the `ExExManager`, with no way for operators to override it. Add a `capacity` field defaulting to `DEFAULT_EXEX_MANAGER_CAPACITY` and a `with_capacity()` builder method following the existing `with_wal_blocks_warning()` pattern.


mattsse (approved):
sure why not

PR #22554, opened 2026-02-25
perf(prewarm): skip prewarming for the first transaction

pepyakin:
## Summary

Skip prewarming for the first transaction in a block. The first transaction is already being executed immediately by the main execution thread, so spawning a prewarm worker for it adds overhead without benefit.

Prewarming now starts from the second transaction (index 1) onwards.

## Changes

- `convert_serial`: skip sending index 0 to the prewarm channel
- The parallel conversion path (`rayon`) always starts at index ≥ `PARALLEL_PREFETCH_COUNT` (4), so no change is needed there


pepyakin:
@decofe bench

pepyakin:
derek bench

decofe:
cc @gakonst

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/22384336107/job/64791700895)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `skip-first-tx-prewarm`

PR #22557, opened 2026-02-25
feat(net): persist richer peer metadata to peers file

mattsse:
Introduces `PersistedPeerInfo` which preserves peer kind, fork ID, and reputation when writing peers to disk, instead of only storing bare `NodeRecord` addresses.

- `write_peers_to_file` now filters out backed-off and banned peers
- The reader supports both the new format and the legacy `Vec<NodeRecord>` format for backwards compatibility on first restart after upgrade
- Adds serde derives to `PeerKind` (feature-gated)

### DIFF
--- a/.changelog/safe-waves-read.md
+++ b/.changelog/safe-waves-read.md
@@ -1,0 +1,7 @@
+---
+reth-network-types: minor
+reth-network: minor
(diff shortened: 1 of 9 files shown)


Rjected (approved):
lgtm

PR #22558, opened 2026-02-25
chore(engine): add debug log in spawned tx iterator after yielding tx index

gakonst:
Adds a `debug!` log in the spawned tx iterator (`convert_serial` + parallel `for_each_ordered` path) that fires after each transaction index is yielded to the execution channel. Useful for tracing tx flow through the payload processor pipeline.

Prompted by: danipopes

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -431,8 +431,13 @@ where
                             tx
                         })
                     })
-                    .for_each_ordered(|tx| {
-                        let _ = execute_tx.send(tx);
+                    .for_each_ordered({
+                        let mut idx = prefetch;
+                        move |tx| {
+                            let _ = execute_tx.send(tx);
+                            debug!(target: "engine::tree::payload_processor", idx, "yielded transaction");
+                            idx += 1;
+                        }
                     });
             });
         }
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
mattsse, line 414: why are we doing this btw

wasnt the entire point of this so that we dont need to do this sequentially?

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -431,8 +431,13 @@ where
...
                     })
-                    .for_each_ordered(|tx| {
-                        let _ = execute_tx.send(tx);
+                    .for_each_ordered({
+                        let mut idx = prefetch;
+                        move |tx| {
+                            let _ = execute_tx.send(tx);
+                            debug!(target: "engine::tree::payload_processor", idx, "yielded transaction");
+                            idx += 1;
+                        }
```
mattsse: this looks a bit weird, is this different from using the .enumerate idx?

DaniPopes (approved):

crates/engine/tree/src/tree/payload_processor/mod.rs
DaniPopes: what do you mean?

DaniPopes: this is just input iterator without any operation collecting to a vec

PR #22560, opened 2026-02-25
fix(evm): use timestamp-based blob params for first post-Cancun block

davidjsonn:
Use `blob_params_at_timestamp` instead of hardcoded `BlobParams::cancun()` for the first post-Cancun block. Fixes incorrect `excess_blob_gas` when Cancun and Prague/Osaka are activated simultaneously (dev mode, testnets).


mattsse (commented):
we can keep this as is

PR #22569, opened 2026-02-25
perf: always chunk proofs

klkvr:


klkvr:
derek bench cores=8 warmup=500 blocks=2000

PR #22574, opened 2026-02-25
feat: Improve txcount binary search

### LINKED ISSUE #22561: Improve txcount binary search
### Describe the feature

https://github.com/paradigmxyz/reth/blob/0bdf6e2f2e81d7466059fe76cc07d901e4d9a947/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L350-L358
this currently does a lot of spawns+ new tx to obtain the txcount

https://github.com/paradigmxyz/reth/blob/0bdf6e2f2e81d7466059fe76cc07d901e4d9a947/crates/rpc/rpc-eth-api/src/helpers/state.rs#L388-L396

this could be more efficient if we had a way to optimize this

we can specialize this as fn find_transaction_by_sender_and_nonce for example that does the binary search over just:

https://github.com/paradigmxyz/reth/blob/0bdf6e2f2e81d7466059fe76cc07d901e4d9a947/crates/rpc/rpc-eth-api/src/helpers/state.rs#L397-L403

@stevencartavia 

### Additional context

_No response_

stevencartavia:
Closes https://github.com/paradigmxyz/reth/issues/22561

Each binary search iteration now reads the account nonce directly via `history_by_block_number().account_nonce()`, bypassing `transaction_count` which had unnecessary overhead (pending block logic, pool checks, state indirection).

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
@@ -24,7 +24,7 @@ use reth_primitives_traits::{
 use reth_rpc_convert::{transaction::RpcConvert, RpcTxReq, TransactionConversionError};
 use reth_rpc_eth_types::{
     block::convert_transaction_receipt,
-    utils::recover_raw_transaction,
+    utils::{binary_search, recover_raw_transaction},
     EthApiError::{self, TransactionConfirmationTimeout},
     FillTransaction, SignError, TransactionSource,
 };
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -335,27 +335,58 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
-            let num = binary_search::<_, _, Self::Error>(1, high, |mid| async move {
-                let mid_nonce =
-                    self.transaction_count(sender, Some(mid.into())).await?.saturating_to::<u64>();
+                    Ok(Some(num))
+                })
+                .await?;
 
-                Ok(mid_nonce > nonce)
-            })
-            .await?;
```
mattsse, line 358: I believe we should be able to resue this if we update the closure code to do the raw history lookups

mattsse (changes requested):

mattsse (commented):
I thought about this a bit and we cant easily make this change because this messes with the include_pending setting

doing a binary search over the spawn inst to bad I think

PR #22575, opened 2026-02-25
fix(rpc): add missing `apply_pre_execution_changes` in `spawn_replay_transaction`

merge-script:
`spawn_replay_transaction` replays a block's transactions from parent state but was missing the `apply_pre_execution_changes` call that sets up EIP-4788 beacon root contract state, EIP-2935 block hash history, and EIP-161 state clearing. The parallel function `spawn_trace_transaction_in_block_with_inspector` in trace.rs correctly includes this call. Without it, `ots_getTransactionError` can return incorrect results for any transaction that depends on system contract state, affecting all post-Cancun blocks.


mattsse (approved):
I believe this is right and we need to do this for replay for sure

stevencartavia (approved):

PR #22576, opened 2026-02-25
fix(storage): return early in RocksDB healing when checkpoint is 0

decofe:
## Summary

Fix infinite healing loop on startup when RocksDB tables are empty and checkpoint is 0 but static files are ahead.

## Motivation

Reported by ethstaker Discord users after upgrading to v1.11.x. When `checkpoint=0` and the RocksDB `TransactionHashNumbers` table is empty but static files have data (e.g. `sf_tip=24500954`, `sf_tip_end_tx=3279227181`), the healer falls through to an expensive loop that iterates 3.2B transactions in 10k batches, deleting nothing. The MDBX read transaction times out after 5 minutes (`error code -96000`), the node crashes, and the loop restarts indefinitely.

The same pattern exists in `StoragesHistory` and `AccountsHistory` healers.

## Changes

- All three RocksDB healers (`TransactionHashNumbers`, `StoragesHistory`, `AccountsHistory`) now unconditionally return early when `checkpoint == 0`. `clear()` on an empty table is a no-op.
- Added test covering checkpoint=0 with empty RocksDB.

## Testing

```
cargo test -p reth-provider --features rocksdb -- check_consistency
```

All 18 invariant tests pass.

Prompted by: alexey

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -112,7 +112,7 @@ impl RocksDBProvider {
             .get_highest_static_file_block(StaticFileSegment::Transactions)
             .unwrap_or(0);
 
-        // Fast path: nothing has been indexed yet — clear any stale data and return.
+        // Fast path: clear any stale data and return.
         if checkpoint == 0 {
             tracing::info!(
                 target: "reth::providers::rocksdb",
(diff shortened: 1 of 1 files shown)


Rjected (approved):
this makes sense to me, although it makes me wonder if we could encounter a situation where a mdbx read tx is held for a long time during healing, _and_ we have checkpoints > 0

PR #22577, opened 2026-02-25
fix(rpc): add missing `apply_pre_execution_changes` in `debug_traceCallMany`

merge-script:
`debug_trace_call_many` replays block transactions from parent state when `replay_block_txs` is true, but was missing the `apply_pre_execution_changes` call. The sibling function `debug_trace_call_at_tx_index` in the same file correctly includes this step with a numbered comment "1. apply pre-execution changes", confirming it was an oversight. Without it, EIP-4788 beacon root and EIP-2935 block hash history are not set up, so any traced call that depends on system contract state can produce incorrect results on post-Cancun blocks.


mattsse (approved):
this also makes sense here actually

PR #22578, opened 2026-02-25
fix(rpc): add missing `apply_pre_execution_changes` in `call_many`

merge-script:
`call_many` replays block transactions from parent state when `replay_block_txs` is true (user specified a `transaction_index`), but was missing the `apply_pre_execution_changes` call before the replay loop. This is the same class of bug as the `debug_traceCallMany` fix — the `replay_block_txs = false` path is correct because it loads state at the block hash which already includes pre-execution changes, but the partial replay path starts from parent state without setting up EIP-4788/EIP-2935 system contract state first.


stevencartavia (commented):
just needs rebase or to update the branch to pass CI

stevencartavia (approved):

merge-script:
> just needs rebase or to update the branch to pass CI

should be good now

mattsse (approved):

PR #22579, opened 2026-02-25
feat(cli): add db stage-checkpoints command

Rjected:
`reth` currently has tooling to inspect and set prune checkpoints, but not stage checkpoints, which makes checkpoint recovery / debug flows harder. This change adds a CLI surface under `reth db stage-checkpoints` with `get`/`set`, including support for setting an arbitrary block number and optionally clearing stage-unit payload or stage progress bytes when needed for recovery (`--clear-stage-unit`, `--clear-progress`).

Example UX:
```
reth db stage-checkpoints set --stage merkle-execute --block-number 18500000 --clear-progress
````

### CODE REVIEW

File: crates/cli/commands/src/db/stage_checkpoints.rs [hunk 1]
```diff
@@ -0,0 +1,314 @@
...
+            checkpoint.stage_checkpoint = None;
+        }
+
+        provider_rw.save_stage_checkpoint(stage_id, checkpoint)?;
+
+        if args.clear_progress {
+            provider_rw
+                .tx_ref()
+                .delete::<tables::StageCheckpointProgresses>(stage_id.to_string(), None)?;
+        }
```
joshieDo: dont we always want this ? when dont we?

mattsse (approved):

joshieDo (approved):

crates/cli/commands/src/db/stage_checkpoints.rs
Rjected: actually yeah i see no reason why we would want to reset this, will remove this feature

PR #22582, opened 2026-02-25
fix(net): validate EIP-1459 hash labels before caching entries

sashass1315:
## Why
DNS entry responses were parsed and cached without verifying that TXT content matched the requested hash label.
This could allow a hash-mismatch response to be accepted and potentially poison the shared entry cache.

## What changed
- Added explicit `LookupError::HashMismatch`.
- Added hash/content verification in query resolution (`base32 hash decode + keccak256 prefix check`) before parsing DNS entries.
- Added a regression test proving that a hash-mismatch response is rejected and not cached, and that a later valid response for the same hash can still be cached.


mattsse (approved):
Looks good to merge. The DNS entry hash validation now matches EIP-1459 semantics, rejects invalid hash labels, and has regression coverage for avoiding cache poisoning.

PR #22583, opened 2026-02-25
fix: default finalized and safe block to genesis on fresh node

decofe:
## Summary
On a fresh node, `finalized_block_num_hash()` and `safe_block_num_hash()` return `None` because the `ChainState` table has no entries yet. This defaults both to the genesis header (block 0) when no prior FCU has been received.

## Changes
- Fall back to `sealed_header(0)` for finalized and safe headers in `BlockchainProvider::with_latest` when the DB has no stored value.

## Testing
`cargo test -p reth-provider` — 103 tests pass.

Prompted by: joshie

### DIFF
--- a/crates/storage/provider/src/providers/blockchain_provider.rs
+++ b/crates/storage/provider/src/providers/blockchain_provider.rs
@@ -90,7 +90,8 @@ impl<N: ProviderNodeTypes> BlockchainProvider<N> {
             .last_finalized_block_number()?
             .map(|num| provider.sealed_header(num))
             .transpose()?
-            .flatten();
+            .flatten()
+            .or(provider.sealed_header(0)?);
         let safe_header = provider
             .last_safe_block_number()?
             .or_else(|| {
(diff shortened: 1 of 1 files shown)


Rjected (approved):

crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -90,7 +90,8 @@ impl<N: ProviderNodeTypes> BlockchainProvider<N> {
             .last_finalized_block_number()?
             .map(|num| provider.sealed_header(num))
             .transpose()?
-            .flatten();
+            .flatten()
+            .or(provider.sealed_header(0)?);
```
mattsse: this should use or else

crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -100,7 +101,8 @@ impl<N: ProviderNodeTypes> BlockchainProvider<N> {
             })
             .map(|num| provider.sealed_header(num))
             .transpose()?
-            .flatten();
+            .flatten()
+            .or(provider.sealed_header(0)?);
```
mattsse: same

mattsse (changes requested):
can we add some docs for this

isn't this the same as using the genesis hash?

PR #22592, opened 2026-02-26
fix(static-file): handle non-zero genesis blocks in static file range calculations

Vui-Chee:
## Summary
When running reth node on snapshot data that begins with custom genesis number, the node panics on static file initialization as it expects static files of certain names but they don't exist. So this PR tries to account static file naming that do not begin with the expected numeric multiple.

- Fix `set_expected_block_start` to preserve the existing end boundary instead of recomputing it via `find_fixed_range`, which produces incorrect end blocks for adjusted (non-zero genesis) ranges
- Adjust `update_index` to clamp the fixed range start to the genesis block number when the canonical range starts below genesis
- Fix static file writer `open()` to use the actual block number for segment lookup and create new files with genesis-aware ranges

## Motivation
When the genesis block is non-zero (common in L2 rollups), the canonical file range computed by `find_fixed_range` starts at 0, but the actual first static file starts at the genesis block number. This mismatch causes:
1. `set_expected_block_start` to recompute an incorrect end boundary from a wrong `blocks_per_file` value
2. `update_index` to attempt loading a file with a range that doesn't exist on disk
3. The static file writer to look up the wrong segment and create files with incorrect ranges

## Changes
| File | Change |
|------|--------|
| `segment.rs` | Preserve existing end boundary in `set_expected_block_start` |
| `manager.rs` | Clamp `fixed_range` start to genesis in `update_index` |
| `writer.rs` | Use actual block for segment lookup; genesis-aware range for new files |


mattsse (commented):
wdyt @joshieDo @yongkangc

Vui-Chee:
@mattsse Will this ever be reviewed?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -303,17 +303,22 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
...
-        let (jar, path) = match static_file_provider.get_segment_provider_for_block(
-            segment,
-            block_range.start(),
-            None,
-        ) {
+        let genesis = static_file_provider.genesis_block_number();
+        let canonical_block_range = static_file_provider.find_fixed_range(segment, block);
+
+        let (jar, path) = match static_file_provider
+            .get_segment_provider_for_block(segment, block, None)
```
mattsse, line 309: why was this changed to `block`?

mattsse: why is this needed here?

dont see how `canonical_block_range` is required here

crates/storage/provider/src/providers/static_file/manager.rs
```diff
@@ -1056,12 +1056,18 @@ impl<N: NodePrimitives> StaticFileProvider<N> {
...
         match segment_max_block {
             Some(segment_max_block) => {
-                let fixed_range = self.find_fixed_range_with_block_index(
+                let mut fixed_range = self.find_fixed_range_with_block_index(
                     segment,
                     indexes.get(segment).map(|index| &index.expected_block_ranges_by_max_block),
                     segment_max_block,
                 );
 
+                let genesis = self.genesis_block_number();
```
mattsse, line 1066: lets please add one line docs why this exists

crates/static-file/types/src/segment.rs
```diff
@@ -421,18 +421,19 @@ impl SegmentHeader {
...
+    /// derived `blocks_per_file` from an adjusted range differs from the canonical one.
+    ///
+    /// For example, if `blocks_per_file` is 500000 and genesis is at block 17000000,
+    /// the canonical range is 17000000..=17499999. Adjusting the start preserves that
+    /// end boundary.
     pub const fn set_expected_block_start(&mut self, block: BlockNumber) {
-        let blocks_per_file =
-            self.expected_block_range.end() - self.expected_block_range.start() + 1;
-        let file_range = find_fixed_range(block, blocks_per_file);
-        self.expected_block_range = SegmentRangeInclusive::new(block, file_range.end());
```
mattsse, line 435: why can we simplify this here?

mattsse (changes requested):

crates/storage/provider/src/providers/static_file/writer.rs
Vui-Chee: Because find_fixed_range computes a canonical start (always a multiple of blocks_per_file, e.g. 0, 500000, …). When genesis is non-zero, that canonical start doesn't correspond to any file on disk — the actual file starts at the genesis block. Passing block directly lets get_segment_provider_for_block find the file that actually contains it, instead of looking up a non-existent one keyed by the canonical range start.

Vui-Chee: Good point, reverting in 47cee25.

crates/storage/provider/src/providers/static_file/manager.rs
Vui-Chee: Done in 0fbfb02.

crates/storage/provider/src/providers/static_file/writer.rs
Vui-Chee: For a standard chain (genesis=0), `block_range.start()` and `block` always resolve to the same file, because find_fixed_range produces a canonical start that aligns with the actual file boundaries. No behavior change.

crates/static-file/types/src/segment.rs
Vui-Chee: The old code derived blocks_per_file from the current range width, but for genesis-adjusted ranges that width is shorter than canonical (e.g. 499,999 instead of 500,000), so find_fixed_range recomputes a wrong end. Since the end boundary is already correct from file creation, we just preserve it directly.

Vui-Chee: The old code assumes `blocks_per_file` are multiples of some number, but if we have custom genesis number that is not aligned, `find_fixed_range` will produce a range with the wrong end.

Here is an example to illustrate this.
```
genesis = 17_000_001, canonical blocks_per_file = 500_000
File on disk: 17_000_001..=17_499_999  (created by writer.rs with genesis clamp)
 
Old code derives: blocks_per_file = 17_499_999 - 17_000_001 + 1 = 499_999  ← wrong!
find_fixed_range(17_000_001, 499_999):
  start = (17_000_001 / 499_999) * 499_999 = 16_999_966
  end   = 16_999_966 + 499_999 - 1 = 17_499_964
 
New range: 17_000_001..=17_499_964  ← WRONG end! Should be 17_499_999
```

PR #22594, opened 2026-02-26
chore(ci): fix `collapsible_match` clippy lint in chainspec

merge-script:
Recent nightly clippy started flagging the `if` inside match arms in `spec.rs` as `collapsible_match`. Collapse them into match guards to satisfy the lint.

couples of PR have the same issue


mattsse (approved):

mattsse:
can you please make sure that all are fixed and not do them one by one

PR #22604, opened 2026-02-26
refactor(engine): PayloadProcessor::spawn_state_root

shekhirin:
Allows to use state root task in payload building without prewarming and caching.


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -725,20 +745,79 @@ fn convert_serial<RawTx, Tx, TxEnv, InnerTx, Recovered, Err, C>(
     }
 }
 
+/// Lightweight handle to a background state root computation task.
```
shekhirin:
```suggestion
/// Handle to a background state root computation task.
```

klkvr (approved):

PR #22605, opened 2026-02-26
feat: implement Bal Client for eth 71

Soubhik-10:


crates/net/p2p/src/block_access_lists/downloader.rs
mattsse, line 1: I dont think we actually need this type here, having the client should be sufficient

crates/net/network/src/fetch/mod.rs
```diff
@@ -535,7 +554,9 @@ impl<N: NetworkPrimitives> DownloadRequest<N> {
     /// Returns the best peer requirements for this request.
     fn best_peer_requirements(&self) -> BestPeerRequirements {
```
mattsse, line 643: we now have a case here where itÄs possible that no connected peer supports eth71, so we need to check for this when we try to find a peer, if we dont find a peer we need to park the request and then revive it once we get a new peer

ideally we also get some tests for this behaviour

mattsse (changes requested):
great

emmajam:
do we want base to be bal-devnet-3 ? or is main ok?

crates/net/network/src/fetch/mod.rs
```diff
@@ -62,6 +65,10 @@ pub struct StateFetcher<N: NetworkPrimitives = EthNetworkPrimitives> {
     download_requests_rx: UnboundedReceiverStream<DownloadRequest<N>>,
     /// Sender for download requests, used to detach a [`FetchClient`]
     download_requests_tx: UnboundedSender<DownloadRequest<N>>,
+    /// Waker to notify when a new peer might satisfy queued requests.
+    ///
+    /// Stored so that `new_active_peer` can wake the fetcher when a peer is added.
+    waker: Option<std::task::Waker>,
```
mattsse: we shouldnt need this, lets remove

crates/net/network/src/fetch/mod.rs
```diff
@@ -151,21 +164,30 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
...
+        });
 
         let mut best_peer = idle.next()?;
 
         for maybe_better in idle {
             // replace best peer if our current best peer sent us a bad response last time
             if best_peer.1.last_response_likely_bad && !maybe_better.1.last_response_likely_bad {
                 best_peer = maybe_better;
-                continue
+                continue;
```
mattsse:
```suggestion
                continue
```

crates/net/network/src/fetch/mod.rs
```diff
@@ -151,21 +164,30 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
...
             }
 
             // replace best peer if this peer meets the requirements better
             if maybe_better.1.is_better(best_peer.1, &requirement) {
                 best_peer = maybe_better;
-                continue
+                continue;
```
mattsse:
```suggestion
                continue
```

crates/net/network/src/fetch/mod.rs
```diff
@@ -104,6 +113,10 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
                 range_info,
             },
         );
+        // waking the fetcher ensures any queued request will be retried
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
```
mattsse: we shouldn't need this

mattsse (changes requested):
if we need the waker for the test then I think we can use the poll_api

mattsse (approved):

mattsse (approved):

Issue #22591, opened 2026-02-26
feat(net): add ReceiptsClient trait and p2p receipt downloading


mattsse:
## Summary

Add a `ReceiptsClient` trait (analogous to [`BodiesClient`](https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/net/p2p/src/bodies/client.rs#L17-L49)) for downloading receipts from peers over the eth protocol. The wire types (`GetReceipts`, `Receipts`, `Receipts69`, `GetReceipts70`, `Receipts70`) already exist but there is no client-side trait or fetcher integration to actually request them.

### 1. Define the `ReceiptsClient` trait

Create `crates/net/p2p/src/receipts/` (or add to `crates/net/p2p/src/bodies/` sibling) with a trait modeled after [`BodiesClient`](https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/net/p2p/src/bodies/client.rs#L17-L49):

```rust
#[auto_impl(&, Arc, Box)]
pub trait ReceiptsClient: DownloadClient {
    type Receipt: TxReceipt;
    type Output: Future<Output = PeerRequestResult<Vec<Vec<Self::Receipt>>>> + Send + Sync + Unpin;

    fn get_receipts(&self, hashes: Vec<B256>) -> Self::Output;
    fn get_receipts_with_priority(&self, hashes: Vec<B256>, priority: Priority) -> Self::Output;
}
```

### 2. Add a `GetReceipts` variant to `DownloadRequest`

Extend the [`DownloadRequest` enum](https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/net/network/src/fetch/mod.rs#L494-L508) with a new variant:

```rust
GetReceipts {
    request: Vec<B256>,
    response: oneshot::Sender<PeerRequestResult<Vec<Vec<N::Receipt>>>>,
    priority: Priority,
}
```

### 3. Implement `ReceiptsClient` for `FetchClient`

Follow the pattern in [`FetchClient`'s `BodiesClient` impl](https://github.com/paradigmxyz/reth/blob/4cec99ed1360b0908ddefa01cd211fe97adb836a/crates/net/network/src/fetch/client.rs#L81-L103) — send a `DownloadRequest::GetReceipts` over the channel.

### 4. Handle the request in `StateFetcher`

[...]

PR #22607, opened 2026-02-26
feat(net): add ReceiptsClient trait and p2p receipt downloading

figtracer:
adds #22591 `ReceiptsClient` extra utility trait for snap sync

### DIFF
--- a/crates/net/network/src/fetch/mod.rs
+++ b/crates/net/network/src/fetch/mod.rs
@@ -1196,104 +1196,97 @@ mod tests {
         );
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_poll_receipts_request_no_peers() {
+    /// Creates a `StateFetcher` with a single idle peer and returns both.
+    fn fetcher_with_peer() -> (StateFetcher<EthNetworkPrimitives>, PeerId) {
         let manager = PeersManager::new(PeersConfig::default());
         let mut fetcher =
             StateFetcher::<EthNetworkPrimitives>::new(manager.handle(), Default::default());
+        let peer_id = B512::random();
+
+        fetcher.new_active_peer(
+            peer_id,
+            Default::default(),
+            Default::default(),
+            Arc::new(Capabilities::from(vec![])),
+            Default::default(),
+            None,
+        );
+        (fetcher, peer_id)
+    }
+
+    /// Inserts an inflight receipts request into the fetcher and returns the
+    /// `oneshot::Receiver` that the final response will be sent through.
+    fn insert_inflight_receipts(
+        fetcher: &mut StateFetcher<EthNetworkPrimitives>,
+        peer_id: PeerId,
+    ) -> oneshot::Receiver<PeerRequestResult<ReceiptsResponse<reth_ethereum_primitives::Receipt>>>
+    {
+        let (tx, rx) = oneshot::channel();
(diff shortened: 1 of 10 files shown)


crates/net/network/src/fetch/mod.rs
```diff
@@ -32,6 +34,25 @@ use tokio_stream::wrappers::UnboundedReceiverStream;
...
+    Request<InflightReceiptsState<R>, PeerRequestResult<ReceiptsResponse<R>>>;
+
+/// Maximum number of eth/70 continuation rounds allowed per receipts request.
+/// Prevents a misbehaving peer from keeping us in an infinite loop.
+const MAX_RECEIPT_CONTINUATIONS: u8 = 16;
+
+/// Tracks the state of an in-flight receipts request, including accumulated
+/// partial results across eth/70 continuation round-trips.
+#[derive(Debug)]
+struct InflightReceiptsState<R> {
```
mattsse: can we move this helper type to end of file, because prrivate helepr

crates/net/network/src/fetch/mod.rs
```diff
@@ -130,6 +157,19 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
         false
     }
 
+    /// Returns the `first_block_receipt_index` to use for an eth/70 `GetReceipts70` request
+    /// to the given peer.
+    ///
+    /// When a previous response was truncated (`last_block_incomplete`), the accumulated
+    /// receipts for the last block are stored in the inflight state and this returns
+    /// their count so the follow-up request skips already-received receipts.
+    pub(crate) fn receipt_continuation_index(&self, peer_id: &PeerId) -> u64 {
```
mattsse: I find this a bit confusing, why do we go through the peerid here when this a requests specific value?

crates/net/network/src/fetch/mod.rs
```diff
@@ -330,6 +380,109 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
...
+    /// Called on a `GetReceipts` response from a peer.
+    ///
+    /// All receipt variants (legacy with bloom, eth/69, eth/70) are expected to be normalized
+    /// to [`ReceiptsResponse`] by the caller before invoking this method.
+    ///
+    /// When an eth/70 peer signals `last_block_incomplete`, this automatically issues a
+    /// follow-up `GetReceipts70` request to the same peer with the appropriate
+    /// `first_block_receipt_index` and merges the partial results. The waiting request future
+    /// is fulfilled only once all continuations complete.
+    pub(crate) fn on_receipts_response(
```
mattsse, line 351: maybe there's an argument for not doing any of this here and instead do this on the client level?

for example the bodiesclient does the same and doesnt handle complete responses in the fetcher. so imo we should simplify this for now and move the complexity to the client/callsite

mattsse (changes requested):
cool, this looks great already

crates/net/network/src/fetch/mod.rs
figtracer: i agree, was a bit confused on where to put this

crates/net/network/src/fetch/mod.rs
figtracer: i assumed this should be fetched against the same peer every time, when in reality there's no need for that

mattsse (approved):

PR #22608, opened 2026-02-26
perf: don't prune extension nodes

klkvr:


klkvr:
derek bench warmup=500 blocks=2000

PR #22609, opened 2026-02-26
chore(txpool): use ValidPoolTransaction methods instead of reaching i…

esorense:
`ValidPoolTransaction` already exposes `is_eip4844()` and `hash()` as delegate methods, but `maintain.rs` bypasses them by accessing `.transaction` directly. This is the same pattern fixed in #22475 for other call sites.

Also adds the missing `max_priority_fee_per_gas()` delegate method to `ValidPoolTransaction` for API symmetry with the existing `max_fee_per_gas()`.


mattsse (approved):
fine

PR #22610, opened 2026-02-26
fix(rpc-provider): state_root delegates to stub that always returns zero

MagicJoshh:
`state_root` was forwarding to `state_root_from_nodes` which unconditionally returns `B256::ZERO`, even when `compute_state_root` is enabled. The working implementation lives in `state_root_with_updates` — just delegate there and discard the trie updates.


mattsse (approved):
LGTM — fixes state_root() routing through the no-op state_root_from_nodes() instead of the actual RPC implementation.

PR #22612, opened 2026-02-26
refactor(engine): move lazy overlay to canonical in-memory state

shekhirin:


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1460,12 +1464,21 @@ where
...
     ///
-    /// Uses a cached overlay if available for the canonical head (the common case).
+    /// Delegates to [`CanonicalInMemoryState::lazy_overlay`] for canonical blocks (which
+    /// also checks the cached overlay), with a fallback to `TreeState` for non-canonical
+    /// fork blocks.
     fn get_parent_lazy_overlay(
         parent_hash: B256,
         state: &EngineApiTreeState<N>,
+        canonical_in_memory_state: &CanonicalInMemoryState<N>,
     ) -> (Option<LazyOverlay>, B256) {
```
mattsse, line 1474: now this fn looks even weirder, can we make this a fn of ctx perhaps?

crates/chain-state/src/in_memory.rs
```diff
@@ -180,6 +186,32 @@ pub struct CanonicalInMemoryState<N: NodePrimitives = EthPrimitives> {
...
+/// the next payload (which typically builds on the canonical head) to reuse
+/// the pre-computed overlay immediately without re-traversing in-memory blocks.
+///
+/// # Invalidation
+///
+/// The cached overlay is invalidated when:
+/// - Persistence completes (anchor changes)
+/// - The canonical head changes to a different block
+#[derive(Debug, Clone)]
+pub struct PreparedCanonicalOverlay {
```
mattsse, line 201: this is mixing type order again

crates/chain-state/src/in_memory.rs
```diff
@@ -395,6 +427,57 @@ impl<N: NodePrimitives> CanonicalInMemoryState<N> {
...
+    /// Returns a [`LazyOverlay`] for the given parent hash.
+    ///
+    /// Checks the cached canonical overlay first, then constructs a new one
+    /// by walking the in-memory chain and collecting [`DeferredTrieData`] handles.
+    ///
+    /// Returns `(None, parent_hash)` if parent is on disk.
+    pub fn lazy_overlay(&self, parent_hash: B256) -> (Option<LazyOverlay>, B256) {
+        let Some(state) = self.state_by_hash(parent_hash) else {
+            return (None, parent_hash);
+        };
```
mattsse, line 439: this isnt entirely euqivalent because this now only works if the parent hash is canonical, which I guess we can assume, but maybe this affects execution of downloaded blocks, case when we buffer more blocks before making them canonical

mattsse (commented):
in general this makes sense

but I'm unsure if this is is the best way to prep access to this, we likely want to give access to more things, and i dont think moving the cache in here would make sense for example so perhaps we introduce some additional container type for this.

PR #22613, opened 2026-02-26
fix(engine): continue downloading head block after making non-head sync target canonical

jsvisa:
Fix the hive `engine-withdrawals` tests, ref: https://hive.ethpandaops.io/#/test/generic/1772094552-fb259e9348320128c32132a6c70dab2c?testnumber=159

> FAIL (Sync after 2 blocks - Withdrawals on Block 2 - Multiple Withdrawal Accounts): Timeout while waiting for secondary client to sync
CLMocker: Closing engine client fbcc082e
CLMocker: Closed engine client fbcc082e
CLMocker: Closing engine client 2247e968
CLMocker: Closed engine client 2247e968
End test (reth_default): Sync after 2 blocks - Withdrawals on Block 2 - Multiple Withdrawal Accounts (Paris)

When the engine receives its first `forkchoiceUpdated` with an unknown head, `handle_missing_block` downloads the safe block first. 
If the safe block is close to the canonical tip, it gets inserted as `Valid` and the code returned `MakeCanonical` early — never downloading the actual head block. 
This caused the secondary client to stall and never complete sync.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2583,6 +2583,51 @@ where
         Some(TreeEvent::Download(request))
     }
 
+    /// Handles a downloaded block that was successfully inserted as valid.
+    ///
+    /// If the block matches the sync target head, returns [`TreeAction::MakeCanonical`].
+    /// If it matches a non-head sync target (safe or finalized), makes it canonical inline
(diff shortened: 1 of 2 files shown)


mattsse (approved):
oh I see, this makes perfect sense, tysm!

PR #22615, opened 2026-02-26
perf: use more multiproof workers

klkvr:
doubles number of threads allocated for proof workers and removes unused `TreeConfig` fields


klkvr:
derek bench warmup=500 blocks=2000

shekhirin (approved):

PR #22617, opened 2026-02-26
fix(rpc): use block executor for transaction replay in call helpers

stevencartavia:
Uses the block executor for replaying transactions in `call_many` and `spawn_replay_transaction` instead of manually creating separate EVMs and committing state.

This also fixes `call_many` not applying pre-execution changes (EIP-4788 beacon root, EIP-2935 blockhashes) when replaying block transactions.


mattsse (approved):

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -705,68 +705,6 @@ pub trait Call:
...
-    /// Retrieves the transaction if it exists and executes it.
-    ///
-    /// Before the transaction is executed, all previous transaction in the block are applied to the
-    /// state by executing them first.
-    /// The callback `f` is invoked with the [`ResultAndState`] after the transaction was executed
-    /// and the database that points to the beginning of the transaction.
-    ///
-    /// Note: Implementers should use a threadpool where blocking is allowed, such as
-    /// [`BlockingTaskPool`](reth_tasks::pool::BlockingTaskPool).
-    fn spawn_replay_transaction<F, R>(
```
mattsse, line 717: doesnt hurt if we keep this

mattsse (changes requested):

mattsse (approved):

PR #22623, opened 2026-02-27
perf(trie): share HashedStorages cursor across storage root calculations

### LINKED ISSUE #18032: Share a `HashedStorages` cursor among storage root calculations

Vui-Chee:
## Summary

Optimize state root calculation by sharing a single `HashedStorages` cursor across all storage root calculations instead of creating a new cursor for each account.

## Changes

- Add `calculate_with_cursor` method to `StorageRoot` that accepts and returns a cursor for reuse
- Update `StateRoot` to create a shared storage cursor on first use and reuse it for all subsequent storage root calculations
- Cursor is lazily initialized to avoid issues with mock cursors in tests
- Cursor is repositioned via `set_hashed_address` before each use

All existing tests pass with no regressions.

## Benchmark Results

Ran a criterion benchmark (`crates/trie/db/benches/shared_storage_cursor.rs`) comparing per-account cursor creation (baseline) vs shared cursor reuse, using real MDBX database cursors:

| Configuration | new_cursor_per_account | shared_cursor | Improvement |
|---|---|---|---|
| 100 accounts, 10 slots | 506.00 µs | 490.36 µs | ~3.1% |
| 500 accounts, 10 slots | 2.43 ms | 2.37 ms | ~2.4% |
| 1000 accounts, 10 slots | 4.87 ms | 4.77 ms | ~2.1% |
| 100 accounts, 100 slots | 4.00 ms | 3.96 ms | ~1.0% |
| 500 accounts, 100 slots | 20.44 ms | 20.31 ms | ~0.6% |

The improvement is larger when there are fewer storage slots per account (cursor creation is a bigger fraction of total work). With many slots, trie computation dominates.

These numbers are conservative for a local test environment with small temporary databases. The original profiling in #18032 showed ~5.7% overhead from cursor creation against production-scale Ethereum state, where MDBX cursor initialization is more expensive due to larger B-tree depths and page counts.

Fixes #18032

---

_Reopening of #22422, which was closed during PR backlog cleanup._


crates/trie/trie/src/trie.rs
```diff
@@ -615,16 +633,42 @@ where
...
+    /// reducing overhead when computing storage roots for many accounts.
+    ///
+    /// # Returns
+    ///
+    /// A tuple of (storage root progress, cursor) where the cursor can be reused for
+    /// subsequent storage root calculations.
+    #[instrument(skip_all, level = "trace", target = "trie::storage_root", name = "storage_trie_with_cursor", fields(addr = %self.hashed_address, storage_root))]
+    pub fn calculate_with_cursor(
+        self,
+        mut hashed_storage_cursor: H::StorageCursor<'_>,
```
mediocregopher: Would it not be possible to pass in a &mut here, rather than taking and returning the cursor?

Vui-Chee: Updated `calculate_with_cursor` to take `&mut H::StorageCursor<'_>` and return just `Result<StorageRootProgress, StorageRootError>`.

mediocregopher (approved):

crates/trie/trie/src/trie.rs
```diff
@@ -615,8 +631,27 @@ where
     pub fn calculate(self, retain_updates: bool) -> Result<StorageRootProgress, StorageRootError> {
         trace!(target: "trie::storage_root", "calculating storage root");
 
-        let mut hashed_storage_cursor =
-            self.hashed_cursor_factory.hashed_storage_cursor(self.hashed_address)?;
+        let factory_clone = self.hashed_cursor_factory.clone();
```
mediocregopher, line 634: Why is this clone necessary?

Vui-Chee: @mediocregopher  The clone is needed because calculate takes self by value — the cursor borrows the factory (part of self), so we can't then move self into calculate_with_cursor. The clone breaks this borrow-move conflict.

DatabaseHashedCursorFactory wraps a &TX reference, so cloning it is just a pointer copy. Even previous authors have adopted the same method.

mediocregopher:
Hey @Vui-Chee I was reviewing the PR again and noticed that also the trie cursor could be reused, and additionally wanted to try doing it without Clone so I put together and merged https://github.com/paradigmxyz/reth/pull/25163. It's based on your branch so you still get co-authorship on it. Thanks for the contribution!

Issue #15394, opened 2025-03-30
Additional indexes for block explorers


paulmillr:
### Describe the bug

I've built [esplr - private block explorer](https://github.com/paulmillr/esplr). It's getting more popular, but still can't be ran on reth.

While #14925 enabled proper trace_filter limits, apparently there are some indexes present in Erigon which Reth doesn't have. It makes Reth less useful for esplr. Initial thoughts on how this can be solved:

a. Add additional indexes. Perhaps target next major version if that requires DB change / resync.
b. Implement additional APIs to get list of token balances and token tx histories.

Esplr uses logic from [archive.ts](https://github.com/paulmillr/micro-eth-signer/blob/main/src/net/archive.ts) to implement robust exploring. Specifically it uses:

```js
// tokenTransfers
this.ethLogs(ERC_TRANSFER.topics({ from: address, to: null, value: null }), opts), // From
this.ethLogs(ERC_TRANSFER.topics({ from: null, to: address, value: null }), opts), // To

// erc1155Transfers
this.ethLogs(
  ERC1155_SINGLE.topics({ operator: null, from: address, to: null, id: null, value: null }),
  opts
),
this.ethLogs(
  ERC1155_SINGLE.topics({ operator: null, from: null, to: address, id: null, value: null }),
  opts
),
// Batch
this.ethLogs(
  ERC1155_BATCH.topics({ operator: null, from: address, to: null, ids: null, values: null }),
  opts
),
this.ethLogs(
  ERC1155_BATCH.topics({ operator: null, from: null, to: address, ids: null, values: null }),
  opts
),
```

### Steps to reproduce

Try using esplr (purely client-side web app) with erigon

### Node logs

```text

```

### Container Type

Not running in a container

### What version/commit are you on?

latest

### What database version are you on?

latest

### Which chain / network are you on?

mainnet

### What type of node are you running?

Archive (default)

### Code of Conduct

- [x] I agree to follow the Code of Conduct

paulmillr:
@jenpaff what could be done here? How hard could the task be?

mattsse:
thanks @paulmillr 

definitely planned, since this requires a db change we will combine this with some other changes that we've planned. can't give you an ETA on this yet though.

but bumping feasibility investigation on this.

PR #22626, opened 2026-02-27
feat(stages): add call trace index for trace_filter optimization

stevencartavia:
Towards #15394. Adds `CallTraceFromIndex` and `CallTraceToIndex` tables that map addresses to block numbers where they appeared as transaction sender or recipient. `trace_filter` now queries these indexes when `fromAddress/toAddress` filters are specified, executing only the candidate blocks instead of every block in the range. Falls back to the existing full-scan behavior when no address filters are given or the index isn't populated yet.
Limitation until Step 2: `trace_filter` results for contract addresses may be incomplete — blocks where an address is only reached via internal calls will be silently skipped. EOA addresses are unaffected. Step 2 closes this gap.

Next Steps:
Step 2 — Internal call coverage
Instrument the execution stage to emit call frame from/to pairs as a side-effect of the EVM run. Those pairs write into the same `CallTraceFromIndex`/`CallTraceToIndex` tables, making fromAddress/toAddress filters on contracts fully correct. Kept separate because it touches the execution stage.

Step 3 — Token transfer history (eth_getLogs)
Add a log address/topic index built from receipts, same sharded `BlockNumberList` pattern as this PR. Unlocks efficient ERC-20/721/1155 transfer history queries without touching trace infrastructure.

Step 4 — Current token balances
Using Step 3's log index, find every token contract an address has interacted with, then batch `eth_call` for `balanceOf`. Expose as a dedicated RPC method.

### DIFF
--- a/crates/config/src/config.rs
+++ b/crates/config/src/config.rs
@@ -128,6 +128,8 @@ pub struct StageConfig {
     pub index_account_history: IndexHistoryConfig,
     /// Index Storage History stage configuration.
     pub index_storage_history: IndexHistoryConfig,
+    /// Index Transaction History stage configuration.
+    pub index_transaction_history: IndexHistoryConfig,
     /// Common ETL related configuration.
     pub etl: EtlConfig,
 }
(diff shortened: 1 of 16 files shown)


paulmillr:
Why was this closed without merging? trace_filter is useless in reth for specific cases - compared to erigon.

PR #22628, opened 2026-02-27
fix(engine/tree): continue sync-target progression for already-seen downloaded blocks

jsvisa:
another part of #22613

As the stall could still happen in long-gap (e.g. 128-block) flows where downloaded blocks are often duplicates.
When the same block arrived again and `insert_block` returned `AlreadySeen(Valid)`, the handler did nothing, which could stop sync progression toward the FCU head.

ref https://github.com/paradigmxyz/reth/pull/22625#issuecomment-3970617541


mattsse (approved):
amazing, these all make sense

tysm!

PR #22637, opened 2026-02-27
fix(net/peers): remove duplicate NodeRecordParseError type

aso20455:
Remove duplicate `NodeRecordParseError` enum from `trusted_peer.rs`. The identical type is already defined in `node_record.rs` and re-exported via `lib.rs`.


mattsse (approved):

aso20455:
@mattsse failed CI

DaniPopes (approved):

PR #22642, opened 2026-02-27
feat(cli): add opcode-level tracing for failed txs in re-execute command

decofe:
## Summary
On receipt/gas mismatch during `reth re-execute`, generate opcode-level traces and output them in logs for debugging.

## Motivation
When re-execute hits a gas usage mismatch, the current output only shows the error and receipt diff. To debug execution mismatches you need the full opcode trace — pc, op, gas, stack, memory, storage — which previously required manually setting up `debug_traceTransaction` against a node.

## Changes
- `trace_failed_transaction`: uses the `BlockExecutor` abstraction (`executor_for_block` + `apply_pre_execution_changes` + `execute_transaction`) to replay prior txs in a chain-agnostic way, then re-executes the failing tx with `TracingInspector(TracingInspectorConfig::all())`, logs the geth-style `DefaultFrame` (structLogs) as JSON
- Hooked into receipt gas mismatch path only (where we know the failing tx index)
- Added `revm-inspectors` dep to `reth-cli-commands`

Zero cost on the happy path — tracing only runs on failure.

## Testing
```
cargo check -p reth-cli-commands
cargo clippy -p reth-cli-commands
```

Prompted by: dragan

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -11,14 +11,15 @@ use reth_chainspec::{EthChainSpec, EthereumHardforks, Hardforks};
 use reth_cli::chainspec::ChainSpecParser;
 use reth_cli_util::cancellation::CancellationToken;
 use reth_consensus::FullConsensus;
-use reth_evm::{execute::Executor, ConfigureEvm};
-use reth_primitives_traits::{format_gas_throughput, BlockBody, GotExpected};
+use reth_evm::{execute::Executor, ConfigureEvm, Evm};
+use reth_primitives_traits::{format_gas_throughput, BlockBody, GotExpected, NodePrimitives};
 use reth_provider::{
     BlockNumReader, BlockReader, ChainSpecProvider, DatabaseProviderFactory, ReceiptProvider,
     StaticFileProviderFactory, TransactionVariant,
 };
-use reth_revm::database::StateProviderDatabase;
+use reth_revm::{database::StateProviderDatabase, State};
 use reth_stages::stages::calculate_gas_used_from_headers;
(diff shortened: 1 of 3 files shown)


rakita (approved):
This is helper fn that would dump tx traces on block exec error. It allows easier debugging. It's touching re exec cli and nothing consensus sensitive.

Waiting for another approval from the team before merging.

crates/cli/commands/src/re_execute.rs
```diff
@@ -65,6 +66,164 @@ impl<C: ChainSpecParser> Command<C> {
...
+            if i >= tx_index {
+                break;
+            }
+            let tx_env = evm_config.tx_env(tx);
+            if let Err(err) = evm.transact_commit(tx_env) {
+                error!(index = i, %err, "Failed to replay transaction before failing tx");
+                return;
+            }
+        }
+    }
```
klkvr: this should use `BlockExecutor` abstraction properly, rn this is ethereum-specific

crates/cli/commands/src/re_execute.rs
```diff
@@ -164,6 +323,17 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
...
+                                // level. Re-execute txs one by one to locate the
+                                // failing index, then trace that tx with an inspector.
+                                let parent_num = block.number() - 1;
+                                find_and_trace_failing_tx(
+                                    &evm_config,
+                                    || db_at(parent_num),
+                                    &block,
+                                    provider_factory.chain_spec(),
+                                );
+
```
klkvr: not sure if this is useful, if `execute_one` returned a error it means something went wrong in executor itself or some of the txs was invalid. imo we can remove this and only keep tracing for invalid txs we know the index for

PR #22644, opened 2026-02-27
perf(engine): sender-affinity routing for prewarm transactions

decofe:
Replaces Rayon work-stealing dispatch with dedicated worker threads where transactions from the same sender are routed to the same worker. This improves DB cache locality for sender account state in blocks with heavy same-sender patterns (e.g. CREATE2 batch minters like XEN).

Prompted by: arsenii


klkvr:
derek bench

PR #22647, opened 2026-02-27
perf: share executed tx counter with prewarming

klkvr:


klkvr:
derek bench warmup=500 blocks=2000 cores=16

shekhirin (approved):

PR #22648, opened 2026-02-27
feat(debug): add FallbackBlockProvider for multi-URL RPC failover

smuu:
## Summary

- Add `FallbackBlockProvider` for ordered RPC failover in debug consensus client
- `--debug.rpc-consensus-url` now accepts comma-separated URLs (backward compatible)
- Staleness detection (30s timeout) triggers automatic failover to next URL


mattsse (changes requested):
I like the idea, but this already exist on the rpc transport level so we should make use of that, irrc this is also call Fallbacktransport or smth

PR #22652, opened 2026-02-27
chore: bump op-alloy to 0.24.0

decofe:
Bumps op-alloy crates from 0.23.1 to 0.24.0 and alloy-op-evm from 0.27.2 to 0.27.3. Removes the temporary alloy-evm fork patch since alloy-evm 0.28.0 now ships with op-alloy 0.24 support.

Supersedes #22611

Prompted by: mattsse


mattsse (commented):
bad bot

PR #22654, opened 2026-02-27
fix(payload): avoid dropping rebuild ticks while payload build is in progress

jsvisa:
Fixes a payload-building timing issue that can surface in Cancun multi-client blob scenarios, where `engine_getPayloadV3` may return a stale 5-blob payload instead of a full 6-blob payload. 

Fix the error case in hive:

```
FAIL: Error verifying blob bundle (payload 1/5): expected 6 blob, got 5
CLMocker: Closing engine client 15820082
CLMocker: Closed engine client 15820082
End test (reth_default): Blob Transaction Ordering, Multiple Clients (Cancun)
```

ref https://hive.ethpandaops.io/#/test/generic/1769588207-23b235db01bd3d466c3c4b706913cd79?testnumber=184

Root cause: `BasicPayloadJob::poll` consumed interval ticks even when a build was already in progress, so when a tick fired during an active build:
  - the tick was consumed,
  - no new build was spawned (because `pending_block` existed),
  - and after the active build completed, the next rebuild could be delayed by a full interval.

In multi-client setups, this increases the chance that `getPayload` resolves an older payload before a refresh includes newly propagated blob txs.

### DIFF
--- a/crates/payload/basic/src/lib.rs
+++ b/crates/payload/basic/src/lib.rs
@@ -376,51 +376,56 @@ where
             return Poll::Ready(Ok(()))
         }
 
-        // Check if a scheduled build tick is reached.
-        //
-        // Important: only poll the interval if there is no active build job.
-        // Polling (and consuming) ticks while a build is still pending can delay follow-up builds
-        // by a full interval even if the current attempt has already finished.
-        if this.pending_block.is_none() &&
-            !this.best_payload.is_frozen() &&
-            this.interval.poll_tick(cx).is_ready()
-        {
-            this.spawn_build_job();
-        }
-
-        // poll the pending block
-        if let Some(mut fut) = this.pending_block.take() {
-            match fut.poll_unpin(cx) {
-                Poll::Ready(Ok(outcome)) => match outcome {
-                    BuildOutcome::Better { payload, cached_reads } => {
-                        this.cached_reads = Some(cached_reads);
-                        debug!(target: "payload_builder", value = %payload.fees(), "built better payload");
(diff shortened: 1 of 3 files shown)


crates/payload/basic/src/lib.rs
```diff
@@ -376,13 +376,16 @@ where
...
-                this.spawn_build_job();
-            }
+        // Check if a scheduled build tick is reached.
+        //
+        // Important: only poll the interval if there is no active build job.
+        // Polling (and consuming) ticks while a build is still pending can delay follow-up builds
+        // by a full interval even if the current attempt has already finished.
+        if this.pending_block.is_none() &&
+            !this.best_payload.is_frozen() &&
+            this.interval.poll_tick(cx).is_ready()
```
mattsse: this is slightly problematic here, becuase poll_tick does not register the waker on the ready response. that's why the while loop exists.

however this fix is directionally correct.

I believe we can achieve the correct/expected behaviour if we restructure this to

```
loop {
   ready!(pending_block);

   
       if this.best_payload.is_frozen() {
        return Poll::Pending;
    }

    ready!(this.interval.poll_tick(cx));
    this.spawn_build_job();
}
```

mattsse (changes requested):
nice, this issue makes sense, especially with longer interval times.

left a different suggestion

crates/payload/basic/src/lib.rs
jsvisa: thanks for the detailed explains, applied, PTAL.

mattsse (approved):

PR #22656, opened 2026-02-27
perf: prepare txenvs in parallel

klkvr:
missed this in #22650, let's also see if this actually has any impact


klkvr:
derek bench

PR #22668, opened 2026-02-27
feat(eth-wire): introduce configurable maximum ETH message size acros…

DeshErBojhaa:
# Summary

Makes the maximum ETH protocol message size (default 10 MiB) configurable instead of hardcoded. Operators can set it via a new CLI flag, and custom setups can set it via the network config builder.

#22302 

## Changes

1. CLI: New optional flag --eth-max-message-size BYTES. When set, the network uses this limit; when omitted, the default 10 MiB is used (no behavioral change).

2. Network builder: NetworkConfigBuilder::eth_max_message_size(max_message_size) sets the limit. Replaces the current handshake with an EthHandshake using the given size (same as before; see doc note about custom handshakes).

3. Wire layer: Single source of truth for the default is reth_eth_wire_types::DEFAULT_MAX_MESSAGE_SIZE. EthHandshake now carries a max_message_size field (with new() and max_message_size()). EthRlpxHandshake has an optional max_message_size() (default impl returns the constant). ETH/Snap streams and handshake flow use this value end-to-end.

## Breaking changes
- EthereumEthHandshake::eth_handshake: Signature now includes max_message_size: usize. Callers must pass the desired limit (e.g. from handshake.max_message_size()).

- reth_eth_wire: The MAX_MESSAGE_SIZE re-export was removed. Use reth_eth_wire_types::DEFAULT_MAX_MESSAGE_SIZE (or the same via reth_network::types).

## Testing

New CLI tests for --eth-max-message-size (valid value, zero rejected, parsing).
Existing network/ef-tests unaffected; no change to default behavior when the flag is not set.

Checklist

- [x] cargo +nightly fmt --all
- [x] cargo +nightly clippy --workspace --all-features
- [x] cargo nextest run --workspace (excluding ef-tests if fixtures not downloaded)
- [ ] CLI docs: make update-book-cli if you changed help text or flags


crates/net/eth-wire-types/src/message.rs
```diff
@@ -24,9 +24,9 @@ use alloy_primitives::{
 use alloy_rlp::{length_of_length, Decodable, Encodable, Header};
 use core::fmt::Debug;
 
-/// [`MAX_MESSAGE_SIZE`] is the maximum cap on the size of a protocol message.
+/// [`DEFAULT_MAX_MESSAGE_SIZE`] is the default maximum cap on the size of a protocol message.
 // https://github.com/ethereum/go-ethereum/blob/30602163d5d8321fbc68afdcbbaf2362b2641bde/eth/protocols/eth/protocol.go#L50
-pub const MAX_MESSAGE_SIZE: usize = 10 * 1024 * 1024;
+pub const DEFAULT_MAX_MESSAGE_SIZE: usize = 10 * 1024 * 1024;
```
mattsse: please dont rename constants

crates/net/eth-wire/src/handshake.rs
```diff
@@ -50,9 +55,30 @@ impl<T> UnauthEth for T where
 /// The Ethereum P2P handshake.
 ///
 /// This performs the regular ethereum `eth` rlpx handshake.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
-pub struct EthHandshake;
+pub struct EthHandshake {
```
mattsse: this we dont need to change because for the handshake we dont need this because we only expect the handshake related message sizes si we can safely keep the defaults for this

crates/node/core/src/args/network.rs
```diff
@@ -708,6 +724,16 @@ impl Default for DiscoveryArgs {
...
 
+/// Parse ETH max message size as raw bytes.
+fn parse_eth_max_message_size(s: &str) -> Result<usize, String> {
+    let value = s.parse::<usize>().map_err(|_| format!("Invalid byte size: {s}"))?;
+    if value == 0 {
+        return Err("ETH max message size must be greater than 0".to_string());
+    }
+
+    Ok(value)
+}
```
mattsse: this fn shouldnt be necessary

mattsse (changes requested):
ty, a few suggestions

crates/node/core/src/args/network.rs
DeshErBojhaa: Don't we need to check if the flag value is 0?

crates/net/eth-wire-types/src/message.rs
DeshErBojhaa: Addressed. Reverted back to `MAX_MESSAGE_SIZE` 
[d7ebaad](https://github.com/paradigmxyz/reth/pull/22668/commits/d7ebaadf3f7118dcdb9ed865eb46525c023e9aea)

crates/net/eth-wire/src/handshake.rs
DeshErBojhaa: Addressed here: [bae0884](https://github.com/paradigmxyz/reth/pull/22668/commits/bae08843119358773d45719cece75cc07b422775)

crates/node/core/src/args/network.rs
DeshErBojhaa: You're correct. We can just use `NonZeroUsize`.
Refactored here: [798c25b](https://github.com/paradigmxyz/reth/pull/22668/commits/798c25b39c64d0598e2f242c311f0ed0999d49a1)

mattsse (approved):

PR #22674, opened 2026-02-28
fix(rpc): propagate provider error for best block number

Elaela22soL:
Propagate the actual provider error from best_block_number, instead of masking it as a HeaderNotFound error.


DaniPopes (approved):
@mattsse

mattsse (approved):

PR #22676, opened 2026-02-28
fix(engine): use IndexSet for deterministic block buffer child ordering

MagicJoshh:
`remove_children()` iterates over a `HashSet<BlockHash>` which has non-deterministic order. Since `try_connect_buffered_blocks` processes blocks in the returned order and may call `make_canonical` on sync targets, this can produce different canonicalization results across runs — which is exactly the kind of thing that causes flaky tests.

Fix: replace `HashSet` with `IndexSet` in `parent_to_child` so children are processed in insertion order (FIFO).

Removes the 3 "Multiple New Payloads Extending Canonical Chain" tests from `ignored_tests.yaml`.


crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -168,13 +172,23 @@ impl<B: Block> BlockBuffer<B> {
         while let Some(parent_hash) = remove_parent_children.pop() {
             // get this child blocks children and add them to the remove list.
             if let Some(parent_children) = self.parent_to_child.remove(&parent_hash) {
-                // remove child from buffer
-                for child_hash in &parent_children {
+                // Sort children deterministically: by block number, then by hash.
+                // HashSet iteration order is non-deterministic and can cause flaky
+                // canonicalization when multiple children exist at the same level.
```
mattsse: this doesnt really make sense to order by number and by hash is a bit strange as well

.github/scripts/hive/ignored_tests.yaml
```diff
@@ -31,5 +30,3 @@ engine-api:
   - Invalid Missing Ancestor Syncing ReOrg, Transaction Signature, EmptyTxs=False, CanonicalReOrg=True, Invalid P9 (Paris) (reth)
   - Invalid Missing Ancestor Syncing ReOrg, Transaction Signature, EmptyTxs=False, CanonicalReOrg=False, Invalid P9 (Paris) (reth)
   - Invalid Missing Ancestor ReOrg, StateRoot, EmptyTxs=True, Invalid P10 (Paris) (reth)
-  - Multiple New Payloads Extending Canonical Chain, Wait for Canonical Payload (Paris) (reth)
-  - Multiple New Payloads Extending Canonical Chain, Set Head to First Payload Received (Paris) (reth)
```
mattsse, line 35: unsure what the expection here is, should this then use the order in which it was received?

because with the current ordering this isnt guaranteed?

mattsse (changes requested):
the better question here would be.

what if we use a vec for this instead

```
pub(crate) parent_to_child: HashMap<BlockHash, HashSet<BlockHash>>,
```

then we get the "which block did we receive first" property

.github/scripts/hive/ignored_tests.yaml
MagicJoshh: Replace HashSet with IndexSet in BlockBuffer::parent_to_child so children are processed in insertion order (FIFO), fixing non-deterministic canonicalization in hive tests. And thanks for the adivce from @DaniPopes

MagicJoshh:
> In that case the simpler replacement is IndexMap

Rebase and replace with IndexMap

mattsse (approved):
pending @DaniPopes

DaniPopes (approved):

DaniPopes:
needs ci

crates/engine/tree/src/tree/block_buffer.rs
```diff
@@ -139,7 +140,7 @@ impl<B: Block> BlockBuffer<B> {
     fn remove_from_parent(&mut self, parent_hash: BlockHash, hash: &BlockHash) {
         // remove from parent to child connection, but only for this block parent.
         if let Some(entry) = self.parent_to_child.get_mut(&parent_hash) {
-            entry.remove(hash);
+            entry.shift_remove(hash);
```
DaniPopes: this is o(n), swap_remove is o(1), order shouldnt matter

MagicJoshh: Make sense

PR #22677, opened 2026-02-28
fix(engine): don't skip head download check when finalized block hash is zero

merge-script:
When `on_backfill_sync_finished` is called with `finalized_block_hash == zero` (common during optimistic sync), the early return skips not only the backfill distance check but also the head block download logic. This means the engine never re-downloads the FCU target head after backfill completes, potentially stalling sync until the next FCU arrives from the CL.

Narrow the `finalized_block_hash.is_zero()` guard to only skip the backfill distance check, so the head download path always executes.


mattsse (changes requested):
I see, this makes sense

needs smol rebase because ` fn valid_outcome` already landed

ShisheerBitgo:
@mattsse @merge-script just following up on this.

I've hit this bug in running `op-reth` v2.2.1 on Base Sepolia.                                                                                                                                           
                                                                                                                                                                                                                         
Out node was offline for ~25 hours during a reth storage migration (v1 to v2). When it restarted, the pipeline synced to its existing checkpoint (block `40,835,717`) but `on_backfill_sync_finished` returns early at the `finalized_block_hash.is_zero()` check. The gap between the pipeline tip and the live chain head (~41M+) is never closed. The node receives Engine API payloads from `op-node` but never applies them.

On OP-stack L2s, `finalized_block_hash` is `B256::ZERO` during optimistic sync, which is the normal operating state. The early return on [line 1840](https://github.com/paradigmxyz/reth/blob/v2.2.0/crates/engine/tree/src/tree/mod.rs#L1840) skips all gap-closing logic, so the pipeline is never restarted to bridge the gap and the node is permanently stuck.

I suspect this PR should've help but it hasn't been merged and released. So at this point is the only recovery path to just delete the data volume and re-syncing from scratch or snapshot? Wondering if you have insights into this?

Note: Below is an aside on a workaround I tried to resolve the aforementioned issue. It failed and I did a deep dive. Thought I'd share to see if the reasoning below on why the workaround did not work makes sense:
`--debug.tip` doesn't help either. The `--debug.tip` lookup occurs in `max_block()`, which seems to be called during engine launch, before the pipeline even starts. It starts [here](https://github.com/paradigmxyz/reth/blob/main/crates/node/builder/src/launch/engine.rs#L141), then goes to `max_block` [here](https://github.com/paradigmxyz/reth/blob/main/crates/node/builder/src/launch/common.rs#L891), which calls `lookup_or_fetch_tip` [here](https://github.com/paradigmxyz/reth/blob/02d1776786abc61721ae8876898ad19a702e0070/crates/node/core/src/node_config.rs#L401), which calls `fetch_tip_from_network` [here](https://github.com/paradigmxyz/reth/blob/02d1776786abc61721ae8876898ad19a702e0070/crates/node/core/src/node_config.rs#L461)). The node then gets stuck in an infinite loop in `fetch_tip_from_network` and so it never reaches the pipeline or Engine API processing. It's blocked at startup.

The reason it is stuck is either because the peers my node has no longer have block (maybe they pruned it)? Or maybe fetching blocks from peers via `--debug.tip` is just not supported. In other words, OP-stack L2 blocks aren't propagated via devp2p and are derived from L1 batch data by the `op-node`? If true, then the pipeline would likely stall again trying to fetch the intermediate blocks.

merge-script:
hi @ShisheerBitgo. Rebased, wait for @mattsse

mattsse:
Closing this because the OP Stack issue is covered by #24159 now.

That PR changed backfill target selection so OP Stack targets the FCU head instead of finalized, and it added post-backfill recheck coverage. The only remaining behavior this PR would affect is the non-OP zero-finalized optimistic-sync path, so if we want to change that as well it should be handled in a fresh, focused PR with an explicit regression test.

PR #22679, opened 2026-03-01
perf(rpc): validate reward percentiles before DB calls in `eth_feeHistory`

stevencartavia:
Moves the reward percentile range and monotonicity check before the provider DB calls in `fee_history`, so malformed requests are rejected without touching the database.


stevencartavia:
Also removes a redundant `into_iter().collect()` in `eth_callBundle`

mattsse (approved):

PR #22682, opened 2026-03-01
fix(rpc-engine-api): enforce FCU SYNCING precedence over V3 payload attr

jsvisa:
This PR fixes `engine_forkchoiceUpdatedV3` behavior so forkchoice status is evaluated before Cancun payload-attributes validation in mixed cases (unknown head + invalid attrs).

Fix the hive test `Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True (Cancun)`: ref https://hive.ethpandaops.io/#/test/generic/1769588207-23b235db01bd3d466c3c4b706913cd79?testnumber=184

```
<< (6649d9c4) {"jsonrpc":"2.0","id":24,"error":{"code":-38003,"message":"Invalid payload attributes","data":{"err":"Payload attributes validation error: no parent beacon block root post-cancun"}}}
FAIL (Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True): Unexpected error on EngineForkchoiceUpdatedV3: Invalid payload attributes, expected=<None>
CLMocker: Closing engine client 6649d9c4
CLMocker: Closed engine client 6649d9c4
End test (reth_default): Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True (Cancun)
```

Refer to https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#specification-1:

> Client software MAY initiate a sync process if forkchoiceState.headBlockHash references an unknown payload or a payload that can't be validated because data that are requisite for the validation is missing. The sync process is specified in the [Sync](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#sync) section.

So the client should immediately response with `SYNCING` instead of checking the payload.

### DIFF
--- a/.github/scripts/hive/expected_failures.yaml
+++ b/.github/scripts/hive/expected_failures.yaml
@@ -20,9 +20,7 @@ engine-withdrawals: [ ]
 
 engine-api: [ ]
 
-# no fix due to https://github.com/paradigmxyz/reth/issues/8732
-engine-cancun:
-  - Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True (Cancun) (reth)
+engine-cancun: [ ]
 
 sync: [ ]
 
--- a/crates/rpc/rpc-engine-api/src/engine_api.rs
+++ b/crates/rpc/rpc-engine-api/src/engine_api.rs
@@ -1339,7 +1339,9 @@ struct EngineApiInner<Provider, PayloadT: PayloadTypes, Pool, Validator, ChainSp
 mod tests {
     use super::*;
     use alloy_primitives::{Address, B256};
-    use alloy_rpc_types_engine::{ClientCode, ClientVersionV1, PayloadAttributes};
+    use alloy_rpc_types_engine::{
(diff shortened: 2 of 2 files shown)


mattsse (commented):
This introduces a double `fork_choice_updated` call on the happy path (synced node + valid attrs) — first without attrs to check status, then again with attrs. That's the hot path for every block build.

The bug only manifests when attrs validation *fails* while the node is syncing. The simpler fix keeps the old structure and just adds the SYNCING check to the existing error branch:

```rust
if let Err(err) = attr_validation_res {
    let fcu_res =
        self.inner.beacon_consensus.fork_choice_updated(state, None, version).await?;
    if fcu_res.is_invalid() || fcu_res.payload_status.is_syncing() {
        return Ok(fcu_res)
    }
    return Err(err.into())
}
```

This avoids the double FCU on the normal path while fixing the hive test — SYNCING now takes precedence over invalid attrs.

DaniPopes (approved):

DaniPopes (approved):
pending @mattsse

mattsse (approved):
tysm!

PR #22687, opened 2026-03-01
fix(net): avoid full HashSet clone in bootstrap

GarmashAlex:
`DnsDiscoveryService::bootstrap()` cloned the entire `bootstrap_dns_networks` HashSet before iterating. This introduced unnecessary allocation and extra work, especially with larger bootstrap sets, while only per-item ownership was required.

Updated `bootstrap()` to iterate with `self.bootstrap_dns_networks.iter().cloned()` and enqueue root resolution directly via a scoped mutable borrow of `self.queries`.


mattsse (commented):
we can keep this as is

PR #22709, opened 2026-03-02
fix(trie): terminate depth-first iterator on database error

MagicJoshh:
Set `self.complete = true` before returning a database error from `DepthFirstTrieIterator::next()`, so the iterator stops instead of retrying the same failing cursor call indefinitely.


DaniPopes (approved):
please add a test or extend an existing one to test this. also implement FusedIterator for the iterator

MagicJoshh:
> please add a test or extend an existing one to test this. also implement FusedIterator for the iterator

Sure thanks for the review.

MagicJoshh:
rebased @DaniPopes

DaniPopes (approved):

PR #22710, opened 2026-03-02
fix(ipc): break out of service loop when response stream is closed

MagicJoshh:
When the sender side of the response channel is dropped, `rx_item.next()` returns `None` immediately on every poll, causing the select loop to busy-spin. Break out of the loop on `None` instead of silently ignoring it.


DaniPopes (approved):
something else would make it break but sure this is fine

PR #22711, opened 2026-03-02
perf(engine): run tx-iterator sequential and parallel conversion concurrently

decofe:
## Summary

Start parallel signature recovery on rayon immediately instead of waiting for the sequential prefetch to finish first.

## Changes

- Spawn parallel `for_each_ordered_in` on a separate thread concurrently with the sequential prefetch
- Use a `Barrier` so the parallel consumer blocks on yielding until the sequential path finishes
- Parallel rayon `.map()` (sig recovery) runs during the sequential prefetch, so results are ready to yield immediately after

Before: sequential prefetch (4 txs) → parallel computation + yield
After: sequential prefetch + parallel computation overlap → parallel yield

Co-Authored-By: DaniPopes <57450786+DaniPopes@users.noreply.github.com>

Prompted by: danipopes


DaniPopes:
derek bench

PR #22714, opened 2026-03-02
feat(provider): add storage_range for paginated storage enumeration

stevencartavia:
Adds `StateProvider::storage_range(address, key_start, limit) -> Vec<(B256, StorageEntry)>` to support efficient paginated storage enumeration, and wires it into `debug_storageRangeAt`.

The historical impl scans `StoragesHistory` to discover slots, then bounds expensive `storage_by_lookup_key` calls to the `limit` candidates. The latest impl uses a cutoff trick on a BTreeMap to cap memory.

Ref #22655

### DIFF
--- a/crates/storage/provider/src/providers/state/historical.rs
+++ b/crates/storage/provider/src/providers/state/historical.rs
@@ -6,12 +6,13 @@ use alloy_eips::merge::EPOCH_SLOTS;
 use alloy_primitives::{Address, BlockNumber, Bytes, StorageKey, StorageValue, B256};
 use reth_db_api::{
     cursor::{DbCursorRO, DbDupCursorRO},
+    models::storage_sharded_key::StorageShardedKey,
     table::Table,
     tables,
     transaction::DbTx,
     BlockNumberList,
 };
-use reth_primitives_traits::{Account, Bytecode};
+use reth_primitives_traits::{Account, Bytecode, StorageEntry};
 use reth_storage_api::{
     BlockNumReader, BytecodeReader, DBProvider, NodePrimitivesProvider, StateProofProvider,
     StorageChangeSetReader, StorageRootProvider, StorageSettingsCache,
@@ -31,6 +32,7 @@ use reth_trie_db::{
     hashed_storage_from_reverts_with_provider, DatabaseProof, DatabaseStateRoot,
     DatabaseStorageProof, DatabaseStorageRoot,
 };
+use std::collections::{BTreeMap, BTreeSet};
 
 use std::fmt::Debug;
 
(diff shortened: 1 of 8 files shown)


mattsse (changes requested):
Two things to address:

1. **Drop `StorageRangeIter`, return `Vec` instead**: The iterator abstraction is misleading — both implementations eagerly materialize everything into a `BTreeMap` before returning, so it's never lazy. The `limit` parameter is doing useful work inside the provider (bounding the BTreeMap via the cutoff trick in latest, capping expensive history lookups in historical), so keep it. But rename `storage_range_iter` to something like `storage_range` and return `Vec<(B256, StorageEntry)>` directly. The `StorageRangeIter` wrapper with `Box<dyn Iterator>` just adds indirection for no benefit.

2. **Use `evm_env_for_header` instead of `evm_env_at`**: `evm_env_at` does a separate header lookup, but you already have the block loaded. Derive the env from the block's header directly — see #22747 for the pattern:
```rust
let evm_env = self.eth_api().evm_env_for_header(block.sealed_block().sealed_header())?;
```

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -570,6 +573,69 @@ impl<
...
+            let hashed = alloy_primitives::keccak256(slot);
+            if hashed >= key_start {
+                candidates.insert(hashed, *slot);
+            }
+        }
+
+        let mut result: Vec<(B256, StorageEntry)> = Vec::new();
+        for (hashed, slot) in candidates {
+            if let Some(value) = self.storage_by_lookup_key(address, slot)? &&
+                value != StorageValue::ZERO
```
mattsse: these should use is_zero()

crates/storage/provider/src/providers/state/latest.rs
```diff
@@ -273,6 +274,46 @@ impl<Provider: DBProvider + BlockHashReader + StorageSettingsCache> StateProvide
...
+        limit: usize,
+    ) -> ProviderResult<Vec<(B256, StorageEntry)>> {
+        if self.0.cached_storage_settings().use_hashed_state() {
+            return Ok(Vec::new());
+        }
+
+        let mut cursor = self.tx().cursor_dup_read::<tables::PlainStorageState>()?;
+        let mut result: BTreeMap<B256, StorageEntry> = BTreeMap::new();
+        let mut cutoff: Option<B256> = None;
+        let walker = cursor.walk_dup(Some(address), None)?;
```
mattsse, line 291: should we use keystart as the subkey here?

mattsse (changes requested):
maybe @Rjected or @joshieDo can help with reviews on this because unsure if this is correct

PR #22715, opened 2026-03-02
perf(trie): wait for pending changeset computation instead of DB fallback

decofe:
When the deferred trie task is computing changesets for a block, concurrent readers now block-wait for that computation to finish instead of falling back to the expensive DB-based state root computation (which triggers the legacy `StorageRoot::calculate` path ~1.7M times for 100 blocks).

## Changes

- Add `PendingChangeset` (Mutex+Condvar) to `ChangesetCache` so in-flight computations are visible to concurrent readers
- `register_pending()` called before spawning the deferred trie task, returns a guard
- `get_or_compute()` checks for pending entries before falling back to DB
- `insert()` resolves the pending entry and wakes all waiters
- `PendingChangesetGuard` cancels on drop (panic safety) so waiters don't block forever

## Secondary Change

The provider used by the deferred trie task to compute trie changesets for a block is now passed in from the engine task. This prevents a second race-condition where persistence could run prior to the trie changesets for a block being computed, resulting in the changesets for parent blocks being evicted, and thus the deferred trie task needing to calculate changesets also for parent blocks, which takes forever and locks up the node.

## Testing

Existing tests pass. DB fallback is preserved for when no pending computation exists.

Prompted by: brian

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -43,7 +43,7 @@ use reth_provider::{
     StateProviderFactory, StateReader, StorageChangeSetReader, StorageSettingsCache,
 };
 use reth_revm::db::{states::bundle_state::BundleRetention, State};
-use reth_trie::{updates::TrieUpdates, HashedPostState, StateRoot};
+use reth_trie::{trie_cursor::TrieCursorFactory, updates::TrieUpdates, HashedPostState, StateRoot};
 use reth_trie_db::ChangesetCache;
 use reth_trie_parallel::root::{ParallelStateRoot, ParallelStateRootError};
 use revm_primitives::Address;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
checks out

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1564,7 +1564,18 @@ where
...
+        // this computation to finish rather than falling back to the expensive DB path.
+        // The guard ensures the pending entry is cancelled if the task panics.
+        let pending_changeset_guard = self.changeset_cache.register_pending(block_hash);
+
+        // Create the overlay provider NOW, while we're on the engine loop thread and
+        // eviction cannot race with us. If we deferred this to the background task,
+        // persistence could advance and evict changeset cache entries between factory
+        // creation and the task actually running, causing expensive DB fallback
+        // computations when building the overlay.
+        let changeset_provider = overlay_factory.database_provider_ro().ok();
```
shekhirin: why not propagate the error further down without `ok()`, so we can return a proper variant instead of `ProviderError::other` ?

shekhirin (approved):
LGTM, one nit

PR #22718, opened 2026-03-02
perf(engine): skip redundant db fetch in prepare_invalid_response

stevencartavia:
Simplifies `prepare_invalid_response` by using a `match` on `sealed_header_by_hash` to directly determine the `latest_valid_hash` — avoids a redundant call to `latest_valid_hash_for_invalid_payload` when the parent header is already known.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2103,19 +2103,14 @@ where
     /// Prepares the invalid payload response for the given hash, checking the
     /// database for the parent hash and populating the payload status with the latest valid hash
     /// according to the engine api spec.
-    fn prepare_invalid_response(&mut self, mut parent_hash: B256) -> ProviderResult<PayloadStatus> {
-        let parent_header = self.sealed_header_by_hash(parent_hash)?;
-
-        // Edge case: the `latestValid` field is the zero hash if the parent block is the terminal
-        // PoW block, which we need to identify by looking at the parent's block difficulty
-        if parent_header.as_ref().is_some_and(|p| !p.difficulty().is_zero()) {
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -352,7 +352,9 @@ where
             .into())
         }
 
-        let (evm_env, _) = self.eth_api().evm_env_at(block.hash().into()).await?;
+        // The block is already in hand; derive the EVM env from its header directly
+        // rather than fetching it from the database again.
```
mattsse: this comment is pretty much useless without the pr's context, please remove.

could you please submit this separately

crates/engine/tree/src/tree/mod.rs
```diff
@@ -2104,15 +2104,20 @@ where
...
+        if parent_header.as_ref().is_some_and(|p| !p.difficulty().is_zero()) {
             parent_hash = B256::ZERO;
         }
 
-        let valid_parent_hash = self.latest_valid_hash_for_invalid_payload(parent_hash)?;
+        let valid_parent_hash = if parent_header.is_some() && !parent_hash.is_zero() {
+            Some(parent_hash)
+        } else {
+            self.latest_valid_hash_for_invalid_payload(parent_hash)?
+        };
```
mattsse: I think we can do one better ehre, because if parent hash is 0 we dont even need to invoke latest_valid_hash_for_invalid_payload

so this entire section can be simplified

can likely use a match over over `self.sealed_header_by_hash(parent_hash)?`

mattsse (changes requested):

mattsse (approved):

PR #22720, opened 2026-03-02
Use `MockTransactionRatio::new()` method in tests

abhi3700:
This PR ensures that tests in `transaction-pool` use `MockTransactionRatio::new(...)` instead of manually specifying each field. This ensures the internal values always sum to zero (matching type invariants).


mattsse (commented):
we can keep as is

PR #22724, opened 2026-03-02
perf(rpc): reduce redundant DB lookups for receipts

stevencartavia:
- `pending_block_and_receipts()` replaces separate `pending_block()` + `receipts_by_block()` calls
- `load_transaction_and_receipt` uses the block receipt cache + `meta.index` instead of `receipt_by_hash`, which was re-resolving the tx hash to a tx number already resolved by `transaction_by_hash_with_meta`

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
@@ -23,7 +23,7 @@ use reth_rpc_eth_types::{
     PendingBlockEnv, PendingBlockEnvOrigin,
 };
 use reth_storage_api::{
-    noop::NoopProvider, BlockReader, BlockReaderIdExt, ProviderHeader, ProviderTx, ReceiptProvider,
+    noop::NoopProvider, BlockReader, BlockReaderIdExt, ProviderHeader, ProviderTx,
     StateProviderBox, StateProviderFactory,
 };
 use reth_transaction_pool::{
@@ -62,11 +62,8 @@ pub trait LoadPendingBlock:
     ///
     /// If no pending block is available, this will derive it from the `latest` block
     fn pending_block_env_and_cfg(&self) -> Result<PendingBlockEnv<Self::Evm>, Self::Error> {
-        if let Some(block) = self.provider().pending_block().map_err(Self::Error::from_eth_err)? &&
-            let Some(receipts) = self
-                .provider()
-                .receipts_by_block(block.hash().into())
-                .map_err(Self::Error::from_eth_err)?
+        if let Some((block, receipts)) =
+            self.provider().pending_block_and_receipts().map_err(Self::Error::from_eth_err)?
         {
             // Note: for the PENDING block we assume it is past the known merge block and
             // thus this will not fail when looking up the total
--- a/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
@@ -23,7 +23,7 @@ use reth_rpc_eth_types::{
     PendingBlockEnv, PendingBlockEnvOrigin,
 };
 use reth_storage_api::{
-    noop::NoopProvider, BlockReader, BlockReaderIdExt, ProviderHeader, ProviderTx,
+    noop::NoopProvider, BlockReader, BlockReaderIdExt, ProviderHeader, ProviderTx, ReceiptProvider,
     StateProviderBox, StateProviderFactory,
 };
 use reth_transaction_pool::{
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
nice find

we can actually simplify this more now that we have:

https://github.com/paradigmxyz/reth/blob/91182f653593f2a81eddd424802c9728f4bd2a8e/crates/rpc/rpc-eth-types/src/cache/mod.rs#L256-L267

similar to how we do:

https://github.com/paradigmxyz/reth/blob/91182f653593f2a81eddd424802c9728f4bd2a8e/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L623-L651

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -256,23 +256,50 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
+                let meta = TransactionMeta {
+                    tx_hash: hash,
+                    index: cached.tx_index as u64,
+                    block_hash: cached.block.hash(),
+                    block_number: cached.block.number(),
+                    base_fee: cached.block.base_fee_per_gas(),
+                    excess_blob_gas: cached.block.header().excess_blob_gas(),
+                    timestamp: cached.block.timestamp(),
+                };
+                return Ok(Some((tx, meta, receipt)));
```
mattsse: can we clean this up abit with functions, all of this info should be available via the `cached` type

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
mattsse, line 1: can you submit this separately

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -256,23 +256,50 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
+                .spawn_blocking_io(move |this| {
+                    this.provider()
+                        .transaction_by_hash_with_meta(hash)
+                        .map_err(Self::Error::from_eth_err)
+                })
+                .await?
+            {
+                Some(res) => res,
                 None => return Ok(None),
             };
```
mattsse: this isnt quite optimal and doesnt cover all cases, for example we might have already fetched the block and only the receipts are missing in `cached`

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -256,23 +256,50 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
 
-            Ok(Some((tx, meta, receipt)))
-        })
+            let receipt = self
+                .cache()
+                .get_receipts(meta.block_hash)
+                .await
+                .map_err(Self::Error::from_eth_err)?
+                .and_then(|receipts| receipts.get(meta.index as usize).cloned());
```
mattsse: if we go through the cache here anyway then we can also add another helper next to

https://github.com/paradigmxyz/reth/blob/cc30b1e6cc62581014a2cc8af716ed85a4d60467/crates/rpc/rpc-eth-types/src/cache/mod.rs#L260-L260

that always fetches the receipts if missing

mattsse (changes requested):
cool, but this still needs more cleanup

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
stevencartavia: Addressed via the new `get_transaction_with_receipts` helper, fetching receipts on demand if missing so `transaction_by_hash_with_meta` is only called on a true cache miss.

crates/rpc/rpc-eth-api/src/helpers/pending_block.rs
stevencartavia: done in https://github.com/paradigmxyz/reth/pull/22760

crates/rpc/rpc-eth-api/src/helpers/transaction.rs
```diff
@@ -256,23 +256,38 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
...
```
mattsse: we should not fetch all receipts here because we're only interested in the single receipt so this should remain the same

mattsse (changes requested):
this now fetches the entire block receipts but we only need the tx's

so we should keep how we do this if the block's receipts arent cached 

and we also can avoid doing 

```
this.provider()
                        .transaction_by_hash_with_meta(hash)
```
if we have `get_transaction_with_receipts` independent of whether we have the rceipts cached or not

mattsse (approved):

PR #22726, opened 2026-03-03
perf(rpc): derive evm env from header in debug trace

stevencartavia:
In `debug_trace_transaction` and `debug_trace_call_at_tx_index`, derive the EVM env directly from the sealed header via `evm_env_for_header` instead of re-fetching the header from the database with `evm_env_at`.


mattsse (approved):
clean

PR #22727, opened 2026-03-03
fix(net): propagate local pending txs during initial sync

jsvisa:
Reth was blocking propagation of locally pending transactions while the node was in initial sync. 

This diverges from geth behavior and caused the Cancun Hive case `engine-cancun/Blob Transaction Ordering, Multiple Clients` to fail.

So propose to remove the guards of outbound tx announcements during the initial syncing.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -849,10 +849,6 @@ where
     /// transactions to a fraction of peers usually ensures that all nodes receive the transaction
     /// and won't need to request it.
     fn on_new_pending_transactions(&mut self, hashes: Vec<TxHash>) {
-        // Nothing to propagate while initially syncing
-        if self.network.is_initially_syncing() {
-            return
-        }
         if self.network.tx_gossip_disabled() {
             return
         }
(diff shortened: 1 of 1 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -849,10 +849,6 @@ where
     /// transactions to a fraction of peers usually ensures that all nodes receive the transaction
     /// and won't need to request it.
     fn on_new_pending_transactions(&mut self, hashes: Vec<TxHash>) {
-        // Nothing to propagate while initially syncing
-        if self.network.is_initially_syncing() {
-            return
-        }
```
mattsse, line 855: oh, I see, can we make a comment here about symmetry: basically we dont need to skip this here because during initial sync we only see local transactions here because we dont import txs over p2p?

mattsse (changes requested):
very, nice, this makes sense

I have one pedantic doc suggestion


btw, really appreciate all the fixes 🙏

crates/net/network/src/transactions/mod.rs
jsvisa: yes, during initial sync we skip importing tx announcements from peers in `on_new_pooled_transaction_hashes`, so tx reached here is local

mattsse (approved):
lgtm!

crates/net/network/src/transactions/mod.rs
jsvisa: @mattsse could you please also take a look at https://github.com/ethereum/hive/pull/1397, thanks

PR #22731, opened 2026-03-03
fix(args): correct MetricArgs default for push_gateway_interval

aso20455:
The `MetricArgs` struct was using `#[derive(Default)]` which meant `push_gateway_interval` defaulted to 0 seconds when created programmatically (e.g. in `NodeConfig::new()`), while the CLI default was 5 seconds.

Added manual `Default` impl to keep both consistent at 5 seconds.


mattsse (approved):
LGTM, fixes the derive(Default) vs CLI default inconsistency for push_gateway_interval.

PR #22739, opened 2026-03-03
chore(downloader): simplify the canonical blocks check

jsvisa:
replace the collect all header nums into Vec, sort and then iterate checking into a simple `max-min+1` gap check.

### DIFF
--- a/crates/net/downloaders/src/file_client.rs
+++ b/crates/net/downloaders/src/file_client.rs
@@ -163,17 +163,10 @@ impl<B: FullBlock> FileClient<B> {
         if self.headers.is_empty() {
             return true
         }
-        let mut nums = self.headers.keys().copied().collect::<Vec<_>>();
-        nums.sort_unstable();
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this makes sense,

mind double checking this @DaniPopes

crates/net/downloaders/src/file_client.rs
```diff
@@ -163,17 +163,10 @@ impl<B: FullBlock> FileClient<B> {
...
-        let mut iter = nums.into_iter();
-        let mut lowest = iter.next().expect("not empty");
-        for next in iter {
-            if next != lowest + 1 {
-                return false
-            }
-            lowest = next;
-        }
-        true
+        let min = self.min_block().expect("not empty");
```
DaniPopes: nit: can get min and max in one iterator instead of two

DaniPopes (approved):

mattsse (approved):

PR #22742, opened 2026-03-03
chore(provider): remove unnecessary collect in changeset readers

jsvisa:
iterate directly over `head_block.chain()` instead of collecting into a `Vec` first, avoiding an unnecessary intermediate allocation.


mattsse (approved):
this is equivlaent because the pointer wont change while we're holding this:

https://github.com/paradigmxyz/reth/blob/c71e946306025d9aa8ec21d8284e4b48145e238b/crates/chain-state/src/in_memory.rs#L686-L689

pending @joshieDo or @Rjected

joshieDo (approved):

PR #22746, opened 2026-03-03
feat(rpc): early check gas_limit in bundle api

jsvisa:
check user's gas_limit against configured first, this avoids wasting a db lookup with `evm_env_at.await?` when gas_limit is invalid:

https://github.com/paradigmxyz/reth/blob/0a2f4f2babc62a23072ab331554c8275c0692b90/crates/rpc/rpc/src/eth/bundle.rs#L97

And this change also avoids double write on the `block_env.block_limit`


mattsse (approved):
nice!

PR #22747, opened 2026-03-03
perf(rpc): derive evm env from loaded block header

jsvisa:
follow on #22726 to eliminate one extra redundant header fetch

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/trace.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/trace.rs
@@ -265,16 +265,11 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
         R: Send + 'static,
     {
         async move {
-            let block = async {
-                if block.is_some() {
-                    return Ok(block)
-                }
-                self.recovered_block(block_id).await
-            };
-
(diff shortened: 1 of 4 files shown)


crates/rpc/rpc-eth-api/src/helpers/trace.rs
```diff
@@ -265,16 +266,13 @@ pub trait Trace: LoadState<Error: FromEvmError<Self::Evm>> + Call {
...
-                self.recovered_block(block_id).await
-            };
-
-            let ((evm_env, _), block) = futures::try_join!(self.evm_env_at(block_id), block)?;
+            let block =
+                if block.is_some() { block } else { self.recovered_block(block_id).await? };
 
-            let Some(block) = block else { return Ok(None) };
+            let Some(block) = block else {
+                return Err(EthApiError::HeaderNotFound(block_id).into())
```
mattsse: this alters behaviour

can we keep this as is?

mattsse (changes requested):
ty! this makes sense, we forgot to clean that up

crates/rpc/rpc-eth-api/src/helpers/trace.rs
jsvisa: Good catch, seems the it/test was wrong, I'll update it in the test crate.

mattsse (approved):

PR #22748, opened 2026-03-03
perf(engine): batch MDBX storage reads in BAL prewarming

stevencartavia:
Adds `storage_range` to the `StateProvider` trait and uses it in BAL prewarming to batch all storage slot reads into a single cursor per account instead of one cursor per slot.

Also fixes `prefetch_bal_account` using `CachedStateProvider::new` instead of `new_prewarm`, which meant reads during prefetch were not populating the cache.

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -458,6 +458,63 @@ impl<S: StateProvider, const PREWARM: bool> StateProvider for CachedStateProvide
             self.state_provider.storage(account, storage_key)
         }
     }
+
+    fn storage_range(
+        &self,
+        account: Address,
+        keys: &[StorageKey],
+    ) -> ProviderResult<Vec<(StorageKey, StorageValue)>> {
+        let mut uncached_keys = Vec::new();
+        let mut result = Vec::with_capacity(keys.len());
+
+        if PREWARM {
+            for &key in keys {
+                if let Some(value) = self.caches.0.storage_cache.get(&(account, key)) {
+                    if !value.is_zero() {
+                        result.push((key, value));
+                    }
+                } else {
+                    uncached_keys.push(key);
+                }
+            }
+
+            // Batch-fetch all uncached keys from the inner provider
+            if !uncached_keys.is_empty() {
+                let fetched = self.state_provider.storage_range(account, &uncached_keys)?;
+                for (key, value) in &fetched {
+                    self.caches.insert_storage(account, *key, Some(*value));
+                }
+                // Insert zero entries for keys that weren't found
+                let fetched_set: std::collections::HashSet<StorageKey> =
+                    fetched.iter().map(|(k, _)| *k).collect();
+                for &key in &uncached_keys {
+                    if !fetched_set.contains(&key) {
+                        self.caches.insert_storage(account, key, Some(StorageValue::ZERO));
+                    }
+                }
+                result.extend(fetched);
+            }
+        } else {
+            for &key in keys {
+                if let Some(value) = self.caches.0.storage_cache.get(&(account, key)) {
+                    self.metrics.storage_cache_hits.increment(1);
+                    if !value.is_zero() {
+                        result.push((key, value));
+                    }
+                } else {
+                    self.metrics.storage_cache_misses.increment(1);
+                    uncached_keys.push(key);
+                }
+            }
+
+            if !uncached_keys.is_empty() {
+                let fetched = self.state_provider.storage_range(account, &uncached_keys)?;
+                result.extend(fetched);
+            }
+        }
+
+        Ok(result)
+    }
 }
 
 impl<S: BytecodeReader, const PREWARM: bool> BytecodeReader for CachedStateProvider<S, PREWARM> {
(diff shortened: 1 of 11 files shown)


crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -351,6 +351,63 @@ impl<S: StateProvider, const PREWARM: bool> StateProvider for CachedStateProvide
...
+        &self,
+        account: Address,
+        keys: &[StorageKey],
+    ) -> ProviderResult<Vec<(StorageKey, StorageValue)>> {
+        let mut uncached_keys = Vec::new();
+        let mut result = Vec::with_capacity(keys.len());
+
+        if PREWARM {
+            for &key in keys {
+                if let Some(value) = self.caches.0.storage_cache.get(&(account, key)) {
```
emmajam: nit: using raw cache get instead of atomic insert

PREWARM path for storage_range doesn't match the semantics of the existing `storage()` method which uses `get_or_try_insert_storage_with` for atomic check-and-insert

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -351,6 +351,63 @@ impl<S: StateProvider, const PREWARM: bool> StateProvider for CachedStateProvide
...
+                        result.push((key, value));
+                    }
+                } else {
+                    self.metrics.storage_cache_misses.increment(1);
+                    uncached_keys.push(key);
+                }
+            }
+
+            if !uncached_keys.is_empty() {
+                let fetched = self.state_provider.storage_range(account, &uncached_keys)?;
```
emmajam: storage_range on the non-PREWARM path will always miss cache for uncached keys on repeated calls, we should consider whether the non-PREWARM path of storage_range is even reachable / useful

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -351,6 +351,63 @@ impl<S: StateProvider, const PREWARM: bool> StateProvider for CachedStateProvide
...
+            }
+
+            // Batch-fetch all uncached keys from the inner provider
+            if !uncached_keys.is_empty() {
+                let fetched = self.state_provider.storage_range(account, &uncached_keys)?;
+                for (key, value) in &fetched {
+                    self.caches.insert_storage(account, *key, Some(*value));
+                }
+                // Insert zero entries for keys that weren't found
+                let fetched_set: std::collections::HashSet<StorageKey> =
```
emmajam: since keys are already sorted+deduped in the caller, a more efficient approach would be to iterate the sorted `uncached_keys` and sorted `fetched` results in tandem (merge-join) to find missing entries, avoiding the allocation entirely

crates/storage/provider/src/providers/state/latest.rs
```diff
@@ -263,6 +263,41 @@ impl<Provider: DBProvider + BlockHashReader + StorageSettingsCache> StateProvide
...
+    ) -> ProviderResult<Vec<(StorageKey, StorageValue)>> {
+        let mut result = Vec::with_capacity(keys.len());
+        if keys.is_empty() {
+            return Ok(result);
+        }
+
+        if self.0.cached_storage_settings().use_hashed_state() {
+            let hashed_address = alloy_primitives::keccak256(account);
+            let mut cursor = self.tx().cursor_dup_read::<tables::HashedStorages>()?;
+            for &key in keys {
```
emmajam: The storage_range implementation in `LatestStateProviderRef` creates a new cursor but calls `seek_by_key_subkey` for each key independently. 

For the hashed path, keys are hashed and lose their sort order, so cursor seeks may jump backwards (non-sequential). 

This is correct but doesn't benefit from cursor locality. 

I think sorting the hashed keys before iteration would be a meaningful improvement.

emmajam (commented):
nice!

emmajam (approved):

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -598,20 +603,25 @@ where
...
+            Vec::with_capacity(account.storage_changes.len() + account.storage_reads.len());
         for slot in &account.storage_changes {
-            let _ = state_provider.storage(account.address, StorageKey::from(slot.slot));
+            slots.push(StorageKey::from(slot.slot));
         }
         for &slot in &account.storage_reads {
-            let _ = state_provider.storage(account.address, StorageKey::from(slot));
+            slots.push(StorageKey::from(slot));
         }
+        slots.sort_unstable();
```
mediocregopher: According to the EIP
```
storage_changes: Slots lexicographic by storage key; within each slot, changes by block access index (ascending)
storage_reads: Lexicographic by storage key
```
So you should be able to use something like [merge_join_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge_join_by) to dedup in a single pass without sorting.

crates/engine/tree/src/tree/cached_state.rs
```diff
@@ -351,6 +351,60 @@ impl<S: StateProvider, const PREWARM: bool> StateProvider for CachedStateProvide
...
+                        continue;
+                    }
+                    // key not returned by inner provider → zero slot
+                    let _ = self.caches.get_or_try_insert_storage_with(account, key, || {
+                        Ok::<_, ProviderError>(StorageValue::ZERO)
+                    });
+                }
+            }
+        } else {
+            // Not called on the execution path; delegate directly.
```
mediocregopher: Why not just always populate the cache?

mediocregopher (approved):

PR #22749, opened 2026-03-03
feat(rpc): implement debug_getModifiedAccountsByNumber/Hash

joohhnnn:
These two endpoints were defined in the API trait but returned empty Ok(()) silently. They should return all account addresses changed between two blocks — geth/besu/nethermind all have working implementations.

reth already has ChangeSetReader and AccountChangeSets storing the needed data, just wasn't wired to the RPC layer. Connects them: iterate change sets over (start, end], collect addresses, done. The hash variant resolves hashes to block numbers then delegates. Also fixes the return type from RpcResult<()> to RpcResult<Vec<Address>>.


joohhnnn:
CI error: 500

crates/rpc/rpc/src/debug.rs
```diff
@@ -631,6 +632,7 @@ where
 impl<Eth> DebugApiServer<RpcTxReq<Eth::NetworkTypes>> for DebugApi<Eth>
 where
     Eth: EthTransactions + TraceExt,
+    Eth::Provider: ChangeSetReader,
```
mattsse: I think we can add this to:

https://github.com/paradigmxyz/reth/blob/fcf66452424b5a2f22db940f544c5a45dc6fc79d/crates/rpc/rpc-eth-api/src/node.rs#L28-L28

then we dont need to enforce the bound here or anywhere else

crates/rpc/rpc/src/debug.rs
```diff
@@ -947,18 +949,41 @@ where
...
+            .provider()
+            .header(start_hash)
+            .to_rpc_result()?
+            .ok_or_else(|| internal_rpc_err(format!("block {start_hash} not found")))?;
+        let end_header = self
+            .provider()
+            .header(end_hash)
+            .to_rpc_result()?
+            .ok_or_else(|| internal_rpc_err(format!("block {end_hash} not found")))?;
+        self.debug_get_modified_accounts_by_number(start_header.number(), end_header.number()).await
```
mattsse: this does blocking io and should be spawned as such

mattsse (changes requested):
an alternative to reading changesets could be replaying the block and look at the output.

but this should also work.

a few nits

crates/rpc/rpc/src/debug.rs
joohhnnn: Yeah, nice

joohhnnn:
cc @mattsse

PR #22750, opened 2026-03-03
perf(rpc): check cache before hitting DB in `eth_getTransactionReceipt`

MagicJoshh:
`transaction_by_hash` already checks the RPC cache before going to disk, but `transaction_receipt` skipped straight to two DB calls (`transaction_by_hash_with_meta` + `receipt_by_hash`). `CachedTransaction::into_receipt()` was literally written for this but never wired up. Now it is.


figtracer:
this is being worked on here
https://github.com/paradigmxyz/reth/pull/22724

mattsse (commented):
yep, this is dup of https://github.com/paradigmxyz/reth/pull/22724

PR #22753, opened 2026-03-03
feat(rpc): implement debug_storageRangeAt

joohhnnn:
Wanted to use `debug_storageRangeAt` to inspect contract storage at a specific transaction index, found it was a noop returning `Ok(())`.

The main challenge was that `StateProvider` only had single-key `storage()`, no range iteration. So I added `storage_range()` through the full storage abstraction chain (`StateProvider` → `EvmStateProvider` → `StateProviderTraitObjWrapper`), with an efficient cursor-based implementation in `LatestStateProviderRef` using the `PlainStorageState` DupSort table.

The RPC handler replays transactions up to `tx_idx`, then merges the in-memory execution cache (modified/deleted slots) with the base storage from the provider to build the paginated result.

### DIFF
--- a/crates/revm/src/database.rs
+++ b/crates/revm/src/database.rs
@@ -1,4 +1,5 @@
 use crate::primitives::alloy_primitives::{BlockNumber, StorageKey, StorageValue};
+use alloc::vec::Vec;
 use alloy_primitives::{Address, B256, U256};
 use core::ops::{Deref, DerefMut};
 use reth_primitives_traits::Account;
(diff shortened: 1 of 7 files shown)


figtracer:
this is being worked on here https://github.com/paradigmxyz/reth/pull/22655

mattsse:
Closing in favor of #22714 which covers this with correct keccak-hashed pagination, historical state support, and v2/RocksDB awareness.

PR #22754, opened 2026-03-03
feat(rpc): implement debug_intermediateRoots

joohhnnn:
Implement debug_intermediateRoots endpoint that executes a block and returns the state root after each transaction.

Replays transactions on the parent state with transition tracking enabled, computing the state root after each transaction via merge_transitions + hashed_post_state + state_root.

### DIFF
--- a/crates/rpc/rpc/src/debug.rs
+++ b/crates/rpc/rpc/src/debug.rs
@@ -1,7 +1,7 @@
 use alloy_consensus::{transaction::TxHashRef, BlockHeader};
 use alloy_eip7928::BlockAccessList;
 use alloy_eips::{eip2718::Encodable2718, BlockId, BlockNumberOrTag};
-use alloy_evm::env::BlockEnvironment;
+use alloy_evm::{env::BlockEnvironment, Evm};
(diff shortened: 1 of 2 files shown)


crates/rpc/rpc/src/debug.rs
```diff
@@ -625,6 +626,43 @@ where
...
     }
+
+    /// Executes a block and returns the state root after each transaction.
+    pub async fn intermediate_roots(&self, block_hash: B256) -> Result<Vec<B256>, Eth::Error> {
+        let ((evm_env, _), block) = futures::try_join!(
+            self.eth_api().evm_env_at(block_hash.into()),
+            self.eth_api().recovered_block(block_hash.into()),
+        )?;
+
+        let block = block.ok_or(EthApiError::HeaderNotFound(block_hash.into()))?;
```
mattsse: this can be simplified see:

https://github.com/paradigmxyz/reth/pull/22747

mattsse (changes requested):

mattsse (approved):
lgtm

PR #22757, opened 2026-03-04
fix: fix clippy warnings

stevencartavia:
fix clippy warnings when running `cargo clippy --workspace --lib --examples --tests --benches --all-features --locked`


mattsse (commented):
I dont see these warnings and the match over a bool is just weird

PR #22759, opened 2026-03-04
fix(rpc): correct call_many block lookup errors

Elaela22soL:
Preserve provider lookup failures as internal errors in `eth_callMany` block resolution, keep `HeaderNotFound` only for missing block hash.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -324,12 +324,14 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
             // if it's not pending, we should always use block_hash over block_number to ensure that
             // different provider calls query data related to the same block.
             if !is_block_target_pending {
-                target_block = self
+                let Some(block_hash) = self
                     .provider()
                     .block_hash_for_id(target_block)
-                    .map_err(|_| EthApiError::HeaderNotFound(target_block))?
-                    .ok_or_else(|| EthApiError::HeaderNotFound(target_block))?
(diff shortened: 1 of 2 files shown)


mattsse (commented):
we can keep this as is

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -324,12 +324,14 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
...
             // different provider calls query data related to the same block.
             if !is_block_target_pending {
-                target_block = self
+                let Some(block_hash) = self
                     .provider()
                     .block_hash_for_id(target_block)
-                    .map_err(|_| EthApiError::HeaderNotFound(target_block))?
-                    .ok_or_else(|| EthApiError::HeaderNotFound(target_block))?
-                    .into();
+                    .map_err(|err| Self::Error::from_eth_err(EthApiError::Internal(err.into())))?
```
mattsse: actually we should use

https://github.com/paradigmxyz/reth/blob/0a7d7e0d663c0ed0222df3ee028efdc3fa2a4c25/crates/rpc/rpc-eth-types/src/error/mod.rs#L500-L500

here

mattsse (changes requested):
this does make sense actually, but I have one suggestion

crates/rpc/rpc-eth-api/src/helpers/call.rs
Elaela22soL: @mattsse Updated, sir, please re-check.

mattsse (approved):

PR #22762, opened 2026-03-04
perf: pin engine and sparse-trie threads to dedicated CPU cores

decofe:
## Summary
Pin the engine and sparse-trie threads to individual CPU cores via `sched_setaffinity`, reducing context-switch overhead and improving cache locality.

## Changes
- Added `pin_current_thread_to_core()` in `crates/tasks/src/utils.rs` — uses `sched_getaffinity`/`sched_setaffinity` to assign each caller a dedicated core in round-robin order from the available CPU set
- Called it in the engine thread spawn (`crates/engine/tree/src/tree/mod.rs`)
- Called it in the sparse-trie thread spawn (`crates/engine/tree/src/tree/payload_processor/mod.rs`)

Respects existing process affinity (cgroups/taskset/isolcpus). No-op on non-Linux. Failures are logged at debug level and are not fatal.

## Testing
`cargo check -p reth-tasks -p reth-engine-tree` — clean
`cargo clippy -p reth-tasks` — clean

Co-Authored-By: DaniPopes <57450786+DaniPopes@users.noreply.github.com>

Prompted by: DaniPopes

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -38,7 +38,7 @@ use reth_provider::{
 };
 use reth_revm::database::StateProviderDatabase;
 use reth_stages_api::ControlFlow;
-use reth_tasks::{spawn_os_thread, utils::increase_thread_priority};
+use reth_tasks::{spawn_os_thread, utils::{increase_thread_priority, pin_current_thread_to_core}};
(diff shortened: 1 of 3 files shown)


DaniPopes:
derek bench

DaniPopes:
derek bench

PR #15884, opened 2025-04-23
chore(metrics): summable reth_info value

danyalprout:
### Description
We run multiple clusters of Reth/Geth nodes. We want to be able to sum the number of `reth_info` metrics by version to have a count of nodes that are on each version, e.g. compare our Geth devnet vs. Reth

Screenshot 2025-04-23 at 11 14 56 AM

For historical context, this was previously set in https://github.com/paradigmxyz/reth/pull/9679, but it was removed in https://github.com/paradigmxyz/reth/pull/9691. If there's an issue with setting this metric to 1, happy to find a workaround.

PR #22764, opened 2026-03-04
fix(metrics): set chain_spec gauge value

reallesee:
Same fix as #15884 for version.rs, but for chain.rs which was missed.


mattsse (approved):
LGTM, same fix as #15884 for the chain_spec gauge.

PR #22766, opened 2026-03-04
perf: LFU-based sparse trie cache

pepyakin:
## Motivation

The sparse trie cache persists between blocks to avoid expensive re-reveals (DB fetches) of nodes needed again. Pruning bounds memory, but the old strategy poorly predicts what to keep:

1. **Depth-based pruning is blind** — it cuts at a fixed depth regardless of which keys are actually being reused across blocks.
2. **Subtrie-level heat tracking is defeated by hashing at scale** — since keys are keccak256 hashes, updates distribute uniformly across all 256 lower subtries under load. With enough touched slots per block, every subtrie looks equally hot, so the heat mechanism can't distinguish frequently-used regions from cold ones. (It could discriminate with very few updates, but that's exactly when pruning pressure is low and the mechanism isn't needed.)
3. **Whole-trie storage retention is too coarse** — `max_storage_tries` kept the top N whole contract-level storage tries, ranked by `node_count × heat_multiplier`. A massive contract like USDT would dominate by sheer size even if only 1 slot was actually hot, while a smaller contract with many hot slots could get evicted entirely.

The net effect: pruning decisions don't correlate with actual re-use, leading to unnecessary re-reveals (cache misses) and wasted memory (retaining cold data).

## Strategy

Track hotness at the individual key level. Every storage slot write records a `(address, slot)` touch and every account write records an account touch in a `BucketedLfu` frequency cache. At prune time, retain only the top-N hottest slots and top-M hottest accounts, then prune tries to keep exactly the paths leading to retained keys, collapsing everything else to hash stubs. This makes pruning proportional to actual re-use rather than trie topology.

### DIFF
--- a/crates/engine/primitives/src/config.rs
+++ b/crates/engine/primitives/src/config.rs
@@ -29,7 +29,7 @@ pub const DEFAULT_SPARSE_TRIE_PRUNE_DEPTH: usize = 4;
 /// Default LFU hot-slot capacity for sparse trie pruning.
 ///
 /// Limits the number of `(address, slot)` pairs retained across prune cycles.
-pub const DEFAULT_SPARSE_TRIE_MAX_HOT_SLOTS: usize = 10000;
+pub const DEFAULT_SPARSE_TRIE_MAX_HOT_SLOTS: usize = 1500;
 
 /// Default LFU hot-account capacity for sparse trie pruning.
 ///
--- a/crates/engine/primitives/src/config.rs
+++ b/crates/engine/primitives/src/config.rs
@@ -26,10 +26,15 @@ pub const DEFAULT_RESERVED_CPU_CORES: usize = 1;
 /// Depth 4 means we keep roughly 16^4 = 65536 potential branch paths at most.
 pub const DEFAULT_SPARSE_TRIE_PRUNE_DEPTH: usize = 4;
 
-/// Default maximum number of storage tries to keep after pruning.
+/// Default LFU hot-slot capacity for sparse trie pruning.
 ///
-/// Storage tries beyond this limit are cleared (but allocations preserved).
-pub const DEFAULT_SPARSE_TRIE_MAX_STORAGE_TRIES: usize = 100;
+/// Limits the number of `(address, slot)` pairs retained across prune cycles.
+pub const DEFAULT_SPARSE_TRIE_MAX_HOT_SLOTS: usize = 10000;
+
+/// Default LFU hot-account capacity for sparse trie pruning.
+///
+/// Limits the number of account addresses retained across prune cycles.
+pub const DEFAULT_SPARSE_TRIE_MAX_HOT_ACCOUNTS: usize = 1000;
 
 /// Default timeout for the state root task before spawning a sequential fallback.
 pub const DEFAULT_STATE_ROOT_TASK_TIMEOUT: Duration = Duration::from_secs(1);
@@ -131,8 +136,10 @@ pub struct TreeConfig {
     disable_cache_metrics: bool,
     /// Depth for sparse trie pruning after state root computation.
     sparse_trie_prune_depth: usize,
-    /// Maximum number of storage tries to retain after pruning.
-    sparse_trie_max_storage_tries: usize,
+    /// LFU hot-slot capacity: max `(address, slot)` pairs retained across prune cycles.
+    sparse_trie_max_hot_slots: usize,
+    /// LFU hot-account capacity: max account addresses retained across prune cycles.
+    sparse_trie_max_hot_accounts: usize,
     /// Whether to fully disable sparse trie cache pruning between blocks.
     disable_sparse_trie_cache_pruning: bool,
     /// Timeout for the state root task before spawning a sequential fallback computation.
@@ -165,7 +172,8 @@ impl Default for TreeConfig {
             allow_unwind_canonical_header: false,
             disable_cache_metrics: false,
             sparse_trie_prune_depth: DEFAULT_SPARSE_TRIE_PRUNE_DEPTH,
-            sparse_trie_max_storage_tries: DEFAULT_SPARSE_TRIE_MAX_STORAGE_TRIES,
+            sparse_trie_max_hot_slots: DEFAULT_SPARSE_TRIE_MAX_HOT_SLOTS,
+            sparse_trie_max_hot_accounts: DEFAULT_SPARSE_TRIE_MAX_HOT_ACCOUNTS,
             disable_sparse_trie_cache_pruning: false,
             state_root_task_timeout: Some(DEFAULT_STATE_ROOT_TASK_TIMEOUT),
         }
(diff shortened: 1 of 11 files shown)


pepyakin:
derek bench warmup=1000 blocks=4000

crates/trie/sparse/src/parallel.rs
```diff
@@ -1157,198 +1150,115 @@ impl SparseTrie for ParallelSparseTrie {
...
+                        .subtrie_for_path_mut_untracked(&parent_path)
+                        .expect("parent subtrie exists")
+                        .nodes
+                        .get_mut(&parent_path)
+                        .expect("expected parent branch node")
+                    else {
+                        panic!("expected branch node at path {parent_path:?}");
+                    };
+
+                    let nibble = path.last().unwrap();
```
pepyakin, line 1206: are we even leaving justifications for things like those, or are they self evident?

decofe:
cc @pepyakin

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/22714667430/job/65861216780)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `pep/hot-slots-trie`

crates/trie/sparse/src/parallel.rs
```diff
@@ -1157,198 +1150,115 @@ impl SparseTrie for ParallelSparseTrie {
...
-    fn prune(&mut self, max_depth: usize) -> usize {
+    fn prune(&mut self, retained_leaves: &[Nibbles]) -> usize {
         #[cfg(feature = "trie-debug")]
         self.debug_recorder.reset();
 
-        // Decay heat for subtries not modified this cycle
-        self.subtrie_heat.decay_and_reset();
+        let mut retained_leaves = retained_leaves.to_vec();
+        retained_leaves.sort_unstable();
+        retained_leaves.dedup();
```
mediocregopher: imo if we don't expect there to be duplicates here (which I don't believe we do, since the LFU cache should already dedup) when I'd rather this be a debug_assert asserting that there's no dupes, or just leave it out completely.

mediocregopher (approved):

crates/trie/sparse/src/parallel.rs
pepyakin: good catch.

Also, it shouldn't affect correctness so I just removed it.

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -594,6 +598,55 @@ where
...
+    /// Computes storage roots for accounts whose storage updates are fully drained.
+    ///
+    /// For each storage trie T that:
+    /// 1. was modified in the current block,
+    /// 2. all the storage updates are fully drained,
+    /// 3. but the storage root hasn't been updated yet,
+    ///
+    /// we trigger state root computation on a rayon pool.
+    fn compute_drained_storage_roots(&mut self) {
+        let span = debug_span!("compute_storage_roots").entered();
```
DaniPopes: nit: can be #[instrument], also needs target

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -594,6 +598,55 @@ where
...
+                ?address
+            )
+            .entered();
+            // SAFETY:
+            // - pointers are created from `storage_tries_mut().get_mut(address)` above;
+            // - `addresses_to_compute_roots` comes from map iteration, so addresses are unique;
+            // - we do not insert/remove entries between pointer collection and use, so pointers
+            //   stay valid and map reallocation cannot occur;
+            // - each pointer is consumed by at most one rayon task, so no aliasing mutable access.
+            unsafe { (*trie).root().expect("updates are drained, trie should be revealed by now") };
```
DaniPopes, line 651: unsafe lgtm

crates/trie/sparse/src/state.rs
```diff
@@ -997,93 +968,164 @@ impl<S: SparseTrieTrait + Clone> StorageTries<S> {
...
 
-    /// Clears all heat tracking state.
-    fn clear(&mut self) {
-        self.state.clear();
-        self.accessed_this_cycle.clear();
+/// Slot-specific helpers for the `BucketedLfu<HotSlotKey>`.
+impl BucketedLfu<HotSlotKey> {
+    /// Returns retained slots grouped by address.
+    fn retained_slots_by_address(&self) -> B256Map<Vec<Nibbles>> {
+        let mut grouped = B256Map::<Vec<Nibbles>>::default();
```
DaniPopes: with_capacity?

crates/trie/sparse/src/parallel.rs
mediocregopher: I usually try to but I'm not super diligent about it

DaniPopes (approved):
lgtm nitZ

crates/engine/primitives/src/config.rs
```diff
@@ -26,10 +26,15 @@ pub const DEFAULT_RESERVED_CPU_CORES: usize = 1;
 /// Depth 4 means we keep roughly 16^4 = 65536 potential branch paths at most.
 pub const DEFAULT_SPARSE_TRIE_PRUNE_DEPTH: usize = 4;
 
-/// Default maximum number of storage tries to keep after pruning.
+/// Default LFU hot-slot capacity for sparse trie pruning.
 ///
-/// Storage tries beyond this limit are cleared (but allocations preserved).
-pub const DEFAULT_SPARSE_TRIE_MAX_STORAGE_TRIES: usize = 100;
+/// Limits the number of `(address, slot)` pairs retained across prune cycles.
+pub const DEFAULT_SPARSE_TRIE_MAX_HOT_SLOTS: usize = 1500;
```
yongkangc, line 32: curious how did we decide on these numbers?

yongkangc (approved):
nice one! lgtm 🚀  just one question!

crates/trie/sparse/src/state.rs
```diff
@@ -997,93 +968,164 @@ impl<S: SparseTrieTrait + Clone> StorageTries<S> {
...
 }
 
-#[allow(dead_code)]
-impl StorageTrieModifications {
-    /// Marks a storage trie as accessed this cycle.
-    /// Heat and `prune_backlog` are updated in [`Self::update_and_reset`].
-    #[inline]
-    fn mark_accessed(&mut self, address: B256) {
-        self.accessed_this_cycle.insert(address);
+impl<K> Default for BucketedLfu<K> {
```
yongkangc: also should we consider about adding tests for `BucketedLfu` ? think thats prob the most complex part of this change

crates/engine/primitives/src/config.rs
pepyakin: Trial and error! If you increase it too high, the maintenance overheads start to dominate. Set it too low and the hit rate starts to plummet.

crates/trie/sparse/src/state.rs
pepyakin: will do!

crates/engine/primitives/src/config.rs
yongkangc: ahh i see, thats a good note

PR #22767, opened 2026-03-04
feat(trie): add sparse trie cache hit rate metrics

decofe:
## Summary

Add metrics to track sparse trie cache hit rate in `update_leaves`.

## Motivation

We need a proper way to measure how effective the sparse trie cache is, rather than using proxies like revealed node counts. Every key removed from the `updates` map after `update_leaves` is a cache hit (the trie path was already revealed), and every key left behind is a cache miss (required a new proof).

## Changes

- Added 4 histogram metrics to `MultiProofTaskMetrics`: `sparse_trie_{account,storage}_cache_{hits,misses}`
- In `process_leaf_updates`, measure `updates.len()` before/after `update_leaves()` for storage tries and record aggregated hits/misses
- In `process_account_leaf_updates`, do the same for the account trie

## Testing

`cargo check -p reth-engine-tree` passes clean.

Prompted by: alexey


mediocregopher (approved):

yongkangc (approved):
Makes sense

PR #22772, opened 2026-03-04
feat(devnet): add cli flags for the different BAL methods

emmajam:
Adding BAL specific cli flags

| Column | Meaning | Flag | 
| - | -| - |
| Exec Par. | Parallel execution using BAL | --engine.disable-bal-parallel-execution |
| State Par. | Parallel state root using BAL | --engine.disable-bal-parallel-state-root |
| Batch IO | Batched storage reads | Steven's PR #22748 |


mattsse (approved):
makes sense

PR #22776, opened 2026-03-04
fix(net): mark transactions as seen in propagate_hashes_to

bobtajson:
Add missing `peer.seen_transactions.insert()` in `propagate_hashes_to`, matching the pattern in `propagate_full_transactions_to_peer` and `propagate_transactions`. Without this, transactions manually propagated as hashes were not marked as seen, so subsequent `propagate_transactions` calls would redundantly re-send them to the same peer.


mattsse (approved):
right, this makes sense

PR #22784, opened 2026-03-04
feat(bench): add `wait_for_*` arguments to `reth_newPayload`

decofe:
## Summary

Adds optional `wait_for_persistence` and `wait_for_caches` boolean arguments to `reth_newPayload`.

Additionally, fully removes `wait_for_persistence` and infrastructure around it — we can now use `reth_newPayload`.

## Motivation

Currently `reth_newPayload` always waits for in-flight persistence and cache locks before executing the payload. This is useful for benchmarking, but sometimes you want to send blocks as fast as possible without wait time in between to test more extreme scenarios.

Co-Authored-By: Alexey Shekhirin <5773434+shekhirin@users.noreply.github.com>

Prompted by: alexey


mattsse (dismissed):
can we rename wait to wait_for_xyz 
because otherwise hard to understand what this is waiting for

mediocregopher (approved):

joshieDo (approved):

PR #22785, opened 2026-03-04
perf(net): reorder filters to run cheap checks first

jsvisa:
Reorder the filtering pipelines in tx handling:

- `on_get_pooled_transactions`: check `tx_gossip_disabled` first, it avoids an unnecessary HashMap lookup when gossip is disabled; 
- `on_new_pooled_transaction_hashes`: check spam first(by checking the tx type), as the `pool.retain_unknown` acquires RwLock on the entire pool


mattsse (approved):
this makes sense, ty!

PR #22786, opened 2026-03-04
perf(engine): prepare canonical overlay after forkchoice update

joohhnnn:
`prepare_canonical_overlay()` was only called after persistence, not after forkchoice updates. Since FCU precedes payload building and persistence lags behind, the overlay cache missed on nearly every payload request.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2501,6 +2501,13 @@ where
         // sends an event to all active listeners about the new canonical chain
         self.canonical_in_memory_state.notify_canon_state(notification);
 
+        // Pre-compute the canonical overlay so it's ready when the next payload arrives.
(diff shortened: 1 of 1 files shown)


mattsse (commented):
@mediocregopher I dont think this is necessary/useful here because we already prepare all this in

https://github.com/paradigmxyz/reth/blob/1ffff8f8e3419c245ec7d88a2b009187ece43c0e/crates/engine/tree/src/tree/payload_validator.rs#L1689-L1692

for each block I believe

and we use this one here to re-anchor after persistence

https://github.com/paradigmxyz/reth/blob/1ffff8f8e3419c245ec7d88a2b009187ece43c0e/crates/engine/tree/src/tree/mod.rs#L1460-L1467

so afaict doing this on fcu is redundant because already taken care of by the deferred trie task?

PR #22798, opened 2026-03-05
fix(prune): correct broken test for set_deleted_entries_limit

rejected-l:
Test was cloning struct and mutating the clone instead of the original - always passed but tested nothing. Fixed to verify actual behavior.


mattsse (approved):
lgtm, good catch

PR #22803, opened 2026-03-05
fix(engine): persist blocks before pruning to survive SIGTERM

skylenet:
## Context / Issue

I took a mainnet snapshot at height 24350000 . While starting from that snapshot, I've noticed that reth was missing 2 blocks (head=24349998). Initially I thought that it was maybe my snapshot process that made some mistake, so I was just trying to do newPayload + fcu for the missing blocks. This worked fine. But once I stop the reth node and start it again, I'm again at block 24349998. So it's not persisting these 2 last blocks. 

I'm running it with `--engine.persistence-threshold=0`. 

So it goes like:

1 ) Start reth, check latest block is 24349998
2) Move head to 24350000 via newPayload + fcus
3) Wait 10s
4) Stop reth (SIGTERM)
5) Start reth
6) Reth comes up with latest block at 24349998 . I was expecting 24350000. 

## Code changes

Closes blocks being lost on restart when using `--engine.persistence-threshold 0`.

The persistence service ran the pruner in the same MDBX transaction as
`save_blocks`. On a full node the first engine-triggered prune can take
60+ seconds (e.g. pruning ~1M blocks of receipts). If SIGTERM arrives
during pruning the transaction is rolled back, losing the newly persisted
blocks.

Split into two transactions: commit blocks first, then prune separately.
Blocks now survive even if the pruner is interrupted.
  
---  
This PR is AI generated using Claude code.


crates/engine/tree/src/tree/mod.rs
```diff
@@ -471,7 +471,12 @@ where
                     }
                 }
                 LoopEvent::Disconnected => {
-                    error!(target: "engine::tree", "Channel disconnected");
+                    // Persist all remaining blocks before shutting down to avoid data
+                    // loss when the incoming channel is dropped (e.g. on SIGTERM).
+                    info!(target: "engine::tree", "Channel disconnected, persisting remaining blocks before shutdown");
+                    if let Err(err) = self.persist_until_complete() {
+                        error!(target: "engine::tree", %err, "Failed to persist blocks during shutdown");
```
joshieDo: this shouldnt be necessary, we already signal the termination to engine which will  persit until completion

https://github.com/paradigmxyz/reth/blob/909157859aac9811b675dbdacbc5af81232009fe/crates/engine/tree/src/tree/mod.rs#L1445-L1451 

https://github.com/paradigmxyz/reth/blob/909157859aac9811b675dbdacbc5af81232009fe/crates/engine/tree/src/tree/mod.rs#L1309-L1320

most likely what happened was a timeout of the graceful shutdown while the task was still persisting or pruning

crates/engine/tree/src/persistence.rs
```diff
@@ -171,14 +171,18 @@ where
                 }
             }
 
+            // Commit blocks before pruning so that persisted data survives even if
+            // the pruner is interrupted (e.g. by SIGTERM).
+            provider_rw.commit()?;
```
joshieDo, line 182: makes sense

skylenet: I've simplified the PR and only kept this change.

PR #22806, opened 2026-03-05
refactor(txpool): extract generic blob lookup helper in DiskFileBlobS…

joohhnnn:
Unify `get_by_versioned_hashes_v1` and `get_by_versioned_hashes_eip7594` into a single generic `get_by_versioned_hashes_inner` method that takes a sidecar-matching closure. 

Eliminates ~100 lines of duplicated two-phase lookup logic and fixes V1 missing the O(1) missing_count early-exit optimization that had drifted in from the eip7594 variant.


mattsse (commented):
I'd prefer keeping this as is, saving a few loc isnt meaningful here

PR #22808, opened 2026-03-05
perf(primitives): cache tx hash in IndexedTx to avoid redundant keccak256

joohhnnn:
find_indexed() already computes trie_hash() to locate the transaction, but the result was discarded. Callers then immediately call meta() or tx_hash(), triggering another RLP encode + keccak256. Cache the hash in the IndexedTx struct to eliminate this redundant computation on RPC hot paths like eth_getTransactionByHash.

### DIFF
--- a/crates/primitives-traits/src/block/recovered.rs
+++ b/crates/primitives-traits/src/block/recovered.rs
@@ -318,7 +318,7 @@ impl<B: Block> RecoveredBlock<B> {
             .transactions_iter()
             .enumerate()
             .find(|(_, tx)| tx.trie_hash() == tx_hash)
-            .map(|(index, tx)| IndexedTx { block: self, tx, index })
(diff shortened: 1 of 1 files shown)


crates/primitives-traits/src/block/recovered.rs
```diff
@@ -318,7 +318,7 @@ impl<B: Block> RecoveredBlock<B> {
             .transactions_iter()
             .enumerate()
             .find(|(_, tx)| tx.trie_hash() == tx_hash)
-            .map(|(index, tx)| IndexedTx { block: self, tx, index })
+            .map(|(index, tx)| IndexedTx { block: self, tx, index, tx_hash })
```
mattsse, line 321: this is already always computed here and stored in the oncelock

mattsse (commented):
we dont need this because the triehash is cashed

PR #22812, opened 2026-03-05
perf(rpc): skip block construction in rpc_block_header

stevencartavia:
`rpc_block_header` called `rpc_block(block_id, false)` which built a full RPC block (collected tx hashes, computed ommer hashes, assembled `Block` struct) only to discard everything except `.header`.

Now calls `recovered_block` + `convert_header` directly, skipping the unnecessary work. Same inputs to `convert_header` (`clone_sealed_header()`, `rlp_length()`) as the old path through `to_rpc_block_with_tx_hashes`.


mattsse (approved):
lgtm

PR #22815, opened 2026-03-05
fix(cli): preserve jwt.hex and discovery-secret during snapshot extraction

joohhnnn:
Snapshot archives may contain `jwt.hex` and `discovery-secret` from the source node. Extracting them into an existing data directory silently overwrites the user's files, breaking Engine API authentication with the consensus layer client.

Encountered this while downloading a snapshot into a directory where Lighthouse was already connected to reth via a shared JWT secret.
After extraction, reth loaded the snapshot's JWT while Lighthouse kept using the original one:

```
ERROR engine::jwt-validator: Invalid JWT: provided signature is invalid
ERROR engine::jwt-validator: Invalid JWT: provided signature is invalid
```

The JWT mismatch produced no obvious error on the reth side — only the Lighthouse logs showed repeated auth failures, making this hard to diagnose.

Iterates tar entries during extraction and skips `jwt.hex` and `discovery-secret` if they already exist in the target directory, so existing node identity and auth credentials are never overwritten.

### DIFF
--- a/crates/cli/commands/src/download/mod.rs
+++ b/crates/cli/commands/src/download/mod.rs
@@ -947,6 +947,35 @@ impl CompressionFormat {
     }
 }
 
+/// Files that should be preserved across snapshot extraction.
+///
+/// These are node-specific identity and authentication files that, if overwritten
+/// by a snapshot archive, would silently break the connection with the consensus
+/// layer client or change the node's P2P identity.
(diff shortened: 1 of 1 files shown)


mattsse (commented):
these are excluded from our snapshotter so I'd rather rely on the snapshotter not including them

PR #22818, opened 2026-03-05
feat(reth-bench): `--reorg` flag

shekhirin:
Adds support for simulating 1-block deep chain reorgs during newPayload/FCU benchmarks.

When `--reorg` is enabled, each block height triggers: send modified payload-> FCU to make it head -> send canonical payload -> FCU to reorg back. The modified block differs only in extra_data (XOR first byte), keeping execution results identical.

### DIFF
--- a/bin/reth-bench/src/bench/new_payload_fcu.rs
+++ b/bin/reth-bench/src/bench/new_payload_fcu.rs
@@ -22,7 +22,8 @@ use crate::{
         },
     },
     valid_payload::{
-        block_to_new_payload, call_forkchoice_updated_with_reth, call_new_payload_with_reth,
+        block_to_new_payload_data, call_forkchoice_updated_with_reth, call_new_payload_with_reth,
+        create_reorg_payload,
     },
 };
 use alloy_provider::{ext::DebugApi, Provider};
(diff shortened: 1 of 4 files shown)


bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -95,6 +96,14 @@ pub struct Command {
     )]
     rpc_block_buffer_size: usize,
 
+    /// Simulate 1-block deep reorgs at each height.
+    ///
+    /// When enabled, for each block the benchmark first sends a modified (non-canonical)
+    /// payload and makes it the head via FCU, then sends the canonical payload and
+    /// switches back. This measures how the engine handles chain reorganizations.
+    #[arg(long, default_value = "false", verbatim_doc_comment)]
+    reorg: bool,
```
mattsse, line 105: should we also support more than 1 depth? so perhaps this can take an option option usize

mattsse (approved):

bin/reth-bench/src/bench/new_payload_fcu.rs
joshieDo: +1

PR #22819, opened 2026-03-05
fix(net): rate-limit "session command buffer full" debug log

joohhnnn:
The "session command buffer full, dropping message" debug log has no rate limiting — it fires on every `try_send` failure. A single slow peer can produce 500+ log lines per second, drowning out all other debug output.

Real data from a mainnet node running ~2 hours:

- 9,300 "buffer full" lines from **one** peer in 8 minutes
- Peak rate: 595 lines/second
- This single peer generated 3.5% of all log lines (2.1 MB)

This adds a per-peer 30-second cooldown: only the first occurrence per peer per 30s window is logged, subsequent drops within the window are silently counted via the existing `total_outgoing_peer_messages_dropped` metric. The rate-limit state is cleaned up when the peer disconnects.


mattsse (commented):
seems weird

PR #22820, opened 2026-03-06
perf(rpc): derive pending base fee from latest header

stevencartavia:
`eip1559_fees()` called `recovered_block(Pending)` just to read `base_fee_per_gas`, which could trigger building a full pending block from the mempool.

The pending base fee is deterministic from the parent header (EIP-1559 formula), so derive it via `chain_spec().next_block_base_fee()` instead. Matches the approach `fee_history` and `gas_price` already use.


mattsse (approved):
nice!

PR #22825, opened 2026-03-06
perf(rpc): deduplicate `pending_block_env_and_cfg` in `local_pending_block`

stevencartavia:
`local_pending_block` calls `pending_block_env_and_cfg()`, then on the `DerivedFromLatest` path calls `pool_pending_block()` which calls `pending_block_env_and_cfg()` again — repeating `provider().pending_block_and_receipts()`, `provider().latest_header()`, and `evm_config().next_evm_env()`.

Extracts the shared build logic into `build_pool_pending_block` so the already-resolved parent and evm_env are passed directly, avoiding the duplicate provider lookups.


mattsse (approved):
lgtm

PR #22826, opened 2026-03-06
perf(rpc): fetch blocks and receipts concurrently in `eth_feeHistory`

stevencartavia:
Uses a buffered stream (4 in flight) instead of sequential fetches in `eth_feeHistory`, processing each result as it arrives via while let Some.

The `expect` is safe because `blocks_and_receipts` is populated under the same `is_some_and(|p| !p.is_empty())` guard that protects the access site.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/fee.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/fee.rs
@@ -190,7 +190,22 @@ pub trait EthFees:
                 }
 
                 let chain_spec = self.provider().chain_spec();
-                for header in &headers {
+
+                // Fetch blocks and receipts concurrently if reward percentiles are
+                // requested, instead of awaiting each one sequentially in the loop.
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/fee.rs
```diff
@@ -190,7 +190,22 @@ pub trait EthFees:
...
+                            self.cache().get_block_and_receipts(header.hash()).await
+                                .map_err(Self::Error::from_eth_err)
+                                .and_then(|opt| opt.ok_or_else(|| EthApiError::InvalidBlockRange.into()))
+                        })
+                    ).await?)
+                } else {
+                    None
+                };
+
+                for (idx, header) in headers.iter().enumerate() {
```
mattsse: this isnt the best we can do here

how this should work is that this is an ordered buffered stream with restricted concurrency (e.g. geth uses 4 in parallel which seems appropriate)

so this should be a futures::stream::iter().buffered() instead here

mattsse (changes requested):

mattsse (approved):
lgtm

PR #22828, opened 2026-03-06
feat(payload): propagate tracing span across payload builder channel

yongkangc:
**Summary**
TLDR: Allows us to see the full latency breakdown in one new payload trace, making searching easier as well

**Explanation:** 
The trace ends at `send_new_payload`, the actual slow work (EVM execution, state root computation) lives in a separate orphaned trace you have to hunt for manually. This is because the mpsc channel between `send_new_payload` and `PayloadBuilderService` drops the tracing context.

**Before**
```
Trace A:
  engine::newPayload ─── send_new_payload
                              │
                         mpsc channel  ← context lost
                              │
Trace B (orphaned):           ╳
  payload_builder::new_payload_job
    └── payload_builder::poll
          └── spawn_blocking
```
Payload builder spans are disconnected — you see the request but not where time was actually spent.

**After**
```
Trace A:
  engine::newPayload ─── send_new_payload
                              │
                         mpsc channel  ← debug_span!("payload_job") carried across
                              │
                         payload_job   ← child span re-entered
                           ├── new_payload_job
                           └── poll
                                 └── spawn_blocking
```
One trace, full latency breakdown from Engine API call through EVM execution.

**Changes**
- Create a child `debug_span!` in `send_new_payload` and attach it to `BuildNewPayload`
- Store the span alongside each active job in `PayloadBuilderService`
- Re-enter it when creating jobs (`new_payload_job`) and polling them (`poll_unpin`)

### DIFF
--- a/crates/payload/builder/src/service.rs
+++ b/crates/payload/builder/src/service.rs
@@ -209,7 +209,9 @@ where
 {
     /// The type that knows how to create new payloads.
     generator: Gen,
-    /// All active payload jobs.
+    /// All active payload jobs, each accompanied by its id and the caller's tracing span
+    /// propagated across the channel so that poll and resolve work appears as children of the
+    /// original Engine API request.
     payload_jobs: Vec<(Gen::Job, PayloadId, Span)>,
     /// Copy of the sender half, so new [`PayloadBuilderHandle`] can be created on demand.
     service_tx: mpsc::UnboundedSender<PayloadServiceCommand<T>>,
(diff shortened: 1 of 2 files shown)


crates/payload/builder/src/service.rs
```diff
@@ -483,6 +490,7 @@ pub enum PayloadServiceCommand<T: PayloadTypes> {
     /// Start building a new payload.
     BuildNewPayload(
         T::PayloadBuilderAttributes,
+        Span,
```
pepyakin, line 498: When the variant is a tuple, it's good to explain in the comment the values. In this case we could say what is the span we are carrying here.

crates/payload/builder/src/service.rs
```diff
@@ -209,7 +210,7 @@ where
     /// The type that knows how to create new payloads.
     generator: Gen,
     /// All active payload jobs.
-    payload_jobs: Vec<(Gen::Job, PayloadId)>,
+    payload_jobs: Vec<(Gen::Job, PayloadId, Span)>,
```
pepyakin, line 215: Also docs.

yongkangc: ahh good point, thanks just added 👍🏻

mediocregopher (approved):

PR #22831, opened 2026-03-06
perf(payload): move sealed block instead of cloning

decofe:
## Summary
Move owned `RecoveredBlock` into `Arc<SealedBlock>` instead of cloning via `.sealed_block().clone()`.

## Motivation
`default_ethereum_payload` runs every ~1s during a 12s slot. The `block` returned by `builder.finish()` is owned and not referenced after the sealed block extraction — the clone is unnecessary overhead.

Co-Authored-By: YK <46377366+yongkangc@users.noreply.github.com>

Prompted by: yk


yongkangc (approved):

mattsse (approved):
easy

PR #22843, opened 2026-03-06
perf(txpool): skip listener locks when no subscribers exist

MagicJoshh:
`pending_transaction_listener` and `transaction_listener` grab a read lock on every new transaction even when nobody is listening. The same struct already does the `AtomicBool` guard trick for `event_listener` 

Extends it to the other two listener types.

Mostly helps during sync when there are no RPC subscribers.


mattsse (commented):
we can ignore these because we expect that we'll always have at least one listener here, (network)

PR #22846, opened 2026-03-06
perf(engine): use realistic avg code size for cache budget estimation

joohhnnn:
The code cache budget uses MAX_CODE_SIZE (48 KiB) to estimate per-entry memory, but this is the EIP-7907 upper bound — mainnet contracts average ~7 KiB. The 7x overestimate causes bytes_to_entries to allocate only 4096 entries for a 228 MB budget.

This replaces the worst-case constant with ESTIMATED_AVG_CODE_SIZE (8 KiB), a value based on observed mainnet contract size distribution.

With the default 4 GB cross-block cache (5.56% = 228 MB for code):

| | Before | After |
|---|---|---|
| Entry size estimate | 49,280 B | 8,320 B |
| Capacity | 4,096 | 16,384 |
| Estimated memory | ~30 MB (13%) | ~117 MB (51%) |

Worst case (all 16,384 entries at MAX_CODE_SIZE): ~768 MB. This is unlikely on mainnet where average is ~7 KiB, and users on constrained machines can reduce the budget via --cross-block-cache-size.

Confirmed on a mainnet node: code cache was 100% full at 4096 entries with 128 collisions and 89% hit rate.

A capacity assertion test is added to guard against accidental regressions to the old estimate.

### DIFF
--- a/crates/engine/tree/src/tree/cached_state.rs
+++ b/crates/engine/tree/src/tree/cached_state.rs
@@ -18,7 +18,6 @@ use reth_trie::{
     updates::TrieUpdates, AccountProof, HashedPostState, HashedStorage, MultiProof,
     MultiProofTargets, StorageMultiProof, StorageProof, TrieInput,
 };
-use revm_primitives::eip7907::MAX_CODE_SIZE;
 use std::{
     mem::size_of,
     sync::{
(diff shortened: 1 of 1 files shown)


Rjected (approved):
makes sense, math checks out + memory impact doesn't seem too bad

yongkangc (approved):

PR #22849, opened 2026-03-06
perf(net): size-based backpressure for session broadcast messages

figtracer:
The `session command buffer full, dropping message` spam happens because backpressure is count-based: 100 messages with 1 tx hash each fill the channel just as fast as 1 message with 100 hashes. This switches to size-based backpressure that tracks the total number of buffered broadcast items (tx hashes, transactions, blocks).

cc: @mattsse 
#22824

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -6,7 +6,10 @@ use std::{
     future::Future,
     net::SocketAddr,
     pin::Pin,
-    sync::{atomic::AtomicU64, Arc},
+    sync::{
+        atomic::{AtomicU64, AtomicUsize},
+        Arc,
+    },
     task::{ready, Context, Poll},
     time::{Duration, Instant},
 };
@@ -28,7 +31,7 @@ use reth_eth_wire::{
     message::{EthBroadcastMessage, MessageError},
     Capabilities, DisconnectP2P, DisconnectReason, EthMessage, NetworkPrimitives, NewBlockPayload,
 };
-use reth_eth_wire_types::{message::RequestPair, RawCapabilityMessage};
+use reth_eth_wire_types::{message::RequestPair, NewPooledTransactionHashes, RawCapabilityMessage};
 use reth_metrics::common::mpsc::MeteredPollSender;
 use reth_network_api::PeerRequest;
 use reth_network_p2p::error::RequestError;
@@ -37,7 +40,7 @@ use reth_network_types::session::config::INITIAL_REQUEST_TIMEOUT;
 use reth_primitives_traits::Block;
 use rustc_hash::FxHashMap;
 use tokio::{
-    sync::{mpsc::error::TrySendError, oneshot},
+    sync::{mpsc, mpsc::error::TrySendError, oneshot},
     time::Interval,
 };
 use tokio_stream::wrappers::ReceiverStream;
@@ -76,6 +79,13 @@ const TIMEOUT_SCALING: u32 = 3;
 /// before reading any more messages from the remote peer, throttling the peer.
 const MAX_QUEUED_OUTGOING_RESPONSES: usize = 4;
 
+/// Soft limit for the total number of buffered outgoing broadcast items (e.g. transaction hashes).
+///
+/// Many small broadcast messages carrying a single tx hash each are equivalent in cost to one
+/// message carrying many hashes. This limit counts individual items (hashes, transactions, blocks)
+/// rather than messages, so that many small messages don't trigger aggressive drops unnecessarily.
+pub(super) const MAX_QUEUED_BROADCAST_ITEMS: usize = 4096;
+
 /// The type that advances an established session by listening for incoming messages (from local
 /// node or read from connection) and emitting events back to the
 /// [`SessionManager`](super::SessionManager).
@@ -101,6 +111,8 @@ pub(crate) struct ActiveSession<N: NetworkPrimitives> {
     pub(crate) session_id: SessionId,
     /// Incoming commands from the manager
     pub(crate) commands_rx: ReceiverStream<SessionCommand<N>>,
+    /// Overflow channel for broadcast messages that couldn't fit in the bounded command channel.
+    pub(crate) broadcast_overflow_rx: mpsc::UnboundedReceiver<PeerMessage<N>>,
     /// Sink to send messages to the [`SessionManager`](super::SessionManager).
     pub(crate) to_session_manager: MeteredPollSender<ActiveSessionMessage<N>>,
     /// A message that needs to be delivered to the session manager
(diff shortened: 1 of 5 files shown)


crates/net/network/src/session/active.rs
```diff
@@ -894,6 +904,52 @@ impl<N: NetworkPrimitives> OutgoingMessage<N> {
...
+                EthBroadcastMessage::NewBlock(_) => 1,
+                EthBroadcastMessage::Transactions(txs) => txs.len(),
+            },
+            Self::Raw(_) => 0,
+        }
+    }
+
+    /// Tries to merge pooled transaction hash announcements into this message if it is the same
+    /// variant (eth66 or eth68).
+    fn try_merge_hashes(&mut self, incoming: &NewPooledTransactionHashes) -> bool {
```
mattsse: I dont like that this effectively copies, when it should consume, and return `Option<NewPooledTransactionHashes>` if they're not the same version

crates/net/network/src/session/mod.rs
```diff
@@ -372,13 +375,23 @@ impl<N: NetworkPrimitives> SessionManager<N> {
...
+    /// session task. Many small messages carrying a single tx hash each are treated equivalently
+    /// to one large message carrying many hashes.
     pub fn send_message(&self, peer_id: &PeerId, msg: PeerMessage<N>) {
         if let Some(session) = self.active_sessions.get(peer_id) {
+            if msg.is_broadcast() && session.queued_broadcast_items() >= MAX_QUEUED_BROADCAST_ITEMS
+            {
+                self.metrics.total_outgoing_peer_messages_dropped.increment(1);
+                return;
+            }
             let _ = session.commands_to_session.try_send(SessionCommand::Message(msg)).inspect_err(
```
mattsse: this doesnt solve the problem entirely because even if the session has capacity, this will still run into the limitations of the bounded `commands_to_session` buffer

so what we should do here instead is, flip this so that we do

1. try_send
2. if err(full) && broadcast && session.queued_broadcast_items() < MAX_QUEUED_BROADCAST_ITEMS try again via `session.commands_to_session.clone().try_send`


but what this doesnt solve is: keeping track of how many are in the channel already

so we ideally need our own sender+receiver wrapper for this

mattsse (changes requested):
this is pretty good already, however it's not quite complete yet,

because this only takes into account the secondary buffer (the outgoing buffer in the session itself, not the channel buffer)

crates/net/network/src/session/mod.rs
figtracer: oh i see, this makes sense

crates/net/network/src/session/handle.rs
```diff
@@ -154,6 +167,106 @@ impl<N: NetworkPrimitives> ActiveSessionHandle<N> {
...
+        tx: mpsc::Sender<SessionCommand<N>>,
+        broadcast_overflow_tx: mpsc::UnboundedSender<PeerMessage<N>>,
+        broadcast_items: Arc<AtomicUsize>,
+    ) -> Self {
+        Self { tx, broadcast_overflow_tx, broadcast_items }
+    }
+
+    /// Sends a disconnect command, cloning the sender to reserve capacity.
+    pub(crate) fn disconnect(&self, reason: Option<DisconnectReason>) {
+        let _ = self.tx.clone().try_send(SessionCommand::Disconnect { reason });
```
mattsse: this should now also use the overflow channel, then we should actually rename the field to smth like `unbounded_tx` or smth

crates/net/network/src/session/handle.rs
```diff
@@ -154,6 +167,106 @@ impl<N: NetworkPrimitives> ActiveSessionHandle<N> {
...
+    }
+
+    /// Sends a disconnect command, awaiting channel capacity.
+    pub(crate) async fn try_disconnect(
+        &self,
+        reason: Option<DisconnectReason>,
+    ) -> Result<(), SendError<SessionCommand<N>>> {
+        self.tx.clone().send(SessionCommand::Disconnect { reason }).await
```
mattsse: same here

crates/net/network/src/session/active.rs
```diff
@@ -919,15 +986,22 @@ fn calculate_new_timeout(current_timeout: Duration, estimated_rtt: Duration) ->
...
+///
+/// Tracks both the total number of queued messages (via a metric gauge) and the total number of
+/// broadcast items (tx hashes, transactions, blocks) via a shared atomic counter. The atomic
+/// counter is shared with [`ActiveSessionHandle`](super::handle::ActiveSessionHandle) so the
+/// [`SessionManager`](super::SessionManager) can apply size-based backpressure.
 pub(crate) struct QueuedOutgoingMessages<N: NetworkPrimitives> {
     messages: VecDeque<OutgoingMessage<N>>,
     count: Gauge,
+    /// Shared counter of buffered broadcast items for size-based backpressure.
+    broadcast_items: Arc<AtomicUsize>,
```
mattsse: could we wrap this type into a new type perhaps, because maybe we need to change this later to track memory as well and it would be nice to not have to change everything from Arc AtomicUisze

crates/net/network/src/session/active.rs
```diff
@@ -946,7 +1041,6 @@ impl<N: NetworkPrimitives> QueuedOutgoingMessages<N> {
 
 impl<N: NetworkPrimitives> Drop for QueuedOutgoingMessages<N> {
     fn drop(&mut self) {
-        // Ensure gauge is decremented for any remaining items to avoid metric leak on teardown.
```
mattsse, line 955: was this wrong?

crates/net/network/src/session/mod.rs
```diff
@@ -551,20 +550,24 @@ impl<N: NetworkPrimitives> SessionManager<N> {
                     interval
                 });
 
+                let broadcast_items = Arc::new(AtomicUsize::new(0));
```
mattsse: I want to see some docs here what the invariants for this types are: that the session must always update this once it has processed broadcasts etc

mattsse (changes requested):
this makes sense,

I only have smol suggestions

crates/net/network/src/session/active.rs
```diff
@@ -364,7 +413,7 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
             }
             PeerMessage::PooledTransactions(msg) => {
                 if msg.is_valid_for_version(self.conn.version()) {
-                    self.queued_outgoing.push_back(EthMessage::from(msg).into());
+                    self.queued_outgoing.push_pooled_hashes(msg);
                 } else {
                     debug!(target: "net", ?msg,  version=?self.conn.version(), "Message is invalid for connection version, skipping");
                 }
```
mattsse, line 422: there's one issue here in case this happens, we need to cleanup the tracker because this bypasses `BroadcastItemCounter::sub`

mattsse (changes requested):
last tiny nit

mattsse (approved):

PR #22851, opened 2026-03-06
fix(cli): fix ctrl-C in reth downloads

Rjected:
This adds cancellation so that ctrl-C properly aborts downloads

### DIFF
--- a/crates/cli/commands/src/download/mod.rs
+++ b/crates/cli/commands/src/download/mod.rs
@@ -17,6 +17,7 @@ use manifest::{
 use reqwest::{blocking::Client as BlockingClient, header::RANGE, Client, StatusCode};
 use reth_chainspec::{EthChainSpec, EthereumHardfork, EthereumHardforks};
 use reth_cli::chainspec::ChainSpecParser;
+use reth_cli_util::cancellation::CancellationToken;
 use reth_db::{init_db, Database};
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/download/mod.rs
```diff
@@ -1395,20 +1458,30 @@ async fn process_modular_archive(
...
-    .await??;
+    });
+
+    tokio::select! {
+        result = handle => result??,
+        _ = tokio::signal::ctrl_c() => {
+            cancelled.store(true, Ordering::Relaxed);
+            eyre::bail!("Download cancelled");
+        }
+    }
```
mattsse: isnt this command spawned via the regular clirunner?

th type should already do this and then terminate this task, so we should get the same if we fire cancelled on drop here

we also have this:

https://github.com/paradigmxyz/reth/blob/5b0297551a2255e0cf89229704665d8673cd8986/crates/cli/util/src/cancellation.rs#L23-L35

which comes with a dropguard

mattsse (changes requested):

crates/cli/commands/src/download/mod.rs
Rjected: good point, the token type seems cleaner

mattsse (approved):

PR #22856, opened 2026-03-07
perf(trie): deduplicate prefix set computation

stevencartavia:
Avoid redundant keccak256 hashing and DB cursor seeks when building trie prefix sets by deduplicating accounts via a HashSet and caching the last hashed address for consecutive storage changeset entries.

### DIFF
--- a/crates/trie/db/src/prefix_set.rs
+++ b/crates/trie/db/src/prefix_set.rs
@@ -1,7 +1,7 @@
 use alloy_primitives::{
     keccak256,
     map::{HashMap, HashSet},
-    BlockNumber, B256,
+    Address, BlockNumber, B256,
 };
 use core::ops::RangeInclusive;
 use reth_db_api::{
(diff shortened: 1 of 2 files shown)


mattsse (commented):
with keccak caching this likely doesnt matter

wdyt @mediocregopher @DaniPopes

crates/trie/db/src/prefix_set.rs
```diff
@@ -39,20 +39,28 @@ where
     // We still need direct access to HashedAccounts table
     let mut account_hashed_state_cursor = tx.cursor_read::<tables::HashedAccounts>()?;
 
+    let mut seen_accounts = HashSet::new();
```
DaniPopes: pls use addressset

stevencartavia: updated, I used HashSet to mirror this one https://github.com/paradigmxyz/reth/blob/main/crates/trie/db/src/state.rs#L308
lmk if you want this one updated to AddressSet too

DaniPopes: yes

DaniPopes (approved):

PR #22859, opened 2026-03-07
feat(download): use snapshots.reth.rs API with --list

figtracer:
Switch default snapshot source from merkle.io to `snapshots.reth.rs`. Manifest discovery queries the snapshots API.

- `--list`: browse available snapshots with manifest URLs for use with `--manifest-url`

### DIFF
--- a/crates/cli/commands/src/download/mod.rs
+++ b/crates/cli/commands/src/download/mod.rs
@@ -43,7 +43,8 @@ use url::Url;
 use zstd::stream::read::Decoder as ZstdDecoder;
 
 const BYTE_UNITS: [&str; 4] = ["B", "KB", "MB", "GB"];
-const MERKLE_BASE_URL: &str = "https://downloads.merkle.io";
+const RETH_SNAPSHOTS_BASE_URL: &str = "https://snapshots-r2.reth.rs";
+const RETH_SNAPSHOTS_API_URL: &str = "https://snapshots.reth.rs/api/snapshots";
 const EXTENSION_TAR_LZ4: &str = ".tar.lz4";
 const EXTENSION_TAR_ZSTD: &str = ".tar.zst";
 const DOWNLOAD_CACHE_DIR: &str = ".download-cache";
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/download/mod.rs
```diff
@@ -602,16 +627,24 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> DownloadCo
...
+                let profile = if self.archive {
+                    Some("archive")
+                } else if self.full {
+                    Some("full")
+                } else if self.minimal {
+                    Some("minimal")
+                } else {
+                    None
+                };
+                discover_manifest_url(chain_id, profile, &self.channel).await
```
Rjected: iirc we should only have one manifest url (generated from an archive node) for a given snapshot, the idea behind the modular manifest is that you can request different things based on whether you want minimal, full, archive. So we shouldnt have to select based on desired profile.

crates/cli/commands/src/download/mod.rs
```diff
@@ -1511,15 +1544,146 @@ fn file_blake3_hex(path: &Path) -> Result<String> {
...
+    use serde::Deserialize;
+    let value = serde_json::Value::deserialize(deserializer)?;
+    match &value {
+        serde_json::Value::Number(n) => {
+            n.as_u64().ok_or_else(|| serde::de::Error::custom("expected u64"))
+        }
+        serde_json::Value::String(s) => {
+            s.parse::<u64>().map_err(|_| serde::de::Error::custom("expected numeric string"))
+        }
+        _ => Err(serde::de::Error::custom("expected number or string")),
```
Rjected, line 1636: is this because we have both legacy manifests and modular manifests in r2? I would prefer we don't deserialize the legacy manifests at all

Rjected (commented):
have a few questions about this, we should definitely update to the reth snapshots as the default

crates/cli/commands/src/download/mod.rs
figtracer: oh i see, misunderstood this completely.. will fix

crates/cli/commands/src/download/mod.rs
figtracer: 👍🏼 will simplify this

crates/cli/commands/src/download/mod.rs
```diff
@@ -1572,13 +1594,114 @@ fn file_blake3_hex(path: &Path) -> Result<String> {
...
-fn get_base_url(chain_id: u64) -> String {
-    let defaults = DownloadDefaults::get_global();
-    match &defaults.default_chain_aware_base_url {
-        Some(url) => format!("{url}/{chain_id}"),
-        None => defaults.default_base_url.to_string(),
+/// Discovers the latest snapshot manifest URL for the given chain from the snapshots API.
+///
+/// Queries `snapshots.reth.rs/api/snapshots` and returns the manifest URL for the most
+/// recent modular snapshot matching the requested chain.
+async fn discover_manifest_url(chain_id: u64, channel: &str) -> Result<String> {
```
Rjected: actually i don't think we need channel at all. edge was the previous name for the db format that the modular snapshots use. so I don't think we should even support stable - id remove the cli flag and all channel selection logic

Rjected:
btw am getting this when running reth download:
```
2026-03-10T18:50:13.569289Z  INFO Discovering latest snapshot manifest api_url=https://snapshots.reth.rs/api/snapshots chain_id=1 channel=stable
Error: error decoding response body

Caused by:
    invalid type: string "1", expected u64 at line 1 column 15

Location:
    /Users/dancline/projects/reth/crates/cli/commands/src/download/mod.rs:1663:42
    
```

crates/cli/commands/src/download/mod.rs
figtracer: makes sense

Rjected:
whoops had some cache issues

Rjected (approved):
lgtm, thanks!

PR #22860, opened 2026-03-07
fix(net): remove redundant PendingPoolImportsInfo allocation in TransactionsManager

bobtajson:
`PendingPoolImportsInfo` was created twice with identical parameters in `TransactionsManager::with_policy` - once for a metrics read, then discarded and recreated for the struct field. Reuse the single instance instead


mattsse (approved):
lgtm

PR #22861, opened 2026-03-07
fix(rpc): exclude Testing module from default IPC modules

MagicJoshh:
`Testing` was listed in `STANDARD_VARIANTS`, causing `default_ipc_modules()` to include it via `all_modules()`. Since IPC is enabled by default, the `testing_buildBlockV1` endpoint was silently registered on every node startup without any user configuration — contradicting the comment in `node.rs` that states it should only be wired "when the hidden testing module is explicitly requested on any transport".

Removes `Testing` from `STANDARD_VARIANTS` so it is excluded from `All`, `all_modules()`, and `default_ipc_modules()`. Users can still explicitly enable it via CLI flags (e.g. `--http.api testing`). `FromStr` parsing is unaffected.


mattsse (commented):
this will likely break some setups, I think its fine if its enabled via all

PR #22862, opened 2026-03-07
fix(txpool): treat NotFound as success in blob store cleanup

joohhnnn:
Observed on a mainnet full node: ~30 "No such file or directory" errors per day from blob store cleanup, with `blobstore_failed_deletes` metric steadily increasing. The errors come from concurrent cleanup tasks (e.g. triggered by both `on_canonical_state_change` and the periodic maintenance task) racing to delete the same expired blob files — the second task fails with `NotFound` for files already removed by the first.

Since the desired outcome (file removed) is already achieved, `NotFound` should count as a successful delete rather than a failure. This avoids inflating the `blobstore_failed_deletes` metric and eliminates spurious debug log noise.

Example log line: `[<tx_hash>] failed to delete blob file at /data/blobstore/<hash>.blob: No such file or directory (os error 2)`

### DIFF
--- a/crates/transaction-pool/src/blobstore/disk.rs
+++ b/crates/transaction-pool/src/blobstore/disk.rs
@@ -179,6 +179,10 @@ impl BlobStore for DiskFileBlobStore {
                     stat.delete_succeed += 1;
                     subsize += filesize;
                 }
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah this makes sense, very nice ty

PR #22867, opened 2026-03-08
refactor(txpool): use ValidPoolTransaction delegate accessors

esorense:
This PR replaces direct `.transaction` field access with `ValidPoolTransaction` delegate methods in txpool hot paths (`best.rs`, `pool/mod.rs`, `validate/mod.rs`), including fee checks, blob checks, sender/gas access, and hash access.

  Why this is needed: these checks should run through the validated transaction API boundary, not internal fields. Keeping access at the delegate layer reduces coupling, prevents refactor-prone call sites, and avoids subtle type/API misuse.

  Correctness: behavior and thresholds are unchanged; only access paths were normalized. Validation evidence: `cargo +nightly fmt --all --check` passed, `cargo +1.93.1 test -p reth-transaction-pool --all-targets` passed (`225 + 1` tests), and `cargo +1.93.1 clippy -p reth-transaction-pool --all-  targets --locked -- -D warnings` passed.

### DIFF
--- a/crates/transaction-pool/src/pool/best.rs
+++ b/crates/transaction-pool/src/pool/best.rs
@@ -4,7 +4,6 @@ use crate::{
     pool::pending::PendingTransaction,
     PoolTransaction, Priority, TransactionOrdering, ValidPoolTransaction,
 };
-use alloy_consensus::Transaction;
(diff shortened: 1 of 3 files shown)


mattsse (commented):
we can keep as is

PR #22874, opened 2026-03-08
fix(rpc): return "pruned history unavailable" instead of "header not found" for pruned blocks

joohhnnn:
Running a mainnet full node initialized from a snapshot (starting around block 0x1770FA0). When calling `trace_block` or `debug_traceBlockByNumber` on historical blocks before the snapshot start, the response is `"header not found"`, which is misleading — the block existed on mainnet but its data was pruned per the node's retention policy.

Reproduction on a snapshot-synced node:
$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"trace_block","params":["0xF4240"],"id":1}' http://127.0.0.1:8545
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"header not found"}}

The expected response should indicate that the history was pruned, not that the header doesn't exist. reth already has `EthApiError::PrunedHistoryUnavailable` (EIP-4444, error code 4444) used by `eth_getLogs`, but `trace_*` and `debug_*` methods weren't using it.

This adds a `header_or_pruned_error` helper to both `TraceApi` and `DebugApi` that checks whether the requested block number is below `earliest_block_number()`, returning the appropriate error variant.

### DIFF
--- a/crates/rpc/rpc-eth-types/src/error/mod.rs
+++ b/crates/rpc/rpc-eth-types/src/error/mod.rs
@@ -1208,6 +1208,25 @@ mod tests {
     }
 
     #[test]
+    fn block_expired_converts_to_pruned_history_unavailable() {
+        // ProviderError::BlockExpired → EthApiError::PrunedHistoryUnavailable
+        let provider_err = reth_errors::ProviderError::BlockExpired {
+            requested: 100,
+            earliest_available: 1_000_000,
+        };
+        let eth_err: EthApiError = provider_err.into();
+        assert!(
+            matches!(eth_err, EthApiError::PrunedHistoryUnavailable),
+            "expected PrunedHistoryUnavailable, got: {eth_err:?}"
+        );
+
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
i dont think we should check for this on the callsite because then we'd need to do this everywehre, 
can we instead add this in the providerror and lookup isntead

joohhnnn:
> i dont think we should check for this on the callsite because then we'd need to do this everywehre, can we instead add this in the providerror and lookup isntead

Moved the check into `BlockIdReader::block_hash_for_id` — when `block_hash(num)` returns `None` and `num < earliest_block_number()`, it now returns `Err(ProviderError::BlockExpired)`. The existing `From<ProviderError> for EthApiError` conversion handles the rest, so no per-callsite checks needed.

Added regression tests for both the provider-level error and the RPC error code (4444) conversion.

PR #22875, opened 2026-03-08
perf(engine): hoist outer map lookups out of per-slot loops

joohhnnn:
In `on_hashed_state_update` and `on_prewarm_targets`, the outer `B256Map` lookups for `new_storage_updates[address]` and `storage_updates[address]` were performed inside the per-slot inner loop, even though the address is constant within each iteration of the outer loop. This caused redundant hashing and SwissTable probing for every storage slot.

Profiling a mainnet full node with `perf record` showed ~55% of engine thread CPU in SwissTable probing (`pcmpeqb`/`pmovmskb`/`tzcnt` loops), with cache misses on B256 bucket key loads as the bottleneck. On typical mainnet blocks (~200 addresses, ~1000 storage slots), this eliminates ~1600 redundant outer-map lookups per block (~80% reduction in outer-map probes).

Fix by destructuring `self` to allow simultaneous mutable borrows of disjoint fields, hoisting the outer-map `entry()`/`get_mut()` calls to once per address rather than once per slot.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -384,13 +384,11 @@ where
         }
 
         for (address, slots) in targets.storage_targets {
+            // Look up outer map once per address instead of once per slot.
+            let new_updates = self.new_storage_updates.entry(address).or_default();
             for slot in slots {
                 // Only touch storages that are not yet present in the updates set.
-                self.new_storage_updates
-                    .entry(address)
-                    .or_default()
-                    .entry(slot.key())
-                    .or_insert(LeafUpdate::Touched);
+                new_updates.entry(slot.key()).or_insert(LeafUpdate::Touched);
             }
 
             // Touch corresponding account leaf to make sure its revealed in accounts trie for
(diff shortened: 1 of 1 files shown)


Rjected (approved):
lgtm

yongkangc (dismissed):
gotta figure out why theres a regression

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -406,45 +406,61 @@ where
         skip_all
     )]
     fn on_hashed_state_update(&mut self, hashed_state_update: HashedPostState) {
-        for (address, storage) in hashed_state_update.storages {
-            for (slot, value) in storage.storage {
-                self.trie.record_slot_touch(address, slot);
+        // Destructure to allow simultaneous mutable borrows of disjoint fields, so we can
+        // hoist per-address outer-map lookups out of the per-slot inner loop.
```
DaniPopes: i don't think this is necessary, can you please verify this by using `self` as before?

joohhnnn: Verified, compiles fine with `self` — removed the destructuring.

Also added `!is_empty()` guards for the hoisted lookups to fix the bench panics (empty slots/storage creating entries in `new_storage_updates` → unrevealed trie panic in `compute_drained_storage_roots`).

DaniPopes (approved):

PR #22878, opened 2026-03-08
feat(txpool): announce blob txs within 1 fee-jump of execution

figtracer:
Blob transactions in the blob sub-pool were never announced to peers, only promoted txs in the pending pool were. This adds early announcements for blob txs with priority >= -1 (within 1 fee-jump of executable), giving peers earlier visibility on near-executable blob txs.

When fees change (new block), txs that cross into the threshold get announced. If fees spike and a tx falls below threshold, its announced flag resets so it can be re-announced when fees recover.

Reference: https://github.com/ethereum/go-ethereum/pull/33893

### DIFF
--- a/crates/transaction-pool/src/pool/blob.rs
+++ b/crates/transaction-pool/src/pool/blob.rs
@@ -9,6 +9,14 @@ use std::{
     sync::Arc,
 };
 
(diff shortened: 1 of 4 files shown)


figtracer:
might not be that relevant in practice but imo this is nice to keep things tidy

PR #22889, opened 2026-03-09
feat(engine): add --engine.proof-jitter option behind trie-debug

mediocregopher:
Adds a random sleep (up to the configured duration) before each proof computation in both storage and account workers. Feature-gated behind trie-debug so there is zero overhead in normal builds. Useful for shaking out race-conditions in the SparseTrieCacheTask.

### DIFF
--- a/crates/trie/parallel/src/proof_task.rs
+++ b/crates/trie/parallel/src/proof_task.rs
@@ -723,9 +723,8 @@ where
 
             #[cfg(feature = "trie-debug")]
             if let Some(max_jitter) = self.task_ctx.proof_jitter {
-                let jitter = Duration::from_nanos(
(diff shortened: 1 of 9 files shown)


crates/trie/parallel/src/proof_task.rs
```diff
@@ -980,6 +1008,20 @@ where
             // Mark worker as busy.
             self.available_workers.fetch_sub(1, Ordering::Relaxed);
 
+            #[cfg(feature = "trie-debug")]
+            if let Some(max_jitter) = self.task_ctx.proof_jitter {
+                let jitter = Duration::from_nanos(
+                    rand::random::<u64>() % (max_jitter.as_nanos() as u64 + 1),
```
pepyakin: Nit: `rand::random_range`? +beauty -bias. Ditto above.

pepyakin (approved):

PR #22893, opened 2026-03-09
fix(rpc): disable EIP-7825 tx gas limit cap in eth_createAccessList 

joohhnnn:
Closes #22841

Since Osaka activated on mainnet (Dec 2025), EIP-7825 enforces a per-tx gas limit cap of 2^24 (~16.7M). When `gas` is omitted from `eth_createAccessList`, alloy-evm falls back to `block_env.gas_limit()` (~36M), which exceeds the cap → revm rejects with `TxGasLimitGreaterThanCap` → `"intrinsic gas too high"`. The same cap also limits `eth_estimateGas` binary search upper bound to ~16.7M.

`prepare_call_env` (used by `eth_call`) already disables this via `tx_gas_limit_cap = Some(u64::MAX)`. 

applies the same override to `create_access_list_with` and `estimate_gas_with`.

Verified on a mainnet node — the EIP-7825 boundary is exact: `gas=0x1000000` (2^24) passes, `gas=0x1000001` (2^24+1) fails before the fix, both pass after.

Sorry about the incorrect root cause in #22841

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -491,6 +491,11 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
             // Disabled because eth_createAccessList is sometimes used with non-eoa senders
             evm_env.cfg_env.disable_eip3607 = true;
 
+            // Disable EIP-7825 transaction gas limit cap so that the gas limit
+            // fallback (block gas limit) is not rejected when it exceeds the
+            // per-tx cap (2^24 ≈ 16.7M post-Osaka).
+            evm_env.cfg_env.tx_gas_limit_cap = Some(u64::MAX);
+
             if request.as_ref().gas_limit().is_none() && tx_env.gas_price() > 0 {
                 let cap = this.caller_gas_allowance(&mut db, &evm_env, &tx_env)?;
                 // no gas limit was provided in the request, so we need to cap the request's gas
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
The `eth_createAccessList` change looks correct — Geth also bypasses the EIP-7825 cap there via `SkipTransactionChecks: true`.

However, the `eth_estimateGas` change diverges from Geth. Setting `tx_gas_limit_cap = Some(u64::MAX)` removes the cap entirely, so the binary search upper bound becomes `block_gas_limit` (~36M) via:

```rust
let max_gas_limit = evm_env.cfg_env.tx_gas_limit_cap.map_or_else(
    || evm_env.block_env.gas_limit(),
    |cap| cap.min(evm_env.block_env.gas_limit()),
);
```

This means `eth_estimateGas` could return values > 2^24 that can't actually be submitted as transactions post-Osaka.

Geth handles this differently in `gasestimator.go` — it explicitly caps the search ceiling to `MaxTxGas` for Osaka blocks:

```go
if hi > params.MaxTxGas {
    // ...check if Osaka is active...
    if opts.Config.IsOsaka(blockNumber, blockTime) {
        hi = params.MaxTxGas
    }
}
```

So Geth ensures the returned estimate is always a valid transaction gas limit. The `tx_gas_limit_cap` should remain at the EIP-7825 value for `eth_estimateGas` — it's already correctly wired into the binary search via `max_gas_limit`.

mattsse (dismissed):
The `eth_createAccessList` change is correct.

The `eth_estimateGas` change should not disable the cap — Geth explicitly caps the search ceiling to `MaxTxGas` for Osaka blocks in `gasestimator.go`:

```go
if hi > params.MaxTxGas {
    if opts.Config.IsOsaka(blockNumber, blockTime) {
        hi = params.MaxTxGas
    }
}
```

With `tx_gas_limit_cap = Some(u64::MAX)`, `eth_estimateGas` could return values > 2^24 that are invalid for real transactions post-Osaka. The cap is already correctly wired into the binary search via `max_gas_limit`, so it should stay.

mattsse (changes requested):
The `eth_createAccessList` change is correct.

The `eth_estimateGas` change should not disable the cap. With `tx_gas_limit_cap = Some(u64::MAX)`, `eth_estimateGas` could return values > 2^24 that are invalid for real transactions post-Osaka. The cap is already correctly wired into the binary search via `max_gas_limit`, so it should stay.

mattsse (approved):

PR #22900, opened 2026-03-09
perf(engine): replace moka with schnellru for precompile cache and add per-precompile sizing

joohhnnn:
Replaces `moka::sync::Cache` with `schnellru::LruMap` for the precompile cache and sizes each precompile's cache based on its actual working-set.

**Motivation**

Production metrics from a mainnet node (block ~22.1M) show that the uniform 10K capacity starves high-throughput precompiles:

| Precompile | Capacity | Entries | Hit Rate |
|-----------|----------|---------|----------|
| ecMul (0x07) | 10,000 | 9,993 | 55.0% |
| modexp (0x05) | 10,000 | 9,999 | 71.6% |
| ecAdd (0x06) | 10,000 | 10,000 | 70.7% |
| SHA256 (0x02) | 10,000 | 18 | 96.3% |

**Changes**

1. **moka → schnellru** (`refactor`): `moka` was the only user in this crate. schnellru is lighter (no background threads) and already used in the same crate (`invalid_headers.rs`). Wrapped in `Arc<Mutex<..>>` to match the shared-cache semantics.

2. **Per-precompile sizing** (`perf`): ecMul → 50K, ecAdd/modexp → 30K, others → 10K. ~6MB additional memory for ecMul at 62GB RSS.

3. **Fix hit metric** (`fix`): `precompile_cache_hits` was incremented on cache lookup even when gas was insufficient, inflating the reported hit rate.


crates/engine/tree/src/tree/precompile_cache.rs
```diff
@@ -34,42 +64,59 @@ where
...
+type PrecompileLruMap<S> = LruMap<Vec<u8>, CacheEntry<S>, ByLength>;
+
 /// Cache for precompiles, for each input stores the result.
-#[derive(Debug, Clone)]
-pub struct PrecompileCache<S>(moka::sync::Cache<Bytes, CacheEntry<S>, DefaultHashBuilder>)
+///
+/// Internally backed by [`schnellru::LruMap`] behind an `Arc<Mutex<..>>` so that multiple
+/// [`CachedPrecompile`] instances for the same address share state.
+#[derive(Clone)]
+pub struct PrecompileCache<S>(Arc<Mutex<PrecompileLruMap<S>>>)
```
DaniPopes, line 56: let's try with RwLock

joohhnnn: Sure. switch from Mutex to RwLock and use `peek` (shared lock) for cache lookups instead of `get` (exclusive lock). Inserts still take an exclusive write lock.  allows concurrent cache reads without blocking.

crates/engine/tree/src/tree/precompile_cache.rs
```diff
@@ -1,30 +1,60 @@
...
+/// - Others: default 10K (sufficient)
+fn default_cache_size_for_address(address: &Address) -> u32 {
+    match address.0[19] {
+        // ecMul — largest working set
+        0x07 => 50_000,
+        // modexp, ecAdd
+        0x05 | 0x06 => 30_000,
+        _ => DEFAULT_CACHE_SIZE,
+    }
+}
```
klkvr: i believe this shouldn't really have any impact. most of the hitrate is coming from prewarming and cross-block cache size shouldn't matter likely

joohhnnn: yaeh, prewarm already runs the same txs through CachedPrecompile so per-block cache is warm regardless of capacity. Dropped the sizing.

joohhnnn:
cc @DaniPopes

PR #22906, opened 2026-03-10
perf(provider): remove unnecessary clones in changeset readers

stevencartavia:
Removes unnecessary `.clone()` calls on `Reverts` in `ConsistentProvider`. `to_plain_state_reverts` takes `&self`, and the count methods can use `.iter()` or `.len()` directly without allocating.


mattsse (approved):
The remaining `.reverts.clone().to_plain_state_reverts()` sites in this file (lines 1314, 1368, 1502, 1548) could also drop the clone since `to_plain_state_reverts` takes `&self`.

PR #22908, opened 2026-03-10
perf(engine): offload DeferredDrops deallocation to a persistent background thread

pepyakin:
Adds a persistent named thread (`reth-deferred-drop`) that receives `DeferredDrops` over an mpsc channel and deallocates them off the critical path.

Dropping proof node buffers after reveal can be expensive (thousands of nested `Vec<RlpNode>` inside `BranchNodeV2`). Previously these were dropped inline on the `sparse-trie` thread immediately after releasing the preserved-trie lock. This change moves that cost to a dedicated background thread so it never blocks state root computation or lock acquisition.

Benchmark (schelk big-blocks, 50 × 1 Ggas blocks, same commit baseline):

| Metric | Baseline | DeferredDropThread |
|--------|----------|--------------------|
| Overall Ggas/s | 1.2022 | 1.1987 (-0.29%) |
| Mean total latency | 832,300 µs | 834,738 µs |
| p50 total latency | 801,541 µs | 784,728 µs (-2.1%) |
| Mean FCU latency | 10,341 µs | 9,850 µs (-4.7%) |

Performance-neutral overall, with a slight improvement in FCU latency.


crates/trie/sparse/src/state.rs
mattsse, line 1: @DaniPopes wdyt about integrating this into the `reth_tasks:Runtime` that accepts `Box<dyn Any>` messages to drop

crates/trie/sparse/src/state.rs
```diff
@@ -28,6 +28,45 @@ pub struct DeferredDrops {
...
+    }
+
+    /// Sends [`DeferredDrops`] to the background thread for deallocation.
+    ///
+    /// If the background thread has exited, the drops are deallocated inline as a fallback.
+    pub fn schedule_drop(&self, drops: DeferredDrops) {
+        if drops.proof_nodes_bufs.is_empty() {
+            return;
+        }
+        // If the receiver is gone, fall back to inline drop.
```
DaniPopes: outdated/unnecessary comment

crates/trie/sparse/src/state.rs
DaniPopes: yeah can be like `runtime::spawn_drop(&self, x: T) { self.spawn_blocking_named("drop", move || drop(x)); }`

crates/trie/sparse/src/state.rs
pepyakin: that's actually a subtle one! `send(x: T)` returns `SendError<T>` so therefore _-ing it means dropping it inline.

DaniPopes (approved):
A single thread is fine, values are buffered in the channel. The only difference with the previous implementation is that the value is always sent to the thread even if empty, don't know if this matters much in practice.

mattsse (approved):

DaniPopes:
derek bench

DaniPopes (approved):
This is fine, we'll want it in other places too. Benches are within noise.

PR #22910, opened 2026-03-10
refactor(txpool): change `EthTransactionValidator::validate_stateless` return type, accept tx by ref

rkrasiuk:
## Description

Return `Result<(), InvalidPoolTransactionError>` instead of `Result<Tx, TransactionValidationOutcome<Tx>>`. Takes &Tx by reference instead of by value, allowing callers to retain ownership and wrap the error into a `TransactionValidationOutcome` themselves if necessary.

## Motivation

`EthTransactionValidator::validate_stateless` provides an ambiguous API by unnecessarily taking ownership of the transaction and returning `TransactionValidationOutcome` as an error, while neither `TransactionValidationOutcome::Valid` or `TransactionValidationOutcome::Error` are constructed.


mattsse (approved):
clean simplification

PR #22915, opened 2026-03-10
fix(metrics): Rename more instances of invalid save_blocks_block_count

bakhtin:
Another instance of Prometheus-incompatible metric was added in https://github.com/paradigmxyz/reth/commit/f012b3391e. See https://github.com/paradigmxyz/reth/pull/21654 for a similar fix of another incompatible metric.

While renaming a Gauge-type metric wasn't strictly necessary (only Histogram was incompatible) I renamed it for consistency.

Updated a Grafana dashboard as well.


mattsse (approved):
LGTM — clean follow-up to #21654, all references updated correctly including the Grafana dashboard.

PR #22916, opened 2026-03-10
fix(bench): retry HTTP 502 errors in block provider

decofe:
The block provider's `RetryBackoffLayer` only retried 429/503 by default. HTTP 502 (Bad Gateway) errors from the RPC would immediately fail the benchmark.

Uses `RateLimitRetryPolicy::or()` to also retry on 502, with max 2 retries.

Prompted by: DaniPopes


shekhirin (approved):

DaniPopes (approved):
LGTM! Thanks -- Derek! Looks good to ME!

PR #22919, opened 2026-03-10
fix(net): log provider errors in eth request handler instead of silently swallowing them

andrewshab3:
All provider calls in `EthRequestHandler` used `unwrap_or_default()` on `ProviderResult<Option<T>>`, which silently converts database errors (IO failures, corruption, missing static files) into `None` — making them indistinguishable from "data not found" responses.

Adds a `eth_req_handler_provider_errors_total` metric counter and `debug!`-level logging to the 5 affected call sites so operators can detect and diagnose provider failures during p2p request serving.


mattsse (commented):
not gonna review this

PR #22921, opened 2026-03-10
feat: add verisions to the reth download metadata

emmajam:
feat: add verisions to the reth download metadata


Rjected (approved):
nice, should be useful

emmajam:
> nice, should be useful

tims idea!

PR #22922, opened 2026-03-10
feat(trie): Use proof v2 in TrieWitness

mediocregopher:
This change refactors the TrieWitness to use V2 proof calculations via `Proof::multiproof_v2` in combination with the `SparseTrie::update_leaves` method which should correctly handle all cases where blinded nodes are required in leaf insert/removal edge-cases.

Should fix the issue brought up by @jsign [here](https://github.com/paradigmxyz/reth/pull/22290#issuecomment-3968580749).

The new implementation is pretty much a rewrite, but it should be more performant as it is no longer relying on the TrieNodeProvider.

This fix supercedes the work done on #22733, which used an alternate but ultimately wrong approach

This PR also includes a `compare-witness-recent.sh` script which is useful for comparing a batch of recent trie witnesses for correctness. It allows for extra nodes being present in the local witness, but not wrong/missing nodes.


jsign:
@mediocregopher, I'll ask Rafeael if he can run the `debug_executionWitness` using this branch in bloatnet, which was the way this problem was uncovered. I can ping you here wheenver we have any news on that front.

crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -525,18 +525,16 @@ impl<
             let nodes_sorted = input.nodes.into_sorted();
             let state_sorted = input.state.into_sorted();
             TrieWitness::new(
-                MaskedTrieCursorFactory::new(
-                    InMemoryTrieCursorFactory::new(
-                        reth_trie_db::DatabaseTrieCursorFactory::<_, A>::new(self.tx()),
-                        &nodes_sorted,
-                    ),
-                    input.prefix_sets.freeze(),
+                InMemoryTrieCursorFactory::new(
```
mediocregopher, line 528: MaskedTrieCursor was not sufficient to implement this, and is removed in favor of passing prefix sets through to the proof v2 calculators (see #22946)

crates/trie/trie/src/witness.rs
```diff
@@ -92,190 +97,281 @@ where
...
+        mut state: HashedPostState,
+    ) -> Result<B256Map<Bytes>, TrieWitnessError> {
         let is_state_empty = state.is_empty();
         if is_state_empty && !self.always_include_root_node {
             return Ok(Default::default())
         }
 
+        // Expand wiped storages into explicit zero-value entries for every existing slot,
+        // so that downstream code can treat all storages uniformly.
+        self.expand_wiped_storages(&mut state)?;
```
mediocregopher, line 111: This use to happen as part of `get_proof_targets`, but now happens as a separate preceding step.

crates/trie/trie/src/witness.rs
```diff
@@ -92,190 +97,281 @@ where
...
+        // first would add a new child that keeps the count above one, masking the need.
+        let mut storage_removals: B256Map<B256Map<LeafUpdate>> = B256Map::default();
+        let mut storage_upserts: B256Map<B256Map<LeafUpdate>> = B256Map::default();
+        for (hashed_address, storage) in &state.storages {
+            for (&hashed_slot, value) in &storage.storage {
+                if value.is_zero() {
+                    storage_removals
+                        .entry(*hashed_address)
+                        .or_default()
+                        .insert(hashed_slot, LeafUpdate::Changed(vec![]));
```
rkrasiuk, line 162: would be cleaner to represent this as `LeafUpdate::Removed`

rkrasiuk (commented):
nice

scripts/compare-witness-recent.sh
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
yongkangc, line 2: is this something that we would have in our nightly runs and workflows?

yongkangc (approved):
I dont see any logical regression and the cleanup is nice! lgtm

scripts/compare-witness-recent.sh
mediocregopher: For now I think running it manually when making changes is enough, if we find ourselves making lots of bug-fixes/changes to proof_v2 or TrieWitness it might make sense to automate then

crates/trie/trie/src/witness.rs
mediocregopher: I had originally thought it would be cleaner to keep it as a single "changed' vs "touched" (prewarm) dichotemy, but in the end I think you're right, we basically always end up checking is_empty or not and branching the logic on that. Not for this PR though

PR #22923, opened 2026-03-10
test(trie): Implement TrieTestHarness

mediocregopher:
This PR implements a TrieTestHarness helper type, and replaces a number of existing test harnesses which duplicate its functionality.

The test harness has the following scope:

* Generating a set of base TrieUpdates on top of a base key/value dataset. We assume a storage trie.
* Creation of trie/hashed cursor factories on top of that base dataset.
* Root and trie updates calculation given a set of changes to the base dataset.
* Application of hashed key changes to the base dataset, with accompanying trie updates
* Proof V2 calculation

This new test harness replaces the ones in:

* Sparse trie test suite
* proof_v2 - tests had to be translated from account trie based to storage trie based

It will also replace the one created for the proptests in the [ArenaParallelSparseTrie PR](https://github.com/paradigmxyz/reth/pull/22381).


crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -2021,423 +1901,26 @@ mod tests {
...
+        let harness = ProofTestHarness::new(storage);
 
-        // Assert the proof (convert B256 to ProofV2Target with no min_len for this test)
         harness
             .assert_proof(targets.into_iter().map(ProofV2Target::new))
             .expect("Proof generation failed");
     }
-
-    #[test]
-    fn test_failing_proptest_case_0() {
```
mediocregopher, line 2050: Translating this test to use storages wasn't viable, so I just deleted it. The proptests remain so we're still covered.

DaniPopes (approved):

PR #22944, opened 2026-03-11
perf: parallel segmented snapshot downloads

decofe:
Use HTTP Range requests with rayon to download snapshots in 16 parallel
segments, matching aria2c-style parallelism. Falls back to sequential
download when the server doesn't support Range requests.

Snapshot downloads were taking ~3 hours with a single connection.
aria2c does the same in ~10 minutes using 16 parallel connections.
This brings reth's download command to parity.

Co-Authored-By: zhygis <5236121+Zygimantass@users.noreply.github.com>

Prompted by: zygis

### DIFF
--- a/crates/cli/commands/src/download.rs
+++ b/crates/cli/commands/src/download.rs
@@ -9,9 +9,12 @@ use reth_fs_util as fs;
 use std::{
     borrow::Cow,
     fs::OpenOptions,
-    io::{self, BufWriter, Read, Write},
+    io::{self, BufWriter, Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
-    sync::{Arc, OnceLock},
+    sync::{
+        atomic::{AtomicBool, AtomicU64, Ordering},
+        Arc, Mutex, OnceLock,
+    },
     time::{Duration, Instant},
 };
 use tar::Archive;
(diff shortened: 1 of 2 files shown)


crates/cli/commands/src/download.rs
```diff
@@ -330,6 +333,8 @@ fn extract_from_file(path: &Path, format: CompressionFormat, target_dir: &Path)
 
 const MAX_DOWNLOAD_RETRIES: u32 = 10;
 const RETRY_BACKOFF_SECS: u64 = 5;
+const DEFAULT_SEGMENTS: usize = 16;
+const SEGMENT_RETRY_ATTEMPTS: u32 = 3;
```
Rjected, line 337: let's make these configurable? at least the 16 number

crates/cli/commands/src/download.rs
```diff
@@ -468,9 +473,211 @@ fn resumable_download(url: &str, target_dir: &Path) -> Result<(PathBuf, u64)> {
...
+            let downloaded = Arc::clone(&downloaded);
+            let errors = Arc::clone(&errors);
+
+            s.spawn(move |_| {
+                if let Err(e) = download_segment(url, part_path, start, end, &downloaded) {
+                    errors.lock().unwrap().push(e);
+                }
+            });
+        }
+    });
```
Rjected, line 658: I think we might want to move to tokio for this, especially if we benefit from increasing the number of segments to ie 32 or even more

Rjected (commented):
have some comments on making the number of concurrent connections configurable

shekhirin:
Superseded by https://github.com/paradigmxyz/reth/pull/23028

PR #22946, opened 2026-03-11
feat(trie): proof_v2 prefix set support

mediocregopher:
Add support for PrefixSets to the proof_v2::ProofCalculator.

PrefixSets are used to indicate that specific leaves are considered "dirty", that is whatever data we have cached about them in trie nodes is invalid and we should fallback to actual leaf data from the hashed cursors.

Invalidated data includes:
* hashes of parent branches
* presence - state_mask bit may indicate a leaf exists when it doesn't, ditto for parent branches which may then be collapsed deleted due to removal of leaves
* absence - state_mask bit may indicate a leaf does _not_ exist when it does, ditto for parent branches which may be created due to the addition of a leaf

The changes in this PR are designed such that if a prefix set is not given there is no difference in behavior from before the changes. Unit tests for specific edge-cases have been added as debugging has progressed.


crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -1,9 +1,10 @@
 //! Generic value encoder types for proof calculation with lazy evaluation.
```
mediocregopher, line 1: Changes in this file are just threading the prefix set through to the inner StorageCalculator calls

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -814,6 +905,12 @@ where
         sub_trie_prefix: &Nibbles,
         uncalculated_lower_bound: &Option<Nibbles>,
     ) -> Result<PopCachedBranchOutcome, StateProofError> {
+        // If the `uncalculated_lower_bound` is None it indicates that there can be no more
```
mediocregopher, line 908: `uncalculated_lower_bound` becomes None when the entire keyspace has been accounted for. In the context of `next_uncached_key_range`, where this method is called from, this means that a cached branch with a path consisting of all `f`s (e.g. 0xfff) has been fully constructed; at that point there can be no further leaves.

This check has been moved up ahead of the pop to get out of an infinite loop which was caused lower down with prefix sets active.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -909,17 +1018,6 @@ where
         sub_trie_upper_bound: Option<&Nibbles>,
         mut uncalculated_lower_bound: Option<Nibbles>,
     ) -> Result<Option<(Nibbles, Option<Nibbles>)>, StateProofError> {
-        // Pop any under-construction branches that are now complete.
-        // All trie data prior to the current cached branch, if any, has been computed. Any branches
-        // which were under-construction previously, and which are not on the same path as this
-        // cached branch, can be assumed to be completed; they will not have any further keys added.
-        // to them.
-        if let Some(cached_path) = self.cached_branch_stack.last().map(|kv| kv.0) {
-            while !cached_path.starts_with(&self.branch_path) {
```
mediocregopher, line 918: This loop got moved into a method and is now called immediately after `try_pop_cached_branch` in all cases. This allows us to be sure that when `next_uncached_key_range` returns the branch_stack is in the correct state to start processing leaves.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -980,12 +1088,62 @@ where
...
+
+            // Also include child nibbles indicated by the prefix set. The prefix set can
+            // indicate children that need recalculation from leaves (e.g. new keys inserted
+            // under this branch). Skip nibbles already set in `curr_state_mask` since those
+            // children have already been constructed.
+            if self.prefix_set.contains(&self.branch_path) {
+                let branch_path_len = self.branch_path.len();
+                let mut child_path = self.branch_path;
+                for nibble in 0u8..16 {
+                    if !curr_state_mask.is_bit_set(nibble) {
```
decofe, line 1101: 🚨 **[SECURITY]** Dirty leaves silently skipped when splitting an extension node with no sibling branches

When discovering prefix-set child nibbles, this `!curr_state_mask.is_bit_set(nibble)` guard unconditionally skips nibbles already set in `curr_state_mask`. If a dirty leaf diverges under an already-visited child nibble (e.g. a new key `[5,c]` that would split a cached extension node at `[5,a,b]`), and that parent branch has no later sibling nibble, `next_child_nibbles` becomes empty, the branch is popped, and `uncalculated_lower_bound` advances past the still-dirty leaves. This silently produces an incorrect state/storage root — a consensus failure.

**Recommended Fix:** Do not unconditionally skip a nibble already in `curr_state_mask` if `uncalculated_lower_bound` still falls under that nibble:
```rust
let is_current_nibble = uncalculated_lower_bound_ref
    .starts_with(&self.branch_path) &&
    uncalculated_lower_bound_ref.len() > branch_path_len &&
    uncalculated_lower_bound_ref.get_unchecked(branch_path_len) == nibble;

if !curr_state_mask.is_bit_set(nibble) || is_current_nibble {
    // ... check prefix set and set next_child_nibbles
}
```

decofe (commented):
:eye: **Cyclops Review**

**Summary:** This PR adds prefix-set-aware cached branch invalidation to V2 proof calculators. Three verified findings survived dedup: one high-severity state-root correctness bug and two medium-severity integration issues (currently dormant in production).

---

### ⚠️ [ISSUE] `multiproof_v2` bypasses new prefix-set invalidation path
**Severity:** Medium
**File:** `crates/trie/trie/src/proof/mod.rs:146-180` (not in diff)

This PR makes V2 proof correctness depend on prefix-set propagation (`ProofCalculator::with_prefix_set`, `SyncAccountValueEncoder::with_storage_prefix_sets`), but `Proof::multiproof_v2` still constructs calculators/encoders without passing `self.prefix_sets`. Calling `Proof::with_prefix_sets_mut(...)` before `multiproof_v2(...)` has no effect on V2 traversal, while legacy `multiproof(...)` does honor prefix sets. Currently dormant because `multiproof_v2` has no in-tree production callers with overlay prefix sets.

**Recommended Fix:** Thread `self.prefix_sets` into all V2 calculators inside `multiproof_v2`: apply the account prefix set to the account `ProofCalculator`, and pass storage prefix sets to `SyncAccountValueEncoder::with_storage_prefix_sets`.

---

### ⚠️ [ISSUE] Skipped prefix-set branches drop `tree_mask` / `hash_mask`, corrupting sparse-trie update tracking
**Severity:** Medium
**File:** `crates/trie/trie/src/proof_v2/mod.rs:486` (pre-existing code, not in diff)

When `should_skip_cached_branch` skips a cached branch for recalculation, the rebuilt branch goes through `push_new_branch`, which hardcodes `masks: None`. The V2 proof-node contract requires branch masks for sparse-trie update derivation. Missing masks cause incomplete `TrieUpdates` even when the root hash is correct. Currently dormant because production proof workers don't call `with_prefix_set`.

**Recommended Fix:** When a cached branch is skipped and rebuilt from leaves, preserve the skipped branch's `tree_mask`/`hash_mask` on the reconstructed `ProofTrieBranch`.

---

<details>
<summary>Reviewer Callouts</summary>

- ⚡ **`next_uncached_key_range` complexity**: This state machine absorbed multiple bug fixes in this PR alone. It remains the highest-value regression target for fuzzing.
- ⚡ **`multiproof_v2` future rollout**: If a future PR starts using `multiproof_v2()` for overlay/post-state proofs, the plumbing gap and mask-loss bugs become immediately reachable.
- ⚡ **`PrefixSet` shared-index correctness**: Both `contains()` and `contains_range()` mutate the same internal `index`. Future changes should maintain regression coverage around mixed query sequences.
- ⚡ **Long-lived worker calculators**: `proof_task.rs` keeps reusable V2 calculators alive across jobs. If per-request prefix sets are needed later, `with_prefix_set` at construction time won't suffice.
</details>

crates/trie/trie/src/proof_v2/mod.rs
mediocregopher: * Confirmed that the "dirty leaves silently skipped issue" is a false positive, that case is handled [here](https://github.com/paradigmxyz/reth/pull/22946/changes#diff-0f5e4d1b41c9f7f212a9fb1ec319846f1489e692499d00bbd356c3a2022b1131R1176-R1186) and there's already a unit test for it.

* `multiproof_v2 bypasses new prefix-set invalidation path` will be dealt with in #22922 once this is merged

* `Skipped prefix-set branches drop tree_mask / hash_mask, corrupting sparse-trie update tracking` is technically true... but we do not use proof_v2 w/ prefix sets combined with the sparse trie, only with TrieWitness, which doesn't need masks. In general masks are only needed to produce TrieUpdates, and it's not possible to correctly produce TrieUpdates using only prefix sets to revert trie state anyway, you must have trie changesets, so this will never be relevant.

crates/trie/trie/src/proof_v2/mod.rs
```diff
@@ -102,8 +104,21 @@ impl<TC, HC, VE: LeafValueEncoder> ProofCalculator<TC, HC, VE> {
...
             rlp_encode_buf: Vec::<_>::with_capacity(RLP_ENCODE_BUF_SIZE),
+            prefix_set: PrefixSet::default(),
         }
     }
+
+    /// Sets the prefix set and returns `self`.
+    ///
+    /// When given, all cached hashes matching the [`PrefixSet`] will be invalidated. When all but
+    /// one of a branch's children match the prefix set then that remaining child's cached hash, if
+    /// any will also be invalidated. This allows for properly handling branch collapse situations,
```
pepyakin:
```suggestion
    /// any, will also be invalidated. This allows for properly handling branch collapse situations,
```

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -91,15 +94,23 @@ impl<T, H> SyncAccountValueEncoder<T, H> {
         Self {
             trie_cursor_factory: Rc::new(trie_cursor_factory),
             hashed_cursor_factory: Rc::new(hashed_cursor_factory),
+            storage_prefix_sets: Rc::new(B256Map::default()),
         }
     }
+
+    /// Set the storage prefix sets.
```
pepyakin: Nit: It would be better if this said how the behavior would be altered if this had been called.

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -145,6 +159,7 @@ where
         SyncAccountDeferredValueEncoder {
             trie_cursor_factory: self.trie_cursor_factory.clone(),
             hashed_cursor_factory: self.hashed_cursor_factory.clone(),
+            storage_prefix_sets: self.storage_prefix_sets.clone(),
```
pepyakin: Nit, tiny: `Rc::clone` would've been clearer (esp for reading without LSP aid)

crates/trie/trie/src/proof_v2/value.rs
```diff
@@ -115,6 +126,9 @@ where
             self.hashed_cursor_factory.hashed_storage_cursor(self.hashed_address)?;
 
         let mut storage_proof_calculator = ProofCalculator::new_storage(trie_cursor, hashed_cursor);
+        if let Some(prefix_set) = self.storage_prefix_sets.get(&self.hashed_address) {
+            storage_proof_calculator = storage_proof_calculator.with_prefix_set(prefix_set.clone());
```
pepyakin, line 132: would the cloning here become an issue?

pepyakin (approved):
lgtm, nits only. none of the issues are blocking the merge!

crates/trie/trie/src/proof_v2/value.rs
mediocregopher: PrefixSet has an Arc internally so shouldn't be an issue

PR #22947, opened 2026-03-11
fix(trie): ArenaParallelSparseTrie: fix merge_subtrie_updates not cancelling updates/removals

mediocregopher:
Includes new case for the test suite

### DIFF
--- a/crates/trie/sparse/src/arena/mod.rs
+++ b/crates/trie/sparse/src/arena/mod.rs
@@ -835,6 +835,10 @@ impl ArenaParallelSparseTrie {
 
     /// Merges updates from a subtrie's buffer into the parent's buffer.
     /// Both `dst` and `src` must be `Some` when updates are being tracked.
+    ///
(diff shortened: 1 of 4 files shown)


crates/trie/sparse/src/arena/mod.rs
```diff
@@ -2961,7 +2975,7 @@ mod tests {
     }
 
     proptest! {
-        #![proptest_config(ProptestConfig::with_cases(5000))]
+        #![proptest_config(ProptestConfig::with_cases(1000))]
```
pepyakin, line 2978: is there a reason for this change or is this accidental?

pepyakin (approved):

crates/trie/sparse/src/arena/mod.rs
mediocregopher: Deliberate, it was taking too long to do 5000 iterations, nearly a minute on my machine

PR #22951, opened 2026-03-11
fix(rpc): zero parentBeaconBlockRoot in eth_simulateV1

jsvisa:
Zeroes `parentBeaconBlockRoot` for simulated blocks in `eth_simulateV1`, 
matching geth behavior and the spec's default block values table where consensus-layer fields default to zero.

Seems this is zero in the first place, and changed in #19645


mattsse (commented):
i think this is a valid request even if it deviates from the spec

https://github.com/paradigmxyz/reth/pull/19645#issuecomment-3516711230

PR #22953, opened 2026-03-11
test(trie): Integrate trie-debug recorder into ArenaParallelSparseTrie

mediocregopher:
When built with `-F trie-debug` reth will output all recorded trie interactions during a block upon a state-root mismatch or a failed comparison when `--engine.state-root-task-compare-updates` is set. This helps immensely during debugging.

This PR integrates this functionality into the ArenaParallelSparseTrie, so that it not only records all operations made against it but also the initial state at the start of a block after `prune` has been called.


crates/trie/sparse/src/debug_recorder.rs
pepyakin, line 1: Rustdoc could be updated:

```
//! Debug recorder for tracking mutating operations on sparse tries.
//!
//! This module is only available with the `trie-debug` feature and provides
//! infrastructure for recording all mutations to a [`crate::ParallelSparseTrie`]
//! for post-hoc debugging of state root mismatches.
```

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -523,6 +529,158 @@ impl ArenaParallelSparseTrie {
...
+        // Collect from the upper arena.
+        let mut cursor = mem::take(&mut self.buffers.cursor);
+        collect_records(
+            &mut self.upper_arena,
+            self.root,
+            Nibbles::default(),
+            &mut cursor,
+            &mut nodes,
+        );
+        self.buffers.cursor = cursor;
```
pepyakin: Is this cursor switcheroo really required?

pepyakin (approved):

pepyakin (approved):

yongkangc (approved):

PR #6587, opened 2024-02-13
chore: add exit future back

mattsse:
Add type back that somehow got lost

#6166

PR #22956, opened 2026-03-11
fix: resolve exit future once engine exits

klkvr:
Right now if consensus engine exits with an `Ok` outcome the `node_exit_future` will never resolve because it will hit the `terminate == false` branch.

I'm guessing that this logic is a left-over from legacy engine that is no longer relevant because consensus engine always stays alive (unless actually terminated) — maybe relevant PR https://github.com/paradigmxyz/reth/pull/6587

This only became an issue now because https://github.com/paradigmxyz/reth/pull/22698 made the engine exit gracefully on runtime shutdown which wasn't the case before and thus we were always entering the `Err(_)` branch


mattsse (approved):
lgtm

maybe we should also remove `config.debug.terminate` ?

PR #22975, opened 2026-03-12
chore(net): trim tx manager allocations

BearSoltokien:
Remove an unnecessary hash Vec collection in transaction announcement tracing, pre-reserve transactions_by_peers capacity before inserting recovered transactions.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -758,7 +758,6 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
         trace!(target: "net::tx::propagation",
             peer_id=format!("{peer_id:#}"),
             hashes_len=valid_announcement_data.len(),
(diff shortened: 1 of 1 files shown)


mattsse (commented):
thanks for looking into this!

the trace log cleanup is a nice improvement — removing the `collect::<Vec<_>>()` avoids an unnecessary allocation on a hot path while `hashes_len` still gives enough context.

for the `reserve` call: `transactions_by_peers` is a long-lived map that accumulates entries over the node's lifetime, so after the first few batches it'll already have enough capacity making `reserve` a no-op most of the time. the benefit here is marginal.

also minor nit: the `reserve` is outside the `if !new_txs.is_empty()` guard at line 1445, so it calls `reserve(0)` when there are no new txs. harmless but slightly inconsistent.

going to close this — the trace cleanup is good but probably not worth a standalone PR, and the reserve doesn't move the needle. feel free to fold the trace change into a broader PR if you'd like.

PR #22976, opened 2026-03-12
fix(rpc): use -38026 error code for "too many blocks"

jsvisa:
align with geth, to let hive test pass, refer https://github.com/ethereum/execution-apis/blob/ffe6c839567f931ece3276d8242963744f09bf67/tests/eth_simulateV1/ethSimulate-big-block-state-calls-array.io#L3


mattsse (approved):
makes sense, ty!

PR #22978, opened 2026-03-12
feat(payload): add resolve and job-creation latency histograms

decofe:
Adds latency histograms to the payload builder service that would have made the deadlock in #22971 visible.

| Metric | Type | What it catches |
|---|---|---|
| `payloads.resolve_duration_seconds` | Histogram | Resolve latency — a stalled resolve (deadlock) shows as an outlier or missing observations |
| `payloads.new_job_duration_seconds` | Histogram | Job creation latency — `new_payload_job()` blocking |

The resolve duration histogram covers end-to-end: for cache hits it records inline, for active jobs it records after the payload future completes. 

A deadlock like #22971 would show as observations stopping.

In the future, alerts can be added.

Prompted by: yk

### DIFF
--- a/crates/payload/builder/src/service.rs
+++ b/crates/payload/builder/src/service.rs
@@ -351,15 +351,7 @@ where
 
         Some(Box::pin(fut))
     }
-}
 
-impl<Gen, St, T> PayloadBuilderService<Gen, St, T>
-where
-    T: PayloadTypes,
-    Gen: PayloadJobGenerator,
-    Gen::Job: PayloadJob<PayloadAttributes = T::PayloadBuilderAttributes>,
-    <Gen::Job as PayloadJob>::BuiltPayload: Into<T::BuiltPayload>,
-{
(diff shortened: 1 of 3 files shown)


crates/payload/builder/src/service.rs
```diff
@@ -348,15 +351,7 @@ where
...
     }
-}
 
-impl<Gen, St, T> PayloadBuilderService<Gen, St, T>
-where
-    T: PayloadTypes,
-    Gen: PayloadJobGenerator,
-    Gen::Job: PayloadJob<PayloadAttributes = T::PayloadBuilderAttributes>,
-    <Gen::Job as PayloadJob>::BuiltPayload: Into<T::BuiltPayload>,
-{
```
yongkangc, line 359: removed redundant bound present in  https://github.com/paradigmxyz/reth/blob/e51520e/crates/payload/builder/src/service.rs#L237-L242

DaniPopes (approved):

PR #22979, opened 2026-03-12
perf(engine): avoid redundant state provider and header lookups

stevencartavia:
- Replace `sealed_header_by_hash(...).is_some()` with `blocks_by_hash.contains_key(...)` in the AlreadySeen check
- Use `executed_block_by_hash` + `provider.header` in `find_disk_reorg` to avoid cloning `SealedHeader`
- Replace `state_provider_builder` with a lightweight `has_parent_state` check the old result was immediately discarded

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -3143,10 +3143,7 @@ where
 
     /// Checks whether the parent state for the given hash is available, either in memory or in the
     /// database.
-    fn has_parent_state(&self, hash: B256) -> ProviderResult<bool>
-    where
-        P: BlockReader,
-    {
+    fn has_parent_state(&self, hash: B256) -> ProviderResult<bool> {
         // Lookup for in-memory blocks
         if self.state.tree_state.contains_hash(&hash) {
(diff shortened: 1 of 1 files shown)


decofe:
cc @yongkangc

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/23048138879/job/66942309836)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `lean-parent-lookups`

crates/engine/tree/src/tree/mod.rs
```diff
@@ -3148,6 +3155,18 @@ where
         Ok(())
     }
 
+    /// Checks whether the parent state for the given hash is available, either in memory or in the
+    /// database.
+    fn has_parent_state(&self, hash: B256) -> ProviderResult<bool> {
```
mattsse: I think we should rename this a bit, because the `_parent_` naming doesnt really make sense here

mattsse (changes requested):
this makes sense

there has been a conflicting change since so this needs a smol rebase

PR #22986, opened 2026-03-12
fix(net): send disconnect on invalid inbound eth messages

jsvisa:
Established eth sessions currently close the connection on some invalid inbound messages without sending a discMsg, from the devp2p spec https://github.com/ethereum/devp2p/blob/master/rlpx.md#disconnect-0x01, it defined the defines Disconnect (0x01) as the control message used to terminate an established session and includes 0x02 = breach of protocol as a standard reason in the disconnect reason table.

So let's explicit disconnect for established invalid inbound messages.

### DIFF
--- a/crates/net/network/src/session/active.rs
+++ b/crates/net/network/src/session/active.rs
@@ -491,7 +491,7 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
     }
 
     /// Returns whether a receive-side stream error should be treated as a protocol breach.
-    fn is_protocol_breach_receive_error(error: &EthStreamError) -> bool {
+    const fn is_protocol_breach_receive_error(error: &EthStreamError) -> bool {
         matches!(
             error,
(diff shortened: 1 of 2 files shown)


crates/net/network/src/session/active.rs
```diff
@@ -490,6 +490,25 @@ impl<N: NetworkPrimitives> ActiveSession<N> {
         self.poll_terminate_message(cx).expect("message is set")
     }
 
+    /// Returns whether a receive-side stream error should be treated as a protocol breach.
+    const fn is_protocol_breach_receive_error(error: &EthStreamError) -> bool {
```
mattsse: can we make this a fn of `EthStreamError` instead

mattsse (changes requested):
cool, this makes sense, tiny style nit

mattsse (approved):
lgtm

DaniPopes (approved):

PR #22993, opened 2026-03-12
fix(tasks): install panic handler on all worker pools

mediocregopher:
Without a panic handler, a panic on any WorkerPool thread (proof-acct, proof-stg, prewarm) aborts the process. Install a handler in WorkerPool::from_builder that logs the panic instead.

### DIFF
--- a/crates/tasks/src/pool.rs
+++ b/crates/tasks/src/pool.rs
@@ -180,10 +180,23 @@ impl WorkerPool {
     }
 
     /// Creates a new `WorkerPool` from a [`rayon::ThreadPoolBuilder`].
+    ///
+    /// Installs a panic handler that logs panics instead of aborting the process.
     pub fn from_builder(
         builder: rayon::ThreadPoolBuilder,
     ) -> Result<Self, rayon::ThreadPoolBuildError> {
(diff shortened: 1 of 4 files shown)


shekhirin (approved):

crates/tasks/src/pool.rs
```diff
@@ -180,10 +180,23 @@ impl WorkerPool {
...
-        Ok(Self { pool: builder.build()? })
+        Ok(Self {
+            pool: builder
+                .panic_handler(|payload| {
+                    let msg = payload
+                        .downcast_ref::<&str>()
+                        .copied()
+                        .or_else(|| payload.downcast_ref::<String>().map(|s| s.as_str()))
+                        .unwrap_or("(no message)");
+                    tracing::error!(target: "reth::tasks", %msg, "panic in worker pool thread");
```
klkvr: should we maybe remove this? rayon produces the backtrace of the panic anyway so i think it might be okay to not log anything else

klkvr (approved):

DaniPopes (approved):

klkvr (approved):

PR #22995, opened 2026-03-12
fix(provider): heal finalized/safe block numbers ahead of highest header

decofe:
## Summary

On 1.10.x shutdown, the finalized block number is always flushed but blocks/state are batched. This can leave finalized (and safe) block numbers ahead of the actual tip in the database. Any node upgrading from 1.10.x starts with this inconsistent state.

## Changes

- Added `heal_chain_state_block_numbers()` to `ProviderFactory`, called as Step 4 at the end of `check_consistency()` when no unwind is needed
- Reads finalized/safe block numbers and compares against `last_block_number()` (highest header from static files)
- If either exceeds the highest header, opens a write tx and corrects them

### Drive-by

- Pipeline unwind was only bringing down finalized but not safe — added `save_safe_block_number` next to the existing `save_finalized_block_number` in the unwind loop

## Testing

- `cargo check -p reth-provider reth-stages-api` ✅
- `cargo clippy -p reth-provider --no-deps -- -D warnings` ✅

Prompted by: joshie

### DIFF
--- a/crates/stages/api/src/pipeline/mod.rs
+++ b/crates/stages/api/src/pipeline/mod.rs
@@ -394,8 +394,8 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
 
                         let last_saved_safe_block_number = provider_rw.last_safe_block_number()?;
 
-                        if last_saved_safe_block_number.is_none()
-                            || Some(checkpoint.block_number) < last_saved_safe_block_number
+                        if last_saved_safe_block_number.is_none() ||
+                            Some(checkpoint.block_number) < last_saved_safe_block_number
                         {
                             provider_rw.save_safe_block_number(BlockNumber::from(
                                 checkpoint.block_number,
--- a/crates/stages/api/src/pipeline/mod.rs
+++ b/crates/stages/api/src/pipeline/mod.rs
@@ -378,7 +378,7 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
                             });
                         }
 
-                        // update finalized block if needed
+                        // update finalized and safe block if needed
                         let last_saved_finalized_block_number =
                             provider_rw.last_finalized_block_number()?;
 
@@ -388,6 +388,16 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
                             Some(checkpoint.block_number) < last_saved_finalized_block_number
                         {
                             provider_rw.save_finalized_block_number(BlockNumber::from(
+                                checkpoint.block_number,
+                            ))?;
+                        }
+
+                        let last_saved_safe_block_number = provider_rw.last_safe_block_number()?;
+
+                        if last_saved_safe_block_number.is_none()
+                            || Some(checkpoint.block_number) < last_saved_safe_block_number
+                        {
+                            provider_rw.save_safe_block_number(BlockNumber::from(
                                 checkpoint.block_number,
                             ))?;
                         }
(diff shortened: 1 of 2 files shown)


crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -367,8 +367,65 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
             },
         );
 
+        drop(provider_ro);
+
```
joshieDo: just use it in `heal_chain_state_block_numbers`

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -367,8 +367,65 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
...
+        // Step 4: Heal finalized/safe block numbers if they are ahead of the
+        // highest header. This can happen on 1.10.x when the node shuts down
+        // after flushing the finalized block number but before persisting the
+        // corresponding blocks/state.
+        self.heal_chain_state_block_numbers()?;
```
joshieDo: should we only call this if we're not unwinding? since unwinding will fix it anyway

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -367,8 +367,65 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
             },
         );
 
+        drop(provider_ro);
+
+        // Step 4: Heal finalized/safe block numbers if they are ahead of the
+        // highest header. This can happen on 1.10.x when the node shuts down
+        // after flushing the finalized block number but before persisting the
+        // corresponding blocks/state.
```
joshieDo: less verbose (still reference its for <=1.10.2 nodes

crates/stages/api/src/pipeline/mod.rs
```diff
@@ -392,6 +392,16 @@ impl<N: ProviderNodeTypes> Pipeline<N> {
...
+                        let last_saved_safe_block_number = provider_rw.last_safe_block_number()?;
+
+                        if last_saved_safe_block_number.is_none()
+                            || Some(checkpoint.block_number) < last_saved_safe_block_number
+                        {
+                            provider_rw.save_safe_block_number(BlockNumber::from(
+                                checkpoint.block_number,
+                            ))?;
+                        }
+
```
joshieDo, line 404: drive-by fix

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -367,8 +367,57 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
             },
         );
 
+        // Step 4: Heal finalized/safe block numbers that may be ahead of the
+        // highest header on nodes coming from <=1.10.2.
```
joshieDo, line 371:
```suggestion
        // Step 4: Heal finalized/safe block numbers that may be ahead of the
        // highest header on nodes coming from <=1.10.2.
        //
        // Unwinds already set it to the target block.
```

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -367,8 +367,59 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
...
+        // Unwinds already set it to the target block.
+        if rocksdb_unwind.is_none() && static_file_unwind.is_none() {
+            self.heal_chain_state_block_numbers(&provider_ro)?;
+        }
+
         Ok((rocksdb_unwind, static_file_unwind))
     }
+
+    /// Checks if the stored finalized or safe block numbers exceed the highest
+    /// known header and corrects them if so.
```
mattsse: can you clarify what correct is here?

mattsse (approved):
smol doc request

yongkangc (approved):
makes sense

PR #22996, opened 2026-03-12
chore(trie): address arena PR review feedback

mediocregopher:
Addresses remaining feedback from https://github.com/paradigmxyz/reth/pull/22381.

- Deduplicate `type Index` / `type NodeArena` aliases — defined once in `mod.rs`, imported via `super::` in `cursor.rs` and `nodes.rs`
- Add `ArenaSparseNodeBranch::sibling_child` for the 2-child branch sibling lookup, replacing the `1 - child_idx.get()` trick
- Check `cursor.depth() == 0` before popping in `remove_leaf` instead of popping then checking `is_empty()`; removes `cursor::is_empty()` entirely
- `cursor::drain` now stops at the root instead of clearing the stack, since there is no parent to propagate dirty state to
- `cursor::reset` asserts the stack is drained (at most the root) before resetting
- Remove redundant dirty-state branch guard in `cursor::pop` — unconditionally write `Dirty` instead of checking first

### DIFF
--- a/crates/trie/sparse/src/arena/branch_child_idx.rs
+++ b/crates/trie/sparse/src/arena/branch_child_idx.rs
@@ -38,13 +38,6 @@ impl BranchChildIdx {
         self.0 as usize
     }
 
-    /// Returns the dense index of the other child in a branch with exactly 2 children.
-    /// With exactly 2 bits set the dense array has indices 0 and 1, so the sibling is
-    /// simply `1 - self`.
-    pub(super) const fn sibling(self) -> Self {
-        Self(1 - self.0)
-    }
-
     /// Counts the number of occupied child slots below `nibble` in the dense children array.
     const fn count_below(state_mask: TrieMask, nibble: u8) -> u8 {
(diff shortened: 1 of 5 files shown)


crates/trie/sparse/src/arena/branch_child_idx.rs
```diff
@@ -38,6 +38,13 @@ impl BranchChildIdx {
         self.0 as usize
     }
 
+    /// Returns the dense index of the other child in a branch with exactly 2 children.
+    /// With exactly 2 bits set the dense array has indices 0 and 1, so the sibling is
+    /// simply `1 - self`.
+    pub(super) const fn sibling(self) -> Self {
```
pepyakin: oh oh I think this may work only in this limited case! so the doc must be updated? maybe debug_assertion check.

mediocregopher: from this context there's nothing to assert, the BranchChildIdx doesn't know the state_mask. But I could move this method to the ArenaSparseNodeBranch, then it could

pepyakin (approved):

PR #22998, opened 2026-03-12
fix(provider): heal finalized/safe block numbers ahead of highest header

decofe:
## Summary
Backport of [294e215](https://github.com/paradigmxyz/reth/commit/294e215) (`fix(provider): heal finalized/safe block numbers ahead of highest header (#22995)`) adapted for `tempo/v1.10.2`.

The upstream commit couldn't be cherry-picked cleanly because it relies on `ProviderFactory::check_consistency` and `ProviderError::MustUnwind` which don't exist on this branch. This PR manually adapts the fix to the branch's architecture.

## Changes

- `crates/node/builder/src/launch/common.rs`: Added Step 4 to the startup consistency check — if the stored finalized or safe block number is ahead of the highest header, reset it to the highest header. Only runs when no unwind is needed (unwinds already set it to the target block).
- `crates/stages/api/src/pipeline/mod.rs`: Update safe block number during pipeline unwind, matching the existing finalized block number logic.

## Testing
- `cargo check -p reth-stages-api -p reth-node-builder` passes
- `cargo +nightly fmt --all --check` passes

Prompted by: joshie

### DIFF
--- a/crates/node/builder/src/launch/common.rs
+++ b/crates/node/builder/src/launch/common.rs
@@ -66,8 +66,9 @@ use reth_node_metrics::{
 };
 use reth_provider::{
     providers::{NodeTypesForProvider, ProviderNodeTypes, RocksDBProvider, StaticFileProvider},
-    BlockHashReader, BlockNumReader, DatabaseProviderFactory, ProviderError, ProviderFactory,
-    ProviderResult, RocksDBProviderFactory, StageCheckpointReader, StaticFileProviderBuilder,
+    BlockHashReader, BlockNumReader, ChainStateBlockReader, ChainStateBlockWriter,
+    DatabaseProviderFactory, ProviderError, ProviderFactory, ProviderResult,
+    RocksDBProviderFactory, StageCheckpointReader, StaticFileProviderBuilder,
     StaticFileProviderFactory,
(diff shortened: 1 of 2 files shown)


.github/workflows/unit.yml
```diff
@@ -26,15 +26,13 @@ jobs:
...
-        type: [ethereum, optimism]
+        type: [ethereum]
         storage: [stable, edge]
         include:
           - type: ethereum
             features: asm-keccak ethereum
             exclude_args: ""
-          - type: optimism
-            features: asm-keccak
-            exclude_args: --exclude reth --exclude reth-bench --exclude "example-*" --exclude "reth-ethereum-*" --exclude "*-ethereum"
```
shekhirin, line 37: do we need to remove this? unrelated, maybe can be put back?

joshieDo: i wanted a passing CI, unsure if the original branch is fucked or if this introduced issues. but i now see that the same test failed  on eth stable

joshieDo: removed, the orignal branch is the one with failing CI

Rjected (approved):
lgtm

PR #23007, opened 2026-03-12
fix(rpc): eth_config returns wrong fork

MagicJoshh:
where genesis is at timestamp 0 and Shanghai doesn't kick in until 1000 — eth_config was happily returning the Shanghai config for blocks that hadn't even reached it yet.

The position().and_then(checked_sub).or_else(len-1) chain conflates two different None paths: "past all forks" (position returns None, fallback to last is correct) vs "before all forks" (position returns Some(0), checked_sub(1) underflows to None, then incorrectly falls into the same fallback). Split it into an explicit match so Some(0) correctly hits the error path instead of silently returning the last fork.


mattsse (approved):
lgtm — consider adding a regression test for the pre-first-fork case

DaniPopes (approved):

PR #23009, opened 2026-03-13
perf: avoid redundant seal_slow when hash is known

stevencartavia:
`sealed_header_by_id` in `ConsistentProvider` and `init_from_state_dump` in `db-common` call `SealedHeader::seal_slow` (RLP encode + keccak256) despite already having the block hash. Use `SealedHeader::new` instead.


mattsse (approved):
lgtm

PR #23011, opened 2026-03-13
perf(rpc): avoid hash_slow in reward traces

stevencartavia:
Uses the pre-cached block hash from `SealedBlock` instead of recomputing `hash_slow()` in `reward_trace`, avoiding redundant keccak256 on every `trace_block`/`trace_filter` call.


mattsse (approved):
lgtm

PR #23012, opened 2026-03-13
fix(txpool): use ceiling division for replacement tx price bump check

jsvisa:
Fixes a rounding issue where truncating integer division allowed underpriced replacement transactions to bypass the minimum price bump.

eg:  with a 10% bump on fee=1, `1 * 110 / 100 = 1` (truncated) required no actual increase. Using `div_ceil` correctly rounds up to 2.

### DIFF
--- a/crates/transaction-pool/src/pool/txpool.rs
+++ b/crates/transaction-pool/src/pool/txpool.rs
@@ -3105,6 +3105,35 @@ mod tests {
     }
 
     #[test]
+    fn insert_replace_underpriced_rounds_up_minimum_bump() {
+        let on_chain_balance = U256::ZERO;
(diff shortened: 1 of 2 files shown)


mattsse (commented):
ty,

imo this makes sense, and is correct.

but doesnt this now technically diverge from   https://github.com/ethereum/go-ethereum/blob/1c9ddee16f925989cc6bda3df39a98cd46c8b1f1/core/txpool/legacypool/list.go#L310-L323

mattsse (approved):
Correct fix — truncating division silently drops the bump requirement at low fee values. `div_ceil` enforces the intended minimum increase. The geth divergence is fine since txpool replacement policy is local, and being stricter is the safer direction.

PR #23014, opened 2026-03-13
fix(net): fully remove disconnected peers from transaction state

jsvisa:
When a peer session closes, the tx manager removed the peer from its main peer map, but tx-local state could still retain references to that peer. This will waste retry/search work against peers that are already gone, and those gone peers should also not treated as fallback candidates. 

Propose to make changes to make peer removal consistent across tx-local state.

### DIFF
--- a/crates/net/network/src/transactions/fetcher.rs
+++ b/crates/net/network/src/transactions/fetcher.rs
@@ -92,6 +92,9 @@ impl<N: NetworkPrimitives> TransactionFetcher<N> {
     /// Removes the peer from the active set.
     pub(crate) fn remove_peer(&mut self, peer_id: &PeerId) {
         self.active_peers.remove(peer_id);
+        for (_, metadata) in self.hashes_fetch_inflight_and_pending_fetch.iter_mut() {
+            metadata.fallback_peers_mut().remove(peer_id);
+        }
     }
 
     /// Updates metrics.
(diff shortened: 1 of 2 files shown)


crates/net/network/src/transactions/mod.rs
```diff
@@ -460,6 +460,18 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
         self.network.reputation_change(peer_id, ReputationChangeKind::AlreadySeenTransaction);
     }
 
+    /// Removes a peer from transaction-local tracking state.
+    fn remove_peer_from_transaction_state(&mut self, peer_id: &PeerId) {
```
mattsse: nit: can we name this on_peer_session_closed so that this matches the event this handles

crates/net/network/src/transactions/mod.rs
```diff
@@ -460,6 +460,18 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
...
+        if let Some(mut peer) = self.peers.remove(peer_id) {
+            self.policies.propagation_policy_mut().on_session_closed(&mut peer);
+        }
+        self.transaction_fetcher.remove_peer(peer_id);
+        self.transactions_by_peers.retain(|_, peers| {
+            peers.remove(peer_id);
+            !peers.is_empty()
+        });
```
mattsse: I'm not sure we need this and I'd like to avoid having to scan every transaction here, which I think we dont need?

crates/net/network/src/transactions/fetcher.rs
```diff
@@ -92,6 +92,9 @@ impl<N: NetworkPrimitives> TransactionFetcher<N> {
     /// Removes the peer from the active set.
     pub(crate) fn remove_peer(&mut self, peer_id: &PeerId) {
         self.active_peers.remove(peer_id);
+        for (_, metadata) in self.hashes_fetch_inflight_and_pending_fetch.iter_mut() {
+            metadata.fallback_peers_mut().remove(peer_id);
+        }
```
mattsse: this also scans all tracked items, why do we need to clean up this up?

I believe this already takes care of this lazily?

https://github.com/paradigmxyz/reth/blob/28e067432ae9d0854e56b70760ed34eaf480fdbd/crates/net/network/src/transactions/fetcher.rs#L178-L184

mattsse (commented):
I'm skeptical that we need this because cleaning of these ephemeral trackers isnt really needed afaict

crates/net/network/src/transactions/fetcher.rs
mattsse: oh actually @jsvisa 

```
let Some(inflight_count) = self.active_peers.peek(peer_id) else { return true };
```

this looks like a bug?

jsvisa: yes, I think so. A disconnected peer that's still in `fallback_peers` would be selected as an "idle" candidate for fetching transactions, and this is what this PR mainly try to fix.

crates/net/network/src/transactions/mod.rs
jsvisa: reverted back

crates/net/network/src/transactions/fetcher.rs
jsvisa: insted of O(n) fallback_peers scan in remove_peer, changed to re-buffer the hashes when `peers.get(peer_id)` failed(caused by disconnected peer selected by is_idle)

mattsse (approved):

PR #23018, opened 2026-03-13
fix(chain-state): correct return type of NewCanonicalChain::tip()

rejected-l:
In commit 13ecd6afa the function body was changed from `.block()` to `.recovered_block()` but the return type was not updated. The function returns `&RecoveredBlock<N::Block>` but was typed as `&SealedBlock<N::Block>`. Only compiled due to Deref coercion.


mattsse (approved):
strange why does this still compile with just the return type change

rejected-l:
RecoveredBlock has Deref to SealedBlock so the old return type worked via deref coercion

PR #23030, opened 2026-03-13
fix(pool): clean up blob sidecars when removing transactions

HuberyJulianay:
`remove_transactions` was the only discard path that didn't call `delete_discarded_blobs`, leaving orphaned blob data in the store.


mattsse (commented):
this is intended as two step op:

https://github.com/paradigmxyz/reth/blob/a20d1fb1ef107291232320d4b8da8782a0b626aa/crates/transaction-pool/src/maintain.rs#L292-L293

PR #23033, opened 2026-03-14
fix(storage): preserve genesis history entries in RocksDB consistency check

decofe:
RIght now if `checkpoint=0` we will always clear `AccountsHistory` and `StoragesHistory` tables entirely which is almost never a correct thing to do because those tables contain genesis state.

This PR changes this logic to never prune the tables directly but instead check how many entries they have and only fallback to the changesets-based pruning if they have any entries beyond genesis data.

Additionally i've moved some initialization helpers around so that we can use them in tests

### DIFF
--- a/crates/storage/provider/src/providers/rocksdb/invariants.rs
+++ b/crates/storage/provider/src/providers/rocksdb/invariants.rs
@@ -9,6 +9,7 @@ use crate::StaticFileProviderFactory;
 use alloy_eips::eip2718::Encodable2718;
 use alloy_primitives::BlockNumber;
 use rayon::prelude::*;
+use reth_chainspec::EthChainSpec;
 use reth_db_api::tables;
 use reth_stages_types::StageId;
 use reth_static_file_types::StaticFileSegment;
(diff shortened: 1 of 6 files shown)


crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -360,12 +383,17 @@ impl RocksDBProvider {
...
+
+            // Only trigger potentially slow healing if we have any entries beyond the genesis data.
+            if self.iter::<tables::AccountsHistory>()?.count() > genesis_entries {
+                tracing::info!(
+                    target: "reth::providers::rocksdb",
+                    "AccountsHistory: checkpoint is 0, clearing stale data"
+                );
+            } else {
+                return Ok(None);
+            }
```
klkvr: here we could also just unconditionally overwrite entries with genesis data, that'd require adding a helper function and likely doesn't matter much as this would only affect cases when `IndexAccountHistory`/`IndexStorageHistory` stages are interrupted almost immediately after starting

crates/storage/provider/src/providers/rocksdb/invariants.rs
```diff
@@ -360,12 +383,17 @@ impl RocksDBProvider {
...
-            tracing::info!(
-                target: "reth::providers::rocksdb",
-                "AccountsHistory: checkpoint is 0, clearing stale data"
-            );
-            self.clear::<tables::AccountsHistory>()?;
-            return Ok(None);
+            let genesis_entries = provider.chain_spec().genesis().alloc.len();
+
+            // Only trigger potentially slow healing if we have any entries beyond the genesis data.
+            if self.iter::<tables::AccountsHistory>()?.count() > genesis_entries {
```
yongkangc: would this be an issue if the entries is really high? same for `iter::<StoragesHistory>().count()`

klkvr: updated to use `.nth`, should be fast enough now

joshieDo (approved):

klkvr (approved):

PR #23035, opened 2026-03-14
fix(net): treat malformed blob sidecar responses as peer misbehavior

jsvisa:
Blob-sidecar-specific import failures were previously folded into generic bad-import handling in the tx manager.

That meant a malformed blob sidecar response could be treated like a globally bad
transaction:

- the sender was only handled through generic bad-transaction penalties
- the transaction hash could be inserted into `bad_imports`
- subsequent announcements for the same hash could be filtered out even if another peer
could serve valid blob data

This change treats response-sourced malformed blob sidecar failures as peer misbehavior instead.

### DIFF
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -468,6 +468,8 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
         self.transaction_fetcher.remove_peer(peer_id);
     }
 
+    /// Penalizes and disconnects a peer that violated the protocol (e.g. sent malformed blob
+    /// sidecar data).
     fn disconnect_peer_for_protocol_violation(&self, peer_id: PeerId) {
         self.report_peer_bad_transactions(peer_id);
         self.network.disconnect_peer_with_reason(peer_id, DisconnectReason::SubprotocolSpecific);
--- a/crates/net/network/src/transactions/mod.rs
+++ b/crates/net/network/src/transactions/mod.rs
@@ -44,8 +44,8 @@ use alloy_primitives::{TxHash, B256};
 use constants::SOFT_LIMIT_COUNT_HASHES_IN_NEW_POOLED_TRANSACTIONS_BROADCAST_MESSAGE;
 use futures::{stream::FuturesUnordered, Future, StreamExt};
 use reth_eth_wire::{
-    DedupPayload, EthNetworkPrimitives, EthVersion, GetPooledTransactions, HandleMempoolData,
-    HandleVersionedMempoolData, NetworkPrimitives, NewPooledTransactionHashes,
+    DedupPayload, DisconnectReason, EthNetworkPrimitives, EthVersion, GetPooledTransactions,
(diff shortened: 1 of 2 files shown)


crates/transaction-pool/src/error.rs
```diff
@@ -145,6 +145,16 @@ impl PoolError {
             }
         }
     }
+
+    /// Returns `true` if the error indicates malformed or incompatible blob sidecar data from a
+    /// peer, without proving that the transaction hash itself is globally invalid.
+    #[inline]
+    pub const fn is_blob_sidecar_misbehavior(&self) -> bool {
```
mattsse: could we maybe make this more generic like 

```
is_protocol_violation
```

or something, just so that we could reuse this for anything that should result in an immediate disconnect?

crates/net/network/src/transactions/mod.rs
```diff
@@ -562,14 +577,19 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
 
 impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
     /// Processes a batch import results.
-    fn on_batch_import_result(&mut self, batch_results: Vec<PoolResult<AddedTransactionOutcome>>) {
+    fn on_batch_import_result(
+        &mut self,
+        peer_id: PeerId,
+        source: TransactionSource,
+        batch_results: Vec<PoolResult<AddedTransactionOutcome>>,
+    ) {
```
mattsse: I dont think we need need any of these changes

crates/net/network/src/transactions/mod.rs
```diff
@@ -488,9 +498,14 @@ impl<Pool: TransactionPool, N: NetworkPrimitives> TransactionsManager<Pool, N> {
     /// - kzg error
     /// - not blob transaction (tx type mismatch)
     /// - wrong versioned kzg commitment hash
-    fn on_bad_import(&mut self, err: PoolError) {
+    fn on_bad_import(&mut self, peer_id: PeerId, source: TransactionSource, err: PoolError) {
         let peers = self.transactions_by_peers.remove(&err.hash);
```
mattsse, line 506: can we not use all the peers here?

hmm maybe not because we only use the first blob tx that we import?

mattsse (commented):
> previously folded into generic bad-import

over the long run this has the same effect: disconnect peer, right?

a bit skeptical that this is worth doing if the existing code has the same effect eventually

jsvisa:
> over the long run this has the same effect: disconnect peer, right?

I think so, the peer will be disconnected, but in this PR the key issue is that the current path also inserts the tx hash into bad_imports, which blocks the transaction from being fetched from any other peer. So we try not to avoid poisoning the `bad_imports` cache

mattsse (changes requested):
I see, this makes sense now.

I believe there's a problem with this change, which is that this could maybe end up penalizing peers that sent the correct sidecar, but we only dedup by hash.

so what if we relax this slightly and instead of `disconnect_peer_for_protocol_violation`, simply dont add the blob errors as bad imports?
and keep the regular `report_peer_bad_transactions`, then we rule out this edge case and we would still disconnect from a peer that keeps sending bad blobs?

mattsse (approved):
lgtm

PR #23039, opened 2026-03-14
chore: add glamsterdam to reth-bench

Soubhik-10:
closes #23038

### DIFF
--- a/bin/reth-bench/src/bench/helpers.rs
+++ b/bin/reth-bench/src/bench/helpers.rs
@@ -107,9 +107,12 @@ pub(crate) fn prepare_payload_request(
     let cancun_active = chain_spec.is_cancun_active_at_timestamp(timestamp);
     let prague_active = chain_spec.is_prague_active_at_timestamp(timestamp);
     let osaka_active = chain_spec.is_osaka_active_at_timestamp(timestamp);
+    let amsterdam_active = chain_spec.is_amsterdam_active_at_timestamp(timestamp); // Todo: Enable by default for bal bench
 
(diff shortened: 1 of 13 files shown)


crates/rpc/rpc/src/testing.rs
```diff
@@ -203,19 +213,40 @@ where
...
+            .try_into_v5()
+            .map_err(RethError::other)
+            .map_err(Eth::Error::from_eth_err)
+    }
+
+    async fn build_block_v2(
+        &self,
+        request: TestingBuildBlockRequestV1,
+    ) -> Result<ExecutionPayloadEnvelopeV6, Eth::Error> {
+        self.build_block(request, false)
```
mattsse: should this be true?

bin/reth-bench/src/bench/helpers.rs
```diff
@@ -260,10 +265,51 @@ pub(crate) async fn get_payload_with_sidecar(
...
 
+pub(crate) async fn fetch_block_access_list(
+    rpc_url: &str,
+    block_hash: B256,
+) -> eyre::Result<Bytes> {
+    let client = reqwest::Client::new();
+
+    let request = serde_json::json!({
+        "jsonrpc": "2.0",
+        "method": "debug_getBlockAccessList",
```
mattsse: I think this needs to use the eth_ endpoint?

for which we should already have alloy provider fns, if not we can add

mattsse (changes requested):
we def need this, I assume maybe the testing_ enpoints will change but all things reth-bench we'll need

mattsse (approved):

PR #23040, opened 2026-03-14
fix(p2p): respect --bootnodes flag in `reth p2p` commands

HuberyJulianay:
`reth p2p header` and `reth p2p body` ignored `--bootnodes` and always used the chain spec defaults. Now uses `resolved_bootnodes()` first, matching the behavior of the main node startup path.


mattsse (approved):
Correct fix — aligns `reth p2p` bootnode resolution with the main node startup path.

PR #23045, opened 2026-03-16
perf(trie): hot/cold node split for arena sparse trie

pepyakin:
`ArenaCursor::seek` is the single largest application-level L3 cache miss source in the arena sparse trie, accounting for 1.64% of all L3 misses across the entire profile. It also consumes 0.74% of total cycles on the sparse trie thread.

The root cause is a serialized pointer-chasing dependency chain inside `seek`. Each trie level descent requires a sequence of dependent loads where every load blocks on the result of the previous one — the CPU cannot issue any of them speculatively.

## The seek loop

Each iteration of the inner loop descends one trie level:

1. Read `head = self.stack.last()` → get `head.index` (arena Index)
2. Read `arena[head_idx]` → SlotMap version check + load node, match discriminant tag
3. If Branch: read `branch.short_key` and extend `entry.path` with it to form the logical path
4. Compare `full_path.starts_with(&logical_path)` — SIMD 256-bit comparison
5. Extract `child_nibble` from `full_path` at the logical path length
6. Read `branch.state_mask`, compute `BranchChildIdx` via popcount
7. Read `branch.children[branch_child_idx]` — into the SmallVec (may chase heap pointer)
8. If Revealed: read the child's Index, compute child path
9. Push new entry onto stack, loop back to step 1

## Serialized dependency chain

Every load below depends on the result of the previous load:

```
stack.last()
  → head.index
    → SlotMap version check                    2.88%  mov 0xf0(%rdx,%r8,1),%ecx
    → arena[head_idx] node discriminant        8.93%  movzbl 0xc6(%r8),%eax
    → branch.state_mask                        5.02%  mov 0xc4(%r8),%esi
    → branch.short_key.len                     4.48%  mov 0x98(%r8),%rax
      ↳ (logical_branch_path: path compare)
    → BranchChildIdx via popcount(state_mask)
    → SmallVec capacity (spill check)          7.62%  mov 0x90(%r8),%rdx
    → SmallVec heap pointer (if spilled)       9.02%  mov 0x8(%r8),%rcx
                                               8.35%  mov (%r8),%r8
    → children[idx] discriminant              14.62%  cmpb $0x0,(%r8,%rdx,4)
    → child_idx from Revealed variant          7.16%  mov 0x4(%rdx),%esi
                                               5.13%  mov 0x8(%rdx),%edx
      ↳ (child_path: reloads branch.short_key) 3.61%  mov 0x98(%rcx),%r8
    → SlotMap version check (child)            2.33%  mov 0xf0(%rdi,%rcx,1),%esi
    → child node discriminant                  2.51%  cmpb $0x3,0xc6(%rcx)
    → [repeat from top for next level]
```

## L3 miss hotspots

Raw L3 miss percentages from profiling, within `ArenaCursor::seek` (32,501 samples total):

| L3 Miss % | Instruction | Source Operation |
|-----------|-------------|-----------------|
| **14.62%** | `cmpb $0x0,(%r8,%rdx,4)` | `children[branch_child_idx]` discriminant (Blinded vs Revealed) |
| **9.02%** | `mov 0x8(%r8),%rcx` | SmallVec spilled heap pointer dereference |
| **8.93%** | `movzbl 0xc6(%r8),%eax` | `arena[head_idx]` node enum discriminant tag |
| **8.35%** | `mov (%r8),%r8` | SmallVec spilled heap pointer (second code path) |
| **7.62%** | `mov 0x90(%r8),%rdx` | SmallVec `.capacity` field (spill check) |
| **7.16%** | `mov 0x4(%rdx),%esi` | Reading `child_idx` from `Revealed(Index)` variant |
| **5.13%** | `mov 0x8(%rdx),%edx` | Reading second field of child entry |
| **5.02%** | `mov 0xc4(%r8),%esi` | `branch.state_mask` field |
| **4.48%** | `mov 0x98(%r8),%rax` | `branch.short_key.len` (first logical path call) |
| **3.61%** | `mov 0x98(%rcx),%r8` | `branch.short_key.len` (second logical path call) |
| **2.88%** | `mov 0xf0(%rdx,%r8,1),%ecx` | SlotMap version check (current node) |
| **2.51%** | `cmpb $0x3,0xc6(%rcx)` | Child node discriminant tag (after descending) |
| **2.33%** | `mov 0xf0(%rdi,%rcx,1),%esi` | SlotMap version check (child node) |

Two structural problems make this chain worse than it needs to be:

**1. SmallVec heap spills dominate L3 misses (39.6% of seek's misses).** Branch nodes store children in a `SmallVec<[ArenaSparseNodeBranchChild; 4]>`. When a branch has >4 children (common near the root), the SmallVec spills to a heap allocation. This inserts an extra dependent cache miss — chasing a heap pointer to a separate allocation with no spatial locality to the arena node. Three dependent misses fire in sequence: capacity check → heap pointer → child entry.

**2. The child type is too fat for the hot path (14.6% of seek's misses).** `ArenaSparseNodeBranchChild` is an enum carrying either `Index` (small) or `RlpNode` (up to 33 bytes). But `seek` only needs to know: is the child revealed? What's its index? Loading a fat enum just to check a discriminant wastes cache lines.

Additionally, `ArenaSparseNode` itself is a large heterogeneous enum (~268 bytes, spanning 5 cache lines). The seek path only needs a small subset of fields (discriminant, state_mask, short_key, child references), but every node access pulls in cold data like cached RLP encodings, leaf values, and branch masks.


pepyakin:
derek bench blocks=big

pepyakin:
derek bench blocks=big feature-args="--engine.enable-arena-sparse-trie" baseline-args="--engine.enable-arena-sparse-trie"

pepyakin:
derek bench blocks=big feature-args="--engine.enable-arena-sparse-trie" baseline-args="--engine.enable-arena-sparse-trie"

PR #23048, opened 2026-03-16
chore(engine-tree): simplify return type of canonical_block_by_hash

duyquang6:
canonical_block_by_hash do not return Ok(None), so can simplify return type of this fn


yongkangc (approved):
think this makes sense @mediocregopher @shekhirin ?

mattsse (approved):
Confirmed — `canonical_block_by_hash` converts missing data to `ProviderError` via `ok_or_else`, so the `Option` wrapper and all `None` handling were unreachable. Clean simplification.

PR #23050, opened 2026-03-16
fix(rpc): clone EthSigner trait objects with generic tx request

BearSoltokien:
Include `TxReq` in the `EthSigner` trait-object clone macro, this restores cloning for custom RPC transaction request types.


mattsse (approved):
Fix is correct — the clone macro needs both type params to support custom `TxReq` types.

PR #23053, opened 2026-03-16
fix(trie): Fix branch collapse edge-cases in ArenaParallelSparseTrie

mediocregopher:
Fixes two cases where ArenaParallelSparseTrie::update_leaves did not correctly handle subtries having all leaves removed and being deleted from the trie. In both cases the bug appears when parallelism is triggered.

## Case 1

```
root (branch)
 └─ 0xd (branch, 3 children)
      ├─ 0xd7 → Subtrie (1 leaf, revealed)
      ├─ 0xd8 → Subtrie (1 leaf, BLINDED — no update touches it)
      └─ 0xdd → Subtrie (1 leaf, revealed)
```

* Both 0xd7 and 0xdd become EmptyRoot
* 0xd is left with a single child, collapsing
* The single remaining child (0xd8) is blinded, preventing collapse

This case did not have a test which previously covered it, now it does.

## Case 2

```
root (branch) 
       ├─ 0x1 (branch, revealed)
       │    ├─ 0x10 → Subtrie (leaf, revealed)  ← emptied
       │    └─ 0x11 → Subtrie (leaf, revealed)  ← emptied
       └─ 0x2 → BLINDED
```

* Both 0x10 and 0x11 become EmptyRoot
* 0x1 is deleted as well
* 0x (root) is left with a single child, collapsing
* The single remaining child (0x2) is blinded, preventing collapse

This case already had a test case covering it, but it wasn't triggered because we weren't testing ArenaParallelSparseTrie with low enough parallelism thresholds.

## Test Suite <> Parallelism Coverage

The SparseTrie test suite was previously only equipped to test ArenaParallelSparseTrie's default value, but these error cases are triggered when parallelism is triggered, which the tests suite's tests wouldn't do.

The test suite is modified so that all test cases accept a constructor closure, and we now test ArenaParallelSparseTrie with both default parallelism thresholds and with thresholds of 1 (ie, always enabled).

### DIFF
--- a/crates/trie/sparse/src/arena/mod.rs
+++ b/crates/trie/sparse/src/arena/mod.rs
@@ -2822,7 +2822,10 @@ impl SparseTrie for ArenaParallelSparseTrie {
                     // can detect blinded siblings and request proofs.
                     let all_removals = subtrie_updates
                         .iter()
-                        .filter(|(_, _, u)| matches!(u, LeafUpdate::Changed(_))) // Ignore Touched
+                        // Filter out Touched, as they don't affect the structure of the trie. So an
+                        // update set with 2 removals and one Touched could still result in an empty
+                        // sub trie.
+                        .filter(|(_, _, u)| matches!(u, LeafUpdate::Changed(_)))
                         .all(|(_, _, u)| matches!(u, LeafUpdate::Changed(v) if v.is_empty()));
                     let subtrie_num_leaves = match &self.upper_arena[child_idx] {
                         ArenaSparseNode::Subtrie(s) => s.num_leaves,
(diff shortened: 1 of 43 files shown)


crates/trie/sparse/src/arena/mod.rs
```diff
@@ -2817,7 +2817,21 @@ impl SparseTrie for ArenaParallelSparseTrie {
 
                     let num_subtrie_updates = update_idx - subtrie_start;
 
-                    if num_subtrie_updates >= threshold {
+                    // If all updates are removals and could empty the subtrie,
+                    // force inline processing so the upper-arena collapse logic
+                    // can detect blinded siblings and request proofs.
+                    let all_removals = subtrie_updates
+                        .iter()
+                        .filter(|(_, _, u)| matches!(u, LeafUpdate::Changed(_))) // Ignore Touched
```
pepyakin: Could you please leave a better documentation as to why we are ignoring touched here? 🙏

pepyakin (approved):

PR #23062, opened 2026-03-17
fix(txpool): ignore ENOENT when removing tx backup after reinsertion

BearSoltokien:
Ignore ENOENT when deleting txpool backup files after successful reinsertion, and preserve all non-NotFound filesystem errors with targeted tests.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -32,6 +32,7 @@ use serde::{Deserialize, Serialize};
 use std::{
     borrow::Borrow,
     hash::{Hash, Hasher},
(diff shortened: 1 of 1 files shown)


mattsse (commented):
Thanks for the fix. I'm going to pass on this as-is.

This is addressing a very small cleanup-path edge case, but the patch adds a dedicated helper plus targeted tests around the helper behavior. That feels like more surface area than we want here for what is essentially "ignore NotFound when deleting the backup file after successful reinsertion."

If we revisit this, I'd prefer a much smaller change that keeps the handling inline at the call site and avoids adding new helper/test weight for it.

PR #23063, opened 2026-03-17
perf(rpc): fast path `transaction_receipt` when block & receipts are cached

stevencartavia:
When both the block and receipts are already in the cache, `eth_getTransactionReceipt` was redundantly calling `get_receipts` a second time inside `build_transaction_receipt`. This adds an early return that converts the receipt in-place from the cached data.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/transaction.rs
@@ -233,6 +233,13 @@ pub trait EthTransactions: LoadTransaction<Provider: BlockReaderIdExt> {
         Self: LoadReceipt + 'static,
     {
         async move {
+            // Fast path: block and receipts both cached — convert in-place.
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
Thanks for identifying the redundant `get_receipts` call — the observation is correct. However this approach adds a second `get_transaction_by_hash` lookup before `load_transaction_and_receipt`, which already does the same cache check. On cache miss the flow now does 3 lookups instead of 2.

The root issue is that `load_transaction_and_receipt` discards the full receipts list and returns a single receipt, forcing `build_transaction_receipt` to re-fetch all receipts for gas accounting. A better fix would be to restructure these two functions so the cached receipts flow through without a second lookup — e.g. having `load_transaction_and_receipt` return something like `CachedTransaction` (or a similar type that optionally carries the full receipts vec), so `build_transaction_receipt` can use them directly when available.

Will open an issue for this.

PR #23064, opened 2026-03-17
perf(rpc): avoid redundant `next_env_attributes` call in `simulate_v1`

stevencartavia:
In `simulate_v1`, `next_env_attributes` was called twice per iteration for the same parent — once for `next_evm_env` and again for `context_for_next_block`. Extracted to a local variable.

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/call.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/call.rs
@@ -127,9 +127,11 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
                         .into());
                     }
 
+                    let attributes = this.next_env_attributes(&parent)?;
+
                     let mut evm_env = this
                         .evm_config()
-                        .next_evm_env(&parent, &this.next_env_attributes(&parent)?)
+                        .next_evm_env(&parent, &attributes)
                         .map_err(RethError::other)
                         .map_err(Self::Error::from_eth_err)?;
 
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -321,23 +323,23 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
...
-            let block = self
-                .recovered_block(target_block)
-                .await?
-                .ok_or(EthApiError::HeaderNotFound(target_block))?;
+                self.cache()
+                    .get_recovered_block(block_hash)
+                    .await
+                    .map_err(Self::Error::from_eth_err)?
+                    .ok_or(EthApiError::HeaderNotFound(target_block))?
+            };
```
mattsse: idk about this, unclear why

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -321,23 +323,23 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
             let mut target_block = block_number.unwrap_or_default();
             let is_block_target_pending = target_block.is_pending();
 
-            // if it's not pending, we should always use block_hash over block_number to ensure that
-            // different provider calls query data related to the same block.
```
mattsse, line 325: why remove this

crates/rpc/rpc-eth-api/src/helpers/call.rs
```diff
@@ -205,7 +207,7 @@ pub trait EthCall: EstimateCall + Call + LoadPendingBlock + LoadBlock + FullEthA
 
                     let ctx = this
                         .evm_config()
-                        .context_for_next_block(&parent, this.next_env_attributes(&parent)?)
+                        .context_for_next_block(&parent, attributes)
```
mattsse, line 210: this is valid, can we submit this separately, I find the other changes harder to evaluate and they arent related afaict

mattsse (changes requested):

mattsse (approved):

PR #23067, opened 2026-03-17
refactor(storage): use RocksReadSnapshot for read-only compatible RocksDB reads

decofe:
The `re-execute` command opens RocksDB with `AccessRights::RO`, but the read path created a write transaction via `RocksTx::tx()` → `db_rw()`, which panics on read-only providers.

Introduces `RocksReadSnapshot` — a point-in-time snapshot enum over `SnapshotWithThreadMode<OptimisticTransactionDB>` and `SnapshotWithThreadMode<DB>` — that works in both read-only and read-write modes. Snapshots are lighter than transactions (no write-conflict tracking) and are `Send + Sync`.

- `EitherReader::RocksDB` now holds `RocksReadSnapshot` instead of `&RocksTx`
- `with_rocksdb_tx` replaced by `with_rocksdb_snapshot` which passes a snapshot instead of creating a write transaction
- Deleted duplicated `RocksTx` history methods (`account_history_info`, `storage_history_info`, `history_info`) — `RocksReadSnapshot` is now the sole owner
- Deleted unused `cf_handle_rw` helper
- Added read-only regression test

Co-Authored-By: Tim <12827757+laibe@users.noreply.github.com>

### DIFF
--- a/crates/storage/provider/src/either_writer.rs
+++ b/crates/storage/provider/src/either_writer.rs
@@ -68,11 +68,11 @@ pub type RocksBatchArg<'a> = crate::providers::rocksdb::RocksDBBatch<'a>;
 /// The raw `RocksDB` batch type returned by [`EitherWriter::into_raw_rocksdb_batch`].
 pub type RawRocksDBBatch = rocksdb::WriteBatchWithTransaction<true>;
 
-/// Helper type for `RocksDB` transaction reference argument in reader constructors.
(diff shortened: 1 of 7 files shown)


yongkangc (approved):
i see, makes sense opens the DB in read-only mode but using write txn

Rjected (approved):
lgtm, enum -> snapshot makes sense for getting around the optimistictxdb / ro panic

PR #23070, opened 2026-03-17
fix(provider): prevent stale state reads during pipeline sync

### LINKED ISSUE #22873: Stale state returned during/after backfill sync via `history_by_block_hash` (missing `ensure_canonical_block` guard)
### Describe the bug

During and after backfill (pipeline) sync, RPC queries that resolve state via `history_by_block_hash` can silently return **stale or incorrect state** for blocks whose headers have been downloaded but whose execution has not yet completed. This affects any RPC method that resolves state by block hash — including `eth_getBalance`, `eth_call`, `eth_getStorageAt`, and all `debug_trace*` methods.

### Root cause

There are two interacting bugs:

**Bug 1: `history_by_block_hash` lacks the `ensure_canonical_block` guard**

[`history_by_block_number`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/provider/src/providers/blockchain_provider.rs#L554-L565) correctly calls `ensure_canonical_block(block_number)` before returning state, rejecting queries for blocks that haven't been fully synced. However, [`history_by_block_hash`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/provider/src/providers/blockchain_provider.rs#L567-L569) skips this check entirely:

```rust
// blockchain_provider.rs L567-569 — NO guard
fn history_by_block_hash(&self, block_hash: BlockHash) -> ProviderResult<StateProviderBox> {
    self.consistent_provider()?.into_state_provider_at_block_hash(block_hash)
}
```

Critically, `history_by_block_hash` is the path used by **most queries**, not `history_by_block_number`. [`state_by_block_number_or_tag(Number(num))`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/provider/src/providers/blockchain_provider.rs#L545-L550) converts the number to a hash first, then routes through `state_by_block_hash` → `history_by_block_hash` — bypassing the guarded `history_by_block_number` path. The same applies to [`state_by_block_id(BlockId::Hash(...))`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/storage-api/src/state.rs#L152-L157) which routes directly to `history_by_block_hash`. This means `history_by_block_number` (with its guard) is effectively dead code for most queries — only used for `Earliest`.

The `debug_trace*` methods are also affected: `debug_traceBlockByNumber` resolves the block, then calls `spawn_with_state_at_block(block.parent_hash().into(), ...)`, which routes through `state_at_block_id(BlockId::Hash(parent_hash))` → `history_by_block_hash` — the unguarded path.

**Bug 2: `ConsistentProvider::best_block_number()` returns header height, not execution height**

Even if bug 1 is fixed by adding `ensure_canonical_block` to `history_by_block_hash`, the guard itself has a gap. [`ensure_canonical_block`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/provider/src/providers/consistent.rs#L631-L638) checks against [`best_block_number()`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/storage/provider/src/providers/consistent.rs#L762-L766), which falls back to `self.last_block_number()` when `head_block` is `None`:

```rust
// consistent.rs L762-763
fn best_block_number(&self) -> ProviderResult<BlockNumber> {
    self.head_block.as_ref().map(|b| Ok(b.number())).unwrap_or_else(|| self.last_block_number())
}
```

`last_block_number()` reads the highest entry in the `CanonicalHeaders` table — the **header download frontier**, which is far ahead of execution during staged sync. This means `ensure_canonical_block` compares against headers-downloaded (not state-executed), making it pass for blocks whose state doesn't exist yet.

`head_block` is `None` after every backfill sync because [`on_backfill_sync_finished`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/engine/tree/src/tree/mod.rs#L1741-L1753) calls `clear_state()` (line 1743), which empties the in-memory state. Backfill is triggered when the gap exceeds [`MIN_BLOCKS_FOR_PIPELINE_RUN = EPOCH_SLOTS = 32`](https://github.com/paradigmxyz/reth/blob/cdeba79590a3c5339b398e5136cdc355cba5f20d/crates/engine/tree/src/tree/mod.rs#L89) blocks.

### Observable symptoms

1. **`eth_getBalance` / `eth_getStorageAt`**: Returns stale balances for recent blocks. For native balances, the mempool rejects transactions based on stale state (loud failure). For ERC20 balances, the stale values propagate silently into EVM execution.

2. **`debug_traceBlockByNumber`**: Re-executes block transactions against stale parent state. Manifests as arithmetic overflow panics (`Panic(0x11)`) in DEX router calls where reserves are stale, or phantom transactions appearing/disappearing in traces vs. stored receipts.

3. **`eth_call` / `eth_estimateGas`**: Returns incorrect results when called against recent block numbers.

## Suggested fix

Both bugs need to be addressed together:

**Fix 1**: Add `ensure_canonical_block` guard to `history_by_block_hash`:

```rust
fn history_by_block_hash(&self, block_hash: BlockHash) -> ProviderResult<StateProviderBox> {
    let provider = self.consistent_provider()?;
    let block_number = provider
        .block_number(block_hash)?
        .ok_or(ProviderError::BlockHashNotFound(block_hash))?;
    provider.ensure_canonical_block(block_number)?;
    provider.into_state_provider_at_block_hash(block_hash)
}
```

**Fix 2**: Change `best_block_number()` fallback from `last_block_number()` (header height) to `storage_provider.best_block_number()` (Finish stage checkpoint = actual execution progress):

```rust
fn best_block_number(&self) -> ProviderResult<BlockNumber> {
    self.head_block
        .as_ref()
        .map(|b| Ok(b.number()))
        .unwrap_or_else(|| self.storage_provider.best_block_number())
}
```

The database provider's `best_block_number()` returns the `StageId::Finish` checkpoint, which reflects the highest block where all pipeline stages (including Execution) have completed.

## Downstream report and fix

This was discovered and is being fixed downstream in bnb-chain/reth (BSC fork):
- **Issue**: [bnb-chain/reth-bsc#273](https://github.com/bnb-chain/reth-bsc/issues/273)
- **Fix PR**: [bnb-chain/reth#101](https://github.com/bnb-chain/reth/pull/101) (includes both fixes)

### Steps to reproduce

1. Run a reth node on a fast-block chain (e.g., BSC at 0.45s blocks)
2. Wait for the node to sync and begin serving traffic
3. Introduce brief network lag (>14 seconds for BSC, >32 blocks for the target chain) to trigger a backfill sync cycle
4. During or immediately after backfill, query `eth_getBalance` or `debug_traceBlockByNumber` for a recent block
5. Compare the result against a fully-synced reference node

The stale state is silent — no error is returned, just incorrect data.

### Node logs

```text

```

### Platform(s)

_No response_

### Container Type

Kubernetes

v1.11.1

2

mainnet

Archive (default)

_No response_

_No response_

### Code of Conduct

decofe:
Closes #22873 (Bug 2)

Companion to #22876 (Bug 1). Three changes to prevent stale state during pipeline sync:

**1. `ConsistentProvider::best_block_number()`** — falls back to `storage_provider.best_block_number()` (`StageId::Finish` checkpoint) instead of `last_block_number()` (header download frontier). Without this, `ensure_canonical_block` passes for blocks whose headers exist but state hasn't been computed.

**2. `try_into_history_at_block`** — when `block_number == best_block` (Finish), checks whether `StageId::Execution` has moved past it. Pipeline stages commit independently, so after Execution commits N+1..N+K the plain/hashed state is at N+K while Finish is at N. In this window neither `LatestStateProvider` nor `HistoricalStateProvider` can safely serve the Finish-tip block — `LatestStateProvider` reads plain state at N+K, and `HistoricalStateProvider`'s `InPlainState` fallback does the same when the history index has no modification after the target. Returns `StateUnavailableDuringSync` instead of wrong data.

**3. `StateUnavailableDuringSync` error variant** — replaces the misleading `BlockNotExecuted` (the block *was* executed, we just can't safely read it while the pipeline is running). Gives callers a clear "try again later" signal.

Prompted by: joshie

### DIFF
--- a/crates/storage/provider/src/providers/consistent.rs
+++ b/crates/storage/provider/src/providers/consistent.rs
@@ -760,7 +760,10 @@ impl<N: ProviderNodeTypes> BlockNumReader for ConsistentProvider<N> {
     }
 
     fn best_block_number(&self) -> ProviderResult<BlockNumber> {
-        self.head_block.as_ref().map(|b| Ok(b.number())).unwrap_or_else(|| self.last_block_number())
+        self.head_block
+            .as_ref()
+            .map(|b| Ok(b.number()))
+            .unwrap_or_else(|| self.storage_provider.best_block_number())
     }
 
     fn last_block_number(&self) -> ProviderResult<BlockNumber> {
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -14,11 +14,11 @@ use crate::{
     BlockReader, BlockWriter, BundleStateInit, ChainStateBlockReader, ChainStateBlockWriter,
     DBProvider, EitherReader, EitherWriter, EitherWriterDestination, HashingWriter, HeaderProvider,
     HeaderSyncGapProvider, HistoricalStateProvider, HistoricalStateProviderRef, HistoryWriter,
-    LatestStateProvider, LatestStateProviderRef, OriginalValuesKnown, ProviderError,
-    PruneCheckpointReader, PruneCheckpointWriter, RawRocksDBBatch, RevertsInit, RocksBatchArg,
-    RocksDBProviderFactory, StageCheckpointReader, StateProviderBox, StateWriter,
-    StaticFileProviderFactory, StatsReader, StorageReader, StorageTrieWriter, TransactionVariant,
-    TransactionsProvider, TransactionsProviderExt, TrieWriter,
+    LatestStateProviderRef, OriginalValuesKnown, ProviderError, PruneCheckpointReader,
+    PruneCheckpointWriter, RawRocksDBBatch, RevertsInit, RocksBatchArg, RocksDBProviderFactory,
+    StageCheckpointReader, StateProviderBox, StateWriter, StaticFileProviderFactory, StatsReader,
+    StorageReader, StorageTrieWriter, TransactionVariant, TransactionsProvider,
+    TransactionsProviderExt, TrieWriter,
 };
 use alloy_consensus::{
     transaction::{SignerRecoverable, TransactionMeta, TxHashRef},
@@ -871,10 +871,13 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
             });
         }
 
-        // If requesting state at the best block, use the latest state provider
-        if block_number == best_block {
-            return Ok(Box::new(LatestStateProvider::new(self)));
-        }
+        // Note: we intentionally do NOT short-circuit with `LatestStateProvider` when
+        // `block_number == best_block`. `best_block` is the `StageId::Finish` checkpoint, but
+        // during pipeline sync the Execution stage may have already committed state changes for
+        // later blocks, advancing the plain/hashed state tables past `best_block`.
+        // `LatestStateProvider` reads those tables directly and would return state from the
+        // execution frontier instead of the requested block. `HistoricalStateProvider` correctly
+        // reconstructs state via changesets regardless of where plain state currently sits.
 
         // +1 as the changeset that we want is the one that was applied after this block.
         block_number += 1;
--- a/crates/storage/provider/src/providers/database/provider.rs
+++ b/crates/storage/provider/src/providers/database/provider.rs
@@ -14,11 +14,11 @@ use crate::{
     BlockReader, BlockWriter, BundleStateInit, ChainStateBlockReader, ChainStateBlockWriter,
     DBProvider, EitherReader, EitherWriter, EitherWriterDestination, HashingWriter, HeaderProvider,
     HeaderSyncGapProvider, HistoricalStateProvider, HistoricalStateProviderRef, HistoryWriter,
-    LatestStateProviderRef, OriginalValuesKnown, ProviderError, PruneCheckpointReader,
-    PruneCheckpointWriter, RawRocksDBBatch, RevertsInit, RocksBatchArg, RocksDBProviderFactory,
-    StageCheckpointReader, StateProviderBox, StateWriter, StaticFileProviderFactory, StatsReader,
-    StorageReader, StorageTrieWriter, TransactionVariant, TransactionsProvider,
-    TransactionsProviderExt, TrieWriter,
+    LatestStateProvider, LatestStateProviderRef, OriginalValuesKnown, ProviderError,
+    PruneCheckpointReader, PruneCheckpointWriter, RawRocksDBBatch, RevertsInit, RocksBatchArg,
+    RocksDBProviderFactory, StageCheckpointReader, StateProviderBox, StateWriter,
+    StaticFileProviderFactory, StatsReader, StorageReader, StorageTrieWriter, TransactionVariant,
+    TransactionsProvider, TransactionsProviderExt, TrieWriter,
 };
 use alloy_consensus::{
     transaction::{SignerRecoverable, TransactionMeta, TxHashRef},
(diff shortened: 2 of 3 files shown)


crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -871,8 +871,23 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
...
+            // If so, plain state is ahead and neither `LatestStateProvider` (reads plain state
+            // directly) nor `HistoricalStateProvider` (falls back to `InPlainState` for keys
+            // without indexed modifications after the target) can safely serve this block.
+            let execution_tip = self
+                .get_stage_checkpoint(StageId::Execution)?
+                .map(|c| c.block_number)
+                .unwrap_or_default();
+
+            if execution_tip > best_block {
+                return Err(ProviderError::BlockNotExecuted {
```
joshieDo: this error variant is technically wrong, but if the `IndexStorage/AccountHistoryStage ` haven't been executed our best bet is either:
* throw some error, since we're are not ready yet (plainstate is at N+K, and historical state hasnt processed changesets) OR
* connect somehow historicalstateprovider to just the existing changesets (at N+K) without relying on history tables.

crates/storage/provider/src/providers/consistent.rs
mattsse, line 1: can we submit this separately?

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -871,8 +871,20 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
...
+            // If so, plain state is ahead and neither `LatestStateProvider` (reads plain state
+            // directly) nor `HistoricalStateProvider` (falls back to `InPlainState` for keys
+            // without indexed modifications after the target) can safely serve this block.
+            let execution_tip = self
+                .get_stage_checkpoint(StageId::Execution)?
+                .map(|c| c.block_number)
+                .unwrap_or_default();
+
+            if execution_tip > best_block {
+                return Err(ProviderError::StateUnavailableDuringSync(block_number));
```
mattsse, line 885: imo always fetching this doesnt feel right

mattsse (changes requested):

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -871,8 +871,20 @@ impl<TX: DbTx + 'static, N: NodeTypes> TryIntoHistoricalStateProvider for Databa
             });
         }
 
-        // If requesting state at the best block, use the latest state provider
         if block_number == best_block {
```
phil-k1, line 874: Just to double-check: Does it suffice to handle the case `block_number == best_block` here?

I.e. if `block_number < best_block` and if `account_history_lookup`/`storage_history_lookup` later resolve to `HistoryInfo::InPlainState | HistoryInfo::MaybeInPlainState`, wouldn't we still use the `PlainState` from `execution_tip > best_block`?

phil-k1: FWIW that's the fix that was merged for addressing the same issue on BSC Reth downstream: https://github.com/bnb-chain/reth/pull/113

PR #23071, opened 2026-03-17
perf(engine): check hashmap instead of clone

jsvisa:
Currently if we want to look up a block by hash, we will clone the full `SealedHeader`, then immediately discarded it, it's a waste of clone: https://github.com/paradigmxyz/reth/blob/fd05a2ee88630711ae0adac4ddb28eb46a4ec5d5/crates/engine/tree/src/tree/state.rs#L83-L86

So we can replace `sealed_header_by_hash().is_some()` with lightweight existence checks in the engine tree.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2057,6 +2057,17 @@ where
         ))
     }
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense ty

in the happy case this will remove a header clone

PR #23081, opened 2026-03-17
fix(provider): fix race between save_blocks and rocksdb pruning

Rjected:
Right now `save_blocks` and the pruner both push to `pending_rocksdb_batches` before committing. The pruner doesn't see the shard that save_blocks wrote (it's still pending), so it reads the old committed shard, filters it, and pushes its own version. When `provider_rw.commit()` drains the batches in order, the pruner's batch overwrites the `save_blocks` batch for the same `ShardedKey(addr, u64::MAX)`:

* committed shard for addr `0xa`: `[100, 200, 300, 400, 500]`
* `save_blocks` pushes batch (pending): `[100, 200, 300, 400, 500, 600, 700]`
* pruner reads committed, pushes batch (pending): `[400, 500]`
* `commit` drains: `save_blocks` batch first, then pruner batch
This results in a final shard of `[400, 500]`, the new blocks are lost and not in the shard.

To prevent this, we now commit the result of `save_blocks` before running the pruner, so the pruner reads committed state that includes the new entries.

### DIFF
--- a/crates/engine/tree/src/persistence.rs
+++ b/crates/engine/tree/src/persistence.rs
@@ -177,17 +177,24 @@ where
                 }
             }
 
+            provider_rw.commit()?;
+            debug!(target: "engine::persistence", first=?first_block, last=?last_block, "Saved range of blocks");
+
+            // Run the pruner in a separate provider so it reads committed RocksDB state
+            // that includes the history entries written by save_blocks above.
+            //
+            // The pruner reads the indices from rocksdb, filters it, and writes to indices, so it
+            // must be able to read anything written by save_blocks.
             if self.pruner.is_pruning_needed(last.number) {
                 debug!(target: "engine::persistence", block_num=?last.number, "Running pruner");
                 let prune_start = Instant::now();
+                let provider_rw = self.provider.database_provider_rw()?;
                 let _ = self.pruner.run_with_provider(&provider_rw, last.number)?;
+                provider_rw.commit()?;
+                debug!(target: "engine::persistence", tip=?last.number, "Finished pruning after saving blocks");
                 self.metrics.prune_before_duration_seconds.record(prune_start.elapsed());
             }
-
-            provider_rw.commit()?;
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
lgtm, this also handles crash between the two commits

crates/engine/tree/src/persistence.rs
```diff
@@ -177,14 +177,20 @@ where
                 }
             }
 
+            provider_rw.commit()?;
+
+            // Run the pruner in a separate provider so it reads committed RocksDB state
+            // that includes the history entries written by save_blocks above.
+            // Previously both ran before a single commit, causing the pruner's batch to
```
mattsse: this comment (previously) will just be confusing with out this pr's context, please rephrase.

mediocregopher (approved):

crates/engine/tree/src/persistence.rs
```diff
@@ -177,14 +177,20 @@ where
...
+                let provider_rw = self.provider.database_provider_rw()?;
                 let _ = self.pruner.run_with_provider(&provider_rw, last.number)?;
+                provider_rw.commit()?;
                 self.metrics.prune_before_duration_seconds.record(prune_start.elapsed());
             }
-
-            provider_rw.commit()?;
         }
 
         debug!(target: "engine::persistence", first=?first_block, last=?last_block, "Saved range of blocks");
```
joshieDo: can we move this above , after line 180 ?

joshieDo: and add a similar one after the pruner run

crates/engine/tree/src/persistence.rs
Rjected: rephrased

crates/engine/tree/src/persistence.rs
Rjected: added another log yep

PR #23082, opened 2026-03-17
feat(prune): make minimum pruning distance configurable

decofe:
## Summary

Make the minimum safe unwind / pruning distance configurable instead of hardcoded to `MINIMUM_UNWIND_SAFE_DISTANCE` (10,064).

## Motivation

Chains with different reorg assumptions need lower pruning distances. E2E tests also need smaller values to test pruning without producing 10k+ blocks.

## Testing

Existing tests pass — the default value (10,064) preserves current behavior. The new field is only active when explicitly set via CLI or config.

Co-Authored-By: Dan Cline <6798349+Rjected@users.noreply.github.com>

Prompted by: Dan

### DIFF
--- a/crates/cli/commands/src/download/config_gen.rs
+++ b/crates/cli/commands/src/download/config_gen.rs
@@ -199,11 +199,7 @@ pub(crate) fn config_for_selections(
         }
 
         return Config {
-            prune: PruneConfig {
-                block_interval: PruneConfig::default().block_interval,
-                minimum_pruning_distance: PruneConfig::default().minimum_pruning_distance,
-                segments,
-            },
+            prune: PruneConfig { segments, ..Default::default() },
             static_files,
             ..Default::default()
         };
--- a/crates/cli/commands/src/download/config_gen.rs
+++ b/crates/cli/commands/src/download/config_gen.rs
@@ -199,7 +199,11 @@ pub(crate) fn config_for_selections(
         }
 
         return Config {
-            prune: PruneConfig { block_interval: PruneConfig::default().block_interval, segments },
+            prune: PruneConfig {
+                block_interval: PruneConfig::default().block_interval,
+                minimum_pruning_distance: PruneConfig::default().minimum_pruning_distance,
+                segments,
+            },
             static_files,
             ..Default::default()
         };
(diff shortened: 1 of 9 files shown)


crates/node/core/src/args/pruning.rs
```diff
@@ -220,21 +225,29 @@ impl PruningArgs {
                     .block_number()
                     .map(PruneMode::Before);
             }
-            config = PruneConfig { block_interval: config.block_interval, segments }
+            config = PruneConfig {
+                block_interval: config.block_interval,
+                segments,
+                ..Default::default()
```
yongkangc:
```suggestion
                minimum_pruning_distance: config.minimum_pruning_distance,
```
should this be `minimum_pruning_distance: config.minimum_pruning_distance` instead? so we dont drop the user's TOML-set minimum_pruning_distance

crates/node/core/src/args/pruning.rs
```diff
@@ -220,21 +225,29 @@ impl PruningArgs {
...
+            }
         }
 
         // If --minimal is set, use minimal storage mode with aggressive pruning.
         if self.minimal {
             config = PruneConfig {
                 block_interval: config.block_interval,
                 segments: DefaultPruningValues::get_global().minimal_prune_modes.clone(),
+                ..Default::default()
```
yongkangc:
```suggestion
                minimum_pruning_distance: config.minimum_pruning_distance,
```
should this be `minimum_pruning_distance: config.minimum_pruning_distance` instead? so we dont drop the user's TOML-set minimum_pruning_distance

same here?

yongkangc (approved):

crates/cli/commands/src/download/config_gen.rs
```diff
@@ -199,7 +199,11 @@ pub(crate) fn config_for_selections(
         }
 
         return Config {
-            prune: PruneConfig { block_interval: PruneConfig::default().block_interval, segments },
+            prune: PruneConfig {
+                block_interval: PruneConfig::default().block_interval,
+                minimum_pruning_distance: PruneConfig::default().minimum_pruning_distance,
```
mattsse: isnt this the same as segments,..Default::default()

Rjected: yep fixed

crates/node/core/src/args/pruning.rs
Rjected: yep fixed

crates/node/core/src/args/pruning.rs
Rjected: yep fixed

PR #23083, opened 2026-03-18
fix(provider): disable read transaction timeout during check_consistency

decofe:
Healing can iterate changesets over millions of blocks, easily exceeding the default 5-minute MDBX read transaction timeout. When the transaction gets killed mid-heal, the node restarts and re-enters healing, creating an infinite crash loop.

Calls `disable_long_read_transaction_safety()` on the read transaction in `check_consistency` so healing can complete regardless of the configured timeout.

Co-Authored-By: Arsenii Kulikov <62447812+klkvr@users.noreply.github.com>

Prompted by: klkvr

### DIFF
--- a/crates/storage/provider/src/providers/database/mod.rs
+++ b/crates/storage/provider/src/providers/database/mod.rs
@@ -18,7 +18,7 @@ use core::fmt;
 use parking_lot::RwLock;
 use reth_chainspec::ChainInfo;
 use reth_db::{init_db, mdbx::DatabaseArguments, DatabaseEnv};
-use reth_db_api::{database::Database, models::StoredBlockBodyIndices};
+use reth_db_api::{database::Database, models::StoredBlockBodyIndices, transaction::DbTx};
 use reth_errors::{RethError, RethResult};
 use reth_node_types::{
     BlockTy, HeaderTy, NodeTypesWithDB, NodeTypesWithDBAdapter, ReceiptTy, TxTy,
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -349,7 +349,12 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
     /// consistency. I.e. this MAY result in writes to the static files.
     #[instrument(err, skip(self))]
     pub fn check_consistency(&self) -> ProviderResult<(Option<u64>, Option<u64>)> {
-        let provider_ro = self.database_provider_ro()?;
+        let mut provider_ro = self.database_provider_ro()?;
+
+        // Healing can run long-lived read transactions (e.g. iterating changesets
+        // over millions of blocks). Disable the default timeout so MDBX doesn't
+        // kill the transaction mid-heal, which causes a crash loop on startup.
+        provider_ro.tx_mut().disable_long_read_transaction_safety();
```
yongkangc: is there any concern with totally disabling timeouts? DB file growth etc?

yongkangc (approved):
left one question, but from my research think is fine

crates/storage/provider/src/providers/database/mod.rs
shekhirin: it's only disabled for healing, so no

PR #23084, opened 2026-03-18
fix(node): fail fast on invalid storage settings metadata

BearSoltokien:
Fail node startup when persisted storage_settings metadata is malformed, while keeping lossy parsing for compatibility tooling paths.


mattsse (commented):
dont think we need this

PR #23085, opened 2026-03-18
perf(rpc): avoid request clone in `eth_createAccessList`

stevencartavia:
Read `gas_limit` and `access_list` from the request upfront to avoid cloning the full `RpcTxReq` in `create_access_list_with`.


mattsse (approved):
Correct and clean — hoists two cheap reads before the move, avoiding the full request clone while preserving semantics.

PR #23090, opened 2026-03-18
fix(ethstats): Re-enable TLS in tokio-tungstenite

MysticRyuujin:
Basically at the same time that #21595 was merged, #21584 was also merged and it broke TLS support for ethstats

Trying to use ethstats with `wss://` right now results in:

```text
ERROR shutting down due to error
Error: WebSocket error: URL error: TLS support not compiled in

Caused by:
   0: URL error: TLS support not compiled in
   1: TLS support not compiled in

Location:
    /app/crates/node/builder/src/launch/common.rs:1101:24
```

Note: I really don't know why #21584 removed `rustls-tls-native-roots`? So open to alternative fixes...


mattsse (approved):
i see, ty!

PR #23094, opened 2026-03-18
fix: use zero gas price for empty blocks

klkvr:
RIght now if any of the sampled blocks are empty in gas price oracle we will use previous gas price estimate for them.

For chains with a lot of activity this logic is effectively unreachable, but for chains with low activity and many empty blocks this might be problematic because priority fee estimate will basically never be updated.


shekhirin (approved):

mattsse (approved):
makes sense, I believe using the last_price here was even a bug that 's just never reachable on mainnet

PR #23104, opened 2026-03-19
fix(download): wire Ctrl+C to cancellation token and fix progress labels

decofe:
Two bugs in `reth download`:

1. **Ctrl+C doesn't stop downloads** — `CancellationToken` was created but never wired to signal handlers, so blocking I/O in `spawn_blocking` threads keeps running after Ctrl+C. Fixed by spawning a task that listens for SIGINT/SIGTERM and cancels the token. Also added cancellation checks at the top of retry loops.

2. **Misleading progress message** — progress always showed `Downloading and extracting...` even during extraction-only phases (resumable two-phase downloads) or download-only phases. Now shows `Downloading`, `Extracting`, or `Downloading and extracting` depending on the actual phase.

Co-Authored-By: Alexey Shekhirin <5773434+shekhirin@users.noreply.github.com>

Prompted by: alexey

### DIFF
--- a/crates/cli/commands/src/download/mod.rs
+++ b/crates/cli/commands/src/download/mod.rs
@@ -279,6 +279,29 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> DownloadCo
         let cancel_token = CancellationToken::new();
         let _cancel_guard = cancel_token.drop_guard();
 
+        // Wire Ctrl+C / SIGTERM to the cancellation token so blocking
+        // download threads actually stop.
+        let signal_token = cancel_token.clone();
+        tokio::spawn(async move {
+            #[cfg(unix)]
+            {
+                let ctrl_c = tokio::signal::ctrl_c();
+                let mut sigterm =
+                    tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
+                        .expect("failed to register SIGTERM handler");
+                tokio::select! {
+                    _ = ctrl_c => {}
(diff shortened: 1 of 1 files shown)


crates/cli/commands/src/download/mod.rs
```diff
@@ -279,6 +279,29 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> DownloadCo
...
+                    _ = ctrl_c => {}
+                    _ = sigterm.recv() => {}
+                }
+            }
+            #[cfg(not(unix))]
+            {
+                let _ = tokio::signal::ctrl_c().await;
+            }
+            info!(target: "reth::cli", "Cancelling download...");
+            signal_token.cancel();
```
Rjected, line 302: why do we need a non unix cfg?

Rjected (commented):
just have one nit about unix cfg

crates/cli/commands/src/download/mod.rs
```diff
@@ -279,6 +279,29 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + EthereumHardforks>> DownloadCo
...
+                        .expect("failed to register SIGTERM handler");
+                tokio::select! {
+                    _ = ctrl_c => {}
+                    _ = sigterm.recv() => {}
+                }
+            }
+            #[cfg(not(unix))]
+            {
+                let _ = tokio::signal::ctrl_c().await;
+            }
```
mattsse, line 300: why do we need this, why isnt the existing token not enough

PR #23106, opened 2026-03-19
perf(rpc): avoid header clone in `logs_for_filter`

stevencartavia:
When the block is cached, only `number()` and `timestamp()` are needed from the header. Extracting those directly avoids cloning the full header on every `eth_getLogs` call by block hash.


mattsse (approved):
lgtm

PR #23117, opened 2026-03-19
perf(reth-bench): fetch RPC blocks in parallel

shekhirin:
Fetches up to `rpc_block_buffer_size` RPC blocks in parallel to keep the buffer full and ready for consumption.

### DIFF
--- a/bin/reth-bench/src/bench/new_payload_fcu.rs
+++ b/bin/reth-bench/src/bench/new_payload_fcu.rs
@@ -169,10 +169,10 @@ impl Command {
         let buffer_size = self.rpc_block_buffer_size;
 
         let mut blocks = Box::pin(
-            stream::iter(next_block..)
+            stream::iter((next_block..)
                 .take_while(|next_block| {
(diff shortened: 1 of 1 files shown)


yongkangc (approved):
cool stuff

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -167,70 +168,71 @@ impl Command {
...
-                    Ok(block) => block,
-                    Err(e) => {
-                        tracing::error!(target: "reth-bench", "Failed to fetch block {next_block}: {e}");
-                        let _ = error_sender.send(e);
-                        break;
+        let mut blocks = Box::pin(
+            stream::iter(next_block..)
+                .take_while(|next_block| {
+                    futures::future::ready(benchmark_mode.contains(*next_block))
+                })
```
mattsse: I think this take_while filter could also be applied directly to the original iter next_block..

but same thing really

mattsse (approved):

PR #23118, opened 2026-03-19
fix(provider): bound ConsistentProvider changeset ranges to requested…

esorense:
Bound ConsistentProvider account and storage changeset readers to the requested block range, and add regression coverage for DB-only  boundary requests.


mattsse (commented):
no

PR #23124, opened 2026-03-19
perf(trie): fused prune+compact with accurate memory_size

pepyakin:
This PR bundles two changes:

1. fold arena compaction into pruning making it a single pass.
2. since we are doing a single pass, we can compute the precise memory consumption while we are at it.

Estimation turned out to be pretty precise, within 1% for the bigger sets. Although it overestimates with smaller trees. Check the memory_size test.

### DIFF
--- a/crates/trie/sparse/src/arena/mod.rs
+++ b/crates/trie/sparse/src/arena/mod.rs
@@ -194,13 +194,13 @@ impl ArenaSparseSubtrie {
         );
     }
 
-    /// Prunes revealed subtrees that are not ancestors of any retained leaf.
+    /// Prunes revealed subtrees that are not ancestors of any retained leaf, compacting the
+    /// arena in a single BFS pass.
     ///
-    /// `retained_leaves` must be sorted and scoped to this subtrie's key range. The method
-    /// walks all nodes depth-first with the cursor, removing non-retained nodes bottom-up.
-    /// Only the boundary nodes (direct children of retained branches) have their parent's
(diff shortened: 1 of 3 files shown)


crates/trie/sparse/tests/memory_size.rs
pepyakin, line 1: I decided not putting it under the `suite` dir.

crates/trie/sparse/src/arena/mod.rs
```diff
@@ -60,6 +60,18 @@ fn prefix_range(
     begin..end
 }
 
+/// Returns the per-slot byte size used by `SlotMap<_, T>`. `SlotMap` wraps each value in a
+/// `Slot<T>` containing the value union + a 4-byte version field, with struct alignment.
+const fn slotmap_slot_size<T>() -> usize {
```
pepyakin, line 65: The cost of such estimation precision is that we are committing this crime. I think it's unlikely that it's going to be changed though. Also the test `test_memory_size_vs_allocator` should catch the divergence if happens.

pepyakin:
derek bench feature-args="--engine.enable-arena-sparse-trie" baseline-args="--engine.enable-arena-sparse-trie"

pepyakin:
derek bench feature-args="--engine.enable-arena-sparse-trie" baseline-args="--engine.enable-arena-sparse-trie"

pepyakin:
derek bench feature-args="--engine.enable-arena-sparse-trie" baseline-args="--engine.enable-arena-sparse-trie"

mediocregopher (approved):

PR #23125, opened 2026-03-19
perf(txpool): add capacity hint in BlobTransactions::satisfy_attributes

andrewshab3:
Pre-allocate `Vec` capacity in `satisfy_attributes` to avoid reallocations


mattsse (commented):
this would mostlikely over allocate so we can keep this as is, its unlikely that this branch is even hot

PR #23128, opened 2026-03-19
perf(rpc): remove redundant block id resolution in `debug_trace_block`

stevencartavia:
`recovered_block` already resolves `BlockId` to a block hash internally, so the manual `block_hash_for_id` lookup before calling it can be removed.


mattsse (approved):
Pending behavior change looks acceptable to me.

PR #23129, opened 2026-03-20
perf(rpc): avoid storage access clone

stevencartavia:
Use `into_accessed_slots()` instead of `accessed_slots().clone()` in `trace_block_storage_access`. Moves the data out of the inspector rather than cloning it.

`take_inspector()` internally clones the fused inspector, but this was already happening unconditionally after the callback in `TracerIter::next()` via `fuse_inspector()`. The new code just moves that clone into the callback by calling `take_inspector()`, which sets `was_fused = true` and causes `TracerIter` to skip the redundant `fuse_inspector()` call.

### DIFF
--- a/crates/rpc/rpc/src/trace.rs
+++ b/crates/rpc/rpc/src/trace.rs
@@ -603,12 +603,14 @@ where
                 block_id,
                 Some(block.clone()),
                 StorageInspector::default,
-                move |tx_info, ctx| {
+                move |tx_info, mut ctx| {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
Correct optimization: moves the storage access HashMap out via take_inspector() + into_accessed_slots() instead of cloning it. The fused inspector clone count stays the same (1), and the fused StorageInspector::default() is cheap to clone (empty maps). Ordering is correct — unique_loads/warm_loads read before take_inspector consumes the inspector.

PR #23134, opened 2026-03-20
fix: change `DEFAULT_IGNORE_GAS_PRICE`

klkvr:
Right now transactions with priority fee less than 2 wei are getting ignored when suggesting a priority fee

It doesn't seem reasonable to ignore 0 wei values but keep 2 wei values and the fact that transactions not paying any priority fee are still landing should definietly be accounted for in calculation


shekhirin (approved):
Affects the gas oracle here https://github.com/paradigmxyz/reth/blob/6259cb86f8d8e05e694b415b18f495a47cf00703/crates/rpc/rpc-eth-types/src/gas_oracle.rs#L253-L258

mattsse (approved):
makes sense

PR #23139, opened 2026-03-20
refactor(rpc): simplify `block_transaction_count`

stevencartavia:
Delegates `block_transaction_count` to `recovered_block`


mattsse (approved):
Correct simplification — recovered_block already handles the full pending/non-pending logic, so the duplication was unnecessary. Also fixes the inverted preference order for pending blocks (local_pending_block before provider) that was inconsistent with all other callers of recovered_block.

PR #23140, opened 2026-03-20
feat: add `reth-bb` binary with multi-segment big block execution support

mediocregopher:
Adds `reth-bb`, a modified reth node that executes "big block" payloads — merged multi-block payloads used for high-gas benchmarking. The `reth_newPayload` endpoint accepts optional `BigBlockData` containing per-segment EVM environments and prior block hashes, enabling correct execution across segment boundaries (system calls, withdrawals, BLOCKHASH resolution, gas accounting).

Also rewrites `generate-big-block` in reth-bench to directly merge consecutive RPC blocks (instead of going through `testing_buildBlockV1`), producing self-contained payload files with environment switches, correct receipts roots, and chained basefee/blob gas derivation. The `replay-payloads` command loads the new format and passes `BigBlockData` via `reth_newPayload`.

Supporting changes:
- `BeaconEngineMessage::ForkchoiceUpdated` now carries `EngineApiMessageVersion` so the version propagates through the engine pipeline
- `EthBeaconConsensus` gains `skip_blob_gas_used_check` and `skip_requests_hash_check` flags (used only by `reth-bb`)
- Removes the `--testing.skip-gas-limit-ramp-check` CLI flag from the main reth binary (no longer needed)
- CI bench workflow updated to build/run `reth-bb` when `big-blocks` mode is selected


decofe:
cc @mediocregopher

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/23536619232/job/68513357758)

⏳ **Status:** Building binaries (snapshot update pending)...

**Config:** blocks: `big`, baseline: `mediocregopher/2k-big-blocks`, feature: `mediocregopher/2k-big-blocks`, no-slack, abba: `disabled`

crates/engine/primitives/src/message.rs
```diff
@@ -162,6 +162,30 @@ pub struct NewPayloadTimings {
...
+    /// Environment switches at block boundaries.
+    /// Each entry is `(cumulative_tx_count, execution_data_of_next_block)`.
+    ///
+    /// The first entry at index 0 represents the **original unmutated** base block's
+    /// `ExecutionData`, which must be used to derive the initial EVM environment.
+    pub env_switches: Vec<(usize, ExecutionData)>,
+    /// Block number → real block hash for blocks covered by previous big blocks in a sequence.
+    /// When replaying chained big blocks, the BLOCKHASH opcode needs real hashes for blocks
+    /// that were merged into earlier big blocks (and thus not individually persisted).
+    pub prior_block_hashes: Vec<(u64, alloy_primitives::B256)>,
```
pepyakin, line 180: Nit: it would be great to distinguish between different types of blocks in these docs. You also used segments somewhere, maybe that's a better term here.

bin/reth-bb/src/evm.rs
```diff
@@ -0,0 +1,570 @@
...
+        ctx.block = new_block_env;
+        ctx.cfg = new_cfg_env;
+
+        // Build a new inner executor for the next segment. gas_used starts
+        // at 0 so the per-transaction gas check uses this segment's real
+        // gas_limit correctly.
+        let mut new_inner = EthBlockExecutor::new(evm, new_ctx, spec, receipt_builder);
+
+        // Carry forward receipts from prior segments.
+        new_inner.receipts = result.receipts;
```
klkvr, line 300: could we override `new_inner.gas_used` here to avoid dealing with offsets for cumulative gas?

bin/reth-bb/src/evm_config.rs
```diff
@@ -0,0 +1,291 @@
...
+        // can reseed State::block_hashes at segment boundaries. The seeder
+        // is a function pointer that knows the concrete State<DB> type,
+        // allowing the generic BbBlockExecutor to reseed without additional
+        // trait bounds on DB.
+        self.executor_factory.create_executor_with_seeder(
+            evm,
+            ctx,
+            Some(seed_state_block_hashes::<DB>),
+        )
+    }
```
klkvr, line 182: ah it's a bit unfortunate that we rely on `State<DB>` type here

i guess one way to get rid of it would be to wrap the db into some kind of wrapper that would overwrite the blockhashes but probably ok to just do this for now

bin/reth-bb/src/evm.rs
mediocregopher: I believe we can't because then we'd have to raise the gas-limit on the inner EVM, and that throws off the GASLIMIT opcode. We could disable gas limit checking on the EVM I think, but this worked with less skipped checks

bin/reth-bb/src/evm_config.rs
mediocregopher: I think I had tried that in a previous iteration but there was a reason it couldn't work, idr now

klkvr (approved):

pepyakin (approved):

PR #23143, opened 2026-03-20
perf(rpc): avoid redundant block fetch in `sim_bundle_inner`

stevencartavia:
`sim_bundle_inner` was fetching the block header twice — once inside `evm_env_at` and again via `recovered_block`. Derive the evm env directly from the already-fetched block via `evm_env_for_header` instead.


mattsse (changes requested):
all of this is a bit problematic for the pending tag, which could change hashes midflight,
so we should keep this as is and operate on ids instead because most of these arent atomic, and spawn_with_state_at_block doesnt resolve a local pending hash

so I think we can keep this as is, or find a better way to deal with the pending tag issue

PR #23144, opened 2026-03-20
perf(rpc): avoid cloning block env in pending block builder

stevencartavia:
Reads gas_limit, basefee, and blob_gasprice directly from the block env instead of cloning it to read three fields.


mattsse (approved):
Looks good to me.

PR #23147, opened 2026-03-21
perf(engine): clone block body instead of full block for tx root task

stevencartavia:
Avoids cloning the block header when spawning the transaction root computation task, since only the body is needed.


mattsse (approved):
Looks good to me.

PR #23151, opened 2026-03-22
fix(download): avoid double-counting resumable progress on retry

### LINKED ISSUE #23146: fix(download): resumable download double-counts progress on retry
## Describe the bug

When using `reth download --resumable` with modular snapshots, the shared progress counter double-counts already-downloaded bytes on retry within the same `resumable_download` call.

## Steps to reproduce

1. Run `reth download --chain mainnet --minimal -y --resumable`
2. Wait for a large single-file archive (e.g., `state.tar.zst` ~157 GB) to partially download
3. Let the connection drop naturally (or interrupt it) — reth's built-in retry handles it

## Observed behavior

Progress jumps by the size of the existing `.part` file on each retry:

```
13:38:47  progress=12.0%  downloaded=18.83 GB  total=156.39 GB
13:38:47  Download interrupted, retrying in 5s...  file=state.tar.zst
13:38:52  Retry attempt 2/10 - resuming from 9896656896 bytes  file=state.tar.zst
13:38:53  progress=12.0%  downloaded=18.83 GB
13:38:56  progress=17.9%  downloaded=28.05 GB  ← jumped by ~9.2 GB (the .part file size)
```

The ~9.2 GB already in the `.part` file gets added to `SharedProgress` again on retry.

## Root cause

In [`crates/cli/commands/src/download/mod.rs`](https://github.com/paradigmxyz/reth/blob/main/crates/cli/commands/src/download/mod.rs), inside `resumable_download`, the retry loop calls `sp.add(start_offset)` on every attempt:

```rust
for attempt in 1..=MAX_DOWNLOAD_RETRIES {
    let existing_size = fs::metadata(&part_path).map(|m| m.len()).unwrap_or(0);
    // ...
    let start_offset = if is_partial { existing_size } else { 0 };

    if let Some(sp) = shared {
        if start_offset > 0 {
            sp.add(start_offset);  // ← called on EVERY retry, not just once
        }
    }
}
```

On attempt 1, it adds the resumed bytes (correct). On attempt 2+, it adds them again (double-count).

## Environment

- Reth version: nightly (built from main, ~March 21 2026)
- Running as init container in Kubernetes (GKE)
- Downloading mainnet minimal snapshot (~156 GB `state.tar.zst`)

## Impact

Progress display only — does not affect actual download integrity or correctness. But inflated progress makes ETA unreliable for long-running downloads.

ozankenangungor:
Fixes #23146

Avoid double-counting resumed bytes in shared progress on retry.


mattsse (commented):
The fix is correct — pre-existing `.part` bytes must not be re-added on retry.

However, the `ResumeSeed` struct can be replaced with an `Option<u64>` + `.take()`:

```rust
// Before the loop, next to total_size / last_error:
let mut resume_pending = fs::metadata(&part_path).map(|m| m.len()).ok();

// Inside the loop, replacing the old sp.add block:
if let Some(bytes) = resume_pending.take() {
    if is_partial {
        sp.add(bytes);
    }
}
```

`.take()` returns the value and leaves `None` — subsequent retries see `None` and skip. The `!is_partial` discard is implicit: we take the value but don't add it.

With that simplification the three new unit tests aren't really needed either — the logic is straightforward enough.

PR #23156, opened 2026-03-22
refactor: remove bincode usage from `HeaderStage`

klkvr:
bincode is unmaintained and can be fully replaced with rlp for primitives, this is a first step towards it


mattsse (approved):
makes sense, the only way how this could be problematic is if the client is upgraded mid headerstage and I think we can ignore this

PR #23158, opened 2026-03-22
refactor: remove SerdeBincodeCompat trait, use RLP for block serialization

decofe:
Removes the `SerdeBincodeCompat` trait and replaces bincode-compatible serde wrappers with direct RLP encoding for block types.

**Motivation**: `SerdeBincodeCompat` was a workaround for bincode's inability to handle optional serde fields. It added `MaybeSerdeBincodeCompat` bounds to `BlockHeader`, `BlockBody`, `FullSignedTx`, `FullReceipt`, and `NodePrimitives::Block` — complexity that's no longer needed since the two production usage sites (header stage ETL and ExEx WAL) can use RLP instead.

**Changes**:
- `Chain` serde: blocks encoded as `(hash, rlp_bytes, senders)` instead of nested `SerdeBincodeCompat` wrappers
- `ExecutionOutcome` serde: receipts encoded as RLP bytes
- Header stage ETL: `hash ++ RLP(header)` instead of `bincode::serialize(SealedHeader)`
- Removed `SerdeBincodeCompat` trait, `MaybeSerdeBincodeCompat`, `RlpBincode`, and all block/header/tx serde_bincode_compat modules (~850 lines)
- Simplified `serde-bincode-compat` Cargo features across crates

**WAL compatibility**: this changes the ExEx WAL serialization format. Old WAL files won't deserialize. The WAL is only non-empty across restarts if the node crashed — clean shutdown drains it.

Co-Authored-By: Arsenii Kulikov <62447812+klkvr@users.noreply.github.com>

Prompted by: klkvr


mattsse (approved):
rip bozo

shekhirin (approved):

PR #23161, opened 2026-03-22
fix(rpc): apply evm_memory_limit to all execution RPC methods

AKABABA-ETH:
Was poking around `cfg_env` setup across the RPC handlers and noticed `memory_limit` only gets set in one place:

`call.rs:899: evm_env.cfg_env.memory_limit = self.evm_memory_limit();  // prepare_call_env — eth_call only`
`eth_estimateGas`, `eth_createAccessList`, and `eth_simulateV1` all skip it, so they run with revm's default of `u64::MAX` — effectively no cap. Someone could craft a call through any of those three endpoints and blow past the configured limit without touching `eth_call`.

Added `evm_env.cfg_env.memory_limit = self.evm_memory_limit()` to `estimate_gas_with`, `create_access_list_with`, and `simulate_v1` to match what `prepare_call_env` already does.


mattsse (commented):
I believe we can keep this as is because this setting is only really relevant for the general purpose eth-call context

PR #23164, opened 2026-03-23
fix(engine): handle stream termination in MiningMode::Trigger

dennsikl:
`MiningMode::Trigger` uses `is_ready()` to check if the stream yielded a value, but this also returns true when the stream terminates (`Ready(None)`). 

Match the pattern used in `MiningMode::Instant` which correctly checks `Poll::Ready(Some(_))` to only trigger on actual stream items.


mattsse (commented):
we can keep this as is

PR #23171, opened 2026-03-23
fix(trie): avoid boundary parent unwrap panic in parallel sparse reveal

pepyakin:


crates/trie/sparse/tests/suite/reveal_nodes.rs
```diff
@@ -385,3 +387,41 @@ pub(super) fn test_insert_then_reveal_does_not_overwrite_branch<T: SparseTrie>(
...
+        state_mask,
+        Some(branch_rlp),
+    ));
+
+    let mut trie = (new_trie)();
+    trie.set_root(root, None, false).expect("set_root should succeed");
+
+    // Before the fix this panicked at `hashes_from_upper` when trying to unwrap
+    // `upper_subtrie.nodes.get_mut([0x1])`.
+    let boundary_path = Nibbles::from_nibbles([0x1, 0x2]);
```
mediocregopher, line 423: Is this right? `0x12` is the branch I believe, which has a child at nibble `0`, so this should be `0x120` I would think.

crates/trie/sparse/tests/suite/reveal_nodes.rs
```diff
@@ -385,3 +387,41 @@ pub(super) fn test_insert_then_reveal_does_not_overwrite_branch<T: SparseTrie>(
...
+///
+/// When root is an extension that crosses the upper/lower boundary, a boundary node path can be
+/// reachable even if there is no explicit upper branch at `path[..UPPER_TRIE_MAX_DEPTH - 1]`.
+/// Revealing such a node should not panic.
+pub(super) fn test_reveal_boundary_node_with_missing_upper_parent_branch<T: SparseTrie>(
+    new_trie: fn() -> T,
+) {
+    // Root reveals as extension [0x1, 0x2] with a branch below it.
+    let child_hash = RlpNode::word_rlp(&B256::repeat_byte(0xAA));
+    let state_mask = TrieMask::new(0b0001);
```
mediocregopher: nit: for consistency it'd be good to give the branch two children, since a single-child branch isn't actually possible

crates/trie/sparse/tests/suite/reveal_nodes.rs
pepyakin: comment updated.

mediocregopher (approved):

PR #23180, opened 2026-03-23
perf(payload): avoid tx clone in block building loop

stevencartavia:
Pre-compute tx fields before moving tx into execute_transaction, avoiding a `Recovered<TransactionSigned>` clone per transaction.


mattsse (approved):
lgtm

PR #23184, opened 2026-03-23
chore: unify InMemorySize

klkvr:
removed reth's `InMemorySize` trait in favor of alloy's one

also changed `SignedTransaction` to be a blanket trait that shouldn't need to be upstreamed to alloy


mattsse (approved):
nice

PR #23187, opened 2026-03-24
perf(provider): avoid full receipt clone in tx-range query

BearSoltokien:
Clone only the requested receipt subrange from in-memory block state, avoiding a full receipt-vector clone before slicing.


crates/chain-state/src/in_memory.rs
```diff
@@ -673,6 +673,14 @@ impl<N: NodePrimitives> BlockState<N> {
         self.receipts()
     }
 
+    /// Returns a vector of `Receipt` cloned only for the requested tx index range.
+    pub fn executed_block_receipts_by_index_range(
+        &self,
+        index_range: std::ops::RangeInclusive<usize>,
+    ) -> Vec<N::Receipt> {
+        self.executed_block_receipts_ref()[index_range].to_vec()
+    }
```
mattsse: lets avoid adding this and update the callsite with `executed_block_receipts_ref()...().cloned().collect` instead

mattsse (changes requested):

mattsse (approved):
Looks good to me.

PR #23188, opened 2026-03-24
chore: fixed changes for reth-bench

Rimeeeeee:


bin/reth-bench/Cargo.toml
```diff
@@ -76,7 +76,7 @@ csv.workspace = true
 [dev-dependencies]
 
 [features]
-default = ["jemalloc"]
```
Soubhik-10: ToDo: re add this feature

emmajam (approved):

PR #23189, opened 2026-03-24
refactor(net): simplify seen-transaction filtering in propagation

mark0-cn:
Replaces manual loops with iterator filtering in both full-transaction and hash propagation paths.
 Keeps behavior unchanged while reducing duplication and improving readability.


mattsse (commented):
we can keep this as is

PR #23190, opened 2026-03-24
fix(init): track actual byte size instead of account count in dump_state

mattsse:
Closes #23155

The `dump_state` chunking used a fixed account-count heuristic (`AVERAGE_COUNT_ACCOUNTS_PER_GB_STATE_DUMP`) to split writes into ~1GB batches. This doesn't account for accounts with large storage maps, which caused batches to far exceed the intended size on state dumps like Gnosis (25GB, with storage-heavy accounts), leading to OOM kills on low-memory devices.

Now tracks actual `(address, account)` byte sizes from the ETL collector entries and flushes when `chunk_byte_size >= DEFAULT_SOFT_LIMIT_BYTE_LEN_ACCOUNTS_CHUNK` (1GB). This constant already existed but was unused.


yongkangc (approved):
ahh this makes sense so our flushing is more accurate

PR #23196, opened 2026-03-24
feat(txpool): add TransactionValidationTaskExecutor::spawn

mattsse:
Extracts the task spawning logic from `EthTransactionValidatorBuilder::build_with_tasks` into a standalone `TransactionValidationTaskExecutor::spawn` method. This allows any validator to be wrapped with task-based validation without going through the Eth builder, making it easier to set up custom validators with the full spawning infrastructure.


yongkangc (approved):
makes sense, much cleaner

PR #23202, opened 2026-03-24
refactor: remove `PayloadBuilderAttributes`

klkvr:
Unifies `PayloadBuilderAttributes` and `PayloadAttributes` so that payload builder and engine API operate on the same type.

This is done by extending `PayloadAttributes` trait with `payload_id` method that is responsible for computing payload ID given parent hash.
https://github.com/paradigmxyz/reth/blob/958b738a42d0734c14fdbdfefe4e60c2bb4bd7ed/crates/payload/primitives/src/traits.rs#L105-L106

This allows to remove `PayloadBuilderAttributes` type which purpose was essentially to combine RPC attributes with a parent hash and payload ID.

### DIFF
--- a/crates/payload/primitives/src/traits.rs
+++ b/crates/payload/primitives/src/traits.rs
@@ -2,11 +2,9 @@
 
 use crate::PayloadBuilderError;
 use alloc::{boxed::Box, sync::Arc, vec::Vec};
-use alloy_eips::{
-    eip4895::{Withdrawal, Withdrawals},
-    eip7685::Requests,
-};
-use alloy_primitives::{Address, B256, U256};
+use alloy_eips::{eip4895::Withdrawal, eip7685::Requests};
+use alloy_primitives::{B256, U256};
+use alloy_rlp::Encodable;
 use alloy_rpc_types_engine::{PayloadAttributes as EthPayloadAttributes, PayloadId};
 use core::fmt;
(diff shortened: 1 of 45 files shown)


crates/payload/primitives/src/traits.rs
```diff
@@ -97,59 +95,16 @@ pub trait BuiltPayload: Send + Sync + fmt::Debug {
...
-
 /// Basic attributes required to initiate payload construction.
 ///
 /// Defines minimal parameters needed to build a new execution payload.
 /// Implementations must be serializable for transmission.
 pub trait PayloadAttributes:
     serde::de::DeserializeOwned + serde::Serialize + fmt::Debug + Clone + Send + Sync + 'static
 {
+    /// Computes the unique identifier for this payload build job.
+    fn payload_id(&self, parent_hash: &B256) -> PayloadId;
```
mattsse, line 106: this removes the version requirement,

however the way this should be mitigated is by rolling a `VersionedAttributes<>` type

mattsse (approved):

crates/payload/primitives/src/traits.rs
klkvr: yep for any nodes that need `version` context from RPC it should be made part of payload attributes and set on FCU request

PR #23207, opened 2026-03-24
fix(engine/tree): Fix canonical hash check in `find_canonical_header` for provider-known blocks

### LINKED ISSUE #23205: engine/tree: `find_canonical_header` can misclassify a known non-canonical header as canonical, causing FCU to return `VALID` without re-canonicalizing the requested head
### Describe the bug

I hit this while debugging a rewind flow in an external integration, but the root cause reduces to `engine/tree`.

After the engine has moved its in-memory canonical head to a new sibling block, a later `ForkchoiceUpdated` back to the original block can be short-circuited as "already canonical" if the original block is merely _known_ by hash in the provider but not actually canonical, leaving the canonical chain un-updated.

The problematic path is from `apply_chain_update(...)` calling into `find_canonical_header(...)`. In `find_canonical_header`, we currently short-circuit to checking if the requested hash is known to the block provider via `provider.header(hash)`, which proves only that the header is known, not that it is canonical.

That can make FCU return `VALID` while leaving the actual canonical head unchanged.

I've implemented a fix for this issue in PR #23207 and in a [local branch](https://github.com/wwared/reth/tree/bug/fcu-sibling-recanonicalization-repro). The PR includes a test that reproduces the issue by performing two FCUs between sibling blocks at the same height.

To give some more context for how we ran into this bug: in `op-supernode`'s  `RewindEngine` ([this function](https://github.com/ethereum-optimism/optimism/blob/e08548e4d6021412b30e7f092738e7a4a4b90115/op-supernode/supernode/chain_container/engine_controller/rewind.go#L26-L90)), we try to rewind the underlying EL to some block B by temporarily inserting a synthetic block S at the same height as B, doing a FCU to S, and then doing another FCU back to B. When running this function under `op-reth`, we observe that the first FCU to the synthetic block updates the canonical head correctly to point to S, but the second FCU does not update the head as expected back to B. With the changes from PR #23207 the issue no longer occurs in `op-supernode`.

### Steps to reproduce

I've added a test that reproduces this case in this commit: https://github.com/wwared/reth/commit/87929d696f9745331c502d1d8515039e410f74b3

It can also be reproduced by running some tests in the optimism monorepo via `op-reth`, for example this one https://github.com/ethereum-optimism/optimism/blob/73b3f0894e42b100e9701a99fad5a72dd24a5005/op-acceptance-tests/tests/supernode/interop/reorg/invalid_message_reorg_test.go#L27-L30 after commenting out the `sysgo.SkipOnOpReth` line and running it with the env var `DEVSTACK_L2EL_KIND=op-reth` set.

### Node logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

I checked this behavior on `v1.11.3` and on current `main` commit 15338b8113d006e24a42a0d556cfeb8bd70ebf0d

N/A, I believe this applies regardless of DB version

dev

Archive (default)

_No response_

_No response_

### Code of Conduct

wwared:
Fixes #23205

If the hash is already known to the provider engine, then `find_canonical_header` would return the block even if it's not actually part of the canonical chain, leading to `apply_chain_update` short-circuit by believing the hash is already canonical and thus not applying a FCU correctly in certain specific edge cases.

This PR changes the fallback path inside the `if canonical.is_none()` branch to do a full round-trip via the provider's canonical chain mapping: first we fetch the block number for the given hash, then we fetch the block hash at that block number, and finally we compare the resulting hash with the expected hash. We only consider the hash to be canonical if it matches after this round-trip.

The `test_fcu_back_from_synthetic_sibling_to_original_head_recanonicalizes_head` test added works by having a `BlockchainProviderTestHarness` that can persist blocks. It sets up the following scenario:

1. Create a new synthetic block at height 9 `s9` to replace a the real block `a9`
2. Perform a FCU so the new head of the chain is `s9`
3. Check that the canonical head changed to `s9` correctly
4. Check that `find_canonical_header(a9.hash())` returns `None` since `a9` is no longer part of the canonical chain (this fails on current `main`)
5. Perform a FCU back to `a9`
6. Check that the canonical head has been updated again

This might not be the correct way to fix this, but issue #23205 talks a bit about where I ran into this bug, and I'd be happy to update the approach or the test as needed

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2966,7 +2966,11 @@ where
 
         if canonical.is_none() {
             if let Some(number) = self.provider.block_number(hash)? {
-                if self.provider.block_hash(number)?.is_some_and(|canonical_hash| canonical_hash == hash) {
+                if self
+                    .provider
+                    .block_hash(number)?
+                    .is_some_and(|canonical_hash| canonical_hash == hash)
+                {
                     canonical =
                         self.provider.header(hash)?.map(|header| SealedHeader::new(header, hash));
                 }
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2965,7 +2965,12 @@ where
         let mut canonical = self.canonical_in_memory_state.header_by_hash(hash);
 
         if canonical.is_none() {
-            canonical = self.provider.header(hash)?.map(|header| SealedHeader::new(header, hash));
+            if let Some(number) = self.provider.block_number(hash)? {
+                if self.provider.block_hash(number)?.is_some_and(|canonical_hash| canonical_hash == hash) {
+                    canonical =
+                        self.provider.header(hash)?.map(|header| SealedHeader::new(header, hash));
+                }
+            }
         }
 
         Ok(canonical)
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -2964,17 +2964,13 @@ where
     ) -> Result<Option<SealedHeader<N::BlockHeader>>, ProviderError> {
         let mut canonical = self.canonical_in_memory_state.header_by_hash(hash);
 
-        if canonical.is_none() {
-            if let Some(number) = self.provider.block_number(hash)? {
-                if self
-                    .provider
-                    .block_hash(number)?
-                    .is_some_and(|canonical_hash| canonical_hash == hash)
-                {
-                    canonical =
-                        self.provider.header(hash)?.map(|header| SealedHeader::new(header, hash));
-                }
-            }
+        if canonical.is_none() &&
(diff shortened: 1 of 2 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -2964,7 +2964,12 @@ where
     ) -> Result<Option<SealedHeader<N::BlockHeader>>, ProviderError> {
         let mut canonical = self.canonical_in_memory_state.header_by_hash(hash);
 
-        if canonical.is_none() {
+        if canonical.is_none() &&
+            let Some(number) = self.provider.block_number(hash)? &&
+            self.provider
+                .block_hash(number)?
+                .is_some_and(|canonical_hash| canonical_hash == hash)
```
mattsse, line 2971: I dont really understand why this is necessary.

this should be the same as fetching the header by its hash, because these fn operate on the same data?
```
self.provider.header(hash)
```

mattsse (commented):
not clear what this fix is supposed to do

PR #23214, opened 2026-03-25
feat: enable jemalloc `override_allocator_on_supported_platforms`

decofe:
Enables `override_allocator_on_supported_platforms` on tikv-jemallocator so jemalloc overrides the system allocator for all linked code — including C/C++ dependencies like MDBX — not just Rust allocations via `#[global_allocator]`.

Changes:
- `jemalloc` feature now enables `override_allocator_on_supported_platforms` (superset of `unprefixed_malloc_on_supported_platforms`)
- Adds `tikv-jemalloc-sys` dep, re-exported from `reth-cli-util` and `use`d as `_` in each binary's `main.rs` (required for linker visibility)
- Removes the now-redundant `jemalloc-unprefixed` feature
- Fixes `jemalloc-prof` to properly depend on `jemalloc` in both binaries
- Updates `malloc_conf` export name and docs to match unprefixed symbols

Prompted by: DaniPopes

### DIFF
--- a/bin/reth/src/main.rs
+++ b/bin/reth/src/main.rs
@@ -3,6 +3,7 @@
 #[global_allocator]
 static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
 
+// Required for "override_allocator_on_supported_platforms".
 #[cfg(all(feature = "jemalloc", unix))]
 use reth_cli_util::allocator::tikv_jemalloc_sys as _;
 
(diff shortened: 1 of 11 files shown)


bin/reth/src/main.rs
```diff
@@ -3,8 +3,11 @@
 #[global_allocator]
 static ALLOC: reth_cli_util::allocator::Allocator = reth_cli_util::allocator::new_allocator();
 
+#[cfg(all(feature = "jemalloc", unix))]
```
DaniPopes, line 7:
```suggestion
// Required for "override_allocator_on_supported_platforms".
#[cfg(all(feature = "jemalloc", unix))]
```

mattsse (approved):

shekhirin (approved):

PR #23223, opened 2026-03-25
fix: deduplicate packed nibble compact codec

anim001k:
Deduplicates the compact encode/decode logic shared by `PackedStoredNibbles` and `PackedStoredNibblesSubKey` in `reth-trie-common`.

This keeps the 33-byte packed nibble format defined in one place and reduces the risk of the two implementations drifting apart.


mattsse (commented):
not worth doing

PR #23228, opened 2026-03-25
chore: bump alloy 1.8.1

mattsse:


bin/reth-bench/src/authenticated_transport.rs
```diff
@@ -52,8 +52,7 @@ impl InnerTransport {
         url: Url,
         jwt: JwtSecret,
     ) -> Result<(Self, Claims), AuthenticatedTransportError> {
-        let mut client_builder =
-            reqwest::Client::builder().tls_built_in_root_certs(url.scheme() == "https");
```
mattsse, line 56: this is no more and i dont think we even needed this here?

DaniPopes (approved):

PR #23242, opened 2026-03-26
feat: share execution cache with payload builder

klkvr:
Refactors payload building logic to use `SavedCache` from the engine instead of `CachedReads`


crates/payload/basic/src/lib.rs
```diff
@@ -808,7 +741,7 @@ impl<Payload> BuildOutcomeKind<Payload> {
 #[derive(Debug)]
 pub struct BuildArguments<Attributes, Payload: BuiltPayload> {
     /// Previously cached disk reads
```
shekhirin, line 822:
```suggestion
    /// Optional execution cache to use for the payload.
```

can we also document when it's `None`?

yongkangc (approved):
This looks much cleaner than what I would have expected

shekhirin (approved):

joshieDo (approved):

mattsse (approved):

PR #23243, opened 2026-03-26
fix(trie): before prune call root

pepyakin:
Remove tolerant prune behavior from 50ce26f since we discovered more cases. It's better just to embrace the tight requirement of calling prune after `root()`.


crates/trie/sparse/src/arena/mod.rs
pepyakin, line 1: This is basically revert of the loosening fix for arena.

mediocregopher (approved):

PR #23244, opened 2026-03-26
feat(engine): backpressure

pepyakin:


crates/engine/primitives/src/forkchoice.rs
```diff
@@ -402,10 +402,10 @@ mod tests {
     #[test]
     fn test_forkchoice_state_hash_find() {
         // Define example hashes
-        let head_hash = B256::random();
-        let safe_hash = B256::random();
-        let finalized_hash = B256::random();
-        let non_matching_hash = B256::random();
+        let head_hash = B256::with_last_byte(1);
```
mediocregopher, line 405: What was this about?

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1363,6 +1568,11 @@ where
         }
     }
 
+    #[cfg(test)]
```
mediocregopher, line 1571: nit: let's just remove the cfg(test) and not have two methods

mediocregopher (approved):

PR #23246, opened 2026-03-26
feat(engine): share sparse trie pipeline with payload builder

joshieDo:
#23242 follow-up

Shares the engine's incremental sparse trie pipeline with the payload builder so FCU-triggered block builds skip the blocking `state_root_with_updates()` call.

## Moved & unified types

`StateRootHandle` (new, in `reth-trie-parallel::state_root_task`):
- Replaces engine-local `StateRootHandle` (was in `payload_processor/mod.rs`)
- Single type used by both engine (`newPayload`) and payload builder (`FCU`)

Also moved to `state_root_task`:
- `MultiProofMessage`, `Source`, `StateRootComputeOutcome`, `StateHookSender`, `evm_state_to_hashed_post_state()`

## `BlockBuilder::finish()` 
- `finish(provider, None)` — computes state root internally (existing behavior)
- `finish(provider, Some((root, trie_updates)))` — uses pre-computed root

## Payload builder wiring

In `default_ethereum_payload`:
- Sets state hook via `handle.state_hook()` before tx loop
- After execution, drops hook → `FinishedStateUpdates`, recvs root, calls `finish` with pre-computed result
- Falls back to sync path when handle is `None`

## Engine convergence

- `spawn_sparse_trie_handle` deleted — `spawn_state_root` now takes `halve_workers: bool`
- `EngineValidator::sparse_trie_handle_for` returns `Option<StateRootHandle>`
- Gated behind `--engine.share-sparse-trie-with-payload-builder` (default `false`)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -271,12 +267,18 @@ where
...
         let span = Span::current();
 
-        let state_root_handle = self.spawn_state_root(multiproof_provider_factory, &env, config);
+        let halve_workers = env.transaction_count <= Self::SMALL_BLOCK_PROOF_WORKER_TX_THRESHOLD;
+        let state_root_handle = self.spawn_state_root(
+            multiproof_provider_factory,
+            env.parent_state_root,
+            halve_workers,
+            config,
+        );
```
yongkangc, line 276: i see, we wanna half worker if its a small block.

crates/evm/evm/src/execute.rs
```diff
@@ -477,18 +482,21 @@ where
     fn finish(
         self,
         state: impl StateProvider,
+        state_root_precomputed: Option<(B256, TrieUpdates)>,
```
klkvr, line 485: makes me wonder if this should just take a closure like `|&db| -> (state_root, trie_updates)`

this should likely allow us to upstream this abstraction into alloy-evm entirely

crates/trie/parallel/src/state_root_task.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl From<StateChangeSource> for Source {
+    fn from(source: StateChangeSource) -> Self {
+        Self::Evm(source)
+    }
+}
+
+/// Messages used internally by the multi proof task.
+#[derive(Debug)]
+pub enum MultiProofMessage {
```
klkvr: let's rename this to `StateRootTaskMessage`/`StateRootMessage`?

klkvr (approved):

shekhirin (approved):

mattsse (changes requested):
this has a few drawbacks/side-effects users should be aware of which arent properly documented

the fact that this advances the trie makes this pretty much unsuitable for the ethereum implementation and we should probably not use it at all to avoid side-effects for the --always-prepare-payload

but also, this setting is opt in so likely not an issue unless misused

mattsse (approved):

PR #23248, opened 2026-03-26
perf: enable revm p256-aws-lc-rs feature by default

decofe:
Enables the revm `p256-aws-lc-rs` feature flag so secp256r1 (RIP-7212) signature verification uses `aws-lc-rs` instead of the pure-Rust `p256` crate. Faster for the P256 precompile.

Context: https://github.com/tempoxyz/tempo/pull/3339

Prompted by: DaniPopes


DaniPopes:
derek bench

rakita (approved):

PR #23256, opened 2026-03-27
feat: add a method to get payload resolve future

klkvr:
Adds a helper method to `PayloadBuilderHandle` to get the payload resolution future

### DIFF
--- a/crates/payload/builder/src/service.rs
+++ b/crates/payload/builder/src/service.rs
@@ -158,6 +158,17 @@ impl<T: PayloadTypes> PayloadBuilderHandle<T> {
         }
     }
 
+    /// Same as [`Self::resolve_kind`] but returns the underlying future.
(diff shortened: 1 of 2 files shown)


crates/payload/builder/src/service.rs
```diff
@@ -158,6 +158,17 @@ impl<T: PayloadTypes> PayloadBuilderHandle<T> {
         }
     }
 
+    /// Same as [`Self::resolve_kind`] but returns the underlying future.
+    pub async fn resolve_kind_fut(
+        &self,
+        id: PayloadId,
+        kind: PayloadKind,
+    ) -> Result<Option<PayloadFuture<T::BuiltPayload>>, PayloadBuilderError> {
```
mattsse, line 166: is this so we can await the payload in two steps?

mattsse (approved):

PR #23263, opened 2026-03-28
chore: exp fix remaining hive tests

Rimeeeeee:


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -1408,8 +1396,8 @@ where
             block,
             output,
             receipt_root_bloom,
-            None,
-            false,
+            built_bal,
+            true,
```
Rimeeeeee: do we have a false case

Soubhik-10: That's true lol

PR #23270, opened 2026-03-29
feat(db): add `create_test_provider_factory_with_chain_spec_and_db_args`

figtracer:
## Summary
- Adds `create_test_provider_factory_with_chain_spec_and_db_args` to `reth-provider` test utils
- Same as `create_test_provider_factory_with_chain_spec` but accepts custom `DatabaseArguments`
- No breaking changes — existing functions are untouched

## Motivation

Downstream consumers like [`paradigmxyz/stateless#11`](https://github.com/paradigmxyz/stateless/pull/11) need larger database geometries for worst-case benchmark witness generation.

### DIFF
--- a/crates/storage/provider/src/test_utils/mod.rs
+++ b/crates/storage/provider/src/test_utils/mod.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use alloy_primitives::B256;
 use reth_chainspec::{ChainSpec, MAINNET};
-use reth_db::{test_utils::TempDatabase, DatabaseEnv};
+use reth_db::{mdbx::DatabaseArguments, test_utils::TempDatabase, DatabaseEnv};
 use reth_errors::ProviderResult;
 use reth_ethereum_engine_primitives::EthEngineTypes;
 use reth_node_types::NodeTypesWithDBAdapter;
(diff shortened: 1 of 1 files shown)


crates/storage/provider/src/test_utils/mod.rs
```diff
@@ -42,19 +46,21 @@ pub type MockNodeTypesWithDB<DB = Arc<TempDatabase<DatabaseEnv>>> =
...
-    create_test_provider_factory_with_node_types::<MockNodeTypes>(chain_spec)
+    create_test_provider_factory_with_node_types::<MockNodeTypes>(chain_spec, db_args)
 }
 
-/// Creates test provider factory with provided chain spec.
+/// Creates test provider factory with provided chain spec and optional database arguments.
 pub fn create_test_provider_factory_with_node_types<N: NodeTypesForProvider>(
     chain_spec: Arc<N::ChainSpec>,
+    db_args: Option<DatabaseArguments>,
 ) -> ProviderFactory<NodeTypesWithDBAdapter<N, Arc<TempDatabase<DatabaseEnv>>>> {
```
mattsse, line 58: I think this makes sense but I'd prefer this as another helper fn or even a builder style input args to avoid breaking existing test code

mattsse (changes requested):

mattsse (approved):

PR #23275, opened 2026-03-30
fix(net): prefer peer-reported block number in session activation

MagicJoshh:
`on_session_activated` always does a local DB lookup for the peer's block number via `blockhash`, even when `UnifiedStatus.latest_block` already carries it from the wire. Worse, if the hash isn't in local chain, `unwrap_or_default()` silently drops it to 0 

uses `status.latest_block` directly when available (eth/69+), falls back to the hash lookup for older protocol versions. Related to eth/69 support work.


mattsse (approved):
this makes sense, ty!

PR #23277, opened 2026-03-30
test(txpool): add regression for parked basefee ancestor handling

0xWeakSheep:
Adds a focused regression test for parked basefee enforcement in the txpool.

The new test locks in current behavior when only the ancestor transaction satisfies the basefee threshold: the ancestor is removed while the descendant remains parked. This prevents ambiguity around ancestor/descendant handling and keeps the behavior explicit in tests.


mattsse (approved):
Clean extraction of the TODO block into a standalone test. Looks good.

PR #23280, opened 2026-03-30
feat(engine): backpressure, take 2.

pepyakin:
This PR adds backpressure handling to the engine tree when persistence falls behind. Motivation is, when the node is node able to persist the blocks as fast as they arrive, then it should slowdown the admission of the blocks in the first place.

When the gap between the canonical tip and the last persisted block exceeds `--engine.persistence-backpressure-threshold` (default: 16 blocks), the main loop stops reading from the engine channel and blocks on the in-flight persistence task instead.

This shifts buffering from the costlier persistence pipeline (where each block carries trie updates and heavy state) to the lighter-weight incoming engine channel (where payloads sit as raw messages). The engine API has no backpressure semantics and standard CLs will timeout and resend after ~8s, so this can't shrink the inbound queue under sustained load but a) it prevents the expensive downstream queue from growing unboundedly and b) allows clients that do care about backpressure to handle it properly.

This PR supersedes https://github.com/paradigmxyz/reth/pull/23244. This takes a simpler approach. We take the L and buffer the payload messages in the channel. This yields a simpler implementation but sacrifies some fidelity of metrics.

### Observability

Two metrics are exposed under `consensus.engine.beacon`:

- **`backpressure_active`** (gauge): `1.0` when the engine loop is stalled waiting on persistence, `0.0` otherwise.
- **`backpressure_stall_duration`** (histogram): wall-clock time spent blocked on each persistence wait. Useful for understanding how long the engine loop is paused per stall event.

### Flag interaction

`--engine.persistence-backpressure-threshold` (default: 16) must be strictly greater than `--engine.persistence-threshold` (default: 2). If you increase the persistence threshold, you likely need to bump the backpressure threshold as well — otherwise the gap between them shrinks and backpressure kicks in too eagerly. The node will refuse to start if the backpressure threshold is not greater than the persistence threshold.

### DIFF
--- a/.changelog/zesty-bees-drink.md
+++ b/.changelog/zesty-bees-drink.md
@@ -1,0 +1,8 @@
+---
+reth-engine-primitives: minor
+reth-engine-tree: major
+reth-node-core: minor
+reth-cli-commands: minor
+---
+
+Added persistence backpressure to the engine tree: when the canonical-minus-persisted block gap exceeds a configurable threshold (`--engine.persistence-backpressure-threshold`, default 16), the engine loop stalls on the persistence receiver instead of processing new incoming messages. Added CLI argument, cross-field validation, metrics (`backpressure_active`, `backpressure_stall_duration`), and tests.
--- a/crates/cli/commands/src/node.rs
+++ b/crates/cli/commands/src/node.rs
@@ -178,6 +178,8 @@ where
             ext,
         } = self;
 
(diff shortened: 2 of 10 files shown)


mediocregopher (approved):

mattsse (approved):

yongkangc (approved):
Sick stuff

PR #23282, opened 2026-03-30
fix(discv5): use Weak reference in kbuckets bg task to release port on shutdown

romanbrodetski-ai:
## Summary

The `spawn_populate_kbuckets_bg` function holds an `Arc<discv5::Discv5>` which prevents the discv5 node from being fully dropped when the `Discv5` handle is dropped. This causes the UDP port to remain bound, making it impossible to restart a node on the same port without waiting for the OS to reclaim it.

This PR:
- Changes `spawn_populate_kbuckets_bg` to accept `Weak<discv5::Discv5>` instead of `Arc<discv5::Discv5>`
- The background task now upgrades the weak reference on each iteration and gracefully exits when the `Discv5` node is dropped
- Adds a test (`discv5_releases_port_on_drop`) that verifies the port is released after dropping the node

### DIFF
--- a/crates/net/discv5/src/lib.rs
+++ b/crates/net/discv5/src/lib.rs
@@ -714,7 +714,6 @@ mod test {
     use rand_08::thread_rng;
     use reth_chainspec::MAINNET;
     use std::{
-        env,
         net::UdpSocket,
         time::{Duration, Instant},
     };
--- a/crates/net/discv5/src/lib.rs
+++ b/crates/net/discv5/src/lib.rs
@@ -799,9 +799,8 @@ mod test {
         let secret_key = SecretKey::new(&mut thread_rng());
         let port = unused_udp_port();
 
-        let (node, updates) = start_discovery_node_with_key(&secret_key, port)
-            .await
-            .expect("should start discv5");
+        let (node, updates) =
+            start_discovery_node_with_key(&secret_key, port).await.expect("should start discv5");
         drop(updates);
         drop(node);
 
--- a/crates/net/discv5/src/lib.rs
+++ b/crates/net/discv5/src/lib.rs
@@ -225,7 +225,7 @@ impl Discv5 {
             bootstrap_lookup_interval,
             bootstrap_lookup_countdown,
             metrics.clone(),
-            discv5.clone(),
+            Arc::downgrade(&discv5),
         );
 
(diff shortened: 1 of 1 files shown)


RomanBrodetski:
@mattsse just a gentle ping on this one for feedback. If there are reasons you decide against merging this - also fine! 
Thanks, 
Roman

crates/net/discv5/src/lib.rs
```diff
@@ -738,6 +749,40 @@ mod test {
...
+        let secret_key = SecretKey::new(&mut thread_rng());
+        let port = 30366;
+
+        let (node, updates) = start_discovery_node_with_key(&secret_key, port)
+            .await
+            .expect("should start discv5");
+        drop(updates);
+        drop(node);
+
+        tokio::task::yield_now().await;
```
DaniPopes: this might be flaky since it doesn't guarantee to run the entire destructor

crates/net/discv5/src/lib.rs
```diff
@@ -573,17 +573,25 @@ pub fn spawn_populate_kbuckets_bg(
...
+        let Some(discv5_handle) = discv5.upgrade() else {
+            return;
+        };
+        let local_node_id = discv5_handle.local_enr().node_id();
+        drop(discv5_handle);
+
         // make many fast lookup queries at bootstrap, trying to fill kbuckets at furthest
         // log2distance from local node
         for i in (0..bootstrap_lookup_countdown).rev() {
+            let Some(discv5_handle) = discv5.upgrade() else {
```
DaniPopes: both of these upgrades are held across `sleep` calls, might not matter much tho

DaniPopes (approved):
makes sense, lgtm, nits

RomanBrodetski:
@DaniPopes thanks for the review - I addressed the nits and - hopefully - resolved the CI. Please approve the CI run and have an another look 🙌

mattsse (approved):
this entire disc5 start code is a bit of a mess -.-
using a weak here is fine at this point

RomanBrodetski:
Thanks a lot for the approval @mattsse !! I merged `main` to this branch and need another CI approval to merge I think..

mattsse:
thank you @RomanBrodetski

PR #23287, opened 2026-03-30
perf: reduce cacheline ping pong in workers availability

pepyakin:
Every time a core updates the availability counter it claims the cacheline invalidating it on other cores. Workers essentially ping pong this cacheline to each others. This showed up during perf c2c profiling.

The solution is to have each worker report its status on its own cacheline. When a non-worker core wants to know the status/counts, it will request the cachelines making them shared temporarily. This path should become a bit more expensive but AFAIU it's also a less frequent one.

At least that's the theory.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
+++ b/crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
@@ -777,8 +777,8 @@ where
             chunking_length,
             self.chunk_size,
             self.max_targets_for_chunking,
-            self.proof_worker_handle.available_account_workers(),
-            self.proof_worker_handle.available_storage_workers(),
+            self.proof_worker_handle.has_multiple_idle_account_workers(),
+            self.proof_worker_handle.has_multiple_idle_storage_workers(),
             MultiProofTargetsV2::chunks,
             |proof_targets| {
                 if let Err(e) =
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10512,6 +10512,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rlp",
  "crossbeam-channel",
+ "crossbeam-utils",
  "derive_more",
  "itertools 0.14.0",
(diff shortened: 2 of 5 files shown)


crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -777,8 +777,8 @@ where
             chunking_length,
             self.chunk_size,
             self.max_targets_for_chunking,
-            self.proof_worker_handle.available_account_workers(),
-            self.proof_worker_handle.available_storage_workers(),
+            self.proof_worker_handle.has_multiple_idle_account_workers(),
+            self.proof_worker_handle.has_multiple_idle_storage_workers(),
```
DaniPopes, line 781: these are both called eagerly but may be discarded inside of the function, can also be optimized to compute lazily if you think that could help

pepyakin:
derek bench otlp=false blocks=big baseline=f0d07c38be40c173abab5879b49a20dc4126c427 feature=f81632a

decofe:
cc @pepyakin

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/23781750665/job/69295853192)

**Config:** blocks: `big`, baseline: `f0d07c38be40c173abab5879b49a20dc4126c427`, feature: `f81632a`, otlp: `disabled`

pepyakin:
↑ to make it clear. I rebased and added a new commit. The 1st and 2nd recent runs are with the Dani's suggestion included, the 3rd is only the original change. As you can see it's showing an opposite result. Now, let's see what would happen if we re-run the original, pre-rebase, change.

pepyakin:
derek bench otlp=false blocks=big baseline=930f2a6eb257854138bb9e291c7013586fcd0223 feature=930f2a6eb257854138bb9e291c7013586fcd0223

mediocregopher (approved):

PR #23291, opened 2026-03-30
fix(engine): defer changeset static file truncation during reorgs

joshieDo:
Defers changeset static file truncation during `RemoveBlocksAbove` to the next `SaveBlocks`, preventing stale mmap panics when concurrent readers hold old memory-mapped handles.

Waits for all MDBX readers from the reorg era to drain (via `oldest_reader_txnid`) before applying the deferred truncation.

---

Follow-up: #23305 (pins RocksDB snapshot for cross-store read consistency, stacked on this PR)

### DIFF
--- a/crates/engine/tree/src/persistence.rs
+++ b/crates/engine/tree/src/persistence.rs
@@ -8,7 +8,7 @@ use reth_primitives_traits::{FastInstant as Instant, NodePrimitives};
 use reth_provider::{
     providers::ProviderNodeTypes, BlockExecutionWriter, BlockHashReader, ChainStateBlockWriter,
     DBProvider, DatabaseProviderFactory, ProviderFactory, SaveBlocksMode,
-    StaticFileProviderFactory, StaticFileWriter,
+    StaticFileProviderFactory, StaticFileWriter, StorageSettingsCache,
 };
 use reth_prune::{PrunerError, PrunerWithFactory};
 use reth_stages_api::{MetricEvent, MetricEventsSender};
@@ -156,9 +156,11 @@ where
         // By taking the strategies here and re-applying them later in `on_save_blocks`,
         // we ensure the truncation only happens when no reader needs the old data:
         // after the reorg completes and the next batch of blocks is persisted.
-        if let Some(target) = self.provider.static_file_provider().take_changeset_prunes() {
-            self.deferred_changeset_prune =
-                Some(self.deferred_changeset_prune.map_or(target, |prev| prev.min(target)));
+        if self.provider.cached_storage_settings().storage_v2 {
+            if let Some(target) = self.provider.static_file_provider().take_changeset_prunes() {
+                self.deferred_changeset_prune =
+                    Some(self.deferred_changeset_prune.map_or(target, |prev| prev.min(target)));
+            }
         }
 
         provider_rw.commit()?;
--- a/crates/engine/tree/src/persistence.rs
+++ b/crates/engine/tree/src/persistence.rs
@@ -156,11 +156,11 @@ where
         // By taking the strategies here and re-applying them later in `on_save_blocks`,
         // we ensure the truncation only happens when no reader needs the old data:
         // after the reorg completes and the next batch of blocks is persisted.
-        if self.provider.cached_storage_settings().storage_v2 {
-            if let Some(target) = self.provider.static_file_provider().take_changeset_prunes() {
-                self.deferred_changeset_prune =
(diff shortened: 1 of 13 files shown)


crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -142,6 +142,42 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
...
+        };
+        let storage = {
+            let mut writer = self.storage_change_sets.write();
+            writer.as_mut().and_then(|w| w.take_changeset_prune())
+        };
+        debug_assert_eq!(
+            account, storage,
+            "account and storage changeset prunes must target the same block"
+        );
+        account.or(storage)
```
joshieDo, line 165: this should be the same block, unsure if a new type for both is necessary

crates/storage/libmdbx-rs/src/environment.rs
```diff
@@ -412,6 +412,17 @@ impl Info {
         self.0.mi_recent_txnid as usize
     }
```
Rjected, line 413: can we document that this is the most recent txn id?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -142,6 +142,49 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
...
+    }
+
+    /// Re-queues previously deferred changeset prunes onto the writers.
+    ///
+    /// This only restores the queued prune-on-commit state. Safety still depends on the caller
+    /// having waited for stale MDBX readers to drain before the next commit executes the prune.
+    pub(crate) fn requeue_changeset_prunes(&self, last_block: BlockNumber) -> ProviderResult<()> {
```
Rjected, line 172: can we rename this from `requeue` to maybe `complete` or something that mentions that this finalizes / flushes the queued prunes?

crates/storage/provider/src/providers/static_file/writer.rs
```diff
@@ -142,6 +142,49 @@ impl<N: NodePrimitives> StaticFileWriters<N> {
         Ok(())
     }
 
+    /// Takes queued changeset prune strategies from the account and storage changeset writers,
+    /// returning the target block number. The writers will no longer execute these prunes on
+    /// commit.
+    ///
+    /// This is used by the deferred reorg-prune flow in the persistence service. Callers must
+    /// only reapply the returned prune after stale MDBX readers from the pre-unwind era have
```
Rjected, line 150:
```suggestion
    /// only apply the returned prune after stale MDBX readers from the pre-unwind era have
```

Rjected (commented):
mostly nits, based on my understanding of the problem this fix should work. I think we should add some docs explaining why we would not get any _new_ readers by the time we call `save_blocks` and wait  for any final readers to complete

crates/engine/tree/src/persistence.rs
```diff
@@ -32,6 +34,21 @@ pub struct PersistenceResult {
     pub commit_duration: Option<Duration>,
 }
 
+/// A deferred changeset prune from a `RemoveBlocksAbove` operation.
+///
+/// Changeset static file truncation is deferred to avoid truncating memory-mapped files while
+/// concurrent readers may still hold handles. The prune is applied at the start of the next
+/// `SaveBlocks`, after waiting for all MDBX readers from the reorg era to drain.
+#[derive(Debug)]
+struct DeferredChangesetPrune {
```
mattsse, line 43: lets please not add private types to top of file :)

crates/engine/tree/src/persistence.rs
```diff
@@ -160,6 +212,35 @@ where
...
+        // Apply any deferred changeset prunes from a previous RemoveBlocksAbove.
+        // Wait for all MDBX readers from the reorg era to complete before truncating,
+        // so no reader can observe a stale memory-mapped file.
+        if let Some(deferred) = self.deferred_changeset_prune.take() {
+            if let Some(prune_txn) = deferred.committed_txn_id {
+                while self
+                    .provider
+                    .db_ref()
+                    .oldest_reader_txnid()
+                    .is_some_and(|oldest| oldest < prune_txn)
```
mattsse, line 224: is this perhaps too relaxed?

this would also include rpc related transactions?
and there's a risk that an ethgetproof or adjacent tx will be open for a while?
should we either introduce a counter, that after like 3 blocks or so we always do this?

PR #23293, opened 2026-03-31
perf(provider): avoid full receipt clone on tx-range reads

BearSoltokien:
Clone only the requested in-memory receipt subrange for tx-range queries.


mattsse (commented):
this has been merged in another pr

PR #23303, opened 2026-03-31
feat(download): make snapshot API URL overridable

onbjerg:
Add `snapshot_api_url` field to `DownloadDefaults` so downstream projects can override the discovery API endpoint. Previously hardcoded to `snapshots.reth.rs`.


mattsse (approved):
pending @Rjected

Rjected (approved):
makes sense to me, this should be configurable

PR #23305, opened 2026-03-31
fix(storage): pin RocksDB snapshot for cross-store read consistency

joshieDo:
Stacked on #23291.

Fixes block gas mismatch during aggressive reorgs with `storage_v2`. `HistoricalStateProvider` takes a fresh RocksDB snapshot per query, but its MDBX read tx is from creation time. If `RemoveBlocksAbove` commits between the two, the reader sees unwound RocksDB indices with pre-unwind MDBX state — `InPlainState` fallback returns wrong values.

Fix: pin the RocksDB snapshot at `DatabaseProvider` creation alongside the MDBX read tx, with a `last_txnid` retry loop to ensure consistency. Thread it through to `HistoricalStateProvider` for all history lookups.

### DIFF
--- a/crates/storage/provider/src/providers/database/mod.rs
+++ b/crates/storage/provider/src/providers/database/mod.rs
@@ -1,7 +1,7 @@
 use crate::{
     providers::{
         state::latest::LatestStateProvider, NodeTypesForProvider, RocksDBProvider,
-        StaticFileProvider, StaticFileProviderRWRefMut,
+        RocksReadSnapshot, StaticFileProvider, StaticFileProviderRWRefMut,
     },
     to_range,
     traits::{BlockSource, ReceiptProvider},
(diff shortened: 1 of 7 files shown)


crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -259,7 +290,13 @@ impl<N: ProviderNodeTypes> ProviderFactory<N> {
             self.runtime.clone(),
             self.db.path(),
         )
-        .with_minimum_pruning_distance(self.minimum_pruning_distance))
+        .with_minimum_pruning_distance(self.minimum_pruning_distance);
+
+        if let Some(snapshot) = pinned_rocksdb_snapshot {
+            provider = provider.with_pinned_rocksdb_snapshot(snapshot);
+        }
```
joshieDo: option since storage.v1 doesnt need/use it

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -239,6 +239,35 @@ impl<N: ProviderNodeTypes<DB = DatabaseEnv>> ProviderFactory<N> {
 }
 
 impl<N: ProviderNodeTypes> ProviderFactory<N> {
+    /// Opens a read-only MDBX transaction and, for `storage_v2`, pins a matching `RocksDB`
+    /// snapshot.
+    ///
+    /// The helper retries if an MDBX write commit lands between opening the read transaction and
+    /// taking the `RocksDB` snapshot. This keeps both stores aligned for historical reads that
+    /// span MDBX state and `RocksDB` history indices.
```
Rjected, line 247: I think we need to note that this:
 1. allows rocksdb to be ahead of mdbx, ie mdbx at block 7 and rocksdb up to block 10, and 
 2. we need docs that outline what behavior is safe, for example the engine is safe because it creates txs sequentially, and we only ever perform SaveBlocks after we have closed a tx and added a block to the tree. I am not sure a single historical provider should outlive both a `remove_blocks` mdbx commit and a subsequent `save_blocks`

Rjected (commented):
just one comment otherwise i think this does fix the stated bug

PR #23308, opened 2026-03-31
feat(engine): include backpressure in reported persistence_wait, make wait-time mimic CL slot time

mediocregopher:
* persistence_wait_us now always reports the total time spent waiting on persistence: channel queuing time from backpressure plus any explicit wait-for-persistence wait.

* persistence_wait_us is no longer optional — it is always populated even when wait-for-persistence is not requested.

* reth_newPayload and wait-for-persistence toggles are removed from the bench GitHub workflow, always using reth_newPayload and relying on backpressure for persistence timing.

* wait-time in reth-bench now controls the interval between nP submissions _excluding_ validation time, mimicking a CL slot time. If nP does not complete within the wait-time then the next nP is submitted immediately.

### DIFF
--- a/crates/rpc/rpc-api/src/reth_engine.rs
+++ b/crates/rpc/rpc-api/src/reth_engine.rs
@@ -13,11 +13,10 @@ pub struct RethPayloadStatus {
     pub status: PayloadStatus,
     /// Server-side execution latency in microseconds.
     pub latency_us: u64,
-    /// Time spent waiting for persistence to complete, in microseconds.
-    ///
-    /// `None` when wasn't asked to wait.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub persistence_wait_us: Option<u64>,
+    /// Time spent waiting on persistence in microseconds, including both time spent
+    /// queued due to persistence backpressure and, when requested, the explicit wait
+    /// for in-flight persistence to complete.
+    pub persistence_wait_us: u64,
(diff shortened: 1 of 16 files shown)


crates/rpc/rpc-api/src/reth_engine.rs
```diff
@@ -13,11 +13,10 @@ pub struct RethPayloadStatus {
...
     pub latency_us: u64,
-    /// Time spent waiting for persistence to complete, in microseconds.
-    ///
-    /// `None` when wasn't asked to wait.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub persistence_wait_us: Option<u64>,
+    /// Time spent waiting on persistence in microseconds, including both time spent
+    /// queued due to persistence backpressure and, when requested, the explicit wait
+    /// for in-flight persistence to complete.
+    pub persistence_wait_us: u64,
```
pepyakin, line 19: Nit: why not break it down/split into persistence / backpressure?

pepyakin (approved):
Looks sensible.

crates/rpc/rpc-api/src/reth_engine.rs
mediocregopher: seemed like a YAGNI situation, i think with backpressure we won't really need wait-for-persistence going forward, and this is less changes to the plumbing

PR #23310, opened 2026-03-31
refactor(storage): remove changeset count APIs

joshieDo:
Removes the account and storage changeset count methods from the storage provider traits and their implementations.

The history indexing stages now log block-based progress while walking static-file changesets instead of precomputing an exact denominator.

This trims API surface that was only used for progress reporting and removes bespoke counting logic from the static-file provider.

alternative to https://github.com/paradigmxyz/reth/pull/23296


mattsse (approved):
cleaner indeed

PR #23318, opened 2026-04-01
fix(rpc): remove 0x prefix from admin_peers id to match go-ethereum format

### LINKED ISSUE #23316: bug: admin_peers[].id has spurious 0x prefix, incompatible with go-ethereum
## Summary

`admin_peers` returns each peer's `id` field with a `0x` prefix, while go-ethereum returns bare hex without any prefix. This makes reth incompatible with tools that parse peer IDs per the go-ethereum convention.

## Reproduction

```bash
curl -X POST http://localhost:8545 \
  -d '{"jsonrpc":"2.0","method":"admin_peers","params":[],"id":1}'
```

**reth response (wrong):**
```json
[{ "id": "0x6b36f791352f15eb3ec4f67787074ab8ad9d487e37c4401d383f0561a0a20507" }]
```

**go-ethereum response (correct):**
```json
[{ "id": "6b36f791352f15eb3ec4f67787074ab8ad9d487e37c4401d383f0561a0a20507" }]
```

## Root Cause

In `crates/rpc/rpc/src/admin.rs`:

```rust
id: keccak256(peer.remote_id.as_slice()).to_string(),
```

`keccak256()` returns `B256` (alloy-primitives `FixedBytes<32>`). Its `Display` / `to_string()` always includes the `0x` prefix:

```rust
// alloy-primitives FixedBytes Display impl
// prefix = true → includes "0x"
self.fmt_hex::<false>(f, true)
```

## Fix

Replace `.to_string()` with `alloy_primitives::hex::encode(...)`:

```rust
id: alloy_primitives::hex::encode(keccak256(peer.remote_id.as_slice())),
```

This is a one-line change that outputs a bare 64-character hex string, matching go-ethereum's format.

Note: the `node_info` handler in the same file already uses this pattern correctly (line 160), so this change also makes the two handlers consistent with each other.

## History

PR #19245 (merged Oct 2025) correctly changed the *value* from raw pubkey to `keccak256`, but left the `0x` prefix issue unfixed.

## Impact

Any tool, library, or test suite that compares peer IDs against go-ethereum output will see a mismatch due to the unexpected `0x` prefix.

asdv23:
## Summary

Removes the spurious `0x` prefix from `admin_peers[].id`, making it consistent with go-ethereum which returns bare hex without any prefix.

## Root Cause

`keccak256(...).to_string()` uses alloy-primitives `B256::Display`, which always prepends `0x` (via `fmt_hex` with `prefix = true`). Replacing it with `alloy_primitives::hex::encode(...)` produces a bare 64-character hex string.

As a bonus, the `node_info` handler in the same file already uses `hex::encode` at line 160 — so this change makes both handlers consistent.

## Change

**File:** `crates/rpc/rpc/src/admin.rs`

```diff
-id: keccak256(peer.remote_id.as_slice()).to_string(),
+id: alloy_primitives::hex::encode(keccak256(peer.remote_id.as_slice())),
```

## Before / After

| | Value | Length | Has `0x`? |
|---|---|---|---|
| Before | `0x6b36f791352f15eb3ec4f67787074ab8ad9d487e37c4401d383f0561a0a20507` | 66 chars | Yes ❌ |
| After | `6b36f791352f15eb3ec4f67787074ab8ad9d487e37c4401d383f0561a0a20507` | 64 chars | No ✅ |

Closes #23316

## Related

- #19245 fixed the hash algorithm (raw pubkey → keccak256) but left the `0x` prefix issue
- go-ethereum reference: https://github.com/ethereum/go-ethereum/blob/master/p2p/server.go

### DIFF
--- a/crates/rpc/rpc/src/admin.rs
+++ b/crates/rpc/rpc/src/admin.rs
@@ -76,7 +76,7 @@ where
 
         for peer in peers {
             infos.push(PeerInfo {
-                id: keccak256(peer.remote_id.as_slice()).to_string(),
+                id: alloy_primitives::hex::encode(keccak256(peer.remote_id.as_slice())),
                 name: peer.client_version.to_string(),
                 enode: peer.enode,
                 enr: peer.enr,


mattsse (approved):
Looks good to me. The admin_peers.id serialization now matches geth's bare lowercase hex format.

PR #9448, opened 2024-07-11
feat(rpc/admin): compatible with geth's admin_nodeInfo

jsvisa:
In this PR, fixed two data inconsistent with geth's response of `admin_nodeInfo`:

1. `id` should be the node's public key, wait for alloy to be released a new version, ref https://github.com/alloy-rs/alloy/pull/1038
2. Add the xxxBlock and yyyTime to `protocols.eth.config`, after this issue fixed, we can use reth to fetch the prestate to run trace tests and any other state-related testing, ref https://github.com/ethereum/go-ethereum/blob/2d9d42376436cd275c28056cffd0eb97cb8daed8/eth/tracers/internal/tracetest/makeTest.js#L30 

It's not clever to extract the xxxBlocks in the admin module, should be implement in chainspec instead.

PR #23319, opened 2026-04-01
fix(rpc): admin_nodeInfo.id now returns keccak256 node ID matching go-ethereum format

### LINKED ISSUE #23317: bug: admin_nodeInfo.id returns compressed secp256k1 pubkey instead of keccak256 node ID (incompatible with go-ethereum)
## Summary

`admin_nodeInfo` returns the `id` field as a **compressed secp256k1 public key** (33 bytes, 66 hex chars, prefixed `02`/`03`), which is incompatible with the [go-ethereum standard](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-admin) where `id` is `keccak256(X||Y)` of the public key (32 bytes, 64 hex chars, no `0x` prefix).

## Reproduction

```bash
curl -X POST http://localhost:8545 \
  -d '{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}'
```

**reth response (wrong):**
```json
{ "id": "02a1b2c3d4e5f6..." }
```
66 characters, starts with `02` or `03` — this is a compressed secp256k1 public key.

**go-ethereum response (correct):**
```json
{ "id": "6b36f791352f15eb3ec4f67787074ab8ad9d487e37c4401d383f0561a0a20507" }
```
64 characters — this is `keccak256(X||Y)` of the node's public key.

## Root Cause

In `crates/rpc/rpc/src/admin.rs`, the `node_info()` handler does:

```rust
id: id2pk(enode.id)
    .map(|pk| pk.to_string())   // ← wrong: secp256k1 LowerHex returns compressed key
    .unwrap_or_else(|_| alloy_primitives::hex::encode(enode.id.as_slice())),
```

The secp256k1 crate's `LowerHex` implementation calls `self.serialize()`, which returns the **compressed** 33-byte form.

**go-ethereum** computes the node ID as ([source](https://github.com/ethereum/go-ethereum/blob/master/p2p/enode/idscheme.go)):
```go
math.ReadBits(pubkey.X, buf[:32])
math.ReadBits(pubkey.Y, buf[32:])
return crypto.Keccak256(buf)
```

## Fix

```rust
id: alloy_primitives::hex::encode(keccak256(enode.id.as_slice())),
```

Since `enode.id` is already the raw 64-byte X||Y coordinates (`PeerId` = `B512`), calling `keccak256` on it directly produces the correct 32-byte node ID, matching go-ethereum's format.

Note: `keccak256` is already imported in this file (used by `admin_peers`). The `id2pk` import can be removed as it is no longer needed.

## Consistency Note

Notably, the `admin_peers` handler in the same file **already uses the correct approach**:

```rust
id: keccak256(peer.remote_id.as_slice()).to_string(),
```

So `admin_nodeInfo` and `admin_peers` were inconsistent with each other.

## Impact

Any tool or library that parses `admin_nodeInfo.id` and expects the go-ethereum format (e.g., monitoring tools, peer discovery utilities, interop test suites) will receive an unexpected 66-char compressed key instead of a 64-char hash.

Previously attempted in #9448 but the format discrepancy was not fully resolved.

asdv23:
## Summary

Fixes `admin_nodeInfo.id` to return the correct node ID format matching go-ethereum: `keccak256(X||Y)` of the public key as a 64-character hex string without `0x` prefix.

Previously, `id2pk(enode.id).map(|pk| pk.to_string())` was used, which calls the secp256k1 crate's `LowerHex` implementation that returns the **compressed** public key (33 bytes, 66 hex chars, `02`/`03` prefix). This is not the node ID format used by go-ethereum or the Ethereum devp2p spec.

This also makes `admin_nodeInfo` **consistent** with `admin_peers`, which already correctly uses `keccak256(peer.remote_id.as_slice())`.

## Changes

- `crates/rpc/rpc/src/admin.rs`: Replace `id2pk(enode.id).map(|pk| pk.to_string()).unwrap_or_else(...)` with `alloy_primitives::hex::encode(keccak256(enode.id.as_slice()))`
- Remove now-unused `id2pk` import from `reth_network_peers`

## Before / After

| | Value | Length | Format |
|---|---|---|---|
| Before | `02a1b2c3...` | 66 chars | compressed secp256k1 pubkey |
| After | `6b36f791...` | 64 chars | `keccak256(X\|\|Y)` — go-ethereum compatible |

## Why this works

`enode.id` is a `PeerId` (`B512`), which already holds the raw 64-byte uncompressed public key coordinates `X||Y`. Calling `keccak256` on those bytes directly produces the correct 32-byte node ID — exactly what go-ethereum does:

```go
// go-ethereum: p2p/enode/idscheme.go
math.ReadBits(pubkey.X, buf[:32])
math.ReadBits(pubkey.Y, buf[32:])
return crypto.Keccak256(buf)
```

`keccak256` is already imported in this file (used by `admin_peers`), so no new dependencies are added.

Closes #23317

## Related

- Attempted fix in #9448 (did not fully resolve the format)
- go-ethereum reference: https://github.com/ethereum/go-ethereum/blob/master/p2p/enode/idscheme.go

### DIFF
--- a/crates/rpc/rpc/src/admin.rs
+++ b/crates/rpc/rpc/src/admin.rs
@@ -9,7 +9,7 @@ use async_trait::async_trait;
 use jsonrpsee::core::RpcResult;
 use reth_chainspec::{EthChainSpec, EthereumHardfork, EthereumHardforks, ForkCondition};
 use reth_network_api::{NetworkInfo, Peers};
-use reth_network_peers::{id2pk, AnyNode, NodeRecord};
+use reth_network_peers::{AnyNode, NodeRecord};
 use reth_network_types::PeerKind;
 use reth_rpc_api::AdminApiServer;
 use reth_rpc_server_types::ToRpcResult;
@@ -155,9 +155,7 @@ where
         ]);
 
         Ok(NodeInfo {
-            id: id2pk(enode.id)
-                .map(|pk| pk.to_string())
-                .unwrap_or_else(|_| alloy_primitives::hex::encode(enode.id.as_slice())),
+            id: alloy_primitives::hex::encode(keccak256(enode.id.as_slice())),
             name: status.client_version,
             enode: enode.to_string(),
             enr: self.network.local_enr().to_string(),


mattsse (approved):
Looks good to me. The admin_nodeInfo.id serialization now matches geth's bare lowercase keccak256 node ID format.

PR #23322, opened 2026-04-01
Fix/engine reorg async task

0xWeakSheep:
Offloads reorg head construction in `EngineReorg` to `tokio::task::spawn_blocking` so `poll_next` no longer performs expensive synchronous work on the stream path.

The wrapper now forwards the triggering `NewPayload` immediately, tracks an in-flight reorg build task, and enqueues reorg `newPayload`/`forkchoiceUpdated` messages only after the background task completes. This reduces stream stalls under heavy reorg construction work.

Adds unit tests for reorg trigger gating logic.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,7 +1,7 @@
 name: Pull Request
 
 on:
-  pull_request:
+  pull_request_target:
     types:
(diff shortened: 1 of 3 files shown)


mattsse (commented):
we can keep this as is

PR #23329, opened 2026-04-01
fix(rpc): avoid evicting cached entries on replacement

apetro2:
Avoids evicting the oldest RPC cache entry when replacing an existing key in `MultiConsumerLruCache`, and keeps tracked `memory_usage` aligned with the actual cached contents.
  Adds regression tests covering replacement at capacity and replacement memory accounting.


mattsse (commented):
we can keep this as is, because we dont insert concurrently

PR #23330, opened 2026-04-01
feat: add bal rpc methods

Soubhik-10:
code taken from [ref](https://github.com/paradigmxyz/reth/pull/22406/changes#diff-66d6ef1fd7427f869a3eac4ee88780ba383406b700e5ee8ab9762d8e1221616bR1-R13)
cc @Rimeeeeee


crates/rpc/rpc-eth-api/src/helpers/bal.rs
```diff
@@ -0,0 +1,79 @@
...
+                .recovered_block(block_hash.into())
+                .await?
+                .ok_or_else(|| EthApiError::HeaderNotFound(block_hash.into()))?;
+
+            // Check if the block has been pruned (EIP-4444)
+            let earliest_block = self.provider().earliest_block_number()?;
+            if block.header().number() < earliest_block {
+                return Err(EthApiError::PrunedHistoryUnavailable.into());
+            }
+            // Check if the block is pre-Amsterdam, as access lists are not available for those
```
mediocregopher: Sorry I'm catching up to all this BAL work, I'm guessing we have some way of pre-enabling this for development/testing?

Soubhik-10: This was in the specs, i dont think there is a way to enable it ..we can just remove the check for now if you want

mediocregopher: will defer to @mattsse, but yeah having some way to get BALs for pre-amsterdam blocks is a requirement for big block testing. Maybe we enable that via a CLI arg or something.

crates/rpc/rpc-eth-api/src/helpers/bal.rs
```diff
@@ -0,0 +1,79 @@
...
+            let block = self
+                .recovered_block(block_hash.into())
+                .await?
+                .ok_or_else(|| EthApiError::HeaderNotFound(block_hash.into()))?;
+
+            // Check if the block has been pruned (EIP-4444)
+            let earliest_block = self.provider().earliest_block_number()?;
+            if block.header().number() < earliest_block {
+                return Err(EthApiError::PrunedHistoryUnavailable.into());
+            }
```
mattsse: I think we can remove this because this is now checked by recovered_block already

crates/rpc/rpc-eth-api/src/helpers/bal.rs
mattsse: @Soubhik-10 lets remove this check for now, because this endpoint will come in handy for pre fork testing

crates/rpc/rpc-eth-api/src/core.rs
```diff
@@ -913,17 +915,31 @@ where
     }
 
     /// Handler for: `eth_getBlockAccessListByBlockHash`
-    async fn block_access_list_by_block_hash(&self, hash: B256) -> RpcResult<Option<Bytes>> {
-        trace!(target: "rpc::eth", ?hash, "Serving eth_getBlockAccessListByBlockHash");
-        Err(internal_rpc_err("unimplemented"))
-    }
+    async fn block_access_list_by_block_hash(&self, block_hash: B256) -> RpcResult<Option<Value>> {
```
mattsse, line 917: I think returning value here makes sense for now, but it has some overhead because we then need to serialize this as string again. 

but lets keep it for now

mattsse (changes requested):
a few nits

@klkvr I think we need this additional trait, also ptal at the how this currently uses the blockexecutor, I believe the final design is still tbd

mattsse (approved):

PR #23334, opened 2026-04-01
perf(rpc): pre-allocate vectors in `eth_feeHistory`

AKABABA-ETH:
Vectors in `eth_feeHistory` were growing dynamically despite `block_count` being known upfront. Switched from `Vec::new()` to `Vec::with_capacity()` for all five vectors, eliminating ~10 reallocations per vector on large requests (block_count=1024).

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/fee.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/fee.rs
@@ -137,13 +137,14 @@ pub trait EthFees:
             let start_block = end_block_plus - block_count;
 
             // Collect base fees, gas usage ratios and (optionally) reward percentile data
-            let mut base_fee_per_gas: Vec<u128> = Vec::new();
-            let mut gas_used_ratio: Vec<f64> = Vec::new();
+            // Pre-allocate capacity: base_fee and blob_fee need +1 for the next block's fee
+            let mut base_fee_per_gas: Vec<u128> = Vec::with_capacity(block_count as usize + 1);
+            let mut gas_used_ratio: Vec<f64> = Vec::with_capacity(block_count as usize);
 
-            let mut base_fee_per_blob_gas: Vec<u128> = Vec::new();
-            let mut blob_gas_used_ratio: Vec<f64> = Vec::new();
+            let mut base_fee_per_blob_gas: Vec<u128> = Vec::with_capacity(block_count as usize + 1);
+            let mut blob_gas_used_ratio: Vec<f64> = Vec::with_capacity(block_count as usize);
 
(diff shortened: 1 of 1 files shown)


crates/rpc/rpc-eth-api/src/helpers/fee.rs
```diff
@@ -137,13 +137,14 @@ pub trait EthFees:
...
+            let mut base_fee_per_gas: Vec<u128> = Vec::with_capacity(block_count as usize + 1);
+            let mut gas_used_ratio: Vec<f64> = Vec::with_capacity(block_count as usize);
 
-            let mut base_fee_per_blob_gas: Vec<u128> = Vec::new();
-            let mut blob_gas_used_ratio: Vec<f64> = Vec::new();
+            let mut base_fee_per_blob_gas: Vec<u128> = Vec::with_capacity(block_count as usize + 1);
+            let mut blob_gas_used_ratio: Vec<f64> = Vec::with_capacity(block_count as usize);
 
-            let mut rewards: Vec<Vec<u128>> = Vec::new();
+            let mut rewards: Vec<Vec<u128>> = Vec::with_capacity(block_count as usize);
```
mattsse: `eth_feeHistory` often runs without reward percentiles, and in that case we never push into `rewards` and still return `reward: None`. This turns the new `with_capacity(block_count as usize)` into a net new allocation on the no-rewards path. Please keep `rewards` as `Vec::new()` unless percentiles were actually requested, or make the capacity conditional on a non-empty percentile list.

mattsse (changes requested):
This is headed in the right direction, but I think we should make one small change before merging. The four fee/gas vectors are good candidates for preallocation, but `rewards` is now preallocated on every call, including the common path where `reward_percentiles` is omitted and no rewards are returned at all. Since this PR is framed as reducing allocations in `eth_feeHistory`, I would avoid introducing a new unconditional allocation on that path.

Please make the `rewards` capacity conditional on actually requesting reward percentiles, and this should be in good shape.

crates/rpc/rpc-eth-api/src/helpers/fee.rs
mattsse: Pushed bd1bd926fa to make `rewards` lazy when `reward_percentiles` is omitted, while keeping the other vector preallocations from this PR intact. The focused `fee_history` tests passed locally in the takeover worktree.

mattsse (approved):
Addressed in bd1bd926fa3ab7d91068ecdc313152d91c075ccf. `rewards` now stays lazy when `reward_percentiles` is omitted, and the focused `fee_history` tests passed locally.

PR #23335, opened 2026-04-01
fix(provider): cap storage_v2 unwind history by MDBX tip

joshieDo:
Alternative to #23291 and #23305.

For `storage_v2` unwind, this keeps the MDBX-first ordering but waits after the MDBX commit and before `RocksDB` so a reader opened before the unwind cannot observe unwound MDBX state and then later route through pre-unwind Rocks history.

It then caps `RocksDB` history lookups by the reader's MDBX-visible tip instead of adding a second wait/fence. Example: if a reader's MDBX snapshot only sees block `N`, a later Rocks snapshot ignores history entries above `N` and falls back to the MDBX-visible prefix instead of routing into a newer changeset.

Also drops the validation RO provider before `unwind_provider_rw()` so the unwind wait cannot block on that provider itself.

### DIFF
--- a/crates/storage/db-api/src/database.rs
+++ b/crates/storage/db-api/src/database.rs
@@ -123,43 +123,39 @@ impl<DB: Database> Database for &DB {
 
 /// Object-safe adapter for reader-txn tracking and unwind fencing.
 pub trait ReaderTxnTracker: Send + Sync {
-    /// Returns the txnid of the oldest active reader snapshot.
-    ///
-    /// This is not a unique per-reader id; it is the committed txnid the reader is pinned to.
-    fn oldest_reader_txnid(&self) -> Option<u64>;
+    /// Waits until all readers older than the latest committed txnid have drained.
+    fn wait_for_pre_commit_readers(&self);
 
-    /// Returns the latest committed MDBX txnid.
-    fn last_txnid(&self) -> Option<u64>;
-
-    /// Forces a real commit so the latest txnid advances, then returns it.
-    fn commit_fence(&self) -> Result<Option<u64>, DatabaseError>;
-
-    /// Waits until all readers pinned to a committed txnid older than `cutoff_txnid` have drained,
-    /// polling every 10ms.
-    fn wait_for_readers_before_txnid(&self, cutoff_txnid: u64) {
-        while self.oldest_reader_txnid().is_some_and(|oldest| oldest < cutoff_txnid) {
-            std::thread::sleep(std::time::Duration::from_millis(10));
-        }
-    }
+    /// Forces a real commit boundary and waits until readers older than that new txnid have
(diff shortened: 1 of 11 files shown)


crates/storage/db-api/src/database.rs
```diff
@@ -25,6 +26,22 @@ pub trait Database: Send + Sync + Debug {
...
     fn path(&self) -> PathBuf;
 
+    /// Returns the transaction ID of the oldest active reader, if available.
+    ///
+    /// This is the committed txnid of the snapshot the reader is pinned to, not a unique per-reader
+    /// identifier, so multiple readers can report the same txnid.
+    ///
+    /// Used to check whether stale readers from a previous write transaction have completed.
+    /// Returns `None` if no readers are active or the backend does not support this query.
+    fn oldest_reader_txnid(&self) -> Option<u64> {
```
joshieDo: this is especially important to understand

crates/storage/provider/src/providers/database/provider.rs
```diff
@@ -212,6 +212,8 @@ pub struct DatabaseProvider<TX, N: NodeTypes> {
     minimum_pruning_distance: u64,
     /// Database provider metrics
     metrics: metrics::DatabaseProviderMetrics,
+    /// Database handle used to inspect active MDBX readers during unwind commits.
+    reader_txn_tracker: Option<Arc<dyn ReaderTxnTracker>>,
```
joshieDo, line 216: this is essentially `Database/DatabaseEnv`, but didn't want to make it such an hard dependency. wdyt

klkvr: yeah having this generic over `DB` would be a mess

crates/storage/provider/src/providers/rocksdb/provider.rs
```diff
@@ -1526,27 +1549,65 @@ impl<'db> RocksReadSnapshot<'db> {
             return fallback();
         };
         let chunk = BlockNumberList::decompress(value_bytes)?;
+
+        // Only the prefix up to `visible_tip` should influence routing for this reader. Any
+        // later Rocks history entries are not yet visible in the companion MDBX snapshot.
+        let visible_count = chunk.rank(visible_tip);
         let (rank, found_block) = compute_history_rank(&chunk, block_number);
```
klkvr, line 1548: hmm couldn't we simplify this by assuming that `visible_tip > block_number`?

i was thinking that given that this always holds we basically have 2 cases
- `found_block > visible_tip` which means that rocksdb is ahead AND we can safely treat this as `found_block = None` because the earliest block the key was changed in is after our mdbx view
- `found_block <= visible_tip` which can be handled as is

klkvr (approved):

PR #23338, opened 2026-04-02
fix(rpc): apply count only after after is consumed

crypomen9:
Apply count only after after has been consumed in trace_filter to preserve pagination semantics.


mattsse (approved):
Looks good to me.

PR #23345, opened 2026-04-02
feat(revm): add `ExecutionWitnessRecord::into_execution_witness` helper

figtracer:
## Summary
- Adds `into_execution_witness` method to `ExecutionWitnessRecord` behind the `witness` feature
- Generates state proofs and serializes ancestor headers into a complete `ExecutionWitness`
- Adds `alloy-rlp` and `alloy-rpc-types-debug` as optional deps gated behind `witness`

## Motivation
The witness assembly pattern (destructure record → generate state proofs → resolve BLOCKHASH headers → build `ExecutionWitness`) is duplicated in reth's `debug_execution_witness_for_block` and downstream in [`paradigmxyz/stateless`](https://github.com/paradigmxyz/stateless/blob/main/testing/ef-tests/src/cases/blockchain_test.rs#L338). This extracts it into a reusable method on `ExecutionWitnessRecord`.

### DIFF
--- a/crates/revm/src/witness.rs
+++ b/crates/revm/src/witness.rs
@@ -80,4 +80,44 @@ impl ExecutionWitnessRecord {
         record.record_executed_state(state);
         record
     }
+
+    /// Converts this record into a complete [`alloy_rpc_types_debug::ExecutionWitness`] by
+    /// generating state proofs and fetching ancestor block headers.
+    ///
+    /// The `block_number` is the number of the block being witnessed. Ancestor headers are
+    /// included based on the lowest block number referenced by BLOCKHASH opcodes during
+    /// execution, or just the parent header if BLOCKHASH was not called.
+    #[cfg(feature = "witness")]
+    pub fn into_execution_witness<SP, HP>(
+        self,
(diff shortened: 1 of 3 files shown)


crates/revm/src/witness.rs
```diff
@@ -1,5 +1,11 @@
 use alloc::vec::Vec;
 use alloy_primitives::{keccak256, Bytes, B256};
+#[cfg(feature = "witness")]
+use alloy_rlp::Encodable;
+#[cfg(feature = "witness")]
+use alloy_rpc_types_debug::ExecutionWitness;
+#[cfg(feature = "witness")]
+use reth_storage_api::{HeaderProvider, StateProofProvider};
```
mattsse: can we use full paths in the fn instead, then we dont need to have these feature gated imports

mattsse (changes requested):

crates/revm/src/witness.rs
figtracer: on it

mattsse (approved):

PR #23346, opened 2026-04-02
refactor(rocksdb): use secondary instances for read only providers

klkvr:
Read-only rocksdb instances are designed for brief queries and do not support catching up with primary instance writing data

This PR changes readonly rocksdb providers to use `Secondary` instances that expose `try_catch_up_with_primary` fn allowing to update the database view to match the new writes


Rjected (approved):
lgtm, assuming we will call catch up etc in a followup

PR #23350, opened 2026-04-02
perf(engine): reduce write lock hold time in save_cache

AKABABA-ETH:
Moved expensive operations (`insert_state`, `update_metrics`, `recv`) outside the write lock in `save_cache()`. The lock is now only held for the final quick assignment, reducing block time from tens of milliseconds to microseconds.


mattsse (commented):
no

PR #23354, opened 2026-04-03
refactor(engine): enable sharing of PayloadProcessor and state caches

sieniven:
## Summary

This PR enables the sharing of the payload processor, so that other external customized components are able to leverage on the payload processor's optimized caches and payload processing logic.

For example, this allows external components building on top of reth, such as flashblocks on optimism chains or other customized payloads, to have their customized validators to share the same payload processor logic and its state caches. The customized validator can encapsulate its own customizable logic for that unique payload, while directly leveraging on the payload processor for optimized payload execution and merklization.

Note - the state consistency is guaranteed on the customizer's end, and the payload processor cache states require an engine validator controller for atomic payload validation operations between the default engine validation and the customized layer.


sieniven:
@mattsse hi, any updates regarding this PR? would appreciate feedback and whether this is possible.

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -87,8 +88,203 @@ type IteratorPayloadHandle<Evm, I, N> = PayloadHandle<
 >;
 
 /// Entrypoint for executing the payload.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct PayloadProcessor<Evm>
+where
+    Evm: ConfigureEvm,
+{
+    inner: Arc<Mutex<PayloadProcessorInner<Evm>>>,
```
mattsse, line 96: I dont think this makes a lot of sense

mattsse (commented):
can you clarify why you need this to be clone

this type's pub fns dont appear to be super useful to me so I dont fully understand your use case

sieniven:
> can you clarify why you need this to be clone
> 
> this type's pub fns dont appear to be super useful to me so I dont fully understand your use case

@mattsse adding a follow up to this PR for better visibility and clarity:

https://github.com/okx/xlayer-reth/pull/176/ is an example of this requirement to expose the underlying Reth engine's payload processor for reth-sdk.

in our use case, for customized payload validators, currently the payload processor is not sharable and encapsulated privately inside the default reth payload validator. however, we hope to be able to use it and work on the layer of abstraction to add a custom flashblocks sequence validator that is able to leverage on these optimizations.

wrapping the payload processor in to implement the sync and send traits also allows us to run multiple validators. in our use case above, our flashblocks RPC layer has the newly defined flashblocks sequence validator that shares the payload processor with the underlying reth's engine validator. we also have a custom flashblocks sync while maintaining the default EL sync, which this allows us to share the sparse trie cache states and the execution cache states between the 2 validators

hope this helps, cheers and thanks for taking the time to review 🫡

crates/engine/tree/src/tree/payload_processor/mod.rs
sieniven: sorry for the lack of context - refer to https://github.com/paradigmxyz/reth/pull/23354#issuecomment-4384850673

PR #23357, opened 2026-04-03
feat: catch-up for read-only `ProviderFactory`s

klkvr:
Implements catch-up logic for read-only `ProviderFactory`. 

It is implemented like the following:
https://github.com/paradigmxyz/reth/blob/6ea9ce81fe7c26781a99acb13263fb01f461fa11/crates/storage/provider/src/providers/database/mod.rs#L269-L297

This function is invoked in 2 cases:
1. For every created `DatabaseProvider` or before every read directly from static files (those bypass mdbx entirely)
2. Every time when mdbx dir changes — this part is identical to static files watcher we have right now and is kept to make sure that read-only factories don't fall too far behind when unused, causing potentially long catch up to happen on next access


crates/storage/db/src/lib.rs
```diff
@@ -230,6 +230,27 @@ pub mod test_utils {
...
+    /// concurrent primary + read-only secondary provider scenarios.
+    ///
+    /// Must be called before any MDBX environment is opened.
+    ///
+    /// # Safety
+    ///
+    /// This uses `MDBX_DBG_LEGACY_MULTIOPEN` which recovers POSIX file locks on close.
+    /// It may cause unexpected pauses and does not perfectly mirror multi-process behavior.
+    /// Use only in tests.
+    pub fn enable_legacy_multiopen() {
```
joshieDo, line 245: we might be able to use this in tempo test suite as well cc @SuperFluffy

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -82,6 +94,10 @@ pub struct ProviderFactory<N: NodeTypesWithDB> {
     runtime: reth_tasks::Runtime,
     /// Minimum distance from tip required before pruning can occur.
     minimum_pruning_distance: u64,
+    /// State for on-demand syncing of `RocksDB` secondary and static file indexes.
+    ///
+    /// Only set for read-only factories.
+    read_only_sync: Option<Arc<ReadOnlySyncState>>,
```
joshieDo, line 101: should we be more specific that this should be used for out of process concurrent access?

crates/storage/provider/src/providers/database/mod.rs
```diff
@@ -180,6 +197,102 @@ impl<N: NodeTypesWithDB> ProviderFactory<N> {
         self
     }
 
+    /// Enables on-demand syncing of `RocksDB` secondary and static file indexes for read-only
+    /// factories. Initializes the tracker to the current MDBX txn ID.
+    ///
+    /// If `watch` is true, also watches the static files directory for `.conf` changes to
+    /// eagerly catch up in the background.
+    pub fn with_read_only_sync(mut self, watch: bool) -> Self
```
joshieDo, line 206: should we be more specific that this should be used for out of process concurrent access?

joshieDo (approved):

PR #23365, opened 2026-04-06
fix(rpc): guard gas_used_ratio against division by zero

KENILSHAHH:
## Summary

- Adds `checked_gas_used_ratio` helper to safely compute `gas_used / gas_limit`, returning `0.0` when `gas_limit` is zero instead of producing `NaN`
- Applies the helper in `FeeHistoryEntry::new` and the `fee_history` fallback path where the ratio was previously computed with raw division
- Adds unit tests covering the zero, partial, and full usage cases

## Motivation

Unguarded `gas_used as f64 / gas_limit as f64` produces `NaN` when `gas_limit` is `0`. `NaN` is not valid JSON per RFC 8259 and would cause serialization failures in `eth_feeHistory` responses. This mirrors the existing `checked_blob_gas_used_ratio` pattern already used for blob gas.

## Test plan

- [x] `cargo +nightly nextest run -p reth-rpc-eth-types -- test_checked_gas_used_ratio` — new test passes
- [x] `cargo +nightly check -p reth-rpc-eth-types` — compiles clean
- [x] `cargo +nightly check -p reth-rpc-eth-api` — downstream consumer compiles clean

### DIFF
--- a/crates/rpc/rpc-eth-api/src/helpers/fee.rs
+++ b/crates/rpc/rpc-eth-api/src/helpers/fee.rs
@@ -10,7 +10,8 @@ use futures::{Future, StreamExt};
 use reth_chainspec::{ChainSpecProvider, EthChainSpec};
 use reth_primitives_traits::BlockBody;
 use reth_rpc_eth_types::{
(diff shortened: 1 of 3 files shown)


mattsse (commented):
gas limit 0 doesnt make sense

PR #23370, opened 2026-04-07
fix(ci): pin CPU frequency to nominal clock in bench workflow

decofe:
The existing boost-disable logic writes to `/sys/devices/system/cpu/cpufreq/boost` which doesn't exist on `amd-pstate-epp` (our GHA runners use AMD EPYC 4585PX). The `echo 0` silently fails, so boost was never actually disabled — cores could swing between 600 MHz and 5.7 GHz during benchmarks.

Reads nominal frequency from CPPC (`acpi_cppc/nominal_freq`) or Intel's `base_frequency` sysfs and clamps `scaling_min_freq`/`scaling_max_freq` to it on all cores. On the EPYC 4585PX runners this pins cores at 4.3 GHz. Also logs the pinned frequencies and restores the full range + powersave governor on cleanup.

Prompted by: pep

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -811,10 +811,27 @@ jobs:
       # System tuning for reproducible benchmarks
       - name: System setup
         run: |
+          # Switch amd-pstate to passive mode so the kernel governor
+          # controls frequency directly (EPP is ignored in passive).
+          echo passive | sudo tee /sys/devices/system/cpu/amd_pstate/status 2>/dev/null || true
           sudo cpupower frequency-set -g performance || true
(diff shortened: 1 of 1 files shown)


decofe:
cc @pepyakin

🚀 Benchmark started! [View job](https://github.com/paradigmxyz/reth/actions/runs/24069996880/job/70204479974)

**Config:** 500 blocks, 100 warmup blocks, baseline: `main`, feature: `pep/fix-bench-frequency-pinning`

laibe (approved):

shekhirin (approved):

mediocregopher (approved):

DaniPopes (approved):

PR #23371, opened 2026-04-07
fix: unmount schelk volume before recover in snapshot script

laibe:
Fixes "Volume is already mounted" failures when a previous run left the volume mounted on the same runner.

### DIFF
--- a/.github/scripts/bench-reth-snapshot.sh
+++ b/.github/scripts/bench-reth-snapshot.sh
@@ -77,8 +77,11 @@ trap 'rm -f -- "$MANIFEST_TMP"' EXIT
 echo "$MANIFEST_CONTENT" \
   | jq --arg base "$BASE_URL" '.base_url = $base' > "$MANIFEST_TMP"
 
-# Prepare mount
-mountpoint -q "$SCHELK_MOUNT" && sudo schelk recover -y || true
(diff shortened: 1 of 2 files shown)


emmajam (approved):

.github/scripts/bench-reth-snapshot.sh
```diff
@@ -77,8 +77,11 @@ trap 'rm -f -- "$MANIFEST_TMP"' EXIT
 echo "$MANIFEST_CONTENT" \
   | jq --arg base "$BASE_URL" '.base_url = $base' > "$MANIFEST_TMP"
 
-# Prepare mount
-mountpoint -q "$SCHELK_MOUNT" && sudo schelk recover -y || true
+# Prepare mount – unmount first if a previous run left the volume mounted
+if mountpoint -q "$SCHELK_MOUNT"; then
+  sudo umount -l "$SCHELK_MOUNT" || true
```
pepyakin: IIRC this would destroy the scratch volume, because dm-era (which tracks changes) would also be destroyed and so we won't know what blocks to restore and thus we will be forced to do full-recover.

So omit the umount!

pepyakin (changes requested):

pepyakin (approved):

shekhirin (approved):

PR #23383, opened 2026-04-08
chore: updated expected failure  and removed default true from bal builder

Rimeeeeee:
* updated expected failures in paris
* instead of using with_bal_builder_if(true),set is seperately if amsterdam is enabled

cc: @mattsse

### DIFF
--- a/crates/evm/evm/src/execute.rs
+++ b/crates/evm/evm/src/execute.rs
@@ -573,11 +573,11 @@ impl<F, DB: Database> BasicBlockExecutor<F, DB> {
     }
 
     /// Creates a new `BasicBlockExecutor` with the given strategy and bal if enabled.
-    pub fn new_with_bal(strategy_factory: F, db: DB) -> Self {
+    pub fn new_with_bal(strategy_factory: F, db: DB, has_bal: bool) -> Self {
         let db = State::builder()
             .with_database(db)
             .with_bundle_update()
-            .with_bal_builder_if(true)
+            .with_bal_builder_if(has_bal)
             .build();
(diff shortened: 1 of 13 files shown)


crates/exex/exex/src/backfill/job.rs
```diff
@@ -80,7 +80,7 @@ where
             self.provider
                 .history_by_block_number(self.range.start().saturating_sub(1))
                 .map_err(BlockExecutionError::other)?,
-        ));
+        )); //TODO: change to batch_executor_with_bal when we have access.
```
Rimeeeeee: as of now it is not possible to check for bal presence without fetching block at first

Soubhik-10: Fixed now

crates/evm/evm/src/execute.rs
```diff
@@ -605,13 +591,25 @@ where
             .executor_for_block(&mut self.db, block)
             .map_err(BlockExecutionError::other)?;
 
-        executor.evm_mut().db_mut().bal_state.bal_builder = Some(Bal::new());
+        let has_bal = block.header().block_access_list_hash().is_some();
+
+        if has_bal {
+            executor.evm_mut().db_mut().bal_state.bal_builder = Some(Bal::new());
+        } else {
+            executor.evm_mut().db_mut().bal_state.bal_builder = None;
```
Soubhik-10, line 599: Have tried seperating the executor but thats messy..hopefully this would not impact perf for pre ams blocks

mattsse (approved):

PR #23385, opened 2026-04-08
chore: remove migrate-trie-to-packed example

decofe:
Removes the `migrate-trie-to-packed` example from the workspace. The example is no longer needed.

Prompted by: pep


mattsse (approved):
🫡

PR #23387, opened 2026-04-08
refactor(engine): use Mutex instead of RwLock in PayloadExecutionCache

decofe:
All access paths (`get_cache_for`, `wait_for_availability`, `update_with_guard`) take exclusive (write) locks — there are no read-only consumers. `Mutex` is a more accurate representation of the actual access pattern.

Prompted by: Sergei


shekhirin (approved):
fair

shekhirin (approved):

PR #23393, opened 2026-04-08
fix(engine): disable payload state hook when BAL is present

mediocregopher:
When BAL is present we hash it and send it directly to the SparseTrieCacheTask, sending updates from main execution is redundant.


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -795,7 +800,9 @@ impl<Tx, Err, R: Send + Sync + 'static> PayloadHandle<Tx, Err, R> {
     ///
     /// If a multiproof task is spawned the hook will notify it about new states.
     pub fn state_hook(&self) -> Option<impl OnStateHook> {
```
shekhirin, line 802: can we update the comment here?

shekhirin (approved):

mattsse (approved):

PR #23401, opened 2026-04-08
chore: deny duplicate reqwest versions

decofe:
Adds `reqwest` to the deny list with `deny-multiple-versions = true` so the cargo-deny bans check errors if multiple versions of reqwest are in the dependency graph.

Prompted by: mattsse


mattsse:
Investigated this on the updated branch head.

`cargo deny` is still failing because the resolved graph contains both `reqwest 0.12.28` and `reqwest 0.13.2`.

The remaining `reqwest 0.12.x` comes from `opentelemetry-http -> opentelemetry-otlp -> reth-tracing-otlp`, so this is currently blocked on an `opentelemetry-otlp` release that stops pulling in `reqwest 0.12`.

PR #23409, opened 2026-04-08
refactor: support modifying `next_available_nonce_for`

klkvr:
changes `next_available_nonce_for` to take a transaction request, allowing for shared logic for nonce filling


crates/rpc/rpc-eth-api/src/helpers/state.rs
```diff
@@ -350,14 +351,22 @@ pub trait LoadState:
...
     ) -> impl Future<Output = Result<u64, Self::Error>> + Send
     where
         Self: SpawnBlocking,
     {
+        let address = request.as_ref().from;
         self.spawn_blocking_io(move |this| {
+            let address = match address {
+                Some(address) => address,
+                None => return Err(SignError::NoAccount.into_eth_err()),
+            };
```
mattsse, line 368: we could lift this and check before we spawn, but this seems fine

mattsse (approved):

PR #23413, opened 2026-04-09
chore(ci): add `set -x` to all GHA bash scripts

decofe:
`bench-reth-run.sh` (and other scripts) are called from `bench.yml` workflows, but without `set -x` the logs only show the script output — not which command produced it.

**Example:** [run 24178299041, baseline (1/2)](https://github.com/paradigmxyz/reth/actions/runs/24178299041) fails with:

```
Error:
   0: Volume is not mounted.

Location:
```

The script does `schelk recover`, `schelk mount`, then starts reth — but the log shows none of these commands, so you have to guess which one failed.

**After this PR**, the log would show:

```
+ sudo schelk recover -y --kill
+ sudo schelk mount -y
Error:
   0: Volume is not mounted.
```

Now it is immediately obvious `schelk mount` returned the error.

Prompted by: pep

### DIFF
--- a/.github/scripts/bench-reth-build.sh
+++ b/.github/scripts/bench-reth-build.sh
@@ -16,7 +16,7 @@
 #
 # Required: mc (MinIO client) with a configured alias
 # Optional env: BENCH_BIG_BLOCKS (true/false) — build reth-bb instead of reth
-set -euo pipefail
+set -euxo pipefail
 
 MC="mc"
 MODE="$1"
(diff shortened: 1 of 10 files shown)


laibe (approved):

DaniPopes:
this is extremely spammy. it logs every single bash line executed and every single variable. it creates unreadable outputs sometimes, like for .github/scripts/bench-reth-snapshot.sh which logs the full snapshot archive listing or whatever

PR #23373, opened 2026-04-07
Mediocregopher/big block bals

mediocregopher:

PR #23423, opened 2026-04-09
perf(engine): improve BAL prewarm sparse-trie streaming

mediocregopher:
With BAL enabled we were seeing a [large regression](https://github.com/paradigmxyz/reth/pull/23373#issuecomment-4206833395) when benching against big blocks.

The source of this regression is in how we were dispatching the BAL to the state root. We were previously running through all accounts in the BAL, loading them into the cached provider sequentially, and only then sending the BAL. This took a significant amount of time, and delaying the start of proof work that long caused nP to suffer.

This PR [partially addresses](https://github.com/paradigmxyz/reth/actions/runs/24195654339) the regression, making it into only a slight regression. Further improvements are to follow.

The fix here is to spawn a separate par_iter of BAL accounts which does the following for each account:

* Immediately hash and send all storage slots to the sparse trie task. This is pure CPU, so it can happen immediately and get the proof workers doing something.
* Lazily create provider, if that rayon thread didn't already have one
* Fetch the account from the db, simultaneously populating the state cache with it
* Send the account to the sparse trie task

This is all done in a separate, new rayon thread pool from the storage slot prewarming, otherwise they contend. The new pool uses a new LazyWorkerPool struct which initializes lazily, so we don't create it unless BAL is actually provided.


crates/engine/tree/src/tree/payload_processor/bal.rs
mattsse, line 1: nice

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -352,67 +346,85 @@ where
...
+        let ctx = &self.ctx;
+        let to_sparse_trie_task = self.to_sparse_trie_task.as_ref();
+        let executor = &self.executor;
+        let parent_span = Span::current();
+        let prefetch_parent_span = parent_span.clone();
+        let stream_parent_span = parent_span;
+        let prefetch_bal = Arc::clone(&bal);
+        let stream_bal = Arc::clone(&bal);
+
+        rayon::join(
```
mattsse: I'm not entirely sure about this,

I suppose this is fine because both closures install on the other pool and wouldnt cause the blocking IO issues mentioned in the join docs, but maybe this this also use the `executor.prewarming_pool()` to join?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -593,18 +605,141 @@ where
...
+            }),
+            bytecode_hash: code_hash.or_else(|| {
+                existing_account
+                    .as_ref()
+                    .and_then(|account| account.bytecode_hash)
+                    .or(Some(alloy_consensus::constants::KECCAK_EMPTY))
+            }),
+        };
+
+        let hashed_address = keccak256(address);
```
mattsse: it's likely that we already have this from L623

it should be cached but we could use an Option.unwrap_or_else here I suppose

mattsse (approved):
all of this makes sense

PR #23431, opened 2026-04-10
docs(rpc): mark admin_peerEvents, debug_traceChain, eth_mining, and eth_coinbase as not implemented

yottaes:
Closes #23240

Add `:::warning` callouts to `admin_peerEvents`, `debug_traceChain`, `eth_mining`, and `eth_coinbase` so docs accurately reflect that calling these methods returns an error.


mattsse (commented):
can keep as is

PR #23438, opened 2026-04-10
fix(transaction-pool): reject underpriced replacements with zero priority fee

cuiweixie:
When the pooled EIP-1559 transaction had a non-zero max priority fee, the replacement bump check was skipped if the replacement's priority fee was zero. That allowed replacing with a zero-tip transaction without meeting the configured bump. Remove the guard so the bump rule applies consistently.

### DIFF
--- a/crates/transaction-pool/src/validate/mod.rs
+++ b/crates/transaction-pool/src/validate/mod.rs
@@ -470,7 +470,6 @@ impl<T: PoolTransaction> ValidPoolTransaction<T> {
 
         // Check max priority fee per gas (relevant for EIP-1559 transactions only)
         if existing_max_priority_fee_per_gas != 0 &&
-            replacement_max_priority_fee_per_gas != 0 &&
             replacement_max_priority_fee_per_gas <
                 required_bumped_fee(existing_max_priority_fee_per_gas)
         {


cuiweixie:
reproduce:
```
# start reth dev node
./target/debug/reth node \
  --dev \
  --dev.block-time 10000s  \
  --http \
  --http.api eth,net,web3,txpool
```
```
➜  ~ export RPC=http://127.0.0.1:8545
export MNEMONIC="test test test test test test test test test test test junk"
export PK="$(cast wallet private-key "$MNEMONIC" 0)"
export TO=0x000000000000000000000000000000000000dEaD

# 1) send：tip=10 wei, maxFee=20 gwei, nonce=0
cast send "$TO" \
  --value 1wei \
  --nonce 0 \
  --gas-price 20gwei \
  --priority-gas-price 10wei \
  --private-key "$PK" \
  --rpc-url "$RPC" \
  --async
0x89f42e0af2ddb7486246fc6f1ae02b51a39f6374c35f651e39cfb77fb9670260
# ok
➜  ~ cast send "$TO" \
  --value 1wei \
  --nonce 0 \
  --gas-price 30gwei \
  --priority-gas-price 1 \
  --private-key "$PK" \
  --rpc-url "$RPC" \
  --async
Error: server returned an error response: error code -32000: replacement transaction underpriced
# reject if --priority-gas-price != 0
➜  ~ cast send "$TO" \
  --value 1wei \
  --nonce 0 \
  --gas-price 30gwei \
  --priority-gas-price 0 \
  --private-key "$PK" \
  --rpc-url "$RPC" \
  --async
0x9f2a75cf0c66575c905c3ca631b24b57cdcf1f0b87c198ea8a18a58601e854d1
# accept if --priority-gas-price == 0
➜  ~
```

mattsse (changes requested):
This change fixes the reported zero-tip replacement case, but it appears to regress mixed non-blob replacements. `max_priority_fee_per_gas()` is `None` for legacy and EIP-2930 transactions, so after removing the guard those replacements are treated as having a literal zero priority fee and can be rejected when replacing a tipped EIP-1559/EIP-7702 transaction.

Please either normalize mixed transaction types before this comparison, or explicitly disallow those replacements if that is the intended behavior.

Please also add tests covering the full replacement matrix here: EIP-1559 -> EIP-1559 with zero priority fee, EIP-1559/EIP-7702 -> legacy or EIP-2930, legacy or EIP-2930 -> EIP-1559, and the existing rounding/bump edge cases. That will make it much easier to validate that all replacement scenarios still behave correctly.

PR #23439, opened 2026-04-10
fix(net): correct full-tx propagation peer boundary

cuiweixie:
Fixes an off-by-one when selecting pooled vs full transaction propagation: peers are enumerated from zero, so the first full_peer_count peers should receive full objects. The previous check used `>` and caused one additional peer to receive full broadcasts.


mattsse (commented):
we can keep this as is, this favors full propagation when the peer count is lower, and for higher peercount this doesnt really matter

PR #23443, opened 2026-04-10
fix(provider): size block_range buffer for inclusive span

cuiweixie:
`block_range` builds blocks for a `RangeInclusive<BlockNumber>`. The element count is `end - start + 1`, but the preallocation used `end - start`, shorting `Vec::with_capacity` by one.


mattsse (approved):
Looks good to me.

PR #23444, opened 2026-04-10
fix(net): seed block range on first BlockRangeUpdate

cuiweixie:
Handles `EthMessage::BlockRangeUpdate` when the session has no `range_info` yet. Previously only `BlockRangeInfo::update` ran for `Some`, so the first update was dropped. Seed `BlockRangeInfo::new` from the message when `range_info` is `None`.


mattsse:
Superseded by #23446.

PR #23457, opened 2026-04-11
chore(ci): add set -x to all shell steps in bench workflows

decofe:
Adds `set -x` to every multi-line shell `run:` block in `bench.yml` and `bench-scheduled.yml` for easier debugging of CI failures.

Steps that already had `set -euo pipefail` were updated to `set -euxo pipefail`.

Prompted by: pep

### DIFF
--- a/.github/workflows/bench-scheduled.yml
+++ b/.github/workflows/bench-scheduled.yml
@@ -80,6 +80,7 @@ jobs:
       - name: Detect mode
         id: mode
         run: |
+          set -x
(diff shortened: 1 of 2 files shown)


DaniPopes:
https://github.com/paradigmxyz/reth/pull/23413#issuecomment-4231256973

decofe:
Closing — the set -x approach is too noisy for scripts like bench-reth-snapshot.sh (162K lines of manifest JSON in a 314K line log). Reverting #23413 as well.

PR #23459, opened 2026-04-11
feat: replace `era1` with `er(a)e`

lean-apple:
Toward #23380 

```
cargo nextest run -p reth-era -p reth-era-utils -p reth-era-downloader --release --run-ignored al
```


lean-apple:
We dont need this pr anymore.

PR #23455, opened 2026-04-11
feat(cli): add receipt overrides for download --full

mattsse:
Closes #23452

Adds `--receipts-all`, `--receipts-distance`, and `--receipts-pre-merge` to `reth download` so `--full` can keep more than the default 10,064 receipts window. The existing full preset behavior for other prune segments stays unchanged; only the receipts selection and generated prune config are overridden.

This mirrors the receipt range controls that already exist in the TUI for the non-interactive CLI path, and updates the generated CLI docs. Validated with `cargo test -p reth-cli-commands download:: --lib`, `cargo test -p reth --test it download_help`, and `make update-book-cli`.

Issue #23452, opened 2026-04-10
Keep more receipts with `--full`


yorickdowne:
### Describe the feature

Change the default receipts kept with `--full`. Currently bodies are post-merge, and receipts just the last `10064`.

Current settings
```
[prune.segments.bodies_history]
before = 15537394
[prune.segments.receipts]
distance = 10064
```

Change this to keep all post-merge receipts as well
```
[prune.segments.bodies_history]
before = 15537394
[prune.segments.receipts]
before = 15537394
```

This will allow node operators for RocketPool, SSV, NodeSet/Stakewise, and other protocols that rely on receipts, to use `reth download --full` and receive a snapshot that is useful to them. All of these protocols rely on post-merge receipts being available.

The additional disk space appears a good trade-off for being able to sync quickly, and run protocols that need receipts.

yorickdowne:
After discussion: Full receipts can be chosen from the `reth download` TUI. Having more control via CLI would be extremely helpful, here, as the download is handled by the tooling I maintain, Eth Docker.

yaowubarbara:
I'd like to take this on. Did a quick dig — the fix turns out to be
  surprisingly surgical:

  - `--with-receipts` already exists (`crates/cli/commands/src/download/mod.rs`
    L225-227) and correctly inserts `ComponentSelection::All` for the Receipts
    component (L545-546).
  - The blocker is purely the clap constraint: `with_receipts` is listed in
    `--full`'s `conflicts_with_all` (L250), and vice versa (L227).
  - When `--full` is used, `full_preset_selections` at L656-658 hard-codes
    `defaults.full_prune_modes.receipts` (which resolves to `Distance(10064)`),
    ignoring the explicit flag intent.

  Proposed change (~30 LoC + tests):

  1. Remove `"full"` from `with_receipts`'s `conflicts_with_all` and
     `"with_receipts"` from `full`'s `conflicts_with_all`, so `--full
     --with-receipts` becomes a legal combination.
  2. In `full_preset_selections`, branch the Receipts case on `self.with_receipts`:
     if set, return `ComponentSelection::All`; otherwise fall through to the
     existing `selection_from_prune_mode(...)` path.
  3. Update the `--with-receipts` doc comment to note the combination and the
     Eth Docker / non-interactive operator use case.
  4. Add a unit test alongside the existing ones near L2019/L2040 that builds
     a `DownloadCommand` with both `full` and `with_receipts` set and asserts
[...]

PR #23465, opened 2026-04-12
fix(download): allow --full --with-txs and --with-senders combination

### LINKED ISSUE #23456: feat(cli): Allow `--with-txs` with `download --full`
### Describe the feature

This is a bit of a niche request, and I may turn out what I am looking for can already be done. I am looking for a CLI, programmatic way to get a snapshot that mirrors `--block-interval 5 --prune.senderrecovery.full --prune.accounthistory.distance 10064 --prune.storagehistory.distance 10064 --prune.receipts.before 0` - a true full node, but not an archive node.

With #23455 I could do `reth download --full --receipts-all`, but I'd still be missing pre-merge bodies. I believe `--with-txs` would give me that, but that can't be combined with `--full`.

This is the one node type where Eth Docker cannot programmatically grab a snapshot right now. It's also quite niche: Most users would either choose a history-expired node or an archive node. If this is a non-trivial ask, feel free to say so, and the issue can be closed.

### Additional context

_No response_

yaowubarbara:
Closes #23456

Same constraint pattern as #23452 / #23455: `--with-txs` and `--with-senders` were listed in `--full`'s `conflicts_with_all` (and vice versa), preventing non-interactive use of `reth download --full --with-txs`.

## Changes

All in `crates/cli/commands/src/download/mod.rs`:

1. Remove `"full"` from `--with-txs` `conflicts_with_all`
2. Remove `"full"` from `--with-senders` `conflicts_with_all` (symmetric — senders require txs, so allowing `--full --with-txs` while rejecting `--full --with-txs --with-senders` would be a confusing trap)
3. Remove `"with_txs"` and `"with_senders"` from `--full` `conflicts_with_all`
4. In `full_selection_for_component`, add `if self.with_txs { ComponentSelection::All }` guard before the existing pre-merge heuristic for `Transactions`
5. In `full_selection_for_component`, add `if self.with_senders { ComponentSelection::All }` guard for `TransactionSenders`
6. Update help text for `--with-txs` and `--with-senders`

## Usage

```bash
# Full node + all transaction bodies since genesis (including pre-merge)
reth download --full --with-txs

# Full node + all txs + all sender indices
reth download --full --with-txs --with-senders

# Non-interactive (e.g. Eth Docker Compose)
reth download --full --with-txs --non-interactive
```

### DIFF
--- a/crates/cli/commands/src/download/mod.rs
+++ b/crates/cli/commands/src/download/mod.rs
@@ -218,8 +218,8 @@ pub struct DownloadCommand<C: ChainSpecParser> {
     #[arg(long, value_name = "PATH", conflicts_with_all = ["url", "manifest_url"])]
     manifest_path: Option<PathBuf>,
 
-    /// Include transaction static files.
-    #[arg(long, conflicts_with_all = ["minimal", "full", "archive"])]
+    /// Include transaction static files for all transactions since chain genesis.
+    #[arg(long, conflicts_with_all = ["minimal", "archive"])]
     with_txs: bool,
 
     /// Include receipt static files.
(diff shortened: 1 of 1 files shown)


Rjected:
superseded by https://github.com/paradigmxyz/reth/pull/23513

PR #23469, opened 2026-04-12
fix(init-state): write accounts directly with chunked commits

figtracer:
Closes #23155

Replaces the batch-accumulate approach in `dump_state` that built `BundleStateInit`/`RevertsInit`/`ExecutionOutcome` (3x memory duplication) with direct per-account cursor writes to all required tables. Commits every 500K storage units to bound MDBX dirty page accumulation.

Adds `compute_state_root_chunked` which opens a fresh transaction per progress step, freeing dirty pages between iterations instead of accumulating them in a single long-running transaction.


mattsse:
derek bench

mattsse (approved):

PR #23470, opened 2026-04-13
fix: remove expect() from miner_fee in payload builder

bobsingh-dev:
What does this change do?
Replaces a production expect() call in the payload builder with proper error handling using ok_or_else.

Why is this needed?
Using expect() in production execution paths can cause node panics if the assumption is violated. This change ensures the node fails safely instead of crashing.

What is the impact?

Improves reliability and safety
Prevents potential runtime panics
Aligns with Rust best practices for error handling in critical systems

How was this tested?
Run existing tests with:
cargo test


mattsse (commented):
we can keep the expect here

PR #23476, opened 2026-04-13
fix(txpool): add accept arm for known tx types in match guard

asdv23:
## Summary

- **Fix EIP-1559 (type 2) transactions being incorrectly rejected as "transaction type not supported"** in the transaction pool validator.
- Clippy's `collapsible_match` refactoring (`f0b9cf0b`) converted `if`-inside-`match` to match guards, but didn't account for Rust's guard fall-through semantics. When `eip1559=true`, the guard `!self.eip1559` evaluates to `false`, causing type 2 txs to fall through to the `other_tx_types` catch-all arm and be rejected.
- Adds an explicit accept arm for all known tx types (`LEGACY | EIP2930 | EIP1559 | EIP4844 | EIP7702`) after the rejection guards, so they are accepted when their respective fork is active. This aligns with upstream paradigmxyz/reth commit `3ac5637bd1` (PR #22594).

## Changes

- `crates/transaction-pool/src/validate/eth.rs`: Removed standalone `LEGACY_TX_TYPE_ID` accept arm and added a combined accept arm for all known tx types after the fork-gated rejection guards.

## Test plan

- [x] Release binary built successfully with the fix
- [ ] Verify EIP-1559 (type 2) transactions are accepted when `eip1559` fork is active
- [ ] Verify legacy, EIP-2930, EIP-4844, and EIP-7702 transactions continue to work correctly
- [ ] Verify transactions are still correctly rejected when their fork is not yet active
- [ ] Verify unknown/custom transaction types still route through `other_tx_types` check


asdv23:
Opened against wrong repo by mistake. Re-creating on mantle-xyz/reth.

PR #23486, opened 2026-04-13
chore(BAL): added newPayloadV5 and getPayloadV6

Soubhik-10:
cc @mattsse


crates/rpc/rpc-engine-api/src/engine_api.rs
```diff
@@ -253,6 +253,38 @@ where
...
+        payload: PayloadT::ExecutionData,
+    ) -> EngineApiResult<PayloadStatus> {
+        let payload_or_attrs = PayloadOrAttributes::<
+            '_,
+            PayloadT::ExecutionData,
+            PayloadT::PayloadAttributes,
+        >::from_execution_payload(&payload);
+        self.inner
+            .validator
+            .validate_version_specific_fields(EngineApiMessageVersion::V6, payload_or_attrs)?;
```
mattsse, line 272: this should be V5?
```suggestion
            .validate_version_specific_fields(EngineApiMessageVersion::V5, payload_or_attrs)?;
```

Soubhik-10: not sure, saw that v5 is used in osaka, thats why we used v6 here
https://github.com/paradigmxyz/reth/blob/bce7368a820ae6118e3946a95c9b2b0fb4581e94/crates/payload/primitives/src/lib.rs#L143
a bit confused how the versions are introduced

mattsse (approved):

PR #23489, opened 2026-04-14
fix(rpc): produce complete execution witnesses in debug_executionWitness for historical blocks

daggy1234:
## Summary

Fixes two bugs in `debug_executionWitness` (`crates/rpc/rpc/src/debug.rs`) that caused incomplete execution witnesses for historical blocks, breaking stateless validators like [`paradigmxyz/stateless`](https://github.com/paradigmxyz/stateless).

- **Missing state trie root node**: The root node (whose `keccak256` equals the parent block's `stateRoot`) was not included for historical blocks. Fix: fetch an account proof for `Address::ZERO` after `witness()` and append the root node if absent.
- **Incomplete account witnesses (~47% failure rate)**: Empty prefix sets were passed to `witness()`, allowing the proof calculator to use stale cached hashes instead of expanding touched subtrees. Fix: construct proper prefix sets from `HashedPostState` before calling `witness()`.
- Also restores `ExecutionWitnessMode` passthrough and ancestor header inclusion for `BLOCKHASH` opcodes.

## Testing

Tested against 675 fixtures from an archive node (Reth 2.0.0, mainnet):

| | Before fix | After fix |
|---|---|---|
| Blocks with root present | 87 / 675 (13%) | 675 / 675 (100%) |
| Stateless validation pass | ~53% | 100% |

## Test plan

- [x] Verified `keccak256` of witness state nodes includes parent `stateRoot` for all 675 test blocks
- [x] Stateless validation passes for all 675 blocks post-fix
- [x] Near-tip blocks unaffected (root already present, no-op path)


mediocregopher:
@daggy1234 can you give a bit more context on how you arrived at these fixes? The prefix sets change doesn't look correct to me, and the others I'm having trouble evaluating. How is it possible for a witness to be produced without the state root node, for example?

daggy1234:
Fair! This is what I noticed and sorta why I pushed this

## Prefix sets:  
- The historical provider's `witness()` already calls `input.prepend(self.revert_state()?.into())`, and prepend extends `prefix_sets` via `construct_prefix_sets()`. 
- So the prefix sets aren't "empty" on the historical path. they cover everything modified in `[block_n .. tip]`. What my change does on top of that is add touched keys from the post-state HashedPostState that r missing
- that  that flipped ~47% of fixture blocks from incomplete to complete witnesses, but I don't have a clean explanation for why. there shouldn't be a key in hashed_state whose path isn't already invalidated by the revert state, unless something about how proof_v2 walks cached branches when a target's path matches a non-prefix-set branch is producing an incomplete subtree?

Unsure why

## Missing root: S
-  By the code `(retain_root = targets[0].min_len == 0, defaults to 0 in ProofV2Target::new)`, the root should be retained whenever hashed_state is non-empty, and it is on near-tip blocks.
-  On 87% of historical fixtures it wasn't, and the Address::ZERO proof reliably produced it.  this implies the multiproof is dropping the top-level root in some cases. Possibly related to the prefix set issue above (e.g. cached root branch reused without retaining).


I run an RETH archive node (using latest main commit) and use it to compute.
https://github.com/daggy1234/zkevm-benchmark-workload
Is the package I used (witness-generator-cli)

It is not in live mode

mediocregopher:
> What my change does on top of that is add touched keys from the post-state HashedPostState that r missing
The only thing that this does is that it blocks even more cached trie nodes from the db from being used, the actual proofs being requested/calculated remain the same. So it's just taking longer. I'm not sure why you saw any behavior change from this, but at this point it's drifted too far for me to evaluate (that's my fault for taking so long to get back, sorry). I'm going to close for now, if you can rebase and maybe provide some more specific examples / repro steps for the old/new behavior you're seeing then we can re-evaluate.

daggy1234:
@mediocregopher following up with the repro/analysis you asked for. TL;DR: you
were right that my prefix-set change couldn't change which proofs are
calculated *given a correct walker* — but the walker wasn't correct at the
revision I was testing, and the extra prefix-set coverage was masking that.
Current main no longer reproduces the bug (the path was rewritten in #23667 /
#25646), so instead of reopening this I've submitted regression tests that pin
the fixed behavior: (this PR).

**How a witness can be produced "without the state root node".** It isn't
missing — it's *wrong*. Structural analysis of a broken witness (mainnet block
24618928, 249 blocks below the node tip, generated by 2.0.0-dev): the 14,860
nodes form one fully connected tree under a single un-referenced branch node
`0x9813f414…`, which differs from the true parent root node (`0xcfe8bbae…`) in
exactly 2 of 16 children:

- nibble `7`: resolves to the USDC account leaf carrying a stale storage root
  (itself a mix of rebuilt and stale storage nodes — it matches no block's
  actual USDC storage root);
- nibble `d`: at path `d91` the witness embeds a blind stale hash
  (`0xb8c96401…`) where the parent trie has `0xa527ce6b…`.

Any stale hash poisons every node RLP above it, so the recomputed root hashes
to a value that matches no real state root — which shows up as "no witness
node hashes to the parent stateRoot".

**Why this implicates the walker, not the reverts.** At that revision the
historical witness ran `input.prepend(self.revert_state()?)`, which extends
the prefix sets from the revert state. `AccountChangeSets` for the revert
range `[24618928 ..= tip]` contains 18 accounts whose hashed address starts
with `d91` (e.g. `0x1e92b09b…` → hashed `d91ef992…`, changed at 24619175), so
path `d91` *was* in the prefix sets — and the proof calculator still served
that subtree from a cached branch node. That also explains why my patch
"worked": force-invalidating more paths reduced how often the broken
cached-branch reuse was exercised (~47% → 0% on my fixtures) without fixing
the actual defect. Sorry for the confusion this caused — the patch was
treating a symptom.

**Current status.** Same archive node, same blocks, current main
(476401f02a): witnesses for all previously-failing sampled blocks are
complete (parent root node present; every touched account/storage path
resolvable). Verified with a standalone checker that walks the returned

PR #23496, opened 2026-04-14
feat: add bal validation in post execution

Soubhik-10:


crates/ethereum/consensus/src/validation.rs
mattsse, line 1: can we make this separately, because I think we should make check this earlier

mattsse (changes requested):
lets keep this pr to just amsterdam time and submit the validation part seprately

crates/ethereum/consensus/src/validation.rs
Soubhik-10: we can but if we check before this fn, we will get error of bal mismatch instead of other error, for example in case of a eest test, we should get request hash mismatch but we will get bal mismatch

crates/consensus/consensus/src/lib.rs
```diff
@@ -85,6 +86,7 @@ pub trait FullConsensus<N: NodePrimitives>: Consensus<N::Block> {
         block: &RecoveredBlock<N::Block>,
         result: &BlockExecutionResult<N::Receipt>,
         receipt_root_bloom: Option<ReceiptRootBloom>,
+        block_access_list: Option<BlockAccessList>,
```
mattsse: lets make this accept the hash instead

mattsse (changes requested):

crates/ethereum/consensus/src/validation.rs
```diff
@@ -79,6 +80,20 @@ where
         }
     }
 
+    // Validate that the block access list hash matches the calculated block access list hash
+    if chain_spec.is_amsterdam_active_at_timestamp(block.header().timestamp()) &&
+        block_access_list_hash.is_some()
+    {
```
mattsse, line 87: I think this can be a `&& let Some(bal_hash)`

crates/ethereum/consensus/src/validation.rs
```diff
@@ -21,6 +21,7 @@ pub fn validate_block_post_execution<B, R, ChainSpec>(
     chain_spec: &ChainSpec,
     result: &BlockExecutionResult<R>,
     receipt_root_bloom: Option<(B256, Bloom)>,
+    block_access_list_hash: Option<B256>,
```
mattsse, line 25: lets update the function docs and document what this does

mattsse (changes requested):
more nits

klkvr (approved):

mattsse (approved):

PR #23497, opened 2026-04-14
feat: implement try into v6

Rimeeeeee:


crates/ethereum/engine-primitives/src/payload.rs
```diff
@@ -36,6 +37,8 @@ pub struct EthBuiltPayload<N: NodePrimitives = EthPrimitives> {
     pub(crate) sidecars: BlobSidecars,
     /// The requests of the payload
     pub(crate) requests: Option<Requests>,
+    /// The block access list of the payload
+    pub(crate) block_access_list: Option<BlockAccessList>,
```
mattsse: lets change this to Bytes

crates/ethereum/engine-primitives/src/payload.rs
```diff
@@ -154,7 +158,34 @@ impl EthBuiltPayload {
...
+            BlobSidecars::Eip7594(sidecars) => BlobsBundleV2::from(sidecars),
+            BlobSidecars::Eip4844(_) => {
+                return Err(BuiltPayloadConversionError::UnexpectedEip4844Sidecars)
+            }
+        };
+        Ok(ExecutionPayloadEnvelopeV6 {
+            execution_payload: ExecutionPayloadV4::from_block_unchecked_with_bal(
+                block.hash(),
+                &Arc::unwrap_or_clone(block).into_block(),
+                alloy_rlp::encode(block_access_list.unwrap_or_default()).into(),
```
mattsse: I think we should make this a hard error because we always need a BAL here so if this is None for some reason then the envelope would be invalid

mattsse (changes requested):

mattsse (approved):
lgtm

klkvr (approved):

PR #23498, opened 2026-04-14
[autoopt] 20260414-003-inline-small-payload-tx-root

mediocregopher:
# Inline transaction roots for small payloads
## Evidence
- The baseline engine thread still spends work in payload validation around post-execution helpers after transaction execution completes.
- For both baseline latency runs, 41 of 500 blocks have `transaction_count <= 128`, so a meaningful slice of payloads are small enough that scheduler and clone overhead can dominate the tx-root helper task.
- The payload validator currently clones the full block body and spawns a dedicated `payload-tx-root` blocking task for every payload, then waits on that handle before post-execution validation.

## Hypothesis
If we compute the transaction root inline for small payloads and only spawn the background task for larger bodies, gas throughput improves by ~0.1-0.4% because we avoid extra body cloning and blocking-task wakeups on the small-payload tail.

## Success Metric
- gas_per_second (mgas_s.pct in summary.json) improves by >0.1%

## Plan
- Change `crates/engine/tree/src/tree/payload_validator.rs` so payloads with at most 128 transactions compute `calculate_tx_root()` inline.
- Keep the existing spawned path for larger payloads to preserve concurrency where it matters.

### DIFF
--- a/crates/engine/tree/src/tree/payload_validator.rs
+++ b/crates/engine/tree/src/tree/payload_validator.rs
@@ -599,17 +599,26 @@ where
                 hashed_state_provider.hashed_post_state(&hashed_state_output.state)
             });
 
+        const INLINE_PAYLOAD_TX_ROOT_THRESHOLD: usize = 128;
+
         let block = convert_to_block(input)?;
         let transaction_root = is_payload.then(|| {
-            let body = block.body().clone();
-            let parent_span = Span::current();
-            let num_hash = block.num_hash();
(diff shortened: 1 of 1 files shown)


mediocregopher:
derek bench

decofe:
cc @mediocregopher

❌ Benchmark failed while unknown step. [View logs](https://github.com/paradigmxyz/reth/actions/runs/24405817826/job/71289460042)

PR #23503, opened 2026-04-14
feat(discv4): add enforce_eip868_neighbours config setting

decofe:
Adds `enforce_eip868_neighbours` to `Discv4Config`. When enabled, `Neighbours` responses only include nodes that have a known EIP-868 ENR fork ID. Off by default.

Gated on `enable_eip868` so the filter is a no-op when EIP-868 is disabled. Sends an empty `Neighbours` when all nodes are filtered out to avoid requester timeouts.

Prompted by: mattsse


mattsse (approved):
Looks good. CI is green and the change is default-off, gated on EIP-868, and narrowly scoped to the Neighbours response path.

PR #23507, opened 2026-04-14
[autoopt] 20260414-009-packed-storage-decode-fastpath

mediocregopher:
# Fast-path packed trie exact decodes
## Evidence
- The baseline samply profile shows `<reth_trie_common::storage::PackedStorageTrieEntry as reth_codecs::Compact>::from_compact` with 6,059 exclusive samples and `BranchNodeCompact` decompression on the same storage-trie lookup path.
- The same profile shows `reth_trie::trie_cursor::in_memory::InMemoryTrieCursor<C>::cursor_seek` and exact trie lookups as a hot deferred-trie path, which means small decode work repeats many times during changeset/proof reads.
- `crates/trie/common/src/nibbles.rs` always unpacked packed nibble buffers via the generic path, and `crates/trie/db/src/trie_cursor.rs` rebuilt exact-match keys from decoded DB entries even when the caller had already provided the exact `Nibbles` being sought.

## Hypothesis
If we fast-path packed nibble decoding for common exact-width cases and stop reconstructing exact trie keys after exact cursor hits, gas throughput improves by ~0.2-0.5% because packed storage trie lookups do less redundant decode and key conversion work on the hot exact-seek path.

## Success Metric
- gas_per_second (mgas_s.pct in summary.json) improves by >0.2%

## Plan
- Update `crates/trie/common/src/nibbles.rs` to use a dedicated packed nibble decode helper with zero/full-width fast paths.
- Update `crates/trie/db/src/trie_cursor.rs` so exact account/storage cursor lookups return the requested key directly and keep exact-subkey semantics covered by a regression test.
- Verify with `cargo check -p reth-trie-db` and `cargo test -p reth-trie-db trie_cursor --lib`.

### DIFF
--- a/crates/trie/common/src/nibbles.rs
+++ b/crates/trie/common/src/nibbles.rs
@@ -205,9 +205,7 @@ impl reth_codecs::Compact for PackedStoredNibbles {
     }
 
     fn from_compact(buf: &[u8], _len: usize) -> (Self, &[u8]) {
-        let nibble_count = buf[32] as usize;
-        let packed_len = nibble_count.div_ceil(2);
-        (Self(Nibbles::unpack(&buf[..packed_len]).slice(..nibble_count)), &buf[33..])
+        (Self(unpack_packed_nibbles(buf)), &buf[33..])
     }
 }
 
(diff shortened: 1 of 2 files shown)


mediocregopher:
derek bench

decofe:
cc @mediocregopher

❌ Benchmark failed while unknown step. [View logs](https://github.com/paradigmxyz/reth/actions/runs/24410995873/job/71307487233)

PR #23517, opened 2026-04-15
feat(BAL): use new engine-api methods in bench

Soubhik-10:


bin/reth-bench/src/bench/helpers.rs
```diff
@@ -69,6 +72,47 @@ pub(crate) fn parse_duration(s: &str) -> eyre::Result<Duration> {
...
+) -> eyre::Result<BlockAccessList> {
+    provider
+        .client()
+        .request("eth_getBlockAccessListByBlockNumber", (BlockNumberOrTag::Number(block_number),))
+        .await
+        .map_err(Into::into)
+        .and_then(|block_access_list: Option<BlockAccessList>| {
+            block_access_list.ok_or_else(|| eyre::eyre!("BAL not found for block {block_number}"))
+        })
+}
```
mattsse, line 87: looks like we need to change:

https://github.com/alloy-rs/alloy/blob/773f70a7968d1cd72508e99eab3fe792b464858c/crates/provider/src/provider/trait.rs#L492-L497

lets also add the raw endpoint?

bin/reth-bench/src/bench/new_payload_only.rs
```diff
@@ -123,12 +126,15 @@ impl Command {
 
             debug!(target: "reth-bench", number=?block.header.number, "Sending payload to engine");
 
+            let bal = fetch_block_access_list(&block_provider, block.header.number).await?;
```
mattsse: does this currently always fetch?

dont think this is right

mattsse (changes requested):

bin/reth-bench/src/bench/helpers.rs
```diff
@@ -69,6 +72,47 @@ pub(crate) fn parse_duration(s: &str) -> eyre::Result<Duration> {
...
+
+/// Fetches the block access list for a given block hash using a direct RPC call.
+pub(crate) async fn fetch_block_access_list_with_rpc(
```
mediocregopher: Why is this one necessary?

Soubhik-10: this was the fn we were using in the devnet..however i see its only used here now
https://github.com/paradigmxyz/reth/pull/23517/changes#diff-d8faa813ed88d90dfb1ce8d764912b5570817e5dd8910e8c814724b75f648080R262
probably we dont need it here?

bin/reth-bench/src/bench/send_invalid_payload/mod.rs
```diff
@@ -227,22 +243,29 @@ impl Command {
...
         let parent_beacon_block_root =
             self.parent_beacon_block_root.or(block.header.parent_beacon_block_root);
         let blob_versioned_hashes =
             block.body.blob_versioned_hashes_iter().copied().collect::<Vec<_>>();
         let use_v4 = block.header.requests_hash.is_some();
+        let use_v5 = block.header.block_access_list_hash.is_some();
         let requests_hash = self.requests_hash.or(block.header.requests_hash);
 
-        let mut execution_payload = ExecutionPayload::from_block_slow(&block).0;
+        let mut execution_payload = ExecutionPayload::from_block_slow_with_bal(&block, bal).0;
```
Soubhik-10: here

bin/reth-bench/src/bench/helpers.rs
Soubhik-10: removed that and used the other

bin/reth-bench/src/bench/send_invalid_payload/mod.rs
```diff
@@ -227,22 +247,32 @@ impl Command {
             })?
             .into_consensus();
 
+        let client = ClientBuilder::default()
+            .layer(alloy_transport::layers::RetryBackoffLayer::new(10, 800, u64::MAX))
+            .http(self.rpc_url.clone().unwrap_or_default().parse()?);
+        let provider = RootProvider::<AnyNetwork>::new(client);
+
+        let bal =
+            alloy_rlp::encode(fetch_block_access_list(&provider, block.header.number()).await?);
```
mattsse: this now always fetches the BAL?

this would be redundant pre amsterdam/v4, so we need to make this conditional based on whether the block has a balhash in the header I think

bin/reth-bench/src/valid_payload.rs
```diff
@@ -198,7 +242,11 @@ pub(crate) fn block_to_new_payload(
             tx.try_into().map_err(|_| eyre::eyre!("unsupported tx type"))
         })?
         .into_consensus();
-    let (payload, sidecar) = ExecutionPayload::from_block_slow(&block);
+
+    let block_access_list = alloy_rlp::encode(bal.unwrap_or_default());
+
+    let (payload, sidecar) =
+        ExecutionPayload::from_block_slow_with_bal(&block, block_access_list.into());
```
mattsse, line 249: looks like we need another alternative helper fn for ExecutionPayload::from_block_slow_with_bal that accepts an optional bal?

bin/reth-bench/src/bench/new_payload_fcu.rs
```diff
@@ -198,12 +198,19 @@ impl Command {
                 finalized_block_hash: finalized,
             };
 
+            let bal = if block.header.block_access_list_hash.is_some() {
+                Some(fetch_block_access_list(&block_provider, block.header.number).await?)
+            } else {
+                None
+            };
```
mattsse, line 205: basically this

mattsse (changes requested):

mattsse (approved):
Looks good with the BAL helper follow-up in #23609 accounted for.

PR #23523, opened 2026-04-15
fix(txpool): log decode and insert failures during backup restoration

AJStonewee:
tl;dr: Backup transaction restoration now logs individual decode/recovery failures and reports accurate success/failure counts instead of a blanket "Successfully reinserted" message.

`load_and_reinsert_transactions` was silently discarding transactions that failed `decode_2718_exact`, `try_into_recovered`, or `try_from_consensus` via `filter_map` with `.ok()`. After `join_all`, the log reported `inserted.len()` as a success count without checking each result. The function would then delete the backup file regardless of how many transactions actually made it into the pool.

Each decode/recovery/conversion step now uses `.inspect_err()` to emit a `warn` log on failure and tracks a running count. The final log distinguishes between `succeeded` and `failed` insertions against the `total` from the backup file, downgrading to `warn` when any insertion fails.

### DIFF
--- a/crates/transaction-pool/src/maintain.rs
+++ b/crates/transaction-pool/src/maintain.rs
@@ -713,44 +713,88 @@ where
         return Ok(())
     }
 
-    let pool_transactions: Vec<(TransactionOrigin, <P as TransactionPool>::Transaction)> =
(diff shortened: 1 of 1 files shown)


mattsse (commented):
dont think we need this

PR #23536, opened 2026-04-15
fix(net): encode block access lists as raw BAL RLP

mattsse:
`BlockAccessLists` was encoding inner `Bytes` entries as RLP strings, which broke eth/71 BAL wire semantics and made empty BAL stubs serialize incorrectly. This switches the wrapper to raw BAL RLP encoding/decoding, returns `0xc0` placeholders for unavailable BALs in `reth-network`, and adds roundtrip coverage against real `alloy-eip7928` structures.

This keeps the existing `Vec<Bytes>` representation while matching the EIP-8159/EIP-7928 response shape.

### DIFF
--- a/crates/net/eth-wire-types/src/block_access_lists.rs
+++ b/crates/net/eth-wire-types/src/block_access_lists.rs
@@ -2,7 +2,7 @@
 
 use alloc::vec::Vec;
 use alloy_primitives::{Bytes, B256};
-use alloy_rlp::{RlpDecodableWrapper, RlpEncodableWrapper};
+use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodableWrapper, RlpEncodableWrapper};
 use reth_codecs_derive::add_arbitrary_tests;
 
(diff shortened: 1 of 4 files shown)


crates/net/eth-wire-types/src/block_access_lists.rs
```diff
@@ -16,12 +16,209 @@ pub struct GetBlockAccessLists(
...
+
+        while !payload.is_empty() {
+            let item_start = payload;
+            let item_header = Header::decode(&mut payload)?;
+            if !item_header.list {
+                return Err(alloy_rlp::Error::UnexpectedString)
+            }
+
+            let item_length = item_start.len() - payload.len() + item_header.payload_length;
+            bals.push(Bytes::copy_from_slice(&item_start[..item_length]));
```
klkvr: could do like this
```suggestion
            let item_length = item_header.length_with_payload();
            bals.push(Bytes::copy_from_slice(&item_start[..item_length]));
```

klkvr (approved):

PR #23542, opened 2026-04-15
feat(BAL): add bal in payload-validator

Soubhik-10:


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -983,14 +992,17 @@ where
         debug_span!(target: "engine::tree", "merge_transitions")
             .in_scope(|| db.merge_transitions(BundleRetention::Reverts));
 
+        // Extract the built bal is payload has bal
```
mediocregopher:
```suggestion
        // Extract the built bal if payload has bal
```

mediocregopher (approved):
LGTM, pending @mattsse

PR #23553, opened 2026-04-15
chore: remove BENCH_RUNNERS and queue position tracking from bench workflow

decofe:
Removes the BENCH_RUNNERS env var and all queue position tracking logic (initial calculation + polling step). The queue display was informational only and didn't actually control concurrency — that's governed by runner availability.

Prompted by: alexey


emmajam:
closing as stale

PR #23557, opened 2026-04-15
refactor(execution-cache): wrap SavedCache in Arc<SavedCacheInner>

decofe:
Replace the separate `usage_guard: Arc<()>` field with a newtype `SavedCache(Arc<SavedCacheInner>)`. The Arc's own strong count now serves as the usage guard, removing one level of indirection.

`with_disable_cache_metrics` and `clear_with_hash` now use `Arc::get_mut` which enforces exclusive access at runtime.

Prompted by: Alexey


mattsse (commented):
this introduces a few panics that are hard to reason about or even track if they are safe or not

emmajam:
closing as stale

PR #23569, opened 2026-04-15
fix: validate against executor output gas used

klkvr:
Comparing against last receipt's `cumulative_gas_used` is no longer correct after Amsterdam and block executors are already expected to set `BlockExecutionResult::gas_used` to the expected header's value so we switch the check to rely on this instead


crates/ethereum/consensus/src/validation.rs
```diff
@@ -18,8 +19,7 @@ use reth_primitives_traits::{
 pub fn validate_block_post_execution<B, R, ChainSpec>(
     block: &RecoveredBlock<B>,
     chain_spec: &ChainSpec,
-    receipts: &[R],
-    requests: &Requests,
+    result: &BlockExecutionResult<R>,
```
mattsse, line 22: I think this makes sense, especially since we expect that this will only be used with the block result

mattsse (approved):

PR #23574, opened 2026-04-16
fix(rpc): restore Osaka gas cap for eth_createAccessList

handfuzzer:
Restores the Osaka tx gas cap for `eth_createAccessList`.

When `gas` is omitted, the RPC now chooses a default gas limit that stays within the current tx gas cap while still respecting the block gas limit and caller allowance. This prevents `eth_createAccessList` from accepting requests that exceed the Osaka per-tx cap.

### DIFF
--- a/crates/ethereum/node/tests/e2e/rpc.rs
+++ b/crates/ethereum/node/tests/e2e/rpc.rs
@@ -20,6 +20,7 @@ use reth_node_core::{
 };
 use reth_node_ethereum::EthereumNode;
 use reth_payload_primitives::BuiltPayload;
+use reth_primitives_traits::constants::MAX_TX_GAS_LIMIT_OSAKA;
 use reth_rpc_api::servers::AdminApiServer;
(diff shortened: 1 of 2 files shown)


mattsse (commented):
whats the motivation for this, any more context why we should do this

handfuzzer:
Motivation is to fix the regression introduced in 94d3445.

That change disabled the Osaka tx gas cap for `eth_createAccessList`, so an explicit `gas` above the per-tx cap could still be executed by the RPC. This patch narrows that back down: oversized explicit requests are rejected again, while requests that omit `gas` still get a valid default capped by the active tx gas cap. @mattsse

mattsse (commented):
I think this now differs from geth which skips this check for createaccesslist

https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/transaction_args.go#L447-L496

handfuzzer:
@mattsse You're right, I re-checked geth and this would diverge from it.

PR #23575, opened 2026-04-16
chore(BAL): enable BAL building in evm

Rimeeeeee:


crates/evm/evm/src/either.rs
```diff
@@ -79,4 +79,11 @@ where
             Self::Right(b) => b.size_hint(),
         }
     }
+
+    fn take_bal(&mut self) -> Option<alloy_eips::eip7928::BlockAccessList> {
+        match self {
+            Self::Left(a) => a.take_bal(),
+            Self::Right(b) => b.take_bal(),
+        }
+    }
```
klkvr, line 88: this seems unused?

crates/evm/evm/src/execute.rs
```diff
@@ -554,11 +589,33 @@ where
...
+        }
+
+        for tx in block.transactions_recovered() {
+            executor.execute_transaction(tx)?;
+            if has_bal {
+                executor.evm_mut().db_mut().bump_bal_index();
+            }
+        }
+
+        let result = executor.apply_post_execution_changes()?;
```
klkvr, line 618: why are those changes needed? do we ever need to build a BAL from context of batch executor?

klkvr (commented):
i think i'm okay with the `BasicBlockBuilder` changes if we can scope this PR to just them

crates/evm/evm/src/either.rs
Rimeeeeee: think we need for consume-rlp testsuite,otherwise the index bump not possible
https://github.com/paradigmxyz/reth/blob/d72d87f2c23cc4491eabbbc53a744cd6269de9fa/crates/stages/stages/src/stages/execution/mod.rs#L363

crates/evm/evm/src/execute.rs
Rimeeeeee: think so,
https://github.com/paradigmxyz/reth/blob/bal-devnet-3/crates/stages/stages/src/stages/execution/mod.rs#L355-L363

Soubhik-10: previously we did the bumping in evm, but it was changed to a generic instead of the StateDB type, so we moved this part here

mattsse:
Closing as superseded by #23924.

PR #23590, opened 2026-04-16
feat(engine): add PersistenceGate and expose on RpcHandle

decofe:
Adds `PersistenceGate`, a shared gate that suppresses new persistence cycles during latency-critical windows (e.g. block proposal on fast-block-time chains).

`RpcHandle.persistence_gate.set_enabled(false)` causes `should_persist()` to return `false`, preventing new `SaveBlocks` from being dispatched. In-flight persistence is unaffected. Defaults to `true`.

Also refactors `build_engine_orchestrator` to accept a `PersistenceHandle` directly instead of spawning it internally.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -405,6 +405,9 @@ where
     ///
     /// Returns the sender through which incoming requests can be sent to the task and the receiver
     /// end of a [`EngineApiEvent`] unbounded channel to receive events from the engine.
+    ///
+    /// Persistence can be suppressed at runtime via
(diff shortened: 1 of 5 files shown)


crates/engine/tree/src/tree/mod.rs
```diff
@@ -405,6 +405,9 @@ where
     ///
     /// Returns the sender through which incoming requests can be sent to the task and the receiver
     /// end of a [`EngineApiEvent`] unbounded channel to receive events from the engine.
+    ///
+    /// Persistence can be suppressed at runtime via
+    /// [`PersistenceHandle::set_persistence_enabled`].
```
joshieDo: rm

PR #23596, opened 2026-04-16
feat(bal): scaffold BAL store abstraction

0xKarl98:
This is the first step of the revamp of PR #22952. 

This PR is intentionally limited in scope:
  - adds a `BalStore` abstraction plus a default `NoopBalStore`
  - threads the BAL store through the node/RPC builder chain into `EngineApi`
  - updates BAL-aware engine paths to read/write through the injected store instead of baking a concrete cache into `rpc-engine-api`

  This PR does **not** implement the final BAL storage strategy yet:
  - no static-file BAL persistence
  - no historical BAL backfill
  - no finalized gap/coverage policy

  Follow-up PRs will add concrete implementations on top of this scaffold:
  1. a transient/recent BAL store for hot-path data
  2. a static-file-backed BAL store for canonical persisted data
  3. backfill and explicit gap/coverage handling for older nodes
  
  cc @mattsse

### DIFF
--- a/crates/node/builder/src/rpc.rs
+++ b/crates/node/builder/src/rpc.rs
@@ -32,6 +32,7 @@ use reth_node_core::{
     version::{version_metadata, CLIENT_CODE},
 };
 use reth_payload_builder::{PayloadBuilderHandle, PayloadStore};
+use reth_provider::BalStoreHandle;
 use reth_rpc::{
     eth::{core::EthRpcConverterFor, DevSigner, EthApiTypes, FullEthApiServer},
     AdminApi,
@@ -44,7 +45,6 @@ use reth_rpc_builder::{
 };
 use reth_rpc_engine_api::{capabilities::EngineCapabilities, EngineApi};
 use reth_rpc_eth_types::{cache::cache_new_blocks_task, EthConfig, EthStateCache};
-use reth_storage_api::BalStoreHandle;
 use reth_tokio_util::EventSender;
 use reth_tracing::tracing::{debug, info};
 use std::{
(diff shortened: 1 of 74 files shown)


crates/storage/storage-api/src/bal.rs
mattsse, line 1: I like this, this is a good start

mattsse (changes requested):
cool, great start.

I'd like to change one thing architecturally, closer to what 

https://github.com/paradigmxyz/reth/pull/22952

did and move the store into the `BlockChainProvider` itself, and leave the engine api untouched.
this is because we also want bals on non engine api nodes:


https://github.com/paradigmxyz/reth/pull/22952/changes#diff-3e907e0480039111c3e6da6ce067cc7512c8086d3f1bc0879c09a645fc87918cR54-R55

crates/node/builder/src/rpc.rs
mattsse, line 1: I think we can undo these

crates/rpc/rpc-engine-api/src/engine_api.rs
mattsse, line 1: lets also keep this unchanged because we should handle all things bal in the tree itself, so that we can also support non engine api paths properly

mattsse (changes requested):

crates/storage/storage-api/src/bal.rs
```diff
@@ -0,0 +1,130 @@
+use alloc::{sync::Arc, vec::Vec};
+use alloy_primitives::{BlockHash, BlockNumber, Bytes};
+use reth_storage_errors::provider::ProviderResult;
+
```
mattsse, line 4: I addition, can we introduce a config type:

```
struct BalStoreConfig {
  pub prune_mode: PruneMode
}

```

klkvr (approved):

mediocregopher (approved):

mattsse (approved):

PR #23597, opened 2026-04-16
chore(BAL): enabled bal building in ethereum payload

Rimeeeeee:


crates/ethereum/payload/src/lib.rs
```diff
@@ -178,8 +178,13 @@ where
...
     }
     let state = StateProviderDatabase::new(state_provider.as_ref());
-    let mut db =
-        State::builder().with_database(cached_reads.as_db_mut(state)).with_bundle_update().build();
+    let chain_spec = client.chain_spec();
+    let is_amsterdam = chain_spec.is_amsterdam_active_at_timestamp(attributes.timestamp());
+    let mut db = State::builder()
+        .with_database(cached_reads.as_db_mut(state))
+        .with_bundle_update()
+        .with_bal_builder_if(is_amsterdam)
```
mattsse, line 186: I think we now also need to use that then?

here:

https://github.com/paradigmxyz/reth/blob/8c826a5cd0fe655a183c13203fc8169835713bed/crates/ethereum/payload/src/lib.rs#L437-L437

mattsse (changes requested):

mattsse (approved):
we def need this

we can take care of extracting the bal from the state next

PR #23600, opened 2026-04-16
fix(rpc): reject trace_filter ranges in pruned history

AKABABA-ETH:
Was chasing a user complaint about `trace_filter` returning a misleading `HeaderNotFound` on an archive node that had EIP-4444 pruning on. Pulled the RPC logs and saw:              
                                                                                             
      `2026-04-16T18:22:41.702Z  WARN reth::rpc: trace_filter error=HeaderNotFound(Number(14500100)) from=14000000 to=15000000                           `
                                                                                             
`from_block` sat below the node's `earliest_block_number`, so the scan tripped on a pruned header partway through instead of rejecting the request up front. Added an `earliest_block_number` check alongside the existing `best_block_number` one so the endpoint returns `PrunedHistoryUnavailable` (JSON-RPC code 4444) when the range dips into pruned state.


crates/rpc/rpc/src/trace.rs
```diff
@@ -360,6 +360,13 @@ where
             return Err(EthApiError::HeaderNotFound(end.into()).into());
         }
 
+        // Check if the requested range overlaps with pruned history (EIP-4444)
+        let earliest_block =
+            self.provider().earliest_block_number().map_err(Eth::Error::from_eth_err)?;
+        if start < earliest_block {
+            return Err(EthApiError::PrunedHistoryUnavailable.into());
+        }
```
mattsse, line 368: oh I see, the `recovered_block_range` was not respecting the earliest block number,

adding this here is a nicer and cheaper way to check against this, ty!

mattsse (approved):

PR #23601, opened 2026-04-16
fix(rpc): don't cap eth_simulateV1 blockOverrides.gasLimit by --rpc.gascap

kustrun:
Remove the pre-check that rejected `blockOverrides.gasLimit` above `max(base_block_gas_limit, --rpc.gascap)` in `eth_simulateV1`, matching geth/op-geth which apply no upfront ceiling on `BlockOverrides.GasLimit`.

`eth_simulateV1` exists to model hypothetical blocks, so forcing operators to raise `--rpc.gascap` globally just to simulate larger blocks defeats the override's purpose. `--rpc.gascap` still bounds defaulted per-call gas in `simulate_v1` and still applies to `eth_call` / `eth_estimateGas` via their own code paths; the existing sum check still rejects per-call gas sums that exceed the (possibly-overridden) block gas limit.

An e2e regression test (`crates/ethereum/node/tests/e2e/simulate.rs::test_simulate_v1_block_gas_limit_override_above_rpc_gascap`) spins up a node with defaults, submits `eth_simulateV1` with `blockOverrides.gasLimit = 100M` (above the default 50M `--rpc.gascap` and the 30M genesis block gas), and asserts the simulation succeeds.

## Reproduction against a live node

Tested against a reth node on Ethereum mainnet running default `--rpc.gascap = 50_000_000`. Both payloads are identical except for `blockOverrides.gasLimit`. The call is a USDC `transfer` from an EOA holding no USDC, which reverts with `ERC20: transfer amount exceeds balance` (actual gas used ~36K) — the revert itself is expected and only demonstrates that the handler actually executed.

`sim_low.json` — `blockOverrides.gasLimit = 0xF4240` (1,000,000). Succeeds both before and after this change.

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_simulateV1",
  "params": [
    {
      "blockStateCalls": [
        {
          "blockOverrides": { "gasLimit": "0xF4240" },
          "calls": [
            {
              "from": "0x0000000000000000000000000000000000001234",
              "to":   "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
              "data": "0xa9059cbb000000000000000000000000111111111111111111111111111111111111111100000000000000000000000000000000000000000000000000000000000F4240",
              "gas":  "0x7A120"
            }
          ]
        }
      ],
      "validation": false,
      "traceTransfers": false
    },
    "latest"
  ]
}
```

`sim_high.json` — identical payload, only `blockOverrides.gasLimit` changed to `0x5F5E100` (100,000,000). On `main` returns `{"code":-38026,"message":"Client adjustable limit reached"}`; on this branch returns a normal simulation result.

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "eth_simulateV1",
  "params": [
    {
      "blockStateCalls": [
        {
          "blockOverrides": { "gasLimit": "0x5F5E100" },
          "calls": [
            {
              "from": "0x0000000000000000000000000000000000001234",
              "to":   "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
              "data": "0xa9059cbb000000000000000000000000111111111111111111111111111111111111111100000000000000000000000000000000000000000000000000000000000F4240",
              "gas":  "0x7A120"
            }
          ]
        }
      ],
      "validation": false,
      "traceTransfers": false
    },
    "latest"
  ]
}
```

```bash
curl -sS -X POST "$RPC_URL" -H 'Content-Type: application/json' -d @sim_low.json
curl -sS -X POST "$RPC_URL" -H 'Content-Type: application/json' -d @sim_high.json
```

### DIFF
--- a/crates/ethereum/node/tests/e2e/simulate.rs
+++ b/crates/ethereum/node/tests/e2e/simulate.rs
@@ -148,6 +148,92 @@ async fn test_simulate_v1_block_gas_limit_override_above_rpc_gascap() -> eyre::R
 }
 
 #[tokio::test]
+async fn test_simulate_v1_explicit_call_gas_clamped_to_rpc_gascap() -> eyre::Result<()> {
+    use alloy_consensus::Transaction as _;
+
+    reth_tracing::init_test_tracing();
+
+    let chain_spec = Arc::new(
+        ChainSpecBuilder::default()
+            .chain(MAINNET.chain)
+            .genesis(serde_json::from_str(include_str!("../assets/genesis.json")).unwrap())
+            .cancun_activated()
+            .build(),
+    );
+
+    let (mut nodes, wallet) = setup_engine::<EthereumNode>(
+        1,
+        chain_spec,
+        false,
+        Default::default(),
+        eth_payload_attributes,
+    )
+    .await?;
+    let node = nodes.pop().unwrap();
+    let provider = ProviderBuilder::new()
+        .wallet(EthereumWallet::new(wallet.wallet_gen().swap_remove(0)))
+        .connect_http(node.rpc_url());
+
+    // Default --rpc.gascap is 50M.
+    let rpc_gascap_default = 50_000_000u64;
+    // Block-level override above rpc.gascap — allowed (matches the first fix).
+    let block_gas_limit = 100_000_000u64;
+    // Per-call explicit gas above rpc.gascap — op-geth clamps this to rpc.gascap.
+    let explicit_call_gas = 80_000_000u64;
+
+    let from: Address = "0xc000000000000000000000000000000000000000".parse()?;
+    let to: Address = "0xc100000000000000000000000000000000000000".parse()?;
+
+    let tx = TransactionRequest::default()
+        .from(from)
+        .to(to)
+        .gas_limit(explicit_call_gas)
+        .max_fee_per_gas(0)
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
Allowing `blockOverrides.gasLimit` above `--rpc.gascap` is directionally right and matches geth/op-geth at the block level. The issue is that geth still applies `RPCGasCap()` to each simulated call via `sanitizeCall`, including calls that explicitly set `gas`, while this patch removes the upfront rejection without adding the per-call clamp.

That means a request can now raise `blockOverrides.gasLimit` and also set `calls[i].gas` above `--rpc.gascap`, which widens the RPC DoS envelope instead of mirroring geth. Please keep the block gas limit override allowed, but add a separate simulate gas budget / per-call cap for individual calls so explicit `gas` values are still bounded by `--rpc.gascap`. A regression test should cover `blockOverrides.gasLimit > rpc.gascap` together with `call.gas > rpc.gascap`, since the current test only exercises a small explicit gas value.

kustrun:
@mattsse Thank you for taking the time to review this. The improvements you recommended are now implemented..

mattsse (commented):
superseded by #24387

PR #8590, opened 2024-06-04
chore: add `EnvironmentArgs` and use across all CLI commands  

joshieDo:
`--chain`, `--config` and `--datadir` are all used to initiate an environment that a CLI command will use (config, providers etc).

However, there's a lot of repetition with some commands missing some relevant flags. This PR attempts to fix that and dedup by adding a `EnvironmentArgs` struct that can initialize all the necessary context.

Feel free to suggest other names other than Environment. `CliContext` was taken.

Follow-up will be a version of https://github.com/paradigmxyz/reth/pull/8553 , and probably moving `AccessRights` to reth_primitives since there's different versions of the same enum across crates.

PR #23602, opened 2026-04-16
fix(cli): open stage dump environment read-write

joohhnnn:
While trying to isolate a regression I ran `reth stage dump merkle` on `main` and it fails immediately with `the environment opened in read-only` — the top-level command opens the source datadir RO, but every subcommand's `unwind_and_copy` asks for a RW provider on the same env (broken since #8590).

Open RW so the subcommands run, and add a top-level note that the command does mutate the source datadir.


mattsse (approved):
Looks correct.  opens the environment read-only today, but every current dump subcommand ends up requesting a writable source provider during unwind/copy, so opening the env read-write at the top level is the right fix.

PR #23611, opened 2026-04-17
feat(net): add snap/2 wire helpers and messages

mattsse:
Adds preparatory snap/2 wire support without changing negotiation or sync behavior yet.

This introduces `SnapVersion`, snap/2 BAL messages, `Capability::snap_*` and `Protocol::snap_*` helpers, and lets `EthSnapStream` carry a snap version while defaulting existing behavior to snap/1.

This is groundwork for EIP-8189 BAL-based healing.

Checks:
- cargo +nightly fmt --all
- cargo test -p reth-eth-wire-types snap --lib
- cargo test -p reth-eth-wire eth_snap_stream --lib

### DIFF
--- a/crates/net/eth-wire-types/src/snap.rs
+++ b/crates/net/eth-wire-types/src/snap.rs
@@ -3,12 +3,40 @@
 //! facilitating the exchange of Ethereum state snapshots between peers
 //! Reference: [Ethereum Snapshot Protocol](https://github.com/ethereum/devp2p/blob/master/caps/snap.md#protocol-messages)
 //!
-//! Current version: snap/1
+//! This module currently includes snap/1 plus preparatory snap/2 message definitions.
 
+use crate::BlockAccessLists;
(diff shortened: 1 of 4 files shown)


crates/net/eth-wire-types/src/snap.rs
```diff
@@ -3,13 +3,41 @@
...
+    /// Returns the number of messages supported by this version.
+    pub const fn message_count(self) -> u8 {
+        match self {
+            Self::V1 => 8,
+            Self::V2 => 10,
+        }
+    }
+
+    /// Returns the highest supported message id for this version.
+    pub const fn max_message_id(self) -> u8 {
```
emmajam, line 36: is this used anywhere @mattsse ?

mattsse: not yet, but we need this value if this is combined with other subprotocols

PR #23613, opened 2026-04-17
feat: support binding discv5 and discv4 to the same port

klkvr:
depeds on https://github.com/sigp/discv5/pull/299

Adds support for binding discv4 and discv5 on the same port by using a shared socket and forwarding `UnrecognizedPacket` messages to discv4


mattsse (approved):

emmajam:
🙏

PR #23616, opened 2026-04-17
feat(provider): add storage range provider

figtracer:
Adds a minimal storage-range provider primitive for historical storage_v2 state.

It pages by hashed storage key and returns plain slot preimages plus values, which is the provider-side building block for `debug_storageRangeAt`.

Tested with `cargo test -p reth-provider history_provider_storage_range_hashed_state`.


crates/storage/provider/src/providers/state/historical.rs
```diff
@@ -587,6 +601,76 @@ impl<
...
+        }
+
+        let visible_tip = self.provider.best_block_number()?;
+        let mut slots_by_hash = BTreeMap::new();
+
+        self.provider.with_rocksdb_snapshot(|rocksdb_ref| {
+            let rocksdb = rocksdb_ref
+                .expect("storage_v2 historical range queries require a RocksDB snapshot");
+
+            rocksdb.walk_visible_storage_history_keys(account, B256::ZERO, visible_tip, |key| {
```
shekhirin: should we pass `start_key` here? Otherwise we would collect all slots first, and only then filter?

shekhirin: ah no, we can't easily, because `startKey` is a hashed storage slot key. Do we need `start_key` argument for `walk_visible_storage_history_keys` then?

figtracer: on it, we don't need it here

PR #23617, opened 2026-04-17
perf(re-execute): relax executor reset thresholds

emmajam:
Increase state cache size limit from 1M to 5M entries and executor lifetime from 120s to 600s. Preserves hot state caches longer during re-execute, reducing redundant DB lookups for frequently accessed accounts and storage slots.

## Benchmark (dev-emma, 3 runs, blocks 21M–21.001M)

| | Baseline (main) | Feature | 
|---|---|---|
| Run 1 | 0.478 Ggas/s | 0.520 Ggas/s |
| Run 2 | 0.486 Ggas/s | 0.520 Ggas/s |
| Run 3 | 0.480 Ggas/s | 0.512 Ggas/s |
| **Mean** | **0.481 Ggas/s** | **0.517 Ggas/s** |

**+7.5% ±1.5% (sig=good)** — statistically significant at 95% CI via Welch's t-test.

## Changes
- `crates/cli/commands/src/re_execute.rs`: `size_hint` threshold 1M → 5M, `executor_lifetime` 120s → 600s

### DIFF
--- a/crates/cli/commands/src/re_execute.rs
+++ b/crates/cli/commands/src/re_execute.rs
@@ -138,7 +138,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
             let cancellation = cancellation.clone();
             let next_block = Arc::clone(&next_block);
             tasks.spawn_blocking(move || {
-                let executor_lifetime = Duration::from_secs(120);
+                let executor_lifetime = Duration::from_secs(600);
 
                 loop {
                     if cancellation.is_cancelled() {
@@ -245,7 +245,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
                         let _ = stats_tx.send(block.gas_used());
 
                         // Reset DB once in a while to avoid OOM or read tx timeouts
-                        if executor.size_hint() > 1_000_000 ||
+                        if executor.size_hint() > 5_000_000 ||
                             executor_created.elapsed() > executor_lifetime
                         {
                             executor =


emmajam:
## 50K block benchmark (dev-emma, 3 runs, blocks 21M–21.05M)

| | Baseline (main) | Feature |
|---|---|---|
| Run 1 | 2.39 Ggas/s | 2.73 Ggas/s |
| Run 2 | 2.49 Ggas/s | 2.80 Ggas/s |
| Run 3 | 2.50 Ggas/s | 2.82 Ggas/s |
| **Mean** | **2.46 Ggas/s** | **2.78 Ggas/s** |

**+13.1% ±3.5% (sig=good)** — stronger effect at scale since more executor resets are avoided over 50K blocks.

crates/cli/commands/src/re_execute.rs
```diff
@@ -138,7 +138,7 @@ impl<C: ChainSpecParser<ChainSpec: EthChainSpec + Hardforks + EthereumHardforks>
             let cancellation = cancellation.clone();
             let next_block = Arc::clone(&next_block);
             tasks.spawn_blocking(move || {
-                let executor_lifetime = Duration::from_secs(120);
+                let executor_lifetime = Duration::from_secs(600);
```
mattsse, line 141: this exceeds the 5min threshold we usually have but in this case it doesnt really matter and I believe we'll always encounter the size hint first

mattsse (approved):
seems fine

crates/cli/commands/src/re_execute.rs
emmajam: just benching the change on tempo also

emmajam:
## Tempo benchmark — no improvement

Tested the same change on Tempo chain data (21 GB, blocks 14M–15M, dev-emma).

**1K blocks (15M–15.001M):**
| | Baseline | Feature |
|---|---|---|
| Mean | 2.26 Ggas/s | 2.31 Ggas/s |

**1M blocks (14M–15M):**
| | Baseline | Feature |
|---|---|---|
| Run 1 | 86.62 Ggas/s | 87.80 Ggas/s |
| Run 2 | 88.28 Ggas/s | 87.54 Ggas/s |
| Run 3 | 88.56 Ggas/s | 87.61 Ggas/s |
| Mean | 87.82 Ggas/s | 87.65 Ggas/s |

No meaningful difference on Tempo. The chain is much smaller (21 GB vs 2.5 TB for Ethereum mainnet) so state fits in cache easily and executor resets are essentially free — the bottleneck is elsewhere.

PR #23618, opened 2026-04-17
feat(engine): suppress persistence during payload building

joshieDo:
Alternative to #23590 (`PersistenceGate` on `RpcHandle`).

Adds `--engine.suppress-persistence-during-build` flag (default: `false`).

When enabled, `should_persist()` returns `false` while a payload is being built — from the moment an FCU with payload attributes arrives until the next FCU clears the build. This avoids persistence I/O competing with block building on latency-sensitive chains.

### DIFF
--- a/crates/engine/tree/src/tree/mod.rs
+++ b/crates/engine/tree/src/tree/mod.rs
@@ -305,6 +305,9 @@ where
     /// Stored here (not in `ExecutedBlock`) to avoid leaking observability concerns into the block
     /// type. Entries are removed when blocks are persisted or invalidated.
     execution_timing_stats: HashMap<B256, Box<ExecutionTimingStats>>,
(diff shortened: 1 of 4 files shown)


klkvr (approved):

crates/engine/tree/src/tree/mod.rs
```diff
@@ -1112,6 +1116,8 @@ where
     ) -> ProviderResult<TreeOutcome<OnForkChoiceUpdated>> {
         trace!(target: "engine::tree", ?attrs, "invoked forkchoice update");
 
+        self.building_payload = attrs.is_some() && self.config.suppress_persistence_during_build();
```
mattsse, line 1119: this should do it

mattsse (approved):

shekhirin (approved):
Clean.

PR #23619, opened 2026-04-17
fix: don't cache stateful precompiles

klkvr:
Precompiles that don't access/create state would still return `reservoir` value which would be exactly the same as input value (because no state gas was consumed) 

Right now we will return cached vaue for `reservoir` while actual input might be different. We need to override the cached value with the one that was in the input

### DIFF
--- a/crates/engine/tree/src/tree/precompile_cache.rs
+++ b/crates/engine/tree/src/tree/precompile_cache.rs
@@ -10,6 +10,7 @@ use reth_primitives_traits::dashmap::DashMap;
 use revm::precompile::{PrecompileId, PrecompileOutput, PrecompileResult};
 use revm_primitives::Address;
 use std::{hash::Hash, sync::Arc};
+use tracing::error;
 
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/precompile_cache.rs
```diff
@@ -175,22 +182,29 @@ where
...
             // Only successful outputs are cacheable. Non-success statuses and errors must execute
             // again instead of poisoning the cache for subsequent calls.
             Ok(output) if output.is_success() => {
-                let size = self.cache.insert(
-                    Bytes::copy_from_slice(calldata),
-                    CacheEntry { output: output.clone(), spec: self.spec_id.clone() },
-                );
-                self.set_precompile_cache_size_metric(size as f64);
-                self.increment_by_one_precompile_cache_misses();
+                if output.reservoir != reservoir {
```
mattsse, line 201: can we document the logic here, I find this hard to follow/reason about what this even checks

mattsse (approved):

PR #23621, opened 2026-04-18
perf(txpool): replace BTreeMap with imbl::OrdMap in BestTransactions

decofe:
Replaces `BTreeMap` with [`imbl::OrdMap`](https://docs.rs/imbl/latest/imbl/) for the `all` map in `BestTransactions` and `by_id` in `PendingPool`.

`OrdMap` is a persistent/immutable B-tree that provides O(1) clone via structural sharing. Since `PendingPool::best()` clones the entire `by_id` map to create a `BestTransactions` snapshot, this is a significant win for large pools — the clone goes from O(n) to O(1).

Prompted by: DaniPopes

### DIFF
--- a/crates/transaction-pool/src/pool/pending.rs
+++ b/crates/transaction-pool/src/pool/pending.rs
@@ -8,13 +8,9 @@ use crate::{
     },
     Priority, SubPoolLimit, TransactionOrdering, ValidPoolTransaction,
 };
+use imbl::OrdMap;
 use rustc_hash::{FxHashMap, FxHashSet};
-use std::{
-    cmp::Ordering,
-    collections::{hash_map::Entry, BTreeMap},
-    ops::Bound::Unbounded,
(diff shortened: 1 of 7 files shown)


crates/transaction-pool/src/pool/pending.rs
```diff
@@ -525,7 +521,7 @@ impl<T: TransactionOrdering> PendingPool<T> {
     }
 
     /// All transactions grouped by id
-    pub const fn by_id(&self) -> &BTreeMap<TransactionId, PendingTransaction<T>> {
+    pub const fn by_id(&self) -> &OrdMap<TransactionId, PendingTransaction<T>> {
```
mattsse, line 524: since this changes public api

could we maybe re-export the type from the crate?

mattsse (approved):

onbjerg (approved):

PR #23623, opened 2026-04-18
chore(BAL): added bal boundary checking

Rimeeeeee:


Soubhik-10:
lgtm! we need to follow this up in stages so that we can reject invalid blocks there too

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -2142,4 +2160,15 @@ impl<T: PayloadTypes> BlockOrPayload<T> {
...
+    /// Returns the gas limit used by the block.
+    pub fn gas_limit(&self) -> u64
+    where
+        T::ExecutionData: ExecutionPayload,
+    {
+        match self {
+            Self::Payload(payload) => payload.gas_limit(),
+            Self::Block(block) => block.gas_limit(),
+        }
+    }
```
mattsse, line 2173: lets submit this separately first

Rimeeeeee: #23624

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -907,6 +907,24 @@ where
...
+
+            let bal_items = total_bal_items(&bal);
+
+            if bal_items > input.gas_limit() / ITEM_COST as u64 {
+                debug!(target: "engine::tree::payload_validator", bal_items, "{} {}", input.gas_limit(), "BAL is invalid since it contains more items than the gas limit allows");
+                return Err(InsertBlockErrorKind::Consensus(
+                    ConsensusError::BlockAccessListCostMoreThanGasLimit,
+                ));
+            }
+        }
```
mattsse: could we make this a standalone fn instead, in consensus instead, then we can keep the diff here smaller

mattsse (changes requested):

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -907,6 +909,15 @@ where
...
         debug!(target: "engine::tree::payload_validator", "Executing block");
 
+        if let Some(bal_opt) = input.block_access_list() {
+            let bal = bal_opt.map_err(BlockExecutionError::other)?;
+            validate_block_access_list_gas(Some(&bal), input.gas_limit())
+                .map_err(|e| {
+                    debug!(target: "engine::tree::payload_validator", "BAL is invalid since it contains more items than the gas limit allows");
+                    InsertBlockErrorKind::Consensus(e)
+                })?
+        }
```
mattsse, line 919: I think we should move this to the `ExecutionEnv` type and give it the decoded bal, then we avoid having to decode it multiple times potentially

crates/consensus/consensus/src/lib.rs
```diff
@@ -471,6 +474,23 @@ impl ConsensusError {
...
+    gas_limit: u64,
+) -> Result<(), ConsensusError> {
+    if let Some(bal) = block_access_list {
+        let bal_items = alloy_eip7928::total_bal_items(bal);
+        if bal_items > gas_limit / alloy_eip7928::ITEM_COST as u64 {
+            return Err(ConsensusError::BlockAccessListCostMoreThanGasLimit)
+        }
+    }
+    Ok(())
+}
```
mattsse, line 540: I think we could submit this separtely first?

mattsse (changes requested):

mattsse:
Closing as superseded by #23924.

PR #23629, opened 2026-04-19
feat(p2p): optionally fetch BAL with full blocks

### LINKED ISSUE #23433: Add support for downloading BAL in FullBlockClient
### Describe the feature

with glam we'll get eth71 that provides the BAL for recent blocks

ideally we integrate this as additional functionality next to 

https://github.com/paradigmxyz/reth/blob/e4cf0ab09be36c1378fe7724b2b3fe6a2f8c52e3/crates/net/p2p/src/full_block.rs#L66-L66

now that we have 

https://github.com/paradigmxyz/reth/blob/608c96791fe0746827bb5fb645aa504ff08586e6/crates/net/p2p/src/block_access_lists/client.rs#L9-L9

this could be helpful speeding up the initial sync of the most recent ~2epoch blocks to tip

for now we can simply add the functionality to maybe optionally request the BAL via the FullBlockClient

@lean-apple 

### Additional context

_No response_

0xKarl98:
Closes #23433 , and optimize upon  #23435 
## Summary 
- Adds FullBlockClient::get_full_block_with_access_lists for clients that also implement BlockAccessListsClient. The new path reuses the existing full-block downloader and requests BAL in parallel, while leaving get_full_block unchanged.
- Includes tests for the success path and retry handling on invalid BAL responses.
## Follow-up 
- higher-level call sites can opt into this API when they want BAL-backed full-block fetches. This PR only adds the
  capability and keeps the policy at the call site.
cc @mattsse

### DIFF
--- a/crates/net/p2p/src/full_block.rs
+++ b/crates/net/p2p/src/full_block.rs
@@ -1,5 +1,6 @@
 use super::headers::client::HeadersRequest;
 use crate::{
+    block_access_lists::client::BlockAccessListsClient,
     bodies::client::{BodiesClient, SingleBodyRequest},
     download::DownloadClient,
     error::PeerRequestResult,
@@ -10,15 +11,17 @@ use crate::{
 use alloy_consensus::BlockHeader;
 use alloy_primitives::{Sealable, B256};
 use core::marker::PhantomData;
+use futures::FutureExt;
 use reth_consensus::Consensus;
-use reth_eth_wire_types::{EthNetworkPrimitives, HeadersDirection, NetworkPrimitives};
+use reth_eth_wire_types::{
+    BlockAccessLists, EthNetworkPrimitives, HeadersDirection, NetworkPrimitives,
+};
 use reth_network_peers::{PeerId, WithPeerId};
 use reth_primitives_traits::{SealedBlock, SealedHeader};
 use std::{
     cmp::Reverse,
     collections::{HashMap, VecDeque},
     fmt::Debug,
-    future::Future,
     hash::Hash,
     ops::RangeInclusive,
     pin::Pin,
@@ -26,6 +29,9 @@ use std::{
     task::{ready, Context, Poll},
 };
 use tracing::debug;
+
+// RLP encoding for an empty list.
+const EMPTY_LIST_CODE: u8 = 0xc0;
 
 /// A Client that can fetch full blocks from the network.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


crates/net/p2p/src/full_block.rs
```diff
@@ -251,6 +287,142 @@ where
...
+        let hash = *self.block.hash();
+        self.access_lists = Some(self.block.client.get_block_access_lists(vec![hash]));
+    }
+
+    fn poll_access_lists(&mut self, cx: &mut Context<'_>) {
+        if self.access_lists_result.is_some() {
+            return
+        }
+
+        let Some(fut) = self.access_lists.as_mut() else { return };
```
mattsse: this looks a bit complex, 

unsure we need all of this?

crates/net/p2p/src/full_block.rs
```diff
@@ -251,6 +287,142 @@ where
...
+                            debug!(
+                                target: "downloaders",
+                                hash = ?self.block.hash(),
+                                expected = 1,
+                                received = access_lists.0.len(),
+                                "Received wrong access list response",
+                            );
+                            self.block.client.report_bad_message(peer);
+                            self.retry_access_lists_request();
+                            cx.waker().wake_by_ref();
```
mattsse: this usually indicates something wrong

and we likely need to wrap this poll into a loop {} instead

crates/net/p2p/src/full_block.rs
```diff
@@ -251,6 +287,142 @@ where
...
+/// header/body logic stays centralized.
+#[must_use = "futures do nothing unless polled"]
+pub struct FetchFullBlockWithAccessListsFuture<Client>
+where
+    Client: BlockClient + BlockAccessListsClient,
+{
+    block: FetchFullBlockFuture<Client>,
+    access_lists: Option<<Client as BlockAccessListsClient>::Output>,
+    block_result: Option<SealedBlock<Client::Block>>,
+    access_lists_result: Option<BlockAccessLists>,
```
mattsse: can we group these based on block,bal

crates/net/p2p/src/full_block.rs
```diff
@@ -126,6 +143,25 @@ where
     body: Option<BodyResponse<Client::Body>>,
 }
 
+impl<Client> FetchFullBlockFuture<Client>
+where
+    Client: BlockClient,
+{
+    fn new(client: Client, consensus: Arc<dyn Consensus<Client::Block>>, hash: B256) -> Self {
```
mattsse: can we move this next to 

```
pub fn get_full_block(&self, hash: B256) -> FetchFullBlockFuture<Client> {
```

mattsse (changes requested):
directionally this looks good

left some pedantic nits

crates/net/p2p/src/full_block.rs
0xKarl98: Timmed the extra state out , only keep one internal AccessListsState now, so pending vs ready lives in one place.

crates/net/p2p/src/full_block.rs
0xKarl98: Have wrapped them into loop , and also apply the budge of 4 to it, just as :
https://github.com/paradigmxyz/reth/blob/main/crates/net/p2p/src/full_block.rs#L192

0xKarl98:
@mattsse  Have applied your nits , please take a look again

mattsse (approved):

PR #23633, opened 2026-04-20
perf(chain-state): flatten in-memory overlay for O(1) state reads

### LINKED ISSUE #20612: Experiment: flatten plain state for MemoryStateProvider
### Describe the feature

When using MemoryStateProvider we pass it an array of `ExecutedBlock`s, which it then iterates over to serve requests:

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/chain-state/src/memory_overlay.rs#L111-L115

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/chain-state/src/memory_overlay.rs#L215-L219

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/chain-state/src/memory_overlay.rs#L227-L231

We should see if we can flatten the BundleState of the `ExecutedBlock`s into a single state, which can be applied as an overlay instead. This can be done asynchronously in a similar manner to the trie input:

https://github.com/paradigmxyz/reth/blob/b79c58d83537b8db52a6a6c02e81d2f868bd4f6d/crates/engine/tree/src/tree/payload_validator.rs#L1050

Note that we can only use this flattened state if its anchor block shares the anchor of the block being validated. If not then we must either compute a new flattened state or fallback to the current behavior.

### Additional context

_No response_

zacharyftw:
Adds a `DeferredBundleState` handle on `ExecutedBlock` that asynchronously                                                                                                                 
  flattens in-memory ancestor `BundleState`s into one cumulative overlay,          
  anchored to the oldest in-memory block's parent hash. `MemoryOverlayStateProvider`                                                                                                         
  reads collapse from O(N) iteration across every in-memory block to a single                                                                                                                
  HashMap probe, falling back to iteration on anchor mismatch (correctness
  preserved across persistence events).                                                                                                                                                      
                                                                                                                                                                                             
  Benchmarks at N=64 (default `--engine.memory-block-buffer-target`):
                                                                                                                                                                                             
  | read                 | legacy    | flattened | speedup |                       
  |----------------------|-----------|-----------|---------|                                                                                                                                 
  | `basic_account` miss | 153.9 ns  | 33.0 ns   | 4.67x   |                       
  | `storage` miss       | 277.5 ns  | 33.3 ns   | 8.33x   |                                                                                                                                 
  | `bytecode` miss      | 93.6 ns   | 34.0 ns   | 2.75x   |                       
                                                                                                                                                                                             
  Flattened timings stay flat regardless of N; legacy scales linearly.
  Bench reproducer: `crates/chain-state/benches/memory_overlay.rs`.                                                                                                                          
                                                                                   
  Follows the existing `DeferredTrieData` pattern (pending/ready state                                                                                                                       
  machine, anchor-hash validation, parent overlay reuse via `Arc::make_mut`).

Closes #20612

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7496,6 +7496,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-signer",
  "alloy-signer-local",
+ "codspeed-criterion-compat",
  "derive_more",
  "metrics",
  "parking_lot",
--- a/crates/chain-state/Cargo.toml
+++ b/crates/chain-state/Cargo.toml
@@ -54,6 +54,7 @@ reth-primitives-traits = { workspace = true, features = ["test-utils"] }
 reth-testing-utils.workspace = true
 alloy-signer.workspace = true
 alloy-signer-local.workspace = true
+criterion.workspace = true
 rand.workspace = true
 revm-state.workspace = true
 
@@ -85,3 +86,7 @@ test-utils = [
     "reth-ethereum-primitives/test-utils",
 ]
 rayon = ["dep:rayon"]
+
+[[bench]]
+name = "memory_overlay"
+harness = false
--- a/crates/chain-state/benches/memory_overlay.rs
+++ b/crates/chain-state/benches/memory_overlay.rs
@@ -1,0 +1,278 @@
+//! Benchmarks for `MemoryOverlayStateProvider` state-read hot paths.
+//!
+//! Two modes are compared:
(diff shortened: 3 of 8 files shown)


mattsse (commented):
ty for this

we're currently experimenting with a few things that are similar to this change but conflicting and we still need to do more testing first

Issue #23632, opened 2026-04-20
perf(stages): parallelize MerkleExecute for pipeline initial sync


joohhnnn:
## Motivation

The pipeline `MerkleStage::Execution` computes per-account storage roots serially on the same thread that walks the account trie. On mainnet, a single XEN-class account can take 60+ seconds for its storage trie alone, blocking everything else.

reth already has `ParallelStateRoot` (used by the engine tree's payload processor) which dispatches per-account storage-root computation to a worker pool. Wiring it into the pipeline `MerkleExecute` lets that single-account work overlap with the rest of the chunk.

## Bench

Hetzner i7-8700 (12 threads), mainnet pipeline sync. The size of the parallelism win scales with the weight of single-account storage tries in the workload — serial must compute every per-account storage root on the main thread, while parallel absorbs large accounts onto workers.

| Workload | Variant | MerkleExecute | Δ |
|---|---|---:|---:|
| 2000 blocks (24885314 → 24887313) | Baseline reth 2.0.0 serial | 38.0 s | — |
| | `ParallelMerkleExecutionStage` (this proposal) | 30.2 s | **−21 %** |
| 7000 blocks (24885314 → 24892314) | Baseline reth 2.0.0 serial | 112.4 s | — |
| | `ParallelMerkleExecutionStage` (this proposal) | 85.8 s | **−24 %** |
| 13811 blocks (24885313 → 24901124), includes an Heavy-class account with ~60 s storage trie | Baseline reth 2.0.0 serial | 814 s | — |
| | `ParallelMerkleExecutionStage` (this proposal) | 494 s | **−39 %** |

The two smaller workloads come from isolated stage-replay runs (no engine API load, warm OS page cache). The 13811-block workload comes from a full pipeline sync with Lighthouse attached (cold cache); the absolute numbers are larger than the isolated runs due to cache state, but the relative speedup is what matters.

Strict isolated A/B (Lighthouse stopped, `--debug.tip` driven, 2000 blocks) of `ParallelMerkleExecutionStage` vs its closest serial-equivalent baseline matches within 0.006 % noise (160.406 s vs 160.415 s). The wall-clock win comes purely from parallelism, not an algorithm change.

## Plan

Four PRs, intentionally small and independently mergeable. The first three are pre-existing bugs that benefit the serial path too; the fourth is the actual perf work and has a compile-time dependency on the second.

[...]

PR #23634, opened 2026-04-20
fix(trie): length-prefix StoredSubNode inner node in compact codec

joohhnnn:
Tracked in #23632.

StoredSubNode writes the inner BranchNodeCompact payload without a length prefix, so the decoder guesses where it ends by reading bytes and checking if they look like a valid branch node. When state_mask (big-endian u16) collides with a plausible compact length (e.g. 6, 38, 70, 550), the decoder eats too far and corrupts the wrapping MerkleCheckpoint.

BranchNodeCompact::from_compact's assert_eq!(buf.len() % 32, 6) masks this today, but one extra field anywhere downstream is enough to flip it into a startup panic on resume.

This isn't hypothetical: the parallel MerkleExecute work tracked in #23632 adds a should_check_walker_key byte to MerkleCheckpoint's trailer and exercises the Progress/Resume path far more often. Without this fix, any user upgrading with a paused MerkleExecute checkpoint whose walker_stack contains a Some(node) would hit the assertion on restart.

Tag the inner node with a version byte — 1 = legacy (no length prefix), 2 = new ([len: u16][payload]). Writers always emit 2; readers accept both, so the decoder always knows the exact byte count while old paused checkpoints still decode. Includes a regression test for the colliding mask values.

### DIFF
--- a/crates/trie/common/src/subnode.rs
+++ b/crates/trie/common/src/subnode.rs
@@ -36,7 +36,12 @@ impl reth_codecs::Compact for StoredSubNode {
         if let Some(node) = &self.node {
             buf.put_u8(1);
             len += 1;
-            len += node.to_compact(buf);
+
+            let mut node_buf = Vec::new();
+            let node_len = node.to_compact(&mut node_buf);
+            buf.put_u16(node_len as u16);
+            buf.put_slice(&node_buf);
+            len += 2 + node_len;
         } else {
             len += 1;
             buf.put_u8(0);
@@ -57,8 +62,9 @@ impl reth_codecs::Compact for StoredSubNode {
 
         let node_exists = buf.get_u8() != 0;
         let node = node_exists.then(|| {
-            let (node, rest) = BranchNodeCompact::from_compact(buf, 0);
-            buf = rest;
+            let node_len = buf.get_u16() as usize;
+            let (node, _) = BranchNodeCompact::from_compact(&buf[..node_len], node_len);
+            buf.advance(node_len);
             node
         });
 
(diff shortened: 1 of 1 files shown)


crates/trie/common/src/subnode.rs
```diff
@@ -36,7 +36,12 @@ impl reth_codecs::Compact for StoredSubNode {
         if let Some(node) = &self.node {
             buf.put_u8(1);
```
mediocregopher: Ideally we would use a new prefix here (`2`) to indicate the slightly changed encoding, and then in the decoding we can support both the older and newer forms.

mediocregopher:
cc @Rjected @mattsse

crates/trie/common/src/subnode.rs
joohhnnn: Sure, switched to the tag-byte versioned codec.

mattsse (commented):
this is breaking, right?

joohhnnn:
> this is breaking, right?

Yeah.

mediocregopher:
@joohhnnn @mattsse I don't think this is actually breaking, the previous form was tagged with a `1` byte and the new is tagged with a `2`, and both are handled on the deserialize path. Or am I missing something?

joohhnnn:
> @joohhnnn @mattsse I don't think this is actually breaking, the previous form was tagged with a `1` byte and the new is tagged with a `2`, and both are handled on the deserialize path. Or am I missing something?

Ah sorry for the confusion — it's not breaking anymore after 031d85f8a2 (the v2 tag-byte fallback), I just forgot to update the comment/PR description. Let me fix it.

PR #23639, opened 2026-04-20
feat: external ip when listen addr is unspecified

### LINKED ISSUE #23577: bug(discv5): ENR missing external IP when listening on 0.0.0.0 with --nat extip
## Description

When running reth with `--addr 0.0.0.0 --nat extip:<IP>`, the discv5 ENR (Ethereum Node Record) is built without any IPv4 address. This means the node advertises an incomplete ENR to the discv5 network, making it initially undiscoverable by other discv5 peers.

## Root Cause

In [`build_local_enr()`](https://github.com/paradigmxyz/reth/blob/main/crates/net/discv5/src/lib.rs#L475-L508), when the listen address is `0.0.0.0` (UNSPECIFIED), the builder skips calling `builder.ip4(ip)`:

```rust
ListenConfig::Ipv4 { ip, port } => {
    if ip != Ipv4Addr::UNSPECIFIED {
        builder.ip4(ip);  // skipped when ip == 0.0.0.0
    }
    // ...
}
```

There is no mechanism to pass the NAT-resolved external IP into the discv5 `Config`, so `build_local_enr()` has no fallback IP to use. The resulting ENR has `ip4() == None`.

## Relationship to #19784

PR #19784 (`fix(net): use external ip for discv5 config`) addressed a related symptom by fixing `NetworkHandle::local_node_record()` to return the correct IP for RPC responses (e.g., `admin_nodeInfo`). However, the **actual discv5 ENR** that gets broadcast to peers via the discv5 protocol is still built without the external IP. This ENR is what other nodes use to discover and connect to us.

## Regression Unit Tests

The following tests demonstrate the bug. They can be added to `crates/net/discv5/src/lib.rs` in the `mod test` block:

```rust
/// Regression test: when RLPx binds to 0.0.0.0 (all interfaces) and the user configures
/// `--nat extip:<IP>`, the discv5 ENR must contain the NAT-resolved external IPv4 address.
#[test]
fn build_enr_includes_external_ip_when_listen_addr_is_unspecified() {
    let nat_external_ip = Ipv4Addr::new(1, 2, 3, 4);
    let tcp_port: u16 = 30303;
    let udp_port: u16 = 9200;

    // Simulate: --addr 0.0.0.0 --nat extip:1.2.3.4
    let rlpx_tcp_socket: SocketAddr = (Ipv4Addr::UNSPECIFIED, tcp_port).into();
    let discv5_listen_config =
        ListenConfig::Ipv4 { ip: Ipv4Addr::UNSPECIFIED, port: udp_port };

    let config = Config::builder(rlpx_tcp_socket)
        .discv5_config(discv5::ConfigBuilder::new(discv5_listen_config).build())
        .build();

    let sk = SecretKey::new(&mut thread_rng());
    let (enr, _, _, _) = build_local_enr(&sk, &config);

    // The ENR MUST contain the external IPv4 address so that other nodes can discover us.
    // With the bug, enr.ip4() returns None because build_local_enr skips setting the IP
    // when the listen address is 0.0.0.0 and there is no mechanism to pass the NAT IP.
    assert_eq!(
        enr.ip4(),
        Some(nat_external_ip),
        "discv5 ENR must advertise the NAT-resolved external IP ({nat_external_ip}) \
         when listen address is 0.0.0.0, but got {:?}",
        enr.ip4()
    );
}

/// Regression test: dual-stack variant of the external IP bug.
#[test]
fn build_enr_includes_external_ip_when_listen_addr_is_unspecified_dual_stack() {
    let nat_external_ip = Ipv4Addr::new(5, 6, 7, 8);
    let tcp_port: u16 = 30303;
    let udp_port_v4: u16 = 9200;
    let udp_port_v6: u16 = 9201;

    let rlpx_tcp_socket: SocketAddr = (Ipv4Addr::UNSPECIFIED, tcp_port).into();
    let discv5_listen_config = ListenConfig::DualStack {
        ipv4: Ipv4Addr::UNSPECIFIED,
        ipv4_port: udp_port_v4,
        ipv6: Ipv6Addr::UNSPECIFIED,
        ipv6_port: udp_port_v6,
    };

    let config = Config::builder(rlpx_tcp_socket)
        .discv5_config(discv5::ConfigBuilder::new(discv5_listen_config).build())
        .build();

    let sk = SecretKey::new(&mut thread_rng());
    let (enr, _, _, _) = build_local_enr(&sk, &config);

    assert_eq!(
        enr.ip4(),
        Some(nat_external_ip),
        "discv5 ENR must advertise the NAT-resolved external IP ({nat_external_ip}) \
         in dual-stack mode when IPv4 listen address is 0.0.0.0, but got {:?}",
        enr.ip4()
    );
}
```

## Steps to Reproduce

1. Apply the above tests to `crates/net/discv5/src/lib.rs` inside the `#[cfg(test)] mod test` block.
2. Add `alloy-primitives = { workspace = true, features = ["rand"] }` to `[dev-dependencies]` in `crates/net/discv5/Cargo.toml` (needed for existing test code to compile with `PeerId::random()`).
3. Run:
   ```bash
   cargo test -p reth-discv5 --lib -- test::build_enr_includes_external_ip
   ```
4. Both tests fail with `assertion left == right failed`:
   ```
   left: None
   right: Some(1.2.3.4)
   ```

## Suggested Fix

Add an `external_ip: Option<Ipv4Addr>` field to the discv5 `Config` / `ConfigBuilder`, and use it in `build_local_enr()` as a fallback when the listen address is UNSPECIFIED:

```rust
ListenConfig::Ipv4 { ip, port } => {
    if ip != Ipv4Addr::UNSPECIFIED {
        builder.ip4(ip);
    } else if let Some(ext_ip) = external_ip {
        builder.ip4(*ext_ip);
    }
    // ...
}
```

The NAT-resolved IP can be passed from `DiscoveryArgs::discovery_v5_builder()` in `crates/node/core/src/args/network.rs`.

## Platform

- Reth version: main branch (HEAD)
- OS: N/A (code-level bug)

cuiweixie:
fixes #23577

### DIFF
--- a/crates/node/core/src/args/network.rs
+++ b/crates/node/core/src/args/network.rs
@@ -957,9 +957,10 @@ impl DiscoveryArgs {
     /// Apply the discovery settings to the given `NetworkConfigBuilder` from `reth_network`.
     ///
     /// `rlpx_tcp_socket` and `boot_nodes` should match the builder's `listener_addr` and
-    /// `boot_nodes`, respectively. The NAT resolver for ENR `external_ip` is taken from the
-    /// builder at the start of this function (before discovery may call `disable_nat`), so
-    /// `--nat extip:...` still applies to the discv5 ENR when appropriate (see issue #14878).
+    /// `boot_nodes`, respectively. The NAT resolver is read from the builder at the start of this
+    /// function (before discovery may call `disable_nat`), so `--nat extip:...` still applies to
+    /// the discv5 ENR when appropriate
+    /// (<https://github.com/paradigmxyz/reth/issues/14878>).
     pub fn apply_to_builder<N>(
         &self,
         mut network_config_builder: NetworkConfigBuilder<N>,
@@ -969,9 +970,15 @@ impl DiscoveryArgs {
     where
         N: NetworkPrimitives,
     {
-        // Read before `disable_nat` mutates the builder: discv5 still uses this for `external_ip`
-        // in the local ENR when `nat:extip:...` is set (see issue #14878).
+        // Read before `disable_nat` mutates the builder: discv5 still uses the resolved IPv4 for
+        // the local ENR when `nat:extip:...` is set
+        // (<https://github.com/paradigmxyz/reth/issues/14878>).
         let nat_for_enr: Option<NatResolver> = network_config_builder.nat_resolver().cloned();
+        let nat_external_ipv4 =
+            nat_for_enr.as_ref().and_then(|nat| match nat.clone().as_external_ip(0) {
+                Some(IpAddr::V4(v4)) if !v4.is_unspecified() => Some(v4),
+                _ => None,
+            });
 
         if self.disable_discovery || self.disable_dns_discovery {
             network_config_builder = network_config_builder.disable_dns_discovery();
@@ -987,8 +994,9 @@ impl DiscoveryArgs {
         }
 
         if self.should_enable_discv5() {
-            network_config_builder = network_config_builder
-                .discovery_v5(self.discovery_v5_builder(rlpx_tcp_socket, boot_nodes, nat_for_enr));
+            network_config_builder = network_config_builder.discovery_v5(
+                self.discovery_v5_builder(rlpx_tcp_socket, boot_nodes, nat_external_ipv4),
+            );
         }
 
         network_config_builder
@@ -999,7 +1007,7 @@ impl DiscoveryArgs {
         &self,
         rlpx_tcp_socket: SocketAddr,
         boot_nodes: impl IntoIterator<Item = NodeRecord>,
-        nat: Option<NatResolver>,
+        nat_external_ipv4: Option<Ipv4Addr>,
     ) -> reth_discv5::ConfigBuilder {
         let Self {
             discv5_addr,
@@ -1037,20 +1045,8 @@ impl DiscoveryArgs {
             discv5_config_builder.disable_enr_update();
         }
 
-        // When the user configured `--nat extip:<IP>` and rlpx binds to an unspecified address,
-        // forward the resolved external IP through the rlpx tcp socket. `build_local_enr` then
-        // falls back to it for the ENR when the discv5 listen address is unspecified.
-        let rlpx_tcp_socket = if let Some(nat) = nat.as_ref() &&
-            let Some(IpAddr::V4(ip)) = nat.clone().as_external_ip(0) &&
-            !ip.is_unspecified() &&
-            rlpx_tcp_socket.ip().is_unspecified()
-        {
-            SocketAddr::new(IpAddr::V4(ip), rlpx_tcp_socket.port())
-        } else {
-            rlpx_tcp_socket
-        };
-
         reth_discv5::Config::builder(rlpx_tcp_socket)
+            .with_nat_external_ipv4_for_unspecified_bind(nat_external_ipv4)
             .discv5_config(discv5_config_builder.build())
             .add_unsigned_boot_nodes(boot_nodes)
             .lookup_interval(*discv5_lookup_interval)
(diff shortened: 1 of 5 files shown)


cuiweixie:
@mattsse Can you help to take a look on this pr?

crates/node/core/src/args/network.rs
```diff
@@ -780,6 +780,7 @@ impl DiscoveryArgs {
         mut network_config_builder: NetworkConfigBuilder<N>,
         rlpx_tcp_socket: SocketAddr,
         boot_nodes: impl IntoIterator<Item = NodeRecord>,
+        nat: &NatResolver,
```
mattsse: why do we need this as additional arg if this is already in `NetworkConfigBuilder`

crates/optimism/node/src/node.rs
mattsse, line 1: dont think we need this?

mattsse (changes requested):
sorry for the delay, stopped looking at it when I saw this was >1k lines -.-

the op code might be some leftovers

crates/optimism/node/src/node.rs
cuiweixie: Sorry, let me make it right, quickly.

crates/node/core/src/args/network.rs
cuiweixie: fix it

cuiweixie: Should not change the func sig, since it's public.

crates/node/core/src/args/network.rs
```diff
@@ -781,7 +780,12 @@ pub struct DiscoveryArgs {
 }
 
 impl DiscoveryArgs {
-    /// Apply the discovery settings to the given [`NetworkConfigBuilder`]
+    /// Apply the discovery settings to the given `NetworkConfigBuilder` from `reth_network`.
+    ///
+    /// `rlpx_tcp_socket` and `boot_nodes` should match the builder's `listener_addr` and
+    /// `boot_nodes`, respectively. The NAT resolver for ENR `external_ip` is taken from the
+    /// builder at the start of this function (before discovery may call `disable_nat`), so
+    /// `--nat extip:...` still applies to the discv5 ENR when appropriate (see issue #14878).
```
mattsse: this should not mention raw issue numbers
especially not without a url

crates/node/core/src/args/network.rs
```diff
@@ -806,17 +814,18 @@ impl DiscoveryArgs {
...
         network_config_builder
     }
 
-    /// Creates a [`reth_discv5::ConfigBuilder`] filling it with the values from this struct.
+    /// Creates a [`reth_discv5::ConfigBuilder`] from CLI discovery args and the network builder.
     pub fn discovery_v5_builder(
         &self,
         rlpx_tcp_socket: SocketAddr,
         boot_nodes: impl IntoIterator<Item = NodeRecord>,
+        nat: Option<NatResolver>,
```
mattsse: this now accepts Nat but we only use its ipv4

crates/node/core/src/args/network.rs
```diff
@@ -851,12 +860,22 @@ impl DiscoveryArgs {
...
+
+        // Set external IP from NAT resolver for the ENR when bind address is 0.0.0.0
+        if let Some(nat) = nat.as_ref() &&
+            let Some(IpAddr::V4(ip)) = nat.clone().as_external_ip(0) &&
+            !ip.is_unspecified()
+        {
+            builder = builder.external_ip(ip);
+        }
+
+        builder
```
mattsse: this is applied to the builder, so I think this should be its own fn on

```
reth_discv5::ConfigBuilder
```

mattsse (changes requested):
more bikeshed, sorry

cuiweixie:
@mattsse any new feedback.  we needs this feature in our node.

mattsse:
Closing in favor of #24013, which implements the `--nat extip:<IP>` to local ENR propagation from current `main` with a focused test.

PR #23651, opened 2026-04-20
perf(stages): parallelize incremental MerkleExecute

joohhnnn:
Adds `ParallelMerkleExecutionStage` driving `ParallelStateRoot` for the incremental path. The pipeline's `MerkleExecute` slot routes to the new stage, which dispatches based on the input:

| Input | Routes to |
|---|---|
| range > `rebuild_threshold` (default 100k, full rebuild) | serial `MerkleStage::Execution` |
| `from_block == 1` (initial merkleization) | serial `MerkleStage::Execution` |
| chunk modifies < 1024 storage tries (parallel overhead too high) | serial `MerkleStage::Execution` |
| empty range | no-op |
| otherwise (incremental, ≥ 1024 modified storage tries) | **`ParallelStateRoot`** |

`MerkleStage::Unwind` keeps the `MerkleUnwind` slot. The serial rebuild path — including its existing Progress/Resume + intra-storage checkpointing that handles XEN-class accounts — is unchanged.

## `ParallelStateRoot::calculate` changes

- `tokio::spawn_blocking` → `rayon::spawn` (work-stealing pool sized to CPU cores, matching `sender_recovery` / `hashing_storage`).
- Eager spawn-all → bounded frontier (`max_in_flight_storage_root_tasks`).
- Worker-miss fallback uses the account's real `storage_prefix_set` instead of `Default::default()` (the empty-set path silently dropped storage updates when the frontier throttled).

## Equivalence

- `parallel_trie_updates_match_serial` — `(root, TrieUpdates)` bit-for-bit identical to serial.
- `parallel_multi_chunk_trie_tables_match_serial` — multi-chunk run leaves `AccountsTrie` / `StoragesTrie` matching serial.

## Bench

Hetzner i7-8700, 12 threads, mainnet:

| Workload | Serial | Parallel | Δ |
|---|---:|---:|---:|
| 2000 blocks (24885314 → 24887313) | 38.0 s | 30.2 s | **−21 %** |
| 7000 blocks (24885314 → 24892314) | 112.4 s | 85.8 s | **−24 %** |
| 13811 blocks (24887313 → 24901124), includes a heavy account | 814 s | 494 s | **−39 %** |

The 13811-block chunk contains one heavy account whose storage root alone takes ~60 s; absorbing it on a worker while the main thread keeps walking is the main driver of the 39 % win. 

## Scope

- Engine tree (sparse trie) live-sync path is untouched.
- Worker granularity is per-account; intra-account storage parallelism is out of scope (XEN-class single-account stalls remain).
- `root_with_progress` intentionally not exposed — bench shows per-`Progress` overhead in parallel is ~100× serial, so chunks run single-shot.

### DIFF
--- a/crates/trie/parallel/src/root.rs
+++ b/crates/trie/parallel/src/root.rs
@@ -117,15 +117,26 @@ where
             self.runtime.cpu_pool().spawn(move || {
                 let result = (|| -> Result<_, ParallelStateRootError> {
                     let provider = factory.database_provider_ro()?;
-                    Ok(StorageRoot::new_hashed(
+                    let storage_root = StorageRoot::new_hashed(
                         &provider,
                         &provider,
                         hashed_address,
                         prefix_set,
                         #[cfg(feature = "metrics")]
                         metrics,
-                    )
-                    .calculate(retain_updates)?)
+                    );
+                    // Use `root_with_updates` / `root`, not `calculate`: those wrappers set
+                    // `with_no_threshold()` first, guaranteeing the storage walker runs to
+                    // completion. `calculate` may return `Progress` for accounts whose
+                    // hashed-entries-walked exceeds the default threshold (XEN-class accounts
+                    // in pipeline chunks), and `ParallelStateRoot` has no mechanism to
+                    // resume a paused worker.
+                    Ok(if retain_updates {
+                        let (root, _, updates) = storage_root.root_with_updates()?;
+                        (root, updates)
+                    } else {
+                        (storage_root.root()?, Default::default())
+                    })
                 })();
                 let _ = tx.send(result);
             });
@@ -155,7 +166,7 @@ where
                     hash_builder.add_branch(node.key, node.value, node.children_are_in_trie);
                 }
                 TrieElement::Leaf(hashed_address, account) => {
-                    let storage_root_result = match storage_roots.remove(&hashed_address) {
+                    let (storage_root, updates) = match storage_roots.remove(&hashed_address) {
                         Some(rx) => rx.recv().map_err(|_| {
                             ParallelStateRootError::StorageRoot(StorageRootError::Database(
                                 DatabaseError::Other(format!(
@@ -174,26 +185,20 @@ where
                                 .get(&hashed_address)
                                 .cloned()
                                 .unwrap_or_default();
-                            StorageRoot::new_hashed(
+                            let storage_root = StorageRoot::new_hashed(
                                 &provider,
                                 &provider,
                                 hashed_address,
                                 fallback_prefix_set,
                                 #[cfg(feature = "metrics")]
                                 self.metrics.storage_trie.clone(),
-                            )
-                            .calculate(retain_updates)?
-                        }
-                    };
-
-                    let (storage_root, _, updates) = match storage_root_result {
-                        reth_trie::StorageRootProgress::Complete(root, _, updates) => (root, (), updates),
-                        reth_trie::StorageRootProgress::Progress(..) => {
-                            return Err(ParallelStateRootError::StorageRoot(
-                                StorageRootError::Database(DatabaseError::Other(
-                                    "StorageRoot returned Progress variant in parallel trie calculation".to_string()
-                                ))
-                            ))
+                            );
+                            if retain_updates {
+                                let (root, _, updates) = storage_root.root_with_updates()?;
+                                (root, updates)
+                            } else {
+                                (storage_root.root()?, Default::default())
+                            }
                         }
                     };
 
--- a/crates/trie/parallel/src/root.rs
+++ b/crates/trie/parallel/src/root.rs
@@ -5,7 +5,6 @@ use alloy_primitives::B256;
 use alloy_rlp::{BufMut, Encodable};
 use itertools::Itertools;
 use reth_execution_errors::{SparseTrieError, StateProofError, StorageRootError};
-use reth_primitives_traits::Account;
 use reth_provider::{DatabaseProviderROFactory, ProviderError};
 use reth_storage_errors::db::DatabaseError;
 use reth_trie::{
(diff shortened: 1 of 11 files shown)


crates/trie/parallel/src/root.rs
```diff
@@ -37,20 +41,17 @@ pub struct ParallelStateRoot<Factory> {
     factory: Factory,
     // Prefix sets indicating which portions of the trie need to be recomputed.
     prefix_sets: TriePrefixSets,
-    /// The runtime handle for spawning blocking tasks.
-    runtime: Runtime,
```
mediocregopher, line 41: I believe we want to continue to use Runtime, it's actually using rayon under-the-hood now, but it has a bunch of builtin niceities like naming of threads for profiling and such.

crates/trie/parallel/src/root.rs
```diff
@@ -66,145 +67,198 @@ where
...
+        self.calculate(true).map(|(root, _, updates)| (root, updates))
     }
 
     /// Computes the state root by calculating storage roots in parallel for modified accounts,
     /// then walking the state trie to build the final state root hash.
     fn calculate(
         self,
         retain_updates: bool,
-    ) -> Result<(B256, TrieUpdates), ParallelStateRootError> {
+    ) -> Result<(B256, usize, TrieUpdates), ParallelStateRootError> {
```
mediocregopher: Is anything using this new `usize` return?

crates/trie/parallel/src/root.rs
```diff
@@ -269,33 +323,95 @@ impl From<StateProofError> for ParallelStateRootError {
...
+struct ParallelStateRootContext {
+    /// Reusable buffer for encoding account data.
+    account_rlp: Vec<u8>,
+    /// Accumulates updates from account and storage root calculation.
+    trie_updates: TrieUpdates,
+    /// Tracks total hashed entries walked.
+    hashed_entries_walked: usize,
+}
+
+impl ParallelStateRootContext {
```
mediocregopher: tbh I don't feel like this is adding a lot of value, over just inlining all this state/logic like before

crates/trie/parallel/src/root.rs
```diff
@@ -66,145 +67,198 @@ where
...
-        let mut storage_roots = HashMap::with_capacity(storage_root_targets.len());
-
-        let handle = self.runtime.handle().clone();
+        tracker.set_precomputed_storage_roots(total_storage_root_targets as u64);
+        debug!(
+            target: "trie::parallel_state_root",
+            len = total_storage_root_targets,
+            "pre-calculating storage roots"
+        );
+        let mut storage_roots = BTreeMap::new();
```
mediocregopher: Why change this to a BTreeMap? It seems like we're still just `remove`ing off of it once the account is encountered

mediocregopher (commented):
I don't mind the general strategy of invoking ParallelStateRoot during incremental trie rebuilds, that makes sense to me. However I'm not sure about some of the changes to ParallelStateRoot itself, it seems to me like it's operating basically exactly like it used to, so we may as well leave it be. And I don't believe it makes sense to have a separate parallel MerkleExecute stage, but maybe that was just for proof-of-concept.

@joohhnnn I'm curious if you've had any further thoughts about the full rebuild case? Imo that's where we have the most to gain, though obviously speeding up the incremental rebuild is also valuable.

crates/trie/parallel/src/root.rs
joohhnnn: nothing consumes it — was speculatively reserving it for a future count_entries::<HashedAccounts>() shortcut in the stage's checkpoint write(not related to this part), dropped in 7145b513

crates/trie/parallel/src/root.rs
joohhnnn: Thought SELFDESTRUCT'd accounts' orphan receivers needed early cleanup, but on a second look each one is a single mpsc slot and there's usually < 10 per chunk — reverted to HashMap in 83cda761.

crates/trie/parallel/src/root.rs
joohhnnn: Thought we could drop the runtime field once we switched to rayon::spawn, but on a second look Runtime already wraps named rayon pools — restored in 62c5231b, now using runtime.cpu_pool().spawn().

crates/trie/parallel/src/root.rs
joohhnnn: Sure, revert in 7145b513. Also found a latent PSR bug while re-benching: the worker called .calculate(), which returns Progress once the storage walker passes the default threshold — PSR has no resume, so upstream's single-block engine-tree usage never trips it, but our 7000-block pipeline chunks with large accounts do; fixed in 0db94616 by switching to .root_with_updates() (which uses with_no_threshold()).

joohhnnn:
> However I'm not sure about some of the changes to ParallelStateRoot itself, it seems to me like it's operating basically exactly like it used to, so we may as well leave it be. 

Resolved in the sub-review

joohhnnn:
> And I don't believe it makes sense to have a separate parallel MerkleExecute stage, but maybe that was just for proof-of-concept.

Agreed. In d23e952265f96a7f9f58e61dcd55b138c9c85fa7 I removed the standalone `ParallelMerkleExecutionStage` and folded the parallel path into `MerkleStage::Execution` behind an optional `MerkleParallelConfig`.

`MerkleStage` is now the single owner of `StageId::MerkleExecute`; setup enables parallel execution with `MerkleStage::new_execution(...).with_parallel(...)`.

The tricky part was multi-chunk execution: `ParallelStateRoot` opens fresh read-only providers, so later chunks cannot see earlier uncommitted trie updates. Instead of exposing each chunk as a stage checkpoint, the stage now carries prior chunk `TrieUpdatesSorted` as an in-memory overlay and only advances the main checkpoint after the final target root is validated.

A small type-erased `IncrementalStateRootDispatcher` keeps `MerkleStage` generic-free while allowing node setup to inject the concrete parallel implementation.

PR #23655, opened 2026-04-20
chore(bench): add cli flag to fetch bal by default

Soubhik-10:


DaniPopes (approved):

mattsse (approved):
that should do it now

PR #23656, opened 2026-04-20
refactor: encapsulate state fetching in db provider

klkvr:
Right now the only way to obtain a `BundleState` from db changesets is by using `ConsistentProvider`. This operation only needs access to database and thus belongs to `DatabaseProvider`

This change allows to remove a duplicated `populate_bundle_state` fn on `ConsistentProvider`. Also as a drive-by I've refactored `populate_bundle_state` on `DatabaseProvider` so that we don't have to add a third implementation of it in addition to hashed/plain duplication

### DIFF
--- a/crates/storage/provider/src/providers/blockchain_provider.rs
+++ b/crates/storage/provider/src/providers/blockchain_provider.rs
@@ -133,16 +133,6 @@ impl<N: ProviderNodeTypes> BlockchainProvider<N> {
         let latest_historical = self.database.history_by_block_hash(anchor_hash)?;
         Ok(state.state_provider(latest_historical))
     }
-
-    /// Return the last N blocks of state, recreating the [`ExecutionOutcome`].
-    ///
-    /// If the range is empty, or there are no blocks for the given range, then this returns `None`.
-    pub fn get_state(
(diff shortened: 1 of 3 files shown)


crates/storage/provider/src/providers/blockchain_provider.rs
```diff
@@ -133,16 +133,6 @@ impl<N: ProviderNodeTypes> BlockchainProvider<N> {
...
-
-    /// Return the last N blocks of state, recreating the [`ExecutionOutcome`].
-    ///
-    /// If the range is empty, or there are no blocks for the given range, then this returns `None`.
-    pub fn get_state(
-        &self,
-        range: RangeInclusive<BlockNumber>,
-    ) -> ProviderResult<Option<ExecutionOutcome<ReceiptTy<N>>>> {
-        self.consistent_provider()?.get_state(range)
-    }
```
klkvr, line 145: this method was not used anywhere in the codebase, do we think someone might be relying on it?

shekhirin (approved):

DaniPopes (approved):

mattsse (approved):

PR #23658, opened 2026-04-20
refactor(trie): remove TrieNodeProvider

mediocregopher:
* Implements a small refactoring in SparseTrieCacheTask to remove the last real usage of TrieNodeProvider.

* Remove update_leaf / remove_leaf methods from the SparseTrie trait, they are no longer used anywhere.

* Delete TrieNodeProvider trait definition, all of its impls, and the associated code-paths in the proof workers.


crates/trie/sparse/src/state.rs
```diff
@@ -622,12 +602,9 @@ where
             storage_trie.root().ok_or(SparseTrieErrorKind::Blind)?
         } else if self.is_account_revealed(address) {
             trace!(target: "trie::sparse", ?address, "Retrieving storage root from account leaf to update account");
-            // The account was revealed, either...
```
mediocregopher, line 625: Why were all these comments removed?

crates/trie/sparse/src/trie.rs
```diff
@@ -202,6 +202,30 @@ impl<T: SparseTrieTrait> RevealableSparseTrie<T> {
         };
     }
 
+    /// Shrinks the capacity of the sparse trie's node storage.
```
mediocregopher, line 202: Why did these methods get moved?

crates/trie/sparse/src/state.rs
mediocregopher: Restored the explanation in provider-free form in e476826930. The fallback to `EMPTY_ROOT_HASH` is still intentional here because a revealed account path can exist before the account leaf does when only the storage trie has been materialized so far.

crates/engine/tree/src/tree/payload_processor/sparse_trie.rs
```diff
@@ -359,10 +364,25 @@ where
         debug!(target: "engine::root", "All proofs processed, ending calculation");
 
         let start = Instant::now();
-        let (state_root, trie_updates) =
-            self.trie.root_with_updates(&self.proof_worker_handle).map_err(|e| {
-                ParallelStateRootError::Other(format!("could not calculate state root: {e:?}"))
-            })?;
+        let (state_root, trie_updates) = match self.trie.root_with_updates() {
```
mediocregopher, line 367: Note: This is the only logical change to this PR. If the block had no state updates then the account trie root node won't necessarily be revealed. Previously we would reveal it on-the-fly using a TrieNodeProvider. Now `root_with_updates` returns a blinded error. In this case we can assume that the SR hasn't changed, and we return the previous block's state root.

crates/trie/sparse/src/parallel.rs
```diff
@@ -432,1043 +432,954 @@ impl SparseTrie for ParallelSparseTrie {
         }
     }
 
-    fn update_leaf<P: TrieNodeProvider>(
```
mediocregopher, line 435: The only real change to this file was to remove the update_leaf/remove_leaf methods out of the SparseTrie impl. This can be seen better using a `--patience ` diff:

```
git diff origin/main..mediocregopher/sparse-trie-parent-root-short-circuit --patience -- crates/
trie/sparse/src/parallel.rs
```

crates/trie/parallel/src/proof_task.rs
pepyakin, line 541: Collapse?

pepyakin (approved):

shekhirin (approved):

crates/trie/parallel/src/proof_task.rs
mediocregopher: Could do yeah, I just didn't want to bloat the diff with any more new code than necessary

PR #23668, opened 2026-04-21
fix(engine): let consensus impls control which errors are transient

joshieDo:
Adds `Consensus::is_transient_error()` (default `false`). The engine tree skips the invalid header cache for errors the consensus impl marks as transient, allowing blocks to be re-validated later.

This lets consumers like Tempo override `is_transient_error` to mark `TimestampIsInFuture` as transient without hardcoding that in reth.


mattsse (approved):
lgtm, pending @klkvr

shekhirin (approved):

klkvr (approved):

PR #23670, opened 2026-04-21
fix(engine): configure invalid header cache hit eviction

mattsse:
Adds a TreeConfig/EngineArgs setting for the invalid header cache hit eviction threshold instead of hardcoding it in the engine tree. Setting the threshold to `0` now evicts entries on the first lookup, which effectively disables the cache, and the generated CLI docs document that behavior.

Validated with:
- cargo +nightly fmt --all
- cargo test -p reth-engine-tree invalid_headers --lib
- cargo test -p reth-node-core engine_args --lib
- cargo test -p reth-node-core test_parse_invalid_header_hit_eviction_threshold --lib
- make update-book-cli


emmajam (approved):
lgtm

PR #23672, opened 2026-04-21
test: remove unsafe env::set_var(RUST_LOG) from tests

AJStonewee:
Two tests mutated process-global `RUST_LOG` via `unsafe env::set_var` without restoring it. The discv5 one was debug leftover — assertions do not read logs. The `parse_env_filter_directives` test still covers directive parsing via `--log.file.filter`; the `RUST_LOG` read path is handled by `EnvFilter::from_env_lossy` itself.


mattsse (approved):
we should indeed to set an env var here

PR #23673, opened 2026-04-21
fix(engine): keep state hook when BAL prewarm is skipped

AJStonewee:
`install_state_hook` was decided solely on `bal.is_some()`, so when BAL was provided but prewarm was skipped (small blocks or `disable_prewarming=true`), neither the BAL-driven prewarm path nor the execution-side hook fed the state root task.

The decision is now routed through a new `bal_drives_prewarm` helper that accounts for `SMALL_BLOCK_TX_THRESHOLD`, `disable_transaction_prewarming`, and `disable_bal_parallel_execution`. The hook is installed whenever BAL will not actually drive execution. `spawn_caching_with` reuses the same helper to keep the skip condition in a single place. Three unit tests cover the skip paths.

### DIFF
--- a/crates/engine/tree/src/tree/payload_processor/mod.rs
+++ b/crates/engine/tree/src/tree/payload_processor/mod.rs
@@ -215,6 +215,25 @@ where
     N: NodePrimitives,
     Evm: ConfigureEvm<Primitives = N> + 'static,
 {
+    /// Returns `true` when prewarming should be skipped entirely for a block with the given
+    /// transaction count.
+    ///
+    /// Prewarming is skipped when the user disabled it globally, or when the block is small
+    /// enough that spawning prewarm workers costs more than it saves.
+    const fn should_skip_prewarm(&self, transaction_count: usize) -> bool {
+        self.disable_transaction_prewarming || transaction_count < SMALL_BLOCK_TX_THRESHOLD
+    }
+
+    /// Returns `true` when BAL will actually drive prewarming for this invocation.
+    ///
+    /// The state hook that forwards execution updates to the state root task must be installed
+    /// whenever this returns `false`, so the task still gets updates via the execution path.
+    const fn bal_drives_prewarm(&self, transaction_count: usize, has_bal: bool) -> bool {
+        has_bal &&
+            !self.should_skip_prewarm(transaction_count) &&
+            !self.disable_bal_parallel_execution
+    }
+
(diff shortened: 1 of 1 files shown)


crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -273,7 +292,7 @@ where
             halve_workers,
             config,
         );
-        let install_state_hook = bal.is_none();
+        let install_state_hook = !self.bal_drives_prewarm(env.transaction_count, bal.is_some());
```
mattsse, line 295: I believe this isnt quite correct, because this bool `install_state_hook` determines whether regular execution outcomes should be streamed to the sparse trie, which is a bal only setting and doesnt have anything to do with prewarming technically.

but we should make this clearer, and at least add a one line doc here

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -215,6 +215,25 @@ where
     N: NodePrimitives,
     Evm: ConfigureEvm<Primitives = N> + 'static,
 {
+    /// Returns `true` when prewarming should be skipped entirely for a block with the given
+    /// transaction count.
+    ///
+    /// Prewarming is skipped when the user disabled it globally, or when the block is small
+    /// enough that spawning prewarm workers costs more than it saves.
+    const fn should_skip_prewarm(&self, transaction_count: usize) -> bool {
```
mattsse, line 223: I think this one we can keep

mattsse (changes requested):
wdyt @mediocregopher

mediocregopher:
This LGTM in principle, but I will defer to @pepyakin who is working on the parallel execution implementation right now and I think is changing a bunch of things in this area anyway.

crates/engine/tree/src/tree/payload_processor/mod.rs
mediocregopher: It has to do with prewarming in that prewarming is currently also responsible for converting the BAL into HashedPostState for the sparse trie task, so the two are intertwined.

pepyakin:
Thank you for your PR. It looks good in principle but I would rather close this PR. The reason is, it currently shapes up as BAL path to go via separate route and not prewarming. Prewarming will be skipped and the state hook will not be engaged at all.

PR #23675, opened 2026-04-22
chore: add DecodedBal in ExecutionEnv

Soubhik-10:


crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -495,6 +499,10 @@ where
             transaction_count: input.transaction_count(),
             gas_used: input.gas_used(),
             withdrawals: input.withdrawals().map(|w| w.to_vec()),
+            decoded_bal: input
+                .block_access_list_raw()
+                .cloned()
+                .and_then(|raw| DecodedBal::from_rlp_bytes(raw).ok()),
```
mattsse, line 505: I think we need to do this fallible and bubble up the error

which means we need to cleanup:

https://github.com/paradigmxyz/reth/blob/f39c47bd11b72ed19c6ec1cba868511a504cb5a7/crates/engine/tree/src/tree/payload_validator.rs#L512-L518

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -936,6 +936,9 @@ pub struct ExecutionEnv<Evm: ConfigureEvm> {
     /// Withdrawals included in the block.
     /// Used to generate prefetch targets for withdrawal addresses.
     pub withdrawals: Option<Vec<Withdrawal>>,
+    /// Optional decoded BAL for the block.
+    /// Used to validate and optimize execution.
+    pub decoded_bal: Option<DecodedBal>,
```
mattsse, line 941: to make this cheaper to clone, I think we should Arc this

mattsse (changes requested):

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -491,6 +491,15 @@ where
             .in_scope(|| self.evm_env_for(&input))
             .map_err(NewPayloadError::other)?;
 
+        // Extract the BAL, if valid and available
+        let decoded_bal = ensure_ok!(input
+            .block_access_list_raw()
```
mattsse: lets add a `fn try_decoded_access_list` fn instead because we want to eventually also give the downloaded block an optional DecodedBal

crates/engine/tree/src/tree/payload_processor/mod.rs
```diff
@@ -507,15 +503,18 @@ where
...
             self.executor.spawn_blocking_named("prewarm", move || {
                 let mode = if skip_prewarm {
                     PrewarmMode::Skipped
-                } else if let Some(bal) = bal.filter(|_| !disable_bal_parallel_execution) {
-                    PrewarmMode::BlockAccessList(bal)
+                } else if let Some(decoded_bal) =
+                    maybe_decoded_bal.filter(|_| !disable_bal_parallel_execution)
+                {
+                    PrewarmMode::BlockAccessList(Arc::new(
+                        decoded_bal.as_bal().clone().into_inner(),
```
mattsse: I think we should avoid this expensive clone here and also clone the decoded bal.

changing this to `PrewarmMode::BlockAccessList(Arc<DecodedBal>)`

mattsse (changes requested):

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -2110,6 +2106,28 @@ impl<T: PayloadTypes> BlockOrPayload<T> {
...
+    ///
+    /// This is kept on [`BlockOrPayload`] so downloaded blocks can eventually surface an optional
+    /// decoded BAL through the same interface.
+    pub fn try_decoded_access_list(&self) -> Result<Option<DecodedBal>, alloy_rlp::Error> {
+        match self {
+            Self::Payload(payload) => payload
+                .block_access_list()
+                .map(|block_access_list| DecodedBal::from_rlp_bytes(block_access_list.clone()))
+                .transpose(),
+            Self::Block(_) => Ok(None),
```
mattsse, line 2115: for context, next we can add Option Decoded bal to the `Block` variant, sicne we can also download the over p2p

fyi @0xKarl98

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -2110,6 +2106,28 @@ impl<T: PayloadTypes> BlockOrPayload<T> {
         }
     }
 
+    /// Returns the raw rlp block access list embedded in a payload, if present.
+    pub fn block_access_list_raw(&self) -> Option<&Bytes> {
+        match self {
+            Self::Payload(payload) => payload.block_access_list(),
+            Self::Block(_) => None,
+        }
+    }
```
mattsse: looks like we dont need this fn anymore?

crates/engine/tree/src/tree/payload_processor/prewarm.rs
```diff
@@ -455,7 +455,7 @@ where
                 self.spawn_txs_prewarm(pending, actions_tx, self.to_sparse_trie_task.clone());
             }
             PrewarmMode::BlockAccessList(bal) => {
-                self.run_bal_prewarm(bal, actions_tx);
+                self.run_bal_prewarm(Arc::new(bal.as_bal().into_inner()), actions_tx);
```
mattsse: this arg we should also change

mattsse (changes requested):
more tiny nits

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -2110,6 +2105,20 @@ impl<T: PayloadTypes> BlockOrPayload<T> {
         }
     }
 
+    /// Returns the decoded block access list, if present and successfully decoded.
+    ///
+    /// This is kept on [`BlockOrPayload`] so downloaded blocks can eventually surface an optional
+    /// decoded BAL through the same interface.
+    pub fn try_decoded_access_list(&self) -> Result<Option<DecodedBal>, alloy_rlp::Error> {
```
emmajam, line 2109: can you verify `DecodedBal::from_rlp_bytes` isn't doing redundant work vs the old path, worried it could be a perf regression on the hot path

crates/engine/tree/src/tree/payload_validator.rs
```diff
@@ -2110,6 +2105,20 @@ impl<T: PayloadTypes> BlockOrPayload<T> {
         }
     }
 
+    /// Returns the decoded block access list, if present and successfully decoded.
+    ///
+    /// This is kept on [`BlockOrPayload`] so downloaded blocks can eventually surface an optional
+    /// decoded BAL through the same interface.
```
emmajam: The Block variant just returns Ok(None) today

mattsse (approved):
lgtm,

one thought @klkvr @pepyakin 

maybe we should wrap this in an arc instead:

https://github.com/alloy-rs/eips/blob/4b803807a73c68e4cbd3ddfa786add30ebfffbdf/crates/eip7928/src/block_access_list.rs#L263-L265

which I believe would be more beneficial for dealing with the evm eventually?

pepyakin:
I am not entirely sure, but seems roughly right ↑

pepyakin (approved):

klkvr (approved):

PR #23682, opened 2026-04-22
feat(net): add BAL requirement to block access list requests

0xKarl98:
This PR adds a BAL request requirement to the network fetch path, so callers can choose between Mandatory and Optional handling.
  Mandatory preserves the existing behavior, while Optional lets the local reth node stop waiting when no connected peer can serve
  BALs.

  Roadmap:

  - validate returned BALs against the block header hash and return them as Sealed<Bytes>
  - extend the same requirement and validation flow to full-block range requests
  
  @mattsse

### DIFF
--- a/crates/net/network/src/fetch/mod.rs
+++ b/crates/net/network/src/fetch/mod.rs
@@ -160,15 +160,10 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
     /// full history available
     fn next_best_peer(&self, requirement: BestPeerRequirements) -> Option<PeerId> {
         // filter out peers that aren't idle or don't meet the requirement
-        let mut idle = self.peers.iter().filter(|(_, peer)| {
-            peer.state.is_idle() &&
-                match &requirement {
-                    BestPeerRequirements::EthVersion(ver) => {
-                        peer.capabilities.supports_eth_at_least(ver)
-                    }
-                    _ => true,
-                }
-        });
+        let mut idle = self
+            .peers
+            .iter()
+            .filter(|(_, peer)| peer.state.is_idle() && peer.satisfies(&requirement));
 
         let mut best_peer = idle.next()?;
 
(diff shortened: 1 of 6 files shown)


crates/net/network/src/fetch/mod.rs
```diff
@@ -195,28 +196,39 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
...
+        })
+    }
+
     /// Returns the next action to return
     fn poll_action(&mut self) -> PollAction {
-        // we only check and not pop here since we don't know yet whether a peer is available.
-        if self.queued_requests.is_empty() {
-            return PollAction::NoRequests
-        }
+        loop {
```
mattsse: I dont think we need to change anything here?

this is hard to review, and I think we can achieve this feature with fewer changes

crates/net/network/src/fetch/mod.rs
```diff
@@ -195,28 +196,39 @@ impl<N: NetworkPrimitives> StateFetcher<N> {
         Some(*best_peer.0)
     }
 
+    /// Returns whether any connected peer can serve BAL requests.
+    fn has_bal_capable_peer(&self) -> bool {
+        self.peers.values().any(|peer| {
+            !matches!(peer.state, PeerState::Closing) &&
+                peer.capabilities.supports_eth_at_least(&EthVersion::Eth71)
+        })
+    }
```
mattsse, line 200: we can rename this to `has_eth71_peer`

mattsse (changes requested):
I think we should instead apply the check only once here:

https://github.com/paradigmxyz/reth/blob/d55458479d003e33affbfa7cb487357e7952dcea/crates/net/network/src/fetch/mod.rs#L234-L235

which should be more efficient than always doing it when we select a peer

there's only a potential edge case when we drop a eth71 peer later but imo we can live with that

0xKarl98:
@mattsse have modified according to your feedback

mattsse (approved):
Looks good. The remaining follow-up request compatibility concern is pre-existing and can be handled separately.

PR #23685, opened 2026-04-22
feat(db): detect and warn about ZFS

emmajam:


crates/storage/db/src/mdbx.rs
```diff
@@ -12,12 +12,77 @@ pub use reth_libmdbx::*;
...
+        if libc::statfs(c_path.as_ptr(), &mut stat) == 0 {
+            let fstype = std::ffi::CStr::from_ptr(stat.f_fstypename.as_ptr());
+            Ok(fstype.to_bytes() == b"zfs")
+        } else {
+            Err(std::io::Error::last_os_error())
+        }
+    }
+}
+
+/// Returns `true` if the given path is on a ZFS filesystem.
```
shekhirin:
```suggestion
/// Returns `false` for operating systems other than Linux and macOS.
```

crates/storage/db/src/mdbx.rs
```diff
@@ -12,12 +12,77 @@ pub use reth_libmdbx::*;
 /// versions. These will be dropped during database initialization.
 const ORPHAN_TABLES: &[&str] = &["AccountsTrieChangeSets", "StoragesTrieChangeSets"];
 
+/// Checks if the given path resides on a ZFS filesystem and logs a warning.
+///
+/// ZFS uses copy-on-write (COW) semantics which conflict with MDBX's write patterns, leading to
+/// significant performance degradation and database bloat.
```
shekhirin:
```suggestion
/// significant performance degradation.
```

ZFS CoW doesn't just copy the whole mdbx.dat AFAIK, so no bloat here, just perf degradation

shekhirin (approved):

joshieDo (approved):

crates/storage/db/Cargo.toml
```diff
@@ -42,6 +42,9 @@ rustc-hash = { workspace = true, optional = true, features = ["std"] }
 sysinfo = { workspace = true, features = ["system"] }
 parking_lot = { workspace = true, optional = true }
 
+[target.'cfg(unix)'.dependencies]
```
makcandrov, line 45: @emmajam This puts `strum` under `[target.'cfg(unix)'.dependencies]` as well. That section should be placed at the end of the dependencies.

PR #23686, opened 2026-04-22
feat(p2p): Discv5 is enabled by default

emmajam:
Will keep both enabled for now. Not all EL clients have discv5 yet, so discv4 is still needed for compatibility with those peers. Running both in parallel is the standard transition approach — Geth does the same.

### DIFF
--- a/crates/node/core/src/args/network.rs
+++ b/crates/node/core/src/args/network.rs
@@ -870,10 +870,12 @@ impl DiscoveryArgs {
         true
     }
 
-    /// Set the discovery port to zero, to allow the OS to assign a random unused port when
-    /// discovery binds to the socket.
+    /// Set the discovery ports to zero, to allow the OS to assign random unused ports when
+    /// discovery binds to the sockets.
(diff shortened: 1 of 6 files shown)


klkvr (approved):

Rjected (approved):

crates/node/core/src/args/network.rs
```diff
@@ -718,9 +718,9 @@ pub struct DiscoveryArgs {
     #[arg(long, conflicts_with = "disable_discovery")]
     pub disable_discv4_discovery: bool,
 
-    /// Enable Discv5 discovery.
+    /// Disable Discv5 discovery.
     #[arg(long, conflicts_with = "disable_discovery")]
-    pub enable_discv5_discovery: bool,
```
mattsse, line 723: we need to keep this, otherwise this can break all setups that currently opt-in

mattsse (changes requested):

mattsse (approved):

PR #23688, opened 2026-04-22
fix(net): track unknown tx types in announcement metrics

wataxiwajojo:
`AnnouncedTxTypesMetrics` and `TxTypesCounter` now include an `other` field. When `TxType::try_from` fails during ETH68 announcement processing, the transaction is counted as `other` instead of being silently ignored. Mirrors the fix already applied in txpool (`total_other_transactions`).


mattsse (approved):
Reviewed the patch in `crates/net/network/src/metrics.rs` and `crates/net/network/src/transactions/mod.rs`. Unknown ETH68 tx type bytes are now counted in the new `other` bucket without changing accept/ignore/reject behavior, and the existing relaxed-filter test still passes.

PR #23692, opened 2026-04-23
fix(cli): use TxTy and ReceiptTy for static-file db get

Veronic85:
## Summary

Use `TxTy<N>` and `ReceiptTy<N>` when decoding `Transactions` and `Receipts` from `reth db get static-file`, matching the generic masks already used for those segments.

This keeps the static-file CLI path aligned with custom `NodePrimitives` implementations instead of decoding through the default table value types.


mattsse (approved):
ah, another one -.-

PR #23695, opened 2026-04-23
fix(era-downloader): infer ERA1 from indexed file names

s0mnathh:
Make `EraClient` derive the effective file type from indexed file names and downloaded file names instead of letting `url.contains("era1")` decide whether the downloader should follow the `.era` or `.era1` path.

This keeps index extraction, download URLs, and checksum verification correct for ERA1 mirrors whose base URLs do not include `era1`, while preserving `with_era_type` as an explicit override.


mattsse (commented):
I wanna keep this as is

PR #23698, opened 2026-04-23
fix(provider): migrate-v2 crash on pruned nodes with no prior changeset static files

### LINKED ISSUE #23691: reth 2.1 crash on pruned full node migrate-v2
### Describe the bug

When attempting to use reth 2.1's migrate-v2 command on a pruned full node, it looks for static files that do not exist and then crashes.

### Steps to reproduce

I only run reth with rocketpool but either of these commands will trigger the crash:

```
$ reth db --datadir /ethclient/reth migrate-v2
# or
$ docker compose -p rocketpool   -f ~/.rocketpool/runtime/eth1.yml   run --rm --no-deps   --entrypoint reth eth1   db --datadir /ethclient/reth   migrate-v2
```

### Node logs

```text
$ docker compose -p rocketpool   -f ~/.rocketpool/runtime/eth1.yml   run --rm --no-deps   --entrypoint reth eth1   db --datadir /ethclient/reth   migrate-v2
Container rocketpool-eth1-run-be3fa59bd02b Creating
Container rocketpool-eth1-run-be3fa59bd02b Created
2026-04-21T20:16:31.136401Z  INFO Initialized tracing, debug log directory: /root/.cache/reth/logs/mainnet
2026-04-21T20:16:31.139591Z  INFO Opening storage db_path="/ethclient/reth/db" sf_path="/ethclient/reth/static_files"
2026-04-21T20:16:31.197170Z  INFO check_consistency{read_only=false}: Verifying storage consistency.
2026-04-21T20:16:31.242578Z  WARN Storage settings mismatch detected. Using the stored settings from the existing database. stored=StorageSettings { storage_v2: false } requested=StorageSettings { storage_v2: true }
2026-04-21T20:16:31.242605Z  INFO Starting v1 → v2 storage migration
2026-04-21T20:16:31.242617Z  INFO Chain tip block number tip=24930625
2026-04-21T20:16:31.242622Z  INFO Migrating AccountChangeSets → static files
2026-04-21T20:16:31.264980Z  WARN Failed to load static file jar path="/ethclient/reth/static_files/static_file_account-change-sets_24000000_24499999" e=No such file or directory (os error 2)
2026-04-21T20:16:31.277432Z ERROR shutting down due to error err=failed to open file "/ethclient/reth/static_files/static_file_account-change-sets_24000000_24499999.conf": No such file or directory (os error 2)
Error: failed to open file "/ethclient/reth/static_files/static_file_account-change-sets_24000000_24499999.conf": No such file or directory (os error 2)

Caused by:
    No such file or directory (os error 2)

Location:
    /app/crates/cli/commands/src/db/migrate_v2.rs:136:13
```

### Platform(s)

Linux (x86)

### Container Type

Docker

2.1

Current database version: 2
Local database version: 2

mainnet

Pruned with custom reth.toml config

```
[prune]
block_interval = 5

[prune.segments]
sender_recovery = "full"

[prune.segments.receipts]
before = 0

[prune.segments.account_history]
distance = 10064

[prune.segments.storage_history]
distance = 10064
```

_No response_

### Code of Conduct

PatrickTulskie:
Closes #23691

When migrate-v2 runs on a pruned full node, `AccountChangeSets` has no prior static files. The writer starts at a non-zero block (the block after the prune checkpoint), which lands at the start of a new static-file range. `update_index` was computing segment_max_block = expected_block_start - 1 to point at the "previous file" — but that file never existed on a v1 node, so `NippyJar::load` hit ENOENT and the migration crashed.

The fix is to just guard the heuristic behind a check that the segment already has static files in the index. When it doesn't, `segment_max_block` is None and no load is attempted.

I will add that while this should fix the crash, tests all pass and things look good, I'm not familiar enough with the v2 database format to know if this causes other problems or side effects. Figured it was worth a shot though.

### DIFF
--- a/crates/storage/provider/src/providers/static_file/writer.rs
+++ b/crates/storage/provider/src/providers/static_file/writer.rs
@@ -704,6 +704,14 @@ impl<N: NodePrimitives> StaticFileProviderRW<N> {
         //
         // If that expected block start is 0, then it means that there's no actual block data, and
         // there's no block data in static files.
+        //
+        // We also skip this heuristic when there are no existing static files for this segment.
+        // This handles migration (e.g. v1→v2) where we open a writer starting at a non-zero block
+        // but there is no previous file on disk. Without this guard, `StaticFileProviderInner::
+        // update_index` would try to load the (non-existent) previous range's `.conf` file and
+        // return an ENOENT error.
+        let segment = self.writer.user_header().segment();
+        let reader = self.reader();
         let segment_max_block = self
             .writer
             .user_header()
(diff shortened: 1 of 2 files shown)


shekhirin:
Superseded by https://github.com/paradigmxyz/reth/pull/23716, found a couple more issues and will be merging it soon, thanks!

PR #23699, opened 2026-04-23
feat(tracing): add `std` feature gate to `reth-tracing` for wasm compatibility

mablr:
tracing-appender 0.2.5 introduced a dependency on the `symlink` crate which uses OS-level symlink APIs unavailable on wasm32-wasip1, breaking the wasm CI job for reth-tracing.

Introduces a `std` feature (enabled by default) that gates all std-requiring deps. The `tracing` re-export remains unconditional as it supports `no_std`. When built with `--no-default-features` for `wasm32-wasip1`, only the tracing crate is pulled in.


mablr (commented):
The following depend on C libraries (secp256k1-sys and zstd-sys) and appear to be incompatible with WASM.
```
reth-cli-util       # secp256k1-sys via enr
reth-db             # zstd-sys via reth-nippy-jar
reth-db-api         # zstd-sys via reth-codecs -> reth-zstd-compressors
reth-ecies          # secp256k1-sys via enr
reth-network-api    # secp256k1-sys via enr
reth-nippy-jar      # zstd-sys (direct dependency)
reth-node-types     # zstd-sys via reth-codecs -> reth-zstd-compressors
reth-rpc-server-types # secp256k1-sys via reth-network-api -> enr
reth-trie-db        # zstd-sys via reth-codecs -> reth-zstd-compressors
```

crates/tracing/src/lib.rs
```diff
@@ -210,6 +231,7 @@ impl Default for LayerInfo {
     }
 }
 
+#[cfg(feature = "std")]
```
DaniPopes: after doc comment everywhere please

DaniPopes (dismissed):

crates/tracing/src/lib.rs
```diff
@@ -39,41 +39,57 @@
...
 pub use tracing;
+#[cfg(feature = "std")]
 pub use tracing_appender;
+#[cfg(feature = "std")]
 pub use tracing_subscriber;
 
 #[cfg(feature = "tracy")]
 tracy_client::register_demangler!();
 
-// Re-export our types
```
mattsse, line 53: this I think we can keep

crates/trie/sparse/Cargo.toml
mattsse, line 1: if this is the only place where we forward, then this means everything else also goes through this crate here, since the binary crate compiles, this means that the features are properly activated, but wouldnt hurt to be explicit in the reth-cli/ or reth crate

mablr: set feat explicitly

crates/tracing/src/lib.rs
emmajam, line 323:
```suggestion
#[cfg(feature = "std")]
///  Initializes a tracing subscriber for tests.
///
///  The filter is configurable via `RUST_LOG`.
///
///  # Note
///
///  The subscriber will silently fail if it could not be installed.
pub fn init_test_tracing() {
let _ = TestTracer::default().init();
}
```

Cargo.toml
```diff
@@ -422,7 +422,7 @@ reth-storage-errors = { path = "crates/storage/errors", default-features = false
 reth-tasks = { path = "crates/tasks" }
 reth-testing-utils = { path = "testing/testing-utils" }
 reth-tokio-util = { path = "crates/tokio-util" }
-reth-tracing = { path = "crates/tracing", default-features = false }
```
mattsse, line 426: why was this removed?

doesnt this effectivel always enable it everywhere?

crates/tracing/src/lib.rs
```diff
@@ -39,45 +39,62 @@
...
 
 // Re-export tracing crates
 pub use tracing;
+#[cfg(feature = "std")]
 pub use tracing_appender;
+#[cfg(feature = "std")]
 pub use tracing_subscriber;
 
 #[cfg(feature = "tracy")]
 tracy_client::register_demangler!();
```
mattsse: I believe this also needs a std gate

```
cargo check --no-default-features --features tracy
```

is failing

mattsse (changes requested):

Cargo.toml
mablr: ~that's right, restored~

crates/tracing/src/lib.rs
mablr: I added `std` gate as well, it seems to work now.

Cargo.toml
mablr: Reverted as it was actually a footgun (inverting the default std), i do think it's fine.

mattsse (approved):
Approved per maintainer request.

mattsse (approved):

PR #23702, opened 2026-04-23
fix(net): retain active session buffer capacity

mattsse:
Avoid shrinking active session buffers to zero on every poll.

`ActiveSession::shrink_to_fit()` runs unconditionally at the end of the poll loop, so the next remote request or outgoing response burst immediately has to allocate again. This keeps one incoming request slot warm and retains the existing `MAX_QUEUED_OUTGOING_RESPONSES` window for queued outgoing messages instead.

This preserves the current backpressure behavior while reducing needless allocation churn in steady state.

Validation:
- `cargo +nightly fmt --all`


Rjected (approved):
makes sense

PR #23703, opened 2026-04-23
fix(cli): preserve `trusted_nodes_only` from config when `--trusted-only` is not set

wataxiwajojo:
`--trusted-only` is a clap flag that defaults to `false` when omitted. The three call sites wrote `config.peers.trusted_nodes_only = self.network.trusted_only` unconditionally, silently clobbering `connect_trusted_nodes_only = true` from the TOML config. Switched to `|=` so the CLI can still force it on while TOML values survive when the flag is absent.

Before:
```
$ cat reth.toml | grep trusted
connect_trusted_nodes_only = true
$ reth node --config reth.toml
# trusted_nodes_only silently becomes false
```

After:
```
$ cat reth.toml | grep trusted
connect_trusted_nodes_only = true
$ reth node --config reth.toml
# trusted_nodes_only stays true; --trusted-only on CLI still forces it on
```

### DIFF
--- a/crates/cli/commands/src/p2p/mod.rs
+++ b/crates/cli/commands/src/p2p/mod.rs
@@ -188,7 +188,7 @@ impl<C: ChainSpecParser> DownloadArgs<C> {
             )
         }
 
-        config.peers.trusted_nodes_only = self.network.trusted_only;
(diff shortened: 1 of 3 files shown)


mattsse (approved):
LGTM

PR #23710, opened 2026-04-24
feat(net): add Basic in-memory BAL store

0xKarl98:
Adds a basic in-memory BAL store and wires it into P2P `GetBlockAccessLists` serving.

The network request handler now reads BALs from the provider's `BalStore` and preserves EIP-8159 fallback semantics by returning an RLP empty BAL (`0xc0`) for missing entries. 

This keeps the first implementation intentionally simple: hash-based in-memory lookup only, with no Engine API changes, persistence, range lookup, or retention policy.
  
cc @mattsse

### DIFF
--- a/crates/net/network/src/eth_requests.rs
+++ b/crates/net/network/src/eth_requests.rs
@@ -18,7 +18,7 @@ use reth_network_api::test_utils::PeersHandle;
 use reth_network_p2p::error::RequestResult;
 use reth_network_peers::PeerId;
 use reth_primitives_traits::Block;
-use reth_storage_api::{BalProvider, BlockReader, HeaderProvider};
+use reth_storage_api::{BalStoreHandle, BlockReader, HeaderProvider};
 use std::{
     future::Future,
     pin::Pin,
@@ -63,6 +63,8 @@ pub struct EthRequestHandler<C, N: NetworkPrimitives = EthNetworkPrimitives> {
     peers: PeersHandle,
     /// Incoming request from the [`NetworkManager`](crate::NetworkManager).
     incoming_requests: ReceiverStream<IncomingEthRequest<N>>,
+    /// Store for block access lists.
+    bal_store: BalStoreHandle,
     /// Metrics for the eth request handler.
     metrics: EthRequestHandlerMetrics,
 }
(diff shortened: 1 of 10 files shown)


crates/net/network/src/eth_requests.rs
```diff
@@ -63,6 +63,8 @@ pub struct EthRequestHandler<C, N: NetworkPrimitives = EthNetworkPrimitives> {
     peers: PeersHandle,
     /// Incoming request from the [`NetworkManager`](crate::NetworkManager).
     incoming_requests: ReceiverStream<IncomingEthRequest<N>>,
+    /// Store for block access lists.
+    bal_store: BalStoreHandle,
```
mattsse: I think, intead of storing this as separate field, lets add the balstore trait bound to client: C instead

crates/node/builder/src/builder/mod.rs
```diff
@@ -915,7 +915,10 @@ impl<Node: FullNodeTypes> BuilderContext<Node> {
     {
         let (handle, network, txpool, eth) = builder
             .transactions_with_policies(pool, tx_config, propagation_policy, announcement_policy)
-            .request_handler(self.provider().clone())
+            .request_handler_with_bal_store(
+                self.provider().clone(),
+                self.provider().bal_store().clone(),
```
mattsse: because these are effectively coming from he same type here: Blockchainprovider

mattsse (changes requested):
possible we need to sprinkle a few more trait bounds over the place to get `C: BalstoreProvider` as well

crates/net/network/src/eth_requests.rs
```diff
@@ -332,13 +311,41 @@ where
...
+        request: GetBlockAccessLists,
+        response: oneshot::Sender<RequestResult<BlockAccessLists>>,
+    ) {
+        let access_lists = self
+            .client
+            .bal_store()
+            .get_by_hashes(&request.0)
+            .unwrap_or_else(|_| request.0.iter().map(|_| None).collect())
+            .into_iter()
+            .map(|bal| bal.unwrap_or_else(|| Bytes::from_static(&[alloy_rlp::EMPTY_LIST_CODE])))
```
mattsse, line 335: what we're missing here is a way to enforce the softlimit (2MB) per request.

what we can do in a followup is adding something similar to these 

https://github.com/paradigmxyz/reth/blob/b27169430154fbdf0f1def8c4156515a70666e03/crates/transaction-pool/src/traits.rs#L355-L374

to the Balstore trait, so that we can call this here with 2MB and then the store only returns the first 2MB

mattsse (approved):

mattsse:
@0xKarl98 I also think we need to add trait bounds to 

https://github.com/paradigmxyz/reth/blob/b27169430154fbdf0f1def8c4156515a70666e03/crates/storage/provider/src/traits/full.rs#L17-L17

perhaps more even

0xKarl98:
@mattsse but currently  trait FullProvider<N: NodeTypesWithDB>:  already has bounded to balProvider here :

  pub trait FullProvider<N: NodeTypesWithDB>:
      ...
      + AccountReader
      + BalProvider
      + StateProviderFactory
      ...
      
  Think they are fine ?

PR #23716, opened 2026-04-24
fix(storage): `reth db migrate-v2` for pruned nodes

shekhirin:
1. For a static file that has a range `0..=10` we can't write block `2` without writing blocks `0..=1 first`. The fix is to write empty data for these blocks, we do it already in Execution stage for receipts / changesets.
2. When we open a static files writer on block `500_001`, it creates a new empty jar for range `500_000..=999_999`, and tries to update the index with the max existing block being `499_999` (thinking that it exists), but it doesn't because it's a pruned node that shouldn't have this static file by design, so it crashes. The fix here is to not crash, skipping the index update.
3. When migrating storage changesets, we initialized the walker as `cursor.walk(Some(Default::default()))?.peekable()`, which is not correct if the first block is non-zero.

Drive-by: add `count` field to receipts migration log

### DIFF
--- a/crates/cli/commands/src/db/migrate_v2.rs
+++ b/crates/cli/commands/src/db/migrate_v2.rs
@@ -255,6 +255,10 @@ impl Command {
             writer.commit()?;
         }
 
+        let before = sf_provider
+            .get_highest_static_file_tx(StaticFileSegment::Receipts)
+            .map_or(0, |tx| tx + 1);
+
         let block_range = first_block..=tip;
 
         let segment = reth_static_file::segments::Receipts;
@@ -262,7 +266,11 @@ impl Command {
 
         sf_provider.commit()?;
 
-        info!(target: "reth::cli", "Receipts migrated");
+        let after = sf_provider
+            .get_highest_static_file_tx(StaticFileSegment::Receipts)
+            .map_or(0, |tx| tx + 1);
+        let count = after - before;
+        info!(target: "reth::cli", count, "Receipts migrated");
         Ok(())
     }
 
--- a/crates/cli/commands/src/db/migrate_v2.rs
+++ b/crates/cli/commands/src/db/migrate_v2.rs
@@ -135,6 +135,14 @@ impl Command {
         let mut writer =
             sf_provider.get_writer(first_block, StaticFileSegment::AccountChangeSets)?;
 
+        // The writer starts at the fixed range boundary (e.g. 2500000) which may be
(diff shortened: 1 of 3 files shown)


crates/cli/commands/src/db/migrate_v2.rs
```diff
@@ -177,6 +185,14 @@ impl Command {
         let mut writer =
             sf_provider.get_writer(first_block, StaticFileSegment::StorageChangeSets)?;
 
+        // The writer starts at the fixed range boundary (e.g. 2500000) which may be
+        // earlier than first_block (e.g. 2603897 from prune checkpoint). Pad with
+        // empty changesets for pruned blocks so the writer is aligned.
+        let writer_start = writer.next_block_number();
+        for block in writer_start..first_block {
+            writer.append_storage_changeset(vec![], block)?;
+        }
```
joshieDo: i think we can call `ensure_at_block` ?

crates/cli/commands/src/db/migrate_v2.rs
```diff
@@ -238,6 +254,18 @@ impl Command {
...
 
+        // The writer starts at the fixed range boundary (e.g. 2500000) which may be
+        // earlier than first_block (e.g. 2603897 from prune checkpoint). Pad with
+        // empty blocks for pruned receipts so the writer is aligned.
+        {
+            let mut writer = sf_provider.get_writer(first_block, StaticFileSegment::Receipts)?;
+            let writer_start = writer.next_block_number();
+            for block in writer_start..first_block {
+                writer.increment_block(block)?;
+            }
```
joshieDo: i think we can call `ensure_at_block` ?

crates/cli/commands/src/db/migrate_v2.rs
```diff
@@ -135,6 +135,14 @@ impl Command {
         let mut writer =
             sf_provider.get_writer(first_block, StaticFileSegment::AccountChangeSets)?;
 
+        // The writer starts at the fixed range boundary (e.g. 2500000) which may be
+        // earlier than first_block (e.g. 2603897 from prune checkpoint). Pad with
+        // empty changesets for pruned blocks so the writer is aligned.
+        let writer_start = writer.next_block_number();
+        for block in writer_start..first_block {
+            writer.append_account_changeset(vec![], block)?;
+        }
```
joshieDo: i think we can call ensure_at_block ?

joshieDo (approved):

PR #23718, opened 2026-04-24
feat: bound memory footprint of p2p messages

klkvr:
Even though we do enforce 10Mb limit on p2p message size, due to RLP encoding being pretty compact, memory footprint of messages might be up to 20x bigger, meaning that a 10MB `Transactions` payload would in fact occupy 200MB of memory, making worst case of a DOS from multiple peers quite bad

This PR adds an extra measure applied when decoding p2p messages which is that while decoding a `Vec` from RLP we track memory footprint of elements decoded so far, and skip decoding the rest once the threshold is exceeded.

For now this is only applied to `Transactions` and `PooledTransactions` messages as RLP-encoded transactions tend to have highest potential amplifiers of RLP vs mem footprint. The current limit of memory usage is set to 2x the message size

### DIFF
--- a/crates/net/eth-wire-types/src/broadcast.rs
+++ b/crates/net/eth-wire-types/src/broadcast.rs
@@ -7,13 +7,14 @@ use alloy_primitives::{
     Bytes, TxHash, B256, U128,
 };
 use alloy_rlp::{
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

PR #23719, opened 2026-04-24
fix(txpool): log failed transaction re-injection on reorg

wataxiwajojo:
Reorg path discarded the `Vec<Result<…>>` from `add_external_transactions`, so re-injection failures were silent while `metrics.inc_reinserted_transactions` had already counted them pre-call.

Before:
```rust
let _ = pool.add_external_transactions(pruned_old_transactions).await;
```

After: count errors and emit `warn!` with failed/total, plus `debug!` per error.
```
WARN txpool: failed=3 total=12 failed to re-inject transactions after reorg
```


mattsse (commented):
we can keep this as is

PR #23720, opened 2026-04-24
fix(rpc): include block numbers in `BlockRangeExceedsHead` error

wataxiwajojo:
`FilterBlockRangeError::BlockRangeExceedsHead` and `EthFilterError::BlockRangeExceedsHead` now carry `requested` and `head`, matching the existing `EthApiError::RequestBeyondHead` pattern. Callers can't know the current head, so surfacing it makes the error actionable.

Before:
```
block range extends beyond current head block
```

After:
```
block range extends beyond current head block: requested 21000100, head 21000000
```


mattsse (approved):
Looks good. This keeps the existing invalid-params behavior and Geth-compatible base message while adding requested/head context for callers.

